PR #848, opened 2019-07-24
[ Sublist.Propositional ] category laws, plus functor laws for All and Any

andreasabel:
I have been needing the category laws for sublists whenever I use them as renamings in formalizations of STLC.

I only added them for the propositional case.  There are generalizations, like for the Setoid-sublist they hold if the Setoid is actually a Category, e.g. having `trans refl = id`.  
In some future, when we have a proper concept of Category in the standard library, we might want such a generalization.

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
@@ -8,7 +8,7 @@
 
 module Data.List.Relation.Binary.Sublist.Propositional.Properties {a} {A : Set a} where
 
-open import Data.List using (map)
+open import Data.List using (List; []; _∷_;  map)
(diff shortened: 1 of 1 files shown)


MatthewDaggitt:
Hi @andreasabel, thank you very much for the pull request. However would it be possible to leave them up for a little longer in future? The general rule of thumb I've stuck by is about a week, that gives everyone who's involved in the library a chance to feedback about the changes. For instance had I got a chance to look at it, I'd have noted:
-  you didn't include the changes in the Changelog.
- it's standard library convention (as noted in the style guide) not to use variables for anything except levels and sets at the moment.
- all variables should be declared at the top of the file
- the terms `lookup-⊆-refl` and `lookup-⊆-trans` should have types attached to them.
- pretty sure one of the cases for `lookup-injective` can be omitted

PR #841, opened 2019-07-15
[ Sublist/Setoid ] use open instead of alias for ⊆-refl etc.

andreasabel:
Gives better jump-to-definition behavior in Emacs.

By using `open M arg renaming (foo to bar)` instead `bar = M.foo arg` I am hoping for better UX.

Issue #556, opened 2018-12-04
Adding an (existentially quantified) sublist


ajrouvoet:
We have `Sublist.Inductive` to relationally define what `xs` are sublists/OPEs of a given `ys`.
It seems useful to me to actually define a notion `Sublist ys` to existentially quantify over the `xs`, with definitions:

```
  ⊤ : ∀ {l} → Sublist l
  ⊥ : ∀ {l} → Sublist l
  Empty : ∀ {l} → Sublist l → Set a
  _∪_ : ∀ {l} → (xs ys : Sublist l) → Sublist l
  _∩_ : ∀ {l} → (xs ys : Sublist l) → Sublist l
```

This is completely analogue to the notion `Fin.Subset n` for natural numbers.

I have these definitions sitting in a branch [here](https://github.com/ajrouvoet/agda-stdlib/commit/b71d00dae5404a8899967a0bcc30dc249ae8f8c9). 
I would like some input on:

- [ ] Where should this be defined?
- [ ] Equality of a sublist to ⊤ appears hard to prove in practice. Proving that the reification of a sublist of xs is equal to xs seems more practical. This begs for the definition of `Every` as a counterpart to `Empty`.

gallais:
```agda
_∪_ : ∀ {l} → (xs ys : Sublist l) → Sublist l
_∩_ : ∀ {l} → (xs ys : Sublist l) → Sublist l
```

Ooooh fancy. I don't think we have meet & joins for the current inductive sublists?

I don't know where this would go.
As for the last point: can't we had a proof that if `reify` equals `xs` then the sublist equals `⊤` whenever the setoid equalities are irrelevant?

ajrouvoet:
We do not have those now indeed, probably exactly because the left operand of the output is then existentially quantified. 

The latter proof I have indeed. So maybe naming the intermediate representation is unnecessary?

After your note on the other issue about closure I realized that this is another example of the closure of a relation, but using the 'somewhere' modality, rather than the everywhere. Might be useful to define that modality and define Sublist as such.

MatthewDaggitt:
As for location I've been thinking about adding a `Predicate` subfolder to match the `Relation` folder for compositional data. This should really live in `Data.List.Predicate.Sublist` right? The only problem is that `Any` and `All` are so widely used that it's going to be backwards compatible nightmare.... I guess we could deprecate the module.

gallais:
[...]

PR #849, opened 2019-07-25
[ Sublist.Setoid ] RawPushout for merging two extensions of a list

andreasabel:
OPE does not have proper pushouts.
For weak pushouts, we would need a groupoid rather than a setoid.
However, we can still have raw pushout diagrams without any proven properties.

I plan to extend this PR to weak pushouts in case of Sublist.Propositional.

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Setoid.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Setoid.agda
@@ -92,7 +92,7 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ �
   }
 
 ------------------------------------------------------------------------
--- Weak pushout
+-- Raw pushout
 --
 -- The category _⊆_ does not have proper pushouts.  For instance consider:
 --
@@ -128,12 +128,11 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ �
 
 private
   variable
-    x y z : A
-    xs ys zs us vs : List A
-    τ τ′ τ₁ τ₂ σ σ′ : xs ⊆ ys
+    x y z    : A
+    xs ys zs : List A
+    τ σ      : xs ⊆ ys
 
 record RawPushout (τ : xs ⊆ ys) (σ : xs ⊆ zs) : Set (c ⊔ ℓ) where
-  constructor rawPushout
   field
     {upperBound} : List A
     leg₁         : ys ⊆ upperBound
--- a/src/Data/List/Relation/Binary/Sublist/Setoid.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Setoid.agda
@@ -146,17 +146,26 @@ open RawPushout
 -- Extending the right upper corner.
 
 _∷ʳ₁_ : ∀ y → RawPushout τ σ → RawPushout (y ∷ʳ τ) σ
-y ∷ʳ₁ rpo = record { leg₁ = refl ∷ leg₁ rpo ; leg₂ = y ∷ʳ leg₂ rpo }
+y ∷ʳ₁ rpo = record
+  { leg₁ = refl ∷ leg₁ rpo
+  ; leg₂ = y   ∷ʳ leg₂ rpo
+  }
 
 -- Extending the left lower corner.
 
 _∷ʳ₂_ : ∀ z → RawPushout τ σ → RawPushout τ (z ∷ʳ σ)
-z ∷ʳ₂ rpo = record { leg₁ = z ∷ʳ leg₁ rpo ; leg₂ = refl ∷ leg₂ rpo }
+z ∷ʳ₂ rpo = record
+  { leg₁ = z   ∷ʳ leg₁ rpo
+  ; leg₂ = refl ∷ leg₂ rpo
+  }
 
 -- Extending both of these corners with equal elements.
 
 ∷-rpo : (x≈y : x ≈ y) (x≈z : x ≈ z) → RawPushout τ σ → RawPushout (x≈y ∷ τ) (x≈z ∷ σ)
-∷-rpo x≈y x≈z rpo = record { leg₁ = sym x≈y ∷ leg₁ rpo; leg₂ = sym x≈z ∷ leg₂ rpo }
+∷-rpo x≈y x≈z rpo = record
+  { leg₁ = sym x≈y ∷ leg₁ rpo
+  ; leg₂ = sym x≈z ∷ leg₂ rpo
+  }
 
 ------------------------------------------------------------------------
 -- Left-biased pushout: add elements of left extension first.
--- a/src/Data/List/Relation/Binary/Sublist/Setoid.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Setoid.agda
@@ -14,6 +14,7 @@ module Data.List.Relation.Binary.Sublist.Setoid
   {c ℓ} (S : Setoid c ℓ) where
 
 open import Level using (_⊔_)
+
 open import Data.List.Base using (List; []; _∷_)
 import Data.List.Relation.Binary.Equality.Setoid as SetoidEquality
 import Data.List.Relation.Binary.Sublist.Heterogeneous as Heterogeneous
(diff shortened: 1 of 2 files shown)


src/Data/List/Relation/Binary/Sublist/Setoid.agda
```diff
@@ -79,3 +79,80 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ 
...
+∷-rpo x≈y x≈z rpo = record { leg₁ = sym x≈y ∷ leg₁ rpo; leg₂ = sym x≈z ∷ leg₂ rpo }
+
+------------------------------------------------------------------------
+-- Left-biased pushout: add elements of left extension first.
+
+⊆-merge : (τ : xs ⊆ ys) (σ : xs ⊆ zs) → RawPushout τ σ
+⊆-merge []        σ         = record { leg₁ = σ ; leg₂ = ⊆-refl }
+⊆-merge (y  ∷ʳ τ) σ         = y ∷ʳ₁ ⊆-merge τ σ
+⊆-merge τ@(_ ∷ _) (z  ∷ʳ σ) = z ∷ʳ₂ ⊆-merge τ σ
+⊆-merge (x≈y ∷ τ) (x≈z ∷ σ) = ∷-rpo x≈y x≈z (⊆-merge τ σ)
```
gallais: Should the name mention the fact it is left-biased (with an `\^l` for instance)?
Should it be more informative than `merge` which could also stand for a pullback?

andreasabel: Yes, I had the `l` superscript but deleted it since the `r` version can be obtained by swapping arguments and results.  However, I am happy to put `l` back.

`merge` alludes to the merging of the two superlists similar to what mergesort does.  Alternative: `join`.  I also thought of `pushout`, but this sounds too categorical for my taste; ideally, I'd like to stay closer to FP and mainstream math vocabulary, where possible.

gallais: Ah, yes, meet and join! Related discussion: #556

src/Data/List/Relation/Binary/Sublist/Setoid.agda
```diff
@@ -79,3 +79,80 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ 
...
+--
+-- for some notion of equality _==_ for equality proofs _≈_.
+-- Such a property is given e.g. if _≈_ is proof irrelevant
+-- or forms a groupoid.
+
+private
+  variable
+    x y z : A
+    xs ys zs us vs : List A
+    τ τ′ τ₁ τ₂ σ σ′ : xs ⊆ ys
```
MatthewDaggitt: Standard library convention is that we're not currently using variables for anything except levels and sets.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
```diff
@@ -79,3 +79,80 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ 
...
+
+open RawPushout
+
+------------------------------------------------------------------------
+-- Extending corners of a raw pushout square
+
+-- Extending the right upper corner.
+
+_∷ʳ₁_ : ∀ y → RawPushout τ σ → RawPushout (y ∷ʳ τ) σ
+y ∷ʳ₁ rpo = record { leg₁ = refl ∷ leg₁ rpo ; leg₂ = y ∷ʳ leg₂ rpo }
```
MatthewDaggitt: Please could layout the records as specified by the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md)

src/Data/List/Relation/Binary/Sublist/Setoid.agda
```diff
@@ -79,3 +79,80 @@ open HeterogeneousProperties.Antisymmetry {R = _≈_} {S = _≈_} (λ x≈y _ 
 ⊆-poset = record
   { isPartialOrder = ⊆-isPartialOrder
   }
+
+------------------------------------------------------------------------
+-- Weak pushout
```
MatthewDaggitt: Do you think this content should go in `Data.List.Relation.Binary.Sublist.Setoid.Categorical`?

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: Ok.
(Maybe except the very last, which is very short.)

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: Just saw this in the style guide.    But it seems to contradict https://github.com/agda/agda-stdlib/pull/841#pullrequestreview-262429559 where the opposite advice was given.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: Currently, the `Category` tree defines functional control structures such as `RawMonad` and relatives. (Imho, this tree would be better named `Control` like in Haskell to free `Category` for proper category theory; but this would be a too invasive change.)
The stuff under `Categorical` seems to confine itself to define Monad instances etc.
Thus, I am not sure whether such a `Categorical` module would be similar enough to the existing ones. Dunno. @gallais, what would you think?

andreasabel:
I am now thinking that `⊆-joinˡ τ σ` should return a derivation of `xs ⊆ upperBound ...` so the original `join` has been renamed to `pushout`.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: If you want to stick to the strict `variable` policy, I can of course give the explicit type signatures, no problem.

MatthewDaggitt: I'm pretty sure @gallais gave the same advice in #841 

> I think this is a good opportunity to add a private variable block at
the top of Data.List.Relation.Binary.Sublist.Heterogeneous.Properties
declaring the **levels & sets** used in all of these modules.

As for a rationale, It's pretty much for the same reason as you mentioned earlier. The more complex interactions between variables is still a little unstable so we don't want to dive into using them too much for the moment.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
MatthewDaggitt: > Imho, this tree would be better named Control like in Haskell to free Category for proper category theory; but this would be a too invasive change.

Hmm I guess we could change it to `Control` if we kept the old modules around for backwards compatibility... I agree `Categorical` was a bit of a weird naming choice on my part. The beauty of hindsight :smile:

gallais: `Categorical` makes sense given the pre-existing `Category.(Functor/Applicative/Monad)` modules.

JacquesCarette: This is why, in part, the current `agda-categories` uses `Categories` as the top-level name space, so that it can be used alongside `stdlib` with no conflicts. If that name should be changed, the sooner we do that, the better.

MatthewDaggitt: So I think using `Categories` as a top-level subsection is fine. Then, just like `Relation`, different parts of the library can have a `Categorical` folder where they use those definitions.

MatthewDaggitt:
Apologies @andreasabel I did some tidying up after your previous commit without realising that this PR was working on the same file. I've merged the changes in, but you'll need to add in the previously generalised `variable`s.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: I removed this variable block now.

src/Data/List/Relation/Binary/Sublist/Setoid.agda
andreasabel: Concerning this PR, I propose to leave the `RawPushout` here until the `Category` tree goes beyond the category of `Set` and `->`.  Then, the definition of `RawPushout` can be moved to the `Category` tree and the `⊆-pushoutˡ` to a `Categorical` submodule.

PR #850, opened 2019-07-27
Standardise numeric predicates

MatthewDaggitt:
I've found it a bit irritating not having a standard notation for these so I thought I'd add them. They're all defined along the lines of the proofs in `Data.Nat.DivMod` so that as long as the constructor is in the right shape, Agda can automatically infer the proof. Comments included explaining this.

I'm not quite sure about the location. `Data.X.Predicate` seems like a compromise but still not ideal. Ideally I'd add them in `Data.X.Base` but of course they rely on decidability proofs and so must depend on `Data.X.Properties`... I thought about the idea of `Data.X.Relation.Unary` but it seems a little overkill. The name is a little misleading as well as there's other predicates not in the file, e.g. `Prime`. If anyone has other suggestions, I'm all ears.


src/Data/Nat/Predicate.agda
```diff
@@ -0,0 +1,40 @@
...
+NonZero x = False (x ≟ 0)
+
+------------------------------------------------------------------------
+-- Constructors
+
+nonZero : ∀ {n} → n ≢ 0 → NonZero n
+nonZero {zero}  0≢0 = 0≢0 refl
+nonZero {suc n} n≢0 = _
+
+nonZero< : ∀ {n} → 0 < n → NonZero n
```
JacquesCarette: Wouldn't it be easier to read/comprehend if this were written as `n > 0` ?

JacquesCarette:
I definitely prefer `Data.X.Predicate` over `Data.X.Relation.Unary` !  Simply `Data.X.Relation` might be more general (you can have different arities than unary without being confusing).

A point for debate (i.e. I am not actively advocating for this, but I think the pros/cons should be discussed) would be to put these in `Data.X.Proposition` too.

src/Data/Nat/Predicate.agda
MatthewDaggitt: Yup, will change.

MatthewDaggitt:
> I definitely prefer Data.X.Predicate over Data.X.Relation.Unary ! Simply Data.X.Relation might be more general (you can have different arities than unary without being confusing).

So these are kind of weird as most relations (in the general sense) I would usually define in `Data.X.Base` where `X` is numeric. Indeed we already have many binary relations in `Data.X.Base`. The only reason these even need their own file is their annoying dependency on decidable equality. I think it's unlikely we'll want to define corresponding versions of binary relations.

> A point for debate (i.e. I am not actively advocating for this, but I think the pros/cons should be discussed) would be to put these in Data.X.Proposition too.

Hmm, I can see where you're coming from but I can't off the top of my head think of other places the library uses the term `Proposition` in this way. Feel free to point out occurrences where I'm wrong :smile:

src/Data/Nat/DivMod.agda
```diff
@@ -15,33 +15,34 @@ open import Data.Fin.Properties using (toℕ-fromℕ≤)
...
+-- Definitions (with automatic non-zeroness)
 
 -- The division and modulus operations are only defined when the divisor
 -- is non-zero. The proof of this is defined as an irrelevant
--- implict argument of type `False (divisor ≟ 0)`. This allows this
--- proof to be automatically inferred when the divisor is of the form
--- `suc n`, and hence minimises the number of these proofs that
--- need be passed around. You can therefore write `m / suc n` without
--- issue.
+-- implict argument of type `NonZero divisor ≡ False (divisor ≟ 0)`.
```
andreasabel: implict --> implicit

`NonZero divisor ≡ False (divisor ≟ 0)` looks like an equality type, probably not what you meant.  Expand to natural language, like 
> `NonZero divisor` which is defined as `False (divisor ≟ 0)` 
?!

riz0id:
started working on updating this pull request today incorporating [`Data.Refinement`](https://github.com/agda/agda-stdlib/blob/master/src/Data/Refinement.agda). work includes fixing [Issue 1061](https://github.com/agda/agda-stdlib/issues/1061), catching the branch up to master, and simplifying any existing proofs where appropriate. figured i'd announce it to avoid conflicts in-case someone else was working on this.

MatthewDaggitt:
Hmm so the reason I abandoned this is that after I raised #921 I began to worry about the compatibility of these definitions with instance arguments... however I really(!) should have written down my concerns here as I can no longer remember what the problem was. Feel free to bring it up to date, and I'll see if I can try and remember.

src/Data/Rational/Unnormalised.agda
```diff
@@ -50,10 +53,14 @@ open ℚᵘ public using ()
...
 -- Equality of rational numbers (does not coincide with _≡_)
 
-infix 4 _≃_
+infix 4 _≃_ _≠_
+
 data _≃_ : Rel ℚᵘ 0ℓ where
   *≡* : ∀ {p q} → (↥ p ℤ.* ↧ q) ≡ (↥ q ℤ.* ↧ p) → p ≃ q
 
+_≠_ : Rel ℚᵘ 0ℓ
+p ≠ q = ¬ (p ≃ q)
```
MatthewDaggitt, line 62: I'm a little unhappy about this symbol. Anyone have any better alternatives?

MatthewDaggitt:
Right I've completely reworked this. I'd appreciate if someone could have another quick look over.

Most importantly I've dropped the dependency on the decidability of equality. This means that the predicates and their constructors can all go in the main `X.Base` file rather than the separate `X.Predicate`. This is a particularly big win as it means that later on we can then use them in division operations that are in the base file.

The downside is that it's not backwards compatible with the existing non-zero types used in division and modulus operations. However we're going to change all the definitions anyway to use irrelevant instance arguments (see #921) so I think switching the existing code over to using the new definitions should also wait for version 2.0.

PR #852, opened 2019-08-02
Add Partial Equivalences, Partial Setoids, and Partial Reasoning

TOTBWF:
This PR adds Partial Equivalences and Partial Setoids, which generalize Equivalence Relations and Setoids by removing the reflexivity requirement. This opens the door to dealing with algebraic substructures, as well as other subset related things.


JacquesCarette:
Looks great to me!

src/Relation/Binary/Partial.agda
```diff
@@ -0,0 +1,58 @@
...
+
+------------------------------------------------------------------------
+-- Simple properties and equivalence relations
+
+open import Relation.Binary.Core public
+
+------------------------------------------------------------------------
+-- Equivalence relations
+
+record IsPartialEquivalence {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ) where
```
MatthewDaggitt: I might be tempted to extract the `{a ℓ} {A : Set a}` out into a private variable declaration at the top of the file.

src/Relation/Binary/Partial.agda
```diff
@@ -0,0 +1,58 @@
...
+  field
+    sym   : Symmetric _≈_
+    trans : Transitive _≈_
+
+------------------------------------------------------------------------
+-- Partial Setoids
+
+record PartialSetoid c ℓ : Set (suc (c ⊔ ℓ)) where
```
MatthewDaggitt: I'd try and keep the level names consistent, i.e. stay with `a` instead of `c`.

src/Relation/Binary/Partial.agda
```diff
@@ -0,0 +1,58 @@
...
+  { sym = sym
+  ; trans = trans
+  }
+  where open IsEquivalence eq
+
+fromSetoid : ∀ {c ℓ} → Setoid c ℓ → PartialSetoid c ℓ
+fromSetoid S = record
+  { Carrier = Carrier
+  ; _≈_ = _≈_
+  ; isPartialEquivalence = fromEquivalence isEquivalence
```
MatthewDaggitt: So you should only need the `isPartialEquivalence` field here.

TOTBWF:
I suppose that those partial versions could be useful to someone, but I am struggling to figure out what the use cases would be.

Perhaps it would be better to add `IsPartialEquivalence` to `Relation.Binary.Core` and `PartialSetoid` to `Relation.Binary`, as that would let us add

MatthewDaggitt:
> Perhaps it would be better to add IsPartialEquivalence to Relation.Binary.Core and PartialSetoid to Relation.Binary, as that would let us add isPartialEquivalence and partialSetoid as definitions in the record modules of IsEquivalence and Setoid, which would make lowering much less painful.

Yup, that would seem to make sense. If there's no plan to extend the hierarchy to the orders then I think it's fine having them in the same modules as the originals.

The only thing left to do then (after the minor changes I mention above) is to add a CHANGELOG entry and then we should be good to go :+1:

TOTBWF:
This should be good to go!

MatthewDaggitt:
Merged, thanks for all the work @TOTBWF

PR #856, opened 2019-08-12
Sublist union

andreasabel:
Compute the disjoint union of two sublists.

Similar to the non-overlapping Cover of Conor McBride's "Everybody's got to be somewhere".

This PR needs squashing upon merge.

TODOs:
- [x] fixities for infix operators
- [x] code and properties for making two sublists disjoint
- [x] example


src/Data/List/Relation/Binary/Sublist/Heterogeneous.agda
```diff
@@ -44,7 +49,8 @@ minimum (x ∷ xs) = x ∷ʳ minimum xs
 ------------------------------------------------------------------------
 -- Conversion to and from Any
 
-toAny : ∀ {a bs} → Sublist R [ a ] bs → Any (R a) bs
+-- Special case: Sublist R [ a ] bs → Any (R a) bs
+toAny : ∀ {a as bs} → Sublist R (a ∷ as) bs → Any (R a) bs
```
MatthewDaggitt: You'll need to document this change in the CHANGELOG

src/Data/List/Relation/Binary/Sublist/Heterogeneous.agda
```diff
@@ -61,3 +67,82 @@ module _ {p q} {P : Pred A p} {Q : Pred B q} (resp : P ⟶ Q Respects R) where
...
+⊆-disjoint? (y ∷ʳ τ) (x≈y ∷ σ) with ⊆-disjoint? τ σ
+... | yes d = yes (x≈y ∷ᵣ d)
+... | no ¬d = no λ{ (_ ∷ᵣ d) → ¬d d }
+⊆-disjoint? (x≈y ∷ τ) (y ∷ʳ σ) with ⊆-disjoint? τ σ
+... | yes d = yes (x≈y ∷ₗ d)
+... | no ¬d = no λ{ (_ ∷ₗ d) → ¬d d }
+-- Present in neither sublist: ok.
+⊆-disjoint? (y ∷ʳ τ) (.y ∷ʳ σ) with ⊆-disjoint? τ σ
+... | yes d = yes (y ∷ₙ d)
+... | no ¬d = no λ{ (_ ∷ₙ d) → ¬d d }
```
MatthewDaggitt: The convention is to put decidability proofs in the `.Properties` file.

src/Data/List/Relation/Binary/Sublist/Heterogeneous.agda
```diff
@@ -61,3 +67,82 @@ module _ {p q} {P : Pred A p} {Q : Pred B q} (resp : P ⟶ Q Respects R) where
...
+
+  -- Element y of zs is in xs as x with x≈y: add to us.
+  _∷ₗ_ : ∀ {xs ys zs us} {τ : xs ⊆ zs} {σ : ys ⊆ zs} {ρ : us ⊆ zs} {x y} →
+         (x≈y : R x y) → DisjointUnion τ σ ρ → DisjointUnion (x≈y ∷ τ) (y  ∷ʳ σ) (x≈y ∷ ρ)
+
+  -- Element y of zs is in ys as x with x≈y: add to us.
+  _∷ᵣ_ : ∀ {xs ys zs us} {τ : xs ⊆ zs} {σ : ys ⊆ zs} {ρ : us ⊆ zs} {x y} →
+         (x≈y : R x y) → DisjointUnion τ σ ρ → DisjointUnion (y  ∷ʳ τ) (x≈y ∷ σ) (x≈y ∷ ρ)
+
+DisjointUnion→Disjoint : ∀ {xs ys zs us} {τ : xs ⊆ zs} {σ : ys ⊆ zs} {ρ : us ⊆ zs} →
```
MatthewDaggitt: Likewise with conversion proofs.

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -444,3 +447,192 @@ module _ {a e r} where
...
+
+  DisjointUnion-fromAny∘toAny-∷ˡ⁻ : ∀ {x xs ys} (τ : (x ∷ xs) ⊆ ys) → DisjointUnion (fromAny (toAny τ)) (∷ˡ⁻ τ) τ
+  DisjointUnion-fromAny∘toAny-∷ˡ⁻ (y  ∷ʳ τ) = y   ∷ₙ DisjointUnion-fromAny∘toAny-∷ˡ⁻ τ
+  DisjointUnion-fromAny∘toAny-∷ˡ⁻ (xRy ∷ τ) = xRy ∷ₗ DisjointUnion-[]ˡ
+
+  -- Disjoint union of three mutually disjoint lists.
+  --
+  -- τᵢⱼ denotes the disjoint union of τᵢ and τⱼ: DisjointUnion τᵢ τⱼ τᵢⱼ
+
+  record DisjointUnion³
```
MatthewDaggitt: Definitions should be in the main file rather than the `.Properties` file

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -444,3 +447,192 @@ module _ {a e r} where
...
+
+module Disjointness {a b r} {A : Set a} {B : Set b} {R : REL A B r} where
+
+  private
+    infix 4 _⊆_
+    _⊆_ = Sublist R
+
+  -- Irreflexivity
+
+  Disjoint-irrefl : ∀{xs ys} {τ : xs ⊆ ys} → Disjoint τ τ → Null xs
```
MatthewDaggitt: Hmm this is the opposite way round to how the standard library usually defines irreflexivity. Normally its "equals to" implies "not related". Obviously we don't have a notion of equivalence here, but I think it would still be more consistent to prove not disjoint.

src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
```diff
@@ -158,3 +193,80 @@ IsWeakPushout {τ = τ} {σ = σ} rpo =
...
+DisjointUnion-diffʳ (x≈y ∷ₗ d) = _ ∷ʳ _   , cong (_  ∷ʳ_) (proj₂ (DisjointUnion-diffʳ d))
+
+------------------------------------------------------------------------
+-- A Union where the triangles commute is a
+-- Cospan in the slice category (_ ⊆ zs).
+
+record IsCospan {xs ys zs : List A} {τ₁ : xs ⊆ zs} {τ₂ : ys ⊆ zs} (u : UpperBound τ₁ τ₂) : Set a where
+  field
+    tri₁ : ⊆-trans (UpperBound.inj₁ u) (UpperBound.sub u) ≡ τ₁
+    tri₂ : ⊆-trans (UpperBound.inj₂ u) (UpperBound.sub u) ≡ τ₂
```
MatthewDaggitt: These definitions should not be in `.Properties`. But as they rely on `⊆-trans` then I think all of this stuff needs to be moved out to it's own module. `Sublist.Heterogeneous/Setoid/Propositional.Disjoint` maybe?

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -444,3 +447,192 @@ module _ {a e r} where
...
+  Disjoint-irrefl []       = []
+  Disjoint-irrefl (y ∷ₙ d) = Disjoint-irrefl d
+
+  -- Symmetry
+
+  DisjointUnion-symmetric : ∀ {xs ys xys} {zs} {τ₁ : xs ⊆ zs} {τ₂ : ys ⊆ zs} {τ : xys ⊆ zs} →
+    DisjointUnion τ₁ τ₂ τ → DisjointUnion τ₂ τ₁ τ
```
MatthewDaggitt: Please could you follow the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) with respect to indentation of types.

andreasabel:
I now added an example in 77b4aa7a712c685a2f772631fbf8dc0a0aa489d5 to demonstrate how sublist union and its properties can be used to reason about lambda-calculus with unique bound variables.  (Gosh, took me days to turn the categorical diagrams into Agda code.)

Thanks for the reviews so far @gallais and @MatthewDaggitt !  I shall see how to reorganize the contribution to match better with the convention of the standard-library.

The separation of operations/definitions one the one hand and their properties on the other hand gets a bit blurred if we use integrated verification (such as I saw fit in some cases).  Basically, such a definition will take a form like
```agda
foo : forall arg -> Foo arg
```
where `Foo` is a record aggregating the outputs of operation `foo` plus the post conditions.  The individual cases of `foo` may use constructions of `Foo` of the form `Foo x -> Foo (cons x)`.  
I guess this is fine, but it seems out of the spirit of the standard library for lemmata that gather the proven properties in a record, like
```agda
lemma : forall arg -> Lemma (foo arg)
```
The `Lemma` needs a record definition whose sole purpose is for the formulation of a property.  The question is whether such a record definition may live in a `.Properties` file.

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
andreasabel: Note that _not Null implies not Disjoint_ is a trivial contraposition of my property.  My property seems strictly more useful than 
```agda
∀{x xs ys} {τ : (x :: xs) ⊆ ys} → Disjoint τ τ → \bot
```  
You get the latter as a special case of my property when instantiating `xs` to a non-empty list.

MatthewDaggitt:
Thanks for the example, I think your contribution has instantly rocketed up to one of the best documented additions to the library :smile: 

> The separation of operations/definitions one the one hand and their properties on the other hand gets a bit blurred if we use integrated verification (such as I saw fit in some cases).

Yup, agreed. 

> The Lemma needs a record definition whose sole purpose is for the formulation of a property. The question is whether such a record definition may live in a .Properties file.

I guess if the record is purely capturing the structure of a particular lemma then it's fine for it to live in `.Properties` file. However if it captures some general relationship that is itself subject to various lemmas then it should definitely live in the main file. Have a great vacation, and we can thrash out the finer details when you get back :+1:

PR #864, opened 2019-08-21
[ fix #826 ] Coercible type class

### LINKED ISSUE #826: Definition of Coercion
It's quite annoying to have to convert an FFI-bound function's arguments
to make it an "internal" function. Something like what follows allows us
to perform the conversion automatically:

```agda
import Data.Maybe as Maybe
import Data.Product as Prod

open import Foreign.Haskell.Pair  using (Pair) public
open import Foreign.Haskell.Maybe using (Maybe) public
open import Function
open import Level

private
  variable
    a b c d : Level
    A : Set a
    B : Set b
    C : Set c
    D : Set d

record Coerce {a b} (A : Set a) (B : Set b) : Set (a ⊔ b) where
  constructor _,_
  field to   : A → B
        from : B → A
open Coerce {{...}} public renaming (to to toForeign; from to fromForeign)

refl : Coerce A A
refl = id , id

instance

  coerce-pair : {{_ : Coerce A B}} {{_ : Coerce C D}} →
                Coerce (A Prod.× C) (Pair B D)
  Coerce.to   coerce-pair =
    Foreign.Haskell.Pair.toForeign ∘′ Prod.map toForeign toForeign
  Coerce.from coerce-pair =
    Prod.map fromForeign fromForeign ∘′ Foreign.Haskell.Pair.fromForeign


  coerce-maybe : {{_ : Coerce A B}} → Coerce (Maybe.Maybe A) (Maybe B)
  Coerce.to   coerce-maybe =
    Foreign.Haskell.Maybe.toForeign ∘′ Maybe.map toForeign
  Coerce.from coerce-maybe =
    Maybe.map fromForeign ∘′ Foreign.Haskell.Maybe.fromForeign


  coerce-function : {{_ : Coerce A B}} {{_ : Coerce C D}} →
                    Coerce (A → C) (B → D)
  Coerce.to   coerce-function f = toForeign ∘′ f ∘′ fromForeign
  Coerce.from coerce-function f = fromForeign ∘′ f ∘′ toForeign


  coerce-nat : Coerce ℕ ℕ
  coerce-nat = refl


postulate
  predFFI : ℕ → Maybe ℕ

pred : ℕ → Maybe.Maybe ℕ
pred = fromForeign predFFI
```

gallais:
To cross FFI boundaries easily, we can use `coerce` to convert
between the FFI-friendly types and the ones actually use in the
core library with all the functions we know and love.

Note that I had to swap Data.Maybe.Base's constructors so that
they would appear in the same order as the ones for Maybe as
defined in ghc's base.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,8 @@ script:
   - agda $RTS_OPTIONS -i . -i src/ --safe EverythingSafeSizedTypes.agda
   # detecting basic compilation errors
   - agda $RTS_OPTIONS -i . -i src/ -c --no-main Everything.agda
+  # compiling & running the examples using the FFI
+  - agda $RTS_OPTIONS -i . -i src/ -c README/Foreign/Haskell.agda && ./Haskell
   # building the docs
   - agda $RTS_OPTIONS -i . -i src/ --html safe.agda
   - agda $RTS_OPTIONS -i . -i src/ --html index.agda
--- a/GenerateEverything.hs
+++ b/GenerateEverything.hs
@@ -38,6 +38,7 @@ unsafeModules = map toAgdaFilePath
   , "Data.Word.Unsafe"
   , "Debug.Trace"
   , "Foreign.Haskell"
+  , "Foreign.Haskell.Coerce"
   , "Foreign.Haskell.Maybe"
   , "Foreign.Haskell.Pair"
   , "IO"
--- a/README/Foreign/Haskell.agda
+++ b/README/Foreign/Haskell.agda
@@ -89,7 +89,7 @@ open import Relation.Nullary.Negation
 -- example program using uncons, catMaybes, and testChar
 
 main = run $
-  ♯ readFiniteFile "Haskell.agda" {- read this file -} >>= λ f →
+  ♯ readFiniteFile "README/Foreign/Haskell.agda" {- read this file -} >>= λ f →
   ♯ let chars   = toList f in
     let cleanup = catMaybes ∘ List.map (λ c → if testChar c then just c else nothing) in
     let cleaned = dropWhile ('\n' ≟_) $ cleanup chars in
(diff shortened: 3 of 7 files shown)


Zambonifofex:
This is really neat! :open_mouth: I didn’t know about `Unsafe.Coerce`. :thinking:

Either way, I was wondering if there is any reason to have a bunch of sequential `instance postulate`  blocks and not a single one. :thinking: I guess it’s a matter of style, sure, but I personally think that indentation is less noise than non‐whitespace when reading code.

I don’t know, maybe I’m just being picky, though. :sweat_smile:

gallais:
> I was wondering if there is any reason to have a bunch of sequential instance postulate blocks and not a single one.

My experiment started with `COMPILE` pragmas attached to each postulated instance
because I was trying to use `Data.Coerce`'s safe `coerce` rather than `unsafeCoerce`.
I thought it was clearer to have the `COMPILE` pragma right after the corresponding
`postulate`, hence the split.

Now that we do not have this interleaving, it is true that we could put them all
in the same block.

MatthewDaggitt:
Looks good to me!

gallais:
I may add some more examples & actually get travis to compile them to make sure
everything works out.

Zambonifofex:
Hello! Sorry for having taken so long to give this feedback, but why does `TrustMe` start with an uppercase letter? Shouldn’t that be reserved for types and type functions?

Also, while I’m commenting, perhaps kind of a minor nitpick, but I feel like the comments inside the `instance` block should be indented like the rest of the code.

MatthewDaggitt:
Hmm that's a good point about `TrustMe`...

gallais:
We need a linter. :sweat_smile:

MatthewDaggitt:
I know. Anyone have any keen undergraduates looking for projects? :smile:

PR #865, opened 2019-08-22
Work-arounds for failing forcing translation

UlfNorell:
See agda/agda#3903. Previously Agda sometimes picked the wrong place to move a forced pattern to, resulting (in the worst case) in segfaulting or badly behaved compiled code. Now the forcing translation gives up when it's not sure, and this happens in a small number of places in the standard library. The workaround is to move the match manually to the correct non-forced position.

### DIFF
--- a/src/Data/Integer/Properties.agda
+++ b/src/Data/Integer/Properties.agda
@@ -876,8 +876,8 @@ m-n≤m m n = ≤-steps-neg n ≤-refl
 m≤n⇒m-n≤0 : ∀ {m n} → m ≤ n → m - n ≤ + 0
 m≤n⇒m-n≤0 (-≤+ {n = n})         = ≤-steps-neg n -≤+
 m≤n⇒m-n≤0 (-≤- {n = n} n≤m)     = ≤-trans (⊖-monoʳ-≥-≤ n n≤m) (≤-reflexive (n⊖n≡0 n))
-m≤n⇒m-n≤0 (+≤+ {n = 0} z≤n)     = +≤+ z≤n
-m≤n⇒m-n≤0 (+≤+ {n = suc n} z≤n) = -≤+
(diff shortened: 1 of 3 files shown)


nad:
Isn't the idea that the experimental branch of the standard library should be compatible with the development version of Agda? This commit is currently not on the experimental branch.

MatthewDaggitt:
Yup, sorry will merge in this evening. Alternatively if it's urgent and you
want to bring it up to date with master before then, feel free.

nad:
I tried to merge, but the library failed to type-check:
```
[…]/Everything.agda:437,1-20
Data.Fin.Dec was deprecated in v0.17.
Use Data.Fin.Properties and Data.Fin.Subset.Properties instead.
  import Data.Fin.Dec
[…]/Everything.agda:475,1-25
Failed to find source of module Data.Float.Unsafe in any of the
following locations:
  […]/src/Data/Float/Unsafe.agda
  […]/src/Data/Float/Unsafe.lagda
  […]/Data/Float/Unsafe.agda
  […]/Data/Float/Unsafe.lagda
  […]/lib/prim/Data/Float/Unsafe.agda
  […]/lib/prim/Data/Float/Unsafe.lagda
  import Data.Float.Unsafe
```

MatthewDaggitt:
Okay, I've fixed the `Dec` problem and pushed the updated branch. As for `Data.Float.Unsafe` module, it should no longer exist on experimental. Maybe your `Everything.agda` file isn't being regenerated properly? It doesn't appear in mine. I guess try deleting it and try again? Anyway I'd wait to see if [Travis](https://travis-ci.org/agda/agda-stdlib/builds/581147255) goes green....

nad:
Thanks. (I did not even try to regenerate `Everything.agda`.)

PR #869, opened 2019-08-28
[ fix #868 ] Add Data.AVL.Map

### LINKED ISSUE #868: Add (non-indexed) Data.AVL.Map
At the moment it's a bit of a pain to use a `Map` à la Haskell because we have
to get acquainted with `Data.AVL.Value` even though there are no dependencies
between the key type and the value type and hence no need for the coherence
requirement imposed by `Value`.

We can of course use `Data.AVL.Value.const` to get such a map, but it would
be nice (and would probably help with type inference) to have a lighter interface.

rubenpieters:
Implements the simple interface for maps in module `Data.AVL.Map`. Fixes #868 .

I kept the same order of the functions as in the `Data.AVL` module.

Let me know if something needs to be changed.

### DIFF
--- a/src/Data/AVL/Map.agda
+++ b/src/Data/AVL/Map.agda
@@ -1,0 +1,94 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Maps from keys to values, based on AVL trees
+-- This modules provides a simpler map interface, without a dependency
+-- between the key and value types.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary using (StrictTotalOrder)
+
+module Data.AVL.Map
+  {a ℓ₁ ℓ₂} (strictTotalOrder : StrictTotalOrder a ℓ₁ ℓ₂)
+  where
+
+open import Data.Bool.Base using (Bool)
+open import Data.List.Base using (List)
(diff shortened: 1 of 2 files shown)


src/Data/AVL/Map.agda
```diff
@@ -0,0 +1,94 @@
...
+open import Data.Product using (_×_)
+open import Level using (_⊔_)
+
+import Data.AVL strictTotalOrder as AVL
+open StrictTotalOrder strictTotalOrder renaming (Carrier to Key)
+
+------------------------------------------------------------------------
+-- The map type
+
+Map : ∀ {i} → (v : Set i) → Set (a ⊔ i ⊔ ℓ₂)
```
MatthewDaggitt: We usually mark types using capitals, so I think it would be more consistent to call the universe level `v` and the set `V : Set v` so the type would become:
```
Map : ∀ {v} (V : Set v) → Set (a ⊔ v ⊔ ℓ₂)
```
Otherwise looks great!

rubenpieters: Thanks, I updated the PR according to this suggestion.

MatthewDaggitt:
Great! Thanks for the PR. I'll leave it open for a few days to see if anyone else has any feedback and then I'll merge it in.

PR #873, opened 2019-08-29
Proofs about identity and composition for the new function hierarchy

MatthewDaggitt:
Transferred across the identity and composition constructs to the new function hierarchy. Seems to work pretty well! 

The only major change is that the packages didn't previously export the equality proofs so I've added two modules `Eq₁` and `Eq₂` that contains the equalities for the domain and codomain in the same way that `Eq` does for the binary relation hierarchy.

Comments welcome as always!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Other non-backwards compatible changes
 
 #### New function hierarchy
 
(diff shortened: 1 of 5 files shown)


MatthewDaggitt:
Not sure why Travis is running out of memory, the branch is passing test locally. Merging in.

PR #878, opened 2019-09-02
[ new ] fresh lists

gallais:


README/Data/Trie/NonDependent.agda
```diff
@@ -148,33 +167,38 @@ module Lexer
...
-  LPAR RPAR : TOK
-  ID : String → TOK
 
-toks : List⁺ (String × TOK)
-toks = ("let" , LET)
-     ∷ ("="   , EQ)
-     ∷ ("in"  , IN)
-     ∷ []
+letIn : Lexer 0ℓ
+letIn = record { LetIn } module LetIn where
```
gallais: Fascinating. This is valid syntax in 2.6.1 but in 2.6.0.1 the
named where-clause `LetIn` is not in scope for the RHS.

JacquesCarette:
This is very cool.  I believe that I have an immediate use for this -- formalizing dependent records, with a proper freeness predicate on labels.  Though it might need a generalization to dependent lists, but I think the general idea ought to work as is.

EdNutting:
I am currently using the following functions/types/properties based on a slightly different version of FreshList that if included in stdlib would be immediately useful to me (I guess some can be generated using existing features of the lib):

* Membership based on Any
* Non-membership as negation of above
* All
* Empty/NonEmpty, as well as:
    * Empty=>not NonEmpty
    * NonEmpty => not Empty
    * Empty? / NonEmpty? (decidables)
    * x `member` xs => NonEmpty
* x `member` xs AND y `not member` xs => x `not equal` y
* toList/fromList
* length
* x `not member` xs => fresh x xs
* fresh x xs => x `not member` xs
* x `not member` xs => x `not equal` head xs
* length xs  `equal` 1 + (length (drop xs))
* concat (_++_)
* (ls rs : FreshList A) -> (Exists x: x `member` rs AND x `not member` ls) -> length ls < length rs
* (ls rs : FreshList A) -> x `member` rs -> x `member` (ls ++ rs)
* (ls rs : FreshList A) -> x `member` ls -> x `member` (ls ++ rs)

gallais:
Is `concat` a partial function? Or do you have some hidden precondition?

EdNutting:
@gallais Sorry, missed off the precondition last night: _++_ : (xs ys : FreshList A) → ({i : A} → i ∈ xs → i ∉ ys) → FreshList A

EdNutting:
I suppose the alternative is to require a decidable equality for A and then test & exclude anything from xs that appears in ys. But that wasn't what I needed.

src/Data/List/Fresh/Properties.agda
```diff
@@ -0,0 +1,41 @@
...
+
+
+open import Data.List.Fresh
+
+private
+  variable
+    a b p r : Level
+    A : Set a
+    B : Set b
+    R : Rel A r
```
MatthewDaggitt: Ooh a forbidden generalisation under the current rules I'm afraid!

src/Data/List/Fresh/Relation/Unary/All.agda
```diff
@@ -0,0 +1,34 @@
...
+open import Data.List.Fresh
+
+private
+  variable
+    a b p q r : Level
+    A : Set a
+    B : Set b
+    P : Pred A p
+    Q : Pred A q
+    R : Rel A r
```
MatthewDaggitt: And some more.

src/Relation/Unary.agda
```diff
@@ -161,6 +161,14 @@ syntax IUniversal P = ∀[ P ]
 Decidable : Pred A ℓ → Set _
 Decidable P = ∀ x → Dec (P x)
 
+-- Erasure: A decidable predicate gives rise to another one, more
+-- amenable to η-expansion
+
+⌊_⌋ : {P : Pred A ℓ} → Decidable P → Pred A ℓ
```
MatthewDaggitt: Hmm I'm a bit worried about this name clashing with `⌊_⌋` in `Relation.Nullary.Decidable.Core`. Are there any other symbols we could use? Maybe `⌞_⌟`?

gallais: Alternatively the `Core` definition could be deprecated & renamed to `isYes`
(and we could throw in `isNo` too for good measure)?

gallais:
I have added most of @EdNutting's results except for `head` and `concat` (and
related properties).

src/Relation/Unary.agda
MatthewDaggitt: Yup sounds good to me. Sorry for the long reply time!

andreasabel:
When writing a type checker for C-- I though about how to represent variable contexts without duplicates.  I first thought to use fresh lists which I new from paper by Catarina Coquand (CSL 93).  But then I realized it would be very awkward to work with them.  For instance, it is inproportionally hard to establish the intuitively trivial property that one can delete elements from a fresh list while preserving the "freshness" conditions.  
I then settled for a notion of fresh list that couples a list `xs` with the predicate
```
UniqueList xs = ∀{x} (i j : x ∈ xs) → i ≡ j
```
The proposition `∀ (i j : x ∈ xs) → i ≡ j` states that all occurrences of `x` in `xs` are the same.  Or in HoTT terms, that the type `x ∈ xs` is a proposition.

From such a notion one trivially gets that `UniqueList` is preserved by taking a sublist.  I expect this to be painful with the inductive-recursive definition of fresh lists.

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -97,3 +100,18 @@ dec-no (no ¬p′) ¬p = ¬p′ , refl
...
+
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 1.2
+
+⌊_⌋ = isYes
```
andreasabel: I'd consider `⌊_⌋` as untouchable.   This operator might have been around since a decade. There are several published papers using it.

JacquesCarette:
@andreasabel Are there some public codes that use `UniqueList`? The part where I'm stuck on is the appropriate notion of renaming that makes preservation of the `UniqueList` property straightforward. This shouldn't be so hard, and yet...

Issue #870, opened 2019-08-28
Deprecating `Data.Table` in favour of `Data.Vec.Functional`


MatthewDaggitt:
@gallais if you get to use James' [blog post](https://personal.cis.strath.ac.uk/james.wood.100/blog/html/VecMat.html) to argue for adding [Reflects](https://github.com/agda/agda-stdlib/issues/225#issuecomment-525622181) into the library, I'm going to use it to argue for fixing `Data.Table` :stuck_out_tongue_closed_eyes: 

I share @James's opinion that the record structure doesn't add anything for inference. Furthermore it's fairy difficult to work with. I've tried twice now to integrate into my own library  and failed because:
-  it makes it makes it very difficult to define recursive functions over it because you have to keep removing and adding the record constructor.
- it makes it much more difficult to compose them together to define matrices as again the record constructor gets in the way.

I propose leaving `Data.Table` in place, deprecating it, and add the non-record version as `Data.Vec.Functional`

gallais:
I agree that insisting on making `Table` a record was a mistake on my part.
As james highlights, it would have made sense for a dependent function space
but the current definition is just needlessly painful to work with.

laMudri:
I'd be happy to work on `Data.Vec.Functional` before AIM. For backwards compatibility, it might make sense to reimplement all of the definitions in the `Data.Table` hierarchy to just unwrap/wrap all `Table`s and then call stuff from `Data.Vec.Functional`. Minor matters: what should the type be called? `Table`? `Vec`? Something else?

MatthewDaggitt:
> I'd be happy to work on Data.Vec.Functional before AIM

That'd be great, thanks! 

> For backwards compatibility, it might make sense to reimplement all of the definitions in the Data.Table hierarchy to just unwrap/wrap all Tables and then call stuff from Data.Vec.Functional

Hmm I'm not sure about that. The idea with backwards compatibility is not to change any observable behaviour. The problem with reimplementing `Data.Table` in terms of the new `Data.Vec.Functional` is that it'll probably change the reduction behaviour. To be honest I'd just leave `Data.Table` as is and slap a deprecated warning on it.

> Minor matters: what should the type be called? Table? Vec? Something else?

I've gone off `Table`. How about `Vector` as i) it distinguishes it from `Vec` and ii) I like `Matrix` better than `Mat`. Open to other suggestions though?

laMudri:
[...]

Issue #697, opened 2019-04-13
Policy on generalised variables


MatthewDaggitt:
The new `variable` keyword has the potential to get rid of a lot of the anonymous modules that are scattered about the library. Thought I should start a discussion about how best to go about using it.

Two main questions immediately spring to mind I guess:

1. What should be put under `variable` and what should be left in the type

In my opinion we should definitely generalise universe levels and sets. After that it becomes a little less obvious to me what should and shouldn't be generalised. Although it might be tempting to generalise everything possible, it might be possible that going overboard will impact readability. For example I'm wondering if the type signature:
```
filter : Decidable {A = A} P → List A → List A
```
is more or less clear than:
```
filter : ∀ {P : Pred A p} → Decidable P → List A → List A
```

2. Should we have a central module of generalised variables or should they be local to each file?

Variables are exported by modules so in the former case the generalised variable file could act as a style guide. However I suspect that such a setup isn't flexible enough and we'd still have to add local variable declarations to each file. In the latter case I guess we have add a section to the prelude of each file where we declare variables in a private scope?

JacquesCarette:
On `filter`, I definitely prefer the second signature. The first asks me to remember too much about what the arguments to `Decidable` are; and the need to specify an implicit also seems to lean in that direction.

Also, before jumping on this, I think it is indispensable to have some kind of emacs command where, when you hover (or select?) a function name, you get its **full** signature.

For second, to me it's clear already in the question: you want a file full of **globals**, seriously?

Both anonymous modules and variables have a common problem: they make reading the library a "whole file" affair, rather than being able to jump to where the function is actually defined. In some ways, "jump to definition" of a function should jump where the definition starts, which is now logically at the first module/variable declaration...

Also note that Coq uses sections to separate things -- in other words, they mix modules (anonymous and otherwise) and variables. So there is likely a lesson in that too.

MatthewDaggitt:
> For second, to me it's clear already in the question: you want a file full of globals, seriously?

Isn't that the point of most of the library? :stuck_out_tongue: Haha, but yes I agree with you. Thought I'd raise it as a possible design decision just in case.

> Both anonymous modules and variables have a common problem: they make reading the library a "whole file" affair, rather than being able to jump to where the function is actually defined. In some ways, "jump to definition" of a function should jump where the definition starts, which is now logically at the first module/variable declaration...

[...]

PR #882, opened 2019-09-06
Functional vectors

laMudri:
Motivation: #870

This pull request introduces vectors defined by index notation. That is, it defines vectors to be functions from `Fin n` to `A`. Assuming function extensionality, this is equivalent to `Vec A n`, but prioritises introduction and elimination via `tabulate` and `lookup`, rather than `[]`/`_∷_` and induction, respectively. This is the same as `Data.Table`, except for not having the unnecessary record wrapper (see #870 and the linked blog post for why it is unnecessary).

I'd like to have a somewhat complete collection of definitions and lemmas before this is merged. This should match use cases for functional vectors, rather than being a copy of what we have for inductive vectors. Remaining issues are as follows:

- [ ] What other type definitions/modules do we need?
- [ ] What other lemmas do we need?
- [ ] Does anything already done need modifying?

An important use case is vector/matrix calculus.

- [ ] Where do we put things about matrices?
- [ ] How do we organise all of the algebraic properties?
- [ ] Do we generalise to arbitrary tensors?

### DIFF
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -14,7 +14,7 @@ module Data.Fin.Base where
 
 open import Data.Empty using (⊥-elim)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; z≤n; s≤s)
-open import Data.Sum as ⊎ using (_⊎_; inj₁; inj₂)
+open import Data.Sum as Sum using (_⊎_; inj₁; inj₂)
 open import Function.Core using (id; _∘_; _on_)
 import Data.Nat.Properties as ℕₚ
 open import Level using () renaming (zero to ℓ₀)
@@ -128,7 +128,7 @@ strengthen (suc i) = suc (strengthen i)
 splitAt : ∀ m {n} → Fin (m ℕ.+ n) → Fin m ⊎ Fin n
 splitAt zero i = inj₂ i
 splitAt (suc m) zero = inj₁ zero
-splitAt (suc m) (suc i) = ⊎.map suc id (splitAt m i)
+splitAt (suc m) (suc i) = Sum.map suc id (splitAt m i)
 
 ------------------------------------------------------------------------
 -- Operations
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -14,7 +14,8 @@ module Data.Fin.Base where
 
 open import Data.Empty using (⊥-elim)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; z≤n; s≤s)
-open import Function.Core using (_∘_; _on_)
+open import Data.Sum as ⊎ using (_⊎_; inj₁; inj₂)
+open import Function.Core using (id; _∘_; _on_)
 import Data.Nat.Properties as ℕₚ
 open import Level using () renaming (zero to ℓ₀)
 open import Relation.Nullary using (yes; no)
@@ -119,6 +120,14 @@ lower₁ {suc n} (suc i) ne = suc (lower₁ i λ x → ne (cong suc x))
 strengthen : ∀ {n} (i : Fin n) → Fin′ (suc i)
 strengthen zero    = zero
 strengthen (suc i) = suc (strengthen i)
+
+-- split+ m "i" = inj₁ "i"      if i < m
+--                inj₂ "i - m"  if i ≥ m
+
+split+ : ∀ m {n} → Fin (m ℕ.+ n) → Fin m ⊎ Fin n
+split+ zero i = inj₂ i
+split+ (suc m) zero = inj₁ zero
+split+ (suc m) (suc i) = ⊎.map suc id (split+ m i)
 
 ------------------------------------------------------------------------
 -- Operations
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -121,13 +121,14 @@ strengthen : ∀ {n} (i : Fin n) → Fin′ (suc i)
 strengthen zero    = zero
 strengthen (suc i) = suc (strengthen i)
 
--- split+ m "i" = inj₁ "i"      if i < m
---                inj₂ "i - m"  if i ≥ m
+-- splitAt m "i" = inj₁ "i"      if i < m
+--                 inj₂ "i - m"  if i ≥ m
+-- This is dual to splitAt from Data.Vec.
 
-split+ : ∀ m {n} → Fin (m ℕ.+ n) → Fin m ⊎ Fin n
-split+ zero i = inj₂ i
-split+ (suc m) zero = inj₁ zero
-split+ (suc m) (suc i) = ⊎.map suc id (split+ m i)
+splitAt : ∀ m {n} → Fin (m ℕ.+ n) → Fin m ⊎ Fin n
+splitAt zero i = inj₂ i
+splitAt (suc m) zero = inj₁ zero
+splitAt (suc m) (suc i) = ⊎.map suc id (splitAt m i)
 
 ------------------------------------------------------------------------
 -- Operations
--- a/src/Data/Vec/Functional.agda
+++ b/src/Data/Vec/Functional.agda
@@ -75,7 +75,7 @@ _⊛_ : ∀ {n} → Vector (A → B) n → Vector A n → Vector B n
 _⊛_ = _ˢ_
 
 zipWith : (A → B → C) → ∀ {n} → Vector A n → Vector B n → Vector C n
-zipWith f xs ys = replicate f ⊛ xs ⊛ ys
+zipWith f xs ys i = f (xs i) (ys i)
 
 zip : ∀ {n} → Vector A n → Vector B n → Vector (A × B) n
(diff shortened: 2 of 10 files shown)


src/Data/Vec/Functional.agda
```diff
@@ -0,0 +1,75 @@
...
+open import Level using (Level)
+
+infixr 5 _∷_
+
+private
+  variable
+    a b : Level
+    A : Set a
+    B : Set b
+    m n : ℕ
```
MatthewDaggitt: At the moment the standard library policy (#697) is to only use variables for levels and sets.

src/Data/Vec/Functional.agda
```diff
@@ -0,0 +1,75 @@
...
+
+Vector : Set a → ℕ → Set a
+Vector A n = Fin n → A
+
+map : (A → B) → (Vector A n → Vector B n)
+map f v = f ∘ v
```
MatthewDaggitt: I think best if we stuck to the convention of using `xs`, `ys` etc. for collections of objects. `v` is usually reserved for a single value (e.g. in membership proofs).

MatthewDaggitt:
>  What other type definitions/modules do we need?
 
Wouldn't mind adding `zip`, `zipWith` functions. For the initial pull request that's probably not too bad.

> What other lemmas do we need?

I'm happy with this for the initial PR. We can always add more later on.

>  Does anything already done need modifying?

The only use of `Data.Table` that I'm aware of is `Algebra.Properties.CommutativeMonoid` but see my comments below about what to do with this.

An important use case is vector/matrix calculus.

>  Where do we put things about matrices?

Definitely `Data.Matrix`

>  How do we organise all of the algebraic properties?

So I would imagine we'd have a `Data.X.Algebra` folder just like we have `Data.X.Relation` folder. We could then have `Data.X.Algebra.Monoid`, `Data.X.Algebra.CommutativeMonoid` etc.? Thoughts? I'd plan to relocate much of the sum lemmas in `Algebra.Properties.CommutativeMonoid` accordingly.

>  Do we generalise to arbitrary tensors?

What would your proposed definition of an arbitrary tensor look like?

laMudri:
Having defined `zipWith` and `zip` in terms of `replicate` and `_⊛_`, and having started to prove things about them, I'm thinking that the problem I highlighted in [#870](https://github.com/agda/agda-stdlib/issues/870#issuecomment-527106628) makes all of the lemmas in `Pointwise` and `All` fairly useless. Agda is unable to infer any relations, so the path of least resistance is often to go back to index notation rather than spell everything out. In contrast, with a record wrapper, things work out okay, and `zipWith⁺` can have the same form as `zipWith`, but with a more complicated type.

laMudri:
> What would your proposed definition of an arbitrary tensor look like?

I guess it's beyond the scope of this pull request, but I imagine something like this.

```agda
Tensor : Set a → Vector ℕ n → Set a
Tensor A ds = All Fin ds → A
```

`n` is the rank, and `ds` are the dimensions in each rank. I don't know what it's like programming with this, though, and I also don't know much about tensors.

src/Data/Vec/Functional.agda
```diff
@@ -0,0 +1,82 @@
...
+fromVec = V.lookup
+
+replicate : ∀ {n} → A → Vector A n
+replicate = const
+
+_⊛_ : ∀ {n} → Vector (A → B) n → Vector A n → Vector B n
+_⊛_ = _ˢ_
+
+zipWith : (A → B → C) → ∀ {n} → Vector A n → Vector B n → Vector C n
+zipWith f xs ys = replicate f ⊛ xs ⊛ ys
```
MatthewDaggitt: Can't this just be `zipWith f xs ys i = f (xs i) (ys i)`

src/Data/Vec/Functional/Relation/Unary/All.agda
```diff
@@ -0,0 +1,126 @@
...
+------------------------------------------------------------------------
+-- Operations
+
+map : {P : Pred A p} {Q : Pred A q} → P ⊆ Q → ∀ {n} → All P {n = n} ⊆ All Q
+map pq ps i = pq (ps i)
+
+------------------------------------------------------------------------
+-- map
+
+map⁺ : {P : Pred A p} {Q : Pred B q} {f : A → B} →
```
MatthewDaggitt: The convention in other parts of the library has been to put all the `⁺` and `⁻` proofs in `All.Properties` rather than `All`.

src/Data/Vec/Functional/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,48 @@
...
+  variable
+    a b p q ℓ : Level
+    A : Set a
+    B : Set b
+
+------------------------------------------------------------------------
+-- Definition
+
+Any : Pred A ℓ → ∀ {n} → Vector A n → Set ℓ
+Any P xs = ∃ \ i → P (xs i)
```
MatthewDaggitt: Can we use a lambda here just for consistency?

MatthewDaggitt:
> Having defined zipWith and zip in terms of replicate and _⊛_, and having started to prove things about them, I'm thinking that the problem I highlighted in #870 makes all of the lemmas in Pointwise and All fairly useless. Agda is unable to infer any relations, so the path of least resistance is often to go back to index notation rather than spell everything out. In contrast, with a record wrapper, things work out okay, and zipWith⁺ can have the same form as zipWith, but with a more complicated type.

Is it fixed by my proposed simpler definition for `zipWith`?

> ```agda
> Tensor : Set a → Vector ℕ n → Set a
> Tensor A ds = All Fin ds → A
> ```
> 
> n is the rank, and ds are the dimensions in each rank. I don't know what it's like programming with this, though, and I also don't know much about tensors.

Yup, maybe something to explore in future.

src/Data/Vec/Functional.agda
laMudri: This is definitionally equal (via η-expansion), so I prefer the one that makes the code duplication obvious.

src/Data/Vec/Functional.agda
```diff
@@ -0,0 +1,82 @@
...
+head : ∀ {n} → Vector A (suc n) → A
+head xs = xs zero
+
+tail : ∀ {n} → Vector A (suc n) → Vector A n
+tail xs = xs ∘ suc
+
+uncons : ∀ {n} → Vector A (suc n) → A × Vector A n
+uncons xs = head xs , tail xs
+
+_++_ : ∀ {m n} → Vector A m → Vector A n → Vector A (m + n)
```
andreasabel: It feels like append should be factored through the isomorphism
`Fin (m + n) ~= Fin m + Fin n`.

laMudri: As far as I can see, we don't have this isomorphism in stdlib. We *really* should have it, so I'll take this as an excuse to implement it.

It changes the computational behaviour, but possibly for the better.

JacquesCarette: Yes, it should. And I have those proofs in https://github.com/JacquesCarette/pi-dual/blob/master/Univalence/FinEquivPlusTimes.agda that I've been meaning to port & submit to stdlib. This might be the push I needed.

JacquesCarette: @laMudri go ahead and 'borrow' any parts of my code if it helps. I wasn't necessarily going to do it today, so it sounds like you'll get to it sooner.

src/Data/Fin/Base.agda
```diff
@@ -120,6 +121,14 @@ strengthen : ∀ {n} (i : Fin n) → Fin′ (suc i)
...
 strengthen (suc i) = suc (strengthen i)
 
+-- split+ m "i" = inj₁ "i"      if i < m
+--                inj₂ "i - m"  if i ≥ m
+
+split+ : ∀ m {n} → Fin (m ℕ.+ n) → Fin m ⊎ Fin n
+split+ zero i = inj₂ i
+split+ (suc m) zero = inj₁ zero
+split+ (suc m) (suc i) = ⊎.map suc id (split+ m i)
+
```
gallais: Maybe `splitAt` instead? It's already used for a similar function on `Vec`.

src/Data/Fin/Base.agda
```diff
@@ -14,7 +14,8 @@ module Data.Fin.Base where
 
 open import Data.Empty using (⊥-elim)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; z≤n; s≤s)
-open import Function.Core using (_∘_; _on_)
+open import Data.Sum as ⊎ using (_⊎_; inj₁; inj₂)
+open import Function.Core using (id; _∘_; _on_)
```
gallais: We usually use `X` as the short name for `Data.X`.

src/Data/Vec/Functional.agda
MatthewDaggitt: Hmm I'm not convinced about this. I definitely prefer the one that makes it clearer to the reader what's actually going on.  At the moment you have to lookup the definition of 3 different functions spread across two modules just to work out what the function actually does with `f`, `xs` and `ys`.

PR #888, opened 2019-09-12
PR Bin.Minus.

mechvel:
Please, consider this my comit of September 12, 2019:  
new file:    src/Data/Bin/Minus.agda

This is subtraction on (new) Bin, and its properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,8 @@ we would encourage to the new hierarchy in the medium to long term.
 #### Re-implementation of `Data.Bin`
 
 * `Data/Bin.agda` and `Data.Bin/*.agda`  of lib-1.0 are removed,
-  added new `Data.Bin.Base, Data.Bin.Properties`.
+   added new `Data.Bin.Base, Data.Bin.Induction, Data.Bin.Minus,
+   Data.Bin.Properties`.
   This total change of the Bin part is done for the following reasons.
   1) Many necessary functions and proofs are added.
   2) After this has been done, the author noticed (decided) that the whole
@@ -107,7 +108,7 @@ The following new modules have been added to the library:
   Data.Bin
   Data.Bin.Base
   Data.Bin.Induction
-  Data.Bin.Ordering
+  Data.Bin.Minus
   Data.Bin.Properties
 
   Data.List.Kleene
--- a/src/Data/Bin/Minus.agda
+++ b/src/Data/Bin/Minus.agda
@@ -1,0 +1,496 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Subtraction on Bin and some of its property proofs.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Bin.Minus where
+
+open import Data.Bin.Base
+open import Data.Bin.Properties
+open import Data.Empty using (⊥)
+open import Data.Nat as ℕ using (ℕ)
+import Data.Nat.Properties as ℕₚ
+open import Data.Product using (_,_; proj₁; proj₂; ∃)
+open import Data.Sum using (inj₁; inj₂)
+open import Function using (_∘_; _$_)
+open import Relation.Binary using (Tri; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
+open import Relation.Binary.PropositionalEquality as P using
+  (_≡_; _≢_; _≗_; refl; sym; trans; cong; cong₂; resp₂; subst)
+import Relation.Binary.Reasoning.Base.Triple as InequalityReasoning
+open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Algebra.FunctionProperties {A = Bin} _≡_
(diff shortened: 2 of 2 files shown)


MatthewDaggitt:
So my main comment is that your definition of minus is 140 lines long (!) which makes it very difficult to read. One of the reasons for this is that your using the "construct the value with the proof" style. As you yourself pointed out in `Data.Nat.GCD` these tend to lead to poor performance as the value cannot be calculated independently of the proof.

I think it'd be much cleaner and shorter to define `_∸_` first and then prove properties about it.

mechvel:
> So my main comment is that your definition of minus is 140 lines long (!) which makes it very 
> difficult to read. One of the reasons for this is that your using the "construct the value with the 
> proof" style

* I see in  ``minus`` only  127 lines, 19 of which are empty or are comments.
  This makes it  108 lines.
* If I separate _∸_ from proofs accumulation, then the code union of _∸_ and the needed proofs
  will not be smaller, probably it will be larger.

> As you yourself pointed out in Data.Nat.GCD these tend to lead to poor performance as the value > cannot be calculated independently of the proof.

For my code proposed to Standard library, I check the performance. If carrying proofs essentially effects the performance, then I would change the code by separating proofs.

Please see, ``http://www.botik.ru/pub/local/Mechveliani/inAgda/gcdProp1.zip``,
and look there into ``Readme.lagda``.
"
NatGCD.agda  provides a simple enough code for  ``gcd : (m n : ℕ) → GCD m n``
which also accumulates all the proofs recursively.
...
The test program  GCDTest.agda for gcd! shows that on the numbers of 10-40 decimal digits it is
        1.6 times slower than gcd of lib-1.1,

on the numbers of 10.000-40.000 decimal digits it is
        1.2 times slower than gcd of lib-1.1.
                        
So that in this case carrying proofs does not effect performance any essentially.
"

If ``minus`` suffered in performance, then ``divMod`` and ``gcd`` also would. 
But they do not, at least the GCDTest for Bin does not show this.

Please, accept my carrying-proofs versions for ``minus, divMod, gcd`` for Bin
(and for general ``gcd`` in the "Generic GCD" proposal).
And the best performance test for Bin.Minus.minus is the test for Bin.GCD.gcd.

MatthewDaggitt:
> If I separate minus from proofs accumulation, then the code union for ∸ and the needed
proofs will not be smaller, probably it will be larger.

That's fine because at least if they were separate then it would be clear how minus was computed. At the moment it takes serious mental effort to pick out the values out from the equality proofs.

> If minus for Bin suffered in performance, then divMod and gcd also would.
But they do not, at least the GCDTest for Bin does not show this.

Hmm I find that surprising. Perhaps it's because the proof being computed is always actually `refl`... Okay, I withdraw my performance argument. Apologies, obviously I was extrapolating without any actual data.

Finally a third reason to separate them is that I think we should be consistent with `Data.Nat` and `Data.Integer` and have `_∸_` in `Data.Nat.Binary.Base` and all the proofs in `Data.Nat.Binary.Properties`. More complicated operations such as `GCD` and `log` can be in their own files as per `Data.Nat`

Let me turn it around then and ask you what is the advantage in this case of having the proof combined with the calculation of the value?

mechvel:
> what is the advantage in this case of having the proof combined with the calculation of the value?

First, look into  ``gcd`` for Nat in 
``http://www.botik.ru/pub/local/Mechveliani/inAgda/gcdProp1.zip``  in  NatGCD.agda.

It implements gcd in a single recursive loop, together with the three proofs implementing the definition. If you separate the main value computation, then it will be _four loops_ to program:
one for the proper value and three for the necessary proofs.
Is not this evident that this NatGCD.gcd is more clear?
The same is with  divMod.

I doubt about Minus.agda. I start separating proofs, but this leads to complication in proofs.

MatthewDaggitt:
> First, look into gcd for Nat in
http://www.botik.ru/pub/local/Mechveliani/inAgda/gcdProp1.zip in NatGCD.agda.
>
> It implements gcd in a single recursive loop, together with the three proofs implementing the definition. If you separate the main value computation, then it will be four loops to program:
one for the proper value and three for the necessary proofs.
Is not this evident that this NatGCD.gcd is more clear?
The same is with divMod.

I realise this is highly subjective topic, but in my opinion it is not clearer. It's like someone saying "Okay I'm going to prove three separate lemmas to you on the whiteboard". Given the choice would you rather have them explained to you sequentially or would you rather they explained them in parallel jumping around between each of the proofs? The latter is how I feel when reading your definition of GCD. For me it requires a lot of extra mental effort as now not only do I have to understand what the code does, but I also have to maintain a mental mapping from the code to each proof... It also seems to require you to have 3 levels of nested `where` blocks which is not much fun to keep track of either.

Sorry I know this is all highly subjective, and what's clearer to one person may not be to another. However I genuinely believe that it's best for the library to stick to one style and use it consistently. If you have a strong technical reason why the definition needs to be in this format then I'm happy to listen.  I'm sorry, I know I play this card a **lot**.

> I doubt about Minus.agda. I start separating proofs, but this leads to complication in proofs.

What are the complications?

MatthewDaggitt:
Superceded by #914

PR #840, opened 2019-07-15
Order for `Data.Bin`

mechvel:
You branch is updated according to «origin/master».

changed:      CHANGELOG.md
changed:      src/Data/Bin.agda
new file:     src/Data/Bin/Ordering.agda
changed:      src/Data/Bin/Properties.agda
changed:      src/Data/Nat/Properties.agda

PR #889, opened 2019-09-12
Remove invalid brackets in function arguments

maltenmuller:
To do with the new telescope syntax and also #840, as far as I know.


MatthewDaggitt:
Thanks, looks good!

PR #890, opened 2019-09-12
Show functions for reflection things

fredrikNordvallForsberg:
I found having some show functions around very useful for debugging when playing around with tactics, so thought I would share my utility.

There's plenty of design decisions to discuss, I'm sure.

### DIFF
--- a/src/Data/String.agda
+++ b/src/Data/String.agda
@@ -22,3 +22,9 @@ open import Data.String.Properties using (_≈?_; _≟_; _<?_; _==_) public
 
 toVec : (s : String) → Vec Char (length s)
 toVec s = Vec.fromList (toList s)
+
+parens : String → String
+parens s = "(" ++ s ++ ")"
+
+braces : String → String
+braces s = "{" ++ s ++ "}"
--- a/src/Data/String.agda
+++ b/src/Data/String.agda
@@ -22,9 +22,3 @@ open import Data.String.Properties using (_≈?_; _≟_; _<?_; _==_) public
 
 toVec : (s : String) → Vec Char (length s)
 toVec s = Vec.fromList (toList s)
-
-parens : String → String
-parens s = "(" ++ s ++ ")"
-
-braces : String → String
-braces s = "{" ++ s ++ "}"
--- a/src/Reflection/Definition.agda
+++ b/src/Reflection/Definition.agda
@@ -8,12 +8,12 @@
 
 module Reflection.Definition where
 
-open import Data.List using (map; intersperse)
+open import Data.List using (map)
 import Data.List.Properties as Listₚ
 import Data.Nat.Properties as ℕₚ
 import Data.Nat.Show as NatShow
 open import Data.Product using (_×_; <_,_>; uncurry)
-open import Data.String as String using (String; braces; parens; _++_)
+open import Data.String as String using (String; _<+>_; intersperse; braces)
 open import Function using (_∘′_)
 open import Relation.Nullary
 import Relation.Nullary.Decidable as Dec
@@ -42,16 +42,14 @@ open import Agda.Builtin.Reflection public
 -- Showing
 
 show : Definition → String
-show (function cs) = "function " ++ braces (Term.showClauses cs)
+show (function cs)       = "function" <+> braces (Term.showClauses cs)
 show (data-type pars cs) =
- "datatype " ++ NatShow.show pars ++ " " ++
- braces (String.concat (intersperse ", " (map Name.show cs)))
-show (record′ c fs) =
- "record " ++ Name.show c ++ " " ++
- braces (String.concat (intersperse ", " (map (Name.show ∘′ Arg.unArg) fs)))
-show (constructor′ d) = "constructor " ++ Name.show d
-show axiom = "axiom"
-show primitive′ = "primitive"
+ "datatype" <+> NatShow.show pars <+> braces (intersperse ", " (map Name.show cs))
+show (record′ c fs)      =
+ "record" <+> Name.show c <+> braces (intersperse ", " (map (Name.show ∘′ Arg.unArg) fs))
+show (constructor′ d)    = "constructor" <+> Name.show d
+show axiom               = "axiom"
+show primitive′          = "primitive"
 
 ------------------------------------------------------------------------
 -- Decidable equality
--- a/src/Reflection/Definition.agda
+++ b/src/Reflection/Definition.agda
@@ -8,9 +8,13 @@
 
 module Reflection.Definition where
 
+open import Data.List using (map; intersperse)
 import Data.List.Properties as Listₚ
 import Data.Nat.Properties as ℕₚ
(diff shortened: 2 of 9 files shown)


src/Data/String.agda
```diff
@@ -22,3 +22,9 @@ open import Data.String.Properties using (_≈?_; _≟_; _<?_; _==_) public
 
 toVec : (s : String) → Vec Char (length s)
 toVec s = Vec.fromList (toList s)
+
```
gallais: Can we have a more general `between : String → String → String → String` too?

gallais: Also, this should probably go in `Base`.

src/Reflection/Definition.agda
```diff
@@ -34,6 +38,21 @@ open import Agda.Builtin.Reflection public
...
+-- Showing
+
+show : Definition → String
+show (function cs) = "function " ++ braces (Term.showClauses cs)
+show (data-type pars cs) =
+ "datatype " ++ NatShow.show pars ++ " " ++
+ braces (String.concat (intersperse ", " (map Name.show cs)))
+show (record′ c fs) =
+ "record " ++ Name.show c ++ " " ++
+ braces (String.concat (intersperse ", " (map (Name.show ∘′ Arg.unArg) fs)))
```
gallais: Please add `intersperse : String → List String → String` to `Data.String.Base`.

src/Reflection/Definition.agda
```diff
@@ -34,6 +38,21 @@ open import Agda.Builtin.Reflection public
...
+show (constructor′ d) = "constructor " ++ Name.show d
```
gallais: We could define `_<+>_ : String → String → String` as the space-introducing append
(similar to what we use in the Agda codebase).

src/Reflection/Pattern.agda
```diff
@@ -25,9 +25,41 @@ open import Reflection.Argument
...
+
+  showPatterns : List (Arg Pattern) → String
+  showPatterns [] = ""
+  showPatterns (a ∷ ps) = showArg a ++ showPatterns ps
+    where
+      showRel : Builtin.Relevance → String
+      showRel Builtin.relevant = ""
+      showRel Builtin.irrelevant = "."
+
+      showArg : Arg Pattern → String
```
gallais: These should probably go in the appropriate `Reflection.Argument(.Relevance/Visibility)`.

fredrikNordvallForsberg: There are more verbose show functions in `.Relevance` and `.Visibility`, here I wanted something more lightweight. I put them in a where-clause since they are not very useful on their own.

gallais: Ah, I see.

src/Data/String.agda
MatthewDaggitt: Is `between` really a useful specialisation of `xs ++ ys ++ zs` though? I'm not 100% sold on the idea.

src/Reflection/Pattern.agda
```diff
@@ -25,9 +25,41 @@ open import Reflection.Argument
...
+------------------------------------------------------------------------
+-- Showing
+
+mutual
+
+  showPatterns : List (Arg Pattern) → String
+  showPatterns [] = ""
+  showPatterns (a ∷ ps) = showArg a ++ showPatterns ps
+    where
+      showRel : Builtin.Relevance → String
```
MatthewDaggitt: Can we get rid of the `Builtin.` here? Not sure they're adding to the clarity...

src/Reflection/Term.agda
```diff
@@ -49,6 +54,66 @@ pattern vΠ[_∶_]_ s a ty    = Π[ s ∶ (vArg a) ] ty
...
+-- messages.
+
+private
+  open ListMembership CP._≟_
+
+  -- enclose string with parens if it contains a space character
+  parensIfSpace : String -> String
+  parensIfSpace s with ' ' ∈? String.toList s
+  ...             | yes _ = parens s
+  ...             | no _  = s
```
MatthewDaggitt: Can we get rid of the spaces between the `...` and the `|`. Sorry I know it seems minor. Really looking forward to the day when someone keen comes up with a linter for Agda!

src/Reflection/Term.agda
```diff
@@ -49,6 +54,66 @@ pattern vΠ[_∶_]_ s a ty    = Π[ s ∶ (vArg a) ] ty
...
+  showTerms (arg i t ∷ ts) = visibilityParen (visibility i) (show t) ++ " " ++ showTerms ts
+
+  show : Term → String
+  show (var x []) = "var " ++ NatShow.show x
+  show (var x args) = "var " ++ NatShow.show x ++ " " ++ showTerms args
+  show (con c []) = Name.show c
+  show (con c args) = Name.show c ++ " " ++ showTerms args
+  show (def f []) = Name.show f
+  show (def f args) = Name.show f ++ " " ++ showTerms args
+  show (lam v (abs s x)) = "lam " ++ visibilityParen v s ++ " → " ++ show x
```
MatthewDaggitt: Is there a reason you're not using unicode here? Just interested!

src/Reflection/Term.agda
```diff
@@ -49,6 +54,66 @@ pattern vΠ[_∶_]_ s a ty    = Π[ s ∶ (vArg a) ] ty
...
+
+mutual
+
+  showTerms : List (Arg Term) → String
+  showTerms [] = ""
+  showTerms (arg i t ∷ []) = visibilityParen (visibility i) (show t)
+  showTerms (arg i t ∷ ts) = visibilityParen (visibility i) (show t) ++ " " ++ showTerms ts
+
+  show : Term → String
+  show (var x []) = "var " ++ NatShow.show x
```
MatthewDaggitt: Sorry I know it's tedious but I think this function in particular does become much clearer to read if you align the `=` signs.

src/Reflection/Definition.agda
fredrikNordvallForsberg: Done! Also noted that `unlines = intersperse "\n"`

src/Reflection/Pattern.agda
fredrikNordvallForsberg: Good point.

src/Reflection/Term.agda
fredrikNordvallForsberg: Of course, thanks for flagging it up!

src/Reflection/Term.agda
fredrikNordvallForsberg: Actually I think unicode resulting in a shorter string is a win, so I'll change it.

src/Reflection/Term.agda
fredrikNordvallForsberg: Agreed!

fredrikNordvallForsberg:
Note that I removed the specific changelog entry for the show functions, since the modules affected are introduced as new further up in the changelog anyway.

PR #895, opened 2019-09-13
[ List.All ] points-to relation and props of updateAt modelled after Vec

andreasabel:
Made updateAt of List.All analogous to updateAt of Data.Vec.

### DIFF
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -146,7 +146,7 @@ module _ {_~_ : REL (List A) B ℓ} where
 --
 -- pxs [ i ]= px  if and only if  lookup pxs i = px.
 
-module _ {a p} {A : Set a} {P : A → Set p} where
+module _ {P : A → Set p} where
 
   -- `i` points to `lookup pxs i` in `pxs`.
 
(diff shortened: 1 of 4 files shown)


src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -115,68 +141,196 @@ module _ {_~_ : REL (List A) B ℓ} where
     (x~y ∷ (All.map All.head pxss)) ∷
     All-swap (x~ys ∷ (All.map All.tail pxss))
 
+------------------------------------------------------------------------
+-- Defining properties of lookup and _[_]=_
+--
+-- pxs [ i ]= px  if and only if  lookup pxs i = px.
+
+module _ {a p} {A : Set a} {P : A → Set p} where
+
```
gallais: We have `variable`s for `a` and `A` so we can drop some of these parameters.

MatthewDaggitt: and `p`

MatthewDaggitt:
Looks good to me.

MatthewDaggitt:
@andreasabel just thought I should mention that unless there's an urgent reason to merge it in quickly or the change is purely cosmetic, we usually leave PRs up for a week or so to give everyone the chance to provide feedback. Stops people from having to log on everyday! No worries in this case but just a general heads-up for future PRs :+1: Actually I should probably write this down somewhere :smile:

andreasabel:
Ok, thanks for letting me know, shall observe this in the future.

Issue #876, opened 2019-09-01
Cleaning up the various hierarchies


MatthewDaggitt:
Having just re-defined the function hierarchy, and as I'm just about to contribute a hierarchy for metric spaces under `Function.Metric`, I'm keen to standardise the files involved in each hierarchy. Currently we have:
- Functions
  - `Function.Core`
  - `Function.Definitions`
  - `Function.Structures`
  - `Function.Packages`
  - `Function` (empty, but re-exports all of the above)
- Algebra
  - `Algebra.FunctionProperties.Core`
  - `Algebra.FunctionProperties` (re-exports the above)
  - `Algebra.Structures`
  - `Algebra`
- Binary relations
  - `Relation.Binary.Core`
  - `Relation.Binary` (full, and re-exports all of the above)

To straighten this out I propose that every hierarchy `X` should have the following files
 - `X.Core` - contains very basic definitions, e.g. `Op`, `Op₂`, `Rel`, `REL`, `Fun` as well as basic combinators if applicable (e.g. function composition)
 - `X.Definitions` - contains named properties, e.g. `Transitive`, `Commutative`, `Injective`
 - `X.Structures` - self-explanatory
 - `X.Packages` - self-explanatory
 - `X` which publicly re-exports all of the above, parameterised by equality relations if applicable.

As the proposed top-level `X` module re-exports the contents of the previous modules, this should be doable in an entirely backwards compatible manner.

The only reservation I have is perhaps the name `Definitions`. I'd definitely prefer `Properties` but that name is already reserved for a certain class of files.  Anyone have any thoughts?

JacquesCarette:
Is `Packages` self-explanatory?  I'm not quite sure what is in there -- and in my usage, I've sometimes had conflicts between the version in `X` and `X.Packages` (I'll report a bug when it occurs again).

In any case, about `Definitions` vs `Properties`. Here a little bit of theory might help: when extending a theory (of `X`) with ground terms, i.e. definable in the background type theory, this creates a "conservative extension" of that theory. It doesn't matter if the things being defined turn out to be 'Properties' or anything else.  So I think that `Definitions` is a better name.

Furthermore, when I see `X.Properties`, I think that this is going to contain proofs of properties that `X` has, not definitions of properties of `X`.

In other words, I prefer the current naming scheme.

mechvel:
> I'm just about to contribute a hierarchy for metric spaces under Function.Metric, 
> []

Metric space theory for standard library?

[...]

PR #897, opened 2019-09-14
[ new ] add definitions of module-like structures

laMudri:
This pull request introduces modules over rings. In detail, we have:

 - `LeftSemimodule` and `RightSemimodule` over semirings
 - `Semimodule` over commutative semirings
 - `LeftModule` and `RightModule` over rings
 - `Module` over commutative rings

Modules are all really R-modules. For example, the parameters of `Module` are the following.

```
Module {r ℓr} (commutativeRing : CommutativeRing r ℓr) m ℓm
```

As far as I understand, this best fits mathematical practice. It also means that `IsModule` depends on `Algebra`, motivating the new modules. The added modules are:

 - `Algebra.Module`, corresponding to `Algebra`
 - `Algebra.Structures.Module`, corresponding to `Algebra.Structures`
 - `Algebra.FunctionProperties.Module.Core`, corresponding to `Algebra.FunctionProperties.Core`
 - `Algebra.FunctionProperties.Module.(Left|Right)`, corresponding to `Algebra.FunctionProperties`

The function properties were named based on what they look like, rather than names more meaningful for linear algebra. This produces names that are familiar from `Algebra.FunctionProperties`, and hopefully are beginner-friendly.

The module exports are very complicated, so there may be some bugs in them.


src/Algebra/Module.agda
```diff
@@ -0,0 +1,283 @@
...
+      ; *ᵣ-cong = flip *ₗ-cong
+      ; *ᵣ-zeroʳ = *ₗ-zeroˡ
+      ; *ᵣ-distribˡ = λ m x y → *ₗ-distribʳ m x y
+      ; *ᵣ-identityʳ = λ m → *ₗ-identityˡ m
+      ; *ᵣ-assoc = λ m x y → M-trans (*ₗ-comm y x m) (M-sym (*ₗ-assoc x y m))
+      ; *ᵣ-zeroˡ = *ₗ-zeroʳ
+      ; *ᵣ-distribʳ = λ x m n → *ₗ-distribˡ x m n
+      }
+    }
+
```
gallais: If you start with `isRightSemimodule : IsRightSemimodule (...)` you should then
be able to define `rightSemimodule = record { isRightSemimodule = isRightSemimodule }`.

src/Algebra/Module.agda
```diff
@@ -0,0 +1,283 @@
...
+  open IsModule isModule public
+
+  semimodule : Semimodule commutativeSemiring m ℓm
+  semimodule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; 0ᴹ = 0ᴹ
+    ; isSemimodule = isSemimodule
```
gallais: No need to spell out `_≈ᴹ_ = _≈ᴹ_; _+ᴹ_ = _+ᴹ_; _*ₗ_ = _*ₗ_; 0ᴹ = 0ᴹ` here:
Agda can figure them out by unification.

src/Algebra/Module.agda
```diff
@@ -0,0 +1,283 @@
...
+  leftModule = record { isLeftModule = isLeftModule }
+
+  open LeftModule leftModule public
+    using (+ᴹ-abelianGroup; +ᴹ-group)
+
+  isRightModule : IsRightModule ring _+ᴹ_ _*ᵣ_ 0ᴹ -ᴹ_
+  isRightModule = record
+    { isRightSemimodule = record
+      { +ᴹ-isCommutativeMonoid = +ᴹ-isCommutativeMonoid
+      ; *ᵣ-cong = *ᵣ-cong
```
gallais: Same here: probably worth lifting the `isRightSemimodule` proof to the toplevel.

src/Algebra/FunctionProperties/Module/Core.agda
```diff
@@ -0,0 +1,25 @@
...
+open import Level
+
+------------------------------------------------------------------------
+-- Binary operations
+
+Opₗ : ∀ {a b} → Set a → Set b → Set (a ⊔ b)
+Opₗ A B = A → B → B
+
+Opᵣ : ∀ {a b} → Set a → Set b → Set (a ⊔ b)
+Opᵣ A B = B → A → B
```
JacquesCarette: If we really want these type abbreviations, they should probably go with the other `Op`. Because, for example, `foldr` and `foldl` in `Data.List` could make use of them.

src/Algebra/FunctionProperties/Module/Left.agda
```diff
@@ -0,0 +1,55 @@
...
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Relation.Binary
+open import Data.Sum
+
+-- The properties are parameterised by the following "equality"
+-- relations
+
+module Algebra.FunctionProperties.Module.Left
```
JacquesCarette: Everything in here feels more general than for `Module`. [This is good!] Thus perhaps should be put somewhere more findable?

laMudri: Yes, I had difficulty naming this file, and I would be happy to make it more generally named. My other working name was `Double` (in reference to having two carrier sets), though I don't think this sounds great.

src/Algebra/FunctionProperties/Module/Core.agda
laMudri: This was also suggested by @wenkokke, and I guess I don't see why not, so I'll make this change.

src/Algebra/FunctionProperties/Module/Left.agda
wenkokke: > Everything in here feels more general than for `Module`. [This is good!] Thus perhaps should be put somewhere more findable?

Maybe we should keep these for now, and come up with a good name once we have other structures which could share the namespace? Unless you have any good examples off the top of your head?

wenkokke: Otherwise maybe try `Algebra.FunctionProperties.Left`?

src/Algebra/Module/Properties.agda
```diff
@@ -0,0 +1,97 @@
...
+
+open import Algebra
+open import Algebra.Module
+open import Level
+
+private
+  variable
+    c ℓ : Level
+
+self-leftSemimodule : (R : Semiring c ℓ) → LeftSemimodule R c ℓ
```
wenkokke: Why not name these `semiring⇒leftSemimodule`? The `self` seems confusing, and it implies that semirings are the only stuctures which construct trivial semimodules.

src/Algebra/FunctionProperties/Module/Left.agda
laMudri: I've gone for `Algebra.FunctionProperties.(Left|Right)Action`, which I'm fairly happy with.

src/Algebra/Module/Properties.agda
```diff
@@ -0,0 +1,97 @@
...
+      isLeftSemimodule
+    }
+  } where open CommutativeSemiring commutativeSemiring
+
+self-leftModule : (R : Ring c ℓ) → LeftModule R c ℓ
+self-leftModule ring = record
+  { -ᴹ_ = -_
+  ; isLeftModule = record
+    { isLeftSemimodule =
+      let open LeftSemimodule (self-leftSemimodule semiring) in
```
wenkokke: This inline let binding is rather horrifying... can't you move this to the where clause?

src/Algebra/Module/Properties.agda
```diff
@@ -0,0 +1,97 @@
...
+  variable
+    c ℓ : Level
+
+self-leftSemimodule : (R : Semiring c ℓ) → LeftSemimodule R c ℓ
+self-leftSemimodule semiring = record
+  { Carrierᴹ = Carrier
+  ; _≈ᴹ_ = _≈_
+  ; _+ᴹ_ = _+_
+  ; _*ₗ_ = _*_
+  ; 0ᴹ = 0#
```
wenkokke: Would it make sense to leave all these out, since Agda can infer them?

src/Algebra/Module.agda
```diff
@@ -0,0 +1,285 @@
...
+    _+ᴹ_ : Op₂ Carrierᴹ
+    _*ₗ_ : Opₗ Carrier Carrierᴹ
+    0ᴹ : Carrierᴹ
+    isLeftSemimodule : IsLeftSemimodule semiring _+ᴹ_ _*ₗ_ 0ᴹ
+
+  open IsLeftSemimodule isLeftSemimodule public
+
+  +ᴹ-commutativeMonoid : CommutativeMonoid m ℓm
+  +ᴹ-commutativeMonoid =
+    record { isCommutativeMonoid = +ᴹ-isCommutativeMonoid }
```
MatthewDaggitt: The record declarations should conform to the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md).

src/Algebra/Module.agda
```diff
@@ -0,0 +1,285 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definitions of algebraic structures defined over some other
+-- structure, like modules and vector spaces
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Module where
```
MatthewDaggitt: Given #876 can we rename this `Algebra.Module.Packages`? The `Packages` part of it may change before the next release, but at least we'll have it in the right place.

src/Algebra/Module.agda
```diff
@@ -0,0 +1,285 @@
...
+  open Semiring semiring
+
+  infixr 7 _*ₗ_
+  infixl 6 _+ᴹ_
+  infix 4 _≈ᴹ_
+
+  field
+    Carrierᴹ : Set m
+    _≈ᴹ_ : Rel Carrierᴹ ℓm
+  open Str _≈ᴹ_
```
MatthewDaggitt: I think it's clearer not to break the fields up by opening a parameterised `Str` here, and instead just open the `Structures` file globally at the top and provide the equality to the structure definitions. It also reduce the number of lines.

src/Algebra/Module.agda
```diff
@@ -0,0 +1,285 @@
...
+
+  open LeftSemimodule leftSemimodule public
+    using ( +ᴹ-commutativeMonoid; +ᴹ-monoid; +ᴹ-semigroup; +ᴹ-magma
+          ; +ᴹ-rawMonoid; +ᴹ-rawMagma)
+
+  _*ᵣ_ : Opᵣ Carrier Carrierᴹ
+  _*ᵣ_ = flip _*ₗ_
+
+  *ₗ-comm : L.Commutative _*ₗ_
+  *ₗ-comm x y m =
```
MatthewDaggitt: I would usually say that these sort of proofs should go in `Algebra.Module.Consequences`, and then be imported. But of course `LeftAction` and `RightAction` aren't under `Algebra.Module`...

For the moment I'd put them under `Algebra.FunctionProperties.LeftAction.Consequences` and I'll sort them out later when I refactor `Algebra` as part of #876.

src/Algebra/Module.agda
```diff
@@ -0,0 +1,285 @@
...
+  *ᵣ-comm m x y = *ₗ-comm y x m
+
+  isRightSemimodule : IsRightSemimodule semiring _+ᴹ_ _*ᵣ_ 0ᴹ
+  isRightSemimodule = record
+    { +ᴹ-isCommutativeMonoid = +ᴹ-isCommutativeMonoid
+    ; *ᵣ-cong = flip *ₗ-cong
+    ; *ᵣ-zeroʳ = *ₗ-zeroˡ
+    ; *ᵣ-distribˡ = λ m x y → *ₗ-distribʳ m x y
+    ; *ᵣ-identityʳ = λ m → *ₗ-identityˡ m
+    ; *ᵣ-assoc = λ m x y → M-trans (*ₗ-comm y x m) (M-sym (*ₗ-assoc x y m))
```
MatthewDaggitt: This one is complicated enough that it should go in `Algebra.Module.Consequences`

src/Algebra/Structures/Module.agda
```diff
@@ -0,0 +1,166 @@
...
+-- The Agda standard library
+--
+-- Some algebraic structures defined over some other structure
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary using (Rel; Setoid; IsEquivalence)
+
+module Algebra.Structures.Module
```
MatthewDaggitt: This should be `Algebra.Module.Structures` I think.

src/Algebra/Module/Properties.agda
```diff
@@ -0,0 +1,98 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definitions of algebraic structures defined over some other
+-- structure, like modules and vector spaces
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Module.Properties where
```
MatthewDaggitt: Hmm hierarchies don't usually have `.Properties` file but instead have a `Properties.S` for every structure `S` (which are usually parameterised by an `S`).

For example `semiring⇒leftSemimodule` should live in `Algebra.Properties.Semiring` and `ring⇒rightModule` should live in `Algebra.Properties.Ring`

MatthewDaggitt:
I wasn't going to mention this is as the current definitions match the standard mathematical theory quite well, but having just read #898 I've realised that this should definitely be discussed.

This is the first time that we have "structures" taking in a "package" as a parameter. Is this something we want to allow? Although it matches the standard mathematical formalisation, it does introduce various problems with respect to the hierarchy. 
1. As far as I can tell the issue in #898 is caused by inheritance in the structure hierarchy not being replicated in the package hierarchy (for obvious reasons). 
2. There's also the standard problem that as not all the fields from the package are exposed, it prevents you from expressing some relationships between modules. For example, you can't easily specify that two modules share all the same operations but the two underlying equality relations differ.

I'm a little worried that there might be more problems lurking out there as well. Whenever we deviate from the `Structures/Packages` pattern problems seem to creep in (e.g. the problems we've been having morphisms #853).

Although a little more verbose, I might be tempted to go with definitions in the style:
```
  record IsLeftSemimodule (+ * : Op₂ A) (0# 1# : A) (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (0ᴹ : M)  : Set (r ⊔ m ⊔ ℓr ⊔ ℓm) where
    open L _≈_ _≈ᴹ_
    field
      +-*-isSemiring : IsSemiring + * 0# 1#
      +ᴹ-isCommutativeMonoid : IsCommutativeMonoid _≈ᴹ_ +ᴹ 0ᴹ
      *ₗ-cong : Congruent *ₗ
      *ₗ-zeroˡ : LeftZero 0# 0ᴹ *ₗ
      *ₗ-distribʳ : *ₗ DistributesOverʳ _+_ ⟶ +ᴹ
      *ₗ-identityˡ : LeftIdentity 1# *ₗ
      *ₗ-assoc : Associative _*_ *ₗ
      *ₗ-zeroʳ : RightZero 0ᴹ *ₗ
      *ₗ-distribˡ : *ₗ DistributesOverˡ +ᴹ
```
which would circumvent both of these problems and maintain consistency. What are people's thoughts?

src/Algebra/FunctionProperties/RightAction.agda
```diff
@@ -0,0 +1,54 @@
...
+
+open import Level
+open import Relation.Binary
+open import Data.Sum
+
+-- The properties are parameterised by the following "equality"
+-- relations
+
+module Algebra.FunctionProperties.RightAction
+  {a b ℓa ℓb} {A : Set a} {B : Set b} (_≈ᴬ_ : Rel A ℓa) (_≈ᴮ_ : Rel B ℓb)
```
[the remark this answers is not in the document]
MatthewDaggitt: Of course if we @gallais and I win the argument in #4057 then we can add it to it's own separate module.

laMudri:
An odd thing is that then an R-module contains a proof that the operations of R form a ring – duplicating what is already contained in R. But still, this might be the way to go.

laMudri:
@MatthewDaggitt Would it work to give `IsLeftSemimodule` a parameter of type `IsSemiring`, rather than `Semiring`?

src/Algebra/Module/Properties.agda
laMudri: Do you think `Algebra.Properties.LeftSemimodule` &c is the way to go? Also, would this imply simplifying the names of lemmas? E.g, `semiring⇒leftSemimodule` becomes `fromSemiring`, or something like this.

wenkokke: This seems like putting a LOT in the module hierarchy, though: {left, right, neuter} * {semi, full}... that's a lot of modules.

laMudri: Yeah, this was also part of the reasoning. All of these lemmas are pretty similar, so it seems easier to keep them synchronised if they are together. Maybe we should think about how large this file is likely to get.

ajrouvoet:
@MatthewDaggitt I've been fighting with Agda to get a nice algebra hierarchy myself lately. The heavily parameterized ones are definitely the one to go for now, but are tedious to work with because you need a ton of quantification everywhere.

I wonder if Intellij's Arend gets it right, which claims to implement a neat trick where parameters and fields of records are actually one and the same thing: https://arend-lang.github.io/about/arend-features#anonymous-extensions

jespercockx:
Sounds like the "frame type theory" by Cyril Cohen, Assia Mahboubi and Xavier Montillet (types 2019) would be useful here.

MatthewDaggitt:
> An odd thing is that then an R-module contains a proof that the operations of R form a ring – duplicating what is already contained in R. But still, this might be the way to go.

So the way I've avoided this in the `Function` hierarchy is not to use the `Is` modules internally but to simply re-list the properties. Obviously this quite a lot more effort here... Otherwise I'm not **too** fussed about the duplication. If you have the proof for `Ring`, it's easy to get `IsRing` out of it. It's a little inconvenient but it doesn't rule out any functionality like the current solution.

> Should I introduce `semiring⇒isLeftSemimodule : (R : Semiring c ℓ) → IsLeftSemimodule _≈_ R _+_ _*_ 0# 1#`? Because of the requirement of a full (commutative) (semi)ring when forming `Is(Left|Right|)(Semi)module`, it's not quite so obvious what to do.

`semiring⇒isLeftSemimodule` sounds good, though I have to confess I'm not quite sure what the problem is? Is it the issue described in #898?

MatthewDaggitt:
> @ajrouvoet I've been fighting with Agda to get a nice algebra hierarchy myself lately. The heavily parameterized ones are definitely the one to go for now, but are tedious to work with because you need a ton of quantification everywhere.

Yes, it's definitely preferable to use packages when you can get away with it, but lots of quantification does seem to be required when you want dependencies between different structures.

> @ajrouvoet  I wonder if Intellij's Arend gets it right, which claims to implement a neat trick where parameters and fields of records are actually one and the same thing: https://arend-lang.github.io/about/arend-features#anonymous-extensions

That sounds really neat! Seems almost too neat to be honest, there must a trade-off somewhere surely? :)

> @jespercockx  Sounds like the "frame type theory" by Cyril Cohen, Assia Mahboubi and Xavier Montillet (types 2019) would be useful here.

I'm afraid I've temporarily fallen out of academia (working on reversing it!) and don't have journal access at the moment, so I'll have to take your word on that :+1:

jespercockx:
https://www.ii.uib.no/~bezem/abstracts/TYPES_2019_paper_51

MatthewDaggitt:
@laMudri I think I'm going to bump this to `v1.3` if that's okay with you? At some point I'll try and find the time to go through it and find a set of module parameters that work better.

JacquesCarette:
It's definitely on my list of things to look at. I should be able to do it tomorrow or Friday.

laMudri:
The definitions of (semi)module are now #985-compliant. An R-module is an R-R-bimodule, defined only when R is commutative. This gives the option to specify both the left scaling and the right scaling operators. There are also helper modules `IsSemimoduleFromLeft` and `IsSemimoduleFromRight` to allow specifying just one of them (it looks like I forgot to do the `Module` versions; will do shortly...).

laMudri:
Oh, I forgot to add that one of the more surprising modifications to incorporate bimodules was the definitions `IsPreleftSemimodule`, `IsPrerightSemimodule`, &c. The idea is that giving both `IsLeftSemimodule` and `IsRightSemimodule` in order to prove `IsBisemimodule` would mean giving twice a proof of `+ᴹ-isCommutativeMonoid`. Instead, `IsPreleftSemiring` and `IsPrerightSemiring` omit this field, and `IsLeftSemimodule` and `IsBisemimodule` look as follows:

```agda
  record IsLeftSemimodule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (0ᴹ : M) : Set _ where
    field
      +ᴹ-isCommutativeMonoid : IsCommutativeMonoid _≈ᴹ_ +ᴹ 0ᴹ
      isPreleftSemimodule : IsPreleftSemimodule +ᴹ *ₗ 0ᴹ

  record IsBisemimodule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (*ᵣ : Opᵣ S M) (0ᴹ : M) : Set _ where
    field
      +ᴹ-isCommutativeMonoid : IsCommutativeMonoid _≈ᴹ_ +ᴹ 0ᴹ
      isPreleftSemimodule : IsPreleftSemimodule R-semiring +ᴹ *ₗ 0ᴹ
      isPrerightSemimodule : IsPrerightSemimodule S-semiring +ᴹ *ᵣ 0ᴹ
      *ₗ-*ᵣ-assoc : ∀ r m s → *ᵣ (*ₗ r m) s ≈ᴹ *ₗ r (*ᵣ m s)
```

These definitions basically abstract over duplicated code (in lieu of being able to abstract over pure rows); not much else.

JacquesCarette:
Very nice development. It is also standard mathematical practice to have things like `Module` be parametric over a `CommutativeRing`.  I think in module-theory is where the limits of listing every operator in the header of the `IsFoo` starts to fall apart (this isn't a criticism of this library, this is a criticism of the current Agda library methodology, but is a bit pointless since there are no good alternatives yet).  What is needed are 'record definition combinators' that allow you to bundle/unbundle at will. [See the PEPM 2020 paper].

The point is that there are a whole bunch of isomorphic ways of defining the same thing in Agda, and they have different usability profiles in different contexts. So it's not like one is better than any of the others, it's that they are each *relatively* superior.  See points 7-10 of [Tom Hales' critique of Lean](https://jiggerwit.wordpress.com/2018/09/18/a-review-of-the-lean-theorem-prover/) for many things which apply to Agda too.

Bottom line: in current Agda, I think what you've done here is the best you can do, and should be merged in. What will be needed is for someone on the core Agda team to get interested in this kind of library development and see that things have gotten unwieldly but could be fixed.  (Hopefully @alhassy and @WolframKahl are paying attention too).

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,398 @@
...
+import Algebra.FunctionProperties as FP
+import Algebra.FunctionProperties.Consequences as Consequences
+import Algebra.FunctionProperties.LeftAction as L
+import Algebra.FunctionProperties.RightAction as R
+open import Algebra.Structures
+open import Data.Product using (_,_; proj₁; proj₂)
+open import Function.Base using (flip)
+open import Level using (_⊔_)
+
+module _ {r ℓr} (semiring : Semiring r ℓr) where
```
MatthewDaggitt: Can we use `variable` to get rid of some of these implicits?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,398 @@
...
+        { isMonoid = +ᴹ-isMonoid
+        ; inverse = -ᴹ‿inverse
+        ; ⁻¹-cong = -ᴹ‿cong
+        }
+      ; comm = +ᴹ-comm
+      }
+
+    open IsAbelianGroup +ᴹ-isAbelianGroup public
+      using ()
+      renaming ( isGroup to +ᴹ-isGroup; inverseˡ to -ᴹ‿inverseˡ
```
MatthewDaggitt: I've tended to go for the style
```agda
  using () renaming
  ( ... to ...
  ; ... to ...
  )
```
as it doesn't end up wasting a load of whitespace.

src/Algebra/Module/Bundles.agda
```diff
@@ -0,0 +1,309 @@
...
+  open Bisemimodule bisemimodule public
+    using ( leftSemimodule; rightSemimodule
+          ; +ᴹ-commutativeMonoid; +ᴹ-monoid; +ᴹ-semigroup; +ᴹ-magma; +ᴹ-rawMagma
+          ; +ᴹ-rawMonoid)
+
+  *ₗ-comm : L.Commutative _*ₗ_
+  *ₗ-comm x y m =
+    M-trans (M-sym (*ₗ-assoc x y m))
+            (M-trans (*ₗ-cong (*-comm _ _) M-refl)
+                     (*ₗ-assoc y x m))
```
MatthewDaggitt: Equational reasoning would make this and the proof below much more readable?

src/Algebra/FunctionProperties/LeftAction.agda
```diff
@@ -0,0 +1,53 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary
+
+-- The properties are parameterised by the two carriers and
+-- the result equality.
+
+module Algebra.FunctionProperties.LeftAction
```
MatthewDaggitt: The `FunctionProperties` subfolder is being deprecated. We could try either `Algebra.Module.Definitions.LeftAction` or `Algebra.Definitions.LeftAction`?

src/Algebra/FunctionProperties/LeftAction.agda
```diff
@@ -0,0 +1,53 @@
...
+  {a b ℓb} (A : Set a) {B : Set b} (_≈_ : Rel B ℓb)
```
MatthewDaggitt: When is the `A` not inferable?

src/Algebra/Module/Properties.agda
MatthewDaggitt: > This seems like putting a LOT in the module hierarchy, though: {left, right, neuter} * {semi, full}... that's a lot of modules.

@wenkokke you're right, but it is the way the library is currently laid out (see `Algebra.Properties`). I think best to try and stay consistent. 

> Do you think Algebra.Properties.LeftSemimodule &c is the way to go? Also, would this imply simplifying the names of lemmas? E.g, semiring⇒leftSemimodule becomes fromSemiring, or something like this.

Yup sounds good to me.

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,398 @@
...
+          ; *ᵣ-identityʳ = *ᵣ-identityʳ
+          ; *ᵣ-assoc = *ᵣ-assoc
+          ; *ᵣ-zeroˡ = *ᵣ-zeroˡ
+          ; *ᵣ-distribʳ = *ᵣ-distribʳ
+          }
+        ; *ₗ-*ᵣ-assoc = *ₗ-*ᵣ-assoc
+        }
+      }
+
+  record IsModuleFromLeft (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (0ᴹ : M) (-ᴹ : Op₁ M)
```
MatthewDaggitt: In #985 I think I suggested that these might go in a separate module?

src/Algebra/Module/Structures.agda
laMudri: Do you do anything different for long `using` lists?

laMudri:
The Christmas break plus a deadline totally distracted me from this, but I should have a little bit of time now.

For properties, I'm wondering what would be the best naming scheme. We currently have names like `semiring⇒leftSemimodule`, which with the addition of more namespaces we're wanting to turn into `fromSemiring`. But this name is not so descriptive, and is basically a duplicate of the type (which is also not that descriptive). These constructions are doing a specific thing, which is constructing the tensor unit in R-semimodules and R-modules. In the non-commutative cases, they are the identity 1-cells of the respective 2-categories of bimodules (where left R-modules are seen as R-⊤-bimodules, &c). I feel as if this is worth putting in the name, especially if we at some point add tensor products. In fact, because, like the tensor product, the tensor unit is a construction, it probably shouldn't be in a `Properties` module, but rather a `Construct` one.

src/Algebra/Module/Properties.agda
laMudri: The reason I haven't done this is because I think seeing these fields being filled in is useful. This definition is pretty subtle, and though it's all inferable by Agda, I don't think it's inferable by humans who don't already know the theory. Furthermore, Agda can only infer these fields after lots of little lemmas are filled in, which makes me believe that it's slightly brittle.

src/Algebra/Module/Bundles.agda
```diff
@@ -0,0 +1,318 @@
...
+  +ᴹ-abelianGroup = record { isAbelianGroup = +ᴹ-isAbelianGroup }
+
+  open AbelianGroup +ᴹ-abelianGroup public
+    using () renaming (group to +ᴹ-group)
+
+------------------------------------------------------------------------
+-- Bimodules
+------------------------------------------------------------------------
+
+record Bisemimodule {r s ℓr ℓs} (R-semiring : Semiring r ℓr)
```
MatthewDaggitt: We could use `variable` to generalise over `r s ℓr ℓs`?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+private
+  variable
+    r ℓr s ℓs : Level
+
+module _ (semiring : Semiring r ℓr) where
+  open Semiring semiring renaming (Carrier to R)
+
+  record IsPreleftSemimodule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (0ᴹ : M)
+                             : Set (r ⊔ m ⊔ ℓr ⊔ ℓm) where
+    open L R _≈ᴹ_
```
MatthewDaggitt: Does this really work? :open_mouth:

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+import Algebra.Module.Definitions.Right as R
+open import Algebra.Structures
+open import Data.Product using (_,_; proj₁; proj₂)
+open import Level using (Level; _⊔_)
+
+private
+  variable
+    r ℓr s ℓs : Level
+
+module _ (semiring : Semiring r ℓr) where
```
MatthewDaggitt: So the problem with this organisation of parameters is that the order goes: `_≈ᴹ_ semiring +ᴹ ...` which is a little bit weird... I know it loses the module parameter but I think it'd be much cleaner to have `semiring _≈ᴹ_ +ᴹ ...`?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+    open IsPreleftSemimodule isPreleftSemimodule public
+
+    open IsCommutativeMonoid +ᴹ-isCommutativeMonoid public
+      using () renaming
+      ( assoc                to +ᴹ-assoc
+      ; comm                 to +ᴹ-comm
+      ; identity             to +ᴹ-identity
+      ; identityʳ            to +ᴹ-identityʳ
+      ; identityˡ            to +ᴹ-identityˡ
+      ; isEquivalence        to M-isEquivalence
```
MatthewDaggitt: Given the relation is called `≈ᴹ` perhaps `≈ᴹ-isEquivalence` etc. would be more consistent with other renamings?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+  open Semiring R-semiring using () renaming (Carrier to R)
+  open Semiring S-semiring using () renaming (Carrier to S)
+
+  record IsBisemimodule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (*ᵣ : Opᵣ S M) (0ᴹ : M)
+                        : Set (r ⊔ s ⊔ m ⊔ ℓr ⊔ ℓs ⊔ ℓm) where
+    field
+      +ᴹ-isCommutativeMonoid : IsCommutativeMonoid _≈ᴹ_ +ᴹ 0ᴹ
+      isPreleftSemimodule : IsPreleftSemimodule R-semiring +ᴹ *ₗ 0ᴹ
+      isPrerightSemimodule : IsPrerightSemimodule S-semiring +ᴹ *ᵣ 0ᴹ
+      *ₗ-*ᵣ-assoc : ∀ r m s → *ᵣ (*ₗ r m) s ≈ᴹ *ₗ r (*ᵣ m s)
```
MatthewDaggitt: This should probably use a name out of `Definitions`?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+      hiding (+ᴹ-isCommutativeMonoid; isPreleftSemimodule)
+    open IsPrerightSemimodule isPrerightSemimodule public
+
+module _ (commutativeSemiring : CommutativeSemiring r ℓr) where
+  open CommutativeSemiring commutativeSemiring renaming (Carrier to R)
+
+  record IsSemimodule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (*ᵣ : Opᵣ R M) (0ᴹ : M)
+                      : Set (r ⊔ m ⊔ ℓr ⊔ ℓm) where
+    field
+      isBisemimodule : IsBisemimodule semiring semiring +ᴹ *ₗ *ᵣ 0ᴹ
```
MatthewDaggitt: This one is going to need a comment? Why is a semimodule the same thing as  bisemimodule? Oh because the underlying semiring is commutative?

src/Algebra/Module/Structures.agda
```diff
@@ -0,0 +1,262 @@
...
+    isRightModule = record
+      { isRightSemimodule = isRightSemimodule
+      ; -ᴹ‿cong = -ᴹ‿cong
+      ; -ᴹ‿inverse = -ᴹ‿inverse
+      }
+
+module _ (commutativeRing : CommutativeRing r ℓr) where
+  open CommutativeRing commutativeRing renaming (Carrier to R)
+
+  record IsModule (+ᴹ : Op₂ M) (*ₗ : Opₗ R M) (*ᵣ : Opᵣ R M) (0ᴹ : M)
```
MatthewDaggitt: Again another comment here would be useful!

src/Algebra/Module/Structures/Biased.agda
```diff
@@ -0,0 +1,122 @@
...
+    isBisemimodule = record
+      { +ᴹ-isCommutativeMonoid = +ᴹ-isCommutativeMonoid
+      ; isPreleftSemimodule = isPreleftSemimodule
+      ; isPrerightSemimodule = record
+        { *ᵣ-cong = flip *ₗ-cong
+        ; *ᵣ-zeroʳ = *ₗ-zeroˡ
+        ; *ᵣ-distribˡ = *ₗ-distribʳ
+        ; *ᵣ-identityʳ = *ₗ-identityˡ
+        ; *ᵣ-assoc = λ m r s → M-trans (M-sym (*ₗ-assoc s r m))
+                                       (*ₗ-cong (*-comm s r) M-refl)
```
MatthewDaggitt: These sort of proofs tend to go in `Algebra.Module.Consequences`

src/Algebra/Module/Structures/Biased.agda
```diff
@@ -0,0 +1,122 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Abbreviations of some of the module-like definitions
```
MatthewDaggitt: Not sure this comment is quite right?

src/Algebra/Module/Properties.agda
MatthewDaggitt: Hmm okay, new plan! Keep this module as is, but rename it to `Algebra.Module.Construct.Trivial`? As you suggested earlier @laMudri, we could then ditch the `semiring⇒` part of the `semiring⇒leftSemimodule` names?

Edit: ah, just seen that you suggested this as well. Great minds think alike :smile:

gallais:
Note that now that we are using 2.6.1 on master, you do not need to use
`\ where` to pattern-match on a pair: a usual lambda abstraction will do.

src/Algebra/Module/Consequences.agda
```diff
@@ -0,0 +1,86 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- TODO
+------------------------------------------------------------------------
```
gallais: todo

laMudri:
On my checklist:

- [ ] There is one `TODO`: moving a definition in `Agda.Module.Construct.Biproduct` into the main `Algebra` hierarchy. This would basically mean creating `Algebra.Construct.Biproduct`, and reasonably populating it.
- [ ] Tensor products are the last remaining standard construct. But they are hard, so I think they should be left for later. Perhaps I'll open an issue with a high-level description when this gets merged.

gallais:
:+1: for later. This is already a pretty big PR.

laMudri:
For the first todo, maybe this should have its own (simple, but possibly moderately large) pull request, partially to test out the `Construct` structure to a wider audience.

src/Algebra/Module/Construct/Zero.agda
```diff
@@ -0,0 +1,71 @@
...
+  ; _≈ᴹ_ = λ _ _ → ⊤
+  }
+
+bimodule : {R : Ring r ℓr} {S : Ring s ℓs} → Bimodule R S zero zero
+bimodule = record
+  { Carrierᴹ = ⊤
+  ; _≈ᴹ_ = λ _ _ → ⊤
+  }
+
+module′ : {R : CommutativeRing r ℓr} → Module R zero zero
```
MatthewDaggitt: The convention when we have names that are keywords is to use angle brackets, e.g. `⟨module⟩`

src/Algebra/Module/Construct/Zero.agda
```diff
@@ -0,0 +1,71 @@
...
+open import Algebra.Bundles
+open import Algebra.Module.Bundles
+open import Data.Unit
+open import Level
+
+private
+  variable
+    r s ℓr ℓs : Level
+
+leftSemimodule : {R : Semiring r ℓr} → LeftSemimodule R zero zero
```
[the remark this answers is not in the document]
laMudri: Ah, I didn't know about that synonym. Will do...

src/Algebra/Module/Construct/TensorUnit.agda
```diff
@@ -0,0 +1,117 @@
...
+
+open import Algebra.Bundles
+open import Algebra.Module.Bundles
+open import Level
+
+private
+  variable
+    c ℓ : Level
+
+leftSemimodule : {R : Semiring c ℓ} → LeftSemimodule R c ℓ
```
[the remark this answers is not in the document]
laMudri: I think we should expect to construct modules over known rings, meaning in this case that we're checking `leftSemimodule` against at least `LeftSemimodule R ? ?`. As an analogy, making `R` explicit here would be like making `A` explicit in the identity function. They serve the same role.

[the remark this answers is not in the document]
MatthewDaggitt: Okay fair enough I guess...

MatthewDaggitt:
Merged in. Thanks for everyone's contributions!

PR #899, opened 2019-09-14
[ Data.List ] function reverse-append _ʳ++_, simplify proofs about reverse

andreasabel:
The properties are much easier to prove for the generalization of
reverse to reverse-append.

Should be backwards compatible.

In the future, we could deprecate `reverseAcc` in favor of `_ʳ++_` (reverse-append).

### DIFF
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -773,21 +773,21 @@ module _ {a} {A : Set a} where
 
   -- Defining property of reverse-append _ʳ++_.
 
-  ʳ++=reverse++ : ∀ (xs : List A) {ys} → xs ʳ++ ys ≡ reverse xs ++ ys
-  ʳ++=reverse++ [] = refl
-  ʳ++=reverse++ (x ∷ xs) {ys} = begin
+  ʳ++-defn : ∀ (xs : List A) {ys} → xs ʳ++ ys ≡ reverse xs ++ ys
+  ʳ++-defn [] = refl
+  ʳ++-defn (x ∷ xs) {ys} = begin
     (x ∷ xs)             ʳ++ ys   ≡⟨⟩
     xs         ʳ++   x     ∷ ys   ≡⟨⟩
-    xs         ʳ++ [ x ]  ++ ys   ≡⟨ ʳ++=reverse++ xs  ⟩
+    xs         ʳ++ [ x ]  ++ ys   ≡⟨ ʳ++-defn xs  ⟩
     reverse xs  ++ [ x ]  ++ ys   ≡⟨ P.sym (++-assoc (reverse xs) _ _) ⟩
-    (reverse xs ++ [ x ]) ++ ys   ≡⟨ P.cong (_++ ys) (P.sym (ʳ++=reverse++ xs)) ⟩
+    (reverse xs ++ [ x ]) ++ ys   ≡⟨ P.cong (_++ ys) (P.sym (ʳ++-defn xs)) ⟩
     (xs ʳ++ [ x ])        ++ ys   ≡⟨⟩
     reverse (x ∷ xs)      ++ ys   ∎
 
   -- Corollary: reverse of cons is snoc of reverse.
 
   unfold-reverse : ∀ (x : A) xs → reverse (x ∷ xs) ≡ reverse xs ∷ʳ x
-  unfold-reverse x xs = ʳ++=reverse++ xs
+  unfold-reverse x xs = ʳ++-defn xs
 
   -- Reverse-append of append is reverse-append after reverse-append.
   -- Simple inductive proof.
@@ -808,7 +808,7 @@ module _ {a} {A : Set a} where
     reverse (xs ++ ys)         ≡⟨⟩
     (xs ++ ys) ʳ++ []          ≡⟨ ʳ++-++ xs ⟩
     ys ʳ++ xs ʳ++ []           ≡⟨⟩
-    ys ʳ++ reverse xs          ≡⟨ ʳ++=reverse++ ys ⟩
+    ys ʳ++ reverse xs          ≡⟨ ʳ++-defn ys ⟩
     reverse ys ++ reverse xs   ∎
 
   -- Reverse-append of reverse-append is commuted reverse-append after append.
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -764,84 +764,152 @@ module _ {P : A → Set p} (P? : Decidable P) where
   ...  | no ¬Px = P.cong (Prod.map id (x ∷_)) (partition-defn xs)
 
 ------------------------------------------------------------------------
--- reverse
+-- reverse is an involution in the list monoid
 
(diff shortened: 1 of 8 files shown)


src/Data/List/Relation/Binary/Equality/Setoid.agda
```diff
@@ -121,7 +121,12 @@ module _ {P : Pred A p} (P? : U.Decidable P) (resp : P Respects _≈_)
   filter⁺ xs≋ys = PW.filter⁺ P? P? resp (resp ∘ sym) xs≋ys
 
 ------------------------------------------------------------------------
--- filter
+-- reverse
+
+infixr 5 _ʳ++⁺_
+
+_ʳ++⁺_ : ∀{xs xs′ ys ys′} → xs ≋ xs′ → ys ≋ ys′ → xs ʳ++ ys ≋ xs′ ʳ++ ys′
```
MatthewDaggitt: The corresponding proof for `++` doesn't use infix notation so I'm not sure this one should either. If we use copy `++` in using `ws xs ys zs` as variables as well then it also fits within the line length!

src/Data/List/Relation/Binary/Pointwise.agda
```diff
@@ -243,6 +243,14 @@ module _ {R : REL A B ℓ} where
   reverseAcc⁺ rs′ []       = rs′
   reverseAcc⁺ rs′ (r ∷ rs) = reverseAcc⁺ (r ∷ rs′) rs
 
+  infixr 5 _ʳ++⁺_
+
+  _ʳ++⁺_ : ∀ {as bs as′ bs′} →
```
MatthewDaggitt: Again probably more consistent to avoid using the infix notation. (And in the various other binary relation modules)

src/Data/List/Properties.agda
```diff
@@ -764,84 +764,152 @@ module _ {P : A → Set p} (P? : Decidable P) where
   ...  | no ¬Px = P.cong (Prod.map id (x ∷_)) (partition-defn xs)
 
 ------------------------------------------------------------------------
--- reverse
+-- reverse is an involution in the list monoid
 
 module _ {a} {A : Set a} where
```
MatthewDaggitt: I know this isn't your code, but might as well remove this anonymous module while we're here :smile:

src/Data/List/Properties.agda
```diff
@@ -764,84 +764,152 @@ module _ {P : A → Set p} (P? : Decidable P) where
...
-                       reverse (xs ++ ys) ≡ reverse ys ++ reverse xs
-  reverse-++-commute []       ys = P.sym (++-identityʳ _)
-  reverse-++-commute (x ∷ xs) ys = begin
-    reverse (x ∷ xs ++ ys)               ≡⟨ unfold-reverse x (xs ++ ys) ⟩
-    reverse (xs ++ ys) ++ [ x ]          ≡⟨ P.cong (_++ [ x ]) (reverse-++-commute xs ys) ⟩
-    (reverse ys ++ reverse xs) ++ [ x ]  ≡⟨ ++-assoc (reverse ys) _ _ ⟩
-    reverse ys ++ (reverse xs ++ [ x ])  ≡⟨ P.sym $ P.cong (reverse ys ++_) (unfold-reverse x xs) ⟩
-    reverse ys ++ reverse (x ∷ xs)       ∎
-    where open P.≡-Reasoning
+  ʳ++=reverse++ : ∀ (xs : List A) {ys} → xs ʳ++ ys ≡ reverse xs ++ ys
```
MatthewDaggitt: I'm aware it's not quite as meaningful, but to be consistent with the naming conventions used in the rest of the file we should probably call this `ʳ++-defn`

andreasabel: Ok!

src/Data/List/Properties.agda
andreasabel: This module allows me to use `open` without polluting the namespace for code further done.  Thus, I think it is good to keep it!

src/Data/List/Relation/Binary/Equality/Setoid.agda
andreasabel: Isn't `++` off here?  I'd expect that if we have an infix operator like `++`, then its predicate/relation lifting is also an infix operator.  Examples of this is `Data.List.All._::_` which is the predicate lifting of `Data.List._::_`.

src/Data/List/Relation/Binary/Pointwise.agda
andreasabel: This one is the relation lifting of `_^r++_` to `Pointwise`.  I think infix makes sense here.

src/Data/List/Properties.agda
MatthewDaggitt: Fair enough :+1:

src/Data/List/Relation/Binary/Equality/Setoid.agda
MatthewDaggitt: > Isn't ++ off here?

Hmm well if it is off, it's off everywhere else in the library: 
- [Any](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/List/Relation/Unary/Any/Properties.agda#L331)
- [All](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/List/Relation/Unary/All/Properties.agda#L374)
- [Unique](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/List/Relation/Unary/Unique/Setoid/Properties.agda#L46)
- [Permutation](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda#L86)
- [Prefix](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/List/Relation/Binary/Prefix/Heterogeneous/Properties.agda#L72)
- [Pointwise](https://github.com/agda/agda-stdlib/blob/c82ed45d6b6b1322b2b1f55c52edd1a38ab56926/src/Data/Vec/Relation/Binary/Pointwise/Inductive.agda#L159) (for Vec)
- etc. (at least 5 more places)

I'd suggest if you do want to change it, then the best course of action is to make this PR consistent with the current form, and then open a new PR which changes it **everywhere** in the library.

src/Data/List/Relation/Binary/Pointwise.agda
MatthewDaggitt: See comment above.

src/Data/List/Relation/Binary/Equality/Setoid.agda
andreasabel: Seems fair for the sake of conservativity.

I guess this decision (infix or not) is also a topic for the style guide...

PR #901, opened 2019-09-17
[ new ] fromBool : Bool → A → A ⊎ A

gallais:


JacquesCarette:
This is really one of the directions from the proof of `Inverse (Bool x A) (A ⊎ A)`. And calling it `fromBool` is perhaps sub-optimal, seen in that light?

gallais:
I used the uncurried version together with `partitionSumsWith` so I would be happy
to implement both directions using `Product`. :D

MatthewDaggitt:
What's the status of this? Is this ready or does it need a change?

gallais:
No hurry for this one either, I ended up using something else in the end.

laMudri:
Wouldn't it make more sense to call this function `fromBool′`, and have one of the following for `fromBool`?

```agda
fromBool₀ : {A : Bool → Set a} (b : Bool) → A b → A true ⊎ A false
fromBool₁ : {A : Set a} {B : Set b} (b : Bool) → (if b then A else B) → A ⊎ B
```

MatthewDaggitt:
@gallais as you ended up using something else, I'm going to close this for now? Feel free to reopen and bring up to date if you'd like it in.

PR #909, opened 2019-09-22
Add `Linked` predicate over lists

### LINKED ISSUE #617: AllNeighbours: Predicate on lists
I've been using the following data type to describe a property applied to all neighbouring pairs of elements. Not sure if it may be useful to others or if there's a better way of writing it?

```agda
data AllNeighbours {p} (P : A → A → Set p) : L.List A → Set (a Level.⊔ p) where
    empty : AllNeighbours P L.[]
    single : (x : A) → AllNeighbours P (x L.∷ L.[])
    pair : {x y : A} → {zs : L.List A} → P x y → AllNeighbours P (y L.∷ zs) → AllNeighbours P (x L.∷ y L.∷ zs)
```

MatthewDaggitt:
Fixes #617. @EdNutting care to take a look? Feedback welcome!

I'm not quite happy with the name `AllNeighbours` as it's take a little too long to type. What do people think about `Linked`? Any other suggestions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,9 @@ The following new modules have been added to the library:
   Data.List.Kleene.AsList
   Data.List.Kleene.Base
 
+  Data.List.Relation.Unary.AllNeighbours
+  Data.List.Relation.Unary.AllNeighbours.Properties
+
   Data.List.Relation.Binary.Sublist.Propositional.Disjoint
 
   Data.Rational.Unnormalised
--- a/src/Data/List/Relation/Unary/AllNeighbours.agda
+++ b/src/Data/List/Relation/Unary/AllNeighbours.agda
@@ -1,0 +1,98 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Lists where every consecutative pair of elements is related.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Relation.Unary.AllNeighbours {a} {A : Set a} where
+
+open import Data.List using (List; []; _∷_)
+open import Data.List.Relation.Unary.All as All using (All; []; _∷_)
+open import Data.Product as Prod using (_,_; _×_)
+open import Function using (id; _∘_)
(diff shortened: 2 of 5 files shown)


gallais:
:+1 for linked.

JacquesCarette:
`Beside` ? `Linked` isn't bad, but it's harder to generalize. For any `Traversable` structure (in the Haskell sense), there is a natural notion of neighbours, from the flattening, and this predicate ought to be applicable (eventually) in that case too. One could even foresee an Optics-based `Traversable` generalization as well. `AllNeighbours` still makes sense then (conceptually and nominally) but `Linked` doesn't. That's my main reservation.

MatthewDaggitt:
Hmm `Beside` for me doesn't quite capture the notion that the elements are actually related. I'd argue that all neighbours in a list are "beside" one another, without necessarily having any further relationship between them.

>  One could even foresee an Optics-based Traversable generalization as well. AllNeighbours still makes sense then (conceptually and nominally) but Linked doesn't. That's my main reservation.

I'm not entirely convinced by this one either. If you have a `Traversable` binary tree for example, I wouldn't expect `AllNeighbours` to say that the neighbouring elements in the pre/in/post-order traversal are related. Instead I'd expect that a parent node would be related to all of its children.

MatthewDaggitt:
> I will however point out that your own arguments apply to AllNeighboors and Linked too: the notion is rather informal, and might clash with one's expectations of its instance on one data-structure or another.

I agree that all the names clash with the flattening of the data structures via a `Traversable` interface. I think that if we went with `Linked`, then `Linked` would be used to refer to the natural unflattened version. We could then come up with a separate name for the relation over the `Traversable` interface (which would happen to coincide with `Linked` for `List` & `Vec`)?

MatthewDaggitt:
@JacquesCarette are you happy with the proposal above? (please do feel free to say if you're not!). If you are then I wouldn't mind getting this in for `v1.2`...

JacquesCarette:
As this is under `List`, I'm fine with this proposal and having it go in v1.2.  If a generalization appears, we can always come back to this when the dust has settled.

PR #914, opened 2019-09-23
Subtraction for Data.Nat.Binary

mechvel:
changed:      src/Algebra/Properties/CommutativeSemigroup.agda
new file:     src/Data/Nat/Binary/Minus.agda
changed:      src/Data/Nat/Binary/Properties.agda

Please, consider this PR for  ``Binary.Minus``  edited according to your notes.

1) The ``ℕᵇ`` denotation and the folder Data/Nat/Binary/ are used.

2) Added to Algebra.Properties.CommutativeSemigroup:
```
-- Some permutation laws for _∙_ for four factors.                                
                                                                                  
xy∙zu≈xz∙yu :  ∀ x y z u → (x ∙ y) ∙ (z ∙ u) ≈ (x ∙ z) ∙ (y ∙ u)                  
...
```
It is used in Binary.Minus.

3) Proposal for  Relation.Binary...DectotalOrder
```
-- These three functons make it possible to write "tri-< lt" instead of           
-- "tri< lt (<⇒≉ lt) (<⇒≯ lt)".  This simplifies the code a lot.                  
                                                                                  
tri-< :  ∀ {x y} → x < y → Tri (x < y) (x ≈ y) (x > y)                            
tri-< lt =  tri< lt (<⇒≢ lt) (<⇒≯ lt)  
                                                                                  
tri-≈ : ∀ {x y} → x ≡ y → Tri (x < y) (x ≈ y) (x > y)                             
...
tri-> : ∀ {x y} → x > y → Tri (x < y) (x ≈ y) (x > y)                             
```
These are used in  Binary.Minus.
And currently they are in  Binary.Properties.
But probably their generalization to ``≈`` needs to be set somewhere to
Relation.Binary...DectotalOrder  - ?

4) Added to Binary.Properties:
```
x+suc[y]≡suc[x]+y : ∀ x y → x + suc y ≡ suc x + y
x≡0⇒double[x]≡0 : ∀ {x} → x ≡ 0B → double x ≡ 0B
pred[x]+y≡x+pred[y] : ∀ {x y} → x ≢ 0ᵇ → y ≢ 0ᵇ → (pred x) + y ≡ x + pred y 
pred-+ :  ∀ {x y} → x ≢ 0ᵇ → y ≢ 0ᵇ → (pred x) + y ≡ pred (x + y)                 
double-suc :  ∀ x → double (suc x) ≡ 2ᵇ + double x                               

-- Lifting the result of _<_ by the constructors 2[1+_], 1+[2_] 

even<even-mono-cmp : ∀ {x y} → Tri (x < y) (x ≡ y) (x > y) →       
                                       let x' = 2[1+ x ];  y' = 2[1+ y ]
                                       in           
                                       Tri (x' < y') (x' ≡ y') (x' > y') 
,
odd<odd-mono-cmp : ∀ {x y} → Tri (x < y) (x ≡ y) (x > y) →          
                                    let x' = 1+[2 x ];  y' = 1+[2 y ]                               
                                    in                                      
                                   Tri (x' < y') (x' ≡ y') (x' > y') 
```
These are used in  Binary.Minus.

5) Data.Nat.Binary.Minus  is rewritten

Now ``minus`` takes 18 lines, and subtraction is defined as
```
Minus : ℕᵇ  → ℕᵇ → Set                             -- result of extended subtraction     
Minus x y =  ℕᵇ × Tri (x < y) (x ≡ y) (x > y)                                    
                                                                                  
minus : ∀ x y → Minus x y    -- extended subtraction

_∸_ x =  proj₁ ∘ minus x      -- traditional subtraction
```
This is arranged so for the following reasons.
The simplest code for ``∸`` is
```
_∸_ : Op₂ ℕᵇ  
zero      ∸ _           =  0ᵇ                                                         
x           ∸ zero      =  x                                                         
2[1+ x ] ∸ 2[1+ y ]  =  double (x ∸ y)                                            
2[1+ x ] ∸ 1+[2 y ]  with y ∸ x                                                   
...                                          | zero = 1+[2 (x ∸ y) ] 
...                                          | _    = 0ᵇ
1+[2 x ] ∸ 2[1+ y ]  with x ∸ y                                                   
...                                          | zero = 0ᵇ                     
...                                          | _    = pred (double (x ∸ y))
1+[2 x ] ∸ 1+[2 y ] =  double (x ∸ y)                                             
```
But the proofs needed further are too difficult to build, at least for me.
For example, try to prove the main property
```
y≤x⇒x≡[x∸y]+y :  ∀ {x y} → y ≤ x → x ≡ (x ∸ y) + y
```
And with ``minus : ∀ x y → Minus x y``, everything is proved simple enough.

I you accept this approach, I would fill in the remaining proofs.


MatthewDaggitt:
Sorry @mechvel, it may take me a week or so until I have the time to look at this, but I will get round to it.

mechvel:
I also need a time-out, for about 2-4 weeks.

mechvel:
Sorry, I have lost the line. Can you, please, explain the situation.
As I recall, my last commit was Nat.Binary.Minus.
And I expected comments on it.
Instead I see "Resolve conflicts", and the conflicting file is  src/Data/Nat/Binary/Properties.agda.
I look into src/Data/Nat/Binary/Properties.agda  of _master_, and see that it differs greatly from such module in my local tree. 
There are added various "monomorphism" things related to ``toℕ, fromℕ``. 
I recall the monomorphism notion of the category theory, and it means a very different thing. So that I am not sure about these names.
All right, suppose that I agree (blindly) with the changes. In what way can I resolve a conflict in this file? 
I can copy this file to my local tree (with saving my version). Then Minus.agda would not type-checked, because, for example, it imports  ``even<even-mono-cmp``, and  this function has disappeared from Nat.Binary.Properties. 
Has it? 
Probably there are more such functions.
Please, advise.

mechvel:
You implement ``<-cmp`` for ``ℕᵇ`` by converting arguments to  ``ℕ``.
Right?
This will ruin performance. Because,  ``<-cmp`` has to apply further and is applied in ordinary computation, for fast comparison.
As to the rest, probably these OrderHomomorphism things look all right, they establish an explicit isomorphism  between the corresponding ordering structures for Bin and for Nat.

mechvel:
Besides the ``<-cmp`` performance, there is another reason to avoid  as possible conversion to Nat when composing proofs for _<_ on Bin 
(still I use such conversion in some proofs, because I am lazy to invent complicated proofs).

This reason is:  possible _evaluation with proofs_.
Regularly, I avoid such evaluation. But suppose someone needs such. 
And it is better to avoid the cost explosion.
For example, let  ``a, b, c : Bin``  are some large numbers, say about  2^30. 
There are given  ``lt1 : a < b;   lt2 : b < c``,
and we set to the program  
```
lt :  a < c
lt = <-trans lt1 lt2
```
Suppose that we need to evaluate with the ``lt`` value.
Each of  ``a, b, c``   has size about 30 (constructors).
The values ``lt1, lt2``  also have size about  30 -- 100.  Because ``_<_`` is defined without conversion to Nat.  
For my version of Binary.Properties,  ``lt`` has size about 300. 
Because ``<-trans`` is implemented without conversion to Nat.

And for your  version of Binary.Properties,  ``lt`` has size about 2^30.
Because  ``<-trans`` applies  ``NatProp.<-trans aN<bN bN<cN``,
where ``aN bN cN : Nat``  are unary codes of size about 2^30,
and a proof for  aN < bN  has about 2^30 occurrences of the ``s≤s`` constructor.

Am I missing something?

MatthewDaggitt:
> Instead I see "Resolve conflicts", and the conflicting file is src/Data/Nat/Binary/Properties.agda.
I look into src/Data/Nat/Binary/Properties.agda of master, and see that it differs greatly from such module in my local tree.
> ...
> Please, advise and explain, a bit.

So yes, various changes have been made to the way that morphisms are used to transfer proofs between types. For that reason you would have to have pulled the latest version from the standard library master branch and merge in your changes. However I've just done this for you. If you now run `git pull` from your repo then you should now get a version that type checks with all the latest changes to the standard library included (assuming you yourself haven't added any new commits on your local version, in which case you may have to resolve your own merge conflicts).

> The matter is that the proposed Bin implementation is fast even when the Nat arithmetic is explicit.
We need not loose this property.

Yup I had forgotten about this, I have opened a new PR #974 which reverts the proofs for decidability and trichotomy. This should return the previous behaviour and I'll make sure it's included in the upcoming release.

> This reason is: possible evaluation with proofs.
Regularly, I avoid such evaluation. But suppose someone needs such.
And it is better to avoid the cost explosion.

I don't think this is a problem as in such cases they can mark the proof as [irrelevant](https://agda.readthedocs.io/en/v2.6.0.1/language/irrelevance.html). As it happens for `Data.Nat.Binary` all proofs of `x < y` are unique and therefore can always be marked as irrelevant. Once marked as irrelevant means that they will never be evaluated hence sidestepping these concerns.

mechvel:
> I don't think this is a problem as in such cases they can mark the proof as irrelevant. As it 
> happens for Data.Nat.Binary all proofs of x < y are unique and therefore can always be marked as > irrelevant. Once marked as irrelevant means that they will never be evaluated hence sidestepping > these concerns.

So far, I never used the ``irrelevant`` construct. But your words  "they will never be evaluated" seem contradicting to something essential. Sometimes a programmer evaluates with proofs deliberately.
For example,  one can write ``f (s≤s m≤n) = ...`` which pattern-match against a certain proof for the relation of ``Nat._≤_``. How can ``f`` evaluate at this argument if this proof argument does not start evaluating? 
Please, explain this point.

On the other hand, my example with the two versions of ``<-trans`` for Bin may occur wrong.
Because both versions return a value of type  ``a < c`` expressed in the same data constructors for
``Binary.Base._<_``, and probably such a value is unique (?), as you write.

I need to think more about this. 
Thank you.

mechvel:
I continue to think that implementing ``<-trans`` on Bin via ``toℕ`` ruins performance -- when the program does evaluation with the values of type ``x < y`` on Bin.
Here is an example (written in old constructors for Bin).
```
size< : {x y : Bin} → x < y → ℕ                      -- number of constructor occurrences 
size< {0#}      {2suc y}  (0<2suc {y})             =  1
size< {0#}      {suc2* y} (0<suc2* {y})            =  1
size< {2suc x}  {2suc y}  (2suc<2suc {x} {y} x<y)  =  1+ (size< x<y) 
size< {2suc x}  {suc2* y} (2suc<suc2* {x} {y} x<y) =  1+ (size< x<y)
size< {suc2* x} {2suc y}  (suc2*<2suc (inj₁ x<y))  =  1+ (size< x<y)
size< {suc2* x} {2suc y}  (suc2*<2suc (inj₂ x≡y))  =  1  
size< {suc2* x} {suc2* y} (suc2*<suc2* x<y)        =  1+ (size< x<y)

<-trans' : Transitive _<_       
<-trans' {x} {y} {z} x<y y<z =  implemented  via   toℕ x, fromℕ,
                                               toℕ-mono-<, fromℕ-mono-<,  ℕp.<-trans
f :  {x y z : Bin} → x < y → y < z → ℕ
f x<y y<z =  size< (<-trans x<y y<z)       -- by <-trans

f' x<y y<z =  size< (<-trans' x<y y<z)      -- by <-trans'
```
The test shows that for  ``a = 2^ 25 : Bin,  b = suc a,  c = suc b``,
evaluation and printing  ``f' a b c``  costs about 10000 times more than for ``f a b c``,
while both values equal 2.
The test is
```
test : ℕ → String
test e =
  let  a = 2^ e;   b = suc a;   c = suc b
  in
  case (<-cmp a b , <-cmp b c)
  of \
  { (tri< a<b _ _ , tri< b<c _ _) → 
          concatStr
          ("e    =  "   ∷ showN e     ∷
           "\na  =  "   ∷ show a      ∷ 
           -- "\nf  =  "   ∷ showN (f a<b b<c)   ∷
           "\nf' =  "   ∷ showN (f' a<b b<c)  ∷
           []
          )
  ; _ → "error"
  }
```
``<-cmp`` is applied in order to obtain the proofs in ``a < b,  b < c``.
 
The size of the final proof (and I believe, the proof itself) is the same. But its run-time cost is very different. So that possible pattern-matching on the value of  ``f x<y y<z`` has a much smaller cost order than for ``f' x<y y<z``.

So that I have the following impression. 
* All old implementation in Binary.Properties that was without using ``toℕ`` needs to remain
  (By the way, I took effort to avoid ``toℕ``).
* There are more proofs there that need to be implemented  without using ``toℕ``
  (for example, distributivity). But these proofs are a bit complicated, and I suggest to delay their 
  design for future.  

What people would think about all this?

mechvel:
For example, it is better to prove ``+-comm``  for Bin without mapping to  ``ℕ``. 
It is simple. I write it in old constructors, because _master_ is under construction, and it is easier for me to debug examples under BFLib-0.2:  
```
+-comm :  FP-Bin.Commutative _+_
+-comm 0#          y             =  sym (+0 y)
+-comm x            0#            =  +0 x
+-comm (2suc x)  (2suc y)  =  cong (2suc ∘ suc) (+-comm x y)
+-comm (2suc x)  (suc2* y) =  cong (suc ∘ 2suc) (+-comm x y)
+-comm (suc2* x) (2suc y)  =  cong (suc ∘ 2suc) (+-comm x y)
+-comm (suc2* x) (suc2* y) =  cong (suc ∘ suc2*) (+-comm x y)
```
Probably ``+-assoc`` is also easy. It worth to try ``*-comm, *-assoc, *-distrib-+``, but I suggest
to delay this for the future stdlib releases.

mechvel:
Concerning evaluation with proofs:  I doubt of whether we can get free of it.

I sometimes use such like in termination proofs. 
To prove termination of some ``f``, the counter  ``cnt : ℕ`` is introduced, with the bound like
``cnt≤foo[x] : cnt ≤ foo x``. At each step of recursion for ``f`` the counter changes from  
``suc cnt`` to ``cnt``, and the bound witness changes from ``s≤s cnt≤foo[x]`` to  ``cnt≤foo[x']``.
This applies pattern matching on a proof for ``Nat._≤_``.
For example, for Bin,  ``cnt`` can be the number of constructors in x : Bin, 
for ``f`` taking a pair of lists, it can be ``length xs + length ys``. 
To prevent the evaluation cost explosion one needs to care of the bound for ``cnt``.

MatthewDaggitt:
> So far, I never used the irrelevant construct.

As you're very concerned with performance, this seems like an important area for you to look into. The whole point of irrelevant arguments is that they are never evaluated.

> For example, one can write f (s≤s m≤n) = ... which pattern-match against a certain proof for the relation of Nat._≤_. How can f evaluate at this argument if this proof argument does not start evaluating?

The actual term for `m ≤ n` is determined entirely by `m` and `n` and hence the function can evaluate `m` and `n` instead of the proof `m ≤ n`.

> Here is an example:

I'm almost certain that with a little bit of re-arranging it is possible to mark the `x < y` proof in your `size<` function as irrelevant as it really it's result is a function of `m` and `n`. 

> I sometimes use such like in termination proofs.

`Acc` proofs are erased at compile time [according to @jespercockx](https://lists.chalmers.se/pipermail/agda/2019/011113.html) and so should not affect performance.

Nonetheless to make life easier for you, I'll role the other ordering proofs back in #974 tomorrow when I have the chance.

mechvel:
>>  For example, one can write ``f (s≤s m≤n) = ...`` which pattern-match against a certain proof for >> the relation of ``Nat.≤``. How can ``f`` evaluate at this argument if this proof argument does not >> start evaluating?

> The actual term for m ≤ n is determined entirely by m and n and hence the function can 
> evaluate m and n instead of the proof m ≤ n.

So, needs standard library to claim: 
"Caution:  evaluating with proofs,  pattern-matching on proofs may ruin performance"
 ?

> I'm almost certain that with a little bit of re-arranging it is possible to mark the x < y proof in 
> your size< function as irrelevant as it really it's result is a function of m and n.

Here is a simpler example.
Let ``<-trans : Transitive _<_``    be implemented as in Binary.Properties in my version.
Let ``<-trans' : Transitive _<_``    be implemented via toℕ, as in your version of  Binary.Properties.
Now, ``f`` applies ``<-trans`` and ``f'`` applies ``<-trans'`` :  
```
f :  {x y z : Bin} → x < y → y < z → ℕ
f x<y y<z  with  <-trans x<y y<z
... | even<even _ =  0
... | _                    =  1

f' :  {x y z : Bin} → x < y → y < z → ℕ
f' x<y y<z  with  <-trans' x<y y<z
... | even<even _ =  0
... | _                   =  1
```
For some arguments about 2^20,  ``f`` evaluates about 1000 times faster than ``f'``.  
What would you advise  to a programmer who wrote this ``f'`` ?
Where to set ``irrelevant`` ?
Is ``irrelevant`` by prefixing something with a dot? Where to set this dot in this particular example?

* Will this denotation help in this example?
* 3-4 years ago I used dot to mark irrelevance somewhere. I failed to type-check DoCon-A with it,
   and nobody was able to help with advice. I was stuck. So, I removed this denotation; the   
  programs have become a bit longer, but they work anyway.

By the way, a similar adventure was with the ``inspect`` denotation and with the ``with`` construct. 
The way out is as follows.
1) Replace ``inspect`` with introducing explicitly an auxiliary function, let it be longer, but it is clear to
    understand and it will type-check.
2) If  ``f`` is implemented via ``with``, and a proof for ``f`` is not type-checked, then
     replace ``with``  with introducing explicitly an auxiliary function (at the same scope level).

Some serious person cried on a certain conference in 2013 "I cannot write proofs in Agda!".
I think, this was due to ``with``, ``inspect``, and may be, due to the irrelevance denotation.

MatthewDaggitt:
> So, needs standard library to claim: "Caution: evaluating with proofs, pattern-matching on proofs may ruin performance"?

I think that probably goes without saying.

> What would you advise to a programmer who wrote this f' ?
Where to set irrelevant ?

In your example the proof of `x < y` and `y < z` are so irrelevant that you don't even need them as arguments at all. I'd advise them to simply write it as:
```agda
f' :  (x z : Bin) → ℕ
f' 2[1+ x ] 2[1+ z ] =  1
f' _        _        =  0
```
and your original `size<` I'd write as:
```agda
size< : (x y : Bin) → ℕ
size< x y = size y ∸ size x
```
As a general rule of thumb, no function that computes a value should depend on proof of `<`, you should always be able to rewrite it without the proof.  Hope these examples illustrate how you may avoid such dependencies.

mechvel:
>> What would you advise to a programmer who wrote this f' ?
>>    Where to set irrelevant ?

> In your example the proof of x < y and y < z are so irrelevant that you don't even need them as 
> arguments at all. I'd advise them to simply write it as:
```
> f' :  (x z : Bin) → ℕ
> f' 2[1+ x ] 2[1+ z ] =  1
> f' _        _        =  0

You wrote about using irrelevance denotation as a way out. And my example was to see how the irrelevance denotation may help (I expected that it would not help). 
Instead you remove the proofs at all. 

In Agda, proofs are data. And programmers are free to choose the style which combines evaluation on proofs with evaluation on "ordinary" data. Call this _EP style_.
For example, a program may have a goal to analyze proofs of some kind.

Each part of standard library may support or not support the EP style. 
Not supporting it for a function f in standard library means that using EP style for the result of f 
ruins performance (as compared to the no-EP style) for some arguments. 
The more functions in standard library support the EP style the better. 
If we manage to implement a function with supporting the EP style, why not to do this?
As to  Nat.Binary,  all of it can be written with supporting the EP style.

I think, we need to return to my last version of  Binary.Properties.

As to my personal experience, so far I use EP only for the counter for termination proof:
``(suc cnt,  (s≤s l≤cnt)) --> (cnt,  l≤cnt)``.
You write about using the WellFounded recursion tool instead.
But in some situations using the counter is easy to arrange and is more clear. 
 WellFounded & Acc matter is brain-twisting. To arrange it is much more difficult than to write an intuitive proof by induction: "this expression is smaller than this in this sense, hence the process 
terminates". A way from this intuitive discourse to the program construct of WellFoulded & Acc is a brain-twisting exercise. Only observe in the library the type expressions for these constructs.

Again, other users of standard library may have more reasons for using the EP style.

**Comment by MatthewDaggitt** [comment]
> In Agda, proofs are data. 

You're 100% right. So let's talk about your Agda coding style in terms of _data_. Imagine you have some toy problem to solve in some classical simply typed programming language, and you've written an algorithm to solve it that uses a priority queue. You've managed to get it really fast because you've put a lot of effort into implementing the priority queue with some fancy data structure, e.g. a van Emde Boas tree.

Now imagine someone comes along and shows you a way to rewrite your algorithm that is exactly the same except that it doesn't need a priority queue. In this situation wouldn't you say "Great, I don't have to maintain a very complicated priority queue structure now"?

This is exactly the same situation as your Agda code. The proofs being passed into your functions correspond to the priority queue structure. You don't _need_ them in your algorithm so why bother spending all that time making them efficient?

Writing efficient code isn't just about fixating on one implementation and trying to optimise that. It's about trying to find better and more efficient algorithms.

> The more functions in standard library support the EP style the better. 

Unfortunately this is not true. Hand-proving all these proofs makes them much more complex and comes with a significant cost in terms of type-checking time. The standard library now takes 6 minutes to type check on my laptop up from 3 minutes a couple of years back. The file `Data.Nat.Binary.Properties` takes 10 seconds alone. This means it takes up 3% of the time every time we type check the library. The tests will be type checked 10,000s of times by many different people, and all of this represents a significant cost to the users.

Having said all this, I have rolled the rest of the ordering proofs back to help you in the short term. However in the long term I think we are going to have to start optimising the library and looking at the  worst offending code in terms of type checking time. Therefore I can't guarantee that we won't switch back to the morphism style of proofs eventually.

**Comment by mechvel** [comment]
For example, what is better: 
(I) to have 20 library functions that type-check in the time T and have the evaluation cost order   
    O(2^n) 
or
(II) to have them as 10 times larger in code, and type-checking in the time 4*T,  and having the evaluation cost O(n^2)
?
I think that (II) is better.

> Having said all this, I have rolled the rest of the ordering proofs back to help you in the short term. > However in the long term I think we are going to have to start optimising the library and looking at > the worst offending code in terms of type checking time. Therefore I can't guarantee that we won't > switch back to the morphism style of proofs eventually.

Let us try this approach.

I understand this so that now it is your turn: to consider the module Binary.Minus as it is, as it has been committed. Because this is the current state of the project.
Right?

**Comment by MatthewDaggitt** [comment]
> DoCon-A-2.02 takes 50 minutes to type-check on a 3 GHz machine

Oh wow. I have to confess I'm a little surprised to find you arguing that optimising type checking time isn't important!

> I think that (II) is better, even assuming that this 2^n is only for the user programs that evaluate with proofs for arithmetic on Bin.

Of which there are precisely none that ever need to be written! Anyway I sense that this isn't exactly going anywhere so lets move swiftly on.

I will try to look at this PR in the next couple of weeks once v1.3 has been released.

**Comment by mechvel** [comment]
> I have to confess I'm a little surprised to find you arguing that optimising type checking time isn't 
> important!

* Reducing exponentially run-time performance is more important than reducing 10 times type 
  checking cost.
* For these two libraries programs, reducing the type checking cost can probably be only by 
  optimizing the type checker.

**Comment by mechvel** [comment]
> I will try to look at this PR in the next couple of weeks once v1.3 has been released.

Really 1.3, not 1.2 ?

File: src/Data/Nat/Binary/Minus.agda [hunk 1]
```diff
@@ -0,0 +1,239 @@
...
+  (_≡_; _≢_; _≗_; refl; sym; trans; cong; cong₂; resp₂; subst)
+import Relation.Binary.Reasoning.Base.Triple as InequalityReasoning
+open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Algebra.FunctionProperties {A = ℕᵇ} _≡_
+open import Algebra.Properties.CommutativeSemigroup +-commSemigroup
+     using (xy∙z≈y∙xz; x∙yz≈y∙xz; xy∙zu≈xz∙yu)
+
+private
+  module IneqReasoning = InequalityReasoning {A = ℕᵇ} {_≈_ = _≡_} ≤-isPreorder
```
MatthewDaggitt: This is just `≤-Reasoning` from `Data.Nat.Binary.Properties`

src/Algebra/Properties/CommutativeSemigroup.agda
```diff
@@ -114,3 +114,14 @@ xy∙z≈yz∙x x y z =  trans (xy∙z≈y∙zx x y z) (sym (assoc y z x))
 
 xy∙z≈zx∙y :  ∀ x y z → (x ∙ y) ∙ z ≈ (z ∙ x) ∙ y
 xy∙z≈zx∙y x y z =  trans (xy∙z≈z∙xy x y z) (sym (assoc z x y))
+
+------------------------------------------------------------------------------
+-- Some permutation laws for _∙_ for four factors.
+------------------------------------------------------------------------------
+
+xy∙zu≈xz∙yu :  ∀ x y z u → (x ∙ y) ∙ (z ∙ u) ≈ (x ∙ z) ∙ (y ∙ u)
```
MatthewDaggitt: I'm don't think we want to go listing all 120 permutation rules for 4 variables. Instead at this level of complexity we should be trying to use the library's monoid solvers instead.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,239 @@
...
+                                   <-trans (resp₂ _<_) <⇒≤ <-≤-trans ≤-<-trans
+open ℕᵇ
+open Tri
+open _<_
```
MatthewDaggitt: Don't need any of these 3 `open` statements

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,239 @@
...
+... | (_ , tri< x<y _  _)   =  (0ᵇ , tri-< (odd<even (inj₁ x<y)))
+... | (_ , tri≈ _   eq _)   =  (0ᵇ , tri-< (odd<even (inj₂ eq)))
+... | (d , tri> _   _  x>y) =  (pred (double d) , tri-> (even<odd x>y))
+minus 1+[2 x ] 1+[2 y ]  with minus x y
+... | (d , cmp) =  (double d , odd<odd-mono-cmp cmp)
+
+infixl 6 _∸_
+
+_∸_ : Op₂ ℕᵇ              -- traditional subtraction
+_∸_ x =  proj₁ ∘ minus x
```
MatthewDaggitt: I think your definition of minus is unnecessarily complicated. I would change it for the much shorter definition:
```
_∸_ :  ℕᵇ → ℕᵇ → ℕᵇ
zero     ∸ _        = 0ᵇ
x        ∸ zero     = x
2[1+ x ] ∸ 2[1+ y ] = double (x ∸ y)
1+[2 x ] ∸ 1+[2 y ] = double (x ∸ y)
... | yes _  = 0ᵇ
... | no  _  = 1+[2 (x ∸ y) ]
1+[2 x ] ∸ 2[1+ y ] with x ≤? y
... | yes _  = 0ᵇ
... | no  _  = pred (double (x ∸ y))
```
This cuts down the number of cases significantly, making subsequent proofs shorter. The proof components from the removed trichotomous component can easily be recovered from the proofs from `yes` and `no`

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,239 @@
...
+... | tri> _ _     gt =  gt
+... | tri≈ _ y∸x≡0 _  =  contradiction (x∸y≡0⇒x≤y y∸x≡0) (<⇒≱ x<y)
+... | tri< y∸x<0 _ _  =  contradiction y∸x<0 x≮0
+
+y≤x⇒x≡[x∸y]+y :  ∀ {x y} → y ≤ x → x ≡ (x ∸ y) + y    -- the defining property for _∸_
+y≤x⇒x≡[x∸y]+y {zero} {y} y≤0 =  sym $ begin-equality
+  (0ᵇ ∸ y) + y    ≡⟨ cong₂ _+_ (cong (0ᵇ ∸_) y≡0) y≡0 ⟩
+  (0ᵇ ∸ 0ᵇ) + 0ᵇ  ≡⟨ +-identityʳ 0ᵇ ⟩
+  0ᵇ              ∎
+  where open IneqReasoning;  y≡0 = x≤0⇒x≡0 y≤0
```
MatthewDaggitt: You can just open `≤-Reasoning` once at the top of the file.

src/Data/Nat/Binary/Properties.agda
```diff
@@ -654,20 +660,35 @@ private
   { isSemigroup = +-isSemigroup
   }
 
++-commSemigroup : CommutativeSemigroup 0ℓ 0ℓ
```
MatthewDaggitt: The standard name is `+-commutativeSemigroup`

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,239 @@
...
+
+infixl 6 _∸_
+
+_∸_ : Op₂ ℕᵇ              -- traditional subtraction
+_∸_ x =  proj₁ ∘ minus x
+
+0∸x≡0 :  ∀ x → 0ᵇ ∸ x ≡ 0ᵇ
+0∸x≡0 zero     =  refl
+0∸x≡0 2[1+ _ ] =  refl
+0∸x≡0 1+[2 _ ] =  refl
```
MatthewDaggitt: For example, with the new definition I propose above no pattern matching is required here. It simply becomes `0∸x≡0 x =  refl`

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,239 @@
...
+x≤y⇒x∸y≡0 {1+[2 _ ]} {2[1+ _ ]} (inj₂ ())
+
+x≤y⇒x∸y≡0 {1+[2 _ ]} {1+[2 _ ]}  (inj₁ (odd<odd x<y)) =  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (inj₁ x<y))
+x≤y⇒x∸y≡0 {1+[2 x ]} {1+[2 .x ]} (inj₂ refl)          =
+  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 {x} {x} (inj₂ refl))
+
+x∸x≡0 :  ∀ x → x ∸ x ≡ 0ᵇ
+x∸x≡0 x =  x≤y⇒x∸y≡0 (≤-refl {x})
+
+y<x⇒x∸y≢0 :  ∀ {x y} → y < x → x ∸ y ≢ 0ᵇ
```
MatthewDaggitt: The convention is that the variables should be allocated names in the order that they appear, i.e. this should be either `x<y⇒y∸x≢0` or `x>y⇒x∸y≢0` rather than `y<x⇒x∸y≢0`

src/Algebra/Properties/CommutativeSemigroup.agda
mechvel: It is written "Some permutations". I set only few ones that I use in Nat.Binary.
If we need many of them, then we could try to use monoid solver.
OK?

src/Data/Nat/Binary/Minus.agda
mechvel: Your statement  for ``2[1+ x ] ∸ 1+[2 y ]`` computes ``x <? y`` and then computes ``x ∸ y`` by new.
And in my program, computing ``minus 2[1+ x ] 1+[2 y ]`` applies ``minus x y``, which also gives the result of  ``<-cmp x  y`` and  ``x <? y``.  This latter result is formed for free as an addition to the difference denoted ``d`` in the code.
My format for ``minus`` is chosen so because I think that a natural way of computing ``x ∸ y``   
includes computing ``x <? y``. So that there is no need to apply ``x <? y`` separately.
I expect that my ``minus`` is about 3/2 times faster than your ``∸``.
But I am not sure.

src/Algebra/Properties/CommutativeSemigroup.agda
MatthewDaggitt: Again it's better to be consistent. Either we include all of them or none of them. Otherwise how can people predict which ones are present and which ones are missing?

mechvel: This leads us to a trap. 
Either to apply a monoid solver for several rather simple cases, with complicating the code with preparing parameters for the solver,
or to implement ``xy+zu≡xz+yu,  xy*zu≡xz*yu`` in Binary.Properties individually for Bin
(and then, probably, to do this in DivMod.agda, GCD.agda -- for CommutativeSemiring). 
??

MatthewDaggitt: > Either to apply a monoid solver for several rather simple cases, with complicating the code with preparing parameters for the solver,

I agree, writing out for the solver doesn't always look nice. On the `experimental` branch we have a new monoid solver from @oisdk which doesn't require you to write out the full equation. This should be available after the next release of Agda.

In the meantime however I would recommend you using the existing solver. I don't think that your second proposal of proving it individually in multiple files is not the way to go.

src/Data/Nat/Binary/Minus.agda
MatthewDaggitt: You may very well be right, but I'm afraid I don't quite see it myself why your version is more efficient than mine. Yes you're right that mine does effectively calculate the difference multiple times. However doesn't yours also do that with all the translations of `tri-<`, `tri->` etc?

I would be interested in seeing the timings...

mechvel: I doubt. 
I shall look into this after may be a week. I need a time out.

src/Algebra/Properties/CommutativeSemigroup.agda
mechvel: In e-mail, your last sentence ends with "is not the way to go",
which makes it a great difference. 
Probably, you have pressed "Comment", and the copy went to e-mail. 
And after this, you have corrected the message by "Edit".
Right?

With this, I conclude the following:
a message copy in e-mail often is not a fresh message on GitHub,
and it is better to read GitHub after seeing the corresponding message in e-mail.
??

gallais: > In the meantime however I would recommend you using the existing solver

Or to target your PRs at `experimental` so that you may already start
using the new solver! It will make a great test case for it too.

src/Data/Nat/Binary/Minus.agda
mechvel: I tried a certain test, and your version for ``∸`` has occurred faster, I do not understand, why.

All right, let us choose your version.
Let us delay this performance analysis, it is better to advance with the project. 
I have changed the whole Minus.agda, accommodating  it to your version of ``∸`.
But I do not recall the situation with this PR. Probably  _master_ has changed greatly.
So, I am going to do the following.
* To save your last notes on this PR,
* To delete this my branch from GitHub.
* To start a new branch from _master_ with a PR with Minus.agda, together with several 
   small evident items of which you suggested me to do a PR.
   Because everything except Minus.agda is already in _master_. Right?

MatthewDaggitt: > Because everything except Minus.agda is already in master. Right?

That is correct.

But you don't necessarily have to delete the branch and start again. You could [pull the changes down from the library and merge them in](https://stackoverflow.com/questions/34656523/git-update-local-branch-with-remote-master/34656923)?

mechvel: I want to replace my local branch with the copy of remote _master_. 
Only this, to start with.
Is this by
```
$ cd .../agda-stdlib-fork        (my local, where CHANGELOG.md is)
$ git checkout master
$ git pull
```
? 
On the Web, it is also added 
```
$ git checkout <your-branch>
$ git merge master
```
Is this needed? If yes, then what need I set for <your-branch> ?  Is it '.' ?

MatthewDaggitt: > I want to replace my local branch with the copy of remote master.

That's not quite what you want to do. You want merge the remote master branch with your master branch.

After your first two commands I would run `git merge origin master`

mechvel: I followed this. It reports "Already updated" (translated from Russian).
Now I am going to commit a PR with Nat.Binary.Minus

src/Algebra/Properties/CommutativeSemigroup.agda
mechvel: Let us delay this and wait until it becomes _master_.
Even with _master_ I have too many things to change.

I use the CommutativeMonoid solver in the coming Binary.Minus.

src/Data/Nat/Binary/Minus.agda
mechvel: ``open Tri`` is used in new Minus.agda, and I remove the other two.

src/Data/Nat/Binary/Minus.agda
mechvel: In the coming Binary.Minus, it is also used ≡-Reasoning for Nat (not fir Bin).
So, to avoid a name clash, I am forced to open the two Reasonings locally.
(?)

src/Data/Nat/Binary/Properties.agda
mechvel: Renamed.

src/Data/Nat/Binary/Minus.agda
mechvel: Done. I remove this lemma, instead one has to apply ``refl``.

src/Data/Nat/Binary/Minus.agda
mechvel: Changed to ``x>y⇒x∸y≢0``. The same done to a certain another function.

mechvel:
Please consider my commit of December 23, 2019.
                                               
All your notes are answered and taken in account.                                               
There are added  
* new  Nat.Binary.Minus  -- the main thing,
* the four minor functions (see CHANGELOG.md).   
                                                                                      
These minor functions are evidently useful, and some of them Matthew invited for a PR,
all of them were discussed here on Issues.
On the other hand most of them are not used in the current Nat.Binary.
I join them in this PR, because they are small and because it is difficult for me to commit many small PR-s.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Nat.Binary.Minus where
+
+import Algebra.FunctionProperties as FuncProp
+open import Data.Empty using (⊥)
+open import Data.Fin using () renaming (zero to 0F; suc to sucF)
+open import Data.Nat as ℕ using (ℕ)
```
gallais: Instead of renaming these and writing `0F`, `sucF 0F` and `sucF (sucF 0F)`,
you could use [`#_`](https://agda.github.io/agda-stdlib/Data.Fin.html#730): all your cases are concrete so Agda should see that
you can indeed embed `0`, `1`, and `2` into `Fin 3`.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+zero     ∸ _        = 0ᵇ
+x        ∸ zero     = x
+2[1+ x ] ∸ 2[1+ y ] = double (x ∸ y)
+1+[2 x ] ∸ 1+[2 y ] = double (x ∸ y)
+2[1+ x ] ∸ 1+[2 y ] with x <? y
+... | yes _ = 0ᵇ
+... | no  _ = 1+[2 (x ∸ y) ]
+1+[2 x ] ∸ 2[1+ y ] with x ≤? y
+... | yes _ = 0ᵇ
+... | no  _ = pred (double (x ∸ y))
```
gallais: Same comments as earlier: avoid `yes / no` whenever you do not need the proofs
and use `does` instead. This will give you more efficient functions (essentially
computing a boolean).

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+
+x∸x≡0 : ∀ x → x ∸ x ≡ 0ᵇ
+x∸x≡0 x =  x≤y⇒x∸y≡0 (≤-refl {x})
+
+x>y⇒x∸y≢0 : ∀ {x y} → x > y → x ∸ y ≢ 0ᵇ
+x>y⇒x∸y≢0 {zero} {_} ()
+x>y⇒x∸y≢0 {x'@(2[1+ x ])} {zero} _ =  subst (_≢ 0ᵇ) x'≡x'∸0 2[1+x]≢0
+  where
+  x'≡x'∸0 = sym (x∸0≡x x')
+
```
gallais: I think we can inline such trivial proofs.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+  2ᵇ + double (x∸y + y)                ≡⟨ cong ((2ᵇ +_) ∘ double) $ sym $
+                                         x≥y⇒x≡[x∸y]+y y≤x ⟩
+  double 1ᵇ + double x                 ≡⟨ sym (double-distrib-+ 1ᵇ x) ⟩
+  double (1ᵇ + x)                      ≡⟨ cong double (sym (suc≗1+ x)) ⟩
+  double (suc x)                       ≡⟨ sym (2[1+_]-double-suc x) ⟩
+  2[1+ x ]                             ∎
+  where
+  open ≤-Reasoning
+  x∸y = x ∸ y
+  eq = let _+_ = _⊕_;  a = var 0F;  b = var (sucF 0F);  c = var (sucF (sucF 0F))
```
gallais: Location where one may use `#_` from `Data.Fin`.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+  toℕ 0ᵇ        ≡⟨ refl ⟩
+  0             ≡⟨ sym (ℕₚ.m≤n⇒m∸n≡0 m≤n) ⟩
+  m ℕ.∸ n       ∎
+  where
+  open ≡-Reasoning
+  m = toℕ x;  n = toℕ y;  x≤y = <⇒≤ x<y;  m≤n = toℕ-mono-≤ x≤y
+
+... | tri≈ _ x≡y _ =  begin
+  toℕ (x ∸ y)  ≡⟨ cong toℕ (x≤y⇒x∸y≡0 {x} {y} x≤y) ⟩
+  toℕ 0ᵇ       ≡⟨ refl ⟩
```
gallais: `≡⟨ refl ⟩` can be replaced with `≡⟨⟩` (also applies to other lines in the file).

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,378 @@
...
+    toℕ (x ∸ y) ℕ.+ n    ≡⟨ sym (toℕ-homo-+ (x ∸ y) y) ⟩
+    toℕ ((x ∸ y) + y)    ≡⟨ cong toℕ (+-comm (x ∸ y) y) ⟩
+    toℕ (y + (x ∸ y))    ≡⟨ cong toℕ (x+[y∸x]≡y y≤x) ⟩
+    m                    ≡⟨ sym (ℕₚ.m+[n∸m]≡n n≤m) ⟩
+    n ℕ.+ (m ℕ.∸ n)      ≡⟨ ℕₚ.+-comm n (m ℕ.∸ n) ⟩
+    (m ℕ.∸ n) ℕ.+ n      ∎
+
+
+fromℕ-homo-∸ :  ∀ m n → fromℕ (m ℕ.∸ n) ≡ (fromℕ m) ∸ (fromℕ n)
+fromℕ-homo-∸ m n =  begin-equality
```
gallais: Given that we have `fromℕ-toℕ`, it feels like we ought to be able to prove a more general
lemma. i.e.

```agda
  (f : ℕ → ℕ) (g : ℕᵇ → ℕᵇ) →
  (∀ m n → toℕ (g m n) ≡ f (toℕ m) (toℕ n)) →
  (∀ m n → fromℕ (f m n) ≡ g (fromℕ m) (fromℕ n))
```

and the converse.

src/Data/Nat/Binary/Minus.agda
MatthewDaggitt: You don't need `open Tri` either.

src/Data/Nat/Binary/Minus.agda
MatthewDaggitt: Hmm the lemma still seems useful to have as one may not realise that `refl` is the answer. I'd probably keep it in.

MatthewDaggitt:
> These minor functions are evidently useful, and some of them Matthew invited for a PR,
all of them were discussed here on Issues.
On the other hand most of them are not used in the current Nat.Binary.
I join them in this PR, because they are small and because it is difficult for me to commit many small PR-s.

@mechvel I'm really sorry but I'm really not keen on adding many different things in the same PR. There are many reasons to avoid doing so, here are few examples:
- it makes it difficult to search for which commit something was added in, e.g. very few people would think `search` was added in a commit named `add minus for Data.Nat.Binary`.
- it makes reverting commits difficult. If we find a problem in `Data.Nat.Binary.Minus` we don't want to have to roll back `search` as well.
- it makes using git bisect more difficult.
- it makes reviewing the PR much easier.

I know you find using Git difficult and I 100% understand that not being a native English speaker doesn't make things any easier, but I promise you you'll save so much more time (both yours and ours) if you spent a day learning the basics of git branches. There's plenty of tutorials out there. I'd recommend setting up a toy repository and just practising for a bit.

mechvel:
So, I have to remove these minor functions and commit once more with Minus.agda ?

src/Data/Nat/Binary/Minus.agda
mechvel: You seem to refer to old Minus.agda while there is suggested the commit of December 23
(which uses the constructors of Tri).

MatthewDaggitt: Your commit on December 23 6b3c586 still opens `Tri` but as far as I can see the file no longer uses it as we've switched to a different definition?

MatthewDaggitt:
> So, I have to remove these minor functions and commit once more with Minus.agda ?

I think that would be the best plan. To be honest given the size of this thread it might be best to address the outstanding comments and then make an entirely new PR.

src/Data/Nat/Binary/Minus.agda
mechvel: The new Minus.agda
* has a simpler implementation for ``∸`` (as you suggested),
* has about twice as much functionality (the first version was not complete, it was to see the reaction and then to decide),
* uses the constructors of Tri, for example, in ``x<y⇒0<y∸x``.

Following your suggestion, I am going to save the fresh comments on GitHub, delete this my branch, and to start a new PR with Minus.agda and without those minor functions mentioned.

mechvel:
> I'm really sorry but I'm really not keen on adding many different things in the same PR. There are 
> many reasons to avoid doing so, here are few examples
> [..]

I do not fear of languages. And probably, there exists a Russian translation. But this would not make easier to read such.
I do not understand another thing. 
For any tiny item you suggest to do a PR. For example, I propose with 
```
Data.list.Base: 
-- Delete the first x such that P x, return the given list if it has not such x. 
delete : ∀ {P : Pred A p} → Decidable P → List A → List A
delete _  []        =  []
delete P? (x ∷ xs)  with P? x
... | yes _ =  xs
... | no _  =  x ∷ (delete P? xs)
```
If you approve it, why do not you put it into _master_, without any bureaucracy?  
Do you wish your library to become better?

All right, you can change ``(yes _)`` to ``true`` silently. Anyway, people look into CHANGELOG.md, and they can issue notes. This will be all the same, but without extra bureaucracy.

gallais:
What you call "without extra bureaucracy" I would call "us doing additional work
because you refuse to learn basic git commands".

Creating a new branch for each PR is **really** not that hard. And if you were
also using git to merge `master` to bring your branch up to speed, you would also
gain a lot time wasted by restarting a PR from scratch 2 or 3 times per feature.

MatthewDaggitt:
> I do not understand another thing. For any tiny item you suggest to do a PR. If you approve it, why do not you put it into master, without any bureaucracy? Do you wish your library to become better?

So I think you slightly misunderstand how contributing to the library works. It's the Agda community's library, not mine. I volunteer my time in order to help keep everyone's contributions consistent and manage releases but beyond that I'm just another contributor. Like every other contributor I pretty much only add things to it that I myself find useful. Adding everything that everyone suggests myself would pretty much be a full time job.

Furthermore the PR system helps maintain code quality and allows everyone to have a say. Even for my own contributions I never add code directly to `master` and always put up my own pull requests (e.g. #1006 #850). Bypassing the PR system would result in a lowering of the quality of code that's contributed to the library. Even for small additions such as you suggest, there are plenty of things that might be picked up in a PR. For example, it might already exist somewhere else in the library.

In short we value your suggestions and contributions but I'm afraid we don't have the time to implement all of your suggestions ourselves.

mechvel:
> but I'm afraid we don't have the time to implement all of your suggestions ourselves.

The function
```
pred-n≤n : ∀ {n} → pred n ≤ n
pred-n≤n = ≤⇒pred≤ ≤-refl
```
is already implemented on this page. And this GitHub place (or on Issues) already allows everyone to have a say.
A maintainer is not an ordinary user, because ordinary users do not decide on merging a commit to _master_. I understand this so: the library maintainer considers the notes on the committed PR and decides on merging it into _master_.  For simple suggestions like ``pred-n≤n``, which code is given on the page, the maintainer's work amount is  the same whether he considers the notes on PR on the _branch_ and then merges it into _master_ or considers the notes on Issues and puts the _given code_ to _master_. 
?

gallais:
Alternatively:

```shell
git checkout master
git pull
git checkout -b my_shiny_feature
emacs src/THE_AGDA_FILE_I_WANT_TO_MODIFY.agda
emacs CHANGELOG.md
git add -p
git commit
git push origin -u my_shiny_feature
```

And then open a PR on github.

* Takes you an additional 1 minute
* CI checks the patch is valid
* No extra work for the maintainers
* PR ready to be reviewed and merged

MatthewDaggitt:
Closed in favour of #1026

PR #918, opened 2019-09-28
Turn binary ++⁺ and related operators infix.

andreasabel:
This is motivated by their proof-relevant use.  For instance
```agda    
      _++⁺_ : All P xs → All P ys → All P (xs ++ ys)
```
is the operation concatenating two well-typed environments, in case `xs` and `ys` are contexts (lists of types) and `P` is mapping a type to its set of values.

Also: deprecate `reverseAcc`.

This PR originated from https://github.com/agda/agda-stdlib/pull/899#discussion_r325096169

### DIFF
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -308,18 +308,17 @@ _─_ : (xs : List A) → Fin (length xs) → List A
 ------------------------------------------------------------------------
 -- Operations for reversing lists
 
-reverseAcc : List A → List A → List A
-reverseAcc = foldl (flip _∷_)
-
-reverse : List A → List A
-reverse = reverseAcc []
-
 -- "Reverse append" xs ʳ++ ys = reverse xs ++ ys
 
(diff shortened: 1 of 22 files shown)


MatthewDaggitt:
So no problems with deprecating `reverseAcc`. My main concern is that the change to infix notation is quite a big non-compatible change and the concatenation function `_++_` is quite widely used. Because we're keeping the same name, I don't think we can even mark the old non-infix version as deprecated right?

What do other people think?

MatthewDaggitt:
@andreasabel not sure if you've had any further thoughts about how to make this change backwards compatible? If you haven't and if it's okay with you, I'm going to bump this to `v1.3`.

MatthewDaggitt:
Alternatively maybe we could just include the deprecation of `reverseAcc` and leave the infix notation for a separate PR?

MatthewDaggitt:
@andreasabel are you still keen to work on this in or should I close this?

Issue #839, opened 2019-07-14
Add a universe polymorphic Unit type?


JacquesCarette:
Issue #71 explains quite well why `⊤` in `Data.Unit` is not universe polymorphic.

But sometimes, one just needs a universe polymorphic unit type - see https://github.com/agda/agda-categories/pull/22 in particular.  Yes, one could use `Lift` all over the place, resulting in very ugly code, while the current code is quite clean.  I've also had to define it in the pi-dual repository, and I think also elsewhere too.

So the idea would be to add a new Unit type, which would be universe polymorphic. It could go into `Data.Unit` under a new name or `Data.Unit.Polymorphic` (or ???). But I think it is sufficiently useful to define it. [And likely the same for `Data.Empty`].

WolframKahl:
I call it `Data.Unit.Generalised`.

MatthewDaggitt:
I think I prefer to `Data.Unit.Polymorphic` but I'm happy with `Data.Unit.Generalised` if people prefer.

TOTBWF:
In a similar vein, it would be quite handy to have a universe polymorphic Empty type as well.

JacquesCarette:
So, in theory, this is now fixed - there is now `Data.Unit.Polymorphic` and `Data.Empty.Polymorphic`.  However the situation is still somewhat sub-optimal: these definitions do not interact very well with the ones in `Data.Unit` and `Data.Empty`.  In the sense that many uses of `Lift` on these in the rest of the library cannot be removed as there would be knock-on effects.

I think that for `Data.Empty`, this can be fixed be having `⊥` be an alias for `⊥ {0ℓ}`. Less sure how well that would work for `Data.Unit` since that's really `Agda.Builtin.Unit`, and I think that would mean having the `Level`-polymorphic `⊤` all be translated uniformly to Haskell.

MatthewDaggitt:
Added in 4aecce2

MatthewDaggitt:
@JacquesCarette feel free to open another issue if you'd like to continue to explore where they can be used.

PR #925, opened 2019-10-03
Add a Level-polymorphic version of Data.Unit 

JacquesCarette:
(in `Data.Unit.Polymorphic`) and accompanying properties.

[`Data.Empty.Polymorphic` will follow later].  Part of #839 .


src/Data/Unit/Polymorphic.agda
```diff
@@ -0,0 +1,38 @@
...
+record ⊤ {ℓ : Level} : Set ℓ where
+  constructor tt
+
+------------------------------------------------------------------------
+-- An ordering relation over the unit type
+
+record _≤_ {ℓ ℓ′ e : Level} (x : ⊤ {ℓ}) (y : ⊤ {ℓ′}) : Set e where
+
+------------------------------------------------------------------------
+-- Decidable Equality and Ordering
```
MatthewDaggitt: Decidability properties usually go in the `Properties` file. If we then put the remaining contents in the `Data.Unit.Polymorphic.Base` file, then `Data.Unit.Polymorphic` can export `Base` and the decidability proofs from `Properties`. I know it seems like overkill in this case, but I think it's best to be consistent on this...

src/Data/Unit/Polymorphic.agda
```diff
@@ -0,0 +1,38 @@
...
+------------------------------------------------------------------------
+-- A unit type defined as a record type
+
+record ⊤ {ℓ : Level} : Set ℓ where
+  constructor tt
+
+------------------------------------------------------------------------
+-- An ordering relation over the unit type
+
+record _≤_ {ℓ ℓ′ e : Level} (x : ⊤ {ℓ}) (y : ⊤ {ℓ′}) : Set e where
```
MatthewDaggitt: Interesting that you need the relation to live at its own level. I guess that makes sense as it can live below the levels of its arguments...

src/Data/Unit/Polymorphic.agda
JacquesCarette: I knew I was taking a chance on using 2 files instead of the 3 that `Data.Unit` uses for this. I'll split it up.

src/Data/Unit/Polymorphic.agda
JacquesCarette: Yep.  Inspired by the same possibility in categories.

src/Data/Unit/Polymorphic.agda
JacquesCarette: Done now.

WolframKahl:
On Wed, Oct 09, 2019 at 05:13:37PM -0700, Jacques Carette wrote:
> JacquesCarette commented on this pull request.
> 
> 
> 
> > +open import Relation.Binary
> +open import Relation.Binary.PropositionalEquality
> +
> +------------------------------------------------------------------------
> +-- A unit type defined as a record type
> +
> +record ⊤ {ℓ : Level} : Set ℓ where
> +  constructor tt
> +
> +------------------------------------------------------------------------
> +-- An ordering relation over the unit type
> +
> +record _≤_ {ℓ ℓ′ e : Level} (x : ⊤ {ℓ}) (y : ⊤ {ℓ′}) : Set e where
> +
> +------------------------------------------------------------------------
> +-- Decidable Equality and Ordering
> 
> Done now.

???

Since the only order on the unit type coincides with the only equivalence relation,
introducing separate names for them is, in my opinion, superfluous,
and using separate implementations for them will occasionally introduce friction,
in places where the fact that they are the same relation is actually used.

And the relation visible above is also actually not an order,
since it is not (necessarily) a homogeneous relation.

JacquesCarette:
I don't necessarily disagree, I was just following the 'lead' of `Data.Unit` which has the same code in it, just not level polymorphic.  I think it is worse for them to be inconsistent [and yes, I'd rather they both be consistently wrong!]

You might want to turn your comment into an Issue on the Tracker about `Data.Unit` itself. If there's agreement, I may well go ahead and fix them both at once.

MatthewDaggitt:
> You might want to turn your comment into an Issue on the Tracker about Data.Unit itself. If there's agreement, I may well go ahead and fix them both at once.

Agreed, I have no problem with deprecating `_≤_` and just using propositional equality everywhere, but it should be done consistently in a different issue :+1:

PR #928, opened 2019-10-03
[ fix #360 ] algebraic structures preserved via equivalence

### LINKED ISSUE #360: Prove stability of various structures under `Function.Equivalence`
Cf. https://stackoverflow.com/questions/36269333/agdas-standard-library-data-avl-sets-containing-data-string-as-values/36271248#36271248 for related code and motivation

gallais:


MatthewDaggitt:
Currently the library calls these proofs `replace-equality` and puts them in the `Properties.X` files, e.g. [lattice](https://github.com/agda/agda-stdlib/blob/3f9958326b7995f3167458d21b607379d1263b46/src/Algebra/Properties/Lattice.agda#L185). Not saying that this is the best option, but whatever we decide to do we should be consistent.

Furthermore, aren't these actually just special cases of `Algebra.Morphism.RawMagma/RawMonoid/RawGroup/RawSemiring/RawRing` where the morphism is `id`? (Okay so the last three don't exist yet, but we're getting there!)

gallais:
I think it is worth having a 'simple' version and one packaged up using
the homomorphism approach. In the same vein, `replace-equality` is using
`Function.Equivalence` with the trivial setoid induced by propositional
equality. It would be nice to have a version not mobilising this machinery.

MatthewDaggitt:
> I think it is worth having a 'simple' version and one packaged up using
the homomorphism approach.

Agree 100%. 

What are your reasons for not wanting to mobilise the homomorphism machinery? Efficiency or something else? It feels like we should have a good reason if we're going duplicate code!

gallais:
> What are your reasons for not wanting to mobilise the homomorphism machinery?

A misguided fear that they would mobilise setoids. But I have just checked
and they don't. So that's good!

MatthewDaggitt:
@gallais I'm happy to merge this in for `v1.2` and then we can replace the implementations with the morphism infrastructure at a later date. The main decision is where to put them. I'm not massively in favour of `Algebra.Structures.Properties` as that's not a standard file. I think it'd probably be more consistent to put it under `Algebra.Construct.ReplaceEquality`?

gallais:
Let's wait.

MatthewDaggitt:
I've found myself needing this so I've brought it up-to-date again. I've renamed the module to `Algebra.Construct.Subst.Equality` as that seems more appropriate and gives us room for other substitutions if we need to. I've also deprecated the existing `replace-equality` proofs in the `Algebra.Properties.X` modules.

> I think it is worth having a 'simple' version and one packaged up using
the homomorphism approach. In the same vein, replace-equality is using
Function.Equivalence with the trivial setoid induced by propositional
equality. It would be nice to have a version not mobilising this machinery.

(Unsuprisingly!) you were right about this. Using the homomorphism machinery is requires a proof of reflexivity for the equality which we don't have for the individual structures.

Issue #225, opened 2018-02-16
Adding ssreflect's `Reflects` idiom


gallais:
Following Phil's question on the Agda ML, I realised we could maybe benefit
from having `Reflects` in the standard library:

```agda
data Reflects (P : Set) : Bool → Set where
  yes : (p : P)    → Reflects P true
  no  : (¬p : ¬ P) → Reflects P false
```

Sample properties:

```agda
map : ∀ {P Q} → (P → Q) → (Q → P) → ∀ {b} → Reflects P b → Reflects Q b
map p⇒q q⇒p (yes p) = yes (p⇒q p)
map p⇒q q⇒p (no ¬p) = no (λ q → ¬p (q⇒p q))

dec-Reflects : ∀ {b P} → Reflects P b → Dec P
dec-Reflects (yes p) = yes p
dec-Reflects (no ¬p) = no ¬p
```

Typical use-case:

```agda
_≤ᵇ_ : ℕ → ℕ → Bool
zero ≤ᵇ n = true
suc m ≤ᵇ zero = false
suc m ≤ᵇ suc n = m ≤ᵇ n

s≤s-inj : ∀ {m n} → suc m ≤ suc n → m ≤ n
s≤s-inj (s≤s p) = p

≤?⇔≤ᵇ : ∀ m n → Reflects (m ≤ n) (m ≤ᵇ n)
≤?⇔≤ᵇ zero    n       = yes z≤n
≤?⇔≤ᵇ (suc m) zero    = no (λ ())
≤?⇔≤ᵇ (suc m) (suc n) = map s≤s s≤s-inj (≤?⇔≤ᵇ m n)
```

MatthewDaggitt:
Do we want to encourage further programming with `Bool`? I feel like we've been making progress against that programming style recently. 

In my opinion your use case isn't really a use case as `_≤ᵇ_` should never be defined!

gallais:
In [this blog post](https://personal.cis.strath.ac.uk/james.wood.100/blog/html/VecMat.html) @laMudri makes the case that it can sometimes be interesting
computationally to use the `Bool`-variant of a decidability proof.

In which case, being able to easily relate it to its proof-carrying counterpart
cannot hurt.

MatthewDaggitt:
Fair enough! I've definitely feel less strongly about the point than I did last year. (As I remember I was in the middle of trying to prove properties about the horrendous boolean-based `filter` functions)

laMudri:
[...]

PR #929, opened 2019-10-03
[ see #225 ] Redefine `Dec` for better computation

laMudri:
Motivation
===

As proposed [here-ish](https://github.com/agda/agda-stdlib/issues/225#issuecomment-525622181), this pull request redefines `Dec` in terms of `Reflects`. The actual series of definitions is the following (`Relation.Nullary`).

```agda
data Reflects {p} (P : Set p) : Bool → Set p where
  true  : ( p :   P) → Reflects P true
  false : (¬p : ¬ P) → Reflects P false

record Dec {p} (P : Set p) : Set p where
  constructor dec
  field
    isYes : Bool
    reflects : Reflects P isYes

open Dec public

pattern yes p = dec true (true p)
pattern no ¬p = dec false (false ¬p)
```

`Reflects P b` is the graph of `if b then P else ¬ P`. A `Dec` value contains a boolean `isYes`, along with a proof that `isYes` reflects the proposition `P` that we're deciding. We are able to branch on `isYes` without pattern matching on `reflects`, which is where the computational advantages of this definition come from.

With the pattern declarations, pretty much everything in the standard library still works if the only change we make is the redefinition of `Dec`. However, this way, all definitions are as strict as they were before, and we gain no benefit.

We can often gain something in decidable equality proofs. For example, consider a standard current definition of decidable equality on natural numbers, and how it interacts with the current `isYes` (recently renamed from `⌊_⌋`). The definition is equivalent to:

```agda
_≟_ : Decidable {A = ℕ} _≡_
zero ≟ zero = yes refl
zero ≟ suc n = no λ ()
suc m ≟ zero = no λ ()
suc m ≟ suc n with m ≟ n
... | yes p = yes (cong suc p)
... | no ¬p = no (¬p ∘ suc-injective)
```

In the `suc`-`suc` case, before we can say what the result is, we must be able to get the full proof coming from `m ≟ n`. We can try getting an answer:

```agda
run : (m n : ℕ) → Set
run m n = {!isYes (3 + m ≟ 2 + n)!}
```

But normalising this expression does not give us anything much simpler.

In contrast, we can now make the following definition also.

```agda
_≟_ : Decidable {A = ℕ} _≡_
zero ≟ zero = yes refl
zero ≟ suc n = no λ ()
suc m ≟ zero = no λ ()
isYes (suc m ≟ suc n) = isYes (m ≟ n)
reflects (suc m ≟ suc n) with m ≟ n
... | yes p = true (cong suc p)
... | no ¬p = false (¬p ∘ suc-injective)
```

Now, we can follow the test case again, and it normalises to `isYes (1 + m ≟ n)`. This behaviour can be useful for functions like `filter` from `Data.List`, which require a decidable proposition but don't actually use the proposition part.

Work done
===

The most interesting changes are in the `Relation.Nullary` hierarchy. In `Relation.Nullary.Decidable.Core`, the closest equivalent to the old `isYes` is `isYes!`. As documented in the file, `isYes` is intended for branching, whereas `isYes!` is intended for proof automation (assuming concrete values). `Relation.Nullary.Reflects` has been introduced, containing some inversion lemmas. The proofs of decidability for logical connectives (`¬?`, `_×-dec_`, `_⊎-dec_`, and `_→-dec_`) have been rewritten to make the connection to the corresponding boolean functions explicit.

Elsewhere, functions branching on decisions (like `filter`) and recursive decision procedures have been rewritten to be less strict. A lot of the time for decision procedures, the tactic has been to rewrite `with` clauses to some combination of `map′` and `_×-dec_`, which would have been a stylistic improvement with the current definition of `Dec`. In addition, lemmas about decision-using functions have also been made less strict, which seems more idiomatic.

I don't have any good way to know whether I've caught every overly strict usage of `Dec` in the library. I have so far been guided by grepping for `yes  *_` and `no  *_`, which suggest only the branching is important and not the proofs. There are probably more things that could be improved. The `Everything` file checks successfully, so at least nothing that is used in the library is broken (proof automation, for example).

### DIFF
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -20,7 +20,7 @@ open import Function.Equivalence
 open import Level using (Level; 0ℓ)
 open import Relation.Binary
 open import Relation.Binary.PropositionalEquality hiding ([_])
-open import Relation.Nullary using (yes; no)
+open import Relation.Nullary using (Dec; isYes; reflects; yes; no; true; false)
 open import Relation.Nullary.Decidable using (True)
 import Relation.Unary as U
 
@@ -626,8 +626,9 @@ T-irrelevant : U.Irrelevant T
 T-irrelevant {true}  _  _  = refl
 
 T? : U.Decidable T
-T? true  = yes _
-T? false = no (λ ())
+isYes (T? x) = x
+reflects (T? false) = false id
+reflects (T? true ) = true  _
 
 T?-diag : ∀ b → T b → True (T? b)
 T?-diag true  _ = _
--- a/src/Relation/Nullary.agda
+++ b/src/Relation/Nullary.agda
@@ -12,6 +12,7 @@ module Relation.Nullary where
 
 open import Agda.Builtin.Equality
 
+open import Data.Bool.Base
 open import Data.Empty hiding (⊥-elim)
 open import Data.Empty.Irrelevant
 open import Level
@@ -23,11 +24,29 @@ infix 3 ¬_
 ¬_ : ∀ {ℓ} → Set ℓ → Set ℓ
 ¬ P = P → ⊥
 
--- Decidable relations.
+-- `Reflects` idiom.
+-- The truth value of P is reflected by a boolean value.
 
-data Dec {p} (P : Set p) : Set p where
-  yes : ( p :   P) → Dec P
-  no  : (¬p : ¬ P) → Dec P
+data Reflects {p} (P : Set p) : Bool → Set p where
+  true  : ( p :   P) → Reflects P true
+  false : (¬p : ¬ P) → Reflects P false
+
+-- Decidable relations.
+-- This version of `Dec` allows the boolean portion of the
+-- value to compute independently from the proof portion.
+-- This often allows good computational properties when we
+-- only care about the boolean portion.
+
+record Dec {p} (P : Set p) : Set p where
+  constructor dec
+  field
+    isYes : Bool
+    reflects : Reflects P isYes
+
+open Dec public
+
+pattern yes p = dec true (true p)
+pattern no ¬p = dec false (false ¬p)
 
 -- Given an irrelevant proof of a decidable type, a proof can
 -- be recomputed and subsequently used in relevant contexts.
--- a/src/Category/Monad/Partiality.agda
+++ b/src/Category/Monad/Partiality.agda
@@ -910,9 +910,9 @@ private
   -- McCarthy's f91:
 
   f91′ : ℕ → ℕ ⊥P
-  f91′ n with n ≤? 100
-  ... | yes _ = later (♯ (f91′ (11 + n) >>= f91′))
-  ... | no  _ = now (n ∸ 10)
+  f91′ n with isYes (n ≤? 100)
+  ... | true  = later (♯ (f91′ (11 + n) >>= f91′))
+  ... | false = now (n ∸ 10)
 
   f91 : ℕ → ℕ ⊥
   f91 n = ⟦ f91′ n ⟧P
--- a/src/Codata/Cofin/Literals.agda
+++ b/src/Codata/Cofin/Literals.agda
@@ -17,7 +17,7 @@ open import Relation.Nullary.Decidable
 
 number : ∀ n → Number (Cofin n)
 number n = record
-  { Constraint = λ k → True (suc k ℕ≤? n)
-  ; fromNat    = λ n {{p}} → fromℕ< (toWitness p)
+  { Constraint = λ k → True! (suc k ℕ≤? n)
+  ; fromNat    = λ n {{p}} → fromℕ< (toWitness! p)
   }
 
--- a/src/Data/Bool/Base.agda
+++ b/src/Data/Bool/Base.agda
@@ -11,7 +11,6 @@ module Data.Bool.Base where
 open import Data.Unit.Base using (⊤)
 open import Data.Empty
 open import Level using (Level)
-open import Relation.Nullary
 
 private
   variable
--- a/src/Data/Char/Properties.agda
+++ b/src/Data/Char/Properties.agda
@@ -16,7 +16,7 @@ import Data.Nat.Properties as ℕₚ
 
 open import Function
 open import Relation.Nullary using (yes; no)
-open import Relation.Nullary.Decidable using (map′; isYes)
+open import Relation.Nullary.Decidable using (map′; isYes!)
 open import Relation.Binary
   using ( _⇒_; Reflexive; Symmetric; Transitive; Substitutive
         ; Decidable; IsEquivalence; IsDecEquivalence
@@ -103,7 +103,7 @@ x ≟ y = map′ ≈⇒≡ ≈-reflexive (x ≈? y)
 
 infix 4 _==_
 _==_ : Char → Char → Bool
-c₁ == c₂ = isYes (c₁ ≟ c₂)
+c₁ == c₂ = isYes! (c₁ ≟ c₂)
 
 private
 
--- a/src/Data/Digit.agda
+++ b/src/Data/Digit.agda
@@ -8,6 +8,7 @@
 
 module Data.Digit where
 
+open import Data.Bool.Base using (if_then_else_)
 open import Data.Nat
 open import Data.Nat.Properties
 open import Data.Nat.Solver
@@ -18,7 +19,7 @@ open import Data.Product
 open import Data.Vec as Vec using (Vec; _∷_; [])
 open import Data.Nat.DivMod
 open import Data.Nat.Induction
-open import Relation.Nullary using (yes; no)
+open import Relation.Nullary using (isYes)
 open import Relation.Nullary.Decidable
 open import Relation.Binary using (Decidable)
 open import Relation.Binary.PropositionalEquality as P using (_≡_; refl)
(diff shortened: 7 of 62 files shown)


WolframKahl:
On Thu, Oct 03, 2019 at 11:35:13AM -0700, James Wood wrote:
> We are able to branch on `isYes` without pattern matching on `reflects`, which is where the computational advantages of this definition come from.

Which “computational advantages...”?
Since branching on `isYes` is, in principle, pattern matching on Bool,
it will depend on further implementation details whether there really are any advantages.

Have you benchmarked the difference?


Wolfram

UlfNorell:
Basically any proof-by-reflection example where you are not interested in the proof will show huge differences. It's the difference between running the decision procedure and matching on the bool and normalising the soundness proof of the decision procedure.

laMudri:
> Which “computational advantages...”?

I should've been clearer that my original motivation was to improve definitional equality when using decision procedures only for the yes/no result. What Ulf says should also be true, though I haven't benchmarked it. I've made sure that `isYes!` only relies on the boolean, so we can avoid all the deep copying usually required with `Dec`.

The original motivating example for me is mostly explained in [this section](https://personal.cis.strath.ac.uk/james.wood.100/blog/html/VecMat.html#behind-the-scenes) of my blog post about vectors and matrices. Briefly, the setup is that `Matrix m n = Fin m → Fin n → Carrier`, with `Carrier` being from a semiring. If one defines the identity matrix `1ₘ : ∀ {n} → Matrix n n` via decidable equality on `Fin n`, then `1ₘ (suc i) (suc j)` does not compute to `1ₘ i j`, which would often be useful. With the new `Dec`, and factoring through `isYes`, we do get that `1ₘ (suc i) (suc j)` is definitionally equal to `1ₘ i j`. The full definition is `1ₘ i j = if isYes (i ≟ j) then 1# else 0#`.

laMudri:
I've also planted an example [in Data.Fin.Properties](https://github.com/agda/agda-stdlib/pull/929/files#diff-3d30f647d768c855434815d99283082eR536) showing that the boolean content of `all?` (which decides whether a predicate on `Fin n` holds for all of `0..n-1`) now normalises to the boolean function you'd expect. As such, it is easy to use and reason about in algorithms. This was impossible with the previous definition of `Dec`.

MatthewDaggitt:
I'm on board with the idea in principle, but I'm a bit scared by the commit named `[fixup] everything that broke` the contents of which seems to be rather large. In #225 you seemed to be suggesting that everything was going to be backwards compatible?

Do you think you could separate the PRs out into multiple PRs?
 1. Changing the definition in an entirely backwards compatible manner?
 2. Make any non-backwards compatible changes
 3. Make any changes to use the new definitions 

I'm also a bit worried about how easy this is going to be to understand for newcomers to the library. We're probably going to have to improve the documentation.

laMudri:
That commit basically does two things:

1. Switch from `isYes` to the stricter version `isYes!` where it is used for proof automation. See, for example, the [changes in `Data.Nat.DivMod`](https://github.com/agda/agda-stdlib/pull/929/commits/13a8aec744142d09ff6e418083c458c731794f34#diff-49f9fc88a6cb39a4d03f8f022d181996).
2. Resolve name clashes, particularly between the constructor `dec` and identifiers with the same name.

Looking back, maybe some of (2) should be worked around differently. Particularly, I renamed some declarations called `dec` to `decidable`, which is not backwards-compatible. There is some precedent for `decidable`, so maybe this is something to clear up globally and manage via deprecation. Maybe the constructor could have a different name, but I can't think of a better one.

In (2), I also did a few refactorings where I introduce a `module _` for a parameter which is constant across recursive calls. This bumps up the edit distance a lot, but does not constitute a major change.

As for (1), the transition could be done in a more backwards-compatible way, but my feeling is that that would be at the expense of good names going forward. Here's a table of what has happened:

```
 v1.1   | master | proposed | alternative | good for...
--------+--------+----------+-------------+------------------
 (none) | (none) | isYes    | isTrue      | branching
 ⌊_⌋    | isYes  | isYes!   | isYes       | proof automation
```

The `alternative` column is what I had been using as working names, but I think they are a bit rubbish. Their advantage is that everything is backwards-compatible, but we would still aim to rename a lot of the `isYes` occurrences in the library to `isTrue`. I like the names this PR has now, because it's clear that `isYes` and `isYes!` are doing the same thing, but the `!` shows that the latter is stricter, and thus only used in special circumstances. Additionally, it can be seen as an advantage that things break, because it tells us exactly where the strict version is required, rather than us having to guess where we can get away with the non-strict version.

I can make the first of the proposed smaller pull requests when we have a name for the `isYes` field that doesn't clash with the existing `isYes` definition. It's quite tempting to use `⌊_⌋`, but I guess this is still undergoing deprecation. I guess I'd also remove or rename the `dec` constructor so that I don't have to go through (2).

src/Relation/Nullary.agda
```diff
@@ -23,11 +24,29 @@ infix 3 ¬_
...
+
 -- Decidable relations.
+-- This version of `Dec` allows the boolean portion of the
+-- value to compute independently from the proof portion.
+-- This often allows good computational properties when we
+-- only care about the boolean portion.
+
+record Dec {p} (P : Set p) : Set p where
+  constructor dec
+  field
```
gallais: For the constructor name, may I suggest `mkDec`? Or maybe `_because_`?

src/Data/Bool/Properties.agda
```diff
@@ -626,8 +626,9 @@ T-irrelevant : U.Irrelevant T
 T-irrelevant {true}  _  _  = refl
 
 T? : U.Decidable T
-T? true  = yes _
-T? false = no (λ ())
+isYes (T? x) = x
+reflects (T? false) = false id
+reflects (T? true ) = true  _
```
gallais: Could we have a smart constructor for this kind of pattern?
Something like (we probably need a different name):
`if_then_else_ : (b : Bool) -> P true -> P false -> Dec P`

src/Relation/Nullary.agda
laMudri: Yeah, I guess `mkDec` would work fine. I'm worried about the name getting long, because `dec false _` appears quite a lot and is already very long. I don't think `_because_` fits exactly; `true because (true ...)` doesn't read well.

MatthewDaggitt:
> Particularly, I renamed some declarations called dec to decidable, which is not backwards-compatible.

So the choice of names seem to be one of the major issues currently preventing backwards compatibility. Running with @gallais's suggestion of `_because_`, how about the following names?
```agda
data Reflects {p} (P : Set p) : Bool → Set p where
  of₁_ : ( p :   P) → Reflects P true
  of₂_ : (¬p : ¬ P) → Reflects P false

record Dec {p} (P : Set p) : Set p where
  constructor _because_
  field
    is : Bool
    proof : Reflects P is

open Dec public

pattern yes p = true because of₁ p
pattern no ¬p = false because of₂ ¬p
```
and "advanced" decidability proofs then look like:
```
_≟_ : Decidable {A = ℕ} _≡_
zero ≟ zero = yes refl
zero ≟ suc n = no λ ()
suc m ≟ zero = no λ ()
is (suc m ≟ suc n) = is (m ≟ n)
proof (suc m ≟ suc n) with m ≟ n
... | yes p = true (cong suc p)
... | no ¬p = false (¬p ∘ suc-injective)
```
Maybe someone can think of a way of avoiding the subscript numbers for `of_`...

Regardless of the names we go with, this change still has the _potential_ to break a ton of code. I'm afraid I'm going to insist on an initial PR which substitutes these definitions in whilst changing the absolute bare minimum (i.e. hopefully nothing!) outside of `Relation.Nullary` and `Relation.Nullary.Decidable`. A second PR can then alter the existing decidability proofs to be more efficient.

MatthewDaggitt:
Actually maybe substituting `is` for `does` results in even better readability:
```
_≟_ : Decidable {A = ℕ} _≡_
zero ≟ zero = yes refl
zero ≟ suc n = no λ ()
suc m ≟ zero = no λ ()
does (suc m ≟ suc n) = does (m ≟ n)
proof (suc m ≟ suc n) with m ≟ n
... | yes p = of₁ (cong suc p)
... | no ¬p = of₂ (¬p ∘ suc-injective)
```

gallais:
> I'm going to insist on an initial PR which substitutes these definitions in whilst changing the absolute bare minimum (i.e. hopefully nothing!) outside of Relation.Nullary and Relation.Nullary.Decidable

Given that `isYes` has not been release yet, I think it should be OK to break it.

MatthewDaggitt:
> Given that isYes has not been release yet, I think it should be OK to break it.

It's not that I'm worried about. It's breaking everyone's decidability proofs in some subtle way, e.g. introducing unsolved metas.

laMudri:
These seem workable. I'll get a new pull request out soon.

It's annoying that we don't have subscript ‘f’, but how about `ofᵗ` and `ofᶠ`?

laMudri:
Alternatively maybe `ofʸ` and `ofⁿ`, so that we at least have some visual distinction.

MatthewDaggitt:
> These seem workable. I'll get a new pull request out soon.

Thanks!

> It's annoying that we don't have subscript ‘f’, but how about ofᵗ and ofᶠ? Alternatively maybe ofʸ and ofⁿ, so that we at least have some visual distinction.

Good idea. Either sounds good to me.

Also for the 2nd PR it occurs to me that the last clauses of the proof should probably match on `proof` right? i.e.
```agda
_≟_ : Decidable {A = ℕ} _≡_
zero ≟ zero = yes refl
zero ≟ suc n = no λ ()
suc m ≟ zero = no λ ()
does (suc m ≟ suc n) = does (m ≟ n)
proof (suc m ≟ suc n) with proof (m ≟ n)
... | of₁ p = of₁ (cong suc p)
... | of₂ ¬p = of₂ (¬p ∘ suc-injective)
```

JacquesCarette:
I came here to say I liked `_because_`, but then read on -- `does` and `proof` with `of` works quite nicely too.

laMudri:
> the last clauses of the proof should probably match on `proof` right?

Turns out that example doesn't check. `does (m ≟ n)` doesn't get abstracted, so we can't do the pattern matching on `proof (m ≟ n) : Reflects (m ≡ n) (does (m ≟ n))`.

MatthewDaggitt:
Closing as superseded by #932 and #935

PR #932, opened 2019-10-05
Add `Reflects` and define `Dec` in terms of it

laMudri:
This pull request contains the first steps of #929. It introduces `Reflects` and the module `Relation.Nullary.Reflects`, which at the moment just contains lemmas converting between `Reflects P b` and `if b then P else ¬ P`. It redefines `Dec` to contain a boolean value and a `Reflects` proof of it. There are patterns `yes` and `no`, which have behaviour similar to the old constructors of `Dec`. The rest of the library checks without modification. Finally, a bit about all of this has been added to the changelog.

I think with this new set of names (in particular `_because_` and `does`), breaking changes should not be necessary. Some of the existing names may look a bit strange (`isYes`, `True`, `toWitness`, and the corresponding negative definitions are more special-purpose than their names suggest), but they can be dealt with via deprecation if necessary. This leaves just a few more tasks to do to catch up to #929:

1. Reprove `map′` in `Relation.Nullary.Decidable` so that the `does` field is a pass-through.
2. Reprove the decidability of logical connectives in `Relation.Nullary.*` so that the `does` field uses named boolean functions. Maybe there should also be `Reflects` proofs, for example, `×-reflects : Reflects P x → Reflects Q y → Reflects (P × Q) (x ∧ y)` in `Relation.Nullary.Product`. If `×-reflects` exists, `_×-dec_` can then use it, in a relationship similar to `isMonoid` and `monoid` in various places.
3. Reprove recursive decision procedures throughout the library (mostly in `Data`). This mostly (but not entirely) consists of rephrasing `with` clauses to applications of `map′`, possibly on `_×-dec_`, `_⊎-dec_`, &c, which could be done independently of even this pull request.
4. Reimplement functions which use `Dec` values for the boolean content only, e.g. `filter`.
5. Reprove lemmas to be only as strict as necessary. This is least important, unless someone is going to depend computationally on these proofs.

All of these changes should only improve definitional equality while not causing any new unification issues or name clashes (this was my experience in #929), so should be safe to do.

### DIFF
--- a/src/Relation/Nullary/Reflects.agda
+++ b/src/Relation/Nullary/Reflects.agda
@@ -1,0 +1,28 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of the `Reflects` construct
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Nullary.Reflects where
+
+open import Data.Bool.Base
+open import Level
+open import Relation.Nullary
+
+variable
+  p : Level
+  P : Set p
+
+------------------------------------------------------------------------
+-- `Reflects P b` is equivalent to `if b then P else ¬ P`.
+
+of : ∀ {b} → if b then P else ¬ P → Reflects P b
+of {b = false} ¬p = ofⁿ ¬p
+of {b = true }  p = ofʸ p
+
+invert : ∀ {b} → Reflects P b → if b then P else ¬ P
+invert (ofʸ  p) = p
+invert (ofⁿ ¬p) = ¬p
--- a/src/Relation/Nullary/Reflects.agda
+++ b/src/Relation/Nullary/Reflects.agda
@@ -9,7 +9,9 @@
 module Relation.Nullary.Reflects where
 
 open import Data.Bool.Base
+open import Data.Empty
 open import Level
+open import Relation.Binary.PropositionalEquality
 open import Relation.Nullary
 
 variable
@@ -32,3 +34,13 @@ of {b = true }  p = ofʸ p
 invert : ∀ {b} → Reflects P b → if b then P else ¬ P
 invert (ofʸ  p) = p
 invert (ofⁿ ¬p) = ¬p
+
+------------------------------------------------------------------------
+-- Other lemmas
+
+-- `Reflects` is deterministic.
+det : ∀ {b b′} → Reflects P b → Reflects P b′ → b ≡ b′
+det (ofʸ  p) (ofʸ  p′) = refl
+det (ofʸ  p) (ofⁿ ¬p′) = ⊥-elim (¬p′ p)
+det (ofⁿ ¬p) (ofʸ  p′) = ⊥-elim (¬p p′)
+det (ofⁿ ¬p) (ofⁿ ¬p′) = refl
--- a/src/Relation/Nullary/Reflects.agda
+++ b/src/Relation/Nullary/Reflects.agda
@@ -19,6 +19,12 @@ variable
 ------------------------------------------------------------------------
 -- `Reflects P b` is equivalent to `if b then P else ¬ P`.
 
+-- These lemmas are intended to be used mostly when `b` is a value, so
+-- that the `if` expressions have already been evaluated away.
+-- In this case, `of` works like the relevant constructor (`ofⁿ` or
+-- `ofʸ`), and `invert` strips off the constructor to just give either
+-- the proof of `P` or the proof of `¬ P`.
+
 of : ∀ {b} → if b then P else ¬ P → Reflects P b
 of {b = false} ¬p = ofⁿ ¬p
 of {b = true }  p = ofʸ p
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,47 @@ from `_↞_` to `_↩_` in order to make room for the new package for right inve
 * The existing modules `Data.Bin` and `Data.Bin.Properties` still exist but have been
   deprecated and may be removed in some future release of the library.
 
+### Addition of the `Reflects` idiom
+
+* A version of the `Reflects` idiom, as seen in SSReflect, has been introduced
+in `Relation.Nullary`. Some properties of it have been added in
+`Relation.Nullary.Reflects`. The definition is as follows
+  ```agda
+  data Reflects {p} (P : Set p) : Bool → Set p where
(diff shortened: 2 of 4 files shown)


src/Relation/Nullary/Reflects.agda
```diff
@@ -0,0 +1,28 @@
...
+open import Relation.Nullary
+
+variable
+  p : Level
+  P : Set p
+
+------------------------------------------------------------------------
+-- `Reflects P b` is equivalent to `if b then P else ¬ P`.
+
+of : ∀ {b} → if b then P else ¬ P → Reflects P b
```
MatthewDaggitt: I would probably call this `if-reflects` and the proof below `reflects-if`

MatthewDaggitt:
Thanks for making this! Looks good to me :+1:

src/Relation/Nullary/Reflects.agda
laMudri: I think maybe the comment is what is wrong here, rather than the names. I only see these two functions being used when `b` is a value, so the user never actually sees an `if` in their goal. I have some examples of `invert` [here](https://github.com/agda/agda-stdlib/pull/929/commits/99b8729b7c830bb424cfeead3b341c2a7fc724e4#diff-3d30f647d768c855434815d99283082eR505-R519), where it is used to get the proof out of a `Reflects` value while being as unstrict as possible. We know, based on which branch we're in, whether `r` has type `Reflects (Q zero) false` or `Reflects (Q zero) true`, and we use `invert` to get out either a proof of `¬ Q zero` or of `Q zero`. I've also got an example [here](https://github.com/agda/agda-stdlib/pull/929/commits/99b8729b7c830bb424cfeead3b341c2a7fc724e4#diff-f720b1ed09bcf5d567bb95b6f63dafa1R49-R55) where I try it out mostly as a stylistic thing.

As for `of`, the potential use is to replace any instance of `ofʸ` or `ofⁿ` in expressions. Again, whether `does` is `true` or `false` will be known. A way to see it is that the `if` and the `b` control a form of overloading.

MatthewDaggitt: Ah okay, that makes sense then. Thanks for the explanation!

gallais:
Is it not the case that the pattern synonyms `yes` and `no` provide 100%
backwards compatibility?

laMudri:
Note that stdlib still checks on this branch, and the entire changes are:

 - add `Reflects` and module `Relation.Nullary.Reflects`
 - redefine `Dec`
 - remove one redundant import from `Data.Bool.Base`
 - changelog

In particular, every decision procedure in stdlib still works with the new definition, entirely without modification. In previous iterations, some minor changes were needed, but with the right system of records this is now fine. There is an argument that automatic case-splitting on `Dec` now no longer does the same simple thing, but that's only an issue for new code.

A general principle with this bunch of changes is that the new definitions should allow us to be more precise, but never force the unexpected updating of old code to be similarly precise. For example, if I reimplement `filter` to only do matching on the boolean value, then all of the old proofs about `filter` that pattern-match via `yes` and `no` will still work.

MatthewDaggitt:
@andreasabel are you convinced by the arguments that it's backwards compatible or have we missed something?

laMudri:
Any chance we could work out any remaining issues before Wednesday afternoon? I'm giving a talk about this work at SPLS, and it would be nice to have this major part merged by then.

src/Relation/Nullary/Reflects.agda
```diff
@@ -32,3 +34,13 @@ of {b = true }  p = ofʸ p
...
+
+------------------------------------------------------------------------
+-- Other lemmas
+
+-- `Reflects` is deterministic.
+det : ∀ {b b′} → Reflects P b → Reflects P b′ → b ≡ b′
+det (ofʸ  p) (ofʸ  p′) = refl
+det (ofʸ  p) (ofⁿ ¬p′) = ⊥-elim (¬p′ p)
+det (ofⁿ ¬p) (ofʸ  p′) = ⊥-elim (¬p p′)
+det (ofⁿ ¬p) (ofⁿ ¬p′) = refl
```
gallais: You could reuse `Injective` here.
This type of lemma is usually called `X-inj`.

MatthewDaggitt:
> Any chance we could work out any remaining issues before Wednesday afternoon? I'm giving a talk about this work at SPLS, and it would be nice to have this major part merged by then.

Sure, no problem. If @andreasabel hasn't replied by tomorrow evening, I'll merge it in.

mietek:
> Is it not the case that the pattern synonyms `yes` and `no` provide 100%
backwards compatibility?

Apparently not. When `yes` and `no` were constructors, it was possible to overload these names for constructors of other types. Now that `yes` and `no` are pattern synonyms, this is no longer possible.

```
Multiple definitions of yes. Previous definition at
/usr/local/Cellar/agda/2.6.1/lib/agda/src/Relation/Nullary.agda:57,9-12
  data RF? {n} where
    yes : ∀ {e} → RF e → RF? e
    no : ∀ {e} → WHNF e → RF? e
```

MatthewDaggitt:
Hmm, that is unfortunate. Apologies that we missed that. Has it proved a major inconvenience?

mietek:
Minor, thanks. It would be nice if Agda treated pattern synonyms the same as constructors, with respect to overloading.

laMudri:
Yet another argument for typed pattern synonyms...

gallais:
Next AIM? :angel:

mietek:
There’s finally an issue to track the proposed improvement to Agda:
https://github.com/agda/agda/issues/7915

Meanwhile, regarding:

> This can have advantages for both performance and reasoning.

Has anyone shown that this change does indeed have advantages for performance or reasoning?

gallais:
@laMudri has a blog post explaining the motivation behind these changes:
https://web.archive.org/web/20230407101512/https://personal.cis.strath.ac.uk/james.wood.100/blog/html/VecMat.html

mietek:
> @laMudri has a blog post explaining the motivation behind these changes: https://web.archive.org/web/20230407101512/https://personal.cis.strath.ac.uk/james.wood.100/blog/html/VecMat.html

Thanks for the reference.  However, I’m unable to find an answer to my question in this blog post; the most relevant part seems to be this sentence:

> Perhaps we should have a type `EDec` for extrinsic decidability, complementing intrinsic decidability from `Dec`.

jamesmckinna:
See also #2741 revisiting this design...

PR #935, opened 2019-10-06
[ refactor ] change decision procedures to use more combinators

laMudri:
This pull request addresses point 3 of #932. Specifically, I've tried to find all instances where a decision procedure could be stated using `map′`, `_×-dec_`, `_⊎-dec_`, &c, and stated them that way. For example, suppose we started with:

```agda
  any : Decidable P → Decidable (Any P)
  any P? []       = no λ()
  any P? (x ∷ xs) with P? x
  ... | yes px = yes (here px)
  ... | no ¬px with any P? xs
  ...   | yes pxs = yes (there pxs)
  ...   | no ¬pxs = no λ { (here px) → ¬px px
                         ; (there pxs) → ¬pxs pxs
                         }
```

This would become:

```agda
  any : Decidable P → Decidable (Any P)
  any P? []       = no λ()
  any P? (x ∷ xs) =
    Dec.map′ [ here , there ]′ (λ { (here px) → inj₁ px
                                  ; (there pxs) → inj₂ pxs
                                  })
             (P? x ⊎-dec any P? xs)
```

In some cases, there are generally useful existing or new lemmas to fill the two directions of the `map′`. In the case of the example, we actually have

```agda
  toSum : Any P (x ∷ xs) → P x ⊎ Any P xs
  fromSum : P x ⊎ Any P xs → Any P (x ∷ xs)
```

This cuts the decision procedure down to just

```agda
  any : Decidable P → Decidable (Any P)
  any P? []       = no λ()
  any P? (x ∷ xs) = Dec.map′ fromSum toSum (P? x ⊎-dec any P? xs)
```

All of these changes should be entirely backwards compatible, and for now not change the definitional behaviour of these procedures at all. The view is towards changing the implementation of `Dec` and the combinators, at which point the behaviour will be better. Nevertheless, I think these changes can be defended just from a stylistic point of view, in that they make clearer the logical operations that happen in the procedure.

Not in scope are complex procedures like `decFinSubset` from `Data.Fin.Properties`, which will require a more thorough rewrite to get the desired behaviour on the `does` field of new `Dec`.

A potential extension is to use `map` instead of `map′`, and state as lemmas all of the equivalences that get used by `map`. For example, we should be able to roll together `toSum` and `fromSum` into a proof of `Any P (x ∷ xs) ⇔ P x ⊎ Any P xs` (and I suppose even `Any P (x ∷ xs) ↔ P x ⊎ Any P xs`).

### DIFF
--- a/src/Data/Fin/Subset/Properties.agda
+++ b/src/Data/Fin/Subset/Properties.agda
@@ -21,7 +21,7 @@ open import Data.Fin.Properties using (any?; decFinSubset)
 open import Data.Nat.Base using (ℕ; zero; suc; z≤n; s≤s; _≤_)
 open import Data.Nat.Properties using (≤-step)
 open import Data.Product as Product using (∃; ∄; _×_; _,_)
-open import Data.Sum as Sum using (_⊎_; inj₁; inj₂)
+open import Data.Sum as Sum using (_⊎_; inj₁; inj₂; [_,_]′)
 open import Data.Vec
 open import Data.Vec.Properties
 open import Function.Core using (_∘_; const; id; case_of_)
@@ -29,8 +29,10 @@ open import Function.Equivalence using (_⇔_; equivalence)
 open import Relation.Binary as B hiding (Decidable)
 open import Relation.Binary.PropositionalEquality
   using (_≡_; refl; cong; cong₂; subst; isEquivalence)
-open import Relation.Nullary.Negation using (contradiction)
 open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Decidable using (map′)
+open import Relation.Nullary.Negation using (contradiction)
+open import Relation.Nullary.Sum using (_⊎-dec_)
 open import Relation.Unary using (Pred; Decidable)
 
 ------------------------------------------------------------------------
@@ -49,9 +51,7 @@ infix 4 _∈?_
 _∈?_ : ∀ {n} x (p : Subset n) → Dec (x ∈ p)
 zero  ∈? inside  ∷ p = yes here
 zero  ∈? outside ∷ p = no  λ()
-suc n ∈? s       ∷ p with n ∈? p
-... | yes n∈p = yes (there n∈p)
-... | no  n∉p = no  (n∉p ∘ drop-there)
+suc n ∈? s       ∷ p = map′ there drop-there (n ∈? p)
 
 ------------------------------------------------------------------------
 -- Empty
@@ -152,13 +152,12 @@ x∈⁅y⁆⇔x≡y {_} {x} {y} = equivalence
 infix 4 _⊆?_
 _⊆?_ : ∀ {n} → B.Decidable (_⊆_ {n = n})
 []          ⊆? []          = yes id
-outside ∷ p ⊆? y ∷ q with p ⊆? q
-... | yes p⊆q = yes λ { (there v∈p) → there (p⊆q v∈p)}
-... | no  p⊈q = no (p⊈q ∘ drop-∷-⊆)
+outside ∷ p ⊆? y ∷ q =
+  map′ (λ { p⊆q (there v∈p) → there (p⊆q v∈p) }) drop-∷-⊆ (p ⊆? q)
 inside  ∷ p ⊆? outside ∷ q = no (λ p⊆q → case (p⊆q here) of λ())
-inside  ∷ p ⊆? inside  ∷ q with p ⊆? q
-... | yes p⊆q = yes λ { here → here ; (there v) → there (p⊆q v)}
-... | no  p⊈q = no (p⊈q ∘ drop-∷-⊆)
+inside  ∷ p ⊆? inside  ∷ q =
+  map′ (λ { p⊆q here → here ; p⊆q (there v∈p) → there (p⊆q v∈p) }) drop-∷-⊆
+       (p ⊆? q)
 
 module _ (n : ℕ) where
 
@@ -576,15 +575,11 @@ Lift? P? p = decFinSubset (_∈? p) (λ {x} _ → P? x)
 -- Other
 
 anySubset? : ∀ {n} {P : Subset n → Set} → Decidable P → Dec (∃ P)
-anySubset? {zero}  P? with P? []
-... | yes P[] = yes (_ , P[])
-... | no ¬P[] = no (λ {([] , P[]) → ¬P[] P[]})
-anySubset? {suc n} P? with anySubset? (P? ∘ (inside ∷_))
-... | yes (_ , Pp) = yes (_ , Pp)
-... | no  ¬Pp      with anySubset? (P? ∘ (outside ∷_))
-...   | yes (_ , Pp) = yes (_ , Pp)
-...   | no ¬Pp'      = no λ
-  { (inside  ∷ p , Pp)  → ¬Pp  (_ , Pp)
-  ; (outside ∷ p , Pp') → ¬Pp' (_ , Pp')
-  }
+anySubset? {zero}  P? = map′ ([] ,_) (λ { ([] , P[]) → P[] }) (P? [])
+anySubset? {suc n} P? =
+  map′ [ Product.map _ id , Product.map _ id ]′
+       (λ { (outside ∷ p , Pp) → inj₂ (p , Pp)
+          ; ( inside ∷ p , Pp) → inj₁ (p , Pp)
+          })
+       (anySubset? (P? ∘ (inside ∷_)) ⊎-dec anySubset? (P? ∘ (outside ∷_)))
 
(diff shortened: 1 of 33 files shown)


src/Data/Fin/Subset/Properties.agda
```diff
@@ -152,13 +152,12 @@ x∈⁅y⁆⇔x≡y {_} {x} {y} = equivalence
 infix 4 _⊆?_
 _⊆?_ : ∀ {n} → B.Decidable (_⊆_ {n = n})
 []          ⊆? []          = yes id
-outside ∷ p ⊆? y ∷ q with p ⊆? q
-... | yes p⊆q = yes λ { (there v∈p) → there (p⊆q v∈p)}
-... | no  p⊈q = no (p⊈q ∘ drop-∷-⊆)
+outside ∷ p ⊆? y ∷ q =
+  map′ (λ { p⊆q (there v∈p) → there (p⊆q v∈p) }) drop-∷-⊆ (p ⊆? q)
```
MatthewDaggitt: Could we split the first direction into it's own lemma for these two?

src/Data/Fin/Subset/Properties.agda
```diff
@@ -576,15 +575,11 @@ Lift? P? p = decFinSubset (_∈? p) (λ {x} _ → P? x)
...
-...   | no ¬Pp'      = no λ
-  { (inside  ∷ p , Pp)  → ¬Pp  (_ , Pp)
-  ; (outside ∷ p , Pp') → ¬Pp' (_ , Pp')
-  }
+anySubset? {zero}  P? = map′ ([] ,_) (λ { ([] , P[]) → P[] }) (P? [])
+anySubset? {suc n} P? =
+  map′ [ Product.map _ id , Product.map _ id ]′
+       (λ { (outside ∷ p , Pp) → inj₂ (p , Pp)
+          ; ( inside ∷ p , Pp) → inj₁ (p , Pp)
+          })
```
MatthewDaggitt: Could we split the second direction out into it's own lemma or a `where` clause?

src/Data/List/Fresh/Relation/Unary/Any.agda
```diff
@@ -69,6 +69,6 @@ module _ {R : Rel A r} {P : Pred A p} (P? : Decidable P) where
 
   any? : (xs : List# A R) → Dec (Any P xs)
   any? []        = no (λ ())
-  any? (x ∷# xs) with P? x
-  ... | yes p = yes (here p)
-  ... | no ¬p = Dec.map′ there (tail ¬p) (any? xs)
+  any? (x ∷# xs) =
+    Dec.map′ [ here , there ]′ (λ{ (here x) → inj₁ x ; (there z) → inj₂ z })
```
MatthewDaggitt: Again maybe split the second direction out?

src/Data/List/Relation/Binary/Lex/Core.agda
```diff
@@ -94,15 +98,16 @@ module _ {a ℓ₁ ℓ₂} {A : Set a} {P : Set}
...
+  module _ (dec-P : Dec P) (dec-≈ : Decidable _≈_) (dec-≺ : Decidable _≺_)
+    where
+
+    decidable : Decidable _<_
+    decidable []       []       =
+      map′ base (λ{ (base p) → p }) dec-P
+    decidable []       (y ∷ ys) = yes halt
+    decidable (x ∷ xs) []       = no λ()
+    decidable (x ∷ xs) (y ∷ ys) =
+      map′ [ this , uncurry next ] (λ { (this x≺y) → inj₁ x≺y
```
MatthewDaggitt: Again maybe split the second direction out?

src/Data/List/Relation/Binary/Pointwise.agda
```diff
@@ -116,11 +121,8 @@ decidable : ∀ {_∼_ : REL A B ℓ} → Decidable _∼_ → Decidable (Pointwi
 decidable dec []       []       = yes []
```
MatthewDaggitt: This would be cleaner if we put `dec` in a module parameter!

src/Data/Sum/Properties.agda
```diff
@@ -23,14 +24,12 @@ module _ {a b} {A : Set a} {B : Set b} where
   inj₂-injective refl = refl
 
   ≡-dec : Decidable _≡_ → Decidable _≡_ → Decidable {A = A ⊎ B} _≡_
```
MatthewDaggitt: Might be worth lifting the two decidable arguments to module parameters?

src/Relation/Binary/Construct/Closure/Reflexive/Properties.agda
```diff
@@ -59,10 +61,10 @@ module _ {_~_ : Rel A ℓ} where
   ... | tri> _ _    c = inj₂ [ c ]
 
   dec : Decidable {A = A} _≡_ → Decidable _~_ → Decidable (Refl _~_)
-  dec ≡-dec ~-dec a b with ≡-dec a b | ~-dec a b
-  ... | _        | yes q = yes [ q ]
-  ... | yes refl | _     = yes refl
-  ... | no ¬p    | no ¬q = no λ { refl → ¬p refl; [ p ] → ¬q p }
+  dec ≡-dec ~-dec a b =
+    map′ Sum.[ (λ { refl → refl }) , [_] ]
+         (λ { refl → inj₁ refl ; [ x∼y ] → inj₂ x∼y })
```
MatthewDaggitt: Extract to separate lemmas?

src/Data/List/Fresh/Relation/Unary/Any.agda
gallais: :+1: these conversion lemmas are useful

src/Data/Fin/Subset/Properties.agda
laMudri: How about these lemmas being equivalences, and then using `map` rather than `map′`?

```agda
outside∷-⊆-⇔ : ∀ {n y} {p q : Subset n} → p ⊆ q ⇔ outside ∷ p ⊆ y ∷ q
outside∷-⊆-⇔ = equivalence (λ { p⊆q (there ∈p) → there (p⊆q ∈p) }) drop-∷-⊆
```

MatthewDaggitt: Hmm except that then you have to get `Setoid`s involved unnecessarily which feels like overkill. Not opposed to adding a lemma that they are an `Equivalence` of course (just make sure you use the new `Function` hierarchy rather than the old one!).

laMudri: Aah, okay. I hadn't seen this comment when I added the commits earlier, so I'll look over them again.

MatthewDaggitt:
Thanks for the changes, although I think perhaps @gallais and I didn't manage to be clear enough, apologies. Having the equivalences is great, but what we were really after was splitting out the more complicated one-direction lemmas. In particular the equivalence directions that require pattern matching! e.g. the forwards direction for `inside∷-⊆-⇔`

src/Data/Fin/Subset/Properties.agda
MatthewDaggitt: Okay, no worries!

laMudri:
Okay, I think that's everything cleared up.

JacquesCarette:
This is indeed very very nice!
What I find fascinating is that `map′` is the eliminator for `Dec`; normally programming with pattern-matching is more pleasant than using raw eliminators. This case seems to be different.

laMudri:
I'm not sure it's quite so fascinating. `map′` only allows elimination into another `Dec` of the same shape, so is quite akin to `map` for lists (rather than `foldr`), which gets plenty of use. It's the action on morphisms of the functor Dec : Type⇔ → Type. The iterator/recursor for `Dec` would have type `(P → Q) → (¬ P → Q) → Dec P → Q`.

gallais:
Probably worth mentioning that in `Relation.Nary` we have [`≟-mapₙ`](https://agda.github.io/agda-stdlib/Relation.Nary.html#2924)
automating the pattern

```agda
mapₙ (uncurryₙ (congₙ c)) c-injective (rec?₁ ×-dec ... ×-dec rec?ₙ)
```

It would be nice to have the lazier version too (or maybe even to only have
the lazier version). Up to you to decide whether it is then worth using in
other modules (it allows you to skip the `cong` bit which is obvious because
everything is congruent wrt propositional equality).

laMudri:
Thanks for mentioning this. I didn't know about it, and it looks handy.

So, as we discussed, it will become lazier once `map′` and `_×-dec_` become lazier. A question is whether it should be defined via `via-injection` from `Relation.Nullary.Decidable`, which has previously seen little use in the library (maybe because the name is bad). This injection would be from `(a₁ , ⋯ , aₙ)` to `con a₁ ⋯ aₙ`, which should be a corollary of `Injectiveₙ n con`.

As for using `≟-mapₙ` more widely, this seems like a good idea. I'd suggest this goes into a new pull request, because it's mostly a separate concern that won't ever affect computational behaviour. Also, the decision procedures changed in this PR and that one will intersect, but there are plenty that will only be changed in one of the two. My only concern with this change is that `Relation.Nary` is fairly heavy on dependencies, possibly causing cyclic dependencies down the line. But maybe `Properties` modules (where decidability of equality is proven) are too, so it doesn't really matter.

PR #936, opened 2019-10-07
Refactored Algebra hierarchy to the new standard format

MatthewDaggitt:
See #876. Minor backwards incompatibilities when importing both `Algebra` and `Algebra.FunctionProperties/Structures` when the latter is parameterised by an equality relation. However it can be fixed easily by importing `Algebra.Bundles` instead of `Algebra` so I hope that's acceptable to people? The one time inconvenience should be offset by the new ability to simply import `Algebra` instead of all the separate parts.


laMudri:
Thanks for standardising all these things!

The changelog talks about `*.Bundles` modules, but what seem to actually exist are `*.Packages` modules. What's going on there?

MatthewDaggitt:
Sorry, I was getting ahead of myself! In #876 the consensus is that they should be called `Bundles` not `Packages`. That change will be my next PR, but in the mean time I was trying to avoid writing `Packages` in the changelog where it can potentially be missed and forgotten about.

PR #948, opened 2019-10-12
Remove leq from Data.Unit(.Polymorphic)

### LINKED ISSUE #946: Deprecate `_≤_` in `Data.Unit(.Polymorphic)`
As discussed in #925 the order `_≤_` actually coincides with `_≡_` for `Data.Unit(.Polymorphic)`. We should deprecate `_≤_` and replace all proofs with `_≡_`.

JacquesCarette:
Fixes #946, but by deletion rather than deprecation. Is unlikely to actually cause trouble.


MatthewDaggitt:
Hmm this wasn't quite what I'd envisaged, as this actually deletes functionality, not just renames it.

Instead of deleting them entirely I think it would be better simply to rename the proofs from e.g. `≤-poset` to  `≡-poset`. I also think it's probably best to deprecate the existing code rather than delete it entirely. I agree it doesn't seem terribly useful but who knows what weird and wacky things people are using it for! It was written in the first place so presumably somewhere, someone is using it!

JacquesCarette:
Right - I'll actually close this PR as it is easier for me to redo this entirely rather than backtrack.  And it will hopefully allow me to rebase too.

PR #950, opened 2019-10-16
Simplify ∸-+-assoc

Kraks:
The original `∸-+-assoc` does pattern matching over `m`, `n` and `o`, which seems not necessary. I guess a simpler program/proof is preferred. 
This PR suggests a version that only destructs `m` and `n`, and just uses a rewrite and recursive invocation.


gallais:
Nice. Do we need a `rewrite` or can the RHS be `0∸n≡0 o` directly?

MatthewDaggitt:
Thanks for the PR! Merging in.

PR #955, opened 2019-10-20
Added more filter lemmas

### LINKED ISSUE #933: More lemmas about `filter` in `Data.List`
The following statement is usable and also trivial, but I fail to use it in a simplest way, somewhat like setting ``refl``. And I implement it as   
```
module _ ... {A : ...} {P ...} ... where
filter-step-positive :  (p? : Decidable P) {x : A} (xs : List A) → P x →         
                                                  filter p? (x ∷ xs) ≡ x ∷ (filter p? xs)      
filter-step-positive p? {x} xs Px  with  p? x                                    
...                               | yes _  =  refl                              
...                               | no ¬Px =  contradiction Px ¬Px              
```  
If a simpler proof is not possible, may be, the standard library needs the above proof?
A similar situation is with  ``filter-step-negatiive``.

**Edit**

- [ ] filter-++ :  filter p? (xs ++ ys) ≡  filter p? xs ++ filter p? ys
- [ ] filter-accept: P x → filter p? (x ∷ xs) ≡ x ∷ (filter p? xs)  
- [ ] filter-reject: \not P x → filter p? (x ∷ xs) ≡ filter p? xs
- [ ] filter-idem: filter p? ∘ filter p? ≗ filter p?

MatthewDaggitt:
Fixes  #933

### DIFF
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -30,7 +30,7 @@ open import Level using (Level)
 import Relation.Binary as B
 import Relation.Binary.Reasoning.Setoid as EqR
 open import Relation.Binary.PropositionalEquality as P
-  using (_≡_; _≢_; _≗_; refl ; sym ; cong; cong₂)
+  using (_≡_; _≢_; _≗_; refl ; sym ; cong; cong₂; inspect)
 open import Relation.Nullary using (¬_; yes; no)
 open import Relation.Nullary.Negation using (contradiction)
 open import Relation.Nullary.Decidable using (isYes; map′)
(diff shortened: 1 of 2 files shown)


src/Data/List/Properties.agda
```diff
@@ -753,10 +753,32 @@ module _ {P : A → Set p} (P? : Decidable P) where
...
+  filter-reject {x} ¬Px with P? x
+  ... | yes Px = contradiction Px ¬Px
+  ... | no  _  = refl
+
+  filter-idem : filter P? ∘ filter P? ≗ filter P?
+  filter-idem []       = refl
+  filter-idem (x ∷ xs) with P? x | inspect P? x
+  ... | no  _ | _                   = filter-idem xs
+  ... | yes _ | P.[ eq ] rewrite eq = cong (x ∷_) (filter-idem xs)
+
```
gallais: Alternatively this could be done with equational reasoning by using `filter-accept`.

MatthewDaggitt: True and probably be more elegant, although it does require an extra pattern match and a bit more space. I think on balance I'm just going to leave it as is,

PR #956, opened 2019-10-21
Deprecate LEQ in Data.Unit, and remove it from Data.Unit.Polymorphic.

### LINKED ISSUE #946: Deprecate `_≤_` in `Data.Unit(.Polymorphic)`

JacquesCarette:
Fixes #946 .  Hopefully this properly deprecates things.


MatthewDaggitt:
Thanks, looks good! The only thing I've changed is that I have just added a commit that reincludes the old definition of `_≤_` to maximise backwards compatibility. This is just in case someone is relying on that particularly definition somehow (for example totality now requires `refl` to prove).

Other than that I'm happy with it now!

src/Data/Unit.agda
```diff
@@ -42,22 +39,23 @@ decSetoid = Data.Unit.Properties.≡-decSetoid
...
-decTotalOrder = Data.Unit.Properties.≤-decTotalOrder
+decTotalOrder = Data.Unit.Properties.≡-decTotalOrder
 {-# WARNING_ON_USAGE decTotalOrder
 "Warning: decTotalOrder was deprecated in v1.1.
-Please use ≤-decTotalOrder from Data.Unit.Properties instead."
+Please use ≡-decTotalOrder from Data.Unit.Properties instead."
 #-}
 preorder = PropEq.preorder ⊤
 {-# WARNING_ON_USAGE decTotalOrder
 "Warning: preorder was deprecated in v1.1."
```
JacquesCarette: There is a problem here (an extra " ) that is causing the build to fail.

JacquesCarette:
Question: how to I `pull` your changes down into my branch? I would have fixed the broken build myself, but I could not figure out how to do that!

gallais:
@JacquesCarette I think a simple `git pull` ought to work: when you open a PR
you implicitly authorise maintainers to push to the branch you started the PR
from. So matthew has pushed directly to your fork. :)

MatthewDaggitt:
Yup, @gallais is right, a simple pull should do it.

> So matthew has pushed directly to your fork. :)

Apologies if that's bad git etiquette!

JacquesCarette:
It seems I had not set up the remote properly, so `git pull origin DeprecateUnitLEQ` was needed, then worked.

I had seen that I gave maintainers the right to push here - glad to do so, and that you fixed things up.

I fixed the comment break just now too.

src/Data/Unit/Polymorphic/Properties.agda
```diff
@@ -11,91 +11,78 @@ module Data.Unit.Polymorphic.Properties where
...
 
+≡-antisym : Antisymmetric {A = ⊤ {ℓ}} _≡_ _≡_
+≡-antisym p _ = p
 
+------------------------------------------------------------------------
 -- Structures
 
-≤-isPreorder : IsPreorder _≡_ (_≤_ {ℓ} {e = e})
-≤-isPreorder = record
+≡-isPreorder : IsPreorder {ℓ} {_} {⊤ {ℓ}} _≡_ _≡_
```
MatthewDaggitt: It occurs to me that for all the Structures and Bundles we probably want the level to be given explicitly rather than implicitly. It's very rare that structures or bundles are used in a context where Agda can infer the levels...

JacquesCarette: The 'problem' here is with `IsPreorder` , right? It should have 2 explicit `Level` parameters?

MatthewDaggitt: Yes, as should `≤-preorder`.

JacquesCarette: But that is independent of this PR, right?  Can't we pull this in as-is, and then make an issue to 'improve' `IsPreorder` and `≤-preorder` appropriately?

MatthewDaggitt: Yup, fair enough! Merging in.

PR #958, opened 2019-10-23
Tidy up and expand morphism hierarchies

MatthewDaggitt:
The library has added a lot of morphism code since `v1.1`. This PR attempts to tidy it up before the release of `v1.2`. In particular it:
-  Firstly it splits the files up into the new standard hierarchy structure, i.e. `Definitions` + `Structures` (with `Bundles` still to come). 
- Secondly it extends the hierarchy upwards to include not just homomorphisms, but also monomorphisms (which are used in `Data.Nat.Binary` and `Data.Rational` to transfer properties) and isomorphisms (yet unused).
- Thirdly it removes the assumption of equality axioms in the algebraic definitions. Assuming people are in favour, it therefore supersedes #853.

@mwageringel what are your thoughts?


src/Algebra/Morphism/Structures.agda
```diff
@@ -0,0 +1,111 @@
...
+-- Morphisms over magma-like structures
+------------------------------------------------------------------------
+
+module _ (_∙_ : Op₂ A) (_◦_ : Op₂ B) (⟦_⟧ : A → B) where
+
+  record IsMagmaHomomorphism : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      isRelHomomorphism : IsRelHomomorphism _≈₁_ _≈₂_ ⟦_⟧
+      homo              : Homomorphic₂ ⟦_⟧ _∙_ _◦_
+
```
gallais: I would expect a MagmaHomomorphism to be between two magmas.
Why are the bundles unpacked here?

MatthewDaggitt: So yes, we could parameterise the structures with `RawMagma`, `RawMonoid`, `RawRing` etc. This would be much neater, particularly once we get up to rings, but I wasn't sure whether consistency was better as we don't have the corresponding "raw" structures for the `Relation.Binary.Morphism` hierarchy.

mwageringel:
Thank you for putting this together. This looks pretty good to me and seems much more consistent than what I tried at #853.

This PR removes the raw morphisms altogether. They did not form a mathematical category, as identity morphisms could not be defined, so I welcome this change.

src/Algebra/Morphism/Structures.agda
mwageringel: If this is possible, I also would be in favor of parametrizing these algebraic morphisms by the raw structures, even if this is inconsistent with morphisms of relations.

Note that they cannot be parametrized by the non-raw structures `Magma` or `Monoid` in their current form, as the main purpose of `Algebra.Morphism.MagmaMonomorphism` is to prove that the domain of a monomorphism is a magma, etc. – although one could argue this is part of the definition of morphisms of magmas or monoids. On the other hand, one might argue that this rather is part of the definition of the corresponding category and not necessary in the definition of the morphisms themselves.

src/Algebra/Morphism/MagmaMonomorphism.agda
```diff
@@ -0,0 +1,125 @@
...
+open import Algebra.Structures
+open import Algebra.Morphism.Structures
+open import Relation.Binary.Core
+
+module Algebra.Morphism.MagmaMonomorphism
+  {a b} {A : Set a} {B : Set b}
+  {ℓ₁ ℓ₂} {_≈₁_ : Rel A ℓ₁} {_≈₂_ : Rel B ℓ₂}
+  {_∙_ : Op₂ A} {_◦_ : Op₂ B} {f : A → B}
+  (isMagmaMonomorphism : IsMagmaMonomorphism _≈₁_ _≈₂_ _∙_ _◦_ f)
+  (◦-isMagma : IsMagma _≈₂_ _◦_)
```
mwageringel: This argument is redundant for the structures `isSemigroup`, `isBand`, etc. below. Perhaps it could be added just for the properties section via an anonymous module, but not for the structures. The same applies for monoids.

src/Algebra/Morphism/MonoidMonomorphism.agda
```diff
@@ -0,0 +1,96 @@
...
+  f x ◦ f ε₁ ≈⟨  ◦-cong refl ε-homo ⟩
+  f x ◦ ε₂   ≈⟨  zeʳ (f x) ⟩
+  ε₂         ≈˘⟨ ε-homo ⟩
+  f ε₁       ∎)
+
+zero : Zero _≈₂_ ε₂ _◦_ → Zero _≈₁_ ε₁ _∙_
+zero = map zeroˡ zeroʳ
+
+------------------------------------------------------------------------
+-- Properties
```
mwageringel: Should this be `Structures`?

MatthewDaggitt: Yup!

src/Algebra/Morphism/MagmaMonomorphism.agda
MatthewDaggitt: Fair enough. I was just trying to make my life easier for myself, but you're right we shouldn't have it everywhere if we don't need it!

MatthewDaggitt:
There we go, addressed everyone's feedback I think.

src/Data/List/#Properties.agda#
```diff
@@ -0,0 +1,1061 @@
+------------------------------------------------------------------------
```
mwageringel: It looks like this emacs file was committed by accident.

PR #968, opened 2019-10-29
Redo Dec-related lemmas in the `Relation.Nullary` hierarchy

laMudri:
This pull request addresses points 1 and 2 of the plan laid out in #932. The key changes are as follows.

1. `map′` now has definition
```agda
map′ : (P → Q) → (Q → P) → Dec P → Dec Q
does  (map′ P→Q Q→P p?) = does p?
proof (map′ P→Q Q→P ( true because  [p])) = ofʸ (P→Q (invert [p]))
proof (map′ P→Q Q→P (false because [¬p])) = ofⁿ (invert [¬p] ∘ Q→P)
```
This is maximally non-strict, and the `does` half is branch-free. Additionally (I don't know whether this is good or bad) when `b` is a value, `proof (map′ id id (b because r))` will reduce to `of (invert r)`. This is a form of η-expansion. If we wanted to avoid this, the proof of `map′` would pattern match on `[p]` and `[¬p]`, instead of using `invert` below the `ofʸ` and `ofⁿ` constructors.

2. `isYes` (with alias `⌊_⌋`) has definition
```agda
isYes : Dec P → Bool
isYes ( true because _) = true
isYes (false because _)  = false
```
I have added a comment above it explaining what it would be used for (i.e, proof automation). Not destructing the proof portion should bring performance improvements to code that doesn't need the proofs.

3. There are two new proofs in `Relation.Nullary.Decidable`.
```agda
dec-true : (p? : Dec P) → P → does p? ≡ true
dec-false : (p? : Dec P) → ¬ P → does p? ≡ false
```
These will now generally be more useful than `dec-yes` and `dec-no`.

4. Each logical connective has a new lemma like
```agda
_×-reflects_ : ∀ {bp bq} → Reflects P bp → Reflects Q bq → Reflects (P × Q) (bp ∧ bq)
```
which is used to prove the corresponding `_×-dec_`. These are proved by pattern matching on the `Reflects` proofs rather than the `Bool`s, which may be inconsistent with how `map′` was proved.

I'll update the changelog, but I'm not sure which of these changes changes should be noted, and to what level of detail. Definitely the new lemmas are included, but the change to `map′` feels like part of the whole `Dec` changes. The rest of the changes I think are unimportant.

I'll address the rest of the hierarchy in a separate pull request. Note, though, that, given #935, most of the decision procedures in the library will already have the improved behaviour after this pull request. IIRC, only a few more complicated decision procedures remain to be redone.

### DIFF
--- a/README/Dec.agda
+++ b/README/Dec.agda
@@ -24,6 +24,7 @@ open import Data.Product
 open import Data.Unit
 open import Function
 open import Relation.Binary.PropositionalEquality
+open import Relation.Nary
 open import Relation.Nullary.Product
 
 infix 4 _≟₀_ _≟₁_ _≟₂_
@@ -106,7 +107,7 @@ _ = λ m n → refl
 -- `map′` can be used in conjunction with combinators such as `_⊎-dec_` and
 -- `_×-dec_` to build complex (simply typed) decision procedures.
 
-module ListDecEq {a} {A : Set a} (_≟ᴬ_ : (x y : A) → Dec (x ≡ y)) where
+module ListDecEq₀ {a} {A : Set a} (_≟ᴬ_ : (x y : A) → Dec (x ≡ y)) where
 
   _≟ᴸᴬ_ : (xs ys : List A) → Dec (xs ≡ ys)
   []       ≟ᴸᴬ []       = yes refl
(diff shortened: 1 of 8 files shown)


src/Relation/Nullary/Reflects.agda
```diff
@@ -8,15 +8,16 @@
...
 open import Data.Empty
 open import Level
-open import Relation.Binary.PropositionalEquality
 open import Relation.Nullary
 
-variable
-  p : Level
-  P : Set p
+private
+  variable
```
gallais: If this PR cannot make it to v1.2, this fix should at least be
included: we should never declare non-`private` `variable`s.

MatthewDaggitt:
@laMudri do you think you could add a CHANGELOG entry for this?

laMudri:
That should all be ready now.

laMudri:
I've added a README file. It requires the new definition of `map′` to check, which is why I didn't make a separate PR for it. If it should be separate, I can pull it out.

README/Dec.agda
```diff
@@ -0,0 +1,119 @@
...
+
+module ListDecEq {a} {A : Set a} (_≟ᴬ_ : (x y : A) → Dec (x ≡ y)) where
+
+  _≟ᴸᴬ_ : (xs ys : List A) → Dec (xs ≡ ys)
+  []       ≟ᴸᴬ []       = yes refl
+  []       ≟ᴸᴬ (y ∷ ys) = no λ ()
+  (x ∷ xs) ≟ᴸᴬ []       = no λ ()
+  (x ∷ xs) ≟ᴸᴬ (y ∷ ys) =
+    map′ (uncurry (cong₂ _∷_)) ∷-injective (x ≟ᴬ y ×-dec xs ≟ᴸᴬ ys)
+
```
gallais: Worth also demonstrating the use of `≟-map` from `Relation.Nary`?

laMudri:
Should there be a link to README.Dec somewhere? README.agda?

gallais:
README/Data at least?

laMudri:
Even though it's in the Relation hierarchy?

MatthewDaggitt:
See #975

PR #971, opened 2019-10-30
Make RawFunctor more Level polymorphic

JacquesCarette:
`Morphism` as well, and little bits of `RawApplicative`.

If this seems useful, I'll add something to the Changelog as well.


src/Category/Functor.agda
```diff
@@ -31,9 +31,9 @@ record RawFunctor {ℓ} (F : Set ℓ → Set ℓ) : Set (suc ℓ) where
 -- A functor morphism from F₁ to F₂ is an operation op such that
 -- op (F₁ f x) ≡ F₂ f (op x)
 
-record Morphism {ℓ} {F₁ F₂ : Set ℓ → Set ℓ}
+record Morphism {ℓ ℓ′ ℓ″} {F₁ : Set ℓ → Set ℓ′} {F₂ : Set ℓ → Set ℓ″}
                 (fun₁ : RawFunctor F₁)
-                (fun₂ : RawFunctor F₂) : Set (suc ℓ) where
+                (fun₂ : RawFunctor F₂) : Set ((suc ℓ) ⊔ (ℓ′ ⊔ ℓ″)) where
```
MatthewDaggitt: Don't think you need any of those brackets for the levels.

MatthewDaggitt:
Looks good to me, but I'd appreciate another opinion as I haven't used this part of the library at all.

laMudri:
Is it a good idea to call it `Morphism` rather than `FunctorMorphism` or `RawFunctorMorphism` or something like that?

JacquesCarette:
This code pre-dates me, but on naming: since this is in the file/module `Category.Functor`, I think it makes sense that it is a `Morphism` in the context of `Category.Functor`.  I think it's a bad idea to try to put all of the context in the name; qualified imports work well if you want to have multiple `Morphism`s around at once.

MatthewDaggitt:
Hi @JacquesCarette, any chance you could add an changelog entry for this? I'll then merge it in.

src/Category/Functor.agda
JacquesCarette: Indeed - fixing.

JacquesCarette:
This should now be ready to merge.

PR #973, opened 2019-11-02
Added properties to `Algebra.Properties.Group`

rei1024:
This PR adds these properties to `Algebra.Properties.Group`.
```agda
⁻¹-injective : ∀ {x y} → x ⁻¹ ≈ y ⁻¹ → x ≈ y
⁻¹-anti-homo : ∀ x y → (x ∙ y) ⁻¹ ≈ y ⁻¹ ∙ x ⁻¹
```
and refactor `Algebra.Properties.Group` and `Algebra.Properties.AbelianGroup`.


MatthewDaggitt:
Looks good, thanks for the PR. It's small enough that I'll include it in `v1.2`

laMudri:
Maybe `⁻¹-anti-homo` should be arranged a bit differently. We also should have that `ε  ⁻¹ ≈ ε`, so the whole property is that `_⁻¹` is a homomorphism from G to Gᵒᵖ. At least, I think `⁻¹-anti-homo` should say what it is a homomorphism over (`_∙_`).

laMudri:
And eventually we should note that `_⁻¹` is not just an injection but a bijection, and so too are the actions `x ∙_` and `_∙ x` of any group element `x`. This can come separately, though. I think the first part should follow from `⁻¹-involutive` (giving an inverse to `_⁻¹`) plus the general principle that Inverse ⇒ Bijection.

rei1024:
@laMudri Do you think `⁻¹-antihomo-∙` is a good name for it?

MatthewDaggitt:
Moving to `v1.3` as it sounds like there's still some more discussion to be had.

laMudri:
Possibly `⁻¹-anti-homo-∙`, to combine the two. I guess we want to keep `⁻¹-homo-∙` free for abelian groups (though it seems to be called `⁻¹-∙-comm` at the moment). Other than that, I think these are good additions. They may be refactored later, but these names should still be exported (if I understand stdlib design well enough).

MatthewDaggitt:
Hi @rei1024, thanks again for the PR. Apologies it's taken so long to merge in!

PR #974, opened 2019-11-03
Reverted decidability procedures in `Data.Nat.Binary.Properties` to improve performance

MatthewDaggitt:


src/Data/Nat/Binary/Properties.agda
```diff
@@ -273,11 +274,35 @@ private
...
+... | tri> _ _ y<x =  tri> (>⇒≮ gt) (>⇒≢ gt) gt  where gt = odd<even (inj₁ y<x)
+<-cmp 1+[2 _ ] zero =  tri> (>⇒≮ gt) (>⇒≢ gt) gt  where gt = 0<odd
+<-cmp 1+[2 x ] 2[1+ y ]  with <-cmp x y
+... | tri< x<y _ _ =  tri< lt (<⇒≢ lt) (<⇒≯ lt)  where lt = odd<even (inj₁ x<y)
+... | tri≈ _ x≡y _ =  tri< lt (<⇒≢ lt) (<⇒≯ lt)  where lt = odd<even (inj₂ x≡y)
+... | tri> _ _ x>y =  tri> (>⇒≮ gt) (>⇒≢ gt) gt  where gt = even<odd x>y
+<-cmp 1+[2 x ] 1+[2 y ]  with <-cmp x y
+... | tri< x<y _  _   =  tri< lt (<⇒≢ lt) (<⇒≯ lt)  where lt = odd<odd x<y
+... | tri≈ _ refl _   =  tri≈ (≡⇒≮ refl) refl (≡⇒≯ refl)
+... | tri> _ _    x>y =  tri> (>⇒≮ gt) (>⇒≢ gt) gt  where gt = odd<odd x>y
```
gallais: Feels like we should be able to play @laMudri's games on `Tri`
just like we have on `Dec`.

MatthewDaggitt: Yup, probably true, although won't we need need a three valued logic? Might be best to open an issue for `v1.3`...

laMudri: At least in this case, we have a clear intended semantics for our three-valued logic (if we choose to see it as such). But yeah, the same pattern clearly applies.

gallais: We can use the classic `LT`, `EQ`, `GT` familiar to Haskellers.

PR #975, opened 2019-11-09
Improve pointers to README files and rename README.Dec

MatthewDaggitt:


src/Relation/Nullary.agda
```diff
@@ -25,18 +26,22 @@ infix 2 _because_
...
--- value to compute independently from the proof portion.
--- This often allows good computational properties when we
--- only care about the boolean portion.
+------------------------------------------------------------------------
+-- Decidability.
+
+-- This definition allows the boolean portion of the value to compute
+-- independently from the proof portion. This translates to better
+-- computational properties when we only care about the result and not
+-- the proof. See README.Decidability for further details.
```
laMudri: Looking back, I think “the value” should be something more informative, like “a decision procedure”.

laMudri:
It's a bit of a shame that these pointers are not clickable/go-to-definition-able (because of cyclic dependencies, I guess).

gallais:
> It's a bit of a shame that these pointers are not clickable/go-to-definition-able (because of cyclic dependencies, I guess).

We could write a small Haskell script that spots special syntax in comments
and turns it into links. This way the html hosted by us would have the relevant
links.

A better solution would of course to extend Agda to do so natively.

src/Relation/Nullary.agda
MatthewDaggitt: Updated the text.

PR #976, opened 2019-11-09
Moved most of Function.Core to Function.Utilities

MatthewDaggitt:
Whilst tidying up the CHANGELOG I noticed that we were advising people to import `Function.Core`. The whole point of the `Core` modules is that they should not be imported. Additionally `Function.Core` contains all sorts of stuff not found in `Algebra.Core` or `Relation.Binary.Core`.

I've therefore moved all the function combinators out of `Function.Core` to `Function.Utilities`, which is still re-exported by `Function`. I picked `Utilities` without a lot of detailed thought. I'm open to other names. `Function.Combinators` would also work perhaps?


laMudri:
I don't fully understand the naming conventions, but could it be called `Function.Base`?

MatthewDaggitt:
`Function.Base` would work. Changing now.

Blaisorblade:
> The whole point of the `Core` modules is that they should not be imported.

Why? Shouldn't that be documented, if it matters?

laMudri:
I guess this is folklore among contributors (who learn when they end up needing to introduce a `Core` module). For users, we hope they see the header comments like:

```agda
-- The contents of this file should usually be accessed from `Function`.
```

Some centralised documentation might help, though.

MatthewDaggitt:
> Why? Shouldn't that be documented, if it matters?

It is documented, albeit fairly far down the main `README.agda` file.

https://github.com/agda/agda-stdlib/blob/c36eae9f1d11f69a08164b2c675fa517bd78d030/README.agda#L320

I do have a branch lurking on my local repo that tries to tidy that file up. Really need to push it at some point.

PR #982, opened 2019-11-18
Rewrite definitions that use Dec to be less strict

laMudri:
This pull request addresses points 4 and 5 from #932. I found that distinguishing programs from proofs (the difference between 4 and 5) was fairly difficult, so I put them all into one PR. The general principles have been:

1. If possible, factor through `does`.
2. If the proof part is needed, try to localise pattern matching as much as possible using `invert`.
3. In cases where (AFAICT) computation doesn't matter, keep with `yes` and `no`. This is mostly proofs via `⊥`.

### DIFF
--- a/src/Axiom/UniquenessOfIdentityProofs.agda
+++ b/src/Axiom/UniquenessOfIdentityProofs.agda
@@ -8,7 +8,9 @@
 
 module Axiom.UniquenessOfIdentityProofs where
 
+open import Data.Bool.Base using (true; false)
 open import Data.Empty
+open import Relation.Nullary.Reflects using (invert)
 open import Relation.Nullary hiding (Irrelevant)
 open import Relation.Binary.Core
 open import Relation.Binary.Definitions
@@ -62,13 +64,13 @@ module Decidable⇒UIP
 
   ≡-normalise : _≡_ {A = A} ⇒ _≡_
   ≡-normalise {a} {b} a≡b with a ≟ b
-  ... | yes p = p
-  ... | no ¬p = ⊥-elim (¬p a≡b)
+  ... | true  because  [p] = invert [p]
+  ... | false because [¬p] = ⊥-elim (invert [¬p] a≡b)
 
   ≡-normalise-constant : ∀ {a b} (p q : a ≡ b) → ≡-normalise p ≡ ≡-normalise q
   ≡-normalise-constant {a} {b} p q with a ≟ b
-  ... | yes _ = refl
-  ... | no ¬p = ⊥-elim (¬p p)
+  ... | true  because   _  = refl
+  ... | false because [¬p] = ⊥-elim (invert [¬p] p)
 
   ≡-irrelevant : UIP A
   ≡-irrelevant = Constant⇒UIP.≡-irrelevant ≡-normalise ≡-normalise-constant
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -674,15 +674,15 @@ module _ {P : Pred A p} (P? : Decidable P) where
 
   takeWhile++dropWhile : ∀ xs → takeWhile P? xs ++ dropWhile P? xs ≡ xs
   takeWhile++dropWhile []       = refl
-  takeWhile++dropWhile (x ∷ xs) with P? x
-  ... | yes _ = cong (x ∷_) (takeWhile++dropWhile xs)
-  ... | no  _ = refl
+  takeWhile++dropWhile (x ∷ xs) with does (P? x)
+  ... | true  = cong (x ∷_) (takeWhile++dropWhile xs)
+  ... | false = refl
 
   span-defn : span P? ≗ < takeWhile P? , dropWhile P? >
   span-defn []       = refl
-  span-defn (x ∷ xs) with P? x
-  ... | yes _ = cong (Prod.map (x ∷_) id) (span-defn xs)
-  ... | no  _ = refl
+  span-defn (x ∷ xs) with does (P? x)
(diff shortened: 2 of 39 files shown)


laMudri:
Minor point: should the convention be ` true because` or `true  because`? I think I've been switching between the two.

laMudri:
So, I've been investigating the checking failure in `Data.Rational.Properties`. One failing lemma is the following:

```agda
normalize-coprime : ∀ {n d-1} .(c : Coprime n (suc d-1)) →
                    normalize n (suc d-1) ≡ mkℚ (+ n) d-1 c
normalize-coprime {n} {d-1} c = begin
  normalize n d              ≡⟨⟩
  mkℚ+ (n ℕ./ g) (d ℕ./ g) _ ≡⟨ mkℚ+-cong n/g≢0 d/1≢0 {c₂ = c₂} (ℕ./-congʳ {n≢0 = g≢0} g≡1) (ℕ./-congʳ {n≢0 = g≢0} g≡1) ⟩
  mkℚ+ (n ℕ./ 1) (d ℕ./ 1) _ ≡⟨ mkℚ+-cong d/1≢0 _ (ℕ.n/1≡n n) (ℕ.n/1≡n d) ⟩
  mkℚ+ n d _                 ≡⟨⟩
  mkℚ (+ n) d-1 _            ∎
  where
  open ≡-Reasoning; d = suc d-1; g = ℕ.gcd n d
  c′ = recomputeCP c
  c₂ : Coprime (n ℕ./ 1) (d ℕ./ 1)
  c₂ = subst₂ Coprime (sym (ℕ.n/1≡n n)) (sym (ℕ.n/1≡n d)) c′
  g≡1 = C.coprime⇒gcd≡1 c′
  g≢0   = fromWitnessFalse (ℕ.gcd[m,n]≢0 n d (inj₂ λ()))
  n/g≢0 = fromWitnessFalse (ℕ.n/gcd[m,n]≢0 n d {_} {g≢0})
  d/1≢0 = fromWitnessFalse (subst (_≢ 0) (sym (ℕ.n/1≡n d)) λ())
```

The reported error, after the changes made so far, is that where `n/g≢0` is used, for `eq : d ≡ 0`, `(λ()) eq` (from the definition of `g≢0`) is not equal to the expected, `ℕ.≡⇒≡ᵇ d 0 eq`. Indeed, replacing `λ()` by `ℕ.≡⇒≡ᵇ d 0` in `g≢0` fixes the issue. I don't want to do this, because it implies that users of stdlib could have similar code, which breaks in this mysterious way.

The odd thing is that even in master, these two open terms are not equal. So somehow, I've introduced an equality constraint. I'll keep investigating this on another day.

laMudri:
It looks like the problem wasn't really to do with anything `Dec`-related. I had unthinkingly changed the implementation of `T?` in `Data.Bool.Properties` so that the proof of `¬ ⊥` required in the `no`/`ofⁿ` case was `id` rather than `λ ()`. Changing it back to `λ ()` seems to fix the problem in `Data.Rational.Properties`.

src/Axiom/UniquenessOfIdentityProofs.agda
```diff
@@ -62,13 +64,13 @@ module Decidable⇒UIP
 
   ≡-normalise : _≡_ {A = A} ⇒ _≡_
   ≡-normalise {a} {b} a≡b with a ≟ b
-  ... | yes p = p
-  ... | no ¬p = ⊥-elim (¬p a≡b)
+  ... | true  because  [p] = invert [p]
+  ... | false because [¬p] = ⊥-elim (invert [¬p] a≡b)
```
MatthewDaggitt: I'm struggling to see the advantage of this change? Can you give an example of where the behaviour differs?

src/Data/List/Properties.agda
```diff
@@ -673,15 +674,15 @@ module _ {P : Pred A p} (P? : Decidable P) where
 
   takeWhile++dropWhile : ∀ xs → takeWhile P? xs ++ dropWhile P? xs ≡ xs
   takeWhile++dropWhile []       = refl
-  takeWhile++dropWhile (x ∷ xs) with P? x
-  ... | yes _ = cong (x ∷_) (takeWhile++dropWhile xs)
-  ... | no  _ = refl
+  takeWhile++dropWhile (x ∷ xs) with does (P? x)
```
MatthewDaggitt: Could you just confirm for me that these proofs still go through without this change? i.e. that changing the definition of `takeWhile` doesn't **require** us to change the proofs.

src/Axiom/UniquenessOfIdentityProofs.agda
laMudri: Here I'm being cautious, and hoping that the principle of making case splits as late as possible might save us in some weird case. In particular, it's possible to imagine a difference in the `false` case, where we now have `(≡-normalise _ | false because <stuck>) = ⊥-elim <stuck>` rather than just being stuck.

Another rough principle was to pretend that old `Dec` never existed, and to rewrite stuff in the newly “obvious” way. This goes some way to explaining the `true` case, as well as wanting to keep consistency with the `false` case.

src/Data/List/Properties.agda
laMudri: Yep. When I was working on these, the file checked before and after the changes to each lemma.

src/Axiom/UniquenessOfIdentityProofs.agda
MatthewDaggitt: Okay fair enough, thanks for the explanation.

MatthewDaggitt:
Any chance you could add a CHANGELOG entry documenting the change and then I'll merge it in.

PR #984, opened 2019-11-26
Proper subset relations

uzytkownik:
Sorry if this is not notable enough changes. I needed to use well-founded induction over the subsets in my project for which I needed to define subset relationship and I decided to attempt to upstream my attempt (I haven't ported the well-founded relationship yet). I'm still learning AGDA so I'm sorry if code is unidiomatic et.c

I added `⊂′` as well to mirror the relation definition though I founded it more difficult to prove properties of.

### DIFF
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -586,6 +586,10 @@ not-involutive : Involutive not
 not-involutive true  = refl
 not-involutive false = refl
 
+not-injective : ∀ {x y} → not x ≡ not y → x ≡ y
+not-injective {false} {false} nx≢ny = refl
+not-injective {true}  {true}  nx≢ny = refl
+
 not-¬ : ∀ {x y} → x ≡ y → x ≢ not y
 not-¬ {true}  refl ()
 not-¬ {false} refl ()
--- a/src/Data/Fin/Subset/Properties.agda
+++ b/src/Data/Fin/Subset/Properties.agda
@@ -19,9 +19,9 @@ open import Data.Bool.Properties
 open import Data.Fin using (Fin; suc; zero)
 open import Data.Fin.Subset
 open import Data.Fin.Properties using (any?; decFinSubset)
-open import Data.Nat.Base using (ℕ; zero; suc; z≤n; s≤s; _≤_)
+open import Data.Nat.Base using (ℕ; zero; suc; z≤n; s≤s; _≤_; _<_)
 open import Data.Nat.Properties using (≤-step)
-open import Data.Product as Product using (∃; ∄; _×_; _,_)
+open import Data.Product as Product using (∃; ∄; _×_; _,_; proj₁)
 open import Data.Sum as Sum using (_⊎_; inj₁; inj₂; [_,_]′)
 open import Data.Vec
 open import Data.Vec.Properties
@@ -42,8 +42,15 @@ open import Relation.Unary using (Pred; Decidable; Satisfiable)
 drop-there : ∀ {s n x} {p : Subset n} → suc x ∈ s ∷ p → x ∈ p
 drop-there (there x∈p) = x∈p
 
+drop-not-there : ∀ {s n x} {p : Subset n} → suc x ∉ s ∷ p → x ∉ p
+drop-not-there x∉sp x∈p = contradiction (there x∈p) x∉sp
+
 drop-∷-⊆ : ∀ {n s₁ s₂} {p₁ p₂ : Subset n} → s₁ ∷ p₁ ⊆ s₂ ∷ p₂ → p₁ ⊆ p₂
 drop-∷-⊆ p₁s₁⊆p₂s₂ x∈p₁ = drop-there (p₁s₁⊆p₂s₂ (there x∈p₁))
+
+drop-∷-⊂ : ∀ {n s} {p₁ p₂ : Subset n} → s ∷ p₁ ⊂ s ∷ p₂ → p₁ ⊂ p₂
+drop-∷-⊂ {s = inside} (_       , zero  ,      x∈sp₂ , x∉sp₁) = contradiction here x∉sp₁
+drop-∷-⊂ {s}          (sp₁⊆sp₂ , suc x , there x∈p₂ , x∉sp₁) = drop-∷-⊆ sp₁⊆sp₂ , x , x∈p₂ , drop-not-there x∉sp₁
 
 module _ {n} {p q : Subset n} where
 
@@ -59,6 +66,28 @@ module _ {n} {p q : Subset n} where
 
   in⊆in-⇔ : p ⊆ q ⇔ inside ∷ p ⊆ inside ∷ q
   in⊆in-⇔ = equivalence in⊆in drop-∷-⊆
+
+  s⊆s : ∀ {s} →  p ⊆ q → s ∷ p ⊆ s ∷ q
+  s⊆s p⊆q here = here
+  s⊆s p⊆q (there ∈p) = there (p⊆q ∈p)
+
+  out⊂ : ∀ {y} → p ⊂ q → outside ∷ p ⊂ y ∷ q
+  out⊂ (p⊆q , x , x∈q , x∉p) = out⊆ p⊆q , suc x , there x∈q , x∉p ∘ drop-there
+
+  out⊂in : p ⊆ q → outside ∷ p ⊂ inside ∷ q
+  out⊂in p⊆q = out⊆ p⊆q , zero , here , λ ()
+
+  out⊂in-⇔ : p ⊆ q ⇔ outside ∷ p ⊂ inside ∷ q
+  out⊂in-⇔ = equivalence out⊂in (drop-∷-⊆ ∘ proj₁)
+
+  out⊂out-⇔ : p ⊂ q ⇔ outside ∷ p ⊂ outside ∷ q
+  out⊂out-⇔ = equivalence out⊂ drop-∷-⊂
+
+  in⊂in : p ⊂ q → inside ∷ p ⊂ inside ∷ q
+  in⊂in (p⊆q , x , x∈q , x∉p) = in⊆in p⊆q , suc x , there x∈q , x∉p ∘ drop-there
+
+  in⊂in-⇔ : p ⊂ q ⇔ inside ∷ p ⊂ inside ∷ q
+  in⊂in-⇔ = equivalence in⊂in drop-∷-⊂
 
 ------------------------------------------------------------------------
 -- _∈_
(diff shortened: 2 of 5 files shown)


CHANGELOG/v1.2.md
```diff
@@ -506,6 +506,11 @@ Other minor additions
   tail : ∀ {i} {j : Size< i} → Stream A i → Stream A j
   ```
 
+* Added new proofs to `Data.Bool`:
```
MatthewDaggitt: Thanks for adding a CHANGELOG entry, most people forget that! The only problems is that `v1.2` has now been released so the entry should go in `CHANGELOG.md` outside of this folder which for old changelogs only.

CHANGELOG/v1.2.md
```diff
@@ -522,17 +527,29 @@ Other minor additions
...
   in⊆in-⇔       : p ⊆ q ⇔  inside ∷ p ⊆ inside ∷ q
+  s⊆s           : p ⊆ q →       s ∷ p ⊆      s ∷ q
 
   ∃-Subset-zero : ∃⟨ P ⟩ → P []
   ∃-Subset-[]-⇔ : P [] ⇔ ∃⟨ P ⟩
   ∃-Subset-suc  : ∃⟨ P ⟩ → ∃⟨ P ∘ (inside ∷_) ⟩ ⊎ ∃⟨ P ∘ (outside ∷_) ⟩
   ∃-Subset-∷-⇔  : (∃⟨ P ∘ (inside ∷_) ⟩ ⊎ ∃⟨ P ∘ (outside ∷_) ⟩) ⇔ ∃⟨ P ⟩
+
+  x∉∁s⇒x∈s : ∀ {n} → {x : Fin n} → {s : Subset n} → x ∉ ∁ s → x ∈ s
+  x∉s⇒x∈∁s : ∀ {n} → {x : Fin n} → {s : Subset n} → x ∉ s → x ∈ ∁ s
```
MatthewDaggitt: To keep things tidy, we usually don't put implicit variables in the types in the CHANGELOG

src/Data/Fin/Subset/Properties.agda
```diff
@@ -203,6 +256,171 @@ p⊆q⇒∣p∣<∣q∣ {p = outside ∷ p} {inside  ∷ q} p⊆q = ≤-step (p
...
+------------------------------------------------------------------------
+-- _⊂_
+
+p⊂q⇒p⊆q : ∀ {n} → {p q : Subset n} → p ⊂ q → p ⊆ q
+p⊂q⇒p⊆q = proj₁
+
+⊂-trans : ∀ {n} → Transitive (_⊂_ {n})
+⊂-trans (p⊆q , x , x∈q , x∉p) (q⊆r , _ , _ , _) = ⊆-trans p⊆q q⊆r , x , q⊆r x∈q , x∉p
+
+⊂-trans-⊆ : ∀ {n} → Trans (_⊂_ {n}) (_⊆_ {n}) (_⊂_ {n})
```
MatthewDaggitt: I think the standard name for this is `⊂-⊆-trans` (e.g. see [Data.Rational](https://github.com/agda/agda-stdlib/blob/b1cdd5d7dddf5251d5eda0cbf3711fa2bbd3bbb5/src/Data/Rational/Properties.agda#L357)). I think(?) you could also make the type clearer as `Trans {A = Subset n} _⊂_ _⊆_ _⊂_`

src/Data/Fin/Subset/Properties.agda
```diff
@@ -203,6 +256,171 @@ p⊆q⇒∣p∣<∣q∣ {p = outside ∷ p} {inside  ∷ q} p⊆q = ≤-step (p
...
+⊂-trans-⊆ (p⊆q , x , x∈q , x∉p) q⊆r = ⊆-trans p⊆q q⊆r , x , q⊆r x∈q , x∉p
+
+⊆-trans-⊂ : ∀ {n} → Trans (_⊆_ {n}) (_⊂_ {n}) (_⊂_ {n})
+⊆-trans-⊂ p⊆q (q⊆r , x , x∈r , x∉q) = ⊆-trans p⊆q q⊆r , x , x∈r , λ x∈p → contradiction (p⊆q x∈p) x∉q
```
MatthewDaggitt: The last term of this can be simplified to `x∉q ∘ p⊆q`

src/Data/Fin/Subset/Properties.agda
```diff
@@ -203,6 +256,171 @@ p⊆q⇒∣p∣<∣q∣ {p = outside ∷ p} {inside  ∷ q} p⊆q = ≤-step (p
...
+p⊂′q⇒∣p∣<∣q∣ {p = []} {[]} (p⊆q , q⊈p) = case q⊈p p⊆q of λ ()
+p⊂′q⇒∣p∣<∣q∣ {p = outside ∷ p} {outside ∷ q} op⊂′oq@(_ , _) = p⊂′q⇒∣p∣<∣q∣ (drop-∷-⊂′ op⊂′oq)
+p⊂′q⇒∣p∣<∣q∣ {p = outside ∷ p} {inside ∷ q} (op⊆iq , _) = s≤s (p⊆q⇒∣p∣<∣q∣ (drop-∷-⊆ op⊆iq))
+p⊂′q⇒∣p∣<∣q∣ {p = inside ∷ p} {outside ∷ q} (ip⊆oq , _) = contradiction (ip⊆oq here) (λ ())
+p⊂′q⇒∣p∣<∣q∣ {p = inside ∷ p} {inside ∷ q} ip⊂′iq@(ip⊆iq , _) = s≤s (p⊂′q⇒∣p∣<∣q∣ (drop-∷-⊂′ ip⊂′iq))
+
+------------------------------------------------------------------------
+-- ∁
+
+x∉∁s⇒x∈s : ∀ {n} → {x : Fin n} → {s : Subset n} → x ∉ ∁ s → x ∈ s
```
MatthewDaggitt: I think these last two will be significantly easier to prove by first proving the contrapositive (i.e. `x∈s⇒x∉∁s` and `x∈∁s⇒x∉s`) by induction.

src/Data/Fin/Subset/Properties.agda
```diff
@@ -203,6 +256,171 @@ p⊆q⇒∣p∣<∣q∣ {p = outside ∷ p} {inside  ∷ q} p⊆q = ≤-step (p
...
+    { isEquivalence = isEquivalence
+    ; irrefl = ⊂-irref
+    ; trans = ⊂-trans
+    ; <-resp-≈ = (λ {refl → id}) , (λ {refl → id})
+    }
+
+  ⊂-strictPartialOrder : StrictPartialOrder _ _ _
+  ⊂-strictPartialOrder = record
+    { _≈_ = _≡_
+    ; _<_ = _⊂_
```
MatthewDaggitt: We usually omit these first two lines entirely from the record as it's nearly always obvious what the underlying relations are. Agda will then infer them for us.

src/Data/Fin/Subset/Properties.agda
```diff
@@ -59,6 +67,51 @@ module _ {n} {p q : Subset n} where
   in⊆in-⇔ : p ⊆ q ⇔ inside ∷ p ⊆ inside ∷ q
   in⊆in-⇔ = equivalence in⊆in drop-∷-⊆
 
+  s⊆s : ∀ {s} →  p ⊆ q → s ∷ p ⊆ s ∷ q
+  s⊆s {.inside} p⊆q here = here
+  s⊆s {s} p⊆q (there ∈p) = there (p⊆q ∈p)
+
+  out⊂ : ∀ {y} → p ⊂ q → outside ∷ p ⊂ y ∷ q
+  out⊂ (p⊆q , x , x∈q , x∉p) = out⊆ p⊆q , suc x , there x∈q , λ {(there x∈p) → contradiction x∈p x∉p}
```
MatthewDaggitt: I think the last part can be simplified to `x∉p ∘ drop-there`

src/Data/Fin/Subset/Properties.agda
```diff
@@ -59,6 +67,51 @@ module _ {n} {p q : Subset n} where
...
+  out⊂out-⇔ = equivalence out⊂ drop-∷-⊂
+
+  in⊂in : p ⊂ q → inside ∷ p ⊂ inside ∷ q
+  in⊂in (p⊆q , x , x∈q , x∉p) = in⊆in p⊆q , suc x , there x∈q , λ {(there x∈p) → contradiction x∈p x∉p}
+
+  in⊂in-⇔ : p ⊂ q ⇔ inside ∷ p ⊂ inside ∷ q
+  in⊂in-⇔ = equivalence in⊂in drop-∷-⊂
+
+drop-∷-⊂′ : ∀ {n s} {p₁ p₂ : Subset n} → s ∷ p₁ ⊂′ s ∷ p₂ → p₁ ⊂′ p₂
+drop-∷-⊂′ (sp₁⊆sp₂ , sp₂⊈sp₁) = drop-∷-⊆ sp₁⊆sp₂ , λ p₂⊂p₁ → sp₂⊈sp₁ (s⊆s p₂⊂p₁)
```
MatthewDaggitt: Again can be simplified to ` sp₂⊈sp₁ ∘ s⊆s`

src/Data/Fin/Subset/Properties.agda
```diff
@@ -59,6 +67,51 @@ module _ {n} {p q : Subset n} where
   in⊆in-⇔ : p ⊆ q ⇔ inside ∷ p ⊆ inside ∷ q
   in⊆in-⇔ = equivalence in⊆in drop-∷-⊆
 
+  s⊆s : ∀ {s} →  p ⊆ q → s ∷ p ⊆ s ∷ q
+  s⊆s {.inside} p⊆q here = here
+  s⊆s {s} p⊆q (there ∈p) = there (p⊆q ∈p)
```
MatthewDaggitt: Don't think you need the implicits here.

src/Data/Bool/Properties.agda
```diff
@@ -586,6 +586,10 @@ not-involutive : Involutive not
 not-involutive true  = refl
 not-involutive false = refl
 
+not-injective : ∀ {x y} → not x ≡ not y → x ≡ y
+not-injective {false} {false} nx≢ny = refl
```
JacquesCarette: This proof is strict on the implicit arguments but lazy on the explicit argument - is this really what we want?

src/Data/Fin/Subset/Properties.agda
```diff
@@ -41,9 +42,16 @@ open import Relation.Unary using (Pred; Decidable; Satisfiable)
...
 drop-there (there x∈p) = x∈p
 
+drop-not-there : ∀ {s n x} {p : Subset n} → suc x ∉ s ∷ p → x ∉ p
+drop-not-there x∉sp x∈p = contradiction (there x∈p) x∉sp
+
 drop-∷-⊆ : ∀ {n s₁ s₂} {p₁ p₂ : Subset n} → s₁ ∷ p₁ ⊆ s₂ ∷ p₂ → p₁ ⊆ p₂
 drop-∷-⊆ p₁s₁⊆p₂s₂ x∈p₁ = drop-there (p₁s₁⊆p₂s₂ (there x∈p₁))
 
+drop-∷-⊂ : ∀ {n s} {p₁ p₂ : Subset n} → s ∷ p₁ ⊂ s ∷ p₂ → p₁ ⊂ p₂
+drop-∷-⊂ {s = inside} (_ , zero , x∈sp₂ , x∉sp₁) = contradiction here x∉sp₁
```
JacquesCarette: Since you're case-splitting on the argument (the 2nd part of the 4-tuple, it seems), and don't use `s` at all, why bother defining/refining it?

src/Data/Bool/Properties.agda
uzytkownik: AGDA rejects being strict in the explicit argument without being strict in implicit here. Do you want me to make it strict in all arguments?

src/Data/Fin/Subset/Properties.agda
uzytkownik: AGDA does not typecheck otherwise.

CHANGELOG/v1.2.md
uzytkownik: I moved to CHANGELOG.md.

CHANGELOG/v1.2.md
uzytkownik: I've removed all implicit variables from CHANGELOG.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done

src/Data/Fin/Subset/Properties.agda
uzytkownik: I'm not sure how to get from there within intuitionistic logic though so I switched to induction for those functions.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done.

src/Data/Fin/Subset/Properties.agda
uzytkownik: Done.

src/Data/Bool/Properties.agda
JacquesCarette: Can you not match just on 'refl' ?

src/Data/Fin/Subset/Properties.agda
JacquesCarette: I see (and I think I get it).

src/Data/Bool/Properties.agda
uzytkownik: I don't think so:

```
not-injective refl = refl
```

Results in:

```
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  not x ≟ not y
when checking that the pattern refl has type not x ≡ not y
```

uzytkownik: You mean:

```agda
not-injective {false} {false} refl = refl
not-injective {true}  {true}  refl = refl
```

That's what I meant by being strict in all arguments.

JacquesCarette: I was just incorrect here.  not-injective really does need to pattern-match on its boolean inputs.  I think your original definition is fine, sorry for the noise.

MatthewDaggitt:
Thanks for the changes. Looks good to me. I'll leave it up for another couple of days so that other people have the chance to give feedback and then I'll merge it in.

PR #993, opened 2019-12-14
Proved a property of Vectors related to lookup and Fin.splitAt

MirceaS:
Hi!
I needed to prove a certain fact about the interaction of lookup, Fin.splitAt and _++_ for a personal project and I thought it might be useful to have it in the standard library as well.
This is my first PR to this repository so I'm still learning about how everything should be structured. I did however read the styling sheet and ran the test suite but if there are any suggestions for making my code prettier or more readable I'm more than happy to oblige.


src/Data/Sum/Properties.agda
```diff
@@ -33,3 +34,19 @@ module _ {a b} {A : Set a} {B : Set b} where
...
+
+module _ where
+  private
+    variable
+      a b c d e : Level
+      A : Set a
+      B : Set b
+      C : Set c
+      D : Set d
+      E : Set e
```
MatthewDaggitt: It would be great if we could move the `variable` declaration to the top of the file and get rid of the first anonymous module?

src/Data/Sum/Properties.agda
```diff
@@ -33,3 +34,19 @@ module _ {a b} {A : Set a} {B : Set b} where
...
+
+  [,]-map-comm : ∀ {f′ : C → E} {f : A → C}
```
MatthewDaggitt: Looking at `Data.List.Properties` it appears we usually write out `commute` in full?

MatthewDaggitt: Also sets should occur in alphabetical order `A`, `B`, `C` etc. and `f` should appear before `f'`. We also normally use \prime instead of `. (Not particularly in favour of that last one but best to be consistent!)

src/Data/Vec/Properties.agda
```diff
@@ -340,6 +342,15 @@ lookup-++ʳ []       ys       zero    = refl
 lookup-++ʳ []       (y ∷ xs) (suc i) = lookup-++ʳ [] xs i
 lookup-++ʳ (x ∷ xs) ys       i       = lookup-++ʳ xs ys i
 
+lookup-splitAt : ∀ m {n} (xs : Vec A m) (ys : Vec A n) i →
+                lookup (xs ++ ys) i ≡ [ lookup xs , lookup ys ]′
+                (Fin.splitAt m i)
+lookup-splitAt zero    []       ys i       = refl
+lookup-splitAt (suc m) (x ∷ xs) ys zero    = refl
+lookup-splitAt (suc m) (x ∷ xs) ys (suc i) = P.subst id
+  (P.cong (_ ≡_) (P.sym ([,]-map-comm (Fin.splitAt m i))))
```
MatthewDaggitt: `P.cong (_ ≡_)` is simply an instance of `trans` and I'm pretty sure you can rewrite it without the `P.subst id`

src/Data/Sum/Properties.agda
MirceaS: I chose to use ′ as this was the apostrophe used in [_,_]′ which I'm proving my fact about, so it seemed appropriate to use that one but I can change it if you think that would be better.

src/Data/Vec/Properties.agda
MirceaS: I don't think it's an instance of trans, or at least I don't see it.
What's happening here is that the type of the inductive call is `LHS == RHS'` but what it needs to be is `LHS == RHS`. Fortunately `RHS' == RHS` which we get by calling `[,]-map-comm`.
We then have to use the `P.cong (_ ≡_)` to get our desired `(LHS == RHS') == (LHS == RHS)`.
`P.subst id` merely turns this equality into a function so that we can get our desired result.
I would be very surprised if this can be made any more concise than what it is now.

src/Data/Sum/Properties.agda
MirceaS: I did consider doing that but I suspect that this has the (unlikely but still) potential to cause a regression. That is in the case where the implicit arguments to the module are provided explicitly.

MirceaS: Screenshot from 2019-12-16 04-36-05
I also get this error when I try to change it.

MatthewDaggitt: > I did consider doing that but I suspect that this has the (unlikely but still) potential to cause a regression. That is in the case where the implicit arguments to the module are provided explicitly.

It shouldn't do as long as we use the variables in alphabetical order.

So in terms of the screenshot above:
1. The top-level anonymous module can be removed as it's empty.
2. You'll need to provide `{A = A}` and `{B = B}` as parameters to the assumptions of decidable equality. Previously Agda could infer them as the same `A` and `B` were used in the `injective` proofs but now they're technically different `A`s and `B`s.

src/Data/Sum/Properties.agda
MatthewDaggitt: Hmm at one point we had a drive to eliminate any `'` from the library but apparently they've crept in again *sigh* I'd change it to prime for the moment. I've opened an issue for it #995 in the general case.

src/Data/Vec/Properties.agda
MatthewDaggitt: > EDIT: Nevermind, I see it now. Just tried it and it works! That's so much better than what it was before!

Haha, great to hear it :smile: it's always satisfying when it just clicks into place!

src/Data/Sum/Properties.agda
MirceaS: I fixed this.
I believe a similar set of changes could be done for Product.Properties as well since the Product map has the same compositional property as the Sum map.

src/Data/Sum/Properties.agda
MirceaS: I changed the apostrophes of my f's and g's to \prime.
I couldn't however have the sets appear in alphabetical order due to the nature of the function arguments but I don't believe this to be a problem. They are now as alphabetical as they can be.

MirceaS: Also, I proved another very similar property along with this one called map-commute.
I don't think this is the proper name to call it as it's not commutativity but I don't really know what to call this binary operation property.
I believe there should be a new entry in Algebra.Definitions for the following Op2 interaction:
```
(w . x) * (y . z) = (w * y) . (x * z)
```

MatthewDaggitt:
Thanks for the changes! Looks good to go. Any chance you could add an entry to the CHANGELOG documenting the new proofs?

I'll then merge it in after a couple of days (just on the off chance anyone else has some comments).

src/Data/Sum/Properties.agda
MatthewDaggitt: > I believe there should be a new entry in Algebra.Definitions for the following Op2 interaction

Hmm that looks useful though I have no idea what to call it. Feel free to open another PR.

MirceaS:
Sorry, I’ve been (and still am) away for the Christmas holidays. I’ll try to add to the changelog one of these days if I get some time. Do you think I should open a new PR for a new entry in Algebra.Definitions rather than just commit to this one?

MatthewDaggitt:
> Sorry, I’ve been (and still am) away for the Christmas holidays. I’ll try to add to the changelog one of these days if I get some time.

No worries, no rush.

> Do you think I should open a new PR for a new entry in Algebra.Definitions rather than just commit to this one?

Yes that would be best!

MirceaS:
> Yes that would be best!
I opened a new PR for that and will update this one once that one gets merged in
https://github.com/agda/agda-stdlib/pull/1020

MirceaS:
I was looking to use the newly defined property of operations in the definition of map-commute (which I was also planning to change the name of) but since f, g, f' and g' it's not as simple as writing `Interchangable _o_ map` as we don't have function extensionality in agda (do we?).
I'm half tempted to just leave it as it is while just changing the name. Any ideas?

MatthewDaggitt:
> we don't have function extensionality in agda (do we?).

We don't. You could probably get around it by parameterising it with pointwise propositional equality `_≗_ ` from `Relation.Binary.PropositionalEquality`. 

> I'm half tempted to just leave it as it is while just changing the name. Any ideas?

I think I'm actually happy with it as is. The name and type agrees with other `commute` lemmas elsewhere in the library. I'm happy to merge it in if you are?

MirceaS:
@MatthewDaggitt Alright, I'm happy for it to be merged in now. 
But first, can you teach me how to add to the changelog? I can't really follow the documentation on this matter.

MatthewDaggitt:
> But first, can you teach me how to add to the changelog? I can't really follow the documentation on this matter.

Ah yup, good spot. In this case it's pretty simple. In the file CHANGELOG.md under the section `Other minor additions` you should add new bullet points `Added new functions to Data.Vec.Properties/Data.Sum.Properties:` and list the types of the lemmas you've added. In practice we tend to simplify the types by removing the implicit and explicit arguments to the lemmas, e.g.

 `[,]-map-commute : [ f′ , g′ ]′ ∘ (map f g) ≗ [ f′ ∘ f , g′ ∘ g ]′`

Hopefully it should be fairly easy to copy the pattern used from other entries in the file. Let me know if you have any questions!

MirceaS:
> > But first, can you teach me how to add to the changelog? I can't really follow the documentation on this matter.
> 
> Ah yup, good spot. In this case it's pretty simple. In the file CHANGELOG.md under the section `Other minor additions` you should add new bullet points `Added new functions to Data.Vec.Properties/Data.Sum.Properties:` and list the types of the lemmas you've added. In practice we tend to simplify the types by removing the implicit and explicit arguments to the lemmas, e.g.
> 
> `[,]-map-commute : [ f′ , g′ ]′ ∘ (map f g) ≗ [ f′ ∘ f , g′ ∘ g ]′`
> 
> Hopefully it should be fairly easy to copy the pattern used from other entries in the file. Let me know if you have any questions!

Thanks! I just wasn't sure whether new additions go on the top or on the bottom but looking at the file more carefully I figured that they go on the top.
It should be all good to merge now

MatthewDaggitt:
Great, merging in now. Thanks for the PR!

PR #994, opened 2019-12-15
Pretty printing tables of strings

gallais:
I am not dead set on the current names. Alternatives are welcome!


src/Data/Vec.agda
```diff
@@ -56,3 +59,20 @@ module _ {P : A → Set p} (P? : Decidable P) where
...
+
+------------------------------------------------------------------------
+-- Padding a list of vectors
+
+rectangle : ∀[ Vec≤ A ⇒ Vec A ] → List (∃ (Vec A)) → ∃ (List ∘ Vec A)
+rectangle f = Prod.map₂ (List.map f)
+            ∘ Vec≤.rectangle
+            ∘ List.map (Prod.map₂ fromVec)
+
+rectangleˡ : A → List (∃ (Vec A)) → ∃ (List ∘ Vec A)
```
MatthewDaggitt: I have to confess these function signatures are really doing my head in. They take a list of pairs `(n_1, Vec A n_1), ...,` and they presumably output a pair `(m, List (Vec A m))` where `m` is the maximum over `n_i`?

These seem like really specialised functions... I'm torn about whether or not they should live in `Data.Vec`? Actually are they even used in `Data.Vec.Show`? I can't seem to trace them through?

src/Data/Vec/Bounded/Base.agda
```diff
@@ -37,11 +42,55 @@ record Vec≤ (A : Set a) (n : ℕ) : Set a where
...
+... | less-than-or-equal refl = vs Vec.++ Vec.replicate a
+
+padLeft : ∀ {n} → A → Vec≤ A n → Vec A n
+padLeft a as@(vs , m≤n)
+  with recompute (_ ℕₚ.≤″? _) (ℕₚ.≤⇒≤″ m≤n)
+... | less-than-or-equal {k} ∣as∣+k≡n
+  with P.trans (ℕₚ.+-comm k (Vec≤.length as)) ∣as∣+k≡n
+... | refl = Vec.replicate a Vec.++ vs
+
+module _ where
```
MatthewDaggitt: Why is the anonymous module needed?

src/Data/List/Show/Core.agda
```diff
@@ -0,0 +1,279 @@
...
+  tr   = lineOf (c. row)
+  lbot = mlineOf (c. bot) header
+
+  th  = ltop ?∷ tr header ∷ []
+  trs = if null rows then id else (maybe _∷_ id lsep)
+      $ maybe intersperse id lsep
+      $ map tr rows
+
+------------------------------------------------------------------------
+-- examples
```
MatthewDaggitt: It might be better to lift these examples to `Data.List.Show` where more people are likely to find them?

src/Data/List/Show/Core.agda
```diff
@@ -0,0 +1,279 @@
...
+ascii .bot = just λ where
+  .left  → just "+"
+  .cont  → just '-'
+  .sep   → "-"
+  .right → just "+"
+
+compact : Config → Config
+compact c = record c { sep = nothing }
+
+module _ where
```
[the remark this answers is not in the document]
gallais: I guess it is useless. In my mind these two definitions where really grouped
together hence the module surrounding them.

src/Data/Vec.agda
gallais: They are not used. I wrote them before I wrote the `rectangle` ones for
`String` because I figured it was better for efficiency to work directly
on strings rather than to go via `Vec Char`. Their types however are more
precise and they could be used as specifications for the ones acting on
strings if we ever wanted to prove them equivalent!

src/Data/List/Show/Core.agda
gallais: I think I will put them all in a README.

src/Data/List/Show/Core.agda
```diff
@@ -0,0 +1,279 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Fancy display functions for List
+-- This is a core module because we have (unenforced) invariants.
+-- If you cannot guarantee that your data respects these invariants,
+-- you should instead use Data.List.Show.
```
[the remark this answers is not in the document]
gallais: I have had a revelation and am moving the whole hierarchy to `Text.Tabular`,
named after the popular LaTeX command! :D

gallais:
The README is breaking the whitespace check (we have trailing whitespace in the
string literals because they have been padded to match the column's width).

I will try to upgrade to https://github.com/agda/fix-whitespace which allows us to
explicitly list files that should not be checked.

src/Data/Vec.agda
MatthewDaggitt: > They are not used.

Should we remove them for the moment then until we find a use for them?

I'm also a bit worried about adding more stuff to `Data.Vec` itself. Eventually after some of the issues with re-exporting are fixed,  in v2.0 I hope to have the main modules for every data type act as summary modules, re-exporting much more of the sub-hierarchy. If you'd like to keep them perhaps we could put them in a submodule, `Data.Vec.Reshape` maybe?

src/Data/Vec/Base.agda
```diff
@@ -284,3 +287,10 @@ init .(ys ∷ʳ y) | (ys , y , refl) = ys
 last : ∀ {n} → Vec A (1 + n) → A
 last xs         with initLast xs
 last .(ys ∷ʳ y) | (ys , y , refl) = y
+
+------------------------------------------------------------------------
+-- Other operations
+
+transpose : ∀ {m n} → Vec (Vec A n) m → Vec (Vec A m) n
```
MatthewDaggitt: Can't help feeling this should be in a `Matrix` class... Don't have any good ideas about what to do with it at the moment though! Just thought I'd put it out there.

src/Data/Vec.agda
MatthewDaggitt: Additionally there doesn't seem to much of a pattern where functions that involve both `Vec` and `Vec≤` live. I kind of feel that either they should all live in the `Vec` folder or in the `Vec≤` folder. Thoughts?

src/Data/Vec/Bounded/Base.agda
```diff
@@ -80,3 +126,31 @@ zip = zipWith _,_
 
 align : ∀ {n} → Vec≤ A n → Vec≤ B n → Vec≤ (These A B) n
 align = alignWith id
+
+-- take and drop
+
+take : ∀ {m} n → Vec≤ A m → Vec≤ A (n ⊓ m)
+take             zero    _                = []
+take             (suc n) (Vec.[] , p)     = []
```
MatthewDaggitt: Do we need to qualify the `[]` and `_∷_` constructors here? Or is it just for clarity?

src/Text/Tabular/Base.agda
```diff
@@ -0,0 +1,166 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Fancy display functions for List-based tables
+--
+-- The functions in this module assume some (unenforced) invariants.
+-- If you cannot guarantee that your data respects these invariants,
+-- you should instead use Text.Tabular.List.
```
MatthewDaggitt: All of the text in the header gets sucked into `Everything.agda` right? Usually I think we just have a summary line in the header, and paragraphs like this are inserted immediately below the header.

src/Text/Tabular/Base.agda
```diff
@@ -0,0 +1,166 @@
...
+
+open String
+  using ( Alignment
+        ; Left
+        ; Center
+        ; Right
+        ) public
+
+record TabularLine : Set where
+  field left  : Maybe String
```
MatthewDaggitt: Could we align the fields below the `field` keyword in the record as we usually do? Need to add this to the style guide...

src/Data/Vec/Bounded/Base.agda
gallais: `Data.Vec.Bounded.Base` redefines `[]` and `_∷_` as *functions* so you
have constructor vs. functions and cannot disambiguate the overloaded names
on the RHS if you happen to import `Vec`'s constructors too.

src/Text/Tabular/Base.agda
gallais: I just had a look and you only get the first paragraph. As soon as you
encounter an empty line, the rest is dropped.

src/Data/Vec.agda
```diff
@@ -18,9 +18,12 @@ module Data.Vec where
 
 open import Level
 open import Data.Bool.Base
+open import Data.List.Base as List using (List)
 import Data.Nat.Properties as ℕₚ
+open import Data.Product as Prod using (∃)
 open import Data.Vec.Bounded.Base as Vec≤
   using (Vec≤; ≤-cast; fromVec)
+open import Function.Base
```
MatthewDaggitt: We don't need these extra imports anymore right?

src/Text/Tabular/Base.agda
```diff
@@ -0,0 +1,168 @@
...
+  .right → just "+"
+
+compact : TabularConfig → TabularConfig
+compact c = record c { sep = nothing }
+
+private
+  dropBorder : TabularLine → TabularLine
+  dropBorder l = record l { left = nothing; right = nothing }
+
+noborder : TabularConfig → TabularConfig
```
MatthewDaggitt: Should these be `noBorder` and `addSpace`?

gallais:
I have decided to change the return type of `display` so that we can use its
output to create a faithful `Doc` (in the `Text.Pretty` sense).

With `unlines`, we would get a single `String` which `Text.Pretty` assumes is
a single line. This means that `indent` for instance will not work properly.
With a list of strings instead, we can generate a proper `Doc` for the table
(via `flush . vcat . map text`).

PR #998, opened 2019-12-17
[ fix #939 ] add lift to RawMonadT

### LINKED ISSUE #939: Add `lift` for all the various monad transformers in the stdlib
I needed `lift : M a -> StateT S M a` and could not find it. I suspect we also
miss it for numerous other monad transformers (e.g. `MaybeT` in `Data.Maybe.Categorical`).

Probably worth extending the whole [`RawIMonadT`](https://agda.github.io/agda-stdlib/Category.Monad.Indexed.html#1339) interface to always require
a `lift`.

gallais:


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ Non-backwards compatible changes
   Definitions that are sensitive to the behaviour of these lemmas, rather than
   just their existence, may need to be revised.
 
+* `RawIMonadT T` has been turned into a record type to allow the addition
+  of a lift field of type `∀ {M i j A} → RawIMonad M → M i j A → T M i j A`.
```
MatthewDaggitt: This seems like a fairly big breaking in change in that users will have to add some new non-trivial code. Any chance you could write something brief giving them some hints about how to fix their code?

src/Category/Monad/State.agda
```diff
@@ -111,6 +111,19 @@ module RawMonadState {f} {S : Set f} {M : Set f → Set f}
...
 StateT S M = IStateT {I = ⊤} (λ _ → S) M _ _
 
+monadT : ∀ {i f} {I : Set i} (S : Set f) →
+         RawIMonadT (λ (M : IFun I f) i j → StateT S (M i j))
+monadT S = record
+  { rawIMonadT = rawIMonadT
+  ; lift       = λ M → let open RawIMonad M in λ m s → (_, s) <$> m
+  } where
+
+  rawIMonadT : ∀ {M} → RawIMonad M → RawIMonad (λ i j → StateT S (M i j))
```
MatthewDaggitt: Should all of these `rawIMonadT` be top-level proofs? (not having used this, I don't know!)

JacquesCarette: They are accessible through the record - I think that's enough.

JacquesCarette:
Well, this says that all (indexed) monad transformers, in the particular style given here, must have the equivalent of `pure` (one level up, here named `lift`). I'm not sure if that's justified categorically, but since this is really the programming version and, for example, programming monads are always assumed strong, this assumption doesn't seem out of place.

src/Category/Monad.agda
```diff
@@ -26,6 +26,8 @@ RawMonadZero M = RawIMonadZero {I = ⊤} (λ _ _ → M)
 RawMonadPlus : ∀ {f} → (Set f → Set f) → Set _
 RawMonadPlus M = RawIMonadPlus {I = ⊤} (λ _ _ → M)
 
+module RawMonadT = RawIMonadT
+
```
laMudri: Is it worth restricting the type of the parameter here, so that we can't use `module RawMonadT` on an indexed one? So, `module RawMonadT {f} {T} (MonT : RawMonadT {f} T) = RawIMonadT MonT`.

JacquesCarette: That is probably wise, else the results might be surprising.

gallais: I started with the restricted type but doing so leads to unsolved meta
variables in other parts of the PR.

MatthewDaggitt:
What's the latest on this PR? Should we bring it up to date or close it?

MatthewDaggitt:
Closing this. @gallais feel free to re-open if desired.

Issue #1003, opened 2019-12-18
Add `_≋⟨_⟩` combinator to `PermutationReasoning`


MatthewDaggitt:

MatthewDaggitt:
Fixed with 06c14ea

Issue #1004, opened 2019-12-18
Add aliases for `refl`, `sym` and `prep` which do not require equalities in `Permutation.Setoid`


MatthewDaggitt:

JacquesCarette:
Could you give an actual description? Right now, I think you might be the only one who really understands the intent of this issue. I feel like I ought to, but I don't.

MatthewDaggitt:
Ah sorry! At the moment I'm writing a lot of code of the form `swap refl refl p` and `prep refl p`. I'd quite like some shortcuts that would just allow me to write `prep′ p` and `swap′ p` for when the first elements are propositionally equal..

MatthewDaggitt:
Fixed with 06c14ea

PR #1006, opened 2019-12-20
[ fix ] change definition of `Permutation.refl`

### LINKED ISSUE #1002: Change definition of `refl` constructor in `Permutation.Heterogeneous`
At the moment it's `refl : Permutation R xs xs`. Although this looks fine, it's causing me severe problems as it breaks the size preserving property for various proofs involve `≋`. e.g. if `xs ≋ ys` and `ys ↭ zs` then the proof of `xs ↭ zs` is counter-intuitively much bigger than the proof of `ys ↭ zs` when it should be the same size.

To fix this I propose changing the definition of `refll` to `refl : Pointwise R xs ys →Permutation R xs ys`

MatthewDaggitt:
Fixes #1002


MatthewDaggitt:
Now also fixes #1003 and #1004.

PR #1009, opened 2019-12-20
add examples of renaming, using directives; minor grammar check

pnlph:
1. Adding some examples of the import directives renaming and using.
1. Small spell/grammar cleanup.
1. Underscore and hyphen explanation.

PR #1010, opened 2019-12-20
enumerated items import statements

pnlph:
in README.Data.Nat


README/Data/Nat.agda
```diff
@@ -12,31 +12,34 @@ module README.Data.Nat where
...
 ex₂ = refl
 
 -- Data.Nat.Properties contains a number of properties about natural
 -- numbers.
 
-import Data.Nat.Properties as Nat
+import Data.Nat.Properties as Nat using (*-comm; +-identityʳ)
 
 ex₃ : ∀ m n → m * n ≡ n * m
 ex₃ m n = Nat.*-comm m n
```
MatthewDaggitt: Should we remove the `Nat.` qualification here then?

README/Data/Nat.agda
```diff
@@ -12,31 +12,34 @@ module README.Data.Nat where
...
 
 -- The module ≡-Reasoning in Relation.Binary.PropositionalEquality
 -- provides some combinators for equational reasoning.
 
-open ≡-Reasoning
+open Relation.Binary.PropositionalEquality using (cong; module ≡-Reasoning)
+open ≡-Reasoning using (begin_; _≡⟨_⟩_; _∎)
```
MatthewDaggitt: Hmm I'm wondering if qualifying import statements for reasoning and solver modules is perhaps overkill? I certainly don't want to have to do so every time I use them in the main library...

README/Data/Nat.agda
pnlph: Yes, because of redundancy. I did not realize.

README/Data/Nat.agda
pnlph: I see you point.
It took me some time to figure out where the constructors and operators exactly come from and what do they exactly mean, so I guess for someone that is starting with the stdlib from the README files makes sense, but perhaps this must not be extensively applied to the library.

MatthewDaggitt: Okay, fair enough, that's a good argument.

README/Data/Nat.agda
MatthewDaggitt: Okay, once you've removed it I'm happy and I'll merge this in.

pnlph: OK, done!

MatthewDaggitt:
@pnlph this one has the same problem as #1009. Any chance you could revert the unneeded commits?

pnlph:
Of course, unneeded commits reverted.

PR #1011, opened 2019-12-20
[ new ] port of @jyp's pretty printing library

gallais:


src/Data/Product/Relation/Unary/All.agda
```diff
@@ -0,0 +1,23 @@
...
+open import Relation.Unary
+
+private
+  variable
+    a b p q : Level
+    A : Set a
+    B : Set b
+
+All : (A → Set p) → (B → Set q) → (A × B → Set (p ⊔ q))
+All P Q (a , b) = P a × Q b
```
MatthewDaggitt: Just a heads-up that this already exists in `Relation.Unary`. Do we want to re-use the definition?

src/Data/Erased.agda
```diff
@@ -0,0 +1,43 @@
...
+--
+-- Wrapper for the erased modality
+-- This allows us to store erased proofs in a record an use projections
+-- to manipulate them without having to turn on the unsafe option
+-- --irrelevant-projections.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Erased where
```
MatthewDaggitt: Not sure about this, but I'm wondering if this might sit better under `Relation.Unary.Erased` rather than `Data.Erased`?

src/Data/Erased.agda
```diff
@@ -0,0 +1,43 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Wrapper for the erased modality
+-- This allows us to store erased proofs in a record an use projections
```
MatthewDaggitt: Is the `an` a typo? Also as mentioned in #994 the second paragraph should probably go just below the header? (added this to the list of things to add to the style guide!)

src/Data/Erased.agda
```diff
@@ -0,0 +1,43 @@
...
+open import Level using (Level)
+
+private
+  variable
+    a b c : Level
+    A : Set a
+    B : Set b
+    C : Set c
+
+record Erased (A : Set a) : Set a where
```
MatthewDaggitt: Could we have a few more subheaders in these new files? e.g.
```agda
-- Definition

-- Functions
```

src/Data/Refinement.agda
```diff
@@ -0,0 +1,45 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Refinement type: a value together with an erased proof.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Refinement where
```
MatthewDaggitt: Again, should this be under `Data`? (I'm not sure, just putting the thought out there!)

src/Data/String/Unsafe.agda
```diff
@@ -21,3 +24,19 @@ toList∘fromList s = trustMe
 
 fromList∘toList : ∀ s → fromList (toList s) ≡ s
 fromList∘toList s = trustMe
+
+toList-++ : ∀ s t → toList (s ++ t) ≡ toList s List.++ toList t
+toList-++ s t = trustMe
```
MatthewDaggitt: Any thoughts on what we need to add to 2.6.2 in order to make these safe?

src/Data/String/Unsafe.agda
```diff
@@ -21,3 +24,19 @@ toList∘fromList s = trustMe
...
+
+length-++ : ∀ s t → length (s ++ t) ≡ length s + length t
+length-++ s t = begin
+  length (s ++ t)                         ≡⟨⟩
+  List.length (toList (s ++ t))           ≡⟨ cong List.length (toList-++ s t) ⟩
+  List.length (toList s List.++ toList t) ≡⟨ Listₚ.length-++ (toList s) ⟩
+  length s + length t                     ∎ where open ≡-Reasoning
```
MatthewDaggitt: Can we open `≡-Reasoning` at the top of the file rather than opening multiple times?

src/Data/Tree/Binary.agda
```diff
@@ -0,0 +1,47 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Binary Trees
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Tree.Binary where
```
MatthewDaggitt: Haha, I sense this is another awkward module organisation conversation coming up. Seems inconsistent to have `Data.AVL` but `Data.Tree.Binary`... Personally I'd be in favour of having `Data.Tree.AVL` but obviously that involves another round of deprecations.

src/Data/Tree/Binary.agda
```diff
@@ -0,0 +1,47 @@
...
+foldr nd lf (node l m r) = nd (foldr nd lf l) m (foldr nd lf r)
+
+------------------------------------------------------------------------
+-- Extraction to lists, depth first.
+
+toDiffList : Tree A → DiffList A
+toDiffList = foldr (λ l m r → l ++ m ∷ r) []
+
+toList : Tree A → List A
+toList = DiffList.toList ∘′ toDiffList
```
MatthewDaggitt: We should probably have a more specific name for this, as we may want to add different `toList` functions in the future that extract the items in different orders.

src/Data/Tree/Binary/Relation/Unary/All.agda
```diff
@@ -0,0 +1,35 @@
...
+
+module _ {P : A → Set p} {Q : A → Set q} where
+
+  map : ∀[ P ⇒ Q ] → ∀[ All P ⇒ All Q ]
+  map f leaf         = leaf
+  map f (node l m r) = node (map f l) (f m) (map f r)
+
+module _ {P : B → Set p} where
+
+  map⁺ : (f : A → B) → ∀[ All (f ⊢ P) ⇒ Tree.map f ⊢ All P ]
```
MatthewDaggitt: I think this usually goes in `All.Properties`? (excepting the odd cases of `List.Relation.Binary.Equality/Pointwise` that I need to fix at some point).

src/Text/Pretty/Core.agda
```diff
@@ -0,0 +1,343 @@
...
+node? (just (x , xs)) y ys = just (x , node xs y ys)
+node? nothing         y ys = just (y , ys)
+
+∣node?∣ : ∀ b y ys → size (node? b y ys)
+                   ≡ size b + suc (Tree.size ys)
+∣node?∣ (just (x , xs)) y ys = refl
+∣node?∣ nothing         y ys = refl
+
+≤-Content : ∀ {m n} {b : Content} → m ≤ n → All≤ m b → All≤ n b
+≤-Content {m} {n} m≤n = Allᴹ.map (Prod.map step (Allᵀ.map step)) where
```
MatthewDaggitt: the `where` should be on the next line

src/Text/Pretty/Core.agda
```diff
@@ -0,0 +1,343 @@
...
+
+  step : ∀ {p} → p ≤ m → p ≤ n
+  step = flip ≤-trans m≤n
+
+All≤-node? : ∀ {l m r n} →
+  All≤ n l → length m ≤ n → Allᵀ.All (λ s → length s ≤ n) r →
+  All≤ n (node? l m r)
```
MatthewDaggitt: The types should be aligned with the colon

src/Text/Pretty/Core.agda
```diff
@@ -0,0 +1,343 @@
...
+
+    pad : Maybe String
+    pad with x.lastWidth
+    ... | 0 = nothing
+    ... | l = just (replicate l ' ')
+
+    size-pad : maybe′ length 0 pad ≡ x.lastWidth
+    size-pad with x.lastWidth
+    ... | 0         = refl
+    ... | l@(suc _) = length-replicate l
```
MatthewDaggitt: Presumably these will go once #994 gets merged in?

src/Text/Pretty/Core.agda
```diff
@@ -0,0 +1,343 @@
...
+  All≤ n l → length m ≤ n → Allᵀ.All (λ s → length s ≤ n) r →
+  All≤ n (node? l m r)
+All≤-node? nothing           py pys = just (py , pys)
+All≤-node? (just (px , pxs)) py pys = just (px , node pxs py pys)
+
+------------------------------------------------------------------------
+-- Appending two documents
+
+private
+  module append (x y : Block) where
```
MatthewDaggitt: These modules `append` and `flush` look big enough to live in their own files.

README/Text/Pretty.agda
```diff
@@ -0,0 +1,189 @@
...
+open import Data.List.NonEmpty as List⁺
+open import Data.Nat.Base
+open import Data.Product
+open import Data.String.Base
+open import Data.Vec.Base as Vec
+open import Function.Base
+
+-- We import the pretty printer and pass 80 to say that we do not want to
+-- have lines longer than 80 characters
+open import Text.Pretty 80
```
MatthewDaggitt: Will there be other options we'll want to pass in later on? If so is there someway we can future proof this so we don't end up breaking code every time we do so?

src/Data/Erased.agda
gallais: It's not saying anything about a value of type `A`, it is wrapping one.
So I'd say it's definitely on the `Data` side of things.

src/Data/Refinement.agda
gallais: I see it as a `Data.Product` where the second component is erased.
But I guess it could also be seen as a new quantifier in `Relation.Unary`.

src/Data/String/Unsafe.agda
gallais: It would make sense to add axioms specifying these BUILTIN functions.
These types are pretty annoying to write in internal syntax though! :D

src/Data/Tree/Binary/Relation/Unary/All.agda
gallais: Yup. I was too lazy to do it in gallais/agda-pretty-notgreedy and it came
back to bite me!

README/Text/Pretty.agda
gallais: Jean-Philippe's algorithm only takes one parameter. It is proven correct
(wrt to the 3 criterion he specifies in the article) only for this parameter.
I am not planning to add any other options.

src/Text/Pretty/Core.agda
gallais: I would have to look into it. Given that we use the same pad for multiple
lines, I would really like to generate the string only once and use it
for all of them.

src/Text/Pretty/Core.agda
gallais: Then again this is a `Core` module and no one should look at it! :D

PR #1012, opened 2019-12-20
Author

JacquesCarette:
I've done enough bits that I can have my name in there now.


gallais:
I think you may have started this PR by forking from #971 rather than master so
the log looks pretty noisy.

JacquesCarette:
Yes, it does look like that.  I'll close this one and restart on master.

Issue #1016, opened 2019-12-23
Add undeprecated versions of deprecated lemmas for _<_ in `Data.Integer.Properties`


MatthewDaggitt:
See #752 

At a minimum includes:
```agda
m+1≤n⇒m<n : ∀ {m n} → sucℤ m ≤ n → m < n
m<n⇒m+1≤n : ∀ {m n} → m < n → sucℤ m ≤ n
m≤pred[n]⇒m<n : ∀ {m n} → m ≤ pred n → suc m ≤ n
m<n⇒m≤pred[n] : ∀ {m n} → suc m ≤ n → m ≤ pred n
```

Blaisorblade:
s/#752/#756.

Blaisorblade:
Also: now that @gallais agda-presburger's compiles against stdlib `master`, I think a good test would be to upgrade it to remove the deprecation warnings. I suspect most lemmas on `<′` need an undeprecated statement — `<′-irrelevant` has one already. And ideally the deprecation warnings should be updated.

MatthewDaggitt:
I've just had a scan through and after #1017 it looks like every deprecated lemma has a non-deprecated counter-part. I'm therefore closing this issue. 

> Also: now that @gallais agda-presburger's compiles against stdlib master, I think a good test would be to upgrade it to remove the deprecation warnings.

Not sure who's in charge of maintaining this, but whoever it is they can take a shot at it now.

Blaisorblade:
> I've just had a scan through and after #1017 it looks like every deprecated lemma has a non-deprecated counter-part.

I think there's a misunderstanding — that issue wasn't fixed after my PR.
Take `m≤pred[n]⇒m<′n : ∀ {m n} → m ≤ pred n → m <′ n`, which is mentioned already in the OP.
I'm asking to add `m≤pred[n]⇒suc[m]≤n : ∀ {m n} → m ≤ pred n → suc m ≤ n`, define `m≤pred[n]⇒m<′n = m≤pred[n]⇒suc[m]≤n`, and mention `m≤pred[n]⇒suc[m]≤n` in the deprecation message.

MatthewDaggitt:
> I think there's a misunderstanding

Ah yes there has been, apologies. Thanks for the example, I think I'm now on the same page as you.

[...]

PR #1017, opened 2019-12-23
Relate < and ≤ (part of #1016)

Blaisorblade:
This is just the first part, and I'm out of time for now. Simpler proofs might exist.


MatthewDaggitt:
Oh and we don't usually put the implicits `{m n}` in the changelog.

MatthewDaggitt:
Great, thanks! I'll leave it up a couple more days in case anyone else has any comments and then I'll merge it in.

PR #1020, opened 2019-12-24
Added additional property of 2 binary operations in Algebra.Definitions

MirceaS:


MatthewDaggitt:
Looks good to me. Though I might as well ask if anyone knows of a more technical name for this property?

JacquesCarette:
I knew I'd see this one before - in monoidal categories, this is a consequence of Functor homomorphism, i.e. tensor distributes over composition (and vice-versa). "distributes over" is not really much better than "commutes with" - [some certainly argue](https://math.stackexchange.com/questions/882206/what-are-some-binary-operations-that-commute) that that terminology is non-standard. 

The [nLab page](https://ncatlab.org/nlab/show/Eckmann-Hilton+argument) for the Eckmann-Hilton lemma says

> if a set is equipped with two monoid structures, such that one is a homomorphism for the other

which argues for "homomorphism for" -- though that seems weak/awkward, since this isn't sufficiently indicative of arity.

Further down that same page, this is called an "exchange law" (Wikipedia calls it an "interchange law"). Which is also the name used for the property of vertical/horizontal composition of natural transformations (when it is named at all).

So maybe `Exchange` ?

MatthewDaggitt:
> Further down that same page, this is called an "exchange law" (Wikipedia calls it an "interchange law"). Which is also the name used for the property of vertical/horizontal composition of natural transformations (when it is named at all).
>
> So maybe Exchange ?

Perhaps it should be an adjective in keeping with the rest of the names? How about `Exchangable` or `Interchangable`?

fredrikNordvallForsberg:
A little bit of further digging suggests the terminology that _∘_ *abides* with _∙_, from 
Richard Bird's [Lectures on Constructive Functional Programming](
https://www.cs.ox.ac.uk/files/3390/PRG69.pdf#page=39) (page 37). Apparently it's an abbreviation of above-besides. 

This could be made into an adjective `Abiding`? It's not great, because it doesn't tell (at least to me) what it means without looking it up, but at least it's from the published literature, and it avoids the false impression that the two operations are interchangeable, which the other two adjectives suggest to me.

JacquesCarette:
If we're going to go with "published" material [wikipedia usually draws from that, as does nLab, but oh well], then
- Mac Lane's CWM 2nd ed., p.  274 calls it the "middle four exchange" axiom. [Yuck!]
- interestingly, p. 164 of CWM, when talking about monoidal categories, calls it the "interchange law"!
- Paul Taylor's Practical Foundations books calls it the "middle four interchange" law.
- Googling for "abides", the [first relevant hit](https://math.stackexchange.com/questions/628236/name-for-this-axiom-x-bullet-r-sqcup-y-bullet-s-equiv-x-sqcup-y-bul) is to a math.SE question that quotes Bird in a comment, but also "interchange" in an answer.
- One can also find a reference by [Fokkinga](https://maartenfokkinga.github.io/utwente/mmf92b.pdf) who also credits Bird.
- In Concurrent Kleene Algebras (with containment instead of equality"), this is called the "exchange law" - [for example](https://www.researchgate.net/publication/220700994_On_Locality_and_the_Exchange_Law_for_Concurrent_Processes)
- Edmunds calls it "interchange law" and create [interchange rings](https://arxiv.org/abs/1402.3699) from it ( [published version](https://www.cambridge.org/core/journals/journal-of-the-australian-mathematical-society/article/interchange-rings/F0C8199A8C638F1D17A929B3D8AE5ADC) )
- under "interchange law", the study of this goes back to at least the 1960s in universal algebra -- see [this paper](https://www.cambridge.org/core/journals/proceedings-of-the-royal-society-of-edinburgh-section-a-mathematics/article/viendomorphisms-of-abstract-algebras/4C6FB097A68F21E13CB6EF4F4847D55A) by Evans, and even the 50s by Godement.

My current preference is `Interchangeable`, mostly because of historical weight, even though I understand that they are not 'interchangeable' operations.

fredrikNordvallForsberg:
Sorry, I didn't mean to imply that your suggestions where not "published". I was rather trying to convince myself that "abides" is not completely crazy. 

Off the top of my head, I would also call this "the interchange law", its mostly the adjective version I'm not so happy with. I think I would be happier with simply `Interchange` or `Exchange`, although of course I understand the original motivation for using an adjective.

JacquesCarette:
Thanks for the clarification - and I'm actually quite happy with the results, which pushed me to investigate more deeply in the literature.  `abide` is not crazy, but still feels too artificial.

If I were naming an axiom, I'd probably lean towards `ExchangeLaw` or `InterchangeLaw`. It's unfortunate that adjectives have been chosen, as not as many concepts have been 'adjectivized' (!) one would prefer...

MirceaS:
That’s some great answers so far. I’m leaning towards something like `_Interchanges_` as well. I considered `_InterchangesWith_` at first but I feel like this might be a bit more ambiguous for the reasons quoted by Fredrik. Just the verb by itself however feels different while still keeping in line with literature.
What do you think professor @JacquesCarette ?

MatthewDaggitt:
Thanks for everyone's contributions! In order of preference I'd go for `Interchangeable`, `InterchangeLaw`, `_InterchangableWith_`. I'm against `_InterchangesWith_`/`_Interchanges_` as they doesn't make much grammatical sense to me.

PR #1026, opened 2019-12-28
Add `Nat.Binary.Minus`

mechvel:
New    src/Algebra/Morphism/ToSetoid.agda,    src/Algebra/Morphism/TwoMagmas.agda
          src/Data/Nat/Binary/Minus.agda;
Changed:   src/Data/Nat/Binary/Properties.agda,   src/Data/Nat/Properties.agda.

Please, consider this commit of December 28, 2019.
 All your notes on the previous  Binary.Minus version are taken in account.
* (1) New Data.Nat.Binary.Minus committed. 
* (2) New Minus requires certain small additions to Binary.Properties. These additions are done.                                                                                
* (3) G. Allais suggested to simplify the proof for `Binary.Minus.fromℕ-homo-∸`   by using a certain generic function for Homomorphism.  This simplification is done for ``_+_, _*_, _∸_`` on `ℕᵇ`.                            
    A similar simplification is expected for future `fromℤ-homo-`.                      
*  (4) To implement (3), there are added the modules  Algebra.Morphism.ToSetoid,    Algebra.Morphism.TwoMagmas.  ``ToSetoid`` introduces `_≈∘_` - a generalization for `_≗_`.                                  
``TwoMagmas`` implements the following lemma on two magmas:   
"If f and g are mutually inverse maps between A and B, g is congruent, f is a homomorphism, 
then g is a homomorphism."


mechvel:
Here are the responses to the notices on the previous commit version.
src/Data/Nat/Binary/Minus.agda                                                            
...
``open import Data.Fin using () renaming (zero to 0F; suc to sucF)``                          
          
@gallais                                                                                
> Instead of renaming these and writing 0F, sucF 0F and sucF (sucF 0F),                   
> you could use ``#_``: all your cases are concrete so Agda should see that                   
> you can indeed embed 0, 1, and 2 into Fin 3.                                            
                                                                                          
Done.

```
... | yes _ = 0ᵇ                                                                          
... | no  _ = 1+[2 (x ∸ y) ] 
```
@gallais:
> Same comments as earlier: avoid yes / no whenever you do not need the proofs            
> and use does instead. This will give you more efficient functions (essentially          
> computing a boolean).                                                                   
                                                                                          
I tried there                                                                             
```                                                                                       
2[1+ x ] ∸ 1+[2 y ] with does (x <? y)                                                    
... | true  = 0ᵇ                                                                          
... | false = 1+[2 (x ∸ y) ]                                                              
```                                                                                       
And it reports                                                                            
"The pattern variable true has the same name as the constructor  Agda.Builtin.Bool.Bool.true                                                              
 when checking the clause   Data.Nat.Binary.Minus.with-234 x y true = 0ᵇ                                             
"                                                                                         
-- no matter whether true, false imported from Bool or not.                                                                            
But a similar code in a certain my application works under  lib-1.2.                      
What makes this difference, I wonder. 
Is this due to --without-K --safe ?

mechvel:
@gallais 
```
src/Data/Nat/Binary/Minus.agda                                                            
x>y⇒x∸y≢0 {x'@(2[1+ x ])} {zero} _ =  subst (_≢ 0ᵇ) x'≡x'∸0 2[1+x]≢0                      
  where                                                                                   
  x'≡x'∸0 = sym (x∸0≡x x')                                                                
```                                                                                      
> I think we can inline such trivial proofs.                                              
                                                                                          
This is only the first pattern for ``x>y⇒x∸y≢0``.            
Sorry, I have separated these patterns with empty lines.                                    
Now I remove these empty lines, to dutifully make this body difficult to read.                                                                                  

And still, if it was a single pattern, and we inline it, then what would we have to write instead of, say,                                              
```  
f x y x>y =  g (x + y) (x>y⇒x∸y≢0 x>y)                                                  
```
?                                                                                         
Probably:
```                                                                                 
  f x y x>y =  g (x + y) x∸y≢0                                                            
               where                                                                      
               x∸y≢0 = subst (_≢ 0ᵇ) (sym (x∸0≡x x')) 2[1+x]≢0  -- (1)                    
```
And each time one will need to invent a proof like (1) ...

mechvel:
@gallais 
```
src/Data/Nat/Binary/Minus.agda                                                                                                 
... | tri≈ _ x≡y _ =  begin                                                               
  toℕ (x ∸ y)  ≡⟨ cong toℕ (x≤y⇒x∸y≡0 {x} {y} x≤y) ⟩                                      
  toℕ 0ᵇ       ≡⟨ refl ⟩                                                                  
```                                
> `≡⟨ refl ⟩`  can be replaced with ≡⟨⟩ (also applies to other lines in the file).           
                                                                                          
Done.

mechvel:
@gallais 
```
fromℕ-homo-∸ :  ∀ m n → fromℕ (m ℕ.∸ n) ≡ (fromℕ m) ∸ (fromℕ n)                           
fromℕ-homo-∸ m n =  begin-equality                                                        
```                                                                                          
> Given that we have fromℕ-toℕ, it feels like we ought to be able to prove a more general lemma. i.e.                                                                
```
>  (f : ℕ → ℕ) (g : ℕᵇ → ℕᵇ) →                                                            
>     (∀ m n → toℕ (g m n) ≡ f (toℕ m) (toℕ n)) →                                            
>     (∀ m n → fromℕ (f m n) ≡ g (fromℕ m) (fromℕ n))                                        
```                                                                                   
> and the converse.                

Probably you mean  ``... (f : Op₂ ℕ) (g : Op₂ ℕᵇ) → ...``.
This is implemented -- see the first message on this PR.
                                                                                          
foo : (f : Op₂ ℕ) (g : Op₂ ℕᵇ) →

src/Data/Nat/Properties.agda
```diff
@@ -554,6 +554,9 @@ suc[pred[n]]≡n {suc n} n≢0 = refl
   { isCommutativeMonoid = +-0-isCommutativeMonoid
   }
 
+∸-magma : Magma 0ℓ 0ℓ
+∸-magma = magma _∸_
```
MatthewDaggitt: Needs an entry in the changelog

src/Data/Nat/Binary/Properties.agda
```diff
@@ -730,6 +740,15 @@ private
   { isCommutativeMonoid = +-0-isCommutativeMonoid
   }
 
+import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup
+  as Bin+CSemigroup
+import Algebra.Morphism.TwoMagmas +-magma ℕₚ.+-magma
+  as Magmas+ℕᵇ-ℕ
```
MatthewDaggitt: `import` statements should be at the top of the file

CHANGELOG.md
```diff
@@ -149,3 +161,5 @@ Other minor additions
   *-1-monoid              : Monoid 0ℓ 0ℓ
   *-1-commutativeMonoid   : CommutativeMonoid 0ℓ 0ℓ
   ```
+
+* Several minor proofs added to `Data.Nat.Binary.Properties`.
```
MatthewDaggitt: Please list these

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+   module ≡-Reasoning)
+open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Algebra.FunctionProperties {A = ℕᵇ} _≡_
+open import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup
+  using (xy∙z≈y∙xz; x∙yz≈y∙xz)
+open import Algebra.Solver.CommutativeMonoid +-0-commutativeMonoid
+
+private module FP-Bin = FuncProp (_≡_ {A = ℕᵇ})
+open Tri
```
MatthewDaggitt: Please import the constructors from `Relation.Binary` directly rather than opening `Tri` again

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+  xN = toℕ x;   yN = toℕ y;   x'N = toℕ x';   y'N = toℕ y'
+
+  xN≤x'N = toℕ-mono-≤ x≤x';   y'N≤yN = toℕ-mono-≤ y'≤y
+
+∸-monoˡ-≤ :  (x : ℕᵇ) → (_∸ x) Preserves _≤_ ⟶ _≤_
+∸-monoˡ-≤ x y≤z =  ∸-mono-≤ y≤z (≤-refl {x})
+
+x∸y≤x :  ∀ x y → x ∸ y ≤ x
+x∸y≤x x y =  begin
+  x ∸ y    ≤⟨ ∸-mono-≤ (≤-refl {x}) (0≤x y) ⟩
```
MatthewDaggitt: You could split this proof out to the right version of `∸-monoˡ-≤`

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+*-distribˡ-∸ :  FP-Bin._DistributesOverˡ_ _*_ _∸_
+*-distribˡ-∸ x y z =  begin-equality
+  x * (y ∸ z)                         ≡⟨ sym (fromℕ-toℕ (x * (y ∸ z))) ⟩
+  fromℕ (toℕ (x * (y ∸ z)))           ≡⟨ cong fromℕ (toℕ-homo-* x (y ∸ z)) ⟩
+  fromℕ (k ℕ.* (toℕ (y ∸ z)))         ≡⟨ cong (fromℕ ∘ (k ℕ.*_)) (toℕ-homo-∸ y z) ⟩
+  fromℕ (k ℕ.* (m ℕ.∸ n))             ≡⟨ cong fromℕ (ℕₚ.*-distribˡ-∸ k m n) ⟩
+  fromℕ ((k ℕ.* m) ℕ.∸ (k ℕ.* n))     ≡⟨ fromℕ-homo-∸ (k ℕ.* m) (k ℕ.* n) ⟩
+  fromℕ (k ℕ.* m) ∸ fromℕ (k ℕ.* n)   ≡⟨ cong₂ _∸_ (fromℕ-homo-* k m)
+                                                   (fromℕ-homo-* k n) ⟩
+    (fromℕ k * fromℕ m) ∸ (fromℕ k * fromℕ n)
```
MatthewDaggitt: Alignment has gone wrong here

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+    toℕ (y + (x ∸ y))    ≡⟨ cong toℕ (x+[y∸x]≡y y≤x) ⟩
+    m                    ≡⟨ sym (ℕₚ.m+[n∸m]≡n n≤m) ⟩
+    n ℕ.+ (m ℕ.∸ n)      ≡⟨ ℕₚ.+-comm n (m ℕ.∸ n) ⟩
+    (m ℕ.∸ n) ℕ.+ n      ∎
+
+∸-magma : Magma 0ℓ 0ℓ
+∸-magma = magma _∸_
+
+import Algebra.Morphism.TwoMagmas ∸-magma ℕₚ.∸-magma
+  as Magmas∸ℕᵇ-ℕ
```
MatthewDaggitt: Import statements should go at the top of the file

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+2[1+ x ] ∸ 2[1+ y ] = double (x ∸ y)
+1+[2 x ] ∸ 1+[2 y ] = double (x ∸ y)
+2[1+ x ] ∸ 1+[2 y ] with x <? y
+... | yes _ = 0ᵇ
+... | no _  = 1+[2 (x ∸ y) ]
+1+[2 x ] ∸ 2[1+ y ] with x ≤? y
+... | yes _ = 0ᵇ
+... | no _  = pred (double (x ∸ y))
+
+x∸0≡x : ∀ x → x ∸ 0ᵇ ≡ x
```
MatthewDaggitt: Please could you have a look at `Data.Nat.Properties` and add sections in the same style?

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+  (_≡_; _≢_; _≗_; refl; sym; trans; cong; cong₂; resp₂; subst; magma;
+   module ≡-Reasoning)
+open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Algebra.FunctionProperties {A = ℕᵇ} _≡_
+open import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup
+  using (xy∙z≈y∙xz; x∙yz≈y∙xz)
+open import Algebra.Solver.CommutativeMonoid +-0-commutativeMonoid
+
+private module FP-Bin = FuncProp (_≡_ {A = ℕᵇ})
```
MatthewDaggitt: Why not just open this directly at the top of the file rather than creating a new module? In fact you are already doing this, so I don't think you need this module at all.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+--
+-- Subtraction on Bin and some of its property proofs.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Nat.Binary.Minus where
+
+open import Algebra using (Magma)
+import Algebra.FunctionProperties as FuncProp
```
MatthewDaggitt: `Algebra.FunctionProperties` is deprecated. Please use `Algebra.Definitions` instead.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+
+odd∸even-for> : ∀ {x y} → x > y → 1+[2 x ] ∸ 2[1+ y ] ≡ pred (double (x ∸ y))
+odd∸even-for> {x} {y} x>y  with x ≤? y
+... | no _    = refl
+... | yes x≤y = contradiction x>y (≤⇒≯ x≤y)
+
+x≤y⇒x∸y≡0 :  ∀ {x y} → x ≤ y → x ∸ y ≡ 0ᵇ
+x≤y⇒x∸y≡0 {zero} {_} _           =  refl
+x≤y⇒x∸y≡0 (inj₁ (even<even x<y)) =  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (inj₁ x<y))
+x≤y⇒x∸y≡0 {2[1+ x ]} {2[1+ .x ]} (inj₂ refl) =
```
MatthewDaggitt: Dot patterns are no longer necessary

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+x≤y⇒x∸y≡0 {1+[2 _ ]} {1+[2 _ ]} (inj₁ (odd<odd x<y)) =
+  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (inj₁ x<y))
+x≤y⇒x∸y≡0 {1+[2 x ]} {1+[2 .x ]} (inj₂ refl) =
+  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (≤-refl {x}))
+
+x∸x≡0 : ∀ x → x ∸ x ≡ 0ᵇ
+x∸x≡0 x =  x≤y⇒x∸y≡0 (≤-refl {x})
+
+x>y⇒x∸y≢0 : ∀ {x y} → x > y → x ∸ y ≢ 0ᵇ
+x>y⇒x∸y≢0 {zero} {_} ()
```
MatthewDaggitt: You should be able to omit this case entirely?

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+x>y⇒x∸y≢0 {x'@(2[1+ x ])} {zero} _ =  subst (_≢ 0ᵇ) x'≡x'∸0 2[1+x]≢0
+  where
+  x'≡x'∸0 = sym (x∸0≡x x')
```
MatthewDaggitt: This proof is short enough that you don't need to inline it. Alternatively you could use the `rewrite` syntax.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+  1+[2 (x∸y + y) ]               ≡⟨ cong 1+[2_] $ sym $ x≥y⇒x≡[x∸y]+y (inj₁ y<x) ⟩
+  1+[2 x ]                       ∎
+  where open ≤-Reasoning;  x∸y = x ∸ y
+x≥y⇒x≡[x∸y]+y {x'@(1+[2 x ])} {1+[2 .x ]} (inj₂ refl) =  sym $ begin-equality
+  (x' ∸ x') + x'  ≡⟨ cong (_+ x') (x∸x≡0 x') ⟩
+  0ᵇ + x'         ≡⟨ sym (+-identityˡ x') ⟩
+  x'              ∎
+  where open ≤-Reasoning
+
+[x+y]∸y≡x :  ∀ x y → (x + y) ∸ y ≡ x
```
MatthewDaggitt: Please could you stop adding the double spaces after the colons and equals signs. You only do so inconsistently. I know it allows aligning in literate Agda but the standard library doesn't tend to be used for literate Agda.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+∸-magma : Magma 0ℓ 0ℓ
+∸-magma = magma _∸_
+
+import Algebra.Morphism.TwoMagmas ∸-magma ℕₚ.∸-magma
+  as Magmas∸ℕᵇ-ℕ
+fromℕ-homo-∸ :  ∀ m n → fromℕ (m ℕ.∸ n) ≡ (fromℕ m) ∸ (fromℕ n)
+fromℕ-homo-∸ =  Magmas∸ℕᵇ-ℕ.IsHomo⇒IsHomo-rev toℕ fromℕ (cong fromℕ)
+                                       fromℕ-toℕ toℕ-fromℕ toℕ-homo-∸
+
+∸-+ :  ∀ x y o → (x ∸ y) ∸ o ≡ x ∸ (y + o)
```
MatthewDaggitt: Please look at `Data.Nat.Properties` for what to call this and the following proofs.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+  k = toℕ x;  m = toℕ y;  n = toℕ o;  n≤m = toℕ-mono-≤ o≤y
+
+x+y∸x+z :  ∀ x y z →  (x + y) ∸ (x + z) ≡ y ∸ z
+x+y∸x+z x y z =  begin-equality
+  (x + y) ∸ (x + z)    ≡⟨ sym (∸-+ (x + y) x z) ⟩
+  ((x + y) ∸ x) ∸ z    ≡⟨ cong (_∸ z) ([x+y]∸x≡y x y) ⟩
+  y ∸ z                ∎
+  where open ≤-Reasoning
+
+suc-x∸suc-y :  ∀ x y → suc x ∸ suc y ≡ x ∸ y
```
MatthewDaggitt: Please follow the style guide and use square brackets instead of `-` to indicate application in proof names.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,364 @@
...
+open import Data.Nat.Binary.Base
+open import Data.Nat.Binary.Properties
+import Data.Nat.Properties as ℕₚ
+open import Data.Product using (_×_; _,_; proj₁; proj₂; ∃)
+open import Data.Sum using (inj₁; inj₂)
+open import Data.Vec using ([]; _∷_)
+open import Function using (_∘_; _$_)
+open import Level using (0ℓ)
+open import Relation.Binary using (Tri; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
+open import Relation.Binary.PropositionalEquality as P using
```
MatthewDaggitt: At this point I think we can just remove the `using` syntax.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,56 @@
...
+                                 _∙_ to _∙'_; ∙-cong to ∙'cong; setoid to setoid')
+
+open Definitions A B _≈'_ using (Homomorphic₂)
+open Definitions B A _≈_  using () renaming (Homomorphic₂ to Homomorphic₂')
+open ToSetoid A setoid using (_≈∘_)
+open ToSetoid B setoid' using () renaming (_≈∘_ to _≈'∘_)
+open EqR setoid
+
+id' : B → B
+id' = id
```
MatthewDaggitt: What's the point of this?

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,56 @@
...
+
+IsHomo :  Pred (A → B) _
+IsHomo f =  Homomorphic₂ f _∙_ _∙'_
+
+IsHomo-rev :  Pred (B → A) _
+IsHomo-rev f =  Homomorphic₂' f _∙'_ _∙_
```
MatthewDaggitt: I don't think adding these new names just for this file is worth it...

src/Algebra/Morphism/ToSetoid.agda
```diff
@@ -0,0 +1,49 @@
...
+
+
+open import Function using (_∘_)
+open import Level using (_⊔_)
+open import Relation.Binary using (Rel; _⇒_; Reflexive; Symmetric;
+                                   Transitive; IsEquivalence; Setoid)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl)
+
+
+module Algebra.Morphism.ToSetoid {α β β=} (A : Set α) (S : Setoid β β=) where
```
MatthewDaggitt: I don't see you using any of the contents of this file except for `_≈∘_` anywhere in this PR, and in the context `_≈∘_` is used the existing `_≗_` can be used just fine.

Therefore could you remove this file from the PR and you can always add it back in a later PR if you need it?

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,56 @@
...
+
+---------------------------------------------------------------------------------
+-- If f and g are mutually inverse maps between A and B, g is congruent,
+-- f is a homomorphism, then g is a homomorphism.
+
+IsHomo⇒IsHomo-rev : (f : A → B)              → (g : B → A) →
+                    (g Preserves _≈'_ ⟶ _≈_) → g ∘ f ≈∘ id → f ∘ g ≈'∘ id' →
```
MatthewDaggitt: You should use `IsInverse` from `Function.Structures` to neatly capture these properties. You then no longer need the magmas as module parameters.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,56 @@
...
+open import Algebra using (Magma)
+import Algebra.Morphism.Definitions as Definitions
+import Algebra.Morphism.ToSetoid as ToSetoid
+open import Function using (id; _∘_)
+open import Relation.Binary using (_Preserves_⟶_)
+import Relation.Binary.EqReasoning as EqR
+open import Relation.Unary using (Pred)
+
+
+module Algebra.Morphism.TwoMagmas {α α= β β=} (M : Magma α α=) (M' : Magma β β=)
```
MatthewDaggitt: After ditching the `Magma` module parameters (described how below) I'd then rename this module `Algebra.Morphism.Consequences` which is more standardised.

src/Algebra/Morphism/ToSetoid.agda
mechvel: IsHomo-rev is proved for arbitrary Magmas, in particular, with  _≈∘_.
Instead you suggest to include in the library only its special case for PropositionalEquality.
This looks like a poor design. Because users will program various generalizations for this lemma instead of using standard library. Agda is such a language in which methods tend to be programmed in a general way.

MatthewDaggitt: > Instead you suggest to include in the library only its special case for PropositionalEquality.

Yes you're right, I was getting slightly confused but my point stands for `IsHomo-Rev`. If you use `IsInverse` in `Function.Structures` which works over arbitrary equalities then you no longer need `_≈∘_`.

src/Data/Nat/Properties.agda
mechvel: Done.

src/Data/Nat/Binary/Properties.agda
mechvel: But they need the parameter values  +-commutativeSemigroup, +-magma               
(for `ℕᵇ`) which are implemented not at the top of this file but only several lines higher.                                                             
- ?

CHANGELOG.md
mechvel: Done.

src/Data/Nat/Binary/Minus.agda
mechvel: Done.

src/Data/Nat/Binary/Minus.agda
mechvel: Done.

src/Data/Nat/Binary/Minus.agda
mechvel: Fixed.

src/Data/Nat/Binary/Minus.agda
mechvel: But it uses a parameter value  ∸-magma  implemented above.                        
 (?)

src/Data/Nat/Binary/Minus.agda
mechvel: Done.

src/Data/Nat/Binary/Minus.agda
mechvel: Fixed. It remains only  ``open import Algebra.Definitions {A = ℕᵇ} _≡_``                                    
at the top.

src/Data/Nat/Binary/Minus.agda
mechvel: Fixed. And now, `Op₂` is imported from Algebra.

src/Data/Nat/Binary/Minus.agda
mechvel: Removed everywhere.

src/Data/Nat/Binary/Minus.agda
mechvel: Omitted.

src/Data/Nat/Binary/Minus.agda
mechvel: Changed to ``... = subst (_≢ 0ᵇ) (sym (x∸0≡x x')) 2[1+x]≢0``

src/Data/Nat/Binary/Minus.agda
mechvel: I did this not due to literate Agda, but in order to make the source more readable.                                                                    
All right, I remove these doubles.

src/Data/Nat/Binary/Minus.agda
mechvel: Renamed to ``∸-+-assoc, +-∸-assoc``.

src/Data/Nat/Binary/Minus.agda
mechvel: Renamed to `suc[x]∸suc[y]`.

src/Data/Nat/Binary/Minus.agda
mechvel: Removed.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: Removed.  Instead ``id {A = B}`` is set in-place (because there are `id` on A and on B, and the type checker reports an unsolved thing).

src/Algebra/Morphism/TwoMagmas.agda
mechvel: Removed.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: This concerns several your last comments.
Please, consider once more the above subject of                                   
Morphism.ToSetoid and Morphism.TwoMagmas.                                         
                                                                                  
* It is strange the you are against `_≈∘_`.                                      
  It is a very natural generalization for `_≗_`,   a generally needed notion.                              
  Its context requires a domain of Set and a codomain of Setoid -- only this.     
  It is evident that one day Standard will introduce such a relation on functions.                                                                      
  With this, there also naturally appears the corresponding Setoid instance on    
  `A → B`.  Generally, a useful thing. You are going to remove this instance.       
  Well, at least `_≈∘_` is used in TwoMagmas.                                       
                                                                                  
* To formulate and to prove the lemma `IsHomo⇒IsHomo-rev`,                      
  there are needed                                                                
  a) two Setoids, b) binary operations ``_∙_, _∙'_`` on these setoids,                
  c) the congruence property for `_∙'_`.                          
                                                                                  
All these requirements together mean just _two magmas_                            
(with adding only a single condition of congruence for  `_∙_`).                       
This is why it is introduced as                                                   
```                                                                               
 module Algebra.Morphism.TwoMagmas ... (M : Magma α α=) (M' : Magma β β=)                       
```                                                                                                
As to f and g in the signature, it is required congruence for g.                  
                                                                                  
Please, see  Algebra.Morphism.TwoMagmas.                                          
It is changed to meet somewhat half of your requirements for this subject.
                                                                                  
You suggest to use `IsInverse` in Function.Structures.                            
It consists of  `isLeftInverse` and  `inverseʳ`. 
And this requires the congruence property for  f.                                  
And this property is not required in the lemma.                                   
So, I use instead AreInverseMaps defined in Algebra.Morphism.TwoMagmas.           
                                                                                  
Please, advice.

mechvel:
Please, consider my commit of January 3.  It overwrites the one of January 2.
It is more close to the requirements by @MatthewDaggitt.

Function.Structures.IsInverse  requires congruence for one of the maps f, g.
But it occurs that  there exists ``Function.Definitions.Inverseᵇ``, which does not require such a congruence.
So, I apply now  Inverseᵇ,  and  Algebra.Morphism.ToSetod  is not needed in this particular PR.
This meets the requirement by @MatthewDaggitt.

(On the other hand,  `_≈∘_` is so evidently useful that one day Standard will include
it and its related Setoid instance -- but as some separate PR).


It occurs that  Function.Definitions.Inverseᵇ
is not as in Function.Structures: it does not require congruece.

src/Algebra/Morphism/TwoMagmas.agda
MatthewDaggitt: > It is a very natural generalization for _≗_, a generally needed notion.

Yes, you're right, it is useful. However a) `Algebra.Morphism.ToSetoid` is not the right place to put it and b) it's not needed by this PR and c) this PR is already huge enough as it is. The smaller this PR is the greater the chance of it getting merged in.

> You suggest to use IsInverse in Function.Structures.
It consists of isLeftInverse and inverseʳ.
And this requires the congruence property for f.
And this property is not required in the lemma.

Hmm well I assumed that adding an extra congruency assumption isn't too harmful. For example you've already done so yourself, note that you've assumed congruency of `_∙_` via `Magma` but not used it. I'm happy with your current solution of using `Inverseᵇ` however.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,43 @@
...
+-- f is a homomorphism, then g is a homomorphism.
+
+IsHomo⇒IsHomo-rev : (f : A → B)              → (g : B → A)  →
+                    (g Preserves _≈'_ ⟶ _≈_) → Inverseᵇ f g →
+                    Homomorphic₂ f _∙_ _∙'_  → Homomorphic₂' g _∙'_ _∙_
+IsHomo⇒IsHomo-rev f g gCong (f∘g=id , g∘f=id) fHomo y y' = begin
+  g (y ∙' y')       ≈⟨ gCong (∙'cong y=fx y'=fx') ⟩
+  g (f x ∙' f x')   ≈⟨ gCong (≈'sym (fHomo x x')) ⟩
+  g (f (x ∙ x'))    ≈⟨ g∘f=id (x ∙ x') ⟩
+  x ∙ x'            ≈⟨ ≈refl ⟩
```
MatthewDaggitt: This should be `≡⟨⟩`.

I'd also recommending expanding out everything. I think the reassignments to `x` just confuse things... You can then just use `x` and `y` instead of `y` and `y'`  which are much easier to read.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,43 @@
...
+open EqR setoid
+
+---------------------------------------------------------------------------------
+-- If f and g are mutually inverse maps between A and B, g is congruent,
+-- f is a homomorphism, then g is a homomorphism.
+
+IsHomo⇒IsHomo-rev : (f : A → B)              → (g : B → A)  →
+                    (g Preserves _≈'_ ⟶ _≈_) → Inverseᵇ f g →
+                    Homomorphic₂ f _∙_ _∙'_  → Homomorphic₂' g _∙'_ _∙_
+IsHomo⇒IsHomo-rev f g gCong (f∘g=id , g∘f=id) fHomo y y' = begin
```
MatthewDaggitt: It would be more conventional to call it `cong` or `g-cong` rather than `gCong`.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,368 @@
...
+x≤y⇒x∸y≡0 (inj₁ (even<even x<y)) = x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (inj₁ x<y))
+x≤y⇒x∸y≡0 {2[1+ x ]} {2[1+ x ]} (inj₂ refl) =
+  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 {x} {x} (inj₂ refl))
+x≤y⇒x∸y≡0 {2[1+ x ]} {1+[2 y ]} (inj₁ (even<odd x<y)) with x <? y
+... | yes _  = refl
+... | no x≮y = contradiction x<y x≮y
+x≤y⇒x∸y≡0 {1+[2 x ]} {2[1+ y ]} (inj₁ (odd<even x≤y)) with x ≤? y
+... | yes _  = refl
+... | no x≰y = contradiction x≤y x≰y
+x≤y⇒x∸y≡0 {1+[2 _ ]} {2[1+ _ ]} (inj₂ ())
```
MatthewDaggitt: This case is not needed.

src/Data/Nat/Binary/Minus.agda
```diff
@@ -0,0 +1,368 @@
...
+
+x≤y⇒x∸y≡0 : ∀ {x y} → x ≤ y → x ∸ y ≡ 0ᵇ
+x≤y⇒x∸y≡0 {zero} {_} _           = refl
+x≤y⇒x∸y≡0 (inj₁ (even<even x<y)) = x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 (inj₁ x<y))
+x≤y⇒x∸y≡0 {2[1+ x ]} {2[1+ x ]} (inj₂ refl) =
+  x≡0⇒double[x]≡0 (x≤y⇒x∸y≡0 {x} {x} (inj₂ refl))
+x≤y⇒x∸y≡0 {2[1+ x ]} {1+[2 y ]} (inj₁ (even<odd x<y)) with x <? y
+... | yes _  = refl
+... | no x≮y = contradiction x<y x≮y
+x≤y⇒x∸y≡0 {1+[2 x ]} {2[1+ y ]} (inj₁ (odd<even x≤y)) with x ≤? y
```
MatthewDaggitt: This is just `odd∸even-for>` is it not?

mechvel:
I am going somehow to implement (or/and answer to) your new notes.
But my last commit has been tested under  _master_ of about January 3.
And now I need to do changes to this commit under the current master.
So I have saved several my .agda files from my local  .../agda-stdlib-fork,
and commanded 
```
> cd agda-stdlib-fork
> git checkout master
> git pull
```
Many files have changed there.
Before setting my new changes, I run the test:
```
> make clean
> fix-agda-whitespace
> make test
```
The old local tree was tested successfully.
But this time it reports
```
cabal clean && cabal build && cabal install
Warning: The clean command is a part of the legacy v1 style of cabal usage.

Please switch to using either the new project style and the new-clean command
or the legacy v1-clean alias as new-style projects will become the default in
the next version of cabal-install. Please file a bug if you cannot replicate a
working v1- use case with the new-style commands.

For more information, see: https://wiki.haskell.org/Cabal/NewBuild
Warning: lib.cabal:0:0: version with tags
./lib.cabal has been changed. Re-configuring with most recently used options.
If this fails, please run configure manually.
Warning: lib.cabal:0:0: version with tags
Configuring lib-1.3...
cabal: Encountered missing dependencies:
directory >=1.0.0.0 && <1.4

GNUmakefile:26: recipe for target 'Everything.agda' failed
```
I have run  _cabal update_,  and it reports  
```
Warning: The update command is a part of the legacy v1 style of cabal usage.

Please switch to using either the new project style and the new-update command
or the legacy v1-update alias as new-style projects will become the default in
the next version of cabal-install. Please file a bug if you cannot replicate a
working v1- use case with the new-style commands.

For more information, see: https://wiki.haskell.org/Cabal/NewBuild

Downloading the latest package list from hackage.haskell.org
Warning: the 'wget' transport currently doesn't support range requests, which
wastes network bandwidth. To fix this, set 'http-transport' to 'curl' or
'plain-http' in '~/.cabal/config'. Note that the 'plain-http' transport
doesn't support HTTPS.
To revert to previous state run:
    cabal update --index-state='2019-12-29T10:21:54Z'
```
After this, ``make clean && make test`` still reports the same error.

Please, how to fix?

Thanks.

MatthewDaggitt:
Closing in favour of #1127

PR #1031, opened 2019-12-30
Add property lift-injective to Data.Fin.Properties.

dominiquedevriese:
lift-injective : ∀ {m n} (f : Fin m → Fin n) →
                 (∀ {x y} → f x ≡ f y → x ≡ y) →
                 ∀ k {x y} → lift k f x ≡ lift k f y → x ≡ y

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,13 @@ Other minor additions
   not-injective : not x ≡ not y → x ≡ y
   ```
 
+* Added new properties to `Data.Fin.Properties`:
+  ```agda
+  lift-injective : ∀ {m n} (f : Fin m → Fin n) →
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Looks good. Though we should probably use the new `Injective` definition from `Function.Definitions`?

dominiquedevriese:
That makes sense, of course, but I'm not sure I would do it in this PR.
This file already contains a number of other injectivity proofs (`suc-injective`, `toℕ-injective`, `inject₁-injective` etc.), none of which use `Function.Definitions.Injective` yet.
I would personally keep the addition of `lift-injective` separate from the migration of all injectivity proofs to use `Function.Definitions.Injective`.
But if you prefer, I'm fine with a different approach.

MatthewDaggitt:
Okay, fair enough!

PR #1035, opened 2020-01-02
[ new ] pretty printing of rose trees

gallais:
We should be able to embed any other finite tree into a rose tree
so this is highly reusable.


MatthewDaggitt:
Looks good to me, no comments here :+1:

PR #1036, opened 2020-01-04
Generalised some permutation lemmas from propositional to setoid equality

MatthewDaggitt:
These are some extremely unpleasant lemmas to prove. I'm sure they can probably be optimised further, particularly `drop-mid` but I think it's more important to save other people the pain and get them in as is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -305,13 +305,36 @@ Other minor additions
   AllPairs-resp-Pointwise : R Respects₂ _∼_ → (AllPairs R) Respects (Pointwise _∼_)
   ```
 
-* Added new combinators and functions to `Data.List.Relation.Binary.Permutation.Setoid.PermutationReasoning`:
+* Added new functions to `Data.List.Relation.Binary.Permutation.Setoid`:
+  ```agda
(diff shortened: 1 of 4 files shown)


gallais:
Ouch, that's a big one. Small suggestion: could we not prove a corollary of these
ones with an arbitrary list `mid` instead of a singleton list `[ v ]` by induction
on `mid`?

```agda
  drop-mid-≋  : ws ++ [ x ] ++ ys ≋ xs ++ [ x ] ++ zs → ws ++ ys ↭ xs ++ zs
  drop-mid    : ws ++ [ v ] ++ ys ↭ xs ++ [ v ] ++ zs → ws ++ ys ↭ xs ++ zs
```

This would suggest renaming these to `X-singleton` and use the `X` name for the
more general lemmas.

JacquesCarette:
I'm one of those who went through some of the pain of such lemmas - without success.  `split` was my eventual goal. So I'm glad that's over with now!

MatthewDaggitt:
> Small suggestion: could we not prove a corollary of these
ones with an arbitrary list mid instead of a singleton list [ v ] by induction
on mid?

Yup, I'll add it in.

Issue #619, opened 2019-02-15
Proposal of reexport 


mechvel:
Suppose that one has to use ``_+_, divMod, gcd, +-comm``   -- all for Nat.
One is forced to write
```
open import Data.Nat ...
open import Data.Nat.DivMod ...
open import Data.Nat.GCD ...
open import Data.Nat.Properties ...
```
But normally, of a common reason, it needs to be only
``open import Data.Nat using (_+_; divMod; gcd; +-comm)``.

A similar approach is desirable to other domain constructors: Integer, Bin, List, Maybe, and so on.
Can standard library organize this?
May be, to import  ``Data.Nat.Properties`` to ``Data.Nat`` as public 
(with hiding unneeded items) 
-- ?

gallais:
Is this #153?

mechvel:
> Is this #153?

Not precisely. 
#153 suggests a single module Data that exports "everything of Data" -- which looks strange to me, something like joining all the standard library into one module. 
I suggest 
Data.Nat  to export all the main items for the domain Nat,
Data.Bin  to export all the main items for the domain Bin,
Data.Maybe  to export all the main items related to the domain constructor Maybe,
and so on.

mechvel:
Suppose that you join all  the standard Data export, 
and that you need  ``+-comm``  for Integer.
Will  ``Data.+-comm``   be for ``Integer._+_``,  for ``Nat._+_``, or for ``Bin._+_``  ?
In my suggestion, it will be written as 
``Data.Integer.+-comm``,  or as  
``open import Data.Integer using (+-comm)``.

HuStmpHrrr:
instead of re-exporting `Properties`, I'd prefer the following

```agda
import Data.Nat.Properties
module ℕₚ = Data.Nat.Properties
```

this keeps the name space clean while provide access to the properties in a desired way while not introducing any name collision.

JacquesCarette:
I think what @mechvel is after is what Bill Farmer and I called [high level theories](http://www.cas.mcmaster.ca/~carette/publications/hlt.pdf). Consider "Group Theory": this denotes two things, 1) the 'presentation of a Group', given by the record in the module `Algebra`, and 2) a large set of tools and definitions, as defined in most books called "Group Theory".

[...]

PR #1038, opened 2020-01-04
Add a metric space hierarchy

MatthewDaggitt:
I've been meaning to add this ever since v0.16. There are a couple of quirks to this one:

1. In full generality metric spaces are quantified over two types: the domain and codomain of the distance function. However in mathematics the codomain usually goes unremarked upon for the most part and is assumed to be e.g. the naturals or the reals. To ensure this flexibility, I've provided both a standard `Function.Metric` hierarchy and a `Function.Metric.Nat` hierarchy which re-exports the definitions specialised to natural numbers. It should be easy to add future sub-hierarchies for the rationals/reals or other more exotic codomains.

2. In full generality it's impossible to distinguish between the stronger types of metrics (e.g. metric/ultrametric/(p-)inframetric) without knowing what the operator is. Therefore the upper part of the hierarchy in `Function.Metric.Structures` simply consists of the record `IsGeneralMetric` which takes in an arbitrary operator. Only when specialised to naturals in `Function.Metric.Nat.Structures` are these subtypes actually distinguished between. This requires a little bit of non-standard module trickery to ensure that `open IsMetric isMetric` actually works as expected.

Finally, the names for some of properties of metrics don't appear to be massively standardised in the literature. I've attempted to do the best I can but I welcome any feedback on the names.


src/Function/Metric/Bundles.agda
```diff
@@ -0,0 +1,139 @@
...
+    Carrier         : Set a
+    Image           : Set i
+    _≈_             : Rel Carrier ℓ₁
+    _≈ᵢ_            : Rel Image ℓ₂
+    _≤_             : Rel Image ℓ₃
+    0#              : Image
+    _∙_             : Op₂ Image
+    d               : DistanceFunction Carrier Image
+    isGeneralMetric : IsGeneralMetric _≈_ _≈ᵢ_ _≤_ 0# _∙_ d
+
```
gallais: Should `_∙_` and `0#` form a monoid?
￼

src/Function/Metric/Structures.agda
```diff
@@ -0,0 +1,85 @@
...
+
+------------------------------------------------------------------------
+-- Quasi-semi-metrics
+
+record IsQuasiSemiMetric (d : DistanceFunction A I)
+                       : Set (a ⊔ i ⊔ ℓ₁ ⊔ ℓ₂ ⊔ ℓ₃) where
+  field
+    isPreMetric      : IsPreMetric d
+    0⇒eq             : ∀ {x y} → d x y ≈ᵢ 0# → x ≈ₐ y
+
```
gallais: Shouldn't you use `Indiscernable` here?

src/Function/Metric/Bundles.agda
MatthewDaggitt: I don't think so...
- the `_∙_` operators for p-relaxed metrics and p-inframetrics are not associative. 
- 0 is not an identity element for ultrametrics over the reals, only over the positive reals.

CHANGELOG.md
```diff
@@ -132,6 +132,25 @@ Automated warnings are attached to all deprecated names to discourage their use.
 Other major additions
 ---------------------
 
+* Added a hierarchy for metric spaces:
+  ```
+  Function.Metric
+  Function.Metric.Core
+  Function.Metric.Definitions
```
MatthewDaggitt: Also need to fix `Bundles` here and below

JacquesCarette:
Note that I was at first very surprised that the "metric space" hierarchy was under `Function/`. Once I read the code, it became clearer: mathematicians tend to talk about 'a space with a metric', but the important part are the metrics as functions. So, post facto, this makes sense.  However, I'm pretty sure a lot of others will also be confused.  Thus I suggest you talk about "metric functions" when introducing this work.

src/Data/Nat/Properties.agda
```diff
@@ -1747,6 +1751,80 @@ m≤n+∣m-n∣ m n = subst (m ≤_) (cong (n +_) (∣-∣-comm n m)) (m≤n+∣
...
+∣-∣-triangle x       y       zero    = begin
+  ∣ x - 0 ∣             ≡⟨ ∣-∣-identityʳ x ⟩
+  x                     ≤⟨ m≤∣m-n∣+n x y ⟩
+  ∣ x - y ∣ + y         ≡⟨ cong₂ _+_ refl (sym (∣-∣-identityʳ y)) ⟩
+  ∣ x - y ∣ + ∣ y - 0 ∣ ∎
+  where open ≤-Reasoning
+∣-∣-triangle (suc x) (suc y) (suc z) = ∣-∣-triangle x y z
+
+------------------------------------------------------------------------
+-- Metric structures
```
JacquesCarette: I think this is a bad place to put these properties - Nats have so many properties, if they all end up in one, eventually huge, file, things will devolve.  Perhaps `Data.Nat.Metric.Properties`?

src/Function/Metric/Bundles.agda
```diff
@@ -0,0 +1,139 @@
...
+    _≈ᵢ_              : Rel Image ℓ₂
+    _≤_               : Rel Image ℓ₃
+    0#                : Image
+    d                 : DistanceFunction Carrier Image
+    isQuasiSemiMetric : IsQuasiSemiMetric _≈_ _≈ᵢ_ _≤_ 0# d
+
+  open IsQuasiSemiMetric isQuasiSemiMetric public
+
+  preMetric : PreMetric a i ℓ₁ ℓ₂ ℓ₃
+  preMetric = record
```
JacquesCarette: would it be ok to lay all these out on a single line instead of 3? I completely get that once you get 2 (or more) fields, it does make sense to use more vertical space -- but just 1?

src/Function/Metric/Bundles.agda
JacquesCarette: I had never heard of p-relaxed metrics and p-inframetrics until now.  Might it make sense to add a link to some reference?

src/Function/Metric/Definitions.agda
```diff
@@ -0,0 +1,60 @@
...
+-- Properties
+
+Congruent : Rel A ℓ₁ → Rel I ℓ₂ → DistanceFunction A I → Set _
+Congruent _≈ₐ_ _≈ᵢ_ d = d Preserves₂ _≈ₐ_ ⟶ _≈ₐ_ ⟶ _≈ᵢ_
+
+Indiscernable : Rel A ℓ₁ → Rel I ℓ₂ → DistanceFunction A I → I → Set _
+Indiscernable _≈ₐ_ _≈ᵢ_ d 0# = ∀ {x y} → d x y ≈ᵢ 0# → x ≈ₐ y
+
+PositiveDefinite : Rel A ℓ₁ → Rel I ℓ₂ → DistanceFunction A I → I → Set _
+PositiveDefinite _≈ₐ_ _≈ᵢ_ d 0# = ∀ {x y} → x ≈ₐ y → d x y ≈ᵢ 0#
```
JacquesCarette: Are you sure that's enough for `PositiveDefinite`? Don't you also need that if they are not equal then the distance is not 0 too?

src/Function/Metric/Definitions.agda
```diff
@@ -0,0 +1,60 @@
...
+TriangleInequality _≤_ _∙_ d = ∀ x y z → d x z ≤ (d x y ∙ d y z)
+
+Bounded : Rel I ℓ → DistanceFunction A I → Set _
+Bounded _≤_ d = ∃ λ n → ∀ x y → d x y ≤ n
+
+TranslationInvariant : Rel I ℓ₂ → Op₂ A → DistanceFunction A I → Set _
+TranslationInvariant _≈_ _∙_ d = ∀ {x y a} → d (x ∙ a) (y ∙ a) ≈ d x y
+
+Contracting : Rel I ℓ → (A → A) → DistanceFunction A I → Set _
+Contracting _≤_ f d = ∀ x y → d (f x) (f y) ≤ d x y
```
JacquesCarette: I somehow feel that `Contracting` on just a `Rel` rather than on a linear order feels wrong. The definition above is more like `Homomorphism` than `Contracting`.  I probably could have made the same comment about `Bounded`.

src/Function/Metric/Definitions.agda
MatthewDaggitt: Yup, nice spot! Renamed this one to `Definite` and added `NonNegative` as well which is then used in `IsProtoMetric`.

src/Function/Metric/Bundles.agda
MatthewDaggitt: I've got into the habit of putting them on 3 because in `Relation.Binary`/`Algebra` hierarchies the names are very long and they don't always fit on one line under the very loosely obeyed 80 characters per line limit. I guess it doesn't matter here?

src/Data/Nat/Properties.agda
MatthewDaggitt: Hmm so I agree that ideally we'd have a `Data.Nat.Metric` subdirectory, just like we have a `Data.List.Relation`. However I'd want it to be done consistently, e.g. move all the properties of `|_-_|` out of `Data.Nat.Properties` at the same time. It may be difficult to do backwards compatibly(?) so I think it's probably best done when we get around to #619.

src/Function/Metric/Definitions.agda
MatthewDaggitt: I completely understand where you're coming from, however I think this is fairly standard practice in the library e.g. look at `Maximum` in `Relation.Binary.Definitions` in which we don't require the relation to be a partial order, or `Antisymmetric` in which we don't require the first relation to be a setoid.

The main reason for this is to avoid passing extraneous things into definitions. None of the proofs for a total order are used in the definition and neither is the equality relation we'd be forced to provide. This means that a) the definition can be used in environments where those assumptions are not yet available (even though they will be later on) and b) it provides better modularity of code. For example imagine we wanted a new structure `IsContractingMetric` which extended `IsMetric`. If we made the `contracting` definition require a total order, we'd have to first build that total order between the field `isMetric : IsMetric ...` and `contracting : Contracting ...` which would be inconvenient and ugly.

JacquesCarette: I think you're right - so I think this means we ought to document that at the top level of the library somewhere.  Something along the lines of "Type abbreviations (such as `Maximum` and `Contracting`) are meant to capture the basic idea of certain definitions -- but are not meant to express the full concept on their own. The complete concepts are captured in various bundles where the abbreviation is correctly used in context. In other words, the bundles are meant to be the place where mathematical abstractions are adequately captured, while type abbreviations are meant to help modular reuse of the concepts that make up those abstractions."  Or something like that (my words above are probably a little too woolly).

MatthewDaggitt:
Just realised that we can't reuse the bundles for `Nat` as we can't constrain that the image to `ℕ` :cry: 

One possible solution is to parameterise all the bundles by some new `LowerBoundedTotalOrder` bundle similar to how #897 is parameterising modules over semirings. However I'm not too keen on this as it would mean that you could no longer define metrics over `ℚ`. Instead you'd have to define the new `ℚ+` type.

In the end I've just gone for redefining the 4 simpler bundles.

JacquesCarette:
Regarding the bundles for `Nat`: I think what you really want here [aka a new feature] is a declaration-level operation that lets you derive new declarations by instantiation of previous ones.  This is a really nice use case for @alhassy 's work.

src/Function/Metric/Structures.agda
```diff
@@ -0,0 +1,93 @@
...
+open import Function.Metric.Definitions
+open import Level using (_⊔_)
+
+------------------------------------------------------------------------
+-- Proto-metrics
+
+record IsProtoMetric (d : DistanceFunction A I)
+                   : Set (a ⊔ i ⊔ ℓ₁ ⊔ ℓ₂ ⊔ ℓ₃) where
+  field
+    isTotalOrder     : IsTotalOrder _≈ᵢ_ _≤_
```
laMudri: I don't know the aim of this development, but surely this property is too strong if we want the image to be real numbers.

MatthewDaggitt: Hmm that's an interesting point... To be honest I've only used it over naturals. I've just checked which assumptions I use in my development, and it turns out I only require it to be a preorder. I  I guess it probably would make sense to weaken it therefore. Anybody have any strong opinions on whether we should drop antisymmetry as well?

MatthewDaggitt:
Messed up the rebase so I landed this manually in 20e9d7a

PR #623, opened 2019-02-16
[ re #622 ] Better reasoning inference

oisdk:
As described in #622

PR #1045, opened 2020-01-09
Change argument order for binary reasoning combinators

MatthewDaggitt:
After further reflection in #922, it seems like inverting the argument order for the reasoning combinators is probably the lesser of two evils in order to get the new solvers working. This is therefore an updated version of @oisdk's original PR (#623) with the changes extended to the new double/triple reasoning.

Whilst doing this I did find one very minor backwards incompatibility in the actual use of the reasoning combinators: an instance whilst using `≡-Reasoning` in `Data.Rational.Properties` where the change in order of arguments seems to mean that one of the coprimality proofs can no longer be inferred. This was easily fixed by giving the implicit argument explicitly. Given the rarity of this vs the potential gains in performance and the use of solvers, this seems like a reasonable trade-off.

Any thoughts?

(Note: I still need to add a changelog entry).


CHANGELOG.md
```diff
@@ -21,6 +21,67 @@ Bug-fixes
...
+  as `_≡⟨_⟩_` etc. are now syntax instead of names. For example instead of:
+  ```agda
+  open SetoidReasoning hiding (_≈⟨_⟩_) renaming (_≡⟨_⟩_ to _↭⟨_⟩_)
+  ```
+  one must now write :
+  ```agda
+  open SetoidReasoning hiding (step-≈; step-≡)
+
+  infixr 2 step-↭
+  step-↭ = step-≡
```
asr: Note that if we hide `step-≡` we cannot write `step-↭ = step-≡`.

CHANGELOG.md
```diff
@@ -21,6 +21,67 @@ Bug-fixes
...
+  syntax step-↭ x y≡z x≡y = x ≡⟨ x≡y ⟩ y≡z
```
asr: Typo: `≡⟨` should be `↭⟨`.

MatthewDaggitt:
Thanks @asr

Issue #1033, opened 2020-01-02
Drop the dependency on `Data.Nat.Properties` from `Data.Fin.Base`? 


gallais:
If you want loading to get snappy, it's usually a good idea to use `.Base` rather
than the toplevel module because they are generally fairly self-contained.

For `Data.Fin.Base` however we need to typecheck the whole of `Data.Nat.Properties`
because of two functions:

```agda
fromℕ< : ∀ {m n} → m ℕ.< n → Fin n
inject≤ : ∀ {m n} → Fin m → m ℕ.≤ n → Fin n
```

This, in turns, means that `Data.List.Base` also depends on `Data.Nat.Properties`
and the same goes for everything depending on `Data.List.Base`.

I suggest that we move these two definitions to `Data.Fin`. The main hurdle being:
what should we do with the proofs involving these definitions? They are currently
in `Data.Fin.Properties` and cannot stay there as this would create a cycle
between `Data.Fin` and `Data.Fin.Properties`.

JacquesCarette:
My first instinct is to move `≤-pred` out of `Data.Nat.Properties` and into `Data.Nat.Base`. That's where `≤` is defined, and `≤-pred` is arguably as much a computation as a proof transformer.  And that's the only function those two above require.

An alternative would be to create a `Data.Nat.Properties.Base` with just that one function in it.

MatthewDaggitt:
Hmm so this was caused by #717. Prior to that we didn't need `Data.Nat.Properties` at all... For such a small change, that PR has had a lot of unintended side effects!

> I suggest that we move these two definitions to Data.Fin. The main hurdle being:
what should we do with the proofs involving these definitions?

I agree we don't have any pattern for dealing with properties of functions in the non-base module. Indeed I'd like to actively discourage us from adding things to them, as in v2.0 I'm hoping to start exporting far more stuff from the top-level modules, including other operations and properties. See #619.

> My first instinct is to move ≤-pred out of Data.Nat.Properties and into Data.Nat.Base. That's where ≤ is defined, and ≤-pred is arguably as much a computation as a proof transformer.

I'm also not terribly keen on this. Once you make an argument that `≤-pred` is a computation then in my mind we're on a slippery slope and it's very hard to justify not moving other stuff such as `suc-injective` from the `Properties` files.

> An alternative would be to create a Data.Nat.Properties.Base with just that one function in it.

[...]

PR #1048, opened 2020-01-09
[ new ] pretty printing of binary trees

gallais:
Question: should we move `toRose` out of `Data.Tree.Binary` because it requires
the addition of the `--sized-types` option? And if so: where?


MatthewDaggitt:
> Question: should we move toRose out of Data.Tree.Binary because it requires
the addition of the --sized-types option? And if so: where?

Hmm I agree the pollution of all the modules with `--sized-types` is unpleasant and could potentially get a lot worse.
- One option would be to rename it `fromBinary` and move it to `Data.Tree.Rose`. The downside is that rose trees are the most general type of tree so it seems a bit backwards to have it depend on the other types.
- Alternatively we could move it to a new module `Data.Tree.Binary.Conversion`? See #1033 in which I propose `Data.Fin.Conversion`...

MatthewDaggitt:
Also another more general question that has just occurred to me is that we have `Data.(Bool/Nat/Integer/Rational).Show` modules which feel like they have significant overlap with the new `Text` hierarchy. How should we resolve these?

To be honest I feel like we should stick with the `Show` modules as in general the pattern we're moving towards has been to keep everything local to the data-type within it's own folder, e.g. we would rename `Text.Tree.Binary.Linear` to `Data.Tree.Binary.Show` or `Data.Tree.Binary.Show.Linear`?

MatthewDaggitt:
Brought this back up-to-date and moved `toRose` to `fromBinary` to get rid of the sized type issues and moved `Text.Tree.Linear` to `Data.Tree.X.Show`. @gallais any thoughts?

gallais:
Looks good to me! Sorry I completely forgot about this PR.

Issue #1055, opened 2020-01-14
inclusion of Function.BiInverse


pnlph:
Hi,

i would like to add the definition of equivalence from [Jesper Cockx's thesis](https://limo.libis.be/primo-explore/fulldisplay?docid=LIRIAS1656778&context=L&vid=Lirias&search_scope=Lirias&tab=default_tab&lang=en_US). There is already a module named `Function.Equivalence` so we could call this module `Function.BiInverse`

What do you think? Any suggestions?

jespercockx:
I want to mention that I suggested the name BiInverse because that is what the concept is called in the HoTT book (https://hott.github.io/book/nightly/hott-online-1239-g9a4dadb.pdf#page=148)

MatthewDaggitt:
Any chance you could give the definition and/or a page reference?

(Probably) no objections to adding it, however we're trying to deprecate the `Function.Equivalence/Injection/Surjection/...` modules in favour of the new `Function.Structures` and `Function.Bundles` modules so it should probably live in there.

pnlph:
Yes, you can take a look at [BiInverse.agda](https://gist.github.com/pnlph/5908275291bc1d62b346962c19b65ffa)

MatthewDaggitt:
Yup, no objections to adding to `IsBiInverse` to `Function.Structures` and `BiInverse` to `Function.Bundles`. 

Unsure what sort of arrow symbol we'd use to represent the propositional bundle. Personally I'd be favour of `_↩↪_` which is the composition of the symbols we use for the left and right inverses

PR #1058, opened 2020-01-15
inclusion of Function.Structures.BiInverse

pnlph:
New records BiEquivalence and BiInverse in Function.Bundles
New record IsBiInverse in Function.Structures
#1055

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,6 +632,13 @@ Other minor additions
   unwords : List String → String
   _<+>_ : String → String → String -- space-introducing append
   ```
+* Added new records IsBiInverse to `Function.Structures` and BiInverse
+  to `Function.Bundles`:
+  ```agda
+  record IsBiInverse
+    (f : A → B) (g₁ : B → A) (g₂ : B → A) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
+  record BiInverse : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
+  ```
 
 Version 2.6.1 changes
 =====================
@@ -675,4 +682,4 @@ Version 2.6.1 changes
 * Added new definitions in `Relation.Binary.Core`:
   ```agda
   DecidableEquality A = Decidable {A = A} _≡_
-  ```
+  ```  ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -682,4 +682,4 @@ Version 2.6.1 changes
 * Added new definitions in `Relation.Binary.Core`:
   ```agda
   DecidableEquality A = Decidable {A = A} _≡_
-  ```  ```
+  ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -640,6 +640,14 @@ Other minor additions
   record BiInverse : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
   ```
 
+* Added new BiInverse bundle `_↩↪_` and a new constructor `mk↩↪`
+  to `Funtion.Bundles`:
+  ```agda
+  _↩↪_ : Set a → Set b → Set _
+  mk↩↪ : ∀ {f : A → B} {g₁ : B → A} {g₂ : B → A}
+    → Inverseˡ f g₁ → Inverseʳ f g₂ → A ↩↪ B
+  ```
+
 Version 2.6.1 changes
 =====================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -632,11 +632,17 @@ Other minor additions
   unwords : List String → String
   _<+>_ : String → String → String -- space-introducing append
   ```
-* Added new records IsBiInverse to `Function.Structures` and BiInverse
-  to `Function.Bundles`:
+* Added new records `IsBiEquivalence` and `IsBiInverse`
+  to `Function.Structures`:
   ```agda
+  record IsBiEquivalence
+    (f : A → B) (g₁ : B → A) (g₂ : B → A) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
   record IsBiInverse
     (f : A → B) (g₁ : B → A) (g₂ : B → A) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
+  ```
+
+* Added new record BiInverse to `Function.Bundles`:
+  ```agda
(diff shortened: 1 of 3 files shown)


src/Function/Bundles.agda
```diff
@@ -189,6 +189,51 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
...
 
 
+  record BiEquivalence : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      f     : A → B
+      g₁    : B → A
+      g₂    : B → A
+      cong₁ : f Preserves _≈₁_ ⟶ _≈₂_
+      cong₂ : g₁ Preserves _≈₂_ ⟶ _≈₁_
+      cong₃ : g₂ Preserves _≈₂_ ⟶ _≈₁_
```
jespercockx: Maybe there could be some more descriptive names for these fields, e.g. `cong-f`, `cong-g₁`, and `cong-g₂`?

src/Function/Bundles.agda
```diff
@@ -189,6 +189,51 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
...
+
+
+  record BiInverse : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      f         : A → B
+      g₁        : B → A
+      g₂        : B → A
+      cong₁     : f Preserves _≈₁_ ⟶ _≈₂_
+      cong₂     : g₁ Preserves _≈₂_ ⟶ _≈₁_
+      cong₃     : g₂ Preserves _≈₂_ ⟶ _≈₁_
```
jespercockx: Same here.

src/Function/Bundles.agda
pnlph: I agree, that notation is more descriptive and clear, maybe @MatthewDaggitt can say something about the impact of this change. In order to preserve consistency, `cong₁` and `cong₂` should be also elsewhere in the library replaced by `cong-f` and `cong-g` respectively.

gallais: I believe we usually name things `f-cong` and use `Congruent`.
Cf. [`Algebra.Structures`](https://agda.github.io/agda-stdlib/Algebra.Structures.html)

src/Function/Bundles.agda
```diff
@@ -189,6 +189,51 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
...
+      inverseˡ  : Inverseˡ f g₁
+      inverseʳ  : Inverseʳ f g₂
+
+    isCongruent : IsCongruent f
```
MatthewDaggitt: I think this needs to be `f-isCongruent` as it is also possible to show `g₁-isCongruent` and `g₂-isCongruent`

src/Function/Structures.agda
```diff
@@ -100,6 +100,19 @@ record IsRightInverse (f : A → B) (g : B → A) : Set (a ⊔ b ⊔ ℓ₁ ⊔
     renaming (cong to cong₁)
 
 
+record IsBiInverse
+  (f : A → B) (g₁ : B → A) (g₂ : B → A) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂) where
+  field
+    isCongruent : IsCongruent f
```
MatthewDaggitt: Same here

pnlph:
Thanks, minor changes sent.

Just an observation: when editing the changelog, i think i introduced a newline at the end of the file. I see that Atom does this by default, so I found [why should text files end with a newline](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

The reason seems to be that this is how [the POSIX standard defines a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).

CHANGELOG.md
```diff
@@ -675,4 +682,4 @@ Version 2.6.1 changes
 * Added new definitions in `Relation.Binary.Core`:
   ```agda
   DecidableEquality A = Decidable {A = A} _≡_
-  ```
\ No newline at end of file
+  ```  ```
```
gallais: duplicated backtick guards?

src/Function/Bundles.agda
```diff
@@ -189,6 +189,51 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
       }
 
 
+  record BiEquivalence : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂) where
```
MatthewDaggitt: Hmm is this definition needed or not? If it is then we should make sure we add a `IsBiEquivalence` record as well and make sure it's added to the CHANGELOG.

pnlph: OK, but `IsEquivalence` is is not a record itself, but defined as a field in the `Setoid` record.

Should we better add a `BiSetoid` record to `Relation/Binary/Bundles.agda` that includes a `IsBiEquivalence` field and the arrow symbol `_↩↪_` that you proposed in [574466050](https://github.com/agda/agda-stdlib/issues/1055#issuecomment-574466050)?

MatthewDaggitt: > OK, but IsEquivalence is not a record itself, but defined as a field in the Setoid record.

I'm not quite sure what you mean by this. `IsEquivalence` is a record:
https://github.com/agda/agda-stdlib/blob/15dc42e1eeb0dbf946961213ffd86b5fa60ccb00/src/Relation/Binary/Structures.agda#L47

> Should we better add a BiSetoid record to Relation/Binary/Bundles.agda that includes a IsBiEquivalence field and the arrow symbol _↩↪_ that you proposed in 574466050?

No I don't think so. A `BiEquivalence` is still talking about functions between two types and therefore belongs in the function hierarchy. The `_↩↪_` symbol I was referring to, should be used for the instantiation of `BiEquivalence` with propositional equality. See the examples at the bottom of `Function.Bundles`.

pnlph: > I'm not quite sure what you mean by this. `IsEquivalence` is a record:

Sorry, when retrieving the `agda-mode:go-to-definition` command, I mixed `isEquivalence` with `IsEquivalence`. Of course, you are right, `IsEquivalence` is a record in `Relation/Binary/Structures.agda`

Thank you for the hint, I added a new BiInverse bundle `_↩↪_` and a new constructor `mk↩↪`  following the notation of the examples at the end of `Funtion.Bundles`

Issue #1040, opened 2020-01-05
Update instructions for `fix-agda-whitespace`


MatthewDaggitt:
In:
- HACKING.md
- GNUmakefile

gallais:
[In this PR](https://github.com/agda/agda-stdlib/pull/1048) I have updated the makefile and it works for me.
In the `HACKING.md` we basically need to tell people to install https://github.com/agda/fix-whitespace

MatthewDaggitt:
Hmm so I've installed `fix-whitespace` from the repository and am trying to run it but it's not picking up any of the whitespace problems. All of the below seem to exit without complaint even when there is a whitespace problem:
```
fix-whitespace
fix-whitespace --check
cabal exec -- fix-whitespace --check
```
Anyone have any idea what I'm doing wrong? Pinging @L-TChen in particular!

L-TChen:
Do you have a configuration `fix-whitespace.yaml` in the working directory?

gallais:
In the PR mentioned above I have put a yaml file in the `travis/` directory.
It should maybe be moved to the main directory (or `fix-whitespace` should be
made to accept a command line argument pointing at the location of such a file).

MatthewDaggitt:
> It should maybe be moved to the main directory

I'd vote for this solution. We don't want to make contributing to the library any more complicated than necessary...

L-TChen:
I will add an option for specifying a location in the future anyway.

PR #1059, opened 2020-01-16
fix whitespace problems before starting new work.

JacquesCarette:
There were whitespace issues on master.


MatthewDaggitt:
So this is actually a bit complicated. The whitespace problems in `README.Text.Tabular` are actually there by design as that is actually the string that is generated.

To fix this we no longer use `fix-agda-whitespace` but instead use the new `fix-whitespace` program in this [repository](https://github.com/agda/fix-whitespace). See issue #1040 for a discussion.

Equally when we switch to `fix-whitespace` we'll no longer check the CHANGELOG for violations.

MatthewDaggitt:
I should add that to get around this in the mean time, I've simply been running `fix-agda-whitespace` and then commenting out `README.Text.Tabular` and not commiting the changes.

gallais:
> Equally when we switch to fix-whitespace we'll no longer check the CHANGELOG for violations.

If we want to, we can keep doing so. It's a matter of editing the `.yaml` file to
include that file too.

JacquesCarette:
The reason I submitted this is that I had downloaded a completely fresh copy of master and did `make test` and it failed on `fix-agda-whitespace`.

So it seems that the infrastructure in the `Makefile` (and probably the `.yaml` file) have not caught up to the current state of `master`?  Or has `master` jumped the gun?

MatthewDaggitt:
Closed, see #1060

PR #1062, opened 2020-01-20
added definitions for abelian group on unnormalised rationals

riz0id:
i win again


src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -374,6 +380,15 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
   ; isGroup = +-0-isGroup
   }
 
++-abelianGroup : AbelianGroup 0ℓ 0ℓ
++-abelianGroup = record
+  { Carrier        = ℚᵘ
+  ; _≈_            = _≃_
+  ; _∙_            = _+_
+  ; ε              = 0ℚᵘ
+  ; _⁻¹            = -_
```
gallais: You do not need to list these: they will be inferred because of
the value you pass to the `isAbelianGroup` field.

MatthewDaggitt: Likewise you can also remove them in the `+-0-group` definition

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -374,6 +380,15 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
   ; isGroup = +-0-isGroup
   }
 
++-abelianGroup : AbelianGroup 0ℓ 0ℓ
```
MatthewDaggitt: This should be `+-0-abelianGroup`

riz0id:
can i go ahead and start work on morphisms and then inject rational into all of these?

MatthewDaggitt:
> can i go ahead and start work on morphisms and then inject rational into all of these?

Yup, please do! I'll leave this one up for a few more days in case anyone has any comments, then merge it in.

PR #1067, opened 2020-01-23
Tidied up types in morphism definitions

### LINKED ISSUE #1025: a strange Algebra.Morphism.Definitions
Algebra.Morphism.Definitions  puts ``Morphism = A → B``,
while there is a more classical word  `Map`.
Further, it puts  ``Homomorphic₁ : Morphism → Fun₁ A → Op₁ B → Set _``

`Fun₁` and `Op₁` are defined in different modules, and their definitions differ only in renaming variables. Am I missing something? What does this all mean?

For example, the definition 
```
  Homomorphic₂ : Morphism → Fun₂ A → Op₂ B → Set _      
  Homomorphic₂ ⟦_⟧ _∙_ _∘_ = ∀ x y → ⟦ x ∙ y ⟧ ≈ (⟦ x ⟧ ∘ ⟦ y ⟧)
```
will be more clear to the reader in the form of 
```
  Homomorphic₂ :  (A → B) → Op₂ A → Op₂ B → Set _      
  Homomorphic₂ ⟦_⟧ _∙_ _∘_ = ∀ x y → ⟦ x ∙ y ⟧ ≈ (⟦ x ⟧ ∘ ⟦ y ⟧)
```
-- ?

MatthewDaggitt:
Fixes #1025

### DIFF
--- a/src/Algebra/Morphism/Definitions.agda
+++ b/src/Algebra/Morphism/Definitions.agda
@@ -23,11 +23,11 @@ open import Function.Core
 Morphism : Set _
 Morphism = A → B
 
-Homomorphic₀ : Morphism → A → B → Set _
+Homomorphic₀ : (A → B) → A → B → Set _
 Homomorphic₀ ⟦_⟧ ∙ ∘ = ⟦ ∙ ⟧ ≈ ∘
 
-Homomorphic₁ : Morphism → Fun₁ A → Op₁ B → Set _
+Homomorphic₁ : (A → B) → Op₁ A → Op₁ B → Set _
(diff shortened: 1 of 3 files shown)


JacquesCarette:
Excellent.

src/Algebra/Morphism/Definitions.agda
```diff
@@ -23,11 +23,11 @@ open import Function.Core
 Morphism : Set _
 Morphism = A → B
```
gallais: Should these `Morphism`s definition be deprecated if we are moving
away from them?

JacquesCarette: That is certainly my opinion.  Making A and B invisible in a signature does not make it clearer.  And `Morphism` is longer than what it stands for!

PR #1077, opened 2020-01-28
Cleanup imports

gallais:
Tired of trying to break down cycles in the dependency graph
when I try to add some things so here we go: massive `sed`-based
effort to use the `Data.X.Base` modules as much as possible.


JacquesCarette:
I would, in general, try to exempt `README/` from such changes.  They shouldn't cause any loops, and it would be good to check that they working without having to dig down to `.Base`.

src/Data/Nat.agda
```diff
@@ -27,6 +29,10 @@ open import Data.Nat.Properties public
   ; _≤″?_; _<″?_; _≥″?_; _>″?_
   )
 
+infix 4 _≢0
+_≢0 : ℕ → Set
```
JacquesCarette: Does adding this function really belong in this PR ?  Or did it slip in by accident?

gallais: I moved it from `Data.Rational.Unnormalised` because I spotted the
code duplication with `Data.Rational.Base` during the refactoring.
Removed from now.

MatthewDaggitt:
I would second @JacquesCarette suggestion of exempting the README files.

gallais:
Turns out it's really easy to do: `git reset HEAD~10 README/` :)

PR #1079, opened 2020-01-29
Proved a property expanding on the relation between inject+, raise and splitAt

MirceaS:
Proved a property which can be seen as an inversion of the 2 properties already proven above it, that relates `inject+`, `raise` and `splitAt` for Finite Integers.
I also added another simple identity to Sum.Properties that I needed to complete the proof and that I feel belongs there as well.
In addition to those I changed `[_,_]′` to `[_,_]` in `[,]-map-commute` (previously added by me as part of PR #993) because it looks nicer without changing the behaviour of the function. (`[_,_]′` is just a more specialised version of `[_,_]` but Agda can figure out that on it's own in the case of `[,]-map-commute`).


src/Data/Fin/Properties.agda
```diff
@@ -426,6 +427,16 @@ splitAt-raise : ∀ m n i → splitAt m (raise {n} m i) ≡ inj₂ i
 splitAt-raise zero n i = refl
 splitAt-raise (suc m) n i rewrite splitAt-raise m n i = refl
 
+inject+-raise-splitAt : ∀ m n i → [ inject+ n , raise {n} m ] (splitAt m i) ≡ i
+inject+-raise-splitAt zero n i = refl
+inject+-raise-splitAt (suc m) n zero = refl
+inject+-raise-splitAt (suc m) n (suc i) = begin
```
MatthewDaggitt: Really minor I know, but any chance we could align these arguments?

MatthewDaggitt:
Oh and a changelog entry please :heart: . This slipped my mind the first time I looked at it, apologies!

MirceaS:
> Oh and a changelog entry please ❤️ . This slipped my mind the first time I looked at it, apologies!

I was writing it while you commented that :)

MirceaS:
Should be ready for merge now, right?

PR #1081, opened 2020-02-02
Add `deduplicate`, `derun` functions and `Grouped` property for lists

uzytkownik:


src/Data/List/Base.agda
```diff
@@ -292,6 +293,16 @@ span P? (x ∷ xs) with does (P? x)
 break : ∀ {P : Pred A p} → Decidable P → List A → (List A × List A)
 break P? = span (∁? P?)
 
+nub-filter : ∀ {R : Rel A p} → B.Decidable R → A → List A → List A
+nub-filter R? y [] = []
+nub-filter R? y (x ∷ xs) with does (R? y x)
+... | true  = nub-filter R? y xs
+... | false = x ∷ nub-filter R? x xs
+
```
gallais: I think we should use `filter (¬? ∘ R? y)` rather than implementing the ad-hoc `nub-filter`.
This should also make the proofs simpler: we can rely on what we already know about `filter`!

uzytkownik: @gallais Those are two different functions:

- `nub [1, 1, 2, 1]` is `[1, 2, 1]`
- `dedup [1, 1, 2, 1]` is `[1, 2]`

I happen to need `nub` for further work (also - I know performance is not agda-stdlib goal - `nub` is O(n) while `dedup` is O(n²).

It's true that:

```
nub-filter y xs with span R? xs
nub-filter y xs | _ , [] = []
nub-filter y xs | _ , (z ∷ zs) = nub-filter z zs
```

But I think that it will be hard to convince termination-checker that `zs < xs`

I will add `dedup` as well but this is different function.

MatthewDaggitt: Ah yup, it is different. However the Haskell [`nub`](https://hoogle.haskell.org/?hoogle=nub) function seems to remove all duplicates, not just consecutive runs so I'm not sure that your function should really be called `nub`.

In general I kind of dislike the name `nub` as it's really not very descriptive, in the Haskell docs they claim it refers to the "essence" but that doesn't make any more sense to me either. I'd be tempted to name the two functions `deduplicate` and `derun`. Thoughts?

> dedup is O(n³)

As an aside, I'm pretty sure recent work on the implementation of decidable equality has got this down to O(n²).

src/Data/List/Base.agda
```diff
@@ -292,6 +294,20 @@ span P? (x ∷ xs) with does (P? x)
...
+nub : ∀ {R : Rel A p} → B.Decidable R → List A → List A
+nub R? []       = []
+nub R? (x ∷ xs) = x ∷ nub-filter R? x xs
```
MatthewDaggitt: Could we implement `nub` (or whatever we end up calling it!) as follows:
```agda
nub : ∀ {R : Rel A p} → B.Decidable R → List A → List A
nub R? []           = []
nub R? (x ∷ [])     = x ∷ []
nub R? (x ∷ y ∷ xs) with does (R? x y) | nub R? (y ∷ xs)
... | true  | nxs = nxs
... | false | nxs = x ∷ nxs
```
which would remove the need to have `nub-filter` and all it's own associated lemmas? (The nastiness of with abstracting the recursive call is needed due to termination checking getting less powerful in the upcoming release of Agda 2.6.1, see agda/agda#3604)

src/Data/List/Properties.agda
```diff
@@ -752,6 +753,64 @@ module _ {P : Pred A p} (P? : Decidable P) where
...
+
+  nub-filter-resp-≈ : ∀ y₁ y₂ xs → y₁ ≈ y₂ → nub-filter _≈?_ y₁ xs ≡ nub-filter _≈?_ y₂ xs
+  nub-filter-resp-≈ y₁ y₂ [] y₁≈y₂ = P.refl
+  nub-filter-resp-≈ y₁ y₂ (x ∷ xs) y₁≈y₂ with y₁ ≈? x | y₂ ≈? x
+  ... | true  because _    | true  because _    = nub-filter-resp-≈ y₁ y₂ xs y₁≈y₂
+  ... | true  because y₁≈x | false because y₂≉x = contradiction (≈-trans (≈-sym y₁≈y₂) (invert y₁≈x)) (invert y₂≉x)
+  ... | false because y₁≉x | true  because y₂≈x = contradiction (≈-trans y₁≈y₂ (invert y₂≈x)) (invert y₁≉x)
+  ... | false because _    | false because _    = P.refl
+
+  nub-resp-≈ : ∀ x₁ x₂ xs → x₁ ≈ x₂ → nub _≈?_ (x₁ ∷ xs) ≡  (nub _≈?_ (x₂ ∷ xs)) [ zero ]∷= x₁
```
MatthewDaggitt: I don't really understand this type at all. From the name of the proof I'd expect it to be `∀ {xs ys} → xs ≋ ys → nub R? xs ≋ nub R? ys`?

src/Data/List/Properties.agda
```diff
@@ -752,6 +753,64 @@ module _ {P : Pred A p} (P? : Decidable P) where
...
+
+  nub-idem : nub R? ∘ nub R? ≗ nub R?
+  nub-idem [] = refl
+  nub-idem (x ∷ xs) = cong (x ∷_) (nub-filter-idem x xs)
+
+module _ {b ℓ} (S : B.DecSetoid b ℓ) where
+
+  open B.DecSetoid S renaming (_≟_ to _≈?_; trans to ≈-trans; sym to ≈-sym)
+
+  nub-filter-resp-≈ : ∀ y₁ y₂ xs → y₁ ≈ y₂ → nub-filter _≈?_ y₁ xs ≡ nub-filter _≈?_ y₂ xs
```
MatthewDaggitt: Several of these arguments (here and elsewhere) can be made implicit.

src/Data/List/Relation/Unary/Grouped.agda
```diff
@@ -0,0 +1,41 @@
...
+open import Relation.Nullary.Product using (_×-dec_)
+open import Relation.Nullary.Negation using (¬?)
+open import Level using (_⊔_)
+
+infixr 5 _∷_ _∷≈_
+
+data Grouped {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) : Pred (List A) (a ⊔ ℓ) where
+  [] : Grouped _≈_ []
+  _∷_ : ∀ {x xs} → All (λ y → ¬ (x ≈ y)) xs → Grouped _≈_ xs → Grouped _≈_ (x ∷ xs)
+  _∷≈_ : ∀ {x y xs} → x ≈ y → Grouped _≈_ (y ∷ xs) → Grouped _≈_ (x ∷ y ∷ xs)
```
MatthewDaggitt: Any chance you could quickly describe the context in which this predicate is useful?

src/Data/List/Relation/Unary/Grouped.agda
```diff
@@ -0,0 +1,41 @@
...
+open import Relation.Nullary.Sum using (_⊎-dec_)
+open import Relation.Nullary.Product using (_×-dec_)
+open import Relation.Nullary.Negation using (¬?)
+open import Level using (_⊔_)
+
+infixr 5 _∷_ _∷≈_
+
+data Grouped {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) : Pred (List A) (a ⊔ ℓ) where
+  [] : Grouped _≈_ []
+  _∷_ : ∀ {x xs} → All (λ y → ¬ (x ≈ y)) xs → Grouped _≈_ xs → Grouped _≈_ (x ∷ xs)
```
MatthewDaggitt: How about calling this `_∷≉_` for symmetry?

src/Data/List/Relation/Unary/Grouped/Properties.agda
```diff
@@ -0,0 +1,82 @@
...
+  [_]⁺ : ∀ x → Grouped P [ x ]
+  [_]⁺ x = [] ∷ []
+
+------------------------------------------------------------------------
+-- nub
+
+module _ (S : DecSetoid a p) where
+
+  open DecSetoid S renaming (Carrier to A)
+  open import Data.List.Relation.Unary.Unique.Setoid setoid using (Unique; []; _∷_)
```
MatthewDaggitt: All import statements should be at the top of the file. You can rename the module something like `Unique` and then open it here.

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -179,6 +179,44 @@ module _ {p} {P : A → Set p} (P? : Decidable P) where
...
+-- nub and dedup
+
+module _ {ℓ} {R : Rel A ℓ} (R? : B.Decidable R) where
+  ∈-nub-filter⁻ : ∀ y {xs z} → z ∈ nub-filter R? y xs → z ∈ xs
+  ∈-nub-filter⁻ y {x ∷ xs} z∈nub-filter[R?,y,xs] with does (R? y x)
+  ∈-nub-filter⁻ y {x ∷ xs} z∈nub-filter[R?,y,xs]         | true  = there (∈-nub-filter⁻ y z∈nub-filter[R?,y,xs])
+  ∈-nub-filter⁻ y {x ∷ xs} (here refl)                   | false = here P.refl
+  ∈-nub-filter⁻ y {x ∷ xs} (there z∈nub-filter[R?,x,xs]) | false = there (∈-nub-filter⁻ x z∈nub-filter[R?,x,xs])
+
+  ∈-nub⁻ : ∀ xs {z} → z ∈ nub R? xs → z ∈ xs
```
MatthewDaggitt: These proofs need to be added to the changelog as well. Thanks!

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -179,6 +179,44 @@ module _ {p} {P : A → Set p} (P? : Decidable P) where
...
+  ∈-nub⁻ (x ∷ xs) (here refl) = here refl
+  ∈-nub⁻ (x ∷ xs) (there z∈nub[R?,xs]) = there (∈-nub-filter⁻ x z∈nub[R?,xs])
+
+  ∈-dedup⁻ : ∀ xs {z} → z ∈ dedup R? xs → z ∈ xs
+  ∈-dedup⁻ (x ∷ xs) (here refl) = here refl
+  ∈-dedup⁻ (x ∷ xs) (there z∈dedup[R,xs]) = there (∈-dedup⁻ xs (proj₁ (∈-filter⁻ (¬? ∘ R? x) z∈dedup[R,xs])))
+
+module _ {R : Rel A ℓ} (_≈?_ : B.Decidable (_≡_ {A = A})) where
```
MatthewDaggitt: It's neater to write this as `B.Decidable {A = A} _≡_`. We also tend to name this as `_≟_`.

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -179,6 +179,44 @@ module _ {p} {P : A → Set p} (P? : Decidable P) where
   ∈-filter⁻ : ∀ {v xs} → v ∈ filter P? xs → v ∈ xs × P v
   ∈-filter⁻ = Membershipₛ.∈-filter⁻ (P.setoid A) P? (P.subst P)
 
+------------------------------------------------------------------------
+-- nub and dedup
+
+module _ {ℓ} {R : Rel A ℓ} (R? : B.Decidable R) where
```
MatthewDaggitt: Any chance you could generalise these results to setoid membership and then simply re-export them here, as is done throughout the rest of the file?

src/Data/List/Membership/Propositional/Properties.agda
MatthewDaggitt: Also what is the extra `{R : Rel A ℓ}` for?

MatthewDaggitt: Finally shouldn't these proofs belong in `Membership.DecPropositional.Properties` if we have decidable equality?

src/Data/List/Relation/Unary/Grouped.agda
```diff
@@ -0,0 +1,41 @@
...
+  _∷≈_ : ∀ {x y xs} → x ≈ y → Grouped _≈_ (y ∷ xs) → Grouped _≈_ (x ∷ y ∷ xs)
+
+module _ {a ℓ} {A : Set a} {_≈_ : Rel A ℓ} where
+  grouped : B.Decidable _≈_ → U.Decidable (Grouped _≈_)
```
MatthewDaggitt: (`All` and `Any` notwithstanding, and which need to be made consistent with rest), the other predicates and relations call this property `grouped?` rather than `grouped`.

src/Data/List/Relation/Unary/Grouped/Properties.agda
```diff
@@ -0,0 +1,82 @@
...
+
+  grouped[y∷xs]⇒unique[y∷nub-filter[y,xs]] : ∀ y xs → Grouped _≈_ (y ∷ xs) → Unique (y ∷ nub-filter _≟_ y xs)
```
MatthewDaggitt: I assume you also meant to include a proof for `nub`?

src/Data/List/Base.agda
uzytkownik: > Ah yup, it is different. However the Haskell nub function seems to remove all duplicates, 

Sorry. My fault - blame my reading comprehention

> In general I kind of dislike the name nub as it's really not very descriptive, in the Haskell docs they claim it refers to the "essence" but that doesn't make any more sense to me either. I'd be tempted to name the two functions deduplicate and derun. Thoughts?

I'd think `deduplication` to be on long side. What about `dedup`? I'll rename `nub` to `derun`.

src/Data/List/Base.agda
uzytkownik: My reading of the agda/agda#3604 was that this used to be allowed but now it will be rejected by termination checker?

At least for recursive calls the agda though that y ∷ xs ≮ x ∷ y ∷ xs and failed the termination checker. My reading of agda/agda#3604 is that this will be true for with abstraction as well?

src/Data/List/Base.agda
MatthewDaggitt: > I'd think deduplication to be on long side.

Looking through `Data.List.Base` it looks like the vast majority of function names are written in full, e.g. `intersperse`, `unalignWith`, `replicate`, `tabulate`, `partition` etc. I think `deduplicate` would be more in keeping with these than `dedup` would be.

src/Data/List/Base.agda
MatthewDaggitt: So it's the following that used to be allowed but isn't.
```agda
nub : ∀ {R : Rel A p} → B.Decidable R → List A → List A
nub R? []           = []
nub R? (x ∷ [])     = x ∷ []
nub R? (x ∷ y ∷ xs) with does (R? x y)
... | true  | nxs = nub R? (y ∷ xs)
... | false | nxs = x ∷ nub R? (y ∷ xs)
```
My variant that is listed above is allowed, because the recursive call comes before the with abstraction. Indeed I've tested it out on the release candidate for `2.6.1` and it's passes termination checking.

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -505,6 +506,52 @@ module _ {P : A → Set p} {Q : A → Set q} (P? : Decidable P) where
...
+  filter⁻ {x ∷ xs}       all⁺        all⁻ with P? x  | ¬? (P? x)
+  filter⁻ {x ∷ xs}       all⁺        all⁻  | yes  Px | yes  ¬Px = contradiction Px ¬Px
+  filter⁻ {x ∷ xs} (qx ∷ all⁺)       all⁻  | yes  Px | no  ¬¬Px = qx ∷ filter⁻ all⁺ all⁻
+  filter⁻ {x ∷ xs}       all⁺  (qx ∷ all⁻) | no    _ | yes  ¬Px = qx ∷ filter⁻ all⁺ all⁻
+  filter⁻ {x ∷ xs}       all⁺        all⁻  | no  ¬Px | no  ¬¬Px = contradiction ¬Px ¬¬Px
+
+------------------------------------------------------------------------
+-- derun and deduplicate
+
+module _ {P : A → Set p} {Q : A → A → Set q} (Q? : B.Decidable Q) where
```
MatthewDaggitt: `Q` is usually reserved for unary relations. `R` is the usual name for binary relations. This applies generally throughout the PR (just made a note to add it to the style guide).

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -505,6 +506,52 @@ module _ {P : A → Set p} {Q : A → Set q} (P? : Decidable P) where
...
+  ... | true  = derun⁺ all[P,y∷xs]
+
+  deduplicate⁺ : ∀ {xs} → All P xs → All P (deduplicate Q? xs)
+  deduplicate⁺ [] = []
+  deduplicate⁺ {x ∷ _} (px ∷ all[P,xs]) = px ∷ filter⁺ (¬? ∘ Q? x) (deduplicate⁺ all[P,xs])
+
+  module _ (P-resp-Q : P B.Respects (flip Q)) where
+
+    private
+      derun⁻-aux : ∀ x xs → All P (derun Q? (x ∷ xs)) → All P (x ∷ xs)
```
MatthewDaggitt: This would be neater if you moved it into a `where` block underneath `derun⁻`. You could then get rid of the second nested anonymous module block.

src/Data/List/Relation/Unary/Grouped/Properties.agda
```diff
@@ -0,0 +1,76 @@
...
+open import Relation.Nullary using (¬_; does; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Level
+
+private
+  variable
+    a b c p q : Level
+    A : Set a
+    B : Set b
+    C : Set p
```
MatthewDaggitt: `Set c`?

src/Data/List/Relation/Unary/Any.agda
```diff
@@ -66,6 +66,10 @@ module _ {P : Pred A p} where
   lookup : ∀ {xs} → Any P xs → A
   lookup {xs} p = List.lookup xs (index p)
 
+  lookup-p : ∀ {xs} → (p : Any P xs) → P (lookup p)
```
MatthewDaggitt: I think this belongs in `Any.Properties` and how about calling it something slightly more meaningful (though I agree, it's a seriously hard one to name!). How about `lookup-result`?

src/Data/List/Relation/Unary/Grouped/Properties.agda
```diff
@@ -0,0 +1,76 @@
...
+
+------------------------------------------------------------------------
+-- map
+
+module _ (P : Rel A p) (Q : Rel B q) where
+
+  map⁺ : ∀ {f xs} → (∀ {x y} → P x y ⇔ Q (f x) (f y)) → Grouped P xs → Grouped Q (map f xs)
```
MatthewDaggitt: Please could you use the new `_⇔_` from `Function` (originally in `Function.Bundles`)?

CHANGELOG.md
```diff
@@ -643,13 +651,49 @@ Other minor additions
...
+  filter⁻ : All Q (filter P? xs) → All Q (filter (¬? ∘ P?) xs) → All Q xs
+  derun⁻ : All P (derun Q? xs) → All P xs
+  deduplicate⁻ : All P (deduplicate Q? xs) → All P xs
+  ```
+
+* Added new proofs to `Data.List.Relation.Unary.Any.Properties`:
+  ```agda
+  lookup-result : (p : Any P xs) → P (lookup p)
+  filter⁺ : (p : Any P xs) → Any P (filter Q? xs) ⊎ ¬ Q (Any.lookup p)
+  derun⁺ : Respects Q → Any P xs → Any P (derun Q? xs)
```
gallais: You are missing a `P` here in `P Respects Q`.

src/Data/List/Properties.agda
```diff
@@ -752,6 +753,36 @@ module _ {P : Pred A p} (P? : Decidable P) where
...
+  length-derun [] = ≤-refl
+  length-derun (x ∷ []) = ≤-refl
+  length-derun (x ∷ y ∷ xs) with does (R? x y) | length-derun (y ∷ xs)
+  ... | true  | r = ≤-step r
+  ... | false | r = s≤s r
+
+  length-deduplicate : ∀ xs → length (deduplicate R? xs) ≤ length xs
+  length-deduplicate [] = z≤n
+  length-deduplicate (x ∷ xs) = ≤-begin
+    1 + length (filter (¬? ∘ R? x) (deduplicate R? xs)) ≤⟨ s≤s (length-filter (¬? ∘ R? x) (deduplicate R? xs)) ⟩
```
gallais: Here you could save a lot of width by naming `deduplicate R? xs`
in a `let` before the `≤-begin`.

src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -459,6 +467,65 @@ module _ {P : A → Set p} where
...
+  filter⁺ {x ∷ xs} (here px) with Q? x
+  ... | true  because _       = inj₁ (here px)
+  ... | false because ofⁿ ¬Qx = inj₂ ¬Qx
+  filter⁺ {x ∷ xs} (there p) with Q? x
+  ... | true  because _       = Sum.map₁ there (filter⁺ p)
+  ... | false because _       = filter⁺ p
+
+  filter⁻ : ∀ {xs} → Any P (filter Q? xs) → Any P xs
+  filter⁻ {x ∷ xs} p with Q? x
+  filter⁻ {x ∷ xs} (here px) | true  because _ = here px
```
gallais: I believe using `with does (Q? x)` would allow you  to leave the `because _` out.

MatthewDaggitt:
Thanks for the changes @uzytkownik, looks great. Merging in now.

Just a heads-up in the future that you shouldn't need to force push every time. A normal push should do it, and preserves the comment history at the same time.

PR #1086, opened 2020-02-06
Add note for where case statement comes from

strangeglyph:
Add a brief comment on the import site of `case_of_` for people like me who are looking for the right module to import

### DIFF
--- a/README/Case.agda
+++ b/README/Case.agda
@@ -15,7 +15,7 @@ open import Data.Nat   hiding (pred)
 open import Data.List
 open import Data.Sum
 open import Data.Product
-open import Function
+open import Function -- case_of_ defined here
 open import Relation.Nullary
 open import Relation.Binary
 open import Relation.Binary.PropositionalEquality


strangeglyph:
As an additional note, the documentation somewhat implies that Agda supports standard case statements in places like [Function Definitions](https://agda.readthedocs.io/en/v2.5.4/language/function-definitions.html#case-trees), it might be worth noting there that this is pseudo-code and add a reference to this README file?

README/Case.agda
```diff
@@ -15,7 +15,7 @@ open import Data.Nat   hiding (pred)
 open import Data.List
 open import Data.Sum
 open import Data.Product
-open import Function
+open import Function -- case_of_ defined here
```
gallais: Why not *explicitly* listing the things imported here?
And then if you want to, you can mention `Function`
again in running text closer to a use site.

laMudri: As a largely orthogonal issue, shouldn't this be `Function.Base` these days? Perhaps this would even make navigation easier.

MatthewDaggitt: > As a largely orthogonal issue, shouldn't this be Function.Base these days? Perhaps this would even make navigation easier.

I guess it depends whether people want ease of use or to cut down on the number of dependencies...

MatthewDaggitt:
> As an additional note, the documentation somewhat implies that Agda supports standard case statements in places like Function Definitions, it might be worth noting there that this is pseudo-code and add a reference to this README file?

Hmm this needs addressing on the [main Agda](https://github.com/agda) repository. Feel free to open an issue or a PR over there.

MatthewDaggitt:
@strangeglyph I hope you don't mind but I pushed the suggested change straight to master in e23ac02. Avoided the hassle of downloading the entire of your repository.

Thank you for the suggestion :+1:

strangeglyph:
That's fine! I couldn't get Agda to build locally, that's why I didn't update the PR with the suggested changes.

PR #1088, opened 2020-02-13
[ new ] alternative closure operator

gallais:
missing CHANGELOG


src/Relation/Unary/Closure/Base.agda
```diff
@@ -32,17 +36,47 @@ open import Relation.Unary using (Pred)
...
--- Closed: whenever we have a value in one context, we can get one in any
+-- Diamond
+
+-- We then have the definition of ◇ ("diamond") which is named after the
+-- diamond modality in modal logic. `◇ T x` states that there exists an
+-- element one step away from x with respect to thre relation R that
+-- satisfies T
+
+◇ : ∀ {t} → Pred A t → Pred A (a ⊔ b ⊔ t)
+◇ T x = Σ[ support ∈ A ] (R support x × T support)
```
JacquesCarette: I was very surprised to see that `support` comes before `x`, i.e. I was expecting `R x support`, i.e. the step is into "the future" rather than in the past, wrt the order defined by the `R` relation. I am not claiming that this code is wrong!  I was just expecting, from the comments to "step from x to support" rather than "step from support to x".  Perhaps a few words of comments (or a link to the literature?) might help dispell that confusion from future readers.

src/Relation/Unary/Closure/Base.agda
```diff
@@ -32,17 +36,47 @@ open import Relation.Unary using (Pred)
 --   `Data.List.Relation.Sublist.Inductive`, `∀ {Γ} → Tm Γ → □ T Γ`
 --   corresponds to the fact that we have a notion of weakening for `Tm`.
 
--- Closed: whenever we have a value in one context, we can get one in any
+-- Diamond
+
+-- We then have the definition of ◇ ("diamond") which is named after the
+-- diamond modality in modal logic. `◇ T x` states that there exists an
+-- element one step away from x with respect to thre relation R that
```
fredrikNordvallForsberg:
```suggestion
-- element one step away from x with respect to the relation R that
```

MatthewDaggitt:
Looks good to me. No comments here!

src/Relation/Unary/Closure/Base.agda
gallais: Right. This is actually a departure from the modal logic metaphor.
I cannot really explain it except for the fact that it's the that
makes things work.

I'll had a reference to Conor's paper though.

MatthewDaggitt:
@gallais what's the status of this? Minus a CHANGELOG entry, it looks good to go? Though I notice you haven't removed the `status:being-worked-on` tag?

Issue #1083, opened 2020-02-03
Add tactics readme to main `README.md` file


MatthewDaggitt:

MatthewDaggitt:
Fixed in 571530b

PR #1090, opened 2020-02-24
Refactor reflection in new RingSolver and MonoidSolver

MatthewDaggitt:
This is mostly renaming, shuffling of things about and adding additional comments as I go through and try and get my head around reflection.

The one thing I'm not sure about is whether reflection utilities for specific data types should go in `Data.Nat.Reflection` or `Reflection.Data.Nat`? The former is much more standard, i.e. we have `Data.List.Relation` / `Data.Nat.Induction` / `Data.Nat.Tactic`, but maybe reflection is suitably "meta" that it deserves to be all grouped together...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,6 +310,9 @@ Other major additions
   Data.Word.Base
   Data.Word.Properties
   
+  Data.Nat.Reflection
+  Data.Fin.Reflection
+  
   Reflection.Abstraction
   Reflection.Argument
   Reflection.Argument.Information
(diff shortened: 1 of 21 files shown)


MatthewDaggitt:
I've had to move all the `show` functions into `Reflection.Show` in order to avoid a nasty dependency cycle via `Tactic.RingSolver` and `Data.Nat.DivMod`. Seems much cleaner to me this way anyway, as the show functionality pulls in large parts of the library that are unnecessary for much of the reflection infrastructure.

MatthewDaggitt:
Now fixes #1083

PR #1091, opened 2020-02-25
Eta-expand definition of `pure` in Data.Erased

jespercockx:
This very small change is required to make the standard library compatible with the fix for https://github.com/agda/agda/issues/4390 (see PR https://github.com/agda/agda/pull/4432). By default, Agda will no longer use subtyping rules for modalities (irrelevance, erasure, flat) so it is now required to eta-expand definitions if the argument is used at a different modality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,13 +56,14 @@ install:
     mkdir $HOME/.cabsnap;
     cp -a $HOME/.ghc $HOME/.cabsnap/ghc;
     cp -a $HOME/.cabal/lib $HOME/.cabal/share $HOME/.cabal/bin $HOME/installplan.txt $HOME/.cabsnap/;
+  - cd -
   # installing fix-agda-whitespace
   - git clone https://github.com/agda/fix-whitespace --depth=1
   - cd fix-whitespace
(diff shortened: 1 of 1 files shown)


gallais:
Looks like merging master into experimental broke the build. I am trying to fix
it & then you should be able to rebase on top of that to get the PR building.

You should probably also change the hash of the commit on Agda's master we are
using to make sure that everything is indeed working with `--no-subtyping`. It's
specified here:

https://github.com/agda/agda-stdlib/blob/experimental/.travis.yml#L34

gallais:
Looks like merging master into experimental broke the build. I am trying to fix
it & then you should be able to rebase on top of that to get the PR building.

You should probably also change the hash of the commit on Agda's master we are
using to make sure that everything is indeed working with `--no-subtyping`. It's
specified here:

https://github.com/agda/agda-stdlib/blob/experimental/.travis.yml#L34

gallais:
Pushed the wrong branch, it closed the PR and now I'm not allowed to push to
your remote anymore (or to reopen the PR)! I'll open another pr. :/

PR #1092, opened 2020-02-25
Eta-expand definition of `pure` in Data.Erased in response to https:/…

gallais:
…/github.com/agda/agda/pull/4432

ping #1091

We should wait until the PR is merged in Agda & we can then update the commit
we are using to run the tests. Hopefully that'll be the only change needed.

### DIFF
--- a/src/Data/Erased.agda
+++ b/src/Data/Erased.agda
@@ -36,7 +36,7 @@ map : (A → B) → Erased A → Erased B
 map f [ a ] = [ f a ]
 
 pure : A → Erased A
-pure = [_]
+pure x = [ x ]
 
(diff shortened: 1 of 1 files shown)


jespercockx:
I think this PR should be merged *before* the one on agda/agda, since this change is valid even for the old version of Agda. Then we can update the commit of the std-lib in the agda repository and merge my PR there.

jespercockx:
I know there is normally a waiting period, but since this is such a small change I will go ahead and merge it so I can run the tests on my PR on agda/agda.

MatthewDaggitt:
Go for it 👍

PR #1095, opened 2020-02-29
Use generalized variables in Category subdirectory

jespercockx:
This PR simply updates all modules in the Category subdirectory to use generalized variables for universe levels and types, hopefully making the type signatures a bit easier to read.

### DIFF
--- a/src/Category/Applicative.agda
+++ b/src/Category/Applicative.agda
@@ -11,27 +11,31 @@
 
 module Category.Applicative where
 
-open import Level using (suc; _⊔_)
+open import Level using (Level; suc; _⊔_)
 open import Data.Unit
 open import Category.Applicative.Indexed
 
-RawApplicative : ∀ {f} → (Set f → Set f) → Set (suc f)
+private
+  variable
+    f : Level
+
+RawApplicative : (Set f → Set f) → Set (suc f)
 RawApplicative F = RawIApplicative {I = ⊤} λ _ _ → F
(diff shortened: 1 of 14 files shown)


jespercockx:
Ah, I generally try to avoid using names that are too generic like `a` for universe levels. This is to avoid awkward situations like when you have `f : {a : A} -> ...`, it gets expanded to `f : {a : Level} {A : Set a} {a : A} -> ...` so `f {a = a}` does not mean what you think it means. Choosing more specific names for generalized variables avoids this problem.

However, I understand that this is technically a breaking change, so if you rather keep the types as they are then I'd be happy to change them.

MatthewDaggitt:
> Ah, I generally try to avoid using names that are too generic like a for universe levels. This is to avoid awkward situations like when you have f : {a : A} -> ..., it gets expanded to f : {a : Level} {A : Set a} {a : A} -> ... so f {a = a} does not mean what you think it means.

Agreed, this can be annoying. The library avoids this problem by reserving `a`, `b`, `c`, `d` purely for universe levels and all variables are named `x`, `y`, z`.

I think in cases where the universe levels can **always** be inferred, it's fair play to change them but in this case it looks like it is necessary to specify them. I'd therefore be tempted to change the name of the generalised variable to match the existing name.

jespercockx:
Ok, the original names are now all restored, so now there should be no more breaking changes.

Issue #559, opened 2018-12-07
Instance argument agda-stdlib policy?


xekoukou:
I would like if agda-stdlib used instance arguments. One could place them in different files so as not to contaminate the other files, as suggested by @jespercockx (1)

According to @UlfNorell , you could also hide instance arguments during import. (2)

(I have code that I want to contribute that has instance arguments)

1. https://lists.chalmers.se/pipermail/agda/2018/010699.html
2. https://lists.chalmers.se/pipermail/agda/2018/010698.html

MatthewDaggitt:
It's clear that there's a lot of demand for this, and putting them in separate modules would alleviate the concern of pollution that has prevented them from being added until now.

Would you like to put together a PR for what `Data.Nat.Instance` would look like? We can then discuss?

xekoukou:
If you are asking me to provide rules on how to organize modules with regard to instance resolution, please note that I am very bad at "organizing". I would prefer if others provided the rules and I simply followed them.

Maybe others could take the lead in this one.

WhatisRT:
Has there been any progress on this issue? I have written a lot of code that uses instance arguments which I would like to put in a pull request (Equality, Monoid, a couple of monad related classes).
My proposal would be to put the definitions (and functions that are defined on top of those classes) in a `Class` directory, such as `Class.Show` and then putting instances in `*.Instance` as previously suggested.

MatthewDaggitt:
> Has there been any progress on this issue?

I'm afraid not. I haven't really used instance arguments extensively myself, but the main requirements as I see it are:
- Instances should be separated out from the non-instance code, because, as I understand it, instances are imported automatically whenever you import a module.
- We should avoid the situation where we end up with parallel `instance` and `non-instance` definitions of common library concepts for which the theory then needs to be ported over. For example I'm a bit worried about the parallel definition of `_≤_` in `Data.Nat.Instance` in PR #560.
 
> My proposal would be to put the definitions (and functions that are defined on top of those classes) in a Class directory, such as Class.Show and then putting instances in *.Instance as previously suggested.

Would you mind fleshing out your proposal a little? Which definitions and functions would be put in the `Class` directory? What would the contents of `Data.Nat.Instance` then look like? Feel free to open a small example PR if that's easier to explain?

WhatisRT:
> * Instances should be separated out from the non-instance code, because, as I understand it, instances are imported automatically whenever you import a module.

[...]

PR #1096, opened 2020-02-29
Add `Instances` modules for exposing instances to instance search

jespercockx:
This is my attempt to get the discussion around #559 started again. So far I just added new modules containing instance declarations for instances that were already defined but not exposed to instance search. I also restricted myself to _unique_ instances, so there are e.g. no Functor instances for sum and product. In a later stage, we could also try to find out what other instances are missing and implement them.

As you can see, the `Instances` modules are really small and minimalistic, so they should not be too much of a burden to maintain.


gallais:
IIRC instances named `_` are only in scope in the file they are declared in,
not once imported. Cf. #2498

MatthewDaggitt:
> IIRC instances named _ are only in scope in the file they are declared in,
not once imported. Cf. agda/agda#2498

Additionally giving them names would allow users to select which ones they want to import? (or is that not a feature yet?)

In general though I'm happy with this design. Easy to generalise and out of the way if you don't want to use them. :+1:

jespercockx:
Ok, I added the names so they are actually available in other modules. Importing specific instances is more tricky because instances are added already on the `import` statement, so whether you open the module or not (and whether you hide specific fields or not) does not have any effect on instance search. But with well-designed instances it should never be a problem to have some extra unused instances floating around, so I do not believe this is a big problem.

jespercockx:
Travis says:
```
/home/travis/build/agda/agda-stdlib/src/Data/List/Instances.agda:21,23-56

Cannot instantiate the metavariable _18 to solution

Set (Agda.Primitive.lsuc ℓ) since it contains the variable ℓ

which is not in scope of the metavariable or irrelevant in the metavariable but relevant in the solution

λ {M} ⦃ inst ⦄ → monadT {ℓ} {M} inst has type

{M : Agda.Builtin.Unit.⊤ → Agda.Builtin.Unit.⊤ → Set ℓ → Set ℓ}

⦃ inst : Category.Monad.Indexed.RawIMonad M ⦄ →

Category.Monad.Indexed.RawIMonad

(λ _ _ →

   M Agda.Builtin.Unit.⊤.tt Agda.Builtin.Unit.⊤.tt Function.Base.∘′

   Agda.Builtin.List.List)
```
I'm not sure what's wrong though, `make test` goes through fine on my PC.

gallais:
Can you try picking a more recent commit hash in `.travis.yml`? It's here:

https://github.com/agda/agda-stdlib/blob/master/.travis.yml#L51

gallais:
That did it! <3

MatthewDaggitt:
@jespercockx only thing missing is a CHANGELOG entry. Any chance you could the entries to the list of new modules?

PR #1078, opened 2020-01-29
[ fix #870 ]  Finish deprecating Data.Table

gallais:

PR #1097, opened 2020-03-04
Use vec.functional for commutative monoids

alexarice:
I was getting warnings about CommutativeMonoid.Propeties using Data.Table which is deprecated. 
I think I have fixed these errors and added the appropriate properties needed to Data.Vec.Functional so that all the proofs still work. 

There could be some style issues as I haven't tried to change anything in the standard library before.


gallais:
Is this #1078?

alexarice:
Ah, probably, I missed that

Issue #1099, opened 2020-03-05
Add Haskell's `Foldable`


MatthewDaggitt:
The results in `Algebra.Operations.CommutativeMonoid` would really benefit from being proved over a suitable `Foldable` structure rather than individually for `List`/`Vec`/`Vector` etc.

See https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Foldable.html

Perhaps someone whose familiar with it, could suggest whether we can leverage dependant types to improve the definition somehow?

gallais:
I don't think it's a low hanging fruit. `Foldable` cannot fit in a record like
Functor or Monad do because of size issues!

jespercockx:
Can't we just have a `Foldable` that is parametrized over the level of the return type? At least that's what @UlfNorell did in https://github.com/UlfNorell/agda-prelude/blob/master/src/Container/Foldable.agda.

jamesmckinna:
Cf. My ICFP2017 talk at TyDe on `Foldable` https://icfp17.sigplan.org/track/tyde-2017-papers (and Tarmo Uustalu's independent discovery of the same thing somewhat earlier): it would be nice to have a suitable setting in which that characterisation is *provable*, and hence, as a by-product, be able to dispense with `Foldable` altogether... But the higher-rank characterisation perhaps hints at what dark level-polymorphic magic might be required to achieve that end.

jamesmckinna:
And yes, let's please try to have this: `sum` and `product` have no business being defined in `Data.List.Base`, being special to `ℕ` ... which means that `Data.List.Properties` ends up importing `Data.Nat.Divisibility` in order to establish a *single* property of `product`, while properties of `sum` such as `sum-++` are *not* defined as instances of the corresponding general property of `foldr`: *not* a rational approach, IMNSVHO.

jamesmckinna:
Meanwhile, is the general issue now closed by #2300 ? I hope so!

Issue #759, opened 2019-05-04
Deprecate old function hierarchy


mechvel:
I propose the following two backwards compatible additions. 

(1) To add to  ``Function.Injection`` 
``` 
InjectiveBy≡ : ∀ {a b} {A : Set a} {B : Set b} → (A → B) → Set _
InjectiveBy≡ f =  ∀ {x y} → f x ≡ f y → x ≡ y
```
To  add to ``Function.Surjection`` 
``` 
SurjectiveBy≡ : ∀ {a b} {A : Set a} {B : Set b} → (A → B) → Set _
SurjectiveBy≡ {_} {_} {_} {B} f =  (y : B) → ∃ (\x → f x ≡ y)
```
To add to  ``Function.Bijection`` 
```
BijectiveBy≡ : ∀ {a b} {A : Set a} {B : Set b} → (A → B) → Set _
BijectiveBy≡ f =  InjectiveBy≡ f × SurjectiveBy≡ f
```
(2) To reexport ``In(Sur,Bi)jective,  In(Sur,Bi)jectiveBy≡`` from ``Function``.

**Motivation.**
(2) is proposed because it is nicer to write a single ``... import Function using (...)``
      instead of the four import decls: 
      Function, Function.Injection, Function.Surjection,  Function.Bijection. 

(1) is proposed because ``In(Sur,Bi)jectiveBy≡`` have a wide application area and also are simple to perceive and to use, while the generic  ``In(Sur,Bi)jective`` are declared so that they are difficult to perceive and to use.
If the user really needs the general version, all right, let him study the general constructs given there by ``record``s.

gallais:
The problem I can see with these definitions is that we would now have two
(essentially the same on propositional equality) distinct notions of injectivity
and we would need to convert between them.

We already have "smart constructors" for injectivity involving propositional
equality, would that solve your problem? https://agda.github.io/agda-stdlib/Function.Injection.html#1399

mechvel:
For example, consider 
```
suc-injective :  ∀ {m n} → suc m ≡ suc n → m ≡ n
suc-injective refl = refl
```
in Data.Nat.Properties. Why does not it use ``Injective`` ? Can you demonstrate the usage of smart constructors for ``suc-injective : Injective suc`` ?

Now, I suggest the following.
```
------------------------------------------------------------ Function/GenericInjSurBi.agda ---------
module Function.GenericInjSurBi where
...
[...]

PR #1100, opened 2020-03-05
Suppress warnings from `Algebra.Properties.CommutativeMonoid`

MatthewDaggitt:
Whilst trying to improve #1078 I've run into several other outstanding issues:
 - (#759) it uses the old function hierarchy via `Data.Fin.Permutation` which we also want to deprecate.
 - (#1099) we want to prove most of these lemmas over some `Foldable` type rather than once each for list/vector/vec

Therefore to me, it feels that if we fix these problems by naively transferring the proofs over to `Vector` we're simply generating more code we're going to need to deprecate in the future. This is therefore an alternative PR to #1078 which simply suppresses the warnings for the mean time, and we can tackle the problem more completely in `v1.4`.

 @gallais thoughts?

### DIFF
--- a/src/Algebra/Operations/CommutativeMonoid.agda
+++ b/src/Algebra/Operations/CommutativeMonoid.agda
@@ -7,6 +7,9 @@
 
 {-# OPTIONS --without-K --safe #-}
 
+-- Disabled to prevent warnings from deprecated Table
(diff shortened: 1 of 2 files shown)


gallais:
Not sure how you'd prove these things generically. Happy to see us taking our
time on this & simply silencing the warning though.

PR #1101, opened 2020-03-05
Get rid of slow 'rewrite'

UlfNorell:
I checked the standard library with the now working `-vprofile.definitions:10` and picked the top entry.

Saves about 5s for the price of a more verbose (but readable IMO) proof.


UlfNorell:
I wanted to do that, but the problem is that `Function.Related.EquationalReasoning` is openend at the top-level.

UlfNorell:
The alternative is to add `where open Related.EquationalReasoning` in 6 places.

PR #1106, opened 2020-03-08
Typeclass instances for TC monad

jespercockx:
This adds instances of the standard typeclasses for the `TC` monad, and exposes them for instance search in a separate `Instances` module (see https://github.com/agda/agda-stdlib/pull/1096).

Possible point of discussion: currently the definitions `functor`, `applicative`, ... are exported directly from `Reflection.agda`. Maybe it would be better to have them in a separate module `Reflection.Categorical`? I'm not really sure what would be the best design.


MatthewDaggitt:
Looks good.

> Possible point of discussion: currently the definitions functor, applicative, ... are exported directly from Reflection.agda. Maybe it would be better to have them in a separate module Reflection.Categorical? I'm not really sure what would be the best design.

I would favour
- renaming the module `Reflection.TCMonad` to `Reflection.TypeChecking`
- moving the functor, applicative... definitions to `Reflection.TypeChecking.Categorical`
- moving the contents of `Reflection.TCMonadSyntax` to `Reflection.TypeChecking.Categorical` as well
- moving `Reflection.TCMonad.Instances` to `Reflection.TypeChecking.Instances`
- only re-export `Reflection.TypeChecking` from `Reflection`

All of these modules have been introduced since v1.3 so it won't have any backwards compatibility issues. Thoughts?

Also it will need a CHANGELOG entry whatever we decide.

PR #1108, opened 2020-03-09
Pull lattice-like structures out of Algebra, into Algebra.Lattice

laMudri:
Following [this comment](https://github.com/agda/agda-stdlib/issues/898#issuecomment-568286520), here is an attempt at moving and fixing the definitions of lattice-like structures found in the Algebra part of the library. What exists now is work towards an ideal hierarchy, which was done for clarity and my own convenience. For this to be merged, I'd expect to have better backwards compatibility.

Comments welcome.


src/Algebra/Lattice/Bundles.agda
```diff
@@ -0,0 +1,184 @@
...
+    _≈_           : Rel Carrier ℓ
+    _∙_           : Op₂ Carrier
+    isSemilattice : IsSemilattice _≈_ _∙_
+
+  open IsSemilattice isSemilattice public
+
+  band : Band c ℓ
+  band = record { isBand = isBand }
+
+  open Band band public using
```
MatthewDaggitt: The updated style guide says `using` should go on the next line down (which I think you're doing elsewhere already)

src/Algebra/Lattice/Bundles.agda
```diff
@@ -0,0 +1,184 @@
...
+  infixr 6 _∨_
+  infix  4 _≈_
+  field
+    Carrier   : Set c
+    _≈_       : Rel Carrier ℓ
+    _∨_       : Op₂ Carrier
+    _∧_       : Op₂ Carrier
+    isLattice : IsLattice _≈_ _∨_ _∧_
+
+  open IsLattice isLattice public
```
MatthewDaggitt: Should you also not be exporting (and renaming to) things like `∧-semilattice` here? Ditto for the two modules below as well.

src/Algebra/Structures.agda
```diff
@@ -128,17 +119,8 @@ record IsIdempotentCommutativeMonoid (∙ : Op₂ A)
...
--- from monoids rather than lattices.
-
-IsBoundedLattice = IsIdempotentCommutativeMonoid
-
-module IsBoundedLattice {∙ : Op₂ A}
-                        {ε : A}
-                        (isIdemCommMonoid : IsIdempotentCommutativeMonoid ∙ ε) =
-       IsIdempotentCommutativeMonoid isIdemCommMonoid
+  isBand : IsBand ∙
+  isBand = record { isSemigroup = isSemigroup ; idem = idem }
```
MatthewDaggitt: Note not to forgot to add this to the CHANGELOG

src/Algebra/Lattice/Structures.agda
```diff
@@ -0,0 +1,106 @@
...
+
+  open IsBand isBand public
+
+module IsMeetSemilattice {∧} (isSemilattice : IsSemilattice ∧) where
+  open IsSemilattice isSemilattice public
+    renaming (∙-cong to ∧-cong; ∙-congˡ to ∧-congˡ; ∙-congʳ to ∧-congʳ)
+
+module IsJoinSemilattice {∨} (isSemilattice : IsSemilattice ∨) where
+  open IsSemilattice isSemilattice public
+    renaming (∙-cong to ∨-cong; ∙-congˡ to ∨-congˡ; ∙-congʳ to ∨-congʳ)
```
MatthewDaggitt: I think at the very least, a comment is needed here to explain that these purely exist to rename the congruence proofs. which to be honest I don't think should ever have had `∙` added to their name in the first place...

I'm not entirely convinced of the utility of having separate join/meet variants. Maybe you can convince me? :)

src/Algebra/Lattice/Structures.agda
```diff
@@ -0,0 +1,106 @@
...
+
+
+IsBoundedSemilattice = IsIdempotentCommutativeMonoid
+module IsBoundedSemilattice
+  {∙ ε} (isBoundedSemilattice : IsBoundedSemilattice ∙ ε) where
+
+  open IsIdempotentCommutativeMonoid isBoundedSemilattice public
+
+  -- TODO: Maybe add isBand to IsIdempotentCommutativeMonoid.
```
MatthewDaggitt: Yup definitely, and then add simply `isBand = isBand` here.

src/Algebra/Lattice/Structures.agda
```diff
@@ -0,0 +1,106 @@
...
+module IsMeetSemilattice {∧} (isSemilattice : IsSemilattice ∧) where
+  open IsSemilattice isSemilattice public
+    renaming (∙-cong to ∧-cong; ∙-congˡ to ∧-congˡ; ∙-congʳ to ∧-congʳ)
+
+module IsJoinSemilattice {∨} (isSemilattice : IsSemilattice ∨) where
+  open IsSemilattice isSemilattice public
+    renaming (∙-cong to ∨-cong; ∙-congˡ to ∨-congˡ; ∙-congʳ to ∨-congʳ)
+
+
+IsBoundedSemilattice = IsIdempotentCommutativeMonoid
```
MatthewDaggitt: A comment above here, saying that a bounded semilattice is the same thing as an idempotent commutative monoid would be useful!

MatthewDaggitt:
@laMudri apologies, was trying to bring this up-to-date and ended up messing up the commits. I've ended up having to force push. I'm happy to take responsibility for merging this in for v2.0.

MatthewDaggitt:
Hmm actually this is more work then I'd first envisaged. @laMudri any chance you could help out by moving all the lattice constructions in `Algebra.Construct` to corresponding files in `Algebra.Lattice.Construct`, e.g.

https://github.com/agda/agda-stdlib/blob/b09525ccb0b0d0993ecbd3123eae1e72046b8263/src/Algebra/Construct/DirectProduct.agda#L93

MatthewDaggitt:
Also a couple of quick points on some of the design decisions. I reverted your change to define `IsLattice` in terms of two `IsSemilattice` proofs. As @Taneb found in https://github.com/agda/agda-stdlib/pull/1540, defining it in this way results in the record having two *different* proofs that the equality relation is an equivalence which is often undesirable. In order to keep the ease of constructing it, I've added your original version to a new `Algebra.Lattice.Structures.Biased`. Hope that's okay?

The second is that aren't we missing the smart constructors for the `IsDistributiveLattice`? Which ones were you planning on adding?

laMudri:
If this is for version 2, can't we find a general solution to these redundancy problems? For example, I'd start by taking the `isEquivalence` field out of `IsMagma` and putting it into `Magma`, `Monoid`, `Semiring`, &c.

I think any further thought or effort from me on this is going to come next weekend at the earliest. I'm happy to discuss the first point earlier, though, given that I have it in mind.

MatthewDaggitt:
> If this is for version 2, can't we find a general solution to these redundancy problems? For example, I'd start by taking the isEquivalence field out of IsMagma and putting it into Magma, Monoid, Semiring, &c.

Eek that sounds rather drastic. Isn't having a well-formed equality relation part of being a magma etc.? To be honest, with the current design of Agda, I don't think you're ever going to find a solution to all redundancy problems + get 100% usability. I think the best we can probably aim for is consistency and eliminate the worst ones. 

More concretely, ignoring the duplicate equality problem, what advantage does your definition of `IsLattice` have over the current one + smart constructors?

MatthewDaggitt:
Well that was unbelievably painful, let's not do that again :cry:

PR #1109, opened 2020-03-09
Add some standard algebraic constructs (unit and pairing)

laMudri:
This pull request (so far) adds zero and direct sum instances of many of the algebraic structures in the library. I'd want to eventually add these for semirings, rings, &c, and add tensor products, but I think that these are better done via (semi)modules.

### DIFF
--- a/src/Algebra/Construct/DirectSum.agda
+++ b/src/Algebra/Construct/DirectSum.agda
@@ -141,3 +141,41 @@ idempotentCommutativeMonoid M N = record
   where
   module M = IdempotentCommutativeMonoid M
   module N = IdempotentCommutativeMonoid N
+
+rawGroup : RawGroup c cℓ → RawGroup d dℓ → RawGroup (c ⊔ d) (cℓ ⊔ dℓ)
+rawGroup G H = record
+  { Carrier = G.Carrier × H.Carrier
+  ; _≈_ = Pointwise G._≈_ H._≈_
+  ; _∙_ = zip G._∙_ H._∙_
+  ; ε = G.ε , H.ε
+  ; _⁻¹ = map G._⁻¹ H._⁻¹
+  }
+  where
+  module G = RawGroup G
+  module H = RawGroup H
+
+group : Group c cℓ → Group d dℓ → Group (c ⊔ d) (cℓ ⊔ dℓ)
+group G H = record
+  { _⁻¹ = map G._⁻¹ H._⁻¹
+  ; isGroup = record
+    { isMonoid = Monoid.isMonoid (monoid G.monoid H.monoid)
+    ; inverse = (λ x → (G.inverseˡ , H.inverseˡ) <*> x)
+              , (λ x → (G.inverseʳ , H.inverseʳ) <*> x)
+    ; ⁻¹-cong = map G.⁻¹-cong H.⁻¹-cong
+    }
+  }
+  where
+  module G = Group G
+  module H = Group H
+
+abelianGroup : AbelianGroup c cℓ → AbelianGroup d dℓ →
+               AbelianGroup (c ⊔ d) (cℓ ⊔ dℓ)
+abelianGroup G H = record
+  { isAbelianGroup = record
+    { isGroup = Group.isGroup (group G.group H.group)
+    ; comm = λ x y → (G.comm , H.comm) <*> x <*> y
+    }
+  }
+  where
+  module G = AbelianGroup G
+  module H = AbelianGroup H
--- a/src/Algebra/Construct/DirectSum.agda
+++ b/src/Algebra/Construct/DirectSum.agda
@@ -1,0 +1,143 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Instances of algebraic structures made by taking two other instances
+-- A and B, and having elements of the new instance be pairs |A| × |B|.
+-- In mathematics, this would usually be written A ⊕ B.
+--
+-- From monoids up, these new instances are biproducts – i.e, both
+-- products and coproducts of the relevant category.
+------------------------------------------------------------------------
+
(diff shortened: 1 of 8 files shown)


laMudri:
> We usually have proofs of the `IsX` properties at the top-level and then manufacture
> bundles by simply feeding this proof. Cf. [Data.Nat.Properties](http://agda.github.io/agda-stdlib/Data.Nat.Properties.html#13411) for instance.

I had this in mind, but it feels like the types will get really awful if I do this. Still, if I need to, I guess I will.

src/Algebra/Construct/DirectSum.agda
```diff
@@ -0,0 +1,181 @@
...
+-- A and B, and having elements of the new instance be pairs |A| × |B|.
+-- In mathematics, this would usually be written A ⊕ B.
+--
+-- From monoids up, these new instances are biproducts – i.e, both
+-- products and coproducts of the relevant category.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.DirectSum where
```
MatthewDaggitt: I would probably be tempted to call this either `Algebra.Construct.Product` or `DirectProduct` rather than `DirectSum`. [As I understand it](https://mathworld.wolfram.com/GroupDirectSum.html), direct sums are a generalisation of direct products which only differs when it is a product over an infinite number of objects. 

This would also be more consistent with the `Sum` and `Product` terminology used elsewhere in the library.

src/Data/Product.agda
```diff
@@ -113,6 +114,18 @@ map₂ : ∀ {A : Set a} {B : A → Set b} {C : A → Set c} →
...
+dmap : ∀ {B : A → Set b} {P : A → Set p} {Q : ∀ {a} → P a → B a → Set q} →
+       (f : (a : A) → B a) → (∀ {a} (p : P a) → Q p (f a)) →
+       (ap : Σ A P) → Σ (B (proj₁ ap)) (Q (proj₂ ap))
+dmap f g (x , y) = f x , g y
+
+-- dmap for non-dependent products
+_<*>_ : ∀ {x y} {X : A → Set x} {Y : B → Set y} →
+        ((a : A) → X a) × ((b : B) → Y b) →
+        (ab : A × B) → X (proj₁ ab) × Y (proj₂ ab)
+(f , g) <*> xy = dmap f g xy
```
MatthewDaggitt: If we stick with the naming conventions used in this file, shouldn't this be called `dmap′` and should go with the other non-dependant synonyms at the bottom of the file?

MatthewDaggitt:
>> We usually have proofs of the IsX properties at the top-level and then manufacture
bundles by simply feeding this proof. Cf. Data.Nat.Properties for instance.
>
> I had this in mind, but it feels like the types will get really awful if I do this. Still, if I need to, I guess I will.

Indeed we actually often have individual proofs for each of the properties as well, e.g. associativity. It's often the case that I'll want an individual property from a `Construct` module without being able to build the entire structure I need to access it. e.g. I have something that fulfils the inverse property without all the other group properties.

I imagine laying them out explicitly also has the advantage that it should speed up type checking. At the moment, there is a **huge** amount of inference going on!

src/Algebra/Construct/DirectSum.agda
laMudri: Right, upon reading a bit more, my terminology is totally off here. Probably “biproduct” is the better description for structures with *commutative* addition (commutative monoids, Abelian groups, semirings, rings), in which case the comment at the top of the file is true. Otherwise, it seems that “direct product” or just “product” is right (and it really is just a product in the relevant category).

Concretely, as a module name, `DirectProduct` is probably good enough to cover both.

src/Data/Product.agda
```diff
@@ -160,6 +154,16 @@ curry′ = curry
...
+dmap′ : ∀ {x y} {X : A → Set x} {Y : B → Set y} →
+        ((a : A) → X a) → ((b : B) → Y b) →
+        (ab : A × B) → X (proj₁ ab) × Y (proj₂ ab)
+dmap′ f g = dmap f g
+
+_<*>_ : ∀ {x y} {X : A → Set x} {Y : B → Set y} →
+        ((a : A) → X a) × ((b : B) → Y b) →
+        (ab : A × B) → X (proj₁ ab) × Y (proj₂ ab)
+_<*>_ = uncurry dmap′
+
```
gallais: Note that you can write `((a , b) : A × B) → X a × Y b` nowadays :D

laMudri: I'm not set up for that yet, but indeed this should be done before merging.

gallais: Ah, fair enough.

src/Algebra/Module/Construct/DirectSum.agda
```diff
@@ -0,0 +1,167 @@
...
+--
+-- This module constructs the biproduct of two R-modules, and similar
+-- for weaker module-like structures.
+-- The intended universal property is that the biproduct is both a
+-- product and a coproduct in the category of R-modules.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Module.Construct.DirectSum where
```
MatthewDaggitt: Direct product again?

src/Data/Product.agda
```diff
@@ -113,6 +114,12 @@ map₂ : ∀ {A : Set a} {B : A → Set b} {C : A → Set c} →
        (∀ {x} → B x → C x) → Σ A B → Σ A C
 map₂ f = map id f
 
+-- A version of map where the output can depend on the input
```
JacquesCarette, line 117: all the changes in `Data.Product` (which are great) are not in the `CHANGELOG`.

laMudri: I'll get onto this now...

PR #1111, opened 2020-03-11
Proved that both [,] and map respect function equality in Data.Sum.Properties

MirceaS:


gallais:
In both `Data.List.Properties` and `Data.Maybe.Properties`, the same properties
for `map` are called `map-cong`. We should stick to that name (and adopt the same
naming convention for `[,]`.

MirceaS:
Addressed comment

src/Data/Sum/Properties.agda
```diff
@@ -44,6 +44,10 @@ module _ (dec₁ : Decidable {A = A} {B = A} _≡_)
 swap-involutive : swap {A = A} {B = B} ∘ swap ≗ id
 swap-involutive = [ (λ _ → refl) , (λ _ → refl) ]
 
+map-id : map (id {A = A}) (id {A = B}) ≗ id
```
MatthewDaggitt: You could get rid of half the brackets via `map {A = A} {B = B} id id ≗ id`?

MirceaS:
I reckon this should be ready to be merged

MirceaS:
Having used `[,]-∘-distr` extensively I found that it's only on very trivial cases that Agda can infer the value of the argument `f`, forcing me to always give it explicitly. Given that, I thought it would be a good idea to make the argument explicit to begin with.

src/Data/Sum/Properties.agda
```diff
@@ -58,6 +62,22 @@ swap-involutive = [ (λ _ → refl) , (λ _ → refl) ]
 
 map-commute : {f : A → B}  {g : C → D}
               {f′ : B → E} {g′ : D → F} →
-              ((map f′ g′) ∘ (map f g)) ≗ map (f′ ∘ f) (g′ ∘ g)
+              (map f′ g′) ∘ (map f g) ≗ map (f′ ∘ f) (g′ ∘ g)
```
MirceaS: Those brackets were pointless so I removed them

gallais: The ones around `map` probably are too.

PR #1123, opened 2020-03-22
Tree zippers

alexarice:
Adds leaf labelled trees, zippers and their properties


gallais:
This looks good to me, however it gets me wondering whether it would not be
better to merge this with `Data.Tree.Binary` and have a structure with both
nodes & leaves (each having their own type).

alexarice:
Then have an alias where either leaves or nodes are given the unit type?

alexarice:
Should I go ahead and implement this?

gallais:
I'd be in favour of it, yes!

MatthewDaggitt:
Closed in favour of #1125

PR #1039, opened 2020-01-04
[ admin ] first steps towards leaving CHANGELOG hell (re #521)

gallais:
I currently only support some basic functionalities. If we are happy with the
design I'll add all of the categories we are currently using. I am open to
renaming the directories (it's a matter of changing a variable):

* each PR is associated to a directory `PRDIR` in `changes/`
* each PR may declare:
  + **deprecations** for module `M.N.P` in `changes/PRDIR/deprecated/M.N.P`
    (two names per line, meaning the first got renamed to the second)
  + **additions** in module `M.N.P` in `changes/PRDIR/new/M.N.P`
    (one declaration per line)
  + **highlights** in `changes/PRDIR/highlights`
    (paragraphs are separated by empty lines)
  + **bug fixes** in `changes/PRDIR/bugfixes/FILE`
    each `FILE` contains a section for the bugfix part of the changelog;
    the `others` file is special: each paragraph (separated by empty lines)
    is an item and these get grouped across PRs under the `#### Others` section header.
  + **breaking** changes in `changes/PRDIR/breaking/FILE`
    with the same special treatment of the file `others` described for bugfixes

The `main` function explores all of the `changes/PRDIR` directories and merges
the contributions of each PR, outputting something like this (based on the current
content of the `changes/testN/` directories I have added as illustrations):

````
Highlights
----------
* A highlight

* My super cool shiny feature

* Btw, did you know that you can
  have highlights
  spanning multiple lines?
  As long as you are not skipping lines,
  it's all good.

* Otherwise It'll start a new one.

Bug fixes
---------

#### `_<_` in `Data.Integer`

* The definition of `_<_` in `Data.Integer` often resulted in unsolved metas
  when Agda had to infer the first argument. This was because it was
  previously implemented in terms of `suc` -> `_+_` -> `_⊖_`.

* To fix this problem the implementation has therefore changed to:
  ```agda
  data _<_ : ℤ → ℤ → Set where
    -<+ : ∀ {m n} → -[1+ m ] < + n
    -<- : ∀ {m n} → (n<m : n ℕ.< m) → -[1+ m ] < -[1+ n ]
    +<+ : ∀ {m n} → (m<n : m ℕ.< n) → + m < + n
  ```
  which should allow many implicit parameters which previously had
  to be given explicitly to be removed.

* All proofs involving `_<_` have been updated correspondingly

* For backwards compatibility the old relations still exist as primed versions
[...]

PR #1125, opened 2020-03-23
Tree zippers attempt 2

alexarice:
Merged leaf based and node based binary trees into one then added zippers

attempt number 2 at #1123 

cc @gallais 

~~wip as I would like to add functions to add to a zipper (like _++l_ and _++r_ in the list zipper file)~~

### DIFF
--- a/src/Data/Tree/Binary.agda
+++ b/src/Data/Tree/Binary.agda
@@ -10,56 +10,70 @@ module Data.Tree.Binary where
 
 open import Level using (Level; _⊔_)
 open import Data.List.Base using (List)
-open import Data.DifferenceList as DiffList using (DiffList; []; _∷_; _∷ʳ_; _++_)
+open import Data.DifferenceList as DiffList using (DiffList; []; _∷_; _∷ʳ_; _++_; [_])
 open import Data.Nat.Base using (ℕ; zero; suc; _+_)
 open import Function.Base
-open import Relation.Binary.PropositionalEquality
 
 private
   variable
-    a b : Level
+    a b c d : Level
     A : Set a
     B : Set b
+    C : Set c
+    D : Set d
 
-data Tree (A : Set a) : Set a where
-  leaf : Tree A
-  node : Tree A → A → Tree A → Tree A
+data Tree (A : Set a) (B : Set b) : Set (a ⊔ b) where
+  leaf : B → Tree A B
+  node : Tree A B → A → Tree A B → Tree A B
 
-map : (A → B) → Tree A → Tree B
-map f leaf         = leaf
-map f (node l m r) = node (map f l) (f m) (map f r)
+map : (A → C) → (B → D) → Tree A B → Tree C D
+map f g (leaf x)     = leaf (g x)
+map f g (node l m r) = node (map f g l) (f m) (map f g r)
 
-size : Tree A → ℕ
-size leaf         = 0
-size (node l m r) = size l + suc (size r)
+#nodes : Tree A B → ℕ
+#nodes (leaf x)     = 0
+#nodes (node l m r) = #nodes l + suc (#nodes r)
 
-foldr : (B → A → B → B) → B → Tree A → B
-foldr nd lf leaf         = lf
-foldr nd lf (node l m r) = nd (foldr nd lf l) m (foldr nd lf r)
+#leaves : Tree A B → ℕ
+#leaves (leaf x)     = 1
+#leaves (node l m r) = #leaves l + #leaves r
+
+foldr : (C → A → C → C) → (B → C) → Tree A B → C
+foldr f g (leaf x)     = g x
+foldr f g (node l m r) = f (foldr f g l) m (foldr f g r)
 
 ------------------------------------------------------------------------
 -- Extraction to lists, depth first and left to right.
 
 module Prefix where
 
-  toDiffList : Tree A → DiffList A
-  toDiffList = foldr (λ l m r → m ∷ l ++ r) []
+  toDiffList : Tree A B → DiffList A
+  toDiffList = foldr (λ l m r → m ∷ l ++ r) (λ _ → [])
 
-  toList : Tree A → List A
+  toList : Tree A B → List A
   toList = DiffList.toList ∘′ toDiffList
 
 module Infix where
 
-  toDiffList : Tree A → DiffList A
-  toDiffList = foldr (λ l m r → l ++ m ∷ r) []
+  toDiffList : Tree A B → DiffList A
+  toDiffList = foldr (λ l m r → l ++ m ∷ r) (λ _ → [])
 
-  toList : Tree A → List A
+  toList : Tree A B → List A
   toList = DiffList.toList ∘′ toDiffList
 
 module Suffix where
 
-  toDiffList : Tree A → DiffList A
-  toDiffList = foldr (λ l m r → l ++ r ∷ʳ m) []
+  toDiffList : Tree A B → DiffList A
+  toDiffList = foldr (λ l m r → l ++ r ∷ʳ m) (λ _ → [])
 
-  toList : Tree A → List A
+  toList : Tree A B → List A
+  toList = DiffList.toList ∘′ toDiffList
+
+module Leaves where
+
+  toDiffList : Tree A B → DiffList B
+  toDiffList (leaf x) = [ x ]
+  toDiffList (node l m r) = toDiffList l ++ toDiffList r
+
+  toList : Tree A B → List B
   toList = DiffList.toList ∘′ toDiffList
--- a/src/Data/Tree/Binary.agda
+++ b/src/Data/Tree/Binary.agda
@@ -32,11 +32,11 @@ map : (N → N₁) → (L → L₁) → Tree N L → Tree N₁ L₁
 map f g (leaf x)     = leaf (g x)
 map f g (node l m r) = node (map f g l) (f m) (map f g r)
 
-map₁ : (N → N₁) → Tree N L → Tree N₁ L
-map₁ f t = map f id t
+mapₙ : (N → N₁) → Tree N L → Tree N₁ L
+mapₙ f t = map f id t
 
-map₂ : (L → L₁) → Tree N L → Tree N L₁
-map₂ f t = map id f t
+mapₗ : (L → L₁) → Tree N L → Tree N L₁
+mapₗ f t = map id f t
 
 #nodes : Tree N L → ℕ
 #nodes (leaf x)     = 0
--- a/src/Data/Tree/Binary.agda
+++ b/src/Data/Tree/Binary.agda
@@ -30,6 +30,12 @@ map : (A → C) → (B → D) → Tree A B → Tree C D
 map f g (leaf x)     = leaf (g x)
 map f g (node l m r) = node (map f g l) (f m) (map f g r)
 
+map₁ : (A → C) → Tree A B → Tree C B
+map₁ f t = map f id t
+
+map₂ : (B → C) → Tree A B → Tree A C
+map₂ f t = map id f t
+
 #nodes : Tree A B → ℕ
 #nodes (leaf x)     = 0
 #nodes (node l m r) = #nodes l + suc (#nodes r)
--- a/src/Data/Tree/Binary.agda
+++ b/src/Data/Tree/Binary.agda
@@ -16,35 +16,37 @@ open import Function.Base
 
 private
   variable
-    a b c d : Level
+    n l n₁ l₁ a : Level
+    N : Set n
+    L : Set l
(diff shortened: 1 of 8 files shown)


src/Data/Tree/Binary.agda
```diff
@@ -10,56 +10,70 @@ module Data.Tree.Binary where
...
-map f (node l m r) = node (map f l) (f m) (map f r)
+map : (A → C) → (B → D) → Tree A B → Tree C D
+map f g (leaf x)     = leaf (g x)
+map f g (node l m r) = node (map f g l) (f m) (map f g r)
 
-size : Tree A → ℕ
-size leaf         = 0
+size : Tree A B → ℕ
+size (leaf x)     = 0
 size (node l m r) = size l + suc (size r)
```
gallais: I think this should be renamed to `nodes` for consistency with the other notions
of size one being called `leaves`. Or rather, it may be a good idea to call it
something like `#nodes` while `nodes` would be a function `Tree A B -> List A`!

src/Data/Tree/Binary/Zipper/Properties.agda
```diff
@@ -0,0 +1,106 @@
...
+      1 + BT.size foc + (BT.size l + sum (List.map (suc ∘ BT.size ∘ getTree) ctx)) ≡˘⟨ +-assoc (1 + BT.size foc) (BT.size l) (sum (List.map (suc ∘ BT.size ∘ getTree) ctx)) ⟩
+      1 + BT.size foc + BT.size l + sum (List.map (suc ∘ BT.size ∘ getTree) ctx) ≡⟨ cong (λ y → y + sum (List.map (suc ∘ BT.size ∘ getTree) ctx)) (+-comm (1 + BT.size foc) (BT.size l)) ⟩
+      size (mkZipper ctx (node l m foc)) ≡⟨ helper ctx (node l m foc) ⟩
+      BT.size (toTree (mkZipper (leftBranch m l ∷ ctx) foc)) ∎
+    helper (rightBranch m r ∷ ctx) foc = begin
+      BT.size foc + (1 + (BT.size r + sum (List.map (suc ∘ BT.size ∘ getTree) ctx))) ≡˘⟨ cong (λ y → BT.size foc + y) (+-assoc 1 (BT.size r) (sum (List.map (λ x → suc (BT.size (getTree x))) ctx))) ⟩
+      BT.size foc + (1 + BT.size r + sum (List.map (suc ∘ BT.size ∘ getTree) ctx)) ≡˘⟨ +-assoc (BT.size foc) (suc (BT.size r)) (sum (List.map (suc ∘ BT.size ∘ getTree) ctx)) ⟩
+      size (mkZipper ctx (node foc m r)) ≡⟨ helper ctx (node foc m r) ⟩
+      BT.size (toTreeHelper ctx (node foc m r)) ∎
```
gallais: The subexpressions `sum (List.map (suc ∘ BT.size ∘ getTree) ctx)` appears a lot,
it is worth naming it in a `let`.

alexarice:
Have changed size to #nodes, leaves to #leaves and fixed the other problems

alexarice:
These changes have broken Text.Pretty.Core . I can fix it though I don't know if it is worth setting up aliases for trees with no node labels or no leaf labels. I also don't know the best way to do this, should both versions be put in different files where the Types, constructors and functions are aliased?

src/Data/Tree/Binary/Zipper.agda
```diff
@@ -0,0 +1,105 @@
...
+
+right : Zipper A B → Maybe (Zipper A B)
+right (mkZipper ctx (leaf x)) = nothing
+right (mkZipper ctx (node l m r)) = just $ mkZipper (leftBranch m l ∷ ctx) r
+
+-- To and from trees
+------------------------------------------------------------------------
+
+toTreeHelper : List (Crumb A B) → Tree A B → Tree A B
+toTreeHelper [] t = t
```
gallais: This is usually called `plug` as we plug the hole in the one-hole context
with the term that is supposed to sit in it.

src/Data/Tree/Binary/Zipper.agda
```diff
@@ -0,0 +1,105 @@
...
+------------------------------------------------------------------------
+
+infixr 5 _⟪_⟫ˡ_
+infixl 5 _⟪_⟫ʳ_
+
+_⟪_⟫ˡ_ : Tree A B → A → Zipper A B → Zipper A B
+l ⟪ m ⟫ˡ mkZipper ctx foc = mkZipper (ctx ++ [ (leftBranch m l) ]) foc
+
+_⟪_⟫ʳ_ : Zipper A B → A → Tree A B → Zipper A B
+mkZipper ctx foc ⟪ m ⟫ʳ r = mkZipper (ctx ++ [ (rightBranch m r) ]) foc
```
gallais: These operations are costly. They should be generalised to a full
`Zipper A B -> List (Crumb A B) -> Zipper A B` so that we can pour
a whole one-hole context on top of the current one without iterating
over `_⟪_⟫ˡ_` (it would be quadratic instead of linear). We can then
have them as a special case.

gallais:
> I don't know if it is worth setting up aliases for trees with no node labels or no leaf labels

I am not sure it is.

gallais:
Btw, if you don't force-push it is easier to see the latest changes (and we will
squash the whole PR at the end anyway so the git history will be kept clean even
if there are multiple commits in the PR).

alexarice:
> 
> 
> Btw, if you don't force-push it is easier to see the latest changes (and we will
> squash the whole PR at the end anyway so the git history will be kept clean even
> if there are multiple commits in the PR).

sorry, am used to trying to keep commit histories clean

alexarice:
Everything should be fixed now I think

src/Text/Pretty/Core.agda
```diff
@@ -154,14 +156,14 @@ private
     size-indent nothing    str = refl
     size-indent (just pad) str = length-++ pad str
 
-    indents : Maybe String → Tree String → Tree String
-    indents = maybe′ (Tree.map ∘ _++_) id
+    indents : Maybe String → Tree String ⊤ → Tree String ⊤
+    indents = maybe′ (flip Tree.map id ∘ _++_) id
```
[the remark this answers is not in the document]
alexarice: Should the original map still exist?
If so should it be called bi-map or something similar to show it is a bifunctor?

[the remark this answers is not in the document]
alexarice: Have copied the names in `Data.Sum.Base`

alexarice:
Have addressed all comments

alexarice:
Do I need to squash the commits or will that be done automatically?

gallais:
That will be done automatically. We will also need an entry in the CHANGELOG but
I need to find the time to get the new system (#1039) merged in first.

src/Data/Tree/Binary.agda
```diff
@@ -10,56 +10,76 @@ module Data.Tree.Binary where
...
-map f leaf         = leaf
-map f (node l m r) = node (map f l) (f m) (map f r)
+map : (A → C) → (B → D) → Tree A B → Tree C D
+map f g (leaf x)     = leaf (g x)
+map f g (node l m r) = node (map f g l) (f m) (map f g r)
 
-size : Tree A → ℕ
-size leaf         = 0
-size (node l m r) = size l + suc (size r)
+map₁ : (A → C) → Tree A B → Tree C B
```
MatthewDaggitt: I'd be tempted to call this `mapₙ` and `mapₗ` rather than `1` and `2`?

src/Data/Tree/Binary.agda
```diff
@@ -10,56 +10,76 @@ module Data.Tree.Binary where
...
+    a b c d : Level
     A : Set a
     B : Set b
+    C : Set c
+    D : Set d
 
-data Tree (A : Set a) : Set a where
-  leaf : Tree A
-  node : Tree A → A → Tree A → Tree A
+data Tree (A : Set a) (B : Set b) : Set (a ⊔ b) where
```
MatthewDaggitt: We should either add a comment explaining which type is which, or we should name types `A` and `B` something a little more meaningful?

src/Data/Tree/Binary/Relation/Unary/All/Properties.agda
```diff
@@ -12,15 +12,24 @@ open import Level
...
+    a b c d p q : Level
     A : Set a
     B : Set b
+    C : Set c
+    D : Set d
+    P : C → Set p
+    Q : D → Set q
 
-module _ {P : B → Set p} where
+map⁺ : (f : A → C) → (g : B → D) → ∀[ All (f ⊢ P) (g ⊢ Q) ⇒ Tree.map f g ⊢ All P Q ]
```
MatthewDaggitt: Hmm, this is another one of those cases where I think we might be using too much notation. I'm really struggling to interpret the exact type of this. Perhaps expanding at least some of the notation out might make it more usable?

src/Data/Tree/Binary/Zipper.agda
```diff
@@ -0,0 +1,108 @@
...
+record Zipper (A : Set a) (B : Set b) : Set (a ⊔ b) where
+  constructor mkZipper
+  field
+    context : List (Crumb A B)
+    focus : Tree A B
+
+open Zipper public
+
+-- Fundamental operations of a Zipper: Moving around
+------------------------------------------------------------------------
```
MatthewDaggitt: We usually put the long `-----------` above the section title, rather than below.

src/Data/Tree/Binary/Zipper/Properties.agda
```diff
@@ -0,0 +1,145 @@
...
+
+-- Tree-like operations indeed correspond to their counterparts
+------------------------------------------------------------------------
+
+toTree-#nodes-commute : ∀ (zp : Zipper A B) → #nodes zp ≡ BT.#nodes (toTree zp)
+toTree-#nodes-commute (mkZipper c v) = helper c v
+  where
+    helper : (cs : List (Crumb A B))
+           → (t : Tree A B)
+           → #nodes (mkZipper cs t) ≡ BT.#nodes (toTree (mkZipper cs t))
```
MatthewDaggitt: The [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) suggests that function arrows go at the end of lines rather than the beginning.

src/Data/Tree/Binary.agda
alexarice: sounds good, I'll fix this

src/Data/Tree/Binary.agda
alexarice: I'll do both

src/Data/Tree/Binary/Relation/Unary/All/Properties.agda
alexarice: I agree though I don't think it is more confusing than before

src/Data/Tree/Binary/Zipper.agda
alexarice: I think I copied this from the list zipper file, though I am happy to change it

MatthewDaggitt: Ah okay, that obviously slipped past me last time! I'd never condone something so obviously wrong as putting the dashes below rather than above :smile: If you wouldn't mind changing it that'd be great, thanks.

src/Data/Tree/Binary/Relation/Unary/All/Properties.agda
alexarice: I've changed this though I don't know if it is any better

MatthewDaggitt: Looks ever so slightly better to me...

MatthewDaggitt:
Everything looks good now. Any chance you could add some entries in the CHANGELOG describing the changes? See the [previous one](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.3.md) for an idea of the style we're going for.

After that I'll merge it in 👍

alexarice:
> Everything looks good now. Any chance you could add some entries in the CHANGELOG describing the changes? See the [previous one](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.3.md) for an idea of the style we're going for.
> 
> After that I'll merge it in +1

I don't need to wait for the changelog pr?

MatthewDaggitt:
> I don't need to wait for the changelog pr?

No. We've decided to merge in outstanding PRs and then we'll refactor the existing entries when we merge the new changelog PR in.

alexarice:
Have done by best attempt, Let me know if more detail is needed.

MatthewDaggitt:
> Have done by best attempt, Let me know if more detail is needed.

Looks great! Merging in. Thanks for the PR!

PR #1127, opened 2020-03-26
[ new ] Nat.Binary.Subtraction

mechvel:
new file:    src/Algebra/Morphism/TwoMagmas.agda
changed:     src/Data/Nat/Binary/Properties.agda
new file:    src/Data/Nat/Binary/Subtraction.agda
changed:     src/Data/Nat/Properties.agda


mechvel:
Please, consider this commit of March 26, 2020  for  `Nat.Binary.Subtraction`.                           
                                                                                      
It is a port of  #1026  to _master_ after lib-1.3.                                    
It takes in account all your notes, the last of which was on  January 25, 2020.                                                                     
                                                                                      
Then, lib-1.3 has done many changes.   So that I have got confused with mechvel-stdlib versions.                             
I tried things like  ``git pull``, and failed.                                                                                                                             
Therefor, I have                                                    
* saved the needed files, and our correspondence on #1026,                            
* deleted the  mechvel-stdlib  fork, and restart it now.                                  
                                                                                      
I commit now with  Binary.Subtraction  (old Minus renamed),                           
with adding to some other files the minimal item set that is used  in Subtraction.

New files:                                                     
src/Algebra/Morphism/TwoMagmas.agda,                                                  
src/Data/Nat/Binary/Subtraction.agda                                                  
                                                                                      
Small  changes are done to                                                             
src/Data/Nat/Binary/Properties.agda, src/Data/Nat/Properties.agda.                    
                                                                                                                                                               
I reproduce below our discussion on  #1026,  with adding some fresh comments,                                                      
and with moving the subject of  Morphism.TwoMagmas  to the end.

mechvel:
@gallais                                                                                                                                                            
src/Data/Nat/Binary/Minus.agda                                                        
...                                                                                   
open import Data.Fin using () renaming (zero to 0F; suc to sucF)                      
                                                                                      
    Instead of renaming these and writing 0F, sucF 0F and sucF (sucF 0F),             
    you could use #_: all your cases are concrete so Agda should see that             
    you can indeed embed 0, 1, and 2 into Fin 3.                                      
                                                                                      
Done.                                                                                 

```                                                                                   
... | yes _ = 0ᵇ                                                                      
... | no  _ = 1+[2 (x ∸ y) ]                                                          
```                                                                                   
                                                                                      
@gallais:                                                                             
> Same comments as earlier: avoid yes / no whenever you do not need the proofs         
> and use does instead. This will give you more efficient functions                    
> (essentially computing a boolean).                                                   
       
                                                                                      
I tried there                                                                         
```                                                                                   
2[1+ x ] ∸ 1+[2 y ] with does (x <? y)                                                
... | true  = 0ᵇ                                                                      
... | false = 1+[2 (x ∸ y) ]                                                          
```                                                                                   
And it reports                                                                        
"The pattern variable true has the same name as the constructor                       
Agda.Builtin.Bool.Bool.true                                                           
when checking the clause Data.Nat.Binary.Minus.with-234 x y true = 0ᵇ                 
"                                                                                     
-- no matter whether true, false imported from Bool or not.                           
But a similar code in a certain my application works under lib-1.2.                   
What makes this difference, I wonder.                                                 
Is this due to --without-K --safe ?

mechvel:
@gallais                                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
x>y⇒x∸y≢0 {x'@(2[1+ x ])} {zero} _ =  subst (_≢ 0ᵇ) x'≡x'∸0 2[1+x]≢0                  
  where                                                                               
  x'≡x'∸0 = sym (x∸0≡x x')                                                            
```                                                                                   
> I think we can inline such trivial proofs.                                          
                                                                                      
This is only the first pattern for ``x>y⇒x∸y≢0``.                                     
Sorry, I have separated these patterns with empty lines.                              
Now I remove these empty lines, to dutifully make this body difficult to read.        
                                                                                      
And still, if it was a single pattern, and we inline it, then what would we           
have to write instead of, say,                                                        
```                                                                                   
f x y x>y =  g (x + y) (x>y⇒x∸y≢0 x>y)                                                
?                                                                                     
Probably:                                                                             
  f x y x>y =  g (x + y) x∸y≢0                                                        
               where                                                                  
               x∸y≢0 = subst (_≢ 0ᵇ) (sym (x∸0≡x x')) 2[1+x]≢0  -- (1)                
```                                                                                   
And each time one will need to invent a proof like (1) ...

mechvel:
@gallais                                                                              
 src/Data/Nat/Binary/Minus.agda                                                        
```                                                                                   
... | tri≈ _ x≡y _ =  begin                                                           
  toℕ (x ∸ y)  ≡⟨ cong toℕ (x≤y⇒x∸y≡0 {x} {y} x≤y) ⟩                                  
  toℕ 0ᵇ       ≡⟨ refl ⟩                                                              
```                                                                                   
> ``≡⟨ refl ⟩`` can be replaced with ``≡⟨⟩`` (also applies to other lines in  the file).                                                                          
                                                                                      
Done.                                                                                 
   
```                                                                                   
fromℕ-homo-∸ :  ∀ m n → fromℕ (m ℕ.∸ n) ≡ (fromℕ m) ∸ (fromℕ n)                       
fromℕ-homo-∸ m n =  begin-equality                                                    
```                                                                                   
> Given that we have fromℕ-toℕ, it feels like we ought to be able to prove            
> a more general lemma. i.e.                                                          
```                                                                                   
>  (f : ℕ → ℕ) (g : ℕᵇ → ℕᵇ) →                                                        
>     (∀ m n → toℕ (g m n) ≡ f (toℕ m) (toℕ n)) →                                     
>     (∀ m n → fromℕ (f m n) ≡ g (fromℕ m) (fromℕ n))                                 
```                                                                                   
> and the converse.                                                                   
                                                                                      
Probably you mean ... ``(f : Op₂ ℕ) (g : Op₂ ℕᵇ) → ....``                             
This is implemented now in Algebra.Morphism.TwoMagmas.

mechvel:
MatthewDaggitt:                                                                       
                                                                                      
> Okay here's the latest round of comments.                                           
src/Data/Nat/Properties.agda                                                          
src/Data/Nat/Binary/Properties.agda                                                   
```                                                                                   
import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup                 
  as Bin+CSemigroup                                                                   
import Algebra.Morphism.TwoMagmas +-magma ℕₚ.+-magma                                  
  as Magmas+ℕᵇ-ℕ                                                                      
```                                                                                   
[..]                                                                                  
                                                                                      
> Yup so the standard thing to do is to use the as syntax to rename the               
> module to something meaningful but short, then open it later on.                    
      
src/Data/Nat/Binary/Minus.agda                                                        
```                                                                                   
∸-magma = magma _∸_``                                                                 
                                                                                      
import Algebra.Morphism.TwoMagmas ∸-magma ℕₚ.∸-magma                                  
  as Magmas∸ℕᵇ-ℕ                                                                      
```                                                                                   
> Import statements should go at the top of the file                                  
                                                                                      
But it uses a parameter value ``∸-magma`` implemented above.                          
(?)

mechvel:
MatthewDaggitt on 29 Dec 2019                                                         
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
  y ∸ z                ∎                                                              
  where open ≤-Reasoning                                                              
                                                                                      
suc-x∸suc-y :  ∀ x y → suc x ∸ suc y ≡ x ∸ y                                          
```                                                                                   
> Please follow the style guide and use square brackets instead of -                  
> to indicate application in proof names.                                             
                                                                                                                                                                            
Renamed to ``suc[x]∸suc[y]``.

mechvel:
MatthewDaggitt reviewed on 25 Jan                                                     
                                                                                      
> Another round of comments.                                                                                                                                                
```
src/Data/Nat/Binary/Minus.agda                                                       
...
module Data.Nat.Binary.Minus where                                                    
```                                                                                      
                                                                                      
> Perhaps Subtraction is a better name than Minus?                                    
                                                                                      
I rename it to  Data.Nat.Binary.Subtraction.                                          
It is all the same for me.                                                            

src/Data/Nat/Binary/Minus.agda                                                        
```                                                                                   
[x+y]∸x≡y x y = trans (cong (_∸ x) (+-comm x y)) ([x+y]∸y≡x y x)                      
                                                                                      
∸-identityʳ : RightIdentity 0ᵇ _∸_                                                    
∸-identityʳ x = trans (cong (_∸ 0ᵇ) $ sym $ +-identityʳ x) ([x+y]∸y≡x x 0ᵇ)           
```                                                                                   
> You've already proved this... It's just x∸0≡x                                       
                                                                                      
                                                                                      
Replaced with ``∸-identityʳ = x∸0≡x``                                                 
                                                                                      
                                                                                      
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
  open ≤-Reasoning;  k = toℕ x;  m = toℕ y;  n = toℕ z                                
                                                                                      
*-distribʳ-∸ : _*_ DistributesOverʳ _∸_                                               
*-distribʳ-∸ x y z = begin-equality                                                   
```                                                                                   
> See comm+distrˡ⇒distrʳ in Algebra.Consequences.Propositional                        
                                                                                      
Changed to                                                                            
``*-distribʳ-∸ = comm+distrˡ⇒distrʳ *-comm *-distribˡ-∸  ...``                        
    
      
src/Data/Nat/Binary/Minus.agda                                                        
```                                                                                   
*-distribˡ-∸ :  _*_ DistributesOverˡ _∸_                                              
*-distribˡ-∸ x y z = begin-equality                                                   
  x * (y ∸ z)                         ≡⟨ sym (fromℕ-toℕ (x * (y ∸ z))) ⟩              
```                                                                                   
> No need to introduce the fromℕ here. Instead start the proof with                   
> ``toℕ-injective``. See proofs in Algebra.Morphism.MagmaMonomorphism                 
> for examples.                                                                       
                                                                                      
Thank you. I replace the code with                                                    
```                                                                                   
... = toℕ-injective $ begin                                                           
  toℕ (x * (y ∸ z))    ≡⟨ toℕ-homo-* x (y ∸ z) ⟩                                      
  k *' (toℕ (y ∸ z))   ≡⟨ cong (k *'_) (toℕ-homo-∸ y z) ⟩                             
  k *' (m ∸' n)                                                                       
  ...                                                                                 
```

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
  where open ≤-Reasoning                                                              
                                                                                      
∸-mono-≤ : _∸_ Preserves₂ _≤_ ⟶ _≥_ ⟶ _≤_                                             
∸-mono-≤ {x} {x'} {y} {y'} x≤x' y'≤y = begin                                          
```                                                                                   
> See comment on ``*-distribˡ-∸``  below, no need for ``fromℕ``                       
                                                                                      
Sorry, I do not guess.                                                                
``*-distribˡ-∸`` is an equality, and there it works ``toℕ-injective``.                
And ``∸-mono-≤`` is about ``≤``.                                                      
Where is the similarity?                                                              
                                                                                       
                                                                    
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                                                                                                    
+-∸-assoc : ∀ x {y o} → o ≤ y → (x + y) ∸ o ≡ x + (y ∸ o)                             
+-∸-assoc x {y} {o} o≤y = begin-equality                                              
  (x + y) ∸ o                  ≡⟨ sym (fromℕ-toℕ ((x + y) ∸ o)) ⟩                     
```                                                                                   
> See comment on *-distribˡ-∸ below, no need for fromℕ                                
                                                                                      
Done, for ``+-∸-assoc, ∸-+-assoc``.

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
[x∸y]+y≡x {x} {y} y≤x = trans (+-comm (x ∸ y) y) (x+[y∸x]≡y y≤x)                      
                                                                                      
toℕ-homo-∸ : ∀ x y → toℕ (x ∸ y) ≡ (toℕ x) ℕ.∸ (toℕ y)                                
toℕ-homo-∸ x y with <-cmp x y                                                         
```                                                                                   
> Splitting on x ≤? y not on <-cmp saves you a case.                                                                                                                        
                                                                                      
Done.                                                                                 
                                                                                      
                                                                                           
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
  x'              ∎                                                                   
  where open ≤-Reasoning                                                              
                                                                                      
[x+y]∸y≡x : ∀ x y → (x + y) ∸ y ≡ x                                                   
```                                    
> You shouldn't need the brackets here?                                                                                                                                    
                                                                                      
Renamed to  ``x+y∸y≡x``.

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                                                                                                              
-- Properties of _∸_ and _+_ and _≤_                                                  
                                                                                     
x≥y⇒x≡[x∸y]+y : ∀ {x y} → x ≥ y → x ≡ (x ∸ y) + y                                     
``                                                                                    
> Usually we have the more complicated expression on the left,  i.e. ``(x ∸ y) + y ≡ x``  
                                                                                      
Replaced with   ``x≥y⇒[x∸y]+y≡x : ∀ {x y} → x ≥ y → (x ∸ y) + y ≡ x``                               
                                                                                      
                                                                                                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
fromℕ-homo-∸ = Magmas∸ℕᵇ-ℕ.IsHomo⇒IsHomo-rev toℕ fromℕ (cong fromℕ)                   
                     (toℕ-fromℕ , fromℕ-toℕ) toℕ-homo-∸                               
                                                                                      
∸-+-assoc : ∀ x y o → (x ∸ y) ∸ o ≡ x ∸ (y + o)                                       
...                                                                                   
+-∸-assoc : ∀ x {y o} → o ≤ y → (x + y) ∸ o ≡ x + (y ∸ o)                             
```                                                                                   
> If you're using x and y for variable names then you should use z instead of o.      
                                                                                      
                                                                                      
Replaced  o -> z.

**Comment by mechvel** [comment]
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
  fromℕ (toℕ (x + y) ℕ.∸ n)    ≡⟨ cong (fromℕ ∘ (ℕ._∸ n)) (toℕ-homo-+ x y) ⟩          
  fromℕ ((k ℕ.+ m) ℕ.∸ n)      ≡⟨ cong fromℕ (ℕₚ.+-∸-assoc k n≤m) ⟩                   
  fromℕ (k ℕ.+ (m ℕ.∸ n))      ≡⟨ fromℕ-homo-+ k (m ℕ.∸ n) ⟩                          
  fromℕ k + fromℕ (m ℕ.∸ n)    ≡⟨ cong (_+ (fromℕ (m ℕ.∸ n))) (fromℕ-toℕ x) ⟩         
``                                                                                    
> No need for brackets around fromℕ (m ℕ.∸ n) in the proof.                           
                                                                                      
Strangely, I do not find this place now.                                              
(?)

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Algebra/Morphism/TwoMagmas.agda                                                   
...                                               
open import Relation.Binary using (_Preserves_⟶_)                                     
import Relation.Binary.EqReasoning as EqR                                             
                                                                                      
> This should be Reasoning.Setoid not EqReasoning                                     
```                                                                                   
                                                                                      
Replaced with ``import Relation.Binary.Setoid as EqR``                                
                                                                                      
                                                                                      
---------------------------------------------------------------------                                                                           
```                                                                                   
src/Data/Nat/Binary/Properties.agda                                                   
@@ -1164,6 +1179,10 @@ x≢0∧y≢0⇒x*y≢0 {x} {_} x≢0 y≢0 xy≡0  with x*y≡0⇒x≡0∨y≡0         
double[x]≡0⇒x≡0 : ∀ {x} → double x ≡ zero → x ≡ zero                                  
double[x]≡0⇒x≡0 {zero} _ = refl                                                       
                                                                                      
x≡0⇒double[x]≡0 : ∀ {x} → x ≡ 0ᵇ → double x ≡ 0ᵇ                                      
x≡0⇒double[x]≡0 =                                                                     
                cong double                                                           
```                                                                                   
> Why the line break here?                                                            
                                                                                      
                                                                                      
Wrote in one line.

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
x≤y⇒x∸y≡0 {1+[2 x ]} {2[1+ y ]} (inj₁ (odd<even x≤y)) with x ≤? y                     
... | yes _  = refl                                                                   
... | no x≰y = contradiction x≤y x≰y                                                  
x≤y⇒x∸y≡0 {1+[2 _ ]} {2[1+ _ ]} (inj₂ ())                                             
```                                                                                   
> This case is not needed.                                                            
                                                                                      
                                                                                      
Removed.

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Data/Nat/Binary/Minus.agda                                                        
x≤y⇒x∸y≡0 {2[1+ x ]} {1+[2 y ]} (inj₁ (even<odd x<y)) with x <? y                     
... | yes _  = refl                                                                   
... | no x≮y = contradiction x<y x≮y                                                  
x≤y⇒x∸y≡0 {1+[2 x ]} {2[1+ y ]} (inj₁ (odd<even x≤y)) with x ≤? y                     
```                                                                                   
                                                                                      
> This is just ``odd∸even-for>`` is it not?                                           
                                                                                      
                                                                                      
The further proof has only two lines.   And using ``odd∸even-for>`` will lead to a proof like                                 
```                                                                                   
  begin  it                                      =[ odd∸even-for> (odd<even x≤y)                             
            pred (double (x ∸ y))       =[ cong (pred ∘ double) (x≤y⇒x∸y≡0 x≤y)                     
           0ᵇ                                                                                
  ∎                                                                                   
```                                                                                   
It will hardly look simpler.

mechvel:
MatthewDaggitt on 29 Dec 2019                                                                                                                                               
src/Algebra/Morphism/TwoMagmas.agda                                                   
```                                                                                   
IsHomo f =  Homomorphic₂ f _∙_ _∙'_                                                   
                                                                                      
IsHomo-rev :  Pred (B → A) _                                                          
IsHomo-rev f =  Homomorphic₂' f _∙'_ _∙_                                              
```                                                                                   
> I don't think adding these new names just for this file is worth it...              
                                                                                      
                                                                                      
Removed.                                                                              
                                                                                     
------------------------------------------------------------------                    
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Algebra/Morphism/TwoMagmas.agda                                                   
...
IsHomo⇒IsHomo-rev : (f : A → B)              → (g : B → A)  →                         
``                                                                                    
> f and g can be implicit                                                             
                                                                                      
                                                                                      
Done.

**Comment by mechvel** [comment]
_Summary on Algebra/Morphism/TwoMagmas.agda._                                           
                                                                                      
We have finally agreed in the end of January that  TwoMagmas.agda  is needed,                    
with its                                                                              
```                                                                                   
IsHomo⇒IsHomo-rev : {f : A → B} → {g : B → A} →                                       
                    (g Preserves _≈'_ ⟶ _≈_) → Inverseᵇ _≈_ _≈'_ f g →                
                    IsHomo f → IsHomo-rev g                                           
```                                                                                   
with its ``Function.Definitions.Inverseᵇ`` which does not require too many congruences.

**Comment by mechvel** [comment]
MatthewDaggitt on 25 Jan                                                              
```
src/Algebra/Morphism/TwoMagmas.agda                                                   
  g (y ∙' y')       ≈⟨ gCong (∙'cong y=fx y'=fx') ⟩                                   
  g (f x ∙' f x')   ≈⟨ gCong (≈'sym (fHomo x x')) ⟩                                   
  g (f (x ∙ x'))    ≈⟨ g∘f=id (x ∙ x') ⟩                                              
  x ∙ x'            ≈⟨ ≈refl ⟩                                                        
 ```                                                                                     
> This should be ≡⟨⟩.                                                                 
                                                                                      
Done.                                                                                 
                                                                                      
                                                                                      
> I'd also recommending expanding out everything. I think the reassignments           
> to x just confuse things... You can then just use x and y instead of y and y'       
> which are much easier to read.                                                      
                                                                                      
I rewrite it to                                                                       
```                                                                                   
IsHomo⇒IsHomo-rev {f} {g} gCong g∘f=id f∘g=id fHomo b b' = begin                      
  g (b ∙' b')       ≈⟨ gCong (∙'cong b=fa b'=fa') ⟩                                   
  g (f a ∙' f a')   ≈⟨ gCong (≈'sym (fHomo a a')) ⟩                                   
  ...                                                                                 
  g b ∙ g b'        ∎                                                                 
  where                                                                               
  a    =  g b;                a'     =  g b'                                          
  b=fa =  ≈'sym (f∘g=id b);   b'=fa' =  ≈'sym (f∘g=id b')                             
```                                                                                   
                                                                                      
we have now  ``f : A → B,  b b' : B,  a a' : A``,                                     
and I hope this way it is more clear.

mechvel:
MatthewDaggitt on 25 Jan                                                              
```                                                                                   
src/Algebra/Morphism/TwoMagmas.agda                                                   
IsHomo⇒IsHomo-rev : (f : A → B)              → (g : B → A)  →                         
                    (g Preserves _≈'_ ⟶ _≈_) → Inverseᵇ f g →                         
                    Homomorphic₂ f _∙_ _∙'_  → Homomorphic₂' g _∙'_ _∙_               
IsHomo⇒IsHomo-rev f g gCong (f∘g=id , g∘f=id) fHomo y y' = begin                      
```                                                                                   
> It would be more conventional to call it cong or g-cong rather than gCong.          
                                                                                      
                                                                                      
Replaced with ``g-cong``.

src/Data/Nat/Binary/Properties.agda
```diff
@@ -1061,7 +1075,7 @@ private
   }
 
 *-1-monoid : Monoid 0ℓ 0ℓ
-*-1-monoid = record
+*-1-monoid   = record
```
MatthewDaggitt, line 1078: These equal signs should not be aligned. Same with `*-+-semiring` below.

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+-- Subtraction on Bin and some of its property proofs.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Nat.Binary.Subtraction where
+
+open import Algebra using (Op₂; Magma)
+open import Algebra.Consequences.Propositional using (comm+distrˡ⇒distrʳ)
+open import Data.Empty using (⊥)
```
MatthewDaggitt, line 13: This import is not needed.

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Subtraction on Bin and some of its property proofs.
```
MatthewDaggitt, line 4: "and some of its properties"

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+open import Relation.Binary using (Tri; tri<; tri≈; tri>; _Preserves_⟶_;
+                                                          _Preserves₂_⟶_⟶_)
+open import Relation.Binary.PropositionalEquality
+open import Relation.Nullary using (Dec; yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Algebra.Definitions {A = ℕᵇ} _≡_
+open import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup
+  using (xy∙z≈y∙xz; x∙yz≈y∙xz)
+open import Algebra.Solver.CommutativeMonoid +-0-commutativeMonoid
+
```
MatthewDaggitt, line 33: Could you add a section heading named "Definition" here?

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+x<y⇒0<y∸x {x} {y} x<y with <-cmp (y ∸ x) 0ᵇ
+... | tri> _ _     gt = gt
+... | tri≈ _ y∸x≡0 _  = contradiction (x∸y≡0⇒x≤y y∸x≡0) (<⇒≱ x<y)
+... | tri< y∸x<0 _ _  = contradiction y∸x<0 x≮0
+
+---------------------------------------------------------------
+-- Properties of _∸_ and _+_ and _≤_
+
+x≥y⇒[x∸y]+y≡x : ∀ {x y} → x ≥ y → (x ∸ y) + y ≡ x
+x≥y⇒[x∸y]+y≡x {zero} {y} y≤0 = begin-equality
```
MatthewDaggitt, line 119: Pattern match on `y≤0` and this whole thing becomes `refl`

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+    (double x∸y) + (double (suc y))  ≡⟨ sym (double-distrib-+ x∸y (suc y)) ⟩
+    double (x∸y + suc y)             ≡⟨ cong double (+-comm x∸y (suc y)) ⟩
+    double (suc y + x∸y)             ≡⟨ cong double (suc-+ y x∸y) ⟩
+    double (suc (y + x∸y))           ≡⟨ cong (double ∘ suc) (+-comm y x∸y) ⟩
+    double (suc (x∸y + y))           ≡⟨ cong (double ∘ suc) $
+                                        x≥y⇒[x∸y]+y≡x (inj₁ y<x) ⟩
+    double (suc x)                   ≡⟨ sym (2[1+_]-double-suc x) ⟩
+    2[1+ x ]                         ∎
+  where open ≤-Reasoning;  x∸y = x ∸ y;  x'∸y' = x' ∸ y'
+x≥y⇒[x∸y]+y≡x {z@(2[1+ x ])} {2[1+ x ]} (inj₂ refl) = begin-equality
```
MatthewDaggitt, line 137: This subcase can be replaced with `cong (_+ z) (x∸x≡0 z)`

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+
+toℕ-homo-∸ : ∀ x y → toℕ (x ∸ y) ≡ (toℕ x) ℕ.∸ (toℕ y)
+toℕ-homo-∸ x y with x ≤? y
+... | yes x≤y = begin
+  toℕ (x ∸ y)   ≡⟨ cong toℕ (x≤y⇒x∸y≡0 x≤y) ⟩
+  toℕ 0ᵇ        ≡⟨⟩
+  0             ≡⟨ sym (ℕₚ.m≤n⇒m∸n≡0 m≤n) ⟩
+  m ℕ.∸ n       ∎
+  where
+  open ≡-Reasoning;  m = toℕ x;  n = toℕ y;  m≤n = toℕ-mono-≤ x≤y
```
MatthewDaggitt, line 229: I'm not sure the `m` and `n` aliases actually help here. I'd just write `toℕ x` and `toℕ y` and `toℕ-mono-≤ x≤y` explicitly

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+  toℕ ((x + y) ∸ z)      ≡⟨ toℕ-homo-∸ (x + y) z ⟩
+  (toℕ (x + y)) ℕ.∸ n   ≡⟨ cong (ℕ._∸ n) (toℕ-homo-+ x y) ⟩
+  (k ℕ.+ m) ℕ.∸ n       ≡⟨ ℕₚ.+-∸-assoc k n≤m ⟩
+  k ℕ.+ (m ℕ.∸ n)       ≡⟨ cong (k ℕ.+_) (sym (toℕ-homo-∸ y z)) ⟩
+  k ℕ.+ toℕ (y ∸ z)     ≡⟨ sym (toℕ-homo-+ x (y ∸ z)) ⟩
+  toℕ (x + (y ∸ z))      ∎
+  where
+  open ≡-Reasoning;  k = toℕ x;  m = toℕ y;  n = toℕ z;  n≤m = toℕ-mono-≤ z≤y
+
+x+y∸x+z : ∀ x y z → (x + y) ∸ (x + z) ≡ y ∸ z
```
MatthewDaggitt, line 278: This should be named `x+y∸x+z≡y∸z`

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+    (m ℕ.∸ n) ℕ.+ n      ∎
+
+∸-magma : Magma 0ℓ 0ℓ
+∸-magma = magma _∸_
+
+import Algebra.Morphism.TwoMagmas ∸-magma ℕₚ.∸-magma
+  as Magmas∸ℕᵇ-ℕ
+
+fromℕ-homo-∸ : ∀ m n → fromℕ (m ℕ.∸ n) ≡ (fromℕ m) ∸ (fromℕ n)
+fromℕ-homo-∸ = Magmas∸ℕᵇ-ℕ.IsHomo⇒IsHomo-rev {toℕ} {fromℕ} (cong fromℕ)
```
MatthewDaggitt, line 254: Can the implicit arguments not be inferred here?

src/Data/Nat/Binary/Subtraction.agda
```diff
@@ -0,0 +1,344 @@
...
+    toℕ ((x ∸ y) + y)    ≡⟨ cong toℕ (+-comm (x ∸ y) y) ⟩
+    toℕ (y + (x ∸ y))    ≡⟨ cong toℕ (x+[y∸x]≡y y≤x) ⟩
+    m                    ≡⟨ sym (ℕₚ.m+[n∸m]≡n n≤m) ⟩
+    n ℕ.+ (m ℕ.∸ n)      ≡⟨ ℕₚ.+-comm n (m ℕ.∸ n) ⟩
+    (m ℕ.∸ n) ℕ.+ n      ∎
+
+∸-magma : Magma 0ℓ 0ℓ
+∸-magma = magma _∸_
+
+import Algebra.Morphism.TwoMagmas ∸-magma ℕₚ.∸-magma
```
MatthewDaggitt, line 250: Import statements should go at the top of the file.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,52 @@
...
+import Algebra.Morphism.Definitions as Definitions
+open import Data.Product using (_,_)
+open import Function using (id; _∘_)
+open import Function.Definitions using (Inverseᵇ)
+open import Relation.Binary using (_Preserves_⟶_)
+import Relation.Binary.Reasoning.Setoid as EqR
+open import Relation.Unary using (Pred)
+
+
+module Algebra.Morphism.TwoMagmas {α α= β β=} (M : Magma α α=) (M' : Magma β β=)
```
MatthewDaggitt, line 20: Please could you rename this module to `Algebra.Morphism.Consequences`

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,52 @@
...
+
+open Definitions A B _≈'_ using (Homomorphic₂)
+open Definitions B A _≈_  using () renaming (Homomorphic₂ to Homomorphic₂')
+open EqR setoid
+
+IsHomo :  Pred (A → B) _
+IsHomo f =  Homomorphic₂ f _∙_ _∙'_
+
+IsHomo-rev :  Pred (B → A) _
+IsHomo-rev f =  Homomorphic₂' f _∙'_ _∙_
```
MatthewDaggitt, line 34: I don't think these aliases add anything. Please use the full expression.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,52 @@
...
+
+
+open import Algebra using (Magma)
+import Algebra.Morphism.Definitions as Definitions
+open import Data.Product using (_,_)
+open import Function using (id; _∘_)
+open import Function.Definitions using (Inverseᵇ)
+open import Relation.Binary using (_Preserves_⟶_)
+import Relation.Binary.Reasoning.Setoid as EqR
+open import Relation.Unary using (Pred)
```
MatthewDaggitt, line 17: All non-necessary imports should go inside the module.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,52 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Some properties of Magma homomorphisms
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+
```
MatthewDaggitt, line 9: Please remove the extra line.

src/Algebra/Morphism/TwoMagmas.agda
```diff
@@ -0,0 +1,52 @@
...
+                    IsHomo f                 → IsHomo-rev g
+
+IsHomo⇒IsHomo-rev {f} {g} g-cong ( f∘g=id , g∘f=id) fHomo b b' = begin
+  g (b ∙' b')       ≈⟨ g-cong (∙'cong b=fa b'=fa') ⟩
+  g (f a ∙' f a')   ≈⟨ g-cong (≈'sym (fHomo a a')) ⟩
+  g (f (a ∙ a'))    ≈⟨ g∘f=id (a ∙ a') ⟩
+  a ∙ a'            ≡⟨⟩
+  g b ∙ g b'        ∎
+  where
+  a    =  g b;                a'     =  g b'
```
MatthewDaggitt, line 51: Please use `g b` and `g b'` rather than `a`. The proof will become longer but much more readable.

MatthewDaggitt:
> And it reports
"The pattern variable true has the same name as the constructor
Agda.Builtin.Bool.Bool.true
when checking the clause Data.Nat.Binary.Minus.with-234 x y true = 0ᵇ"

Are you sure you had `true` and `false` imported?

mechvel:
Please, consider my commit of April 11, 2020.
All your notes were taken in account.
I start now to fill in my responses to these notes.

mechvel:
src/Data/Nat/Binary/Minus.agda    
@gallais:                                                                                   
    Same comments as earlier: avoid yes / no whenever you do not need the proofs            
    and use does instead. This will give you more efficient functions                       
    (essentially computing a boolean).                                                                                                                                                                         
```                                                                                         
2[1+ x ] ∸ 1+[2 y ] with does (x <? y)                                                      
... | true  = 0ᵇ                                                                            
... | false = 1+[2 (x ∸ y) ]                                                                
``                                                                                          
And it reports    ....
                        
> Are you sure you had `true` and `false` imported?
                                                                                                                                 
Indeed, this time it works.                                                                 
I do not recall now of what was the mistake                                                  
(for any occasion: ``does`` can be imported from two places, and also one un-existing place is reported).

File: CHANGELOG.md [hunk 16]
```diff
@@ -18,11 +22,36 @@ Deprecated modules
...
+	
 Other major changes
 -------------------
 
+
+	
 Other minor additions
 ---------------------
+
+The following functions are added to Data.Nat.Properties:
```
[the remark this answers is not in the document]
mechvel: Done

src/Data/Nat/Binary/Properties.agda
```diff
@@ -728,6 +739,9 @@ private
   { isCommutativeMonoid = +-0-isCommutativeMonoid
   }
 
+import Algebra.Properties.CommutativeSemigroup +-commutativeSemigroup
```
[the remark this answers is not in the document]
mechvel: I add                                                                                       
```                                                                                         
import Algebra.Properties.CommutativeSemigroup as CommSemigProp                             
```                                                                                         
to the head decls in the module, and replace the second import with                         
```                                                                                         
module Bin+CSemigroup = CommSemigProp +-commutativeSemigroup                                
```                                                                                         
Will this do?

src/Data/Nat/Binary/Properties.agda
mechvel: Fixed.

src/Data/Nat/Binary/Subtraction.agda
mechvel: Removed.

src/Data/Nat/Binary/Subtraction.agda
mechvel: Done

src/Data/Nat/Binary/Subtraction.agda
mechvel: Done

src/Data/Nat/Binary/Subtraction.agda
mechvel: This sentence is replaced with                                                              
```                                                                                         
x≥y⇒[x∸y]+y≡x {zero} {_} (inj₂ refl) = refl                                                 
```

src/Data/Nat/Binary/Subtraction.agda
mechvel: Done

src/Data/Nat/Binary/Subtraction.agda
mechvel: Done as you recommend.

src/Data/Nat/Binary/Subtraction.agda
mechvel: Done

src/Data/Nat/Binary/Subtraction.agda
mechvel: I remove ``{fromℕ}`` from this line.                                                        
Further removing ``{toℕ}`` leads to "unsolved metas".

src/Data/Nat/Binary/Subtraction.agda
mechvel: I set ``import Algebra.Morphism.Consequences`` to the head imports,                         
and change the subject line to                                                              
```                                                                                         
module Magmas∸ℕᵇ-ℕ = Algebra.Morphism.Consequences ∸-magma ℕₚ.∸-magma                       
```                                                                                         
(assuming that you require the "Consequences" renaming below).

src/Algebra/Morphism/TwoMagmas.agda
mechvel: Done.
But the name ``Morphism.TwoMagmas`` looks much more informative than                        
``Morphism.Consequences``. The reader would not guess what kind of                          
consequences this module keeps in it.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: I set the full expressions.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: I change this head to                                                                       
```                                                                                         
open import Algebra using (Magma)                                                           
                                                                                            
module Algebra.Morphism.Consequences {α α= β β=} (M : Magma α α=) (M' : Magma β β=)         
  where                                                                                     
import Algebra.Morphism.Definitions as Definitions                                          
open import Data.Product using (_,_)                                                        
...                                                                                                                                   
import Relation.Binary.Reasoning.Setoid as EqR                                              
open Magma M using (_≈_; _∙_; setoid) renaming (Carrier to A; refl to ≈refl)                
open Magma M' using () renaming ...                                                         
open Definitions A B _≈'_ using (Homomorphic₂)                                              
open Definitions B A _≈_  using () renaming (Homomorphic₂ to Homomorphic₂')                 
open EqR setoid                                                                             
...                                                                           
```                                                                                                                                            
But you see, this is not natural: setting inside a parametric module an import              
that does not depend on the parameters of this module. This gives the reader a              
wrong impression of the dependences.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: Done

src/Algebra/Morphism/TwoMagmas.agda
mechvel: Done

src/Algebra/Morphism/TwoMagmas.agda
MatthewDaggitt: This is the standard name for such modules in the library, e.g. `Relation.Binary.Conquences`, `Relation.Algebra.Consequences`. Better to be consistent.

src/Algebra/Morphism/TwoMagmas.agda
MatthewDaggitt: > But you see, this is not natural: setting inside a parametric module an import
that does not depend on the parameters of this module.

I agree with you on this one, but if we want to change this we should change it consistently across the library. Feel free to open an issue.

src/Algebra/Morphism/TwoMagmas.agda
mechvel: 1) My last commit does not contain  ... TwoMagmas. 
2) Instead it has Algebra.Morphism.Consequences, and this was your suggestion.
3) And now you suggest Relation.Algebra.Consequences. 
There is kind of a confusion.
Need I now to move Algebra.Morphism.Consequences to Relation.Algebra.Consequences ?

MatthewDaggitt: No I mistyped. I was trying to give examples: `Relation.Binary.Conquences` and `Algebra.Consequences`. The file should stay named `Algebra.Morphism.Consequences`.

MatthewDaggitt:
I have added some changes @mechvel. In particular I've moved `toℕ-homo-∸` much earlier up the file by removing the dependency on the other proofs. This has allowed me to shorten the file by ~100 lines by exploiting the homomorphism proof, which you were already doing at the bottom of the file.

I've also renamed a couple of proofs to make it more consistent with the names in `Data.Nat.Properties` and removed two proofs which were in there twice under different names.

Have a look and let me know if I've missed anything. Otherwise I'm happy to merge it in now.

mechvel:
Yes, please, merge it.
But I think that things for Bin need to be evaluated without conversion to ``ℕ``
(``toℕ-homo-∸ and such``). The very reason for introducing Bin is to avoid computation in unary system. This effects even proofs, when there happens evaluation with proofs.
So that if anything is already implemented for Bin without using conversion to ``ℕ``, we do not need to re-design it by using this conversion.
We shall see this further whether your re-design effects the performance of divMod and gcd.
For any occasion I save my version of Subtraction.agda 

Another point. In  Nat.Binary.Base you write in comments
"
This module contains an alternative formulation of ℕ that is still reasonably computationally efficient without having to call out to Haskell.
"
I suggest
"This module contains an alternative formulation of ℕ implemented in Agda (without using 
built-in functions) that is still reasonably computationally efficient."

Because if Agda was done without using Haskell, then there may be built-ins in another language.

mechvel:
The only reason for introducing Bin is to avoid computation in unary system. 
Is not it?

MatthewDaggitt:
> Because if Agda was done without using Haskell, then there may be built-ins in another language.

Fair enough. I've changed it to remove the mention of Haskell.

> But I think that things for Bin need to be evaluated without conversion to ℕ

We've discussed this before (though I know I didn't manage to convince you), it's definitely possible to have our cake and eat it with both simple yet inefficient proofs and efficient functions that don't evaluate the proofs. If it turns out to be a problem for you, we can discuss it again. As you're happy for the moment, I'll go ahead and merge it in.

PR #1129, opened 2020-03-27
add symmetric reflexive closures for relations

HuStmpHrrr:


src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
```diff
@@ -0,0 +1,121 @@
...
+
+private
+  variable
+    a ℓ ℓ′ : Level
+    A B    : Set a
+
+module _ {A : Set a} (_≤_ : Rel A ℓ) where
+  private
+    variable
+      x y z : A
```
MatthewDaggitt: Do bad things happen when you lift these to the top-level `variable` declaration? You could get rid of some of the arguments to the equational reasoning combinators?

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+    setoid : Setoid _ _
+    setoid = record
+      { Carrier       = A
+      ; _≈_           = Plus⇔ _∼_
+      ; isEquivalence = isEquivalence
+      }
+
+  module _ {c e} (S : Setoid c e) where
```
MatthewDaggitt: Can these levels be included in the `variable` declaration?

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
```diff
@@ -0,0 +1,121 @@
...
+    private
+      module S = Setoid S
+
+    minimal : (f : A → Setoid.Carrier S) →
+              _∼_ =[ f ]⇒ Setoid._≈_ S →
+              Plus⇔ _∼_ =[ f ]⇒ Setoid._≈_ S
```
MatthewDaggitt: What's the intuition behind the name `minimal`?

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
```diff
@@ -0,0 +1,121 @@
...
+  syntax forth⁺-syntax x x≤y y⇔z = x ⇒⟨ x≤y ⟩ y⇔z
+
+  back⁺-syntax : ∀ x {y z} → y ≤ x → Plus⇔ _≤_ y z → Plus⇔ _≤_ x z
+  back⁺-syntax _ = back⁺
+
+  syntax back⁺-syntax x y≤x y⇔z = x ⇐⟨ y≤x ⟩ y⇔z
+
+module _ {_≤_ : Rel A ℓ} {_≼_ : Rel B ℓ′} (f : A → B) where
+
+  module _ (inj : _≤_ =[ f ]⇒ _≼_) where
```
MatthewDaggitt: Double anonymous module can be simplified to one here?

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
gallais: Have we lifted the restriction on using `variable` only for levels and Sets?

MatthewDaggitt: Ah no, we haven't! Good point :laughing: Yup, these should be removed!

HuStmpHrrr: so what's the recommendation here?

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
HuStmpHrrr: sure

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
MatthewDaggitt: To remove the `x`,`y`,`z` variables entirely and add them manually :+1:

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
HuStmpHrrr: if `f` converts `_∼_` to an equivalence, it can also convert the symmetric transitive closure to the equivalence. `minimal` might not be the best name for this function.

src/Relation/Binary/Construct/Closure/SymmetricTransitive.agda
HuStmpHrrr: yes, sure

PR #1139, opened 2020-03-31
≗-Reasoning - a simple and elegant way to reason about function equality

MirceaS:
Here is a first draft of a little thing that I put together after being frustrated with having to work with long chains of functions composition'd together. Because of how function composition works in Agda, I believe this has the potential to make some proofs a lot more clear, with a minimal amount of machinery that has to be defined.
As noted before, this is just a Proof of Concept and I would be very happy for people to change this or provide me with any feedback.
Having tried to use this "in the wild" I found that this Reasoning system fails to infer implicit terms a lot more often than if I just used ≡-Reasoning but that may be fixable.
I also made an extension for Sum functions that should come in handy for certain situations where pattern matching is not so easy. (See the examples in `README/Function/Reasoning.agda`).
Thanks for taking the time to look at this!

### DIFF
--- a/README/Function/Reasoning.agda
+++ b/README/Function/Reasoning.agda
@@ -66,3 +66,51 @@ _ = refl
 
 _ : subpalindromes "elle-meme" ≡ "ll" ∷ "elle" ∷ "mem" ∷ "eme" ∷ []
 _ = refl
+
+
+
+-- Example of using ≗-Reasoning
+module _ where
+
+  open import Data.Sum
+  open import Data.Sum.Properties
+  open import Relation.Binary.PropositionalEquality using (_≡_; _≗_)
+
+  variable
+    A B C D E F : Set
+
+  -- Use scenario:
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
This is pretty neat. Just wondering if the `SumReasoning` should be merged into the main reasoning? It seems a little odd to be `SumProperties` as really it's reasoning about functions over sums, not sums themselves. Is there any reason not to merge them?

MatthewDaggitt:
This is pretty neat. Just wondering if the `SumReasoning` should be merged into the main reasoning? It seems a little odd to be `SumProperties` as really it's reasoning about functions over sums, not sums themselves. Is there any reason not to merge them? Oh and a Changelog entry would be great, thanks!

MirceaS:
> This is pretty neat. Just wondering if the `SumReasoning` should be merged into the main reasoning? It seems a little odd to be `SumProperties` as really it's reasoning about functions over sums, not sums themselves. Is there any reason not to merge them? Oh and a Changelog entry would be great, thanks!

That's a good point. As I mentioned in my original post, this is just a messy proof of concept. I'm a bit busy these days but I'll get back to refactoring this once I have some time if people are interested in it.

MatthewDaggitt:
Going to close this as nothing has been added to it in over a year.

PR #1144, opened 2020-04-02
Indexed binary trees

alexarice:
Adds whatever the equivalent of vectors is for binary trees.
I am not sure what they are actually called so I can fix that.

Will put in a changelog entry if this is a wanted feature


gallais:
Do you have a use case for it? You could write a `README.Data.Tree.Binary.Indexed`
explaining to users why one may want to use such a family.

alexarice:
In the process of writing another pr which uses this, it should perhaps all be in one pr

alexarice:
I think I'm going to close this until the whole thing is ready

gallais:
Alternatively we have the not-ready-yet label.

PR #1145, opened 2020-04-03
finish deprecating MonadSyntax

alexarice:
`Everything.agda` will not build at the moment.
The full stop is necessary for `generateEverything.hs` to realise the module is deprecated


MatthewDaggitt:
Eek, not sure how that one slipped through. I though Travis was going green. Thanks for the fix!

gallais:
> I though Travis was going green.

It was. Raising a user warning is not a fatal error. We could turn `-Werror` on
to force Agda to fail if any warning is raised. I'll open a PR to see whether it
breaks the current master :)

PR #1148, opened 2020-04-03
Add stack file and CI test for it

vlopezj:
Added stack-8.4.4.yaml file to compile the .hs scripts, and a CI job to test the stack.yaml file.
I didn't know where or if to add this to the Changelog, as it does not change anything about the library itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,148 +6,6 @@ branches:
 
 dist: xenial
 
-cache:
-  directories:
-    - $HOME/.cabsnap
-
-before_install:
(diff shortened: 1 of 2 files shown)


gallais:
Why is there a massive unrelated change to `.travis.yml`?

vlopezj:
@gallais The original `.travis.yml` had all the steps for the (single) CI job at the top level. If I want to add a second CI job while leaving the first job at the top level (as in commit 5d1e70f), I have to make sure that all the steps from the first job are overiden by the second job, which is fragile.

It's essentially a change in indentation, but the diff tool doesn't seem clever enough to realise it.

PR #1156, opened 2020-04-09
Strict inverses

alexarice:
Adds versions of various inverses that compose in a strictly unital and associative way.
i.e. Given `{A : Setoid} (f g h : Inverse A A)` we have
```agda2
(f ∘ g) ∘ h = f ∘ (g ∘ h)
```
and
```agda2
f ∘ id = f = id ∘ f
```
by definition.


MatthewDaggitt:
@alexarice I have to make an apology here. The `Function.Inverse`/`Bijection` etc, modules are actually tacitly deprecated in favour of `Function.Structure`/`Bundle` modules (see issue #759). 

I'm __really__ sorry, I should have put a message in the files. I hope you understand that we don't want to add further work on top of the old structures

alexarice:
That is understandable, would the content still be wanted if it were moved to the correct place (note this was far less work than it looks as I mainly copied existing files and changed their imports)

MatthewDaggitt:
>  (note this was far less work than it looks as I mainly copied existing files and changed their imports)

That's good to hear.

> That is understandable, would the content still be wanted if it were moved to the correct place

It's not so much that it's in the wrong place. It's that the design of the function hierarchy has been standardised to match the other hierarchies in the library.

I'm unsure whether the same can be done with the new function hierarchy and what the changes necessary would be? If the changes were minimal, then I might be tempted to try and alter the main definitions. If not minimal then I guess we'd have to see what they looked like.

alexarice:
There are two differences with the strict version.
Firstly the definition of `f LeftInverseOf g` is changed from 
```
\forall x -> f (g x) = x
```
to
```
\forall x y -> y = g x -> f y = x
```
The second change is to change composition so that it uses the new definition.

alexarice:
The issue with having this change the main definitions is that I expect other parts of the library use this a lot. It would probably be worth making helpers that allow you to use them in the old way.
The alternative would be to add `StrictInverse` Types to the hierachy alongside the originals

alexarice:
I worked out everything that would need to be changed and it is virtually nothing as nothing seems to rely on the new function infrastructure yet. The downside is that the definitions do not quite align with how people usually state them. I think a good compromise could be to maintain two sets of definitions, have the bundles use the new versions, and have builders that build the bundles from the old versions.

any thoughts?

MatthewDaggitt:
Sorry it's taken me so long to reply! 

I really like the new definitions! The fact that the existing proofs are greatly simplified using them is a huge plus in their favour.

> I think a good compromise could be to maintain two sets of definitions, have the bundles use the new versions

I agree. We could append `\prime` to the old ones, and have the bundles both the primed and the non-primed versions for ease of use. A comment explaining the reason behind the two different definitions would be great to add in `Function.Definitions` as well.

JacquesCarette:
Two sets of definitions seems like a good idea to me as well. The simplifications in the proofs are impressive. I would slightly prefer having these defined in sub-modules so that I can be more precise about which I grab [if I want to].  And I'm not a big fan of differentiating things with prime, as one never really knows which is the primed and which is the unprimed version.  I'm partial to using the word 'Strict' here, but that's a mild opinion.

alexarice:
> 
> 
> Two sets of definitions seems like a good idea to me as well. The simplifications in the proofs are impressive. I would slightly prefer having these defined in sub-modules so that I can be more precise about which I grab [if I want to]. And I'm not a big fan of differentiating things with prime, as one never really knows which is the primed and which is the unprimed version. I'm partial to using the word 'Strict' here, but that's a mild opinion.

Do you have an alternative name to strict? I was using the terminology from strict vs weak (monoidal) categories

JacquesCarette:
I've thought about it some - and no, no better name, `Strict` seems good.

The problem is that `Strict` is overloaded. For example, in category theory, it seems that "having equality on objects" and "natural isos are actually equalities" are both described as 'strict'. That's where my reluctance comes from.

alexarice:
> The problem is that `Strict` is overloaded. For example, in category theory, it seems that "having equality on objects" and "natural isos are actually equalities" are both described as 'strict'. That's where my reluctance comes from.

I'd argue that this is a case of natural isos are actually equalities

JacquesCarette:
That's a fair point. I guess `Strict` it is!

MatthewDaggitt:
Okay, so hypothetically if we did call the new definitions `StrictInverse` what would we call the new definition of `Surjective`? `StrictSurjective`? `StrictlySurjective`?

alexarice:
I might have been confused but I thought the idea was to make a `Function.Definition.Strict` that mirrors `Function.Definition` but with the strict versions (and similar for structures and bundles)

MatthewDaggitt:
Ah I'd missed that. Hmm just trying to think through the consequences. We want to avoid duplication as much as possible, i.e. we don't want to end up duplicating all the `Construct` modules. We would probably want to commit to one most of the time.

I kind of imagined us using the strict versions everywhere and then just have the records exporting the non-strict versions for users. We could then have biased constructors for the non-strict versions of the records like we have `Algebra.Structures.Biased`...

MatthewDaggitt:
Okay finally returning to this now. Going to try and bring it up to date.

jamesmckinna:
I would welcome this PR being (eventually) merged, cf. PR #1914

MatthewDaggitt:
Okay, I've finally pushed this all the way through :tada: 

I'd like to get people's feedback on the `Strictly` prefix for the new names for the old definitions. It was the best I could come up with, but not super happy with it. If anyone has any better ideas for what it could be, please do suggest them. Any ideas @JacquesCarette , @jamesmckinna , @Taneb etc.? Otherwise, I'll aim to finally merge this in sometime next week. 

Apologies for the 3 year delay @alexarice :smile:

Taneb:
I can't think of a better prefix than `Strictly`. It does highlight the minor problem that some of these bundles have an adjective as their name (e.g. `Surjective`) and some a noun (e.g. `Inverse`) which makes it hard to have a grammatically sound prefix across swathes of them

JacquesCarette:
My imagination fails be too: `Strictly` seems to be what seems to stick.

jamesmckinna:
Sorry to have returned to the party too late! 
Congrats to @alexarice and to @MatthewDaggitt for the (ongoing) efforts with the `Function` hierarchy re-engineering effort. Good to see this particular PR finally closed!

jamesmckinna:
Regarding `Strictly`, I think this is a reasonable compromise choice of name. In the interests of (eventually) closing #759 and friends, suggest that we stick with this for now (and at least until the v2.0 milestone ;-)) and having drawn sufficient breath, open a *long* `label:naming` issue ... not least because we are going to have accumulated a great many deprecations, and with them, associated suppressed deprecation warnings, which we may well wish gradually to prune away as we move forward...

PR #1161, opened 2020-04-14
Fixed function names in comments

Isti115:
The comments for fromℕ< and fromℕ<″ still had ≤ in their names remaining from an older version.


MatthewDaggitt:
Thanks for the contribution!

PR #1163, opened 2020-04-17
[ new ] wordsBy & words

gallais:
Following Philippe de Rochambeau's request on the mailing list.


src/Data/String/Base.agda
```diff
@@ -90,6 +92,24 @@ intersperse sep = concat ∘′ (List.intersperse sep)
 unlines : List String → String
 unlines = intersperse "\n"
 
+module _ {p} {P : Pred Char p} (P? : Decidable P) where
+
+  wordsBy : String → List String
+  wordsBy = go [] ∘ toList where
```
MatthewDaggitt, line 98: This can be fairly easily generalised to arbitrary types rather than just `Char` right? Might be worth putting in `Data.List.Base`?

src/Data/List/Base.agda
```diff
@@ -369,6 +369,22 @@ initLast (x ∷ xs)         with initLast xs
 ... | []       = [] ∷ʳ' x
 ... | ys ∷ʳ' y = (x ∷ ys) ∷ʳ' y
 
+------------------------------------------------------------------------
+-- Splitting a list
+
+wordsBy : ∀ {P : Pred A p} → Decidable P → List A → List (List A)
```
MatthewDaggitt, line 375: I was just about to merge this in when it occurred to me that this is usually called `split` rather than `wordsBy` in most languages right?

gallais:
There are a lot of different strategies depending on whether you decide to retain
the separator, whether you are happy to split on a `List A` and not just a single
`A`, etc. The [`split`](https://hackage.haskell.org/package/split-0.2.3.4/docs/Data-List-Split.html) package on hackage implements a lot of these
strategies.

I think `wordsBy` is unambiguous: it transparently is a generalisation of `words`
just like e.g. `sortBy` is a generalisation of `sort` in Haskell.

MatthewDaggitt:
Haha, fair enough :+1:

PR #1164, opened 2020-04-19
Added `pair` and `pairWith` functions for lists

MatthewDaggitt:
Adds functions that generate all possible pairs from two lists. Would welcome feedback on a couple of points:

1. Are people happy with the name `pairWith` and `pair`? It really generates _all_ pairs but I'm struggling to turn that into a verb. I'm a bit worried people might get confused with `zip`. I was also considering either `pairAll(With)` or `combine(With)`...

2. I found it necessary to add a version of `tabulate` over setoids in `Data.List.Relation.Unary.All`. I was at a bit of a loss about what to call this or where to place it. In the end I've gone for `tabulateₛ` but that feels ungainly and inconsistent with the rest of the library. I would definitely like to hear any better suggestions!


gallais:
Maybe `cartesianProduct` for `pair`?

PR #1182, opened 2020-04-30
Add lexicographic ordering on vectors

uzytkownik:


src/Data/Vec/Relation/Binary/Lex/Core.agda
```diff
@@ -0,0 +1,122 @@
...
+
+-- Properties
+
+module _ {a ℓ₁ ℓ₂} {A : Set a} {P : Set} {_≈_ : Rel A ℓ₁} {_≺_ : Rel A ℓ₂} where
+
+  private
+    _≋_ = Pointwise _≈_
+    _<_ = Lex P _≈_ _≺_
+
+  length-equal : ∀ {m n P} {xs : Vec A m} {ys : Vec A n} → Lex P _≈_ _≺_ xs ys → m ≡ n
```
gallais, line 49: If this is provable, why do we have the length polymorphism in the first place?

JacquesCarette: For the same reason that permutations have 2 length parameters too: the proof isn't always `refl` and it's not always definitionally `refl`.  Things compose a whole lot better if you don't conflate two things that are equal because it's a non-trivial property. [At least, that's my guess.]

uzytkownik: I think it was mostly because pointwise did it and I assumed that design choices there were reasonable (but I assume @JacquesCarette is correct).

fredrikNordvallForsberg: Another question is if it is desirable: I was expecting e.g. [0, 1, 2] < [0, 1, 3, 4, 5].

JacquesCarette: That's tricky. It's kind of lifting the domain to always have a (virtual) infimum, even when it doesn't. I'm not saying it's wrong, just requires thought. And, of course, the above doesn't do that. I think you'd at least need to drop the `m≡n` term from `this` to get that behaviour.

MatthewDaggitt: I agree with @fredrikNordvallForsberg that [0,1,2] < [0, 1, 3, 4, 5] should be true. As does our implementation of [lexicographic orderings for lists](https://github.com/agda/agda-stdlib/blob/0f437abd5d4c4e9b35429ce6a3f66ff2c708f373/src/Data/List/Relation/Binary/Lex/Core.agda#L34).

JacquesCarette: Nice!

gallais: We should maybe have both?

* lexicographic for Vec-as-lists
* lexicographic for Vec-as-(homogeneous)-n-tuples

Now to find names...

uzytkownik: My thinking was that if you order Vec-as-lists you can use the list in the ffirst place (or at least bijection to lists).

MatthewDaggitt: > My thinking was that if you order Vec-as-lists you can use the list in the ffirst place (or at least bijection to lists).

That's a fair point :+1: 

> Now to find names...

From reading the [wikipedia page](https://en.wikipedia.org/wiki/Lexicographical_order) on the various versions, I'd suggest `Lex.Cartesian.(Strict/NonStrict)` for this variant where lengths have to be the same, `Lex.Dictionary.(Strict/NonStrict)` for the existing variant under `List` where the shorter sequences can be below in the order and `Lex.Short.(Strict/NonStrict)` for the third variant which apparently exists where shorter sequences are always below in the order.

Thoughts?

Anyway now that I think that we're sure we're keeping this one, I'll go through and add some minor comments on the naming.

src/Data/Vec/Relation/Binary/Lex/NonStrict.agda
```diff
@@ -0,0 +1,196 @@
...
+------------------------------------------------------------------------
+-- Publicly re-export definitions from Core
+
+open import Data.Vec.Relation.Binary.Lex.Core as Core public
+  using (base; this; next; ¬≤-this; ¬≤-next)
+
+----------------------------------------------------------------------
+-- Strict lexicographic ordering.
+
+module _ {a ℓ₁ ℓ₂} {A : Set a} where
```
MatthewDaggitt, line 37: I would make `{a ℓ₁ ℓ₂}` into variables and you can save an indentation level.

uzytkownik: I don't save the indent level without `{A : Set a}` and for moving everything into variable there seems to be distinction between `A.a` and `a`:

```agda
variable
  a ℓ₁ ℓ₂ : Level
  A : Set a
```

MatthewDaggitt: Yup, when you have `A : Set a` as a variable, you'll still need to put the `{A : Set a}` into the definition as follows:
```agda
data Lex {A : Set a} (P : Set) (_≈_ : Rel A ℓ₁) (_≺_ : Rel A ℓ₂) : ∀ {m n} → REL (Vec A m) (Vec A n) (a ⊔ ℓ₁ ⊔ ℓ₂) where
```
but you can remove it elsewhere.

uzytkownik:
@MatthewDaggitt  - I fixed whitespace issues and squashed the commits.

MatthewDaggitt:
Thanks for the PR @uzytkownik :+1:

PR #1190, opened 2020-05-03
Setoid sigma

mechvel:
Please, consider this PR.
Relation.Binary.Construct.Sigma  added, which for the given setoid implements  Setoid on 
``Σ Carrier P``  induced by the projection to Carrier.


MatthewDaggitt:
This doesn't appear to have any new content in so closing.

PR #1191, opened 2020-05-05
[ new ] Division with remainder for binary naturals

mechvel:
Your branch is updated according to «origin/master».

changed:     CHANGELOG.md
new file:    src/Data/Nat/Binary/DivMod.agda

Please, consider this commit of Data.Nat.Binary.DivMod

### DIFF
--- a/src/Algebra/Properties/Semiring.agda
+++ b/src/Algebra/Properties/Semiring.agda
@@ -1,0 +1,32 @@
+------------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Some basic items for arbitrary Semiring
+------------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Algebra
+
+module Algebra.Properties.Semiring {α α≈} (R : Semiring α α≈) where
+
+open import Data.Product using (Σ)
+open import Relation.Binary using (Setoid) renaming (Decidable to Decidable₂)
+open import Relation.Nullary using (yes; no)
+open import Relation.Unary using (Pred; Decidable)
+open Semiring R
+open Setoid setoid using (_≉_)
+
+------------------------------------------------------------------------------
+-- The general notion of a nonzero element
+
+IsNonzero : Pred Carrier α≈
+IsNonzero = (_≉ 0#)
+
+Nonzero : Set _
+Nonzero = Σ Carrier IsNonzero
+
+nonzero? : (_≟_ : Decidable₂ _≈_) → Decidable IsNonzero
+nonzero? _≟_ x  with x ≟ 0#
+... | no ne   =  yes ne
+... | yes x≈0 =  no (\x≉0 → x≉0 x≈0)
--- a/src/Data/Nat/Binary/DivMod.agda
+++ b/src/Data/Nat/Binary/DivMod.agda
@@ -13,16 +13,15 @@ open import Data.Nat.Binary.Induction
 open import Data.Nat.Binary.Properties
 open import Data.Nat.Binary.Subtraction
 open import Function using (_∘_)
-open import Relation.Binary using (Tri)
 import Relation.Binary.Reasoning.Base.Triple as InequalityReasoning
 open import Relation.Binary.PropositionalEquality as P using
-  (_≡_; _≢_; refl; sym; cong; resp₂)
+  (_≡_; _≢_; refl; sym; cong; cong₂; resp₂)
+open import Relation.Nullary using (Dec; yes; no)
 open import Induction.WellFounded using (Acc)
 open Acc using (acc)
 open InequalityReasoning {A = ℕᵇ} {_≈_ = _≡_} ≤-isPreorder <-trans (resp₂ _<_)
                          <⇒≤ <-≤-trans ≤-<-trans
 open ℕᵇ
-open Tri
 
 ------------------------------------------------------------------------------
 -- DivMod  is the definition of what is division with remainder,
@@ -50,7 +49,60 @@ record DivMod (dividend divisor : ℕᵇ) : Set where  -- result of division wit
 -- otherwise                      (2q+1, 2r - b).
 -- In the latter case it is needed to prove  2r - b < b.
 -- This is equivalent to  2r < 2b  ~  r < b.
+
 ------------------------------------------------------------------------------
+-- Below the module  Even  helps the equality proof for  divMod 2[1+ a ] b.
+-- The module  Odd  helps the equality proof for  divMod 1+[2 a ] b.
+
+module Even (a b q r : ℕᵇ) (suc-a≡r+qb : suc a ≡ r + q * b)
+  where
+  2r   = double r;   2q   = double q;  qb   = q * b
+  2*qb = double qb;  2q*b = 2q * b;    2r-b = 2r ∸ b
+
+  2[1+a]≡2r+[2q]b : 2[1+ a ] ≡ double r + (double q) * b
+  2[1+a]≡2r+[2q]b = begin-equality
+    2[1+ a ]               ≡⟨ 2[1+_]-double-suc a ⟩
+    double (suc a)         ≡⟨ cong double suc-a≡r+qb ⟩
+    double (r + q * b)     ≡⟨ double-distrib-+ r qb ⟩
+    double r + double qb   ≡⟨ cong (2r +_) (sym (double-*-assoc q b)) ⟩
+    2r + 2q*b              ∎
+
+  eqFor≮ : double r ≮ b → 2[1+ a ] ≡ ((double r) ∸ b) + 1+[2 q ] * b
+  eqFor≮ 2r≮b = begin-equality
+    2[1+ a ]                   ≡⟨ 2[1+a]≡2r+[2q]b ⟩
+    2r + 2q*b                  ≡⟨ cong (_+ 2q*b) (sym ([x∸y]+y≡x b≤2r)) ⟩
+    ((2r ∸ b) + b) + 2q*b      ≡⟨ +-assoc 2r-b b _ ⟩
+    2r-b + (b + 2q*b)          ≡⟨ cong (2r-b +_) (sym (suc-* 2q b)) ⟩
+    2r-b + (suc 2q) * b        ≡⟨ cong ((2r-b +_) ∘ (_* b))
+                                       (sym (1+[2_]-suc-double q)) ⟩
+    (2r ∸ b) + 1+[2 q ] * b    ∎
+    where
+    b≤2r = ≮⇒≥ 2r≮b
+
+module Odd (a b q r : ℕᵇ) (a≡r+qb : a ≡ r + q * b)
+  where
+  2r = double r;  qb = q * b;  2q = double q;  2q*b = 2q * b
+
+  1+[2a]≡1+[2r]+2q*b :  1+[2 a ] ≡ 1+[2 r ] + 2q * b
+  1+[2a]≡1+[2r]+2q*b = begin-equality
+    1+[2 a ]                 ≡⟨ 1+[2_]-suc-double a ⟩
+    suc (double a)           ≡⟨ cong (suc ∘ double) a≡r+qb ⟩
+    suc (double (r + qb))    ≡⟨ cong suc (double-distrib-+ r qb) ⟩
+    suc (2r + (double qb))   ≡⟨ sym (suc-+ 2r (double qb)) ⟩
+    suc 2r + (double qb)     ≡⟨ cong₂ _+_ (sym (1+[2_]-suc-double r))
+                                          (sym (double-*-assoc q b)) ⟩
+    1+[2 r ] + 2q * b        ∎
+
+  eqFor≮ :  1+[2 r ] ≮ b → 1+[2 a ] ≡ (1+[2 r ] ∸ b) + 1+[2 q ] * b
+  eqFor≮ 1+2r≮b = begin-equality
+    1+[2 a ]                        ≡⟨ 1+[2a]≡1+[2r]+2q*b ⟩
+    1+[2 r ] + 2q*b                 ≡⟨ cong (_+ 2q*b) (sym ([x∸y]+y≡x b≤1+2r)) ⟩
+    ((1+[2 r ] ∸ b) + b) + 2q*b     ≡⟨ +-assoc c b 2q*b ⟩
+    (1+[2 r ] ∸ b) + (b + 2q*b)     ≡⟨ cong (c +_) (sym (suc-* 2q b)) ⟩
+    (1+[2 r ] ∸ b) + (suc 2q) * b   ≡⟨ cong ((c +_) ∘ (_* b)) (sym (1+[2_]-suc-double q)) ⟩
+    (1+[2 r ] ∸ b) + 1+[2 q ] * b   ∎
+    where
+    c = 1+[2 r ] ∸ b;   b≤1+2r = ≮⇒≥ 1+2r≮b
 
 divMod : (a b : ℕᵇ) → b ≢ 0ᵇ → DivMod a b
 divMod a b b≢0 =  dm a (<-wellFounded a)
@@ -61,9 +113,9 @@ divMod a b b≢0 =  dm a (<-wellFounded a)
 
   dm : (a : ℕᵇ) → Acc _<_ a → DivMod a b             -- b is fixed in this loop
   dm zero          _        =  divModᶜ 0ᵇ 0ᵇ refl (x≢0⇒x>0 b≢0)
-  dm a'@(2[1+ a ]) (acc wf) =  correct (dm (suc a) (wf _ suc-a<2[1+a]))
+  dm a'@(2[1+ a ]) (acc wf) =  recurseEven (dm (suc a) (wf _ suc-a<2[1+a]))
     where
-    2b = double b;  1+a = suc a     -- wf :  ∀ x → x < 2[1+ a ] → DivMod x b
+    -- wf :  ∀ x → x < 2[1+ a ] → DivMod x b
 
     suc-a<2[1+a] :  suc a < 2[1+ a ]
     suc-a<2[1+a] =  begin-strict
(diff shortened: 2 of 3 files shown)


mechvel:
I have erroneously merged the commits of Binary.DivMod and Algebra.Properties.Semiring.
If you can consider this joint commit of May 10, then, please, do this.
Otherwise, please, consider instead only the commit of DivMod of May 5 (6f0af2e).
If this is not possible, then I can edit the commit of May 10 so that it will remain only Binary.DivMod.
Please, advise.

src/Algebra/Properties/Semiring.agda
```diff
@@ -0,0 +1,32 @@
...
+-- The general notion of a nonzero element
+
+IsNonzero : Pred Carrier α≈
+IsNonzero = (_≉ 0#)
+
+Nonzero : Set _
+Nonzero = Σ Carrier IsNonzero
+
+nonzero? : (_≟_ : Decidable₂ _≈_) → Decidable IsNonzero
+nonzero? _≟_ x  with x ≟ 0#
```
MatthewDaggitt, line 30: This is just `¬? (x ≟ 0#)` from `Relation.Nullary.Decidable`

src/Data/Nat/Binary/DivMod.agda
```diff
@@ -0,0 +1,189 @@
...
+-- By recursion, let      a = qb + r,    r < b.
+-- Then  2a = 2qb + 2r,   r ≤ 2r < 2b,   r < b  < 2b.
+-- If  2r < b  then the result is (2q , 2r),
+-- otherwise                      (2q+1, 2r - b).
+-- In the latter case it is needed to prove  2r - b < b.
+-- This is equivalent to  2r < 2b  ~  r < b.
+------------------------------------------------------------------------------
+
+divMod : (a b : ℕᵇ) → b ≢ 0ᵇ → DivMod a b
+divMod a b b≢0 =  dm a (<-wellFounded a)
```
MatthewDaggitt, line 56: @mechvel this approach is not practical. The fact that this function takes takes 150 lines to define vs the 6 lines it takes to define `div` and `mod` for `ℕ` should be a strong indication of this. However the main problem is that the values can't be calculated independently from the proofs, and makes using this definition extremely difficult. For example the goal in the first hole in the proof below is is **1552** lines.
```
test : ∀ a → quot a 1ᵇ (λ()) ≡ a
test zero     = refl
test 2[1+ a ] = {!!}
test 1+[2 a ] = {!!}
```
The way to go about this is:
1. define functions calculating only the values
2. prove equivalence to division and remainder for `ℕ`
3. transfer all the proofs across using the equivalence.

I know you have concerns about the performance of step 3, but it's definitely possible to avoid evaluating the proofs if you read up about irrelevance and take a little care about how you define your functions.

src/Algebra/Properties/Semiring.agda
mechvel: All right. I am going to put ``nonzero? _≟_ x = ¬? (x ≟ 0#)``
Only `¬?` is in Relation.Nullary.Negation.

src/Data/Nat/Binary/DivMod.agda
mechvel: Well, let me try.

mechvel:
Please, consider the improved commit of Binary.DivMod 
(b29f447, May 17, 2020). 
                                                                                                      
The implementation for  divMod  has now  115 lines  of code (including empty ones)                    
+ 7 lines of comments.                                                                                
The main difference to the previous version is as follows.                                            
1. ``<-cmp foo foo'``  is changed to  ``foo <? foo'``,                                                
   with considering the cases of yes/no instead of the three cases for `<-cmp`.                       
2. Certain four common equality proofs are separated to the modules Even and Odd.                     
                                                                                                      
Now let me consider your notes.                                                                       
                                                                                                      
> this approach is not practical. The fact that this function takes                            
> 150 lines to define vs the 6 lines it takes to define div and mod for ℕ                             
> should be a strong indication of this.                                                              
                                                                                                      
It is strange to compare this to  div and mod for ℕ.                                                  
Because divMod (div, mod) for ℕ is built-in in lib-1.3, its properties are postulated.                
Are they?                                                                                             
It uses Agda.Builtin.helper etc.                                                                      
I look into Data.Nat.DivMod.Core and do not understand anything there,                                
do not understand the idea.   
It is evident to me that for the current state of the standard library,                               
divMod for ℕ can be only by using ``built-in div and mod`` and by postulating their                   
main properties.                                                                                      
I do not see any other way to implement fast divMod for unary arithmetic.                             
                                                                                                      
Therefore your points 1 and 2 are not possible to do.                                                 
                                                                                                      
Your point 1 is possible to do.                                                                       
But the related proofs for equality and inequality cannot be done by using toℕ.                       
Because such proofs are not supplied for divMod for ℕ.                                                                                                                                
These related proofs for ℕᵇ can be implemented, but I expect, the code will not  be any essentially shorter than the current 115 lines of my divMod.  
I tried to define  divMod!  for computing only the proper values, and so on,  and stopped. 
Because I feel that adding proofs will be too complex.                                   
                                                                                                      
I am going to send the file for  divMod!  to Matthew  -- to report that I indeed tried this.                                                                
                                                                                                      
I do not see of how to add proofs to this divMod! in a reasonable way.                                
Maybe someone would demonstrate how to implement your approach to divMod  for Binary ?    

> However the main problem is that the values can't be calculated  independently from the proofs, 
> and makes using this definition  extremely difficult. For example the goal in the first hole in the                                  
> proof below is 1552 lines.                                                                  
```                                                                                                   
> test : ∀ a → quot a 1ᵇ (λ()) ≡ a                                                                    
> test zero     = refl                                                                                
> test 2[1+ a ] = {!!}                                                                                
> test 1+[2 a ] = {!!}                                                                                
```                                                                                                   
I do not believe holes are helpful, never use them. Why do you need them for `quot` ?                                 
If you need to prove this `test`, then it is somewhat like this:                                      
```                                                                                                   
test a = let (divModᶜ q r a≡r+q*1 r<1) = divMod a 1ᵇ (λ())                                            
                  r≡0 = x<1⇒x≡0 r<1                                                                        
             in                                                                                           
             begin q               ≡⟨ sym (+-identityL q) ⟩                                                  
                       0ᵇ + q       ≡⟨ cong₂ (sym r≡0) (sym (*-identityR q)) ⟩                                
                       r + q * 1ᵇ   ≡⟨ sym a≡r+q*1 ⟩                                                          
                       a                                                                                      
            ∎                                                                                            
```                                                                                                   
If you mean that proofs mixed with values slow down the run-time performance,                                     
then I think this is not so for this particular divMod. I have tests for divMod,                           
and gcd. We can consider the performance tests and discuss them.                                                  
                               
> but it's definitely possible to avoid evaluating the proofs if you read up about irrelevance and take a > little care about how you define your functions.                                                                
                                                                                                      
a) I do not see any reasonable way to separate proofs for Binary.divMod.                              
b) What irrelevance you are talking of?                                                                      
In my programs, I prove irrelevance of some functions on some proofs,                                
and then use such irrelevance lemmas.                                                                
Also I recall a certain irrelevance denotation in the Agda language.                                  
I used such in a certain program version, but this lead to errors which I was not able to handle. 
After this I stopped using such denotation. Instead I formulate explicitly the irrelevance statement - when it is needed, prove it, and use it.                                                                                 
I do not understand the last two lines in your comment.

mechvel:
In other words: I understand this so that you suggest to port the two main proofs for divMod for Binary by using the homomorphism property for toNat/fromNat and the main properties of divMod 
for Nat.  
But these properties for Nat are not proved in the library, they are postulated
(how anything can be proved for a built-in function?).
It comes out that you suggest to prove the things for Binary by referring to certain postulates about certain built-ins for Nat.
?
My main idea for  Binary is that it has to avoid any built-ins and any postulates for built-ins.
Binary arithmetic is implemented  only in Agda and proved in Agda.

mechvel:
A correct way to separate proofs for Binary divMod by  using ``toℕ/fromℕ`` is as follows.

1. To program in Agda  divMod-Nat  returning ``ℕ × ℕ``.
     It is by repeatedly subtracting  divisor  from dividend until the remainder is less than divisor.
The quotient is accumulated by applying suc at each subtraction.
This is exponentially slow. But I do not know of any faster method for divMod in unary system,
and also its performance is not essential for the goal.
2. To prove the equality and inequality statements for the result of divMod-Nat.
    I expect this is simple.
3. To program fast `divMod!` for Binary that returns ``ℕᵇ × ℕᵇ`` and does not carry proofs, except for 
     termination.
4. To prove the Nat-equivalence. That is that `divMod!` composed with `toℕ` returns the same result     
    as divMod-Nat composed with `fromℕ`.
5. To derive the needed proofs for `divMod!` from this equivalence and from the proofs for 
    divMod-Nat.
This will provide fast  `divMod!`  for Binary with the proofs separated from the `value' computation.

But  a) the code will probably be not smaller than my current code for divMod, 
b) I like my current divMod much more.

Concerning your example with 
```
test : ∀ a → quot a 1ᵇ (λ()) ≡ a                                                                    
test zero     = refl                                                                                
test 2[1+ a ] = {!!}                                                                                
test 1+[2 a ] = {!!}                                                                                
```
Probably the first ``{!!}`` invites Agda to substitute any proof for `test` for the case of  ``2[1+ a ]``.
Trying to find this proof the prover starts to normalize the term 
``DivMod.quotient (divMod  2[1+ a ] 1ᵇ (λ()))``.
Right?
In rare lucky cases, such normalization can derive the needed proof.
But why do you expect, or require,  of the prover  to show something small in this situation?
To my mind, it is natural for normalization to return a result having a large print-out.
Also when a term is large on printing, this does not necessarily mean that it takes much memory.   
Because (I hope) sharing is used.  

Another point.
Your complain on a large proof shown in the hole.
On the other hand, you suggested the  `toℕ` conversion to bring proofs from unary arithmetic to binary. For example, this conversion exponentially increases the proof size for `≤`. And you said, 
that proofs are not printed in normal situation, etc. (of which I doubt greatly).
Now, set  ``{!!}``  for some proof for   ``x * z ≤ y * z`` for Binary that uses the conversion  `toℕ`.
It will place into the hole a term that may include billions of `s≤s` constructors.
Am I missing something?
Maybe, treating a hole in the prover can be improved, I do not know.

Theoretically, a prover that attempts to find a proof by substituting cases and by normalizing, is a normal part  of  a (semi-)automatic prover.
But personally, I have written many proofs in Agda, and never felt that I need a hint from a hole.
And when a proof is difficult to invent, a hole still would not help. 
Such is the current practice.

I think there are more essential examples than this  ``{!!}``, but I would not search for them right now.

jespercockx:
I'm not an expert on this part of the library, but you can find the definitions of division and modulus in `Data.Nat.DivMod`, which make use of two functions `div-helper` and `mod-helper` that are defined in `Agda.Builtin.Nat` as Agda code (not as postulates!) The `{-# BUILTIN ... #-}` pragmas are only there to make sure they are evaluated using fast Haskell arithmetic rather than the unary representation.

The main correctness theorems for division is also given in `Data.Nat.DivMod` as `m≡m%n+[m/n]*n` and `m%n<n`, but the actual proofs of the helper lemmas can be found in `Data.Nat.DivMod.Core`. These lemmas are about 20 lines in total, so they seem to be quite a bit simpler than for the binary representation (as I would expect).

mechvel:
> Data.Nat.DivMod, which make use of two functions div-helper and mod-helper that are defined in 
> Agda.Builtin.Nat as Agda code (not as postulates!) The {-# BUILTIN ... #-} pragmas are only there > to make sure they are evaluated using fast Haskell arithmetic rather than the unary representation.

Thank you for help.
Actually my question was: what precisely is postulated in this whole approach?

> div-helper and mod-helper that are defined in Agda.Builtin.Nat as Agda code (not as postulates!) 

1. I do not see, where is this Agda.Builtin.Nat.
2. "not as postulates!"  means  that no their properties are postulated. 

But what is postulated in the whole this approach?
How can one prove in Agda  statements about a function from the GHC library (probably written 
in C) ?
Some postulate has to be used. But such not named. 
And I am trying to find it.
As I understand from your remark,  `div-helper` is programmed in Agda, and its main properties are proved in Agda. But it is evaluated as built-in. So that there it not any proof for that the result of this built-in evaluation equals to the result of the Agda interpreter at `div-helper`.
Instead of such (impossible and senseless) proof, it is `silently presumed` that these two results are equal. Actually this means a _postulate_ "forall x (div-helper x == div-helper-built-in x)".  
Right?
Using this equality `postulate` (why is not it explicitly named?), the properties of `div-helper-built-in` are derived from the properties of the function `div-helper` written in Agda.
Is this so?
Now, the Matthew's  suggestion was to prove that divMod! for Binary returns the same number pair as is returned by the pair of div-helper and mod-helper for Nat. As all these functions are written in Agda, this is indeed possible.
Only I need to find the code for `div-helper` and see ...

jespercockx:
The module `Builtin.Nat` is distributed together with Agda itself so it is not part of the standard library. Its location depends on how you installed Agda, on my system it is `/home/jesper/.cabal/share/x86_64-linux-ghc-8.8.3/Agda-2.6.2/lib/prim/Agda/Builtin/Nat.agda`. It contains two helper functions `div-helper` and `mod-helper` that are used to implement a fast `divMod` for unary numbers in the standard library.

You are correct that this approach is essentially postulating that the Agda implementation of `div-helper` and `mod-helper` is equivalent to the native Haskell implementation which you can find at https://github.com/agda/agda/blob/6e17bfdf163dadf0491fc98c20897081e616f340/src/full/Agda/TypeChecking/Primitive.hs#L754-L759. There is no explicit `postulate` since this `div-helper-builtin` is a Haskell function, so it is not possible to refer to it inside an Agda type.

Anyway, by using Agda you are implicitly trusting many parts of the (unverified) Haskell implementation of the Agda typechecker, so I do not see why this particular piece of Haskell code should be cause for concern.

mechvel:
This all does not somehow look clear.
Why not to arrange this as follows?
Put to  ``Nat.DivMod.agda`` : 
```
divₕ : Op₂ ℕ             -- quotient of m by (suc n)  - the helper
divₕ m n =  < code in Agda >

modₕ : Op₂ ℕ             -- remainder of m by (suc n) - the helper
modₕ m n =  < code in Agda >

-- A theorem proof for helpers:
divMod-helper-eq : (m n : ℕ) → m ≡ (modₕ m n) + (divₕ m n) * (suc n)
divMod-helper-eq m n =   < code in Agda >

div : Op₂ ℕ                -- only signature in Agda
div = Built-in.Nat.div  -- a reference to Haskell library

mod : Op₂ ℕ                   -- only signature in Agda
mod = Built-in.Nat.mod

As the signatures are in Agda, we can set "postulate" explicitly: 

postulate
   div-built-in-eq    : (m n : ℕ) → divₕ m n ≡ div m n
   mod-built-in-eq : (m n : ℕ) → modₕ m n ≡ mod m n
```
As the signatures are in Agda, and "postulate" set explicitly, we can provide a proof in Agda for a theorem for  `div`  and  `mod`  (which are built-in):  
```
divMod-eq : (m n : ℕ) → m ≡ (mod m n) + (div m n) * (suc n)
divMod-eq m n =
  begin
     m                                                      ≡⟨ divMod-helper-eq m n  ⟩
     (modₕ m n) + (divₕ m n) * (suc n)
                                      ≡⟨ cong₂ _+_ (mod-built-in-eq m n)
                                                           (cong (_* (suc n)) (div-built-in-eq m n))
                                        ⟩           
     (mod m n) + (div m n) * (suc n)
   ∎ 
``` 
Similarly can proved the inequality theorem for `div` and `mod`. 
With this approach it is clearly shown what is postulated, the signatures and `postulate` are 
set in Agda. And all the remaining things are programmed in Agda.
Am I missing something?

jespercockx:
The reason for having actual Agda definitions in the `Builtin.Nat` file (instead of postulates) is precisely *because* it allows us to prove things about those definitions in Agda itself. If they were postulates, then we would also have to postulate their properties, and this would prevent them from being used in `--safe` mode. You may not like it, but a lot of people are perfectly happy with this since it gives both relatively simple proofs *and* fast execution of natural number arithmetic. The only cost is having to trust that this small bit of Haskell code has the same computational behaviour on closed natural numbers as the naive Agda version. And as I said, since you are anyway trusting Agda as a whole I would say this is no cost at all.

mechvel:
> Anyway, by using Agda you are implicitly trusting many parts of the (unverified) Haskell 
> implementation of the Agda typechecker, so I do not see why this particular piece of Haskell code > should be cause for concern.

If so, then we should program everything in Agda without proofs.
I think we need to explicitly point the lower bound, so that everything beyond this bound is programmed in Agda and can be proved in Agda. And this bound is marked by the `postulate` word in Agda. 
The type checker implementation is not written in Agda, and it is not (so far) a subject for Agda  proofs. But the standard library functions are written in Agda, and are subjected to Agda proofs.
The lower bound is set for built-ins with their "postulates" - the better.

mechvel:
Consider the summary.
The matter with unary arithmetic in lib-1.3 is that
(a) the helper functions are given in Agda, with proofs,
(b) helpers are not in reality evaluated, instead evaluated are built-ins,
(c) no connection between helpers and built-ins is declared, even in comments. 
     Is it?

And I suggested above 
```
div : Op₂ ℕ                 -- only signature in Agda
div = Built-in.Nat.div     -- a reference to Haskell library
...
postulate
    div-built-in-eq   :  (m n : ℕ) → div-helper m n ≡ div m n
```
which declares in Agda this connection and even allows Agda proofs based on this postulate.
May be I am missing something, I do not know.

jespercockx:
> Consider the summary.
> The matter with unary arithmetic in lib-1.3 is that
> (a) the helper functions are given in Agda, with proofs,
> (b) helpers are not in reality evaluated, instead evaluated are built-ins,
> (c) no connection between helpers and built-ins is declared, even in comments.
> Is it?

That looks correct.

> And I suggested above [...] which declares in Agda this connection and even allows Agda proofs based on this postulate.
> May be I am missing something, I do not know.

This would certainly be possible. However, I believe it would be a step back from the current situation, especially since the `postulate` would disallow the fast division from being used in `--safe` mode. But since I am not a maintainer of the standard library, perhaps others can also give their opinion on this matter.

mechvel:
> the postulate would disallow the fast division from being used in --safe mode.

What does it mean: --safe mode? 
Can applying a built-in function be considered as _safe_?

jespercockx:
Safe Agda is descibed in the Agda documentation: https://agda.readthedocs.io/en/v2.6.1/language/safe-agda.html

> By using the option --safe (as a pragma option, or on the command-line), a user can specify that Agda should ensure that features leading to possible inconsistencies should be disabled.

Since fast unary arithmetic does not lead to inconsistencies, I believe it should be allowed under `--safe` mode.

src/Algebra/Properties/Semiring.agda
mmontin: Or even simpler: nonzero? _≟_ = ¬? ∘ (_≟ 0#)

mechvel:
Please, consider my suggestion.

Our Nat.Binary thing has somehow stuck.
My ``Data.Nat.Binary.DivMod`` mixes the proper gcd evaluation and a proof construction.
And Matthew requires to totally separate them.
I dislike such a separation in this case.

May be the team accepts my version as temporary?

So that I would add my DivMod and GCD for Binary, and Nat.Binary will be done, as intended, with all proofs. 
And it will be usable in the next release of the library.

Then, the team can change the implementation for divMod and GCD as they like. It is possible to do without changing their signatures, so that the backwards compatibility will be preserved. 
Only - with setting the authorship in the comments on these two functions.

?

MatthewDaggitt:
> May be the team accepts my version as temporary?

> Then, the team can change the implementation for divMod and GCD as they like. 

Hmm I'm not very happy with this suggestion I'm afraid. We definitely want to avoid accepting code that we're not happy with on the vague notion that some day we're going to rewrite it. Not to mention that none of the regular team has the time or the inclination to continue to fix-up this contribution.

> It is possible to do without changing their signatures, so that the backwards compatibility will be preserved.

Alas changing the implementation will break backwards compatibility even if the signature stays the same.

In conclusion I think that unless you're willing to rewrite the implementation of this contribution, then perhaps it's best that this code stays part of your own binary library?

mechvel:
I do not like separating proofs from divMod.
As you do not agree, we could, for example, leave Nat.Binary as it is in lib-1.4, without divMod and gcd.
Anybody can add there the implementation for divMod and gcd, and we could discuss them.
Or, I do not know, you can replace the whole Nat.Binary with any code your like, or just remove it.

MatthewDaggitt:
> As you do not agree, we could, for example, leave Nat.Binary as it is in lib-1.4, without divMod and gcd.
Anybody can suggest/add there the implementation for divMod and gcd, and we could discuss them.

Sure!

PR #1196, opened 2020-05-10
Split Relation.Binary.PropositionalEquality into finer grained pieces

JacquesCarette:
that can be used internally to lower the weight of dragging in all of R.B.PE.

### DIFF
--- a/src/Relation/Binary/PropositionalEquality/Bundles.agda
+++ b/src/Relation/Binary/PropositionalEquality/Bundles.agda
@@ -1,0 +1,68 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Propositional (intensional) equality:
+--   Equivalence, Order, Setoid
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Binary.PropositionalEquality.Bundles where
+
+open import Function.Base using (id)
+open import Level using (Level)
+
+open import Relation.Binary
+open import Relation.Binary.PropositionalEquality.Core
+
+private
+  variable
+    a : Level
+    A : Set a
+
+------------------------------------------------------------------------
+-- Structure of equality as a binary relation
+
+isEquivalence : IsEquivalence {A = A} _≡_
+isEquivalence = record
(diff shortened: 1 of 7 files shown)


src/Relation/Binary/PropositionalEquality/Bundles.agda
```diff
@@ -0,0 +1,68 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Propositional (intensional) equality:
+--   Equivalence, Order, Setoid
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Binary.PropositionalEquality.Bundles where
```
MatthewDaggitt, line 10: I'm not sure `Bundles` is quite the right word as we also have structures in here. If we were being consistent then we'd either put it in `Relation.Binary.PropositionalEquality.Properties` or call this `Relation.Binary.PropositionalEquality.Relation.Binary` 😆

I assume the former is problematic?

JacquesCarette: I struggled with naming this one. The double `Relation.Binary` does seem to be a bit much. I could rename the current `Properties` to `Rearrange` and then `Bundles` could be `Properties`.

MatthewDaggitt: You definitely need them to be separate? Up until now we've never distinguished between these two types of properties...

JacquesCarette: Definitely? No. I split them mostly because they depend on somewhat different imports.

MatthewDaggitt: Hmm there's only one import difference. If you don't need them to be different, then I'd be tempted to combine them and save us the trouble of coming up with a new naming convention?

JacquesCarette: Ok, I'm probably going overboard there. I'll merge them.

MatthewDaggitt:
Great, thanks! Merging in.

PR #1167, opened 2020-04-22
Make use of apostrophe and \prime consistent.

elsanussi-s-mneina:
I am not sure of whether the changes I made in lines 54 to 58 of Colist.agda will affect code outside the library. This is because it uses the GHC directive. Please double check the changes I made to that file.

Issue #995, opened 2019-12-16
Consistent use of ` and \prime (again!)


MatthewDaggitt:
I think at one point we tried to standardise the use of these but apparently (#993) it's crept back into the library again.

Actually maybe it was `~` and \tilde (#428) that we tried to standardise...

MatthewDaggitt:
Closed with #1205 1205

PR #1202, opened 2020-05-10
(apostrophe to prime) Deprecate _∷ʳ'_ (both versions) and decidable' functions.

elsanussi-s-mneina:
This commit partly resolves the issue with inconsistent use of apostrophe and primes: #995

It resolves it for three functions.
This pull request is related to pull request: #1167 . This pull request is simpler, and will therefore be easier to maintain.

I decided to deal with the functions that would have to be deprecated first. Other identifiers will have to be dealt with later. The hope is that this will make use of apostrophe's and primes consistent to the user of the library, which is probably most important. Later on we can make it consistent internally.

Deprecated:
Data.List.Base.InitLast._∷ʳ'_
Data.List.NonEmpty.SnocView._∷ʳ'_
Relation.Binary.Construct.StrictToNonStrict.decidable'

If this looks good, then I can work on making use of apostrophe and primes consistent for local variable names, and identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ Deprecated names
 * The proofs `replace-equality` from `Algebra.Properties.(Lattice/DistributiveLattice/BooleanAlgebra)`
   have been deprecated in favour of the proofs in the new `Algebra.Construct.Subst.Equality` module.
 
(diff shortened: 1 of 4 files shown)


MatthewDaggitt:
Looks good! Thanks for the smaller PRs!

PR #1203, opened 2020-05-10
WIP: port old TypeIsomorphisms over to the new Function hierarchy. 

JacquesCarette:
Put in what I think are some improvements while at it too.  I'm doing a PR now, to get a code review.


src/Data/Sum/Algebra.agda
```diff
@@ -0,0 +1,64 @@
...
+open import Level
+
+open import Function.Bundles using (_↔_; Inverse; mk↔)
+import Function.Definitions as FuncDef
+
+open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
+
+------------------------------------------------------------------------
+
+module _ {a b : Level} {A : Set a} {B : Set b} where
```
MatthewDaggitt, line 26: Using variables for the levels and the later sets and would clear things up a bit I think?

JacquesCarette: I can use variables for the levels, but not the sets, as they are explicitly passed to an `open` statement.

JacquesCarette:
This should not be good to go.

CHANGELOG.md
```diff
@@ -159,6 +166,32 @@ Other minor additions
   filter⁺ : Transitive R → Linked R xs → Linked R (filter P? xs)
   ```
 
+* Add new functions to `Data.Product`:
+  ```agda
+  assocʳ : {B : A → Set b} {C : (a : A) → B a → Set c} →
```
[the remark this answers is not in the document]
JacquesCarette: Done.

JacquesCarette:
So the real work is done - but it will make sense to merge #1224 first, and then I can clean up the private versions.  I'll do a proper CHANGELOG entry then as well.  But if there are other things I should clean up, I'd like to know now, so I can do them all at once.

src/Data/Product/Algebra.agda
```diff
@@ -0,0 +1,278 @@
...
+  ×-⊎-commutativeSemiring : CommutativeSemiring (suc ℓ) ℓ
+  ×-⊎-commutativeSemiring = record
+    { isCommutativeSemiring = ⊎-×-isCommutativeSemiring
+    }
+
+------------------------------------------------------------------------
+-- Some reordering lemmas
+
+ΠΠ↔ΠΠ : (P : A → B → Set p) → ((x : A) (y : B) → P x y) ↔ ((y : B) (x : A) → P x y)
+ΠΠ↔ΠΠ _ = inverse flip flip cong′ cong′
```
MatthewDaggitt, line 202: Hmm the proofs below this point shouldn't live in `Data.Product.Algebra` but I'm at a bit of a loss about where to put them... Anyone have any suggestions?

MatthewDaggitt: Well I've had a stab at moving them to "sensible" places. If no-one has any more feedback I'll merge it in next time I'm in the area.

JacquesCarette: These places seem eminently sensible to me!  Note: the CHANGELOG hasn't been updated for that.

JacquesCarette:
Thanks for the tidying up!

PR #1204, opened 2020-05-10
Polymorphic unit and empty rethought

JacquesCarette:
The previous definitions in `Data.Unit.Polymorphic` and `Data.Empty.Polymorphic` did not inter-operate very well with the rest of the library. I saw two ways forward (https://github.com/agda/agda/issues/4646). The one approach got some immediate push-back, so I figured I would try the other one - and it worked surprisingly well.

This PR is the result of that. It isn't ready to be merged, as it's going to need a thorough review. I'll write something for the CHANGELOG once the dust has settled.

### DIFF
--- a/src/Category/Applicative.agda
+++ b/src/Category/Applicative.agda
@@ -12,7 +12,7 @@
 module Category.Applicative where
 
 open import Level using (Level; suc; _⊔_)
-open import Data.Unit
+open import Data.Unit.Polymorphic
 open import Category.Applicative.Indexed
 
 private
@@ -20,21 +20,21 @@ private
     f : Level
 
 RawApplicative : (Set f → Set f) → Set (suc f)
-RawApplicative F = RawIApplicative {I = ⊤} λ _ _ → F
+RawApplicative {i} F = RawIApplicative {I = ⊤ {i}} λ _ _ → F
 
 module RawApplicative {F : Set f → Set f}
                       (app : RawApplicative F) where
   open RawIApplicative app public
 
 RawApplicativeZero : (Set f → Set f) → Set _
-RawApplicativeZero F = RawIApplicativeZero {I = ⊤} (λ _ _ → F)
+RawApplicativeZero {i} F = RawIApplicativeZero {I = ⊤ {i}} (λ _ _ → F)
 
 module RawApplicativeZero {F : Set f → Set f}
                           (app : RawApplicativeZero F) where
   open RawIApplicativeZero app public
 
 RawAlternative : (Set f → Set f) → Set _
-RawAlternative F = RawIAlternative {I = ⊤} (λ _ _ → F)
+RawAlternative {i} F = RawIAlternative {I = ⊤ {i}} (λ _ _ → F)
 
 module RawAlternative {F : Set f → Set f}
                       (app : RawAlternative F) where
--- a/src/Category/Applicative.agda
+++ b/src/Category/Applicative.agda
@@ -12,7 +12,7 @@
 module Category.Applicative where
 
 open import Level using (Level; suc; _⊔_)
-open import Data.Unit.Polymorphic
+open import Data.Unit
 open import Category.Applicative.Indexed
 
 private
@@ -20,21 +20,21 @@ private
     f : Level
 
 RawApplicative : (Set f → Set f) → Set (suc f)
-RawApplicative {i} F = RawIApplicative {I = ⊤ {i}} λ _ _ → F
+RawApplicative F = RawIApplicative {I = ⊤} λ _ _ → F
 
 module RawApplicative {F : Set f → Set f}
                       (app : RawApplicative F) where
   open RawIApplicative app public
 
-RawApplicativeZero : (Set f → Set f) → Set _
-RawApplicativeZero {i} F = RawIApplicativeZero {I = ⊤ {i}} (λ _ _ → F)
+RawApplicativeZero : (Set f → Set f) → Set (suc f)
+RawApplicativeZero F = RawIApplicativeZero {I = ⊤} (λ _ _ → F)
 
 module RawApplicativeZero {F : Set f → Set f}
                           (app : RawApplicativeZero F) where
   open RawIApplicativeZero app public
 
 RawAlternative : (Set f → Set f) → Set _
-RawAlternative {i} F = RawIAlternative {I = ⊤ {i}} (λ _ _ → F)
+RawAlternative F = RawIAlternative {I = ⊤} (λ _ _ → F)
 
 module RawAlternative {F : Set f → Set f}
                       (app : RawAlternative F) where
--- a/src/Category/Monad/State.agda
+++ b/src/Category/Monad/State.agda
@@ -13,7 +13,7 @@ open import Category.Monad
 open import Function.Identity.Categorical as Id using (Identity)
 open import Category.Monad.Indexed
 open import Data.Product
-open import Data.Unit.Polymorphic
+open import Data.Unit
 open import Function
 open import Level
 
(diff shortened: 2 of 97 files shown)


src/Category/Applicative.agda
```diff
@@ -12,29 +12,29 @@
...
+open import Data.Unit.Polymorphic
 open import Category.Applicative.Indexed
 
 private
   variable
     f : Level
 
 RawApplicative : (Set f → Set f) → Set (suc f)
-RawApplicative F = RawIApplicative {I = ⊤} λ _ _ → F
+RawApplicative {i} F = RawIApplicative {I = ⊤ {i}} λ _ _ → F
```
MatthewDaggitt, line 23: It's a bit confusing to call the level `f` in the type and then `i` in the definition! Same issue below, but not in the other similar files.

src/Category/Applicative.agda
```diff
@@ -12,29 +12,29 @@
...
 
 module RawApplicative {F : Set f → Set f}
                       (app : RawApplicative F) where
   open RawIApplicative app public
 
 RawApplicativeZero : (Set f → Set f) → Set _
-RawApplicativeZero F = RawIApplicativeZero {I = ⊤} (λ _ _ → F)
+RawApplicativeZero {i} F = RawIApplicativeZero {I = ⊤ {i}} (λ _ _ → F)
```
MatthewDaggitt, line 30: Actually, any chance you could explain the point of this change? I'm probably being slow but why does it matter what level the indexing type is at? Given that the indexing is trivial anyway...

src/Data/Empty/Polymorphic.agda
```diff
@@ -8,9 +8,12 @@
...
 open import Level
 
-data ⊥ {ℓ : Level} : Set ℓ where
+⊥ : {ℓ : Level} → Set ℓ
+⊥ {ℓ} = Lift ℓ ⊥.⊥
 
-⊥-elim : ∀ {w ℓ} {Whatever : Set w} → ⊥ {ℓ} → Whatever
-⊥-elim ()
+-- make ⊥-elim-P dependent too, as it does seem useful
+⊥-elim-P : ∀ {w ℓ} {Whatever : ⊥ {ℓ} → Set w} → (witness : ⊥ {ℓ}) → Whatever witness
```
MatthewDaggitt, line 18: Hmm do we need the `-P`. We shouldn't really need to import both the polymorphic and the non-polymorphic variety at the same time right?

src/Data/List/Relation/Unary/All.agda
```diff
@@ -60,7 +60,7 @@ data _[_]=_ {A : Set a} {P : Pred A p} :
 -- A list is empty if having an element is impossible.
 
 Null : Pred (List A) _
-Null = All (λ _ → ⊥)
+Null {ℓ} = All (λ _ → ⊥ {ℓ})
```
MatthewDaggitt, line 63: Again what does this add?

src/Relation/Nullary.agda
```diff
@@ -26,6 +26,10 @@ infix 2 _because_
 ¬_ : ∀ {ℓ} → Set ℓ → Set ℓ
 ¬ P = P → ⊥
 
+-- level polymorphic version of ¬
+¬′ : ∀ {ℓ} (b : Level) → Set ℓ → Set (ℓ ⊔ b)
+¬′ {_} b P = P → Lift b ⊥
```
MatthewDaggitt, line 31: Hmm this name needs to change for two reasons. Firstly we're now using `\prime` instead of `'`. But more importantly, the convention is that `\prime` signifies something is less polymorphic, not more.

The fact that you have to provide the level as well is a bit upsetting...

MatthewDaggitt:
Oh the other thing I'm a little bit worried about is that by having an extra `lift` constructor, we might be exceeding the pattern matching search depth in someone's code. @gallais I think you implemented some of it? But I can't actually think of anywhere you'd wrap either top or bottom more than once so maybe we're safe...

src/Relation/Nullary.agda
MatthewDaggitt: It occurs to me that my browser doesn't actually render the difference between `'` and `\prime` so apologies if it's already `\prime`!

src/Category/Applicative.agda
JacquesCarette: Yes, that was a mistake.  Plus, here in particular, there is very little point in making it level polymorphic.  I'll revert the changes to the various `Monad` functions.

src/Category/Applicative.agda
JacquesCarette: You're quite right, it's over-generalized.

src/Data/Empty/Polymorphic.agda
JacquesCarette: I'll remove the `-P` and see. There was one file with a weird problem, but I'm sure I can fix it.

src/Data/List/Relation/Unary/All.agda
JacquesCarette: I think this was needed because it was lifted somewhere else that needed this. I'll investigate and revert if it doesn't add anything.

src/Relation/Nullary.agda
JacquesCarette: (it was already `\prime`). Yeah, I'm not happy with the name either, not that the level is manditory. I tried to make it implicit, and then had to actually give it too often.

`\neg\_p` ?

src/Data/List/Relation/Unary/All.agda
JacquesCarette: Yeah, that was just pointless.

src/Relation/Nullary.agda
JacquesCarette: Turns out that looks awful in the usual fonts. `¬ᵢ` for "`¬` indexed by a level" looks better.

JacquesCarette:
Ok, all issues that were pointed out in the comments have been taken care of.

gallais:
> we might be exceeding the pattern matching search depth in someone's code

As long as it's a record constructor with eta, the emptyness analysis will be
happy to go under it as many times as needed.

src/Category/Monad/State.agda
```diff
@@ -105,15 +105,15 @@ StateTIMonadState S Mon = record
...
+RawMonadState : Set f → (Set f → Set f) → Set (suc f)
 RawMonadState S M = RawIMonadState {I = ⊤} (λ _ → S) (λ _ _ → M)
 
 module RawMonadState {S : Set f} {M : Set f → Set f}
                      (Mon : RawMonadState S M) where
   open RawIMonadState Mon public
 
 StateT : Set f → (Set f → Set f) → Set f → Set f
-StateT S M = IStateT {I = ⊤} (λ _ → S) M _ _
+StateT {f} S M = IStateT {I = ⊤} (λ _ → S) M _ _
```
MatthewDaggitt, line 116: No longer need the implicit argument here

src/Data/Empty/Polymorphic.agda
```diff
@@ -8,9 +8,12 @@
 
 module Data.Empty.Polymorphic where
 
+import Data.Empty as ⊥
```
MatthewDaggitt, line 11: I think it would be clearer if we renamed this to `Empty` not `⊥`

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -335,8 +334,8 @@ True↔ : ∀ {p} {P : Set p}
         (dec : Dec P) → ((p₁ p₂ : P) → p₁ ≡ p₂) → True dec ↔ P
 True↔ ( true because  [p]) irr =
   inverse (λ _ → invert [p]) (λ _ → _) (λ _ → P.refl) (irr _)
-True↔ (false because [¬p]) _   =
-  inverse (λ()) (invert [¬p]) (λ()) (⊥-elim ∘ invert [¬p])
+True↔ {a} (false because ofⁿ ¬p) _ =
```
MatthewDaggitt, line 337: `{a}` is unnecessary

src/Relation/Nullary.agda
```diff
@@ -26,6 +26,11 @@ infix 2 _because_
 ¬_ : ∀ {ℓ} → Set ℓ → Set ℓ
 ¬ P = P → ⊥
 
+-- level polymorphic version of ¬
+-- i.e. ¬ 'i'ndexed by a level.
+¬ᵢ : ∀ {ℓ} (b : Level) → Set ℓ → Set (ℓ ⊔ b)
+¬ᵢ b P = P → Lift b ⊥
```
MatthewDaggitt, line 32: Now that I think about it, given that we have `Relation.Binary.Indexed` and `Relation.Unary.Indexed`, this should probably live in `Relation.Nullary.Indexed`? Then we wouldn't even have to have the subscript?

JacquesCarette: Makes sense.

PR #1205, opened 2020-05-11
Make \prime apostrophe usage consistent (change to prime) (part 2, local identifiers)

elsanussi-s-mneina:
Re: issue #995:
So this changes the use of apostrophe's in identifiers to use prime symbols where possible. 
No CHANGELOGs were modified.


MatthewDaggitt:
Thanks for the PRs!

PR #1211, opened 2020-05-20
Decidablity of dependent products

alexarice:
Moved some proofs from Data.Product.Properties.WithK to Data.Product.Properties. Probably needs a changelog entry

### DIFF
--- a/src/Data/Product/Properties.agda
+++ b/src/Data/Product/Properties.agda
@@ -8,12 +8,14 @@
 
 module Data.Product.Properties where
 
+open import Axiom.UniquenessOfIdentityProofs
 open import Data.Product
-open import Function using (_∘_)
+open import Function using (_∘_;_∋_)
 open import Relation.Binary using (Decidable)
 open import Relation.Binary.PropositionalEquality
 open import Relation.Nullary.Product
 import Relation.Nullary.Decidable as Dec
+open import Relation.Nullary using (Dec; yes; no)
 
 ------------------------------------------------------------------------
 -- Equality (dependent)
@@ -23,7 +25,17 @@ module _ {a b} {A : Set a} {B : A → Set b} where
   ,-injectiveˡ : ∀ {a c} {b : B a} {d : B c} → (a , b) ≡ (c , d) → a ≡ c
   ,-injectiveˡ refl = refl
 
-  -- See also Data.Product.Properties.WithK.,-injectiveʳ.
+  ,-injective≡ʳ : ∀ {a b} {c : B a} {d : B b} → UIP A → (a , c) ≡ (b , d) → (q : a ≡ b) → subst B q c ≡ d
+  ,-injective≡ʳ {c = c} {.c} u refl q = cong (λ x → subst B x c) (u q refl)
+
+  ,-injectiveUIPʳ : ∀ {a} {b c : B a} → UIP A → (Σ A B ∋ (a , b)) ≡ (a , c) → b ≡ c
+  ,-injectiveUIPʳ u p = ,-injective≡ʳ u p refl
+
+  ≡-dec : Decidable {A = A} _≡_ → (∀ {a} → Decidable {A = B a} _≡_) →
+          Decidable {A = Σ A B} _≡_
+  ≡-dec dec₁ dec₂ (a , x) (b , y) with dec₁ a b
+  ... | no [a≢b] = no ([a≢b] ∘ ,-injectiveˡ)
+  ... | yes refl = Dec.map′ (cong (a ,_)) (,-injectiveUIPʳ (Decidable⇒UIP.≡-irrelevant dec₁)) (dec₂ x y)
 
 ------------------------------------------------------------------------
 -- Equality (non-dependent)
@@ -35,8 +47,3 @@ module _ {a b} {A : Set a} {B : Set b} where
 
   ,-injective : ∀ {a c : A} {b d : B} → (a , b) ≡ (c , d) → a ≡ c × b ≡ d
   ,-injective refl = refl , refl
-
-  ≡-dec : Decidable {A = A} _≡_ → Decidable {A = B} _≡_ →
-          Decidable {A = A × B} _≡_
-  ≡-dec dec₁ dec₂ (a , b) (c , d) =
-    Dec.map′ (uncurry (cong₂ _,_)) ,-injective (dec₁ a c ×-dec dec₂ b d)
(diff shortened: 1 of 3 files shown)


src/Data/Product/Properties.agda
```diff
@@ -23,7 +25,17 @@ module _ {a b} {A : Set a} {B : A → Set b} where
...
   ,-injectiveˡ refl = refl
 
-  -- See also Data.Product.Properties.WithK.,-injectiveʳ.
+  ,-injective≡ʳ : ∀ {a b} {c : B a} {d : B b} → UIP A → (a , c) ≡ (b , d) → (q : a ≡ b) → subst B q c ≡ d
+  ,-injective≡ʳ {c = c} {.c} u refl q = cong (λ x → subst B x c) (u q refl)
+
+  ,-injectiveUIPʳ : ∀ {a} {b c : B a} → UIP A → (Σ A B ∋ (a , b)) ≡ (a , c) → b ≡ c
+  ,-injectiveUIPʳ u p = ,-injective≡ʳ u p refl
+
+  ≡-dec : Decidable {A = A} _≡_ → (∀ {a} → Decidable {A = B a} _≡_) →
```
gallais, line 34: Probably worth using this opportunity to use [`DecidableEquality`](http://agda.github.io/agda-stdlib/Relation.Binary.Definitions.html#4883)
rather than the `Decidable {A = ?} _≡_` pattern.

alexarice: Sure

alexarice:
Is it worth adding aliases in the `WithK` module to make this non-breaking?

gallais:
I am not sure how common it is for people to import `X.WithK` without having
imported `X` already but I would be in favour of defaulting to being nice to
user & adding deprecation warnings anyway.

alexarice:
If someone imports `X` and `X.withK` and they both contain `\==-dec` but the one in `withK` is an alias will agda throw an error? In this case it might not be worth adding aliases

gallais:
Hmm good point.

One thing you could do is re-export the `old` definition from `X` inside of
`X.WithK` and declare the warning for it in `X.WithK` so that only the
re-exported version triggers the warning. But then users have to either be
able to drop the `X.WithK` dependency entirely or are forced to have an
explicit `using` list not containing the deprecated symbol.

Not sure that's better than just breaking in this case. @MatthewDaggitt?

alexarice:
Nothing else in the library seems to use any of the functions as far as I can tell if that influences the decision

src/Data/Product/Properties/WithK.agda
```diff
@@ -8,33 +8,14 @@
...
-open import Relation.Nullary.Reflects
-open import Relation.Nullary using (Dec; _because_; yes; no)
-open import Relation.Nullary.Decidable using (map′)
 
 ------------------------------------------------------------------------
 -- Equality
 
-module _ {a b} {A : Set a} {B : Set b} where
-
-  ,-injective : ∀ {a c : A} {b d : B} → (a , b) ≡ (c , d) → a ≡ c × b ≡ d
```
MatthewDaggitt, line 26: I would keep this proof here, I don't see it replicated in `Data.Product.Properties`

src/Data/Product/Properties.agda
```diff
@@ -23,7 +25,17 @@ module _ {a b} {A : Set a} {B : A → Set b} where
   ,-injectiveˡ : ∀ {a c} {b : B a} {d : B c} → (a , b) ≡ (c , d) → a ≡ c
   ,-injectiveˡ refl = refl
 
-  -- See also Data.Product.Properties.WithK.,-injectiveʳ.
+  ,-injective≡ʳ : ∀ {a b} {c : B a} {d : B b} → UIP A → (a , c) ≡ (b , d) → (q : a ≡ b) → subst B q c ≡ d
+  ,-injective≡ʳ {c = c} {.c} u refl q = cong (λ x → subst B x c) (u q refl)
```
MatthewDaggitt, line 29: The `.` is no longer needed

src/Data/Product/Properties.agda
```diff
@@ -23,7 +25,17 @@ module _ {a b} {A : Set a} {B : A → Set b} where
   ,-injectiveˡ : ∀ {a c} {b : B a} {d : B c} → (a , b) ≡ (c , d) → a ≡ c
   ,-injectiveˡ refl = refl
 
-  -- See also Data.Product.Properties.WithK.,-injectiveʳ.
+  ,-injective≡ʳ : ∀ {a b} {c : B a} {d : B b} → UIP A → (a , c) ≡ (b , d) → (q : a ≡ b) → subst B q c ≡ d
```
MatthewDaggitt, line 28: I think it would be more consistent with the naming conventions in the library to call this one `,-injectiveʳ-≡` and the one below `,-injectiveʳ-UIP`

src/Data/Product/Properties/WithK.agda
alexarice: Looks identical to the one in `Data.Product.Properties` to me

MatthewDaggitt: Oh yes, doh. Didn't occur to me that it was duplicated. In that case we should re-export it anyway to preserve compatibility.

alexarice: sure

alexarice:
> (with an appropriate comment that mentioned v1.4 so we can find it again when it comes to remove deprecated warnings)

I'm not sure I understand what you mean here

MatthewDaggitt:
>> (with an appropriate comment that mentioned v1.4 so we can find it again when it comes to remove deprecated warnings)
>
>I'm not sure I understand what you mean here

Ah sorry! Just a comment along the lines of "This export was deprecated in v1.4 and therefore should not be relied upon". Mainly so that when we come to remove it as a deprecated feature at some point in the future, we can find it with a quick grep for v1.4.

alexarice:
This warning should be in the `WithK` file or in the Changelog?

src/Data/Product/Properties/WithK.agda
```diff
@@ -8,33 +8,18 @@
...
-  ,-injective refl = refl , refl
+,-injective = P.,-injective
+≡-dec = P.≡-dec
```
MatthewDaggitt, line 20: Hmm don't quote me on this, but I think this will still clash if you import both modules? Even if it doesn't the more idiomatic way to do this is:
```
open import Data.Product.Properties public
  using (,-injective, ≡-dec)
```

MatthewDaggitt:
> This warning should be in the WithK file or in the Changelog?

In the `WithK` file above the `public` export

alexarice:
Think everything is fixed + Changelog added

MatthewDaggitt:
Thanks, looks good to me 👍. I'll leave it up for a few more days in case anyone else has any comments and then merge it in.

PR #1072, opened 2020-01-25
finishing out rational properties

riz0id:
- added ring definitions for unnormalised and normalised rationals as well as all of their prerequisites. 
- added morphisms for rings
- cried because numbers make me sad 

will add/remove/fix whatever just let me know

PR #1212, opened 2020-05-20
Add `+-*-commutativeRing` structure to `Data.Rational` (ℚ) via ℚᵘ

rwe:
Tiny missing `+-*-isCommutativeRing` and `+-*-commutativeRing` structures from #1072, which were included for [`ℚᵘ` (`Data.Rational.Unnormalised`)](https://github.com/agda/agda-stdlib/pull/1072/files#diff-9b7cc0a9d4127a4b2f3940952508a60aR586) but [not for `ℚ`](https://github.com/agda/agda-stdlib/pull/1072/files#diff-4d7a7de4d2025327784740eb7436dbe8R847).

The reason it was omitted is probably that it was missing from `Algebra.Morphism.RingMonomorphism`, which I've added in the same manner as `isCommutativeMonoid` is done in `Algebra.Morphism.MonoidMonomorphism`.

### DIFF
--- a/src/Algebra/Morphism/RingMonomorphism.agda
+++ b/src/Algebra/Morphism/RingMonomorphism.agda
@@ -127,3 +127,10 @@ isRing isRing = record
   ; distrib          = distrib R.+-isGroup R.*-isMagma R.distrib
   ; zero             = zero R.+-isGroup R.*-isMagma R.zero
   } where module R = IsRing isRing
+
+isCommutativeRing : IsCommutativeRing _≈₂_ _⊕_ _⊛_ ⊝_ 0#₂ 1#₂ →
+                    IsCommutativeRing _≈₁_ _+_ _*_ -_ 0# 1#
+isCommutativeRing isCommRing = record
+  { isRing = isRing C.isRing
+  ; *-comm = *-comm C.*-isMagma C.*-comm
+  } where module C = IsCommutativeRing isCommRing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,7 +103,7 @@ New modules
   ```
   Relation.Nullary.Indexed
   ```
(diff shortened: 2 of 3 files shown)


src/Algebra/Morphism/RingMonomorphism.agda
```diff
@@ -127,3 +127,9 @@ isRing isRing = record
   ; distrib          = distrib R.+-isGroup R.*-isMagma R.distrib
   ; zero             = zero R.+-isGroup R.*-isMagma R.zero
   } where module R = IsRing isRing
+
+isCommutativeRing : IsCommutativeRing _≈₂_ _⊕_ _⊛_ ⊝_ 0#₂ 1#₂ → IsCommutativeRing _≈₁_ _+_ _*_ -_ 0# 1#
```
gallais, line 131: At this point we can split the type signature on two lines, I think.

rwe: Done and fixup'd into the original commit.

MatthewDaggitt:
Looks good to me, thanks for the PR! Just a heads up that you need to run `fix-whitespace` see [HACKING.md](https://github.com/agda/agda-stdlib/blob/master/HACKING.md#how-to-enforce-whitespace-policies). Also a CHANGELOG entry would be great!

rwe:
Got it. The whitespace violations are not from this PR, they're in current master. Will fix them up as part of the changelog entry.

rwe:
Ok, please take a look! Feel free to edit changelog entries inline if necessary, this PR is open for edits by maintainers.

MatthewDaggitt:
Thanks. No further comments from me. I'll leave it up for a few days in case someone else has comments and then I'll merge it in 👍

Issue #1206, opened 2020-05-17
Definition of morphism


pnlph:
Hi, I read the [definition of morphism in nlab](https://ncatlab.org/nlab/show/morphism) and tried to find a counterpart in our `stdlib`.

When reading [the changelog v1.3](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.3.md) I found that the type `Morphism A B` has been deprecated from the modules `Algebra.Morphism.Definitions` and `Relation.Binary.Morphism.Definitions` in favor of the standard function notation `A → B`.

Could we not keep the definition of morphism, just for pedagogical purposes?

I would also propose the following formulation making explicit the source and target objects, so it can be read as the morphism from A to B:
```agda
Morphism : ∀ {a} → ∀ {b} → Set a → Set b → Set _
Morphism A B = A → B
```
Just a suggestion, don't know if it complies with the direction towards that you aim to develop the `stdlib`.

MatthewDaggitt:
The discussion that lead to it being deprecated is in https://github.com/agda/agda-stdlib/issues/1025. The feeling was that actually using the definition outside of the module was more trouble than it was worth. I guess if there are users for it, we can de-deprecate it again 😄 
I'm not sure I'd want to go back to using it extensively in the standard library though.

You're right, the problem with the current definition is that it has an extra equality argument. I think the best option would be to move it to `Function.Core` and then re-export it publicly in `Algebra.Morphism.Definitions` and `Relation.Binary.Morphism.Definitions`. Does that sound reasonable?

pnlph:
Yes, makes a lot of sense moving `Morphism` to `Function.Core`.

By publicly you mean adding this to `Algebra.Morphism.Definitions` and `Relation.Binary.Morphism.Definitions`?
```
open import Function.Core using (Morphism) public
```

pnlph:
Another question, would you make explicit the sort of the codomain?
```agda
Morphism : ∀ {a} → ∀ {b} → Set a → Set b → Set a ⊔ b
Morphism A B = A → B
```
I think it is more clean, but in that case, the `_⊔_` operator should be in scope.

MatthewDaggitt:
> By publicly you mean adding this to Algebra.Morphism.Definitions and Relation.Binary.Morphism.Definitions?

Yup, though the style guide prefers
```agda
open import Function.Core public
  using (Morphism) 
```

> Another question, would you make explicit the sort of the codomain?

[...]

PR #1213, opened 2020-05-21
added `Morphism` definition to Function.Core

pnlph:
- added `Morphism` definition to Function.Core (de-deprecation)
- imported publicly in `Algebra.Morphism.Definitions` and `Relation.Binary.Morphism.Definitions`
- removed `Algebra.Core` import from `Relation.Binary.Morphism.Definitions`
- added Morphism to Changelog Version 1.4


CHANGELOG/v1.4.md
```diff
@@ -12,15 +12,23 @@ Non-backwards compatible changes
...
+`map` now requires a function on both nodes and leaves. There are also
+`mapₙ` and `mapₗ` which only map on nodes and leaves respectively. There
+are similar changes for `Data.Tree.Relation.Unary.All`.
 
 Deprecated modules
 ------------------
 
+* In `Algebra.Morphism.Definitions` and `Relation.Binary.Morphism.Definitions`
+  the type `Morphism A B` were recovered by publicly importing its
+  definition from `Function.Core`. See discussion in issue #1206.
```
MatthewDaggitt, line 30: This would more appropriately live under `Deprecated names` rather than `Deprecated modules`

MatthewDaggitt:
If you look at the [test log](https://travis-ci.org/github/agda/agda-stdlib/jobs/689573729), it appears that it's clashing with `Morphism` elsewhere. Could you take a look a look and try and resolve it? You can run `make test` locally to check that it goes through correctly.

pnlph:
Yes, the new definition clashes with the records `Morphism` in `Category.Functor` and `Morphism` in `Category.Applicative.Index`.

I renamed the first record to `CategoryMorphism` and the second one to `ApplicativeMorphism`; similarly to the way in which e.g. `Data.Container.Indexed` declares `ContainerMorphism`.

I tried to add a `WARNING_ON_USAGE` but it was clashing when defining:
```
Morphism = CategoryMorphism
{-# WARNING_ON_USAGE returnTC
"Warning: Morphism was deprecated in v1.4.
Please use CategoryMorphism instead."
#-}
```
Is there any other way of adding warnings?

pnlph:
Still failing.

It seems like the records in `Algebra.Morphism` like `IsSemigroupMorphism`, `IsMonoidMorphism`, etc. fail to disambiguate `Morphism`, but they are actually pointing to the same definition:
```
  Algebra.Morphism.Morphism bound at
    /home/travis/build/agda/agda-stdlib/src/Function/Core.agda:27,1-9
  Function.Morphism bound at
    /home/travis/build/agda/agda-stdlib/src/Function/Core.agda:27,1-9
```

pnlph:
I finally opted for removing the import in `Algebra.Morphism.Definitions`, I also hid `Morphism` when importing `Function` in `Algebra.Morphism`.

If it looks good to you, I will update the changelog accordingly.

src/Category/Applicative/Indexed.agda
```diff
@@ -94,9 +94,10 @@ record RawIAlternative
 -- Applicative functor morphisms, specialised to propositional
 -- equality.
 
-record Morphism {I : Set i} {F₁ F₂ : IFun I f}
-                (A₁ : RawIApplicative F₁)
-                (A₂ : RawIApplicative F₂) : Set (i ⊔ suc f) where
+record ApplicativeMorphism
```
gallais, line 97: Can we keep these names & simply hide the `Morphism` import?

pnlph: Yes, you both know much better the intricacies of the `agda-stdlib` but please note that @MatthewDaggitt was [asking explicitly](https://github.com/agda/agda-stdlib/issues/1206#issuecomment-631807844) for the import.

I can change it to whatever you agree.

gallais: Re-exporting the definition from various modules is different from
hiding it when importing `Function.Core` inside of the `Category.*` modules.

I don't think it's worth introducing breaking changes for a simple type alias.

pnlph: I was hiding `Morphism` when importing `Function` in some modules, Travis is happy now :relieved:

MatthewDaggitt:
Looks good to me. The only problem is that you've added the CHANGELOG entry to the wrong file, it should be in the top-level file `CHANGELOG.md`. It's my fault as I'm not quite sure why `CHANGELOG/v1.4md` exists... it shouldn't do until we release v1.4!

PR #1217, opened 2020-05-25
More Fin properties

EdNutting:
I've added a bunch of the properties I'm using in my proof work. I don't know if these are useful for the stdlib or not - open to suggestions for improvement or just moving them into a utility library for my project.

Not entirely sure if anything has become muddled between versions of the stdlib. I hope I'm not re-introducing anything that got moved elsewhere!

Ed


src/Data/Maybe/Properties.agda
```diff
@@ -65,6 +65,13 @@ map-cong₂ : ∀ {f g : A → B} {mx} →
 map-cong₂ (just eq) = cong just eq
 map-cong₂ nothing   = refl
 
+map-injective : ∀ {f : A → B} {mx my : Maybe A}
```
MatthewDaggitt, line 68: The type can be written as `∀ {f : A → B} → Injective f _≡_ _≡_ → Injective (map f) _≡_ _≡_`

src/Data/Fin/Properties.agda
```diff
@@ -145,6 +152,24 @@ fromℕ-def : ∀ n → fromℕ n ≡ fromℕ< ℕₚ.≤-refl
 fromℕ-def zero    = refl
 fromℕ-def (suc n) = cong suc (fromℕ-def n)
 
+fromℕ<-irrelevant : ∀ m n {o} → m ≡ n
+                  → (m<o : m ℕ.< o)
```
MatthewDaggitt, line 156: Arrows should go at the end of the lines rather than the beginning of lines as per the style guide (here and below)

src/Data/Fin/Properties.agda
```diff
@@ -128,6 +131,10 @@ fromℕ-toℕ : ∀ {n} (i : Fin n) → fromℕ (toℕ i) ≡ strengthen i
 fromℕ-toℕ zero    = refl
 fromℕ-toℕ (suc i) = cong suc (fromℕ-toℕ i)
 
+≤fromℕ : ∀ {n} → (i : Fin (ℕ.suc n)) → i ≤ fromℕ n
+≤fromℕ {n} i = ℕₚ.≤-trans (ℕₚ.≤-pred (toℕ<n i))
+                           (ℕₚ.≤-reflexive (sym (toℕ-fromℕ n)))
```
MatthewDaggitt, line 136: Using `subst` instead of `ℕₚ.≤-trans` should shorten the proof.

src/Data/Fin/Properties.agda
```diff
@@ -145,6 +152,24 @@ fromℕ-def : ∀ n → fromℕ n ≡ fromℕ< ℕₚ.≤-refl
 fromℕ-def zero    = refl
 fromℕ-def (suc n) = cong suc (fromℕ-def n)
 
+fromℕ<-irrelevant : ∀ m n {o} → m ≡ n
+                  → (m<o : m ℕ.< o)
+                  → (n<o : n ℕ.< o)
+                  → fromℕ< m<o ≡ fromℕ< n<o
+fromℕ<-irrelevant .0 .0 r (s≤s z≤n) (s≤s z≤n) = refl
```
MatthewDaggitt, line 159: The dots can be removed (here and below)

src/Data/Fin/Properties.agda
```diff
@@ -365,6 +390,31 @@ toℕ-inject₁ (suc i) = cong suc (toℕ-inject₁ i)
 toℕ-inject₁-≢ : ∀ {n}(i : Fin n) → n ≢ toℕ (inject₁ i)
 toℕ-inject₁-≢ (suc i) = toℕ-inject₁-≢ i ∘ ℕₚ.suc-injective
 
+inject₁≢suc : ∀ {n} → (i : Fin n) → inject₁ i ≢ suc i
```
MatthewDaggitt, line 393: Hmm I'm not sure this one is that generally applicable? Do you have a use case for it?

src/Data/Fin/Properties.agda
```diff
@@ -365,6 +390,31 @@ toℕ-inject₁ (suc i) = cong suc (toℕ-inject₁ i)
...
+inject₁≢suc (suc i) = inject₁≢suc i ∘ suc-injective
+
+inject₁ℕ< : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.< n
+inject₁ℕ< i = ℕₚ.<-transʳ (ℕₚ.≤-reflexive (toℕ-inject₁ i)) (toℕ<n i)
```
MatthewDaggitt, line 397: Again `subst` should allow you to shorten this.

src/Data/Fin/Properties.agda
```diff
@@ -365,6 +390,31 @@ toℕ-inject₁ (suc i) = cong suc (toℕ-inject₁ i)
...
+
+inject₁ℕ≤ : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.≤ n
+inject₁ℕ≤ = ℕₚ.<⇒≤ ∘ inject₁ℕ<
+
+≤̄⇒inject₁< : ∀ {n} → {i i' : Fin n} → i' ≤ i → inject₁ i' < suc i
+≤̄⇒inject₁< {i' = i'} p = s≤s (ℕₚ.≤-trans (ℕₚ.≤-reflexive (toℕ-inject₁ i')) p)
```
MatthewDaggitt, line 403: Another `subst`

src/Data/Fin/Properties.agda
```diff
@@ -365,6 +390,31 @@ toℕ-inject₁ (suc i) = cong suc (toℕ-inject₁ i)
...
+inject₁≢suc : ∀ {n} → (i : Fin n) → inject₁ i ≢ suc i
+inject₁≢suc (suc i) = inject₁≢suc i ∘ suc-injective
+
+inject₁ℕ< : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.< n
+inject₁ℕ< i = ℕₚ.<-transʳ (ℕₚ.≤-reflexive (toℕ-inject₁ i)) (toℕ<n i)
+
+inject₁ℕ≤ : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.≤ n
+inject₁ℕ≤ = ℕₚ.<⇒≤ ∘ inject₁ℕ<
+
+≤̄⇒inject₁< : ∀ {n} → {i i' : Fin n} → i' ≤ i → inject₁ i' < suc i
```
MatthewDaggitt, line 402: Can we use `j` instead of `i'`? (here and below)

src/Function/Base.agda
```diff
@@ -50,6 +50,14 @@ _∘_ : ∀ {A : Set a} {B : A → Set b} {C : {x : A} → B x → Set c} →
       ((x : A) → C (g x))
 f ∘ g = λ x → f (g x)
 
+_∘₂_ : ∀ {A₁ : Set a} {A₂ : A₁ → Set d}
+         {B : (x : A₁) → A₂ x → Set b}
+         {C : {x : A₁} → {y : A₂ x} → B x y → Set c}
+     → ({x : A₁} → {y : A₂ x} → (z : B x y) → C z)
+     → (g : (x : A₁) → (y : A₂ x) → B x y)
+     → ((x : A₁) → (y : A₂ x) → C (g x y))
```
MatthewDaggitt, line 58: Impressive type signature!

src/Data/Fin/Properties.agda
```diff
@@ -365,6 +390,31 @@ toℕ-inject₁ (suc i) = cong suc (toℕ-inject₁ i)
...
+ℕ<⇒inject₁< : ∀ {n} → {i : Fin (ℕ.suc n)} → {i' : Fin n}
+            → toℕ i' ℕ.< toℕ i → inject₁ i' < i
+ℕ<⇒inject₁< {n} {suc i} {i'} (s≤s p) = ≤̄⇒inject₁< p
+
+------------------------------------------------------------------------
+-- lower₁
+------------------------------------------------------------------------
+
+toℕ-lower₁ : ∀ {m} x → (p : m ≢ toℕ x) → toℕ (lower₁ x p) ≡ toℕ x
+toℕ-lower₁ {ℕ.zero} zero p     = ⊥-elim (p refl)
```
MatthewDaggitt, line 414: We tend to use `contradiction` rather than `⊥-elim` for readability

CHANGELOG.md
```diff
@@ -116,6 +116,30 @@ Other minor additions
 * Added new proof to `Data.Fin.Induction`:
   ```agda
   <-wellFounded : WellFounded _<_
+        ```
+
+* Added new properties to `Data.Fin.Properties`:
+  ```agda
+  toℕ≤n : ∀ {n} → (i : Fin n) → toℕ i ℕ.≤ n
```
MatthewDaggitt, line 123: Thanks for adding the CHANGELOG entry! We tend to leave the initial implicit and explicit arguments out of the types to improve readability. You just need to give a flavour of the type. See some of the other entries for examples.

CHANGELOG.md
```diff
@@ -116,6 +116,30 @@ Other minor additions
...
+  fromℕ<-injective : ∀ m n {o} → (m<o : m ℕ.< o) → (n<o : n ℕ.< o) → fromℕ< m<o ≡ fromℕ< n<o → m ≡ n
+  inject₁≢suc : ∀ {n} → (i : Fin n) → inject₁ i ≢ suc i
+  inject₁ℕ< : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.< n
+  inject₁ℕ≤ : ∀ {n} → (i : Fin n) → toℕ (inject₁ i) ℕ.≤ n
+  ≤̄⇒inject₁< : ∀ {n} → {i i' : Fin n} → i' ≤ i → inject₁ i' < suc i
+  ℕ<⇒inject₁< : ∀ {n} → {i : Fin (ℕ.suc n)} → {i' : Fin n} → toℕ i' ℕ.< toℕ i → inject₁ i' < i
+  toℕ-lower₁ : ∀ {m} x → (p : m ≢ toℕ x) → toℕ (lower₁ x p) ≡ toℕ x
+  inject₁≡⇒lower₁≡ : ∀ {n} → {i : Fin n} → {i' : Fin (ℕ.suc n)} → (≢p : n ≢ (toℕ i')) → inject₁ i ≡ i' → lower₁ i' ≢p ≡ i
+  pred< : ∀ {n} → (i : Fin (ℕ.suc n)) → i ≢ zero → pred i < i
+  map-injective : ∀ {f : A → B} {mx my : Maybe A} → (∀ {x y : A} → f x ≡ f y → x ≡ y) → map f mx ≡ map f my → mx ≡ my
```
MatthewDaggitt, line 135: These last 3 aren't in `Data.Fin.Properties`

src/Data/Fin/Properties.agda
JacquesCarette: Are you sure you'd get the same proof? `≤` has computational content and `subst` basically acts like a permutation in cases like this.

MatthewDaggitt: Err I'm afraid I'm not quite sure what you mean? Yes it will change the computational behaviour of the proof (more efficient I think, though don't quote me on this) but the type of the resulting term will be the same no? What do you mean by "act like a permutation". A permutation on what?

EdNutting:
Great, thanks for all the feedback, I'll fix it up and then we can look at it again :)

src/Data/Fin/Properties.agda
EdNutting: I don't know about performance, but 
```agda
subst (toℕ i ℕ.≤_) (sym (toℕ-fromℕ n)) (ℕₚ.≤-pred (toℕ<n i))
```
doesn't seem much better in terms of length?

EdNutting: (I'm happy to change it anyway)

src/Data/Fin/Properties.agda
EdNutting: I have in old commits but having refactored my code and cleaned up unusued imports, I realise I'm now not using it at all - so probably not worth having.

EdNutting:
@MatthewDaggitt Okay, I've applied all the changes. Please let me know if there's anything else to tweak :)

src/Data/Fin/Properties.agda
JacquesCarette: What I meant above about `subst`: think of what `subst Fin _ (+-comm m n)` does: Not only does it change `Fin (n +m)` into `Fin (m + n)` [or the other way around?], but it does this by 'relabelling' the inhabitants of `Fin` (which is a no-op, I know). In a way, this is a lie. The witnesses that `Fin (n + m)` used to be 0..n-1 and n..n+m-1, but now 0..m-1 and m..m+n-1. In their totality, they are the same (and so the types are).  It's easiest to think with a concrete case, say n=2 and m=5. 

Perhaps a better explanation is to compare two different ways of doing this
1. `subst`
2. via the chain `Fin (n + m) ↔ Fin n ⊎ Fin m ↔ Fin m ⊎ Fin n ↔ Fin (m + n)` where the middle one actually **does** something (swaps `inj₁` and `inj₂`). That actually does permute things.

This is the sense in which `subst` is a permutation, though it is a 'virtual' permutation, that changes the meaning of the underlying elements without actually performing a computation. An elaborate, somewhat convenient, lie.

MatthewDaggitt:
@EdNutting, apologies this one somehow slipped through my net. It looks good to go so merging in :+1: Thanks for the PR!

PR #1152, opened 2020-04-06
[WIP] Added missing properties to Data.Rational.Properties

flupe:
I wanted to use rationals but realized most properties were missing, so I went ahead and implemented them. There are still quite a few missing but I'm going there.

I basically follow the same structure as in `Data.Integer.Properties`.

I was wondering whether it was reasonable to do as suggested in #921 and use instance arguments for definitions depending on sign predicates. For now these constraints are implicit, see `*-monoˡ-≤-pos : ∀ r {r>0 : r > 0ℚ} → (_* r) Preserves _≤_ ⟶ _≤_`.

My only issue so far is that I am not very fond of the naming I chose for some of the definitions.
For example, predicate `neg` can be confused with `-_` as some properties over `-_` use "neg" in their name.

If anyone has time to give feedback and suggestions, that would be much appreciated!

-------
I list below the changes:
## Data.Integer.Properties
<details>

- added `+-suc>0 : ∀ {n} → +[1+ n ] > +0`
- added `+≰- : ∀ {m n} → + m ≰ -[1+ n ]`
- added `≯⇒≤ : ∀ {x y} → x ≯ y → x ≤ y`

</details>

## Data.Rational.Unnormalised
<details>

- added `non-neg : ℚᵘ → Set`
- added `neg : ℚᵘ → Set`
- added `pos : ℚᵘ → Set` Set`
- added `non-pos : ℚᵘ → Set`

These are predicates on uninterpreted rationals. Note that they are NOT defined
as `_≥ 0`, `_< 0`, `_> 0`, `_≤ 0` as their definition was chosen to make writing proofs easier.
But they are indeed equivalent to the latter predicates. As suggested by @MatthewDaggitt in #1118 ,
I use predicates to talk about positive and non-negative rationals, rather than making them separate and deal with a bunch of conversions.

- added `ℚᵘ+ : Set`, the set of rationals satisfying `non-neg`.
- added `ℚᵘ*+ : Set`, the set of rationals satisfying `pos`.

Don't think these are very useful, but eh.

- added `mkℚᵘ+ : ∀ (n : ℕ) dm → ℚᵘ`
- added `mkℚᵘ*+ : ∀ (n : ℕ) dm → ℚᵘ`
- added `mkℚᵘ*- : ∀ (n : ℕ) dm → ℚᵘ`

Helpers to create non-neg, pos, and neg rationals.

</details>

## Data.Rational.Unnormalised.Properties
<details>

[...]

PR #1218, opened 2020-05-25
Adding properties on unnormalised rationals

flupe:
I should have made this PR a long time ago but I guess it's never too late.
Part 1 of splitting #1152 into digestible chunks to ease the review process.

This one is dedicated to properties on unnormalised rationals. I started using the new predicates @MatthewDaggitt  introduced in #850. I fixed the `Positive` predicate on integers as it stated `+0` was Positive. My French self was very happy but it's probably not what you intended.

Reviews welcome!


src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -301,32 +563,175 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
...
+                    ≡ (↥ r ℤ.* ↧ r) ℤ.* (↧ p ℤ.* ↧ q) ℤ.+ (↧ r ℤ.* ↧ r) ℤ.* (↥ p ℤ.* ↧ q)
+  lemma r p q = solve 5 (λ ↥r ↧r ↧p ↥p ↧q →
+                          (↥r :* ↧p :+ ↥p :* ↧r) :* (↧r :* ↧q) :=
+                          (↥r :* ↧r) :* (↧p :* ↧q) :+ (↧r :* ↧r) :* (↥p :* ↧q))
+                      refl (↥ r) (↧ r) (↧ p) (↥ p) (↧ q)
+    where open ℤ-solver
+
++-monoʳ-≤ : ∀ r → (r +_) Preserves _≤_ ⟶ _≤_
++-monoʳ-≤ r {p} {q} (*≤* x≤y)
+  = let r₂ = ↥ r ℤ.* ↧ r in *≤* $ begin
```
MatthewDaggitt, line 624: This might read a little better if you moved the the contents of the `let` into the `where` clause (use a `;` to write multiple assignments on one line). Otherwise this line is very hard to parse.

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -486,6 +891,20 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
...
+*-inverseˡ : ∀ p {p≢0 : ℤ.∣ ↥ p ∣ ≢0} → 1/_ p {p≢0} * p ≃ 1ℚᵘ
+*-inverseˡ p@(mkℚᵘ +[1+ n ] d) = *≡* $ cong +[1+_] $ begin
+  (n ℕ.+ d ℕ.* suc n) ℕ.* 1 ≡⟨ ℕ.*-identityʳ _ ⟩
+  (n ℕ.+ d ℕ.* suc n)       ≡⟨ cong (n ℕ.+_) (ℕ.*-suc d n) ⟩
+  (n ℕ.+ (d ℕ.+ d ℕ.* n))   ≡⟨ solve 2 (λ n d → n :+ (d :+ d :* n) := d :+ (n :+ n :* d)) refl n d ⟩
+  (d ℕ.+ (n ℕ.+ n ℕ.* d))   ≡⟨ cong (d ℕ.+_) (sym (ℕ.*-suc n d)) ⟩
+  d ℕ.+ n ℕ.* suc d         ≡˘⟨ ℕ.+-identityʳ _ ⟩
+  d ℕ.+ n ℕ.* suc d ℕ.+ 0   ∎
+  where open ≡-Reasoning; open ℕ-solver
+*-inverseˡ p@(mkℚᵘ -[1+ n ] d) = *-inverseˡ (mkℚᵘ +[1+ n ] d)
```
MatthewDaggitt, line 903: Could we move this case up above the previous one. We tend to all easier cases first!

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -514,6 +933,93 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
...
+private
+  reorder₁ : ∀ a b c d → a ℤ.* b ℤ.* (c ℤ.* d) ≡ a ℤ.* c ℤ.* b ℤ.* d
+  reorder₁ = solve 4 (λ a b c d → a :* b :* (c :* d) := a :* c :* b :* d) refl
+    where open ℤ-solver
+
+  reorder₂ : ∀ a b c d → a ℤ.* b ℤ.* (c ℤ.* d) ≡ a ℤ.* c ℤ.* (b ℤ.* d)
+  reorder₂ = solve 4 (λ a b c d → a :* b :* (c :* d) := a :* c :* (b :* d)) refl
+    where open ℤ-solver
+
+*-cancelʳ-≤-pos : ∀ {r} → Positive r → ∀ {p q} → p * r ≤ q * r → p ≤ q
```
MatthewDaggitt, line 958: Hmm I'm wondering if we need the `pos`/`neg` suffixes here? Thoughts?

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -514,6 +933,93 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
...
+  ↥ p ℤ.* +[1+ n ] ℤ.* (↧ q ℤ.* +[1+ dm ]) ≤⟨ x≤y ⟩
+  ↥ q ℤ.* +[1+ n ] ℤ.* (↧ p ℤ.* +[1+ dm ]) ≡⟨ reorder₂ (↥ q) _ _ (ℤ.+ (suc dm)) ⟩
+  l₂ ℤ.* (+[1+ n ] ℤ.* +[1+ dm ])          ∎ where open ℤ.≤-Reasoning
+
+*-cancelˡ-≤-pos : ∀ {r} → Positive r → ∀ {p q} → r * p ≤ r * q → p ≤ q
+*-cancelˡ-≤-pos {r} r>0 {p} {q}
+  rewrite *-comm-≡ r p
+        | *-comm-≡ r q = *-cancelʳ-≤-pos r>0
+
+*-monoˡ-≤-non-neg : ∀ {r} → NonNegative r → (_* r) Preserves _≤_ ⟶ _≤_
```
MatthewDaggitt, line 972: If we do decide to use them this suffix should be `nonNeg` rather than `non-neg`

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -166,11 +211,223 @@ p ≤? q = Dec.map′ *≤* drop-*≤* (↥ p ℤ.* ↧ q ℤ.≤? ↥ q ℤ.* 
...
+                                (ℤ.≤-reflexive $ ℤ.*-identityʳ n)
+
+>0⇒↥>0 : ∀ {n dm} → mkℚᵘ n dm > 0ℚᵘ → n ℤ.> 0ℤ
+>0⇒↥>0 {n} {dm} r>0 = ℤ.<-≤-trans (drop-*<* r>0)
+                                  (ℤ.≤-reflexive $ ℤ.*-identityʳ n)
+
+------------------------------------------------------------------------
+-- Properties of Positive/NonPositive/Negative/NonNegative and _≤_/_<_
+
+Positive⇒>0 : ∀ {q} → Positive q → q > 0ℚᵘ
```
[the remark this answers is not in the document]
flupe: In #1152, `positive` was named `>0⇒Positive` to make the purpose of this conversion explicit.
What's your stance on this?
If we stay with `positive` then I'm fine with `positive⁻¹`.

src/Data/Rational/Unnormalised/Properties.agda
flupe: Re: this comment and the one above, I tried being consistent with how these properties are named in `Data.Integer.Properties`. However these properties do not seem to use the predicates you introduced, and directly construct positive (or negative) integers from implicit arguments.

The way I see it, there are two ways to go about it:
- Rewrite theses properties in `Data.Integer.Properties` to use `Positive/...` predicates, but keep the naming.
- Rename here to what you're suggesting: `nonNeg` etc.

So it's really a matter of consistency, and I have no idea which path to follow.

flupe: Or we could remove the `Positive/Negative` predicates from these, and directly construct positive & negative rationals as is done in `Data.Integer.Properties`. Though he type signatures will suffer.

src/Data/Rational/Unnormalised/Properties.agda
flupe: Agreed! Made the change in many other places.

src/Data/Rational/Unnormalised/Properties.agda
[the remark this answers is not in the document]
MatthewDaggitt: I'd keen to keep `positive` as a "constructor" for `Positive`, as I imagine it'll be used rather a lot once we switch over to using the predicates for `_/_` and `_%_`, and simplicity and a lack of unicode characters will make it easier to remember and quicker to type.

src/Data/Rational/Unnormalised/Properties.agda
MatthewDaggitt: > However these properties do not seem to use the predicates you introduced

Yup, it's on the road map for when we switch over to using these predicates as instance arguments everywhere in v2.0. I think I'd also be keen to rename them to match my suggestion here to maintain consistency. Unlike switching to using predicates, this can be done in v1.4 via deprecation warnings.

src/Data/Rational/Unnormalised/Properties.agda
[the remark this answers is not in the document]
flupe: Good point, `positive⁻¹` it is!

src/Data/Rational/Unnormalised/Properties.agda
flupe: > I think I'd also be keen to rename them to match my suggestion here to maintain consistency
Perfect! I'll rename like you suggested and call it a day.

flupe:
So I made the changes and added a CHANGELOG entry.
Tell me if there's anything more to do.

Re #1219 : there is one property where it's not possible to drop the suffix: `*-monoʳ-≤` for which there
is a definition for both `Positive` and `nonNegative`. Should we still remove the suffix for all properties but this one?

Also, it should be noted that since the denominator is not constrained (not even mentioned) in the `Positive` predicate (and the others), you often have to give the implicit rational along with the proof it's positive. Don't know how to feel about that.

MatthewDaggitt:
Apologies, missed that this had been updated and not merged. As soon as Travis goes green, I'll merge it in.

PR #1220, opened 2020-05-27
[ fix #1128 ] Use the BUILTIN Maybe

### LINKED ISSUE #1128: Use the BUILTIN Maybe
Following https://github.com/agda/agda/commit/1eec63b1c5566b252c0a4a815ce1df99a772c475 we have a BUILTIN maybe type constructor.
We should re-export this one rather than defining our own. We can then get rid
of the support for `Maybe` in `Foreign/Haskell`.

This work should be done on the **experimental** branch.

gallais:


MatthewDaggitt:
Looks good! CHANGELOG entry?

gallais:
I am  getting an error:

```agda
  Resolve instance argument
    _97
      : {A.a : Level} {A = A₁ : Set A.a} →
        Coercible (List (Maybe _A_99) → List _A_99)
        (List (Maybe A₁) → List A₁)
  Candidates
    coerce-fun :
      {A.a : Level} {A = A₁ : Set A.a} {B.b : Level} {B : Set B.b}
      {C.c : Level} {C : Set C.c} {D.d : Level} {D : Set D.d}
      ⦃ _ : Coercible A₁ B ⦄ ⦃ _ : Coercible C D ⦄ →
      Coercible (B → C) (A₁ → D)
    coerce-refl : {A.a : Level} {A = A₁ : Set A.a} → Coercible A₁ A₁
```

But taking any of the two would yield the same solution: `TrustMe` so it
should not matter whether we are getting the first or the second solution.
Opinion @UlfNorell?

UlfNorell:
Surely that would depend on the choices of coerce instances for the domain and codomain? Instances are checked for equality before solving recursive goals, if I'm not mistaken.

cc @jespercockx

gallais:
Everything is eventually just `TrustMe`: http://agda.github.io/agda-stdlib/Foreign.Haskell.Coerce.html

UlfNorell:
Hm, that's strange then. Something else must be preventing the search from proceeding.

jespercockx:
When exactly are you getting this error? I'm trying to reproduce it so I can take a look at the debug output.

gallais:
@jespercockx It's in README.Foreign.Haskell on the branch for this PR.
Cf. https://travis-ci.org/github/agda/agda-stdlib/jobs/692191061

jespercockx:
From the debug output, it seems that the first candidate is a "fresh meta" (i.e. one that was created during checking of the candidate) rather than the actual term `TrustMe`. This might just be a case of a missing call to `instantiate`.

jespercockx:
There actually *is* a call to `instantiateFull` on the term before the call to `dropSameCandidates`, so now I have no idea what's actually going on. But I think it's a safe bet that this is a bug in Agda itself and not with how you're using the instance search.

jespercockx:
The issue in Agda has been fixed now.

gallais:
There's a new one (for me): it seems that I have forgotten to automatically
attach compilation pragmas to the builtin `Maybe` like we have for `List`.
So I'll need to fix that first.

gallais:
I have fixed the compilation problem but now there are issues with the
termination checker. :(

jespercockx:
You could try to fix this by adding some `{-# INLINE #-}` pragmas (see https://github.com/agda/agda/issues/4681).

gallais:
You cannot inline anonymous pattern-matching functions! I have filed agda/agda#4702

src/Foreign/Haskell/Maybe.agda
```diff
@@ -1,38 +0,0 @@
...
-
-private
-  variable
-    a : Level
-    A : Set a
-
-------------------------------------------------------------------------
--- Definition
-
-data Maybe (A : Set a) : Set a where
```
MatthewDaggitt, line 22: For backwards compatibility we should really keep this module, but mark the internal terms as deprecated and have the warning explain they're no longer needed?

PR #1223, opened 2020-05-29
Remove `coerce-refl` and replace it with instances for base types

jespercockx:
These are the requires changes to the standard library to not rely on Agda's feature of checking instance candidates for definitional equality.

### DIFF
--- a/README/Foreign/Haskell.agda
+++ b/README/Foreign/Haskell.agda
@@ -63,10 +63,10 @@ postulate
 -- We additionally benefit from the congruence rules for List, Char,
 -- Bool, and a reflexivity principle for variable A.
 
-uncons : List A → Maybe (A × List A)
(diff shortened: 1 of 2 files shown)


README/Foreign/Haskell.agda
```diff
@@ -63,10 +63,10 @@ postulate
 -- We additionally benefit from the congruence rules for List, Char,
 -- Bool, and a reflexivity principle for variable A.
 
-uncons : List A → Maybe (A × List A)
+uncons : {{Coercible A A}} → List A → Maybe (A × List A)
 uncons = coerce primUncons
```
gallais, line 67: This is a pretty terrible type signature to have.

jespercockx:
I think I've found a better solution to fix this in Agda itself so I'm closing this PR.

Issue #1198, opened 2020-05-10
Where to put a rather convenient function?


JacquesCarette:
```agda
  refl₁ : {f : A → B} (x : A) → f x ≡ f x
  refl₁ _ = P.refl
```
A lot of proofs of simple things can infer `f` easily, and `(λ _ → refl)` gets tiresome.

JacquesCarette:
(sorry about the initial blank issue, I hit 'enter' prematurely)

ajrouvoet:
Isn't this reflexivity of the pointwise function equality that is defined in the PropositionalEquality module, with the function argument implicit? Perhaps use the `refl` of the IsEquivalence instance of that equality with `_` for the function argument instead?

gallais:
I have been thinking about this for a bit and I am reminded of the `congruence`
in Idris' standard library which takes its function argument implicitly. I am
tempted to propose this design (to be generalised to all the `cong` and `subst`):

```agda
open import Level
open import Function.Base
open import Relation.Binary.PropositionalEquality

variable
  a b : Level
  A : Set a
  B : Set b

-- i for implicit
icong : ∀ {f : A → B} {x y} → x ≡ y → f x ≡ f y
icong = cong _

-- ′ for simplified à la ∘′?
icong′ : ∀ {f : A → B} x → f x ≡ f x
icong′ {f = f} x = cong f refl
```

If we want we can also play the game in the other direction by introducing:

```agda
-- e for explicit
pattern erefl x = refl {x = x}
```

JacquesCarette:
`icong` and `icong′` would definitely work too.   I would tend to create it at a low level (maybe even define these straight in `Relation.Binary.PropositionalEquality.Base`).  The other-direction game is likely worthwhile for those cases where being explicit does clarify (at least for the reader).

I won't opine on `subst` variants -- I avoid `subst` like the plague, so as a non-user, I shouldn't have much of a say on it!

MatthewDaggitt:
The proposed `icong` (as long as we [use `\prime` and not `'`](https://github.com/agda/agda-stdlib/pull/1205)!) looks good to me.

>  I am tempted to propose this design (to be generalised to all the cong and subst):

I've _never_ come across a scenario where the function argument for `subst` can be inferred? Maybe I'm using it wrong!

gallais:
I don't know if it happens often but see e.g.

```agda
open import Relation.Binary.PropositionalEquality.Core

[...]

PR #1224, opened 2020-06-01
add irefl, icong and icong'

JacquesCarette:
As per discussion on #1198, some useful combinators.  Not all are done here, as some are still being discussed.


src/Relation/Binary/PropositionalEquality/Core.agda
```diff
@@ -66,6 +66,18 @@ respʳ _∼_ refl x∼y = x∼y
...
+-- Some implicit variants of refl and cong; 'i' for implicit
+irefl : {f : A → B} (x : A) → f x ≡ f x
+irefl _ = refl
+
+icong : ∀ {f : A → B} {x y} → x ≡ y → f x ≡ f y
+icong = cong _
+
+-- ′ for simplified à la ∘′
+icong′ : ∀ {f : A → B} x → f x ≡ f x
+icong′ {f = f} x = cong f refl
```
MatthewDaggitt, line 79: I have to confess I'm struggling to see the difference between `irefl` and `icong′`...

JacquesCarette: Indeed. I didn't look hard enough when I grabbed stuff from that thread! I definitely prefer the name `irefl` though.

gallais: I think `irefl` should drop the `f` argument and be `erefl`
(for 'explicit' `refl`) and be defined via a pattern synonym
so it is available both on the left & on the right.

JacquesCarette: I don't quite understand: `irefl` is a function, so defining `erefl` as a pattern would be trying to match against a function when used on the left, no?

JacquesCarette: Also, you can't drop the `f` argument, as you want to provide an explicit input at one type, and then have the equality be at another. That's how it shows up in lots of applications.

JacquesCarette:
In various code (lots in agda-categories), I find myself writing "`\_ -> refl` all the time. I'd like to cut down a little bit on that noise and have a standard idiom for it. It occurs in stdlib a lot too.

laMudri:
I'm quite surprised that it unifies okay. Does it perform as well as `λ _ → refl`? I'm imagining that what happens during type-checking them is quite different.

JacquesCarette:
It's a single delta-expansion away. I have not noticed any slowdowns or any problems with it.

laMudri:
I'm talking about the checking of a use of `irefl`, which shouldn't involve any δ-expansions. But it will involve solving of the implicit argument `f`, which I don't see happening if we were to check `λ _ → refl` instead. But maybe you end up ultimately doing the same amount of unification?

JacquesCarette:
Ah, I misunderstood.  The point of `irefl` is that there are a lot of proof obligations of the shape `∀ {x} → C₁ [ x ] ≡ C₂ [ x ]` for contexts `C₁` and `C₂` that are equal after enough reductions once a specific `x` has been plugged in. So, yes, if figuring out `f` is a lot of work, that could be a problem.   But as Agda is non-strict, maybe that doesn't matter? Figuring out `f` is not needed to figure out that the applications of `irefl` are well-typed.

src/Relation/Binary/PropositionalEquality/Core.agda
```diff
@@ -66,6 +66,17 @@ respʳ _∼_ refl x∼y = x∼y
 resp₂ : ∀ (∼ : Rel A ℓ) → ∼ Respects₂ _≡_
 resp₂ _∼_ = respʳ _∼_ , respˡ _∼_
 
+------------------------------------------------------------------------
+-- Some variants of refl and cong; 'i' for implicit
+-- 'e' for explicit
+pattern erefl x = refl {x = x}
+
+irefl : ∀ {f : A → B} x → f x ≡ f x
```
MatthewDaggitt, line 74: Just been having another think about this... I'm happy with `erefl` as it is the explicit version of `refl`. However I'm less keen on `irefl`, as according the comment above it stands for "implicit refl" which to my mind is what the existing `refl` already is (`refl : ∀ {x} → x ≡ x`).

Is it not actually another degenerate version of `icong`? It seems to me that we have two dimensions over `cong`. The first is whether the `f` is implicit or explicit, the second is whether the equality `x ≡ y` is present or not. Under this analysis we have:

`cong :  ∀ (f : A → B) {x y} → x ≡ y → f x ≡ f y`
`icong :  ∀ {f : A → B} {x y} → x ≡ y → f x ≡ f y`
`??? : ∀ (f : A → B) x → f x ≡ f x` (doesn't exist)
`??? : ∀ {f : A → B} x → f x ≡ f x` (currently `irefl`)

I'm not quite sure what to suggest for the last two names. Obviously `cong'` and `icong'` aren't very meaningful, but we have used them before for degenerate versions of functions...

JacquesCarette: Note that the second `???` is not really a 'cong', as it does not have any equality input. ["congruence" maps equalities to equalities](https://en.wikipedia.org/wiki/Congruence_relation).  By the same reasoning, the first `???` isn't either.

Bikeshedding:
- `crefl`, for "equality of **c**ontexts around x"?
- ` refl₁` (where we started), as it abstracts out over 1 argument. We could generalize. The normal `refl` is the 0-ary version.

gallais: > Obviously cong' and icong' aren't very meaningful, but we have used them before for degenerate versions of functions...

That was my reasoning too in https://github.com/agda/agda-stdlib/issues/1198#issuecomment-626394293

JacquesCarette: The reasoning for the ' is clear. I'm not convinced by the reasoning for calling it `icong`. It's not a congruence!

DrPhil:
I think it's a bit unlucky that `icong` can mean both [`indexed congruence`](https://github.com/agda/agda-stdlib/blob/master/src/Relation/Binary/HeterogeneousEquality.agda#L103) and now also `implicit congruence`. Not sure how big a problem it would be since they would be hard to mistake for each other. I could see renaming the old one to `indexed-cong` making sense, since the new `icong` will probably see more usage.

MatthewDaggitt:
> I think it's a bit unlucky that icong can mean both indexed congruence and now also implicit congruence. Not sure how big a problem it would be since they would be hard to mistake for each other. I could see renaming the old one to indexed-cong making sense, since the new icong will probably see more usage.

I agree. In the past we've tended to use a subscript `i` suffix for indexed variants, e.g. https://github.com/agda/agda-stdlib/blob/b4dc8bd493e96e71a195fd715a6d9f092443514f/src/Relation/Binary/Indexed/Homogeneous/Structures.agda#L36 which should avoid the ambiguity. I hadn't spotted those lemmas in `HeterogeneousEquality` before. At some point they should probably be moved to an `Indexed` subdirectory and their names made consistent right?

src/Relation/Binary/PropositionalEquality/Core.agda
MatthewDaggitt: Hmm I guess I think of it as a congruence with the equality `x ≡ x` omitted.

JacquesCarette: Fine. I much prefer having this in the library than not, so `icong` it is.

JacquesCarette:
Thanks a lot for finishing this. I had it on my list of things to return to, but I've been quite caught up in formalizing (co)ends...  But I'll start using this in various proofs now.

PR #1225, opened 2020-06-02
Move `≤-pred` to Data.Nat.Properties.Base

### LINKED ISSUE #1033: Drop the dependency on `Data.Nat.Properties` from `Data.Fin.Base`?

JacquesCarette:
fixes #1033.

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -17,6 +17,7 @@ open import Data.Fin.Base
 open import Data.Fin.Patterns
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; s≤s; z≤n; _∸_)
 import Data.Nat.Properties as ℕₚ
+open import Data.Nat.Properties.Base using (≤-pred)
 open import Data.Unit using (tt)
 open import Data.Product using (∃; ∃₂; ∄; _×_; _,_; map; proj₁; uncurry; <_,_>)
 open import Data.Sum.Base using (_⊎_; inj₁; inj₂; [_,_])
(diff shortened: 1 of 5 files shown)


src/Data/Fin/Properties.agda
```diff
@@ -17,6 +17,7 @@ open import Data.Fin.Base
 open import Data.Fin.Patterns
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; s≤s; z≤n; _∸_)
 import Data.Nat.Properties as ℕₚ
+open import Data.Nat.Properties.Base using (≤-pred)
```
gallais, line 20: We are already importing `Data.Nat.Properties` here so would it not
make sense to use `≤-pred` via it? It is publically re-exported after
all!

JacquesCarette: Indeed.  Will fix.

MatthewDaggitt:
I think it'd probably be better to make this a `Core` rather than `Base` module, users shouldn't be importing this and it means we would be allowed to get rid of it later on if we find a better solution.

Issue #207, opened 2018-01-12
Set the fixity of _<?_ in Data.Nat.Properties


MatthewDaggitt:
Set the fixity of _<?_ in Data.Nat.Properties.

MatthewDaggitt:
Closed with 5f4c67eca643120a275bea894716c76cc9f2defc

gallais:
Using the new `-Wall` option which I have just introduced, I have
found a lot more mixfix definitions which do not come with a fixity
declaration:

https://gist.github.com/gallais/79386a6ba8826620dc82b3f5d5af2bde

andreasabel:
Well, bracketing constructs like `[_]` or `[_,_]` do not need a fixity, in fact, since they are neither pre/post nor infix operators, a precedence or associativity value for these does not make sense.
The warning mechanism should be adapted to disregard these.

gallais:
I've updated the gist with the new set of warnings ignoring well-bracketed mixfix decls

MatthewDaggitt:
As of Agda 2.6.0 the warnings for the missing fixities should now work properly, so this should be much more doable.

MatthewDaggitt:
Particularly egregious example in `Relation.Binary.Construct.NonStrictToStrict`

JacquesCarette:
@gallais can you update the gist please? I think I'm going to spend some of my time working on this.

gallais:
@JacquesCarette It's just a matter of running `agda` on Everything with `-Wall`.

Saransh-cpp:
Hi, (I don't know if I am doing it right) I ran -
```
agda -Wall Everything.agda
```
locally, but could not reproduce the list of warnings, instead I got an error saying -
```
/Users/saransh/Code/Mitacs/agda-stdlib/Everything.agda:11,8-18
The name of the top level module does not match the file name. The
module Everything should be defined in one of the following files:
  /Users/saransh/Code/Mitacs/agda-stdlib/src/Everything.agda
  /Users/saransh/Code/Mitacs/agda-stdlib/src/Everything.lagda
  /opt/homebrew/Cellar/agda/2.6.3/libexec/ghc-9.4.4/Agd-2.6.3-d4728884/share/lib/prim/Everything.agda
  /opt/homebrew/Cellar/agda/2.6.3/libexec/ghc-9.4.4/Agd-2.6.3-d4728884/share/lib/prim/Everything.lagda
```

Should I be running `agda` on something else or with some additional flags? Thanks!

[...]

PR #1228, opened 2020-06-05
Fixity

JacquesCarette:
A tiny part of issue #207 (about 15% of the missing cases done).


MatthewDaggitt:
Nice! I particularly like the README file :+1: 

You're going to hate me for this, but each of these probably need a changelog entry (particularly as these are breaking changes)

JacquesCarette:
Oh my, CHANGELOG entries. That does increase the workload too.  Oh well, I'll work on plowing through that as well.

gallais:
:heart:

MatthewDaggitt:
> Oh my, CHANGELOG entries. That does increase the workload too. Oh well, I'll work on plowing through that as well.

Apologies... You could add them in for these changes, and then I can merge this lot in as is. That way you don't have to do the whole thing in one go?

JacquesCarette:
That was my plan, otherwise there's just too much. I just haven't had the time to circle back.

PR #1232, opened 2020-06-09
Expose `*-zero`, `*-zeroˡ`, `*-zeroʳ` from `Data.Rational.Properties`

rwe:
Super minor additional export for consistency. `CHANGELOG.md` updated.

These were technically already available by importing and then opening the exposed `+-*-isRing` bundle (transported from `ℚᵘ`). However this was inconsistent with the other properties (`*-identity`, `*-distrib-+`, etc.) which could be imported directly from `Data.Rational.Properties` as usual.

The definition of these properties is identical to the one created in the `+-*-isRing` structure and consistent with the other declarations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -330,10 +330,14 @@ Other minor additions
   nonNegative : p ≥ 0ℚ → NonNegative p
   ```
 
-* Added proofs to `Data.Rational.Properties`:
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Thanks, looks great! I'll leave it up a couple more days in case anyone has any comments and then I'll merge it in.

PR #1234, opened 2020-06-10
Generalize Text.Format to allow custom % codes

UlfNorell:
Step 1 towards a generic printf.

Should be backwards compatible (aside from a few new names exported from `Text.Format`).

Not sure if the module name is the right one.


UlfNorell:
Closing this. I want more refactorings for #1235, and I think it makes more sense to do all of it in one go.

PR #1227, opened 2020-06-05
New module Reflection.TypeChecking.Format

UlfNorell:
Implements printf-style variants of `typeError` and `debugPrint`.

PR #1235, opened 2020-06-10
Generalize printf

UlfNorell:
Part 2 of printf generalization.

On top of #1234, can rebase when that's merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,11 @@ New modules
   Relation.Nullary.Indexed
   ```
 
+* Generic printf format specs
+  ```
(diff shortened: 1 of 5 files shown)


UlfNorell:
Yes! #1227 

There you need `A = ErrorPart`.

UlfNorell:
I'm considering separating the spec and the rendering in `PrintfSpec`. You might want multiple renderings for the same spec.

Also, the `List A → B` continuation to the generic `printf` can be debated. The alternative is to implement a map function for `Printf`.

MatthewDaggitt:
Nice preserving of backwards compatibility :+1:

Issue #278, opened 2018-04-13
Standardisation of folds


MatthewDaggitt:
At the moment naming and typing conventions for folds differ quite a lot across the various different data types. As far as I can see there are three basic types of folds present in the standard library (feel free to suggest more if I missed any!):

1. Basic folds (applicable to everything `List`, `Vec`, `Table` etc.)
  ```agda
  foldr : ∀ {a b} {A : Set a} {B : Set b} → (A → B → B) → A → ∀ {n} → Vec A n → A
  ```
2. Dependent folds (applicable to everything)
  ```agda
  foldr : ∀ {a b} {A : Set a} (B : ℕ → Set b) {m} → 
          (∀ {n} → A → B n → B (suc n)) → B zero → Vec A m → B m
  ```
3. Non-empty folds (applicable to `Vec`, `Table` etc)
  ```agda
  foldr : ∀ {a} {A : Set a} → (A → A → A) → ∀ {n} → Vec A (suc n) → A
  ```
Obviously 1. can be implemented in terms of 2. and 3. can be implemented in terms of 1.

The current state of play is:

Datatype | Name for 1. | Name for 2. | Name for 3.
--- | --- | --- | ---
List | `foldr` | - | N/A
Vec | - | `foldr` | `foldr₁`
Table | `foldr` | - | -
Product.N-ary | - | `foldr` | - | -

It would be great if we could add in some of the missing ones and come up with some consistent naming scheme for these across all data types for v1.0. 
1. In my opinion the name `foldr` should always be reserved for this "normal" fold.
2. I don't have strong opinions or any good suggestions. Maybe something like `dfoldr` or `foldrᵈ`?  
3. I would suggest `foldr₊` to represent the non-emptiness (rather than `foldr⁺` which would clash for property intoduction proofs).

gallais:
I'd suggest also adding:
```
induction : ∀ {a b} {A : Set a} (B : ℕ → Vec A n -> Set b) → 
                  (∀ {n} x xs → B n xs → B (suc n) (x ∷ xs)) → B zero [] →
                 ∀ {m} (xs : Vec A m) → B m xs
```

JacquesCarette:
My instincts for naming (in Agda) goes in the opposite direction for 1 vs 2. The standard library is quite a mixed bag in this respect, but for simpler types (sums, products), the basic operations are all deeply dependent but use the 'classical' name anyways. Then there is often a non-dependent synonym.

[...]

PR #1241, opened 2020-06-19
Data.Vec.Functional operations and properties

calintat:
I added a bunch of new functions in `Data.Vec.Functional` and also made a `Data.Vec.Functional.Properties` module. This is mostly stuff that I needed in one of my projects, but I'm happy to add more stuff if people are interested (either in this PR or a future one). 

I think `Data.Vec` and `Data.Vec.Functional` should ideally have the same functions, so people can swap one module for the other easily (like I needed to do in my project). However, some existing functions in `Data.Vec.Functional` have slightly different types compared to their analogues in `Data.Vec`. For example, the order of arguments of `remove` is different:

``` agda
-- Data.Vec.Base:
remove : ∀ {n} → Vec A (suc n) → Fin (suc n) → Vec A n

-- Data.Vec.Functional:
remove : ∀ {n} → Fin (suc n) → Vector A (suc n) → Vector A n
```

and `foldr/foldl` in `Data.Vec` take a dependent function, whereas `foldr/foldl` in `Data.Vec.Functional` take a non-dependent function. I think these functions in `Data.Vec.Functional` should be changed to match the ones in `Data.Vec`, but such changes would break backwards-compatibility. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -386,7 +386,7 @@ Other minor additions
 
 * Added new functions to `Data.Fin.Base`:
   ```agda
-  group : ∀ n k → Fin (n ℕ.* k) → Fin k × Fin n
+  quotRem : ∀ {n} k → Fin (n ℕ.* k) → Fin k × Fin n
   opposite : ∀ {n} → Fin n → Fin n
   ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,10 +148,16 @@ New modules
   ```agda
   Function.Properties.Inverse
   Function.Properties.Equivalence
+  ```
 
 * Indexed nullary relations/sets:
   ```
   Relation.Nullary.Indexed
+  ```
+
+* Properties for functional vectors:
+  ```
+  Data.Vec.Functional.Properties
   ```
 
 Other major changes
@@ -376,6 +382,36 @@ Other minor additions
 * Added new operator to `Relation.Binary`:
   ```agda
   _⇔_ : REL A B ℓ₁ → REL A B ℓ₂ → Set _
+  ```
+
+* Added new functions to `Data.Fin.Base`:
+  ```agda
+  group : ∀ n k → Fin (n ℕ.* k) → Fin k × Fin n
+  opposite : ∀ {n} → Fin n → Fin n
+  ```
+
+* Added new proofs to `Data.Fin.Properties`:
+  ```agda
+  splitAt-< : ∀ m {n} i → (i<m : toℕ i ℕ.< m) → splitAt m {n} i ≡ inj₁ (fromℕ< i<m)
+  splitAt-≥ : ∀ m {n} i → (i≥m : toℕ i ℕ.≥ m) → splitAt m {n} i ≡ inj₂ (reduce≥ i i≥m)
+  ```
+
+* Added new functions to `Data.Vec.Functional`:
+  ```agda
+  length : ∀ {n} → Vector A n → ℕ
+  insert : ∀ {n} → Vector A n → Fin (suc n) → A → Vector A (suc n)
+  updateAt : ∀ {n} → Fin n → (A → A) → Vector A n → Vector A n
+  _++_ : ∀ {m n} → Vector A m → Vector A n → Vector A (m ℕ.+ n)
+  concat : ∀ {m n} → Vector (Vector A m) n → Vector A (n ℕ.* m)
+  _>>=_ : ∀ {m n} → Vector A m → (A → Vector B n) → Vector B (m ℕ.* n)
+  unzipWith : ∀ {n} → (A → B × C) → Vector A n → Vector B n × Vector C n
+  unzip : ∀ {n} → Vector (A × B) n → Vector A n × Vector B n
+  take : ∀ m {n} → Vector A (m ℕ.+ n) → Vector A m
+  drop : ∀ m {n} → Vector A (m ℕ.+ n) → Vector A n
+  reverse : ∀ {n} → Vector A n → Vector A n
+  init : ∀ {n} → Vector A (suc n) → Vector A n
+  last : ∀ {n} → Vector A (suc n) → A
+  transpose : ∀ {m n} → Vector (Vector A n) m → Vector (Vector A m) n
   ```
 
 Refactorings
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -388,6 +394,36 @@ Other minor additions
   _⇔_ : REL A B ℓ₁ → REL A B ℓ₂ → Set _
   ```
 
+* Added new functions to `Data.Fin.Base`:
+  ```agda
+  group : ∀ n k → Fin (n ℕ.* k) → Fin k × Fin n
```
JacquesCarette, line 399: I don't think this is a good name. `factor` maybe? It is part of an important equivalence between the two given types, so it is rather important to give it an expected name.

src/Data/Fin/Base.agda
```diff
@@ -130,6 +131,14 @@ splitAt zero    i       = inj₂ i
 splitAt (suc m) zero    = inj₁ zero
 splitAt (suc m) (suc i) = Sum.map suc id (splitAt m i)
 
+-- group n k "i" = "i / k" , "i % k"
+-- This is dual to group from Data.Vec.
+
+group : ∀ n k → Fin (n ℕ.* k) → Fin k × Fin n
+group (suc n) k i with splitAt k i
+... | inj₁ j = j , zero
+... | inj₂ j = Product.map id suc (group n k j)
```
JacquesCarette, line 140: Isn't that `Product.map_2` ?

JacquesCarette: (sorry for the lack of unicode)

src/Data/Vec/Functional.agda
```diff
@@ -103,8 +122,35 @@ rearrange r xs = xs ∘ r
...
+unzipWith : ∀ {n} → (A → B × C) → Vector A n → Vector B n × Vector C n
+unzipWith f xs = proj₁ ∘ f ∘ xs , proj₂ ∘ f ∘ xs
+
 zip : ∀ {n} → Vector A n → Vector B n → Vector (A × B) n
 zip = zipWith _,_
+
+unzip : ∀ {n} → Vector (A × B) n → Vector A n × Vector B n
+unzip = unzipWith id
+
+take : ∀ m {n} → Vector A (m ℕ.+ n) → Vector A m
```
JacquesCarette, line 140: Love how `take` to `init` are all defined as "change of variables" on the indexing function, and do not change the underlying `Vector` at all. [As it should be, but that's just my opinion]

src/Data/Vec/Functional/Properties.agda
```diff
@@ -0,0 +1,265 @@
...
+             (xs ++ ys) (inject+ n i) ≡ xs i
+lookup-++ˡ {m = m} {n = n} xs ys i = cong Sum.[ xs , ys ] (Finₚ.splitAt-inject+ m n i)
+
+lookup-++ʳ : ∀ {m n} (xs : Vector A m) (ys : Vector A n) i →
+             (xs ++ ys) (raise m i) ≡ ys i
+lookup-++ʳ {m = m} {n = n} xs ys i = cong Sum.[ xs , ys ] (Finₚ.splitAt-raise m n i)
+
+module _ {m} {ys ys′ : Vector A m} where
+
+  ++-cong : ∀ {n} (xs xs′ : Vector A n) →
```
JacquesCarette, line 178: This can be generalized to two lengths, I believe. I don't think the proof even needs to change.

CHANGELOG.md
gallais: `quotRem`?

gallais:
Standardisation of folds is a long standing issue (#278).
I agree that all of the discrepancies between the order of arguments should
be solved. Could you document the one you found by opening an issue?

CHANGELOG.md
calintat: I renamed it to `quotRem` since Haskell has a similarly named function. I also made `n` implicit.

src/Data/Fin/Base.agda
calintat: Yes, thanks for the suggestion.

src/Data/Vec/Functional.agda
calintat: I agree.

src/Data/Vec/Functional/Properties.agda
calintat: Can you elaborate please?

src/Data/Vec/Functional/Properties.agda
```diff
@@ -0,0 +1,265 @@
...
+module _ {n} {xs ys : Vector A (suc n)} where
+
+  ∷-cong : head xs ≡ head ys × tail xs ≗ tail ys → xs ≗ ys
+  ∷-cong (eq , _) zero    = eq
+  ∷-cong (_ , eq) (suc i) = eq i
+
+  ∷-injective : xs ≗ ys → head xs ≡ head ys × tail xs ≗ tail ys
+  ∷-injective eq = eq zero , eq ∘ suc
+
+≗-dec : B.Decidable {A = A} _≡_ →
```
MatthewDaggitt, line 50: This can be updated to use the new `DecidableEquality` from `Relation.Binary`

src/Data/Vec/Functional/Properties.agda
```diff
@@ -0,0 +1,265 @@
...
+    a b c : Level
+    A : Set a
+    B : Set b
+    C : Set c
+
+------------------------------------------------------------------------
+
+module _ {n} {xs ys : Vector A (suc n)} where
+
+  ∷-cong : head xs ≡ head ys × tail xs ≗ tail ys → xs ≗ ys
```
MatthewDaggitt, line 43: The convention in the library is to have the curried form of these functions, and then uncurry them if need be.

src/Data/Vec/Functional/Properties.agda
MatthewDaggitt: Aren't there already two lengths? In the module parameters?

JacquesCarette: My mistake, sorry, I was reading the code incorrectly.

src/Data/Vec/Functional/Properties.agda
calintat: Done

src/Data/Vec/Functional/Properties.agda
calintat: Done

PR #1242, opened 2020-06-29
Set precedences for Float arithmetic operators.

jinwoo:
Without this, it's inconvenient to do calculations with multiple
operators.

### DIFF
--- a/src/Data/Float/Base.agda
+++ b/src/Data/Float/Base.agda
@@ -29,11 +29,11 @@ open import Agda.Builtin.Float public
   ; primFloatToWord64 to toWord
   ; primNatToFloat    to fromℕ
   -- Operations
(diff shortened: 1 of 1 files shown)


MatthewDaggitt:
Hi @jinwoo, thanks for the useful PR! Just to note that as of Agda 2.6.1 you can now set the [precedence ](https://agda.readthedocs.io/en/v2.6.1/language/module-system.html?highlight=rename#name-modifiers) of misfix operators in the rename statement itself. It would probably be cleaner to use that new syntax!

In the mean time I'll try and work out why Travis is erroring when building Agda...

jinwoo:
Good to know. Thanks, @MatthewDaggitt . I redid this PR. PTAL.

PR #1250, opened 2020-07-05
Add proofs take-step and lookup-inject≤-take

### LINKED ISSUE #1247: Lemmas related to `Data.Vec.take`
Would it be interesting to add the following lemmas?

```
take-step : ∀ n {m} x (xs : Vec A (n ℕ.+ m)) → take (suc n) (x ∷ xs) ≡ x ∷ take n xs
take-lookup-inject : ∀ (i : Fin n) (xs : Vec A (n ℕ.+ m)) → lookup xs (inject≤ i (m≤m+n _ _)) ≡ lookup (take n xs) i
```

umazalakain:
Closes #1247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -442,6 +442,13 @@ Other minor additions
   splitAt-≥ : ∀ m {n} i → (i≥m : toℕ i ℕ.≥ m) → splitAt m {n} i ≡ inj₂ (reduce≥ i i≥m)
   ```
 
+* Added new proofs to `Data.Vec.Properties`:
+  ```agda
+  unfold-take : ∀ n {m} x (xs : Vec A (n + m)) → take (suc n) (x ∷ xs) ≡ x ∷ take n xs
+  lookup-inject≤-take : ∀ m {n} (m≤m+n : m ≤ m + n) (i : Fin m) (xs : Vec A (m + n)) →
+                        lookup xs (Fin.inject≤ i m≤m+n) ≡ lookup (take m xs) i
+  ```
+
 * Added new functions to `Data.Vec.Functional`:
   ```agda
   length : ∀ {n} → Vector A n → ℕ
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -442,6 +442,13 @@ Other minor additions
   splitAt-≥ : ∀ m {n} i → (i≥m : toℕ i ℕ.≥ m) → splitAt m {n} i ≡ inj₂ (reduce≥ i i≥m)
   ```
 
+* Added new proofs to `Data.Vec.Properties`:
+  ```agda
+  take-step : ∀ n {m} x (xs : Vec A (n + m)) → take (suc n) (x ∷ xs) ≡ x ∷ take n xs
```
gallais, line 447: Looking at the other PR, I was reminded that the similar-looking
unfolding lemma for `reverse` is called `unfold-reverse` so this
one should probably be called `unfold-take`.

src/Data/Vec/Properties.agda
```diff
@@ -102,6 +109,15 @@ lookup⇒[]= (suc i) (_ ∷ xs) p    = there (lookup⇒[]= i xs p)
   []=⇒lookup∘lookup⇒[]= (x ∷ xs) (suc i) p    =
     []=⇒lookup∘lookup⇒[]= xs i p
 
+lookup-inject≤-take : ∀ m {n} (i : Fin m) (xs : Vec A (m + n)) →
+                      lookup xs (Fin.inject≤ i (m≤m+n m n)) ≡ lookup (take m xs) i
```
MatthewDaggitt, line 113: Probably best to extract the proof `m ≤ m + n` to its own variable and not assume that the proof was created via the `m≤m+n` lemma. Otherwise if someone proves the inequality a different way it's a bit of a pain for them to apply the lemma. You could still have this form as a special case with a `\prime` on the end if you wanted to this in for convenience.

umazalakain:
Thanks both for the reviews ^_^ Updated and squashed.

Issue #1249, opened 2020-07-05
Predicate `AllAcc (λ x xs → P x xs) xs` on vectors


umazalakain:
Would it be of interest to add a pointwise predicate on vectors that also depends on the tail? This is similar to `Data.List.Fresh`, but it's on vectors and it doesn't bake in the induction-recursion. It allows for vectors with fresh elements to be modelled as AllAcc \_∉\_ xs.

gallais:
We talked a bit about something like this for lists in #467;
nothing came of it but I think it makes sense to have something
like this. I would perhaps name it `AllCons` as it's applying `P`
to all cons nodes.

umazalakain:
A possible course of action is to implement this for vectors, see how it goes, then decide whether to implement it for lists, and if so whether to rewrite List# and AllPairs in terms of AllCons

umazalakain:
If we were to define `AllCons`, where would it be? Under `Data.Vec.Relation` I imagine, but `Binary` seems to usually be used for operands of the same type.

gallais:
That would be a `Data.Vec.Relation.Unary` as it's talking about a single vector.

umazalakain:
Closing in favor of `AllPairs` as in #1257

Issue #467, opened 2018-09-16
A predicate on lists that knows the tails


HuStmpHrrr:
Please suggest a proper name on this predicate. It's a reinforced `All` which I find missing.

Consider a predicate on checking `List` to be unique:

```agda
data Unique : List Carrier → Set (c ⊔ ℓ) where
  nil : Unique []
  cons : ∀ {x xs} → x ∉ xs → Unique xs → Unique (x ∷ xs)
```

This pattern is very general, and can be expressed using a more general prediate:

```agda
data Predicate (P : Carrier → List Carrier → Set ℓ₂) : List Carrier → Set (c ⊔ ℓ) where
  nil : Unique []
  cons : ∀ {x xs} → P x xs → Predicate P xs → Predicate P (x ∷ xs)
```

This is effectively,

```agda
Predicate′ P l = All (λ { [] → ⊤ ; (h ∷ t) → P h t}) (tails l)
```

but this definition is obscure and its meaning is not as immediate as an inductive definition. I have several occasions requiring this predicate so I guess I am not alone. How do you think such relation should be named?

gallais:
Given that it's equivalent to an `All` applied to `tails`... `AllTails`?

MatthewDaggitt:
See #147 as well for an alternative predicate for formulating `Unique` (and `Sorted`, `Ascending`, `Descending` etc.).

HuStmpHrrr:
wow that was a year ago. so i guess a small technical decision to make is whether we want to make it a relation between heads and tails, or adjacent pairs. I find heads and tails is better, because if we want this predicate, the chances are we want to express relations between any pairs, and it's immediate from this predicate.

gallais:
Thinking about this and II/IR, I realised this pattern can also be used to define
the inductive-(inductive/recursive) versions.

The typical example of such a list is one where `R` is a (decidable) equivalence
relation. We get the definition of fresh lists (what Jason calls `Unique`).

```agda
open import Level
open import Data.Empty
open import Data.Unit
open import Data.Bool
open import Relation.Binary using (Rel; Decidable)
open import Relation.Unary using (Pred)
open import Relation.Nullary.Decidable using (⌊_⌋)

module II {a r} {A : Set a} (R : Rel A r) where

  data ListR : Set (a ⊔ r)
  data AllR (a : A) : ListR → Set r

  data ListR where
    []     : ListR
    _∷_by_ : (a : A) (as : ListR) → AllR a as → ListR

[...]

Issue #147, opened 2017-06-29
AllPairs (or something similar)


MatthewDaggitt:
I've personally found the following predicate over lists very useful:
```agda
data AllPairs {a ℓ} {A : Set a} (_~_ : Rel A ℓ) : List A → Set (a ⊔ ℓ) where
    []  : AllPairs _~_ []
    _∷_ : ∀ {x xs} → All (x ~_) xs → AllPairs _~_ xs → AllPairs _~_ (x ∷ xs)
```
Using it you can define such useful things as:
```agda
Sorted : List A → Set _
Sorted xs = AllPairs _≤_ xs

Unique : List A → Set _
Unique xs = AllPairs _≉_ xs
```
Before I add it though, I was wondering if anyone had any opinions on:

1. The name. `AllPairs` suggests that every pair of elements in the list is related. However when the relation `_~_` is not symmetric then it only guarantees that forward pairs are related (e.g. in `a :: b :: []` it says that `a ~ b` but not that `b ~ a`). Can anyone think of a snappier or more descriptive name?

2. Its location. Should this go in `Data.List.All` or in its own file? I'm leaning towards it's own file as the `Data.List.All.Properties` file will get pretty messy.

gallais:
@bwesterb suggests `AllSubPairs` where `Sub` is a reference to the "sub" in "sublists".

bwesterb:
AllSubPairs or simply Ascending.

andreasabel:
Cannot this decomposed into a function computing a list of pairs and an application of All?
```agda
triangle : List A -> List A
triangle [] = []
triangle (x :: xs) = map (x ,_) xs ++ triangle xs
```

gallais:
This seems relevant: https://agda.github.io/agda-stdlib/Data.Vec.html#allPairs

MatthewDaggitt:
Closed with 6965180

PR #1256, opened 2020-07-07
WIP: AllCons, or vectors where every head and tail pair satisfies a predicate

umazalakain:
Proposal and work in progress, brought up in #1249. A similar proposal for lists was brought up in #467 (stale). A more specific version for lists was developed in #147 and exists in `Data.List.Relation.Unary.AllPairs`.

A proof that all the elements of a vector are distinct can be obtained with `AllCons _∉_ xs`. This comes in handy if you want to map between de Bruijn indices and fresh names.

### DIFF
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -10,7 +10,7 @@ module Data.Vec.Base where
 
 open import Data.Bool.Base
 open import Data.Nat.Base
-open import Data.Fin.Base using (Fin; zero; suc)
+open import Data.Fin.Base as Fin using (Fin; zero; suc)
 open import Data.List.Base as List using (List)
 open import Data.Product as Prod using (∃; ∃₂; _×_; _,_)
 open import Data.These.Base as These using (These; this; that; these)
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Hi @umazalakain, thanks for the PR and apologies for the delay in a response. The code itself looks good! I'm just wondering if for your use case of vectors with unique elements, it would be better to stay consistent with `Data.List` and implement it via `AllPairs`? `AllPairs` also has a much more natural predicate (i.e. a `Rel A`) instead of your `Cons P` which has to quantify over all vectors of length `n`?

umazalakain:
I guess `AllCons` is slightly more general because the recursion isn't baked in, but I cannot think of any interesting predicates that couldn't be expressed pointwise with `AllPairs`. If no one else can think of anything (@gallais?) I'm happy to follow `AllPairs`. Thanks @MatthewDaggitt for the input!

umazalakain:
Hah I almost forget about my own use case! In https://github.com/umazalakain/typing-linear-pi/blob/fromFin/src/PiCalculus/Syntax.agda#L97 I use `AllCons` to assign to each variable a natural number that corresponds to the number of times the variable is present in the tail of the context.

umazalakain:
(Turns out I was doing things in a *very* roundabout way and that use of `AllCons` is completely unnecessary)

umazalakain:
Closing in favor of  https://github.com/agda/agda-stdlib/pull/1257

MatthewDaggitt:
> (Turns out I was doing things in a very roundabout way and that use of AllCons is completely unnecessary)

Definitely know that feeling!

gallais:
> If no one else can think of anything (@gallais?)

It's fairly easy to construct an artificial example, e.g. the characterisation of
"Fibonacci vectors":

```agda
Fib : ConsP ℕ 0ℓ
Fib a []          = a ≡ 1
Fib a (x ∷ [])    = a ≡ 1
Fib a (x ∷ y ∷ _) = a ≡ x ℕ.+ y

FibVec : AllCons Fib (8 ∷ 5 ∷ 3 ∷ 2 ∷ 1 ∷ 1 ∷ [])
FibVec = refl ∷ refl ∷ refl ∷ refl ∷ refl ∷ refl ∷ []
```

The more interesting question is whether anyone has a practical use case. I guess we
can wait for one to show up and revive this PR if it ever happens.

PR #1257, opened 2020-07-14
WIP: Introduce AllPairs for vectors

umazalakain:
Mostly as taken from `Data.List.Relation.Unary.AllPairs`. I had to add a couple of proofs to `Data.Vec.Properties` and `Data.Vec.Relation.Unary.All.Properties`. The work on `Disjoint` is still missing (not sure how important that is). I believe `lookup-injective` is missing from the list counterpart, might be a useful addition.

PR #1258, opened 2020-07-14
WIP: Show that `Data.Nat.Show.charsInBase` is injective

umazalakain:
This depends on #1257. The reason why I split `showInBase` into `charsInBase` and then `fromList` is because 1) `fromList` and `toList` are *not* shown to be inverses in *safe* Agda and 2) the proof that `fromList` is injective has not yet been released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -470,6 +470,7 @@ Other minor additions
 * Added new proofs to `Data.Vec.Properties`:
   ```agda
   unfold-take : ∀ n {m} x (xs : Vec A (n + m)) → take (suc n) (x ∷ xs) ≡ x ∷ take n xs
+  unfold-drop : ∀ n {m} x (xs : Vec A (n + m)) → drop (suc n) (x ∷ xs) ≡ drop n xs
   lookup-inject≤-take : ∀ m {n} (m≤m+n : m ≤ m + n) (i : Fin m) (xs : Vec A (m + n)) →
                         lookup xs (Fin.inject≤ i m≤m+n) ≡ lookup (take m xs) i
   ```
@@ -490,6 +491,15 @@ Other minor additions
   init : ∀ {n} → Vector A (suc n) → Vector A n
   last : ∀ {n} → Vector A (suc n) → A
   transpose : ∀ {m n} → Vector (Vector A n) m → Vector (Vector A m) n
+  ```
+
+* Added new proofs to `Data.Vec.Relation.Unary.All.Properties`:
+  ```agda
+  All-swap : ∀ {xs ys} → All (λ x → All (x ~_) ys) xs → All (λ y → All (_~ y) xs) ys
+  tabulate⁺ : ∀ {f : Fin n → A} → (∀ i → P (f i)) → All P (tabulate f)
+  tabulate⁻ : ∀ {f : Fin n → A} → All P (tabulate f) → (∀ i → P (f i))
+  drop⁺ : ∀ {n} m {xs} → All P {m + n} xs → All P {n} (drop m xs)
+  take⁺ : ∀ {n} m {xs} → All P {m + n} xs → All P {m} (take m xs)
   ```
 
 Refactorings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -464,7 +464,7 @@ Other minor additions
   ```agda
   splitAt-< : ∀ m {n} i → (i<m : toℕ i ℕ.< m) → splitAt m {n} i ≡ inj₁ (fromℕ< i<m)
   splitAt-≥ : ∀ m {n} i → (i≥m : toℕ i ℕ.≥ m) → splitAt m {n} i ≡ inj₂ (reduce≥ i i≥m)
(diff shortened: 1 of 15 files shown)


src/Data/List/Properties.agda
```diff
@@ -106,6 +106,13 @@ map-compose : {g : B → C} {f : A → B} → map (g ∘ f) ≗ map g ∘ map f
 map-compose []       = refl
 map-compose (x ∷ xs) = cong (_ ∷_) (map-compose xs)
 
+map-injective : ∀ {f : A → B} (xs ys : List A) → (∀ {x y} → f x ≡ f y → x ≡ y) →
```
MatthewDaggitt, line 109: See the discussion in #1141 where we discuss using the `Injective` predicate, and putting this proof under `Data.List.Function.Map` and then re-exporting `map-injective` publicly in this file if its deemed useful. Given that PR has been sitting there for a while, I'm happy for you to go ahead and move it there in this one.

umazalakain: This `map-injective` proof uses `Data.List.Properties.∷-injective`, is this circular dependency avoidable in some way?

umazalakain: I've update the type for now :)

MatthewDaggitt: Hmm excellent point! Maybe it needs to be the other way around. Have `map-injective` in `Data.List.Properties` and then re-export it in `Data.List.Function.Map`...

Either that or have `Data.List.Properties.Core` which I'm personally against (though I know those sort of files have their supporters!)

MatthewDaggitt: Okay sounds good. I'm happy to leave it at that! I'll wait for Travis to go green and then merge it in 👍

MatthewDaggitt:
Hmm actually any chance you can merge your branch with the latest version of `master` and push so that we cancheck we're not adding any unintended changes that crept in due to changes in #1257?

umazalakain:
Done!

src/Data/List/Properties.agda
umazalakain: I guess if this kind of properties keep pilling up it might be worth revisiting this.

CHANGELOG.md
```diff
@@ -277,6 +277,7 @@ Other minor additions
 * Added new proofs to `Data.List.Properties`:
   ```agda
   reverse-injective : reverse xs ≡ reverse ys → xs ≡ ys
+  map-injective : ∀ {f : A → B} → Injective _≡_ _≡_ f → Injective _≡_ _≡_ (map f)
```
gallais, line 280: You can drop the `∀ {f : A → B} → ` from the CHANGELOG.
We usually only bother listing the interesting part of a type.

src/Data/Digit/Properties.agda
```diff
@@ -0,0 +1,36 @@
...
+open import Relation.Binary.PropositionalEquality as P using (_≡_; refl)
+open import Function using (_∘_)
+
+module Data.Digit.Properties where
+
+digitCharsUnique : Unique digitChars
+digitCharsUnique = from-yes (allPairs? (λ x y → ¬? (x Charₚ.≟ y)) digitChars)
+
+module _ (base : ℕ) where
+  module _ {base≥2 : True (2 ≤? base)} where
```
gallais, line 29: Technically it is not necessary for the `base≥2` and `base≤16` proofs to
be equal on both sides for this to be true. I wonder whether it's worth
generalising it.

Or to prove that for `toDigits` and `showDigit` these arguments are irrelevant.

umazalakain: Good catch! I prefer the former, I find it more ergonomic to use. (Also, for the latter, could we not just use the fact that all lte proofs are equal?) I've updated the PR accordingly, making `inject≤-injective` accept two distinct lte proofs as well.

src/Data/Fin/Properties.agda
```diff
@@ -464,10 +464,10 @@ inject≤-idempotent {_} {suc n} {suc k} zero    _   _   _ = refl
 inject≤-idempotent {_} {suc n} {suc k} (suc i) m≤n n≤k _ =
   cong suc (inject≤-idempotent i (ℕₚ.≤-pred m≤n) (ℕₚ.≤-pred n≤k) _)
 
-inject≤-injective : ∀ {n m} (n≤m : n ℕ.≤ m) x y → inject≤ x n≤m ≡ inject≤ y n≤m → x ≡ y
-inject≤-injective (s≤s p) zero zero eq = refl
-inject≤-injective (s≤s p) (suc x) (suc y) eq =
-  cong suc (inject≤-injective p x y (suc-injective eq))
+inject≤-injective : ∀ {n m} (n≤m n≤m′ : n ℕ.≤ m) x y → inject≤ x n≤m ≡ inject≤ y n≤m′ → x ≡ y
```
MatthewDaggitt, line 467: We should probably therefore update the CHANGELOG entry for this as well!

umazalakain:
Should be ready to merge I think  ^^

MatthewDaggitt:
Thanks for the PR!

PR #1259, opened 2020-07-19
Additional compositions with binary functions

mmontin:
This pull request proposes the addition of the composition of a binary function with two unary functions, a features which was missing, although very useful. It is defined using the composition of three binary functions, and allows a new definition of the composition of a binary function with a unary function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -517,7 +517,7 @@ Other minor additions
   _-⟨_⟩-_ : (A → C) → (C → D → E) → (B → D) → (A → B → E)
   _on₂_   : (C → C → D) → (A → B → C) → (A → B → D)
   ```
-  
+
 * Added new functions to `Data.Vec.Relation.Unary.All`:
   ```agda
   reduce : (f : ∀ {x} → P x → B) → All P xs → Vec B n
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,17 +186,17 @@ Other minor additions
 
 * Added new properties to `Data.Fin.Properties`:
   ```agda
-  toℕ≤n : (i : Fin n) → toℕ i ℕ.≤ n
-  ≤fromℕ : (i : Fin (ℕ.suc n)) → i ≤ fromℕ n
+  toℕ≤n             : (i : Fin n) → toℕ i ℕ.≤ n
+  ≤fromℕ            : (i : Fin (ℕ.suc n)) → i ≤ fromℕ n
   fromℕ<-irrelevant : m ≡ n → (m<o : m ℕ.< o) → (n<o : n ℕ.< o) → fromℕ< m<o ≡ fromℕ< n<o
-  fromℕ<-injective : fromℕ< m<o ≡ fromℕ< n<o → m ≡ n
-  inject₁ℕ< : (i : Fin n) → toℕ (inject₁ i) ℕ.< n
-  inject₁ℕ≤ : (i : Fin n) → toℕ (inject₁ i) ℕ.≤ n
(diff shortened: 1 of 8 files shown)


MatthewDaggitt:
Thanks for the PR! Looks good. Any chance you could add a CHANGELOG entry?

mmontin:
Thanks for your feedback. The CHANGELOG entry is done. Please inform me if further details on the change are needed. In the process, I took the liberty to fix the indentation of other entries, which I can undo if needed. 
On another note, I was wondering about the names of the compositions. The new name which I added with double brackets appears to be more akin to compositing with binary function, while the other with simple brackets appears more logically suited to composing with unary functions. In the current state of the library, this is the other way around and I did not change it to maintain backwards compatibility. What are your thoughts on the matter ?

MatthewDaggitt:
> On another note, I was wondering about the names of the compositions. The new name which I added with double brackets appears to be more akin to composing with binary function, while the other with simple brackets appears more logically suited to composing with unary functions. In the current state of the library, this is the other way around and I did not change it to maintain backwards compatibility. What are your thoughts on the matter ?

Hmm that's a good point. As you say swapping them would be a major incompatible change so I'm afraid that's out of the question. I guess we could choose [a different bracket](https://github.com/agda/agda/blob/master/src/data/emacs-mode/agda-input.el#L562) than the double square one to avoid the double vs unary implication? Perhaps `〈` or something similar?

gallais:
Or some floor brackets?

* Two-pronged bracket: two argument functions
* One-pronged bracket: one argument functions

(which of course opens the door for 2-arg on one side & 1-arg on the other)

mmontin:
I took both your ideas to come up with a system where the number of brackets matches the number of parameters the function on a given side of the composition has. This leads to the definition of the following compositions with a binary function:
● _-⟪_⟫-_ : Two binary functions
● _-⟪_⟩-_ : A binary function and a unary function
● _-⟨_⟫-_ : A unary function and a binary function
● _-⟨_⟩-_ : Two unary functions
● _on₂_ : A single binary function
● _on_ : A single unary function
I have put a warning on deprecation of _-[_]-_ which is to be replaced with _-⟪_⟫-_.
I am waiting your feedback to update the CHANGELOG file.
On another note, I was wondering if you agree with the fixity and the priority I set for these new compositions.

mmontin:
I added yet other compositions, when one of the two functions is the identity function.
I might have gotten carried away, although I feel like these one-sided compositions can be very useful to write neat functional terms. The new compositions are as follows:
● `_-⟪_∣` : A single binary function on the left
● `∣_⟫-_` : A single binary function on the right
● `_-⟨_∣` : A single unary function on the left
● `∣_⟩-_` : A single unary function on the right

mmontin:
After giving it some thoughts, I adapted the changelog and propagated a few changes right away. I'm still wondering if the priority and associativity I set for these new operators are optimal, but overall I think this PR is almost complete.

MatthewDaggitt:
@mmontin the extra functions look really good. My only remaining comment is that could you revert the definitions of `Injective` and `Congruent` back to the old versions? I think it's important to maximise the readability of key definitions like those.

mmontin:
It's done. Before merging this PR, I would very much enjoy having feedback on the priority of the operators I added because, to be honest, I did it mostly by mimicking what was around, which can possibly be perfected.

MatthewDaggitt:
> Before merging this PR, I would very much enjoy having feedback on the priority of the operators I added because, to be honest, I did it mostly by mimicking what was around, which can possibly be perfected.

They look fine to me. Even if it turns out that they're not, since we now have floating point fixity there won't be any problem in tweaking them later.

PR #1262, opened 2020-07-24
Sort `.gitignore` and ignore `GenerateEverything`

rwe:
Very minor annoyance fix.

I frequently notice untracked files in the `std-lib` submodule in `agda`. It's caused by the `GenerateEverything` binary being built, which isn't (and oughtn't be) checked in.
This adds it to the `.gitignore` file.

Additionall, this sorts the `.gitignore` file consistently as reqeusted in the comment, and adds an additional comment specifying the command to do so (using `sort -uf`, which means "unique and ignore case", and works the same for GNU/BSD/macOS).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,22 +1,22 @@
 # Keep this file in alphabetic order please!
-
-*~
-.*.swp
-*.agdai
+# Sort with the command `sort -uf`
 *.agda.el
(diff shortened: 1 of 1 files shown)


MatthewDaggitt:
Thanks, looks good!

PR #1263, opened 2020-07-27
Add function to reduce All into Vec

umazalakain:


MatthewDaggitt:
Looks great, thanks for the PR!

PR #1265, opened 2020-07-31
Fix level for FixPoint in Induction.WellFounded

JorisCeulemans:
I found out I couldn't use Induction.WellFounded.FixPoint.unfold-wfRec on a type family defined by well-founded recursion on the natural numbers. It should be possible now.


MatthewDaggitt:
Nice spot. Thanks for the PR! Any chance you could add a CHANGELOG entry under the bugs section?

JorisCeulemans:
Thanks! I've just updated the CHANGELOG.

PR #1267, opened 2020-08-02
documentation for subset relation on lists

doyougnu:
Hello,

I'm just getting into agda but noticed that subsets over lists were not documented even for simple usage. So I've added some simple examples to `README/`. Any comments appreciated.


README/Data/List/Relation/Binary/Subset.agda
```diff
@@ -0,0 +1,44 @@
...
+import Data.List.Membership.DecPropositional as DecPropMembership
+
+open import Data.String using (_≟_) -- Decidable equality over Strings
+open DecPropMembership _≟_          -- Open the decidable membership module
+                                    -- using Decidable ≡ over Strings
+
+
+-- Simple cases are just inductive proofs
+
+open import Data.List.Relation.Unary.Any using (there)
```
mmontin, line 29: You should place all imports on top of the file.

README/Data/List/Relation/Binary/Subset.agda
```diff
@@ -0,0 +1,44 @@
...
+
+lem₁ : ∀ {x : Set} {xs : List x} → xs ⊆ xs
+lem₁ p = p
+
+lem₂ : "A" ∷ [] ⊆ "B" ∷ "A" ∷ []
+lem₂ p = there p
+
+-- For anything more complicated then basic use cases see
```
mmontin, line 37: then -> than

README/Data/List/Relation/Binary/Subset.agda
```diff
@@ -0,0 +1,44 @@
...
+-- `Data.List.Membership.Propositional.Properties` for useful properties
```
mmontin, line 38: you should remove "for useful properties", as it's redundant

README/Data/List/Relation/Binary/Subset.agda
```diff
@@ -0,0 +1,44 @@
...
+
+lem₃ : "A" ∷  [] ⊆ "B" ∷ "A" ∷ "C" ∷ []
+lem₃ p = ∈-++⁺ˡ (there p)
+
+lem₄ : "A" ∷  [] ⊆ "A" ∷ "B" ∷ "C" ∷ []
+lem₄ p = ∈-++⁺ˡ p
```
mmontin, line 44: Instead of 3 examples where the subset is a singleton, maybe have it have several elements so that the examples are more diverse.

doyougnu: Great idea. It's pretty easy to show subset in the same order:
```
lem₄ : "B" ∷ "S" ∷ "E" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₄ p = ∈-++⁺ˡ (there (there p))
```

but I am struggling to show:
```
lem₅ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₅ p = ?
```

am I missing something like `∈-cong`? I saw `Any-cong` but have no idea how to make use of it.

mmontin: I found 2 ways to prove your property. You can prove it by using the direct definition of subsets (ie each element of the first list appears in the second one):

```
lem₆ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₆ (here refl) = there (there (there (there (here px))))
lem₆ (there (here refl)) = here refl
lem₆ (there (there (here refl))) = there (there (here refl))

Or you can use the property of insertion when the given element is not the head of the second list:

```
lem₆ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₆ (here refl) = ∈-insert ("S" ∷ "U" ∷ "B" ∷ "S" ∷ []) 
lem₆ (there (here refl)) = here refl
lem₆ (there (there (here refl))) = ∈-insert ("S" ∷ "U" ∷ [])
```

```

gallais: That being said, if you find you wish you had a specific function
when writing documentation, it is probably a good idea to add it! ;)

mmontin: You can prove it by using the direct definition of subsets (ie each element of the first list appears in the second one):

```
lem₆₀ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₆₀ (here refl) = there (there (there (there (here refl))))
lem₆₀ (there (here refl)) = here refl
lem₆₀ (there (there (here refl))) = there (there (here refl))
```
Or you can use the property of insertion when the given element is not the head of the second list:

```
lem₆₁ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₆₁ (here refl) = ∈-insert ("S" ∷ "U" ∷ "B" ∷ "S" ∷ []) 
lem₆₁ (there (here refl)) = here refl
lem₆₁ (there (there (here refl))) = ∈-insert ("S" ∷ "U" ∷ [])
```

And you can define the following operator (I have not found it anywhere in the std-lib but I've probably missed it, otherwise, it should be added):

```
_|>_ : ∀ {xs ys x} → x ∈ ys → xs ⊆ ys → x ∷ xs ⊆ ys
_|>_ x∈ys _ (here refl) = x∈ys
_|>_ _ xs⊆ys (there x∈xs) = xs⊆ys x∈xs

infixr 3 _|>_
```

Which leads to the more elegant proof:

```
lem₆₂ : "E" ∷ "S" ∷ "B" ∷ [] ⊆ "S" ∷ "U" ∷ "B" ∷ "S" ∷ "E" ∷ "T" ∷ []
lem₆₂ = ∈-insert ("S" ∷ "U" ∷ "B" ∷ "S" ∷ []) |> here refl |> ∈-insert ("S" ∷ "U" ∷ []) |> λ ()
```

gallais: Yep, let's add that to the lib!

README/Data/List/Relation/Binary/Subset.agda
```diff
@@ -0,0 +1,44 @@
...
+
+open import Data.List.Membership.Propositional.Properties using (∈-++⁺ˡ)
+open import Data.List.Relation.Binary.Subset.Propositional using (_⊆_)
+open import Data.List using (List; _∷_; [])
+
+------------------------------------------------------------------------
+-- Subset Relation
+
+-- The Subset relation is a wrapper over `Any` and so is parameterized over an
+-- equality relation. Thus to use the subset relation we must tell agda which
```
MatthewDaggitt, line 17: Capitalise "Agda"?

README/Data/List/Relation/Binary/Subset.agda
doyougnu: I've added the better examples but @gallais I am not sure which file you are referring to. I'd prefer to add the operator and the examples of its use case in this PR for cohesion.

gallais: I meant https://agda.github.io/agda-stdlib/Data.List.Relation.Binary.Subset.Setoid.html

doyougnu: I'm still struggling to add this operator to the setoid file, I feel out of my depth here. The example above works because `here refl` correctly pattern matches to tell agda that the subsets are equal. But this `refl` is based on propositional equality, and so in the setoid file I find myself unable to pattern match on `(x₁ Setoid.≈ x)`. I think I would need a function that has the type: `go : ∀ {x y xs} → x ≈ y → x ∈ xs → y ∈ xs` but am missing this in the standard library. So it is unclear to me how to proceed with adding the operator to the `.Setoid` file when it seems like it needs propositional equality.

Sorry for the spam this using `Setoid` turned out to be much harder than I expected.

gallais: No need to apologize, this is indeed slightly trickier than I anticipated.
You can add:

```agda
cast : ∀ {xs} → (_∈ xs) Respects _≈_
cast x≈y = map (trans (sym x≈y))
```

to `Data.List.Membership.Setoid` (before `mapWith∈` would be a good spot)
and then use that to make your proof go!

MatthewDaggitt: See `∈-resp-≈` in `Data.List.Membership.Setoid.Properties`

doyougnu: I chose to use `∈-resp-≈` rather than create `cast` and cause some code duplication. I've updated the README file and added `_|>_` to `Data.List.Relation.Binary.Subset.Setoid`.

README/Data/List/Relation/Binary/Subset.agda
doyougnu: done.

doyougnu:
I believe I've made the requested changes, any further comments are very much appreciated!

MatthewDaggitt:
Hi @doyougnu, apologies for leaving this so long. I've just gone through and updated this. Unfortunately I noticed the new operator you added, resulted in `List.Relation.Binary.Subset` files depending on `Properties` files from other relations. We try to avoid this because otherwise we would get to a position where importing basic definitions ends up type-checking most of the library.

I've therefore moved to `Subset.(Setoid/Propositonal).Properties`. Unfortunately this means that we can no longer use infix notation for it, so I've renamed it to match other similar proofs.

Thank you once again for the excellent addition to the README files!

doyougnu:
Hi @MatthewDaggitt no worries. I'm glad I could help and appreciate the fixes.

PR #1271, opened 2020-08-06
Add transposition decomposition for permutations

alexarice:
Adds a type for lists of transpositions and a way to evaluate them.
Also gives a way to get a list of transpositions from any permutation and a proof that the evaluation of this list is the original permutation


alexarice:
Have added a changelog entry and also lined up some equational reasoning operations and added an explaination on what lift does

alexarice:
I thought it could be just of interest as a general theorem that any permutation can be represented as a composition of transpositions. My original motivation was that I wanted to prove was stable under permutation and it would be easier to instead prove that it was stable under transposition. I didn't intend `TranspositionList` as a replacement to `Permutation`.

alexarice:
Perhaps I could add a comment to make clear that this is not a unique representation of a permutation?

MatthewDaggitt:
> it would be easier to instead prove that it was stable under transposition

Ah yes, I can see why it would be easier to prove the property by induction over the length of the list.

> Perhaps I could add a comment to make clear that this is not a unique representation of a permutation?

Yes, that sounds like a good idea.

src/Data/Fin/Permutation.agda
```diff
@@ -183,3 +223,61 @@ fromPermutation π = P.subst (Permutation _) (sym (↔⇒≡ π)) π
...
+... | true = refl
+... | false with does (k ≟ j)
+...   | false = refl
+...   | true = refl
+
+------------------------------------------------------------------------
+-- Decomposition into transpositions
+
+TranspositionList : ℕ → Set
+TranspositionList n = L.List (Fin n × Fin n)
```
MatthewDaggitt, line 252: This should probably live in it's own module `Data.Fin.Permutation.Transposition.List`?

src/Data/Fin/Permutation.agda
```diff
@@ -152,6 +156,31 @@ remove {m} {n} i π = permutation to from
     punchOut {i = πʳ i} {punchIn (πʳ i) j}                         _  ≡⟨ punchOut-punchIn (πʳ i) ⟩
     j                                                                 ∎
 
+-- lift: takes a permutation m → n and creates a permutation (suc m) → (suc n)
+-- by mapping 0 to 0 and applying the input permutation to everything else
+liftₚ : ∀ {m n} → Permutation m n → Permutation (suc m) (suc n)
```
MatthewDaggitt, line 161: Could we call this `lift₀` so that we have room for other `lifts` as well in future?

src/Data/Fin/Permutation.agda
```diff
@@ -152,6 +156,31 @@ remove {m} {n} i π = permutation to from
...
+liftₚ {m} {n} π = permutation to from
+  record
```
MatthewDaggitt, line 163: The `record` keyword usually lives on the end of the previous line

src/Data/Fin/Permutation.agda
```diff
@@ -152,6 +156,31 @@ remove {m} {n} i π = permutation to from
...
+  { left-inverse-of = left-inverse-of
+  ; right-inverse-of = right-inverse-of
+  }
+  where
+    to : Fin (suc m) → Fin (suc n)
```
MatthewDaggitt, line 168: No need to indent again after the `where`. Same comment in other functions in this file

src/Data/Fin/Permutation.agda
```diff
@@ -183,3 +223,61 @@ fromPermutation π = P.subst (Permutation _) (sym (↔⇒≡ π)) π
 
 refute : ∀ {m n} → m ≢ n → ¬ Permutation m n
 refute m≢n π = contradiction (↔⇒≡ π) m≢n
+
+lift-id : ∀ {n} (i : Fin (suc n)) → liftₚ id ⟨$⟩ʳ i ≡ i
+lift-id zero = refl
+lift-id (suc i) = refl
+
+lift-comp : ∀ {m n o} → (π : Permutation m n) → (ρ : Permutation n o)
+          → ∀ i → liftₚ π ∘ₚ liftₚ ρ ⟨$⟩ʳ i ≡ liftₚ (π ∘ₚ ρ) ⟨$⟩ʳ i
```
MatthewDaggitt, line 232: First `→` should go on the previous line according to the style guide.

src/Data/Fin/Permutation.agda
alexarice: All the existing `record` in this file are like this one, shall I change them all?

MatthewDaggitt: Yes please!

alexarice:
Think I have addressed all the comments

src/Data/Fin/Permutation.agda
```diff
@@ -183,3 +218,24 @@ fromPermutation π = P.subst (Permutation _) (sym (↔⇒≡ π)) π
 
 refute : ∀ {m n} → m ≢ n → ¬ Permutation m n
 refute m≢n π = contradiction (↔⇒≡ π) m≢n
+
+lift-id : ∀ {n} (i : Fin (suc n)) → lift₀ id ⟨$⟩ʳ i ≡ i
```
MatthewDaggitt, line 222: I think these proofs should all now be name `lift₀-id` rather than `lift-id`?

src/Data/Fin/Permutation.agda
```diff
@@ -172,6 +196,17 @@ module _ {m n} (π : Permutation (suc m) (suc n)) where
     punchIn (πʳ (πˡ i)) (remove (πˡ i) π ⟨$⟩ʳ j)  ≡⟨ cong₂ punchIn (inverseʳ π) refl ⟩
     punchIn i (remove (πˡ i) π ⟨$⟩ʳ j)            ∎
 
+  lift-remove : zero ≡ πʳ zero → ∀ i → lift₀ (remove zero π) ⟨$⟩ʳ i ≡ πʳ i
```
MatthewDaggitt, line 199: The code would probably read easier if you replaced `zero` everywhere with https://github.com/agda/agda-stdlib/blob/f831cb2bda71f0e06f2b664d006a981be8bdc853/src/Data/Fin/Patterns.agda#L16

src/Data/Fin/Permutation.agda
```diff
@@ -183,3 +218,24 @@ fromPermutation π = P.subst (Permutation _) (sym (↔⇒≡ π)) π
 
 refute : ∀ {m n} → m ≢ n → ¬ Permutation m n
 refute m≢n π = contradiction (↔⇒≡ π) m≢n
+
+lift-id : ∀ {n} (i : Fin (suc n)) → lift₀ id ⟨$⟩ʳ i ≡ i
+lift-id zero = refl
+lift-id (suc i) = refl
+
+lift-comp : ∀ {m n o} → (π : Permutation m n) → (ρ : Permutation n o) →
```
MatthewDaggitt, line 226: No need for the first couple of `→` here and in the type of the proof below.

src/Data/Fin/Permutation/Transposition/List.agda
```diff
@@ -0,0 +1,71 @@
...
+------------------------------------------------------------------------
+-- Types
+
+TranspositionList : ℕ → Set
+TranspositionList n = L.List (Fin n × Fin n)
+
+------------------------------------------------------------------------
+-- Operations on transposition lists
+
+evalₜ : ∀ {n} → TranspositionList n → Permutation′ n
```
MatthewDaggitt, line 32: The nice thing about the new file is that we can now just call these `eval` and `lift₀`

CHANGELOG/v1.4.md
```diff
@@ -154,3 +166,8 @@ Other minor additions
   ```
 
 * Added module `≤-Reasoning` to `Data.Rational.Unnormalised.Properties`
+
+* Added the function `lift₀` to `Data.Fin.Permutation` which takes a
```
MatthewDaggitt, line 170: I would just add the type here rather than a description, and you also need to add the types of the other proofs you've added.

alexarice:
Have hopefully addressed everything

MatthewDaggitt:
Thanks again for the PR @alexarice, apologies it's taken so long to merge in.

PR #1278, opened 2020-08-16
Update reflection for 2.6.2

wenkokke:
[Changes](https://github.com/agda/agda/blob/master/CHANGELOG.md#reflection) were made to reflection, but the standard library wasn't updated. I've updated my fork of the standard library to fix these changes.


wenkokke:
Apparently, my text editor changes `-+-` to `-|-`?

PR #1279, opened 2020-08-17
Added stack.yaml files to match agda/agda

wenkokke:
Added stack.yaml files for the same GHC versions Agda supports:

- stack-8.0.2.yaml
- stack-8.2.2.yaml
- stack-8.4.4.yaml
- stack-8.6.5.yaml
- stack-8.8.2.yaml
- stack-8.8.3.yaml
- stack-8.10.1.yaml

### DIFF
--- a/stack-8.0.2.yaml
+++ b/stack-8.0.2.yaml
@@ -1,5 +1,5 @@
-resolver: lts-12.26
-compiler: ghc-8.4.4
+resolver: lts-9.21
+compiler: ghc-8.0.2
 compiler-check: match-exact
 
 packages:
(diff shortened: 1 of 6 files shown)


MatthewDaggitt:
Never used `stack` before but I find it a bit surprising that you need a new file for each version? If it does require all the files, any chance they can be placed in a `stack` subdirectory rather than having them clutter up the top-level folder?

wenkokke:
I figured it would be best to mirror the [agda/agda](https://github.com/agda/agda) repository.

MatthewDaggitt:
Fair enough! I'll leave this up for a few days in case anyone with more stack knowledge wanders along. Otherwise thanks for the PR :smile:

wenkokke:
I think it's possible to override the resolver and exact compiler version from the command-line, but it's a little bit less user-friendly than just having files for separate GHC versions, and won't work at all if some particular version has different dependencies, e.g., GHC 8.10.1.

PR #1285, opened 2020-08-28
First step of float revision

wenkokke:
For when <https://github.com/agda/agda/pull/4885> is merged.

- Changed Data.Float.Base to import the new version of primitives.
- Changed Data.Float.Properties to remove the strict inequality induced via Word64.

PR #1291, opened 2020-09-14
Add module with utilities for making system calls.

wenkokke:
Adds a new module for making system calls during type checking, `Reflection.External`, which re-exports `Agda.Builtin.Reflection.External`.


MatthewDaggitt:
Looks good! Any chance you could add a brief example in `README.Reflection.External` calling some simple system utility for documentation purposes?

MatthewDaggitt:
Also it looks like it's failing because the Agda commit for the `experimental` branch hasn't been updated sufficiently. I guess best to wait until https://github.com/agda/agda-stdlib/pull/1285 is merged in, otherwise it'll just fall over with the floating point changes instead.

MatthewDaggitt:
@wenkokke I've added a README file. Is there anything I've missed?

wenkokke:
Looks good to me!

PR #1294, opened 2020-09-22
Generic traversals of reflected terms

UlfNorell:
Using it to implement weakening, strengthening and free variable check.

Targeting `experimental` since the reflected term syntax has changed. Once we're happy I can backport to master.


MatthewDaggitt:
No comments from me :+1:

PR #1296, opened 2020-09-25
Tiny build utils enhancements

rwe:
Happy to split this into multiple PRs, let me know. This contains three small changes:
  * Renames the package `lib.cabal` to `agda-stdlib-utils.cabal`, since that is `cabal install`ed and not a good global package name.
  * Fixed some tiny hlint suggestions in `GenerateEverything.hs` and `AllNonAsciiChars.hs`
  * Fixed `AllNonAsciiChars`, which was failing with "too many open files" on macOS. Some light research indicated this was caused by the lazy IO and that the easiest fix was to use the strict IO from `Data.Text`. I've verified that now works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,7 +106,7 @@ matrix:
 
         # generating Everything.agda
         - cd agda-stdlib
-        - cabal install lib.cabal
+        - cabal install agda-stdlib-utils.cabal
         - runghc GenerateEverything.hs
 
         # setting up travis-specific scripts and files
--- a/AllNonAsciiChars.hs
+++ b/AllNonAsciiChars.hs
@@ -24,8 +24,7 @@ main = do
                     "src"
   nonAsciiChars <-
     filter (not . isAscii) . concat <$> mapM readUTF8File agdaFiles
-  let table = reverse $
-              L.sortBy (compare `on` snd) $
+  let table = L.sortBy (flip compare `on` snd) $
               map (\cs -> (head cs, length cs)) $
               L.group $ L.sort $ nonAsciiChars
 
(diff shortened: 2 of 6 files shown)


gallais (approved):

rwe:
@gallais Thanks for fixing that travis break! I didn't see the notification.

rwe:
By the way, following from that discussion on merge/squashing. Normally I'd squash your last commit on top of my recent one as a fixup (+ `Co-authored-by`) and merge. I can handle that if you're comfortable with it; otherwise I'll let you or @MatthewDaggitt etc to squash or merge you see fit. No strong feelings here.

gallais:
Always happy to have everything squashed together.

MatthewDaggitt:
Thanks for the PR!7

> I can handle that if you're comfortable with it; otherwise I'll let you or @MatthewDaggitt etc to squash or merge you see fit.

I'm happy to squash and merge. It's a one-button operation in the Github UI.

I think we probably need to add a note to the CHANGELOG under the "Breaking changes" saying that we've changed the name of the cabal package right? I'm not sure if anyone currently uses it, but best to be on the safe side.

rwe:
@MatthewDaggitt - It's not published (it was just named `lib`) and only exists to declare the dependencies of the build scripts, so I would be both surprised and dismayed if someone relied on it as a package. Your call but I would not consider it a breaking change. Even the main Agda repo does not even use the cabal file here, it just does `runghc GenerateEverything.hs` directly.

> It's a one-button operation in the Github UI.
Please do! I was going to per @gallais request but I couldn't get GitHub to fill out the automatic combined commit message for the squash. Probably edited the default while "Merge pull request" was selected; couldn't get it to reset…

rwe:
Ah, actually Refined GitHub was the problem. It has a default that isn't helpful for squashing. CC @gallais may also want to check that this isn't turned on for you.

Here's what I was seeing:
Screen Shot 2020-09-28 at 1 59 52 PM

Found this toggle by snooping around:
Screen Shot 2020-09-28 at 1 58 13 PM

Which restored the more useful initial squash message:
Screen Shot 2020-09-28 at 2 00 45 PM

PR #1299, opened 2020-09-28
Generic traversals of reflected terms

UlfNorell:
Using it to implement weakening, strengthening and free variable check.

This is the `Agda-2.6.1` version of #1294 suitable for `master`.


MatthewDaggitt:
Thanks for the PR!

Issue #1297, opened 2020-09-25
Add module homomorphisms


conal:
Are there supporting definitions for module homomorphisms (likewise for left semimodule, etc)? I been searching and haven’t found them yet.

conal:
I’m not finding semiring homomorphisms either. Maybe work in progress?

gallais:
Ping @wenkokke @laMudri

laMudri:
As far as I know, there's no work in progress in the literal sense. I think with algebraic stuff, making just the definitions you need is quite easy, but developing a coherent hierarchy can be unexpectedly difficult (and also tedious).

I *think*, if I were to make a stdlib-worthy definition of module homomorphisms, I'd start with the double category of bimodules. That is, we define how to fill a square:

```
       M
  R ──────→ S
  │         │
f │    F    │ g
  ↓         ↓
  R′──────→ S′
       N
```

where R, R′, S, and S′ are (semi)rings, f and g are ring homomorphisms, and M and N are bimodules between the rings suggested in the diagram. The filler F maps elements of M to elements of N, and IIRC obeys laws `F 0 = 0`, `F (x + y) = F x + F y`, `F (r *ₗ x) = f r *ₗ F x`, and `F (x *ᵣ s) = F x *ᵣ g s`. nLab calls this an “intertwiner”, which might be a good enough name.

The usual (“globular”) definition of homomorphism between R-S-bimodules is the definition of intertwiner instantiated with `f = id {R}` and `g = id {S}`. The indirection shouldn't cause any problems, because identity homomorphisms should be definitional identities.

MatthewDaggitt:
> Are there supporting definitions for module homomorphisms (likewise for left semimodule, etc)?

As @laMudri says, no not yet. They would live under `Algebra.Module.Morphism`...

> I’m not finding semiring homomorphisms either. Maybe work in progress?

Yup, they are missing. Feel free to add them!

conal:
> Yup, they are missing. Feel free to add them\!

I’d like to give it a go.

I see that there are five abstractions to handle: `SemiringWithoutAnnihilatingZero`, `Semiring`, `CommutativeSemiring`, `Ring`, `CommutativeRing`. I can start with `SemiringWithoutAnnihilatingZero` and make a PR to ask for feedback; then when it’s looking good, add the other four. How does this plan sound?

conal:
I think these five would go into `Algebra.Morphism`.

conal:
[...]

PR #1300, opened 2020-09-29
(Near)Semiring & Lattice morphisms

conal:
Part of issue #1297. Work in progress. Feedback most appreciated (as picky as you want!).

### DIFF
--- a/src/Algebra/Bundles.agda
+++ b/src/Algebra/Bundles.agda
@@ -359,6 +359,9 @@ record RawNearSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
     ;  ε  = 0#
     }
 
+  open RawMonoid +-rawMonoid public
+    using () renaming (rawMagma to +-rawMagma)
+
   *-rawMagma : RawMagma c ℓ
   *-rawMagma = record
     { _≈_ = _≈_
@@ -493,7 +496,7 @@ record RawSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
     }
 
   open RawNearSemiring rawNearSemiring public
-    using (+-rawMonoid; *-rawMagma)
+    using (+-rawMonoid; +-rawMagma; *-rawMagma)
 
   *-rawMonoid : RawMonoid c ℓ
   *-rawMonoid = record
--- a/src/Algebra/Bundles.agda
+++ b/src/Algebra/Bundles.agda
@@ -341,6 +341,30 @@ record DistributiveLattice c ℓ : Set (suc (c ⊔ ℓ)) where
 -- Bundles with 2 binary operations & 1 element
 ------------------------------------------------------------------------
 
+record RawNearSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
+  infixl 7 _*_
+  infixl 6 _+_
+  infix  4 _≈_
+  field
+    Carrier    : Set c
+    _≈_        : Rel Carrier ℓ
+    _+_        : Op₂ Carrier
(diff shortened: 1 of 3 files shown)


src/Algebra/Bundles.agda
```diff
@@ -341,6 +341,30 @@ record DistributiveLattice c ℓ : Set (suc (c ⊔ ℓ)) where
...
+  infixl 6 _+_
+  infix  4 _≈_
+  field
+    Carrier    : Set c
+    _≈_        : Rel Carrier ℓ
+    _+_        : Op₂ Carrier
+    _*_        : Op₂ Carrier
+    0#         : Carrier
+
+  +-rawMonoid : RawMonoid c ℓ
```
MatthewDaggitt, line 355: I would probably open this publicly and export `+-rawMagma` as well?

src/Algebra/Morphism/Structures.agda
```diff
@@ -177,6 +177,90 @@ module GroupMorphisms (G₁ : RawGroup a ℓ₁) (G₂ : RawGroup b ℓ₂) wher
...
+      ; injective           = injective
+      }
+
+    *-isMagmaMonomorphism : *.IsMagmaMonomorphism ⟦_⟧
+    *-isMagmaMonomorphism = record
+      { isMagmaHomomorphism = *-isMagmaHomomorphism
+      ; injective            = injective
+      }
+
+    -- TODO: Why isRelMonomorphism specifically here?
```
MatthewDaggitt, line 233: Because it's the only thing not exported by `open IsNearSemiringHomomorphism isNearSemiringHomomorphism public` above

MatthewDaggitt (commented):
Looks good! After making `+-rawMagma` change, we can either merge this in immediately or you can add the semiring ones to this PR. If you'd prefer the former, then you'll need to add an entry to the CHANGELOG.

src/Algebra/Bundles.agda
conal: Good catch. Fixed. Thanks.

conal:
@MatthewDaggitt Thanks for the guidance! I addressed the issues you mentioned in review comments, plus a few cosmetics. Added Semiring morphisms and CHANGELOG entry. I think this PR is ready to merge if you (and others?) see fit.

conal:
Since my head is in this code, I made changes for `Lattice` as well and updated the CHANGELOG and PR name accordingly. Please let me know how this latest commit looks to you, and add review comments or merge if ready.

BTW, in `Algebra.Bundles`, I noticed that `Lattice` has a `setoid` operation, which seems useful. Do you know why other records defined in that module don’t have `setoid` (say, via `Magma`)?

MatthewDaggitt:
> Since my head is in this code, I made changes for Lattice as well and updated the CHANGELOG and PR name accordingly. Please let me know how this latest commit looks to you, and add review comments or merge if ready.

Looks good to me! Thanks for the great PR! We usually leave the PR up for a few days in case anyone else has some comments. Assuming no one does, I'll merge it in early next week :+1: 

> BTW, in Algebra.Bundles, I noticed that Lattice has a setoid operation, which seems useful. Do you know why other records defined in that module don’t have setoid (say, via Magma)?

I don't know. Seems probably like an omission. Feel free to open another PR adding it in if you would find it useful!

conal:
> BTW, in `Algebra.Bundles`, I noticed that `Lattice` has a `setoid` operation, which seems useful. Do you know why other records defined in that module don’t have `setoid` (say, via `Magma`)?

I found `setoid` in `IsMagma`.

PR #1303, opened 2020-09-30
[ fix #1302 ] Redefine `_⊖_` to use `_≤ᵇ_` and `_∸_`

### LINKED ISSUE #1302: Redefine Data.Integer._⊖_ to use Data.Nat._≤?_ and Data.Nat._∸_
This function can be defined as:
```agda
_⊖_ : ℕ → ℕ → ℤ
m ⊖ n with m ℕ.≤? n
m ⊖ n | yes p = - + (n ℕ.∸ m)
m ⊖ n | no ¬p = + (m ℕ.∸ n)
```
(although either `Data.Nat` or `Data.Nat.Properties` needs to be imported for `_≤?_`)

This would mean that all of the integer arithmetic operations are backed by built-ins, so should be a lot faster in compiled code, as well as improving the speed of compiled code using Data.Rational, too.

Making this change would require changing a lot of proofs in Data.Integer.Properties.

gallais:


Taneb:
I've completed the failing proofs in `Data.List.Membership.Propositioal.Properties` and `Data.Integer.DivMod` to get the program that prompted #1302 working again and I can confirm this is a significant speed-up.

I don't have permissions to push them here so they're in https://github.com/Taneb/agda-stdlib/tree/minus-shenanigans , feel free to grab them

src/Data/Integer/Base.agda
```diff
@@ -190,9 +191,9 @@ signAbs +[1+ n ] = Sign.+ ◂ ℕ.suc n
 -- Subtraction of natural numbers.
 
 _⊖_ : ℕ → ℕ → ℤ
-m       ⊖ ℕ.zero  = + m
-ℕ.zero  ⊖ ℕ.suc n = -[1+ n ]
-ℕ.suc m ⊖ ℕ.suc n = m ⊖ n
+m ⊖ n with m ℕ.<ᵇ n
+... | true  = - + (n ℕ.∸ m)
+... | false = + (m ℕ.∸ n)
```
MatthewDaggitt, line 196: We should add a comment about why it is defined this way and not inductively?

src/Data/Integer/Properties.agda
```diff
@@ -458,71 +461,120 @@ neg◃-cancel-< {suc m} {suc n} (-<- n<m) = s≤s n<m
...
+∣⊖∣-< {m} {n} p = begin
+  ∣ m ⊖ n ∣         ≡⟨ cong ∣_∣ (⊖-< p) ⟩
+  ∣ - (+ (n ∸ m)) ∣ ≡⟨ ∣-n∣≡∣n∣ (+ (n ∸ m)) ⟩
+  ∣ + (n ∸ m) ∣     ≡⟨⟩
+  n ∸ m             ∎ where open ≡-Reasoning
 
 ∣⊖∣-≰ : ∀ {m n} → n ℕ.≰ m → ∣ m ⊖ n ∣ ≡ n ∸ m
 ∣⊖∣-≰ = ∣⊖∣-< ∘ ℕₚ.≰⇒>
 
+-m+n≡n⊖m : ∀ m n → - (+ m) + + n ≡ n ⊖ m
```
MatthewDaggitt, line 509: These proofs should be added to the CHANGELOG

src/Data/Nat/Base.agda
```diff
@@ -39,6 +40,10 @@ open import Agda.Builtin.Nat public
 open import Agda.Builtin.Nat public
   using () renaming (_<_ to _<ᵇ_)
 
+_≤ᵇ_ : (m n : ℕ) → Bool
```
MatthewDaggitt, line 43: Need to set the fixity of this operator and add it to the CHANGELOG

src/Data/Nat/Properties.agda
```diff
@@ -104,6 +104,28 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 <⇒<ᵇ (s≤s z≤n)       = tt
 <⇒<ᵇ (s≤s (s≤s m<n)) = <⇒<ᵇ (s≤s m<n)
 
+<ᵇ-reflects-< : ∀ m n → Reflects (m < n) (m <ᵇ n)
+<ᵇ-reflects-< m n with m <ᵇ n | <ᵇ⇒< m n | <⇒<ᵇ {m} {n}
```
MatthewDaggitt, line 108: This and the `≤ᵇ-reflects-≤` lemma below feel like they follow a generic bi-implication pattern that should be extracted to a helper reflects lemma?

MatthewDaggitt (commented):
Looks good! It's missing a breaking CHANGELOG entry discussing the change and the reasons for it.

src/Data/Nat/Properties.agda
gallais: Good catch. It's literally the proof that if the test is sound & complete for `P`
then it reflects it!

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -331,16 +332,17 @@ finite inj (x ∷ xs) fᵢ∈x∷xs = excluded-middle helper
     lemma i≤j i≰1+j refl = i≰1+j (≤-step i≤j)
 
     f′ⱼ∈xs : ∀ j → f′ j ∈ xs
-    f′ⱼ∈xs j with i ≤? j
-    ... | yes i≤j = ∈-if-not-i (<⇒≢ (s≤s i≤j))
-    ... | no  i≰j = ∈-if-not-i (<⇒≢ (≰⇒> i≰j) ∘ sym)
+    f′ⱼ∈xs j with i ≤ᵇ j | Reflects.invert (≤ᵇ-reflects-≤ i j)
```
MatthewDaggitt, line 335: Hmm I'm not sure I quite understand the purpose of this change. Yes, `i ≤ᵇ j` is more efficient to pattern match on than `i ≤? j`, but don't we lose all the benefits when we call `Reflects.invert (≤ᵇ-reflects-≤ i j)` which presumably has to reconstruct the unary proof?

If there is a purpose for the change, are you proposing that all uses of `i ≤? j` should be replaced by `Reflects.invert (≤ᵇ-reflects-≤ i j)`

MatthewDaggitt (commented):
Sorry for taking so long to act on this one. Once my last comment is addressed I'll merge it in.

src/Data/List/Membership/Propositional/Properties.agda
gallais: Without this change there is a type error.
We did not go hunting for uses of `i ≤? j`, the build failed and
@Taneb fixed it.

PR #783, opened 2019-05-21
Data.Bin II

mechvel:
Intital modules Base, Properties  for the New Bin part of the library.
Declaration of Bin, decidable equality, the operations ``suc, _+_, _*_`` and some others.
The algebraic structure instances for Bin up to CommutativeSemiring, most of such are put to 
Bin.Properties.

Please, advise.

PR #1308, opened 2020-10-07
Add CancellativeCommutativeSemiring

mechvel:
Add CancellativeCommutativeSemiring.

changed :      CHANGELOG.md
changed :      src/Algebra/Bundles.agda
changed :      src/Algebra/Definitions.agda
changed :      src/Algebra/Structures.agda

record CancellativeCommutativeSemiring added.


mechvel:
The commit of CommutativeCancellativeSemiring.
This is a base for defining a general GCD notion and several proofs for it.
There are defined  LeftNZCancellative, record IsCommutativeCancellativeSemiring, 
record CommutativeCancellativeSemiring.

src/Algebra/Definitions.agda
```diff
@@ -118,5 +119,11 @@ RightCancellative _•_ = ∀ {x} y z → (y • x) ≈ (z • x) → y ≈ z
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+LeftNZCancellative : A → Op₂ A → Set _
+LeftNZCancellative 0# _*_ = ∀ {x y z} → ¬ x ≈ 0# → (x * y) ≈ (x * z) → y ≈ z
```
MatthewDaggitt, line 123: I would call this `Almost(Left/Right)Cancellative`, as nothing says that the element has to be a zero. Likewise I would just call the element `e` rather than `0#` as is used elsewhere in the file.

src/Algebra/Definitions.agda
```diff
@@ -118,5 +119,11 @@ RightCancellative _•_ = ∀ {x} y z → (y • x) ≈ (z • x) → y ≈ z
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+LeftNZCancellative : A → Op₂ A → Set _
+LeftNZCancellative 0# _*_ = ∀ {x y z} → ¬ x ≈ 0# → (x * y) ≈ (x * z) → y ≈ z 
+
+RightNZCancellative : A → Op₂ A → Set _
+RightNZCancellative 0# _*_ = ∀ {x} y z → ¬ x ≈ 0# → (y * x) ≈ (z * x) → y ≈ z 
+
```
MatthewDaggitt, line 127: It would be good to add `AlmostCancellative` as well.

src/Algebra/Structures.agda
```diff
@@ -425,6 +425,15 @@ record IsCommutativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
     }
 
 
+record IsCommutativeCancellativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ)
+  where
+  field isCommutativeSemiring : IsCommutativeSemiring + * 0# 1#
+
+  open IsCommutativeSemiring isCommutativeSemiring public
```
MatthewDaggitt, line 432: No need to put the `open` statement in the middle of the fields, can go at the end.

src/Algebra/Structures.agda
```diff
@@ -425,6 +425,15 @@ record IsCommutativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
     }
 
 
+record IsCommutativeCancellativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ)
```
MatthewDaggitt, line 428: Please follow the layout of the other record declarations in the file. The problems currently are:
 - the `where` not being on the same line
 - the fields not being indented underneath the `field` keyword

src/Algebra/Bundles.agda
```diff
@@ -670,6 +670,25 @@ record CommutativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
...
+  infix  4 _≈_
+  field
+    Carrier                           : Set c
+    _≈_                               : Rel Carrier ℓ
+    _+_                               : Op₂ Carrier
+    _*_                               : Op₂ Carrier
+    0#                                : Carrier
+    1#                                : Carrier
+    isCommutativeCancellativeSemiring : IsCommutativeCancellativeSemiring
+                                          _≈_ _+_ _*_ 0# 1#
```
MatthewDaggitt, line 685: The line is already breaching the character limit, so you might as well put the parameters on the same line.

src/Algebra/Bundles.agda
```diff
@@ -670,6 +670,25 @@ record CommutativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
     }
 
 
+record CommutativeCancellativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
```
MatthewDaggitt, line 673: Looking at the example of `IdempotentCommutativeMonoid`, if you're adding the cancellative property on top of the commutative semiring properties, this should really be called `CancellativeCommutativeSemiring` instead of `CommutativeCancellativeSemiring`. The same goes for the structure as well.

src/Algebra/Bundles.agda
```diff
@@ -670,6 +670,25 @@ record CommutativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
...
+    _+_                               : Op₂ Carrier
+    _*_                               : Op₂ Carrier
+    0#                                : Carrier
+    1#                                : Carrier
+    isCommutativeCancellativeSemiring : IsCommutativeCancellativeSemiring
+                                          _≈_ _+_ _*_ 0# 1#
+
+  open IsCommutativeCancellativeSemiring isCommutativeCancellativeSemiring public
+  
+CCancellativeSemiring = CommutativeCancellativeSemiring  -- a shorthand
```
MatthewDaggitt, line 689: We don't have shorthands for anything else in this file and I don't the proliferation of aliases helps. Please remove.

MatthewDaggitt (changes requested):
Thanks for the PR. I have some comments that need addressing. Also please can you make the title of your pull request and your commits more meaningful than "On the branch master". Something like "Add CancellativeCommutativeSemiring" would be great.

MatthewDaggitt:
Also PR has whitespace issues. Please run `fix-whitespace`.

src/Algebra/Structures.agda
mechvel: I am going to implement all your today notes.
And I have a couple of questions.
1) In Algebra.Definitions.AlmostLeftCancellative  it is desirable to use  ≉,  but I do not find where to import it from
     without obtaining cyclic module dependency.

2) In principle, it needs to be first CancellativeSemiring, and then, CancellativeCommutativeSemiring on top of it
    (with adding the *-comm axiom).
I skip the former, because a) I currently deal only with commutative case,  b) the PR goal is GCD, and standard library needs not theoretical complications related to GCD for a non-commutative case, and I do not know anything about
GCD for a non-commutative case.
Do we need to declare  CancellativeSemiring and then CancellativeCommutativeSemiring on top of it? 

3) Imagine CancellativeCommutativeSemiring with decidable equality, decidable divisibility, and trivial mutual divisibility relation. Do we need to call this record
  CancellativeCommutativeSemiring-with≟,DecDiv,trivialMutualDiv ?
 
And it will need three more levels in records, each one with a single property law, and the three Is-structures, and 
Is-structures having repetitions of several fields like, say, carrier.    
?

src/Algebra/Bundles.agda
mechvel: Please, consider my commit of October 8.
All your current notes are implemented.

MatthewDaggitt:
> In Algebra.Definitions.AlmostLeftCancellative it is desirable to use ≉, but I do not find where to import it from
without obtaining cyclic module dependency.

Yes, you cannot obtain `_≉_` here without generating a cyclic dependency so I think you're approach of using negation directly works. Alternatively if you wanted to make it a little bit neater you could define `_≉_` privately at the top of the file.

> In principle, it needs to be first CancellativeSemiring, and then, CancellativeCommutativeSemiring on top of it
(with adding the *-comm axiom). I skip the former, because a) I currently deal only with commutative case, b) the PR goal is GCD, and standard library needs not theoretical complications related to GCD for a non-commutative case, and I do not know anything about GCD for a non-commutative case. Do we need to declare CancellativeSemiring and then CancellativeCommutativeSemiring on top of it?

This is the problem of diamonds which @JacquesCarette has written extensively about, including in some your issues. At the moment such diamonds (here whether to go via commutativity or cancellativity first) are unavoidable in the Agda language. Therefore the current situation of prioritising commutativity is fine (and probably the best way to go as commutativity is a more common property than cancellativity).

> Imagine CancellativeCommutativeSemiring with decidable equality, decidable divisibility, and trivial mutual divisibility relation. Do we need to call this record CancellativeCommutativeSemiring-with≟,DecDiv,trivialMutualDiv ? And it will need three more levels in records, each one with a single property law, and the three Is-structures, and
the corresponding records in Bundles having several fields repeated like: carrier, 0#, 1# ...

I agree that would be an unpleasant name. I'm guessing that not all of those properties are suitable for bundling up in a structure though. They might be passed alongside a `CancellativeCommutativeSemiring` as separate module parameters for instance.

Please run `fix-whitespace`

src/Algebra/Structures.agda
```diff
@@ -425,6 +425,15 @@ record IsCommutativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
     }
 
 
+record IsCancellativeCommutativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
+  field
+    isCommutativeSemiring : IsCommutativeSemiring + * 0# 1#
+    cancelNZˡ             : AlmostLeftCancellative 0# *
```
MatthewDaggitt, line 431: The name of this should change to `*-cancelˡ` or `*-almostCancelˡ`

mechvel: Please, consider this point once more.
I think, ``*-cancelNonzeroˡ`` is better.
This field means the law of ``cancelling by any nonzero element from left``. It is not cancelling by a non-unity, nor cancelling by any positive element, and so on. 
For the signature, AlmostLeftCancellative is all right. But when a reader reads in the code  ``(*-cancelNonzeroˡ ...)`` it is easier to understand than  ``(*-almostCancelˡ ...)``. Because one thinks "what kind of almost?". Then, one has a chance to  guess  of this - by observing the further arguments like ``x/=0`` -  if  such is sensibly denoted. 
?

mechvel:
Please, consider my commit of October 9.
I have
1) changed the field name to   *-cancelNonzeroˡ
    (if you do not agree with my comment above, change this name as you like),
2) run fix-whitespace.

src/Algebra/Structures.agda
MatthewDaggitt: Hmm I understand your point. How about `*-cancelˡ-nonZero` or `*-cancelˡ-≉0`?

MatthewDaggitt:
> run fix-whitespace.

This doesn't appear to have worked. Please could you try again. [Line 2661](https://travis-ci.org/github/agda/agda-stdlib/jobs/734342432#L2661) suggests that there's still violations in `Algebra.Bundles` and `Algebra.Definitions`.

src/Algebra/Structures.agda
mechvel: I set there ``*-cancelˡ-nonZero``
(though I like   ``*-cancelNonzeroˡ``).
Also I run 
```
> cd agda-stdlib-fork
> fix-whitespace. 
```
(before ``make clean; make test ...``). Why does not it work completely, I wonder.
May be it is needed  ``fix-agda-whitespace`` ?
I tried now, and its messages show several fixed places.

Maybe HACKING.ms needs correction at this point?

So I am going to repeat my today's commit.

Another point: I understand this so that ``*-cancelˡ-nonZero`` is the only request of yours, so that after setting 
``*-cancelˡ-nonZero``, the commit is accepted.
Right?
So, I do now the next commit, adding  Algebra.Properties.Magma, Algebra.Properties.Monoid.
They define the divisibility relation and prove a couple of its properties.
They are needed to define irreducibiity and GCD. 
 Further there will be neded more proofs in these two modules and in
 Algebra.Properties.CommutativeMonoid, Algebra.Properties.Semiring,  
Algebra.Properties.CancellativeCommutativeSemiring.

mechvel:
I run
```
> ffix-agda-whitespace 
(it has fixed several places)
> make clean
> make test
```
It reports
/home/mechvel/agda/stLib/agda-stdlib-fork/README/Text/Tabular.agda:106,5-9
"foo bar\n  1 2  \n  4 3  " != "foo bar\n  1 2\n  4 3  " of type
String
when checking that the expression refl has type
unlines (Tabularᵛ.display whitespace (Right ∷ Left ∷ []) foobar) ≡
"foo bar\n  1 2\n  4 3  "

?

gallais:
You are using the old program to fix whitespace but we have moved
to a different one.
Cf. https://github.com/agda/agda-stdlib/blob/master/HACKING.md#installing-fix-whitespace

mechvel:
On 2020-10-12 21:40, G. Allais wrote:
> You are using the old program to fix whitespace but we have moved
> to a different one.
> Cf.
> https://github.com/agda/agda-stdlib/blob/master/HACKING.md#installing-fix-whitespace
> 

So, by running old fix-agda-whitespace I have spoiled some files in my 
stdlib clone?
And how can I undo this, how to return the source to the correct state?

gallais:
> And how can I undo this, how to return the source to the correct state?

`git checkout -p` will give you an interactive prompt asking you for each
change whether you want to revert it or keep it.

mechvel:
I have saved it to fork.sav/,  cloned `master` by new to `agda-stdlib-fork/`, then copied there from fork.sav/ all the modules that where changed in my commit (there are few of them).
Then I run ``fix-whitespace, make clean, ..., git push``. 
So, please, consider the latest PR of October 12, 2020.

MatthewDaggitt:
> Thank you, I shall reserve this git checkout -p for future.
It happened so that before reading your answer, I have done the following.
Saved it to fork.sav/, cloned master by new to agda-stdlib-fork/, then copied there from fork.sav/ all the modules that where changed in my commit (there are few of them).
Then I run fix-whitespace, make clean, ..., git push.
I hope now it is all right (?).

Please, please, **please** spend a couple of hours with a Git tutorial in your language. @gallais and I spend so much of our time trying to fix these problems for you.

Unfortunately what you've done is not correct. You haven't let me merge the reviewed cancellative semiring changes into the library before adding your new divisibility content. The divisibility changes should be made as a separate pull request.

This means that you'll have to remove the divisibility changes from this PR. If you had a basic familiarity with Git, it's simply a matter of a couple commands but at this stage I know giving you those won't help you.

mechvel:
> Add more commits by pushing to the master branch on mechvel/agda-stdlib.

Does this mean that I can continue with adding divisibility items right here?
That is I would add a couple of modules to my currently unique home repository (home tree)
(that has CancellativeCommutativeSemiring in it) and do ``git add .; git commit;  git push`` ?

MatthewDaggitt (approved):

src/Algebra/Definitions.agda
```diff
@@ -118,5 +119,14 @@ RightCancellative _•_ = ∀ {x} y z → (y • x) ≈ (z • x) → y ≈ z
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+AlmostLeftCancellative : A → Op₂ A → Set _
+AlmostLeftCancellative e _•_ = ∀ {x y z} → ¬ x ≈ e → (x • y) ≈ (x • z) → y ≈ z
+
+AlmostRightCancellative : A → Op₂ A → Set _
+AlmostRightCancellative e _•_ = ∀ {x} y z → ¬ x ≈ e → (y • x) ≈ (z • x) → y ≈ z
```
MatthewDaggitt, line 126: Just looking at this again, I notice the lack of symmetry between implict vs explicit arguments for the left and right cases.  i.e. `{x y z}` vs `{x} y z`. Whether or not `y` and `z` are inferrable depends on the definition of the `_•_`. Because we usually pattern match on the lhs arguments first in our definitions of `_+_`, `_*_` etc it's inferrable where as the rhs arguments are not. However there's no reason that users are pattern matching on the lhs first in their own operator definitions.

It seems to me that we should make these definitions symmetric, i.e. make them both `{x} y z` so they work for both lhs and rhs pattern matching definitions. I know `LeftCancellative` and `RightCancellative` don't do this but I feel like we should probably correct this mistake here.

MatthewDaggitt:
> Does this mean that I can continue with adding divisibility items right here?
That is I would add a couple of modules to my currently unique home repository (home tree)
(that has CancellativeCommutativeSemiring in it) and do git add .; git commit; git push ?

It means you _could_ but you shouldn't. Once my last comment is addressed above then I'll merge it in and only then can you add new commits about divisibility to your `master` branch of your fork.

Alternatively if you don't want to wait until this one is merged in, you could learn about how to use "branches" in Git and then you can submit multiple pull requests at the same time (as long as they don't interfere with each other).

src/Algebra/Definitions.agda
mechvel: I agree.
How do we organize this change?

I am confused with the state of the commit. 
1. There was a message  "Merge branch 'master' into master.   Verified  v 9303d9d",
2. I do not think that this 9303d9d is already in master. (?)
3. Now, I do not know who needs to push this change for "{x} y z" there.
 
When you suggest something of such sort can you, please, add such details? 
Because when I ask about a detail, I need to wait at least one day for answer.
So each line of the program takes 1-2 days processing.
And a simple project of one week would take 1-2 years to agree.

Is not it better to put at least half of the GCD project at a single PR ?
I would add about 5 not so large modules to Algebra.Properties and 2-3 records to Algebra.Bundles.
Also I would write in what succession to consider this.
May be this way there will be less questions for me of when to push each current thing - ?

mechvel:
I recall that the axiom  
 ```
0≉1 : 0# ≉ 1# 
```
needs to be added to IsCancellativeCommutativeSemiring.
Without it, GCD is senseless to consider. And I recall that  0# ≉ 1#  cannot be derived from Algebra.Semiring.
The condition of  0# ≉ 1# means a nonzero semiring. Because if  0# ≈ 1#, then all elements equal 0#.

src/Algebra/Definitions.agda
MatthewDaggitt: > How do we organize this change?

I have done it.

> Is not it better to put at least half of the GCD project at a single PR ?
I would add about 5 not so large modules to Algebra.Properties and 2-3 records to Algebra.Bundles.
Also I would write in what succession to consider this.
May be this way there will be less questions for me of when to push each current thing - ?

No it is not. Let's look at some your past large PRs.
- Adding new binary definitions: https://github.com/agda/agda-stdlib/pull/783 took 3 months to merge in
- Adding orderings over binary naturals: https://github.com/agda/agda-stdlib/pull/840 took 3 months to merge in.
- Adding subtraction over binary naturals: https://github.com/agda/agda-stdlib/pull/888 - https://github.com/agda/agda-stdlib/pull/1127 took **7** months to merge in.

In contrast this PR has taken 10 days to merge in. Hopefully think the numbers speak for themselves! Please continue to make small PRs.

MatthewDaggitt: > But I do not know, I recall, may be I can do this myself by making a branch ...

Yes you could do this with branches, and open many small independent PRs at the same time. That would indeed be quicker.

MatthewDaggitt:
> I recall that the axiom `0≉1 : 0# ≉ 1#`  needs to be added to IsCancellativeCommutativeSemiring.

The trivial semiring with one element is a valid cancellative commutative semiring. Therefore that axiom should not be added to cancellative commutative semirings in general. If it is truly necessary then I think it would be better passed in as a module argument alongside the semiring to your generic GCD module.

I'm now happy with this PR so I'm merging it in. Please submit your next small PR.

PR #1312, opened 2020-10-13
Add version to library name

neosimsim:
Since Agda supports [library versioning](https://agda.readthedocs.io/en/v2.6.1.1/tools/package-system.html#version-numbers), I suggest the standard-library should make use of it.


neosimsim:
Is there some post release check box I have to edit as a reminder to bump the version after a release?

gallais:
That'd be the [release guide](https://github.com/agda/agda-stdlib/blob/master/notes/release-guide.txt)

neosimsim:
Oh, in notes. Excuse the question.

neosimsim:
I unsure when to update the version. I guess it should be done at the same time lib.cabal is update.

MatthewDaggitt:
Thanks!

PR #1317, opened 2020-10-15
Add some properties relating to List.concat

Taneb:
I was looking for these when trying to prove that `List` is a `Monad` for `categories-examples`.

We might desire a property similar to `concat-[-]` with type `concat ∘ [_] ≗ id` but that turns out to be `++-identityʳ`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ Other minor additions
 
 * Added new properties to `Data.List.Properties`:
   ```agda
-<<<<<<< HEAD
+  concat-++ : concat xss ++ concat yss ≡ concat (xss ++ yss)
   concat-concat : concat ∘ map concat ≗ concat ∘ concat
   concat-[-] : concat ∘ map [_] ≗ id
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,9 @@ Other minor additions
 
 * Added new properties to `Data.List.Properties`:
   ```agda
+<<<<<<< HEAD
   concat-concat : concat ∘ map concat ≗ concat ∘ concat
+  concat-[-] : concat ∘ map [_] ≗ id
   ```
 
 * Add version to library name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,4 +56,9 @@ Other minor additions
   IsLatticeIsomorphism   (⟦_⟧ : A → B) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
   ```
 
+* Added new properties to `Data.List.Properties`:
+  ```agda
+  concat-concat : concat ∘ map concat ≗ concat ∘ concat
+  ```
+
 * Add version to library name
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -496,11 +496,17 @@ concat-map {f = f} xss = begin
   foldr (λ ys → map f ys ++_) [] xss         ≡⟨ sym (foldr-fusion (map f) [] (map-++-commute f) xss) ⟩
   map f (concat xss)                         ∎
 
+concat-++ : (xss yss : List (List A)) → concat xss ++ concat yss ≡ concat (xss ++ yss)
+concat-++ [] yss = refl
+concat-++ ([] ∷ xss) yss = concat-++ xss yss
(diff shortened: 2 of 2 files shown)


src/Data/List/Properties.agda
```diff
@@ -496,6 +496,16 @@ concat-map {f = f} xss = begin
...
 
+concat-concat : concat {a} {A} ∘ map concat ≗ concat ∘ concat
+concat-concat [] = refl
+concat-concat ([] ∷ xsss) = concat-concat xsss
+concat-concat (([] ∷ xss) ∷ xsss) = concat-concat (xss ∷ xsss)
+concat-concat (((x ∷ xs) ∷ xss) ∷ xsss) = cong (x ∷_) (concat-concat ((xs ∷ xss) ∷ xsss))
+
+concat-[-] : concat {a} {A} ∘ map [_] ≗ id
+concat-[-] [] = refl
+concat-[-] (x ∷ xs) = {!cong (x ∷_) (concat-[-] xs)!}
```
gallais, line 507: You've left a hole

src/Data/List/Properties.agda
```diff
@@ -496,6 +496,16 @@ concat-map {f = f} xss = begin
   foldr (λ ys → map f ys ++_) [] xss         ≡⟨ sym (foldr-fusion (map f) [] (map-++-commute f) xss) ⟩
   map f (concat xss)                         ∎
 
+concat-concat : concat {a} {A} ∘ map concat ≗ concat ∘ concat
+concat-concat [] = refl
+concat-concat ([] ∷ xsss) = concat-concat xsss
+concat-concat (([] ∷ xss) ∷ xsss) = concat-concat (xss ∷ xsss)
+concat-concat (((x ∷ xs) ∷ xss) ∷ xsss) = cong (x ∷_) (concat-concat ((xs ∷ xss) ∷ xsss))
+
```
gallais, line 504: Feels like these last cases could be made cleaner by first proving
`concat-++ : concat (xs ++ ys) ≡ concat xs ++ concat ys`
and then using equational reasoning.

src/Data/List/Properties.agda
Taneb: Ack, I forgot to save before committing... I'll sort that out

CHANGELOG.md
```diff
@@ -56,4 +56,11 @@ Other minor additions
   IsLatticeIsomorphism   (⟦_⟧ : A → B) : Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
   ```
 
+* Added new properties to `Data.List.Properties`:
+  ```agda
+  concat-++ : concat xss ++ concat yss ≡ concat (xss ++ yss)
+  concat-concat : concat-concat : concat ∘ map concat ≗ concat ∘ concat
```
MatthewDaggitt, line 62: Repeated name on this line

src/Data/List/Properties.agda
```diff
@@ -496,6 +496,22 @@ concat-map {f = f} xss = begin
   foldr (λ ys → map f ys ++_) [] xss         ≡⟨ sym (foldr-fusion (map f) [] (map-++-commute f) xss) ⟩
   map f (concat xss)                         ∎
 
+concat-++ : (xss yss : List (List A)) → concat xss ++ concat yss ≡ concat (xss ++ yss)
+concat-++ [] yss = refl
+concat-++ ([] ∷ xss) yss = concat-++ xss yss
+concat-++ ((x ∷ xs) ∷ xss) yss = cong (x ∷_) (concat-++ (xs ∷ xss) yss)
+
+concat-concat : concat {a} {A} ∘ map concat ≗ concat ∘ concat
```
MatthewDaggitt, line 504: I'm surprised `{a}` is necessary here, shouldn't it be inferrable from `{A}`?

CHANGELOG.md
Taneb: Good spot, thanks

src/Data/List/Properties.agda
Taneb: Indeed it is, I've made that `{_}` now in `concat-concat` and `concat-[-]`

MatthewDaggitt: The usually idiomatic way of doing this instead of having `{_} {A}` would be to instead write `{A = A}`

Taneb: Amended

PR #1322, opened 2020-10-19
Add generic divisibility

mechvel:
The minimal set of items is added which are necessary to define the
notion of GCD.
This commit can be considered in the following succession.

1. It is added  ``+-*-commutativeSemiring``  export to
   Algebra.Bundles.CancellativeCommutativeSemiring.

2. LeftQuotient and RightQuotient added to Algebra.Definitions.

3. Divisibilty ``_|_`` is defined and several its properties proved in
   Algebra.Properties.Magma,  Algebra.Properties.Semigroup,
   Algebra.Properties.Monoid, Algebra.Properties.Semiring.

4. IsIrreducible, Coprime, and GCD are defined, and several their properties
   proved in Algebra.Properties.CancellativeCommutativeSemiring.

changed:   CHANGELOG.md
changed:   src/Algebra/Bundles.agda
changed:   src/Algebra/Definitions.agda
changed:   src/Algebra/Properties/Semigroup.agda
new file:  src/Algebra/Properties/CancellativeCommutativeSemiring.agda
new file:  src/Algebra/Properties/Magma.agda
new file:  src/Algebra/Properties/Monoid.agda
new file:  src/Algebra/Properties/Semiring.agda


src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+
+open Magma M
+
+------------------------------------------------------------------------------
+-- The divisibility relation
+------------------------------------------------------------------------------
+
+infix 5 _∣_ _∤_ _∣∣_ _¬∣∣_
+
+private C = Carrier
```
MatthewDaggitt, line 27: You should do this via renaming rather than a private definition. It's also more standard to call it A rather than C.

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+
+_∣_ : Rel C (a ⊔ ℓ)                   -- x ∣ y  denotes  x divides y
+x ∣ y =  LeftQuotient _≈_ _∙_ y x     -- as  y ≈ q ∙ x  for some q.
```
MatthewDaggitt, line 30: These definitions don't depend on the magma axioms so I think it would make more sense for the contents of this file to live in it's own file (`Algebra.Divisibility`?) parameterised by the equality relation and operator. The proofs will then be parameterised by the exact properties they need (e.g. transitivity, congruency). This will make the whole development much more flexible to use.

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+------------------------------------------------------------------------------
+
+∣-respˡ : {y : C} → (_∣ y) Respects _≈_
+∣-respˡ x≈x' (q , y≈qx) =  (q , trans y≈qx (∙-congˡ x≈x'))
+
+∣-respʳ : {x : C} → (x ∣_) Respects _≈_
+∣-respʳ y≈y' (q , y≈qx) =  (q , trans (sym y≈y') y≈qx)
+
+∣-resp : _∣_ Respects₂ _≈_
+∣-resp = ((\{x y y'} → ∣-respʳ {x} {y} {y'}) , (\{y x x'} → ∣-respˡ {y} {x} {x'}))
```
MatthewDaggitt, line 60: Use proper lambas please.

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+
+_¬∣∣_ : Rel C (a ⊔ ℓ)
+_¬∣∣_ x =  ¬_ ∘ _∣∣_ x
+
+------------------------------------------------------------------------------
+-- Properties of divisibility _∣_ in Magma.
+------------------------------------------------------------------------------
+
+∣-respˡ : {y : C} → (_∣ y) Respects _≈_
+∣-respˡ x≈x' (q , y≈qx) =  (q , trans y≈qx (∙-congˡ x≈x'))
```
MatthewDaggitt, line 54: No need for brackets. Same below

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+-- g g' : GCD a b → g ∣∣ g'.
+
+_¬∣∣_ : Rel C (a ⊔ ℓ)
+_¬∣∣_ x =  ¬_ ∘ _∣∣_ x
+
+------------------------------------------------------------------------------
+-- Properties of divisibility _∣_ in Magma.
+------------------------------------------------------------------------------
+
+∣-respˡ : {y : C} → (_∣ y) Respects _≈_
```
MatthewDaggitt, line 53: Can be written ∀ {y}. Same below.

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+∣-respˡ x≈x' (q , y≈qx) =  (q , trans y≈qx (∙-congˡ x≈x'))
+
+∣-respʳ : {x : C} → (x ∣_) Respects _≈_
+∣-respʳ y≈y' (q , y≈qx) =  (q , trans (sym y≈y') y≈qx)
+
+∣-resp : _∣_ Respects₂ _≈_
+∣-resp = ((\{x y y'} → ∣-respʳ {x} {y} {y'}) , (\{y x x'} → ∣-respˡ {y} {x} {x'}))
+
+x∣yx : {x : C} → (y : C) → x ∣ (y ∙ x)
```
MatthewDaggitt, line 62: Both arguments should be explicit for the same reasons as we discussed for `AlmostLeftCancellative`. Same below.

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,75 @@
...
+x∣xy comm {x} y = (y , comm x y)
+
+bothFactors-∣ : Commutative _≈_ _∙_ → (x y : C) → x ∣ (y ∙ x) × y ∣ (y ∙ x)
+bothFactors-∣ comm x y = ((y , refl) , (x , comm y x))
+
+bothFactors-∣≈ : Commutative _≈_ _∙_ → (x y z : C) → z ≈ y ∙ x → x ∣ z × y ∣ z
+bothFactors-∣≈ comm x y z z≈yx =
+  let (x∣yx , y∣yx) = bothFactors-∣ comm x y
+  in
+  (∣-respʳ (sym z≈yx) x∣yx , ∣-respʳ (sym z≈yx) y∣yx)
```
MatthewDaggitt, line 75: Can be written in one line with `map₂` from `Data.Product`

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,29 @@
...
+open import Relation.Binary using (Reflexive)
+
+module Algebra.Properties.Monoid {a ℓ} (M : Monoid a ℓ) where
+
+open Monoid M
+open Algebra.Properties.Semigroup semigroup public
+
+------------------------------------------------------------------------------
+-- Several divisibility properties
+------------------------------------------------------------------------------
```
MatthewDaggitt, line 21: We usually have a blank line after section headings. I'd also just write "Properties of divisibility"

src/Algebra/Properties/Semigroup.agda
```diff
@@ -7,10 +7,26 @@
...
-x∙yz≈xy∙z x y z =  sym (assoc x y z)
+x∙yz≈xy∙z x y z = sym (assoc x y z)
+
+∣-trans : Transitive _∣_
+∣-trans {x} {y} {z} (q , y≈qx) (q' , z≈q'y) = (q' ∙ q , z≈q'qx)
+ where
+ z≈q'qx = begin
+   z              ≈⟨ z≈q'y ⟩
+   q' ∙ y         ≈⟨ ∙-congˡ y≈qx ⟩
+   q' ∙ (q ∙ x)   ≈⟨ sym (assoc q' q x) ⟩
```
MatthewDaggitt, line 30: Can use the symmetric version of `≈⟨` instead to get rid of `sym`

src/Algebra/Properties/Semigroup.agda
```diff
@@ -7,10 +7,26 @@
...
 open Semigroup S
+open EqReasoning setoid
+open Algebra.Properties.Magma magma public
 
 x∙yz≈xy∙z : ∀ x y z → x ∙ (y ∙ z) ≈ (x ∙ y) ∙ z
-x∙yz≈xy∙z x y z =  sym (assoc x y z)
+x∙yz≈xy∙z x y z = sym (assoc x y z)
+
+∣-trans : Transitive _∣_
+∣-trans {x} {y} {z} (q , y≈qx) (q' , z≈q'y) = (q' ∙ q , z≈q'qx)
```
MatthewDaggitt, line 25: Please use `p` and `q` rather than `q` and `q'`

CHANGELOG.md
```diff
@@ -67,6 +72,9 @@ Other minor additions
   AlmostCancellative      e _•_ = AlmostLeftCancellative e _•_ × AlmostRightCancellative e _•_
   ```
 
+* Added export of `+-*-commitativeSemiring` to the record of
+  of `Algebra.Bundles.CancellativeCommutativeSemiring`.
```
MatthewDaggitt, line 76: This entry is not needed as you added `CancellativeCommutativeSemiring` in this version of the library.

MatthewDaggitt (commented):
Thanks for the PR. I've added some initial comments. 

Unfortunately it's still too big though. I haven't reviewed `Algebra.Properties.Semiring` and `Algebra.Properties.CancellativeCommutativeSemiring` so please remove them and submit them as the next PR after this one is merged in.

src/Algebra/Properties/Semigroup.agda
```diff
@@ -7,10 +7,26 @@
...
+ where
+ z≈q'qx = begin
```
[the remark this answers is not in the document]
mechvel: > Algebra.Properties.Semiring and Algebra.Properties.CancellativeCommutativeSemiring so please remove them and 
> submit them as the next PR 

Removed, so far.

> No need to use the where block at all. Just replace z≈q'qx with the actual proof [..]

Done.

src/Algebra/Properties/Magma.agda
mechvel: Done.

src/Algebra/Properties/Magma.agda
mechvel: Done.

src/Algebra/Properties/Magma.agda
mechvel: Do you mean to write  ``a , b``  instead of  ``(a , b)`` everywhere where it is possible?
You see, a mathematical tradition is to use parentheses to denote a pair.

Anyway I have removed these parentheses.

src/Algebra/Properties/Magma.agda
mechvel: Done.

src/Algebra/Properties/Magma.agda
mechvel: I failed to do this.
``map₂ suc (1 , 3)``  does work.
But I do not guess how to apply here the function ``f = (∣-respʳ (sym z≈yx))`` to the needed pair by  ``map₂``.
The components have different types but `f` needs to apply to both of them.

src/Algebra/Properties/Semigroup.agda
mechvel: Please, where it is documented,  where to find the usage example?
(generally, please, give details when you expect a possible difficulty for the author).

src/Algebra/Properties/Semigroup.agda
mechvel: Done.

CHANGELOG.md
mechvel: Removed.
This is needed to remove - if PR is considered relatively to the last official lib version (say, 1.4).
But if it is considered  relatively to the current `master`, then its is needed to stay - because it changes the record of
 CancellativeCommutativeSemiring.
(?)

src/Algebra/Properties/Magma.agda
mechvel: Please, consider this once more. If you insist, then I would change this.
Meanwhile consider the following.
``_∣_`` needs all the relations and operations of Magma ``(_≈_, _∙_), and the proofs for  ``_∣_`` in this module also use
`` ∙-cong`` and the properties of ``≈``, both  coming from Magma.
So that ``_∣_`` needs exactly the whole environment of Magma.
And if in such an evident situation we do not use a parameter of Magma, then why Magma is at all introduced?
Ii is introduced exactly for the aim of replacing the environment of several relations and operation with a single word of Magma.  For the same reason other classical categories are introduced: Semigroup, Group, Ring etc.

mechvel: Divisibility has sense only in Magma. So, Magma needs to be the parameter type of the module.
In textbooks, divisibility is most often defined for Ring (for its multiplication), which is rather restrictive.  In our design, divisibility in a  (Semi)ring R  will be imported from Algebra.Properties.Magma (Semiring.*-magma R).
Please, agree with this.

mechvel:
```
>>  x ∣ y = ∃ (λ q → y ≈ q ∙ x)       -- Def
``` 
> These definitions don't depend on the magma axioms so I think it would make more sense for the contents > of this file to live in it's own file (Algebra.Divisibility?) parameterised by the equality relation and operator. 

Now, I doubt.

Do you expect the definition ``Def`` to be used in the environment that does not include the Magma laws? 
I do not know, may be you think of RawMagma and "raw-divisibility" _|_ that can be used in provers?
Can they?
(I do not recall, why provers may tend to use Raw<structure> instead  of <structure>).

1) ``Def`` depends on all the Magma _signature_ (``Carrier, _≈_, _∙_``). 
2) I do not expect this _∣_ to be used for `≈` being not an equivalence, or for _∙_ being not congruent.
    If it is used so, then it would be called differently, say, pseudo-divisibility, and would have another identifier.
3) Mathematicians always suppose that a relation  ``_∣_``  called _divisibility_ needs to satisfy reflexivity and
    transitivity. And their proof already requires the laws of Monoid. So that the parameter Magma is already    
    a very generic environment. 
4) The "Raw" definitions complicate and increase the code.

?

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,74 @@
...
+------------------------------------------------------------------------------
+
+∣-respˡ : ∀ {y} → (_∣ y) Respects _≈_
+∣-respˡ x≈x' (q , y≈qx) = q , trans y≈qx (∙-congˡ x≈x')
+
+∣-respʳ : ∀ {x} → (x ∣_) Respects _≈_
+∣-respʳ y≈y' (q , y≈qx) = q , trans (sym y≈y') y≈qx
+
+∣-resp : _∣_ Respects₂ _≈_
+∣-resp = (λ {x y y'} → ∣-respʳ {x} {y} {y'}) , (λ {y x x'} → ∣-respˡ {y} {x} {x'})
```
MatthewDaggitt, line 59: Not sure why you're not simply writing `∣-respʳ, ∣-respˡ`?

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,74 @@
...
+------------------------------------------------------------------------------
+
+infix 5 _∣_ _∤_ _∣∣_ _¬∣∣_
+
+
+_∣_ : Rel Carrier (a ⊔ ℓ)          -- x ∣ y  denotes  x divides y
+x ∣ y = ∃ (\q → y ≈ (q ∙ x))
+
+_∤_ : Rel Carrier (a ⊔ ℓ)
+_∤_ x = ¬_ ∘ _∣_ x
```
MatthewDaggitt, line 32: Much more readable as `x ∤ y = ¬ x ∣ y`. Same for `_¬∣∣_` below.

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,30 @@
...
+
+open Monoid M
+open Algebra.Properties.Semigroup semigroup public
+
+------------------------------------------------------------------------------
+-- Properties of divisibility
+------------------------------------------------------------------------------
+
+ε∣ : ∀ x → ε ∣ x
+ε∣ x = (x , sym (identityʳ x))
```
MatthewDaggitt, line 24: Again unnecessary brackets here and below.

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,30 @@
...
+
+∣-refl : Reflexive _∣_
+∣-refl {x} = (ε , sym (identityˡ x))
+
+∣-reflexive≈ : ∀ {x y} → x ≈ y → x ∣ y
+∣-reflexive≈ {x} {y} x≈y = (ε , sym (trans (identityˡ x) x≈y))
```
MatthewDaggitt, line 30: Shorter: `∣-reflexive≈ x≈y = (ε , sym (trans (identityˡ _) x≈y))`

src/Algebra/Properties/Semigroup.agda
```diff
@@ -7,10 +7,27 @@
...
 open Semigroup S
+open EqReasoning setoid
+open Algebra.Properties.Magma magma public
 
 x∙yz≈xy∙z : ∀ x y z → x ∙ (y ∙ z) ≈ (x ∙ y) ∙ z
-x∙yz≈xy∙z x y z =  sym (assoc x y z)
+x∙yz≈xy∙z x y z = sym (assoc x y z)
+
+∣-trans : Transitive _∣_
+∣-trans {x} {y} {z} (p , y≈px) (q , z≈qy) =
```
MatthewDaggitt, line 25: More consistent and shorter to lay this out as:
```
∣-trans {x} {y} {z} (p , y≈px) (q , z≈qy) = q ∙ p , (begin
  z             ≈⟨ z≈qy ⟩
  q ∙ y         ≈⟨ ∙-congˡ y≈px ⟩
  q ∙ (p ∙ x)   ≈⟨ sym (assoc q p x) ⟩
  (q ∙ p) ∙ x   ∎)
```

src/Algebra/Properties/Magma.agda
```diff
@@ -0,0 +1,74 @@
...
+
+open Magma M
+
+------------------------------------------------------------------------------
+-- The divisibility relation
+------------------------------------------------------------------------------
+
+infix 5 _∣_ _∤_ _∣∣_ _¬∣∣_
+
+
```
MatthewDaggitt, line 27: Only single empty line required

src/Algebra/Properties/Magma.agda
MatthewDaggitt: Sorry, my mistake it should be `map` rather than `map₂`

src/Algebra/Properties/Semigroup.agda
MatthewDaggitt: The symmetric variant `≈˘⟨` is in every equational reasoning file.

CHANGELOG.md
MatthewDaggitt: If you take a moment to think about it, I'm pretty sure you'll agree with me that it wouldn't make much sense to have a CHANGELOG entry for every change relative to the current state of master!

src/Algebra/Properties/Magma.agda
MatthewDaggitt: > Do you expect the definition Def to be used in the environment that does not include the Magma laws? I do not know, may be you think of RawMagma and "raw-divisibility" | that can be used in provers?
 
There's certainly solver theories that involve divisibility out there. That's exactly the point: who knows how this code will be used in the future? Given that, it makes sense to ensure the code is as modular and reusable as possible. This approach is followed across the library.

> The "Raw" definitions complicate and increase the code.

The proofs would then be re-exported and correctly parameterised in `Algebra.Properties.(Magma/Semigroup/Monoid)`. This means that the public facing API for the users of the library remains unchanged.

MatthewDaggitt (commented):
Some more feedback

mechvel:
```
 src/Algebra/Properties/Magma.agda
bothFactors-∣≈ comm x y z z≈yx =
  let (x∣yx , y∣yx) = bothFactors-∣ comm x y
  in
  (∣-respʳ (sym z≈yx) x∣yx , ∣-respʳ (sym z≈yx) y∣yx)
```
> Sorry, my mistake it should be map rather than map₂

I tried
```  
  let (x∣yx , y∣yx) = bothFactors-∣ comm x y
  in
  map (∣-respʳ (sym z≈yx)) (x∣yx , y∣yx)
```
And it reports 
```
Data.Product.Σ _A_100 _P_105 → Data.Product.Σ _B_103 _Q_106 !=<
Data.Product.Σ (x ∣ z) (λ x₁ → y ∣ z)
  Data.Product.Σ _A_100 _P_105 → Data.Product.Σ _B_103 _Q_106
matches the expected type   x ∣ z × y ∣ z
```
Can you, please, suggest a concrete code with this ``map`` ?

src/Algebra/Properties/Magma.agda
mechvel: Please, consider my commit of October 26, 2020.
All your notes are implemented - except few ones to which I provide questions.
```
>>  x ∣ y = ∃ (λ q → y ≈ q ∙ x)       -- Def
```
> These definitions don't depend on the magma axioms so I think it would make more sense for the contents 
> of this file to live in it's own file (Algebra.Divisibility?) parameterised by the equality relation and operator.

So far, I put it to  ``Algebra.Definitions`` as
```
module Divisibility (_∙_ : Op₂ A)                                                 
  where                                                                           
  infix 5 _∣_ _∤_ _∣∣_ _¬∣∣_                                                      
                                                                                  
  _∣_ : Rel A (a ⊔ ℓ)           -- x ∣ y  denotes  x divides y                    
  x ∣ y = ∃ (λ q → y ≈ (q ∙ x))                                                   
  ...
```
Respectively its import to other modules in Algebra/Properties has changed. For example,
 ```
...
open import Algebra.Definitions using (Commutative; module Divisibility)          
...                                                                                   
module Algebra.Properties.Magma {a ℓ} (M : Magma a ℓ) where                       
                                                                                  
open Magma M                                                                      
...                                                                                
open Divisibility {A = Carrier} _≈_ _∙_ public using (_∣_; _∤_; _∣∣_; _¬∣∣_)      
                                                                                  
∣-respˡ : ∀ {y} → (_∣ y) Respects _≈_                                             
∣-respˡ x≈x' (q , y≈qx) = q , trans y≈qx (∙-congˡ x≈x')                           
...
```          
I am going to convince you that this is appropriate.
Divisibility is not such an outstanding definition as to take a separate file. It is not any more remarkable than, say,
`LeftInverse`. It is similar to several definitions in ``Algebra.Definitions``. For example,
```
RightIdentity : A → Op₂ A → Set _                                                 
RightIdentity e _∙_ = ∀ x → (x ∙ e) ≈ x
```
defines what is a right identity with respect to  `≈` and  ``_∙_``.
Similarly
```
Divides : Op₂ A → Rel A _
Divides _∙_ x y = ∃ (λ q → y ≈ (q ∙ x)) 
```
could define what does it mean "x divides y relatively to `≈` and  ``_∙_``".
So, is this an appropriate place to set `Divides` ?
Further, we need to use an infix ``_∣_`` instead of divides. But how to agree the infix syntax for  ``_∣_`` with the third argument ``_∙_`` which is not in the parameters of ``Algebra.Definitions`` ?  
I do not find any other way as to introduce the above ``module Divisibility``, with declaring  ``_∣_`` inside it.
Will this do?

File: src/Algebra/Properties/Magma.agda (hunk 6 above)
Comment (mechvel) [reply src/Algebra/Properties/Magma.agda]: Done.

File: src/Algebra/Properties/Magma.agda (hunk 7 above)
Comment (mechvel) [reply src/Algebra/Properties/Magma.agda]: For example, 
```
  let (x∣yx , y∣yx) = bothFactors-∣ comm x y                                      
  in                                                                              
  map (∣-respʳ (sym z≈yx)) (x∣yx , y∣yx)                           
```
Is not type-checked. Could you, please, provide the precise code line that you intend?

File: src/Algebra/Properties/Semigroup.agda (hunk 9 above)
Comment (mechvel) [reply src/Algebra/Properties/Semigroup.agda]: I set ``≈˘⟨``.

File: CHANGELOG.md (hunk 11 above)
Comment (mechvel) [reply CHANGELOG.md]: Probably  CHANGELOG.md in each my commit  needs to list the changes relatively to the last official library release,
and for this GCD PR, this release is lib-1.4.
Right?

File: src/Algebra/Properties/Magma.agda (hunk 13 above)
Comment (mechvel) [reply src/Algebra/Properties/Magma.agda]: Done.

File: src/Algebra/Properties/Magma.agda (hunk 14 above)
Comment (mechvel) [reply src/Algebra/Properties/Magma.agda]: Done.

File: src/Algebra/Properties/Monoid.agda (hunk 15 above)
Comment (mechvel) [reply src/Algebra/Properties/Monoid.agda]: Done.

File: src/Algebra/Properties/Monoid.agda (hunk 16 above)
Comment (mechvel) [reply src/Algebra/Properties/Monoid.agda]: Done.

File: src/Algebra/Properties/Semigroup.agda (hunk 17 above)
Comment (mechvel) [reply src/Algebra/Properties/Semigroup.agda]: Done (with the symmetric  version for ``≈⟨``).

File: src/Algebra/Properties/Magma.agda (hunk 18 above)
Comment (mechvel) [reply src/Algebra/Properties/Magma.agda]: Done.

**Comment by mechvel** [comment]
> Sorry, my mistake it should be map rather than map₂

It does not check neither. See my earlier comment.

File: src/Algebra/Properties/Magma.agda (hunk 7 above)
Comment (MatthewDaggitt) [reply src/Algebra/Properties/Magma.agda]: No longer relevant thanks to changes in latest commit

**Comment by MatthewDaggitt** [comment]
@mechvel I've had a go at reorganising it along the lines that I was thinking of. We now have one top-level module `Algebra.Divisibility` containing both the definitions and the proofs which is parameterised by the equality relation and the binary operation. The relevant parts of which are re-exported publicly by `Algebra.Properties.X.Divisibility` modules. This should give us maximum flexibility while preserving a nice interface for users who are only interested in divisibility over a particular bundle.

I've also:
 - flipped the definition of divisibility from `∃ λ q → y ≈ q ∙ x` to `∃ λ q → q ∙ x ≈ y ` as it gets rid of around 4/5ths of the requirements for symmetry.
 - renamed some of the proofs to make them more consistent with the library conventions

Let me know if you have any comments?

**Comment by mechvel** [comment]
1) 
> flipped the definition of divisibility from ``∃ λ q → y ≈ q ∙ x to ∃ λ q → q ∙ x ≈ y`` 
> as it gets rid of around 4/5ths of the requirements for symmetry.

This 4/5 is probably by occasion. May be in other code examples it will occur 1/5.

My large application (similar to DoCon-A-2.02) probably has thousands of patterns like ``(q , y≈qx)`` for divisibility.
And for this application I need to replace there possibly more definitions with standard ones.
Flipping these equality parts is not automatic by "find and replace".
A thousand of proofs will have to be changed by inserting/removing `sym` in many places.

It is desirable for me that you return the definition to ``y ≈ q ∙ x`` or allow me to do these changes myself.

But I leave the decision up to you.

2) You set
```
module Algebra.Properties.Semigroup.Divisibility
module Algebra.Properties.CommutativeSemigroup.Divisibility
module Algebra.Properties.Monoid.Divisibility
...
```
Too much of files.
Properties.magma, Properties.Semigroup, Properties.CommutativeSemigroup
are poor, they have too few items and proofs. It is better to add the divisibility proofs to them.

I leave the decision up to you.

3) Why do you remove Algebra.Properties.Magma ?

Let the programmer write a code in a module parameterized by ``(M : Magma _ _)``
and by ``(comm : Commutative _∙_)``.
How does one use
```
    x∣xy,  bothFactors-∣,  bothFactors-∣≈
``` 
?

These are proved in my code for Magma, with adding the argument ``(comm : Commutative _∙_)``.

You move these to  Algebra.Properties.CommutativeSemigroup.Divisibility.
(Commutative)Semigroup includes ``assoc : Associative _∙_``, but it is not used in the above three proofs.

In algebra there are sometimes considered non-associative operations, for example, in Lie algebras.

I suggest to return all the above proofs to  Algebra.Properties.Magma.
The Magma properties and items are very poor, it is not likely to be much of them in this module.
So that it does worth to separate Magma.Divisibility from it, neither it worth to add CommutativeMagma. It is better to add the `comm` argument to the proofs which need it, and just put all the stuff into Algebra.Properties.Magma.

Anyway I leave the decision up to you. 

4)
```
module Algebra.Properties.CommutativeSemigroup.Divisibility {a ℓ} (CS : CommutativeSemigroup a ℓ)
...
∣-factors ...
∣-factors-≈ : ∀ x y {z} → x ∙ y ≈ z → x ∣ z × y ∣ z
```
a)  ``factors-∣``  is likely to read as "each factor divides the product".
   ``∣-factors``  is likely read as "something divides each factor".

b) Applications would  introduce for `Monoid` the function  ``Π : List A → A``  for product of an element list.
   And there will be ``factors-∣`` proving that each element of the list divides the product.
   For this reason I wrote ``bothFactors-∣``.
   With your variant, I would probably write `` ∣-factorsIn-Π``.

I do not know, I leave the decision up to you.

MatthewDaggitt:
>> flipped the definition of divisibility from ∃ λ q → y ≈ q ∙ x to ∃ λ q → q ∙ x ≈ y
>> as it gets rid of around 4/5ths of the requirements for symmetry.
> 1. This 4/5 is probably by occasion. May be in other code examples it will occur 1/5. My large application (similar to DoCon-A-2.02) probably has thousands of patterns like (q , y≈qx) for divisibility. And in my applications I need to minimize the diversion from standard. Flipping these equality parts is not automatic by "find and replace". A thousand of proofs will have to be changed by inserting/removing sym in many places.

The reason why using `q ∙ x ≈ y` instead of `y ≈ q ∙ x`  simplifies a lot of the proofs is that all properties in the library dealing with equations are defined so that the LHS is "complicated" and the RHS is "simple", e.g.

https://github.com/agda/agda-stdlib/blob/77c31ffcf3b37732b3774354bb3f89bac47f027f/src/Algebra/Definitions.agda#L46

https://github.com/agda/agda-stdlib/blob/77c31ffcf3b37732b3774354bb3f89bac47f027f/src/Algebra/Definitions.agda#L55

https://github.com/agda/agda-stdlib/blob/77c31ffcf3b37732b3774354bb3f89bac47f027f/src/Algebra/Definitions.agda#L64

https://github.com/agda/agda-stdlib/blob/77c31ffcf3b37732b3774354bb3f89bac47f027f/src/Algebra/Definitions.agda#L114

https://github.com/agda/agda-stdlib/blob/77c31ffcf3b37732b3774354bb3f89bac47f027f/src/Algebra/Definitions.agda#L93

In order to get similar reductions in the complexity of proofs, we should continue this pattern when defining divisibility. I'm really do feel your pain about having to update your development, I'm sorry. However no matter how big a one-off cost it is, it's best to optimise the library's design for everyone in the future.

> 2. Too much of files. Properties.magma, Properties.Semigroup, Properties.CommutativeSemigroup
are poor, they have too few items and proofs, and in standard, they would not grow much.
It is better to move the divisibility proofs to them and to remove their X.Divisibility files.

I'm afraid I disagree. The study of semirings or groups for examples make up entire fields of research in mathematics. And therefore leaving them as single monolithic files has the potential for them to be enormous. Furthermore they would drag in modules from all over the library (for example `Algebra.Properties.CommutativeMonoid` already has a huge import list). By splitting them into smaller files, it greatly reduces the number of dependencies and makes them easier to use and compose. In the long term I would like to see `Algebra.Properties.X` follow the same pattern as `Data.X`, i.e. be mostly empty, only publicly re-exporting the contents of its submodules.

> 3. Let the programmer write a code in a module parameterized by (M : Magma _ _)
and by (comm : Commutative _∙_). How does one use x∣xy,  bothFactors-∣,  bothFactors-∣≈.

This is an excellent point, I'd missed this.

> So that it does worth to separate Magma.Divisibility from it, neither it worth to add CommutativeMagma

Nope, as always, it's better to be consistent and add a `CommutativeMagma` and `Algebra.Properties.CommutativeMagma.Divisibility`. I've just added a suitable commit.

> 4. a) factors-∣ is likely to read as "each factor divides the product". ∣-factors is likely read as "something divides each factor".
b) Applications would introduce for Monoid the function Π : List A → A for product of an element list.
And there will be factors-∣ proving that each element of the list divides the product. With your variant, I would probably write ∣-factorsIn-Π.

The convention is that names of properties for operators or relations that don't explicitly try to describe the type, are prefixed with the name of the operator or relation. As you say if you proving a property about the hypothetical property `Π` then you would use `Π` in the name so a name clash would not arise.

mechvel:
>>  Too much of files. Properties.magma, Properties.Semigroup, 
>> Properties.CommutativeSemigroup   are poor, they have too few items and proofs, and in 
>> standard, they would not grow much.
>>  It is better to move the divisibility proofs to them and to remove their X.Divisibility files.

> I'm afraid I disagree. The study of semirings or groups for examples make up entire fields of
> research in mathematics. And therefore leaving them as single monolithic files has the potential > for them to be enormous. 

My note was individual for ``Magma, Semigroup, CommutativeSemigroup``
(and now, let us add there `CommutativeMagma`).
1) In mathematics, they are provided with a much smaller theory than Group or Ring. 
2) Though the existing mathematics for Semigroup is large, standard library needs probably less        
    than 1/500 of it. The remaining part is for the _applied libraries_.
3) When ``Algebra.Properties.Foo``  becomes large (which would not happen with the four categories listed above), it is natural to split it to a few number of volumes: 
``Properties.Foo-I, Properties.Foo-II`` - such that  the first one  does not import the second. 
For example, "Van Der Waerden.  Algebra-I, Algebra-II " (two volumes).
3.1) About importing modules of unneededly large size: 
for example, ``Nat.Properties`` has more than 2000 lines. If the programmer uses only 
``+-comm``, is one forced to import all ``Nat.Properties`` ?
May be split It to ``Nat.Properties-I, Nat.Properties-II`` ?

mechvel:
Note that even though ``_|_`` is called "divisibility", it is both for multiplicative case and additive one.
For example, it one imports ``_|_`` from ``+-magma`` for `Nat`, then it occurs that   
``2 | 3``  ("2 divides 3"), and this means that  "there exists q such that 3 == q + 2".

MatthewDaggitt:
> My note was individual for Magma, Semigroup, CommutativeSemigroup
(and now, let us add there CommutativeMagma).
> 1. In mathematics, they are provided with a much smaller theory than Group or Ring.

Yes, but we want to use the same pattern for all of the bundles. Having one pattern for `Semigroup`s and a totally different pattern for `Semiring`s is needlessly confusing.

> 1. Though the existing mathematics for Semigroup is large, standard library needs probably less
than 1/500 of it. The remaining part is for the applied libraries.

1/500th of it is still more than can comfortably fit in one file though. Even if it's not, better to build in flexibility and redundancy now, than be forced to deprecate existing modules and move everything around later on.

> 3. When Algebra.Properties.Foo becomes large (which would not happen with the four categories listed above), it is natural to split it to a few number of volumes:
Properties.Foo-I, Properties.Foo-II - such that the first one does not import the second.
For example, "Van Der Waerden. Algebra-I, Algebra-II " (two volumes).

I _really_ don' think naming the files `Properties-I`, `Properties-II`, `Properties-III` etc. is better than the much more meaningful `Properties.Divisibility`, `Properties.Permutations`. In the former you have no idea what they contain or which one you need to import.

> 4. May be split It to Nat.Properties-I, Nat.Properties-II ?

Yes, at some point I would like to split up `Data.Nat.Properties` to make it less monolithic.

mechvel:
OK, can we proceed with GCD ?

MatthewDaggitt:
> OK, can we proceed with GCD ?

I'll merge this in. Please could your next PR contain the contents of `Algebra.Properties.Semiring` and `Algebra.Properties.CancellativeCommutativeSemiring` that you added in the original version of this PR. Obviously you'll need to rearrange them appropriately to match what's happened in this PR.

Issue #1321, opened 2020-10-19
Level-polymorphic Algebra.Module.Construct.Zero


conal:
The `Algebra.Module.Construct.Zero` module provides zero-dimensional (singleton) module variants (left/right/bi, semi/full) only at level zero.

For instance,

``` agda
leftSemimodule : {R : Semiring r ℓr} → LeftSemimodule R 0ℓ 0ℓ
leftSemimodule = record
  { Carrierᴹ = ⊤
  ; _≈ᴹ_ = λ _ _ → ⊤
  }
```

It’s easy to generalize by adding level parameters to the module and importing *polymorphic* unit:

``` agda
module Algebra.Module.Construct.Zero {c ℓ : Level} where

open import Algebra.Bundles
open import Algebra.Module.Bundles
open import Data.Unit.Polymorphic

private
  variable
    r s ℓr ℓs : Level

leftSemimodule : {R : Semiring r ℓr} → LeftSemimodule R c ℓ
leftSemimodule = record
  { Carrierᴹ = ⊤
  ; _≈ᴹ_ = λ _ _ → ⊤
  }
```

and so on for the other definitions.

I can imagine three options for agda-stdlib:

1.  Add this polymorphic module, say as `Algebra.Module.Construct.Zero.Polymorphic`. (I’ve been using this variant.)
2.  Replace the current `Algebra.Module.Construct.Zero` module with this generalization as above.
3.  Add `Algebra.Module.Construct.Zero.Polymorphic`, and redefine `Algebra.Module.Construct.Zero` by importing the generalization, while specializing to level zero, and re-exporting.

Thoughts?

JacquesCarette:
In the past, there were great sweeps of the library to generalize as many things as possible to be level polymorphic. Seems like that new module did not follow that trend. I think your 2nd option is the way to go.

MatthewDaggitt:
Given that the modules in question were only added in the last version, I think I'm happy with option 2 as well. Not to mention `Relation.Binary.Construct.Always/Never` follow the same pattern.

@conal if you opened a PR making the change that'd be great!

MatthewDaggitt:
I should add that `Algebra.Construct.Zero` will also need to be changed as well.

PR #1323, opened 2020-10-20
Level polymorphism for Algebra.Construct.Zero and Algebra.Module.Construct.Zero

conal:
Addresses issue #1321.

### DIFF
--- a/src/Algebra/Construct/Zero.agda
+++ b/src/Algebra/Construct/Zero.agda
@@ -15,7 +15,7 @@
 
 open import Level using (Level)
 
-module Algebra.Construct.Zero {ℓ : Level} where
+module Algebra.Construct.Zero {c ℓ : Level} where
 
 open import Algebra.Bundles
 open import Data.Unit.Polymorphic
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
Looks good, thanks! I'll leave it up for a few days and then merge it in.

src/Algebra/Module/Construct/Zero.agda
```diff
@@ -10,61 +10,62 @@
...
-open import Data.Unit
-open import Level
+open import Data.Unit.Polymorphic
 
 private
   variable
     r s ℓr ℓs : Level
 
-leftSemimodule : {R : Semiring r ℓr} → LeftSemimodule R 0ℓ 0ℓ
+leftSemimodule : {R : Semiring r ℓr} → LeftSemimodule R c ℓ
```
gallais, line 25: Is there a reason all of these have level indices `c ℓ`
whereas in `Algebra.Construct` all of the definition return
constructs using `ℓ ℓ`?

src/Algebra/Construct/Zero.agda
```diff
@@ -13,53 +13,54 @@
...
 open import Algebra.Bundles
-open import Data.Unit
-open import Level using (Level; 0ℓ)
+open import Data.Unit.Polymorphic
 
 ------------------------------------------------------------------------
 -- Raw bundles
 
-rawMagma : RawMagma 0ℓ 0ℓ
+rawMagma : RawMagma ℓ ℓ
```
JacquesCarette, line 26: Is there something that forces these to be the same level?

src/Algebra/Module/Construct/Zero.agda
conal: Thanks for the eyes and the question! Now I'm asking myself the same question. I just changed `Algebra.Construct` to use `c ℓ` as well. When `make clean && make test` finish, I'll update this pull request.

src/Algebra/Construct/Zero.agda
conal: I don't think so. Thanks for the review. See previous reply.

PR #1325, opened 2020-10-21
Annotated reflected syntax

UlfNorell:
On top of #1324.

Defines annotated reflected syntax and functions for computing and traversing it.
Use case: free variable annotations.

Note: not compatible with `--without-K`

### DIFF
--- a/src/Reflection/Annotated.agda
+++ b/src/Reflection/Annotated.agda
@@ -42,7 +42,6 @@ private
   variable
     ℓ             : Level
     u             : Univ
-    t             : ⟦ u ⟧
     Ann Ann₁ Ann₂ : Annotation ℓ
 
 -- ⟪_⟫ packs up an element of an annotated type with a top-level annotation.
@@ -50,7 +49,7 @@ infixr 30 ⟨_⟩_
 data ⟪_⟫ {u} (Tyₐ : Typeₐ ℓ u) : Typeₐ ℓ u where
   ⟨_⟩_ : ∀ {t} → Ann t → Tyₐ Ann t → ⟪ Tyₐ ⟫ Ann t
 
-ann : {Tyₐ : Typeₐ ℓ u} → ⟪ Tyₐ ⟫ Ann t → Ann t
+ann : {Tyₐ : Typeₐ ℓ u} {t : ⟦ u ⟧} → ⟪ Tyₐ ⟫ Ann t → Ann t
 ann (⟨ α ⟩ _) = α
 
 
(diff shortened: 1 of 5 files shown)


src/Reflection/Annotated.agda
```diff
@@ -0,0 +1,335 @@
...
+
+-- An annotated type is a family over an Annotation and a reflected term.
+Typeₐ : ∀ ℓ → Univ → Set (suc ℓ)
+Typeₐ ℓ u = Annotation ℓ → ⟦ u ⟧ → Set ℓ
+
+private
+  variable
+    ℓ             : Level
+    u             : Univ
+    t             : ⟦ u ⟧
```
MatthewDaggitt, line 45: At the moment the style guide for the library says that only levels and sets should be used as variables.

src/Reflection/Annotated.agda
```diff
@@ -0,0 +1,335 @@
...
+--       over a universe of types is not supported by --without-K.
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --with-K #-}
+
+module Reflection.Annotated where
+
+open import Level                               using (Level; 0ℓ; suc; _⊔_)
+open import Category.Applicative                using (RawApplicative)
+open import Data.Bool                           using (Bool; false; true; if_then_else_)
```
MatthewDaggitt, line 17: Again, importing the `Base` modules will prevent us from dragging in too many other files.

src/Reflection/Annotated.agda
UlfNorell: I would argue for updating the style guide. Is there a reason for limiting it to levels and sets? In this case it will make the data type definitions a lot more verbose. Compare

```agda
-- with variables
data Argₐ′ (Tyₐ : Typeₐ ℓ u) : Typeₐ ℓ (⟨arg⟩ u) where
  arg : ∀ i → Tyₐ Ann t → Argₐ′ Tyₐ Ann (arg i t)

-- without variables
data Argₐ′ {u} (Tyₐ : Typeₐ ℓ u) : Typeₐ ℓ (⟨arg⟩ u) where
  arg : ∀ {Ann : Annotation ℓ} {t : ⟦ u ⟧} i → Tyₐ Ann t → Argₐ′ Tyₐ Ann (arg i t)
```

MatthewDaggitt: Hmm looking back at the discussion in #697, we didn't want people going overboard with variables at the expense of readability. In the back of my mind at the time was also the concern that variables were very new, and therefore _may_ have had bugs in so perhaps we didn't want to jump in to using them too deeply.

The latter concern has obviously been addressed by time, but the former is still valid. The problem is that to the author of a piece of code, it's blindingly obvious what the types of variables are, but to readers it may be much more difficult to suss out. I guess it's also partly me being lazy and not wanting to have to get into an argument about readability in every PR that contains an overzealous use of variables.

What are other people's thoughts? @JacquesCarette? @gallais?

UlfNorell: If you think it improves readability I think it's fair to drop the `t` variable. I don't see any readability argument for `u` since without the variable there would just be a bunch of `∀ {u}` everywhere instead (and it plays the role of a Set here, so could be covered by the "only sets and levels" restriction). The `Ann` variable removes a lot of noise from the primed datatypes, and since it's a central concept of the module I would argue that it should not lead to any confusion to make it a variable.

gallais: I think we could extend the criterion to "values in the module's universe of discourse".
So `Nat` variables in `Data.Nat.Properties`, `List` ones in `Data.List.Relation.Unary.Any`
and Reflection's `Universe` ones in `Reflection`.

This way we still have a clear criterion to avoid infinite debates but are a bit more
liberal when the context ought to make what these are obvious.

MatthewDaggitt: > If you think it improves readability I think it's fair to drop the t variable. I don't see any readability argument for u since without the variable there would just be a bunch of ∀ {u} everywhere instead (and it plays the role of a Set here, so could be covered by the "only sets and levels" restriction). The Ann variable removes a lot of noise from the primed datatypes, and since it's a central concept of the module I would argue that it should not lead to any confusion to make it a variable.

> I think we could extend the criterion to "values in the module's universe of discourse".
So Nat variables in Data.Nat.Properties, List ones in Data.List.Relation.Unary.Any
and Reflection's Universe ones in Reflection.

Both of these seem reasonable to me :+1:

JacquesCarette: Sorry for the slowness in answer. I agree with @gallais 's suggestion on "values in the module's universe of discourse".

However, I am still a tiny bit wary: in agda-categories, we had modules that quantified over arrows that broke because different versions of Agda inserted the parameters in different orders, and that led to problems: you can't reliably specify those, even by name, at call sites!

So the rule perhaps ought to be "you can use a variable to abstract over a parameter that is in the domain of discourse of the module AND you never expect that variable to be hand-instantiated in user code".  At least, until Agda treats its named telescopes more DAG-like than as strict lists.

PR #1326, opened 2020-10-21
Changes related to SizeUniv [agda/agda#5000]

andreasabel:
PR https://github.com/agda/agda/pull/5000 puts `Size` into `SizeUniv` instead of `Set`, so we need special versions of `Pred` one could call `SizedType`.

This patch has to be merged at the same time as PR agda/agda#5000.  Not sure how to CI-test such a scenario.
I am also not sure whether I placed it on the correct branch (browsed `HACKING.md` for that situation, but did not find instructions).

### DIFF
--- a/src/Codata/Thunk.agda
+++ b/src/Codata/Thunk.agda
@@ -8,7 +8,6 @@
 
 module Codata.Thunk where
 
-open import Level
 open import Size; open SizedType
 
 ------------------------------------------------------------------------
--- a/src/Codata/Thunk.agda
+++ b/src/Codata/Thunk.agda
@@ -8,11 +8,26 @@
 
 module Codata.Thunk where
 
+open import Level
 open import Size
-open import Relation.Unary
+
+private
+  variable
+    ℓ ℓ₁ ℓ₂ : Level
 
 ------------------------------------------------------------------------
 -- Basic types.
+
+SizedType : (ℓ : Level) → Set (suc ℓ)
+SizedType ℓ = Size → Set ℓ
+
+∀ˢ[_] : SizedType ℓ → Set ℓ
+∀ˢ[ F ] = ∀{i} → F i
+
+infixr 8 _⇒ˢ_
+
+_⇒ˢ_ : SizedType ℓ₁ → SizedType ℓ₂ → SizedType (ℓ₁ ⊔ ℓ₂)
+F ⇒ˢ G = λ i → F i → G i
 
 record Thunk {ℓ} (F : Size → Set ℓ) (i : Size) : Set ℓ where
   coinductive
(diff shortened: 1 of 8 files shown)


src/Codata/Thunk.agda
```diff
@@ -8,12 +8,27 @@
...
+  variable
+    ℓ ℓ₁ ℓ₂ : Level
 
 ------------------------------------------------------------------------
 -- Basic types.
 
+SizedType : (ℓ : Level) → Set (suc ℓ)
+SizedType ℓ = Size → Set ℓ
+
+∀ˢ[_] : SizedType ℓ → Set ℓ
```
gallais, line 24: Can't we give these the same name as the `Relation.Unary` counterparts?
If people want to use both in the same file they can rename them / use
qualified names at the point of use.

src/Codata/Thunk.agda
```diff
@@ -8,12 +8,27 @@
...
+
+private
+  variable
+    ℓ ℓ₁ ℓ₂ : Level
 
 ------------------------------------------------------------------------
 -- Basic types.
 
+SizedType : (ℓ : Level) → Set (suc ℓ)
+SizedType ℓ = Size → Set ℓ
```
MatthewDaggitt, line 22: Shouldn't these additions go in `Size` rather than `Codata.Thunk`?

MatthewDaggitt:
> This patch has to be merged at the same time as PR agda/agda#5000. Not sure how to CI-test such a scenario.

The version of Agda that the `experimental` branch is tested against is determined by this line:
https://github.com/agda/agda-stdlib/blob/6f78e0333686268849bb84aacafac414b7c3dec5/.travis.yml#L59
so once you've merged in the change upstream you should update the commit hash here to point to the relevant one.

> I am also not sure whether I placed it on the correct branch (browsed HACKING.md for that situation, but did not find instructions).

This is the correct branch. I've opened #1327 to update HACKING.

gallais:
> so once you've merged in the change upstream you should update the commit hash here to point to the relevant one.

I don't even think you need to merge it. You should be able to check
it out by hash only no matter which branch it's in (e.g. that's how
cherry picking works AFAIU).

src/Codata/Thunk.agda
andreasabel: Ok, but I found that these are only used in `Codata.Thunk`.  In other modules, they would clash, e.g.
```
  Checking Codata.M (Codata/M.agda).
Codata/M.agda:31,18-19
Ambiguous name _⇒_. It could refer to any one of
  Size._⇒_ bound at
    Size.agda:33,1-4
  C._⇒_ bound at
    Data/Container/Core.agda:41,8-11
```
I think it is not good that these are in scope when just `open import Size`.  This module is known to be slim and to not generate conflicts.
As a compromise, I put these definition into a submodule `SizedType`.

src/Codata/Thunk.agda
andreasabel: Makes sense, thanks.

andreasabel:
Thanks, @MatthewDaggitt and @gallais for the great instructions how to get the travis test to run!

I updated the PR to address your comments.

In the end, this PR needs squashing before merging.

src/Codata/Thunk.agda
MatthewDaggitt: Hmm I think a better solution is to mimic the normal module hierarchy underneath `Size` just as we do elsewhere in the library. For example we would put `∀[_]` in `Size.Relation.Unary` and `_⇒_` in `Size.Function`. That would make it so much easier to expand these later on if we found it necessary.

Btw, I haven't seen `⇒` used to denote functions very much. Did you consider using `→` instead which would be significantly less likely to clash with other parts of the library?

andreasabel: Another solution would be to strip the uses of `∀[_]` and `_⇒_` from `Codata.Thunk` by expanding to ordinary uses of `∀` and `->`.  It is just a few characters more. Then we do not have to discuss this problem further.  

> put `∀[_]` in `Size.Relation.Unary` and `_⇒_` in `Size.Function`

We should maybe not have two top-level modules containing one definition each...  (That are then used in a single other module, `Codata.Thunk`).

andreasabel:
Please forgive me that I merge this now for the sake of merging agda/agda#5000.
I'll open a new PR for continuing the bikeshedding.

PR #1341, opened 2020-11-01
Expose instances for decidable equality

jespercockx:
This PR exposes all the instances for decidable equality as `instance` to make them available for instance search. After some experimentation I found the nicest way to do this that fits well with the current design is to expose the elements of `IsDecEquivalence` rather than exposing the `Decidable` proofs directly. I also added a module with some examples of how these instances can be used.


jespercockx:
Note that this currently creates a bunch of single-definition modules, but these modules could be populated further with other instances at a later point, basically all the record types in `Relation.Binary.Structures` that start with `Is...` could be exposed as typeclasses following the template I put in `Relation.Binary.TypeClasses`.

src/Data/Vec/Instances.agda
```diff
@@ -9,7 +9,16 @@
...
+  using (isDecEquivalence)
+open import Data.Vec.Relation.Binary.Equality.DecPropositional
+open import Relation.Binary.TypeClasses
+
 
 instance
   vecFunctor = functor
   vecApplicative = applicative
+
+  ≡-isDecEquivalence-Vec = λ {a} {A} {{≡-isDecEquivalence-A}} {n} → isDecEquivalence (≡-dec {a} {A} (_≟_ {{≡-isDecEquivalence-A}}) {n})
```
JacquesCarette, line 24: Out curiosity: here and in a few other places, you are very explicit about implicits - why?

jespercockx: It's mainly laziness; the alternative is writing down the actual type signature of `≡-isDecEquivalence-Vec` (which might actually be a better idea now that I think about it..)

src/Relation/Binary/TypeClasses/Examples.agda
```diff
@@ -0,0 +1,47 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Usage examples of typeclasses for binary relations
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Binary.TypeClasses.Examples where
```
MatthewDaggitt, line 9: I think we're moving away from `Example` files and towards `README` files. I think I'd be tempted to move this to `README.TypeClasses` or `README.Relation.Binary.TypeClasses`. If you could also add an entry for `X.TypeClasses` in the `README.Design.Hierarchies` file that would be great.

src/Relation/Binary/TypeClasses/Examples.agda
```diff
@@ -0,0 +1,47 @@
...
+
+open import Relation.Nullary
+open import Relation.Binary.PropositionalEquality
+open import Relation.Binary.TypeClasses
+
+open import Data.Bool hiding (_≟_)
```
MatthewDaggitt, line 15: Instead of using `hiding`, you could import the base modules instead?

src/Relation/Binary/TypeClasses/Examples.agda
```diff
@@ -0,0 +1,47 @@
...
+open import Data.Vec
+open import Data.Vec.Instances
+
+test-Dec≡-Bool : Dec (true ≡ true)
+test-Dec≡-Bool = true ≟ true
+
+test-Dec≡-Nat : Dec (0 ≡ 1)
+test-Dec≡-Nat = 0 ≟ 1
+
+test-Dec≡-List : Dec (_≡_ {A = List ℕ} (1 ∷ 2 ∷ []) (1 ∷ 2 ∷ []))
```
MatthewDaggitt, line 34: Probably more readable to write this as `Dec {A = List ℕ} (1 ∷ 2 ∷ [] ≡ 1 ∷ 2 ∷ [])`. Same for the ones below

src/Reflection/Abstraction.agda
```diff
@@ -19,7 +20,8 @@ open import Relation.Binary.PropositionalEquality
 
 private
   variable
-    A B : Set
+    a b : Level
+    A B : Set a
```
MatthewDaggitt, line 24: These changes seem unnecessary?

src/Data/Vec/Instances.agda
MatthewDaggitt: Yup, I'd vote for having the type signature instead. Much more readable!

MatthewDaggitt, line 24: Hmm we should also try and make the naming conventions match. So far we have `vecFunctor` and `vecApplicative` but here we have `≡-isDecEquivalence-Vec`. Would people object to `vecIsDecEquivalence`? It's not like we're going to have non-propositional equality instances right? Or should we be going for some other convention?

MatthewDaggitt (changes requested):
> After some experimentation I found the nicest way to do this that fits well with the current design is to expose the elements of IsDecEquivalence rather than exposing the Decidable proofs directly. I also added a module with some examples of how these instances can be used.

Is there a reason for using `IsDecEquivalence` rather than `DecSetoid`? If so, then it'd be great to document in `README.TypeClasses` or something similiar? I don't use instances at all, so apologies if this is a silly question. 

> Note that this currently creates a bunch of single-definition modules, but these modules could be populated further with other instances at a later point, basically all the record types in Relation.Binary.Structures that start with Is... could be exposed as typeclasses following the template I put in Relation.Binary.TypeClasses.

Hmm it would be nice to test this theory out before we commit to it. Would it be possible for you to add `IsDecTotalOrder` instances as well, while you're going back through them? Just so that we can test that this design really does work more multiple type classes :+1: 

Finally Travis is throwing up an error in the deprecated `Data.Bin`

src/Data/Vec/Instances.agda
jespercockx: I just added the type signatures and it really does make a big difference. Not sure *what* I was thinking yesterday!

src/Relation/Binary/TypeClasses/Examples.agda
jespercockx: If only `Dec` would take an argument `{A = ...}` then I could do that.

src/Reflection/Abstraction.agda
jespercockx: Without these changes, the function `≡-dec` below is not level-polymorphic, but I need the polymorphic version in order to define the instance for `Abs` in `Reflection.Instances`.

src/Relation/Binary/TypeClasses/Examples.agda
MatthewDaggitt: Haha doh, yes sorry! Pattern recognition failure there. I spend a lot of time fixing `Decidable (_≡_ {A = List A})` patterns in PRs.

src/Data/Vec/Instances.agda
jespercockx: `vecIsDecEquivalence` is strange though because it reads as if `Vec` itself would be a decidable equivalence, which is nonsense. In fact it does not even mention the name of the relation that is decidable! However, I could go for something like `vec≡-isDecidableEquivalence`, which can be read naturally as "Vector equality is a decidable equivalence".

MatthewDaggitt: Hmm that's a good point. `vec≡-isDecidableEquivalence` is better, but we usually use `-` to separate components which would make it `vec-≡-isDecidableEquivalence`. But I also like your capitalisation of the types in your original proposal as well. Hmm....

I'm aware we've almost gone full circle, but how about `Vec-≡-isDecidableEquivalence`, and I later go back and deprecate `vecFunctor` etc. in favour of `Vec-functor`?

jespercockx:
> Is there a reason for using IsDecEquivalence rather than DecSetoid? If so, then it'd be great to document in README.TypeClasses or something similiar? I don't use instances at all, so apologies if this is a silly question.

That's a great question. In fact there is a good reason, namely that having parameters visible in the type of the instance makes the job of doing instance search a lot easier for Agda. The difference is between asking Agda "Do you have any instance of decidable equality for natural numbers for me?" and "Do you have any instance of a decidable equivalence for me?" In Haskell, only the former is even a well-formed question; there's no "zero-argument type classes". Agda's instance resolution mechanism has a second step after filtering the candidate instances to check if the candidate actually typecheck when you fill it in, but it's good practice to design type classes in such a way that most candidates can be ruled out without this extra step.

If we would use instances of type `DecSetoid`, we would end up with something quite unlike Haskell's type classes and more akin to Coq's canonical structures. However, even in Coq they're considered hard to use correctly and are generally replaced with type classes in modern projects (though canonical structures are still present in a lot of legacy code).

jespercockx:
> I'm aware we've almost gone full circle, but how about Vec-≡-isDecidableEquivalence, and I later go back and deprecate vecFunctor etc. in favour of Vec-functor?

I'm fine with that naming scheme; anyway you're not really supposed to reference these names in user code since they are filled in by instance search.

jespercockx:
> Would it be possible for you to add IsDecTotalOrder instances as well, while you're going back through them? Just so that we can test that this design really does work more multiple type classes

This raises the question of how to deal properly with subclasses, something I haven't thought about very carefully yet. In @UlfNorell 's prelude, these are handled through carefully constructed `super` fields (see e.g. https://github.com/UlfNorell/agda-prelude/blob/master/src/Prelude/Applicative/Indexed.agda#L18), but I don't think it's feasible to implement this solution in the standard library. Just turning the projection into an instance itself has the undesired effect of making instance search loop, so that's not an option either. It'd be nice if `super` could be applied to arbitrary definitions rather than just fields, since that would allow declaring the superclass structure for instance search outside of the record type definition itself.

In the meanwhile, I can implement a typeclass for `IsDecTotalOrder` like I did for `IsDecEquivalence`; however if you want to use both `_≟_` and `_≤?_` on the same (abstract) type you'd need to have instance arguments for both classes.

src/Data/Vec/Instances.agda
gallais: Why do we need the `vec` prefix in `Data.Vec.Instances`?

jespercockx: In a typical use case of these type classes, you'd want to open multiple of these `Instances` modules at the same time (otherwise, what's the point?) so the names should not cause conflicts.

JacquesCarette: Thanks for reinforcing those instincts of mine that underlie asking the question. :smirk:

jespercockx:
I've added decidable ordering instances for the base types and for the lexicographic ordering on lists; more instances could be added but I hope this shows the idea is viable.

src/Data/Vec/Instances.agda
gallais: But that's what qualified names are for?

jespercockx: I find it difficult to really care either way since the whole point of instance search is that you're not going to give these names explicitly. However, one argument in favour of the unambiguous names is that they can be used together with the `--no-qualified-instances` flag, which can help with controlling the scope of instances (see https://github.com/agda/agda/pull/4522).

UlfNorell: The ambiguous names should have no problems with `--no-qualified-instances` either. It's perfectly fine to say
```agda
open import Data.Vec.Instances using (functor)
open import Data.List.Instances using (functor)
```
as long as you don't write `functor` yourself. Instance search will pick up the correct unqualified instance.

MatthewDaggitt:
Thanks for the updates and the informative answers! I'm glad to see it works for decidable orderings as well. I don't have any further questions. Any chance you could add a CHANGELOG entry and then I'll merge it in :+1:

PR #1342, opened 2020-11-02
[ fix agda#1338 ] Made Data.Vec._⊛_ stricter

### LINKED ISSUE #1338: Data.Vec._⊛_ is overly/underly lazy
`_⊛_` is defined as

```agda
_⊛_ : ∀ {n} → Vec (A → B) n → Vec A n → Vec B n
[]       ⊛ _        = []
(f ∷ fs) ⊛ (x ∷ xs) = f x ∷ (fs ⊛ xs)
```

and the lack of pattern matching on `[]` in the first clause caused me some head scratching, 'cause the second argument wasn't getting inferred where I was expecting it to (thanks to @andreasabel for pointing out what was causing the problem).

On the other hand it's kinda nice to be lazy and make `_⊛_` compute when the second argument is not definitionally `[]`. But then why stop there and not match only on `n` like that:

```agda
_⊛_ : ∀ {n} → Vec (A → B) n → Vec A n → Vec B n
_⊛_ {0}     _  _  = []
_⊛_ {suc _} fs xs = head fs (head xs) ∷ (tail fs ⊛ tail xs)
```

This would compute even better than the current definition.

So I think the current definition is a bit inconsistent as it's somewhere between these two options and it'll perhaps be an improvement to stick to a single strategy: either favor inference or computability.

effectfully:
Fixes #1338. Note that `zipWith` is already defined as

```agda
zipWith : ∀ {n} → (A → B → C) → Vec A n → Vec B n → Vec C n
zipWith f []       []       = []
zipWith f (x ∷ xs) (y ∷ ys) = f x y ∷ zipWith f xs ys
```

so we're only making things consistent here.

### DIFF
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -149,7 +149,7 @@ _⋎_ : ∀ {m n} → Vec A m → Vec A n → Vec A (m +⋎ n)
 infixl 4 _⊛_
 
 _⊛_ : ∀ {n} → Vec (A → B) n → Vec A n → Vec B n
-[]       ⊛ _        = []
+[]       ⊛ []       = []
 (f ∷ fs) ⊛ (x ∷ xs) = f x ∷ (fs ⊛ xs)
 
 -- Multiplication


MatthewDaggitt:
Thanks for the PR!

PR #1343, opened 2020-11-03
[ new ] le for chars

gallais:


src/Data/Char/Base.agda
```diff
@@ -47,6 +48,10 @@ infix 4 _<_
 _<_ : Rel Char zero
 _<_ = ℕ._<_ on toℕ
 
+infix 4 _≤_
+_≤_ : Rel Char zero
+_≤_ = ReflClosure _<_
```
MatthewDaggitt, line 53: Hmm so there are many different ways to define this:
```agda
1) _≤_ = ReflClosure _<_
2) _≤_ = ℕ._<_ on toℕ
3) _≤_ = StrictToNonStrict _<_
```
We already use 2) in `Data.Fin.Base` so it seems to me we should be consistent unless there are excellent reasons not to?

While we're here and discussing the differences between 1) and 3), what are the differences between `_≡_` and `_≈_`? It would be nice to have a bit of explanatory text added to the definition of `_≈_`.

src/Data/Char/Properties.agda
```diff
@@ -35,6 +37,9 @@ open import Agda.Builtin.Char.Properties
 ≈⇒≡ : _≈_ ⇒ _≡_
 ≈⇒≡ = toℕ-injective _ _
 
+≉⇒≢ : ∀ {x y} → ¬ (x ≈ y) → x ≢ y
```
MatthewDaggitt, line 40: It'd be nice to define `_≉_` in the `Data.Char.Base` and then use it here?

src/Relation/Binary/Construct/Closure/Reflexive.agda
```diff
@@ -43,6 +43,9 @@ drop-refl : {R : Rel A ℓ} → Reflexive R → ReflClosure R ⇒ R
 drop-refl rfl [ xRy ] = xRy
 drop-refl rfl refl    = rfl
 
+reflexive : {R : Rel A ℓ} → _≡_ ⇒ ReflClosure R
+reflexive refl = refl
```
MatthewDaggitt, line 47: Missing from CHANGELOG

src/Data/Char/Base.agda
gallais: With `2) _≤_ = ℕ._<_ on toℕ` I assume you mean `2) _≤_ = ℕ._≤_ on toℕ`.
I used `ReflClosure` because that is literally how I was thinking about it:
the reflexive closure of the strict order.

> what are the differences between `_≡_` and `_≈_`?

`_≈_` has worse type reconstruction (which made me consider wrapping `_≤_`
in a record)? :speak_no_evil: 
I did not reallly want to pick one over the other (cf. the two versions of
`<-cmp` for instance) but perhaps it's not that big of a deal given they are
equivalent?

src/Data/Char/Properties.agda
gallais: Won't that clash with setoid also defining it?

src/Data/Char/Base.agda
MatthewDaggitt: > With 2) _≤_ = ℕ._<_ on toℕ I assume you mean 2) _≤_ = ℕ._≤_ on toℕ.

Yes!

> _≈_ has worse type reconstruction (which made me consider wrapping _≤_ in a record)?

If `_≈_` is equivalent, but has worse type reconstruction, why do we have it?

src/Data/Char/Properties.agda
MatthewDaggitt: We don't have any opened `Setoid` in scope here right? If you mean in general if we imported `Data.Char` with an opened `Setoid` in scope, then aren't we already getting clashes with `_≈_` so I'm not sure an extra one is problematic?

src/Data/Char/Base.agda
gallais: Because until we had primitive proofs about `toℕ` being injective it was
impossible to prove that it was equivalent. :D

MatthewDaggitt: So should we deprecate it?

gallais: done

gallais:
Retiring `_≈_` means changing the notions of equality and order on Strings obtained
by converting them to lists of chars. This is now a breaking change (but for the better,
I believe).

MatthewDaggitt:
Yup, that looks like a massive improvement! Thanks for propagating the changes through.

PR #1345, opened 2020-11-04
Generic GCD. Properties of Irreducible, Coprime, GCD.

mechvel:
changed:     CHANGELOG.md
changed:     src/Algebra/Bundles.agda
new file:    src/Algebra/GCD.agda
new file:    src/Algebra/Properties/CancellativeCommutativeSemiring.agda
new file:    src/Algebra/Properties/Semiring.agda

Added exports of `semiring` and `_≉_` to `Bundles.CancellativeCommutativeSemiring`.

Algebra.GCD  contains the raw definitions for  IsIrreducible, IsPrime, Coprime, GCD.

`Algebra.Properties.CancellativeCommutativeSemiring` contains proofs for
``IsIrreducible, IsPrime, Coprime, GCD``.


mechvel:
Please, consider this commit of  November 4, 2020.

I do not split `.Divisibility` from `Properties.Semiring`  because this module does not contain, so far, anything besides the divisibility properties.

Algebra.GCD  contains the raw definitions for  IsIrreducible, IsPrime,  Coprime, GCD. 

`Algebra.Properties.CancellativeCommutativeSemiring` contains proofs
for ``IsIrreducible, IsPrime, Coprime, GCD``.
The proofs for GCD are there under an anonymous module.

This is for start. More properties will have to be proved for the above items.
Then, ``record GCDDomain`` to be declared in Algebra.Bundles.
Then, `Nat` and  `Integer`  to be supplied with the instances of 
``CancellativeCommutativeSemiring,  GCDDomain``.

mechvel:
Also I set somewhere ``private A = Carrier``, 
while you required import with renaming. 
The matter is that Carrier is imported earlier by ``open Foo ... public``.
If I add there ``renaming (Carrier to A)``, then `A` will occur public, which is not the point.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,73 @@
...
+open import Data.Sum using (_⊎_; inj₁; inj₂)
+open import Function using (case_of_; flip)
+open import Level using (_⊔_)
+open import Relation.Binary using (Rel; Symmetric; Setoid)
+open import Relation.Unary using (Pred)
+
+module Algebra.Properties.CancellativeCommutativeSemiring
+  {α ℓ} (R : CancellativeCommutativeSemiring α ℓ)
+  (open CancellativeCommutativeSemiring R)
+  (0≉1 : 0# ≉ 1#)
```
MatthewDaggitt, line 23: This should not be module parameter as it is not required by all the proofs in this file. Instead it should be an argument to only the proofs that need it.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,73 @@
...
+  where
+
+open SemigroupDiv *-semigroup using (_∣_; _∤_; ∣-trans)
+open SemiringProp semiring using (∣nonzero⇒≉0)
+open Algebra.GCD _≈_ 0# 1# _*_ using (IsIrreducible; IsPrime; Coprime; GCD)
+
+private A = Carrier
```
MatthewDaggitt, line 30: Do this via the `rename` keyword when opening the bundle.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,62 @@
...
+
+open import Algebra.Divisibility _≈_ _*_
+
+------------------------------------------------------------------------------
+-- Definitions
+
+infix 4 _≉_
+
+_≉_ : Rel A _
+x ≉ y = ¬ (x ≈ y)
```
MatthewDaggitt, line 35: This definition should be private.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,62 @@
...
+
+IsIrreducible : Pred A (β ⊔ ℓ)
```
MatthewDaggitt, line 37: I'm not sure the definitions of `IsIrreducible`, `IsPrime`, and `Coprime` belong in the `GCD` module. I'd move them to a separate `Algebra.Primality` module.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,62 @@
...
+
+Coprime : Rel A (β ⊔ ℓ)
+Coprime a b = ∀ {c} → c ∣ a → c ∣ b → c ∣ 1#
+
+record GCD (a b : A) :  Set (β ⊔ ℓ) where      -- a result of gcd
+  constructor gcdᶜ
+  -- Greatest common divisor.
+
+  field
+    proper   : A                 -- the proper gcd value
```
MatthewDaggitt, line 53: You could call this field `gcd` instead, which would make it much more readable and remove the need for a comment.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,62 @@
...
+open import Data.Product using (_×_)
+open import Data.Sum using (_⊎_)
+open import Function using (flip)
+open import Level using (_⊔_)
+open import Relation.Binary using (Rel; Symmetric)
+open import Relation.Nullary using (¬_)
+open import Relation.Unary using (Pred)
+
+module Algebra.GCD
+  {β ℓ} {A : Set β} (_≈_ : Rel A ℓ) (0# 1# : A) (_*_ : Op₂ A)
```
MatthewDaggitt, line 24: It's standard to use `a` rather than `β` for the level.

MatthewDaggitt, line 24: We also usually put the binary operation before the specialised elements in the telescope.

src/Algebra/Properties/Semiring.agda
```diff
@@ -0,0 +1,33 @@
...
+-- Some basic items and theory for Semiring.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Algebra using (Semiring)
+import Algebra.Properties.Monoid.Divisibility as MonoidDiv
+open import Data.Product using (_,_)
+
+module Algebra.Properties.Semiring {r₁ r₂} (R : Semiring r₁ r₂) where
```
MatthewDaggitt, line 13: This should be `Algebra.Properties.Semiring.Divisibility`

src/Algebra/Properties/Semiring.agda
```diff
@@ -0,0 +1,33 @@
...
+
+open Semiring R
+open MonoidDiv *-monoid using (_∣_; ∣-respˡ; ε∣_)
```
MatthewDaggitt, line 16: The entire of `MonoidDiv` should be opened publicly.

src/Algebra/Properties/Semiring.agda
```diff
@@ -0,0 +1,33 @@
...
+
+------------------------------------------------------------------------------
+-- Divisibility properties specific to Semiring.
+
+1∣ : ∀ x → 1# ∣ x
+1∣ = ε∣_
+
+∣0 : ∀ x → x ∣ 0#
```
MatthewDaggitt, line 24: This is `∣-max` from `Algebra.Divisibility`

src/Algebra/Properties/Semiring.agda
```diff
@@ -0,0 +1,33 @@
...
+∣0 x = (0# , zeroˡ x)
+
+0∣x⇒x≈0 : ∀ {x} → 0# ∣ x → x ≈ 0#
+0∣x⇒x≈0 (q , q*0≈x) = trans (sym q*0≈x) (zeroʳ q)
+
+∣nonzero⇒≉0 : ∀ {x y} → x ∣ y → y ≉ 0# → x ≉ 0#
+∣nonzero⇒≉0 x∣y y≉0 x≈0 = y≉0 (0∣x⇒x≈0 0∣y)
+ where
+ 0∣y = ∣-respˡ x≈0 x∣y
```
MatthewDaggitt, line 33: No need for the `where`. Just inline the expression.

MatthewDaggitt (commented):
> I do not split .Divisibility from Properties.Semiring because this module does not contain, so far, anything besides the divisibility properties.

Regardless, we should be consistent.

> The matter is that Carrier is imported earlier by open Foo ... public.
If I add there renaming (Carrier to A), then A will occur public, which is not the point.

I don't think it is? Where is the `open ... public` statement?

mechvel:
Please, consider this commit of November 11, 2020.
Your notes are implemented, with a couple of questions formulated -
see the replies above.

src/Algebra/GCD.agda
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
mechvel: Done.

> Regardless, we should be consistent.
All right, moved to Algebra.Properties.Semiring.Divisibility.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
mechvel: Done.

src/Algebra/GCD.agda
mechvel: Done.

src/Algebra/GCD.agda
mechvel: Moved to Algebra.Primality.

src/Algebra/GCD.agda
mechvel: I intended the function  `gcd`  to return a record of GCD,
and  `gcd!`  to return the proper gcd value ``: A``, like this: 
```
record GCDSemiring ...
  field
    isGCDSemiring : ...
    gcd           : (x y : A) → GCD x y 
  ...
  gcd! : Op₂ A
  gcd! x y = GCD.proper (gcd x y)
```
So, how to name the corresponding field in GCD ?

src/Algebra/Properties/Semiring.agda
mechvel: Done.

src/Algebra/Properties/Semiring.agda
mechvel: Done.

src/Algebra/Properties/Semiring.agda
mechvel: Do you suggest to replace
```
∣0 x = (0# , zeroˡ x)
```
with
```
∣0 x = ∣-max zeroˡ x
```
?
The former is easy to understand, and the latter is brain-twisting.
The reader would wonder at the proof: in what way the maximum notion is related to the fact that everything divides 0 ?
Also this requires to add
```
open import Algebra.Divisibility _≈_ _*_ using (∣-max)
```
To my mind, it is much better to tell "my cat eats everything" than
"my cat is maximal in the relation of eating". The reason is evident.

Well, the effect is weakened by that the reader still would pay attention only to the type 
``∣0 : ∀ x → x ∣ 0#``.

I set your version.
But I shall be glad if you return to the original version.

src/Algebra/Properties/Semiring.agda
mechvel: Done.

src/Algebra/GCD.agda
MatthewDaggitt: Hmm, okay fair enough. How about `value` instead of `gcd`?

src/Algebra/Properties/Semiring.agda
MatthewDaggitt: You have a point. If you want you can remove `∣-max` and `∣-min` from `Algebra.Divisibility` and just write the proofs where needed.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,40 @@
...
+open import Algebra.Divisibility _≈_ _*_
+
+------------------------------------------------------------------------------
+-- Definitions
+
+private
+  infix 4 _≉_
+
+  _≉_ : Rel A _
+  x ≉ y = ¬ (x ≈ y)
```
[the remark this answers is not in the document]
mechvel: Removed.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,40 @@
...
+------------------------------------------------------------------------
+
+-- You're unlikely to want to use this module directly. Instead you
+-- probably want to be importing the appropriate module from
+-- `Algebra.Properties.CancellativeCommutativeSemiring`.
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Algebra.Core using (Op₂)
+import Algebra.Divisibility
```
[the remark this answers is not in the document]
mechvel: Removed.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,40 @@
...
+open import Level using (_⊔_)
+open import Relation.Binary using (Rel)
+open import Relation.Nullary using (¬_)
```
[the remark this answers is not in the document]
mechvel: Removed the line with  `¬_`.

src/Algebra/Primality.agda
```diff
@@ -0,0 +1,54 @@
...
+-- Definitions
+
+private
+  infix 4 _≉_
+
+  _≉_ : Rel A _
+  x ≉ y = ¬ (x ≈ y)
+
+IsIrreducible : Pred A (a ⊔ ℓ)
+IsIrreducible p = p ∤ 1#  ×  (∀ {x y} → (p ≈ (x * y)) → x ∣ 1# ⊎ y ∣ 1#)
```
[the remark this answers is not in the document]
mechvel: I think records are not good for small definitions.

1) My version does not require any new names to remember.
2) ``proj₁ prime-x``  is a simple proof for a prime element being nonzero. 
3) Look how simply it works:
```
  irreducible≉0 0≉1 (_ , chooseInvertible) x≈0 =  ...
```
Just pattern-match against `_,_` for `Irreducible`.

And with record, one needs to remember the two particular words for the
fields and one for the constructor. Am matching will be as ``(irredᶜ _ chooseinvertible)``.

Please, consider this and announce the decision.

src/Algebra/Primality.agda
```diff
@@ -0,0 +1,54 @@
...
+
+IsPrime : Pred A (a ⊔ ℓ)
+IsPrime p = p ≉ 0#  ×  p ∤ 1#  ×  ∀ {x y} → p ∣ (x * y) → p ∣ x ⊎ p ∣ y
+
+-- In a GCDDomain,  IsIrreducible is equivalent to IsPrime.
+-- But in the general situation IsPrime is stronger.
+
+Coprime : Rel A (a ⊔ ℓ)
```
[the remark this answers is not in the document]
mechvel: I agree. Done.

src/Algebra/Properties/Semiring/Divisibility.agda
```diff
@@ -0,0 +1,32 @@
...
+
+open import Algebra using (Semiring)
+import Algebra.Properties.Monoid.Divisibility as MonoidDiv
+open import Data.Product using (_,_)
+
+module Algebra.Properties.Semiring.Divisibility {r₁ r₂} (R : Semiring r₁ r₂) where
+
+open Semiring R
+open import Algebra.Divisibility _≈_ _*_ using (∣-max)
+open MonoidDiv *-monoid public
```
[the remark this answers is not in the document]
mechvel: I set the comments like
```
-- Re-exporting contents of divisibility over monoids
--
open MonoidDiv *-monoid public renaming (ε∣_ to 1∣_)
```
This looks better because the dashes line inside a sequence the `open` declarations makes it difficult to read.

If you write that you insist on your version, then I shall change it to your version (or you do this yourself).

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,72 @@
...
+open Algebra.Primality _≈_ _*_ 0# 1# using (IsIrreducible; Coprime)
+open Algebra.GCD _≈_ _*_ using (GCD)
+
+0∤1 : 0# ≉ 1# → 0# ∤ 1#
+0∤1 0≉1 (q , q*0≈1) = 0≉1 (trans (sym (zeroʳ q)) q*0≈1)
+
+------------------------------------------------------------------------------
+-- Properties of Irreducibilty, Coprime.
+
+irreducible≉0 : 0# ≉ 1# → (∀ {p} → IsIrreducible p → p ≉ 0#)
```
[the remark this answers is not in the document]
mechvel: I remove these parentheses.

About ``a, b`` vs ``x, y``.
In textbooks, prime (or irreducible) elements are usually denoted `p`, `q`.
This a mnemonic carried by a variable name.

Also ``x, y, z`` usually denote more local values that are quantified and may be are often changed. 
When there are the environment parameters and there are local values ``x, y, z``, where the former are stable,
and the latter are quantified inside,
it is usulally written ``a, b, c, p, q ...`` for the environment parameters
and ``x, y, z`` for local values.
For example, in the body of ``record GCD (a b : A)``,
there may be foralled ``x, y`` to expressed the properties of ``GCD a b``.

These denotations make it easier to follow the discourse with formulas.

Please, consider this and announce the decision.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,72 @@
...
+irreducible≉0 0≉1 (_ , chooseInvertible) p≈0 =
+  let p≈0*0 = trans p≈0 (sym (zeroˡ 0#)) in
+  case chooseInvertible p≈0*0 of λ
+  { (inj₁ 0∣1) → 0∤1 0≉1 0∣1
+  ; (inj₂ 0∣1) → 0∤1 0≉1 0∣1
+  }
+
+coprimeWithInvertible : ∀ {x} y → x ∣ 1# → Coprime x y
```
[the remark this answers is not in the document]
mechvel: Because both of the notions ``Invertible x`` and ``x ∣ 1#``
(``x ∣ ε`` of monoid) are defined in algebra as ``∃ \q → q * x ≈ 1#``.  
(``∃ \q → q ∙ x ≈ ε``).  

"coprimeToInvertible" is shorter, but will not do, because the reader could think of "To" as of conversion to something.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,72 @@
...
+
+module _ (a b : A) (struc : GCD a b)
+  where
+  open GCD struc
+
+  nz⊎nz⇒≉0 : a ≉ 0# ⊎ b ≉ 0# → proper ≉ 0#
+  nz⊎nz⇒≉0 (inj₁ a≉0) = ∣nonzero⇒≉0 divides₁ a≉0
+  nz⊎nz⇒≉0 (inj₂ b≉0) = ∣nonzero⇒≉0 divides₂ b≉0
+
+  quot₁ quot₂ : A
```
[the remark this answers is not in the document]
mechvel: Moved to the GCD declaration.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,72 @@
...
+  ; (inj₂ 0∣1) → 0∤1 0≉1 0∣1
+  }
+
+coprimeWithInvertible : ∀ {x} y → x ∣ 1# → Coprime x y
+coprimeWithInvertible {x} y x∣1 z∣x _ = ∣-trans z∣x x∣1
+
+------------------------------------------------------------------------------
+-- Properties of GCD
+
+module _ (a b : A) (struc : GCD a b)
```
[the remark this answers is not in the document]
mechvel: See my earlier question.

mechvel:
Please, consider this commit of November 12, 2020.
Your notes are implemented, except the ones accompanied by the questions above.

src/Algebra/Primality.agda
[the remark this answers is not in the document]
MatthewDaggitt: > 1. My version does not require any new names to remember.

Not sure this is a valid argument. Taking it to the extreme we should just call everything `lemma1`, `lemma2`, `lemma3` which means no new names to remember.

> 2. proj₁ prime-x is a simple proof for a prime element being nonzero.

Wouldn't `Prime.nonZero` be much more readable better?

> 3. Just pattern-match against _,_ for Irreducible. And with record, one needs to remember the two particular words for the fields and one for the constructor. And the pattern for matching will be as (irredᶜ _ chooseinvertible).

As a counter-point to this, the record constructor does not need you to import `Data.Product` which you may not have in scope. Additionally records provide much better interaction with the editor. It gives you the names, individual holes with types with a single keyboard action. Sorry, but I'm pretty sure a record is the right solution in this instance.

src/Algebra/Properties/Semiring/Divisibility.agda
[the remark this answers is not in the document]
MatthewDaggitt: Please stick to the style that is used everywhere else throughout the library and do not invent your own!

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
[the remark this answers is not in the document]
MatthewDaggitt: What is this notion of `Invertible`? I'm not aware of anything called that in the library?

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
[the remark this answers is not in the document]
MatthewDaggitt: Well leaving aside the fact there is an implicit quantifier over `a` and `b` as well, maybe the best way to think about the standard library is that the library itself is a textbook. It too has it's own conventions. You wouldn't want to read a textbook that switched notation on every page right? In the same way, we don't want the library to switch the notation used in every module.

There might be some valid argument for using `p` and `q` for primes, but in the library the names `a`, `b` are reserved for universe levels. Therefore they are not appropriate to use as elements in an algebra.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,72 @@
...
+  {α ℓ} (R : CancellativeCommutativeSemiring α ℓ)
+  where
+
+open CancellativeCommutativeSemiring R renaming (Carrier to A)
+
+open SemigroupDiv *-semigroup using (_∣_; _∤_; ∣-trans)
+--
+-- Reexporting Divisibility over Semigroup, Semiring
+--
+open SemiringDiv semiring using (∣nonzero⇒≉0)
```
MatthewDaggitt, line 31: This should be opened `public`ly and everything should be re-exported. Not just `∣nonzero⇒≉0`. It should also go underneath the opening of `Algebra.Primality` and `Algebra.GCD`.

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,43 @@
...
+
+record GCD (arg1 arg2 : A) :  Set (a ⊔ ℓ) where      -- a result of gcd
+  constructor gcdᶜ
+  -- Greatest common divisor.
+
+  field
+    value    : A                  -- the proper gcd value
+    divides₁ : value ∣ arg1
+    divides₂ : value ∣ arg2
+    greatest : ∀ {x} → (x ∣ arg1) → (x ∣ arg2) → (x ∣ value)
```
MatthewDaggitt, line 33: No brackets needed here

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,43 @@
...
+
+  quot₁ quot₂ : A
+  quot₁ = proj₁ divides₁
+  quot₂ = proj₁ divides₂
+
+  quot₁*value≈arg1 : (quot₁ * value) ≈ arg1
```
MatthewDaggitt, line 39: No brackets needed here, same below.

src/Algebra/Primality.agda
```diff
@@ -0,0 +1,55 @@
...
+-- Definitions
+
+private
+  infix 4 _≉_
+
+  _≉_ : Rel A _
+  x ≉ y = ¬ (x ≈ y)
+
+Irreducible : Pred A (a ⊔ ℓ)
+Irreducible p = p ∤ 1#  ×  (∀ {x y} → (p ≈ (x * y)) → x ∣ 1# ⊎ y ∣ 1#)
```
MatthewDaggitt, line 39: No brackets needed for `p ≈ (x * y)`

src/Algebra/GCD.agda
```diff
@@ -0,0 +1,43 @@
...
+open import Relation.Binary using (Rel)
+
+module Algebra.GCD {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) (_*_ : Op₂ A) where
+
+open import Algebra.Divisibility _≈_ _*_
+
+------------------------------------------------------------------------------
+-- Definitions
+
+record GCD (arg1 arg2 : A) :  Set (a ⊔ ℓ) where      -- a result of gcd
```
MatthewDaggitt, line 25: Would it make sense to have an `IsGCD x y z` record that had the GCD value in the type as well? Then you could define `GCD` in terms of `IsGCD`?

src/Algebra/Primality.agda
```diff
@@ -0,0 +1,55 @@
...
+
+Prime : Pred A (a ⊔ ℓ)
+Prime p = p ≉ 0#  ×  p ∤ 1#  ×  ∀ {x y} → p ∣ (x * y) → p ∣ x ⊎ p ∣ y
+
+-- * It is easy to prove        Prime ==>  Irreducible.
+-- * In a GCDSemiring it holds  Prime <==> Irreducible
+--   (for example, in ℕ, ℤ).
```
MatthewDaggitt, line 46: I'm not sure these comments add anything. If it's easy to prove then it will be done so somewhere in the library. We also don't yet have a notion of `GCDSemiring` so we shouldn't have comments referring to it yet!

src/Algebra/Properties/Semiring/Divisibility.agda
```diff
@@ -0,0 +1,31 @@
...
+------------------------------------------------------------------------------
+-- Divisibility properties specific to Semiring.
+
+_∣0 : ∀ x → x ∣ 0#
+x ∣0 = 0# , zeroˡ x
+
+0∣x⇒x≈0 : ∀ {x} → 0# ∣ x → x ≈ 0#
+0∣x⇒x≈0 (q , q*0≈x) = trans (sym q*0≈x) (zeroʳ q)
+
+∣nonzero⇒≉0 : ∀ {x y} → x ∣ y → y ≉ 0# → x ≉ 0#
```
MatthewDaggitt, line 30: I think `x∣y∧y≉0⇒x≉0` would be a more consistent and descriptive name for this property.

src/Algebra/Primality.agda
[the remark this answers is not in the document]
mechvel: All right, I define  Irreducible, Prime  as records.

src/Algebra/Properties/Semiring/Divisibility.agda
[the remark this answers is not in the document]
mechvel: Sorry, I have forgotten to return these headings to your style.
Let us do this in the next commit.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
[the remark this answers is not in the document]
mechvel: So,  `a` and `b ` are reserved for Level
(probably, changing them will break backwards compatibility). 
I have set this in the commit.

But I understand your sentence about `p` and `q` so that you allow them in the definitions  of
Irreducible and Prime.

> maybe the best way to think about the standard library is that the library itself is a textbook.

It is natural for the _algebra part_ of the library to preserve
_some_ traditions of the algebra textbooks, just not to neglect everything in a rough way.

This is for the following reason.
a) Among the users of this algebra part 99% are mathematicians by education
(if you ask why, I would explain).
And they are used to follow traditions. For example, formally, the sentence
"Let x be any _group_ and y any its subgroup."  is all right. But this will cause a disgust,
unless these denotations are explained by  a very particular reason.
Rather people expect expressions like
"Let G (A, B ...) be any group and H (G', A' ...) any its subgroup."
These traditions also have a certain sense in them, mathematical denotations have a five century experience.
We only need not to neglect everything in without a technical reason.
If we have an opportunity to denote ``(G : Group _ _)``, then all right.
If some technical reasons require (like in Haskell) a small first letter,
OK, write, say, ``(gG : Group _ _)``. And the last and the worst solution will be, say, 
``(x1 : Group _ _)``, not a great tragedy, but just unpleasant.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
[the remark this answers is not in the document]
mechvel: Textbooks write like this:
An element `x` in a Monoid is called left-invertible iff
there exists q such that  ``q * x ≈ ε``.

And this coincides with our definitions of  ``_∣_,  x ∣ ε``.

Firther, in CommutativeMonoid, and in any Group,
left-invertible <==> right-invertible,
but not in the general case.

An element  x  in a Monoid  is called invertible iff
```
∃ \q → q ∙ x ≈ ε  ×  x ∙ q ≈ ε
```
, that is x has a left inverse q which is also a right inverse for x.

Now I think:
left-invertibe x  is expressed as  ``x ∣ ε``
- which is "x divides ε from right"
``(∃ \q → q ∙ x ≈ ε)``.  

So, (right-invertibe x)  needs to correspond to "x divides ε from left".

Assuming all this, I introduce _divisibility from left_   ``_∣ˡ_`` 
and divisibilty from right ``_∣ʳ_``.
They are equivalent in any commutative magma and in any group, but not in the general case.
I put
```
x ∣ˡ y = ∃ λ q → (x ∙ q) ≈ y    -- x divides y from left                          
x ∣ʳ y = ∃ λ q → (q ∙ x) ≈ y    -- x divides y from right                         

_∣_ = _∣ʳ_
```.
The latter is set because in many domains left and right divisibilities
are equivalent (as in ℕ and ℤ), and there it is simpler to write `_∣_`.
See Algebra.Divisibility.
Is this all right?

File: src/Algebra/Properties/CancellativeCommutativeSemiring.agda (hunk 21 above)
Comment (mechvel) [reply src/Algebra/Properties/CancellativeCommutativeSemiring.agda]: I set this as 
```
open Algebra.Primality _≈_ _*_ 0# 1# using (Irreducible; irredᶜ; Coprime)
open Algebra.GCD _≈_ _*_ using (GCD)
open SemigroupDiv *-semigroup public
--
-- Reexporting divisibility over Semigroup, Semiring
--
open SemiringDiv semiring public using (∣nonzero⇒≉0)
```
(as I understand you, this commentary needs to change to your style).

If I delete the last `using`, then there is an overlap with
``open SemigroupDiv *-semigroup public``.

Please, advise.

File: src/Algebra/GCD.agda (hunk 22 above)
Comment (mechvel) [reply src/Algebra/GCD.agda]: Removed brackets from this line.
But below, we have
```
quot₁*value≈x : (quot₁ * value) ≈ x
```
, and Agda insists on brackets, probably because the priority of
`*` and `≈` is not visible here.

File: src/Algebra/GCD.agda (hunk 23 above)
Comment (mechvel) [reply src/Algebra/GCD.agda]: See the preceding comment.

File: src/Algebra/Primality.agda (hunk 24 above)
Comment (mechvel) [reply src/Algebra/Primality.agda]: Following your record requirement, I set
```
record Irreducible (p : A) : Set (a ⊔ ℓ) where
  constructor irredᶜ
  field
    ∤1                     : p ∤ 1#
    chooseInvertibleFactor : ∀ {x y} → p ≈ (x * y) → x ∣ 1# ⊎ y ∣ 1#
```
As to parentheses in the last line, Agda insists of them, because the priorities are not visible.

File: src/Algebra/GCD.agda (hunk 25 above)
Comment (mechvel) [reply src/Algebra/GCD.agda]: I was glad that you did not not recall of such a thing. And now you do.
Do you suggest 
```
record IsGCD (x y z : A) : Set _ where                                            
  field                                                                         
    commonDivisor : z ∣ x × z ∣ y 
    greatest      : ∀ {u} → u ∣ x → u ∣ y → u ∣ z
```
(similar to Nat.GCD in `master`) ?
And then, how do you define the GCD type and the function
```gcd : (x y : A) → GCD x y
```
in the GCDSemiring record?  IsGCD and GCD - to much of them, complication ...

Traditionally, there is only GCD with two parameters, and  gcd x y  returns  GCD x y.
Classically, it returns a value in the type A,
but in constructive algebra, it should return a structure which also contains witnesses.

Can you, please, remain GCD as it is: 
```
record GCD (x y : A) :  Set (a ⊔ ℓ) where  
  constructor gcdᶜ                                                                
  field                                                                           
    value    : A          
    divides₁ : value ∣ x  
    divides₂ : value ∣ y     
    greatest : ∀ {z} → z ∣ x → z ∣ y → z ∣ value
  ...
```
?

src/Algebra/Primality.agda
mechvel: I set now
```
-- * It will be proved  Prime ==> Irreducible.
-- * In many special cases it holds  Prime <==> Irreducible
--   (for example, in ℕ, ℤ).
```
I think this is helpful, because many people know what is Prime for ℕ and ℤ.
And they can see that there it is equivalent to Irreducible.
And they will doubt of why Irreducible is at all introduced.

If you insist, then I shall remove this comment.

src/Algebra/Properties/Semiring/Divisibility.agda
mechvel: Replaced.

mechvel:
Please, consider this my commit of November 14, 2020.
Your notes are implemented except the ones commented and provided with questions.

src/Algebra/GCD.agda
MatthewDaggitt: Yes I suggest doing it the same way we do it in `Data.Nat.GCD`.

>  And then, how do you define the GCD type and the function, gcd : (x y : A) → GCD x y

I would either go
```agda
  gcd : A → A → A
  gcd-isGCD : IsGCD x y (gcd x y)
```
or
```agda
  gcd : ∃ λ g → IsGCD x y g
```
I prefer the first because it allows you to access the raw `gcd` function without having to project out of the pair.

The advantage of having the value as a parameter to the record is that it's much easier to write types for the proofs. For example consider the following result:
```agda
GCD-* : ∀ {m n d c} → IsGCD (m * c) (n * c) (d * c) → IsGCD m n d
```
Using your current definition, you would have to write something like:
```agda
GCD-* : ∀ {m n c} (g1: GCD (m * c) (n * c)) (g2 : GCD m n d) → GCD.value g1 ≡ c * GCD.value g2
```
naming the input records, and extracting them out of the bundle, and resulting in an equality rather than a new `IsGCD` instance.

MatthewDaggitt:
Apologies for the delay in getting back to you on this @mechvel. Once we've settled the `GCD` vs `IsGCD` question I think it's pretty much ready to go.

mechvel:
> Once we've settled the GCD vs IsGCD question I think it's pretty much ready to go.

1) I doubt in English:  by saying "Once we've settled Foo it's  ready to go."
do you mean
"It remains for us to settle Foo, and then it will be ready to go."
?
2)  About your GCD version: 
Your last note about IsGCD looks reasonable. But I hesitate.
Can you change the all the current GCD items as you like, to remain my version or to rewrite it to your version,  and push it
?
Then, I would consider the new `master` and would try to do the next PR?
(by deleting my current fork, cloning down `master`, and creating the next PR to continue the GCD items).
?

3) Also I have observed occasionally the new suggested code for  exponentiation `_^_`,  and `sum` in 
    CommutativeMonoid, and in Semiring.  
It needs a great improvement.
a) `Commutative` is not necessary for most of these items.
   Because  each submonoid  {ε, x, x^2, x^3 ...}  is commutative itself. So that the proofs can be formed by implicitly using this fact.
  
b) The functions `_^_`,  and `sum` in Semiring can be imported from the Monoid theory module,
     the first one from the `*-monoid` instance, the second  from the `+-monoid` instance (with renaming * to +, 
product  `Π` to `sum`), together with the property proofs.

You can fix it yourself.
Or if you ask me, I  can provide all this -- only after the GCD project is completed.
I already have all the proofs for  ` `_^_, product, sum`` for the approach of (a), (b).
But I know that pushing any evident thing to standard takes long.

mechvel:
> Because each submonoid {ε, x, x^2, x^3 ...} is commutative itself. So that the proofs can be formed by implicitly using this fact.

Briefly, one has to first prove for Monoid 
```
xᵐ-comm-xⁿ :  ∀ x m n → (x ^ m) ∙ (x ^ n) ≈ (x ^ n) ∙ (x ^ m)
``` 
, and all the rest will be more or less easily derived from Monoid instead of CommutativeMonoid.
In fact I have all this proved for  `_^_` implemented by the binary method. First prove the properties for 
`^naive`, then prove   ``^naive x n ≈  ^ = ^binary``, and ...  
And ``^binary`` is not so difficult, due to the arithmetic contained in Nat.Binary.

MatthewDaggitt:
> Do you mean "It remains for us to settle Foo, and then it will be ready to go."?

Yes

> Please, take a decision yourself: to remain my version or to rewrite it to your version.

I have changed it to `IsGCD`

> Also I have observed occasionally the new suggested code for exponentiation _^_, and sum in
CommutativeMonoid, and in Semiring.
It needs a great improvement.
a) Commutative is not necessary neither for defining them nor for most of these proofs.

See `Algebra.Properties.Monoid.(Exponentiation/Summation)` for properties of the two operations which do not require commutativity.

MatthewDaggitt:
Merged in @mechvel

PR #1348, opened 2020-11-08
[ new ] inductive definition of Infix

gallais:


src/Data/List/Relation/Binary/Infix/Heterogeneous/Properties.agda
```diff
@@ -0,0 +1,70 @@
...
+fromSuffix (there p) = there (fromSuffix p)
+
+∷⁻ : ∀ {as b bs} → Infix R as (b ∷ bs) → Prefix R as (b ∷ bs) ⊎ Infix R as bs
+∷⁻ (here pref) = inj₁ pref
+∷⁻ (there inf) = inj₂ inf
+
+------------------------------------------------------------------------
+-- length
+
+module _ {a b r} {A : Set a} {B : Set b} {R : REL A B r} where
```
MatthewDaggitt, line 56: Given the variables, do we need the anonymous module?

gallais: Good catch. This is more or less directly lifted from the `Prefix` version of
the properties file and it still uses the old style. :(

PR #1360, opened 2020-11-13
[ new ] unsnoc

gallais:
It's pretty convenient to have `uncons` and `unsnoc` with very similar types
(e.g. to use them both in a `Maybe` do block).


MatthewDaggitt:
I've only just realised now that `unsnoc` is `uncons` reversed. It would be great if we could add a comment about `snoc` and `unsnoc` to that effect?

PR #1364, opened 2020-11-23
[ refactor ] IO module

gallais:
Once again: given that sized IO is not around the corner, we may as
well make the best of what we have. This means making it easier to
use `do`-notations when we don't actually need to rely on guardedness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@ Non-backwards compatible changes
   were causing cyclic import dependencies.
 
 * Clean up of `IO` to make it more friendly:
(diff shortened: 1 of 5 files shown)


gallais:
Not sure whether the infix names I have picked are very good.
Maybe these constructors can just be prefix and we can define
a ``_`_`_`` combinator in `Function.Base` for people who want
to use them infix.

JacquesCarette:
Nice. This looks like a definite improvement. I don't have suggestions for better names, other than I don't think your suggested combinator is a good idea.

PR #1365, opened 2020-11-24
A verified mergesort implementation

MatthewDaggitt:
After 13 years, it finally feels like the right time for the standard library to provide a sorting algorithm :laughing: I've had an implementation of insertion sort lurking around in my own repo for forever, but finally thought I'd take the time to put together something more standard library worthy. It's surprising how much extra machinery was actually needed...

I've gone for a fairly naive implementation so I'd appreciate someone taking a look over the code, and pointing out anywhere where it could be improved. In particular I've taken the first element as the pivot element, as I'm not aware of a safe way of picking a random element, but let me know if anyone know's of any tricks to get around that.

Users shouldn't import `Data.List.Sort.Quicksort` module directly, but instead import `Data.List.Sort` which exposes the function and proofs inside an `abstract` block to hide the implementation details. This means that hopefully we should be able to switch out the "standard" sorting algorithm if someone comes along and implements something better...


jespercockx:
Thanks for this PR, a proper sorting algorithm has indeed been missing for a very long time! I'm a bit surprised though that you went for quicksort rather than mergesort. I thought that the real benefits of quicksort don't really show up unless you do it in-place and without those benefits mergesort would be easier to implement and verify. It also doesn't require the choice of a pivot element. With the current implementation (choosing the first element as the pivot) I expect performance will be very bad on lists that are already (almost) sorted, which is a shame.

For an important algorithm like this, we should probably try to benchmark it somehow to ensure that it actually has the intended complexity class, both when running inside Agda and when compiling to Haskell. The current gratuitous use of list concatenation worries me a bit on that front; it might be necessary to use difference lists instead (do we have those in the library?) It would also make sense to look at the generated Haskell code for the function `sort'` and make sure that the code is doing something more or less reasonable.

jespercockx:
One more remark: it seems to me that the current algorithm is not stable (i.e. it might swap the position of elements that are equal according to the underlying equality of the order but not definitionally equal).

src/Data/List/Base.agda
```diff
@@ -252,6 +250,11 @@ tail : List A → Maybe (List A)
 tail []       = nothing
 tail (_ ∷ xs) = just xs
 
+last : List A → Maybe A
+last []           = nothing
+last (x ∷ [])     = just x
+last (x ∷ y ∷ xs) = last (y ∷ xs)
```
gallais, line 256:
```suggestion
last (x ∷ xs) = last xs
```

MatthewDaggitt:
Thanks for the feedback @jespercockx, it's exactly what I was after :+1: For myself, I'm purely interested that a provably correct sorting algorithm that exists so I've overlooked all these performance related issues.

> I'm a bit surprised though that you went for quicksort rather than mergesort. I thought that the real benefits of quicksort don't really show up unless you do it in-place and without those benefits mergesort would be easier to implement and verify. It also doesn't require the choice of a pivot element. With the current implementation (choosing the first element as the pivot) I expect performance will be very bad on lists that are already (almost) sorted, which is a shame.

All excellent points. Maybe I should have another go at merge sort.

> For an important algorithm like this, we should probably try to benchmark it somehow to ensure that it actually has the intended complexity class, both when running inside Agda and when compiling to Haskell.

Agreed. The standard library test suite doesn't yet have any performance benchmarks. Presumably there's some in the Agda repo somewhere that I can take and adapt?

> The current gratuitous use of list concatenation worries me a bit on that front; it might be necessary to use difference lists instead (do we have those in the library?) 

We do have difference lists
https://github.com/agda/agda-stdlib/blob/edba5a5150a385246df5c145900b3926ff839c1d/src/Data/DifferenceList.agda#L26

but the problem is that we don't have any of the infrastructure required to prove properties about them (e.g. `All`, `Permutation`, `Sorted` etc.). Adding the infrastructure and the associated proofs would be a very large project...

We also probably wouldn't want people to have to translate their lists into difference lists in order to sort them properly. How is this handled in Haskell?

src/Data/List/Base.agda
MatthewDaggitt: Out of interest, are there any advantages to this aside from it being shorter?

gallais: It may save a tiny bit by not having to reallocate a cons
node that is virtually equal to the one taken apart but it's
mostly a readability concern tbh.

jespercockx:
> Agreed. The standard library test suite doesn't yet have any performance benchmarks. Presumably there's some in the Agda repo somewhere that I can take and adapt?

I'm not sure what would be the best way to benchmark this, so I'm summoning @UlfNorell as the expert on this topic.

> We also probably wouldn't want people to have to translate their lists into difference lists in order to sort them properly. How is this handled in Haskell?

It's not a problem to convert from/to difference lists *once*, so the function exposed to the user could keep the same type. Since `DiffList A` is just a synonym for `List A -> List A`, I'd hope that you would be able to use a lot of the metatheory of regular functions. OTOH, it could be a problem that this type allows many more elements that do not represent a list, so it might be necessary to keep around proofs of important invariants of difference lists, i.e. that for each `f : DiffList A`, we have `f xs = f [] ++ xs` for all `xs` (or more generally, that `f (xs ++ ys) = f xs ++ ys`). If that's indeed necessary, then we would have to show that this invariant is preserved by all the operations on difference lists, which would be quite an undertaking.

UlfNorell:
*(appears)* 🧞‍♂️

For compile-time performance I use [agda-bench](https://github.com/UlfNorell/agda-bench). It might be tricky to get that into CI (since it's implemented as an Agda backend), but for benchmarking-at-home it works well. For run-time benchmarking compile to Haskell and use criterion. I have Agda bindings to criterion here: [UlfNorell/agda-criterion](https://github.com/UlfNorell/agda-criterion).

gallais:
> OTOH, it could be a problem that this type allows many more elements
  that do not represent a list, so it might be necessary to keep around
  proofs of important invariants of difference lists,

I am reminded of [this thread](https://lists.chalmers.se/pipermail/agda/2011/003420.html) started by @asajeffrey.
This work may have been done already and we could integrate it to the stdlib.

src/Data/List/Properties.agda
```diff
@@ -818,6 +819,13 @@ module _ {P : Pred A p} (P? : Decidable P) where
   ...  | true  = cong (Prod.map (x ∷_) id) (partition-defn xs)
   ...  | false = cong (Prod.map id (x ∷_)) (partition-defn xs)
 
+  length-partition : ∀ xs → (let (ys , zs) = partition P? xs) →
+                     length ys ≤ length xs × length zs ≤ length xs
+  length-partition []       = z≤n , z≤n
+  length-partition (x ∷ xs) with does (P? x) | length-partition xs
```
JacquesCarette, line 825: why do you have to `with` on `length-partition`? In fact, why not use `if_then_else` here?

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -335,6 +337,20 @@ module _ {P : Pred A p} {Q : Pred A q} where
...
+module _ {P : A → Set p} where
+
+  head⁺ : ∀ {xs} → All P xs → Maybe.All P (head xs)
+  head⁺ []       = nothing
+  head⁺ (px ∷ _) = just px
+
+  last⁺ : ∀ {xs} → All P xs → Maybe.All P (last xs)
+  last⁺ []             = nothing
+  last⁺ (px ∷ [])      = just px
+  last⁺ (px ∷ py ∷ xs) = last⁺ (py ∷ xs)
```
JacquesCarette, line 352: same pattern here as @gallais pointed out above.

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -533,6 +549,16 @@ module _ {P : A → Set p} {Q : A → Set q} (P? : Decidable P) where
...
 
+------------------------------------------------------------------------
+-- partition
+
+module _ {P : A → Set p} (P? : Decidable P) where
+
+  partition-All : ∀ xs → (let ys , zs = partition P? xs) →
+                  All P ys × All (∁ P) zs
+  partition-All xs rewrite partition-defn P? xs =
+    all-filter P? xs , all-filter (∁? P?) xs
```
JacquesCarette, line 560: This does two full passes. Yes, it's going to be a lot easier to reason about. It still feels... wrong.

JacquesCarette (changes requested):

JacquesCarette:
The usual cheap thing to do for a better-behaved Quicksort is to use 'median of 3' as pivot, where the 3 are usually taken as first, middle and last.  The sets that have worst-case behaviour for that are truly quite weird, and it behaves quite well for both almost-sorted and almost-reversed sets.  Of course, that's pathetically slow on linked lists, but there are so (so!) many sources of slowness here, it's probably not worth worrying about either way.

src/Data/List/Properties.agda
gallais: We need the `with` on `does (P? x)` for `partition P? (x ∷ xs)` to reduce.
The additional `with` on `length-partition xs` is a standard technique to get
on the good side of the termination checker now that the (buggy) `with`-inliner
has been removed.

src/Data/List/Relation/Unary/All/Properties.agda
gallais: Here we will not be able to avoid the split because we need `last`
to reduce in the type. The best we can hope for is something like
`last⁺ (px ∷ pxs@(_ ∷ _)) = last⁺ pxs` which is not necessarily
more readable.

JacquesCarette: Thanks - makes sense.

src/Data/List/Properties.agda
JacquesCarette: Thanks for the first answer, it was not obvious to me that things in the type needed to reduce. This is subtle, and we ought to have a better way to do that than `with`. I was afraid that termination was indeed the reason for the other `with`; sad, but oh well.

JacquesCarette (approved):

MatthewDaggitt:
Okay so I've switched from quick sort to merge sort, whose implementation is a lot less likely to have asymptotically sub-optimal behaviour. I'll see what I can do on the benchmarking front in the next couple of days.

MatthewDaggitt:
Okay, so the answer of whether we have `O(nlog(n))` behaviour is a resounding "No" :sob: 

List length, time
10, 1sec
100, 1sec
1000, 5 sec
10000, 21 minutes and counting

The problem appears to be the merge function:
```agda
merge : {R : Rel A ℓ} → B.Decidable R → List A → List A → List A
merge R? []       ys       = ys
merge R? xs       []       = xs
merge R? (x ∷ xs) (y ∷ ys) = if does (R? x y)
  then x ∷ merge R? xs (y ∷ ys)
  else y ∷ merge R? (x ∷ xs) ys
```
where `R? = Data.Nat._≤?_`. I've tested `does (m ≤? n)` and as hoped it appears to be fast even for arbitrarily large integers, so I think it must be something to do with the definition itself. I'm a bit lost as to what it might be though.

MatthewDaggitt:
Another issue is that I had hoped to use `abstract` in `Data.List.Sort` to hide the implementation details and so we could swap in different sorting algorithms later, but of course you then can't get `sort` to compute in types. Do people think this trade-off is worth it? I guess people who do need it to compute in types, can just import `Data.List.Sort.MergeSort` instead?

MatthewDaggitt:
Still haven't managed to track down why the performance is so bad so I think this is going to have to be pushed to v1.6...

gallais:
In case someone else wants to have a look, this is what I came up with.
You just have to install agda-bench (@UlfNorell: It's great! I should have
tried it earlier!) and call it on the file (tweaking `n`).

```agda
module Test where

import Data.List.Sort.MergeSort as Sort
open import Data.List
open import Data.Nat
open import Data.Nat.DivMod
open import Relation.Nullary.Decidable using (False)

sort : List ℕ → List ℕ
sort = Sort.sort ≤-decTotalOrder

-- Feels like we should have pseudo-random generator in the lib
record Generator : Set where
  field multiplier  : ℕ
        increment   : ℕ
        modulus     : ℕ
        {modulus≢0} : False (modulus ≟ 0)

step : Generator → ℕ → ℕ
step gen x =
  let open Generator gen in
  ((multiplier * x + increment) % modulus) {modulus≢0}

-- Taken from
-- https://en.wikipedia.org/wiki/Linear_congruential_generator#Parameters_in_common_use

random0 : Generator
random0 = record
  { multiplier = 8121
  ; increment  = 28411
  ; modulus    = 134456
  }

list : ℕ → Generator → ℕ → List ℕ
list zero    gen x = []
list (suc k) gen x = x ∷ list k gen (step gen x)

n  = 10000
x₀ = 666

bench-list    = list n random0 x₀
bench-sum     = sum bench-list
bench-reverse = sum (reverse bench-list)
bench-sort    = sum (sort bench-list)

-- n     | sum      | reverse   | sort
-- 10    | 65.35 μs | 86.04 μs  | 666.2 μs
-- 100   | 569.9 μs | 851.1 μs  | 22.94 ms
-- 1000  | 4.998 ms | 9.641 ms  | 2.147 s
-- 10000 | 59.75 ms | 119.5 ms  | (still waiting)
```

gallais:
```agda
mergeAll′ : ℕ → List (List A) → List A
mergeAll′ 0 xss = []
mergeAll′ _ (xs ∷ []) = xs
mergeAll′ (suc n) xss = mergeAll′ n (mergePairs xss)

sort′ : List A → List A
sort′ xs = mergeAll′ (length xs) (map [_] xs)
```

yield massive improvements:

```
n     | sum      | reverse  | sort     | sort′
10    | 65.35 μs | 86.04 μs | 666.2 μs | 506.4 μs
100   | 569.9 μs | 851.1 μs | 22.94 ms | 6.110 ms
1000  | 4.998 ms | 9.641 ms | 2.147 s  | 90.05 ms
10000 | 59.75 ms | 119.5 ms | 293.6 s  | 1.341 s
```

So I suspect the cost is coming from the `<-wellfounded` part of the definition.
Note that this is all erased

MatthewDaggitt:
Thanks for the analysis @gallais. I guess it makes sense that constructing the unary proof for `_<_` required for `Acc` is causing quadratic behaviour at type-checking time.

> do we care enough
about compile time evaluation of sort to change the definition to the less elegant one?

It's only moderately less elegant right? Feed in the fuel, and a proof that the fuel is greater than the size of the list to the proofs? I guess I'm probably in favour.

Been trying to install `agda-bench` to verify your results and get some automated tests up and running but currently blocked by https://github.com/agda/agda/issues/5167 as cabal takes the development version of Agda. Anyone know of a neat way to force cabal to use the Agda 2.6.1 library rather than 2.6.2-dev library when installing it?

UlfNorell:
The way I've done this in [agda-prelude](https://github.com/UlfNorell/agda-prelude/blob/master/src/Control/WellFounded.agda) is to provide a proof of well-foundedness that has a large number of free steps.

MatthewDaggitt:
> The way I've done this in agda-prelude is to provide a proof of well-foundedness that has a large number of free steps.

That is incredibly sneaky :laughing:

UlfNorell:
One thing you have to be a little bit careful about is to always be strict in some other argument than the accessibility proof, otherwise you will have neutral terms unfolding a billion times before getting stuck.

MatthewDaggitt:
@UlfNorell's suggestion does indeed fix the problem. Execution of time of `sort` drops down from 2.147s for n=1000 to 117.9ms.

src/Data/Nat/Induction.agda
```diff
@@ -83,6 +83,22 @@ module _ {ℓ} where
...
+-- A version of `<-wellFounded` that cheats by skipping building
+-- the first billion proofs. Use this when you require the function
+-- needs to evaluate fast.
+--
+-- IMPORTANT: You have to be a little bit careful to always make the
+-- function be strict in some other argument than the accessibility proof,
+-- otherwise you will have neutral terms unfolding a billion times before
+-- getting stuck.
+<-wellFounded-fast : WellFounded _<_
+<-wellFounded-fast = <-wellFounded-skip {1000000000}
```
gallais, line 95: Can we make the argument to `<-wellFounded-skip` explicit given that we
explicitly pass it & then explicitly match on it? :D

MatthewDaggitt: Done!

PR #1366, opened 2020-12-02
[ new ] postulate length-tail lemma

gallais:


MatthewDaggitt:
That's a pretty unreadable type, but I can't think how to improve it so...

JacquesCarette:
If readability is what you want, then a pattern-matching lambda and being pointed instead of the current composition. And, if you want to go all out, use `1 +` instead of `suc` for extra readability.

gallais:
Using an anonymous pattern-matching lambdas has its inconveniences
e.g. you can't copy/paste the RHS when doing equational reasoning
because every single pattern-matching lambda gets a unique internal
name and two similar-looking one will end up not being equal.

So I'd rather stick to the call to `maybe`.

MatthewDaggitt:
Hmm coming back to this again, we could define a `head` and `tail` operations for `String`s in terms of `uncons`? That would then make the type  `length s ≡ maybe′ (suc ∘′ length) zero (tail s)` which is definitely a bit of an improvement?

PR #1367, opened 2020-12-04
[ new ] bindings for Haskell's System.Exit

gallais:
Continuing on my quest to write executables in Agda.


gallais:
Blocked by #1369: once `putStrLn` is level-polymorphic, we can clean up
the definition of `die`!

PR #1371, opened 2020-12-07
Various AVL goodies

gallais:


gallais:
Let me put this back to being-worked-on and extend this PR to adding:
* union for Sets
* (inefficient) intersection for all of these

And any other thing I find I wish was there while doing the advent of code in Agda.

PR #1372, opened 2020-12-07
[ fix ] lines should not drop empty lines

gallais:
Fix for https://github.com/agda/agda-stdlib/commit/eb338fd27861f8c9db1359fda6e0ad24df555ae0#r44894245

### DIFF
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -385,6 +385,17 @@ unsnoc as with initLast as
 ------------------------------------------------------------------------
 -- Splitting a list
 
+-- linesBy preserves the empty lines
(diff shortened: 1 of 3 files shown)


src/Data/List/Base.agda
```diff
@@ -385,6 +385,17 @@ unsnoc as with initLast as
 ------------------------------------------------------------------------
 -- Splitting a list
 
+-- linesBy preserves the empty lines
```
MatthewDaggitt, line 388: I think these comments for `linesBy` and `wordsBy` could be improved. How do "empty lines" and "empty words" relate to the decidable predicate `P`? They probably shouldn't mention whitespace when we're not talking about `String` or `Char`?

PR #1374, opened 2020-12-08
Add min/max/absolute value for ℚ

uzytkownik:
Various improvements to ℚ and ℤ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -264,6 +264,163 @@ Other minor additions
   applyUpTo⁺ : m ≤ n → applyUpTo f m ⊆ applyUpTo f n
   ```
 
+* Add new functions to `Data.Rational`:
+  ```agda
+  _⊔_ : (p q : ℚ) → ℚ
+  _⊓_ : (p q : ℚ) → ℚ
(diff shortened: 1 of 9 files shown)


MatthewDaggitt:
Thanks for the PR @uzytkownik, this is fantastic. However it is a huge amount to review in one go! Please could you split it into a few PRs. That should allow us to actually do a reasonable review on each of the parts and not have this PR stay open forever. The split I would suggest is as follows:

1. Lattice morphisms
2. Naturals and integers
3. Codata
3. Unnormalised rationals
4. Normalised rationals

The first three PRs can be submitted at the same time :+1:

uzytkownik:
It looks like I made some mistakes during rebase so I need to fix them first. I will split the changes.

MatthewDaggitt:
@uzytkownik given everything else has been merged in, I've rebased this branch to only contain the changes for rationals so that you still get the git credit for adding them. I'll go through and make the changes to the new style of min/max proofs.

MatthewDaggitt:
Thanks for the PRs @uzytkownik, glad we managed to merge all of this one in!

PR #1375, opened 2020-12-09
Add several Data.Nat.Properties

uzytkownik:


src/Data/Nat/Properties.agda
```diff
@@ -1185,6 +1190,46 @@ m⊔n≤m+n m n with ⊔-sel m n
...
 ... | inj₂ m⊔n≡n rewrite m⊔n≡n = m≤n+m n m
 
+------------------------------------------------------------------------
+-- Other properties of _⊔_ and _*_
+
+*-distribˡ-⊔ : _*_ DistributesOverˡ _⊔_
+*-distribˡ-⊔ m zero o = begin-equality
+  m * (zero ⊔ o)                                        ≡⟨⟩
+  m * o                                                 ≡⟨⟩
+  zero ⊔ m * o                                          ≡˘⟨ cong (_⊔ m * o) (*-zeroʳ m) ⟩
```
MatthewDaggitt, line 1200: I think it's probably overkill to expand out the cases that don't require multiple steps. This case can probably just be `cong (_⊔ m * o) (*-zeroʳ m)`.

src/Data/Nat/Properties.agda
```diff
@@ -1185,6 +1190,46 @@ m⊔n≤m+n m n with ⊔-sel m n
 ... | inj₁ m⊔n≡m rewrite m⊔n≡m = m≤m+n m n
 ... | inj₂ m⊔n≡n rewrite m⊔n≡n = m≤n+m n m
 
+------------------------------------------------------------------------
+-- Other properties of _⊔_ and _*_
+
+*-distribˡ-⊔ : _*_ DistributesOverˡ _⊔_
+*-distribˡ-⊔ m zero o = begin-equality
+  m * (zero ⊔ o)                                        ≡⟨⟩
```
MatthewDaggitt, line 1198: Cleaner and easier to read I think if you use `0` here and elsewhere in the equational reasoning rather `zero`.

src/Data/Nat/Properties.agda
```diff
@@ -1185,6 +1190,46 @@ m⊔n≤m+n m n with ⊔-sel m n
...
+*-distribˡ-⊔ m (suc n) (suc o) = begin-equality
+  m * (suc n ⊔ suc o)                                   ≡⟨⟩
+  m * suc (n ⊔ o)                                       ≡⟨ *-suc m (n ⊔ o) ⟩
+  m + m * (n ⊔ o)                                       ≡⟨ cong (m +_) (*-distribˡ-⊔ m n o) ⟩
+  m + (m * n ⊔ m * o)                                   ≡⟨ +-distribˡ-⊔ m (m * n) (m * o) ⟩
+  (m + m * n) ⊔ (m + m * o)                             ≡˘⟨ cong₂ _⊔_ (*-suc m n) (*-suc m o) ⟩
+  (m * suc n) ⊔ (m * suc o)                             ∎
+
+*-distribʳ-⊔ : _*_ DistributesOverʳ _⊔_
+*-distribʳ-⊔ m zero o = refl
```
MatthewDaggitt, line 1216: Much easier to use `comm+distrˡ⇒distrʳ` from `Algebra.Consequences.Propositional` for the right versions of distributivity.

https://github.com/agda/agda-stdlib/blob/eb338fd27861f8c9db1359fda6e0ad24df555ae0/src/Algebra/Consequences/Propositional.agda#L76

src/Data/Nat/Properties.agda
```diff
@@ -286,6 +286,11 @@ n≤0⇒n≡0 z≤n = refl
 ≤∧≢⇒< {_} {suc n} (s≤s m≤n) 1+m≢1+n =
   s≤s (≤∧≢⇒< m≤n (1+m≢1+n ∘ cong suc))
 
+≤∧≮⇒≡ : ∀ {m n} → m ≤ n → m ≮ n → m ≡ n
+≤∧≮⇒≡ {zero} {zero} z≤n m≮n = refl
+≤∧≮⇒≡ {zero} {suc n} z≤n m≮n = contradiction (s≤s z≤n) m≮n
+≤∧≮⇒≡ {suc m} {suc n} (s≤s m≤n) m≮n = cong suc (≤∧≮⇒≡ m≤n (m≮n ∘ s≤s))
```
MatthewDaggitt, line 292: Please can you align the function arguments as per the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#function-arguments) as it tends to make proofs easier to read. At some point someone is going to write a linter for Agda and everyone will be happy :smile:

CHANGELOG.md
```diff
@@ -323,6 +323,19 @@ Other minor additions
   ++↭ʳ++ : xs ++ ys ↭ xs ʳ++ ys
   ```
 
+
+* Added new proofs to `Data.Nat.Properties`:
```
MatthewDaggitt, line 327: Please merge these into the existing `Data.Nat.Properties` block below!

MatthewDaggitt (changes requested):
Thanks! A few minor comments and simplifications, but otherwise looks great :+1:

CHANGELOG.md
gallais: The (now empty) block should be removed.

src/Data/Nat/Properties.agda
gallais: Why not use `≤-antisym` in combination with `≮⇒≥`?

PR #1376, opened 2020-12-09
Add Algebra.Morphism.LatticeMonomorphism

uzytkownik:


MatthewDaggitt:
Thanks for the PR. Apologies, I've gone ahead and tidied up some of the module usage as I couldn't get my head around how to simplify it simply from looking at the PR. In particular note the useful `where module L = IsLattice isLattice` trick which tends to simply record declarations of the form at the bottom.

I'll leave this up a couple of days in case anyone else has any comments and then merge it in.

PR #1378, opened 2020-12-09
[ fix #667 ] readMaybe for natural numbers

### LINKED ISSUE #667: parsing Nat
To parse to ``ℕ`` from String I write
``stringToNat (filterStr decimalDigit? str)``,

where ``stringToNat, filterStr, decimalDigit?``  are my home-made functions.
Is there a `standard' way to parse to ``ℕ`` from String ? 
If it is not, then standard library probably needs such.

gallais:


gallais:
I would have added `readMaybe` to `Data.Integer` too if only `uncons` was available
on `master` already. It will have to wait a little bit. :)

src/Data/Nat.agda
```diff
@@ -28,6 +38,35 @@ open import Data.Nat.Properties public
   ; _<‴?_; _≤‴?_; _≥‴?_; _>‴?_
   )
 
+------------------------------------------------------------------------
+-- Reading a number
+
+readMaybe : ∀ base {base≤16 : True (base ≤? 16)} → String → Maybe ℕ
```
MatthewDaggitt, line 44: I feel that this should live in `Data.Nat.Show` rather than here. I'm not sure we really want to add these dependencies,  for example `Data.List.Categorical`, to `Data.Nat`?

src/Data/Nat.agda
```diff
@@ -28,6 +38,35 @@ open import Data.Nat.Properties public
...
+readMaybe _ "" = nothing
+readMaybe base = Maybe.map convert
+              ∘′ TraversableA.mapA Maybe.applicative readDigit
+              ∘′ String.toList
+
+  where
+
+    convert : List ℕ → ℕ
+    convert = List.foldl (λ acc d → base * acc + d) 0
```
MatthewDaggitt, line 53: I was wondering if we could make use of some of the machinery from `Data.Digit` to help us, but it looks like that goes via `Fin` rather than `ℕ` so maybe not...

gallais: Yeah I had started working in `Data.Digit` but gave up because of `Fin`.
We may want a `read` there too but the one for `Nat` should definitely
be implemented efficiently IMO.

src/Data/Nat.agda
JacquesCarette: I like the idea of `Data.Nat.Show`, and other such `.Show` sub-modules.

gallais:
I threw in `Data.Integer.Show` that I happened to need today.

src/Data/Integer/Show.agda
```diff
@@ -0,0 +1,24 @@
...
+open import Data.Nat.Show using () renaming (show to showℕ)
+open import Data.String.Base using (String; _++_)
+
+------------------------------------------------------------------------
+-- Show
+
+-- Decimal notation: conversion back and forth with characters
+-- Time complexity is O(log₁₀(n))
+
+show : ℤ → String
```
MatthewDaggitt, line 22: Err, it turns out we already have it in `Data.Integer`

https://github.com/agda/agda-stdlib/blob/e4a715f757766294ab1d417e946ee1f7a9560903/src/Data/Integer.agda#L28

I think we should deprecate that one and keep this one. For the reasons outlined above, we don't really want all the string dependencies cluttering `Data.Integer`

PR #1379, opened 2020-12-13
Update variable policy in style guide

MatthewDaggitt:
This PR updates the standard library's policy for using the `variable` keyword in the style guide as per the discussion in #1325. I've tried it out on `Data.List.Relation.Unary.(Any/All)` and I think I'm happy that the result improves readability without harming understandability too much. Comments welcome obviously!

### DIFF
--- a/notes/style-guide.md
+++ b/notes/style-guide.md
@@ -327,7 +327,7 @@ line of code, indented by two spaces.
   predicates are not used in the definition of `List`, even though they are used
   in may list functions such as `filter`.
 
-* Example 2: the main type in `Data.List.Relation.Unary.All` is `Any P xs` where
+* Example 2: the main type in `Data.List.Relation.Unary.All` is `All P xs` where
   `A : Set a`, `P : Pred A p`, `xs : List A`. It therefore may declare variables
   over `Level`, `Set a`, `A`, `List A`, `Pred A p`. It may not declare, for example,
   variables of type `Rel` or `Vec`.
--- a/notes/style-guide.md
+++ b/notes/style-guide.md
@@ -311,10 +311,21 @@ line of code, indented by two spaces.
 * If there are lots of implicit arguments that are common to a collection
   of proofs they should be extracted by using an anonymous module.
 
-* Implicit of type `Level` and `Set` can be generalized using the keyword
-  `variable`. At the moment the policy is *not* to generalize over any other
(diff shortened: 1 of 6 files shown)


JacquesCarette:
I'm happy enough with the rule - but found the PR a bit odd. It explains carefully what is / is not allowed, and then the changed files seem to do the opposite, i.e. all of them abstract over some relations `P` and `Q`. I think this is actually justified, as I had not problem reading the code. What seems to be missing is a finer explanation of the rule that would allows this, while disallowing the case that it currently documents.

gallais:
> the changed files seem to do the opposite

The changed files are about `Data.List.Relation.Unary.(Any/All)` whose
main datatype is `(Any/All) P xs` where:

* `a` and `p` are `Level`s
* `A` is a `Set a`
* `P` is a `Pred A p`
* `xs` is a `List A`

By the policy of

>  A file may only declare variables of other types if those types are used
  in the definition of the main type that the file concerns itself with.

it is fine to generalise over all of the `All/Any` parameters above.

I guess this could be added to the doc to highlight the distinction between
the current example (`Data.List.Properties`) and the `Data.X.Relation` one.

JacquesCarette:
That's what I meant by "seem to do the opposite" and "finer explanation of the rule that would allow this" - but I now realize that that was not clear!  What I really wanted is exactly what you say in your last sentence: an addition to the doc that clarifies this point.

notes/style-guide.md
```diff
@@ -321,11 +321,16 @@ line of code, indented by two spaces.
...
+* Example 1: the main type in `Data.List.Properties` is `List A` where `A : Set a`.
   Therefore it may declare variables over `Level`, `Set a`, `A`, `List A`. It may
-  not declare variables over predicates (e.g. `P : Pred A l`) as predicates are
-  not used in the definition of `List`, even though they are used in may list
-  functions such as `filter`.
+  not declare variables, for example, over predicates (e.g. `P : Pred A p`) as
+  predicates are not used in the definition of `List`, even though they are used
+  in may list functions such as `filter`.
+
+* Example 2: the main type in `Data.List.Relation.Unary.All` is `Any P xs` where
```
gallais, line 330:
```suggestion
* Example 2: the main type in `Data.List.Relation.Unary.All` is `All P xs` where
```

PR #1382, opened 2020-12-22
Add various lemmas about list relations

MatthewDaggitt:
Clearing out my own development of various bits and pieces about lists.

### DIFF
--- a/src/Data/List/Relation/Binary/Subset/Propositional/Properties.agda
+++ b/src/Data/List/Relation/Binary/Subset/Propositional/Properties.agda
@@ -33,17 +33,17 @@ open import Function.Inverse as Inv using (_↔_; module Inverse)
 open import Function.Equivalence using (module Equivalence)
 open import Level using (Level)
 open import Relation.Nullary using (¬_; yes; no)
-open import Relation.Unary using (Decidable; Pred)
+open import Relation.Unary using (Decidable; Pred) renaming (_⊆_ to _⋐_)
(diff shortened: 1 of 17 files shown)


src/Data/List/Relation/Binary/Subset/Propositional/Properties.agda
```diff
@@ -231,6 +231,10 @@ module _ {P : Pred A p} (P? : Decidable P) where
   filter-⊆ : ∀ xs → filter P? xs ⊆ xs
   filter-⊆ = Setoidₚ.filter-⊆ (setoid A) P?
 
+  module _ {Q : Pred A q} (Q? : Decidable Q) where
+
+    filter⁺′ : P ⋐ Q → ∀ {xs ys} → xs ⊆ ys → filter P? xs ⊆ filter Q? ys
+    filter⁺′ = Setoidₚ.filter⁺′ (setoid A) P? (resp P) Q? (resp Q)
```
gallais, line 237: Is there any reason not to call this one `filter⁺` and shadow the setoid
version [like we did for `map⁺`](http://agda.github.io/agda-stdlib/Data.List.Relation.Binary.Sublist.Propositional.Properties.html#1230)?

MatthewDaggitt: The same reason why we don't call the setoid version `filter⁺`. There is a lemma in this file that incorrectly had that name before. Currently that name is deprecated. Can't use the name `filter⁺` without breaking backwards compatibility...

PR #1385, opened 2020-12-29
[ new ] Data.Tree.AVL.*.Relation.Unary.Any

gallais:
A first batch of relations and properties for `Data.Tree.AVL`
This introduces a breaking change. `K&_` used to be defined like so:

```agda
K&_ : ∀ {v} (V : Value v) → Set (a ⊔ v)
K& V = Σ Key (Value.family V)
```

and it is now its own standalone record:

```agda
record K&_ {v} (V : Value v) : Set (a ⊔ v) where
  constructor _,_
  field key   : Key
        value : Value.family V key
```

The advantage is that `V` can be reconstructed from a `Pred (K& V) p` type
which means that instead of having to pass it explicitly everywhere I can
have *way* cleaner types because a lot of things can be:
* reconstructed for function calls
* inferred to be equal for implicitly generalised `variable`s

I'm happy with the current status. It's not a huge PR and yet it provides a
slice of the functionalities I want that is both wide & deep enough to function
as a sanity check on the design. The only thing I would consider experimental
is `Data.Tree.AVL.Indexed.Relation.Unary.Any.Properties`.


src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+    P : Pred (K& V) p
+    Q : Pred (K& V) q
+    l u : Key⁺
+    n : ℕ
+    t : Tree V l u n
+
+------------------------------------------------------------------------
+-- Definition
+
+-- Given a predicate P, Any P t is a path to one element in t that satisfies P.
```
MatthewDaggitt, line 44: Very minor, but we should at least keep the comments under the character limit :smile:

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+
+------------------------------------------------------------------------
+-- Operations on Any
+
+map : P ⊆ Q → Any P t → Any Q t
+map f (here  p) = here (f p)
+map f (left  p) = left (map f p)
+map f (right p) = right (map f p)
+
+lookup : {t : Tree V l u n} → Any P t → Key
```
MatthewDaggitt, line 76: Why not define a `lookup` function returning a key and value here? This function could then be defined in terms of it and called `lookupKey`...

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+-- Properties of predicates preserved by Any
+
+any? : Decidable P → (t : Tree V l u n) → Dec (Any P t)
+any? P? (leaf _)          = no λ ()
+any? P? (node kv l r bal) = map′ fromSum toSum
+  (P? kv ⊎-dec any? P? l ⊎-dec any? P? r)
+
+satisfiable : ∀ {k l u} → l <⁺ [ k ] → [ k ] <⁺ u →
+              Satisfiable (P ∘ (k ,_)) →
+              Satisfiable {A = Tree V l u 1} (Any P)
```
MatthewDaggitt, line 116: Why are we not quantifying over the height here even though we are quantifying over the height in `any?`?

src/Data/Tree/AVL/Map/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,69 @@
...
+    v p q : Level
+    V : Set v
+    P : Pred V p
+    Q : Pred V q
+    m : Map V
+
+------------------------------------------------------------------------
+-- Definition
+
+-- Given a predicate P, Any P t is a path to one element in t that satisfies P.
```
MatthewDaggitt, line 41: Too long a comment

src/Data/Tree/AVL/Indexed/Relation/Unary/All.agda
```diff
@@ -0,0 +1,93 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- AVL trees where at least one element satisfies a given property
```
MatthewDaggitt, line 4: Incorrectly copied from `Any`?

src/Data/Tree/AVL/Indexed/Relation/Unary/All.agda
```diff
@@ -0,0 +1,93 @@
...
+    P : Pred (K& V) p
+    Q : Pred (K& V) q
+    l u : Key⁺
+    n : ℕ
+    t : Tree V l u n
+
+------------------------------------------------------------------------
+-- Definition
+
+-- Given a predicate P, Any P t is a path to one element in t that satisfies P.
```
MatthewDaggitt, line 48: Too long comment

src/Data/Tree/AVL/Indexed/Relation/Unary/All.agda
```diff
@@ -0,0 +1,93 @@
...
+decide : Π[ P ∪ Q ] → (t : Tree V l u n) → All P t ⊎ Any Q t
+decide p⊎q (leaf l<u)        = inj₁ leaf
+decide p⊎q (node kv l r bal) with p⊎q kv | decide p⊎q l | decide p⊎q r
+... | inj₂ q | _ | _ = inj₂ (here q)
+... | _ | inj₂ q | _ = inj₂ (left q)
+... | _ | _ | inj₂ q = inj₂ (right q)
+... | inj₁ pv | inj₁ pl | inj₁ pr = inj₁ (node pv pl pr)
+
+unnode : ∀ {hˡ hʳ h} {kv : K& V}
+  {lk : Tree V l [ kv .key ] hˡ} {ku : Tree V [ kv .key ] u hʳ} {bal : hˡ ∼ hʳ ⊔ h} →
```
MatthewDaggitt, line 77: According to the style guide, the following lines should be indented to the same point as the `∀` on the line before.

src/Data/Tree/AVL.agda
```diff
@@ -36,12 +36,12 @@ private
 
 open StrictTotalOrder strictTotalOrder renaming (Carrier to Key)
 import Data.Tree.AVL.Indexed strictTotalOrder as Indexed
-open Indexed using (K&_ ; ⊥⁺ ; ⊤⁺; ⊥⁺<⊤⁺; ⊥⁺<[_]<⊤⁺; ⊥⁺<[_]; [_]<⊤⁺)
+open Indexed using (K&_; toPair; ⊥⁺; ⊤⁺; ⊥⁺<⊤⁺; ⊥⁺<[_]<⊤⁺; ⊥⁺<[_]; [_]<⊤⁺)
```
MatthewDaggitt, line 39: Don't need the `toPair` import here.

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+open import Function.Base
+open import Level using (Level; _⊔_)
+
+open import Relation.Nullary using (Dec; no)
+open import Relation.Nullary.Decidable using (map′)
+open import Relation.Nullary.Sum using (_⊎-dec_)
+open import Relation.Unary
+
+open StrictTotalOrder strictTotalOrder renaming (Carrier to Key)
+open import Data.Tree.AVL.Indexed strictTotalOrder as AVL
```
MatthewDaggitt, line 27: Don't need the `as AVL`?

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+    t : Tree V l u n
+
+------------------------------------------------------------------------
+-- Definition
+
+-- Given a predicate P, Any P t is a path to one element in t that satisfies P.
+-- There may be others.
+-- See `Relation.Unary` for an explanation of predicates.
+
+data Any {V : Value v} (P : K& V → Set p) {l u}
```
MatthewDaggitt, line 48: `Pred (K& V) p`?

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+    ∀ {l u h} (k : Key) (v : Val k) → (Pv : P (k , v))
+    (t : Tree V l u h) (seg : l < k < u) →
+    Any P (proj₂ (insert k v t seg))
+  Any-insert P≈ k v Pv = Any-insertWith k (F.const v) $ λ where
+    eq nothing   → Pv
+    eq (just v′) → P≈ v eq Pv
+
+  insert⁺ : ∀ {l u h} (k : Key) (f : Maybe (Val k) → Val k) (t : Tree V l u h) (seg : l < k < u) →
+     (p : Any P t) → k ≉ Any.lookup p → Any P (proj₂ (insertWith k f t seg))
+  insert⁺ k f (node (k′ , v′) l r bal) (l<k , k<u) p k≉ with compare k k′
```
MatthewDaggitt, line 165: The trick I usually use here is to write `with compare k k′ | p` which allows you to use ellipses to greatly reduce the size of the proof. Unsure what the computational implications are, but it does massively increase readability.

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+    = joinˡ⁺-left⁺ kv (insertWith k f l (l<k , [ k<k′ ]ᴿ)) r bal (insert⁺ k f l _ p k≉)
+  insert⁺ k f (node kv@(k′ , v′) l r bal) (l<k , k<u) (right p) k≉ | tri< k<k′ _ _
+    = joinˡ⁺-right⁺ kv (insertWith k f l (l<k , [ k<k′ ]ᴿ)) r bal p
+  insert⁺ k f (node kv@(k′ , v′) l r bal) (l<k , k<u) (here p)  k≉ | tri> _ _ k′<k
+    = joinʳ⁺-here⁺ kv l (insertWith k f r ([ k′<k ]ᴿ , k<u)) bal p
+  insert⁺ k f (node kv@(k′ , v′) l r bal) (l<k , k<u) (left p)  k≉ | tri> _ _ k′<k
+    = joinʳ⁺-left⁺ kv l (insertWith k f r ([ k′<k ]ᴿ , k<u)) bal p
+  insert⁺ k f (node kv@(k′ , v′) l r bal) (l<k , k<u) (right p) k≉ | tri> _ _ k′<k
+    = joinʳ⁺-right⁺ kv l (insertWith k f r ([ k′<k ]ᴿ , k<u)) bal (insert⁺ k f r _ p k≉)
+  insert⁺ k f (node (k′ , v′) l r bal) (l<k , k<u) (here p)  k≉ | tri≈ _ k≈k′ _ = ⊥-elim (k≉ k≈k′)
```
MatthewDaggitt, line 178: `contradiction` rather than `⊥-elim`?

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
```diff
@@ -0,0 +1,118 @@
...
+satisfied (right p) = satisfied p
+
+module _ {hˡ hʳ h} {kv : K& V}
+  {lk : Tree V l [ kv .key ] hˡ}
+  {ku : Tree V [ kv .key ] u hʳ}
+  {bal : hˡ ∼ hʳ ⊔ h}
+  where
+
+  toSum : Any P (node kv lk ku bal) →
+          P kv ⊎ Any P lk ⊎ Any P ku
```
MatthewDaggitt, line 95: This type and the one below it can go on one line?

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+  Any-insert :
+    (P≈ : ∀ {k k′} v (eq : k ≈ k′) → P (k , v) → P (k′ , Value.respects V eq v)) →
+    ∀ {l u h} (k : Key) (v : Val k) → (Pv : P (k , v))
+    (t : Tree V l u h) (seg : l < k < u) →
+    Any P (proj₂ (insert k v t seg))
+  Any-insert P≈ k v Pv = Any-insertWith k (F.const v) $ λ where
+    eq nothing   → Pv
+    eq (just v′) → P≈ v eq Pv
+
+  insert⁺ : ∀ {l u h} (k : Key) (f : Maybe (Val k) → Val k) (t : Tree V l u h) (seg : l < k < u) →
```
MatthewDaggitt, line 163: This should be called `insertWith⁺`?

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+
+------------------------------------------------------------------------
+-- insert
+
+module _ {V : Value v} where
+
+  private Val = Value.family V
+
+  Any-insertWith :
+    ∀ {l u h} (k : Key) →
```
MatthewDaggitt, line 141: Indentation, here and below is non-standard

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+    (f : Maybe (Val k) → Val k) →
+    (Pf : ∀ {k′} → (eq : k ≈ k′) (mk′ : Maybe (Val k′)) →
+          maybe′ (λ v → P (k′ , Value.respects V eq (f (just (Value.respects V (sym eq) v)))))
+                 (P (k , f nothing)) mk′)
```
MatthewDaggitt, line 145: Ugh, this type is pretty unpleasant. Feels like it might be a symptom of some suboptimal or missing definition? At a minimum you can pull out `Value.respects V` into the private block above...

MatthewDaggitt: Hmm maybe I've found the problem. The proof doesn't actually require the `eq : k ≈ k′` argument to go through (`refl` is always provided`) right? So it's really combining two proofs in one. Firstly that `insertWith` can introduce `Any` and secondly that `Any` respects the underlying equality. The second proof should be split out and proved separately right?

MatthewDaggitt (commented):
> It's not a huge PR

Yup only 600 new lines of code, definitely not huge :laughing:

src/Data/Tree/AVL/Sets.agda
```diff
@@ -68,7 +68,7 @@ toList : ⟨Set⟩ → List A
 toList = List.map proj₁ ∘ AVL.toList
 
 foldr : (A → B → B) → B → ⟨Set⟩ → B
```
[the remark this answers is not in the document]
gallais: I guess it does matter if you're visiting the keys in increasing, decreasing,
or no particular order.

src/Data/Tree/AVL/Indexed/Relation/Unary/Any.agda
gallais: It's not clear how to populate the subtrees. I guess we could
take subtrees as arguments too but then this function is basically
just `node` so I'm not convinced it's of any use.

The `List` equivalent generates a singleton list rather than taking
an arbitrary tail for instance.

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
gallais: The problem is that you don't want to end up with a proof of
`P (k , Value.respects V (sym refl) (f nothing))` when you
need a proof of `P (k , f nothing)`.

I did not want to add an extra coherence constraint stating that
`Value.respects V (sym refl)` preserves proofs.

I should perhaps just split this `Any-insertWith` into two distinct
lemmas depending on whether the key is already present in the tree
or not.

gallais:
> Yup only 600 new lines of code, definitely not huge

By only having one of the `.Any.Properties` files (and a small one at that),
and similarly only one of the `.All` files. :D

src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
MatthewDaggitt: > I should perhaps just split this Any-insertWith into two distinct
lemmas depending on whether the key is already present in the tree
or not.

That sounds like a good plan!

gallais:
Well, that took longer than expected. And led to many more changes than I had anticipated.
Which means I'll admit that this PR is now fairly large. :D

MatthewDaggitt:
@gallais is it okay with you if I push this to v1.6? 

I'd quite like to have some time to go through the new additions, and in particular I'm a bit worried about the addition of `begin-irrefl_` to the reasoning modules, as 1) it doesn't function the same way as the rest of the reasoning combinators and 2) it's not clear that there's actually a contradiction going on underneath and 3) its not backwards compatible.

gallais:
Sure!

MatthewDaggitt:
@gallais I would really like to merge this in as it's a great piece of work, but I am still a bit dubious about the inclusion of `begin-irrefl`. As it's not an integral part of this PR, would you mind if I stripped it out to unblock this PR? If you're still keen to include it, we can then make a separate PR with it in, in which we can better discuss the pros and cons?

gallais:
I think I have a plan to get us out of this pickle.

MatthewDaggitt:
> I think I have a plan to get us out of this pickle.

Ooh, exciting :grin:

MatthewDaggitt:
Thanks @gallais, I agree it is an improvement as it no longer breaks backwards compatibility with `Reasoning.Base.Triple`, and I also agree that the combinator is very convenient and it does remove a fair bit of code. 

However I'm still reluctant to include this combinator. The reasoning APIs' primary purpose is to improve the readability of proofs, but in my opinion `begin-irrefl_` actively harms readability:

- Firstly, there is no indication that there's a contradiction being used underneath the hood. Knowing whether or not a proof uses contradiction is fairly crucial to understanding it.
- Secondly, unlike when using the other combinators, there is no way to infer the final type of the reasoning proof simply by looking at it. From the reader's perspective the final type is magicked out thin air by the (invisible) contradiction.

I would argue that it's better to keep the contradiction explicit in the code?

Other opinions welcome!

gallais:
I think `begin-irrefl` is clearer than:

```agda
flip contradiction (<-irrefl refl) $ begin (...)
```

I guess we could make the return type `⊥` to force the use of `⊥-elim`
as documentation that there's an explosion somewhere in there but I'm
not particularly convinced.

JacquesCarette:
Rather than `begin-irrefl_`, what about `by-contradiction_` ? This would take care of @MatthewDaggitt 's first point. It doesn't fully take care of the second, though it becomes less surprising since we all know we can conjure anything we want via `⊥-elim`.

Tongue-in-cheek: also have `conjure-by-contradiction` with 2 arguments, the first the conjured type, the second the falsity proof, that might work.

MatthewDaggitt:
Hmm, yup I'd be on-board with something that had contradiction in it. Would `begin-irrefl-contradiction_` or `begin-contradiction-by-irrefl_` be too long?

MatthewDaggitt:
Okay, as this is the only outstanding PR/issue blocking the v1.6 release candidate, I've temporarily removed the new combinator from this PR. Once Travis goes green, I'll merge this in and then open a new PR with the combinator restored where we can continue this discussion.

PR #1387, opened 2020-12-30
Re-organisation of algebra operation definitions

MatthewDaggitt:
So I was trying to fit ring exponentiation into the new library design for auxiliary definitions over different types of algebraic bundles (multiplication/summation/divisibility etc.), finishing off the reorganisation started in #1351 and #1377. However, it turns out that the design isn't general enough as solvers need access to the operation even when they don't have access to the accompanying ring axioms. Hence this reorganisation:

- A new set of modules `Algebra.Definitions.(RawMagma/RawMonoid/RawSemiring)` each now contain all the auxiliary definitions that can be defined for the corresponding raw bundle. This gets rid of the `Algebra.Primality` and `Algebra.Divisibility` modules.

- The modules `Algebra.Properties.(Magma/Semigroup/Monoid/Semiring).(Multiplication/Summation/Divisibility)` now each re-export the relevant auxiliary definition and contain the properties about it for each bundle respectively.

This should provide enough flexibility to add more theory about algebraic structures in future.

This PR doesn't yet deprecate `Algebra.Operations.Ring`.


MatthewDaggitt:
Pinging @mechvel as it touches code they've been working on recently.

Issue #392, opened 2018-07-24
Cancellation for List.++


andreasabel:
I could not find laws stating that List.++ make a left- and right-cancellative monoid, i.e., we have
```
xs ++ ys == xs ++ zs -> ys == zs
xs ++ zs == ys ++ zs -> xs == ys
```
The second could be proven from the first by going through `reverse`, maybe:
```
xs ++ zs == ys ++ zs
reverse (xs ++ zs) == reverse (ys ++ zs)
reverse zs ++ reverse xs == reverse zs ++ reverse ys
reverse xs == reverse ys
xs == ys
```
which would need an injectivity lemma for `reverse`.

`Data.Nat.Properties` defines cancellation for addition and multiplication. 

On a related note, `List.++` is also a positive monoid, i.e. we have
```
xs ++ ys == [] -> xs == [] & ys == []
```
Such properties are also missing.  
They are defined for `Nat.+`, but not given a name, like `LeftPositive`.
I do not know what would be the most common terminology here.

MatthewDaggitt:
> On a related note, List.++ is also a positive monoid, i.e. we have
> `xs ++ ys == [] -> xs == [] & ys == []`
> Such properties are also missing.
> They are defined for Nat.+, but not given a name, like LeftPositive.
> I do not know what would be the most common terminology here.

I agree we should settle on a standardised name for these sort of properties. Where is `LeftPositive` from? And why is it Left? The `Positive` might also be a bit confusing for numeric datatypes.

gallais:
As usual there seems to be a few names for these: https://en.wikipedia.org/wiki/Zerosumfree_monoid

andreasabel:
The term `Positive` is intuitive for additive monoids and thus would make sense for numeric instances.  However, for multiplicative ones we would rather speak of being free of unit dividers.

Abstractly maybe the best way is to think of the monoid operation as "composition" where the unit is the `Identity`.  Thus, positivity expresses that the identity cannot be decomposed, or is atomic.  Thus, maybe one of
- `HasAtomicIdentity` 
- `AtomicIdentity`
- `IdentityAtomic`

optionally with superscripts `l` and `r`

JacquesCarette:
`IrreducibleIdentity` ?  Although `Atomic` does make sense, it clashes with both the notion from measure theory and from logic. Something is `Atomic` when it has *no* decompositions; here the `Identity` has a (trivial) decomposition.

andreasabel:
[...]

PR #1391, opened 2020-12-31
[ fix #835 ] Text.Regex

### LINKED ISSUE #835: Add Text.Regex
Inspiration sources:

* https://github.com/sweirich/dth/tree/master/regexp for regular expressions *with (scoped) capture*  in dependent Haskell
* https://github.com/gallais/agdarep for basic regular expressions in Agda (the optimisation described Agular and Mannaa's report does make a big difference)

It would be really nice to have regular expressions with capture (à la sed)
like Weirich does.

Bonus points: I believe we should be able to parse the input regular expression
using the shunting yard algorithm. So no need for a parser combinators library
quite yet.

gallais:
A late Xmas present.


src/Data/List/Relation/Ternary/Appending/Propositional.agda
```diff
@@ -0,0 +1,41 @@
...
+import Data.List.Relation.Ternary.Appending.Setoid as General
+import Data.List.Relation.Ternary.Appending.Properties as Appendingₚ
+open import Data.Product using (_,_)
+open import Relation.Binary.PropositionalEquality
+  using (_≡_; setoid; refl; trans; cong; cong₂; module ≡-Reasoning)
+
+------------------------------------------------------------------------
+-- Re-export the basic combinators
+
+open General hiding (Appending; _++_; break) public
```
MatthewDaggitt, line 23: Style guide says that `public` should go before `using`/`hiding`/`renaming`. Same for other files below.

src/Data/List/Relation/Ternary/Appending/Properties.agda
```diff
@@ -0,0 +1,42 @@
...
+    R : REL A B r
+    as : List A
+    bs : List B
+    cs ds : List C
+
+module _  {e} {E : Rel C e} {L′ : REL A C l} {R′ : REL B C r}
+          (LEL′ : Trans L E L′) (RER′ : Trans R E R′)
+          where
+
+  transitive : Appending L R as bs cs → Pointwise E cs ds → Appending L′ R′ as bs ds
```
MatthewDaggitt, line 34: Hmm I think this name (and `trans` in other files) could be improved? I would expect `transitive` to have a type something like

`Appending L R as bs cs → Appending L R cs ds es→ Appending L R (as ++ bs) ds es`?

This is more like a "respects Pointwise" property. Indeed there are two other properties we could add where the pointwise equality is over the first and second list arguments right?

src/Text/Regex/Base.agda
```diff
@@ -0,0 +1,102 @@
...
+  _∣_  : (e f : Exp) → Exp
+  _∙_  : (e f : Exp) → Exp
+  _⋆   : (e : Exp) → Exp
+
+-- A regular expression has additional parameters:
+-- * should the match begin at the very start of the input?
+-- * should it span until the very end?
+
+record Regex : Set a where
+  field fromStart  : Bool
```
MatthewDaggitt, line 58: Field names should be indented two spaces underneath the `field` keyword.

src/Text/Regex/Search.agda
```diff
@@ -0,0 +1,218 @@
...
+... | true  | false = here
+... | false | true  = Infixₚ.fromSuffix
+... | false | false = id
+
+-- A match is a list, a proof it matches the regular expression,
+-- and a proof it is the right sort of sublist.
+record Match {s} (R : Rel (List A) s) (xs : List A) (exp : Exp)
+       : Set (a ⊔ e ⊔ r ⊔ s) where
+  constructor MkMatch
+  field list    : List A
```
MatthewDaggitt, line 69: Indent constructors.

src/Text/Regex/Search.agda
```diff
@@ -0,0 +1,218 @@
...
+
+open import Relation.Binary using (DecPoset)
+
+module Text.Regex.Search {a e r} (P? : DecPoset a e r) where
+
+open import Level using (_⊔_)
+open import Data.Bool.Base using (if_then_else_; true; false)
+open import Data.List.Base using (List; []; _∷_)
+open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂; [_,_]′)
+open import Function
```
MatthewDaggitt, line 17: `Function.Base`?

src/Text/Regex/Search.agda
```diff
@@ -0,0 +1,218 @@
...
+  []ᴹ : ∀ {xs e} → [] ∈ e → Match (Prefix _≡_) xs e
+  []ᴹ p = MkMatch [] p []
+
+  []⁻¹ᴹ : ∀ {e} → Match (Prefix _≡_) [] e → [] ∈ e
+  []⁻¹ᴹ (MkMatch .[] p []) = p
+
+  _∷ᴹ_ : ∀ {xs e} x → Match (Prefix _≡_) xs (eat x e) → Match (Prefix _≡_) (x ∷ xs) e
+  x ∷ᴹ (MkMatch ys ys∈e\x ys≤xs) = MkMatch (x ∷ ys) (eat-sound x _ ys∈e\x) (refl ∷ ys≤xs)
+
+  _∷⁻¹ᴹ_ : ∀ {xs x e} → ¬ ([] ∈ e) →
```
MatthewDaggitt, line 93: Very minor but why can't we use `∉` everywhere in this file?

MatthewDaggitt (commented):
Looks pretty awesome :smile: I think it would probably benefit from:
- a README file?
- a `Text.Regex` module re-exporting both `Text.Regex.Base` and `Text.Regex.Search`?
- a module that specialises it to `String`s?

src/Data/List/Relation/Ternary/Appending.agda
```diff
@@ -0,0 +1,57 @@
...
+  []       ++ rs = []++ rs
+  (l ∷ ls) ++ rs = l ∷ (ls ++ rs)
+
+  -- extract the "proper" equality split from the pointwise relation
+  break : Appending L R as bs cs → ∃₂ λ cs₁ cs₂ →
+          cs₁ List.++ cs₂ ≡ cs × Pointwise L as cs₁ × Pointwise R bs cs₂
+  break ([]++ rs) = -, -, (refl , [] , rs)
+  break (l ∷ lrs) =
+    let (_ , _ , eq , ls , rs) = break lrs in
+    -, -, (cong (_ ∷_) eq , l ∷ ls , rs)
```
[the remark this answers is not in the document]
gallais: I'd rather not use a `with` as this means the function would get
needlessly stuck on the recursive call in proofs.

ajrouvoet:
Random question out of interest: where does the terminology 'conicalˡ' come from?

In the ternary library I call `[]++⁻¹` the left identity of the monoidal ternary relation `Appending` and `conicalˡ` the inverse identity, for lack of a better name.

gallais:
This was discussed in #392

ajrouvoet:
Thank you! Awesome PR.

PR #1392, opened 2021-01-01
Continuing with generic GCD.

mechvel:
Please, consider this my commit of January 2, 2021.
It continues with proving properties of generic GCD, and introduces GCDSemiring 
(over CancellativeCommutativeSemiring). 

new file:    src/Algebra/Properties/CancellativeCommutativeSemiring.agda
new file:    src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
changed:     CHANGELOG.md
changed:     src/Algebra/Bundles.agda
changed:     src/Algebra/Structures.agda
changed:     src/Algebra/Divisibility.agda
changed:     src/Algebra/Properties/Monoid/Divisibility.agda
changed:     src/Algebra/Properties/Semiring/GCD.agda

* Algebra.Divisibility.IsGCD :             added two values.
* Algebra.Properties.Monoid.Divisibility : added ∣∣-reflexive.
* Algebra.Properties.Semiring.GCD :        added a couple of GCD properties.
* Algebra.Properties.CancellativeCommutativeSemiring :
                                 certain important GCD properties proved.
* Algebra.Bundles.GCDSemiring  introduced.


src/Algebra/Properties/Monoid/Divisibility.agda
```diff
@@ -34,6 +34,9 @@ open import Algebra.Properties.Semigroup.Divisibility semigroup public
 ∣-reflexive : _≈_ ⇒ _∣_
 ∣-reflexive = Div.∣-reflexive trans identityˡ
 
+∣∣-reflexive : _≈_ ⇒ _∣∣_
+∣∣-reflexive x≈y = ∣-reflexive x≈y , ∣-reflexive (sym x≈y)
```
MatthewDaggitt, line 38: Please put this under its own section in the file, rather than in the middle of the divisibility properties.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,45 @@
...
+open import Relation.Nullary using (yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+
+module Algebra.Properties.CancellativeCommutativeSemiring
+  {a ℓ} (R : CancellativeCommutativeSemiring a ℓ)
+  (open CancellativeCommutativeSemiring R) (_≟_ : Decidable _≈_) where
+
+open EqReasoning setoid
+
+*-cancelʳ-nonZero : AlmostRightCancellative _≈_ 0# _*_
```
MatthewDaggitt, line 24: A generalisation of this lemma (and its opposite lemma) should be in `Algebra.Consequences.Setoid`. The generalisation should then be used here.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+  d'∣x = ∣-trans d'∣x' x'∣x
+  d'∣y = ∣-trans d'∣y' y'∣y
+  d'∣d = IsGCD.greatest isGCD d'∣x d'∣y
+
+------------------------------------------------------------------------------
+-- gcd-distr  is an important lemma of the gcd distributivity:
+-- gcd (c*a) (c*b)  assoc-equal  c * (gcd a b).
+
+gcd-distr : ∀ {a b c d d'} → IsGCD a b d → IsGCD (c * a) (c * b) d' → d' ∣∣ (c * d)
+gcd-distr {a} {b} {c} {d} {d'} isGCD-a-b-d isGCD-ca-cb-d' =  aux (c ≟ 0#)
```
MatthewDaggitt, line 104: This proof is pretty messy and unreadable. There must be a way of breaking it up into meaningful constituent parts, in the same way we have done so in `Data.Nat.GCD`!

src/Algebra/Properties/Semiring/GCD.agda
```diff
@@ -7,24 +7,45 @@
...
+x≉0⊎y≉0⇒gcd≉0 (isGCDᶜ _ d∣y _) (inj₂ y≉0) = x∣y∧y≉0⇒x≉0 d∣y y≉0
+
+x≈0∧y≈0⇒gcd≈0 : ∀ {x y d} → IsGCD x y d → x ≈ 0# → y ≈ 0# → d ≈ 0#
+x≈0∧y≈0⇒gcd≈0 (isGCDᶜ _ _ greatest) x≈0 y≈0 =  0∣x⇒x≈0 0∣d
+  where
+  0∣x = ∣-respʳ (sym x≈0) (0# ∣0)
+  0∣y = ∣-respʳ (sym y≈0) (0# ∣0)
+  0∣d = greatest {0#} 0∣x 0∣y
+
+coprime⇒gcd∣1 : ∀ {x y d} → IsGCD x y d → Coprime x y →  d ∣ 1#
```
MatthewDaggitt, line 47: Given the name, the argument order should really be swapped.

src/Algebra/Properties/Semiring/GCD.agda
```diff
@@ -7,24 +7,45 @@
...
+                                   -- (gcd x y) is invertible for (coprime x y)
+coprime⇒gcd∣1 isGCD coprime-x-y =  coprime-x-y divides₁ divides₂
+  where
+  open IsGCD isGCD
```
MatthewDaggitt, line 51: Please use the constructor rather than `where open`

MatthewDaggitt (commented):
First review.

mechvel:
What is the final solution for the record constructor names :   `isGCDᶜ`  or  `mkIsGCD` ?

mechvel:
Please, consider my commit of January 5, 2021.
It improves the code according to your notes.
See my (coming) replies to the above notes.
The two points about  the lemma of  ``.../CancellativeCommutativeSemiring/GCD.gcd-distr`` are questionable
-- see  my  coming replies.
I am going to improve the code assuming your answers to these two questions.

src/Algebra/Divisibility.agda
```diff
@@ -1,4 +1,4 @@
-------------------------------------------------------------------------
+-----------------------------------------------------------------------
```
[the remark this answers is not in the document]
mechvel: Fixed.

> Comments should not exceed the 80 character limit. Please wrap around to
> the next line, and go through and check all the other comments in this PR.

In the file headers I preserve now 72 characters length - as in  stdlib. 
In other places I sometimes set comments of different width, but not more than 80 characters.
OK?

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,45 @@
...
+open import Data.Sum.Base using (_⊎_; inj₁; inj₂)
+open import Function using (_$_)
+open import Relation.Binary using (Decidable)
+import Relation.Binary.Reasoning.Setoid as EqReasoning
+open import Relation.Nullary using (yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+
+module Algebra.Properties.CancellativeCommutativeSemiring
+  {a ℓ} (R : CancellativeCommutativeSemiring a ℓ)
+  (open CancellativeCommutativeSemiring R) (_≟_ : Decidable _≈_) where
```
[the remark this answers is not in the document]
mechvel: Done as you suggest.

src/Algebra/Properties/Monoid/Divisibility.agda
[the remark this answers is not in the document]
mechvel: Done.
Also I add certain properties of  ``_∣∣_``  in Algebra.Divisibility,
Algebra.Properties.Magma.Divisibility, Algebra.Properties.Semigroup.Divisibility,
Algebra.Properties.Monoid.Divisibility.
The latter provides ``||-isEquivalence``.
Their proofs are as simple as the proofs for the properties of ``_∣_`` set in the same files.

src/Algebra/Properties/Semiring/GCD.agda
```diff
@@ -7,24 +7,45 @@
...
-import Algebra.Properties.Monoid.Divisibility as MonoidDiv
-open import Data.Product using (_,_)
+import Algebra.Primality as Primality
 open import Data.Sum.Base using (_⊎_; inj₁; inj₂)
 
 module Algebra.Properties.Semiring.GCD {a ℓ} (R : Semiring a ℓ) where
 
 open Semiring R
+open Primality _≈_ _*_ 0# 1# using (Coprime)
+-- open import Algebra.Properties.Monoid.Divisibility *-monoid using (∣-refl)
```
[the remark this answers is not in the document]
mechvel: Done

src/Algebra/Properties/Semiring/GCD.agda
```diff
@@ -7,24 +7,45 @@
...
+x≈0∧y≈0⇒gcd≈0 : ∀ {x y d} → IsGCD x y d → x ≈ 0# → y ≈ 0# → d ≈ 0#
+x≈0∧y≈0⇒gcd≈0 (isGCDᶜ _ _ greatest) x≈0 y≈0 =  0∣x⇒x≈0 0∣d
+  where
+  0∣x = ∣-respʳ (sym x≈0) (0# ∣0)
+  0∣y = ∣-respʳ (sym y≈0) (0# ∣0)
+  0∣d = greatest {0#} 0∣x 0∣y
+
+coprime⇒gcd∣1 : ∀ {x y d} → IsGCD x y d → Coprime x y →  d ∣ 1#
+                                   -- (gcd x y) is invertible for (coprime x y)
+coprime⇒gcd∣1 isGCD coprime-x-y =  coprime-x-y divides₁ divides₂
```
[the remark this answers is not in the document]
mechvel: Done.
By the way, what is the final decision:  `isGCDᶜ`  or  `mkIsGCD` ?

src/Algebra/Properties/Semiring/GCD.agda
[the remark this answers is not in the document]
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+open import Data.Product using (_,_; proj₁; proj₂)
+open import Data.Sum.Base using (_⊎_)
+open import Relation.Binary using (Decidable)
+import Relation.Binary.Reasoning.Setoid as EqReasoning
+open import Relation.Nullary using (Dec; yes; no)
+
+module Algebra.Properties.CancellativeCommutativeSemiring.GCD
+  {a ℓ} (R : CancellativeCommutativeSemiring a ℓ)
+  (open CancellativeCommutativeSemiring R)
+  (_≟_ : Decidable _≈_)
```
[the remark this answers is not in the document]
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+  (∣-refl; ∣-trans; ∣-respˡ; ∣-respʳ; ∣∣-reflexive)
+import Algebra.Properties.CommutativeSemigroup *-commutativeSemigroup as Of*CSemig
+open import Algebra.Properties.CommutativeSemigroup.Divisibility
+  *-commutativeSemigroup using (x∣xy)
+
+------------------------------------------------------------------------
+-- Re-exporting definition of GCD, primality, properties of GCD in semiring
+
+open import Algebra.Divisibility _≈_ _*_ public
+  using (_∣_; _∣∣_)
```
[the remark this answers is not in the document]
mechvel: Fixed.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+             d * 1#        ∎
+
+           qc≈1 = *-cancelˡ-nonZero {d} (q * c) 1# d≉0 d*qc≈d*1
+        in
+        q , qc≈1
+
+------------------------------------------------------------------------------
+-- Recall that  x ∣∣ y  means that x and y are equal as association classes by
+-- division. For example, for ℕ  this means  x ≡ y,
+-- for ℤ  this means (x ≡ y  or  x ≡ -y).
```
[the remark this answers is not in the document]
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+
+GCD-unique : ∀ {x x' y y' d d'} → x ∣∣ x' → y ∣∣ y' →
+             IsGCD x y d → IsGCD x' y' d' → d ∣∣ d'
+                                            -- gcd-s for assoc-equal pairs are assoc-equal.
+GCD-unique {x} {x'} {y} {y'} {d} {d'} (x∣x' , x'∣x) (y∣y' , y'∣y) isGCD isGCD' =
+  d∣d' , d'∣d
+  where
+  open IsGCD isGCD  using () renaming (divides₁ to d∣x; divides₂ to d∣y)
+  open IsGCD isGCD' using () renaming (divides₁ to d'∣x'; divides₂ to d'∣y')
```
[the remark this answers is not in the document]
mechvel: Done

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+
+------------------------------------------------------------------------------
+-- Recall that  x ∣∣ y  means that x and y are equal as association classes by
+-- division. For example, for ℕ  this means  x ≡ y,
+-- for ℤ  this means (x ≡ y  or  x ≡ -y).
+
+GCD-unique : ∀ {x x' y y' d d'} → x ∣∣ x' → y ∣∣ y' →
+             IsGCD x y d → IsGCD x' y' d' → d ∣∣ d'
+                                            -- gcd-s for assoc-equal pairs are assoc-equal.
+GCD-unique {x} {x'} {y} {y'} {d} {d'} (x∣x' , x'∣x) (y∣y' , y'∣y) isGCD isGCD' =
```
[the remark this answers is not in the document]
mechvel: Removed implicits.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
...
+open import Algebra.Divisibility _≈_ _*_ public
+  using (_∣_; _∣∣_)
+open import Algebra.Primality _≈_ _*_ 0# 1# public
+  using (Coprime)
+open import Algebra.Properties.Semiring.GCD semiring public
+
+------------------------------------------------------------------------
+-- Properties of GCD
+
+x≈0⊎y≉0⇒Coprime[x/gcd,y/gcd] :
```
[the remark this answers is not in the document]
mechvel: Fixed.

src/Algebra/Properties/Semiring/GCD.agda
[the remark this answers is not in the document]
mechvel: Done.

src/Algebra/Properties/Semiring/GCD.agda
```diff
@@ -7,24 +7,45 @@
...
 -- Re-exporting definition of GCD
 
 open import Algebra.Divisibility _≈_ _*_ public
-  using (IsGCD; gcdᶜ)
+  using (IsGCD; isGCDᶜ)
 
 ------------------------------------------------------------------------
 -- Properties of GCD
 
+isGCD-0-x-x : ∀ x → IsGCD 0# x x
```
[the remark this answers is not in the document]
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,183 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of the Greatest Common Divisor in CancellativeCommutativeSemiring.
```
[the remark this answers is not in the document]
mechvel: Done. See my earlier reply.

src/Algebra/Properties/Monoid/Divisibility.agda
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
mechvel: Done.

src/Algebra/Properties/Semiring/GCD.agda
mechvel: Done.

src/Algebra/Properties/Semiring/GCD.agda
mechvel: Done.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
mechvel: This is not so simple!
I do not know of how to simplify it any essentially.
Data.Nat.GCD  would not help. Because it uses the properties of division with remainder (``_/_`` relies on division with remainder), on the properties of ``_/_`` on `ℕ`.
And generic  CancellativeCommutativeSemiring  does not possess such an operation with such properties. 
Among ``R : CancellativeCommutativeSemiring`` there are some domains R for which there is not possible a division operation with the properties of Nat.DivMod... In particular there is not possible in general to solve  ``x ∣? y``  in R,  
while for `ℕ` it is solved via divMod.

gcd-distr  is a subtle and a particular lemma. It nicety is in that its proof does not use anything besides the three
axioms of GCD (and the properties of generic ``_∣_``).
The first impression is that  gcd-distr  cannot be derived only from these laws. But it occurs that the goal is achieved by combining these three laws repeatedly, and in a certain way.

I do not find this lemma in textbooks. This is a folklore.  A certain referee has pointed at its formulation. 
Then, I searched for the question at a mathematical forum, and found there a proof sketch.
The true author of the lemma is not known. It is almost the same as the generic Euclid's lemma (which is also 
a folklore). 
I have rewritten this proof in detail, with generalizing it from IntegralDomain to  CancellativeCommutativeSemiring,
and programmed it.
Hear is this short and clear in-formal proof:

[proof.lagda.zip](https://github.com/agda/agda-stdlib/files/5771220/proof.lagda.zip)

Can you provide a simpler proof?

In Agda, it takes some code. Thus, the line
```
If c = 0 then ``gcd ac bc  ∥  gcd 0 0  ∣∣  0  ∣∣  c ∙ (gcd a b)``.
```
of 65 characters is expressed in Agda in 240 characters (not counting trailing blanks).

This is, for example,  due to the following reasons.
a) Agda needs all details.
b) For readability, these assignments with '=' need to be one per line.
c) In comparison to Nat,  `≡` is replaced in many cases with ``∣∣``, which is expanded to the expressions
    containing two pairs of values, the two equalities to be proved, the two factor values to be introduced.  

Please, consider the issue once more.

MatthewDaggitt:
Hi @mechvel I've updated this PR after having merged in the changes from PR #1387. You will need to run `git pull` in your local repository to get the changes. I'll address the rest of your comments in a bit.

mechvel:
> [..] 
> You will need to run git pull in your local repository to get the changes.  

All right, I have applied  ``git pull; make clean; make test``. 
It shows OK.

src/Algebra/Consequences/Setoid.agda
```diff
@@ -25,7 +25,7 @@ open import Relation.Unary using (Pred)
 ------------------------------------------------------------------------
```
MatthewDaggitt, line 25: Please fix the 2 style-guide violations in this file.

src/Algebra/Definitions/RawMagma.agda
```diff
@@ -1,4 +1,4 @@
-------------------------------------------------------------------------
```
MatthewDaggitt, line 1: Please fix the 5 style guide violations in this file.

src/Algebra/Structures.agda
```diff
@@ -460,8 +460,6 @@ record IsCancellativeCommutativeSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a 
 
   open IsCommutativeSemiring isCommutativeSemiring public
 
-
-
```
MatthewDaggitt, line 464: Please do not unnecessarily delete whitespace.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
MatthewDaggitt: Here's an example of one of your earlier proofs that I've just rewritten from:
```agda
x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd] : ∀ {x y d} → (x ≉ 0#) ⊎ (y ≉ 0#) →
                               ((isGCDᶜ (q₁ , _) (q₂ , _) _) : IsGCD x y d) →
                               Coprime q₁ q₂     
x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd] {x} {y} {d} nz⊎nz
  isGCD@(isGCDᶜ (quot₁ , quot₁∙gcd≈x) (quot₂ , quot₂∙gcd≈y) greatest)
  {c} (q₁ , q₁c≈quot₁) (q₂ , q₂c≈quot₂) = c∣1
  where
  d≉0     = x≉0∨y≉0⇒gcd≉0 isGCD nz⊎nz
  q₁*dc≈x = begin
    q₁ * (d * c)    ≈⟨ Of*CSemig.x∙yz≈xz∙y q₁ d c ⟩
    (q₁ * c) * d    ≈⟨ *-congʳ q₁c≈quot₁ ⟩
    quot₁ * d       ≈⟨ quot₁∙gcd≈x ⟩
    x               ∎

  q₂*dc≈y = begin
    q₂ * (d * c)    ≈⟨ Of*CSemig.x∙yz≈xz∙y q₂ d c ⟩
    (q₂ * c) * d    ≈⟨ *-congʳ q₂c≈quot₂ ⟩
    quot₂ * d       ≈⟨ quot₂∙gcd≈y ⟩
    y               ∎

  dc∣x = q₁ , q₁*dc≈x
  dc∣y = q₂ , q₂*dc≈y
  c∣1  = let
           (q , q*dc≈d) = greatest dc∣x dc∣y
           d*qc≈d*1     = begin
             d * (q * c)   ≈⟨ Of*CSemig.x∙yz≈y∙xz d q c ⟩
             q * (d * c)   ≈⟨ q*dc≈d ⟩
             d             ≈⟨ sym (*-identityʳ d) ⟩
             d * 1#        ∎

           qc≈1 = *-cancelˡ-nonZero {d} (q * c) 1# d≉0 d*qc≈d*1
        in
        q , qc≈1
```
to
```agda
x∣y∧z∣x/y⇒x*z∣y : ∀ {x y z} → ((x/y , _) : x ∣ y) → z ∣ x/y → x * z ∣ y 
x∣y∧z∣x/y⇒x*z∣y {x} {y} {z} (x/y , x/y*x≈y) (p , p*z≈x/y) = p , (begin
  p * (x * z)  ≈⟨ Of*CSemig.x∙yz≈xz∙y p x z ⟩
  (p * z) * x  ≈⟨ *-congʳ p*z≈x/y ⟩
  x/y * x      ≈⟨ x/y*x≈y ⟩
  y            ∎)

x*y∣x⇒y∣1 : ∀ {x y} → x ≉ 0# → x * y ∣ x → y ∣ 1#
x*y∣x⇒y∣1 {x} {y} x≉0 (q , q*xy≈x) = q , *-cancelˡ-nonZero (q * y) 1# x≉0 (begin
  x * (q * y) ≈⟨  Of*CSemig.x∙yz≈y∙xz x q y ⟩
  q * (x * y) ≈⟨  q*xy≈x ⟩
  x           ≈˘⟨ *-identityʳ x ⟩
  x * 1#      ∎)

x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]2 : ∀ {x y d} → x ≉ 0# ⊎ y ≉ 0# →
                                ((isGCDᶜ (q₁ , _) (q₂ , _) _) : IsGCD x y d) →
                                Coprime q₁ q₂
x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]2 x≉0∨y≉0 gcd@(isGCDᶜ d∣x d∣y greatest) x/d∣z y/d∣z =
  x*y∣x⇒y∣1 (x≉0∨y≉0⇒gcd≉0 gcd x≉0∨y≉0) (greatest
    (x∣y∧z∣x/y⇒x*z∣y d∣x x/d∣z)
    (x∣y∧z∣x/y⇒x*z∣y d∣y y/d∣z))
```
Can you see the difference? There's significantly less code duplication and I've split out the sub-results into useful lemmas that can be re-used by other proofs. In contrast your proof is brittle as it relies on implementation details of the divisibility relation, hard to understand and therefore difficult to refactor.

Please do the same for this proof.

MatthewDaggitt (commented):
Just brought back up to date with master. Just to note that from now on I'm going to be more strict in not providing a detailed review until the code conforms to the conventions of the style guide. I'm also not going to be providing in depth explanations of style-guide violations that we've discussed with you before.

MatthewDaggitt:
> What is the final solution for the record constructor names : isGCDᶜ or mkIsGCD ?

We decided on `mkIsGCD`

mechvel:
Please, consider my commit of January 30, 2021.
All your notes are taken in account.
My replies are coming soon.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
mechvel: This is an improvement for ``x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]`` . Thank you.
I must have introduced this ``x∣y∧z∣x/y⇒x*z∣y``.
Also I did not know of the possibility to set things like `` ((foo  , _) : x ∣ y) →`` 
to signatures. 
``x*y∣x⇒y∣1``  is also useful.
Now,  I 
a) rename  `x∣y∧z∣x/y⇒x*z∣y`  to  `x∣y∧z∣x/y⇒xz∣y`
    (in mathematics, they often write `xy` for `x * y`, and I suggest to do this in names
     - - when it is clear that this means   `x * y` or `x  ∙ y`),
b) rename   `x*y∣x⇒y∣1` to `xy∣x⇒y∣1`,
c) move  `x∣y∧z∣x/y⇒xz∣y` to Algebra.Properties.CommutativeSemigroup.Divisibility,
    because it is more generic, it is for CommutativeSemigroup and its `_∙_`.

src/Algebra/Consequences/Setoid.agda
mechvel: I hope that I have found this and fixed (but not very sure).

src/Algebra/Definitions/RawMagma.agda
mechvel: I hope that I have found this and fixed (but not very sure).

src/Algebra/Structures.agda
mechvel: Sorry, I do not understand what is wrong there.

mechvel:
1) `isGCDᶜ`  replaced with `mkIsGCD` everywhere.
2) A couple of your notes occur repeated above, I do not know why.

src/Algebra/Structures.agda
MatthewDaggitt: You have deleted empty lines for no reason, in a file you have changed nothing in. Please do not change files unnecessarily.

src/Algebra/Structures.agda
```diff
@@ -23,7 +23,7 @@ module Algebra.Structures
 open import Algebra.Core
 open import Algebra.Definitions _≈_
 import Algebra.Consequences.Setoid as Consequences
-open import Data.Product using (_,_; proj₁; proj₂)
+open import Data.Product using (_,_; proj₁; proj₂; ∃)
```
MatthewDaggitt, line 26: No need for this change in imports.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
```diff
@@ -0,0 +1,115 @@
...
+    (x∣y∧z∣x/y⇒xz∣y d∣x x/d∣z)
+    (x∣y∧z∣x/y⇒xz∣y d∣y y/d∣z))
+
+------------------------------------------------------------------------------
+-- gcd-s for two division-equivalent pairs
+-- are division-equivalent
+
+GCD-unique : ∀ {x x' y y' d d'} → x ∣∣ x' → y ∣∣ y' →
+             IsGCD x y d → IsGCD x' y' d' → d ∣∣ d'
+GCD-unique (x∣x' , x'∣x) (y∣y' , y'∣y)
```
MatthewDaggitt, line 52: This property does not require cancellation and therefore should not be in this file.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -0,0 +1,51 @@
...
+open import Algebra.Consequences.Setoid setoid
+  using (comm+cancelˡ-nonZero⇒cancelʳ-nonZero)
+open import Algebra.Properties.CommutativeSemigroup *-commutativeSemigroup
+  using (x∙yz≈y∙xz)
+open import Algebra.Properties.Semiring.Divisibility semiring
+
+*-cancelʳ-nonZero : AlmostRightCancellative _≈_ 0# _*_
+*-cancelʳ-nonZero = comm+cancelˡ-nonZero⇒cancelʳ-nonZero *-comm 0# *-cancelˡ-nonZero
+
+x*y≈0⇒x≈0⊎y≈0 : Decidable _≈_ → ∀ {x y} → x * y ≈ 0# → x ≈ 0# ⊎ y ≈ 0#
```
MatthewDaggitt, line 33: By the naming changes you've made, this should be called `xy≈0⇒x≈0⊎y≈0`

src/Algebra/Definitions/RawSemiring.agda
```diff
@@ -9,6 +9,7 @@
 open import Algebra.Bundles using (RawSemiring)
 open import Data.Sum.Base using (_⊎_)
 open import Data.Nat using (ℕ)
+open import Data.Product using (proj₁; proj₂)
```
MatthewDaggitt, line 12: No need for this extra import.

src/Algebra/Definitions/RawMagma.agda
```diff
@@ -54,15 +56,38 @@ _∤_ : Rel A (a ⊔ ℓ)
...
+-- Greatest common divisor (GCD)
+
+record IsGCD (x y gcd : A) : Set (a ⊔ ℓ) where
+  constructor mkIsGCD
+  field
+    divides₁ : gcd ∣ x
+    divides₂ : gcd ∣ y
+    greatest : ∀ {z} → z ∣ x → z ∣ y → z ∣ gcd
+
+  quot₁ : A               -- a complementory quotient  x/gcd
```
MatthewDaggitt, line 83: I don't think there's any need for these extra projections, as you can always extract them via pattern matching.

src/Algebra/Consequences/Setoid.agda
MatthewDaggitt: No, they're still there. Read the section about indentation.

https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#indentation

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
MatthewDaggitt: > Putting foo1 = x∣y∧z∣x/y⇒x*z∣y d∣x x/d∣z, who divides who according to foo1 ?
Putting foo2 = x∣y∧z∣x/y⇒x*z∣y d∣y y/d∣z, does it return the same value as foo1 ?

I'm afraid I do not understand what you mean.

> Now it is somewhat simper and shorter. I wonder of what else can be done.

Nowhere should the proof explicitly rely on the fact that divisibility is implemented as a pair. It should never directly construct or deconstruct a divisibility proof. Anything else breaks the abstraction boundary. Please fix.

CHANGELOG.md
```diff
@@ -52,15 +52,29 @@ Deprecated names
 New modules
 -----------
 
-* Specifications for min and max operators
+* Specifications and properties for greater common divisor
```
gallais, line 55:
```suggestion
* Specifications and properties for greatest common divisor
```

CHANGELOG.md
```diff
@@ -138,3 +152,16 @@ Other minor additions
   *-inverseˡ : ∀ p → 1/ p * p ≡ 1ℚ
   *-inverseʳ : ∀ p → p * 1/ p ≡ 1ℚ
   ```
+
+* Added new proofs in `Algebra.Properties.CommutativeSemigroup.Divisibility`:
+  ```agda
+  x∣y∧z∣x/y⇒xz∣y : ∀ {x y z} → ((x/y , _) : x ∣ y) → z ∣ x/y → x ∙ z ∣ y
+  x∣y⇒zx∣zy      : ∀ {x y} z → x ∣ y → z ∙ x ∣ z ∙ y
```
gallais, line 159:
```suggestion
  x∣y∧z∣x/y⇒xz∣y : ((x/y , _) : x ∣ y) → z ∣ x/y → x ∙ z ∣ y
  x∣y⇒zx∣zy      : x ∣ y → z ∙ x ∣ z ∙ y
```

CHANGELOG.md
```diff
@@ -138,3 +152,16 @@ Other minor additions
...
+  ```
+
+* Added new proofs in `Algebra.Properties.CancellativeCommutativeSemiring`:
+  ```agda
+  x*y≈0⇒x≈0⊎y≈0 : Decidable _≈_ → ∀ {x y} → x * y ≈ 0# → x ≈ 0# ⊎ y ≈ 0#
+  x≉0∧y≉0⇒xy≉0  : Decidable _≈_ → ∀ {x y} → x ≉ 0# → y ≉ 0# → x * y ≉ 0#
+  xy∣x⇒y∣1      : ∀ {x y} → x ≉ 0# → x * y ∣ x → y ∣ 1#
```
gallais, line 166:
```suggestion
  x*y≈0⇒x≈0⊎y≈0 : Decidable _≈_ → x * y ≈ 0# → x ≈ 0# ⊎ y ≈ 0#
  x≉0∧y≉0⇒xy≉0  : Decidable _≈_ → x ≉ 0# → y ≉ 0# → x * y ≉ 0#
  xy∣x⇒y∣1      : x ≉ 0# → x * y ∣ x → y ∣ 1#
```

src/Algebra/Consequences/Setoid.agda
```diff
@@ -48,6 +48,24 @@ module _ {_•_ : Op₂ A} (comm : Commutative _•_) where
     x • z ≈⟨ comm x z ⟩
     z • x ∎)
 
+  comm+cancelˡ-nonZero⇒cancelʳ-nonZero :
+    (0# : A) → AlmostLeftCancellative 0# _•_ → AlmostRightCancellative 0# _•_
+  comm+cancelˡ-nonZero⇒cancelʳ-nonZero 0# cancelˡ-nonZero {x} y z x≉0 yx≈zx =
+    cancelˡ-nonZero y z x≉0 (begin
```
gallais, line 54: You can use `$` instead of parentheses around the `begin (...) ∎` block.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
gallais: Aren't we using `_∨_` in names?

src/Algebra/Consequences/Setoid.agda
mechvel: With `$`, it is not parsed. 
In the cases when  `begin` directly follows after "=",  `$` is parsed.

mechvel:
> Aren't we using _∨_ in names?

Probably, yes, but  I am not sure.
Now I replace `⊎` with `∨` (disjunction) everywhere in names.

src/Algebra/Consequences/Setoid.agda
gallais: Really? It's used in many places such as
* Data.Nat.Properties
* Data.Nat.Divisibility
* Data.Nat.Binary.Subtraction
* Data.Digit
etc., etc.

mechvel: Well, try to find anywhere  ``= <foo> $ begin ``, 
where <foo> is not empty.

gallais: I gave a list of modules where you had such examples.
For instance http://agda.github.io/agda-stdlib/Data.Nat.Divisibility.html#10002

```agda
nonZeroDivisor-lemma m zero r r≢zero (divides (suc q) eq) =
  m+1+n≰m m $ begin
    m + suc (q * suc m) ≡⟨ +-suc m (q * suc m) ⟩
    suc (m + q * suc m) ≡⟨ sym eq ⟩
    1 * toℕ r           ≡⟨ *-identityˡ (toℕ r) ⟩
    toℕ r               ≤⟨ FP.toℕ≤pred[n] r ⟩
    m                   ∎
```

mechvel: So, I do not know, why this $ is parsed in one examples and is not in others.

gallais: Did you forget to add `open import Function.Base using (_$_)` in the import list?

mechvel: Yes, I did. I am sorry.
Now I set this `$` to all these places, except the ones after comma
(because ``foo , $ begin ...`` does not look nice, and also I am not sure that this will be parsed,
also the parentheses often allow to skip the import of  the module Function).
?

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
mechvel: > Nowhere should the proof explicitly rely on the fact that divisibility is implemented as a pair. It should never directly 
> construct or deconstruct a divisibility proof. Anything else breaks the abstraction boundary. Please fix.

It looks like you contradict yourself and also the pattern matching tool propagated by Agda.
For example, in  Algebra.Properties.CancellativeCommutativeSemiring.GCD you write
```
x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd] : ∀ {x y d} → x ≉ 0# ⊎ y ≉ 0# →                           
                                ((mkIsGCD (q₁ , _) (q₂ , _) _) : IsGCD x y d) → Coprime q₁ q₂
```     
Here ``(q₁ , _)`` directly deconstructs the divisibility proof.
Do you require the same for `∃` ? 
Divisibility is defined via `∃`.  And 
``(a , Pa) :   ∃ \x -> P x)``  is all right but  ``(q , q∙a≈b) : a ∣ b)``
is not?
Agda programs (I expect standard library also) are full  with deconstructing and constructing pairs,
and the same is with other records.
Many functions use pattern matching that deconstructs some record, whether it is a pair or some other 
record.
If we define 
```
module _ {α} (A : Set α) (_∙_ : Op₂ A α)  where 
  data _∣_ A → A → Set _
     where
     divides : ∀ {q} → q ∙ x ≈ y → x ∣ y                     (II)
```
will this be good to pattern-match   ``(divides {q} qa≈b) :  a | b``    (II).
Suppose that for the next library version you decide to define  divisibility as (II).
Then a great amount of code will break.
Further, one could decide to change  ``divides`` to ``div``. Again, a great amount of code will break.
Further, one could decide to change there  ``y ≈ q ∙ x`` to  ``q ∙ x ≈ y`` (as you have done).
Again, a great amount of code will break.

The requirement of preserving the user code under changing the program format for the same mathematical definition 
is too strong for a language that uses pattern matching intensively.
For example, I see three naturally looking definitions in Agda for the divisibility notion `_|_`.
None of them is better than others.

I do not understand you.
Am I missing something?

mechvel:
Please, consider my commit of February 2, 2021.
All your note are implemented, except the ones for which I provide questions.
My notes are coming above soon.

The first note is that  as I see now,  IsGCD occurs defined in two different modules (!?).
(probably it needs to be only in RawMagma).
The next commit will have too fix this and may be other issues.

src/Algebra/Consequences/Setoid.agda
mechvel: Now, I have fixed there at least two deviations. 
?

src/Algebra/Structures.agda
mechvel: I copied this module back from to-day's _master_.

src/Algebra/Structures.agda
mechvel: fixed.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
mechvel: Indeed. Thank you.
It is moved to Algebra.Properties.Semiring.GCD.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
mechvel: Renamied to `xy≈0⇒x≈0∨y≈0`.

src/Algebra/Definitions/RawSemiring.agda
mechvel: Fixed.

src/Algebra/Definitions/RawMagma.agda
mechvel: These lines starting from No 82 are removed.

CHANGELOG.md
mechvel: Changed as suggested.

CHANGELOG.md
mechvel: Changed as suggested.

CHANGELOG.md
mechvel: Changed as suggested, with removing `*` from names and replacing `⊎` with disjunction in names.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
MatthewDaggitt: >> Nowhere should the proof explicitly rely on the fact that divisibility is implemented as a pair. It should never directly
construct or deconstruct a divisibility proof. Anything else breaks the abstraction boundary. Please fix.
>
> It looks like you contradict yourself and also the pattern matching construct propagated by Agda.
For example, in Algebra.Properties.Semigroup.Divisibility we write (and you silently agreed with this)
∣-trans (p , px≈y) (q , qy≈z) = ....
If we wrote ∣-trans x∣y y∣z = foo where p = proj₁ x∣y ...,
then would you decide that this proj₁ does not destruct divisibility by using that this is a pair?

@mechvel there's a huge difference between using the concrete implementation of divisibility in a proof about divisibility vs using the concrete implementation of divisibility in a proof about greatest common divisors. 

Unfortunately I do not have the time to give you a crash course in the concepts of abstraction and encapsulation via a Github pull request, nor the time to rewrite this proof myself. If you don't understand the difference, then I would recommend that you read up about "Software abstraction".

In the meantime we can either 1) remove this proof from the PR and we can merge the rest in, or 2) put this PR on hold?

src/Algebra/Definitions/RawMagma.agda
```diff
@@ -54,15 +56,26 @@ _∤_ : Rel A (a ⊔ ℓ)
...
 x ∣∣ y = x ∣ y × y ∣ x
 
 _∤∤_ : Rel A (a ⊔ ℓ)
-x ∤∤ y =  ¬ x ∣∣ y
+x ∤∤ y = ¬ x ∣∣ y
+
+------------------------------------------------------------------------
+-- Greatest common divisor (GCD)
+
+record IsGCD (x y gcd : A) : Set (a ⊔ ℓ) where
```
MatthewDaggitt, line 76: You've now added two copies of this definition, one in `Algebra.Definitions.RawMagma` and one in `Algebra.Definitions.RawSemiring` :cry:

src/Algebra/Consequences/Setoid.agda
MatthewDaggitt: No, the changes you've just made were not style guide violations and you've now changed existing code for no reason. The **new code** that you've added still has two style guide violations, of a type that I've discussed with you twice before. In particular look at the section I linked you to in the style guide and the types of your new proof.

src/Algebra/Properties/CancellativeCommutativeSemiring/GCD.agda
mechvel: > here's a huge difference between using the concrete implementation of divisibility in a proof about divisibility vs using 
> the concrete implementation of divisibility in a proof about greatest common divisors.

I am trying to understand you. 
Probably you mean that for a higher level item `Item` (functions, type) it is better not to use an explicit construct for any lower level item `Item'`. 
For example,  gcd-distr is about `_|_` and IsGCD. A higher level IsGCD is expressed via a lower level   `_|_` .
And it is desirable for a proof for any property P of IsGCD not to use that `_|_` is expressed as a pair.
Rather it needs to use lemmas in which `_|_` takes part as whole. For example, 
 ``lemma : Foo1 x (x | y) -> Foo2  x y (x | z)``.  Here `lemma` is a separate function considered as of a lower level,
and its  implementation has a bit more rights to use the explicit construction for ` _|_`.
Then, if another program version uses a different  construct for  `_|_`, some higher level parts of the program would not need to be changed.  The wider is applied this approach the more stable fragments has the program.
Also reasoning expressed via statements like ``Foo1 x (x | y) -> Foo2  x y (x | z)`` looks more clear than reasoning that uses pattern matching of  ``x | y`` against a pair. 

Is this what you mean? Is this a "concept of abstraction and encapsulation" ?

May be a common-reason people write "abstract-encapsulated programs" without knowing this?
This is like writing and speaking prose without knowing of a concept of prose and without looking into a crash course about this concept.
 
Another point. Probably it is not true that totally following this requirement always leads to a better program. Is it? 
For example, you succeeded with it in the implementation of  `x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]`,
but  its signature pattern-matches divisibility against a pair:  
``... ((mkIsGCD (q₁ , _) (q₂ , _) _) : IsGCD x y d) → Coprime q₁ q₂``.

Probably, there are many functions that cannot be splited into the abstract-encapsulated form in a reasonable and nice way. This is also a matter of taste.
Currently I do not see how to further "abstract-encapsulate" the function  `gcd-distr`.
May be I shall find this later. 
But my impression is that this requirement needs not be applied totally and bureaucratically.
For example,  gcd-distr has a clear signature (without breaking "abstraction"), its proof is not large and is type-checked.
If we merge it and the generic GCD project to standard, this will be a great improvement. This is more important than  whether some proof is more abstracted or less. Anyway, someone could replace it later with a nicer proof.

MatthewDaggitt: > Is this what you mean? Is this a "concept of abstraction and encapsulation" ?

Yes.

> May be a common-reason people write "abstract-encapsulated programs" without knowing this?
This is like writing and speaking prose without knowing of a concept of prose and without looking into a crash course about this concept.

I'm afraid I don't understand what you mean here.

> Probably, there are many functions that cannot be splited into the abstract-encapsulated form in a reasonable and nice way.

Yes this is true. For example you say...

> For example, you succeeded with it in the implementation of x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd],
but its signature pattern-matches divisibility against a pair:
... ((mkIsGCD (q₁ , _) (q₂ , _) _) : IsGCD x y d) → Coprime q₁ q₂.

In the case of `x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]` it is impossible to avoid breaking the abstraction in the type as it refers to one of the implementation details. Notice that the abstraction remains in place in the proof.

However `gcd-distr` does not require you to break the abstraction either in the type or in the proof. It is entirely possible to rewrite in the same way as I showed you how to rewrite the proof of `x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]`.

> There is is also a matter of taste in this. ... But my impression is that this requirement needs not be applied totally and bureaucratically.

I respectfully disagree, writing clean, maintainable and reusable code that doesn't hinder further development is not a matter of taste but good software design!

> For example, gcd-distr has a clear signature (without breaking "abstraction"), its proof is not large and is type-checked.

At the moment this proof is _huge_ by the standards of the library. I think it's easily in the top 5 largest proofs we have, and it's possibly the 2nd largest! That's no small feat in a ~100,000 lines of code.

> If we merge it and the generic GCD project to standard, this will be a great improvement. This is more important than whether some proof is more abstracted or less. Anyway, someone could replace it later with a nicer proof.

Unfortunately I'm not a fan of this course of action. If everyone did this, then the standard library would soon end up completely unmaintainable. It's also not reasonable to expect other people to come along and fix code later on. It is simply accruing technical debt.

> Currently I do not see how to further "abstract-encapsulate" the function gcd-distr.
May be I shall find this later. ... 

That's fine that you don't understand how to do it.  All that means is that this isn't a suitable contribution to the standard library at the moment, apologies.

mechvel: >> Probably, there are many functions that cannot be splited into the abstract-encapsulated form in a 
>> reasonable and nice way.

> Yes this is true. For example you say...

Another example: 
introduce to the module where _|_ is defined the functions
`quotient` and `divEquality` that extract the parts of the data of  ``(x | y)``.
Then, instead of writing   ``f = let (q , q*a==b) = a|b in <use q and q*a==b`` 
 it is possible to write
``` 
q = quotient a|b;   q*a==b = divEquality a|b   
<use q  and  q*a==b>
```
Does this satisfies the approach of abstraction and encapsulating?

MatthewDaggitt: Partially, because you're right you could then change the definition of `_|_` underneath without breaking the code. But it would still be the wrong level of abstraction to use in this proof, because you should be able to prove the properties of GCD simply from building up via lemmas about `_||_` and `_|_`.

mechvel: In mathematical textbooks a discourse like
"Suppose ``a | b``. Then, let `q` be the corresponding quotient, so that ``q * a == b``. 
Hence ``q | (b * c)``, and so on ...
"
is a common place. There are similar fragments longer than my gcd-distr code.
The thing is natural, no reader would think of "abstraction and encapsulation" at this point.  
Anyway, if it is rewritten to your style of  ``gcd-uni x≉0⊎y≉0⇒Coprime[x/gcd,y/gcd]``,  then it would be better  for the library   as well as for reading.
I am going to consider this after a certain delay.

JacquesCarette: If you head over to the Lean Zulip, you'd see that if there's one thing that the mathematicians have learned about formalizing things in Lean's mathlib is... APIs and the value of encapsulation!  What's going in mathematics is that math texts set up a "local context" or scope, where some definitions and equations are valid and visible, and then use them freely. So they use this local scope mechanism instead of setting up an explicit record to bundle up this information.

In other words, CS has a few things to teach to math about how to be precise about what the mathematical vernacular in common use actually means.

mechvel: 1. The whole subject is evident, and has nothing to do with mathematics.
2. If we define divisibility as a record `Divides` with constructor `divides`, the fields `quotient` and
`equality`, then ``(divides q eq) : Divides a b`` is an abstract enough thing.
3. Probably Matthew is right in that gcd-distr can be written in a better way.
4. Even the current version of divisibility, IsGCD,  and gcd-distr is still much better than the whole design in lib-1.5 for divisibility, GCD, Rational and such - even if the latter's are programmed very "encapsulatedly". 
Generally it is desirable to distinguish intuitively important things from not so important.

mechvel: More point: 
when a language relies heavily on pattern matching, this drives it from the abstraction you are so eager of. 
For example, I expect to find in standard library programs similar to 
```
f (_ :: (_ , y) :: ps) = y + ( f  ps)
f  _ = 0
```
What do you write for the "abstract & encapsulated" style? 
Is it
```
f ps  = case null? ps
           of \
           { (yes _) -> 0 
           ; (no notNull-ps) -> 
                   case null? (tail ps notNull-ps)
                   of \
                    { (yes _) -> 0
                    ; (no notNull-tlPs) -> proj-2  (head (tail ps notNull-ps) notNull-tlPs) + (f ...)
                    } 
              }
```
?

JacquesCarette: That second is not at all "abstract and encapsulated". I would call that imperative-style programming.  It is an ugly version of the first code, with pretty much nothing going for it.

Relying heavily on pattern-matching makes things representation-dependent. That is fine for implementing the base routines, and then gets problematic, eventually, for higher-level code.

The code for `f` clearly is a kind of fold that works on even length lists. The "abstract and encapsulated" way of doing this would be to provide such a fold abstraction, and then instantiate it with `0` and `\x acc -> proj-2 x + acc`.

mechvel: > That second is not at all "abstract and encapsulated".

Is it representation dependent? I think, not. 
It uses only  1)  `null?` to decide of emptiness,  
2) `head` and `tail` for a nonempty  a list, these actually constitute a definition of what is a list, 
3) second projection for a direct product.
Only basic mathematical notions, the representation details are not visible.  
?
> to provide such a fold abstraction, and then instantiate it with 0 and \x acc -> proj-2 x + acc.

There is a concrete example. Can you write the function source?

mechvel: > and then gets problematic, eventually, for higher-level code.

Operating with  ``gcd , IsGCD`` is higher-level.
Does this mean that one needs to avoid writing things like
```
let (mkIsGCD g g|a g|b greatest) = gcd a b in ...
```
?

MatthewDaggitt:
@mechvel if it's okay with you I'm going to remove the GCD parts of this PR and merge in the stuff that's ready to go. If you come back to the GCD proofs at some later point, you can open a new PR.

mechvel:
> if it's okay with you I'm going to remove the GCD parts of this PR and merge in the stuff that's ready to go. 

All right.

PR #1393, opened 2021-01-03
Implement all the missing pieces to proving that Fin (m ℕ.+ n) ↔ (Fin m ⊎ Fin n)

JacquesCarette:
Fully implement that + and ⊎ are inverses over `Fin`. Add `join-⊎` to `Data.Fin.Base` as the inverse to `splitAt`.


src/Data/Fin/Base.agda
```diff
@@ -131,6 +131,10 @@ splitAt zero    i       = inj₂ i
 splitAt (suc m) zero    = inj₁ zero
 splitAt (suc m) (suc i) = Sum.map suc id (splitAt m i)
 
+-- inverse of above function
+join-⊎ : ∀ m n → Fin m ⊎ Fin n → Fin (m ℕ.+ n)
```
MatthewDaggitt, line 135: Any objections to calling this simply `join`?

src/Data/Fin/Properties.agda
```diff
@@ -491,7 +492,11 @@ splitAt-raise : ∀ m n i → splitAt m (raise {n} m i) ≡ inj₂ i
 splitAt-raise zero    n i = refl
 splitAt-raise (suc m) n i rewrite splitAt-raise m n i = refl
 
-inject+-raise-splitAt : ∀ m n i → [ inject+ n , raise {n} m ] (splitAt m i) ≡ i
+splitAt-join-⊎ : ∀ m n i → splitAt m (join-⊎ m n i) ≡ i
+splitAt-join-⊎ m n (inj₁ x) = splitAt-inject+ m n x
+splitAt-join-⊎ m n (inj₂ y) = splitAt-raise m n y
+
+inject+-raise-splitAt : ∀ m n i → join-⊎ m n (splitAt m i) ≡ i
```
MatthewDaggitt, line 499: We should deprecate this name and call it `join-⊎-splitAt`/`join-⊎-splitAt`

src/Data/Fin/Properties.agda
```diff
@@ -22,6 +22,7 @@ open import Data.Product using (∃; ∃₂; ∄; _×_; _,_; map; proj₁; uncur
 open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂; [_,_])
 open import Data.Sum.Properties using ([,]-map-commute; [,]-∘-distr)
 open import Function.Base using (_∘_; id; _$_)
+open import Function.Bundles using (_↔_; mk↔′; Inverse)
```
MatthewDaggitt, line 25: No need to import `Inverse`

src/Data/Fin/Base.agda
JacquesCarette: No objection at all. I was just worried with name clashes with other uses of 'join' in the library. I'll rename.

src/Data/Fin/Properties.agda
JacquesCarette: Agreed, will do.

src/Data/Fin/Properties.agda
JacquesCarette: Probably. I think this reflects an intermediate state where I needed it to convince Agda of something.

JacquesCarette:
All comments address in 2e5e471. Documented deprecation too.

Issue #1318, opened 2020-10-17
Tactic.RingSolver doesn't work with multiplication


Taneb:
```agda
module MinimalFailing where

open import Data.Integer
open import Data.Integer.Tactic.RingSolver
open import Relation.Binary.PropositionalEquality

works : ∀ x y → x + y ≡ y + x
works = solve-∀

doesn'twork : ∀ x y → x * y ≡ y * x
doesn'twork = {! solve-∀ !}
```

If I try to fill in that hole, I get this error:

```
/home/taneb/Documents/Agda/antlion/src/MinimalFailing.agda:11,17-24
Tactic.RingSolver.Core.Expression.Expr
(Tactic.RingSolver.Core.AlmostCommutativeRing.AlmostCommutativeRing.Carrier
 ring)
2
!=< ℤ
when checking that the expression x has type ℤ
```

I would expect this, and more complicated examples involving multiplication, to work.

oisdk:
Generally to use the solve you need to use the definitions bound in the `AlmostCommutativeRing` record, i.e.:

```agda
open import Data.Integer.Tactic.RingSolver
open import Relation.Binary.PropositionalEquality

open import Tactic.RingSolver.Core.AlmostCommutativeRing using (AlmostCommutativeRing)
open AlmostCommutativeRing ring

works : ∀ x y → x + y ≡ y + x
works = solve-∀

also-works : ∀ x y → x * y ≡ y * x
also-works = solve-∀
```

Now, if you use definitions that *aren't* from the record, then the ring solver will still try its best to figure out what the operation is, by evaluating the record entry and comparing it to each function in the expression. That works for `_+_` on `ℤ` here, because the definition of `_+_` is:

```agda
_+_ : ℤ → ℤ → ℤ
-[1+ m ] + -[1+ n ] = -[1+ ℕ.suc (m ℕ+ n) ]
-[1+ m ] + +    n   = n ⊖ ℕ.suc m
+    m   + -[1+ n ] = m ⊖ ℕ.suc n
+    m   + +    n   = + (m ℕ+ n)
```

So it can't reduce, since it does pattern matching, so the solver can see that this function is the same as the function referred to in the `AlmostCommutativeRing` record. For `_*_`, on the other hand, the definition *can* reduce:

```agda
_*_ : ℤ → ℤ → ℤ
i * j = sign i S* sign j ◃ ∣ i ∣ ℕ* ∣ j ∣
```

[...]

Issue #1116, opened 2020-03-15
Tactic.RingSolver doesn't work over generic rings


MatthewDaggitt:
Currently the reflective interface for Tactic.RingSolver currently only works over concrete instances of rings (see https://github.com/oisdk/agda-ring-solver/issues/12). It would be useful to get this working for generic rings as well.

felixwellen:
I think I solved the same issue for the RingSolver in the cubical library, which made the examples here work:

[https://github.com/agda/cubical/blob/master/Cubical/Algebra/RingSolver/Examples.agda](https://github.com/agda/cubical/blob/master/Cubical/Algebra/RingSolver/Examples.agda)

I had to sacrifice the generality of AlmostRings for that, which lead to a separate solver for naturals. Under the hood, it is now a solver for algebras over the integers. And the examples rely on normalization of integers.
Is this still an issue? It might be not too much work to implement my approach together with an stdlib-expert...

MatthewDaggitt:
Hey, thanks for the heads-up! Yes I've been vaguely following your PRs on it and meant to have a deep dive to grok what your changes were doing at some point. When I was hacking on the standard library one earlier this year, I couldn't quite understand @oisdk's comment in the thread linked above about why computation on the coefficients was relevant? Even when the underlying ring is some generic ring, I don't quite understand why the coefficient ring can't stay the same and be the natural numbers which still compute?

Any chance you could provide a brief high-level explanation of what your fix was?

felixwellen:
Here is short recap of the problem (all rings commutative):

What (my version of) the ring solver essentially does is to use equality of multivariate polynomials (over the integers) to show that two expressions in a ring are equal. Equality of polynomials, even if they are multivariate, amounts to checking that all coefficients are equal. Also for the Horner forms used in the ring solver, that amounts to checking equality in the given ring.
Now a problem, that might be specific to how I or others implemented translation of ring expressions to Horner forms, is that you sometimes end up with a coefficients of some form like '1-1'. This should be zero, but in an abstract ring, its normal form cannot be the same as that of '0'.
[...]

Issue #1071, opened 2020-01-25
Ring solver doesn't appear to normalise terms?


MatthewDaggitt:
@oisdk I've just been testing the new ring solver and I've discovered that the following doesn't work
```agda
test : (x : ℕ) → x ≡ x
test x = begin
  x ≡⟨ solve (x ∷ []) ⟩
  z ∎
  where z = x
```
Unfortunately it did work with the old ring solver. From my (very!) limited knowledge, I'm guessing this might be fixable if we normalise the expressions first?

MatthewDaggitt:
It also doesn't work for compound expressions, e.g. below:
```agda
test : (x : ℕ) → suc x ≡ suc x
test x = begin
  z ≡⟨ solve (z ∷ []) ⟩
  z ∎
  where z = suc x
```
I seem to remember @gallais discussing this at some point. I think this probably harder to solve. Again the old ring solver could handles this case.

oisdk:
Unfortunately for this kind of solver it would probably be impossible to normalise terms. An expression like this:

```agda
1 + x
```

Can be parsed in the ring solver into the correct AST. But if we normalise it, we would get:

```agda
suc x
```

Which can't, because it doesn't contain the ring operators. (it actually can handle this case, because I have put in a special case for `Nat`, but it isn't able to handle similar stuff for integers, etc.) In general, we have to avoid normalisation because we lose the names of the ring methods. 

To fix the problem, there are two things we can do:

1. Rewrite the goal to be in the form that suits the solver. This would mean renaming `f x` to some variable in the second example, or avoiding the variable in the first.
2. You could use the old interface with the new backend (Tactic.Ringsolver.NonReflective).

At any rate, it would probably be worth putting together some documentation on these two cases in particular and how to avoid them.

MatthewDaggitt:
Thanks for the clear explanation of why we can't normalise them! So example 1 is probably out. Mildly irritating but not too much of a problem as arguably it doesn't make the proofs very easy to read if you have aliases all over the place.

However couldn't we do something about the second set of examples? We're actually passing the "unit" we want to solve for, i.e. `f x` to the solver. So could we carry out your first suggestion automatically via reflection, by going through the expression and rewriting every instance of `f x` with a fresh variable, solving it and then reinserting `f x` afterwards?

oisdk:
I'm not sure how you could spot occurrences of `f x` around the place: I'll have a think about it and see if I can come up with a solution.

[...]

Issue #1070, opened 2020-01-25
Problem with exponentiation in `Data.Nat.Tactic.RingSolver`


MatthewDaggitt:
The solver defines its own version of exponentiation in `Algebra.Operations.Ring` that differs from that in `Data.Nat.Base`. This means that the solver does not work for equations involving exponentiation in `Data.Nat.Base`.

We may have to change the implementation of exponentiation in `Data.Nat.Base`...

gallais:
That seems dodgy: the solver should not rely on one particular implementation.
As long as the two functions are extensionally equal, it should be fine!

oisdk:
I think it should be workable by just having Data.Nat call an exponentiation operator defined on semirings. It could be defined the same way (i.e. rather than the weird way the solver defines it) it just would need to be a function defined on all semirings, not just Nat.

PR #1394, opened 2021-01-03
Add solvers for rationals

uzytkownik:


MatthewDaggitt:
Thanks for the PR. We were hoping to deprecate this solver in favour of the new tactic enabled solver in `Tactic.RingSolver` (#1069) but given the current issues with the new solver (#1070, #1071, #1116, #1318) that no-one is tackling, it seems like we're stuck with this one for a while...

PR #1396, opened 2021-01-05
Deprecate Algebra.Operations.Ring and refactor Tactics.RingSolver

### LINKED ISSUE #1389: Define exponentiation over a semiring as multiplication over the multiplicative monoid.
module Algebra.Properties.Semiring.Exponentiation  
has
```
_^_ : Carrier → ℕ → Carrier                                                            
x ^ zero  = 1#                                                                         
x ^ suc n = x * x ^ n                                                                  
```
It is much more sensible to define this in a more general case, in  Algebra.Properties.Monoid, like this:
```
module Algebra.Properties.Monoid.Foo  {a ℓ} (M : Monoid a ℓ) where 
...
_^_ : Carrier → ℕ → Carrier                                                            
x ^ zero  =  ε                                                                       
x ^ suc n = x ∙ x ^ n                                                                  
```
``^-congˡ`` and many other useful properties to be proved there. 
As to Semiring R,  it has +-monoid and  *-monoid. And exponentiation over R is imported
from  ``Algebra.Properties.Monoid.Foo *-monoid``  with renaming `ε`  to 1#,   `_∙_` to `_*_`.
Further, multiplication  ``_×_``  by  a natural in R, is actually ``_^_`` imported from 
``Algebra.Properties.Monoid.Foo +-monoid``   with renaming ``_^_`` to ``_×_``,
only we need to swap the arguments in the  current definition of  ``_×_`` (as I recall).
There is no need to define these operations several times, one needs only to import them from one  place with  
renaming. The same is for the proofs for the properties of ``_^_``  and ``_×_``.

Further, product (Π) to be defined over Monoid, with using  `ε`  and ``_∙_``.
Importing it from    ``Algebra.Properties.Monoid.Foo *-monoid``    makes it a product (by `_*_`) 
in Semiring.
Importing it from    ``Algebra.Properties.Monoid.Foo +-monoid``   with renaming `Π` to `sum`
makes it a sum (by `_+_`) in Semiring. The proofs also are imported this way.

Useful  properties for exponentiation are likely to be, for example, 
``x ^ (m + n) ≈ x ^ m ∙ x ^ n``, and some of them related to divisibility.
 Useful  properties for product (Π) are likely to be 
``Π (xs ++ ys) ≈ ...``, and some of them related to divisibility.
If we define `^` and `Π` in one place, and prove their properties in one place, and import these items
to other places with renaming, we save proofs and avoid multiplying notions.

MatthewDaggitt:
This PR:
 - finishes off deprecating `Algebra.Operations` by deprecating `Algebra.Operations.Ring`
 - introduces new definitions for exponentation (both optimised and non-optimised) over semirings that are simply multiplication defined over multiplicative monoid (fixes #1389)
 - extracts many of the lemmas about exponentiation from `Tactics.RingSolver`


MatthewDaggitt:
The one remaining issue is that I think the module names are currently very long, e.g. `Algebra.Properties.CommutativeSemiring.Exponentiation.TCOptimised`. I think I'd probably vote in favour of renaming `Exponentiation/Multiplication/Summation` to `Exp/Mult/Sum`?

src/Algebra/Definitions/RawMonoid.agda
```diff
@@ -55,7 +55,9 @@ infixl 8 _×′_
 _×′_ : ℕ → Carrier → Carrier
 0     ×′ x = 0#
 1     ×′ x = x
-suc n ×′ x = x + n ×′ x
+suc n ×′ x = n ×′ x + x
+
```
gallais, line 59: Should this breaking change be mentioned in the CHANGELOG?

MatthewDaggitt: No need as this definition is new to this version of the library. The old version of the operator still exists in `Algebra.Operations.Monoid`.

PR #1397, opened 2021-01-05
add property Fin0↔⊥

JacquesCarette:
This is another useful piece of showing that finite types are a skeleton of the semiring of types. I've needed it independently too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -723,3 +723,8 @@ Other minor additions
   ```agda
   resp : (P : Pred A ℓ) → P Respects _≡_
   ```
+
+* Added new proof to `Data.Fin.Properties`:
+  ```agda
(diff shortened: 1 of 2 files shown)


gallais:
Should this (and the content of #1393) go into `Data.Fin.Algebra`
just like their equivalents go into `Data.(Sum/Product).Algebra`?

JacquesCarette:
Good question. The Bundles most likely should. I don't know about the pieces though.

MatthewDaggitt:
Hmm I don't think I'd be in favour of putting them there, as the results do not have anything to do with the algebra hierarchy...
`Fin` is not a binary op and therefroe does not form any algebraic structures (e.g. magma/semigroup). Personally, I'm happy with their current location.

JacquesCarette:
In which case, I think this is ready to go, same as with #1393.

Issue #1370, opened 2020-12-05
No proof that _*_ has an inverse for ℚ


Taneb:
`Data.Rational.Unnormalised.Properties` has `*-inverseˡ : ∀ p {p≢0 : ℤ.∣ ↥ p ∣ ≢0} → 1/_ p {p≢0} * p ≃ 1ℚᵘ` and `*-inverseʳ : ∀ p {p≢0 : ℤ.∣ ↥ p ∣ ≢0} → p * 1/_ p {p≢0} ≃ 1ℚᵘ` but no corresponding proofs exist in `Data.Rational.Properties`.

I made a brief attempt at proving `*-inverseˡ : ∀ p {p≢0 : ∣ ↥ p ∣ ≢0} → 1/_ p {p≢0} * p ≡ 1ℚ` in terms of the unnormalised version but I didn't succeed.

uzytkownik:
It seems to be added by #1218 ?

MatthewDaggitt:
> It seems to be added by #1218 ?

That only added it for unnormalised rationals.

PR #1399, opened 2021-01-11
Add *-inverseˡ and *-inverseʳ to Data.Rational.Properties

Taneb:
Solves #1370 

I'm not happy with my proof of `*-inverseˡ`, I'd appreciate some advice for refactoring it


src/Data/Rational/Properties.agda
```diff
@@ -810,6 +814,23 @@ private
 *-distrib-+ : _*_ DistributesOver _+_
 *-distrib-+ = *-distribˡ-+ , *-distribʳ-+
 
+*-inverseˡ : ∀ p {p≢0 : ∣ ↥ p ∣ ≢0} → 1/_ p {p≢0} * p ≡ 1ℚ
+*-inverseˡ p {p≢0} = toℚᵘ-injective (begin-equality
+  toℚᵘ (1/ p * p)
+    ≈⟨ toℚᵘ-homo-* (1/ p) p ⟩
+  toℚᵘ (1/ p) ℚᵘ.* toℚᵘ p
+    ≈⟨ ℚᵘ.*-cong (toℚᵘ-homo-1/ p {p≢0}) ℚᵘ.≃-refl ⟩
```
MatthewDaggitt, line 822: This could be simplified by adding `*-congˡ` and `*-congʳ` to `Unnormalised.Properties`. Which would probably then allow you to move the justifications onto the same line as the terms without breaking the line limit?

src/Data/Rational/Properties.agda
```diff
@@ -810,6 +814,23 @@ private
...
+  ℚᵘ.1/ (toℚᵘ p) ℚᵘ.* toℚᵘ p
+    ≈⟨ ℚᵘ.*-inverseˡ (toℚᵘ p) {p≢0} ⟩
+  ℚᵘ.1ℚᵘ
+    ∎
+  )
+  where
+    open ℚᵘ.≤-Reasoning
```
MatthewDaggitt, line 829: The `open` should go on the same line as the `where`. No need to take up too many lines :smile:

src/Data/Rational/Properties.agda
```diff
@@ -810,6 +814,23 @@ private
 *-distrib-+ : _*_ DistributesOver _+_
 *-distrib-+ = *-distribˡ-+ , *-distribʳ-+
 
+*-inverseˡ : ∀ p {p≢0 : ∣ ↥ p ∣ ≢0} → 1/_ p {p≢0} * p ≡ 1ℚ
```
MatthewDaggitt, line 817: Marginally nicer to write this as `(1/ p) {p≢0}` rather than `1/_ p {p≢0}`?

MatthewDaggitt (commented):
> I'm not happy with my proof of *-inverseˡ, I'd appreciate some advice for refactoring it

I can't see anything particular wrong with it! Why are you not happy?

Taneb:
> I can't see anything particular wrong with it! Why are you not happy?
 
When I first opened this PR I wasn't using the setoid reasoning syntax, and had the proof as a very long line. It's better now :)

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -846,6 +846,12 @@ p≤q⇒0≤q-p {p} {q} p≤q = begin
   (↥ y ℤ.* ↥ v) ℤ.* (↧ x ℤ.* ↧ u) ∎)
   where open ≡-Reasoning; open ℤ-solver
 
+*-congʳ : ∀ p {q r} → q ≃ r  → p * q ≃ p * r
```
MatthewDaggitt, line 849: Sorry just spotted that these proofs should have `RightCongruent` and `LeftCongruent` as types to ensure consistency with the rest of the library. Will need to be updated in the CHANGELOG as well.

Taneb: While making this change I realised that I had the names the wrong way around! (At least relative to `LeftCongruent` and `RightCongruent`. I can't remember what I referred to when I named them the first time round)

MatthewDaggitt: Awesome, that's exactly why we have the standardise names :smile:

PR #1402, opened 2021-01-18
Add grammar hierarchy and associated structures

uzytkownik:
This is draft for now letting people know WIP. It doesn't integrate withText.Regex for now as it wasn't present when I started. No changelog entires yet as they are hard to merge.


src/Data/Nat/Properties.agda
```diff
@@ -243,6 +243,9 @@ n≤1+n _ = ≤-step ≤-refl
 n≤0⇒n≡0 : ∀ {n} → n ≤ 0 → n ≡ 0
 n≤0⇒n≡0 z≤n = refl
 
+1+m≤1+n⇒m≤n : ∀ {m n} → 1 + m ≤ 1 + n → m ≤ n
```
gallais, line 246: FYI this is a special case of both:
* `pred-mono`
* `+-cancelˡ-≤`

MatthewDaggitt: Or indeed it's simply `≤-pred`

MatthewDaggitt:
Hi @uzytkownik, thanks for the suggested contribution. While it's great code, having read through it, I think the `Text.Language` files would be better suited as a standalone library rather than as part of the standard library. Grammars and regular languages, while useful, are not required for every or even most developments.

(I should note I'm aware of modules like `Data.Fin.Substitution` which is very similar. If we were starting over, I would probably argue for not including that either.)

As for the other contributions we'd be more than happy to include them as their own pull request :+1:

MatthewDaggitt:
I'm going to close this for now. As mentioned above, please do feel free to submit the non-grammar contributions in their own PR if you would find it useful!

PR #1390, opened 2020-12-31
Add min, max and absolute value for unnormalised rationals

uzytkownik:

PR #1403, opened 2021-01-19
Add specification and properties for min/max operators

MatthewDaggitt:
This PR is motivated by @uzytkownik having to reprove the same properties of min/max for integers, unnormalised rationals, rationals etc. over and over again despite the proofs being identical.

I introduce a specification for the min/max operators in `Algebra.Construct.NaturalChoice.(MinOp/MaxOp)` in terms of the underlying ordering and then prove properties against that specification. Therefore for each numeric datatype you simply prove that it obeys the spec, and then import the properties for free (modulo a little bit of renaming) without having to reprove them.

So far I've only refactored `Data.Nat.Properties`. I'll move onto `Data.Integer.Properties` and `Data.Rational.Unnormalised.Properties` in #1390 if people are happy with this approach?

@mechvel I'm planning on playing the same trick with GCD so we don't end up reproving the same properties over and over again.


mechvel:
Without looking into this, it is evident that `min` and `max` and their main properties can easily be defined/implemented in general (in PartialOrder or in DecTotalOrder ...), and then these items to be imported to Nat, Integer, Rational by 
substituting the parameter structure instance.
As to GCD, one needs a generic IsGCD notion and a generic EuclideanDomain notion,
and then  gcd  for ``ℕ, ℕᵇ, ℤ, ℤᵇ``, and its proofs  can be done in a more or less an unique way.
This is also for some other domains which are not likely to join standard library, but are used 
in applications. But 
a) this needs introducing EuclideanDomain,
b) there remain important domains with  gcd  satisfying the needed GCD properties but
    that need  gcd  implemented by a different method and provided with a different kind of proofs,
c) this all is under the same IsGCD notion.

As to Rational, most of its arithmetic and proofs need to be replaced with the generic Fraction 
domain constructor based on GCDSemiring, GCDDomain (using IsGCD).  And most of
Rational items to be imported from OfFraction with substititing the GCDSemiring instance.
Then we have
 ```
Rational = Fraction gcdDomain-ℤ                                 -- for standard
RationalFunction = Fraction gcdDomain-Polynomial    -- for applied programs
```

**Comment by mechvel** [comment]
Now, with looking into this (generic min/max).
My first impression is that 
``Algebra.Construct.NaturalChoice.Min  {a ℓ₁ ℓ₂} (totalOrder : TotalOrder a ℓ₁ ℓ₂)``
already contains everything needed.
```
_⊓_ : Op₂ A
x ⊓ y with total x y
... | inj₁ x≤y = x
... | inj₂ y≤x = y

⊓-sel : Selective _⊓_
⊓-sel x y with total x y
... | inj₁ x≤y = inj₁ Eq.refl
... | inj₂ y≤x = inj₂ Eq.refl
...
```
All the proofs implemented in general.
Now, for ℕ and  ℤ, you only have import Algebra.Construct.NaturalChoice.Min
with substituting  totalOrder-ℕ,  totalOrder-ℤ.
What is the problem?
Why does it introduce Algebra.Construct.NaturalChoice.Base.MinOperator ?
Without this, what is lost?
Is this complication needed?

Another question: why such a strange name of NaturalChoice in Algebra.Construct.NaturalChoice.Min ? 
According to the library style, this module needs probably to  be somewhat
 Relation.Binary.Properties.TotalOrder.Min
-- "operations and the properties related to TotalOrder in the part of `minumum`".
?

**Comment by mechvel** [comment]
Continuing looking-in.
I observe `master` and do not see  Algebra.Construct.NaturalChoice.Base.
Earlier I have seen Algebra.Construct.NaturalChoice.Base.MinOperator  somewhere.
And as I recall, its  ``record MinOperator ...``
is actually a notion of  a _domain the min operator_.
Actually it could be a bundle, if only the context of (Is)TotalOrder is in Structures.IsDomainWithMin.
And I do not think that this DomainWithMin is needed, as I write earlier, the module
``Relation.Binary.Foo1.Foo2.Min {a ℓ₁ ℓ₂} (totalOrder : TotalOrder a ℓ₁ ℓ₂)`` 
is probably sufficient.
`min` is not such big thing as to provide a bundle domain for it.

As to GCD, I am frightened with your initiative. 
If you introduce ``record GCDOperator ...`` similar as you are going to do with  ``record MinOperator ...``,
it will be actually the same as suggested earlier ``record IsGCDSemiring``,
because MinOperator includes the operation field `min`, and  IsGCDSemiring includes the field `gcd`.
Also in my PR IsGCDSemiring relies on IsGCD.
It is important that IsGCD does not mention the gcd operation, so that some important properties of the GCD 
_relation_ are proved without having a function that finds the gcd value.
I fear of that you are going to bring a confusion.

**Comment by mechvel** [comment]
Continuing looking into the generic min/max suggestion.

As I wrote, it is important that IsGCD defines a _relation_ of GCD, without mentioning any way to evaluate gcd.
And `Structures.GCDSemiring` is separate, it includes a field for the algorithm of `gcd`.
Now, what can be a similar approach to the minimum notion ``min`` ?
For example, the module
```
module OfMin {c ℓ₁ ℓ₂} (ord : TotalOrder c ℓ₁ ℓ₂)  where                                                                           
open TotalOrder ord   
                                                                                  
data Min (x y z : A) : Set _                                                    
    where                                                                         
    ≤ : x ≤ y → Min x y x                                                         
    ≰ : x ≰ y → Min x y y                                                         

defines a notion (_relation_) of minimum without mentioning a way to evaluate the minimum value.
Then, it could be introduced  `Structures.WithMinimum` 
that has the fields of  totalOrder and ``min : ∀ x y →  ∃ \z → Min x y z``.
How will this `min` be evaluated in the instances?
In DecTotalOrder, this is simply by applying ``_≤?_`` once.
So,  (min code + its proof) is trivial.
Has it sense to introduce these two things (in appropriately named modules and with appropriate names!).
This WithMinimum contains somewhat too few to introduce a bundle for it.
Probably 
``Relation.Binary.Foo1.Foo2.Min {a ℓ₁ ℓ₂} (totalOrder : TotalOrder a ℓ₁ ℓ₂)``
as the result of moving the current file ``Algebra.Construct.NaturalChoice.Min`` of _master_
does all that it is needed.

With IsGCD and GCDSemiring, it is different. Because it is possible to prove important properties of the GCD  _relation_
using only IsGCD. And this _does not even presume that for all x and y there exists GCD(x, y)_.
Further, how to compute gcd in instances?
For Integer, it is an algorithm that applies recursively division with remainder (divMod).
For polynomials over Integer, there are various messy methods.
So, in instances, there is a problem of how to compute gcd. Even in simple cases of Nat and Integer
the code for  (gcd  + its proof)  is not trivial.
So that it has sense to introduce separately IsGCD for the relation and a bundle for GCDSemiring.

MatthewDaggitt:
> I do not understand: wat is the problem? Why does it introduce further Algebra.Construct.NaturalChoice.Base.MinOperator ? Without this, what is lost?

The problem with `Algebra.Construct.NaturalChoice.Min` is that enforces a single implementation that uses a proof of totality which is inefficient in many cases (e.g. `Data.Nat`). `Algebra.Construct.NaturalChoice.MinOp` leaves the implementation details up to the user and so you can write an efficient implementation of your choice but still get all the proofs for free.

> Another question: why such a strange name of NaturalChoice in Algebra.Construct.NaturalChoice.Min ?
According to the library style, this module needs probably to be somewhat
Relation.Binary.Properties.TotalOrder.Min
-- "operations and the properties related to TotalOrder in the part of minumum".?

We're constructing an algebraic operation with algebraic laws and therefore it lives in the algebra hierarchy. Going the opposite direction by inducing an ordering from an operator lives in `Relation.Binary.Construct.NaturalOrder.Left`. 

https://github.com/agda/agda-stdlib/blob/f0c53d657d65a200be0879eeb92feaebdd3089fb/src/Relation/Binary/Construct/NaturalOrder/Left.agda#L30

> As to GCD, I am frightened with your initiative. ... It is important that IsGCD does not mention the gcd operation, so that some important properties of the GCD relation are proved without having a function that finds the gcd value.

> And this does not even presume that for all x and y there exists GCD(x, y)

Okay thanks for explanation, the second point above is the crucial point that I was missing. In some domains there may not be always be a GCD for every pair of elements. Okay, so I guess a combination of the two approaches is appropriate. I'll put a PR after this.

> Now, what can be a similar approach to the minimum notion min ? ... Has it sense to introduce these two things (in appropriately named modules and with appropriate names!).

I agree such a ternary relation could be added so that you could still have properties of min and max over a poset for example. However they can be added in a backwards compatible manner, so if no-one needs them then for the moment I think we can leave them out.

mechvel:
Looking at Algebra.Construct.NaturalChoice.Min,  Algebra/Construct/NaturalChoice
I wonder:  why is it chosen such a strange word: NaturalChoice ? 
Is not the word "Ordering" ("OfOrdering") more meaningful , more related?
Because min and max are about an ordering `_≤_`.

PR #1405, opened 2021-01-23
Implement version of Data.Nat.Binary.fromℕ that runs in O(log n)

Taneb:
This makes the dependency footprint of this module a fair bit heavier I'm afraid.

I'll add a changelog entry after the 1.5 release to avoid rebasing difficulties.

I tested this with a little program that prints 10000000 round-tripped through `ℕᵇ`. With the old one, this took 1.5s, with the new version 0.002s.

### DIFF
--- a/src/Data/Nat/Binary/Base.agda
+++ b/src/Data/Nat/Binary/Base.agda
@@ -13,12 +13,17 @@
 module Data.Nat.Binary.Base where
 
 open import Algebra.Core using (Op₂)
+import Data.Fin.Base as Fin
 open import Data.Nat.Base as ℕ using (ℕ)
+open import Data.Nat.DivMod using (_divMod_; result)
+open import Data.Nat.Properties
+open import Data.Nat.Induction using (<-rec)
+open import Data.Product using (_,_)
 open import Data.Sum.Base using (_⊎_)
-open import Function.Base using (_on_)
+open import Function.Base using (case_of_; _on_; _$_)
 open import Level using (0ℓ)
 open import Relation.Binary.Core using (Rel)
-open import Relation.Binary.PropositionalEquality using (_≡_)
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
 open import Relation.Nullary using (¬_)
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


oisdk:
Instead of using well-founded recursion, you can use [the number itself to prove termination](https://oisdk.github.io/agda-playground/Data.Binary.Conversion.Fast.html). I have found this to be much quicker in practice. (and it doesn't require as many dependencies)

I also have proofs that the fast version is the same as the slow, if that is of any help: https://oisdk.github.io/agda-playground/Data.Binary.Conversion.Fast.Properties.html

src/Data/Nat/Binary/Base.agda
```diff
@@ -112,10 +117,29 @@ toℕ zero     =  0
...
+        q ℕ.* 1 ≤⟨ *-monoʳ-≤ q (ℕ.s≤s ℕ.z≤n) ⟩
+        q ℕ.* 2 ≡˘⟨ prop ⟩
+        n       ∎
+      ]
+    ; (result q (Fin.suc Fin.zero) prop) →
+      -- n is odd so suc n is even
+      2[1+ p q $ ℕ.s≤s $ begin
+        q       ≡˘⟨ *-identityʳ q ⟩
+        q ℕ.* 1 ≤⟨ *-monoʳ-≤ q (ℕ.s≤s ℕ.z≤n) ⟩
+        q ℕ.* 2 <⟨ ≤-reflexive (≡.sym prop) ⟩
```
Taneb, line 137: I'm not happy with this line. Is there a cleaner way?

MatthewDaggitt: I think it can be written as `≡˘⟨ prop ⟩`

MatthewDaggitt (commented):
@Taneb thanks for the PR! You'll also need to update all the relevant proofs as well which may prove to be a bit harder given the increase in complexity. @oisdk's version does look a lot neater with less dependencies and the former might be a significant advantage when it comes to proving properties about it...

src/Data/Nat/Binary/Base.agda
Taneb: Sadly not. That gives:

```agda
Goal: n ≡ q ℕ.* 2
Have: n ≡ ℕ.suc (q ℕ.* 2)
```

What I have is to convince Agda that the extra `suc` is OK to have in an inequality proof. Note I'm using `≤-reflexive` and then using it as a proof of `_<_`.

Taneb:
I'm not a huge fan of @oisdk's version, but if I'm being honest with myself, about 70% of that is "I didn't come up with it". 
The remaining 30% is that it makes me sad there's an auxiliary variable to prove termination, that doesn't actually match what it means for the function to terminate, and also that I wanted to learn how to use the `Induction` machinery.
I accept that these don't add up to anything near a reason not to use it, but I still want to try to prove the properties of this form, as a learning exercise if nothing else.

src/Data/Nat/Binary/Base.agda
MatthewDaggitt: Well in that case, I'd split it into two lines:
```
q ℕ.* 2        <⟨ n<1+n _ ⟩
suc (q ℕ.* 2)  ≡˘⟨ prop ⟩
n              ∎
```

MatthewDaggitt:
> I accept that these don't add up to anything near a reason not to use it, but I still want to try to prove the properties of this form, as a learning exercise if nothing else.

Definitely an admirable goal! Though fair warning I suspect it's going to pretty messy :smile: 

We probably want the best code in the standard library though so just a heads up that if you continue with this definition, I probably won't merge the results in! Not a problem of course, but I thought I should give you an explicit heads up :+1:

MatthewDaggitt:
Okay, closing this PR. However, I've opened an issue https://github.com/agda/agda-stdlib/issues/1468 to keep track of the problem.

PR #1417, opened 2021-02-05
Prove that for m n integer, if m * n is 0 then either m or n is 0

Taneb:


gallais (approved):

src/Data/Integer/Properties.agda
```diff
@@ -1601,6 +1601,11 @@ suc-* m n = begin
 -1*n≡-n +0       = refl
 -1*n≡-n +[1+ n ] = cong -[1+_] (ℕₚ.+-identityʳ n)
 
+m*n≡0⇒m≡0∨n≡0 : ∀ m {n} → m * n ≡ 0ℤ → m ≡ 0ℤ ⊎ n ≡ 0ℤ
+m*n≡0⇒m≡0∨n≡0 m p with ℕₚ.m*n≡0⇒m≡0∨n≡0 ∣ m ∣ (abs-cong {s₂ = Sign.+} p)
```
Taneb, line 1605: Would it be better to use `Data.Sum.map` here, or does it not make much difference?

MatthewDaggitt: Probably not much difference to be honest. It's not like it would fit on one line...

Taneb:
```agda
m*n≡0⇒m≡0∨n≡0 m p = Sum.map ∣n∣≡0⇒n≡0 ∣n∣≡0⇒n≡0 (ℕₚ.m*n≡0⇒m≡0∨n≡0 ∣ m ∣ (abs-cong {s₂ = Sign.+} p))
```
Wouldn't be the longest line in the file

gallais: It would be a lot less readable though

Taneb:
Is there anything else that needs doing here?

MatthewDaggitt:
Nope, all good! Thanks for the PR!

Issue #1422, opened 2021-02-15
Add inverse to `Data.Fin.quotRem` and properties


Taneb:
It would be nice to be able to demonstrate that `Fin (x * y)` is isomorphic to `Fin x × Fin y`. This would be useful for proving, for example, in `agda-categories` that `FinSetoids` is cartesian, or in something that I'm working on, that you can combine two finite state machines in various way

The inverse function can be implemented as:

```agda
combine : ∀ {n k} → Fin k → Fin n → Fin (n ℕ.* k)
combine x zero = inject+ _ x
combine x (suc y) = raise _ (combine x y)
```

With one direction of inverse proof as:

```agda
 : ∀ {n k} x (y : Fin n) → quotRem k (combine x y) ≡ (x , y)
quotRem-combine {suc n} {k} x 0F rewrite splitAt-inject+ k (n ℕ.* k) x = refl
quotRem-combine {suc n} {k} x (suc y) rewrite splitAt-raise k (n ℕ.* k) (combine x y) = cong (Data.Product.map₂ suc) (quotRem-combine x y)
```

But I haven't been able to prove the other direction

MatthewDaggitt:
Without looking at much, @uzytkownik opened a recent PR (https://github.com/agda/agda-stdlib/pull/1402) with some results about `quotRem` in. They might be relevant?

MatthewDaggitt:
In fact I think they implemented exactly the results you're looking for!

Taneb:
Ah! I hadn't seen that. The types match roughly what I had in mind, but the proofs seem very complicated... I'm going to hunt around for a simper proof

Taneb:
Finding a simpler proof turned out to be way easier than expected. I'll make a PR this evening

gallais:
Can we please make `combine` take its arguments in the same order as `_*_`?
So `combine : Fin n → Fin k → Fin (n ℕ.* k)` rather than `Fin k → Fin n → ...`

Taneb:
> Can we please make `combine` take its arguments in the same order as `_*_`?
> So `combine : Fin n → Fin k → Fin (n ℕ.* k)` rather than `Fin k → Fin n → ...`

I have no objections to this. I was going with consistency with `quotRem`, which reverses them. Maybe that should also change?

JacquesCarette:
I have a horrible proof in `pi-dual` that I meant to clean up and add too. A nice proof is better!

[...]

PR #1423, opened 2021-02-15
Add Data.Fin.combine as inverse of quotRem, and properties thereof

Taneb:
See also #1422 

There has been discussion about whether the type of `combine` should be altered to match the multiplication and the return type. Currently it matches the type of `quotRem`. I'm not certain how to resolve this.

### DIFF
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -143,10 +143,14 @@ quotRem {suc n} k i with splitAt k i
 ... | inj₁ j = j , zero
 ... | inj₂ j = Product.map₂ suc (quotRem {n} k j)
 
--- inverse of above function
-combine : ∀ {n k} → Fin k → Fin n → Fin (n ℕ.* k)
-combine {suc n} {k} x zero = inject+ (n ℕ.* k) x
-combine {suc n} {k} x (suc y) = raise k (combine x y)
+-- a variant of quotRem the type of whose result matches the order of multiplication
+remQuot : ∀ {n} k → Fin (n ℕ.* k) → Fin n × Fin k
+remQuot k = Product.swap ∘ quotRem k
+
+-- inverse of remQuot
+combine : ∀ {n k} → Fin n → Fin k → Fin (n ℕ.* k)
+combine {suc n} {k} zero y = inject+ (n ℕ.* k) y
+combine {suc n} {k} (suc x) y = raise k (combine x y)
 
 ------------------------------------------------------------------------
 -- Operations
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -145,8 +145,8 @@ quotRem {suc n} k i with splitAt k i
 
 -- inverse of above function
 combine : ∀ {n k} → Fin k → Fin n → Fin (n ℕ.* k)
-combine x zero = inject+ _ x
-combine x (suc y) = raise _ (combine x y)
+combine {suc n} {k} x zero = inject+ (n ℕ.* k) x
+combine {suc n} {k} x (suc y) = raise k (combine x y)
 
 ------------------------------------------------------------------------
 -- Operations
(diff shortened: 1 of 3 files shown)


src/Data/Fin/Properties.agda
```diff
@@ -530,6 +530,35 @@ splitAt-≥ (suc m) (suc i) (s≤s i≥m) = cong (Sum.map suc id) (splitAt-≥ m
...
+... | inj₂ j | P.[ eq ] = begin
+  raise {n ℕ.* k} k (uncurry combine (quotRem {n} k j)) ≡⟨ cong (raise k) (combine-quotRem {n} k j) ⟩
+  join k (n ℕ.* k) (inj₂ j)                             ≡˘⟨ cong (join k (n ℕ.* k)) eq ⟩
+  join k (n ℕ.* k) (splitAt k i)                        ≡⟨ join-splitAt k (n ℕ.* k) i ⟩
+  i                                                     ∎
+  where open ≡-Reasoning
+
+------------------------------------------------------------------------
+-- Bundles
+*↔× : ∀ {m n} → Fin (m ℕ.* n) ↔ (Fin n × Fin m)
```
JacquesCarette, line 559: I get why this is flipped. But this doesn't match what's advertised at the top... and is also quite counter-intuitive. It would make a lot more sense to keep this name for the version obtained via post-composing this with `swap`.

JacquesCarette (changes requested):

src/Data/Fin/Base.agda
```diff
@@ -143,6 +143,11 @@ quotRem {suc n} k i with splitAt k i
 ... | inj₁ j = j , zero
 ... | inj₂ j = Product.map₂ suc (quotRem {n} k j)
 
+-- inverse of above function
+combine : ∀ {n k} → Fin k → Fin n → Fin (n ℕ.* k)
+combine x zero = inject+ _ x
+combine x (suc y) = raise _ (combine x y)
```
MatthewDaggitt, line 149: Can we fill in the `_` here? Even if it means we have to add the implicit arguments, I think it'll make it more readable?

src/Data/Fin/Properties.agda
```diff
@@ -18,7 +18,7 @@ open import Data.Fin.Patterns
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; s≤s; z≤n; _∸_)
 import Data.Nat.Properties as ℕₚ
 open import Data.Unit using (tt)
-open import Data.Product using (∃; ∃₂; ∄; _×_; _,_; map; proj₁; uncurry; <_,_>)
+open import Data.Product using (∃; ∃₂; ∄; _×_; _,_; map; proj₁; proj₂; uncurry; <_,_>)
```
MatthewDaggitt, line 21: Don't need to add the `proj₂` import?

src/Data/Fin/Properties.agda
MatthewDaggitt: Yes, I would define `remQuot` and prove everything in terms of that. Unsure whether it should be defined in terms of `quotRem` or whether we should redefine it and deprecate `remQuot` entirely? Thoughts?

(`remQuot` is a relatively new addition so deprecating it shouldn't cause too much pain)

src/Data/Fin/Properties.agda
Taneb: Ah, I did in a slightly earlier version where I'd needlessly inlined uncurry. I'll fix that

src/Data/Fin/Properties.agda
Taneb: I've added `remQuot` in terms of `quotRem`. I don't think we should remove `quotRem`, I think it's a natural thing to look for.

JacquesCarette:
Looks great to me now!

MatthewDaggitt:
Thanks for the PR @Taneb!

PR #1426, opened 2021-02-17
Remove dependy on Data.Nat.Tactic.RingSolver in Data.Nat.DivMod

Taneb:
I've proven the one lemma that we were using the ring solver for by hand.


MatthewDaggitt:
Thanks looks awesome!

PR #1428, opened 2021-02-17
Fix proof name with | in Data.Nat.Properties

Taneb:


Taneb:
What's the convention with CHANGELOGing this kind of change?

MatthewDaggitt:
I'd add a deprecation warning under the `Deprecated names` section. In this case as the change is visually indistinguishable you probably want to explicitly mention that it's the last character that has changed from `|` to `\|` in both the CHANGELOG entry and the deprecation message. That way we don't baffle users to much...

MatthewDaggitt:
Thank you for cleaning up after me by the way :+1:

Taneb:
I've added a note and updated the changelog. While doing so, I noticed that another identifier, recently added, had a similar issue, so I've also renamed that.

PR #1435, opened 2021-02-23
[ refactor ] Clean up the dependency graph

gallais:
Following https://github.com/agda/agda/issues/5231#issuecomment-783772927
I investigated using Agda's `--dependency-graph` option and managed to
reorganise the dependency graph quite a bit.

We now only need to compile 34 library modules (of which 12 are Builtin
modules!) to run hello world.

### DIFF
--- a/src/Codata/Musical/Colist/Base.agda
+++ b/src/Codata/Musical/Colist/Base.agda
@@ -1,0 +1,66 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Coinductive lists: base type and functions
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --sized-types --guardedness #-}
+
+module Codata.Musical.Colist.Base where
+
+open import Codata.Musical.Notation
+open import Codata.Musical.Conat.Base using (Coℕ; zero; suc)
+open import Data.Bool.Base using (Bool; true; false)
+open import Data.List.Base using (List; []; _∷_)
+open import Data.Maybe.Base using (Maybe; nothing; just)
+open import Data.Nat.Base using (ℕ; zero; suc)
+
+------------------------------------------------------------------------
+-- The type
+
+infixr 5 _∷_
+
+data Colist {a} (A : Set a) : Set a where
+  []  : Colist A
+  _∷_ : (x : A) (xs : ∞ (Colist A)) → Colist A
+
+{-# FOREIGN GHC
+  data AgdaColist a    = Nil | Cons a (MAlonzo.RTE.Inf (AgdaColist a))
+  type AgdaColist' l a = AgdaColist a
+  #-}
+{-# COMPILE GHC Colist = data AgdaColist' (Nil | Cons) #-}
+{-# COMPILE UHC Colist = data __LIST__ (__NIL__ | __CONS__) #-}
+
+------------------------------------------------------------------------
+-- Some operations
+
+null : ∀ {a} {A : Set a} → Colist A → Bool
+null []      = true
+null (_ ∷ _) = false
+
+length : ∀ {a} {A : Set a} → Colist A → Coℕ
+length []       = zero
+length (x ∷ xs) = suc (♯ length (♭ xs))
+
+map : ∀ {a b} {A : Set a} {B : Set b} → (A → B) → Colist A → Colist B
+map f []       = []
+map f (x ∷ xs) = f x ∷ ♯ map f (♭ xs)
+
+fromList : ∀ {a} {A : Set a} → List A → Colist A
+fromList []       = []
+fromList (x ∷ xs) = x ∷ ♯ fromList xs
+
+replicate : ∀ {a} {A : Set a} → Coℕ → A → Colist A
+replicate zero    x = []
+replicate (suc n) x = x ∷ ♯ replicate (♭ n) x
+
+lookup : ∀ {a} {A : Set a} → ℕ → Colist A → Maybe A
+lookup n       []       = nothing
+lookup zero    (x ∷ xs) = just x
+lookup (suc n) (x ∷ xs) = lookup n (♭ xs)
+
+infixr 5 _++_
+
+_++_ : ∀ {a} {A : Set a} → Colist A → Colist A → Colist A
+[]       ++ ys = ys
+(x ∷ xs) ++ ys = x ∷ ♯ (♭ xs ++ ys)
--- a/src/Codata/Musical/Conat/Base.agda
+++ b/src/Codata/Musical/Conat/Base.agda
@@ -1,0 +1,40 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Coinductive "natural" numbers: base type and operations
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --without-K --guardedness #-}
+
+module Codata.Musical.Conat.Base where
+
+open import Codata.Musical.Notation
+open import Data.Nat.Base using (ℕ; zero; suc)
+open import Function.Base using (_∋_)
+
+------------------------------------------------------------------------
+-- The type
+
+data Coℕ : Set where
+  zero : Coℕ
+  suc  : (n : ∞ Coℕ) → Coℕ
+
+------------------------------------------------------------------------
+-- Some operations
+
+pred : Coℕ → Coℕ
+pred zero    = zero
+pred (suc n) = ♭ n
+
+fromℕ : ℕ → Coℕ
+fromℕ zero    = zero
+fromℕ (suc n) = suc (♯ fromℕ n)
+
+∞ℕ : Coℕ
+∞ℕ = suc (♯ ∞ℕ)
+
+infixl 6 _+_
+
+_+_ : Coℕ → Coℕ → Coℕ
+zero  + n = n
+suc m + n = suc (♯ (♭ m + n))
--- a/src/Data/Bool/Base.agda
+++ b/src/Data/Bool/Base.agda
@@ -11,6 +11,7 @@ module Data.Bool.Base where
 open import Data.Unit.Base using (⊤)
 open import Data.Empty
 open import Level using (Level)
+open import Relation.Nullary using (Dec; does; proof; ofʸ; ofⁿ)
 
 private
   variable
@@ -72,3 +73,7 @@ T : Bool → Set
 T true  = ⊤
 T false = ⊥
 
+T? : ∀ b → Dec (T b)
+does  (T? b) = b
+proof (T? true ) = ofʸ _
+proof (T? false) = ofⁿ λ()
--- a/src/Data/Char/Base.agda
+++ b/src/Data/Char/Base.agda
@@ -10,9 +10,10 @@ module Data.Char.Base where
 
 open import Level using (zero)
 import Data.Nat.Base as ℕ
-open import Function
-open import Relation.Binary using (Rel)
-open import Relation.Binary.PropositionalEquality
+open import Data.Bool.Base using (Bool)
+open import Function.Base using (_on_)
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.PropositionalEquality.Core
 open import Relation.Binary.Construct.Closure.Reflexive
 
 ------------------------------------------------------------------------
@@ -46,6 +47,10 @@ _≈_ = _≡_ on toℕ
 
 _≉_ : Rel Char zero
 _≉_ = _≢_ on toℕ
+
+infix 4 _≈ᵇ_
+_≈ᵇ_ : (c d : Char) → Bool
+c ≈ᵇ d = toℕ c ℕ.≡ᵇ toℕ d
 
 infix 4 _<_
 _<_ : Rel Char zero
--- a/src/Data/List/Relation/Binary/Lex.agda
+++ b/src/Data/List/Relation/Binary/Lex.agda
@@ -1,0 +1,92 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Lexicographic ordering of lists
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Relation.Binary.Lex where
+
+open import Data.Empty using (⊥; ⊥-elim)
+open import Data.Unit.Base using (⊤; tt)
+open import Data.Product using (_×_; _,_; proj₁; proj₂; uncurry)
+open import Data.List.Base using (List; []; _∷_)
+open import Data.Sum.Base using (_⊎_; inj₁; inj₂; [_,_])
+open import Function.Base using (_∘_; flip; id)
+open import Function.Equivalence using (_⇔_; equivalence)
+open import Level using (_⊔_)
+open import Relation.Nullary using (Dec; yes; no; ¬_)
+import Relation.Nullary.Decidable as Dec
+open import Relation.Nullary.Product using (_×-dec_)
+open import Relation.Nullary.Sum using (_⊎-dec_)
+open import Relation.Binary hiding (_⇔_)
+open import Data.List.Relation.Binary.Pointwise.Core
+   using (Pointwise; []; _∷_; head; tail)
+
+------------------------------------------------------------------------
+-- Re-exporting the core definitions and properties
+
+open import Data.List.Relation.Binary.Lex.Core public
+
+------------------------------------------------------------------------
+
+-- Properties
+
+module _ {a ℓ₁ ℓ₂} {A : Set a} {P : Set}
+         {_≈_ : Rel A ℓ₁} {_≺_ : Rel A ℓ₂} where
+
+  private
+    _≋_ = Pointwise _≈_
+    _<_ = Lex P _≈_ _≺_
+
+  transitive : IsEquivalence _≈_ → _≺_ Respects₂ _≈_ → Transitive _≺_ →
+               Transitive _<_
+  transitive eq resp tr = trans
+    where
+    trans : Transitive (Lex P _≈_ _≺_)
+    trans (base p)         (base _)         = base p
+    trans (base y)         halt             = halt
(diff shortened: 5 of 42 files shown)


jmchapman:
Nice work!

It would be nice to get rid of the musical codata stuff from IO too and cut down the builtin modules. Correct me if I am wrong but I think supporting finite strings only would cover quite a lot of use cases.

This is roughly all the code that's actually needed to run 'hello world' after all:
```
% cat Hello.agda           
module Hello where

open import Agda.Builtin.IO
open import Agda.Builtin.Unit
open import Agda.Builtin.String

postulate putStrLn : String → IO ⊤
{-# FOREIGN GHC import qualified Data.Text.IO as T #-}
{-# COMPILE GHC putStrLn = T.putStrLn #-}

main : IO ⊤
main = putStrLn "hello world"

% agda --compile Hello.agda
Compiling Agda.Primitive in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Primitive.agdai to MAlonzo/Code/Agda/Primitive.hs
Compiling Agda.Builtin.Unit in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/Unit.agdai to MAlonzo/Code/Agda/Builtin/Unit.hs
Compiling Agda.Builtin.IO in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/IO.agdai to MAlonzo/Code/Agda/Builtin/IO.hs
Compiling Agda.Builtin.Bool in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/Bool.agdai to MAlonzo/Code/Agda/Builtin/Bool.hs
Compiling Agda.Builtin.Nat in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/Nat.agdai to MAlonzo/Code/Agda/Builtin/Nat.hs
Compiling Agda.Builtin.Char in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/Char.agdai to MAlonzo/Code/Agda/Builtin/Char.hs
Compiling Agda.Builtin.List in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/List.agdai to MAlonzo/Code/Agda/Builtin/List.hs
Compiling Agda.Builtin.String in /nix/store/6rd59zf2v561w8rfshjnkpgpwrv2ic4g-Agda-2.6.1.2-data/share/ghc-8.10.2/x86_64-osx-ghc-8.10.2/Agda-2.6.1.2/lib/prim/Agda/Builtin/String.agdai to MAlonzo/Code/Agda/Builtin/String.hs
Compiling Hello in Hello.agdai to MAlonzo/Code/Hello.hs
Calling: /nix/store/drw76lkmh4l8mm8q1navr7r8bx4ygbvw-ghc-8.10.2-with-packages/bin/ghc -O -o Hello -Werror -i. -main-is MAlonzo.Code.Hello MAlonzo/Code/Hello.hs --make -fwarn-incomplete-patterns -fno-warn-overlapping-patterns
[1 of 5] Compiling MAlonzo.RTE      ( MAlonzo/RTE.hs, MAlonzo/RTE.o )
[2 of 5] Compiling MAlonzo.Code.Agda.Builtin.Unit ( MAlonzo/Code/Agda/Builtin/Unit.hs, MAlonzo/Code/Agda/Builtin/Unit.o )
[3 of 5] Compiling MAlonzo.Code.Agda.Builtin.String ( MAlonzo/Code/Agda/Builtin/String.hs, MAlonzo/Code/Agda/Builtin/String.o )
[4 of 5] Compiling MAlonzo.Code.Agda.Builtin.IO ( MAlonzo/Code/Agda/Builtin/IO.hs, MAlonzo/Code/Agda/Builtin/IO.o )
[5 of 5] Compiling MAlonzo.Code.Hello ( MAlonzo/Code/Hello.hs, MAlonzo/Code/Hello.o )
Linking Hello ...
% ./Hello                  
hello world
```

src/Data/Bool/Base.agda
```diff
@@ -72,3 +73,7 @@ T : Bool → Set
 T true  = ⊤
 T false = ⊥
 
+T? : ∀ b → Dec (T b)
```
JacquesCarette, line 76: This move represents a non-trivial design decision.  Is `T?` a property or not?  If it's not, then possibly other such moves may be allowed. More likely, it **is** a property, but somehow a very 'basic' one.  So maybe `Data.Bool.Properties.Base` needs to exist for these.

JacquesCarette (commented):
This is really nice. Hopefully we can get better tools to continue analyzing the dependency graph and keep cleaning it up like this. Once this term-from-hell is over, I'm hoping to dig back in to such things.

src/Data/Bool/Base.agda
jmchapman: I use `blah?` things quite often in programs (i.e., not proofs). E.g., to check if an integer equals `0` before dividing by it. I see that the instances of `=?`  for `Nat` and `Integer` are in their respective properties files. But, I think you could make a case for moving them as they have frequently used computational context, they are decision procedures. Especially if it turns out to be useful to do so.

JacquesCarette: I can definitely warm up to that argument.

MatthewDaggitt: Hmm so this story is a little more complicated. 

Originally it use to be the case that the decision procedures lived in the `Base` files rather than the `Property` files. The problem was that most decision procedures depend on properties, e.g. the decision procedure for the equality of natural numbers depends on the property that the `suc` constructor is injective, the decision procedure for list permutations depends on the decidedly non-trivial property that if `x :: xs` is a permutation of `ys` then you can extract `x` from `ys` and the result is a permutation of `xs`. All decision procedures were therefore moved to the `Properties` files.

However about a year ago, @laMudri tweaked our definition of `Dec` such that it has both a boolean component and a proof component. @jmchapman when you say that you are using decision procedures in programs rather than proofs, I take it that you're really using the boolean component? I think there's a strong argument that the boolean component of the decision procedure to be in the `Base` file and the proof component to be in the `Properties` file. Indeed we already organise things this way in the numeric classes where we have things like `_≡ᵇ_` and `_<ᵇ_` defined in the `Base` file...

In this case, the proof component of `T?` is trivial so it looks like it can be moved to the `Base` file. However for consistency I'd still think we should have the decision procedure in the `Properties` file. At some point in the next couple of days I'll have a look at how to achieve this...

gallais: I can re-export the `T?` in the properties file.

The motivation for `T?`'s special status is that `filter` & friends use a decidability
result instead of a boolean function. The only way to recover the `Bool`-based list
functions is to post-compose the filtering boolean predicate with `T?`. Being forced to
pull in all of `Algebra` & `Algebra.Structures` just for that seems a tad heavy.

gallais:
> Hopefully we can get better tools to continue analyzing the dependency graph
  and keep cleaning it up like this. 
  
Yep. I had to start from `Data.Char.Base` because the full `IO` dependency graph
was too big for `dot` to handle. It would be nice to have a tool spitting out some
of the graph's properties (a collection of the longest paths for instance) rather
than computing a graphical representation of it!

src/Data/Bool/Base.agda
MatthewDaggitt: In that case, how about we put `T?` in a new file `Data.Bool.Properties.Base` (ala `Data.Nat.Properties.Core`) rather than in `Data.Bool.Base`?

CHANGELOG.md
```diff
@@ -12,6 +12,16 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
```
MatthewDaggitt, line 13: I think it's possibly worth adding a CHANGELOG entry mentioning that the dependencies of certain key modules have been significantly reduced? We should at least get some credit for all the hard work we do under the surface :smile:

src/Codata/Musical/Colist/Base.agda
```diff
@@ -0,0 +1,66 @@
...
+module Codata.Musical.Colist.Base where
+
+open import Codata.Musical.Notation
+open import Codata.Musical.Conat.Base using (Coℕ; zero; suc)
+open import Data.Bool.Base using (Bool; true; false)
+open import Data.List.Base using (List; []; _∷_)
+open import Data.Maybe.Base using (Maybe; nothing; just)
+open import Data.Nat.Base using (ℕ; zero; suc)
+
+------------------------------------------------------------------------
```
MatthewDaggitt, line 18: As this is a new module, it would greatly benefit from a couple of variables for `a`, `A` and `B`?

src/Codata/Musical/Conat/Base.agda
```diff
@@ -0,0 +1,40 @@
...
+
+pred : Coℕ → Coℕ
+pred zero    = zero
+pred (suc n) = ♭ n
+
+fromℕ : ℕ → Coℕ
+fromℕ zero    = zero
+fromℕ (suc n) = suc (♯ fromℕ n)
+
+∞ℕ : Coℕ
```
MatthewDaggitt, line 33: Very nitpicky but strictly speaking this isn't an operation. Can it go under it's own `Constants` section?

src/Data/Char/Base.agda
```diff
@@ -47,6 +48,10 @@ _≈_ = _≡_ on toℕ
 _≉_ : Rel Char zero
 _≉_ = _≢_ on toℕ
 
+infix 4 _≈ᵇ_
+_≈ᵇ_ : (c d : Char) → Bool
```
MatthewDaggitt, line 52: Not mentioned in CHANGELOG?

src/Data/List/Relation/Binary/Lex.agda
```diff
@@ -0,0 +1,92 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Lexicographic ordering of lists
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Relation.Binary.Lex where
```
MatthewDaggitt, line 9: I have to confess I'm a bit confused by this reorganisation. To me it would make more sense to lift all properties up into this module, rather than just some of them? This would be more in keeping with other binary relations over lists where we have core modules just containing the definition?

You could then get rid of `Lex.Strict.Core` and move all four of the basic definitions into `Lex.Core`?

src/Data/List/Relation/Binary/Lex.agda
```diff
@@ -0,0 +1,92 @@
...
+open import Data.List.Relation.Binary.Pointwise.Core
+   using (Pointwise; []; _∷_; head; tail)
+
+------------------------------------------------------------------------
+-- Re-exporting the core definitions and properties
+
+open import Data.List.Relation.Binary.Lex.Core public
+
+------------------------------------------------------------------------
+
```
MatthewDaggitt, line 33: Extra empty line

src/Data/List/Relation/Binary/Lex/NonStrict.agda
```diff
@@ -23,10 +23,12 @@ open import Relation.Nullary
 open import Relation.Binary
 import Relation.Binary.Construct.NonStrictToStrict as Conv
 
+import Data.List.Relation.Binary.Lex as Core
+
 ------------------------------------------------------------------------
 -- Publically re-export definitions from Core
 
-open import Data.List.Relation.Binary.Lex.Core as Core public
+open import Data.List.Relation.Binary.Lex.Core public
```
MatthewDaggitt, line 31: Should this either be `open import Data.List.Relation.Binary.Lex.**NonStrict**.Core public` or `open Core public`?

src/Data/List/Relation/Binary/Pointwise/Core.agda
```diff
@@ -0,0 +1,98 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Pointwise lifting of relations to lists
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Relation.Binary.Pointwise.Core where
```
MatthewDaggitt, line 9: Should this be a `Base` module rather than a `Core` module?

src/Data/Bool/Base.agda
gallais: My bad: this was not actually needed!

src/Relation/Binary/Construct/Constant.agda
```diff
@@ -8,13 +8,19 @@
...
+    A : Set a
+    B : Set b
+
 ------------------------------------------------------------------------
--- Definition
+-- Re-export definition
 
-Const : ∀ {a b c} {A : Set a} {B : Set b} → Set c → REL A B c
-Const I = λ _ _ → I
+open import Relation.Binary.Construct.Constant.Core public
```
MatthewDaggitt, line 23: Can't quite spot why do we need a `Core` module for this?

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -0,0 +1,91 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Core properties related to negation
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Nullary.Negation.Core where
```
MatthewDaggitt, line 9: Ugh, the whole `Relation.Nullary` directory needs a serious rethink at some point... Now might not be the time though.

MatthewDaggitt (commented):
Looks great! A plethora of minor comments though...

src/Data/Bool/Base.agda
jmchapman: >  @jmchapman when you say that you are using decision procedures in programs rather than proofs, I take it that you're really using the boolean component?

Yes - for the case I was thinking of.

src/Relation/Binary/Construct/Constant.agda
gallais: `Setoid` induces a big dependency

src/Codata/Musical/Colist/Base.agda
gallais: I used this as a pretext to hack away at `Codata.Musical.Colist` until not much was left.

src/Data/List/Relation/Binary/Lex/NonStrict.agda
gallais: As it turns out, `Lex` defines both strict & non-strict lexicographic ordering.
The adjective in the module paths `Lex.Strict` and `Lex.NonStrict` refers to
whether the *lifted* order is one or the other. Basically only the `Strict` one
is of interest: the nonstrict one is using the non-strict-to-strict construction
to reuse the other.

This is probably worth revamping in the future.

gallais:
Latest state of affairs: compiling `IO.Finite` pulls only 11 modules.
io

Of course you'll have to write you hello world using `Agda.Builtin.String` to keep
such a small dependency graph. As you as you pull `Data.String.Base`, you'll get
the `Base` for `Char`, `List`, `Maybe`, `These`, etc.

jmchapman:
`IO.Finite` ❤️

graph.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
MatthewDaggitt, line 1: Looks awesome! Can you add a brief comment describing what it does and how to use it?

CHANGELOG.md
```diff
@@ -65,9 +75,25 @@ New modules
   Data.Nat.PseudoRandom.LCG
   ```
 
+* Broke up `Codata.Musical.Colist` into a multitude of modules:
```
MatthewDaggitt, line 78: Missing entries for the new IO modules?

CHANGELOG.md
```diff
@@ -116,6 +130,30 @@ New modules
...
+  ```
+  Data.List.Relation.Binary.Pointwise.Base
+  Data.List.Relation.Binary.Pointwise.Properties
+  ```
+
+* Broke up `Codata.Musical.Colist` into a multitude of modules:
+  ```
+  Codata.Musical.Colist.Base
+  Codata.Musical.Colist.Properties
+  Codata.Musical.Colist.Bisimilarity
```
MatthewDaggitt, line 143: I know the `Musical` folder is morally deprecated, but should this be `Relation.Binary.Bisimilarity.Propositional`?

gallais: I followed the structure we have in the sized `Codata` directories
where `Bisimilarity` is put front and center. Should we deprecate
these modules and put it all under `Codata.(...).Relation.Binary`?

MatthewDaggitt: Ah yes. Hmm, probably I guess to be consistent. However, let's leave it for this PR. Already enough changes going on.

PR #1438, opened 2021-02-24
Update Properties.agda

negiizhao:
Add some properties for rationals.

### DIFF
--- a/src/Data/Rational/Properties.agda
+++ b/src/Data/Rational/Properties.agda
@@ -28,7 +28,7 @@ import Data.Nat.DivMod as ℕ
 open import Data.Product using (_×_; _,_)
 open import Data.Rational.Base
 open import Data.Rational.Unnormalised.Base as ℚᵘ
-  using (ℚᵘ; *≡*; *≤*) renaming (↥_ to ↥ᵘ_; ↧_ to ↧ᵘ_; _≃_ to _≃ᵘ_; _≤_ to _≤ᵘ_)
+  using (ℚᵘ; *≡*; *≤*; *<*) renaming (↥_ to ↥ᵘ_; ↧_ to ↧ᵘ_; _≃_ to _≃ᵘ_; _≤_ to _≤ᵘ_; _<_ to _<ᵘ_)
 import Data.Rational.Unnormalised.Properties as ℚᵘ
 open import Data.Sum.Base
 open import Data.Unit using (tt)
@@ -54,6 +54,16 @@ private
   infix 4 _≢0
   _≢0 : ℕ → Set
   n ≢0 = False (n ℕ.≟ 0)
+
+------------------------------------------------------------------------
+-- Properties of Positive/Negative/NonPositive/NonNegative predicates
+------------------------------------------------------------------------
+
+positive⇒nonNegative : ∀ {q} → Positive q → NonNegative q
+positive⇒nonNegative {q} = ℚᵘ.positive⇒nonNegative {toℚᵘ q}
+
+negative⇒nonPositive : ∀ {q} → Negative q → NonPositive q
+negative⇒nonPositive {q} = ℚᵘ.negative⇒nonPositive {toℚᵘ q}
 
 ------------------------------------------------------------------------
 -- Propositional equality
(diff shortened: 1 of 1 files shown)


src/Data/Rational/Properties.agda
```diff
@@ -55,6 +55,16 @@ private
   _≢0 : ℕ → Set
   n ≢0 = False (n ℕ.≟ 0)
 
+------------------------------------------------------------------------
+-- Properties of Positive/Negative/NonPositive/NonNegative predicates
+------------------------------------------------------------------------
+
+positive⇒nonNegative : ∀ {q} → Positive q → NonNegative q
```
MatthewDaggitt, line 62: So for rationals, the underlying number cannot be inferred just from `Positive/Negative etc`. This means that you need to make the `q` argument explicit rather than implicit in many places. I realise we haven't been great about enforcing that so far in this file, but better to start now!

src/Data/Rational/Properties.agda
```diff
@@ -686,6 +718,65 @@ private
...
+neg-distrib-+ p q = toℚᵘ-injective (begin-equality
+  toℚᵘ(- (p + q))                    ≈⟨ toℚᵘ-homo‿- (p + q) ⟩
+  ℚᵘ.- toℚᵘ(p + q)                   ≈⟨ ℚᵘ.-‿cong (toℚᵘ-homo-+ p q) ⟩
+  ℚᵘ.- (toℚᵘ(p) ℚᵘ.+ toℚᵘ(q))        ≡⟨ ℚᵘ.neg-distrib-+ (toℚᵘ p) (toℚᵘ q) ⟩
+  (ℚᵘ.- toℚᵘ(p)) ℚᵘ.+ (ℚᵘ.- toℚᵘ(q)) ≈⟨ ℚᵘ.≃-sym (ℚᵘ.+-cong (toℚᵘ-homo‿- p) (toℚᵘ-homo‿- q)) ⟩
+  toℚᵘ(- p) ℚᵘ.+ toℚᵘ(- q)           ≈⟨ ℚᵘ.≃-sym (toℚᵘ-homo-+ (- p) (- q)) ⟩
+  toℚᵘ((- p) + (- q))             ∎
+  )
+  where
+  open ℚᵘ.≤-Reasoning
```
MatthewDaggitt, line 734: According to the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md), the last four lines should be condensed to two lines as follows:
```
  toℚᵘ((- p) + (- q))                ∎)
  where open ℚᵘ.≤-Reasoning
```
and the `∎` should be aligned to the equational reasoning. There are other places in this PR where this applies as well!

src/Data/Rational/Properties.agda
```diff
@@ -880,6 +971,71 @@ private
   { isCommutativeRing = +-*-isCommutativeRing
   }
 
+------------------------------------------------------------------------
+-- Properties of _*_ and -_
+
+neg-distribˡ-* : ∀ p q → - (p * q) ≡ - p * q
```
MatthewDaggitt, line 977: It would be nice if these two proofs could be lifted to the [RingMonomorphism module](https://github.com/agda/agda-stdlib/blob/master/src/Algebra/Morphism/RingMonomorphism.agda) and then used from there by constructing the `+-*-ringMonomorphism`?

MatthewDaggitt (commented):
Thanks for the great PR @negiizhao! I've got a couple of minor comments but nothing major.

CHANGELOG.md
```diff
@@ -269,8 +269,31 @@ Other minor additions
   toℚᵘ-homo-1/ : toℚᵘ (1/ p) ℚᵘ.≃ ℚᵘ.1/ (toℚᵘ p)
   *-inverseˡ   : 1/ p * p ≡ 1ℚ
   *-inverseʳ   : p * 1/ p ≡ 1ℚ
+  
+  positive⇒nonNegative : ∀ {q} → Positive q → NonNegative q
```
MatthewDaggitt, line 273: In the CHANGELOG for brevity we usually get rid of the variables being quantified over in the type, e.g. `∀ {q} → Positive q → NonNegative q` becomes `Positive q → NonNegative q`

negiizhao:
When I try to fix it, I found more problems. So I made some changes about function names and implicit arguments, but I'm not sure if it's a breaking change.

MatthewDaggitt:
Hi @negiizhao, thanks for your enthusiasm, but some of those changes are a little scary! In particular, as you mention, renaming record fields like that will break other people's code. We try and do our best to avoid that except when absolutely necessary.

I think the best course of action is to revert your previous commit  (`git revert 9d7ad1f`) and then fix just the issues I raised with your new code. We can then open another, different PR to address e.g. the inconsistencies with implicit arguments. Does that sound like a plan?

negiizhao:
Thanks for the review. It has been reverted and fixed.

I'm new to source code developing. Sorry for that.

MatthewDaggitt:
> Thanks for the review. It has been reverted and fixed.

Thanks!

> I'm new to source code developing. Sorry for that.

Don't worry! It's a great first contribution to the library, thank you very much. I totally understand the temptation of trying to fix everything in one go :smile: 

I'll leave the PR open for another couple of days in case anyone else has any comments and then I'll merge it in.

PR #1444, opened 2021-03-04
Add lexicographic product over algebras

MatthewDaggitt:


src/Algebra/Construct/LexProduct/Inner.agda
```diff
@@ -0,0 +1,268 @@
...
+import Relation.Binary.Reasoning.Setoid as SetoidReasoning
+
+import Algebra.Construct.LexProduct.Base as Base
+
+module Algebra.Construct.LexProduct.Inner
+  {ℓ₁ ℓ₂ ℓ₃ ℓ₄} (M₁ : Magma ℓ₁ ℓ₂) (M₂ : Magma ℓ₃ ℓ₄)
+  (_≟₁_ : Decidable (Magma._≈_ M₁))
+  where
+
+open Magma M₁
```
JacquesCarette, line 27: Rather than these long renamings, which are a bit difficult to remember, have you considered using local modules to qualify all the names? I find that using `M` and `N`, instead of subscripts, then is typographically nicer.

MatthewDaggitt: In 90% of cases I agree and much prefer using local modules! The problem with this case is that there is a significant amount of equational reasoning in these files. Qualifying the operators and the equality relations makes this reasoning significantly less readable in my opinion.

I've therefore gone for a half-way house, I've used local modules for all the properties (e.g. `M.refl`) but kept the renamings for the operators and relations. This hopefully gets the best of both worlds by keeping the reasoning readable, but cutting in half the amount of renaming necessary!

JacquesCarette: This does seem like an excellent compromise.

PR #1447, opened 2021-03-06
readjust reasoning APIs

### LINKED ISSUE #1446: a little annoyance in PER reasoning
At this moment, https://github.com/agda/agda-stdlib/blob/master/src/Relation/Binary/Reasoning/PartialSetoid.agda uses 
https://github.com/agda/agda-stdlib/blob/911b52d54d39939a78ba63a387e30cc90ad64163/src/Relation/Binary/Reasoning/Base/Partial.agda#L75-L76

as the terminator for reasoning. This is rather annoying, because consider a proof using transitivity in PER:
```
trans p1 p2
```
That's already a valid proof. But if we use reasoning API, we will get
```
begin
  foo ≈⟨ p1 ⟩
  bar ≈⟨ p2 ⟩
  baz ∎⟨ ??? ⟩
```
All of a sudden, we are obliged to provide another proof which is not necessary in the original proof. Are we able to get rid of this part of the API?

That actually points out a fact that all proofs using this reasoning API requires an unnecessary proof at the very end, which could be a redundant `refl` proof for `Setoid`.

HuStmpHrrr:
Closes #1446 

one nice thing about this modification is that no change to use sites is required.

### DIFF
--- a/src/Relation/Binary/Reasoning/Base/Single.agda
+++ b/src/Relation/Binary/Reasoning/Base/Single.agda
@@ -13,13 +13,72 @@ module Relation.Binary.Reasoning.Base.Single
   (refl : Reflexive _∼_) (trans : Transitive _∼_)
   where
 
+-- TODO: the following part is copied from Relation.Binary.Reasoning.Base.Partial
+-- in order to avoid larger refactors. We will refactor this part later
+-- so taht we use the same framework as Relation.Binary.Reasoning.Base.Partial.
+
+open import Level using (_⊔_)
+open import Relation.Binary.PropositionalEquality.Core as P
+  using (_≡_)
+
+infix  4 _IsRelatedTo_
+infixr 2 step-∼ step-≡ step-≡˘
+infixr 2 _≡⟨⟩_
+infix  1 begin_
+
+------------------------------------------------------------------------
+-- Definition of "related to"
+
+-- This seemingly unnecessary type is used to make it possible to
+-- infer arguments even if the underlying equality evaluates.
+
+data _IsRelatedTo_ (x y : A) : Set ℓ where
+  relTo : (x∼y : x ∼ y) → x IsRelatedTo y
+
 ------------------------------------------------------------------------
 -- Reasoning combinators
 
--- Re-export combinators from partial reasoning
+-- Note that the arguments to the `step`s are not provided in their
+-- "natural" order and syntax declarations are later used to re-order
+-- them. This is because the `step` ordering allows the type-checker to
+-- better infer the middle argument `y` from the `_IsRelatedTo_`
+-- argument (see issue 622).
+--
+-- This has two practical benefits. First it speeds up type-checking by
+-- approximately a factor of 5. Secondly it allows the combinators to be
+-- used with macros that use reflection, e.g. `Tactic.RingSolver`, where
+-- they need to be able to extract `y` using reflection.
 
-open import Relation.Binary.Reasoning.Base.Partial _∼_ trans public
-  hiding (_∎⟨_⟩)
+-- Beginning of a proof
+
+begin_ : ∀ {x y} → x IsRelatedTo y → x ∼ y
+begin relTo x∼y = x∼y
+
+-- Standard step with the relation
+
+step-∼ : ∀ x {y z} → y IsRelatedTo z → x ∼ y → x IsRelatedTo z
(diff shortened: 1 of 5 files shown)


MatthewDaggitt:
Hmm this has broader reaching consequences than I'd imagined. In particular I'm a bit scared of how the changes propagate out into permutation reasoning. This means that when someone has defined their own reasoning combinators based on `Relation.Binary.Reasoning.Base.Single` then its going to get broken by this change.

Given that the whole point of `Reasoning.Base.Partial` is to work around the case where we don't have reflexivity then I don't think it should affect `Reasoning.Base.Single` where we do have reflexivity. I think the best approach is to decouple the two different types of reasoning entirely, i.e. copy the old content of `Reasoning.Base.Partial` into `Reasoning.Base.Single` verbatim.

HuStmpHrrr:
this update should have less impact than the previous one.

src/Relation/Binary/Reasoning/Base/Single.agda
```diff
@@ -13,17 +13,59 @@ module Relation.Binary.Reasoning.Base.Single
...
+
 ------------------------------------------------------------------------
 -- Reasoning combinators
 
 -- Re-export combinators from partial reasoning
 
-open import Relation.Binary.Reasoning.Base.Partial _∼_ trans public
-  hiding (_∎⟨_⟩)
+data _IsRelatedTo_ : A → A → Set (a ⊔ ℓ) where
+  emp   : ∀ x → x IsRelatedTo x
```
MatthewDaggitt, line 32: Hmm sorry just re-read your original issue and I missed the bit about changing setoid reasoning as well. I agree that in the case of partial setoids, having to provide the unnecessary refl proof was problematic as a) extra typing for the user and b) the relation may not even be reflexive. 

However in the single/setoid case, we do have access to the `refl` proof and therefore the user never needs to explicitly provide it. Even if using it is technically unnecessary, is there a compelling reason for why we should break backwards compatibility to avoid using it?

Even if there is, I would be more tempted to put off changing it until v2.0 and then change all the reasoning APIs at once. After all the sample problem happens with poset/preorder reasoning right? Changing those may be more complicated...

HuStmpHrrr: does it actually break anything if we omit an unnecessary proof? the fact that people doesn't complain about an extra refl proof seems to indicate that usually people just don't care.

we can then postpone this PR to 2.0 if you find it better this way. any workaround would complicate the PR and introduce later technical debt to pay back.

MatthewDaggitt: > does it actually break anything if we omit an unnecessary proof? 

Well adding the extra `emp` constructor "unnecessarily" breaks the definition of any reasoning API that a user has built on top of `Relation.Binary.Reasoning.Base.Single` right? For example in the standard library it breaks `Reasoning.MultiSetoid` and `PermutationReasoning`. Note, it's not breaking the use of such reasoning APIs, only their definition.

> we can then postpone this PR to 2.0 if you find it better this way. any workaround would complicate the PR and introduce later technical debt to pay back.

Well ideally I'd like to keep the changes to `Relation.Binary.Reasoning.Base.Partial` and merge them in v1.6 as there is a clear advantage to incorporating them now as it directly simplifies the user-facing API. 

In contrast any changes to eliminate the redundant reflexive cases in `Relation.Binary.Reasoning.Base.(Single/Double/Triple)` do not simplify the user-facing API and therefore if we are to make them, they should be made consistently across all three of those modules and we should wait until v2.0 to do it? Does that make sense?

HuStmpHrrr:
sorry to be late. I refactored the API again. this one copies the old `Partial` API so that there is no impact on `Single` and subsequent `Setoid` based reasoning, but the reasoning API for `Partial` will be improved.

MatthewDaggitt:
> I refactored the API again. this one copies the old Partial API so that there is no impact on Single and subsequent Setoid based reasoning, but the reasoning API for Partial will be improved.

Awesome, thank you! I've made a minor modification so that the code uses the same multi-constructor pattern as used in the other `Reasoning.Base` modules which eliminates the need to pass the "Flag" object as an argument everywhere.

Once it passes the tests, I'll merge it in!

MatthewDaggitt:
Awesome, thanks for the PR!

PR #1448, opened 2021-03-06
[ new ] more interleaving functions in Codata.Stream

gallais:
Came across this cute `cantor` function while formalising Liam's paper in
Idris so thought I'd see how it fairs with sized codata.

Also threw in `interleave⁺` which is very easy 	to define with sized types :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -829,6 +829,8 @@ Other minor additions
 
 * Added new functions to `Codata.Stream`:
   ```agda
+  nats : Stream ℕ ∞
+
   interleave⁺ : List⁺ (Stream A i) → Stream A i
   cantor      : Stream (Stream A ∞) ∞ → Stream A ∞
   plane       : Stream A ∞ → ((a : A) → Stream (B a) ∞) → Stream (Σ A B) ∞
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -200,7 +200,7 @@ Other minor additions
 
 * Added new proofs to `Data.List.Relation.Binary.Pointwise`:
   ```agda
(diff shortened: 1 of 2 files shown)


src/Codata/Stream.agda
```diff
@@ -13,89 +13,159 @@ open import Codata.Thunk as Thunk using (Thunk; force)
...
+drop : ℕ → Stream A ∞ → Stream A ∞
+drop n xs = proj₂ (splitAt n xs)
 
-module _ {a} {A : Set a} where
+chunksOf : (n : ℕ) → Stream A ∞ → Stream (Vec A n) ∞
+chunksOf n = chunksOfAcc n id module ChunksOf where
 
-  iterate : (A → A) → A → Stream A ∞
-  iterate f = unfold < f , id >
+  chunksOfAcc : ∀ {i} k (acc : Vec A k → Vec A n) →
```
MatthewDaggitt, line 115: Can this `i` be left to be a variable?

CHANGELOG.md
```diff
@@ -213,7 +213,7 @@ Other minor additions
   ```agda
   *-pres-∣ : o ∣ m → p ∣ n → o * p ∣ m * n
   ```
-  
+
```
MatthewDaggitt, line 216: Why is your `fix-whitespace` picking up whitespace in the CHANGELOG when it's supposed to be ignored?

https://github.com/agda/agda-stdlib/blob/911b52d54d39939a78ba63a387e30cc90ad64163/fix-whitespace.yaml#L12

MatthewDaggitt:
Otherwise, looks very cool!

CHANGELOG.md
gallais: My emacs is configured to automatically fix whitespace issues. :x

PR #1458, opened 2021-03-29
[ fixed #1452 ] more universe-polymorphic combinators for indexed containers

### LINKED ISSUE #1452: Union of indexed containers not maximally universe-polymorphic
The union operation requires two containers of the same universe levels:
https://github.com/agda/agda-stdlib/blob/48ee0073e1fa0cda2327f00a3ec37068859a6bec/src/Data/Container/Indexed/Combinator.agda#L109-L110
However, the same code would work for containers of different levels.
I have an application where I need different levels.

TODO:
- check for similar restrictions other container operations

RFC.

andreasabel:
[ fixed #1452 ] more universe-polymorphic combinators for indexed containers

Also: 
- simplify some type signatures via `variable`
  NB: I thought I could simplify further, but quantifications like `{I : Set i}` cannot be removed for some reason (!?)

- use copattern notation where appropriate

Discussion point: should the new, more general disjoint sum of indexed containers replace \uplus (with the old, more specialized version being \uplus' then), or should for the sake of backwards compatibility, the old stay \uplus and the new one become \uplus'?
For now, I modeled this after `Data.Sum` where the more special version of `[_,_]` has the prime.


src/Data/Sum/Relation/Unary/All.agda
```diff
@@ -0,0 +1,32 @@
...
+open import Data.Sum       using (_⊎_); open _⊎_
+open import Level          using (Level; _⊔_)
+open import Relation.Unary using (Pred)
+
+private
+  variable
+    a b c p q : Level
+    A B : Set _
+
+data All₂ {A : Set a} {B : Set b} (P : Pred A p) (Q : Pred B q)
```
MatthewDaggitt, line 20: I think this should just be called `All` to be consistent with everywhere else?

src/Data/Sum/Relation/Unary/All.agda
```diff
@@ -0,0 +1,32 @@
...
+       : Pred (A ⊎ B) (a ⊔ b ⊔ p ⊔ q) where
+  inj₁ : ∀ {a} → P a → All₂ P Q (inj₁ a)
+  inj₂ : ∀ {b} → Q b → All₂ P Q (inj₂ b)
+
+-- Elimination
+
+[_,_] : ∀ {P : Pred A p} {Q : Pred B q} {C : (x : A ⊎ B) → All₂ P Q x → Set c} →
```
MatthewDaggitt, line 27: Minor thing, but FYI our new variable policy, outlined in the style guide, allows one to use variables over `P` and `Q` in this file as well.

MatthewDaggitt (commented):
> Discussion point: should the new, more general disjoint sum of indexed containers replace \uplus (with the old, more specialized version being \uplus' then), or should for the sake of backwards compatibility, the old stay \uplus and the new one become \uplus'?
For now, I modeled this after Data.Sum where the more special version of [_,_] has the prime.

I think in the past, we've viewed universe-level generalisation of a function as sufficient grounds to break backwards compatibility. I think your current approach is the right one :+1:

MatthewDaggitt:
Oh and a CHANGELOG entry describing the breaking change would be great.

MatthewDaggitt:
> NB: I thought I could simplify further, but quantifications like {I : Set i} cannot be removed for some reason (!?)

Weird, seems to work for me on Agda 2.6.1. Maybe it's a problem with the development version of Agda? Anyway, I've updated those sets to use variables as wells.

andreasabel:
@MatthewDaggitt : Thanks a lot for finishing my PR!! (And sorry for not getting back in time.)

PR #1459, opened 2021-03-30
adds keep and drop to subset

carlostome:


carlostome:
Sorry I didn't follow up, wouldn't have known where to put it anyway. Thanks, @MatthewDaggitt !

PR #1462, opened 2021-04-02
Add Algebra.Properties.Semiring.Sum showing mulitplication distributes over sums

Taneb:


Taneb:
I haven't added this to Everything.agda yet. Is the "correct" way to do that with GenerateEverything.hs, or can I do it manually?

MatthewDaggitt:
I think we should add `-*` prefix onto the end of these names? Otherwise looks great!

Taneb:
> I think we should add `-*` prefix onto the end of these names? Otherwise looks great!

I ended up with `*-distrib?-sum` which feels the most consistent with existing names to me

MatthewDaggitt:
Yup, looks good!

Issue #921, opened 2019-09-30
`_/_` and `_%_` with irrelevant instance arguments


MatthewDaggitt:
Given #862 and the general policy to avoid using currently unstable features of Agda I know this isn't going to make it in anytime soon, but I thought I'd at least document for some point in the future that there are far superior definitions of `_/_` and `_%_` using instance arguments:

```
_/_ : (m n : ℕ) → .{{NonZero n}} → ℕ
m / suc n = div-helper 0 n m n
```

Current we have to write horrible things like this:
```
∀ m n {gcd≢0 : NonZero (gcd m n)} → Coprime ((m / gcd m n) {gcd≢0}) ((n / gcd m n) {gcd≢0})
```
but with the above the definition this becomes:
```
∀ m n {{NonZero (gcd m n)}} → Coprime (m / gcd m n) (n / gcd m n)
```

JacquesCarette:
Clear, concrete and useful examples such as the above will hopefully drive features forward.

I wonder what the above would look like in cubical agda?

mechvel:
As to me, it is better to write things plainly, with a proof argument and with proving irrelevance.
For example, I have
```
record EuclideanDomain α α≈ :  Set (sucL (α ⊔ α≈))                                
  where                                                                           
   field                                                                           
    divRem       : C → (y : C) → y ≉ 0# → C × C     -- Euclidean division: (quotient , remainder)          
    divRemEq< : (a b : C) → (nzb : b ≉ 0#) →                                      
                            let (q , r) = divRem a b nzb                          
                            in                                                    
                            a ≈ (q * b) + r  ×   ((nzr : r ≉ 0#) → ∣ r ,, nzr ∣ <ₙ ∣ b ,, nzb ∣)       
             -- again, ∣_,,_∣  is the norm operator, which is only for nonzeroes
             
    -- axiom of congruence and irrelevance of the quotient:                                                                                  
    quot-cong : {a a' b b' : C} (nz : b ≉ 0#) (nz' : b' ≉ 0#) → a ≈ a' →  b ≈ b' →         
                                                                  proj₁ (divRem a b nz) ≈ proj₁ (divRem a' b' nz')    
```
It works in many my examples.
Somehow unavoidable complication.
We have to patiently suffer of these irrelevance+congruence proofs.
Other approaches may give a shorter code, but will not be clear to understand.

And concerning relying on `suc`: 
[...]

PR #1463, opened 2021-04-04
Add `Data.Nat.(Factorial/Combinatorics)`

MatthewDaggitt:
This PR results from me trying to refactor #1287. In this PR, the definition of `_C_` (and the newly added `_P_`) are more efficient than the definition in the original PR. Firstly the operators no longer use `Fin` but instead requires you pass a proof that `k ≤ n` when using the properties. Secondly it no longer evaluates `_C_` recursively which took exponential time. I haven't made my way all the way up to the binomial theorem yet, but I thought this was a good point to merge in.

@uzytkownik any feedback you have would be more than welcome!

One niggle that I did encounter is that the current dependency cycles between `Data.Nat.Divisibility` and `Data.Nat.DivMod` make the proof `m/n/o≡m/[n*o]` seriously hard to place, in that it currently requires proofs from both of them and therefore can't live in either. I can't immediately see how to fix it without either adding in a load more nearly empty modules or carrying out the planned reorganisation of the files which will occur in #921. Given that, I've added the proof as a private definition in the new files and will open a new issue to get it fixed in v2.0.


JacquesCarette:
That definition of binomial is still pretty slow.  See the 2nd answer to https://math.stackexchange.com/questions/202554/how-do-i-compute-binomial-coefficients-efficiently for good ways to compute it.

MatthewDaggitt:
> That definition of binomial is still pretty slow. See the 2nd answer to https://math.stackexchange.com/questions/202554/how-do-i-compute-binomial-coefficients-efficiently for good ways to compute it.

Hmm, yes I was foolishly only counting number of operations, not the magnitude of the numbers involved. I'll see if I can tweak it.

subttle:
Not sure if this is of any direct help to you (I haven't looked closely at your code yet) but perhaps it is worth mentioning that the factorial function can be written as a paramorphism for good efficiency.

See [slide](https://github.com/willtim/recursion-schemes/blob/master/slides-final.pdf) 58/106.
```Haskell
fact :: Integer -> Integer
fact = para alg where
  alg Zero = 1
  alg (Succ (f, n)) = f * (n + 1)
```

JacquesCarette:
For the record: I agree that an (inefficient) implementation is better than no implementation. That so many proofs break is a symptom of something discussed already: many fragile proofs that rely on implementation details, rather than being more conceptual. Something for a later day.

PR #1464, opened 2021-04-04
Add bundles for morphisms over binary relations

### LINKED ISSUE #1407: Add bundles for morphisms
While updating the category library to work with v1.5 in https://github.com/agda/agda-categories/pull/243 it's become clear that it is useful to have bundles for the new morphism definitions (which are parameterised by bundles rather than raw bundles). Luckily, we have a natural place to put them:
- `Relation.Binary.Morphism.Bundles`
- `Algebra.Morphism.Bundles`

We should also introduce smart constructors for the morphism bundles in which some of the fields can be inferred automatically (such as congruence over posets).

MatthewDaggitt:
Fixes #1407. 

Pinging @JacquesCarette and @sstucki to let you know that this can be used to undo some of the ugliness I introduced in https://github.com/agda/agda-categories/pull/243 when `v1.6` releases.

Unfortunately I had to remove the module-level arguments in the files I introduced last release, but I think it's unlikely anyone has got round to using them seriously yet!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,10 @@ Non-backwards compatible changes
   `rectangle`, `rectangleˡ`, `rectangleʳ`, `rectangleᶜ`) have been moved to
   `Data.String`.
 
+* The new modules `Relation.Binary.Morphism.(Constant/Identity/Composition)` that
+  were added in the last release no longer have module-level arguments. This is in order
+  to allow proofs about newly added morphism bundles to be added to these files.
+
 Deprecated modules
 ------------------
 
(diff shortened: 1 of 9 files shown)


sstucki:
Thanks for the ping @MatthewDaggitt!

Question: I didn't spot the "smart constructor" for Poset homomorphisms that we discussed back then (see last part of #1407). Did I miss it?

>  We should also introduce smart constructors for the morphism bundles in which some of the fields can be inferred automatically (such as congruence over posets).

Might be worth including in this PR since it's mentioned in the description of the issue. Or maybe you wanna open a separate issue for this?

MatthewDaggitt:
> Might be worth including in this PR since it's mentioned in the description of the issue. Or maybe you wanna open a separate issue for this?

Thanks for the spot! Yup added it as `mkPosetHomo`.

sstucki:
LGTM, thanks!

PR #1465, opened 2021-04-05
Sized types will no longer be safe in 2.6.2

jespercockx:
Due to the continuing soundness issues with sized types, we (the Agda developers) have decided that in the next release of Agda (2.6.2), sized types will unfortunately no longer be considered acceptable under `--safe`. This PR implements the necessary changes to the standard library that are required for this change. On the plus side, this also allows us to get rid of the distinction between safe modules that use `--guardedness` and safe modules that use `--sized-types`, as only the former is allowed now.


MatthewDaggitt:
@gallais thoughts on removing the use of `Stream` from `Data.Nat.PseudoRandom` so that it's safe? Or at least splitting it out into a different module?

gallais:
I was going to suggest using `Codata.Musical.Stream` but it's not safe either
because it has `COMPILE` pragmas. :sweat_smile: 

I think you're right that an `.Unsafe` module is the right way to go.

PR #1470, opened 2021-04-12
Add `begin-irrefl` reasoning combinator

MatthewDaggitt:
This PR extracts the new `begin-irrefl_` combinator proposed in #1385, and continues the conversation started in that PR.

Summary so far:

> @MatthewDaggitt  
> However I'm still reluctant to include this combinator. The reasoning APIs' primary purpose is to improve the readability of proofs, but in my opinion begin-irrefl_ actively harms readability:
>
> Firstly, there is no indication that there's a contradiction being used underneath the hood. Knowing whether or not a proof uses contradiction is fairly crucial to understanding it.
Secondly, unlike when using the other combinators, there is no way to infer the final type of the reasoning proof simply by looking at it. From the reader's perspective the final type is magicked out thin air by the (invisible) contradiction.
I would argue that it's better to keep the contradiction explicit in the code?

> @gallais 
> I think begin-irrefl is clearer than:
> 
> `flip contradiction (<-irrefl refl) $ begin (...)`
>
> I guess we could make the return type ⊥ to force the use of ⊥-elim
> as documentation that there's an explosion somewhere in there but I'm
> not particularly convinced.

> @JacquesCarette 
> Rather than begin-irrefl_, what about by-contradiction_ ? This would take care of @MatthewDaggitt 's first point. It doesn't fully take care of the second, though it becomes less surprising since we all know we can conjure anything we want via ⊥-elim.
>
> Tongue-in-cheek: also have conjure-by-contradiction with 2 arguments, the first the conjured type, the second the falsity proof, that might work.

> @MatthewDaggitt 
> Hmm, yup I'd be on-board with something that had contradiction in it. Would begin-irrefl-contradiction_ or begin-contradiction-by-irrefl_ be too long?

Note that this addition is still currently breaking as it changes the interface to the module `≤-Reasoning` in `Data.Vec.Relation.Binary.Lex.Strict`


MatthewDaggitt:
Brought this up to date, and am going to merge in.

PR #1472, opened 2021-04-12
Add cong₃

carlostome:
Too many times I find myself in need of congruence for ternary constructors, but I also know we cannot have infinitely many definitions of n-ary `cong`, so if there's a better solution than adding `cong₃` I'm open for suggestions.


carlostome:
Oh, just found `congₙ` in `Nary` :sweat_smile:

PR #1475, opened 2021-04-14
add variant of `first?`

damhiya:
`first?` works on `First P (∁ P)` but it seems there is no equivalent for `First (∁ P) P`.
So I added `first?′` which is a swapped variant of `first?`.

I'll request better name for `first?′`, by the way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -448,6 +448,11 @@ Other minor additions
   all-upTo           : All (_< n) (upTo n)
   ```
 
+* Added new proof in `Data.List.Relation.Unary.First.Properties`:
+  ```agda
+  cofirst? : Decidable P → Decidable (First (∁ P) P)
+  ```
+
 * Added new operations in `Data.List.Relation.Unary.Linked`:
   ```agda
   head′ : Linked R (x ∷ xs) → Connected R (just x) (head xs)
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Hey, thanks for the great PR! Yes, your version does seem more natural in many contexts.

I can't help but wondering if the right answer is to rename both of them, so the existing one becomes `first[P]?` and the second `first[∁P]?` The existing name `first?` would then be deprecated, e.g. see the bottom of `Data.Nat.Properties`.

@gallais any thoughts as this is your code originally?

gallais:
> @gallais any thoughts as this is your code originally?

When `Decidable P` holds it is the case that `∁ (∁ P) ⊆ P` so I guess
that's why I was not too bothered about getting the parameters one way
around or the other: you can use `first?` on the negated `P` and then
map double negation elimination over the result if you want to get the
other style.

I wouldn't be opposed to changing the definition or introducing a new
one if we can find a good name. Maybe `first?` vs. `first¿` for the
flipped version?

damhiya:
> Hey, thanks for the great PR! Yes, your version does seem more natural in many contexts.
> 
> I can't help but wondering if the right answer is to rename both of them, so the existing one becomes `first[P]?` and the second `first[∁P]?` The existing name `first?` would then be deprecated, e.g. see the bottom of `Data.Nat.Properties`.
> 
> @gallais any thoughts as this is your code originally?

What about renaming existing one as `first[∁P]?` and new one as `first[P]?`, since value of `First P (∁ P)` is a proof of existence of first element that satisfies `∁ P`?

gallais:
`first` and `cofirst`?

MatthewDaggitt:
Awesome, thanks for changing it @damhiya. One last thing, please could you add a CHANGELOG entry for the new addition. 

As this is so small, I'll include it in the upcoming `v1.6` release.

MatthewDaggitt:
Great, thanks for the PR!

PR #1477, opened 2021-04-14
More properties of absolute value in Data.Rational

bobatkey:
Some additional properties of absolute value for `Data.Rational(.Unnormalised)`.

In `Data.Rational.Properties`:

1. Rename `∣p∣≡p⇒p≡0` to `∣p∣≡0⇒p≡0` so the name matches the type.
2. Add `∣p∣≡p⇒0≤p` because the corresponding result was in `Data.Rational.Unnormalised.Properties`
3. Add `∣∣p∣∣≡∣p∣`

In `Data.Rational.Properties.Unnormalised`:

1. Add `p+p≃0⇒p≃0`
2. Add `0≤∣p∣` -- corresponding result is already in `Data.Rational.Properties`.
3. Rename `∣p∣≡p∨∣p∣≡-p` to `∣p∣≡p⊎∣p∣≡-p` to match the type and the naming in `Data.Rational.Properties`.
4. Add `∣p∣≃p⇒0≤p`
5. Add `∣∣p∣∣≡∣p∣`
6. Add `∣∣p∣∣≃∣p∣`
7. Add `∣-∣-nonNeg`


src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -1647,6 +1650,11 @@ neg-distrib-⊓-⊔ = antimono-≤-distrib-⊓ neg-mono-≤
 ∣p∣≃0⇒p≃0 {mkℚᵘ (ℤ.+ n)  d-1} p≃0ℚ = p≃0ℚ
 ∣p∣≃0⇒p≃0 {mkℚᵘ -[1+ n ] d-1} (*≡* ())
 
+0≤∣p∣ : ∀ p → 0ℚᵘ ≤ ∣ p ∣
+0≤∣p∣ (mkℚᵘ (ℤ.+ ℕ.zero) _)    = *≤* (ℤ.+≤+ ℕ.z≤n)
+0≤∣p∣ (mkℚᵘ (ℤ.+ (ℕ.suc n)) _) = *≤* (ℤ.+≤+ ℕ.z≤n)
```
MatthewDaggitt, line 1655: Can write `+0` instead `(ℤ.+ ℕ.zero)`, `+[1+ _ ]` instead of `(ℤ.+ (ℕ.suc n))` and `-[1+ _ ]` instead of `(ℤ.-[1+ n ])` to tidy things up a bit

src/Data/Rational/Properties.agda
```diff
@@ -1638,6 +1638,10 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
 ∣-p∣≡∣p∣ (mkℚ (+ zero) d-1 _) = refl
 ∣-p∣≡∣p∣ (mkℚ -[1+ n ] d-1 _) = refl
 
+∣p∣≡p⇒0≤p : ∀ {p} → ∣ p ∣ ≡ p → 0ℚ ≤ p
+∣p∣≡p⇒0≤p {p} ∣p∣≡p =
+  toℚᵘ-cancel-≤ (ℚᵘ.∣p∣≃p⇒0≤p (ℚᵘ.≃-trans (ℚᵘ.≃-sym (toℚᵘ-homo-∣-∣ p)) (ℚᵘ.≃-reflexive (cong toℚᵘ ∣p∣≡p))))
```
MatthewDaggitt, line 1643: Might be worth using the equational reasoning here after `toℚᵘ-cancel-≤ (ℚᵘ.∣p∣≃p⇒0≤p ...)` to make it a little bit more readable?

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -1670,9 +1678,27 @@ neg-distrib-⊓-⊔ = antimono-≤-distrib-⊓ neg-mono-≤
   ℤ.+ n ℤ.* 1ℤ       ∎)
   where open ℤ.≤-Reasoning
 
-∣p∣≡p∨∣p∣≡-p : ∀ p → (∣ p ∣ ≡ p) ⊎ (∣ p ∣ ≡ - p)
-∣p∣≡p∨∣p∣≡-p (mkℚᵘ (ℤ.+ n)    d-1) = inj₁ refl
-∣p∣≡p∨∣p∣≡-p (mkℚᵘ (-[1+ n ]) d-1) = inj₂ refl
+∣p∣≡p⊎∣p∣≡-p : ∀ p → (∣ p ∣ ≡ p) ⊎ (∣ p ∣ ≡ - p)
```
MatthewDaggitt, line 1681: Nice spot! Although actually this name was correct and the other name is wrong, as we use logical operators such as `∨` in names rather than type-level operators such as `⊎` (see the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#preconditions-and-postconditions)).

Would it be possible to change this one back, and then change the one in the normalised rationals to match?

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -1670,9 +1678,27 @@ neg-distrib-⊓-⊔ = antimono-≤-distrib-⊓ neg-mono-≤
...
+∣p∣≡p⊎∣p∣≡-p (mkℚᵘ (ℤ.+ n)    d-1) = inj₁ refl
+∣p∣≡p⊎∣p∣≡-p (mkℚᵘ (-[1+ n ]) d-1) = inj₂ refl
+
+
+∣p∣≃p⇒0≤p : ∀ {p} → ∣ p ∣ ≃ p → 0ℚᵘ ≤ p
+∣p∣≃p⇒0≤p {p} ∣p∣≃p with ∣p∣≡p⊎∣p∣≡-p p
+∣p∣≃p⇒0≤p {p} ∣p∣≃p | inj₁ ∣p∣≡p  = ∣p∣≡p⇒0≤p ∣p∣≡p
+∣p∣≃p⇒0≤p {p} ∣p∣≃p | inj₂ ∣p∣≡-p =
+  ≤-reflexive (≃-sym (p+p≃0⇒p≃0 p eq))
+  where open ≤-Reasoning
```
MatthewDaggitt, line 1691: Any chance this could be laid out as:
```agda
≤-reflexive (≃-sym (p+p≃0⇒p≃0 p (begin-equality
  p + p      ≈⟨ +-congʳ p (≃-sym ∣p∣≃p) ⟩
  p + ∣ p ∣  ≈⟨ +-congʳ p (≃-reflexive ∣p∣≡-p) ⟩
  p - p      ≈⟨ +-inverseʳ p ⟩
  0ℚᵘ        ∎)))
  where open ≤-Reasoning
```

Though I suspect this case could be made simpler and more reusable by adding another simple lemma `p≃-p⇒p≃0` and then this case would become simply `rewrite ∣p∣≡-p = ≤-reflexive (p≃-p⇒p≃0 ∣p∣≃p)`

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -1670,9 +1678,27 @@ neg-distrib-⊓-⊔ = antimono-≤-distrib-⊓ neg-mono-≤
...
-∣p∣≡p∨∣p∣≡-p (mkℚᵘ (ℤ.+ n)    d-1) = inj₁ refl
-∣p∣≡p∨∣p∣≡-p (mkℚᵘ (-[1+ n ]) d-1) = inj₂ refl
+∣p∣≡p⊎∣p∣≡-p : ∀ p → (∣ p ∣ ≡ p) ⊎ (∣ p ∣ ≡ - p)
+∣p∣≡p⊎∣p∣≡-p (mkℚᵘ (ℤ.+ n)    d-1) = inj₁ refl
+∣p∣≡p⊎∣p∣≡-p (mkℚᵘ (-[1+ n ]) d-1) = inj₂ refl
+
+
+∣p∣≃p⇒0≤p : ∀ {p} → ∣ p ∣ ≃ p → 0ℚᵘ ≤ p
+∣p∣≃p⇒0≤p {p} ∣p∣≃p with ∣p∣≡p⊎∣p∣≡-p p
+∣p∣≃p⇒0≤p {p} ∣p∣≃p | inj₁ ∣p∣≡p  = ∣p∣≡p⇒0≤p ∣p∣≡p
```
MatthewDaggitt, line 1688: In the two cases you can omit everything to the left of the `|` symbol by replacing it with `...`

MatthewDaggitt (commented):
Thanks for the PR! Particularly awesome work finding the naming bugs. As these were introduced since the last release, I'm bumping this to be included in v1.6.

I've added a few minor comments, but nothing too major.

Any chance you could update the CHANGELOG with the new proofs, and the change the entries for the renamed proofs?

src/Data/Rational/Unnormalised/Properties.agda
bobatkey: Fixed

src/Data/Rational/Properties.agda
bobatkey: Fixed

src/Data/Rational/Unnormalised/Properties.agda
bobatkey: Fixed, and altered the one in `Data.Rational.Properties`.

src/Data/Rational/Unnormalised/Properties.agda
bobatkey: I made the lemma as you suggested, and altered the proof. It wasn't quite as nice due to the need for some uses of symmetry, but it is a bit better.

src/Data/Rational/Unnormalised/Properties.agda
bobatkey: Fixed

bobatkey:
The second commit applies all the fixes in the comments above, and adds entries in `CHANGELOG.md`.

PR #1483, opened 2021-04-22
Added weak-induction schemes for `Data.Fin` and `Data.Nat`

MatthewDaggitt:
At the moment, if you want to prove something by induction that the termination checker doesn't agree is terminating you are forced to pollute the type of your proof with an `Acc` predicate. These new proofs internalise the `Acc` predicate which often makes them much easier and shorter to use, in particular the `Fin` variants which often require a lot of messing around with `inject` and `lower`.

I've left the predicate arguments explicit (to help readability?) but let me know if you think they should be implicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,3 +23,22 @@ New modules
 
 Other minor additions
 ---------------------
+
+* Added new relations to `Data.Fin.Base`:
+  ```agda
+  _≥_ = ℕ._≥_ on toℕ
+  _>_ = ℕ._>_ on toℕ
+  ```
+
+* Added new proofs to `Data.Fin.Induction`:
+  ```agda
+  >-wellFounded   : WellFounded {A = Fin n} _>_
+  
+  <-weakInduction : P zero      → (∀ i → P (inject₁ i) → P (suc i)) → ∀ i → P i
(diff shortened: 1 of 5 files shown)


gallais (approved):

JacquesCarette:
I understand the induction principles on Fin, and how this is a nice way to do it. But on Nat, I don't see the point of `<-weakInduction`, since it seems like the usual induction that's directly provable (by pattern matching). Why would I ever want to use `<-weakInduction` instead?

MatthewDaggitt:
> I understand the induction principles on Fin, and how this is a nice way to do it. But on Nat, I don't see the point of <-weakInduction, since it seems like the usual induction that's directly provable (by pattern matching). Why would I ever want to use <-weakInduction instead?

Excellent point! This could be done by pattern matching and in my definition I could even drop the `Acc` predicate. I guess I was envisaging adding `<-strongInduction` and even things like `<-forwardsBackwardsInduction` eventually, which would be useful for `Nat`. At that point would it better to have a (simplified) definition of `<-weakInduction` for completeness's sake?

MatthewDaggitt:
I've removed weak induction over naturals.

JacquesCarette:
For completeness, you might want to define `<-weakInduction` as a synonym for the 'normal' induction principle, indeed.

MatthewDaggitt:
Okay, well we can revisit the question if and when we add more induction schemes :+1:

PR #1492, opened 2021-04-28
Define 'pinch' as a counterpart to 'punchIn'

TOTBWF:
## Patch Description

During the process of implementing https://github.com/agda/agda-categories/pull/274, @JacquesCarette pointed out that one of the functions I implemented existed in the stdlib under the name `punchIn`. However, this injective function has a surjective counterpart that seems to be missing. This PR adds this function under the name `pinch`, and proves some simple properties about it.

## Notes
I'm not 100% attached to the name, perhaps `pinchIn` would be more appropriate to highlight the connection to `punchIn`. There are also a handful of identities that we can prove relating `pinch` to `punchIn` (See https://en.wikipedia.org/wiki/Simplicial_set#Face_and_degeneracy_maps and https://github.com/agda/agda-categories/pull/274/files#diff-d758c95623d24b748cbba2280dcde56f96e72c64c271bab9afdc4e068efe8117R114), though these seem a bit domain specific for the stdlib.

Furthermore, any map `Fin m -> Fin n` can be factored into a canonical form of a bunch of calls to `pinch` followed a bunch of calls to `punchIn`, though after some preliminary attempts this is proving to be Very Difficult. Perhaps someone braver than I could give it a go.

### DIFF
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -240,13 +240,12 @@ punchIn zero    j       = suc j
 punchIn (suc i) zero    = zero
 punchIn (suc i) (suc j) = suc (punchIn i j)
 
--- The surjection f(i,j) such that f(i,i) = f(i,i+1) = i
+-- The function f(i,j) such that f(i,j) = if j≤i then j else j-1
 
-pinch : ∀ {m} → Fin m → Fin (suc m) → Fin m
-pinch zero zero = zero
-pinch zero (suc j) = j
-pinch (suc i) zero = zero
-pinch (suc i) (suc j) = suc (pinch i j)
+pinch : ∀ {n} → Fin n → Fin (suc n) → Fin n
+pinch {suc n} _       zero    = zero
+pinch {suc n} zero    (suc j) = j
+pinch {suc n} (suc i) (suc j) = suc (pinch i j)
 
 ------------------------------------------------------------------------
 -- Order relations
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -240,6 +240,14 @@ punchIn zero    j       = suc j
 punchIn (suc i) zero    = zero
 punchIn (suc i) (suc j) = suc (punchIn i j)
 
+-- The surjection f(i,j) such that f(i,i) = f(i,i+1) = i
+
+pinch : ∀ {m} → Fin m → Fin (suc m) → Fin m
+pinch zero zero = zero
+pinch zero (suc j) = j
+pinch (suc i) zero = zero
+pinch (suc i) (suc j) = suc (pinch i j)
+
 ------------------------------------------------------------------------
 -- Order relations
 
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -18,7 +18,7 @@ open import Data.Fin.Patterns
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; s≤s; z≤n; _∸_)
 import Data.Nat.Properties as ℕₚ
 open import Data.Unit using (tt)
-open import Data.Product using (∃; ∃₂; ∄; _×_; _,_; map; proj₁; uncurry; <_,_>)
+open import Data.Product using (Σ-syntax; ∃; ∃₂; ∄; _×_; _,_; map; proj₁; uncurry; <_,_>)
 open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂; [_,_]; [_,_]′)
 open import Data.Sum.Properties using ([,]-map-commute; [,]-∘-distr)
 open import Function.Base using (_∘_; id; _$_)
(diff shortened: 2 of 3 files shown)


JacquesCarette:
Small comment: any *order preserving* map `Fin m -> Fin n` can be so factored. Otherwise, all non-identity permutations are counter-examples...

TOTBWF:
Doh! Been thinking too much about simplices, I forgot that non-monotonic maps even existed for a second!

src/Data/Fin/Base.agda
```diff
@@ -240,6 +240,14 @@ punchIn zero    j       = suc j
 punchIn (suc i) zero    = zero
 punchIn (suc i) (suc j) = suc (punchIn i j)
 
+-- The surjection f(i,j) such that f(i,i) = f(i,i+1) = i
+
+pinch : ∀ {m} → Fin m → Fin (suc m) → Fin m
+pinch zero zero = zero
```
MatthewDaggitt, line 246: Should probably collapse 1st and 3rd cases into `pinch _ zero = zero`?

src/Data/Fin/Base.agda
```diff
@@ -240,6 +240,14 @@ punchIn zero    j       = suc j
 punchIn (suc i) zero    = zero
 punchIn (suc i) (suc j) = suc (punchIn i j)
 
+-- The surjection f(i,j) such that f(i,i) = f(i,i+1) = i
+
+pinch : ∀ {m} → Fin m → Fin (suc m) → Fin m
```
MatthewDaggitt, line 245: We usually use `n` for natural numbers when we're only using one of them. I realise we're not that consistent in this particular file!

src/Data/Fin/Base.agda
```diff
@@ -240,6 +240,14 @@ punchIn zero    j       = suc j
 punchIn (suc i) zero    = zero
 punchIn (suc i) (suc j) = suc (punchIn i j)
 
+-- The surjection f(i,j) such that f(i,i) = f(i,i+1) = i
```
MatthewDaggitt, line 243: This comment doesn't explain the behaviour of the function when j != i or i+1?

src/Data/Fin/Properties.agda
```diff
@@ -675,6 +675,23 @@ punchOut-punchIn (suc i) {suc j} = cong suc (begin
   j                                                           ∎)
   where open ≡-Reasoning
 
+
+------------------------------------------------------------------------
+-- pinch
+------------------------------------------------------------------------
+
+pinch-surjective : ∀ {m} i (j : Fin m) → Σ[ k ∈ Fin (suc m) ] (pinch i k ≡ j)
```
MatthewDaggitt, line 683: Can we use the actual `Surjective` type from `Function`? Helps ensure consistency and therefore compatibility between different parts of the library!

src/Data/Fin/Properties.agda
```diff
@@ -675,6 +675,23 @@ punchOut-punchIn (suc i) {suc j} = cong suc (begin
...
+pinch-surjective 0F 0F = zero , refl
+pinch-surjective 0F (suc j) = (suc (suc j)) , refl
+pinch-surjective (suc i) 0F = zero , refl
+pinch-surjective (suc i) (suc j) = map suc (cong suc) (pinch-surjective i j)
+
+pinch-mono-≤ : ∀ {m} (i : Fin m) → (pinch i) Preserves _≤_ ⟶ _≤_
+pinch-mono-≤ 0F {0F} {k} 0≤n = z≤n
+pinch-mono-≤ 0F {suc j} {suc k} (s≤s j≤k) = j≤k
+pinch-mono-≤ (suc i) {0F} {k} 0≤n = z≤n
+pinch-mono-≤ (suc i) {suc j} {suc k} (s≤s j≤k) = s≤s (pinch-mono-≤ i j≤k)
```
MatthewDaggitt, line 693: I don't think you need to pattern match on the `j` and `k` implicits in this proof?

src/Data/Fin/Properties.agda
```diff
@@ -675,6 +675,23 @@ punchOut-punchIn (suc i) {suc j} = cong suc (begin
...
   where open ≡-Reasoning
 
+
+------------------------------------------------------------------------
+-- pinch
+------------------------------------------------------------------------
+
+pinch-surjective : ∀ {m} i (j : Fin m) → Σ[ k ∈ Fin (suc m) ] (pinch i k ≡ j)
+pinch-surjective 0F 0F = zero , refl
+pinch-surjective 0F (suc j) = (suc (suc j)) , refl
```
MatthewDaggitt, line 685: No need for the outer brackets round `suc (suc j)`

MatthewDaggitt (commented):
> I'm not 100% attached to the name, perhaps pinchIn would be more appropriate to highlight the connection to punchIn

Eh, I don't really like the names `punchIn` or `punchOut` either. I think `pinch` is fine for the moment, unless you think there's an analogous `pinch` function for `punchOut` as well?

> There are also a handful of identities that we can prove relating pinch to punchIn ... though these seem a bit domain specific for the stdlib.

Anything that directly relates the two functions is fair game for inclusion into the library. If you think some of them would be useful then feel to add them to the PR.

> Furthermore, any map Fin m -> Fin n can be factored into a canonical form of a bunch of calls to pinch followed a bunch of calls to punchIn, though after some preliminary attempts this is proving to be Very Difficult. Perhaps someone braver than I could give it a go.

Yup, permutations are always a bit nightmarish to work with :cry:

src/Data/Fin/Properties.agda
TOTBWF: Surprisingly, you do! If you don't, agda gets very confused and yells about not being able to solve some unification problems.

MatthewDaggitt: Ah yes because you're matching on `Fin (suc n)` rather than `Fin n`!

MatthewDaggitt:
Thanks for the changes!

PR #1498, opened 2021-05-06
[ re #1465 ] Two more modules that are no longer --safe

jespercockx:
It seems there's two modules that slipped through the cracks in #1465 and that should also be made no longer --safe.


MatthewDaggitt:
Yup I think these were added concurrently to previous pull request :+1:

PR #1499, opened 2021-05-10
[ fixes #1497 ] emulate `[ x , y , z ]` syntactic sugar for `List/Vec/Vector`

### LINKED ISSUE #1497: Possible enhancement for `[_]` in `Data.List/Vec/Vector`
There has been some informal discussion about changing the singleton list/vector `[_]` to a pattern synonym so that you can use it on the left hand side and write things like:
```agda
intersperse : List A -> A -> List A
intersperse [] _ = []
intersperse [ x ] _ = [ x ]
intersperse (x :: xs) y = x :: y :: intersperse xs y
```
However, swapping functions for (untyped) pattern synonyms is not backwards compatible so it hasn't been carried out yet.

In the meantime I propose that we redefine `[_]` to make it more powerful so we can write an arbitrary number of objects inside of the brackets as in other programming languages, e.g. `[ 5 , 2 , 3 ]`. This can be achieved by using `_^_` from `Data.Vec.Recursive` as follows:
```
[_] : A ^ (suc n) → Vec A (suc n)
[_] x = toVec _ x
```
This _is_ a backwards compatible change, and I don't think precludes making it a pattern in v2.0? @gallais 

(although it may need a bit of rejigging to fix the dependency cycles)

MatthewDaggitt:


MatthewDaggitt:
Eugh, maybe this wasn't such a great idea after all. It's causing serious problems with variables and metavariables's. For example:
```agda
↭-singleton-inv : ∀ {x : A} {xs : List A} → xs ↭ [ x ] → xs ≡ [ x ]
```
is giving me an unsolved meta for `_A.a_780` and is only fixed by changing it to:
```agda
↭-singleton-inv : ∀ {a} {A : Set a} {x : A} {xs : List A} → xs ↭ [ x ] → xs ≡ [ x ]
```

@gallais any ideas why it can't solve for the level in the above solution? My only idea is that it's somehow unifying `A ^ n` and is not sure that the level information gets preserved in the `A ^ 0` case and is therefore unable to solve for `A.a`, even though we're in the `A ^ 1` case. Do you think it's a bug in Agda?

MatthewDaggitt:
Okay, made it into a `HasListSyntax` type-class which means that implementations for different types can be used concurrently :tada: 

Still got the annoying inference bug though. Posted on the Zulip about it.

MatthewDaggitt:
Hmm okay, so unfortunately at least one issue (https://github.com/agda/agda/issues/5487) with Agda is currently blocking this change. It can be worked around the changes the really disruptive and I don't think we want to put the user's through that. Reluctantly adding this to the `Agda-2.6.3` milestone rather than `v2.0`.

MatthewDaggitt:
Closing as (https://github.com/agda/agda/issues/5487) has been added to the icebox.

PR #1502, opened 2021-05-14
Add algebra morphism identity and composition constructions

### LINKED ISSUE #1500: Proofs that identity and composition of homomorphisms are homomorphisms
I'd like there to be modules (maybe `Algebra.Morphism.Construction.Identity` and `Algebra.Morphism.Construction.Composition` but this is an unpainted bike shed) which have witnesses that the various things in `Algebra.Morphism.Structures` are satisfied by `id` and `_∘_` from `Function.Base`.

I think this requires the algebraic structures to be the actual algebraic structure, not `RawX`

Taneb:
Closes #1500 

I made slight changes to the `IsLattice` and `Lattice` records.

Parts of this PR worry me that the definition of `RingHomomorphism` etc is incorrect, as they carry two proofs that the function is congruent modulo the equivalence relation, and as such don't provide any directly


src/Algebra/Morphism/Construct/Composition.agda
```diff
@@ -0,0 +1,404 @@
...
+open import Function.Base using (_∘_)
+open import Level using (Level)
+open import Relation.Binary.Morphism.Construct.Composition
+
+private
+  variable
+    a b c ℓ₁ ℓ₂ ℓ₃ : Level
+
+module _
+  {M₁ : Magma a ℓ₁} {M₂ : Magma b ℓ₂} {M₃ : Magma c ℓ₃}
```
MatthewDaggitt, line 23: I don't think you need the full `Magma` structures right? You should be able to make do with only `RawMagma`s + the transitivity of the 3rd equality? The same goes for `Monoid` etc.

I say this because one of the main usages of these modules is proving that e.g. a given `RawGroup` is a full `Group` because it's monomorphic with another `Group`.

src/Algebra/Morphism/Construct/Composition.agda
```diff
@@ -0,0 +1,404 @@
...
+      module f-mono = IsMagmaMonomorphism f-mono
+      module g-mono = IsMagmaMonomorphism g-mono
+
+  isMagmaIsomorphism
+    : IsMagmaIsomorphism M₁.rawMagma M₂.rawMagma f
+    → IsMagmaIsomorphism M₂.rawMagma M₃.rawMagma g
+    → IsMagmaIsomorphism M₁.rawMagma M₃.rawMagma (g ∘ f)
+  isMagmaIsomorphism f-iso g-iso = record
+    { isMagmaMonomorphism = isMagmaMonomorphism f-iso.isMagmaMonomorphism g-iso.isMagmaMonomorphism
+    ; surjective = λ x →
```
MatthewDaggitt, line 64: `Function.Construct.Composition.surjective` perhaps?

src/Algebra/Morphism/Construct/Composition.agda
```diff
@@ -0,0 +1,404 @@
...
+  isMagmaHomomorphism
+    : IsMagmaHomomorphism M₁.rawMagma M₂.rawMagma f
+    → IsMagmaHomomorphism M₂.rawMagma M₃.rawMagma g
+    → IsMagmaHomomorphism M₁.rawMagma M₃.rawMagma (g ∘ f)
+  isMagmaHomomorphism f-homo g-homo = record
+    { isRelHomomorphism = isRelHomomorphism f-homo.isRelHomomorphism g-homo.isRelHomomorphism
+    ; homo = λ x y → M₃.trans (g-homo.⟦⟧-cong (f-homo.homo x y)) (g-homo.homo (f x) (f y))
+    }
+    where
+      module f-homo = IsMagmaHomomorphism f-homo
```
MatthewDaggitt, line 43: You could shorten these modules just to `F` and `G` which would cut out a lot of noise in the record construction.

MatthewDaggitt (commented):
Thanks for the PR! Looks like quite a tedious one so cheers for doing it.

I've added a few comments that should help reduce some the code needed.

src/Algebra/Morphism/Construct/Composition.agda
Taneb: Should the same be done for Identity? I think that just needs `refl`

MatthewDaggitt: Yes, you should avoid using the full bundles in all similar code!

MatthewDaggitt:
@Taneb apologies I somehow missed your comment. I've now answered that. Just to say that if you would like to get this included in the upcoming v1.7 release, it'd be great to get this wrapped up relatively soon so I can include it in the release candidate :+1:

MatthewDaggitt:
> Parts of this PR worry me that the definition of RingHomomorphism etc is incorrect, as they carry two proofs that the function is congruent modulo the equivalence relation, and as such don't provide any directly

Yes, the situation is not very pleasing, I agree. If you have any practical suggestions to fix it, please do open an issue or a PR?

PR #1504, opened 2021-05-17
Add explicit --guardedness and --sized-types flags

jespercockx:
In the light of making `--sized-types` no longer safe, we're considering making both `--sized-types` and `--guardedness` no longer enabled by default. The reason is that the two are inconsistent when used together, so it is better to make the choice explicitly. Since we previously had this rule already when `--safe` was enabled, the changes required to the standard library are only minor.

Aside from just adding the flags, we could maybe also consider refactoring the modules that use both `--sized-types` and `--guardedness` to use one or the other.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,8 +128,7 @@ matrix:
         - echo $AGDA_OPTIONS
         - echo $RTS_OPTIONS
         # checking safe modules build with --safe
-        - agda $AGDA_OPTIONS $RTS_OPTIONS -i . -i src/ --safe EverythingSafeGuardedness.agda
-        - agda $AGDA_OPTIONS $RTS_OPTIONS -i . -i src/ --safe EverythingSafeSizedTypes.agda
+        - agda $AGDA_OPTIONS $RTS_OPTIONS -i . -i src/ --safe EverythingSafe.agda
         # detecting basic compilation errors
         - agda $AGDA_OPTIONS $RTS_OPTIONS -i . -i src/ -c --no-main Everything.agda
         # compiling & running the examples using the FFI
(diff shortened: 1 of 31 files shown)


jespercockx:
Actually, it would be very easy to avoid the combination of `--guardedness` and `--sized-types` in (almost?) all cases by removing the two functions `fromMusical` and `toMusical` from the modules `Codata.Musical.Colist` and `Codata.Musical.Conat`. I could add that to this PR, or create a separate PR for it, please let me know what you prefer.

MatthewDaggitt:
> Actually, it would be very easy to avoid the combination of --guardedness and --sized-types in (almost?) all cases by removing the two functions fromMusical and toMusical from the modules Codata.Musical.Colist and Codata.Musical.Conat. I could add that to this PR, or create a separate PR for it, please let me know what you prefer.
>
> (If you think these functions should be kept, perhaps they could be moved to a separate module.)

Neat spot. Yes, let's isolate them all in a single module `Codata.Musical.Conversion` (a little bit of renaming required, e.g. `toMusical` -> `toMusicalColist`) and add a suitably scary comment at the top that you really shouldn't be mixing sized types and guardedness?

MatthewDaggitt:
As you're making a PR to experimental any chance you could bump the dev Agda commit being used?

https://github.com/agda/agda-stdlib/blob/28cdb06fb5d2d91b597b26293a892643e0618a9b/.travis.yml#L59

I think the bug that was stopping the tests from passing has now been fixed upstream...

jespercockx:
This is ready to be merged now I believe.

MatthewDaggitt:
Looks great! Thanks for the changes :+1:

PR #1511, opened 2021-05-30
Setoid version of indexed containers.

andreasabel:
Following the structure for non-indexed containers.

Putting this out for a first review.

I also added an `Example` development: multi-sorted algebras.
This should be refactored after also adding a setoid version of `Data.W.Indexed` to the library.

Where in the hierachy would proofs be placed that the definitions in `Data.W.Indexed` are `Func`s, i.e., respect setoid-equality?


src/Data/Container/Indexed/Examples/MultiSortedAlgebra.agda
```diff
@@ -0,0 +1,433 @@
...
+      t  [ σ₀ ]          ≈˘⟨ evaluation t σ₀ ⟩
+      ⦅ t   ⦆ .apply σ₀  ≈⟨ V (M Γ) satisfies σ₀ ⟩
+      ⦅ t'  ⦆ .apply σ₀  ≈⟨ evaluation t' σ₀ ⟩
+      t' [ σ₀ ]          ≈⟨ identity t' ⟩
+      t'                 ∎
+      where open SetoidReasoning (TmSetoid Γ s)
+
+{- Q.E.D 2021-05-28 -}
+
+-- -}
```
MatthewDaggitt, line 429: Unnecessary comments?

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
```diff
@@ -0,0 +1,59 @@
...
+--
+-- Equality over indexed container extensions parametrised by a setoid
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary using (Setoid)
+
+module Data.Container.Indexed.Relation.Binary.Equality.Setoid
+  {ℓⁱ ℓᶜ ℓᵉ} {I : Set ℓⁱ} (S : I → Setoid ℓᶜ ℓᵉ)
```
MatthewDaggitt, line 12: I think this should be `IndexedSetoid` from `Relation.Binary.Indexed.Homogeneous(.Bundles)?`

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
```diff
@@ -0,0 +1,59 @@
...
+open Setoid using (Carrier) renaming (_≈_ to Equality)
+
+private variable
+  ℓˢ ℓᵖ : Level
+  O : Set _
+
+------------------------------------------------------------------------
+-- Definition of equality
+
+module _ (C : Container I O ℓˢ ℓᵖ) (o : O) where
```
MatthewDaggitt, line 32: This can technically be lifted up to be a module argument. I have no idea whether it _should_ be or not, just throwing it out there :+1:

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
```diff
@@ -0,0 +1,59 @@
...
+
+  Eq : Rel (⟦ C ⟧ (Carrier ∘ S) o) (ℓᵉ ⊔ ℓˢ ⊔ ℓᵖ)
```
MatthewDaggitt, line 34: If you do lift it the above arguments to be module arguments, this could then become infix?

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
```diff
@@ -0,0 +1,53 @@
...
+open import Data.Container.Indexed.Core using (Container; Subtrees; ⟦_⟧)
+
+private variable
+  ℓᵉ ℓᵉ′ ℓᵖ ℓˢ ℓˣ ℓʸ : Level
+  I O : Set _
+
+-- Equality, parametrised on an underlying relation.
+
+module _ (C : Container I O ℓˢ ℓᵖ)
+         {X : I → Set ℓˣ} {Y : I → Set ℓʸ} (R : (i : I) → REL (X i) (Y i) ℓᵉ)
```
MatthewDaggitt, line 26: Modulo an implict `i` then the type of R is:

https://github.com/agda/agda-stdlib/blob/358ad8bf673f0b292f5e5e86b50f0f6ec1d245cd/src/Relation/Binary/Indexed/Homogeneous/Core.agda#L32

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
```diff
@@ -0,0 +1,53 @@
...
+  -- Maybe we need to update module Data.Container.Relation.Binary.Pointwise accordinly
+  --
+  -- record Pointwise  : Set (ℓˢ ⊔ ℓᵖ ⊔ ℓᵉ) where
+  --   constructor _,_
+  --   field shape    : c ≡ c'
+  --         position : Eqs shape xs ys
+
+module _ {C : Container I O ℓˢ ℓᵖ}
+         {X : I → Set ℓˣ} {Y : I → Set ℓʸ}
+         {R  : (i : I) → REL (X i) (Y i) ℓᵉ}
```
MatthewDaggitt, line 48: Again maybe `IREL`?

src/Data/Container/Indexed/Relation/Binary/Pointwise/Properties.agda
```diff
@@ -0,0 +1,48 @@
...
+open import Relation.Binary
+open import Relation.Binary.PropositionalEquality as P
+  using (_≡_; subst; cong)
+
+private variable
+  ℓᵉ ℓᵖ ℓˢ ℓˣ : Level
+  I O : Set _
+
+module _ (C : Container I O ℓˢ ℓᵖ) {X : I → Set ℓˣ}
+         (R : (i : I) → Rel (X i) ℓᵉ)
```
MatthewDaggitt, line 25: https://github.com/agda/agda-stdlib/blob/358ad8bf673f0b292f5e5e86b50f0f6ec1d245cd/src/Relation/Binary/Indexed/Homogeneous/Core.agda#L37

src/Data/Container/Indexed/Relation/Binary/Pointwise/Properties.agda
```diff
@@ -0,0 +1,48 @@
...
+         {o : O}
+         where
+
+  refl : (∀ i → Reflexive (R i)) → Reflexive (Pointwise C R o)
```
MatthewDaggitt, line 29: (Again modulo an implicit argument) we have the type of indexed reflexivity/symmetry etc. already defined:

https://github.com/agda/agda-stdlib/blob/358ad8bf673f0b292f5e5e86b50f0f6ec1d245cd/src/Relation/Binary/Indexed/Homogeneous/Definitions.agda#L36

src/Data/Container/Indexed/Examples/MultiSortedAlgebra.agda
```diff
@@ -0,0 +1,433 @@
...
+-- multi-sorted algebra C.
+
+-- We need to interpret indexed containers on Setoids.
+-- This definition is missing from the standard library v1.7.
+
+⟦_⟧s : (C : Container I S ℓᵒ ℓᵃ) (ξ : I → Setoid ℓᵐ ℓᵉ) → S → Setoid _ _
+⟦ C ⟧s ξ = ICSetoid.setoid ξ C
+
+-- Multi-sorted algebras
+-- =====================
```
MatthewDaggitt, line 61: Bikeshedding, but we try to keep a consistent comment style in the library. Double header:
```agda
-- Multi-sorted algebras
```
and single headers:
```agda
-- Model
```

src/Data/Container/Indexed/Examples/MultiSortedAlgebra.agda
```diff
@@ -0,0 +1,433 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Multi-sorted algebras as indexed containers
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --without-K --guardedness #-}  -- forced by Agda-2.6.2-rc1
+
+module Data.Container.Indexed.Examples.MultiSortedAlgebra where
```
MatthewDaggitt, line 9: I think we are now putting all `Examples` under the `README` directory. So I think this could be `README.Data.Container.Indexed.MultiSortedAlgebra`?

You should also import it in the `README.Data.Container.Indexed` module so that it gets type-checked.

MatthewDaggitt (commented):
Looks great, thanks for the awesome documentation! Added a few comments.

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
andreasabel: I tried your suggestion, but passing an `IREL` to `Pointwise` then needs an eta-expansion, like:
```agda
map : (R⇒R′ : ∀{i : I} → R {i} ⇒ R′ {i}) {o : O} → Pointwise C (λ{i} → R {i}) o ⇒ Pointwise C (λ{i} → R′ {i}) o
````
Maybe `IREL`s would be better off with a non-hidden argument?

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
andreasabel: Btw, I could not find `_⇒_` for `Homogeneous.IREL`.  I think it should be simply the pointwise-subrelation.

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
andreasabel: What is the advantage of packing this as `IndexedSetoid`?  
Note that `Func` does not exists for `IndexedSetoid`.  With
`S T : I -> Setoid` I can get a function as `(i : I) -> Func (S i) (T i)`, but there seems nothing as convenient for `S T : IndexedSetoid`, as far as I could see.

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
andreasabel: I see.  I followed the style of the same module for non-indexed containers, in `Data.Container.Relation.Binary.Equality.Setoid`.

src/Data/Container/Indexed/Relation/Binary/Equality/Setoid.agda
MatthewDaggitt: Hmm from my perspective, the first advantage is that we keep a consistent interface in the library, the second is that `IndexedSetoid` is a record object that you can open and get everything useful you want immediately (with the right implicits/explicits). In contrast `I -> Setoid _ _` is difficult to access the individual indexed proofs.

 Yup, `IndexedFunc` doesn't exist yet but that doesn't mean it shouldn't exist in the future.

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
MatthewDaggitt: Hmm interesting. I've used `IRel` (rather than `IREL`) a reasonable amount, and Agda's always been able to infer the hidden arguments fine. Hmm I'll open an issue to address it.

MatthewDaggitt:
@andreasabel if you would like this merged in for v2.0 please could you address the comments? Otherwise I'll mark it as abandoned.

MatthewDaggitt:
Marking as abandoned and closing

andreasabel:
Sorry for not getting back.

Our last status was that we disagreed on whether to use `I -> Setoid` or `IndexedSetoid`, and on this fundamental issue there was no resolve, so that PR is stuck.

src/Data/Container/Indexed/Relation/Binary/Pointwise.agda
jamesmckinna: Of course, I might prefer the eta-expansion in the examples above to look like `λ {i} → R′ {i = i}` rather than `λ{i} → R′ {i}`, but is there a technical reason why Agda doesn't/can't eta-expand implicit functions? 

I've been meaning to raise an issue for Agda itself for a while, and it came up in the context of this discussion when we were reviewing outstanding issues for the `std-lib` v2.0 milestone, so I'll ask it here as a shortcut...

jamesmckinna:
And the distinction.conflict arising between the use of `I -> Setoid` or `IndexedSetoid` turns on two things, IIUC:
- [ ] the absence of a suitable function space between `IndexedSetoid`s
- [ ] the contrast between explicit/implicit quantification over `I` in the two approaches

Can each of these be easily overcome? or is my question above about implicit eta-expansion one reason why the second issue is a deal-breaker?

jamesmckinna:
I had promised, a loooong time ago, to try to understand (and fix) this, but it fell completely off my worklist. But I don't think we should abandon it (yet): v2.0 or v3.0?

MatthewDaggitt:
Okay, brought this up to date. I no longer care so much about unifying the different versions of indexing so I propose to merge this in once the checks passed. Sorry for such a long wait @andreasabel

MatthewDaggitt (approved):

MatthewDaggitt:
Eek still needs a changelog entry

PR #1513, opened 2021-05-31
Made `Data.Container(.Any/.FreeMonad)` safe by splitting out fixed points

### LINKED ISSUE #1466: Why does `Data.Container.FreeMonad` use --sized-types?
It does not seem to be used internally to the code. This infects `Data.Container` too.

MatthewDaggitt:
Fixes #1466


src/Data/Container.agda
```diff
@@ -40,11 +35,3 @@ module Morphism where
 
   open import Data.Container.Morphism
     using (id; _∘_) public
-
--- The least and greatest fixpoints of a container.
-
-μ : ∀ {s p} → Container s p → Set (s ⊔ p)
```
JacquesCarette, line 46: This seems "too far" for me: `Data.W` is safe. It's only `Codata.M` which requires `--sized-types`. So I'd be tempted to be finer in the splitting, i.e. keep `μ` here.  And then only split off GreatestFixpoint.

src/Data/Container/FreeMonad.agda
```diff
@@ -38,7 +38,7 @@ _⋆C_ : ∀ {x s p} → Container s p → Set x → Container (s ⊔ x) p
 C ⋆C X = const X ⊎ C
 
 _⋆_ : ∀ {x s p} → Container s p → Set x → Set (x ⊔ s ⊔ p)
-C ⋆ X = μ (C ⋆C X)
+C ⋆ X = W (C ⋆C X)
```
JacquesCarette, line 41: And indeed, here you use `W` safely!

JacquesCarette (changes requested):

src/Data/Container.agda
MatthewDaggitt: I was trying to avoid to much fragmentation, but you're right it's better to fragment than be precise here. I've pushed changes that do as you suggest :+1:

gallais:
I am perversely tempted to suggested adding a sized version of the
least fixpoint and a guarded version of the greatest one. This way
users can pick their poison (lack of safety but more compositional
termination checking vs. safety but annoying checks).

I can prepare the commits doing that & push it to this PR if you
approve the idea.

andreasabel:
> adding a sized version of the
> least fixpoint and a guarded version of the greatest one.

It seems that atm there are different choices of greatest fp for indexed vs. non-indexed:
- indexed: musical M
- non-indexed: sized M

I think it would make sense to have all four variants in each case, as @gallais suggests:
- `W` sized and non-sized
- `M` sized and musical

andreasabel (approved):

MatthewDaggitt:
Yup, happy to have all the variants included if you want to update the PR @gallais

JacquesCarette:
Multiplicity of choice, in this case, seems like a good idea to me as well.

MatthewDaggitt:
@gallais are you going to have time to look at this today? Otherwise let me know and I'll have a crack at it, as I think we'd probably better make a 2nd release candidate sooner rather than later.

gallais:
Yep, doing it today during AIM.

gallais:
This is ready I think. But please merge #1517 first.

src/Data/Container/Sized.agda
```diff
@@ -0,0 +1,27 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Sized containers, based on the work of Abbott and others
```
MatthewDaggitt, line 4: Hmm I'm not sure about this module name... It's not really about sized containers per se, only the implementation of the fixpoints. Would this be better as `Data.Container.Fixpoints.Sized`?

src/Data/W/Sized.agda
```diff
@@ -0,0 +1,68 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Sized W-types
```
MatthewDaggitt, line 4: Needs a CHANGELOG entry

src/Data/Container/Indexed/GreatestFixpoint.agda
```diff
@@ -0,0 +1,25 @@
...
+open import Codata.Musical.M.Indexed
+open import Data.Container.Indexed using (Container)
+open import Relation.Unary using (Pred)
+
+private
+  variable
+    o c r : Level
+    O : Set o
+
+-- This lives in its own module due to its unsafe use of sized types.
```
MatthewDaggitt, line 21: Pretty sure this is my fault, but this comment is wrong! Doesn't use sized types...

src/Data/Container/GreatestFixpoint.agda
```diff
@@ -0,0 +1,23 @@
...
+
+open import Level using (Level; _⊔_)
+open import Codata.Musical.M using (M)
+open import Data.Container using (Container)
+
+private
+  variable
+    s p : Level
+
+-- This lives in its own module due to its unsafe use of sized types.
```
MatthewDaggitt, line 19: This comment is wrong as well, my fault, sorry!

src/Data/Container/GreatestFixpoint.agda
```diff
@@ -0,0 +1,23 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Greatest fixpoint for containers
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --without-K --guardedness #-}
+
+module Data.Container.GreatestFixpoint where
```
MatthewDaggitt, line 9: To have a consistent naming convention with the sized variant, should this go under `Data.Container.Fixpoint.Guarded`? (and then publicly re-export `\mu` from `Data.Container`?)

MatthewDaggitt (commented):
So to summarise the comments, I think the most consistent structure would be:
`Data.Container(.Indexed).Fixpoint.(Sized/Guarded)`?

where the `Guarded` modules re-export the least fixpoints that live in `Data.Container(.Indexed)`?

src/Data/Container/GreatestFixpoint.agda
gallais: It used to be correct but then I moved this greatest fixpoint from sizes to guardedness. :D

src/Data/Container/GreatestFixpoint.agda
gallais: I like this.

PR #1514, opened 2021-06-01
[ Improved fix #1510 ] Get rid of `error`, use `throwError` and `die`.

### LINKED ISSUE #1510: GenerateEverything: use of `last` falsely compiled by GHC >= 8.2
With a module not following the convention for the header, in particular using a block comment in the beginning, `GenerateEverything` crashes ungracefully:
```
cabal run GenerateEverything
Up to date
GenerateEverything: Prelude.last: empty list
```

For a start, purge all unsafe uses of `Prelude.last`, or better even _all_ uses of `last`, employing non-empty lists...

andreasabel:
Solidifying the fix of #1510.  Just don't use `error` for reporting errors to the user.

The discussion at https://gitlab.haskell.org/ghc/ghc/-/issues/19917 strongly discouraged me from ever considering `error` as a means to report errors to the user.  I already knew it was bad style, but now I know that it is not guaranteed to deliver the correct message at all.

### DIFF
--- a/GenerateEverything.hs
+++ b/GenerateEverything.hs
@@ -1,10 +1,13 @@
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE PatternSynonyms #-}
 
 import Control.Applicative
 import Control.Monad
+import Control.Monad.Except
 
 import qualified Data.List as List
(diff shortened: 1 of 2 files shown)


gallais:
Thanks for looking into this!

Issue #417, opened 2018-08-07
Codata.Musical.Stream cannot be compiled


gallais:
re @wenkokke's remark on #112 

> This pull request didn't update the associated Haskell code meaning that compiling Data.Stream with MAlonzo now fails

This bug is still in the standard library! :(
Serge Leblanc (re)uncovered it today. I'll push a fix ASAP.

@MatthewDaggitt Should we consider releasing 0.16.1 with this bug-fix cherry-picked? If so, we should make sure all the other `Codata.Musical.*` modules do compile.

gallais:
Reopening to discuss bug-fix release.

MatthewDaggitt:
If people are trying to use it, then we probably should release v0.16.1. I'll try and get round to it in the next few days.

Issue #370, opened 2018-07-07
Foreign.Haskell: COMPILE pragma lacks data keyword


semenov-vladyslav:
I guess https://github.com/agda/agda-stdlib/blob/master/src/Foreign/Haskell.agda#L28 should be
`{-# COMPILE GHC Pair = data MAlonzo.Code.Foreign.Haskell.AgdaPair ((,)) #-}`

gallais:
Sorry about that. Thanks for the report!

PR #288, opened 2018-04-21
Drop incorrect module qualifier in FFI binding

UlfNorell:
See agda/agda#3040.

There's no need (neither in 2.5.3 or above) to qualify local names in FFI bindings, but if you do it should be with the correct module name 😁.

PR #1518, opened 2021-06-08
[ new ] golden testing framework

gallais:
This is a port of the Idris2 golden testing library: https://github.com/idris-lang/Idris2/tree/master/libs/test

We can use it to compile and run some programs using the standard library's less
certified corners to check we get sensible results. This should help us:

* catch bugs like agda/agda#5288
* make sure the standard library's modules do compile (#288, #370, #417)

Coverage: the tests are currently compiling 434 out of 943 modules

TODO
----

* [x] CHANGELOG
* [x] Turn off the interactive flag when running on CI
* [x] Reorder phases to typecheck the whole lib first
* [x] Test for the pretty-printer!
* [x] Add `no-colour` option to golden testing (& check `TERM` env variable is not `DUMB`)


.github/workflows/ci-ubuntu.yml
```diff
@@ -73,13 +73,19 @@ jobs:
           ${{ env.CABAL_INSTALL }}
           cd ..
 
-     # Download and test stdlib
+      # Download and test stdlib
       - name: Checkout stdlib
         uses: actions/checkout@v2
 
-      - name: Test stdlib
+      - name: Install agda-stdlib-utils
```
andreasabel, line 80: Why do you need to install `agda-stdlib-utils` when you use `cabal run GenerateEverything`?  This builds the executable before running it...

.github/workflows/ci-ubuntu.yml
```diff
@@ -73,13 +73,19 @@ jobs:
           ${{ env.CABAL_INSTALL }}
```
andreasabel, line 73: You are installing `agda` with a `v1` command, would that work with a `v2` command?
You need to pass the `--installdir` then, though.

andreasabel (commented):
Comments in response to your zulip prompt:
> So I've gotten the CI to successfully run the golden tests but I had to use v1-install on agda-stdlib-utils.cabal which is a legacy command. Any experts on this sort of stuff can help make the build more future proof?

MatthewDaggitt (commented):
This is amazing @gallais :tada: when you said you were going to write a few tests I wasn't expecting anything like this!

A few minor comments here and there. The main thing is to make sure we have all the (breaking) changes in the CHANGELOG once you're finished.

.github/workflows/ci-ubuntu.yml
gallais: Sorry I should have been clearer: `runtests` depends on `System.Clock`
and unless I install the `clock` package globally I get the following
error when running `agda -c runtests.agda`

> Could not find module ‘System.Clock’

[cf. this log](https://github.com/agda/agda-stdlib/runs/2808733510#step:10:565)

My solution was to add `clock` as a dependency to the utils & to use
`v1-install` to force the installation of the dependencies.

.github/workflows/ci-ubuntu.yml
gallais: This is a side effect of changing `install` to `v1-install`. In practice
I don't really care about how agda is installed as long as it's available
in `~/.cabal/bin` (which the previous `install --overwrite-policy=always`
guaranteed AFAICT).

src/Test/Golden.agda
```diff
@@ -0,0 +1,379 @@
...
+  open Summary public
+
+  init : Summary
+  init = mkSummary [] []
+
+  merge : Summary → Summary → Summary
+  merge (mkSummary ws1 ls1) (mkSummary ws2 ls2) =
+   mkSummary (ws2 ++ ws1) (ls2 ++ ls1)
+
+  update : List Result -> Summary -> Summary
```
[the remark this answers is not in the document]
gallais: Copy/pasted from the Idris2 code :D

src/Data/Rational/Unnormalised/Show.agda
```diff
@@ -0,0 +1,16 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Showing unnormalised rational numbers
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Rational.Unnormalised.Show where
```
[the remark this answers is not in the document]
gallais: I think I'll wait until v1.7 is released and the CHANGELOG is purged. No need to
create merge conflicts for my future self.

src/System/Clock.agda
```diff
@@ -0,0 +1,101 @@
...
+
+import Data.Nat.Show as ℕ
+open import Data.Nat.DivMod using (_div_)
+import Data.Nat.Properties as ℕₚ
+open import Data.Fin.Base using (Fin; toℕ)
+open import Data.String.Base using (String; _++_; padLeft)
+open import Data.Char.Base using (Char)
+open import Relation.Nullary.Decidable using (False; fromWitnessFalse; toWitnessFalse)
+
+show : Time → Fin 9 → String
```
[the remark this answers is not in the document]
gallais: We should pass a `Fin 10` :see_no_evil:

gallais:
:rocket:

PR #1522, opened 2021-06-17
[ ci ] doc for the actions script

gallais:
I wrote some doc for https://github.com/agda/agda-categories/pull/285
so we may as well have some in the stdlib repo too!


MatthewDaggitt:
Awesome, thank you!

PR #1525, opened 2021-06-20
The start of a small matrix library

MatthewDaggitt:
Someone on the mailing list was asking about matrices, so I thought I'd start the process of porting some of what I have. This is the first very small step.

My main question is that I'm using `Fin` indexed functions as the basic representation. Vectors which use a similar representation live in the `Data.Vec.Functional`. Do people think that we should therefore have these matrices live under `Data.Matrix.Functional` in case we ever want to add a representation based-off of `Data.Vec`?

My instinct is no. When I was a novice Agda user I wasted several months trying to get a matrix library off the ground using `Data.Vec` and it was unbelievably hard. As 99% of the time is spent querying into matrices and you almost never want to alter their size (inductively), I think that the argument for a `Data.Vec` based matrix library is pretty weak. Nonetheless I thought I'd pose the question?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ New modules
   Algebra.Morphism.Construct.Identity
   ```
 
+* The start of a small section about matrices
+  ```
+  Data.Matrix
+  ```
+
 Other minor additions
 ---------------------
 
@@ -46,3 +51,14 @@ Other minor additions
 
 Other minor additions
 ---------------------
+
+* Added new operation in `Data.Fin.Base`:
+  ```agda
+  _≡ᵇ_ : Fin m → Fin n → Bool
+  ```
+
+* Added new operation in `Data.Vec.Functional`:
+  ```agda
+  foldr+ : Op₂ A → Vector A (suc n) → A
+  foldl+ : Op₂ A → Vector A (suc n) → A
+  ```
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -12,6 +12,7 @@
 
 module Data.Fin.Base where
 
+open import Data.Bool.Base using (Bool)
 open import Data.Empty using (⊥-elim)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; z≤n; s≤s)
 open import Data.Nat.Properties.Core using (≤-pred)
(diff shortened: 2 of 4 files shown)


jespercockx:
Asides from proving things about matrices, there's also the consideration of actually running and/or compiling your code (*gasp*). For this neither representation is really good regarding performance, but at least the vector-based one gives some hope of predictable performance (when using a strict backend).

jespercockx:
IMHO the 'right' thing to do would be to provide an abstract interface for (operations on) matrices, and then provide one or more implementations of this interface (e.g. as functions, as vectors, ...) However, Agda is not very well suited to this style of programming/proving, as basically nothing holds by computation. And consequently, the standard library doesn't really support this style either.

MatthewDaggitt:
> Asides from proving things about matrices, there's also the consideration of actually running and/or compiling your code (gasp).

Yes, I always forget about that one :facepalm: 

> IMHO the 'right' thing to do would be to provide an abstract interface for (operations on) matrices, and then provide one or more implementations of this interface (e.g. as functions, as vectors, ...) However, Agda is not very well suited to this style of programming/proving, as basically nothing holds by computation. And consequently, the standard library doesn't really support this style either.

Hmm, I agree that that definitely sounds like a more appealing approach. But defining such an interface is tricky right? Once you've created it, it's impossible to add new operations to it without breaking backwards compatibility. So you have to be sure you've got a minimal set of operations (and axioms!) to begin with. Looking at other matrix libraries in other theorem provers, Lean doesn't seem to have an interface at all (although it does abstract over the indices) and Coq doesn't seem to have matrices in the standard library although it does have a misnamed(?) [specification for vectors](https://coq.inria.fr/library/Coq.Vectors.VectorSpec.html) which seem to be just proofs rather than actual spec. 

Anyone have any pointers on this?

JacquesCarette:
First, pet peeve: what you have are not **matrices** but **arrays**, a rectangular bunch of elements of a type. Matrices are representations, in a basis, of a linear operator. CS has forever confused these, because they forget that they always work over the standard basis, so that in that case, representations of finite linear operators happen to coincide with arrays.

Second, even as arrays, there are many different potential representations for a given API. Maple has `rtable` (for *rectangular table*) to represent this underlying idea. Note that `rtable`s do not promise to be represented via a particular memory representation. They often are contiguous blocks of memory (for the boring cases). A whopping 16 year ago, I wrote [a long message](https://fa.haskell.narkive.com/4p1rg4QK/haskell-cafe-matrix-computations-based-on-the-gsl#post7) to the Haskell list about what we'd done at Maplesoft in this area during 1998-2000. There's some additional information in the [help pages](https://www.maplesoft.com/support/help/maple/view.aspx?path=rtable) for *rtable*.  For example, the 0 matrix and the identity matrix both take O(1) storage because they go through 'indexing functions' that preprocess indexing before lookup. This also how things like sparse, symmetric, block, etc arrays are all handled uniformly and efficiently.

So the first thing to decide is whether you're doing an array library or a matrix library. [Don't get me started on the horror of a name that `Vec` is.]  They have radically different APIs.

MatthewDaggitt:
Thanks for the pointers @JacquesCarette and @jespercockx. It looks like I'm vastly unprepared for formalising matrices in an extensible and performant way. As it's not my primary focus at the moment, just something I had in my own code that I thought might be useful, I'll close this PR and open an issue documenting this discussion for any braver souls than I coming this way.

PR #1531, opened 2021-07-02
Add a version of `Function.Related` for the new function hierarchy

WhatisRT:
This includes conversion functions from the old module and is written for compatibility with the old module. Some choices made for this purpose may be questionable, especially the use of `⤖` (Bijection) instead of `↔` (Inverse) adds some complexity and may not be the best. I'd be happy to discuss these and make changes as required.

I've based this on 1.6, as I haven't switched to Agda 2.6.2 yet, and 1.7 seems to require that. If there are any incompatibilities, I could resolve them once 2.6.2 lands in nixpkgs.


src/Function/Bundles.agda
```diff
@@ -420,3 +421,55 @@ module _ {A : Set a} {B : Set b} where
...
+------------------------------------------------------------------------
+-- Conversion functions
+
+module _ (A : Setoid a ℓ₁) (B : Setoid b ℓ₂) where
+
+  Inverse⇒Bijection : Inverse A B → Bijection A B
+  Inverse⇒Bijection record { f = f ; f⁻¹ = f⁻¹ ; cong₁ = cong₁ ; cong₂ = cong₂ ; inverse = inverse } = record
+    { f =         f
+    ; cong =      cong₁
+    ; bijective = (λ {x} {y} x≈y → begin
```
MatthewDaggitt, line 434: The proofs of the relationship between the individual properties (e.g. `Bijective`, `Inverseᵇ`) rather than bundles (i.e. `Bijection`, `Inverse`) should all live in `Function.Consequences` (doesn't currently exist).

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,290 @@
...
+open import Function.Properties.Inverse
+import Function.Construct.Symmetry as Symmetry
+import Function.Construct.Identity as Identity
+import Function.Construct.Composition as Composition
+
+open import Data.Product
+
+infix 4 _∼[_]_
+
+_∼[_]_ : ∀ {ℓ₁ ℓ₂} → Set ℓ₁ → R.Kind → Set ℓ₂ → Set _
```
MatthewDaggitt, line 34: As this file is going to be the one that we recommend using going forwards, then I think we should move all the shared definitions, (e.g. `Kind`) to _this_ file and then re-export them publicly from `Function.Related` to maintain backwards compatibility.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,290 @@
...
+A ∼[ R.bijection           ] B = A ⤖ B
+
+-- A non-infix synonym.
+
+Related : R.Kind → ∀ {ℓ₁ ℓ₂} → Set ℓ₁ → Set ℓ₂ → Set _
+Related k A B = A ∼[ k ] B
+
+module _ {ℓ₁ ℓ₂} {A : Set ℓ₁} {B : Set ℓ₂} where
+
+  toRelated : {K : R.Kind} → A ∼[ K ] B → A R.∼[ K ] B
```
MatthewDaggitt, line 51: I think these conversions should live in the soon-to-be deprecated file `Function.Related`.

MatthewDaggitt (changes requested):
Hi @WhatisRT thank you very much for the PR, it's greatly appreciated! All the proofs are great, but I think their locations need a little bit of rejigging to keep everything consistent with the rest of the library. I've added some comments describing where I think they should live. Please do let me know if you'd like me to clarify any of them?

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,290 @@
...
+A ∼[ R.left-inverse        ] B = A ↪ B
+A ∼[ R.surjection          ] B = A ↠ B
+A ∼[ R.bijection           ] B = A ⤖ B
+
+-- A non-infix synonym.
+
+Related : R.Kind → ∀ {ℓ₁ ℓ₂} → Set ℓ₁ → Set ℓ₂ → Set _
+Related k A B = A ∼[ k ] B
+
+module _ {ℓ₁ ℓ₂} {A : Set ℓ₁} {B : Set ℓ₂} where
```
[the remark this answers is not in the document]
WhatisRT: This would be a bit annoying, as we pattern-match on the `Kind` here, and because it's an implicit argument we'd have to add four `{_}`'s to every branch. The anonymous module seems like a cleaner solution to me.

src/Function/Bundles.agda
WhatisRT: I'll move this one, but I don't think I can move the proof in `Bijection⇒Inverse` right now without a bunch of extra work. Doing this direction cleanly would probably require a version of `Inverseᵇ` for setoid morphisms.

src/Function/Bundles/Related.agda
WhatisRT: Makes sense, I'll do so!

src/Function/Bundles/Related.agda
WhatisRT: I'll move this as well!

src/Function/Bundles/Related.agda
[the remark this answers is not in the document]
MatthewDaggitt: You'll be happy to learn you can name anonymous arguments to avoid exactly this problem, e.g. `toRelated {K = R.implication}` :smile:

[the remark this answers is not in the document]
WhatisRT: I think I remember trying this in the past in other situations and in didn't work, but it could have been that I didn't use spaces... Thanks!

WhatisRT:
One problem I've noticed is that the way `Function.Bundles.Related` is constructed right now will not generalize to setoids, which I think would be nice to have. The issue is that some constructions only work for def. equality, because properties like `Surjective` don't respect the equivalence in the setoid (meaning that the inverse one can construct will not be a setoid map in general).

One way to fix this would be to rejuggle definitions like `Kind` and `_∼[_]_` and all the specialized versions, and probably adding some kinds. This does of course have the downside that the conversion functions for `Relation.Bundles` would become a bit more complicated, so I don't really know when a good time to do this would be.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+-- Relatedness
+
+-- There are several kinds of "relatedness".
+
+-- The idea to include kinds other than equivalence and bijection came
+-- from Simon Thompson and Bengt Nordström. /NAD
+
+data Kind : Set where
+  implication         : Kind
+  reverse-implication : Kind
```
MatthewDaggitt, line 44: We should change all the instances of kebab case to camel case to be consistent with the rest of the library. For the kinds, we should rename them when opening this module in the old file.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+open import Function.Properties.Bijection
+open import Function.Properties.Surjection
+open import Function.Properties.Inverse
+open import Function.Properties.RightInverse
+
+import Function.Construct.Symmetry as Symmetry
+import Function.Construct.Identity as Identity
+import Function.Construct.Composition as Composition
+
+open import Data.Product
```
MatthewDaggitt, line 32: Any reason this import is living by itself?

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+  surjection          : Kind
+  bijection           : Kind
+
+-- Interpretation of the codes above. The code "bijection" is
+-- interpreted as Inverse rather than Bijection; the two types are
+-- equivalent.
+
+infix 4 _∼[_]_
+
+_∼[_]_ : ∀ {ℓ₁ ℓ₂} → Set ℓ₁ → Kind → Set ℓ₂ → Set _
```
MatthewDaggitt, line 58: We usually use the `a` and `b` for levels of sets `A` and `B`. We could also make them `variable`s to improve readability.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+A ∼[ bijection           ] B = A ⤖ B
+
+-- A non-infix synonym.
+
+Related : Kind → ∀ {ℓ₁ ℓ₂} → Set ℓ₁ → Set ℓ₂ → Set _
+Related k A B = A ∼[ k ] B
+
+-- The bijective equality implies any kind of relatedness.
+
+⤖⇒ : ∀ {k x y} {X : Set x} {Y : Set y} →
```
MatthewDaggitt, line 75: We should stick with `A` and `B` to be consistent rather than switching to `X` and `Y`.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+⇒← {reverse-injection}   = Injection.f
+⇒← {left-inverse}        = RightInverse.g
+⇒← {surjection}          = RightInverse.f ∘ ↠⇒↪
+⇒← {bijection}           = Inverse.f⁻¹ ∘ ⤖⇒↔
+
+-- Kinds whose interpretation include functions going in both
+-- directions.
+
+data Equivalence-kind : Set where
+    equivalence  : Equivalence-kind
```
MatthewDaggitt, line 173: Should be indented 2 rather than 4 spaces.

src/Function/Bundles/Related.agda
```diff
@@ -0,0 +1,399 @@
...
+
+K-preorder : Kind → (ℓ : Level) → Preorder _ _ _
+K-preorder k ℓ = record { isPreorder = K-isPreorder k ℓ }
+
+------------------------------------------------------------------------
+-- Equational reasoning
+
+-- Equational reasoning for related things.
+
+module EquationalReasoning where
```
MatthewDaggitt, line 301: The other reasoning APIs have come a long way since this one was created. We can probably use `Relation.Binary.Reasoning.Base.Double` as a base module for this? And should update the style accordingly.

src/Function/Consequences.agda
```diff
@@ -0,0 +1,35 @@
...
+open import Relation.Binary
+open import Level
+open import Data.Product
+import Relation.Binary.Reasoning.Setoid as SetoidReasoning
+
+private
+  variable
+    a b ℓ₁ ℓ₂ : Level
+
+module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
```
MatthewDaggitt, line 22: I don't think `To` needs to be a setoid right? It could just be a relation?

src/Function/Consequences.agda
```diff
@@ -0,0 +1,35 @@
...
+
+  open Setoid From using () renaming (Carrier to A; _≈_ to _≈₁_)
+  open Setoid To   using () renaming (Carrier to B; _≈_ to _≈₂_)
+  open FunctionDefinitions _≈₁_ _≈₂_
+
+  Inverseᵇ⇒Bijective : ∀ {f f⁻¹} → FunctionDefinitions.Congruent _≈₂_ _≈₁_ f⁻¹ → Inverseᵇ f f⁻¹ → Bijective f
+  Inverseᵇ⇒Bijective {f} {f⁻¹} cong₂ (invˡ , invʳ) = (λ {x} {y} x≈y → begin
```
MatthewDaggitt, line 29: We could split this proof out into two separate proofs of injective and surjective right? The name of the proof should also be lowercase i.e. `inverseᵇ⇒bijective`.

src/Function/Consequences.agda
```diff
@@ -0,0 +1,35 @@
...
+                    x         ≈˘⟨ invʳ x ⟩
+                    f⁻¹ (f x) ≈⟨ cong₂ x≈y ⟩
```
MatthewDaggitt, line 31: The equational reasoning should only be indented 2 spaces from the function definition.

src/Function/Properties/Bijection.agda
```diff
@@ -0,0 +1,52 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Conversions for bijections
+--   This file is meant to be imported qualified.
```
MatthewDaggitt, line 5: Why is this module supposed to be imported qualified out of interest?

src/Function/Properties/Inverse.agda
```diff
@@ -42,3 +45,26 @@ isEquivalence = record
...
   ; trans = Composition.inverse
   }
+
+------------------------------------------------------------------------
+-- Conversion functions
+
+module _ (A : Setoid a ℓ₁) (B : Setoid b ℓ₂) where
+
+  Inverse⇒Bijection : Inverse A B → Bijection A B
+  Inverse⇒Bijection record { f = f ; f⁻¹ = f⁻¹ ; cong₁ = cong₁ ; cong₂ = cong₂ ; inverse = inverse } = record
```
MatthewDaggitt, line 55: Instead of pattern matching on the record like this, it's usually neater to simply name it `I` and append a `where open Inverse I` at the end of the proof.

src/Function/Bundles.agda
```diff
@@ -21,6 +21,7 @@ module Function.Bundles where
 
 import Function.Definitions as FunctionDefinitions
 import Function.Structures as FunctionStructures
+import Relation.Binary.Reasoning.Setoid as SetoidReasoning
```
MatthewDaggitt, line 24: No need to add this as an import now

MatthewDaggitt (commented):
Thanks @WhatisRT, those changes look great. I'm thinking that while we're here we should tidy up the code to bring it more into line with the style of the rest of the library. I've added some more comments where I think it can be improved. No pressure to do them all, feel free to take the ones you're comfortable with and then drop me a line and I can do the rest.

> One way to fix this would be to rejuggle definitions like Kind and _∼[_]_ and all the specialized versions, and probably adding some kinds. This does of course have the downside that the conversion functions for Relation.Bundles would become a bit more complicated, so I don't really know when a good time to do this would be.

Hmm so I think the best course is to rename `Function.Bundles.Related` to `Function.Related.Propositional` which leaves room for `Function.Related.Setoid`? If you're keen then you can have a go at generalising it to `Setoid`s after we've merged this PR in?

src/Function/Properties/Bijection.agda
WhatisRT: No reason, I just copied some stuff around and accidentally left that there

WhatisRT:
I marked all the things I've just done as resolved

src/Function/Consequences.agda
WhatisRT: Do you mean that the requirement that it is an equivalence relation can be dropped?

WhatisRT:
A quick search revealed that nothing in the stdlib seems to import `Function.Related`, should I add a deprecation warning to the module?

src/Function/Bundles/Related.agda
WhatisRT: Because one can reason with inverses and bijections, `Triple` seems appropriate. However, these modules all seem to have the interpretation that the extra relations are not symmetric. I don't know if this would cause any trouble though.

WhatisRT: Actually, it seems to me like the most natural form of this would be some form of `n-ary` reasoning, with a bunch of equivalence relation and orderings and implications between them. I'm not sure if that's possible though.

MatthewDaggitt:
Hi @WhatisRT, thanks for bringing it up to date. I've tried to merge this in to bring the PR in line with the current master, but it looks like I don't have the permissions to do it. Any chance you could enable those permissions as described here?

https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

WhatisRT:
I don't seem to have that button, probably because it's not a user-owned repository. I could either try to resolve the conflicts on my side (which may result in broken code if I can't get my hands on agda 2.6.2), or you could make a branch off of 1.6 for me to merge it into

MatthewDaggitt:
Ah yes, it looks like a company owned repo? Okay, so I think the simplest thing might be to simply open a new PR with the same commits, sourced from this repo. You'd still get the credit for the commits. Would this be okay with you?

WhatisRT:
Sure, that's fine! Next time I'll have a bigger PR with potential conflicts I think I'll make a personal repo so that this issue won't appear again.

MatthewDaggitt:
Continued in #1561

PR #1532, opened 2021-07-04
Add Data.Product.Properties.map-cong₂

Taneb:


src/Data/Product/Properties.agda
```diff
@@ -53,6 +55,9 @@ module _ {B : A → Set b} where
 ,-injective : ∀ {a c : A} {b d : B} → (a , b) ≡ (c , d) → a ≡ c × b ≡ d
 ,-injective refl = refl , refl
 
+map-cong₂ : ∀ {f g : A → C} {h i : B → D} → f ≗ g → h ≗ i → map f h ≗ map g i
+map-cong₂ p q (x , y) = cong₂ _,_ (p x) (q y)
```
MatthewDaggitt, line 59: Very minor bikeshedding but could we rename `p` and `q` to `f≗g` and `h≗i` instead?

src/Data/Product/Properties.agda
```diff
@@ -53,6 +55,9 @@ module _ {B : A → Set b} where
 ,-injective : ∀ {a c : A} {b d : B} → (a , b) ≡ (c , d) → a ≡ c × b ≡ d
 ,-injective refl = refl , refl
 
+map-cong₂ : ∀ {f g : A → C} {h i : B → D} → f ≗ g → h ≗ i → map f h ≗ map g i
```
MatthewDaggitt, line 58: Hmm I think this should be called just `map-cong`? And then the subscripts can be reserved for mapping over either the first or second projections?

Taneb: I think you might be right. Data.Sum.Properties uses that naming convention.

src/Data/Product/Properties.agda
Taneb: Sure

Issue #1486, opened 2021-04-24
Should `_$!_` be moved out of `Function.Base` ?


JacquesCarette:
Perhaps into `Function.Strict`, which could also be imported from `Function`. It's the only function that uses `Strict`, and it is rather rarely used, so this would help simplify the dependency graph.

gallais:
Can we use this opportunity to add `_!|>_`?

MatthewDaggitt:
> It's the only function that uses Strict, and it is rather rarely used, so this would help simplify the dependency graph.

`Strict` doesn't have many import dependencies, so this would only remove one module from the dependency graph right? Just checking I haven't misunderstood?

To be honest, I would like to consolidate everything and move the entire contents of `Strict` to `Function.Strict` **and** move all the strict combinators currently in `Function.Base` to `Function.Strict`.

This isn't a backwards compatible change though so I might be tempted to wait until version 2.0?

JacquesCarette:
It's a minor thing, yes. I was going to try to be a bit more systematic about exploring the lower parts of the dependency graph, but I only had time to get started.

We could actually move `_$!_` to `Function.Strict` now, as that would be a compatible change if that's its eventual home anyways. It could be done in a non-breaking (but non-improving too) way, or as a slight break.

MatthewDaggitt:
Yup, sounds good. A PR doing the following would be welcome:
 - Deprecate `Strict` and move contents to `Function.Strict`
 - Move `_$!_` from `Function.Base` to `Function.Strict`
 - Add `_!|>_` in `Function.Strict` as a special favour for @gallais 
 - Re-export `_$!_` from `Function.Strict` publicly in `Function.Base` to maintain backwards compatibility.

In v2.0 we can then cut the dependency between `Function.Base` and `Function.Strict` by removing the public export of `_$!_`.

MatthewDaggitt:
To-do in v2.0
 - reverse the dependency between `Function.Strict` and `Function.Base` 
 - can then use `flip` more extensively in `Function.Strict`

PR #1533, opened 2021-07-05
Move `Strict` to `Function.Strict` and move `_$!_` there

MatthewDaggitt:
In response to #1486. Haven't yet cut the dependency of `Function.Base` on `Function.Strict` in order to preserve backwards compatibility.

### DIFF
--- a/src/Strict.agda
+++ b/src/Strict.agda
@@ -1,31 +1,20 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Strictness combinators
+-- This module is DEPRECATED, please use `Data.Record` directly.
 ------------------------------------------------------------------------
 
 {-# OPTIONS --without-K --safe #-}
 
 module Strict where
 
-open import Level
-open import Agda.Builtin.Equality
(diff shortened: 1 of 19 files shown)


MatthewDaggitt:
Question - should `Function` re-export `Function.Strict`?

MatthewDaggitt:
> Question - should Function re-export Function.Strict?

In the absence of any comment, I've gone with "yes".

src/Strict.agda
```diff
@@ -1,31 +1,20 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Strictness combinators
+-- This module is DEPRECATED, please use `Data.Record` directly.
```
JacquesCarette, line 4: `Data.Record` ?

JacquesCarette (changes requested):
Other than a minor typo, this is great.

src/Strict.agda
MatthewDaggitt: Thanks, good spot!

PR #1536, opened 2021-07-05
[ new ] levelOf

gallais:


MatthewDaggitt:
What do you think about putting this in `Level`? I'm thinking it'll probably be more discoverable when someone is looking for it?

gallais:
I'm opening `Level` and...

```agda
levelOfType : ∀ {a} → Set a → Level
levelOfType {a} _ = a

levelOfTerm : ∀ {a} {A : Set a} → A → Level
levelOfTerm {a} _ = a
```

:rofl:

PR #1540, opened 2021-07-06
Remove redundancy in ring morphisms etc

### LINKED ISSUE #1539: Algebra.Morphism.Structures: redundant fields in ring-like morphisms
Consider the `IsRingHomomorphism` structure. It winds up having two different, not-necessarily-equal definitions of `IsRelHomomorphism _≈₁_ _≈₂_ ⟦_⟧`, one via the proof that addition is a group homomorphism and one via the proof that multiplication is a monoid homomorphism.

I'm working on a fix for this that defines, for example, a ring homomorphism as a semiring homomorphism plus a proof that negation is homomorphic.

Taneb:
This redefines NearSemiring, Semiring, and Ring homomorphisms so that
they each build up by adding a single new property to a smaller
structure. This ensures there's only one proof that they're a Rel
homomorphism. For the same reason, Lattice homomorphisms are defined
directly rather than showing it's two magma homomorphisms.

I've tried to make it so that each of these structures provides any
smaller structure at its top level, so if we have a RingMonomorphism we
can easily say that it's a MagmaHomomorphism by multiplication or
whatever, but I might have missed one or two.

Closes #1539

### DIFF
--- a/src/Data/Rational/Properties.agda
+++ b/src/Data/Rational/Properties.agda
@@ -785,6 +785,23 @@ private
   ; _⁻¹ = -_
   }
 
++-*-rawNearSemiring : RawNearSemiring 0ℓ 0ℓ
++-*-rawNearSemiring = record
+  { _≈_ = _≡_
+  ; _+_ = _+_
+  ; _*_ = _*_
+  ; 0#  = 0ℚ
+  }
+
++-*-rawSemiring : RawSemiring 0ℓ 0ℓ
++-*-rawSemiring = record
+  { _≈_ = _≡_
+  ; _+_ = _+_
+  ; _*_ = _*_
+  ; 0#  = 0ℚ
+  ; 1#  = 1ℚ
+  }
+
 +-*-rawRing : RawRing 0ℓ 0ℓ
 +-*-rawRing = record
   { _≈_ = _≡_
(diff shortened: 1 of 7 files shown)


src/Data/Rational/Properties.agda
```diff
@@ -1068,10 +1085,36 @@ toℚᵘ-isMonoidMonomorphism-* = record
...
+  { isSemiringHomomorphism = toℚᵘ-isSemiringHomomorphism-+-*
+  ; injective              = toℚᵘ-injective
+  }
+
 toℚᵘ-isRingHomomorphism-+-* : IsRingHomomorphism +-*-rawRing ℚᵘ.+-*-rawRing toℚᵘ
 toℚᵘ-isRingHomomorphism-+-* = record
-  { +-isGroupHomomorphism  = toℚᵘ-isGroupHomomorphism-+
-  ; *-isMonoidHomomorphism = toℚᵘ-isMonoidHomomorphism-*
+  -- { +-isGroupHomomorphism  = toℚᵘ-isGroupHomomorphism-+
+  --; *-isMonoidHomomorphism = toℚᵘ-isMonoidHomomorphism-*
```
MatthewDaggitt, line 1115: Comments can be removed?

MatthewDaggitt (commented):
Looks like a much better design, thank you! Only thing missing is a CHANGELOG entry :smile:

MatthewDaggitt:
@Taneb given the number of great PRs recently, I've also added you to the list of Agda developers. You should now have permission to create PRs on this repo directly if you want to :+1:

src/Data/Rational/Properties.agda
Taneb: Oops, didn't mean to leave them in

Taneb:
I'm going to make a subsequent PR adding modules like `Algebra.Morphism.NearSemiringMonomorphisms` and proving that Rationals satisfy some more intermediate algebraic structures, because I've added some of the `RawX` bundles, but this PR is already bigger than I set out to do so it can wait a little bit.

PR #1541, opened 2021-07-07
add `diagonal` and some properties about `Vec`

damhiya:
I added some function and properties about `Vec`.
The `diagonal` function is a join for monad instance of `Vec` (that is compatible for zip applicative instance of `Vec`), so I added it to `Data.Vec.Categorical`
Though it seems correct monad instance, I couldn't proof that `diagonal` satisfies associative law for monad in agda yet.

### DIFF
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -167,6 +167,17 @@ fs ⊛* xs = fs >>= λ f → map f xs
 allPairs : ∀ {m n} → Vec A m → Vec B n → Vec (A × B) (m * n)
 allPairs xs ys = map _,_ xs ⊛* ys
 
+-- Diagonal
+
+diagonal : ∀ {n} → Vec (Vec A n) n → Vec A n
+diagonal [] = []
+diagonal (xs ∷ xss) = head xs ∷ diagonal (map tail xss)
+
+infixl 1 _>>=′_
+
+_>>=′_ : ∀ {n} → Vec A n → (A → Vec B n) → Vec B n
+xs >>=′ f = diagonal (map f xs)
+
 ------------------------------------------------------------------------
 -- Operations for reducing vectors
 
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -154,15 +154,16 @@ _⊛_ : ∀ {n} → Vec (A → B) n → Vec A n → Vec B n
 
 -- Multiplication
 
-infixl 1 _>>=_
+module Product where
+  infixl 1 _>>=_
 
-_>>=_ : ∀ {m n} → Vec A m → (A → Vec B n) → Vec B (m * n)
-xs >>= f = concat (map f xs)
+  _>>=_ : ∀ {m n} → Vec A m → (A → Vec B n) → Vec B (m * n)
+  xs >>= f = concat (map f xs)
 
 infixl 4 _⊛*_
 
 _⊛*_ : ∀ {m n} → Vec (A → B) m → Vec A n → Vec B (m * n)
-fs ⊛* xs = fs >>= λ f → map f xs
+fs ⊛* xs = fs Product.>>= λ f → map f xs
 
 allPairs : ∀ {m n} → Vec A m → Vec B n → Vec (A × B) (m * n)
 allPairs xs ys = map _,_ xs ⊛* ys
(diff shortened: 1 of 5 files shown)


src/Data/Vec/Base.agda
```diff
@@ -167,6 +167,17 @@ fs ⊛* xs = fs >>= λ f → map f xs
...
 
+-- Diagonal
+
+diagonal : ∀ {n} → Vec (Vec A n) n → Vec A n
+diagonal [] = []
+diagonal (xs ∷ xss) = head xs ∷ diagonal (map tail xss)
+
+infixl 1 _>>=′_
+
+_>>=′_ : ∀ {n} → Vec A n → (A → Vec B n) → Vec B n
```
gallais, line 178: Any reason not to call this `_>>=_` like we do in e.g. `Data.Maybe.Base`?

damhiya: There is already function called `_>>=_` in `Data.Vec.Base` defined as cartesian product. i.e. similar to `_>>=_` for `List`.
This is actually not a monad for `Vec`. However, agda's do-notation works syntactically. So this naming might be pretty useful.
Since diagonal-based one is genuine definition of monad for `Vec`, I also think the old `_>>=_` definition should yield it's name to new one, but it will cause breaking change.

src/Data/Vec/Properties.agda
```diff
@@ -364,6 +364,10 @@ map-id : ∀ {n} → map {A = A} {n = n} id ≗ id
 map-id []       = refl
 map-id (x ∷ xs) = P.cong (x ∷_) (map-id xs)
 
+map-const : ∀ {n} (xs : Vec A n) (x : B) → map {n = n} (const x) xs ≡ replicate x
```
MatthewDaggitt, line 367: Very minor comment, but could we rename the constant value from `x`, e.g. `v`. Makes it clearer that it's not an element of the list `xs`...

src/Data/Vec/Properties.agda
```diff
@@ -695,6 +713,19 @@ map-replicate :  ∀ (f : A → B) (x : A) n →
 map-replicate f x zero = refl
 map-replicate f x (suc n) = P.cong (f x ∷_) (map-replicate f x n)
 
+transpose-replicate : ∀ {m n} (xs : Vec A m) → transpose (replicate {n = n} xs) ≡ map replicate xs
+transpose-replicate {n = zero} _ = P.sym (map-const _ [])
+transpose-replicate {n = suc n} xs = begin
+  transpose (replicate xs)
```
MatthewDaggitt, line 719: I think best to lay this out in the usual style, even if it violates the line length (which is more of a convention anyway)

src/Data/Vec/Properties.agda
```diff
@@ -389,6 +393,10 @@ map-[]≔ : ∀ {n} (f : A → B) (xs : Vec A n) (i : Fin n) {x : A} →
           map f (xs [ i ]≔ x) ≡ map f xs [ i ]≔ f x
 map-[]≔ f xs i = map-updateAt xs i refl
 
+map-⊛ : ∀ {n} (f : A → B → C) (g : A → B) (xs : Vec A n) → (map f xs ⊛ map g xs) ≡ map (f ˢ g) xs
```
MatthewDaggitt, line 396: Maybe put the last term on the next line?

MatthewDaggitt:
Any chance you could add a CHANGELOG entry as well @damhiya ?

src/Data/Vec/Base.agda
MatthewDaggitt: Hmm, so if we have two different possible definitions for the bind operator, and both require to be named `_>>=_` to work with do-notation we appear to be in trouble. Do notation doesn't work with `_>>=′_` right? So maybe, both `_>>=_` and this should be called something else more meaningful and the user can decide?

 I'm really not sure, and would appreciate other people's thoughts...

> but that's a breaking change.
We're currently working on v2.0 which can have breaking changes, but we want to be sure we're making the _right_ breaking changes!

damhiya: Maybe we can just define a sub module `Zip` and `Cartesian` so that each module has their own definition of `_>>=_`.

MatthewDaggitt: That seems like an excellent suggestion :+1:

damhiya:
I chose the name `Product` and `Diagonal` rather than `Cartesian` and `Zip`.
And totally removed definition of `Data.Vec.Base._>>=_` instead of deprecating it, since that will cause name collision if we open `Data.Vec.Base` publicly.

damhiya:
I'm not sure whether it's good to use sub module here instead of making new file.

MatthewDaggitt:
Thanks, we're definitely getting there! The only problem with naming the module `Product` is that it might conflict with `import Data.Product as Product` which is a relatively common idiom. Perhaps naming the module `ProductBind` and `DiagonalBind` might be both less likely to break code and improve readability?

gallais:
I liked `Cartesian` and `Zip` :(

damhiya:
> I liked `Cartesian` and `Zip` :(

Okay I'll back to `Cartesian`, but I think `Diagonal` is more appropriate than `Zip`. Coinciding Applicative instance for `Diagonal` is `Zip`, but the monad instance `Diaognal` is derived from `diagonal` operation.

MatthewDaggitt:
Great, glad we've reached consensus! Merged in. Thanks again for the PR @damhiya

damhiya:
@MatthewDaggitt
It seems the PR closed without merging. Is this intended?

MatthewDaggitt:
Apologies, yes! Misclicked. Nice spot

PR #1548, opened 2021-07-12
Simplify definition of Data.Nat.Properties.<⇒≤

dsheets:
I don't see any benefit to the definition with transitivity and reflexivity.


MatthewDaggitt:
Thanks for the PR, nice spot!

Issue #1542, opened 2021-07-09
Adding priority queue implementation


damhiya:
Hi.
I implemented [leftist heap based priority queue](https://github.com/damhiya/AgdaHeap/blob/master/Tree/WBLT/Base.agda) in agda.
So I'm going to contribute this code to agda-stdlib if such thing was desired.

However, this code is a bit messy and I didn't care so much about library design while writing it.
I would appreciate if someone give a comment about the code.

MatthewDaggitt:
Hi @damhiya, thanks for opening the issue, yes we'd definitely be interested in a priority queue implementation!

However, the first thing we'd want to do I think is to define an interface (operations + laws) for priority queues. We'd then show that your implementation satisfied that interface.  Would you be interested/willing to put together such an interface?

damhiya:
> Hi @damhiya, thanks for opening the issue, yes we'd definitely be interested in a priority queue implementation!
> 
> However, the first thing we'd want to do I think is to define an interface (operations + laws) for priority queues. We'd then show that your implementation satisfied that interface. Would you be interested/willing to put together such an interface?

Of course!

MatthewDaggitt:
Great! If you'd like to open a PR with your proposed interface (probably in `Data.PriorityQueue.Interface`?) that'd be the best course. Once we've got the interface nailed down and merged in, we can then open a second PR with your implementation :+1:

MatthewDaggitt:
See feedback in #1552 for a design discussion.

damhiya:
I'm really sorry for inactivity. I was doing other works. I'll work on this in the future unless someone do it.

MatthewDaggitt:
Absolutely no worries, not everything has to be added. Our discussion alone has really helped clarify my thoughts on this, so thanks!

andreasabel:
Btw, I played around with a leftist heap implementation in Agda and ended up with this: https://gist.github.com/andreasabel/352fee52a39c0bebbd5059bff86d9b6e

gallais:
We do have a generic min construction: http://agda.github.io/agda-stdlib/Algebra.Construct.NaturalChoice.Min.html
And also the extension of an existing total order with a supremum: http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Add.Supremum.NonStrict.html

Taneb:
Here's a different implementation I made a while back, using pairing heaps: https://gist.github.com/Taneb/7e502c708e2e6eabcea5e7638f42b593

[...]

PR #1552, opened 2021-07-15
PriorityQueue interface

damhiya:
Hi I wrote an interface of priority queue as discussed in https://github.com/agda/agda-stdlib/issues/1542.

What I focused was describing soundness of priority queue with small number of fields as possible,
so that not to blowing up record size with declarations of consistency between fields.
For example, I defined `toList` using `popMin`, and used permutation relation over result of `toList` to
guarantee correctness of queue implementation.

If someone has better design idea, please notice me.
I'll also implement instance for `PriorityQueue` soon, in order to convince interface definition.

### DIFF
--- a/src/Data/Maybe/Properties.agda
+++ b/src/Data/Maybe/Properties.agda
@@ -31,6 +31,9 @@ private
 ------------------------------------------------------------------------
 -- Equality
 
+just≢nothing : ∀ {x} → (Maybe A ∋ just x) ≢ nothing
+just≢nothing ()
+
 just-injective : ∀ {x y} → (Maybe A ∋ just x) ≡ just y → x ≡ y
 just-injective refl = refl
 
--- a/src/Data/PriorityQueue.agda
+++ b/src/Data/PriorityQueue.agda
@@ -1,0 +1,11 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Priority queue
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.PriorityQueue where
+
+open import Data.PriorityQueue.Base public
--- a/src/Data/PriorityQueue/Base.agda
+++ b/src/Data/PriorityQueue/Base.agda
@@ -1,0 +1,144 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Interface definition of priority queue
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary using (TotalOrder)
+
+module Data.PriorityQueue.Base
+  {a ℓ₁ ℓ₂} (totalOrder : TotalOrder a ℓ₁ ℓ₂) where
+
+open import Data.Empty
+open import Data.Maybe as Maybe
+open import Data.Maybe.Properties
+open import Data.Nat using (ℕ; zero; suc; _+_)
+open import Data.Nat.Properties
+open import Data.Product
+open import Data.List
+open import Data.List.Relation.Binary.Permutation.Propositional
+open import Data.List.Relation.Unary.All using (All)
+open import Function.Base
+open import Level renaming (suc to ℓ-suc)
+open import Relation.Binary
+open import Relation.Binary.PropositionalEquality as P
+open import Relation.Unary hiding (Empty)
+open import Relation.Nullary
+open import Induction.WellFounded
+
+open TotalOrder totalOrder renaming (Carrier to A)
+
+record RawPriorityQueue b ℓ : Set (a ⊔ ℓ-suc (b ⊔ ℓ)) where
+  field
+    Queue  : Set b
+    size   : Queue → ℕ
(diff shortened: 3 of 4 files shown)


damhiya:
Since I used permutation relation of lists defined over propositional equality,
one should think equivalence between `Queue` as a relation defined over `_≡_`,
not underlying equivalence relation of `totalOrder` (`_≈_`).

MatthewDaggitt:
Thanks for putting together a draft @damhiya! It's a fascinating design problem. 

From an implementor's perspective, I can definitely see the attraction of having a thin interface with the other operations derivable. However, from a user's perspective I think a fat interface might be more desirable. 

For example with your definitions, any notion of equality (i.e permutations of `toList` applied to each queue) is going to unfold to a huge jumble of stuff when normalised, which is going to be very difficult to parse to say the least. Equally from a performance perspective, converting to a list by repeatedly removing the minimum is going to go to at least `O(nlogn)` right? I think it would probably be better to expose a `toList` method that can be implemented efficiently underneath the hood?

With a fat interface, yes you're going to get some blowup in the number of proofs required for the implementation but at least that burden is during the implementation (which only happens once) rather than on the user (of which there will be many).

If we do go for a fat interface then I'd probably have at least the following additional fields:
- Equality type + proof of decidability
- Membership type + proof of decidability
- `toList`

What are your thoughts?

MatthewDaggitt:
Hmm from a performance perspective, maybe we should have a separate `findMin` as well which can be `O(1)` instead of the `O(log n)` of  `popMin`?

damhiya:
In my opinion, ideal solution is providing minimal, but enough operations so that one can derive all general properties of priority queue without having specific instance of priority queue implementation. What I mean by general properties are properties which is true for any implementation of priority queue. Then, we should prove such properties parametrically, not to prove such things for multiple implementation of priority queues.

So I'm looking forward that users may use such proofs constructed on top of minimal interface. However, as you said, it might not very convenient due to complex permutation related things. So I think this decision should made after testing some application of priority queues.

Current implementation of `toList` is pretty important, not only because it's a heap sort, but it is a connection between whole element of priority queue and result of `popMin`. If we provide another function `elems` that return all the elements of the queue without sorting it, then we need `toList` anyway to convince repeated use of `popMin` will give the same result. I agree that it would be useful to have `elems` when we don't need sorted list, but wants efficient function. However, on proving aspects, I don't see much reason to having `elems`. If we're working on specific implementation, we can just use `elems` defined for that implementation. When we're working parametrically, since we don't have the definition of `elems`, there is no reason to prefer `elems` over `toList`.

damhiya:
> Hmm from a performance perspective, maybe we should have a separate `findMin` as well which can be `O(1)` instead of the `O(log n)` of `popMin`?

I thought one may implement `popMin` lazily,  so make it works `O(1)` if we demand only for first item. But I'm not sure about evaluation strategy of agda.

MatthewDaggitt:
Apologies for taking so long to reply, I've been mulling this over in my head.

> In my opinion, ideal solution is providing minimal, but enough operations so that one can derive all general properties of priority queue without having specific instance of priority queue implementation. What I mean by general properties are properties which should be true for any implementation of priority queue. Then, we should prove such properties parametrically, not to prove such things for multiple times.

So what are the advantages of this solution for the _users_ of the standard library? I understand that from the _implementor_'s perspective it's desirable to have a minimal interface so as to only have to prove the properties once. I ask, because the point of the standard library is to make the life of the users as easy possible, not the lift of the implementors.

> So I'm looking forward that users may use such proofs constructed on top of minimal interface. However, as you said, it might not very convenient due to complex permutation related things. I think this decision should be made after examining some application of priority queues.

Hmm having tried it out, the type messages are still pretty bad even when we're considering an abstract definition. I'm getting things like:
```
x ∷
      SizeLaws.toListAux sizeLaws (meld p q)
      (Data.PriorityQueue.Base.SizeLaws.Size⇒Acc sizeLaws
       (size (meld p q)) (meld p q) refl)
      ≡
      SizeLaws.toListAux sizeLaws (insert x p)
      (Data.PriorityQueue.Base.SizeLaws.Size⇒Acc sizeLaws
       (size (insert x p)) (insert x p) refl)
      ++
      SizeLaws.toListAux sizeLaws q
      (Data.PriorityQueue.Base.SizeLaws.Size⇒Acc sizeLaws (size q) q
       refl)
```
Having `toList` as a primitive would make this into:
```
x ∷ toList (meld p q) ≡ toList (insert x p) ++ toList q
```
The situation is going to get much worse once you're using a concrete backing implementation.


> Current implementation of toList is pretty important, not only because it's a heap sort, but it is a connection between whole element of priority queue and result of popMin. If we provide another function elems that return all the elements of the queue without sorting it, then we need toList anyway to convince repeated use of popMin will give the same result.

I'm not suggesting removing the current definition of `toList` merely renaming it something like `popAll` and making it part of the derived definitions, rather than using it as part of the laws.

MatthewDaggitt:
Marking this as abandoned as I don't think we're going to make progress on it.

PR #1556, opened 2021-07-19
[ new ] Data.Default

gallais:
Essentially:

```agda
  -- `inc` increments its argument by the value `step`, defaulting to 1
  inc : {{step : WithDefault 1}} → Nat → Nat
  inc {{step}} n = step .value + n

  -- and indeed incrementing 2 gives you 3
  _ : inc 2 ≡ 3
  _ = refl

  -- but you can also insist that you want to use a bigger increment by
  -- passing the `step` argument explicitly
  _ : inc {{10 !}} 2 ≡ 12
  _ = refl
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,15 +69,15 @@ Non-backwards compatible changes
 
 ### Strict functions
 
-* The module `Strict` has been deprecated in favour of `Function.Strict` 
+* The module `Strict` has been deprecated in favour of `Function.Strict`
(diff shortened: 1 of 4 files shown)


MatthewDaggitt:
Looks great! No comments from me, except maybe that the example should go in a README file instead?

gallais:
I was thinking that it was so small it was maybe not worth the separate README
file but you may be right that it would be better to have it anyways for discoverability
purposes.

Issue #1527, opened 2021-06-22
Names for propositional proofs in `Function.Construct.(Identity/Symmetry/Composition)` are backwards


MatthewDaggitt:
Should be `⤖-sym` rather than `sym-⤖` to be consistent with the rest of the library. The old names should be deprecated.

Akshobhya1234:
F.C.Identity
 ` ↣-id : A ↣ A `
 `↣-id = injection (setoid A)`
F.C.Symmetry
 `⇔-sym : A ⇔ B → B ⇔ A`
`⇔-sym = equivalence`
F.C.Composition 
`↠_-_∘ : A ↠ B → B ↠ C → A ↠ C`
`↠_-_∘ = surjection`

@MatthewDaggitt  Is this the right way to do it?

MatthewDaggitt:
Yes, except composition, which I think should be `_↠-∘_` rather than `↠_-_∘`

PR #1564, opened 2021-07-27
Names for propositional proofs in Function.Construct.X

Akshobhya1234:
This will resolve issue #1527


MatthewDaggitt:
Thanks so much for this! The only things left to do are:
1. add deprecation warnings for the old names so that code using the old names still works and the authors get a nice warning guiding them to the new name, see e.g.
https://github.com/agda/agda-stdlib/blob/27db7955f8abdb640162effb7a81d1eb9e3935aa/src/Data/Nat/Properties.agda#L2026
for how to do this.

2. add an entry to the [CHANGELOG](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG.md) under the `Deprecated names` subheading. See https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.6.md#deprecated-names for an example.

src/Function/Construct/Composition.agda
```diff
@@ -204,28 +204,85 @@ module _ {R : Setoid a ℓ₁} {S : Setoid b ℓ₂} {T : Setoid c ℓ₃} where
...
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 1.7
+
+_∘-⟶_ = _⟶-∘_
+{-# WARNING_ON_USAGE _∘-⟶_
+"Warning: _∘-⟶_ was deprecated in v1.7.
```
MatthewDaggitt, line 244: Perfect, except that it should be `v2.0` everywhere, not `v1.7` :smile:

Akshobhya1234: Oh my bad I will correct it 😄

PR #1567, opened 2021-07-31
Adapted proofs in `Data.Nat` and `Data.Integer` that used the old `suc` trick to use `NonZero`

MatthewDaggitt:
I've gone through all of `Data.Nat` and `Data.Integer` folders looking for proofs which use the old way of ensuring something was `NonZero` (by taking an argument `n` and only using it in the form `suc n`),  and adapted them to use the new `NonZero` instance argument approach.

This has eliminated a lot of places where we were using these proofs and having to use `suc (pred x)` instead of `x` merely to manipulate into the form `suc _`.


JacquesCarette:
Very nice!

PR #1568, opened 2021-07-31
[ new ] GenerateEverything --include-deprecated flag

laMudri:
When `Everything.agda` is used to build the entire contents of the library, we want to include the deprecated modules too, as they are still a usable part of the library. In particular, currently the nixpkgs version of agda-stdlib leaves deprecated modules unusable.


laMudri:
@alexarice and @turion might be interested. Minimal non-working example with the nxipkgs setup:

```agda
import Data.Star
```

MatthewDaggitt:
Thanks for the PR @laMudri!

PR #1569, opened 2021-07-31
Remove everything that was deprecated before v1.0

MatthewDaggitt:
Does what it says on the tin. We can have a separate discussion about whether or not to remove things deprecated between v1.0 and v1.7.


JacquesCarette (approved):
(Don't know if officially 'approving' these is the right thing to do, but I do want to support this effort quite emphatically!)

JacquesCarette:
About discussing things beyond 1.0: probably these should be discussed incrementally, i.e. each version should get a separate issue. For example, the consensus might emerge that up to v1.1 stuff can be removed, but really not v1.2 (say). Then there's no point in creating further discussion issues.

MatthewDaggitt:
Approving it is fine! Always nice to know that people are casting an eye over things :smile: 

> About discussing things beyond 1.0: probably these should be discussed incrementally, i.e. each version should get a separate issue. For example, the consensus might emerge that up to v1.1 stuff can be removed, but really not v1.2 (say). Then there's no point in creating further discussion issues.

Sure. One thing that I should say is that I'm keen that we only remove things on major version numbers, to at least keep up the fiction that we're consistent with semantic versioning. So anything that's not removed in v2.0 will stick around to a far-future v3.0.

PR #1576, opened 2021-08-11
Tidy up `Data.Integer.Properties`

MatthewDaggitt:
Given the file a spring clean in preparation for trying to play the same irrelevant instance argument game with `Positive`, `Negative` etc. 

It looks scary but really I've just done the following:
 - Consistently use `i`, `j`, `k` for integers and `m`, `n`, `o` for naturals as per the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#variables-1). We were using `ijk`/`mno`/`xyz`/`abc` in various places in the file. This was particularly a problem in names of proofs where you couldn't tell whether the name would use `m/n` or `i/j`. Should now be predictable.
 - Use `variable`s for the above
 - Increase usage of -1ℤ, 0ℤ, 1ℤ
 - Increase usage of ≡˘⟨_⟩


JacquesCarette (approved):

MatthewDaggitt:
Merging this in early as it's blocking another bigger PR, and I think it's relatively uncontroversial and it's been approved

Issue #1565, opened 2021-07-29
Explore further improving instance search for `NonZero` instances


MatthewDaggitt:
Now that it looks very likely we're going to use instance arguments for resolving `NonZero` proofs we should explore whether it's possible and/or desirable to automatically create derived instances to further improve the user's experience:

Candidates:
- `NonZero m → NonZero n → NonZero (m * n)`
- `NonZero m → NonZero n → NonZero (m / n)`
- `NonZero p → NonZero (1/ p)`
- `NonZero m → NonZero (gcd m n)`
- `NonZero n → NonZero (gcd m n)`

(the latter two are an example of where one might get multiple instances for the same proof and where irrelevance would be useful)

A list of proofs that would benefit from these:

- `Data.Nat.LCM` - all of it
- `Data.Rational.Unnormalised.Properties` - *-cancelˡ-/ *-cancelʳ-/ 1/-involutive-≡ 1/-involutive
- `Data.Rational.Normalised.Properties` - normalize-coprime, ↥-normalize, ↧-normalize, normalize-cong, normalize-nonNeg, normalize-pos, normalize-injective-≃, 0/n≡0

JacquesCarette:
Question indirectly from this: does Agda allow **irrelevant** instance arguments to be filled by 'any' proof? Normally, instance search wants to find a single candidate for a hole.

MatthewDaggitt:
> Question indirectly from this: does Agda allow irrelevant instance arguments to be filled by 'any' proof? Normally, instance search wants to find a single candidate for a hole.

Making an instance argument irrelevant, still means that Agda has to find a valid proof. However unlike for relevant arguments, instance search won't error if it finds multiple valid proofs.

https://agda.readthedocs.io/en/v2.6.2/language/irrelevance.html#irrelevant-instance-arguments

MatthewDaggitt:
Blocked by agda/agda#5494

MatthewDaggitt:
Likewise with `Positive` etc.

Candidates:
- `NonPositive p → NonNegative (- p)`

Proofs that would benefit:
- `Data.Rational.Unnormalised.Properties`: *-monoˡ-≤-nonPos

MatthewDaggitt:
As pointed out by @technicalguy in https://github.com/agda/agda-stdlib/pull/1581 we should benchmark the library when and if we implement these changes.

jamesmckinna:
[...]

Issue #1175, opened 2020-04-28
generic Nonzero proposal


mechvel:
the master lib of April 28 has NonZero for ℚ and maybe for other special domains.

I suggest to generalizethis to   ``IsNonzero : Pred Carrier _``  
and     ``Nonzero =  Σ Carrier IsNonzero``,

to declare this for arbitrary Semiring. This can be put to Algebra.Properties.Semiring.
Because 
1) IsNonzero "happens" in various semirings, 
2) the Nonzero subset in any semiring is a monoid by ``_*_``, this sub-monoid is often used,
    it is an usual construction applied to a (semi)ring.

Also I have an impression that they write in literature  Nonzero rather than NonZero,
Nonempty rather than NonEmpty.
?

MatthewDaggitt:
I'm happy for you to add the definition to `Agelbra.Properties.Semiring` if you'd like.

> I suggest to generalize this to IsNonzero : Pred Carrier _
and Nonzero = Σ Carrier IsNonzero,

We should not generalise the existing examples in `Nat`, `Integer` and `Rational` however as using them would then rely on first proving that they're a semiring. They therefore would be unable to live in e.g. `Data.Nat.Base`.

> Also I have an impression that they write in literature Nonzero rather than NonZero,

I think `Non-zero` appears the most in the literature, but that doesn't follow the naming conventions. If you feel strongly about it and submit a PR before v1.4 releases then I'm not bothered by you changing it to `Nonzero`.

mechvel:
The generic  ``IsNonzero : Pred Carrier _``  can be defined only as   ``_≉ 0#``.
Because in an arbitrary semiring  it does not hold ``∀ x → 1 + x ≉ 0#``.
So, that  for Nat, there will be the two versions of IsNonzero and Nonzero.
May be, we rename NonZero of lib-1.3 to IsNonzero' ?
As to the `Is' prefix, probably it is needed to distinguish a predicate and a data of kind
``Σ _ IsNonzero'``.

mechvel:
I have committed this PR. 
I hope that it does not override the PR of Binari.DivMod.
I  have entered ``https://github.com/agda/agda-stdlib`` and clicked at Fork.
Then I created the folder  ``.../NonzeroPR/`` on my machine and commanded
```
cd NonzeroPR
git  clone https://github.com/mechvel/agda-stdlib agda-stdlib-fork
...
```
And now there are the folders
```
foo/adga-stlib-fork/           
[...]

PR #1581, opened 2021-08-19
Switch `Positive/Negative/etc.` to use irrelevant instance arguments

MatthewDaggitt:
This plays the same game with `Positive`/`Negative`/`NonPositive`/`NonNegative` as https://github.com/agda/agda-stdlib/pull/1538 did with `NonZero`. We get all the same benefits as the `NonZero` case. Will post something about on Zulip and the mailing list tomorrow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,21 +79,27 @@ Non-backwards compatible changes
 
 * All modules and names that were deprecated prior to v1.0 have been removed.
 
-#### Proofs of non-zeroness as instance arguments
+#### Proofs of non-zeroness/positivity/negativity as instance arguments
 
-* Many numeric operations in the library require their arguments to be non-zero.
-  The previous way of constructing and passing round these proofs resulted in clunky code.
-  As described on the [mailing list](https://lists.chalmers.se/pipermail/agda/2021/012693.html)
-  we have converted the operations to take the proofs of non-zeroness as irrelevant 
-  [instance arguments](https://agda.readthedocs.io/en/latest/language/instance-arguments.html).
+* Many numeric operations in the library require their arguments to be non-zero,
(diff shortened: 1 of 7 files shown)


mechvel:
Positivity/negativity has sense for any Ordered (Abelian) Group.
There `_<_` is agreed with `_+_` in a certain way. 
It is desirable to have the corresponding general definition. 
For Integer, the Negative predicate can be expressed as the `-[1+_]` constructor.
But this does not work for Rational and some other domains.
``_< zero`` for Negative has sense for any ordered Abelian group.

technicalguy:
In my Agda work I've found that instance arguments can sometimes slow type checking down a lot. Are there any benchmarks of before and after this change? That aside, sounds great!

MatthewDaggitt:
> Positivity/negativity has sense for any Ordered (Abelian) Group.

@mechvel I agree, it should be possible to generalise the definitions if done carefully. However this PR is very much about using the definitions and the changes are almost entirely agnostic to the definitions themselves. I'll make a note of your comment in #1175 so it doesn't get forgotten when we come around to generalising the other definitions.

> In my Agda work I've found that instance arguments can sometimes slow type checking down a lot. Are there any benchmarks of before and after this change? That aside, sounds great!

That's an excellent question @technicalguy . Looking  at the test suite, it took `11m 58s` to type-check this PR0, whereas  looking at the last 7 runs of the test on master I get:
```
12m 26s
11m 25s
11m 53s
12m 00s
11m 53s
11m 13s
11m 11s
```
Lot's of variability there, but I think it's clear that it hasn't significantly boosted type-checking time, even with 100+ uses of instance arguments. I think primarily this is because the instance search is never more than single layer deep at the moment. However even with that single layer of search, there's a ton of benefits. If and when we address #1565 and try and make instance search deeper we should re-evaluate the performance impact.

PR #1582, opened 2021-08-24
Define Field and FieldMorphisms

Akshobhya1234:


MatthewDaggitt:
Apologies for the delay in replying @Akshobhya1234, I've been away.

I think the correct way of doing this is by defining `1/_` to take a (irrelevant instance) proof that it is `NonZero`, e.g. very similar to 
https://github.com/agda/agda-stdlib/blob/bfbb6399356d00812552ec9b6eef41048d123af2/src/Data/Rational/Base.agda#L228

Once we've done that, we should define an `AlmostLeftInverse` etc. in the same way that we do an `AlmostLeftCancellative` 
https://github.com/agda/agda-stdlib/blob/bfbb6399356d00812552ec9b6eef41048d123af2/src/Algebra/Definitions.agda#L122

The problem is that we don't have a generic definition of `NonZero` yet as discussed in #1175. However, the issue with defining a generic version of `NonZero` along the lines of `NonZero x = x ≉ 0#` as proposed in #1175, is that it's not going to be compatible with the definitions of `NonZero` used in `Data.Nat/Integer/Rational` that have far better reduction behaviour and therefore are more usable. I'm still pondering how best to work around this.

Do you mind if I put this PR on the backburner for a bit until we resolve #1175? I'm happy to take over responsibility for it.

Akshobhya1234:
> Do you mind if I put this PR on the backburner for a bit until we resolve #1175? I'm happy to take over responsibility for it.

Sure, Now I know what I was missing. Thanks!

Akshobhya1234:
Unfortunately this is on my master. I have to close this PR get latest pull and create a new PR (on different branch) for me to work on other things.

MatthewDaggitt:
No worries. Any chance you could create a new branch, and put this PR back up on that one? I promise I am making progress towards it, see recent discussions in #1175.

Issue #1063, opened 2020-01-21
Add rounding operations to `Data.Rational`


gallais:
[This interesting article](https://ren.zone/articles/safe-money) suggests useful operations we may want to have on rational numbers:

```agda
round floor ceiling truncate : ℚ → ℤ × ℚ
```

together with the proofs of their properties, of course.

MatthewDaggitt:
Hmm I'd be tempted to have the standard signature `ℚ → ℤ` for those operations then have another operator `frac : ℚ →ℚ` that returns the fractional part. Presumably they're simply going to be implemented by the `_/_` and `_%_` operators respectively so it's not like they have any computational overlap.

barrettj12:
I've just opened a PR #1585 which implements the suggested functions. I have not been able to prove any properties about said functions. Should we open a new issue for the implementation of those properties?

PR #1585, opened 2021-09-07
Add new module "Data.Rational.Rounding"

barrettj12:
Implements functions for rounding rationals to integers, thereby solving #1063.

### DIFF
--- a/src/Data/Rational/Rounding.agda
+++ b/src/Data/Rational/Rounding.agda
@@ -1,0 +1,47 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Rounding rational numbers
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Rational.Rounding where
+
+open import Data.Bool.Base using (Bool; true; false)
+open import Data.Nat.Base using (ℕ)
+open import Data.Integer.Base as ℤ using (ℤ)
+open import Data.Integer.DivMod using (_div_)
+open import Data.Rational.Base
+  using (ℚ; ↥_; ↧_; _≤ᵇ_; -_; _/_; 0ℚ; ½; _+_; _-_)
+
+------------------------------------------------------------------------
+-- ROUNDING FUNCTIONS
+
+-- Floor (round towards -∞)
+floor : ℚ → ℤ
+floor p = (↥ p) div (↧ p)
+
+-- Ceiling (round towards +∞)
+ceiling : ℚ → ℤ
+ceiling p = ℤ.- floor (- p)
+
+-- Truncate  (round towards 0)
+truncate : ℚ → ℤ
+truncate p with (p ≤ᵇ 0ℚ)
+...           | true  = ceiling p
+...           | false = floor p
+
+-- Round (to nearest integer)
+round : ℚ → ℤ
+round p = floor (p + ½)
+
+-- Fractional part (remainder after floor)
+fracPart : ℚ → ℚ
+fracPart p = p - floor p / 1
+
+--- Extra notations  ⌊ ⌋ floor,  ⌈ ⌉ ceiling,  [] truncate
+⌊_⌋ ⌈_⌉ [_] : ℚ → ℤ
+⌊ p ⌋ = floor p
+⌈ p ⌉ = ceiling p
+[ p ] = truncate p
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -344,6 +344,16 @@ New modules
   Algebra.Morphism.Construct.Identity
   ```
 
+* A small library for function arguments with default values:
+  ```
+  Data.Default
+  ```
+
+* Functions for rounding rationals to integers:
+  ```
+  Data.Rational.Rounding
+  ```
+
 * Show module for unnormalised rationals:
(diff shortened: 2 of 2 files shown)


src/Data/Rational/Rounding.agda
```diff
@@ -0,0 +1,47 @@
...
+floor p = (↥ p) div (↧ p)
+
+-- Ceiling (round towards +∞)
+ceiling : ℚ → ℤ
+ceiling p = ℤ.- floor (- p)
+
+-- Truncate  (round towards 0)
+truncate : ℚ → ℤ
+truncate p with (p ≤ᵇ 0ℚ)
+...           | true  = ceiling p
```
MatthewDaggitt, line 32: No need to indent these, the library style is just to go with `... | true =`

src/Data/Rational/Rounding.agda
```diff
@@ -0,0 +1,47 @@
...
+floor : ℚ → ℤ
+floor p = (↥ p) div (↧ p)
+
+-- Ceiling (round towards +∞)
+ceiling : ℚ → ℤ
+ceiling p = ℤ.- floor (- p)
+
+-- Truncate  (round towards 0)
+truncate : ℚ → ℤ
+truncate p with (p ≤ᵇ 0ℚ)
```
MatthewDaggitt, line 31: No need for these brackets

src/Data/Rational/Rounding.agda
```diff
@@ -0,0 +1,47 @@
...
+
+open import Data.Bool.Base using (Bool; true; false)
+open import Data.Nat.Base using (ℕ)
+open import Data.Integer.Base as ℤ using (ℤ)
+open import Data.Integer.DivMod using (_div_)
+open import Data.Rational.Base
+  using (ℚ; ↥_; ↧_; _≤ᵇ_; -_; _/_; 0ℚ; ½; _+_; _-_)
+
+------------------------------------------------------------------------
+-- ROUNDING FUNCTIONS
```
MatthewDaggitt, line 19: Headings are normally written using standard capitalisation, i.e. Rounding functions

MatthewDaggitt (changes requested):
Thanks for the great PR! I've added a few minor comments about library style, feel free to check out the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) for more detail.

On the more major questions side, these don't have any particularly fancy import requirements so I think they could probably go in `Data.Rational.Base`?

It'd also be good to add their counter-parts to `Data.Rational.Unnormalised` while you're here? As an aside that might help you prove some properties around them as well, as the standard proof style is first to prove properties of the unnormalised operations, then prove they're isomorphic to the normalised versions and then transfer the proofs across.

gallais:
Can we add some test cases to https://github.com/agda/agda-stdlib/tree/master/tests/data
to make sure we're getting the appropriate behaviour on edge cases?

barrettj12:
> Can we add some test cases to https://github.com/agda/agda-stdlib/tree/master/tests/data
> to make sure we're getting the appropriate behaviour on edge cases?

Sure, that sounds like a good idea. Is there a guide on how to do this? I don't have any experience with that.

gallais:
Not really: it's a fairly new addition to the stdlib.
The easiest way is probably to:

1. `cp -R` an existing directory
2. Edit the main agda file to add your tests [1]
3. Add the new test case to the the appropriate pool in the [test runner](https://github.com/agda/agda-stdlib/blob/master/tests/runtests.agda)
4. Run the test suite and accept the update `expected` file
5. Commit all of that

[1] They get run using the `run` file in the directory so you could have unit tests
of the form

```
_ : a \== b
_ = refl
```

and `run` would just make sure the file typechecks or you could have full blown
programs that get compiled & run. As you want.

barrettj12:
Side note: what should the fractional part of -3.7 be? I reckon 0.7.

JacquesCarette:
I'm sure there are coherent ways of making it 0.7, -0.7, 0.3 or -0.3. To me, it all depends on what relations with `ceil`, `floor` and/or `round` and addition you want to hold.

barrettj12:
Yeah, there doesn't seem to be a convention, so I will redefine `round` and `fracPart` to get symmetry, i.e.
```agda
round p = round (- p)
fracPart p = fracPart (- p)
```

barrettj12:
@gallais I've put in the new test cases under `tests/data/rational`. I think I've done it right.
@MatthewDaggitt I've implemented the changes you suggested, except I still haven't been able to prove anything useful about these functions. I'm happy to open a new issue for proving the properties.

gallais:
Looks like you happen to have a trailing newline in your `expected` file.

barrettj12:
> Looks like you happen to have a trailing newline in your `expected` file.

No, it appears that I _didn't_ have a trailing newline, but now I've added one to match the other `expected` files. Now the tests are passing fine, but it doesn't want to play nice with GHC. The error was
```
Cannot add PPA: 'ppa:~hvr/ubuntu/ghc'.
ERROR: '~hvr' user or team does not exist.
```
which might have been [caused by a faulty internet connection](https://askubuntu.com/a/1181499).

MatthewDaggitt:
Looks good! Thanks for the PR!

PR #1594, opened 2021-09-24
Add a couple of missing equality operators

jespercockx:
I'm not at all sure about the names, so feel free to make a suggestion. I'll write the changelog once we've agreed on the names. Also let me know if there's other similar properties that you'd like to have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -577,6 +577,25 @@ Other minor changes
   isFailure : ExitCode → Bool
   ```
 
+* Added new operations in
+  `Relation.Binary.PropositionalEquality.Properties`:
+  ```
+  J : {A : Set} {x : A} (B : (y : A) → x ≡ y → Set b)
+      {y : A} (p : x ≡ y) → B x refl → B y p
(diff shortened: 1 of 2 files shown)


JacquesCarette:
These have been sorely missing - I've got variants of this all over my libraries. [Yes, I know, I should have submitted them...]

As far as names: these seem as good as any. It's kind of annoying to use `d` as a prefix for dependent (since a lot of the library works the other way around, i.e. the default is dependent and the non-dependent version has a `'`). But I think that for `cong` and `subst`, it makes sense to make an exception!

jespercockx:
I'm glad to hear I'm not the only one who finds them essential from time to time!

JacquesCarette:
Though I also must admit that every time I use one of these, that usually causes me to "think harder" and change my definitions so that I no longer have to use them!

jespercockx:
Sure, but thinking hard is a limited resource :)

MatthewDaggitt:
I'm also happy with the names :+1:

MatthewDaggitt:
Probably time for the CHANGELOG entry!

PR #1597, opened 2021-10-04
Add Foreign.Haskell.NonEmpty

fgaz:


src/Foreign/Haskell/NonEmpty.agda
```diff
@@ -0,0 +1,38 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The NonEmpty type which calls out to Haskell via the FFI
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K #-}
+
+module Foreign.Haskell.NonEmpty where
```
fgaz, line 9: Is `Foreign.Haskell.NonEmpty` ok or should I use `Foreign.Haskell.List.NonEmpty`?

gallais: I'd be in favour of closely matching the hierarchy in `Data` so `List.NonEmpty`.

src/Foreign/Haskell/NonEmpty.agda
```diff
@@ -0,0 +1,38 @@
...
+data NonEmpty (A : Set a) : Set a where
+  _∷_ : A → List A → NonEmpty A
+
+{-# FOREIGN GHC type AgdaNonEmpty l a = NE.NonEmpty a #-}
+{-# COMPILE GHC NonEmpty = data AgdaNonEmpty ((|:)) #-}
+
+------------------------------------------------------------------------
+-- Conversion
+
+toForeign : Data.List⁺ A → NonEmpty A
```
gallais, line 34: I'm not sure we need these conversion functions these days now that we have the `coerce` mechanism.

fgaz: I'd prefer to mirror the other `Foreign` modules, by either adding those functions here too, or removing/deprecating the ones in the other modules. Should I switch to the second option?

gallais: I think it's the right thing to do, yes. Deprecate it in favour of `coerce`.

PR #1600, opened 2021-10-04
Add a proof of `subst Q eq (f x p) ≡ f y (subst P eq p)` in `Relation.Binary.PropositionalEquality.Properties`

bafain:
Note that this is interderivable with `subst-application`, using `cong-id` and `subst-∘`, but perhaps the more common one? The arguments are as in `subst-application`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -588,6 +588,11 @@ Other minor changes
   untilJust : IO (Maybe A) → IO A
   ```
 
+* Added new proofs in `Relation.Binary.PropositionalEquality.Properties`:
+  ```
+  subst-application′ : subst Q eq (f x p) ≡ f y (subst P eq p)
+  ```
+
 * Equality of predicates
   ```
   _≐_ : Pred A ℓ₁ → Pred A ℓ₂ → Set _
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
This feels like you should be able to use the existing `subst-application` and have Agda infer that `f` is equal to `id`. Does it not work?

bafain:
MatthewDaggitt writes:
> This feels like you should be able to use the existing
> `subst-application` and have Agda infer that `f` is equal to `id`.
> Does it not work?

Thanks for taking the time to consider this addition.

Do you mean as in `subst-application′ B₁ = subst-application B₁`? As you
say, Agda infers that `f` is equal to `id`, but then also raises the
type error that `cong id eq` is *not* equal to `eq`.

bafain:
Matthew writes:
> This feels like you should be able to use the existing `subst-application` and have Agda infer that `f` is equal to `id`. Does it not work?

Thanks for taking the time to consider this addition.

Do you mean as in `subst-application′ B₁ = subst-application B₁`? As you say, Agda infers that `f` is equal to `id`, but then also raises the type error that `cong id eq` is *not* equal to `eq`.

MatthewDaggitt:
> Do you mean as in subst-application′ B₁ = subst-application B₁? As you say, Agda infers that f is equal to id, but then also raises the type error that cong id eq is not equal to eq.

Ah thanks, yup that's what I was missing!

PR #1602, opened 2021-10-05
Add Fin1↔⊤ to Data.Fin.Properties

conal:
Add `Fin1↔⊤ : Fin 1 ↔ ⊤`, rounding out `Fin0↔⊥`, `+↔⊎`, and `*↔×` with the missing identity for `*`/`×`

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -894,7 +894,7 @@ Please use ≡-decSetoid instead."
 
 inject+-raise-splitAt = join-splitAt
 {-# WARNING_ON_USAGE inject+-raise-splitAt
-"Warning: decSetoid was deprecated in v1.5.
+"Warning: inject+-raise-splitAt was deprecated in v1.5.
 Please use join-splitAt instead."
 #-}
 


src/Data/Fin/Properties.agda
```diff
@@ -53,6 +53,9 @@ open import Relation.Unary.Properties using (U?)
 Fin0↔⊥ : Fin 0 ↔ ⊥
 Fin0↔⊥ = mk↔′ ¬Fin0 (λ ()) (λ ()) (λ ())
 
+Fin1↔⊤ : Fin 1 ↔ ⊤
+Fin1↔⊤ = mk↔′ (λ { 0F → tt }) (λ { tt → 0F }) (λ { tt → refl }) λ { 0F → refl }
```
gallais, line 57: Do we even need these to be pattern-matching functions as opposed to
```suggestion
Fin1↔⊤ = mk↔′ (const tt) (const 0F) (λ _ → refl) (λ _ → refl)
```
?

conal: We certainly don't need to use pattern-matching, though I personally like the explicit reminder that there is only one domain value. Are there operational differences? If we do use `const`, is there any reason not to use it in all four arguments?

If there's a standard, consistent style used in agda-stdlib for such definitions, I'm happy to adopt it.

conal: Since I'm a relative newcomer to Agda, I may be missing some historical, operational, or even semantic context.

conal: Thanks for the explanation. What a shame that these general lambdas have ergonomic drawbacks! Were they hastily implemented? Can the implementation be improved? I'm trying and failing to relate to your remark that the generated names "show up in a type". Can you give an example (copy & paste from Agda preferably)? Maybe I saw that behavior and dismissed it out of general bewilderment in using an unfamiliar-to-me language.

gallais: We have this problem here for instance: http://agda.github.io/agda-stdlib/Codata.Stream.Properties.html#1941

If you try to replace the `_` in `as ⁺++ _ ` with the expected pattern-matching
lambda you'll get an error telling you that it's not equal to another pattern-matching
that looks exactly the same.

jespercockx: > Thanks for the explanation. What a shame that these general lambdas have ergonomic drawbacks! Were they hastily implemented? Can the implementation be improved? I'm trying and failing to relate to your remark that the generated names "show up in a type". Can you give an example (copy & paste from Agda preferably)? Maybe I saw that behavior and dismissed it out of general bewilderment in using an unfamiliar-to-me language.

I'd say it isn't any more hastily implemented than most parts of Agda. The problem is a fundamental tradeoff one has to make when adding a new feature between either extending the core language (and thus increasing the "attack surface" where unsoundness could be introduced) or desugaring the new feature to existing core syntax (thus running into encoding issues like the one @gallais described here). Ideally we'd have a super-small core language with perfect abstractions on top that never leak any of their implementation details, but we're not there yet (and it is not at all obvious how to get there).

conal: Makes sense to me. Thanks!

conal: When I try the suggested replacement code (or replace `λ _ → refl` with `(const refl)`), I get the following error message:

    .../src/Data/Fin/Properties.agda:57,57-61
    0F != x of type Fin (suc _n_24)
    when checking that the expression refl has type
    const 0F (const tt x) ≡ x

Repairing this error seems to require restoring all but the third pattern lambda.

conal:
It just occurred to me that the names “`Fin0↔⊥`” and “`Fin1↔⊤`” are inconsistent with their counterparts “`+↔⊎`” and “`*↔×`” and so might be better renamed to “`0↔⊥`” and “`1↔⊤`”. Alternatively, rename the other two to “`Fin+↔⊎`” and “`Fin*↔×`” Thoughts?

JacquesCarette:
Well-spotted!  While I rather like short names, I think in this case that the `Fin`-prefixed versions might make the use clearer as to what's going on.

conal:
> I'd be happy to rename "`Fin0↔⊥`" and "`Fin1↔⊤`" to "`0↔⊥`" and "`1↔⊤`" as part of this PR. ...

@MatthewDaggitt How do such questions get resolved?

conal:
> I'd be happy to rename "`Fin0↔⊥`" and "`Fin1↔⊤`" to "`0↔⊥`" and "`1↔⊤`" as part of this PR. ...

It looks like the usual practice would include adding a deprecation warning at the end of the module.

MatthewDaggitt:
Consensus usually! I'm in favour of being consistent and dropping the 'Fin'
👍

MatthewDaggitt:
Looks great @conal, thanks for the PR! I'll leave it up a few more days in case anyone else has any comments and then merge it in.

JacquesCarette:
Core is the functor that goes from a category to its (core!) Groupoid, keeping just the invertible morphisms. So Core(FinSet) is (finite sets, bijections). FinType is the category of finite types and type isomorphisms. 

More details in my ESOP 2016 paper with Amr Sabry. The Agda code for it has bit-rotted somewhat. If you 'name' the finite types (via the language of semirings), then those are exactly equivalent to Core(FinSet). Amr and 2 students have a very beautiful 2022 POPL paper [long version just hit the arXiv yesterday] that really explores this deeply.

PR #1604, opened 2021-10-08
Morphisms for module-like structures

### LINKED ISSUE #1297: Add module homomorphisms

Taneb:
Closes #1297 

- [x] Morphisms for left semimodules
- [x] Morphisms for left modules
- [x] Morphisms for right semimodules
- [x] Morphisms for right modules
- [x] Morphisms for bisemimodules
- [x] Morphisms for bimodules
- [x] Morphisms for semimodules over a commutative semiring
- [x] Morphisms for modules over a commutative ring
- [x] Identity morphisms
- [x] Composite morphisms

### DIFF
--- a/src/Algebra/Module/Morphism/Construct/Composition.agda
+++ b/src/Algebra/Module/Morphism/Construct/Composition.agda
@@ -30,28 +30,25 @@ module _
   {g : Carrierᴹ M₂ → Carrierᴹ M₃}
   where
 
-  isLeftSemimoduleHomomorphism
-    : IsLeftSemimoduleHomomorphism M₁ M₂ f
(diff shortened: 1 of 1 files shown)


MatthewDaggitt:
Glanced over it and looks good so far. I'll give an in-depth review when you're done.

Taneb:
A big difference between these and the morphisms in `Algebra.Morphism.Structures` is that these are defined over the law-carrying structures, as we don't have definitions for things like `RawModule`.

I've only done the simple, underlying-ring-stays-the-same, morphisms, not the double category that @laMudri suggested in #1297. If there's a strong desire for those I can implement them too.

PR #1609, opened 2021-10-15
Two monoidal associativity properties for sum types

conal:


gallais:
Looks good to me! We just need CHANGELOG entries

conal:
I'm done tweaking.

PR #1615, opened 2021-10-30
Adding apartness relations and the beginnings of local rings and fields

cspollard:
Motivated by eventually having Fields in the standard library, I've started to implement apartness relations, local rings, and heyting fields, largely based on

https://ncatlab.org/nlab/show/apartness+relation
https://ncatlab.org/nlab/show/field#constructive

A couple of important (!!) notes:

1) This is my first PR to the stdlib. If I've broken established best practices in any way please let me know!

2) I've chosen here to have sub-modules for Apartness Structures and Bundles in one file separately from Relation.Binary.Structures/Bundles (similarly for LocalCommutativeRing and HeytingField outside of Algebra.Structures/Bundles). I am happy to change this or to leave it, although I suspect it will change in the future (see (4) below).

3) I would appreciate a fairly careful code review. I'm not yet a confident agda programmer. I'll point this PR out on zulip chat to the people who provided input.

4) Is there some way to mark the APIs as not yet stable? That feels appropriate.

### DIFF
--- a/src/Algebra/HeytingField.agda
+++ b/src/Algebra/HeytingField.agda
@@ -1,38 +1,96 @@
-{-# OPTIONS --without-K --safe #-}
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Local commutative rings and heyting fields
+------------------------------------------------------------------------
 
-open import Algebra
-open import Relation.Binary
-open import Level using (_⊔_)
+{-# OPTIONS --without-K --safe #-}
 
 module Algebra.HeytingField where
 
-open import Relation.Binary.Apartness
-open import Data.Product using (∃-syntax)
+open import Algebra.Core using (Op₁; Op₂)
+open import Algebra.Structures using (IsCommutativeRing)
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Level using (Level; _⊔_) renaming (suc to lsuc)
+open import Data.Product using (∃-syntax; _×_)
+import Relation.Binary.Apartness as A
 
-module _
-  {a r1} {A : Set a}
-  (_≈_ : Rel A r1)
-  (_+_ _*_ _/_ : Op₂ A) (-_ : Op₁ A) (0# 1# : A)
+
+module Structures
+  {c ℓ₁ ℓ₂} {Carrier : Set c}
+  (_≈_ : Rel Carrier ℓ₁)
+  (_#_ : Rel Carrier ℓ₂)
+  (_+_ _*_ : Op₂ Carrier) (-_ : Op₁ Carrier) (0# 1# : Carrier)
   where
 
+  open A.Structures hiding (¬#-equiv)
 
-  _#_ : A → A → Set _
-  x # y = ∃[ z ] (z * (x + (- y))) ≈ 1#
+  invertible : Carrier → Set _
+  invertible x = ∃[ 1/x ] (x * 1/x) ≈ 1#
 
 
-  record IsLocalRing : Set (a ⊔ r1) where
+  record IsLocalCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
     field
       isCommutativeRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
       isApartness : IsApartness _≈_ _#_
+      #⇔invertible : ∀ {x y} → let x-y = x + (- y) in (x # y → invertible x-y) × (invertible x-y → x # y)
 
     open IsCommutativeRing isCommutativeRing public
     open IsApartness isApartness public
 
+    open import Relation.Nullary using (¬_)
+
+    _¬#_ : Carrier → Carrier → Set _
+    x ¬# y = ¬ (x # y)
+
+    ¬#-equiv : IsEquivalence _¬#_
+    ¬#-equiv = A.Structures.¬#-equiv _≈_ _#_ refl isApartness
 
 
-  record IsHeytingField : Set (a ⊔ r1) where
+  record IsHeytingField : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
     field
-      isLocalRing : IsLocalRing
+      isLocalCommutativeRing : IsLocalCommutativeRing
       isTight : Tight _≈_ _#_
 
-    open IsLocalRing isLocalRing public
+    open IsLocalCommutativeRing isLocalCommutativeRing public
+
+
+module Bundles where
+  open Structures
+  open A.Bundles
+
+  record LocalCommutativeRing (c ℓ₁ ℓ₂ : Level) : Set (lsuc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+    field
+      Carrier                : Set c
+      _≈_                    : Rel Carrier ℓ₁
+      _#_                    : Rel Carrier ℓ₂
+      _+_                    : Op₂ Carrier
+      _*_                    : Op₂ Carrier
+      -_                     : Op₁ Carrier
+      0#                     : Carrier
+      1#                     : Carrier
+      isLocalCommutativeRing : IsLocalCommutativeRing _≈_ _#_ _+_ _*_ -_ 0# 1#
+
+    open IsLocalCommutativeRing isLocalCommutativeRing public
+
+    apart : Apartness c ℓ₁ ℓ₂
+    apart = record { isApartness = isApartness }
+
+
+  record HeytingField (c ℓ₁ ℓ₂ : Level) : Set (lsuc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+    field
+      Carrier        : Set c
+      _≈_            : Rel Carrier ℓ₁
+      _#_            : Rel Carrier ℓ₂
+      _+_            : Op₂ Carrier
+      _*_            : Op₂ Carrier
+      -_             : Op₁ Carrier
+      0#             : Carrier
+      1#             : Carrier
+      isHeytingField : IsHeytingField _≈_ _#_ _+_ _*_ -_ 0# 1#
+
+    open IsHeytingField isHeytingField public
+
+    apart : Apartness c ℓ₁ ℓ₂
+    apart = record { isApartness = isApartness }
--- a/src/Algebra/HeytingField.agda
+++ b/src/Algebra/HeytingField.agda
@@ -14,6 +14,7 @@ open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Structures using (IsEquivalence)
 open import Level using (Level; _⊔_) renaming (suc to lsuc)
 open import Data.Product using (∃-syntax; _×_)
+
 import Relation.Binary.Apartness as A
 
 
@@ -26,8 +27,10 @@ module Structures
 
   open A.Structures hiding (¬#-equiv)
 
-  invertible : Carrier → Set _
-  invertible x = ∃[ 1/x ] (x * 1/x) ≈ 1#
+  -- does this need to be public?
+  private
+    invertible : Carrier → Set _
+    invertible x = ∃[ 1/x ] (x * 1/x) ≈ 1#
 
 
   record IsLocalCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
--- a/src/Algebra/HeytingField.agda
+++ b/src/Algebra/HeytingField.agda
@@ -1,0 +1,38 @@
+{-# OPTIONS --without-K --safe #-}
+
+open import Algebra
+open import Relation.Binary
+open import Level using (_⊔_)
+
+module Algebra.Properties.HeytingField where
+
+open import Relation.Binary.Apartness
+open import Data.Product using (∃-syntax)
+
+module _
+  {a r1} {A : Set a}
+  (_≈_ : Rel A r1)
+  (_+_ _*_ _/_ : Op₂ A) (-_ : Op₁ A) (0# 1# : A)
+  where
+
+
+  _#_ : A → A → Set _
+  x # y = ∃[ z ] (z * (x + (- y))) ≈ 1#
+
+
+  record IsLocalRing : Set (a ⊔ r1) where
+    field
+      isRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
+      isApartness : IsApartness _#_ _≈_
+
+    open IsCommutativeRing isRing public
+    open IsApartness isApartness public
+
+
+
+  record IsHeytingField : Set (a ⊔ r1) where
+    field
+      isLocalRing : IsLocalRing
+      isTight : Tight _≈_ _#_
+
+    open IsLocalRing isLocalRing public
--- a/src/Algebra/HeytingField.agda
+++ b/src/Algebra/HeytingField.agda
@@ -4,7 +4,7 @@ open import Algebra
 open import Relation.Binary
 open import Level using (_⊔_)
 
-module Algebra.Properties.HeytingField where
+module Algebra.HeytingField where
 
 open import Relation.Binary.Apartness
 open import Data.Product using (∃-syntax)
@@ -22,10 +22,10 @@ module _
 
   record IsLocalRing : Set (a ⊔ r1) where
     field
-      isRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
-      isApartness : IsApartness _#_ _≈_
+      isCommutativeRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
+      isApartness : IsApartness _≈_ _#_
 
-    open IsCommutativeRing isRing public
+    open IsCommutativeRing isCommutativeRing public
     open IsApartness isApartness public
 
 
--- a/src/Data/Nat/Apartness.agda
+++ b/src/Data/Nat/Apartness.agda
@@ -1,0 +1,59 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- An apartness relation for ℕ
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Nat.Apartness where
+
+open import Relation.Binary.Definitions using (Irreflexive; Symmetric)
+open import Data.Nat.Base using (ℕ; zero; suc)
+open import Relation.Binary.PropositionalEquality
+open import Data.Sum using (inj₁; inj₂; map)
+open import Data.Product using (_,_)
+open import Function using (_∘_)
+
+import Relation.Binary.Apartness as A
+open A.Structures
+
+
+data _#_ : ℕ → ℕ → Set where
+  z#s : ∀ {m : ℕ} → zero # suc m
+  s#z : ∀ {m : ℕ} → suc m # zero
+  s#s : ∀ {m n : ℕ} → m # n → suc m # suc n
+
+open A.Properties _#_
+
+#-reduce : ∀ {m n : ℕ} → suc m # suc n → m # n
+#-reduce (s#s x) = x
+
+#-irrefl : Irreflexive _≡_ _#_
+#-irrefl {zero} {zero} refl ()
+#-irrefl refl (s#s x) = #-irrefl refl x
+
+#-sym : Symmetric _#_
+#-sym z#s = s#z
+#-sym s#z = z#s
+#-sym (s#s x) = s#s (#-sym x)
+
+#-comp : Comparison
+#-comp {y = zero} z#s = inj₂ z#s
+#-comp {y = suc _} z#s = inj₁ z#s
+#-comp {y = zero} s#z = inj₁ s#z
+#-comp {y = suc _} s#z = inj₂ s#z
+#-comp {y = zero} (s#s x#z) = inj₁ s#z
+#-comp {y = suc _} (s#s x#z) = map s#s s#s (#-comp x#z)
+
+#-apart : IsApartness _≡_ _#_
+#-apart = record { irrefl = #-irrefl ; sym = #-sym ; comp = #-comp }
+
+#-tight : Tight _≡_ _#_
+#-tight {zero} {zero} _ = refl
+#-tight {suc x} {zero} (¬x#y , _) with ¬x#y s#z
+... | ()
+#-tight {zero} {suc y} (_ , ¬y#x) with ¬y#x s#z
+... | ()
+#-tight {suc x} {suc y} (¬sx#sy , ¬sy#sx) =
+  cong suc (#-tight (¬sx#sy ∘ s#s , ¬sy#sx ∘ s#s))
--- a/src/Relation/Binary/Apartness.agda
+++ b/src/Relation/Binary/Apartness.agda
@@ -1,0 +1,66 @@
+open import Relation.Binary
+
+module Relation.Binary.Apartness where
+
+open import Level using (Level; _⊔_)
+open import Relation.Binary.PropositionalEquality using (_≡_; _≢_) renaming (sym to ≡-sym)
+open import Relation.Nullary using (¬_)
+open import Data.Sum using (_⊎_; [_,_]′; inj₁; inj₂)
+open import Data.Product using (_×_)
+
+
+module _ {a ℓ} {A : Set a} (_#_ : Rel A ℓ) where
+
+  _¬#_ : A → A → Set ℓ
+  x ¬# y = ¬ (x # y)
+
+
+  irrefl⇒refl-¬ : Irreflexive _≡_ _#_ → Reflexive _¬#_
+  irrefl⇒refl-¬ irr = irr _≡_.refl
+
+
+  sym⇒sym-¬ : Symmetric _#_ → Symmetric _¬#_
+  sym⇒sym-¬ sym# x¬#y y#x = x¬#y (sym# y#x)
+
+
+  Comparison : Set _
+  Comparison = ∀ {x y z} → x # z → (x # y) ⊎ (y # z)
+
+
+  comp⇒trans-¬ : Comparison → Transitive _¬#_
+  comp⇒trans-¬ comp x¬#y y¬#z x#z = [ x¬#y , y¬#z ]′ (comp x#z)
+
+
+  record IsApartness : Set (a ⊔ ℓ) where
+    field
+      irrefl : Irreflexive _≡_ _#_
+      sym    : Symmetric _#_
+      comp   : Comparison
+
+
+    ¬#-equiv : IsEquivalence _¬#_
+    ¬#-equiv = record { refl = irrefl⇒refl-¬ irrefl ; sym = sym⇒sym-¬ sym ; trans = comp⇒trans-¬ comp }
+    
+
+  Tight : Set _
+  Tight = ∀ {x y} → (x # y) × (y # x) → x ≡ y
+
+
+module _ {a} {A : Set a} where
+
+  -- _≢_ : A → A → Set _
+  -- x ≢ y = ¬ (x ≡ y)
+
+  ≢-irrefl : Irreflexive {A = A} _≡_ _≢_
+  ≢-irrefl x≡y x≢y = x≢y x≡y
+
+  ≢-sym : Symmetric {A = A} _≢_
+  ≢-sym x≢y y≡x = x≢y (≡-sym y≡x)
+
+  -- I think these require decidability.
+
+  ≢-comp : Comparison {A = A} _≢_
+  ≢-comp {x = x} {y = y} {z = z} x≢z = inj₁ {!   !}
+
+  ≢-apart : IsApartness {A = A} _≢_
+  ≢-apart = record { irrefl = ≢-irrefl ; sym = ≢-sym ; comp = {!   !} } 
--- a/src/Relation/Binary/Apartness.agda
+++ b/src/Relation/Binary/Apartness.agda
@@ -1,22 +1,20 @@
+{-# OPTIONS --without-K --safe #-}
+
 open import Relation.Binary
 
 module Relation.Binary.Apartness where
 
 open import Level using (Level; _⊔_)
-open import Relation.Binary.PropositionalEquality using (_≡_; _≢_) renaming (sym to ≡-sym)
 open import Relation.Nullary using (¬_)
-open import Data.Sum using (_⊎_; [_,_]′; inj₁; inj₂)
-open import Data.Product using (_×_)
+open import Data.Sum using (_⊎_; [_,_]′; inj₁; inj₂; map)
+open import Data.Product using (_×_; _,_)
+open import Function using (_∘_)
 
 
-module _ {a ℓ} {A : Set a} (_#_ : Rel A ℓ) where
+module Local {a r1} {A : Set a} (_#_ : Rel A r1) where
 
-  _¬#_ : A → A → Set ℓ
+  _¬#_ : A → A → Set r1
   x ¬# y = ¬ (x # y)
-
-
-  irrefl⇒refl-¬ : Irreflexive _≡_ _#_ → Reflexive _¬#_
-  irrefl⇒refl-¬ irr = irr _≡_.refl
 
 
   sym⇒sym-¬ : Symmetric _#_ → Symmetric _¬#_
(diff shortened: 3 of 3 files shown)


src/Relation/Binary/Apartness.agda
```diff
@@ -0,0 +1,68 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Apartness relations
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Relation.Binary.Apartness where
```
MatthewDaggitt, line 9: Unless I'm mistaken, there's not a deep hierarchy of apartness relations? If that's the case then I think we should just fold the definitions into the correct places in the standard `Relation.Binary` modules. i.e.

- `Comparison`/`Tight` in `Relation.Binary.Definitions`
- `sym⇒sym-¬`/`comp⇒trans-¬` etc. in `Relation.Binary.Consequences`
- `IsApartness` in `Relation.Binary.Structures`
- `Apartness` in `Relation.Binary.Bundles`
- `¬#-equiv` in `Relation.Binary.Properties.Apartness`

src/Relation/Binary/Apartness.agda
```diff
@@ -0,0 +1,68 @@
...
+      irrefl : Irreflexive _≈_ _#_
+      sym    : Symmetric _#_
+      comp   : Comparison
+
+  irrefl⇒refl-¬ : Reflexive _≈_ → Irreflexive _≈_ _#_ → Reflexive _¬#_
+  irrefl⇒refl-¬ re irr = irr re
+
+  ¬#-equiv : Reflexive _≈_ → IsApartness → IsEquivalence _¬#_
+  ¬#-equiv re apart =
+    record { refl = irrefl⇒refl-¬ re irrefl ; sym = sym⇒sym-¬ sym ; trans = comp⇒trans-¬ comp }
```
MatthewDaggitt, line 50: See the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#layout-of-record-instances) for how to layout record instances

src/Data/Nat/Apartness.agda
```diff
@@ -0,0 +1,59 @@
...
+open import Relation.Binary.PropositionalEquality
+open import Data.Sum using (inj₁; inj₂; map)
+open import Data.Product using (_,_)
+open import Function using (_∘_)
+
+import Relation.Binary.Apartness as A
+open A.Structures
+
+
+data _#_ : ℕ → ℕ → Set where
```
MatthewDaggitt, line 22: I'm a bit confused about why this definition is needed. Is it not just the negation of propositional equality?

src/Relation/Binary/Apartness.agda
```diff
@@ -0,0 +1,68 @@
...
+module Properties {c ℓ₁} {Carrier : Set c} (_#_ : Rel Carrier ℓ₁) where
+
+  _¬#_ : Carrier → Carrier → Set _
+  x ¬# y = ¬ (x # y)
+
+  sym⇒sym-¬ : Symmetric _#_ → Symmetric _¬#_
+  sym⇒sym-¬ sym# x¬#y y#x = x¬#y (sym# y#x)
+
+  Comparison : Set _
+  Comparison = ∀ {x y z} → x # z → (x # y) ⊎ (y # z)
```
MatthewDaggitt, line 29: I suspect that in order to make this usable you're going to have to make `y` explicit, as Agda won't be able to infer it. We also tend to name variables in the order that they're used, so it would be `∀ {x y} z → x # y → (x # z) ⊎ (z # y)`

src/Algebra/HeytingField.agda
```diff
@@ -0,0 +1,99 @@
...
+
+    open IsCommutativeRing isCommutativeRing public
+    open IsApartness isApartness public
+
+    open import Relation.Nullary using (¬_)
+
+    _¬#_ : Carrier → Carrier → Set _
+    x ¬# y = ¬ (x # y)
+
+    ¬#-equiv : IsEquivalence _¬#_
```
MatthewDaggitt, line 50: What's the relationship between `_≈_` and `_#_`? Do they coincide?

src/Algebra/HeytingField.agda
```diff
@@ -0,0 +1,99 @@
...
+  (_#_ : Rel Carrier ℓ₂)
+  (_+_ _*_ : Op₂ Carrier) (-_ : Op₁ Carrier) (0# 1# : Carrier)
+  where
+
+  open A.Structures hiding (¬#-equiv)
+
+  -- does this need to be public?
+  private
+    invertible : Carrier → Set _
+    invertible x = ∃[ 1/x ] (x * 1/x) ≈ 1#
```
MatthewDaggitt, line 33: Yes, I think it should be public and it should live in `Algebra.Definitions`. It's also a type so by convention it should be capitalised.

src/Algebra/HeytingField.agda
```diff
@@ -0,0 +1,99 @@
...
+
+
+  record IsLocalCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
```
MatthewDaggitt, line 36: In contrast to apartness relations, it seems to me there probably is a deep hierarchy of "local" structures that use the apartness relation, so maybe they should live in their own hierarchy. What do you think about them living under `Algebra.Local`? Open to better naming suggestions!

Just like the main algebra hierarchy they would once again have separate modules e.g. `Algebra.Local.Structures`/`Algebra.Local.Bundles` etc.

src/Algebra/HeytingField.agda
```diff
@@ -0,0 +1,99 @@
...
+    field
+      isCommutativeRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
+      isApartness : IsApartness _≈_ _#_
+      #⇔invertible : ∀ {x y} → let x-y = x + (- y) in (x # y → invertible x-y) × (invertible x-y → x # y)
+
+    open IsCommutativeRing isCommutativeRing public
+    open IsApartness isApartness public
+
+    open import Relation.Nullary using (¬_)
```
MatthewDaggitt, line 45: All import statements should go at the top of files :+1:

MatthewDaggitt (changes requested):
Thanks for the PR @cspollard. The content is excellent, but as you say there's a few organisational improvements that could be made to match the style of the rest of the library. I've added some comments about the big changes, and then we can have a second round of reviews to tidy up some of the smaller things. 

> 4. Is there some way to mark the APIs as not yet stable? That feels appropriate.

Unfortunately not... apart from adding a comment at the top of the files.

src/Relation/Binary/Apartness.agda
cspollard: These have now been implemented, with some renaming to make them a bit more general (they're not all specific to apartness relations).

src/Relation/Binary/Apartness.agda
cspollard: Fixed!

src/Data/Nat/Apartness.agda
cspollard: You are 100% correct --- I could easily prove that prop equality and the negation of this relation were isomorphic (that's what made it tight in fact...). I've removed this file.

src/Relation/Binary/Apartness.agda
cspollard: I've done this. It makes the logic in comp=>trans slightly more convoluted, but the original definition is indeed easier.

src/Algebra/HeytingField.agda
cspollard: Done. But please take a look --- I added Left/RightInvertible as well as Invertible.

src/Algebra/HeytingField.agda
cspollard: I've done this, but I won't resolve this thread as it's not obvious what the best route is.

src/Algebra/HeytingField.agda
cspollard: Fixed.

cspollard:
Two general comments after the latest changes:

1) I hope it's ok to squash commits. I think only one commit is really warranted for these changes.

2) I've made liberal use of some private definitions in order to clean up definitions. Sometimes this leads to repetitions. If there are suggestions for how to improve this, I'm happy to hear them.

src/Relation/Binary/Consequences.agda
```diff
@@ -44,6 +44,31 @@ module _ {_∼_ : Rel A ℓ} {P : Pred A p} where
...
 
+------------------------------------------------------------------------
+-- Proofs for negation
+
+module _ {_∼_ : Rel A ℓ} where
+  private
+    _≁_ : A → A → Set _
+    x ≁ y = ¬ (x ∼ y)
+
+  sym⇒≁-sym : Symmetric _∼_ → Symmetric _≁_
```
MatthewDaggitt, line 55: Looking at the rest of the file, I think the most consistent names for these two properties would be `sym⇒¬-sym` and `comp⇒¬-trans`

src/Relation/Binary/Consequences.agda
```diff
@@ -44,6 +44,31 @@ module _ {_∼_ : Rel A ℓ} {P : Pred A p} where
...
+  sym⇒≁-sym sym≁ x≁y y∼x = x≁y (sym≁ y∼x)
+
+  comp⇒≁-trans : Comparison _∼_ → Transitive _≁_
+  comp⇒≁-trans comp {i = x} {j = z} {k = y} x≁z z≁y x∼y = [ x≁z , z≁y ]′ (comp z x∼y)
```
MatthewDaggitt, line 59: I think you can remove the implicit arguments here.

src/Relation/Binary/Bundles.agda
```diff
@@ -297,3 +297,14 @@ record StrictTotalOrder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) wh
   "Warning: decSetoid was deprecated in v1.3.
   Please use Eq.decSetoid instead."
   #-}
+
+
+
+record Apartness c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
```
MatthewDaggitt, line 303: This should go above the deprecated names in the file!

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
...
+    x ¬# y = ¬ (x # y)
+
+    ¬#-equiv : IsEquivalence _¬#_
+    ¬#-equiv = A.¬#-equiv refl isApartness
+
+
+  record IsHeytingField : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      isLocalCommutativeRing : IsLocalCommutativeRing
+      isTight                : Tight _≈_ _#_
```
MatthewDaggitt, line 58: The `is` in names is reserved for structures, so this would just be called `tight` rather than `isTight`

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
...
+    field
+      isCommutativeRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
+      isApartness       : IsApartness _≈_ _#_
+      #⇔invertible      : ∀ {x y} → Iso (x # y) (Invertible _≈_ 1# _*_ (x - y))
+
+    open IsCommutativeRing isCommutativeRing public
+    open IsApartness isApartness public
+
+    _¬#_ : Carrier → Carrier → Set _
+    x ¬# y = ¬ (x # y)
```
MatthewDaggitt, line 49: I feel like this definition should be in `IsApartness` rather than here.

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
...
+
+    ¬#-equiv : IsEquivalence _¬#_
+    ¬#-equiv = A.¬#-equiv refl isApartness
```
MatthewDaggitt, line 52: This should be in `Relation.Binary.Properties.Apartness`

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
...
+open import Relation.Binary.Structures using (IsEquivalence; IsApartness)
+open import Relation.Binary.Definitions using (Tight)
+import Relation.Binary.Properties.Apartness as A
+open import Relation.Nullary using (¬_)
+
+
+module _
+  {c ℓ₁ ℓ₂} {Carrier : Set c}
+  (_≈_ : Rel Carrier ℓ₁)
+  (_#_ : Rel Carrier ℓ₂)
```
MatthewDaggitt, line 26: We can play the same trick as in `Algebra.Structures` and lift `{a ℓ₁ ℓ₂} {A : Set a}  (_≈_ : Rel Carrier ℓ₁)  (_#_ : Rel Carrier ℓ₂)` up to be parameters of the top-level module, which allows users to import all it's contents already parameterised :+1:  The other arguments should then just be part of the main definition.

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Local algebraic structures
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Local.Structures where
```
MatthewDaggitt, line 9: We should have a module `Algebra.Local` that re-exports both `Structures` and `Bundles`.

src/Algebra/Local/Structures.agda
```diff
@@ -0,0 +1,60 @@
...
+  {c ℓ₁ ℓ₂} {Carrier : Set c}
+  (_≈_ : Rel Carrier ℓ₁)
+  (_#_ : Rel Carrier ℓ₂)
+  (_+_ _*_ : Op₂ Carrier) (-_ : Op₁ Carrier) (0# 1# : Carrier)
+  where
+
+  -- private to avoid clash with Monoid._-_
+  private
+    _-_ : Op₂ Carrier
+    x - y = x + (- y)
```
MatthewDaggitt, line 33: This definition of `_-_` should already be available in `IsCommutativeRing` inherited from `IsGroup`. That would allow us to do the following:
```agda
  record IsLocalCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
    field
      isCommutativeRing : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
      isApartness       : IsApartness _≈_ _#_
    
    open IsCommutativeRing isCommutativeRing public
    open IsApartness isApartness public

    field
      #⇔invertible      : ∀ {x y} → Iso (x # y) (Invertible _≈_ 1# _*_ (x - y))
```
If it's not available, then it really should be and we've messed up the transitive imports in the main algebra hierarchy somewhere. If you're not up to finding where, let me know and I'll try and find the place where it goes wrong.

MatthewDaggitt (commented):
Thanks for the changes! Already looking much better :+1: Here's the next batch.

> I hope it's ok to squash commits. I think only one commit is really warranted for these changes.

No need for you to squash them manually as they get squashed automatically when I merge the final thing in.

src/Algebra/HeytingField.agda
cspollard: They are isomorphic only in the case of a HeytingField, but not for a HeytingCommutativeRing.

src/Algebra/HeytingField.agda
cspollard: ok I've put this in Algebra.WithApartness now. See zulip discussions (I'll bring it up again there before resolving this thread).

src/Relation/Binary/Structures.agda
```diff
@@ -278,3 +278,12 @@ record IsStrictTotalOrder (_<_ : Rel A ℓ₂) : Set (a ⊔ ℓ ⊔ ℓ₂) wher
 
   open IsStrictPartialOrder isStrictPartialOrder public
     using (irrefl; asym; <-respʳ-≈; <-respˡ-≈; <-resp-≈)
+
+
+-- Apartness relations
+
+record IsApartness {ℓ₁} (_#_ : Rel A ℓ₁) : Set (a ⊔ ℓ ⊔ ℓ₁) where
```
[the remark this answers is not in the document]
cspollard: I agree. I was following the naming from "IsEquivalence" vs "IsEquivalenceRelation" but you're right "equivalence" can be a noun where I don't think apartness can. Fixed.

cspollard:
Thanks @MatthewDaggitt for the close review, and apologies for some of the obvious mistakes you had to spot! I've made a lot of progress but it's not yet complete. I'll slowly resolve the threads and write back here when they're ready to be looked at again.

MatthewDaggitt:
>  Apologies for some of the obvious mistakes you had to spot!

Absolutely no need to apologise! It's a hefty and complicated first PR and we appreciate the effort :grin:

src/Relation/Binary/Definitions.agda
```diff
@@ -208,6 +208,15 @@ record NonEmpty {A : Set a} {B : Set b}
 
 
 
+-- Definitions for apartness relations
+
+Comparison : Rel A ℓ → Set _
```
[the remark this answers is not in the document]
cspollard: Yep done.

src/Relation/Binary/Consequences.agda
cspollard: Done.

src/Relation/Binary/Consequences.agda
cspollard: In fact I kind of prefer keeping them because it highlights the difference in ordering between Comparison and Transitive (notice j <-> z and k <-> y). I can add a comment pointing that out if you think it could help?

src/Algebra/Local/Structures.agda
cspollard: Agree. This is done.

src/Algebra/Local/Structures.agda
cspollard: It's now in `Relation.Binary.Properties.ApartnessRelation` (following the name change).

src/Algebra/Local/Structures.agda
cspollard: Yep done.

src/Algebra/Local/Structures.agda
cspollard: I've put this in `Algebra.WithApartness`. One thing I wasn't sure of is whether this should import their definitions publicly. I assume that's the point?

src/Algebra/Local/Structures.agda
cspollard: That worked! I thought imports were not allowed interleaved with field declarations. It's a nice trick.

src/Relation/Binary/Bundles.agda
cspollard: Hang on I think that deprecated name is local to `StrictTotalOrder`. ;-)

src/Algebra/Local/Structures.agda
cspollard: Done.

cspollard:
@MatthewDaggitt I think I've mostly resolved your previous comments. I think we still need a few more iterations on naming and structure, and I am considering more properties that might be interesting to have for the algebraic structures with apartness relations (`1 # 0` is the only one I've provided so far).

src/Relation/Binary/Definitions.agda
```diff
@@ -130,6 +130,14 @@ Min R = Max (flip R)
 Minimum : Rel A ℓ → A → Set _
 Minimum = Min
 
+-- Definitions for apartness relations
+
+Comparison : Rel A ℓ → Set _
```
MatthewDaggitt, line 135: Okay, more bikeshedding incoming. All the other properties names are adjectives (`Transitive`/`Symmetric`/`Total` etc.), but `Comparison` is a noun. Would it be reasonable to call this `Comparitive` instead? Or if you don't like that we could use the alternative name in the literature `Cotransitive`? The latter would also have the advantage that the shortening of it, `cotrans`, wouldn't clash with the shortening of `Trichotomous` which is `cmp`. Having two very different properties called `cmp` and `comp` probably isn't great.

src/Relation/Binary/Properties/ApartnessRelation.agda
```diff
@@ -0,0 +1,33 @@
...
+
+
+private
+  _¬#_ : A → A → Set _
+  x ¬# y = ¬ (x # y)
+
+¬#-equiv : Reflexive _≈_ → IsApartnessRelation _≈_ _#_ → IsEquivalence _¬#_
+¬#-equiv re apart = record
+  { refl = irrefl re
+  ; sym = λ x¬#y y#x → x¬#y (sym y#x)
```
MatthewDaggitt, line 30: Is this `sym⇒¬-sym`?

src/Relation/Binary/Properties/ApartnessRelation.agda
```diff
@@ -0,0 +1,33 @@
...
+  ; trans = comp⇒¬-trans comp
+  }
+  where open IsApartnessRelation apart
```
MatthewDaggitt, line 33: Super minor, but we usually save a line by putting the `where` after the `}`

src/Algebra/WithApartness/Structures.agda
```diff
@@ -0,0 +1,56 @@
...
+    isCommutativeRing   : IsCommutativeRing _≈_ _+_ _*_ -_ 0# 1#
+    isApartnessRelation : IsApartnessRelation _≈_ _#_
+
+  open IsCommutativeRing isCommutativeRing public
+  open IsApartnessRelation isApartnessRelation public
+
+  field
+    #⇔invertible : ∀ {x y} → Iso (x # y) (Invertible _≈_ 1# _*_ (x - y))
+
+  ¬#-equiv : IsEquivalence _¬#_
```
MatthewDaggitt, line 46: The standard name for this would be `¬#-isEquivalence`

src/Algebra/WithApartness/Structures.agda
```diff
@@ -0,0 +1,56 @@
...
+{-# OPTIONS --without-K --safe #-}
+
+open import Algebra.Core using (Op₁; Op₂)
+open import Relation.Binary.Core using (Rel)
+
+module Algebra.WithApartness.Structures
+  {c ℓ₁ ℓ₂} {Carrier : Set c}
+  (_≈_ : Rel Carrier ℓ₁)
+  (_#_ : Rel Carrier ℓ₂)
+  (_+_ _*_ : Op₂ Carrier) (-_ : Op₁ Carrier) (0# 1# : Carrier)
```
MatthewDaggitt, line 16: So the parameters on this line should be local to each definition. Otherwise when in future we add more structures which have a different set of operations, we'll have to break compatibility.

src/Algebra/WithApartness/Structures.agda
```diff
@@ -0,0 +1,56 @@
...
+  where
+
+
```
MatthewDaggitt, line 19: As per the style-guide, generally we only use single rather than double line breaks. The exception is after new record declarations where we use double.

src/Algebra/WithApartness/Bundles.agda
```diff
@@ -0,0 +1,54 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.WithApartness.Bundles where
+
+open import Level using (_⊔_; suc)
+open import Algebra.Core using (Op₁; Op₂)
+open import Algebra.Bundles using (CommutativeRing)
+open import Algebra.WithApartness.Structures using (IsHeytingCommutativeRing; IsHeytingField)
```
MatthewDaggitt, line 14: I wouldn't bother with a `using` here, as we will by default need everything in this module.

src/Relation/Binary/Bundles.agda
```diff
@@ -297,3 +297,14 @@ record StrictTotalOrder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) wh
   "Warning: decSetoid was deprecated in v1.3.
   Please use Eq.decSetoid instead."
   #-}
+
+
+
+record ApartnessRelation c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  field
+    Carrier     : Set c
+    _≈_         : Rel Carrier ℓ₁
```
MatthewDaggitt, line 306: We're missing fixity definitions for these relations (see other definitions above). I wonder what fixity would be best for `_#_`? I'm guessing as it's the counterpart for equality it should also be 4?

src/Algebra/WithApartness/Structures.agda
```diff
@@ -0,0 +1,56 @@
...
+open import Level using (_⊔_) renaming (suc to lsuc)
+open import Data.Product using (∃-syntax; _×_; _,_; proj₂)
+open import Algebra.Definitions using (Invertible)
+open import Algebra.Structures using (IsCommutativeRing)
+open import Relation.Binary.Structures using (IsEquivalence; IsApartnessRelation)
+open import Relation.Binary.Definitions using (Tight)
+import Relation.Binary.Properties.ApartnessRelation as A
```
MatthewDaggitt, line 26: Could we name `A` something slightly more meaningful? Either `Apartness` or `ApartnessRelation`?

src/Algebra/WithApartness/Bundles.agda
```diff
@@ -0,0 +1,54 @@
...
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Bundles using (ApartnessRelation)
+
+record HeytingCommutativeRing c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  field
+    Carrier                  : Set c
+    _≈_                      : Rel Carrier ℓ₁
```
MatthewDaggitt, line 21: Again, we need fixity declarations for all of these (see `Algebra.Bundles` for examples)

src/Algebra/Properties/HeytingCommutativeRing.agda
```diff
@@ -0,0 +1,27 @@
...
+open HeytingCommutativeRing HRC
+open CommutativeRing cring using (ring)
+
+open import Algebra.Properties.Ring ring using (-0#≈0#)
+
+
+1#0 : 1# # 0#
+1#0 = proj₂ #⇔invertible (1# ,  1*[x-0]≈x , [x-0]*1≈x)
+  where
+    x-0≈x : ∀ {x} → (x - 0#) ≈ x
```
MatthewDaggitt, line 20: Once you've added fixity declarations, we should be able to get rid of a lot of these brackets in this proof.

MatthewDaggitt (commented):
Some more comments!

src/Relation/Binary/Consequences.agda
MatthewDaggitt: I guess I'd prefer a comment than the unused implicit arguments, and I think the comment should go on the main definition rather than here, but if you're keen to keep both then that's fine with me :+1:

src/Algebra/HeytingField.agda
MatthewDaggitt: I think we've reached consensus on `Algebra.Apartness` in the thread.

src/Relation/Binary/Bundles.agda
MatthewDaggitt: Yes, you're right! My fault for not expanding upwards and only relying on the limited Github view.

src/Algebra/Local/Structures.agda
MatthewDaggitt: Yup, it should allow people to import all the definitions in the hierarchy with only one import statement.

src/Algebra/Properties/HeytingCommutativeRing.agda
cspollard: true! one question: is this file better placed at `Algebra.Apartness.Properties.HeytingCommutativeRing`?

src/Relation/Binary/Definitions.agda
cspollard: Yep agree. I've renamed the member of `IsApartnessRelation` to `cotrans`.

src/Relation/Binary/Properties/ApartnessRelation.agda
cspollard: Yes, and I've replaced it. One odd thing that maybe you could comment on: for some reason I have to give agda the implicit arguments to `sym`.

```agda
¬#-isEquivalence : Reflexive _≈_ → IsApartnessRelation _≈_ _#_ → IsEquivalence _¬#_
¬#-isEquivalence re apart = record
  { refl = irrefl re
  ; sym = λ {a} {b} → sym⇒¬-sym sym {a} {b} 
  ; trans = cotrans⇒¬-trans cotrans
  } where open IsApartnessRelation apart 
 ```

I'm not sure why...

File: src/Relation/Binary/Properties/ApartnessRelation.agda (hunk 22 above)
Comment (cspollard) [reply src/Relation/Binary/Properties/ApartnessRelation.agda]: fixed.

File: src/Algebra/WithApartness/Structures.agda (hunk 23 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Structures.agda]: I've changed this both in `Algebra.Apartness.Structures` and also in `Relation.Binary.Properties.ApartnessRelation`. Was that the intention?

File: src/Algebra/WithApartness/Structures.agda (hunk 24 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Structures.agda]: done.

File: src/Algebra/WithApartness/Structures.agda (hunk 25 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Structures.agda]: ok I've fixed this particular instance. I'll have to double-check the rest of the files once we're happy with the structure and naming.

File: src/Algebra/WithApartness/Bundles.agda (hunk 26 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Bundles.agda]: agree. removed.

File: src/Relation/Binary/Bundles.agda (hunk 27 above)
Comment (cspollard) [reply src/Relation/Binary/Bundles.agda:307]: Right I completely forgot fixities throughout. I've set `_#_` to `infix 4` as you suggested. I think that does make the most sense.

File: src/Algebra/WithApartness/Structures.agda (hunk 28 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Structures.agda]: Yes. done.

File: src/Algebra/WithApartness/Bundles.agda (hunk 29 above)
Comment (cspollard) [reply src/Algebra/WithApartness/Bundles.agda]: Thanks. I've taken all the algebraic operation fixities from CommutativeRing.

File: src/Relation/Binary/Consequences.agda (hunk 10 above)
Comment (cspollard) [reply src/Relation/Binary/Consequences.agda]: I've put comments in both places and removed the implicit arguments. Hopefully that's clear enough for future readers.

File: src/Algebra/Local/Structures.agda (hunk 16 above)
Comment (cspollard) [reply src/Algebra/Local/Structures.agda]: Great. As suggested on Zulip and noted below, everything is now at Algebra.Apartness.

File: src/Relation/Binary/Properties/ApartnessRelation.agda (hunk 21 above)
Comment (MatthewDaggitt) [reply src/Relation/Binary/Properties/ApartnessRelation.agda]: Yes, I also run into this problem now and again. I've never managed to find an explanation I can understand. We run into the problem in other places in the library as well, so it's fine to leave the implicits in.

File: src/Algebra/WithApartness/Structures.agda (hunk 23 above)
Comment (MatthewDaggitt) [reply src/Algebra/WithApartness/Structures.agda]: Yes :+1:

File: src/Algebra/Properties/HeytingCommutativeRing.agda (hunk 30 above)
Comment (MatthewDaggitt) [reply src/Algebra/Properties/HeytingCommutativeRing.agda]: Yes, good spot it should be there!

File: src/Relation/Binary/Definitions.agda [hunk 31]
```diff
@@ -130,6 +130,15 @@ Min R = Max (flip R)
 Minimum : Rel A ℓ → A → Set _
 Minimum = Min
 
+-- Definitions for apartness relations
+
+-- Note that Cotransitive's arguments are permuted with respect to Transitives.
```
MatthewDaggitt, line 135: Typo: `Transitive's`

src/Algebra/Properties/HeytingCommutativeRing.agda
```diff
@@ -0,0 +1,27 @@
...
+open CommutativeRing cring using (ring)
+
+open import Algebra.Properties.Ring ring using (-0#≈0#)
+
+
+1#0 : 1# # 0#
+1#0 = proj₂ #⇔invertible ( 1# ,  ( 1*[x-0]≈x , [x-0]*1≈x ) )
+  where
+    x-0≈x : ∀ {x} → x - 0# ≈ x
+    x-0≈x {x} = trans (+-cong refl -0#≈0#) (+-identityʳ x)
```
MatthewDaggitt, line 21: This could be lifted out to a top-level proof and be given the type `RightIdentity ... ...`?

src/Algebra/Properties/HeytingCommutativeRing.agda
```diff
@@ -0,0 +1,27 @@
...
+open import Algebra.Bundles using (CommutativeRing)
+
+open HeytingCommutativeRing HRC
+open CommutativeRing cring using (ring)
+
+open import Algebra.Properties.Ring ring using (-0#≈0#)
+
+
+1#0 : 1# # 0#
+1#0 = proj₂ #⇔invertible ( 1# ,  ( 1*[x-0]≈x , [x-0]*1≈x ) )
```
MatthewDaggitt, line 18: I don't think we need the inner brackets, and we can certainly remove some of the whitespace!

src/Algebra/Apartness/Structures.agda
```diff
@@ -0,0 +1,58 @@
...
+import Relation.Binary.Properties.ApartnessRelation as ApartnessRelation
+
+private
+  Iso : ∀ {a b} → Set a → Set b → Set _
+  Iso A B = (A → B) × (B → A)
+
+record IsHeytingCommutativeRing
+  {c ℓ₁ ℓ₂} {Carrier : Set c}
+  (_≈_ : Rel Carrier ℓ₁)
+  (_#_ : Rel Carrier ℓ₂)
```
MatthewDaggitt, line 29: I think you went too far the other way again. Every parameter above this comment will be common to _all_ structures defined in this file, and therefore should be top-level module parameters, whereas everything below it will vary depending on the structure in question and therefore should parameters to the individual records. My fault for not being clearer, apologies!

src/Algebra/Apartness/Bundles.agda
```diff
@@ -0,0 +1,65 @@
...
+    _+_                      : Op₂ Carrier
+    _*_                      : Op₂ Carrier
+    -_                       : Op₁ Carrier
+    0#                       : Carrier
+    1#                       : Carrier
+    isHeytingCommutativeRing : IsHeytingCommutativeRing _≈_ _#_ _+_ _*_ -_ 0# 1#
+
+  open IsHeytingCommutativeRing isHeytingCommutativeRing public
+
+  cring : CommutativeRing c ℓ₁
```
MatthewDaggitt, line 36: We tend to use full names for such bundles so this should be `commutativeRing`. Similarly, I know it's verbose, but I would go for `apartnessRelation` rather than `apart`.

src/Algebra/Apartness/Bundles.agda
```diff
@@ -0,0 +1,65 @@
...
+    _+_            : Op₂ Carrier
+    _*_            : Op₂ Carrier
+    -_             : Op₁ Carrier
+    0#             : Carrier
+    1#             : Carrier
+    isHeytingField : IsHeytingField _≈_ _#_ _+_ _*_ -_ 0# 1#
+
+  open IsHeytingField isHeytingField public
+
+  heyting-cring : HeytingCommutativeRing c ℓ₁ ℓ₂
```
MatthewDaggitt, line 61: Similarly this should be `heytingCommutativeRing`

src/Algebra/Apartness/Structures.agda
MatthewDaggitt: If you do that then you can also import `Algebra.Definitions`, `Algebra.Structures` and `Relation.Binary.Structures` parameterised by the equality relation which will cut down a bit on the boilerplate!

cspollard:
Hi @MatthewDaggitt. Just to say this is still on my todo list, but the end-of-term rush is overwhelming any time I have to look at your last round of comments. My impression is that we are very nearly there, so I will try to finish all of this off at some point during the Christmas break.

cspollard:
ok @MatthewDaggitt did I screw up the rebase somehow? I am seeing many "changes" that are already in master after a rebase, which is unexpected... should I not have done a rebase?

src/Algebra/Apartness/Bundles.agda
cspollard: done now.

src/Algebra/Apartness/Bundles.agda
cspollard: done

MatthewDaggitt:
> ok @MatthewDaggitt did I screw up the rebase somehow? I am seeing many "changes" that are already in master after a rebase, which is unexpected... should I not have done a rebase?

Yes it does look like it's gone wrong somewhere... Err in theory a rebase should have been fine, but we strongly discourage them in favour of merges, as the latter don't screw up other people's version of the branch. As to what went wrong in this one, it looks like somehow you've still got the existing commits at the start, i.e. you've made your changes, added all the intervening changes to the library on top of those in the rebase and then added the same commits again.

As to how to fix it, maybe others have better ideas but I would start a new branch off an up-to-date version of `master`, and cherry-pick the commits from this branch that you want to keep (possibly squashing them beforehand to make the process less time consuming and error prone). Let me know if you'd like me to do it?

cspollard:
Thanks @MatthewDaggitt I'll give that a try tomorrow.

cspollard:
Hi @MatthewDaggitt ,

Sorry for all the trouble. I've restarted and have a working branch that is at the same location, but apparently github closes the PR when the branch is replaced. Would you prefer to re-open this PR, or should I make a new one?

Best,

Chris

MatthewDaggitt:
Whatever is easiest for you 👍

PR #1616, opened 2021-11-01
Make J universe-polymorphic on A

flupe:
Needed `J` to be universe-polymorphic, surprised it wasn't.
Would it be considered a breaking change in the changelog?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -699,7 +699,7 @@ Other minor changes
 * Added new operations in
   `Relation.Binary.PropositionalEquality.Properties`:
   ```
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Nice spot! No, it hasn't yet been released so no entry in the changelog needed.

jespercockx:
I only added it recently, must've been a typo :s

jespercockx:
The changelog also has the (non-fully universe polymorphic) type of J though, so that should be updated as well.

flupe:
> The changelog also has the (non-fully universe polymorphic) type of J though, so that should be updated as well.

Updated.

> I only added it recently, must've been a typo :s

Yes haha, I figured as much.

PR #1619, opened 2021-11-09
Define field and field morphism

Akshobhya1234:
Same as #1582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -450,6 +450,8 @@ Other minor changes
   record UnitalMagma c ℓ : Set (suc (c ⊔ ℓ))
   record Quasigroup  c ℓ : Set (suc (c ⊔ ℓ))
   record Loop c ℓ : Set (suc (c ⊔ ℓ))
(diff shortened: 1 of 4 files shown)


iblech:
Thank you for your contribution! Am I reading this wrong or does this require that `1 / x` is defined for any `x`, even for `x = 0`?

Constructively, there are several non-equivalent definitions of the notion of a field. Particularly important are the following ones:
1. Any element is zero or invertible.
2. Any nonzero element is invertible.
3. Any noninvertible element is zero.

Akshobhya1234:
> Thank you for your contribution! Am I reading this wrong or does this require that `1 / x` is defined for any `x`, even for `x = 0`?
> 
> Constructively, there are several non-equivalent definitions of the notion of a field. Particularly important are the following ones:
> 
> 1. Any element is zero or invertible.
> 2. Any nonzero element is invertible.
> 3. Any noninvertible element is zero.

Yes @iblech, you are right. @MatthewDaggitt  is currently working on #1175

PR #1621, opened 2021-11-14
associativity for vectors

sergey-goncharov:
Unclear, why it was not there. But it is provable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -716,6 +716,15 @@ Other minor changes
          → dsubst₂ C p q (f x₁ y₁) ≡ f x₂ y₂
   ```
 
+* Added vector associativity proof to 
+  `Data/Vec/Relation/Binary/Equality/Setoid.agda`:
+  ```
+  ++-assoc : ∀ {n m k} (xs : Vec A n) → (ys : Vec A m) 
+      → (zs : Vec A k) → (xs ++ ys) ++ zs ≋ xs ++ (ys ++ zs)
+  ++-assoc [] ys zs = ≋-refl
+  ++-assoc (x ∷ xs) ys zs = refl ∷ ++-assoc xs ys zs
+  ```
+
 NonZero/Positive/Negative changes
 ---------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -721,8 +721,6 @@ Other minor changes
   ```
   ++-assoc : ∀ {n m k} (xs : Vec A n) → (ys : Vec A m) 
       → (zs : Vec A k) → (xs ++ ys) ++ zs ≋ xs ++ (ys ++ zs)
-  ++-assoc [] ys zs = ≋-refl
-  ++-assoc (x ∷ xs) ys zs = refl ∷ ++-assoc xs ys zs
   ```
 
 NonZero/Positive/Negative changes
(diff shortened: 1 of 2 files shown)


src/Data/Vec/Relation/Binary/Equality/Setoid.agda
```diff
@@ -68,6 +68,11 @@ open PW public using (++⁺ ; ++⁻ ; ++ˡ⁻; ++ʳ⁻)
 ++-identityʳ []       = []
 ++-identityʳ (x ∷ xs) = refl ∷ ++-identityʳ xs
 
+++-assoc : ∀ {n m k} (xs : Vec A n) → (ys : Vec A m) 
+    → (zs : Vec A k) → (xs ++ ys) ++ zs ≋ xs ++ (ys ++ zs)
```
MatthewDaggitt, line 72: Minor formatting comments:
- You don't need the arrows between named quantified variables so you can write `{n m k} (xs : Vec A n) (ys : Vec A m)  (zs : Vec A k)` on one line.
- According to the style guide arrows go at the end of lines.
- Could you align the `xs` and `zs` in the proof below? Helps with readability!

MatthewDaggitt (commented):
> Unclear, why it was not there.

Nothing is there until someone adds it! 

Thanks, looks like a great addition. In addition to the formatting, please could you add a corresponding entry to the `CHANGELOG.md` file?

sergey-goncharov:
> Nothing is there until someone adds it!

Right. I've been unsure, if there are any reasons for not having it there. Incidentally https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Properties/WithK.agda contains a different proof under K, which confused me, because I nearly though that K is needed for associativity.

MatthewDaggitt:
>  Incidentally https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Properties/WithK.agda contains a different proof under K, which confused me, because I nearly though that K is needed for associativity.

So the difference is that the proof under `WithK` is for (heterogeneous) propositional equality, whereas you have proved it for pointwise equality which is a slightly weaker notion. You haven't proved the results are equal, you've proved that all the elements are equal.

CHANGELOG.md
```diff
@@ -716,6 +716,15 @@ Other minor changes
          → dsubst₂ C p q (f x₁ y₁) ≡ f x₂ y₂
   ```
 
+* Added vector associativity proof to 
+  `Data/Vec/Relation/Binary/Equality/Setoid.agda`:
+  ```
+  ++-assoc : ∀ {n m k} (xs : Vec A n) → (ys : Vec A m) 
+      → (zs : Vec A k) → (xs ++ ys) ++ zs ≋ xs ++ (ys ++ zs)
+  ++-assoc [] ys zs = ≋-refl
+  ++-assoc (x ∷ xs) ys zs = refl ∷ ++-assoc xs ys zs
```
MatthewDaggitt, line 725: We don't tend to add implementations to the CHANGELOG. For consistency's sake, any chance you could remove it?

sergey-goncharov:
> > Incidentally https://github.com/agda/agda-stdlib/blob/master/src/Data/Vec/Properties/WithK.agda contains a different proof under K, which confused me, because I nearly though that K is needed for associativity.
> 
> So the difference is that the proof under `WithK` is for (heterogeneous) propositional equality, whereas you have proved it for pointwise equality which is a slightly weaker notion. You haven't proved the results are equal, you've proved that all the elements are equal.

OK, I am seeing it now. Ineed, I've been just assuming that the pointwise equality is the only desirable notion of equality for vectors, but yes, propositional equality always remains a (potentially problematic) alternative. Thank you for clarification!

MatthewDaggitt:
Thanks, I'm marking this as done, but we're currently having issues with the test suite so I won't merge it in until we resolve that, apologies.

PR #1622, opened 2021-11-14
Bump cabal file package upper bounds

### LINKED ISSUE #1618: Upper bounds of dependencies in `agda-stdlib-utils.cabal`
The test suite is running into problems with the upper bounds of the dependencies (see the build problems in #1613). Is there a reason we have specific minor-version upper bounds (e.g `base >= 4.9.0.0 && < 4.16`) rather than major version upper bounds (e.g. `base >= 4.9.0.0 && < 5.0`)?

MatthewDaggitt:
First attempt at fixing issue #1618 (not linked so as to avoid closing it).

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -148,7 +148,7 @@ jobs:
 
       - name: Golden testing
         run: |
-          ${{ env.CABAL_INSTALL }} --lib clock
+          ${{ env.CABAL_INSTALL }} clock
           make testsuite INTERACTIVE='' AGDA_EXEC='~/.cabal/bin/agda'
 
 
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -30,6 +30,11 @@ on:
 ## To be experimented with again in the future to see if things have
 ## gotten better.
 ##
+## We use `v1-install` rather than `install` as Agda as a community
+## hasn't figured out how to manage dependencies with the new local
+## style builds (see agda/agda#4627 for details). Once this is resolved
+## we should upgrade to `install`.
+##
 ## The AGDA variable specifies the command to use to build the library.
 ## It currently passes the flag `-Werror` to ensure maximal compliance
 ## with e.g. not relying on deprecated definitions.
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -43,7 +43,8 @@ on:
 env:
   GHC_VERSION: 8.6.5
   CABAL_VERSION: 3.2.0.0
-  CABAL_INSTALL: cabal install --overwrite-policy=always --ghc-options='-O1 +RTS -M6G -RTS'
+  CABAL_INSTALL: cabal v1-install --ghc-options='-O1 +RTS -M6G -RTS'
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Okay well that didn't work so I'll simply bump the version of `base` as requested by @andreasabel in #1627 and leave the main issue open.

MatthewDaggitt:
Okay this is very confusing. In the setup for the Golden Tests we're definitely installing `clock`:
```
Starting     clock-0.8.2 (lib)
Building     clock-0.8.2 (lib)
Installing   clock-0.8.2 (lib)
Completed    clock-0.8.2 (lib)
```
but when we try to run the tests we get:
```
_build/MAlonzo/Code/System/Clock/Primitive.hs:16:1: error:
    Could not find module ‘System.Clock’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
16 | import System.Clock
   | ^^^^^^^^^^^^^^^^^^^
```

@gallais any ideas?

gallais:
Pretty sure that's https://github.com/agda/agda/issues/4627

Basically cabal moving to local builds has screwed us and none of us have enough
of an idea / time to change the way the GHC backend compiles & runs the generated
Haskell code.

This new interface is a huge pain and it's not limited to code generated by Agda's
GHC backend. I had a multi-hours struggle yesterday when finally moving my website
from travis to github actions and being forced to upgrade to newer version of ghc &
cabal. I couldn't find any documentation addressing the issue so in the end I just
gave up and changed the recipe so that I now rebuild the whole thing from scratch
every time instead of installing dependencies via `apt`. This has taken my build
times from just a couple of minutes to half an hour.

andreasabel:
If I understand this correctly, you need to set up a `GHC_ENVIRONMENT` that tells which packages shown by `ghc-pkg list` are actually in scope.  E.g. my machine has default environment.
```
$ cat .ghc/x86_64-darwin-9.2.1/environments/default
clear-package-db
global-package-db
package-db /Users/abel/.cabal/store/ghc-9.2.1/package.db
package-id ghc-9.2.1
package-id bytestring-0.11.1.0
package-id nx-2.7.2.2-f15b70f9
package-id base-4.16.0.0
package-id time-1.11.1.1
...
```
Packages installed via `cabal v1-install` and `cabal install --lib` should end up there.

This default environment can be ignored with setting `GHC_ENVIRONMENT=-`.

When running some executable via `cabal`, e.g. `cabal run ...`, then `cabal` sets up the environment appropriatly, according to the package dependencies declared in the `.cabal` file.  
(I am still trying to understand the environment concept fully, these are the bits I know so far.)

MatthewDaggitt:
Okay, so I understand very little about the workings of cabal I'm afraid, so a few naive questions in the interests of getting the test suite up and running again:

- Can we (as in the standard library) retrieve the old behaviour of cabal by prefixing our commands with `v1-`? At least until #4627 gets resolved...
- If not then, can we use @gallais old [hack](https://github.com/gallais/agda-stdlib/pull/1/commits/811113456b69b17242cf47e9e08b2362f5628754) in the first version of the workflow?

If neither then what's the best option? Temporarily disable the golden test suites for the standard library?

gallais:
> If not then, can we use @gallais old hack in the first version of the workflow?

I gave up on this because it's not stable: `cabal init`'s output is version-dependent and it wouldn't work with newer versions

MatthewDaggitt:
Okay switching back to `cabal v1-install` did the trick for the moment. I'm going to merge and then once agda/agda#4627 is resolved we can revisit it.

Issue #1623, opened 2021-11-15
Non-empty Fresh Lists?


fangyi-zhou:
Is there any way to use Data.List.Fresh and Data.List.NonEmpty together for non-empty fresh lists?

MatthewDaggitt:
Not obviously as `Data.List.Fresh` doesn't even use the `List` type. Maybe @gallais has a better idea(?) but I think your best bet that won't end up with you reproving a ton of freshness lemmas would be to lift the predicate `Data.List.Relation.Unary.Unique` to `NonEmpty` lists in a new module `Data.List.NonEmpty.Relation.Unary.Unique`. You could then from a `NonEmptyFreshList` by combining `Data.List.NonEmpty` with an (irrelevant?) uniqueness proof.

It would still require a fair bit of boilerplate but it would at least be mechanical. 

If you do go through with it, we'd welcome any PRs implementing the above suggestion :+1:

gallais:
The easiest thing is probably to create `Data.List.Fresh.NonEmpty`
modeled after `Data.List.NonEmpty`: a simple record type corresponding
to a cons.

MatthewDaggitt:
Doh, yes that would also work.

fangyi-zhou:
So would it be easier to do Fresh.NonEmpty instead of NonEmpty.Fresh?

MatthewDaggitt:
Yes.

MatthewDaggitt:
Fixed with #1624

PR #1624, opened 2021-11-15
Data.List.Fresh.NonEmpty

fangyi-zhou:
#1623

### DIFF
--- a/src/Data/List/Fresh/NonEmpty.agda
+++ b/src/Data/List/Fresh/NonEmpty.agda
@@ -1,0 +1,53 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- A non-empty fresh list
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Fresh.NonEmpty where
+
+open import Level using (Level; _⊔_)
+open import Data.List.Fresh as List# using (List#; []; cons; fresh)
+open import Data.Maybe.Base using (Maybe; nothing; just)
+open import Data.Nat using (ℕ; suc)
+open import Data.Product using (_×_; _,_)
+open import Data.Unit.Polymorphic.Base using (⊤; tt)
+open import Relation.Binary as B using (Rel)
+
+private
+  variable
+    a r : Level
+    A : Set a
+    R : Rel A r
+
+module _ {a} (A : Set a) (R : Rel A r) where
+
+  infixr 5 _∷#⁺_
+  record List#⁺ : Set (a ⊔ r) where
+    constructor _∷#⁺_
+    field
+      head : A
+      tail : List# A R
+      {rel} : fresh A R head tail
+
+open List#⁺
+
+uncons : List#⁺ A R → A × List# A R
+uncons (x ∷#⁺ xs) = x , xs
+
+[_] : A → List#⁺ A R
+[ x ] = x ∷#⁺ []
+
+length : List#⁺ A R → ℕ
+length (x ∷#⁺ xs) = suc (List#.length xs)
+
+-- Conversion
+
+toList# : List#⁺ A R → List# A R
+toList# l = cons (l .head) (l .tail) (l .rel)
+
+fromList# : List# A R → Maybe (List#⁺ A R)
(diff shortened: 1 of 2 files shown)


fangyi-zhou:
I've put down some initial code for the proposal, could you let me know what other features are needed before it's ready for ship?

(Also: I'm stilll learning agda, please let me know if I did something wrong)

src/Data/List/Fresh/NonEmpty.agda
```diff
@@ -0,0 +1,47 @@
+{-# OPTIONS --without-K --safe #-}
```
gallais, line 1: You'll need a header briefly describing the module here (cf. the start of other files).

src/Data/List/Fresh/NonEmpty.agda
```diff
@@ -0,0 +1,47 @@
...
+module _ {a} (A : Set a) (R : Rel A r) where 
+
+  record List#⁺ : Set (a ⊔ r) where 
+    field
+      head : A
+      tail : List# A R
+      rel : fresh A R head tail
+
+    infixr 5 _∷#_
+    pattern _∷#_ x xs = record { head = x; tail = xs }
```
gallais, line 27: Unfortunately this is likely to clash with the `List#` one because we
do not have (yet) type-directed disambiguation of pattern synonyms.
You could maybe instead have:

```suggestion
  infixr 5 _∷#⁺_
  record List#⁺ : Set (a ⊔ r) where
    constructor _∷#⁺_
    field
      head  : A
      tail  : List# A R
      {rel} : fresh A R head tail
```
?

fangyi-zhou:
Actually I realised there is a NonEmpty constructor in Data.List.Fresh.Base, I'm no longer sure whether this module is necessary?

MatthewDaggitt:
So the difference is that the existing `NonEmpty` definition is a predicate over lists rather than a type in of itself. I think your new definition still has value :+1:

src/Data/List/Fresh/NonEmpty.agda
```diff
@@ -0,0 +1,52 @@
...
+  record List#⁺ : Set (a ⊔ r) where
+    constructor _∷#⁺_
+    field
+      head : A
+      tail : List# A R
+      {rel} : fresh A R head tail
+
+  open List#⁺
+
+  uncons : List#⁺ → A × List# A R
```
MatthewDaggitt, line 36: I don't think you mean to put these under `module _ {a} (A : Set a) (R : Rel A r) where`. Currently it means that every time you use operations like `uncons` you'll have to pass in the `A` and `R` explicitly, and the singleton list syntax `[_]` will be unusable. Better to make `R` a variable and instead write:
```
uncons : List#⁺ A R → A × List# A R
```
outside the anonymous module. This is more readable as well.

src/Data/List/Fresh/NonEmpty.agda
```diff
@@ -0,0 +1,52 @@
...
+  uncons (x ∷#⁺ xs) = x , xs
+
+  [_] : A → List#⁺
+  [ x ] = record { head = x; tail = []; rel = tt}
```
MatthewDaggitt, line 40: Can you not use the `_∷#⁺_` constructor here? I would think that Agda should be able to infer `tt`?

src/Data/List/Fresh/NonEmpty.agda
```diff
@@ -0,0 +1,52 @@
...
+open import Data.Maybe.Base using (Maybe; nothing; just)
+open import Data.Nat using (ℕ; suc)
+open import Data.Product using (_×_; _,_)
+open import Data.Unit.Polymorphic.Base using (⊤; tt)
+open import Relation.Binary as B using (Rel)
+
+private
+  variable
+    a r : Level
+    A : Set a
```
MatthewDaggitt, line 22: I would add `R` to the set of variables as well.

MatthewDaggitt:
Other than that you'll also need to an entry in the `CHANGELOG.md` file.

I should have said thank you for the PR at the start as well, it's a great first contribution :smile:

MatthewDaggitt:
Merging in, thanks for the PR!

PR #1638, opened 2021-12-02
Fix implicit arguments in +-distrib-/-∣ʳ

tonyxty:
In `+-distrib-/-∣ʳ`,  `m` should be explicit and `n` should be implicit.


MatthewDaggitt:
Thank you, nice find! Can you add a note to `CHANGELOG.md` under the bug fix section?

MatthewDaggitt:
Awesome, thank you for the PR.

PR #1641, opened 2021-12-04
Refactor reflective interface for ring solver

MatthewDaggitt:
I made the mistake of trying to use the new ring solver in anger over rational numbers and found that it's not very usable at all. So my mission this weekend is to try and fix some of the outstanding issues with it.

This initial PR is simply a refactoring to massively clean up the code and doesn't actually change anything. In particular I've got rid of `Tactic.RingSolver.Core.ReflectiveHelper`, pushing some of it into `Reflection.Term` and new `Reflection` modules for individual types such as `Data.List.Reflection`. 

I've also added the convention that names of reflected variable should be preceded by a single back-tick. Anyone have any objections?


JacquesCarette (commented):
Very nice.

MatthewDaggitt:
As people have had a look and okayed it I'm going to break my own rules and merge this in now, as I have a series of PRs blocked up behind it that I would quite like reviewed individually.

PR #1647, opened 2021-12-08
Quasigroup and Loop morphism

Akshobhya1234:


MatthewDaggitt:
Looks good to me  :+1:  Again I'll leave it up for a little bit in case anyone has comments.

Issue #602, opened 2019-01-21
 add proof for ``¬ lex< xs []`` ?


mechvel:
``Data.List.Relation.Lex.Core``  defines `` _<_ = Lex P _≈_ _≺_``,

and I discover that ``\_ -> \lambda()`` does not prove   ``∀ xs → ¬ _<_  xs  []``
May be, it worth to add there (or somewhere)
```
 lex≮[] : ∀ xs → ¬ _<_ xs []
 lex≮[] [] (base ()) 
```
?

gallais:
I'd write the type as `¬ (xs < [])` and would call it `xs≮[]` but otherwise yes that seems reasonable.

MatthewDaggitt:
Yup, I agree, useful addition and `xs≮[]` is the best name. @mechvel would you like to put together a pull request?

mechvel:
> @mechvel would you like to put together a pull request?

I do not understand (do not recall) what is a pull request.
Probably standard library has an administrator (Matthew?). Has it?
As he agrees with the proposal of
```
 xs≮[] : ∀ xs → ¬ (xs < [])
 xs≮[] [] (base ()) 
```
, then he adds this code to Data.List.Relation.Lex.Core for the next library version, and will add a notice to CHANGELOG.md.
Is this a way, for the library to develop?

MatthewDaggitt:
@mechvel a pull request (PR) is exactly what we helped you through in #65. You may also find out how to prepare one in the [HACKING](https://github.com/agda/agda-stdlib/blob/master/HACKING.md) file.

Although I could do it myself, it works better if everyone else helps to. This way you yourself can contribute to the library, and in future can help make bigger changes such as incorporating your Bin library.

mechvel:
> @mechvel a pull request (PR) is exactly what we helped you through in #65. You may also find out 
> how  to prepare one in the HACKING file if you've forgotten.

I had difficulties with pull request, was not able to accomplish it without your help.
After Agda + library becomes close enough to an official release, I shall decide whether to try to move with the  Binary-4* proposal. Then I shall need to revisit the pull request matter (may be this will be easier this time, because I have now a fresh browser).

jamesmckinna:
While I am at it: PR #1648 only resolved this issue for `List`. There's a corresponding lemma for `Vec` to add... another day.

jamesmckinna:
[...]

PR #1648, opened 2021-12-08
Resolve Issue #602:new lemma added

jamesmckinna:
Proof `xs≮[] : ∀ xs → ¬ xs < []` added as per discussion. 
CHANGELOG.md suitably updated. 
Old proof of `¬[]<[] : ¬ [] < []` now defined to be`xs≮[] []` (and definitionally equal to previous definition)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -664,6 +664,11 @@ Other minor changes
   sign-cong′ : s₁ ◃ n₁ ≡ s₂ ◃ n₂ → s₁ ≡ s₂ ⊎ (n₁ ≡ 0 × n₂ ≡ 0)
   ```
 
+* Added new proofs in `Data.List.Relation.Binary.Lex.Strict`:
+  ```agda
+  xs≮[] : ∀ xs → ¬ xs < []
+  ```
+
 * Added new proofs in `Data.Nat.Properties`:
   ```agda
   n+1+m≢m   : n + suc m ≢ m
--- a/src/Data/List/Relation/Binary/Lex/Strict.agda
+++ b/src/Data/List/Relation/Binary/Lex/Strict.agda
@@ -45,8 +45,11 @@ module _ {a ℓ₁ ℓ₂} {A : Set a} where
       _≋_ = Pointwise _≈_
       _<_ = Lex-< _≈_ _≺_
 
+    xs≮[] : ∀ xs → ¬ xs < []
+    xs≮[] _ (base ())
+
     ¬[]<[] : ¬ [] < []
-    ¬[]<[] (base ())
+    ¬[]<[] = xs≮[] []
 
     <-irreflexive : Irreflexive _≈_ _≺_ → Irreflexive _≋_ _<_
     <-irreflexive irr (x≈y ∷ xs≋ys) (this x<y)     = irr x≈y x<y


jamesmckinna:
Aaargh!
Trying to pick off low-hanging fruit, and practise PR-fu, and I have left-over commits from last PR? How? Why? Sorry!

gallais:
It's probably better to create a new branch for each new feature. This is my workflow
(my `main` points to the `agda/agda-stdlib` remote and I have `gallais` as the remove for `gallais/agda-stdlib`)

```
git checkout main
git pull
git checkout -b my_new_feature
git add -p
git commit
git push gallais -u my_new_feature
```

jamesmckinna:
OK, thanks for the advice. 
Meanwhile, FWIW, only files actually touched by this PR are `CHANGELOG.md` and `src/Data/List/Relation/Binary/Lex/Strict.agda`
Will attempt to tidy up tomorrow or Friday.

gallais:
After adding your remote to my local copy, I've:

```
git checkout master
git pull
git checkout -b issue602
git cherry-pick 837b9b3dbacc1b420977d9e974182fcfa93a9c7e 
git push jamesmckinna -u issue602:master
```
and now that should have fixed the PR. :crossed_fingers:

jamesmckinna:
is the simplest thing to do to close this request, and start a fresh one, as per @gallais hint?

On which point, is it possible to achieve the same effects as he suggested, 
by editing my `.git/config` so that (IIUC)
the 'origin' points to `agda/agda-stdlib`, while  the 'remote' points to my own repo `jamesmckinna/agda-stdlib`?

If so, a nudge in that direction would be helpful!

jamesmckinna:
(sorry, further out-of-order comments). [crawls back under rock]

MatthewDaggitt:
> is the simplest thing to do to close this request, and start a fresh one, as per @gallais hint?

So @gallais's first sequence of commands to create a new branch off of `master` and then work on that is for if you are trying to start a *new* PR and you haven't done any work yet. In general if your Git-fu isn't terribly strong and it's a small PR that's often the easier thing to do. It may result in you manually having to copy your changes back over on the new branch though.

To avoid having to do that copying, what @gallais has just done in his second sequence of commands is to create a new branch off of `master`, but instead take the relevant commit you made to your branch and add it to his new branch, then overwrite the reference to your branch with his branch. Probably not recommended for the unconfident.

> On which point, is it possible to achieve the same effects as he suggested,
by editing my .git/config so that (IIUC)
the 'origin' points to agda/agda-stdlib, while the 'remote' points to my own repo jamesmckinna/agda-stdlib?

I'm afraid I don't know about the internals of `.git/config`, but you _could_ do that via a selection of commands from the `git remote` command line mode. However I don't think that it would achieve the effect you're after, and it would be *very* confusing to anyone else trying to help you in future :sweat_smile: So in summary, please don't do that :smile:

MatthewDaggitt:
Other than that the content of the PR looks great :+1:

gallais:
I'll revise the `HACKING` so that it suggests the branch-based method instead of telling people to checkout master
& work on that.

MatthewDaggitt:
> I'm afraid I don't know about the internals of .git/config, but you could do that via a selection of commands from the git remote command line mode. However I don't think that it would achieve the effect you're after, and it would be very confusing to anyone else trying to help you in future sweat_smile So in summary, please don't do that smile

Ah sorry, @jamesmckinna please ignore this comment about not doing this, I got slightly confused about your use of the word "remote". In technical Git parlance both `agda/agda-stdlib` and `jamesmckinna/agda-stdlib` are "remotes". 

Rather than doing it by `.git/config`, you can add a remote by doing `git remote add NAME https://github.com/agda/agda-stdlib` and you can rename a remote using `git remote rename OLD_NAME NEW_NAME`

Issue #1350, opened 2020-11-09
`reverse` in `Data.Fin.Permutation.Components` duplicates `opposite` in `Data.Fin.Base`


MatthewDaggitt:
The former should be deprecated, and any proofs transferred over to the latter.

PR #1650, opened 2021-12-09
Deprecate `reverse` in `Data.Fin.Permutation.Components`

Akshobhya1234:
Will resolve issue #1350

### DIFF
--- a/src/Data/Fin/Permutation/Components.agda
+++ b/src/Data/Fin/Permutation/Components.agda
@@ -51,28 +51,6 @@ transpose-inverse i j {k} with k ≟ j
 ...   | false because [k≢i] rewrite dec-false (k ≟ i) (invert [k≢i])
                                   | dec-false (k ≟ j) (invert [k≢j]) = refl
 
-reverse-prop : ∀ {n} → (i : Fin n) → toℕ (opposite i) ≡ n ∸ suc (toℕ i)
-reverse-prop {suc n} zero = toℕ-fromℕ n
-reverse-prop {suc n} (suc i) = begin
-  toℕ (inject₁ (opposite i)) ≡⟨ toℕ-inject₁ (opposite i) ⟩
-  toℕ (opposite i)           ≡⟨ reverse-prop i ⟩
-  n ∸ suc (toℕ i)            ∎
-
-reverse-involutive : ∀ {n} → Involutive _≡_ (opposite {n})
-reverse-involutive {suc n} i = toℕ-injective (begin
-  toℕ (opposite (opposite i)) ≡⟨ reverse-prop (opposite i) ⟩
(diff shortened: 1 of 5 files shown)


src/Data/Fin/Permutation/Components.agda
```diff
@@ -56,23 +51,38 @@ transpose-inverse i j {k} with k ≟ j
 ...   | false because [k≢i] rewrite dec-false (k ≟ i) (invert [k≢i])
                                   | dec-false (k ≟ j) (invert [k≢j]) = refl
 
-reverse-prop : ∀ {n} → (i : Fin n) → toℕ (reverse i) ≡ n ∸ suc (toℕ i)
-reverse-prop {suc n} i = begin
-  toℕ (inject≤ (n ℕ- i) _)  ≡⟨ toℕ-inject≤ _ (ℕₚ.m∸n≤m (suc n) (toℕ i)) ⟩
-  toℕ (n ℕ- i)              ≡⟨ toℕ‿ℕ- n i ⟩
-  n ∸ toℕ i                 ∎
+reverse-prop : ∀ {n} → (i : Fin n) → toℕ (opposite i) ≡ n ∸ suc (toℕ i)
```
MatthewDaggitt, line 54: It feels like these properties should be deprecated as well? If they're not in `Data.Fin.Properties` then they should also be copied there with updated names?

The names of the other proofs are fine once you substitute `opposite` for `reverse`, but I think this one should really be called `toℕ-opposite`?

src/Data/Fin/Permutation/Components.agda
```diff
@@ -56,23 +51,38 @@ transpose-inverse i j {k} with k ≟ j
...
+reverse-prop {suc n} zero = toℕ-fromℕ n
+reverse-prop {suc n} (suc i) = begin
+  toℕ (inject₁ (opposite i)) ≡⟨ toℕ-inject₁ (opposite i) ⟩
+  toℕ (opposite i)           ≡⟨ reverse-prop i ⟩
+  n ∸ suc (toℕ i)            ∎
 
-reverse-involutive : ∀ {n} → Involutive _≡_ (reverse {n})
+reverse-involutive : ∀ {n} → Involutive _≡_ (opposite {n})
```
MatthewDaggitt, line 61: I know it's not your code, but we've since found the slightly nicer way of writing these as: `Involutive {A = Fin n} _≡_ opposite` which it would be nice to update to.

MatthewDaggitt (commented):
Thanks!

src/Data/Fin/Permutation/Components.agda
Akshobhya1234: Done

MatthewDaggitt:
Awesome thanks, the only thing to do is to update the changelog with the new changes!

Issue #526, opened 2018-11-09
Add `(run/eval/exec)State` to `Monad.State`


gallais:
```haskell
runState :: State s a -> s -> (a, s)
evalState :: State s a -> s -> a
execState :: State s a -> s -> s
```

jamesmckinna:
I'll bite... added as branch 'issue526' to my agda-stdlib on GitHub. I'll make a PR.

PR #1652, opened 2021-12-09
Tutorial on setting up agda-stdlib for a first contribution

gallais:
ping @jamesmckinna


jamesmckinna:
Thanks, I attempted to follow the instructions, but as at #526, seem to have ended up with commits from #602 on the branch. Part of the problem was I think that `jamesmckinna/agda-stdlib` would not initially accept accept the `push -u issue526` without rejecting the commits. I will try again with everything fresh, once my own sense of frustration/bewilderment/sense of inadequacy has abated sufficiently.

jamesmckinna:
Recommend to all (especially beginners!) these new instructions (which, it turns out, are not so very dissimilar from those already on GitHub for managing forks and PRs), but localised to the `agda/agda-stdlib` situation. After a drastic cleanup of my broken setup, things went smothly.

MatthewDaggitt:
Thanks @gallais !

MatthewDaggitt:
Hmm actually should this actually be moved into `HACKING.md` and replace it's first section? Feels a bit like duplication to have it in a separate file entirely. @gallais ?

gallais:
I see HACKING is already linking to the style guide in `notes/` so we could
perhaps remove the git section from it and similarly link to the corresponding note?

MatthewDaggitt:
Yup, that sounds reasonable!

MatthewDaggitt:
I tried to separate out the Git part out of it, but it just didn't work and I ended up moving most of it to your notes. At which point `HACKING.md` was pretty empty... So I just moved it all back into `HACKING.md`.

MatthewDaggitt:
It's much improved though, thank you @gallais!

PR #1660, opened 2021-12-16
Add new exponentiation operator to `Data.Integer` (fixes #490)

### LINKED ISSUE #490: Missing properties in Data.Integer.Properties
Looking at https://github.com/MrChico/Reals-in-agda, the author needed some extra
properties on the integers. It looks like we are still missing them:

* [x] `∣⊖∣-≤ : ∀ {m n} → m ℕ.≤ n → + ∣ m ⊖ n ∣ ≡ n ⊖ m`
* [x] `∣-∣-≤ : ∀ {m n} → m ℤ.≤ n → + ∣ m - n ∣ ≡ n - m`
* [x] `Trans ≤ < <`
* [x] `Trans < ≤ <`
* [x] `_+_ Preserves₂ _≤_ ⟶ _≤_ ⟶ _≤_`
* [x] `_+_ Preserves₂ _<_ ⟶ _<_ ⟶ _<_`
* [x] `_^_ : ℤ -> ℕ -> ℤ` (+ properties?)

You can have a look at [the fork](https://github.com/MrChico/agda-stdlib) but I think I have identified a good chunk of the missing
lemmas.

jamesmckinna:
Some of the names in `Algebra.Properties.Semiring.Exp.TailRecursiveOptimised` might not be as conformant as desirable/desired... but hey?!


CHANGELOG.md
```diff
@@ -632,6 +637,12 @@ Other minor changes
                     CommutativeRing (a ⊔ b) (ℓ₁ ⊔ ℓ₂)
  ```
 
+* Added new functions to `Algebra.Definitions.RawSemiring`:
+  ```agda
+  _^[_]*_ : A → ℕ → A → A
+  _^ᵗ_     : A → ℕ → A
+  ```
```
MatthewDaggitt, line 644: I think we're missing entries for the addition of `_^_` to `Data.Integer.Base` and the proofs to `Data.Integer.Properties`?

src/Algebra/Properties/Semiring/Exp/TCOptimised.agda
```diff
@@ -29,7 +29,7 @@ open import Algebra.Definitions.RawSemiring rawSemiring public
   using () renaming (_^′_ to _^_)
 
 ------------------------------------------------------------------------
--- Properties of _×_
+-- Properties of _^_
```
MatthewDaggitt, line 32: Nice spot, thanks!

MatthewDaggitt (commented):
Thanks for the PR, looks great!

src/Algebra/Properties/Semiring/Exp/TailRecursiveOptimised.agda
```diff
@@ -0,0 +1,82 @@
...
+open Semiring S renaming (zero to *-zero)
+open import Relation.Binary.Reasoning.Setoid setoid
+open import Algebra.Properties.Semiring.Exp S as U
+  using (_^_)
+
+------------------------------------------------------------------------
+-- Re-export definition from the monoid
+
+open import Algebra.Definitions.RawSemiring rawSemiring public
+  using (_^[_]*_; _^ᵗ_)
```
MatthewDaggitt, line 26: It occurs to me that it would be better to simply rename the operator to `_^ᵗ_` here (as well as the accompanying properties)? It's pretty unlikely users will want to import multiple different versions of exponentiation so let's make their life easier for them. In the unlikely case they do want to import multiple versions they can always qualify them.

jamesmckinna:
So... indeed my attempts to avoid the `library-design` discussion did leave open exactly which definitions would be wanted where. And I wasn't clear if each of `Data.Nat` and `Data.Integer` would want to import the definition `Algebra.Definitions.RawSemiring._^_` or define it directly... (which they do at the moment) never mind the properties...
(sorry huge context switch between tasks at the moment, so I'm not clear what you're asking for)

jamesmckinna:
And indeed not sure how/why the changes hadn't propagated to`Data.Integer.*`... and I don't want to screw up the PR by doing anything clumsy right now. Offline, if you spelt out what's missing and where, I'll fix it... probably after Christmas at this rate

src/Algebra/Properties/Semiring/Exp/TailRecursiveOptimised.agda
jamesmckinna: Do you mean rename to `_^_`?

MatthewDaggitt: Yes, sorry `_^ᵗ_` to `_^_`

MatthewDaggitt:
> So... indeed my attempts to avoid the library-design discussion did leave open exactly which definitions would be wanted where. And I wasn't clear if each of Data.Nat and Data.Integer would want to import the definition Algebra.Definitions.RawSemiring._^_ or define it directly... (which they do at the moment) never mind the properties...
(sorry huge context switch between tasks at the moment, so I'm not clear what you're asking for)

I think it's fine to for `Data.Integer.Base` to define `_^_` separately for now, as that's what `Data.Nat.Base` does.

> And indeed not sure how/why the changes hadn't propagated toData.Integer.*... and I don't want to screw up the PR by doing anything clumsy right now. Offline, if you spelt out what's missing and where, I'll fix it... probably after Christmas at this rate

It's simply that you haven't added an entry in the CHANGELOG for the changes to `Data.Integer.*`?

jamesmckinna:
A-ha. Adding the properties to `Data.Integer.Properties` had I think got derailed by moving the tail-recursive proofs to `Algebra.Properties.Semiring.Exp.TailRecursiveOptimised`, where, incidentally (?) the proof that to `_^ᵗ_` is equal to `_^_` resides, so it seems as though both are required? (I may have got confused about what gets exported where in the end of all this)

MatthewDaggitt:
So generally we don't list the individual proofs in new modules, only new proofs to existing modules, so your current entry to the CHANGELOG for `Algebra.Properties.Semiring.Exp.TailRecursiveOptimised` is sufficient :+1:

jamesmckinna:
So... I think how this started was that proving properties of exponentiation (homomorphism properties of various kinds) was/would be a pain without introducing the tail-recursive form. And then you had suggested splitting the work between `...Integer...` and `...rawSemiring...` which actually now leaves `Data.Nat.Properties` with a lot of handrolled proofs about `_^_` and `Data.Integer.Properties` with... so far none (and their proofs are rather different, both to state and to prove, than the `Nat` case)... so perhaps opening a PR was premature, unless we agree (?) that proving them via `Algebra.Properties.Semiring.Exp.TailRecursiveOptimised` might be worthwhile...?

jamesmckinna:
is there anywhere a proof that ` ¬(0ℤ ≡ 1ℤ)`?

jamesmckinna:
That may now be enough for a while...
Currently still missing: 
```agda
* [ ] i*j≡1⇒[i≡1∧j≡1]∨[i≡-1∧j≡-1] : ∀ i {j} → i * j ≡ 1ℤ → (i ≡ 1ℤ × j ≡ 1ℤ) ⊎ (i ≡ - 1ℤ × j ≡ - 1ℤ)
* [ ] i^n≡1⇒[n≡0]∨[i≡1∧n≡2k]∨[i≡-1∧n≡2k+1] : ∀ i {n} → i ^ n ≡ 1ℤ → (n ≡ 0) ⊎ Σ [ k ∈ ℕ] (i ≡ 1ℤ × n ≡ 2 k +2) ⊎ (i ≡ - 1ℤ × n ≡ 2 k +1)
```

MatthewDaggitt:
> And then you had suggested splitting the work between ...Integer... and ...RawSemiring... which actually now leaves Data.Nat.Properties with a lot of existing, but handrolled, proofs about _^_ and Data.Integer.Properties with... so far none (and their proofs are rather different, both to state and to prove, than the Nat case)... so perhaps opening a PR was premature, unless we agree (?) that proving them via Algebra.Properties.Semiring.Exp.TailRecursiveOptimised might be worthwhile...? Or else rethink?

So as I said in the issue, I think in the interests of not dragging in the dreaded `library-design` tag, what you've added here is fine! Yes, there's a lot of duplication, but it's consistent with `Data.Nat` so they can be fixed together at some later date. What you've added is definitely an improvement :+1: 

> is there anywhere a proof that ¬(0ℤ ≡ 1ℤ)?

No, because you can always prove it directly by `λ()` so I don't think we should add it. There's a lot of similar inequalities one could add and it might be a slippery slope :wink: 

> it still leaves open the _^_ vs. _^ᵗ_ question for now

I would be strongly in favour of exposing `_^_` in `Exp.TailRecursiveOptimised`, not least in order to be consistent with `Exp.TCOptimised` which also renames the operator it exports.

> Currently still missing from Data.Integer.Properties:

I think leave those two for the moment, this PR is already moderately large with some non trivial decisions in it so let's focus on getting this in :smile:

jamesmckinna:
OK, I'll revert the addition of `¬(0ℤ ≡ 1ℤ)` and push the tail-recursive export. Try that now.

jamesmckinna:
And yes, statement of the two 'missing' lemmas should be revised to be arithmetically correct... ;-)

CHANGELOG.md
```diff
@@ -680,9 +691,22 @@ Other minor changes
...
 
+* Added new functions in `Data.Integer.Base`:
+  ```
+  _^_ : ℤ → ℕ → ℤ
+  ```
+
 * Added new proofs in `Data.Integer.Properties`:
   ```agda
   sign-cong′ : s₁ ◃ n₁ ≡ s₂ ◃ n₂ → s₁ ≡ s₂ ⊎ (n₁ ≡ 0 × n₂ ≡ 0)
+  ¬[1≡0] : ¬ 1ℤ ≡ 0ℤ
```
MatthewDaggitt, line 702: This entry should be removed and entries for `∣-∣-≤`, `∣⊖∣-≤` and `≤-⊖` added.

MatthewDaggitt (commented):
Once that's fixed, looks good to go!

CHANGELOG.md
jamesmckinna: Working on that now.

jamesmckinna: Good to go now, I think.

PR #1661, opened 2021-12-16
Add definitions of compositeness

### LINKED ISSUE #1659: Data.Nat.Primality: add idea of Compositeness
Currently `Data.Nat.Primality` defines:

```agda
Prime : ℕ → Set
Prime 0             = ⊥
Prime 1             = ⊥
Prime (suc (suc n)) = (i : Fin n) → 2 + toℕ i ∤ 2 + n
```

This is fine, but in many situations a negative proof of `Prime n` is a lot less useful than knowing even one factor that causes it to fail the condition.

I'm proposing to add something like

```agda
Composite : ℕ → Set
Composite 0 = ⊥
Composite 1 = ⊥
Composite (suc (suc n)) = Σ[ i ∈ Fin n ] 2 + toℕ i ∣ 2 + n
```

with a corresponding

```agda
composite? : Decidable Composite
```
implemented similarly to `prime?`, along with proofs relating `Composite` and `Prime` together.

Taneb:
Closes #1659

### DIFF
--- a/src/Data/Nat/Primality.agda
+++ b/src/Data/Nat/Primality.agda
@@ -49,17 +49,17 @@ prime? (suc (suc n)) = all? (λ _ → ¬? (_ ∣? _))
 
 -- Relation between compositeness and primality
 
-composite⇒¬prime : ∀ n → Composite n → ¬ Prime n
+composite⇒¬prime : ∀ {n} → Composite n → ¬ Prime n
 composite⇒¬prime (suc (suc n)) (i , 2+i∣2+n) 2+n-prime = 2+n-prime i 2+i∣2+n
 
-¬composite⇒prime : ∀ n → 2 ≤ n → ¬ Composite n → Prime n
+¬composite⇒prime : ∀ {n} → 2 ≤ n → ¬ Composite n → Prime n
 ¬composite⇒prime (suc (suc n)) (s≤s (s≤s _)) ¬n-composite i 2+i∣2+n = ¬n-composite (i , 2+i∣2+n)
 
-prime⇒¬composite : ∀ n → Prime n → ¬ Composite n
+prime⇒¬composite : ∀ {n} → Prime n → ¬ Composite n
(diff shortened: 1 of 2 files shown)


src/Data/Nat/Primality.agda
```diff
@@ -32,9 +47,28 @@ prime? 0             = no λ()
 prime? 1             = no λ()
 prime? (suc (suc n)) = all? (λ _ → ¬? (_ ∣? _))
 
+-- Relation between compositeness and primality
+
+composite⇒¬prime : ∀ n → Composite n → ¬ Prime n
```
MatthewDaggitt, line 52: Can the `n` be made implicit in these 4 proofs or does the use of `⊥` prevent that?

src/Data/Nat/Primality.agda
```diff
@@ -10,14 +10,29 @@ module Data.Nat.Primality where
...
+Composite 0 = ⊥
+Composite 1 = ⊥
+Composite (suc (suc n)) = Σ[ i ∈ Fin n ] 2 + toℕ i ∣ 2 + n
+
+-- Decision procedure for compositeness
+
+composite? : Decidable Composite
+composite? 0 = no λ()
+composite? 1 = no λ()
+composite? (suc (suc n)) = any? λ _ → _ ∣? _
```
MatthewDaggitt, line 34: Minor point, but can you bracket the lambda? I know it's not necessary, but it greatly helps reading it!

MatthewDaggitt (commented):
Otherwise, looks great, thanks!

src/Data/Nat/Primality.agda
Taneb: I think it should be fine! All of these proofs have _something_ positive, either a proof of compositeness or primality or a proof of being greater than or equal to 2, that can determine n, as an argument.

src/Data/Nat/Primality.agda
Taneb: Sure thing! I agree, it does make it a but more readable

PR #1662, opened 2021-12-17
private variable in Maybe

guilhermehas:
With this modification, variables will be at the top like in most files of this repository.

### DIFF
--- a/src/Data/Maybe/Categorical.agda
+++ b/src/Data/Maybe/Categorical.agda
@@ -8,6 +8,7 @@
 
 module Data.Maybe.Categorical where
 
+open import Level
 open import Data.Maybe.Base
 open import Category.Functor
 open import Category.Applicative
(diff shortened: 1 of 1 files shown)


src/Data/Maybe/Categorical.agda
```diff
@@ -8,34 +8,43 @@
...
 open import Category.Monad
 import Function.Identity.Categorical as Id
 open import Function
 
+private
+  variable
+    a b f m : Level
+    A : Set a
+    B : Set b
+    F : Set f → Set f
```
MatthewDaggitt, line 24: So the current [policy](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#variables) for `variable`s is that they should only be declared over levels/sets and the "main" types that the file concerns itself with, which in this case is `Maybe`. So I don't think `F` and `M` qualify under this policy. Indeed they are only used once in the file at the moment, so I think far better to keep them local!

Otherwise happy to keep the levels and the sets!

MatthewDaggitt (commented):
Thanks for the first PR though, greatly appreciate the changes!

guilhermehas:
I removed F and M from variables.

MatthewDaggitt:
Thanks for the changes!

PR #1667, opened 2021-12-22
Switch to new function hiearchy in several places

Taneb:
I was annoyed that `Data.Fin.Properties` depended on both the old and the new hierarchy, and pulled the thread.

Part of #759

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
@@ -346,7 +346,6 @@ module _ {a b r} {A : Set a} {B : Set b} {R : REL A B r} where
   toAny∘fromAny≗id (there p) = P.cong there (toAny∘fromAny≗id p)
 
   Sublist-[x]-bijection : ∀ {x xs} → (Sublist R [ x ] xs) ⤖ (Any (R x) xs)
-  -- Sublist-[x]-bijection = mk⤖ toAny fromAny toAny-injective toAny∘fromAny≗id
   Sublist-[x]-bijection = mk⤖ (toAny-injective , < fromAny , toAny∘fromAny≗id >)
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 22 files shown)


src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -347,7 +346,8 @@ module _ {a b r} {A : Set a} {B : Set b} {R : REL A B r} where
   toAny∘fromAny≗id (there p) = P.cong there (toAny∘fromAny≗id p)
 
   Sublist-[x]-bijection : ∀ {x xs} → (Sublist R [ x ] xs) ⤖ (Any (R x) xs)
-  Sublist-[x]-bijection = bijection toAny fromAny toAny-injective toAny∘fromAny≗id
+  -- Sublist-[x]-bijection = mk⤖ toAny fromAny toAny-injective toAny∘fromAny≗id
```
MatthewDaggitt, line 349: Rogue comment?

src/Data/Vec/Relation/Binary/Pointwise/Extensional.agda
```diff
@@ -142,7 +141,7 @@ Pointwise-≡⇒≡ {xs = x ∷ xs} {y ∷ ys} xs∼ys     =
 
 Pointwise-≡↔≡ : ∀ {n} {xs ys : Vec A n} → Pointwise _≡_ xs ys ⇔ xs ≡ ys
 Pointwise-≡↔≡ {ℓ} {A} =
-  Equiv.equivalence Pointwise-≡⇒≡ ≡⇒Pointwise-≡
+  mk⇔ Pointwise-≡⇒≡ ≡⇒Pointwise-≡
```
MatthewDaggitt, line 144: This now fits up on to the previous line now!

MatthewDaggitt (commented):
Thanks @Taneb this really looks amazing. I guess before we release v2.0 we should relaly provide some conversion functions from the old to the new bundles, so that if people want to stick with the old definitions, they can prepend these proofs with `fromNewEquivalence` or something similar.

Issue #1610, opened 2021-10-15
cast vs subst in Data.Fin


conal:
I’m working on a collection of proofs for submission to `Data.Fin.Properties`. I can phrase & prove via `cast` or via `subst Fin`. For my own use, I need the `subst` versions, to fit into a setting more general than `Fin`. If `cast` is a better fit for `Fin`, I’m happy to adapt via an equivalence property `cast≗subst : ∀ {m n} (m≡n : m ≡ n) → cast m≡n ≗ subst Fin m≡n` (which could also go into `Data.Fin.Properties`).

Going with `cast`, I also need some properties of `cast` corresponding to ones about `subst` in `Relation.Binary.PropositionalEquality.Properties`, such as `subst-sym-subst` (which I’m calling `cast-sym-cast` for now). I don’t imagine we’d want to replicate all `subst` properties for `cast`, though.

Thoughts?

JacquesCarette:
To me, the biggest advantage of `cast` over `subst` is that `cast` computes regardless of the actual proof of equality, while `subst` only reduces for `refl`.  Working `--without-K` and with `subst` is **amazingly tedious**.

Side commentary: Using either of them is also misleading; we know that `Fin n` has an automorphism group  of size n!. By singling out the `id` one as being the only one worth talking about when doing type isomorphism on `Fin n` hides a lot of the beauty of `Id` types.

conal:
> To me, the biggest advantage of `cast` over `subst` is that `cast` computes regardless of the actual proof of equality, while `subst` only reduces for `refl`.

Ah, interesting. Thanks for mentioning this difference.

From the definitions of `cast` and `subst`, I gather that reducing `cast` requires evaluating only the `Fin` argument (as you mentioned), while reducing `subst` requires evaluating only the equality argument. Do you have any insight to share about how `cast`’s requirement tends to be less burdensome than `subst`’s?

conal:
For concreteness, here are two of the properties (in their `cast` form) I intend to submit in a PR (where `⊎` is `Data.Sum`). I’ll probably need the analogous properties for products as well.

``` agda
join-assocˡ : ∀ {m n p} →
  join (m + n) p ∘ ⊎.map₁ (join m n) ∘ ⊎.assocˡ
    ≗ cast (sym (+-assoc m n p)) ∘ join m (n + p) ∘ ⊎.map₂ (join n p)

join-assocʳ : ∀ {m n p} →
[...]

Issue #1252, opened 2020-07-06
Naming of unfolding lemmas


gallais:
In `Data.List.Properties` we have:

```agda
unfold-reverse : ∀ (x : A) xs → reverse (x ∷ xs) ≡ reverse xs ∷ʳ x

ʳ++-defn : ∀ (xs : List A) {ys} → xs ʳ++ ys ≡ reverse xs ++ ys
```

Both lemmas are filling the same essential role: explain a complex
function (for efficiency reasons) in terms of more basic building
blocks. I think they should follow the same naming scheme & am in
favour of the `unfold-X` pattern.

MatthewDaggitt:
Okay so I agree `unfold` is nicer than `defn` :+1: 

But I think that definition of `unfold-reverse` shouldn't be called either variant, as it's not really unfolding `reverse`, it's instead talking about the interaction of `reverse` and `∷` and therefore should be called `reverse-∷` according to our existing conventions.

In contrast `ʳ++-defn` really is just unfolding the definition of `ʳ++` and so I'm happy with calling it `unfold-ʳ++`. Depending on how many existing lemmas there are, I might be tempted to argue for switching the order so calling those lemmas `ʳ++-unfold` instead to make it more searchable...

Issue #568, opened 2018-12-21
Generalise fold-fusion results


gallais:
Cf. [this blog post](https://doisinkidney.com/posts/2018-12-21-balancing-scans.html) by @oisdk 

> We’re not using the proofs in Agda’s standard library because these are tied to propositional equality. In other words, instead of using an abstract binary relation, they prove things over actual equality. That’s all well and good, but as you can see above, we don’t need propositional equality: we don’t even need the relation to be an equivalence, we just need transitivity and reflexivity.

jamesmckinna:
Well, we now have some more systematic `≡`-based `fold*-fusion` results in `Data.Vec.Properties` (and soon, some more heterogeneous ones about `reverse` to go... somewhere), so part of that reconsideration might be to generalise these things. 

One day.

Issue #465, opened 2018-09-14
Inconsistent naming convention: id-is-fold vs. idIsFold


gallais:
The first one is in `Data.Nat.GeneralisedArithmetic`, the other in `Data.List.Properties`.

MatthewDaggitt:
Hmm so the second variant is definitely wrong. I'm pretty sure the first is as well. I think they should probably be `id-isFold` as we're not referring to a specific fold function. Or if we think that we need both variants then we should use `id-is-foldr` and `id-is-foldl`...

MatthewDaggitt:
Closed with d77a0ed

PR #1668, opened 2021-12-23
Port reverse lemmas to `Data.Vec` (fixes #942)

### LINKED ISSUE #942: Port Andreas' work on reverseAcc to vectors too
Vectors in the stdlib could benefit from the work done by @andreasabel in #899.
Looks like we have none of the properties of reverse for vectors.

(Some may be hard to prove because of the nat index? So I will not tag this
one as a low-hanging-fruit)

jamesmckinna:
Up to, but not including, those lemmas of Andreas which would require heterogeneous equality. 
And the `CHANGELOG`...


jamesmckinna:
I renamed `idIsFold` to ``idIsFoldr`: suggest possible deprecation warning neded?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+    ≡⟨ P.sym (unfold-reverse (f x) (map f xs)) ⟩
+  reverse (f x ∷ map f xs)
+    ≡⟨⟩
+  reverse (map f (x ∷ xs)) ∎
+    where open P.≡-Reasoning
+
+-- _ʳ++_ and map
+-- map distributes over reverse-append.
+
+map-ʳ++ : ∀ {A : Set a} {B : Set b} {m n} (f : A → B) (xs : Vec A m) {ys : Vec A n} →
```
MatthewDaggitt, line 843: `{A}` and `{B}` are declared as variables so can be removed.

src/Data/Vec/Properties/Heterogeneous.agda
```diff
@@ -0,0 +1,210 @@
...
+    a b c d : Level
+    A : Set a
+    B : Set b
+    C : Set c
+    D : Set d
+
+------------------------------------------------------------------------
+-- Heterogeneous propositional equality over vectors
+
+data ≡ᵥ {a} {A : Set a} : {m : ℕ} (xs : Vec A m) {n : ℕ} (ys : Vec A n) → Set (Level.suc a) where
```
MatthewDaggitt, line 35: So there's a lot of great work in this file, but I think that the crux of this definition already exists in `Data.Vec.Relation.Binary.Pointwise.Inductive`. If you then look at it's instantiation in `Data.Vec.Relation.Binary.Equality.Propositional` to a propositional setoid then, modulo requiring a proof that the heads of the lists are equal, the resulting definition of `_≋_` is identical to this one.

Any chance you could you port the contents to the correct places in those two files?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+
+reverse-foldr : ∀ {B : ℕ → Set b} {n} (f : ∀ {n} → A → B n → B (suc n)) b →
+                foldr B f b ∘ reverse ≗ foldl B {n} (λ {n} → flip (f {n})) b
+reverse-foldr {A = A} {B = B} f b xs =
+  foldl-fusion {B = Vec A} (foldr B f b) [] b refl (λ b x → refl) xs
+
+-- reverse and foldl
+
+reverse-foldl : ∀ {B : ℕ → Set b} {n} (f : ∀ {n} → B n → A → B (suc n)) b →
+                foldl B {n} f b ∘ reverse ≗ foldr B (λ {n} → flip (f {n})) b
```
MatthewDaggitt, line 867: Is there a reason that the equality in this is flipped compared to `reverse-foldr`? They should probably be consistent?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+
+------------------------------------------------------------------------
+-- _∷ʳ_
+
+module _ {x y : A} where
+
+  ∷ʳ-injective : ∀ {n} (xs ys : Vec A n) → xs ∷ʳ x ≡ ys ∷ʳ y → xs ≡ ys × x ≡ y
+  ∷ʳ-injective []          []          refl = (refl , refl)
+  ∷ʳ-injective (x ∷ xs)    (y  ∷ ys)   eq   with ∷-injective eq
+  ... | refl , eq′ = Prod.map (P.cong (x ∷_)) id (∷ʳ-injective xs ys eq′)
```
MatthewDaggitt, line 792: Can use `Prod.map_1` here (sorry for the latex notation!)

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+foldl-∷ʳ B f e y (x ∷ xs) = foldl-∷ʳ (B ∘ suc) f (f e x) y xs
+
+-- A foldr after a reverse is a foldl.
+
+foldr-ʳ++ : ∀ (B : ℕ → Set b) (f : ∀ {n} → A → B n → B (suc n))
+           {m} {n} b (xs : Vec A m) {ys : Vec A n} →
+           foldr B f b (xs ʳ++ ys)
+           ≡
+           foldl (B ∘ (_+ n)) ((λ {m} → flip (f {m + n}))) (foldr B f b ys) xs
+foldr-ʳ++ {A = A} B f {m} {n} b xs {ys} =
```
MatthewDaggitt, line 781: Don't need to match on implicits `A`, `m` or `n` here, and I think `ys` will need to be an explicit argument. I doubt Agda can infer it underneath the `foldr` *and* the `ʳ++`

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+                  (f : ∀ {n} → B n → A → B (suc n)) {e}
+                  (h : ∀ {c} (C : ℕ → Set c) →
+                       (g : ∀ {n} → C n → A → C (suc n)) → C zero →
+                       ∀ {n} → Vec A n → C n) →
+                  (∀ {c} {C} {g : ∀ {n} → C n → A → C (suc n)} e →
+                   h {c} C g e [] ≡ e) →
+                  (∀ {c} {C} {g : ∀ {n} → C n → A → C (suc n)} e →
+                   ∀ {n} x →
+                   (h {c} C g e {suc n}) ∘ (x ∷_) ≗ h (C ∘ suc) (λ {n} → g {suc n}) (g e x)) →
+                  ∀ {n} → h B f e ≗ foldl B {n} f e
```
MatthewDaggitt, line 723: Congrats, I think you've just won the coveted award for "the scariest type added in the library" :ghost:

1. Can we let-bind a type called `let Op T = ∀ {n} → T n → A → T (suc n) in` after the first line? We could then write `(f : Op T)` and `(g : Op C)` etc. The name `Op` is pretty arbitrary, if you can come up with something better go for it!

2. If you're making `f` explicit, then `e` should also be explicit?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+  foldl B f e (x ∷ xs)
+    ∎
+  where open P.≡-Reasoning
+
+foldl-[] : ∀ {A : Set a} (B : ℕ → Set b)
+           (f : ∀ {n} → B n → A → B (suc n)) {e} → foldl B f e [] ≡ e
+foldl-[] B f = refl
+
+foldl-fusion : ∀ {A : Set a}
+               {B : ℕ → Set b} {C : ℕ → Set c}
```
MatthewDaggitt, line 741: I suspect `B` and `C` need to explicit, unless you have evidence to the contrary?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+  map f (reverse xs) ++ map f ys
+    ≡⟨ P.cong (_++ map f ys) (reverse-map-commute f xs) ⟩
+  reverse (map f xs) ++ map f ys
+    ≡⟨ P.sym (unfold-ʳ++ {xs = map f xs} {ys = map f ys}) ⟩
+  map f xs ʳ++ map f ys ∎
+    where open P.≡-Reasoning
+
+-- reverse and foldr
+
+reverse-foldr : ∀ {B : ℕ → Set b} {n} (f : ∀ {n} → A → B n → B (suc n)) b →
```
MatthewDaggitt, line 859: Again, I suspect `B` needs to be explicit?

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+                foldr B f b ∘ reverse ≗ foldl B {n} (λ {n} → flip (f {n})) b
+reverse-foldr {A = A} {B = B} f b xs =
+  foldl-fusion {B = Vec A} (foldr B f b) [] b refl (λ b x → refl) xs
+
+-- reverse and foldl
+
+reverse-foldl : ∀ {B : ℕ → Set b} {n} (f : ∀ {n} → B n → A → B (suc n)) b →
```
MatthewDaggitt, line 866: Again `B` should probably be explicit?

MatthewDaggitt (commented):
Awesome work! Added some thoughts :smile:

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
   foldr-universal C _ _ refl (λ x xs → fuse x (foldr B f e xs))
 
-idIsFold : ∀ {n} → id ≗ foldr (Vec A) {n} _∷_ []
-idIsFold = foldr-universal _ _ id refl (λ _ _ → refl)
+idIsFoldr : ∀ {n} → id ≗ foldr (Vec A) {n} _∷_ []
+idIsFoldr = foldr-universal _ _ id refl (λ _ _ → refl)
+
+foldr-[] : ∀ {A : Set a} (B : ℕ → Set b)
+           (f : ∀ {n} → A → B n → B (suc n)) {e} → foldr B f e [] ≡ e
+foldr-[] B f = refl
```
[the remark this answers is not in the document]
MatthewDaggitt: Eh, I've thought about it some more and it's fine to include. Slightly improves abstraction I guess.

jamesmckinna:
Many thanks for the fine-toothed comb! Agree, there are lots of things which need improvement/reworking esp. wrt implicit/explicit arguments, so I'm grateful for your second look. I'll try to do that slowly over the next few days. 

And yes, there is a reason (those scary types!) why these lemmas were *not* `low-hanging fruit`, and I'm still not entirely satisfied with the current formulation. But I would rather these were merged and then had a chance for others to look/play, before any further drastic reorganisation, eg merging the lemmas in from `WithK`, which I am sure could be ported to the `Heterogeneous` developing *without* having to appeal to K...

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
 foldr-fusion {B = B} {f} e {C} h fuse =
   foldr-universal C _ _ refl (λ x xs → fuse x (foldr B f e xs))
 
-idIsFold : ∀ {n} → id ≗ foldr (Vec A) {n} _∷_ []
-idIsFold = foldr-universal _ _ id refl (λ _ _ → refl)
+idIsFoldr : ∀ {n} → id ≗ foldr (Vec A) {n} _∷_ []
+idIsFoldr = foldr-universal _ _ id refl (λ _ _ → refl)
```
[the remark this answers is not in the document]
jamesmckinna: Pushed in commit `3ff93b6`.

src/Data/Vec/Properties.agda
```diff
@@ -373,6 +373,11 @@ map-const : ∀ {n} (xs : Vec A n) (y : B) → map {n = n} (const y) xs ≡ repl
 map-const [] _ = refl
 map-const (_ ∷ xs) y = P.cong (y ∷_) (map-const xs y)
 
+map-++-commute : ∀ {A : Set a} {B : Set b} {m} {n} (f : A → B) xs ys →
```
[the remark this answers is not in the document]
jamesmckinna: Actually, I think I had encountered some weird type-checking error messages (to do with the levels, IIRC) which I seemed only able to resolve by explicitly adding these here... so if you had time to experiment, I'd welcome your insight.

src/Data/Vec/Properties.agda
```diff
@@ -373,6 +373,11 @@ map-const : ∀ {n} (xs : Vec A n) (y : B) → map {n = n} (const y) xs ≡ repl
...
+                 map f (xs ++ ys) ≡ map {n = m} f xs ++ map {n = n} f ys
```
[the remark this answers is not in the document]
jamesmckinna: Yes, that makes sense. At some point I must have been thinking "what's the minimal annotation needed here to ensure type correctness", given how complex and multi-line the types of these things were already getting...

src/Data/Vec/Properties.agda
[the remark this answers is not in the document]
jamesmckinna: Well, I was/am in two minds. I appeciate Andreas' arguments in favour of/inclination towards 'fat' libraries, and I had in the back of my mind that the corresponding lemmas for `foldl` are precisely what makes reasoning for left-folds so painful. So I was trying to isolate some common abstractions, but the `++` case for `foldl` still eludes me. `foldl-fusion` is very powerful, and painfully mysterious, given that stating and proving the analogue of `foldr-universal` also eluded me.

src/Data/Vec/Properties.agda
jamesmckinna: 1. Seems like a good idea (an Agda issue I never dare open is not being able to have where clauses in types... I guess that's what anonymous modules are for,  but it always seems a weird anomaly to have to use `let` in this one case). It might be that `Op` is a sufficiently pervasive abstraction in this whole module that it deserves a top-level name. I'll ponder that.

MatthewDaggitt: I was also thinking about introducing a top level name for it as well. If we do then we'd have to distinguish between "left" and "right" `Op`s.

src/Data/Vec/Properties/Heterogeneous.agda
jamesmckinna: OK, I suspect that's going to require a bit more time and patience as I work my way round those other definitions. I'll focus on the main (homogeneous) properties first, so you'll want to hang on a bit for this collection of commits ;-)

src/Data/Vec/Properties.agda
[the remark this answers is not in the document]
jamesmckinna: OK, done (and will be on the lookout elsewhere). Meanwhile, similarly, can remove the `{n = n}` annotation to the preceding lemma, too!

src/Data/Vec/Properties.agda
jamesmckinna: Have pushed a commit to `Data.Vec.Base` which adds `Opʳ` and `Opˡ`...

src/Data/Vec/Properties.agda
jamesmckinna: Nice tip! Thanks.

src/Data/Vec/Properties.agda
jamesmckinna: So I think that `reverse-foldr` and `reverse-foldl` are consistent now: both of these lemmas now put the length argument on the left of the equality, and each has the 'flipped' function on the right. I have made considerable reorganisation of the `Properties` file, putting properties of `foldl` *before* those of `foldr`, and making heavy use of anonymous modules to group lemmas together. But there may still be opportunities for improvement!

src/Data/Vec/Properties.agda
jamesmckinna: Agreed. In one use (with `map`), inference worked OK, but in one property in `Properties.Heterogeneous` I needed to supply explicit arguments, so you win this one ;-)

src/Data/Vec/Properties.agda
jamesmckinna: And subsequent commits simplify this type quite a bit (well: a little...) with use of `Opˡ`

src/Data/Vec/Properties.agda
jamesmckinna: It turns out that all uses of `foldl-fusion` seem able to infer both `B` and `C`: in the only non-trivial instance, in `Properties.Heterogeneous`, they get inferred by supplying an explicit type signature for the fused function `h` (and this might well be a useful pattern to bear in mind when invoking `foldl-fusion h`...?)

src/Data/Vec/Properties.agda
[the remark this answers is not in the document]
jamesmckinna: Well... whatever that was, it works without the `A`, `B` bindings now.

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+foldl-[] : ∀ {A : Set a} (B : ℕ → Set b)
+           (f : ∀ {n} → B n → A → B (suc n)) {e} → foldl B f e [] ≡ e
+foldl-[] B f = refl
+
+foldl-fusion : ∀ {A : Set a}
+               {B : ℕ → Set b} {C : ℕ → Set c}
+               (h : ∀ {n} → B n → C n) →
+               {f : ∀ {n} → B n → A → B (suc n)} (d : B zero) →
+               {g : ∀ {n} → C n → A → C (suc n)} (e : C zero) →
+               (h {zero} d ≡ e) →
```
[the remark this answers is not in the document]
jamesmckinna: I think I'd been in two minds about giving explicit types to `d` and `e`. Not sure which choice is better: explicitly type an argument, or supply an implicit specialisation to a function?

src/Data/Vec/Properties/Heterogeneous.agda
jamesmckinna: Hmmm. I realise I may have once again done things out-of-order...
... because you won't be able to close the issue with this PR unless these problems with `Heterogeneous` are fixed first. 
Should I push a commit which removes that file (and you could then merge the 'reduced' PR), and then add it back with a view to resolving this (non-trivial?) issue of reorganisation? Hmmm. Hmmm. Hmmm. 
Food needed!

MatthewDaggitt: Yes, that sounds like a good plan :+1:

src/Data/Vec/Properties.agda
```diff
@@ -659,40 +664,246 @@ zipWith-is-⊛ f (x ∷ xs) (y ∷ ys) = P.cong (_ ∷_) (zipWith-is-⊛ f xs ys
...
+unfold-reverse : ∀ {n} x (xs : Vec A n) → reverse (x ∷ xs) ≡ reverse xs ∷ʳ x
+unfold-reverse x xs = P.sym (foldl-fusion (_∷ʳ x) [] [ x ] refl (λ b x → refl) xs)
+
+-- reverse-append is append of reverse.
+
+unfold-ʳ++ : ∀ {m n} (xs : Vec A m) (ys : Vec A n) → xs ʳ++ ys ≡ reverse xs ++ ys
+unfold-ʳ++ xs ys = P.sym (foldl-fusion (_++ ys) [] ys refl (λ b x → refl) xs)
+
+------------------------------------------------------------------------
+module _ {b} (B : ℕ → Set b) (f : Opˡ {A = A} B) where
```
MatthewDaggitt, line 730: I think `b` is a variable and therefore not needed here?

MatthewDaggitt (commented):
A few more (minor) comments

src/Data/Vec/Properties.agda
```diff
@@ -369,10 +369,15 @@ map-id : ∀ {n} → map {A = A} {n = n} id ≗ id
 map-id []       = refl
 map-id (x ∷ xs) = P.cong (x ∷_) (map-id xs)
 
-map-const : ∀ {n} (xs : Vec A n) (y : B) → map {n = n} (const y) xs ≡ replicate y
+map-const : ∀ {n} (xs : Vec A n) (y : B) → map (const y) xs ≡ replicate y
 map-const [] _ = refl
 map-const (_ ∷ xs) y = P.cong (y ∷_) (map-const xs y)
 
+map-++-commute : ∀ {m} {n} (f : A → B) (xs : Vec A m) (ys : Vec A n) →
```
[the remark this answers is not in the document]
jamesmckinna: Fair enough. As with some of the other style comments above, I think I found myself attempting, not always faithfully, to emulate the naming conventions of `Data.List.Properties`... and I confess I'd prefer short names if possible, but your remark about `reverse-foldr` etc. had me wondering, in such a 'commute'/'complex=>simple' unfolding, which order even the commuting operations should be named. I had also wondered whether (see `Data.Vec.Base` for example) whether reorgansing this file so that everything was stated in terms of `foldr` and `foldl` might be a good idea, or a terrible one...

src/Data/Vec/Base.agda
```diff
@@ -280,14 +286,30 @@ fromList (List._∷_ x xs) = x ∷ fromList xs
...
+-- snoc
 
 infixl 5 _∷ʳ_
 
-_∷ʳ_ : ∀ {n} → Vec A n → A → Vec A (1 + n)
+_∷ʳ_ : ∀ {n} → Vec A n → A → Vec A (suc n)
 []       ∷ʳ y = [ y ]
 (x ∷ xs) ∷ʳ y = x ∷ (xs ∷ʳ y)
+{- better ?
+_∷ʳ_ {A = A} xs y = foldr ((Vec A) ∘ suc) (λ x r → x ∷ r) [ y ] xs
```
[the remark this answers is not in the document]
jamesmckinna: But see my comment above. BTW, `_∷ʳ_` was already defined; it was the `foldl`-based definition of `_ʳ++_`  which had me reconsidering those functions defined as `foldr`s... towards some grand (folly of a) general `foldl`/`foldr` commutation result.

src/Data/Vec/Properties.agda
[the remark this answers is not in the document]
jamesmckinna: Hmm. Originally, there was also a `sum-++-commute`... another deprecation opportunity?

src/Data/Vec/Properties/Heterogeneous.agda
jamesmckinna: OK, will do that now.

src/Data/Vec/Properties.agda
jamesmckinna: And they have each had their names flipped now: `foldr-reverse`, `foldl-reverse`.

src/Data/Vec/Properties.agda
jamesmckinna: And indeed, arguments `d` and `e` to this function can also be made implicit.

src/Data/Vec/Properties.agda
jamesmckinna: Hmmm. Apologies for the emoji, but I disagree on this last point, as follows:
* `h` is first declared with its arguments (all save `{c}`) *fully decorated*; 
* so thereafter, one should only declare the *minimum necessary* for any `h`-instance to be well-typed.
* NB interestingly, the `Op` arguments do need to be explicitly typed... don't see how to fix that!
Now, your comment indeed would apply, if one wants a more verbose type signature, but your original comment... that it's bad/scary enough already, applies too. 
Not a hill on which I wish to take a stand, but... this meta-principle of 'minimality where possible' (as an instance of DRY) seems at least debatable? I'm not sure what the library style-guide says about such things; but I guess it's a fingerprint for my style of DTFP: announce your intentions *once*, then let the typechecker work hard(er) for you...

src/Data/Vec/Properties/Heterogeneous.agda
jamesmckinna: So... is it the case that the relation defined in `Data.Vec.Relation.Binary.Pointwise.Inductive` is heterogeneous wrt the lengths of the `Vec`s concerned? I perhaps hadn't previously appreciated that (and in any case, while developing new stuff, I still find it easier to start from a fresh context... not perhaps the best way to do library development, though ;-))

src/Data/Vec/Base.agda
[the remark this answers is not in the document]
jamesmckinna: So here's a strange thing: I was doing some more reorganising, to try to see if `unfold-reverse` and `unfold-ʳ++` could be rewritten in terms of `foldr-reverse` (you'd think, right? at least in view of the above ;-)) and find myself in a situation where agda  seems reluctant to unfold the recursion equations for `foldl` correctly, even with every argument fully written out... starting to wonder if there is a bug?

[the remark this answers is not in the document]
jamesmckinna: Or then again, it perhaps would *require* `_∷ʳ_` to be written as a `foldr` for that idea to work... instead of being, clause for clause, identical to a `foldr` instance... hmmm.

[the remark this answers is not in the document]
jamesmckinna: Above conjecture correct. So... functions defined by pattern-matching are not intensionally equal to (suitable instances of slightly more higher-order) functions with the same pattern-matching behaviour. Hmmm. Generativity bites again? Oh well.

jamesmckinna:
If you are able to successfully merge these changes, I will then return to the material in `Properties.Heterogeneous`... in the New year perhaps!

jamesmckinna:
Hmmm... you'll be wanting me to rectify the `CHANGELOG` to reflect the new names of everything ;-) Later...

src/Data/Vec/Properties.agda
```diff
@@ -691,8 +696,209 @@ foldr-fusion : ∀ {A : Set a}
...
+
+------------------------------------------------------------------------
+-- _∷ʳ_, _ʳ++_ and reverse
+
+-- reverse of cons is snoc of reverse.
+
+unfold-reverse : ∀ {n} (x : A) xs → reverse (x ∷ xs) ≡ reverse {n = n} xs ∷ʳ x
+unfold-reverse x xs =  P.sym (foldl-fusion (_∷ʳ x) [] [ x ] refl (λ b x → refl) xs)
+
+unfold-ʳ++ : ∀ {m n} {xs : Vec A m} {ys : Vec A n} → xs ʳ++ ys ≡ reverse xs ++ ys
```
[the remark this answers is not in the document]
jamesmckinna: Actually, I think it is sufficient to make only `xs` explicit (?).

src/Data/Vec/Properties.agda
jamesmckinna: Oh, and 2. above? It seems that Agda can infer `e` just fine in any goal mentioning `fold* B f e`, AFAICT.

jamesmckinna:
When I get around to the former `Heterogeneous` proofs... there is also Sergei Goncharov's PR #1621 which was successfully merged. But, AFAICT, `Pointwise` extensional equality is weaker than `Inductive` equality. Should/can the two proofs (of 'associativity' of `++`) coexist?

jamesmckinna:
Hmmm. @MatthewDaggitt Issue #465, but now for vectors? I'll rename `idIsFoldr` shall I?
Cf. Issue #602 ...

src/Data/Vec/Base.agda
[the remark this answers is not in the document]
jamesmckinna: And... `List.Properties` again leads the way with a bunch of `*-is-foldr` lemmas. I'll leave those to another day perhaps

src/Data/Vec/Properties/Heterogeneous.agda
jamesmckinna: OK, I think I will need a bit more hand-holding about the library structure here before embarking on the `Heterogeneous` commits... but hopefully you can merge this PR anyway ahead of doing that, while still leaving issue #942 open.

JacquesCarette:
(For when you get back to `Heterogeneous`: `Data.Fin.cast` can be your friend, because it computers. `subst` is evil and not a good way to do anything.)

jamesmckinna:
OK, thanks @JacquesCarette for the tip... but you might need to make the hint more explicit. As it is, I end up using `subst` only in two/three places: once/twice to show the equivalence of `xs [ m ]≡ᵥ[ n ]  ys` with the `transport` definition on the Sigma-type, and only once 'in anger', to resolve the very painful 'unfolding reverse-append on cons' lemma

```agda
ʳ++-∷ : ∀ {m n} y (xs : Vec A m) {ys : Vec A n} →
        (xs ʳ++ (y ∷ ys)) [ m + suc n ]≡ᵥ[ suc m + n ] ((y ∷ xs) ʳ++ ys)
```

I take, reading between the lines, @MatthewDaggitt 's hint that much of what was in `Heterogeneous` a) belongs elsewhere (but nailing that down precisely is making me scratch my head a bit), and b) only a fraction is actually new (to do with `_ʳ++_`...), or c) even requires explicit index arithmetic (the above result may be the only one; all the rest seem to come 'for free'). That leaves the equational reasoning syntax, and I'm not quite sure where that belongs, either (or if it clashes with/is made redundant by existing usage)

So if you can see a way to prove the above lemma without appealing to `subst`... I'd welcome a nudge! 
Happy New Year! (This won't get fixed before then ;-))

JacquesCarette:
`cast` is basically a `subst` that computes -- see https://agda.zulipchat.com/#narrow/stream/238741-general/topic/subst.20puzzle for the originating discussion, suitable for when you're fighting with weird `Fin` lemmas that also struggle with `Nat` equalities.

For your `≡ᵥ`, have you considered parametrizing the equality not by 2 (explicit) numbers, but instead by 2 implicit numbers and 1 explicit proof that they are equal?  That's what I ended up doing when formalizing multicategories, and it worked rather nicely.

src/Data/Vec/Properties/Heterogeneous.agda
MatthewDaggitt: Sounds like a plan!

src/Data/Vec/Properties.agda
MatthewDaggitt: Thanks, definitely the right way to go!

src/Data/Vec/Properties.agda
MatthewDaggitt: Okay, fair enough, I'm convinced. Thinking about it more, I think the real problem is the lack of support for easily finding out the type of `C`, because there's no Agda editor that provides support for displaying the type of a term when you mouse over it...

src/Data/Vec/Properties.agda
MatthewDaggitt: Thanks for investigating!

src/Data/Vec/Properties.agda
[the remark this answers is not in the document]
MatthewDaggitt: Yes, in a perfect world we would be consistent about the ordering as well. I think lets hold off any reorganisation in terms of `foldr` and `foldl` for at least in this PR!

MatthewDaggitt:
Thanks for all the changes @jamesmckinna. I've gone through and had a very minor tidy-up. The things I did do were:
- rename `foldr₀` and `foldl₀` as per the conclusions in https://github.com/agda/agda-stdlib/issues/278 and generalise their types so that the output type can differ from the input type
- renamed `unfold-reverse` to `reverse-::`

Once the tests go green, I'll merge it in!

jamesmckinna:
@MatthewDaggitt Happy New Year! with many thanks for all the shepherding through last year of my various PRs, not least this particular beast. Here's to many more such... ;-)

jamesmckinna:
Oh... regarding `unfold-reverse` (and its friend `unfold-ʳ++`): I agree these two items now stick out a bit, but I had been trying to follow @gallais's precept in issue #1252 but perhaps that has not enjoyed widespread uptake? But your choice of name is also a good one... so what about its friend?

jamesmckinna:
@JacquesCarette thanks for the suggestion regarding `_[_]≡ᵥ[_]_` which I had not previously considered. 

I think that my only (defensible?) reasons for introducing the new notation were:
* to experiment with introducing new syntax;
* and hence, to be able to have top-level lemma *statements* (moreover wanting only *syntax* for such statements) which draw attention to these lemmas being 'special', because they involve 'tricky' index arithmetic;
* to see what was/might be needed for heterogeneous reasoning combinators; 
* to see whether explicitly mentioning `Vec A` indices helped make the proofs clearer: this it didn't seem to do, or at least was not *necessary*, in view of the previous point.

All in all, the eventual reworking of `Heterogeneous` might see all of this vanish in a puff of blue smoke... although @MatthewDaggitt 's point about seeing these lemmas as instances of `Relation.Binary.Pointwise.Inductive` reasoning might see a role of the syntax, at least... and then your observation becomes 'too interesting' for the definitions already there. 

Much food for thought, however, for which my thanks. But the mention of issue #568 above indicates that perhaps even more thought is required...

MatthewDaggitt:
> @MatthewDaggitt Happy New Year! with many thanks for all the shepherding through last year of my various PRs, not least this particular beast. Here's to many more such... ;-)

And to you too :tada: 

> Oh... regarding unfold-reverse (and its friend unfold-ʳ++): I agree these two items now stick out a bit, but I had been trying to follow @gallais's precept in issue #1252 but perhaps that has not enjoyed widespread uptake? But your choice of name is also a good one... so what about its friend?

Will reply in the issue #1252 but long story short, I think I'm happy the names as they are now!

MatthewDaggitt:
Let's get this landed!

JacquesCarette:
@jamesmckinna I'm all for exploration. Especially around where people currently pull out `Heterogeneous` equality (and, worse, `subst`) when it's not needed. I just happened to have explored this a bit myself, so I reported on what I'd learned.

Your instincts perfectly matched my previous thinking (on `Permutation`) where making the indices explicit (and having 2 of them, even though they are always provably equal) was key to making everything go smoothly. And failed me when I tried the same thing with multicategories.

I think the main lesson (for me) is/was that none of the constructions around permutations actually care that the two indices are the same, but they do preserve that 'sameness'. Whereas in other settings, why they are the same matters. There are, after all, `n!` proofs that `n = n`.  Permutations don't care because they are exactly a record of how to shuffle things around.

In other cases, such as vector operations, 'index shuffling' matters - so recording which permutation you've used may indeed matter. At least, that's my current thinking. [It's also an insight that was there in my work with Brent Yorgey in Species a long time ago... but we got referees who didn't like Species, so it never got published. One day, people will see that there's more to the world than just polynomials...]

jamesmckinna:
Thanks @JacquesCarette for the reinforcement of the lessons imparted in issue #1610... on which I (still) need to meditate. As indeed on Species... having failed to learn the lessons of Brent's "Oh My!" Pearl a few years back. 

Interesting to learn, albeit second-hand, the contrasting experiences of `Permutation` and those for multicategories, though your account makes sense (and might the situation be any 'better'/'worse', in the sense of drawing attention to the *necessary* structure, in the operadic case?). 

Concretely, I still have two open questions about the `Heterogeneous` experiment: 
* where should the results actually best go (as instances of `Pointwise _≡_` reasoning, or more generally, for arbitrary `Pointwise _∼_`; 
* is there (still) any value in retaining the `syntax` extensions? 

Happy to take this discussion offline!

MatthewDaggitt:
Sorry I've taken so long to reply to this. 

> - where should the results actually best go (as instances of Pointwise _≡_ reasoning, or more generally, for arbitrary >> (reflexive!) Pointwise _∼_;

So everything should go to the most general place it can live. If it holds for arbitrary binary relations then it should go in the `Pointwise` module. If it only holds for equality then it should live in the `Equality` modules. You can add specialisations of the `Pointwise` results in the `Equality` modules if you like.

> - is there (still) any value in retaining the syntax extensions?

Err I have to confess you can probably answer that one better than I can right now, as I haven't looked closely at these modules for a while. I suspect that you probably don't need it as much.

jamesmckinna:
No worries, it's a slightly stake thread for me atm, so this is a 'keepalive' as much to myself as anything: OK re 'most general' policy, but I haven't had the time to investigate how general the results are; and... duly noted regarding the syntax extensions, although are the 'usual' `≡-Reasoning` combinators available for arbitrary refl, sym, trans, relations, or only for `_≡_`? (I suppose that in part depends on answers such as 'reflexive wrt what underlying setoid 'equality'', as well as whether they make sense for heterogeneous relations... it seems as though they should; nothing special about `Vec` here, I think?)

MatthewDaggitt:
Yes, reasoning combinators are available over many arbitrary binary relations, see the `Relation.Binary.Reasoning` subfolder. However there also reasoning over heterogeneous equality [here](https://github.com/agda/agda-stdlib/blob/4c5be8622893017403c3ef251ae47ad2610574ea/src/Relation/Binary/HeterogeneousEquality.agda#L226), but as you point out, it's only over a very specific propositional equality. I guess it should be possible to generalise your reasoning combinators over an indexed setoid from `Relation.Binary.Indexed`?

PR #1669, opened 2021-12-24
Private variables in Nat directory

guilhermehas:
I put private variables in Nat directory to make it more concise.

### DIFF
--- a/src/Data/Nat/Base.agda
+++ b/src/Data/Nat/Base.agda
@@ -38,6 +38,10 @@ open import Agda.Builtin.Nat public
 open import Agda.Builtin.Nat public
   using () renaming (_<_ to _<ᵇ_)
 
+private
+  variable
+    n : ℕ
+
 infix 4 _≤ᵇ_
 _≤ᵇ_ : (m n : ℕ) → Bool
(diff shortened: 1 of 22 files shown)


src/Data/Nat/Base.agda
```diff
@@ -38,6 +38,10 @@ open import Agda.Builtin.Nat public
 open import Agda.Builtin.Nat public
   using () renaming (_<_ to _<ᵇ_)
 
+private
+  variable
+    n : ℕ
```
MatthewDaggitt, line 43: Sorry to be difficult but I think this one is a) against the style-guide (types should only be variablised if they are *used in* the definition of the main type of the file and `ℕ` isn't used, it *is* the main definition) and b) probably overkill for the number of uses, especially as 2 out of the 5 uses are going to be removed when I get round to fixing https://github.com/agda/agda-stdlib/issues/1670.

MatthewDaggitt (commented):
Other than that, the others look good to me! Thanks!

guilhermehas:
I removed my changes from this file.

MatthewDaggitt:
Great, thanks!

PR #1673, opened 2022-01-01
Modernise `Data.Vec(.Properties)`

MatthewDaggitt:
While working on #1668 with @jamesmckinna I noticed that some of the code in `Data.Vec(.Properties)` is pretty ancient by the library standards. I've tried to modernise the code by:
- introducing variables for sizes/elements/vectors as now allowed by the style guide
- importing `Relation.Binary.PropositionalEquality` unqualified
- using new combinators in propositional reasoning
- reorganising the file slightly to match the layout of `Base`

Should be a very nearly a no-op, modulo the movement of a couple of implicit arguments due to variables, a breakage which is acceptable in v2.0 I think.

### DIFF
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -10,7 +10,6 @@ module Data.Vec.Properties where
 
 open import Algebra.Definitions
 open import Data.Bool.Base using (true; false)
-open import Data.Empty using (⊥-elim)
 open import Data.Fin.Base as Fin using (Fin; zero; suc; toℕ; fromℕ; _↑ˡ_; _↑ʳ_)
 open import Data.List.Base as List using (List)
 open import Data.Nat.Base
@@ -31,6 +30,7 @@ open import Relation.Unary using (Pred; Decidable)
 open import Relation.Nullary using (Dec; does; yes; no)
 open import Relation.Nullary.Decidable using (map′)
 open import Relation.Nullary.Product using (_×-dec_)
+open import Relation.Nullary.Negation using (contradiction)
 
 private
   variable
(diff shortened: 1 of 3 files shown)


src/Data/Vec/Properties.agda
```diff
@@ -460,7 +460,7 @@ module _ {f : A → A → A} where
   zipWith-assoc assoc (x ∷ xs) (y ∷ ys) (z ∷ zs) =
     cong₂ _∷_ (assoc x y z) (zipWith-assoc assoc xs ys zs)
 
-  zipWith-idem : ∀ {f : A → A → A} → Idempotent _≡_ f →
+  zipWith-idem : Idempotent _≡_ f →
                  Idempotent {A = Vec A n} _≡_ (zipWith f)
```
Taneb, line 464: Would this be simpler as `Idempotent _≡_ (zipWith {n = n} f)`?

MatthewDaggitt: Yes :+1:

jamesmckinna:
What is the corresponding situation ('modernised', or otherwise) for `Data.List`? 
Does this latter (set of) module(s) get much more attention than those for `Data.Vec`?

src/Data/Vec/Properties.agda
```diff
@@ -281,388 +270,383 @@ updateAt-cong i f≗g xs = updateAt-cong-relative i xs (f≗g (lookup xs i))
...
   zipWith-absorbs abs (x ∷ xs) (y ∷ ys) =
-    P.cong₂ _∷_ (abs x y) (zipWith-absorbs abs xs ys)
+    cong₂ _∷_ (abs x y) (zipWith-absorbs abs xs ys)
 
 module _ {f : A → A → B} where
 
-  zipWith-comm : (∀ x y → f x y ≡ f y x) → ∀ {n}
-                 (xs ys : Vec A n) → zipWith f xs ys ≡ zipWith f ys xs
+  zipWith-comm : ∀ (comm : ∀ x y → f x y ≡ f y x) (xs ys : Vec A n) →
+                 zipWith f xs ys ≡ zipWith f ys xs
```
gallais, line 532: It's kind of strange to be insisting on `f` being commutative as opposed to
the more general constraint `∀ x y → f x y ≡ g y x`.

MatthewDaggitt: Generalised. At the time I was trying to lift algebraic structures over vectors, so I was thinking in more general terms.

MatthewDaggitt:
> What is the corresponding situation ('modernised', or otherwise) for Data.List?
Is it ('simply'?) the case that this latter (set of) module(s) get(s) much more attention than that/those for Data.Vec?

Better, as they do get more attention. You could definitely go through `Data.List.Properties` and add variables for `x y xs ys` though.

src/Data/Vec/Properties.agda
gallais: Makes sense. I guess now that we're not insisting on commutativity, we can also relax the type and have:
```agda
f : A → B → C
g : B → A → C
```

MatthewDaggitt: Generalised further!

PR #1674, opened 2022-01-02
Improve definition of `Prime` and add Euclid's lemma

Taneb:
This is something that'll be very useful for number theory proofs involving prime numbers.

There's two reasons I'm not quite happy with this addition:

* The proof is a little messy in parts
* The proof depends on Bézout's lemma, and `Data.Nat.GCD` is a heavy dependency to add to the `Data.Nat.Primality` module.

I'd like to discuss these a bit to see if they can be resolved or mitigated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,6 +285,27 @@ Non-backwards compatible changes
   ∣↥p∣≢0⇒p≄0 : ∀ p → ℤ.∣ (↥ p) ∣ ≢0 → p ≠ 0ℚᵘ
   ```
 
+### Change in the definition of `Prime`
+
+* The definition of `Prime` in `Data.Nat.Primality` was:
+  ```agda
+  Prime 0             = ⊥
+  Prime 1             = ⊥
+  Prime (suc (suc n)) = (i : Fin n) → 2 + toℕ i ∤ 2 + n
+  ```
+  which was very hard to reason about as not only did it involve conversion
+  to and from the `Fin` type, it also required that the divisor was of the form
+  `2 + toℕ i`, which has exactly the same problem as the `suc n` hack described
+  above used for non-zeroness.
+  
+* To make it easier to use, reason about and read, the definition has been
+  changed to:
+  ```agda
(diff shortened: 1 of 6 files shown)


JacquesCarette:
You could have `Data.Nat.Primality.Base` with the basic stuff, then `Data.Nat.Primality.Properties` for the theorems, and then public import them both in `Data.Nat.Primality`. That way upstream users (in the library) can be more precise about what they want, if they so wish.

MatthewDaggitt:
Okay I'm getting too sucked into this. The messiness is caused by two problems:
- once again we have `Fin`s pervading our definitions of `Prime` and `Composite`. As with divisibility and modulus, they make working with these proofs incredibly difficult. 
- once again we are representing inequalities in our definitions using additive constraints (i.e. `2 +...`) instead of simply saying `2 <= ...`, which makes them very difficult to use if your arguments are not in that form.

Long story short, we need to change the definitions of `Prime`.

MatthewDaggitt:
Okay, I've made changed the definition of `Prime` and `Composite`, which I hope everyone agrees is a) more readable and b) greatly simplifies the proofs. Thoughts @Taneb/ @gallais anyone else?

MatthewDaggitt:
> You could have Data.Nat.Primality.Base with the basic stuff, then Data.Nat.Primality.Properties for the theorems, and then public import them both in Data.Nat.Primality. That way upstream users (in the library) can be more precise about what they want, if they so wish.

As for this, @Taneb I'll leave it up to you as to whether you do it.

Taneb:
> > You could have Data.Nat.Primality.Base with the basic stuff, then Data.Nat.Primality.Properties for the theorems, and then public import them both in Data.Nat.Primality. That way upstream users (in the library) can be more precise about what they want, if they so wish.
> 
> As for this, @Taneb I'll leave it up to you as to whether you do it.

I think for now I'll leave it where it is.

MatthewDaggitt:
Look at my and @Taneb's collective shortening of the proof `Prime` implies `Coprime` by 15 lines :tada:

PR #1677, opened 2022-01-07
Add a difference operator for predicates

stefanholdermans:
Adds `_∖_  : Pred A ℓ₁ → Pred A ℓ₂ → Pred A _` in
`Relation.Unary`. The operator is implemented directly as a function on
elements rather than in terms of `∁` and `_∩_` in order to

*  be consistent with how the other operators on predicates in
   `Relation.Unary` are implemented; and
*  express the meaning of the operator as clearly as possible.


MatthewDaggitt:
Amazed we don't have that! Thanks for the contribution :smile:

Issue #1676, opened 2022-01-05
Generalize Algebra.Definitions to use Opᵣ and Opₗ where applicable.


phadej:
e.g.

```agda
LeftZero : A → Op₂ A → Set _
LeftZero z _∙_ = ∀ x → (z ∙ x) ≈ z
```

could work with `Opᵣ` etc.

Was this considered?

(Edit; `Opᵣ` definition looks weird, but I probably will get used to it).

MatthewDaggitt:
It's a good question. The same thing can be done with `Commutative` right, but with an even more general `A -> B -> C`! But would such a definition even really belong in an `Algebra` module? 

In general I think the design decision here was not to over-generalise, otherwise we lose locality in the library. Defining it again in the context of `Algebra.Module.Definitions.Right` is not such a hardship, and at least it makes the intention clear to the reader of the code that they're in module land.

phadej:
My motivation is indeed formalizing module like structures. scalar * vector multiplication has plenty of properties, zeros, identities, distributivity, cancellativity. OTOH, i also define properties for vector + vector addition in the same Agda module, which also has zeros and identities. Different `LeftZero` predicates will name clash, `LeftZero` and  `LeftZeroᵣ` would be confusing.

One option is to have `Algebra.Definitions.Generalized` (or some other name) with more general definitions, and specializing them in `Algebra.Definitions`.

That would allow  

```agda
open import Algebra.Definitions {A = MyType} _≡_
```

to still work. (and indeed I use that for some structures too).

---

commutative needs to be `A -> A -> B`,  but indeed it is good point

Taneb:
@phadej are you aware of the `Algebra.Module` hierarchy in stdlib, in particular `Algebra.Module.Definitions.{Left,Right}`? There exists versions of `LeftZero` and several others generalized like you suggest, for use in defining modules and similar algebraic structures

phadej:
@Taneb i'm not, that seems to not be reachable from `README` module (it is through Everything, but ...) i'll try that.

MatthewDaggitt:
Hmm why are `Opᵣ` and `Opl` in `Algebra.Core` and not in `Algbra.Module.Core`?

MatthewDaggitt:
@laMudri ?

laMudri:
I don't remember there being a particular reason, other than that they could be and `Algebra.Module.Core` doesn't exist yet. Given that they're intended for module-like structures, I'd support a move.

PR #1680, opened 2022-01-09
Move definitions from Algebra.Core to Algebra.Module.Core

Taneb:
Based on discussion in #1676.

I don't know how to represent this in the changelog.


MatthewDaggitt:
I'd add a new entry to the new module list, and then make a quick note in the `Other` subsection of the `Breaking changes` section.

PR #1684, opened 2022-01-11
Remove duplication in Ring-like structures

### LINKED ISSUE #1617: Redundancy in ring-like structures
Consider, from `Algebra.Structures`

```agda
record IsNearSemiring (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
  field
    +-isMonoid    : IsMonoid + 0#
    *-isSemigroup : IsSemigroup *
    distribʳ      : * DistributesOverʳ +
    zeroˡ         : LeftZero 0# *
```

`+-isMonoid` and `*-isSemigroup` both contain a field `isEquivalence : IsEquivalence _≈_`. There is nothing forcing these to be equal (which might matter if `_≈_` is proof-relevant). Also, it makes it a pain to implement a ring-like structure "top-down" (writing out nested records explicitly).

Taneb:
Closes #1617 

Currently various ring-like structures carry two different proofs that their underlying equivalence relation is an equivalence relation, because they're built out of a group for + and a monoid for * or similar. This changes the definitions so that they're only defined in terms of the + structure, with the * properties as extra laws. This makes their structure more linear and prevents the two isEquivalence proofs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -322,6 +322,22 @@ Non-backwards compatible changes
 
 * The contents of `Function.Strict` is now re-exported by `Function`.
 
+### Changes to ring structures
+
+* Several ring-like structures now have the multiplicative structure defined by
+  its laws rather than as a substructure, to avoid repeated proofs that the
+  underlying relation is an equivalence. These are:
+  * `IsNearSemiring`
+  * `IsSemiringWithoutOne`
+  * `IsSemiringWithoutAnnihilatingZero`
+  * `IsRing`
+* To aid with migration, structures matching the old style ones have been added
+  to `Algebra.Structures.Biased`, with conversionFunctions:
+  * `IsNearSemiring*` and `isNearSemiring*`
+  * `IsSemiringWithoutOne*` and `isSemiringWithoutOne*`
+  * `IsSemiringWithoutAnnihilatingZero*` and `isSemiringWithoutAnnihilatingZero*`
+  * `IsRing*` and `isRing*`
+
 ### Other
 
 * The first two arguments of `m≡n⇒m-n≡0` (now `i≡j⇒i-j≡0`) in `Data.Integer.Base`
--- a/src/Algebra/Construct/DirectProduct.agda
+++ b/src/Algebra/Construct/DirectProduct.agda
@@ -211,7 +211,9 @@ ring R S = record
   { -_     = uncurry (λ x y → R.-_ x , S.-_ y)
   ; isRing = record
       { +-isAbelianGroup = AbelianGroup.isAbelianGroup A
-      ; *-isMonoid       = Semiring.*-isMonoid Semi
+      ; *-cong           = Semiring.*-cong Semi
+      ; *-assoc          = Semiring.*-assoc Semi
+      ; *-identity       = Semiring.*-identity Semi
       ; distrib          = Semiring.distrib Semi
       ; zero             = Semiring.zero Semi
       }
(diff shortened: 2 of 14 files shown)


JacquesCarette:
This is one solution route. 

Another is to notice that the current `Setoid` structure is internalized (in `isMonoid` etc). If we had a parallel `isMonoid` hierarchy that was parametrized over the carrier `Setoid`, then for something like `NearSemiring`, the `+` structure could be the one over the internalized `Setoid`, while the `*` structure over the parametrized version.

I've seen similar duplication of structures in the 1Lab, and it seemed to work nicely. I've done something like it in `agda-categories` to let me smoothly define "identity on objects" functors.

MatthewDaggitt:
I have to confess I'm a tiny bit reluctant about this, as its a massive breaking change. It's not that I don't feel it's not an improvement, but I guess that @JacquesCarette's has got to me about having no good principled solution to the whole record mess.

Also you've missed out fixing:
- `IsSemiringWithoutAnnihilatingZero`
- `IsSemiringWithoutOne`
- `IsNearSemiring`

Having said that I don't really see an alternative solution. I see the appeal of your solution @JacquesCarette, but in my opinion I think we should probably avoid committing ourselves to maintaining a third parallel hierarchy of records. We have enough problems keeping 2 of them straight as it is.

Taneb:
> I have to confess I'm a tiny bit reluctant about this, as its a massive breaking change. It's not that I don't feel it's not an improvement, but I guess that @JacquesCarette's has got to me about having no good principled solution to the whole record mess.

When changing record fields needs to be done, I don't think Agda has a way of fixing things _without_ there being big breaking changes. That said, there's a pretty easy migration path here, and the changes in stdlib were a lot less than I was expecting.

> Also you've missed out fixing:
> 
>     * `IsSemiringWithoutAnnihilatingZero`
> 
>     * `IsSemiringWithoutOne`
> 
>     * `IsNearSemiring`

Are you sure you're not just looking at the last commit?

> Having said that I don't really see an alternative solution. I see the appeal of your solution @JacquesCarette, but in my opinion I think we should probably avoid committing ourselves to maintaining a third parallel hierarchy of records. We have enough problems keeping 2 of them straight as it is.

I'm pretty against adding a third record hierarchy here. It's a lot more effort and as far as I can tell doesn't have much benefit for either the definers of these structures, the implementors of specific rings, or people trying to use them.

There's a variation on what I've done that I flip-flopped between for a while, which is to build some of the structures out of smaller ring-like structures rather than group-like structures.

JacquesCarette:
I actually agree that adding a full-fledged third record hierarchy has a very poor weight-to-benefit ratio. But I wanted to make sure to explore the design space more thoroughly.

Fundamentally I think a good solution will need non-trivial changes to Agda itself (along the lines of what Arend does). So not until stdlib 3.0, I'm afraid.

MatthewDaggitt:
> When changing record fields needs to be done, I don't think Agda has a way of fixing things without there being big breaking changes. 

I'd second that. I guess it's a feature not a bug of theorem provers that implementation matters.

> Are you sure you're not just looking at the last commit?

Ah yes, I am!

> There's a variation on what I've done that I flip-flopped between for a while, which is to build some of the structures out of smaller ring-like structures rather than group-like structures.

Okay, what we can do to ease the pain for people is to add old versions of the structures to `Algebra.Structures.Biased`, along with conversion functions, just like we did with the left and right versions. That way they only have to add in a function application, rather than fully rewrite their records.

PR #1689, opened 2022-01-14
Injective functions on fin

EgbertRijke:
Changes in `Function.Consequences`
* Injective functions map unequal elements to unequal elements

Changes in `Data.Fin.Properties`
* Definition `0≢1+n` for `0` is not a successor.
* Definition `pinch-injective` proving that `pinch` is almost injective.
* Definition `injective⇒≤` proving that if `f : Fin k → Fin l` is injective, then `k ≤ l`.
* Definition `<⇒notInjective` proving that `l < k` implies that no function `f : Fin k → Fin l` is injective.
* Definition `ℕ→Fin-notInjective` proving that no function `f : ℕ → Fin k` is injective
* Definition `cantor-schroder-bernstein` proving that if `f : Fin k → Fin l` and `g : Fin l → Fin k` are both injective, then `k = l`.


src/Data/Fin/Properties.agda
```diff
@@ -761,6 +767,24 @@ pinch-mono-≤ 0F      {suc j} {suc k} (s≤s j≤k) = j≤k
 pinch-mono-≤ (suc i) {0F}    {k}     0≤n       = z≤n
 pinch-mono-≤ (suc i) {suc j} {suc k} (s≤s j≤k) = s≤s (pinch-mono-≤ i j≤k)
 
+pinch-injective :
+  {k : ℕ} (x : Fin k) {y z : Fin (ℕ.suc k)} →
```
JacquesCarette, line 771: because of how Agda add implicits, it's usually a bad idea to alternate explicits and implicits, especially when they are independent like this.  Best to put `x` last.

src/Data/Fin/Properties.agda
```diff
@@ -761,6 +767,24 @@ pinch-mono-≤ 0F      {suc j} {suc k} (s≤s j≤k) = j≤k
...
+  suc x ≢ y → suc x ≢ z → pinch x y ≡ pinch x z → y ≡ z
+pinch-injective _ {zero} {zero} f g p = refl
+pinch-injective zero {zero} {suc z} f g p =
+  ⊥-elim (g (cong suc p))
+pinch-injective zero {suc y} {zero} f g p =
+  ⊥-elim (f (cong suc (sym p)))
+pinch-injective zero {suc y} {suc z} f g p =
+  cong suc p
+pinch-injective (suc x) {suc y} {suc z} f g p =
+  cong
```
JacquesCarette, line 781: `cong suc` on one line, and without the extra () would be more usual.

JacquesCarette (changes requested):

src/Data/Fin/Properties.agda
EgbertRijke: I realize that `x` can also be implicit here. Would it be ok if I make it implicit, or do you prefer to keep it explicit and making it the last one?

JacquesCarette: If they are all implicit, they can stay in the same order as they are in now.

Side comment: I'm always uncomfortable with proofs by induction on implicits. It always makes me feel like I'm doing something wrong, like maybe there's some intermediate lemmas I'm missing or something. I'm not asking you to do anything with this, just something to ponder.

EgbertRijke: There are some cases where agda can't infer the implicit argument if a construction was made by induction on that implicit argument.

JacquesCarette: Right - I've evolved my code to never do constructions by recursion on implicits, that's just asking for trouble. Proofs are another matter.

EgbertRijke:
Using `<⇒notInjective` and `¬∀⟶∃¬` we can conclude `pigeonhole`. This would simplify the construction in `pigeonhole`. Would that change be welcome?

src/Function/Consequences.agda
```diff
@@ -44,3 +46,12 @@ module _ (From : Setoid a ℓ₁) {≈₂ : Rel B ℓ₂} where
   inverseᵇ⇒bijective : ∀ {f f⁻¹} → Congruent ≈₂ ≈₁ f⁻¹ → Inverseᵇ ≈₁ ≈₂ f f⁻¹ → Bijective ≈₁ ≈₂ f
   inverseᵇ⇒bijective cong₂ (invˡ , invʳ) =
     (inverseʳ⇒injective cong₂ invʳ , inverseˡ⇒surjective ≈₁ ≈₂ invˡ)
+
+module _
+  {A : Set a} {B : Set b} {f : A → B}
```
MatthewDaggitt, line 51: `{A : Set a} {B : Set b}` are variable so can be removed here (then the other module arguments can live on a single line).

src/Function/Consequences.agda
```diff
@@ -44,3 +46,12 @@ module _ (From : Setoid a ℓ₁) {≈₂ : Rel B ℓ₂} where
...
+  (_≈₁_ : Rel A ℓ₁) (_≈₂_ : Rel B ℓ₂)
+  where
+
+  contraInjective : Injective _≈₁_ _≈₂_ f →
+    {x y : A} → ¬ (x ≈₁ y) → ¬ (f x ≈₂ f y)
```
MatthewDaggitt, line 56: Again can use a `∀` here and omit the type annotation on `x` and `y`.

src/Data/Fin/Properties.agda
```diff
@@ -849,6 +872,27 @@ pigeonhole (s≤s (s≤s m≤n)) f with any? (λ k → f zero ≟ f (suc k))
   suc i , suc j , i≢j ∘ suc-injective ,
   punchOut-injective (f₀≢fₖ ∘ (i ,_)) _ fᵢ≡fⱼ
 
+-- Cantor-Schröder-Bernstein for finite sets
+
+injective⇒≤ : {k l : ℕ} {f : Fin k → Fin l} → Injective _≡_ _≡_ f → k ℕ.≤ l
```
MatthewDaggitt, line 877: You can use a `∀` here and omit the type annotation on k and l, e.g. `∀ {k l}`

src/Data/Fin/Properties.agda
```diff
@@ -849,6 +872,27 @@ pigeonhole (s≤s (s≤s m≤n)) f with any? (λ k → f zero ≟ f (suc k))
...
+injective⇒≤ {ℕ.zero} {l} {f} H = z≤n
```
MatthewDaggitt, line 878: Can we call this `inj` instead of `H` everywhere? It's a little bit more meaningful when you're reading the proofs.

src/Data/Fin/Properties.agda
```diff
@@ -761,6 +767,23 @@ pinch-mono-≤ 0F      {suc j} {suc k} (s≤s j≤k) = j≤k
...
+pinch-injective {x = zero} {zero} {suc z} f g p =
+  contradiction (cong suc p) g
+pinch-injective {x = zero} {suc y} {zero} f g p =
+  contradiction (cong suc (sym p)) f
+pinch-injective {x = zero} {suc y} {suc z} f g p =
+  cong suc p
+pinch-injective {x = suc x} {suc y} {suc z} f g p =
+  cong suc
+    ( pinch-injective
+      ( λ q → f (cong suc q))
```
MatthewDaggitt, line 783: Can use function composition:
https://github.com/agda/agda-stdlib/blob/1f87ca2605635bbe05799e322d9ec0c43531873f/src/Function/Base.agda#L52

to rewrite these as `f ∘ cong suc` which should mean that the case's proof fits on 1 line?

src/Data/Fin/Properties.agda
```diff
@@ -761,6 +767,23 @@ pinch-mono-≤ 0F      {suc j} {suc k} (s≤s j≤k) = j≤k
 pinch-mono-≤ (suc i) {0F}    {k}     0≤n       = z≤n
 pinch-mono-≤ (suc i) {suc j} {suc k} (s≤s j≤k) = s≤s (pinch-mono-≤ i j≤k)
 
+pinch-injective :
+  {k : ℕ} {x : Fin k} {y z : Fin (ℕ.suc k)} →
+  suc x ≢ y → suc x ≢ z → pinch x y ≡ pinch x z → y ≡ z
+pinch-injective {y = zero} {zero} f g p = refl
```
MatthewDaggitt, line 773: Can we give more meaningful names to `f` and `g`, e.g. `1+x≢y`? And primary equalities like `p` are usually called `eq`...

MatthewDaggitt (commented):
Thanks for the PR! Probably worth reading through the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) as well, and adjusting the code if you spot any violations.

MatthewDaggitt:
> Using <⇒notInjective and ¬∀⟶∃¬ we can conclude pigeonhole. This would simplify the construction in pigeonhole. Would that change be welcome?

Go for it!

EgbertRijke:
Thanks for your comments! I've made your suggested changes and some changes suggested by the style guide. I think I won't be redoing the pigeonhole principle anymore, because it will require decidability of injectiveness, and an alternative phrasing of `¬∀⟶∃¬` that has the `∀` with implicit arguments. All in all it would be a lot of trouble and I'm not sure if we will gain anything.

If it is not necessary to make anymore style changes, then this PR is done.

EgbertRijke:
Update: There are no further changes I have in mind for this PR, so unless there are some more requests I think this is done.

MatthewDaggitt:
Sorry for the long reply time! Looks great. All that's left to do is to add a note to the `CHANGELOG.md` file in the correct place under the "Minor additions" section towards the end.

Issue #1687, opened 2022-01-13
Add IsRingWithoutOne and RingWithoutOne


Taneb:
These are useful for talking about ideals of rings, which in general have this form.

Akshobhya1234:
@Taneb  Is it Rng as defined here? If required I can migrate it to stdlib

> record IsRng (+ * : Op₂ A) (-_ : Op₁ A) (0# : A) : Set (a ⊔ ℓ) where
  field
    +-isAbelianGroup : IsAbelianGroup + 0# -_
    *-isSemigroup    : IsSemigroup *
    distrib          : * DistributesOver +
    zero             : Zero 0# *

https://github.com/Akshobhya1234/Agda-Algebra/blob/c358d3aa0179a9b6e287d2a06b184856b081d721/src/Structures.agda#L23-L28

Taneb:
@Akshobhya1234 yes, that looks right, but note the changes in #1684

Akshobhya1234:
@MatthewDaggitt or @Taneb  I think we can close this issue. Or is there something that is still missing?

MatthewDaggitt:
Yes, you're right! I forgot to link the PR to the issue. Thanks for the heads up!

PR #1690, opened 2022-01-15
Define Rng

Akshobhya1234:
Define Rng as mentioned in #1687 

Should we call it RingWithoutOne or Rng?

### DIFF
--- a/src/Algebra/Bundles.agda
+++ b/src/Algebra/Bundles.agda
@@ -751,24 +751,24 @@ record CancellativeCommutativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
     )
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Taneb:
It should definitely be called `RingWithoutOne` for consistency with the other definitions (e.g. we have `Semiring`, not `Rig`), and that it's clearer what it means.

src/Algebra/Bundles.agda
```diff
@@ -750,6 +750,40 @@ record CancellativeCommutativeSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
     ; _≉_
     )
 
+------------------------------------------------------------------------
+-- Bundles with 2 binary operations, 1 unary operation & 1 elements
```
MatthewDaggitt, line 754: Typo `1 element`

PR #1692, opened 2022-01-17
Add missing instance to quasigroup and loop.

Akshobhya1234:
Should the binary operations be congruent? Like: 

> record IsQuasigroup (∙ \\ // : Op₂ A) : Set (a ⊔ ℓ) where
  field
    isEquivalence :  IsEquivalence _≈_
    ∙-cong        : Congruent₂ ∙
    \\\\-cong        : Congruent₂ \\\\
    //-cong        : Congruent₂ //
    leftDivides  :  LeftDivides ∙ \\\\
    rightDivides :  RightDivides ∙ //

### DIFF
--- a/src/Algebra/Bundles.agda
+++ b/src/Algebra/Bundles.agda
@@ -945,12 +945,6 @@ record Quasigroup c ℓ : Set (suc (c ⊔ ℓ)) where
   open RawQuasigroup rawQuasigroup public
     using (_≈_; //-rawMagma; \\-rawMagma; ∙-rawMagma)
 
-  setoid : Setoid _ _
-  setoid = record { isEquivalence = isEquivalence }
-
-  open Setoid setoid public
-    using (_≉_)
-
 record RawLoop  c ℓ : Set (suc (c ⊔ ℓ)) where
   infixl 7 _∙_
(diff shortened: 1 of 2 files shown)


src/Algebra/Bundles.agda
```diff
@@ -945,12 +945,6 @@ record Quasigroup c ℓ : Set (suc (c ⊔ ℓ)) where
   open RawQuasigroup rawQuasigroup public
     using (_≈_; //-rawMagma; \\-rawMagma; ∙-rawMagma)
 
-  setoid : Setoid _ _
-  setoid = record { isEquivalence = isEquivalence }
-
-  open Setoid setoid public
-    using (_≉_)
```
MatthewDaggitt, line 952: Where does this now get re-exported from? Looks to me as if it's just gone?

MatthewDaggitt:
> Should the binary operations be congruent?

Yes, it does look like the congruence properties are missing...

src/Algebra/Bundles.agda
Akshobhya1234: This was a mistake so I removed it. It should be in structure and not bundle

Issue #1588, opened 2021-09-16
A `Codata` library based on `--guardedness`


XiaohuWang0921:
Hi, I have been playing around with different ways of implementing coinductive types lately and have encountered some problems with sized types which had me convinced that coinductive records should be preferred to sized types, and therefore the library should support it. Here are the problems I encountered:

1. For some odd reason Agda (I am using 2.6.2) currently does not support `--sized-types` with `--safe` turned on, which means any code using coinduction with sized types cannot use `--safe`.

2. I have come up with a very simple implementation for the Cantor pairing function for streams as coinductive records:
```agda
{-# OPTIONS --safe --guardedness #-}

open import Level using (Level)

private
  variable
    a b c : Level
    A : Set a
    B : Set b
    C : Set c

record Stream (A : Set a) : Set a where
  coinductive
  constructor _∷_
  field
    head : A
    tail : Stream A
open Stream

zipWith : (A → B → C) → Stream A → Stream B → Stream C
zipWith f s t .head = f (s .head) (t .head)
zipWith f s t .tail = zipWith f (s .tail) (t .tail)

cantor : Stream (Stream A) → Stream A
cantor s .head = s .head .head
cantor s .tail = cantor (zipWith _∷_ (s .head .tail) (s .tail))
```
which is way simpler than `Codata.Stream.cantor` which involves two auxiliary functions. And this definition cannot be adapted to sized types, unless one defines an additional type for two-dimensional streams in the following way
```agda
private
  variable
    a : Level
    A : Set a
    i : Size

data Stream2 (A : Set a) (i : Size) : Set a where
  _∷_ : Stream A i → Thunk (Stream2 A) i → Stream2 A i

zipWith∷ : Stream A i → Stream2 A i → Stream2 A i
zipWith∷ (x ∷ xs) (l ∷ ls) =
  (x ∷ λ where .force → l) ∷ λ where .force → zipWith∷ (xs .force) (ls .force)

cantor : Stream2 A i → Stream A i
cantor ((x ∷ xs) ∷ ls) =
  x ∷ λ where .force → cantor (zipWith∷ (xs .force) (ls .force))
```
which is obviously less elegant. And there might be other useful functions that can be easily implemented with coinductive records but not with sized types.

One obvious problem for coinductive records though is that it currently does not support multiple constructors. The solution I came up with is to use sum types or `Data.Maybe.Maybe`. For example one could define `Colist`s like this:
```agda
[...]

PR #1693, opened 2022-01-18
Codata streams by coinductive records

Taneb:
I'm using this for a bunch of personal projects so I'd like to get it in stdlib. However, it's nowhere near ready, really.

I'm only defining Streams for now because they have a regular structure, I don't need to worry about Maybes like I would with Colists or Conats defined this way.

I don't know how far I want to take various relations and properties. "All the way" might be the right answer for this.

I tried to be at least as complete as the other two definitions of streams, but I couldn't figure out how to implement some things, such as `interleave⁺ : List⁺ (Stream A) → Stream A`

### DIFF
--- a/src/Codata/Guarded/Stream.agda
+++ b/src/Codata/Guarded/Stream.agda
@@ -110,10 +110,16 @@ _⁺++_ : List⁺ A → Stream A → Stream A
 head (xs ⁺++ ys) = List⁺.head xs
 tail (xs ⁺++ ys) = List⁺.tail xs ++ ys
 
-{-
 interleave⁺ : List⁺ (Stream A) → Stream A
-interleave⁺ xss = {!!}
--}
+interleave⁺ {A = A} (xs ∷ xss) = go [] xs xss
+  where
+    --    ,- tails of the streams we have already dealt with
+    --   |                 ,-- current focus
+    --   v                 v          v-- future work
+    go : List (Stream A) → Stream A → List (Stream A) → Stream A
+    head (go rev xs yss)        = head xs
+    tail (go rev xs [])         = interleave⁺ (List⁺.reverse (tail xs ∷ rev))
+    tail (go rev xs (ys ∷ yss)) = go (tail xs ∷ rev) ys yss
 
 cycle : List⁺ A → Stream A
 cycle {A = A} (x ∷ xs) = cycleAux List.[]
--- a/src/Codata/Guarded/Stream.agda
+++ b/src/Codata/Guarded/Stream.agda
@@ -8,11 +8,11 @@
 
 module Codata.Guarded.Stream where
 
-open import Data.List.Base using (List)
-open import Data.List.NonEmpty.Base as List⁺ using (List⁺)
+open import Data.List.Base using (List; []; _∷_)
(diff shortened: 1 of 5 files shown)


src/Codata/Guarded/Stream/Relation/Binary/Pointwise.agda
```diff
@@ -0,0 +1,70 @@
...
+open import Codata.Guarded.Stream as Stream using (Stream; head; tail)
+open import Data.Nat.Base using (ℕ)
+open import Function.Base using (_∘_)
+open import Level
+open import Relation.Binary
+
+private
+  variable
+    a b c d ℓ : Level
+    A : Set a
```
MatthewDaggitt, line 14: Could define this as `A B C D : Set a`

src/Codata/Guarded/Stream.agda
```diff
@@ -0,0 +1,138 @@
+{-# OPTIONS --safe --without-K --guardedness #-}
+------------------------------------------------------------------------
```
MatthewDaggitt, line 2: Missing proper standard library module headers. Also the pragmas should go underneath the module headers.

MatthewDaggitt (commented):
See https://github.com/agda/agda-stdlib/issues/1588 for related discussion.

This looks fine to me! Happy to merge things in piece wise if you'd like?

gallais:
This should work:

```agda
interleave⁺ : List⁺ (Stream A) → Stream A
interleave⁺ {A = A} (x ∷ xs) = go [] x xs where

  go : List (Stream A) → Stream A → List (Stream A) → Stream A
  head (go rev xs yss) = head xs
  tail (go rev xs [])         = interleave⁺ (List⁺.reverse (tail xs ∷ rev))
  tail (go rev xs (ys ∷ yss)) = go (tail xs ∷ rev) ys yss
```

src/Codata/Guarded/Stream.agda
```diff
@@ -0,0 +1,142 @@
...
+--
+-- Infinite streams defined as coinductive records
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --without-K --guardedness #-}
+
+module Codata.Guarded.Stream where
+
+open import Data.List.Base using (List)
+open import Data.List.NonEmpty.Base as List⁺ using (List⁺)
```
gallais, line 12:
```suggestion
open import Data.List.Base using (List; []; _∷_)
open import Data.List.NonEmpty.Base as List⁺ using (List⁺; _∷_)
```
Agda does type-directed disambiguation of constructors so you can import all of these and use them unqualified
in the module.

Taneb:
I started implementing non-empty colists like

```agda
record Colist+ (A : Set a) : Set a where
  coinductive
  field
    head : a
    tail : Maybe (Colist+ A)
```

and while it's _possible_ to do things that way, it's frustrating that the termination checker forces me to pattern-match on the Maybe in the tail a lot, I can't just use (e.g.) `Maybe.map (map f) (tail xs)`

MatthewDaggitt:
What's the plan for this? Happy to merge it in if a changelog entry gets added?

Taneb:
Yeah. I'll try to get a changelog entry sorted today

gallais (approved):

Issue #1695, opened 2022-01-23
Enrich properties on Data.List.NonEmpty


HuStmpHrrr:
I find that properties about Data.List.NonEmpty are very limited. I am working on a project which heavily uses it, so I am thinking to prove some properties as well as add some operations to it. I am writing this issue in case some other properties are placed in other folders for some whatever reasons.

HuStmpHrrr:
I am not sure if v2.0 is the next release, but I would like to get this in before that happens.

MatthewDaggitt:
Yes, v2.0 is the next release but its not imminent at the moment.

PR #1696, opened 2022-01-23
add properties of nonempty list

HuStmpHrrr:
#1695

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1473,18 +1473,18 @@ This is a full list of proofs that have changed form to use irrelevant instance 
 
 * In `Data.List.NonEmpty.Base`:
   ```agda
-  drop : ℕ → List⁺ A → List⁺ A
+  drop+ : ℕ → List⁺ A → List⁺ A
   ```
-  When dropping more than the size of the length of the list, the last element remains.
+  When drop+ping more than the size of the length of the list, the last element remains.
   
 * Added new proofs in `Data.List.NonEmpty.Properties`:
   ```agda
   length-++⁺ : length (xs ++⁺ ys) ≡ length xs + length ys
   length-++⁺-tail : length (xs ++⁺ ys) ≡ suc (length xs + length (tail ys))
   ++-++⁺ : (xs ++ ys) ++⁺ zs ≡ xs ++⁺ ys ++⁺ zs
-  ++⁺-cancelˡ′ : xs ++⁺ zs ≡ ys ++⁺ ws → length xs ≡ length ys → zs ≡ ws
-  ++⁺-cancelˡ : xs ++⁺ zs ≡ xs ++⁺ ys → zs ≡ ys
-  drop-++⁺ : drop (length xs) (xs ++⁺ ys) ≡ ys
+  ++⁺-cancelˡ′ : xs ++⁺ zs ≡ ys ++⁺ zs′ → List.length xs ≡ List.length ys → zs ≡ zs′
+  ++⁺-cancelˡ : xs ++⁺ ys ≡ xs ++⁺ zs → ys ≡ zs
+  drop+-++⁺ : drop+ (length xs) (xs ++⁺ ys) ≡ ys
   map-++⁺-commute : map f (xs ++⁺ ys) ≡ map f xs ++⁺ map f ys
   length-map : length (map f xs) ≡ length xs
   map-cong : f ≗ g → map f ≗ map g
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1475,6 +1475,7 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   ```agda
   drop : ℕ → List⁺ A → List⁺ A
   ```
+  When droping more than the size of the length of the list, the last element remains.
   
 * Added new proofs in `Data.List.NonEmpty.Properties`:
   ```agda
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1480,8 +1480,7 @@ This is a full list of proofs that have changed form to use irrelevant instance 
 * Added new proofs in `Data.List.NonEmpty.Properties`:
   ```agda
   length-++⁺ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ List.length l + length l′
-  length-++⁺ [] l′          = refl
-  length-++⁺′ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ suc (List.length l + List.length (List⁺.tail l′))
+  length-++⁺-tail : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ suc (List.length l + List.length (List⁺.tail l′))
   ++-++⁺ : (l : List A) → ∀ {l′ l″} → (l ++ l′) ++⁺ l″ ≡ l ++⁺ l′ ++⁺ l″
   ++⁺-cancelˡ′ : ∀ l l′ {l″ l‴ : List⁺ A} → l ++⁺ l″ ≡ l′ ++⁺ l‴ → List.length l ≡ List.length l′ → l″ ≡ l‴
   ++⁺-cancelˡ : ∀ l {l″ l‴ : List⁺ A} → l ++⁺ l″ ≡ l ++⁺ l‴ → l″ ≡ l‴
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1479,14 +1479,14 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   
 * Added new proofs in `Data.List.NonEmpty.Properties`:
   ```agda
-  length-++⁺ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ List.length l + length l′
-  length-++⁺-tail : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ suc (List.length l + List.length (List⁺.tail l′))
-  ++-++⁺ : (l : List A) → ∀ {l′ l″} → (l ++ l′) ++⁺ l″ ≡ l ++⁺ l′ ++⁺ l″
-  ++⁺-cancelˡ′ : ∀ l l′ {l″ l‴ : List⁺ A} → l ++⁺ l″ ≡ l′ ++⁺ l‴ → List.length l ≡ List.length l′ → l″ ≡ l‴
-  ++⁺-cancelˡ : ∀ l {l″ l‴ : List⁺ A} → l ++⁺ l″ ≡ l ++⁺ l‴ → l″ ≡ l‴
-  drop-++⁺ : ∀ (xs : List A) ys → drop (List.length xs) (xs ++⁺ ys) ≡ ys
-  map-++⁺-commute : ∀ (f : A → B) xs ys → map f (xs ++⁺ ys) ≡ List.map f xs ++⁺ map f ys
-  length-map : ∀ (f : A → B) xs → length (map f xs) ≡ length xs
-  map-cong : ∀ {f g : A → B} → f ≗ g → map f ≗ map g
-  map-compose : {g : B → C} {f : A → B} → map (g ∘ f) ≗ map g ∘ map f
+  length-++⁺ : length (xs ++⁺ ys) ≡ length xs + length ys
+  length-++⁺-tail : length (xs ++⁺ ys) ≡ suc (length xs + length (tail ys))
+  ++-++⁺ : (xs ++ ys) ++⁺ zs ≡ xs ++⁺ ys ++⁺ zs
+  ++⁺-cancelˡ′ : xs ++⁺ zs ≡ ys ++⁺ ws → length xs ≡ length ys → zs ≡ ws
+  ++⁺-cancelˡ : xs ++⁺ zs ≡ xs ++⁺ ys → zs ≡ ys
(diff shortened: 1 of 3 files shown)


src/Data/List/NonEmpty/Base.agda
```diff
@@ -85,6 +85,11 @@ map f (x ∷ xs) = (f x ∷ List.map f xs)
 replicate : ∀ n → n ≢ 0 → A → List⁺ A
 replicate n n≢0 a = a ∷ List.replicate (pred n) a
 
+drop : ℕ → List⁺ A → List⁺ A
+drop zero    xs           = xs
+drop (suc n) (x ∷ [])     = x ∷ []
```
JacquesCarette, line 90: This makes sense, post facto, given the constraints, but is highly non-obvious given just the signature. I would add a comment both in the CHANGELOG and here about that.

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -49,6 +58,46 @@ toList->>= : ∀ {ℓ} {A B : Set ℓ}
...
+    ≡⟨ cong List.concat $ Listₚ.map-compose {g = List⁺.toList} (x ∷ xs) ⟩
   List.concat (List.map List⁺.toList (List.map f (x ∷ xs)))
     ∎
+
+length-++⁺ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ List.length l + length l′
+length-++⁺ [] l′          = refl
+length-++⁺ (x ∷ l) l′
+  rewrite length-++⁺ l l′ = refl
+
+length-++⁺′ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ suc (List.length l + List.length (List⁺.tail l′))
```
JacquesCarette, line 70: the ' is not very indicative here. I think `tail` should appear in the name of the property. Plus, it also feels like this ought to be derived, i.e. the proof should combine a property of `length` and one of `+`.

JacquesCarette (changes requested):

src/Data/List/NonEmpty/Properties.agda
HuStmpHrrr: I will change the name. however, proving it using other theorems is rather cumbersome. this lemma can be proved using the previous one, and then reason `a + (1 + b) = (1 + a + b)`  while using induction it's pretty much the same proof.

src/Data/List/NonEmpty/Base.agda
HuStmpHrrr: will add a comment.

Taneb: I'm not happy with drop returning a non-empty list like this. It doesn't feel right at all to me

HuStmpHrrr: this is not a surprising handling:

https://github.com/agda/agda-stdlib/blob/e3533e8ccbef6ebc1558f79f12661f482e57d49e/src/Data/List/Base.agda#L271-L274

there can be another implementing using `Fin` but this implementation is handy in cases where extra dependent types are unwanted.

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -49,6 +58,46 @@ toList->>= : ∀ {ℓ} {A B : Set ℓ}
...
+map-++⁺-commute f [] ys       = refl
+map-++⁺-commute f (x ∷ xs) ys = cong (λ zs → f x ∷ toList zs) (map-++⁺-commute f xs ys)
+
+length-map : ∀ (f : A → B) xs → length (map f xs) ≡ length xs
+length-map f (_ ∷ xs) = cong suc (Listₚ.length-map f xs)
+
+map-cong : ∀ {f g : A → B} → f ≗ g → map f ≗ map g
+map-cong f≗g (x ∷ xs) = cong₂ _∷_ (f≗g x) (Listₚ.map-cong f≗g xs)
+
+map-compose : {g : B → C} {f : A → B} → map (g ∘ f) ≗ map g ∘ map f
```
MatthewDaggitt, line 102: `f` and `g` should be explicit right? Can can we introduce them in alphabetical order?

src/Data/List/NonEmpty/Base.agda
```diff
@@ -85,6 +85,12 @@ map f (x ∷ xs) = (f x ∷ List.map f xs)
 replicate : ∀ n → n ≢ 0 → A → List⁺ A
 replicate n n≢0 a = a ∷ List.replicate (pred n) a
 
+-- when droping more than the size of the length of the list, the last element remains
+drop : ℕ → List⁺ A → List⁺ A
```
MatthewDaggitt, line 89: This signature feels wrong. Why would you keep the last element? I would imagine that it would return a list rather than a non-empty list?

Can you describe your motivating example for this?

CHANGELOG.md
```diff
@@ -1471,3 +1471,22 @@ This is a full list of proofs that have changed form to use irrelevant instance
...
 
+* In `Data.List.NonEmpty.Base`:
+  ```agda
+  drop : ℕ → List⁺ A → List⁺ A
+  ```
+  When droping more than the size of the length of the list, the last element remains.
+  
+* Added new proofs in `Data.List.NonEmpty.Properties`:
+  ```agda
+  length-++⁺ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ List.length l + length l′
```
MatthewDaggitt, line 1482: We generally drop the individual arguments in the changelog, just give people the "flavour of the type" rather than the full thing which is usually pretty unreadable.
e.g.
```
 length-++⁺ : length (l ++⁺ l′) ≡ length l + length l′
 ++⁺-cancelˡ′ : l ++⁺ l″ ≡ l′ ++⁺ l‴ → length l ≡ length l′ → l″ ≡ l‴
```

MatthewDaggitt (commented):
Thanks for the PR! I've added a few comments.

src/Data/List/NonEmpty/Base.agda
HuStmpHrrr: the point is one shouldn't drop even the length of the list. essentially the second case shouldn't get hit. the same idea applies for the drop function in List; one shouldn't drop too many, so that an undesired base case is never hit. if drop returns a list, then it is rather useless; as one should have been working with a list to begin with. and even if it returned a list, it's going to be empty if dropped too much, so the problem is really not fixed. In my case, List+ is used for a context, and nonemptyness is part of the preconditions. dropping it should return another genuine context, which is at least nonempty.

HuStmpHrrr: consider the counterpart in List for a reference:
https://github.com/agda/agda-stdlib/blob/e3533e8ccbef6ebc1558f79f12661f482e57d49e/src/Data/List/Base.agda#L271-L274

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -49,6 +58,46 @@ toList->>= : ∀ {ℓ} {A B : Set ℓ}
              (List⁺.toList (xs >>= f))
 toList->>= f (x ∷ xs) = begin
   List.concat (List.map (List⁺.toList ∘ f) (x ∷ xs))
-    ≡⟨ cong List.concat $ map-compose {g = List⁺.toList} (x ∷ xs) ⟩
+    ≡⟨ cong List.concat $ Listₚ.map-compose {g = List⁺.toList} (x ∷ xs) ⟩
   List.concat (List.map List⁺.toList (List.map f (x ∷ xs)))
     ∎
+
+length-++⁺ : (l : List A) (l′ : List⁺ A) → length (l ++⁺ l′) ≡ List.length l + length l′
```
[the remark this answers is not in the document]
HuStmpHrrr: sure. it has been a while so I lost track of the naming convention.

src/Data/List/NonEmpty/Properties.agda
HuStmpHrrr: this property is directly taken from the list counterpart:
https://github.com/agda/agda-stdlib/blob/eec2bdf81779ba3e8c7a04e55398a3e3f4bf2e3a/src/Data/List/Properties.agda#L106-L108

CHANGELOG.md
HuStmpHrrr: done

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -49,6 +58,46 @@ toList->>= : ∀ {ℓ} {A B : Set ℓ}
...
+length-++⁺ [] l′          = refl
+length-++⁺ (x ∷ l) l′
+  rewrite length-++⁺ l l′ = refl
```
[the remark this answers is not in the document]
HuStmpHrrr: done

src/Data/List/NonEmpty/Properties.agda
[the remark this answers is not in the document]
HuStmpHrrr: done

CHANGELOG.md
```diff
@@ -1471,3 +1471,22 @@ This is a full list of proofs that have changed form to use irrelevant instance
   1/neg⇒neg : ∀ p .{{_ : NonZero p}} → (1/p : Negative (1/ p)) → Negative p
   ```
 
+* In `Data.List.NonEmpty.Base`:
+  ```agda
+  drop : ℕ → List⁺ A → List⁺ A
+  ```
+  When droping more than the size of the length of the list, the last element remains.
```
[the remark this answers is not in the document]
HuStmpHrrr: done

src/Data/List/NonEmpty/Base.agda
JacquesCarette: Call it `drop-all-but-one`? [I don't like that name, but it might spur better suggestions]

JacquesCarette: sounds like that would drop 1 element!  `drop-n-1` ? The second minus should probably be the one with a dot over it. That's exactly what it does...

HuStmpHrrr: it's not. it's dropping really n. it's just that if you drop too many, it must keep the last element to make the type happy.

JacquesCarette: Indeed, you're right. I don't know, I think I'd prefer `drop` to take a `n < length l` argument. Then have a kind of 'wild' drop that is made artificially total (i.e. your current implementation).

HuStmpHrrr: this is what pauses me. `drop` in list is the wild version, so `drop` in nonempty list should also be wild. in fact, typesafe versions don't exist in list.

MatthewDaggitt: I have to confess I don't understand what the problem is with the existing implementation of `drop` in `Data.List.Base`, I think the semantics of returning the empty list when you drop more than the number of elements in the list is perfectly reasonable. I also think I stand by my feeling that the standard operation `drop` should return a list rather than a non-empty list here. So I'm totally on board with @JacquesCarette's suggestion that we find an alternative name for this variant.

Ideally my first suggestion would be to call it `drop⁺`, but we already use `⁺` to denote introduction forms in predicates and relations over data types, so eventually we'd end up with names like `drop⁺⁺`. Wondering if that's a problem... Alternatively we could go for `drop+`?

src/Data/List/NonEmpty/Properties.agda
MatthewDaggitt: Hmm okay, well that really needs to be fixed too! Definitely swap it here, and I'll open an issue to swap the one in `Data.List.Properties`.

src/Data/List/NonEmpty/Base.agda
HuStmpHrrr: I am ok with any of these.

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -49,6 +58,44 @@ toList->>= : ∀ {ℓ} {A B : Set ℓ}
...
+
+length-++⁺-tail : (xs : List A) (ys : List⁺ A) → length (xs ++⁺ ys) ≡ suc (List.length xs + List.length (List⁺.tail ys))
+length-++⁺-tail [] ys                                     = refl
+length-++⁺-tail (x ∷ xs) ys rewrite length-++⁺-tail xs ys = refl
+
+++-++⁺ : (xs : List A) → ∀ {ys zs} → (xs ++ ys) ++⁺ zs ≡ xs ++⁺ ys ++⁺ zs
+++-++⁺ []      = refl
+++-++⁺ (x ∷ l) = cong (x ∷_) (cong toList (++-++⁺ l))
+
+++⁺-cancelˡ′ : ∀ xs ys {zs ws : List⁺ A} → xs ++⁺ zs ≡ ys ++⁺ ws → List.length xs ≡ List.length ys → zs ≡ ws
```
MatthewDaggitt, line 77: Again can we introduce the variables in alphabetical order (same for the lemma below)?

src/Data/List/NonEmpty/Properties.agda
HuStmpHrrr: also the name is actually a bit awkward. not sure why `compose` needs to be spelled out instead of `∘`

MatthewDaggitt: Happy for that to be changed as well :+1:

src/Data/List/NonEmpty/Properties.agda
HuStmpHrrr: adjusted

Issue #1179, opened 2020-04-29
Add structures/bundles for pairs of strict and non-strict orders


MatthewDaggitt:
Currently, whenever we have a generic proof that involves both a non-strict and a strict order we translate from one to the other using `Relation.Binary.Construct.NonStrictToStrict`/`StrictToNonStrict`. This works fine for generic data, but runs into serious problems when dealing with data where the non-strict and strict orders exploit some structure of the data and are therefore not defined in that way. The most common example is our numeric datatypes. 

Currently we've been getting around this by adding specialised modules for each of the numeric datatypes, e.g. `Data.List.Extrema.Nat`, but this doesn't scale very well as it requires a module for each problematic datatype.

I propose adding some additional relational structures/bundles to the library along the following lines (names are definitely open for discussion!):
```agda
record IsStrictNonStrictPair (_≤_ : Rel A ℓ₂) (_<_ : Rel A ℓ₃) : Set (a ⊔ ℓ ⊔ ℓ₂ ⊔ ℓ₃) where
   field
    <⇒≉    : _<_ ⇒ _≉_
    <⇒≤    : _<_ ⇒ _≤_
    ≤∧≉⇒<  : ∀ {x y} → x ≤ y → x ≉ y → x < y
```

```agda
record IsTotalOrderPair (_≤_ : Rel A ℓ₂) (_<_ : Rel A ℓ₃) : Set (a ⊔ ℓ ⊔ ℓ₂ ⊔ ℓ₃) where
  field
    isTotalOrder          : IsTotalOrder _≤_
    isStrictTotalOrder    : IsStrictTotalOrder _<_
    isStrictNonStrictPair : IsStrictNonStrictPair _≤_ _<_
```
This would then allow us to prove these properties once, rather than having a specialisation for each datatype.

gallais:
Is `_≉_` from a setoid? So it's more like a triple than a pair then.

jamesmckinna:
> Is `_≉_` from a setoid? So it's more like a triple than a pair then.

Well, our methodology, when it's working, sees `...Structures` parametrised on the ambient `_ ≈_`, for which it is indeed canonical that `_≉_` be defined as its negation, either locally in-place, or else a defintional extension of the underlying `RawRelation`/`RawSetoid` defined by `_≈_`. Cf. #2274 .

[...]

Issue #675, opened 2019-03-30
Small change to inject≤


xekoukou:
```agda
inject≤ : ∀ {m n} → Fin m → m ℕ.≤ n → Fin n
inject≤ {n = zero} zero ()
inject≤ {n = suc n} zero le = zero
inject≤ (suc i) (s≤s le) = suc (inject≤ i le)
```
https://agda.github.io/agda-stdlib/Data.Fin.Base.html#3219

This way, we avoid to need to pattern match on ```le``` , which can be hard to do on a proof, we would need to abstact over ```le``` so as to pattern match.

gallais:
You could also use [`≤-pred`](https://agda.github.io/agda-stdlib/Data.Nat.Properties.html#2803) in the `suc i` case.

xekoukou:
Yes, it is much better now : 
```agda
inject≤ : ∀ {m n} → Fin m → m ℕ.≤ n → Fin n
inject≤ {n = zero} zero ()
inject≤ {n = suc n} zero le = zero
inject≤ {n = zero} (suc i) ()
inject≤ {n = suc n} (suc i) le = suc (inject≤ i (≤-pred le))

```

MatthewDaggitt:
Sounds good.

gallais:
Note that with 2.6.0 we should be able to leave out the impossible cases.

xekoukou:
The same can be done to ```fromℕ≤ ``` . 
I can upload the changes now, or later after 2.6.0

MatthewDaggitt:
Probably best to wait until after 2.6.0 and I reset the CHANGELOG

andreasabel:
I think this should wait until 2.6.0.1 is out, see https://github.com/agda/agda/issues/3692. ;)

MatthewDaggitt:
Fixed with 1690f69

PR #1698, opened 2022-01-24
Lighten dependencies of `Data.Nat.Induction`

MatthewDaggitt:
Removed a bunch of unused imports. Would really really like to get rid of the last remaining dependency on `Data.Nat.Properties`...


jamesmckinna:
There's an potential alternative proof using `m≤n⇒m<n∨m≡n` also from `Data.Nat.Properties` ... for the cost of an import of `Data.Sum`. Potayto, potahto. 

Part of the issue seems to be how much `Data.Nat.Properties` covers *everything* and isn't factored into properties of the order, properties of the operations...

... and the fact that `m≤n⇒m<n∨m≡n` itself is a property, not a definition (which would then eliminate the dependency entirely). I guess that's a separate can of worms.

jamesmckinna:
And indeed... the lemma above doesn't really belong, as now, in 
```
-- Properties of _+_ and _≤_/_<_
```
but rather in
```
-- Other properties of _<_
```
in that vast collection...

JacquesCarette:
If I had to guess: accident of history. Various pieces of code evolving independently, and only later clashing by using opposite conventions. As you say, there's no right/wrong here, just lack of harmony.

JacquesCarette:
I've been on a bit of a simplify-the-dependency graph kick (actively so some months ago, on pause right now mostly for lack of time.)  Weird dependencies, in my opinion, will have a real effect down the road when growth becomes increasingly difficult because of cycles.

Axiom went down the route to allow cycles... and ended up with cycles with over 150 modules needing to be type-checked "as one" !

jamesmckinna:
Hmmm. Maybe I spoke too soon. The lemma one actually requires for _<_ would be
m<1+n=>m<n+m=n, which isn't in the library, except as an accident of being definitionally equal to the above lemma, for the particular choices made. But it probably should be a thing in its own right, as a property of _<_. And it isn't definitionally the case, for any of the choices made in Nat.Base, so some dependency might still be required @MatthewDaggit. Ah well.

MatthewDaggitt:
Thanks for the investigations! Yes I agree, the way that the orders are defined in `Data.Nat.Base` is not very pleasing. Indeed see https://github.com/agda/agda-stdlib/issues/1179 for a proposal to get round some of the problems that are caused. As you guessed and @JacquesCarette confirmed, the reasons are indeed historical and long pre-date my involvement with the library. The dependency on `Data.Sum` wouldn't be a problem, as sums are considered "more primitive" than natural numbers in the library in terms of dependecies. I'm not sure it's something we can change at this point though, as _so_ much code relies on the way it's setup currently.

> The lemma one actually requires for _<_ would/should be
m<1+n⇒m<n∨m≡n, which isn't in the library,

That's curious. I've currently broken my installation of Agda, but will have a look at your branch once I've got it running again.

jamesmckinna:
So... the 'curious' lemma indeed has the *identical* proof structure to `m≤n⇒m<n∨m≡n` (which is then a trivial corollary), but is (deliberately) phrased purely in terms of `_<_`, and *mutatis mutandis* would exhibit the analogous relationship between `_<_` and `_<'_` (were they to be given inductively) as between `_≤_` and `_≤'_` currently exploited in `Induction.WF`.

jamesmckinna:
So... there's a discussion perhaps to take offline, until I can get my ideas straight in advance of raising an issue/PR. Some points, more or less in dependency order:

1. `Data.Nat.Properties`, as previously observed, is both heavily dependent on particular choices of definitions in `Data.Nat.Base`, but also on redundant repetition of what could otherwise be reified smart constructors associated with those choices, eg
```
z<s : ∀ {n} → zero < suc n
z<s = s≤s z≤n

s<s : ∀ {m n} (m<n : m < n) → suc m < suc n
s<s = s≤s
```
Such smart constructors could usefully be put in `Data.Nat.Properties.Core`?

2. After @MatthewDaggitt 's pushing of `NonZero` instances everywhere, these now appear in `Data.Nat.Base`, even though they are entirely self-contained...
3. There is also `Data.Nat.Properties.Core`, now containing only a single definition, which once upon a time was needed to resolve a dependency issue (I haven't chased the history of this module); this definition can be eliminated entirely, changing lemmas in 3 places to in-line the pattern match...

So there are (at least) two possible directions to take this towards one (or more) PRs:
* minimalist: eliminate `Data.Nat.Properties.Core` entirely;
* maximalist: enrich `Data.Nat.Properties.Core` to incorporate the smart constructors for `_<_` etc. *and* the `NonZero` instance, and its properties (I'm not clear whether the philosophy regarding `NonZero` instances is to declare them in `*.Base`, but build up their properties elsewhere, or lump them all together; in any case, the proofs of these properties, too, could benefit from the smart constructors...?), and then make that a `public import` in either `Data.Nat.Base` or `Data.Nat.Properties`
* a middle way, but in some ways more drastic: break up `Data.Nat.Properties` into those of the ordering relations, and then those of the operations, and then those concerning their interactions... (and regarding the `NonZero` work as somehow a 'preview' of such factorisation... either inlined, as as `Core` properties)

Matthew's concerns about dependency on the existing 'fat' module which prompted this PR suggests that, sooner or later, it is this size issue which needs dealing with, rather than isolating the actual dependencies on which `Induction`, `Induction.WF` might depend. Of course, *some* isolation of properties of the ordering relations, separated out from `Data.Nat.Properties`, seems the way to go... But I'm interested in what the/a 'best' way forward is/might be?

MatthewDaggitt:
Hmm so we have the original issue discussing why `Data.Fin.Base` depends on `Data.Nat.Properties.Core` here: https://github.com/agda/agda-stdlib/issues/675

And the original PR that implemented it, along the discussion of irrelevance here: https://github.com/agda/agda-stdlib/commit/1690f69e3e9fc82fc4dc026f00186967ee0f310d#r34691463

In my memory there were much stronger reasons for implementing it, but it looks like I've been getting confused with `cast` in more ways than one. Looking back at it now with the benefit of hindsight, I'm really not convinced that @xekoukou's stated reason that the pattern matching on the `<` proof prevents reduction is very convincing. One produces far more robust code by preventing reduction and relying on actual explicit proofs to do the work. And that's all without even mentioning the benefit of eliminating the nasty dependency cycle.

So in answer I think we should open a PR with your minimalist approach to change the definitions of `inject<=` and `fromN<` back to pattern matching explicitly.

As I've mentioned the middle-way would result in an (unacceptable?) hit to searchability of the library, and I'm still hoping that with a smart design, we can avoid the maximalist approach as it feels... unprincipled somehow, as I'm unsure how it would generalise, e.g. what would go in `Data.List.Properties.Core`.

> I don't (yet) see how the library designers really expected such a choice to cash out, or to have formalised the choice in the style guide?

To be honest I'm not aware of anyone has ever really thought about it in those terms before, so no it's not formalised anywhere.

>  @MatthewDaggitt is there now a YouTube feed for the WITS workshop, or else your slides available, to see what you might have said about such issues there?

Unfortunately I can't find any link to my talk. It certainly didn't go into this amount of depth however...

xekoukou:
I am not currently using the agda-stdlib, so it is up to you. 

In general though, I believe that I have spent too much effort to make agda happy, by understanding its quirks. Instead we should spend time on the mathematical problem at hand. 

For example, take _+_ of naturals, from @jespercockx  's [taming of the rew](https://dl.acm.org/doi/abs/10.1145/3434341) .
It is unnatural that the way we split the arguments determines which equality is definitional or propositional.

I have spent so much time optimizing this. So maybe the solution is extensive use of rewrite rules everywhere.

jamesmckinna:
Thanks @xekoukou for the comment, and with sympathy over the trials and tribulations (and development costs) of library development vs. application of such a library to "mathematical problem"s. 

My current version doesn't fundamentally change what was already there; rather, it simplifies one anomalous dependency in favour of a (more) deliberate use of the choices of constructors/admissible pattern-matches already present. The cost seems to be that some definitions/proofs (in `Data.Fin.Base`, `Data.Fin.Properties`) now ostensibly use 'relevant' arguments of the ordering relation, but my understanding (limited as it is) of the evolution of the `Irrelevant` analysis is that the present proofs of the irrelevance of those relations would also devolve to the erasability of such arguments (If this analysis is incorrect, it might be good to have that explained somewhere). 

The new versions, such as they are, also pave the way (I hope) to support removing (some of) the dependency of properties of `Nat` on the slightly mismatched choices of definitions of the orderings, which I find as unnatural as you do (and with good reason!) the choices regarding `_+_` indicated above. Supporting additional rewrites to mitigate biases in representation is one potential part of the solution; supporting more flexible programmer-definable eliminations/pattern matches ('views') on such representations is the/an other side of that story, as Conor and I first advanced a long time ago.

gallais:
> I'm unsure how it would generalise, e.g. what would go in Data.List.Properties.Core.

To compare equals for equals we would need to move the comparison relations
to `Data.Nat.Relation.Binary.Comparison` just like we have `Data.List.Relation.Binary.Prefix`.

The numeric types are a bit special in that we include *way* more in `Base` than for other types.

JacquesCarette:
My feeling is that the module sub-hierarchy has gone a little crazy. `Data.Nat.Relation` makes sense to me (instead of `Data.Nat.Properties.Core`). Splitting it up by arity and then 'kind' (aka `Comparison` - I'd love a better descriptor than 'kind' here) seems over-zealous? It does please my minimalist heart, but for this particular library, seems a bit much.

Minimalist me would certainly be happy with a meta-rule along the lines: non-fat modules can't have more than 20 definitions in them (type, data types, properties, what have you; basically 20 exported names). Fat modules should then go crazy in the opposite direction.

But as to @xekoukou tribulations: in theory, we know what to do. Define Nat whichever way you want, prove the basic properties that make it a rig / ordered rig / lattice / etc, then prove everything else on top of that. Don't allow "proofs by induction" on Nat for properties that are provable from that core. That 'core' becomes the interface to Nat. In theory, we could then change Nat's definition without altering anything else. In practice, I seriously doubt we're ready for that. I do think it is something worth aiming for.

jamesmckinna:
Re @JacquesCarette 's last comment (with which I broadly agree, modulo... history, stdlib as it now is): why is there no distinction made in the library between `zero` and `suc` as *constructors*, and their *values* as ordinary terms? I could imagine better encapsulation of quite a lot reasoning with `Nat` if, instead of `cong suc`, one used `cong 1+_` for example.

jamesmckinna:
Just doing a last `make test` then should be able to open a PR from `pr1698-extended`. But you could always look at the comparisons first... my 'solution' is to expand the range of 'constructors' of the various relations with pattern synonyms, and then aggressively deploy those, under nevertheless quite controlled conditions. Isolating eg the equivalence of `_<_` and `_<′_` in some submodule of `Data.Nat.*` or other would then yield Matthew's holy grail of `Data.Nat.Induction` without appealing to `Data.Nat.Properties`. But I'm open to (further) discussion.

MatthewDaggitt:
Merging this in as it still zaps a few unused dependencies.

PR #1700, opened 2022-01-25
Minor fixes to quasigroup

Akshobhya1234:
I am sorry for missing this. But this change is essential when we try to define direct product for the quasigroup.


MatthewDaggitt:
That's okay! Very rarely is anything committed to this library right the first time :sweat_smile: just happy you caught it in the same release!

PR #1701, opened 2022-01-26
[ new ] NonZero is decidable

gallais:


src/Data/Nat/Properties.agda
```diff
@@ -45,6 +45,14 @@ open import Algebra.Definitions
   using (LeftCancellative; RightCancellative; Cancellative)
 open import Algebra.Structures {A = ℕ} _≡_
 
+------------------------------------------------------------------------
+-- Properties of NonZero
+------------------------------------------------------------------------
+
+_≢0? : U.Decidable NonZero
```
MatthewDaggitt, line 52: Hmm would `nonZero?` be more in keeping with the names in the rest of the library? The underlying definition doesn't really have anything to do with propositional equality symbol used...

Issue #1380, opened 2020-12-14
Fix the mentions of defunct `README` modules


gallais:
In [Data.Nat.Tactic.RingSolver](http://agda.github.io/agda-stdlib/Data.Nat.Tactic.RingSolver.html) we tell people to

> See README.Nat for examples of how to use this solver

But `README.Nat` does not exist anymore. We probably meant to point to
[README.Tactic.RingSolver](http://agda.github.io/agda-stdlib/README.Tactic.RingSolver.html) instead.

Ideally this kind of sanity check should be part of `GenerateEverything`.

jamesmckinna:
And, less ideally because not machine-checkable, advice in `HACKING` or `style-guide` on checking this manually?
Or: for every module `X`, which mentions a `README.*` module, associate an additional module... `X.READMEs` (?) which only imports `X` and all the `README`s that `X` mentions? Can this (easily) be done automagically by `GenerateEverything`?

JacquesCarette:
Any solution to this really ought to be automatable, otherwise it's going to grow into a nightmare.

If our source files were literate, the the reference would be a proper link, which would have broken and thus we would have known to fix it. Is that maybe the best path forward?

jamesmckinna:
That would be a *big* upgrade in ux (making stdlib literate), but also at a big cost, not least in (working out to to trade off) the perhaps-competing forces in the design of the library?

JacquesCarette:
I was thinking it would be done incrementally, i.e. simply start by allowing `.lagda` as input. In fact, we should take a serious look at all of the 1lab's infrastructure -- it is much nicer than stdlib's and sure seems to work well.

jamesmckinna:
Suggest closing this after #2285 in favour of opening a fresh issue on `README`/CI?

JacquesCarette:
Agreed, once there is such a fresh issue, this one should be closed.

silas-hw:
After a scan with `rgrep` I've found the only remaining defunct README reference is in [README.Data.Trie.NonDependent](https://agda.github.io/agda-stdlib/master/README.Data.Trie.NonDependent.html), referencing a `README.Trie.Dependent`. 

I can't seem to find any renamed or new file this should now be referencing. @gallais since it seems you added these files in the git blame, is there anything else it should point to, or is this safe just to remove?

[...]

PR #1704, opened 2022-01-31
[ re #1380 ] small doc correction in Induction.agda

cspollard:
Just pointing documentation in Induction.agda to the right place: Induction.Nat -> Data.Nat.Induction; the former doesn't exist (anymore?) AFAICT.


MatthewDaggitt:
Thank you, great spot!

PR #1708, opened 2022-02-07
Update `Data.Fin.Permutation` to use new function hierarchy

Taneb:
A `Permutation` is now a `Function.Bundles.Inverse`, and `permutation` has been changed accordingly.

I haven't updated the modules that depend on `Data.Fin.Permutation` yet, most frustratingly `Data.Table.Properties`, which is deprecated so I don't want to touch it _too_ much.

In addition, there's now a bit of a mess with names. `_⟨$⟩ʳ_` and `_⟨$⟩ˡ_` don't make much sense with the new function hierarchy, and `inverseˡ` and `inverseʳ` are backwards.


MatthewDaggitt:
> I haven't updated the modules that depend on Data.Fin.Permutation yet, most frustratingly Data.Table.Properties, which is deprecated so I don't want to touch it too much.

I think I'm going to take advantage of v2.0 to nuke everything deprecated before at least v1.2 so I wouldn't worry too much. In particular that'll get rid of all the annoying deprecated `Data.List.Relation.X` modules floating around. If you fix any other broken non-deprecated dependencies, I'll put up a PR that does so.

MatthewDaggitt:
@Taneb removed all the deprecated uses of `Table`. Hopefully should make finishing this much easier!

MatthewDaggitt:
All that's missing is a CHANGELOG entry for the two new properties of injections!

PR #1715, opened 2022-02-14
Removed all deprecated elements up to v1.2

MatthewDaggitt:
Choosing `v1.2` is a strategic move that eliminates 35+ deprecated modules, including `Data.Table`. 

It released nearly 2.5 years ago now, so by the time we release v2.0 it'll probably be closer to 3 years which is a reasonable amount of time I think to expect any active developments to have migrated.

Not currently planning go any further in version numbers in v2.0.


gallais:
I've fixed a bunch of dangling calls to `invIsPartialOrder`. Looking for `invIs` in the lib returns a few
additional hits which suggests we may have a naming discrepancy if we're retiring this name for PO
but are still hanging on to it for other notions?

```shell
$ find src/ -name "*.agda" | xargs grep invIs
src/Relation/Binary/Properties/Poset.agda:  ( invIsPreorder to ≥-isPreorder
src/Relation/Binary/Properties/Poset.agda:  { isPreorder   = PreorderProperties.invIsPreorder
src/Relation/Binary/Properties/Preorder.agda:invIsPreorder : IsPreorder _≈_ (flip _∼_)
src/Relation/Binary/Properties/Preorder.agda:invIsPreorder = record
src/Relation/Binary/Properties/Preorder.agda:  { isPreorder = invIsPreorder
```

MatthewDaggitt:
> Looking for invIs in the lib returns a few
additional hits which suggests we may have a naming discrepancy if we're retiring this name for PO
but are still hanging on to it for other notions?

Nice spot, have deprecated the remaining ones.

PR #1717, opened 2022-02-16
[ new ] Codata.Guarded.Stream.Properties

gallais:
Following #1693, a bunch of properties.

I renamed some of the functions in `Codata.Guarded.Stream.Relation.Binary.Pointwise`
to match the definitions in `Codata.Sized.Stream.Bisimilarity`. This highlights an issue:
`Relation.Binary.Pointwise` vs. `Bisimilarity`. We should probably settle on one naming
scheme and use it throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -457,7 +457,7 @@ Major improvements
 * Previously the division and modulus operators were defined in `Data.Nat.DivMod`
   which in turn meant that using them required importing `Data.Nat.Properties`
   which is a very heavy dependency.
-  
+
 * To fix this, these operators have been moved to `Data.Nat.Base`. The properties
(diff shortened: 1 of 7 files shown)


gallais:
I have a really cool little DSL based on http://www.cse.chalmers.se/~nad/publications/danielsson-productivity.html
to fix the productivity issue in the proof of `chunksOf-repeat`. :partying_face:

Taneb:
That DSL is pretty cool! I'm going to have to mess around with it

Taneb:
Could you add a `lookup-zipWith`?

gallais:
Sure. I'll move it to ready though as I think this can get merged as is and then we can open more PRs to
throw in even more properties.

PR #1720, opened 2022-02-22
Add an Inverse for `swap` in module `Data.Sum.Properties`

sseefried:
Recently I found myself wanting to write this function

```agda
  splitPermute↔ : (m : ℕ) {n : ℕ} → 𝔽 (m + n) ↔ 𝔽 (n + m)
  splitPermute↔ m {n} = +↔⊎ {m} {n} ∘-↔ (swap↔ ∘-↔ flip +↔⊎)
```

where `flip` is defined as

```agda
  flip : ∀ {a b} {A : Set a} {B : Set b} → A ↔ B → B ↔ A
  flip A↔B =
    record { f = f⁻¹ A↔B
           ; f⁻¹ = f A↔B
           ; cong₁ = cong₂ A↔B
           ; cong₂ = cong₁ A↔B
           ; inverse = ×-swap (inverse A↔B) }
    where
      open import Data.Product using () renaming (swap to ×-swap)
```

`swap↔` wasn't available so I've added it in module `Data.Sum.Properties`. It's similar to `+↔⊎` in `Data.Fin.Properties`. 

I would also be very interested in suggestions about whether you think there is a need for the `flip` function somewhere in the Standard Library. Or perhaps there is already a function which does the same thing?

### DIFF
--- a/src/Data/Sum/Properties.agda
+++ b/src/Data/Sum/Properties.agda
@@ -11,10 +11,12 @@ module Data.Sum.Properties where
 open import Level
 open import Data.Sum.Base
 open import Function
+open import Function.Bundles using (mk↔′)
 open import Relation.Binary using (Decidable)
 open import Relation.Binary.PropositionalEquality
 open import Relation.Nullary using (yes; no)
 open import Relation.Nullary.Decidable using (map′)
+
 
 private
   variable
@@ -113,3 +115,6 @@ map₂-cong : {g g′ : C → D} →
             g ≗ g′ →
             map₂ {A = A} g ≗ map₂ g′
 map₂-cong g≗g′ = [,-]-cong ((cong inj₂) ∘ g≗g′)
+
+swap↔ : (A ⊎ B) ↔ (B ⊎ A)
+swap↔ = mk↔′ swap swap swap-involutive swap-involutive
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
Thanks for the PR! To answer your question, you're looking for the following:

https://github.com/agda/agda-stdlib/blob/732d438c431f26e6af17e963b54f0ebf506fb3f9/src/Function/Construct/Symmetry.agda#L199

src/Data/Sum/Properties.agda
```diff
@@ -113,3 +115,6 @@ map₂-cong : {g g′ : C → D} →
             g ≗ g′ →
             map₂ {A = A} g ≗ map₂ g′
 map₂-cong g≗g′ = [,-]-cong ((cong inj₂) ∘ g≗g′)
+
+swap↔ : (A ⊎ B) ↔ (B ⊎ A)
```
MatthewDaggitt, line 119: I think this should be named `swap-↔`. Also it would be great if you could move it next to the `swap-involutive` lemma!

sseefried:
Can do

-------- Original Message --------

MatthewDaggitt:
Oh, and you'll need to add an entry in a suitable place to the CHANGELOG.md file as well :+1:

sseefried:
Thanks Matthew for walking me through these steps.

-------- Original Message --------

sseefried:
Okay, I've made the changes you suggested.

MatthewDaggitt:
Merged in. Thanks again for the great PR :+1:

PR #1721, opened 2022-02-22
Use variables in remaining `Relation.Binary.Construct` modules

MatthewDaggitt:
Not very exciting, just mindlessly using variables to clean up some code.


MatthewDaggitt:
Merging in as the test errors appear to be a temporary glitch in the Github workflows, and the main test suite has passed.

PR #1630, opened 2021-11-20
Attempt at deprecating inspect idiom

MatthewDaggitt:
Attempts to fix #1580.

There's one small fly in the ointment and that's the following use in `Data.List.NonEmpty` 

https://github.com/agda/agda-stdlib/blob/5ac8015bccb973a02275ae8d1e6cf0550807ac49/src/Data/List/NonEmpty.agda#L68

where I get the dreaded `!= w` error below if I try and use `with ... in` instead. Had a hack at it for the last 15 minutes but no matter what I try, I can't seem to get rid of it. @gallais if you have a spare 10 minutes at some point any chance you could take a look?

```
p x != w of type Bool
when checking that the type
{A.a.1 : Level} {A : Set A.a.1} (p : A → Bool) (x : A)
(xs : List A) (w : Bool)
(w₁
 : List
   (List⁺ (Prod.Σ A (λ x₁ → T (p x₁))) ⊎
    List⁺ (Prod.Σ A (λ x₁ → T (not (p x₁)))))) →
List.foldr List._++_ []
(List.map
 Sum.[ (λ x₁ → proj₁ (head x₁) ∷ List.map proj₁ (tail x₁)) ,
 (λ x₁ → proj₁ (head x₁) ∷ List.map proj₁ (tail x₁)) ]
 w₁)
≡ xs →
List.foldr List._++_ []
(List.map
 Sum.[ (λ x₁ → proj₁ (head x₁) ∷ List.map proj₁ (tail x₁)) ,
 (λ x₁ → proj₁ (head x₁) ∷ List.map proj₁ (tail x₁)) ]
 (split p (x ∷ xs) | w | refl | w₁))
≡ x ∷ xs
of the generated with function is well-formed
```

If we can't fix it, then I'll have a go at rewriting the functions. They're currently not very compliant with the standard library style anyway as they mix both computations and proofs.

PR #1723, opened 2022-02-26
Reimplement `split` and `flatten` in `Data.List.NonEmpty`

MatthewDaggitt:
Reasons for this change:
1. Should live in `Data.List.NonEmpty.Base` but because they mixed computation and proofs they could not.
2. The names weren't very descriptive of what they did. N.B. I'm open to shortening them to `groupSeqs` and `ungroupSeqs` if people would prefer.
3. They were blocking the deprecation of the inspect idiom in https://github.com/agda/agda-stdlib/pull/1630

### DIFF
--- a/src/Data/List/NonEmpty/Relation/Unary/All.agda
+++ b/src/Data/List/NonEmpty/Relation/Unary/All.agda
@@ -35,5 +35,5 @@ data All {A : Set a} (P : Pred A p) : Pred (List⁺ A) (a ⊔ p) where
 ------------------------------------------------------------------------
 -- Functions
 
-toListAll : ∀ {xs : List⁺ A} → All P xs → List.All P (toList xs)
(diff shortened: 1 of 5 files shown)


src/Data/List/NonEmpty/Relation/Unary/All.agda
```diff
@@ -0,0 +1,39 @@
...
+
+infixr 5 _∷_
+
+data All {A : Set a} (P : Pred A p) : Pred (List⁺ A) (a ⊔ p) where
+  _∷_ : ∀ {x xs} (px : P x) (pxs : List.All P xs) → All P (x ∷ xs)
+
+------------------------------------------------------------------------
+-- Functions
+
+toListAll : ∀ {xs : List⁺ A} → All P xs → List.All P (toList xs)
```
gallais, line 38: `toList⁺`?

PR #1724, opened 2022-02-26
Propositional equality is a partial order with respect to itself

MatthewDaggitt:


Taneb:
Isn't this true of any equivalence relation? Could we make this more general?

JacquesCarette:
That sure would be nice!

PR #1746, opened 2022-03-31
Show that strict orders over some datatypes are well-founded.

sstucki:
This PR adds proofs that strict partial orders defined over `Data.Bool` and `Data.Fin` are well-founded.

For `Data.Bool`, the proof is just about the "canonical" strict partial order (`_<_`); for `Data.Fin` there is a proof that _any_ strict partial order defined over `Fin` is well-founded (and Noetherian). That proof is more substantial and requires some additions to `Relation.Binary.Properties.StrictPartialOrder` (to show that the induced strict partial order of the inverse of a partial order is again a strict partial order).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -490,9 +490,9 @@ Non-backwards compatible changes
   * `fold-pull`
 
 * In `Data.Fin.Properties`:
-  + the `i` argument to `opposite-suc` has been made explicit
-  + `pigeonhole` has been strenghtened: wlog, we return a proof that `i < j` rather
-    than a mere `i ≢ j`.
+  + the `i` argument to `opposite-suc` has been made explicit;
+  + `pigeonhole` has been strengthened: wlog, we return a proof that
+    `i < j` rather than a mere `i ≢ j`.
 
 Major improvements
 ------------------
@@ -1061,12 +1061,25 @@ Other minor changes
   execState : State s a → s → s
   ```
 
+* Added new proofs in `Data.Bool.Properties`:
+  ```agda
+  <-wellFounded : WellFounded _<_
+  ```
+
 * Added new functions in `Data.Fin.Base`:
   ```
   finToFun  : Fin (m ^ n) → (Fin n → Fin m)
   funToFin  : (Fin m → Fin n) → Fin (n ^ m)
   quotient  : Fin (m * n) → Fin m
   remainder : Fin (m * n) → Fin n
+  ```
+
+* Added new proofs in `Data.Fin.Induction`:
+  every (strict) partial order is well-founded and Noetherian.
+
+  ```agda
+  spo-wellFounded : ∀ {r} {_⊏_ : Rel (Fin n) r} → IsStrictPartialOrder _≈_ _⊏_ → WellFounded _⊏_
+  spo-noetherian  : ∀ {r} {_⊏_ : Rel (Fin n) r} → IsStrictPartialOrder _≈_ _⊏_ → WellFounded (flip _⊏_)
   ```
 
 * Added new definitions and proofs in `Data.Fin.Permutation`:
@@ -1099,7 +1112,7 @@ Other minor changes
   combine-injectiveʳ : combine i j ≡ combine k l → j ≡ l
   combine-injective  : combine i j ≡ combine k l → i ≡ k × j ≡ l
   combine-surjective : ∀ i → ∃₂ λ j k → combine j k ≡ i
-  combine-monoˡ-<    :  i < j → combine i k < combine j l
+  combine-monoˡ-<    : i < j → combine i k < combine j l
 
   lower₁-injective   : lower₁ i n≢i ≡ lower₁ j n≢j → i ≡ j
   pinch-injective    : suc i ≢ j → suc i ≢ k → pinch i j ≡ pinch i k → j ≡ k
@@ -1509,6 +1522,11 @@ Other minor changes
   return : _⟶_ ⇒ SymClosure _⟶_
   join   : SymClosure (SymClosure _⟶_) ⇒ SymClosure _⟶_
   _⋆     : _⟶₁_ ⇒ SymClosure _⟶₂_ → SymClosure _⟶₁_ ⇒ SymClosure _⟶₂_
+  ```
+
+* Added new proofs in `Relation.Binary.Properties.StrictPartialOrder`:
+  ```agda
+  >-strictPartialOrder : StrictPartialOrder s₁ s₂ s₃
   ```
 
 * Added new proofs in `Relation.Binary.PropositionalEquality.Properties`:
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -191,7 +191,7 @@ true  <? _     = no  (λ())
 <-wellFounded _ = acc <-acc
   where
     <-acc : ∀ {x} y → y < x → Acc _<_ y
-    <-acc .false f<t = acc (λ _ → λ())
+    <-acc false f<t = acc (λ _ → λ())
 
 -- Structures
 
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -19,6 +19,7 @@ open import Function.Base
 open import Function.Equality using (_⟨$⟩_)
 open import Function.Equivalence
   using (_⇔_; equivalence; module Equivalence)
+open import Induction.WellFounded using (WellFounded; Acc; acc)
 open import Level using (Level; 0ℓ)
 open import Relation.Binary hiding (_⇔_)
 open import Relation.Binary.PropositionalEquality hiding ([_])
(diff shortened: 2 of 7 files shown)


gallais:
I wonder whether there's a shorter proof reusing [pigeonhole](http://agda.github.io/agda-stdlib/Data.Fin.Properties.html#35821).

sstucki:
I don't know. Do you have a concrete idea of how it might be used? The proof in the PR turned out to be more complex than I initially thought, but I couldn't come up with a simpler one. Feel free to simplify it if you see a way to do so!

gallais:
> Do you have a concrete idea of how it might be used?

For `Fin n`, pile `suc n` constructors `acc` on top of each other which gives you a `Vec (suc n) (Fin n)` of
decreasing values (and so a `Fin (suc n) -> Fin n` function via `lookup`). From pigeonhole you can find two
indices such that the `Fin n` values match and so the `Vec` cannot be strictly decreasing.

I'd probably formalise decreasing vec by porting [`Data.List.Relation.Unary.Linked`](http://agda.github.io/agda-stdlib/Data.List.Relation.Unary.Linked.html) to `Data.Vec`.

sstucki:
I see. Yea, it would be interesting to see if it ends up being shorter. I guess a definite advantage would be that the `Vec`-version of `Linked` would be re-usable (where as prefix-tables don't seem to be).

But there would be some extra work to implement that new data structure properly, which I'll leave to you or anyone else with some extra time on their hands. 😉

sstucki:
Actually, maybe one can use `Linked` directly to represent descending chains `xs` and use `n - length xs` as the measure for such chains (to do do induction on them). The proof that the measure is always positive would then be via the pigeonhole principle. That would avoid adding an extra data structure. WDYT?

sstucki:
Oh, but then you have to write a custom `lookup` which might again be nontrivial without the explicit length-index. Dunno. The devil is in the details, I guess.

sstucki:
That certainly is a lot tidier! 😅 

I could merge your commit into my feature branch and update the PR, but maybe it's easier to just reject the PR and open a new one based on your branch?

gallais:
Merging may not be too bad? There will be a big conflict on that one `module StrictWf` but it should
hopefully not be too intricate to drop one and keep the other. Should I open a PR on your fork?

sstucki:
> Should I open a PR on your fork?

No need, I did the merge (with some very minor changes). Thanks a lot for the help @gallais, the new proof is really much nicer!

sstucki:
The new proof doesn't actually require decidability of the underlying equivalence, so I removed that requirement.

src/Data/Fin/Induction.agda
```diff
@@ -87,3 +98,56 @@ module _ {ℓ} where
...
+  -- The inverse order is also well-founded, i.e. every (strict)
+  -- partial order is also Noetherian.
+
+  spo-noetherian : ∀ {r} {_⊏_ : Rel (Fin n) r} →
+                   IsStrictPartialOrder _≈_ _⊏_ → WellFounded (flip _⊏_)
+  spo-noetherian isSPO = spo-wellFounded >-isStrictPartialOrder
+    where
+      spo : StrictPartialOrder _ _ _
+      spo = record { isStrictPartialOrder = isSPO }
+      open SPO spo using (_>_; >-isStrictPartialOrder)
```
MatthewDaggitt, line 148: We could avoid all the opening and closing of records if you went directly via `Relation.Binary.Construct.Converse`?

https://github.com/agda/agda-stdlib/blob/eb89d28ac6c12eaa15cce6f9ca707c247273f6da/src/Relation/Binary/Construct/Converse.agda#L130-L131

MatthewDaggitt (commented):
Other than that, it's just missing CHANGELOG entries for all the new proofs.

src/Data/Fin/Induction.agda
sstucki: Makes sense. I wasn't aware of that module. But then I'm wondering whether the changes to src/Relation/Binary/Properties/StrictPartialOrder.agda actually make sense. It seems like they're just duplicates. Maybe I should remove them from the PR, WDYT?

src/Data/Bool/Properties.agda
```diff
@@ -186,6 +187,12 @@ true  <? _     = no  (λ())
 <-irrelevant : Irrelevant _<_
 <-irrelevant f<t f<t = refl
 
+<-wellFounded : WellFounded _<_
+<-wellFounded _ = acc <-acc
+  where
+    <-acc : ∀ {x} y → y < x → Acc _<_ y
+    <-acc .false f<t = acc (λ _ → λ())
```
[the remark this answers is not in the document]
sstucki: OK, I'll remove it. But I'm curious: is there an advantage to removing the `.`?

src/Relation/Binary/Properties/StrictPartialOrder.agda
```diff
@@ -11,9 +11,11 @@ open import Relation.Binary
 module Relation.Binary.Properties.StrictPartialOrder
        {s₁ s₂ s₃} (SPO : StrictPartialOrder s₁ s₂ s₃) where
 
-open Relation.Binary.StrictPartialOrder SPO
-  renaming (trans to <-trans)
-open import Relation.Binary.Construct.StrictToNonStrict _≈_ _<_
+open import Data.Product using (_,_)
+open import Function using (flip; _∘_)
+private module SPO = Relation.Binary.StrictPartialOrder SPO
+open SPO hiding (Carrier; _≈_; trans; isEquivalence; module Eq)
```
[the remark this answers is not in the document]
sstucki: OK, I refactored that module and separated the two as you requested. Let me know if looks good to you now.

src/Data/Vec/Relation/Unary/Linked/Properties.agda
```diff
@@ -0,0 +1,67 @@
...
+           Linked R xs → R (Vec.lookup xs i) (Vec.lookup xs j)
+  lookup {i = zero}  {j = suc j} i<j (rx ∷ rxs) = All.lookup j (Linked⇒All rx rxs)
+  lookup {i = suc i} {j = suc j} i<j (_  ∷ rxs) = lookup (<-pred i<j) rxs
+
+------------------------------------------------------------------------
+-- Introduction (⁺) and elimination (⁻) rules for vec operations
+------------------------------------------------------------------------
+-- map
+
+module _ {R : Rel A ℓ} {f : B → A} where
```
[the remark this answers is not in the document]
sstucki: Fixed as requested.

src/Data/Fin/Induction.agda
```diff
@@ -87,3 +98,56 @@ module _ {ℓ} where
     ; wfRec        to ≺-rec
     )
     hiding (wfRec-builder)
+
+
+------------------------------------------------------------------------
+-- Well-foundedness of other (strict) partial orders on Fin
+
+module _ {n e} {_≈_ : Rel (Fin n) e} where
```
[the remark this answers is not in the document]
sstucki: Done.

src/Data/Fin/Induction.agda
sstucki: OK, as a compromise, I left the new proofs but refactored them to use `Converse`. If you feel that the new stuff is superfluous, I don't mind reverting the changes altogether.

sstucki:
I made an attempt to address your comments @MatthewDaggitt. Let me know what you think.

src/Data/Bool/Properties.agda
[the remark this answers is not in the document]
MatthewDaggitt: Not as far as I'm aware :+1:

src/Relation/Binary/Properties/StrictPartialOrder.agda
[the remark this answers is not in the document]
MatthewDaggitt: Yup!

src/Data/Fin/Induction.agda
MatthewDaggitt: So yes in some sense they are just duplicates, but I still think they're valuable as they're optimised for different dimensions. `Construct.Converse` is for flipping multiple structures, whereas the idea of the `Relation.Properties.X` module is that you can pass in a specific structure `X` and get all the relation constructions for that structure with meaningful names instantly in scope.

CHANGELOG.md
```diff
@@ -489,7 +489,19 @@ Non-backwards compatible changes
   * `fold-*`
   * `fold-pull`
 
-* In `Data.Fin.Properties` the `i` argument to `opposite-suc` has been made explicit.
+* In `Data.Fin.Properties`:
+  + the `i` argument to `opposite-suc` has been made explicit;
+  + `pigeonhole` has been strenghtened: wlog, we return a proof that `i < j` rather
+    than a mere `i ≢ j`.
+
+* In `Data.Fin.Induction`:
```
MatthewDaggitt, line 497: This particular entry isn't a breaking change, so should go under the `Other minor changes` section of the CHANGELOG instead of the breaking changes sections. Other than that, it all looks good, so will merge it in after this :+1:

sstucki: Fixed. (I assume you mean the entry about `Data.Fin.Induction`, not the one about `pigeonhole`, which is breaking AFAIK).

MatthewDaggitt: Yup, thanks!

MatthewDaggitt:
Merging in, thanks for the PR!

PR #1747, opened 2022-04-01
Add some facts about decidability of partial orders/equivalences.

sstucki:


src/Data/Fin/Properties.agda
```diff
@@ -1008,8 +1008,25 @@ module _ {f} {F : Set f → Set f} (RF : RawFunctor F) where
 -- If there is an injection from a type A to a finite set, then the type
 -- has decidable equality.
 
+module FromInjection {n a e} {A : Setoid a e}
+                     (inj : Injection A (≡-setoid n)) where
```
MatthewDaggitt, line 1012: Hmm so we don't usually use named inner modules, and if we do we always open them publicly afterwards. The definition of `FromInjection.eq?` is compatible with `eq` right?  `eq?` is also a terrible name for the existing property. So how about we deprecate `eq?` and then add the following:

```agda
module _ {ℓ} {S : Setoid a ℓ} (inj : Injection S (≡-setoid n)) where
  
  inj⇒≟ : -> B.Decidable _≈_
  inj⇒≟ inj = Dec.via-injection inj _≟_

  inj⇒decSetoid : DecSetoid a e
  inj⇒decSetoid = ...
``

File: src/Relation/Binary/Lattice/Properties/JoinSemilattice.agda [hunk 2]
```diff
@@ -101,3 +107,18 @@ dualMeetSemilattice = record
...
+
+------------------------------------------------------------------------
+-- If ≈ is decidable then so is ≤
+
+≈-dec⇒≤-dec : Decidable _≈_ → Decidable _≤_
+≈-dec⇒≤-dec _≟_ x y with (x ∨ y) ≟ y
+... | yes x∨y≈y = yes (trans (x≤x∨y x y) (reflexive x∨y≈y))
+... | no  x∨y≉y = no (contraposition x≤y⇒x∨y≈y x∨y≉y)
+
+≈-dec⇒dec : Decidable _≈_ → IsDecPartialOrder _≈_ _≤_
```
MatthewDaggitt, line 119: I know it's a bit verbose, but I think the names would be a bit more meaningful if we changed this (and the corresponding ones in the other modules) to `≈-dec⇒isDecPartialOrder`

MatthewDaggitt (commented):
Thanks for the PR! Other than those two comments, I think we just need some entries in `CHANGELOG.md`.

src/Data/Fin/Properties.agda
sstucki: Done.

src/Relation/Binary/Lattice/Properties/JoinSemilattice.agda
sstucki: Done.

sstucki:
Made all the suggested changes.

CHANGELOG.md
```diff
@@ -489,7 +489,12 @@ Non-backwards compatible changes
   * `fold-*`
   * `fold-pull`
 
-* In `Data.Fin.Properties` the `i` argument to `opposite-suc` has been made explicit.
+* In `Data.Fin.Properties`:
+  + the `i` argument to `opposite-suc` has been made explicit;
+  + the proof that an injection from a type `A` into `Fin n` induces a
+    decision procedure for `_≡_` on `A` has been generalized to other
+    equivalences over `A` (i.e. to arbitrary setoids), and renamed
+    from `eq?` to the more descriptive `inj⇒≟` and `inj⇒decSetoid`.
```
MatthewDaggitt, line 497: Sorry another niggly organisation point. The generalisation is backwards compatible so should go under `Other minor changes`, the deprecation of `eq?` should live under the `Deprecated names` section. Other than that all looks good.

sstucki: Fixed. I hope it's as it should be now. 😄

MatthewDaggitt: Yes, thanks!

MatthewDaggitt:
Merging in, thanks again for the PR!

Issue #845, opened 2019-07-22
Missing Monoid for Maybe and Add Point


gallais:
For Maybe: empty is nothing and append is the first-just.

For [Add.Point](https://agda.github.io/agda-stdlib/Relation.Nullary.Construct.Add.Point.html), empty is nothing and append is assuming the set we added a point to is a semigroup.

lexvanderstoep:
I'm planning to have a go at this issue. Could you explain what the first-just operator refers to?

gallais:
`Maybe` is typically used for representing errors. As a consequence the sensible
notion of monoid is one where unit is `nothing` and multiplication is the function
which returns the first (left to right) value to be a `just`.

lexvanderstoep:
If I'm not mistaken, it seems to be that the _<|>_ operator is the first-just operator. The corresponding monoid is <|>-isMonoid in Data.Maybe.Properties.

jamesmckinna:
So... should this issue now be closed, since #1751 and @lexvanderstoep 's observation above?
Notwithstanding my own concerns under #1957 and the abandoned (for the time being) #1958 and #1963 , which we might revisit in the future...

JacquesCarette:
It feels like there is nothing left to do for this particular issue, so that keeping it open does not help us?

jamesmckinna:
@JacquesCarette i agree, but had been leaving this open for someone to reach the same conclusion...
... I think this falls under a time when issues weren't linked for automatic closure when PRs were merged addressing them. I've wasted some time on issues that seemed open, but were in fact already 'morally' closed... so let's do this here...

PR #1751, opened 2022-04-12
Add Monoid for Add Point (#845)

Boarders:
Adds Monoid for Pointed and also adds an idempotent proof for `<∣>`.

Note: I don't know the standard library very well (nor much module-fu) so please feel free to make suggestions for improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -772,6 +772,11 @@ Deprecated names
 New modules
 -----------
 
+* Algebraic structures when freely adding an identity element:
+```
+  Algebra.Construct.Add.Identity
+```
+
 * Operations for module-like algebraic structures:
   ```
   Algebra.Module.Core
@@ -941,6 +946,11 @@ New modules
 
 Other minor changes
 -------------------
+
+* Added new proof to `Data.Maybe.Properties`
+```agda
+    <∣>-idem : Idempotent _<∣>_
+```
 
 * The module `Algebra` now publicly re-exports the contents of
   `Algebra.Structures.Biased`.
--- a/src/Algebra/Construct/Add/Identity.agda
+++ b/src/Algebra/Construct/Add/Identity.agda
@@ -1,0 +1,103 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definition of algebraic structures we get from freely adding an
+-- identity element
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.Add.Identity where
+
+open import Algebra.Bundles
+open import Algebra.Core using (Op₂)
+open import Algebra.Definitions
+open import Algebra.Structures
+open import Relation.Binary.Construct.Add.Point.Equality renaming (_≈∙_ to lift≈)
+open import Data.Product using (_,_)
+open import Level using (Level; _⊔_)
+open import Relation.Binary.Core
+open import Relation.Binary.Definitions
+open import Relation.Binary.Structures
+open import Relation.Nullary.Construct.Add.Point
+
+private
+  variable
+    a ℓ : Level
+    A : Set a
+
+liftOp : Op₂ A → Op₂ (Pointed A)
+liftOp op [ p ] [ q ] = [ op p q ]
+liftOp _  [ p ] ∙     = [ p ]
+liftOp _  ∙     [ q ] = [ q ]
+liftOp _  ∙     ∙     = ∙
+
+module _ {_≈_ : Rel A ℓ} {op : Op₂ A} (refl-≈ : Reflexive _≈_) where
+  private
+    _≈∙_ = lift≈ _≈_
+    op∙ = liftOp op
+
+    lift-≈ : ∀ {x y : A} → x ≈ y → [ x ] ≈∙ [ y ]
+    lift-≈ = [_]
+
+  cong₂ : Congruent₂ _≈_ op → Congruent₂ _≈∙_ (op∙)
+  cong₂ R-cong [ eq-l ] [ eq-r ] = lift-≈ (R-cong eq-l eq-r)
+  cong₂ R-cong [ eq   ] ∙≈∙      = lift-≈ eq
+  cong₂ R-cong ∙≈∙      [ eq   ] = lift-≈ eq
+  cong₂ R-cong ∙≈∙      ∙≈∙      = ≈∙-refl _≈_ refl-≈
+
+  assoc : Associative _≈_ op → Associative _≈∙_ (op∙)
+  assoc assoc [ p ] [ q ] [ r ] = lift-≈ (assoc p q r)
+  assoc _     [ p ] [ q ] ∙     = ≈∙-refl _≈_ refl-≈
+  assoc _     [ p ] ∙     [ r ] = ≈∙-refl _≈_ refl-≈
+  assoc _     [ p ] ∙     ∙     = ≈∙-refl _≈_ refl-≈
+  assoc _     ∙     [ r ] [ q ] = ≈∙-refl _≈_ refl-≈
+  assoc _     ∙     [ q ] ∙     = ≈∙-refl _≈_ refl-≈
+  assoc _     ∙     ∙     [ r ] = ≈∙-refl _≈_ refl-≈
+  assoc _     ∙     ∙     ∙     = ≈∙-refl _≈_ refl-≈
+
+  identityˡ : LeftIdentity _≈∙_ ∙ (op∙)
+  identityˡ [ p ] = ≈∙-refl _≈_ refl-≈
+  identityˡ ∙     = ≈∙-refl _≈_ refl-≈
+
+  identityʳ : RightIdentity _≈∙_ ∙ (op∙)
+  identityʳ [ p ] = ≈∙-refl _≈_ refl-≈
+  identityʳ ∙     = ≈∙-refl _≈_ refl-≈
+
+  identity : Identity _≈∙_ ∙ (op∙)
+  identity = identityˡ , identityʳ
+
+module _ {_≈_ : Rel A ℓ} {op : Op₂ A} where
+  private
+    _≈∙_ = lift≈ _≈_
+    op∙ = liftOp op
+
+  isMagma : IsMagma _≈_ op → IsMagma _≈∙_ op∙
+  isMagma M =
+    record
+    { isEquivalence = ≈∙-isEquivalence _≈_ M.isEquivalence
+    ; ∙-cong        = cong₂ M.refl M.∙-cong
+    } where module M = IsMagma M
+
+  isSemigroup : IsSemigroup _≈_ op → IsSemigroup _≈∙_ op∙
+  isSemigroup S = record
+    { isMagma = isMagma S.isMagma
+    ; assoc   = assoc S.refl S.assoc
+    } where module S = IsSemigroup S
+
+  isMonoid : IsSemigroup _≈_ op → IsMonoid _≈∙_ op∙ ∙
+  isMonoid S = record
+    { isSemigroup = isSemigroup S
+    ; identity    = identity S.refl
+    } where module S = IsSemigroup S
+
+semigroup : Semigroup a (a ⊔ ℓ) → Semigroup a (a ⊔ ℓ)
+semigroup S = record
+  { Carrier = Pointed S.Carrier
+  ; isSemigroup = isSemigroup S.isSemigroup
+  } where module S = Semigroup S
+
+monoid : Semigroup a (a ⊔ ℓ) → Monoid a (a ⊔ ℓ)
+monoid S = record
+  { isMonoid = isMonoid S.isSemigroup
+  } where module S = Semigroup S
--- a/src/Data/Maybe/Properties.agda
+++ b/src/Data/Maybe/Properties.agda
@@ -114,6 +114,10 @@ module _ {A : Set a} where
   <∣>-identity : Identity nothing _<∣>_
   <∣>-identity = <∣>-identityˡ , <∣>-identityʳ
 
+  <∣>-idem : Idempotent _<∣>_
+  <∣>-idem (just x) = refl
+  <∣>-idem nothing = refl
+
 module _ (A : Set a) where
 
   open Structures {A = Maybe A} _≡_


src/Relation/Nullary/Construct/Add/Point.agda
```diff
@@ -8,8 +8,84 @@
 
 module Relation.Nullary.Construct.Add.Point where
 
+open import Algebra.Bundles
```
MatthewDaggitt, line 11: This content should be in a new module `Algebra.Construct.Add.Point` rather than here I think!

src/Relation/Nullary/Construct/Add/Point.agda
```diff
@@ -8,8 +8,84 @@
...
+
+private
+  variable
+    a : Level
+
+module _ {A : Set a} {op : Op₂ A} where
+  open Definitions {A = Pointed A} _≡_
+  private
+    module Str = Structures {A = A} _≡_
+  module _ {semigroup : Str.IsSemigroup op} where
```
MatthewDaggitt, line 32: Hmm not every proof in this file requires the underlying op to be a semigroup right? Most of the basic definitions simply require reflexivity of the equality. Could you parameterise the proofs by just the locally required properties in instead?

src/Relation/Nullary/Construct/Add/Point.agda
```diff
@@ -8,8 +8,84 @@
...
   using (≡-dec) renaming (just-injective to []-injective)
+open import Data.Product using (_,_)
+open import Relation.Binary.PropositionalEquality
+
+private
+  variable
+    a : Level
+
+module _ {A : Set a} {op : Op₂ A} where
+  open Definitions {A = Pointed A} _≡_
```
MatthewDaggitt, line 29: Using propositional equality everywhere is a bit restrictive. I think we should take the underlying equality of the structure without the point, and then use pointed equality defined in `Relation.Binary.Construct.Add.Point` as the equality for the new structure?

MatthewDaggitt (commented):
Thank you for the great first PR, we really appreciate it! Just a heads-up that this isn't the easiest `low-hanging-fruit` on the tree, so it might require a couple of review iterations :sweat_smile: 

I've started off with the most important points and we can go from there.

src/Relation/Nullary/Construct/Add/Point.agda
Boarders: Yes! Roughly, what is the best way to do that? Shall I just add the semigroup assumption as an argument to associativity or something else?

MatthewDaggitt: I would just add associativity as an assumption right? Unless I'm misunderstanding something we don't need the full semigroup?

Boarders: Ah, got it.

Boarders:
Have another draft of this attempting to address your points so far @MatthewDaggitt .

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,97 @@
...
+  lift-op-identityʳ [ p ] = refl refl-≈
+  lift-op-identityʳ ∙     = refl refl-≈
+
+  lift-op-identity : Identity ∙ lift-op
+  lift-op-identity = lift-op-identityˡ , lift-op-identityʳ
+
+  open Structures {A = Pointed A} (Pointwise _≈_)
+
+  module _ {equiv : Rel.IsEquivalence _≈_}{op-cong₂ : Underlying.Congruent₂ op} where
+    lift-op-isMagma : IsMagma lift-op
```
MatthewDaggitt, line 71: These proofs should be taken out from under the anonymous modules above and take the form:
 - `IsMagma → IsMagma`
 - `IsSemigroup → IsSemigroup`
 - `IsSemigroup → IsMonoid`

and 
 - `Magma → Magma`
 - `Semigroup → Semigroup`
 - `Semigroup → Monoid`

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,97 @@
...
+open import Relation.Binary.Core
+open import Relation.Binary.Definitions
+import Relation.Binary.Structures as Rel
+open import Relation.Nullary.Construct.Add.Point
+
+private
+  variable
+    a ℓ : Level
+
+module _ {A : Set a} {_≈_ : Rel A ℓ}{refl-≈ : Reflexive _≈_} {op : Op₂ A} where
```
MatthewDaggitt, line 27: Reflexivity won't be able to be inferred so will need to be explicit.

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,97 @@
...
+  open Definitions {A = Pointed A} (Pointwise _≈_)
+  private
+    module Underlying = Definitions {A = A} _≈_
+
+  lift-≈ : ∀ {x y : A} → x ≈ y → Pointwise _≈_ [ x ] [ y ]
+  lift-≈ = just
+
+  lift-op : Pointed A → Pointed A → Pointed A
```
MatthewDaggitt, line 35: The operation itself doesn't need the equality or proof of reflexivity to define right? And it won't be able to infer the underlying op, so it should take `op` as an explicit parameter.

src/Relation/Nullary/Construct/Add/Point.agda
```diff
@@ -8,8 +8,16 @@
 
 module Relation.Nullary.Construct.Add.Point where
 
+open import Algebra.Bundles
```
MatthewDaggitt, line 11: Can we get rid of these now unnecessary imports?

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,97 @@
...
+open import Relation.Nullary.Construct.Add.Point
+
+private
+  variable
+    a ℓ : Level
+
+module _ {A : Set a} {_≈_ : Rel A ℓ}{refl-≈ : Reflexive _≈_} {op : Op₂ A} where
+  open Definitions {A = Pointed A} (Pointwise _≈_)
+  private
+    module Underlying = Definitions {A = A} _≈_
```
MatthewDaggitt, line 30: I think if there are two different underlying equalities, probably more readable to just open `Definition` once unparameterised and provide the two equalities explicitly. You could give the second equality a meaningful name like `_≈∙_ = Pointwise _≈_`

MatthewDaggitt (commented):
Thanks, here are a few more!

src/Algebra/Construct/Add/Point.agda
Boarders: Would it be best to lift it out of the anonymous module?

src/Relation/Nullary/Construct/Add/Point.agda
Boarders: whoops! some sort of git error in not reverting this file

src/Algebra/Construct/Add/Point.agda
MatthewDaggitt: Yes, probably!

src/Algebra/Construct/Add/Point.agda
Boarders: Gave this a go but not sure if the structure is exactly what you had in mind since the `Is*` proofs are still parameterized.

Boarders:
Have another draft whenever you have the time @MatthewDaggitt

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,102 @@
...
+
+module _ {A : Set a} {_≈_ : Rel A ℓ} {op : Op₂ A} (refl-≈ : Reflexive _≈_) where
+  open Definitions
+  private
+    _≈∙_ = Pointwise _≈_
+
+  lift-≈ : ∀ {x y : A} → x ≈ y → Pointwise _≈_ [ x ] [ y ]
+  lift-≈ = just
+
+  lift-op-cong₂ : Congruent₂ _≈_ op → Congruent₂ _≈∙_ (lift-op op)
```
MatthewDaggitt, line 41: I think we've decided in general that the best way to import and use these `Construct` modules is qualified, e.g.
```
import Algebra.Construct.Add.Point as Point
```
therefore we usually omit the `lift-op` prefix because a) it doesn't add any information, what sort of lifting? what sort of op? and b) it's cleaner to be able to write `Point.cong` rather than `Point.lift-op-cong₂` and c) because you can easily chance the name given to the module whereas you can't rename `lift-op` easily for every proof in the file.

So essentially I think we can drop the `lift-op` in every name.

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,102 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definition of algebraic structures we get from adding a disjoint point
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.Add.Point where
```
MatthewDaggitt, line 9: You could in theory add a point element that was the zero instead of the identity element right? So this module should probably be called `Add.Identity` rather than `Add.Point` so that we have space for `Add.Zero` at some later date. You should also adjust the comment in the top header to match.

src/Algebra/Construct/Add/Point.agda
```diff
@@ -0,0 +1,102 @@
...
+module _ {A : Set a} {_≈_ : Rel A ℓ} {op : Op₂ A} where
+  open Structures
+  private
+    _≈∙_ = Pointwise _≈_
+
+  lift-op-isMagma : IsMagma _≈_ op → IsMagma _≈∙_ (lift-op op)
+  lift-op-isMagma M =
+    record
+    { isEquivalence = isEquivalence M.isEquivalence
+    ; ∙-cong        = lift-op-cong₂ (IsEquivalence.refl M.isEquivalence) M.∙-cong
```
MatthewDaggitt, line 77: I think all of these `IsEquivalence.refl M.isEquivalence` here and below can just be `M.refl`?

MatthewDaggitt (commented):
Thanks. Here's the final(?) batch.

You will also need to an entry in the CHANGELOG.md file the top level describing the changes. One entry in the new modules section, and one entry in the minor changes section.

src/Algebra/Construct/Add/Point.agda
MatthewDaggitt: Yup, exactly what I had in mind!

Boarders:
Updated with the suggested improvements.

src/Algebra/Construct/Add/Identity.agda
```diff
@@ -0,0 +1,103 @@
...
+open import Relation.Binary.Definitions
+open import Relation.Binary.Structures
+open import Relation.Nullary.Construct.Add.Point
+
+private
+  variable
+    a ℓ : Level
+    A : Set a
+
+lift-op : Op₂ A → Op₂ (Pointed A)
```
MatthewDaggitt, line 29: Could we rename this `liftOp`? We reserve dashes in names for separating distinct functions.

src/Algebra/Construct/Add/Identity.agda
```diff
@@ -0,0 +1,103 @@
...
+lift-op op [ p ] [ q ] = [ op p q ]
+lift-op _  [ p ] ∙     = [ p ]
+lift-op _  ∙     [ q ] = [ q ]
+lift-op _  ∙     ∙     = ∙
+
+module _ {_≈_ : Rel A ℓ} {op : Op₂ A} (refl-≈ : Reflexive _≈_) where
+  private
+    _≈∙_ = Pointwise _≈_
+
+    lift-≈ : ∀ {x y : A} → x ≈ y → Pointwise _≈_ [ x ] [ y ]
```
MatthewDaggitt, line 39: Could write `[ x ] ≈∙ [ y ]` instead of `Pointwise _≈_ [ x ] [ y ]`

src/Algebra/Construct/Add/Identity.agda
```diff
@@ -0,0 +1,103 @@
...
+
+lift-op : Op₂ A → Op₂ (Pointed A)
+lift-op op [ p ] [ q ] = [ op p q ]
+lift-op _  [ p ] ∙     = [ p ]
+lift-op _  ∙     [ q ] = [ q ]
+lift-op _  ∙     ∙     = ∙
+
+module _ {_≈_ : Rel A ℓ} {op : Op₂ A} (refl-≈ : Reflexive _≈_) where
+  private
+    _≈∙_ = Pointwise _≈_
```
MatthewDaggitt, line 37: Again could add `op∙ = lift-op op` here?

src/Algebra/Construct/Add/Identity.agda
```diff
@@ -0,0 +1,103 @@
...
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.Add.Identity where
+
+open import Algebra.Bundles
+open import Algebra.Core using (Op₂)
+open import Algebra.Definitions
+open import Algebra.Structures
+open import Data.Maybe.Relation.Binary.Pointwise
```
MatthewDaggitt, line 16: Hmm I've just realised that we should really be using

https://github.com/agda/agda-stdlib/blob/650e05f3a5ef06cb938c3b91a4fc8b9b8d7c5ef2/src/Relation/Binary/Construct/Add/Point/Equality.agda#L27

as the underlying equality rather than the definition from `Data.Maybe.Relation.Binary.Pointwise`.

I can't remember why the former isn't implemented using the latter... It probably should be, but that's for another issue.

MatthewDaggitt (commented):
Thanks! Apologies I've found a few more small points, but then it really is finished, promise!

Boarders:
Life got in the way of this, will make these changes this week!

CHANGELOG.md
```diff
@@ -942,6 +947,26 @@ New modules
 Other minor changes
 -------------------
 
+* Added new proofs to `Algebra.Construct.Add.Identity`
```
MatthewDaggitt, line 950: Oh also no need to add this entry, as it's already covered under the "New modules" header :+1:

MatthewDaggitt:
@Boarders any prospect of you finding the time to make the last tweaks so I can merge this in?

Boarders:
Ah apologies, this completely slipped off my radar. Should be able to get these changes to you today!

Boarders:
I made these changes but let me know if you don't like the style of naming I introduced and I'm happy to change it or anything else I overlooked. Thank you for such through reviewing and apologies for the delay.

MatthewDaggitt:
Looks great! As soon as the tests pass I'll merge it in :+1:  Thanks for the PR!

Issue #481, opened 2018-09-26
Look into graded monads


gallais:
We could readily make use of them with `Data.Vec` and `Data.List.All`.

iitalics:
Is this a suitable interface?

```agda
module _ {a r} (monoid : Monoid a r) where
  open Monoid monoid

  record GradedMonad {ℓ} (_*_ : Carrier → Set ℓ → Set ℓ)
                     : Set (a ⊔ r ⊔ suc ℓ) where
    field
      subst : ∀ {x y A} → (x ≈ y) → (x * A) → (y * A)
      fmap : ∀ {x A B} → (A → B) → (x * A) → (x * B)
      return : ∀ {A} → A → (ε * A)
      join : ∀ {x y A} → (x * (y * A)) → (x ∙ y) * A
```

And Vec implemenetation:
```agda
VecGMonad : ∀ {ℓ} → GradedMonad *-1-monoid {ℓ} (flip Vec)
VecGMonad = record
  { subst = PEq.subst _
  ; fmap = map
  ; return = [_]
  ; join = concat }
```

However I'm not sure how `Data.List.All` fits into this.

gallais:
> Is this a suitable interface?

A few remarks:

* The first constraint can be formulated in terms of [`Respects`](https://agda.github.io/agda-stdlib/Relation.Binary.Core.html#4263)
* I think this would be a `RawGradedMonad` as we are not enforcing any laws.
* I think we could have a `∀ {x} → Functor (x *_)` constraint rather than the inlined `fmap`.
* Similarly, we could ask whether it makes sense to talk about a graded applicative
* Is there a reason to pick a join-based presentation rather than the bind-based one we have for `RawMonad`?

> However I'm not sure how `Data.List.All` fits into this.

I'm not sure whether they do. These "look into" issues (cf. #480 too) where thrown
together during ICFP whenever I saw someone talk about a topic that seemed relevant
to us but we had not formalized yet.

It's 100% exploratory work (at least for me) and to pin down a design probably implies
including many examples and formalizing results that prove they are usable in practice.

EdNutting:
Waking up seemingly a very old issue here - have graded monads made their way into the stdlib anywhere? If not, what would be required to do so?

JacquesCarette:
They're not even in agda-categories yet. A PR there would be great!

EdNutting:
[...]

Issue #1238, opened 2020-06-12
Setoid parameter in algebra records


mechvel:
May be it is better for standard library to improve the record design for algebra?
We wrote about this recently in the Agda list.
1) I complain on that (Magma.setoid +-magma) and  (Magma.setoid *-magma) occurs not the same
    in Semiring. The same is for decSetoid. I do not find any better way out as introducing 
    +-setoid and *-setoid in the application. And this does not look nice, it misleads the user.
2)  James Wood wrote there about certain tricks to  fix the effect.
3)  I suggested (in the Agda mail list) the approach to improve the design, with attaching a piece    
     of code.

Breaking backwards compatibility is not so bad at the current stage.
Because
a) so far, each new lib version still breaks almost all the user's application modules,
b) it is not difficult to port the application to a new version,
c) each application specifies the needed Agda version (and hence the lib version),
   install this version and enjoy,
Probably, we need to continue this way until it becomes stable in somewhat a natural way.

JacquesCarette:
The problem here is one of sharing: there is a `Setoid` that appears twice, but should really only appear once, structurally.

One way to deal with it is via having both bundled and unbundled versions of some structures. That way the sharing can be expressed via parameters instead of fields. The drawback is that even though the bundled and unbundled versions are semantically equivalent, Agda doesn't know that, and so 'work' needs to be done to go back and forth. [In theory, cubical solves this, but I am not convinced if this is yet true in practice.]

Note that the Arend theorem prover has solved this problem really nicely already. And its solution seems very Agda-compatible to me. The Agda developers are aware of this, but they may not yet be aware of the impact it has on the current library development.

mechvel:
A simple and natural solution is needed.
My suggestion is concrete.
A small sample code is attached here as 
[Structures.agda.zip](https://github.com/agda/agda-stdlib/files/4774437/Structures.agda.zip)
to demonstrate the approach. 
IsMagma is parameterized by Setoid,
IsSemiring is parameterized by Magma, which is taken as +-magma, and the items for `*` are 
tied there to (Magma.setoid +-magma) . And *-magma is formed over this setoid. 
And so on.
I think the `base' setoid is unique in this design. Is it?
Is not all this natural and simple?
What people think of this?  
May be this approach is well-known, all right, but can the library be fixed this way?
May be this implements of what you call  "sharing can be expressed via parameters instead of fields",  but it is a concrete version of this general idea. The details matter.

mechvel:
[...]

PR #1752, opened 2022-04-13
[ fix #1694 ] Add ordered algebraic structures.

### LINKED ISSUE #1694: Ordered algebraic structures?
I was looking for _partially ordered monoid_ (`POMonoid`) but could not find it.  This is a chimera of a monoid and a partial order such that the composition operation of the monoid is monotone, i.e., we have `x*y <= x'*y'` whenever `x<=x'` and `y<=y'`.

Do we have ordered algebraic structures yet?
If not where should we place them in the module hierarchy?

sstucki:
This is a starting point for addressing #1694 and eventually #481 (since we need promonoids for grading).

I implemented this a while ago but never got around to submitting a PR so I refactored it to fit current stdlib conventions as best I could, but please let me know if anything looks wrong. The PR is also missing a CHANGELOG patch, but I need some advice on what to put where there. (Is there a CHANGELOG guide somewhere?)

### DIFF
--- a/src/Algebra/Ordered/Bundles.agda
+++ b/src/Algebra/Ordered/Bundles.agda
@@ -1,0 +1,346 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definitions of ordered algebraic structures like promonoids and
+-- posemigroups (packed in records together with sets, orders,
+-- operations, etc.)
+------------------------------------------------------------------------
+
+-- The contents of this module should be accessed via `Algebra.Order`.
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Ordered.Bundles where
+
+open import Algebra.Core
+open import Algebra.Bundles
+open import Algebra.Ordered.Structures
+open import Level using (suc; _⊔_)
+open import Relation.Binary using (Rel)
+open import Relation.Binary.Bundles using (Preorder; Poset)
+
+------------------------------------------------------------------------
+-- Bundles of preordered structures
+
+-- Preordered magmas (promagmas)
+
+record Promagma c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier    : Set c
+    _≈_        : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_        : Rel Carrier ℓ₂  -- The preorder.
+    _∙_        : Op₂ Carrier     -- Multiplication.
+    isPromagma : IsPromagma _≈_ _≤_ _∙_
+
+  open IsPromagma isPromagma public
+
+  preorder : Preorder c ℓ₁ ℓ₂
+  preorder = record { isPreorder = isPreorder }
+
+  magma : Magma c ℓ₁
+  magma = record { isMagma = isMagma }
+
+-- Preordered semigroups (prosemigroups)
+
+record Prosemigroup c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier        : Set c
+    _≈_            : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_            : Rel Carrier ℓ₂  -- The preorder.
+    _∙_            : Op₂ Carrier     -- Multiplication.
+    isProsemigroup : IsProsemigroup _≈_ _≤_ _∙_
+
+  open IsProsemigroup isProsemigroup public
+
+  promagma : Promagma c ℓ₁ ℓ₂
+  promagma = record { isPromagma = isPromagma }
+
+  open Promagma promagma public using (preorder; magma)
+
+  semigroup : Semigroup c ℓ₁
+  semigroup = record { isSemigroup = isSemigroup }
+
+-- Preordered monoids (promonoids)
+
+record Promonoid c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier     : Set c
+    _≈_         : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_         : Rel Carrier ℓ₂  -- The preorder.
+    _∙_         : Op₂ Carrier     -- The monoid multiplication.
+    ε           : Carrier         -- The monoid unit.
+    isPromonoid : IsPromonoid _≈_ _≤_ _∙_ ε
+
+  open IsPromonoid isPromonoid public
+
+  prosemigroup : Prosemigroup c ℓ₁ ℓ₂
+  prosemigroup = record { isProsemigroup = isProsemigroup }
+
+  open Prosemigroup prosemigroup public
+    using (preorder; magma; promagma; semigroup)
+
+  monoid : Monoid c ℓ₁
+  monoid = record { isMonoid = isMonoid }
+
+-- Preordered commutative monoids (commutative promonoids)
+
+record CommutativePromonoid c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier : Set c
+    _≈_     : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_     : Rel Carrier ℓ₂  -- The preorder.
+    _∙_     : Op₂ Carrier     -- The monoid multiplication.
+    ε       : Carrier         -- The monoid unit.
+    isCommutativePromonoid : IsCommutativePromonoid _≈_ _≤_ _∙_ ε
+
+  open IsCommutativePromonoid isCommutativePromonoid public
+
+  promonoid : Promonoid c ℓ₁ ℓ₂
+  promonoid = record { isPromonoid = isPromonoid }
+
+  open Promonoid promonoid public
+    using (preorder; magma; promagma; semigroup; prosemigroup; monoid)
+
+  commutativeMonoid : CommutativeMonoid c ℓ₁
+  commutativeMonoid = record { isCommutativeMonoid = isCommutativeMonoid }
+
+  open CommutativeMonoid commutativeMonoid public using (commutativeSemigroup)
+
+-- Preordered semirings (prosemirings)
+
+record Prosemiring c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _*_
+  infixl 6 _+_
+  field
+    Carrier       : Set c
+    _≈_           : Rel Carrier ℓ₁
+    _≤_           : Rel Carrier ℓ₂
+    _+_           : Op₂ Carrier
+    _*_           : Op₂ Carrier
+    0#            : Carrier
+    1#            : Carrier
+    isProsemiring : IsProsemiring _≈_ _≤_ _+_ _*_ 0# 1#
+
+  open IsProsemiring isProsemiring public
+
+  +-commutativePromonoid : CommutativePromonoid c ℓ₁ ℓ₂
+  +-commutativePromonoid = record
+    { isCommutativePromonoid = +-isCommutativePromonoid }
+
+  open CommutativePromonoid +-commutativePromonoid public
+    using (preorder)
+    renaming
+    ( magma                to +-magma
+    ; promagma             to +-promagma
+    ; semigroup            to +-semigroup
+    ; prosemigroup         to +-prosemigroup
+    ; monoid               to +-monoid
+    ; promonoid            to +-promonoid
+    ; commutativeSemigroup to +-commutativeSemigroup
+    ; commutativeMonoid    to +-commutativeMonoid
+    )
+
+  *-promonoid : Promonoid c ℓ₁ ℓ₂
+  *-promonoid = record { isPromonoid = *-isPromonoid }
+
+  open Promonoid *-promonoid public
+    using ()
+    renaming
+    ( magma        to *-magma
+    ; promagma     to *-promagma
+    ; semigroup    to *-semigroup
+    ; prosemigroup to *-prosemigroup
+    ; monoid       to *-monoid
+    )
+
+  semiring : Semiring c ℓ₁
+  semiring = record { isSemiring = isSemiring }
+
+------------------------------------------------------------------------
+-- Bundles of partially ordered structures
+
+-- Partially ordered magmas (pomagmas)
+
+record Pomagma c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier   : Set c
+    _≈_       : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_       : Rel Carrier ℓ₂  -- The partial order.
+    _∙_       : Op₂ Carrier     -- Multiplication.
+    isPomagma : IsPomagma _≈_ _≤_ _∙_
+
+  open IsPomagma isPomagma public
+
+  poset : Poset c ℓ₁ ℓ₂
+  poset = record { isPartialOrder = isPartialOrder }
+
+  promagma : Promagma c ℓ₁ ℓ₂
+  promagma = record { isPromagma = isPromagma }
+
+  open Promagma promagma public using (preorder; magma)
+
+-- Partially ordered semigroups (posemigroups)
+
+record Posemigroup c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier       : Set c
+    _≈_           : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_           : Rel Carrier ℓ₂  -- The partial order.
+    _∙_           : Op₂ Carrier     -- Multiplication.
+    isPosemigroup : IsPosemigroup _≈_ _≤_ _∙_
+
+  open IsPosemigroup isPosemigroup public
+
+  pomagma : Pomagma c ℓ₁ ℓ₂
+  pomagma = record { isPomagma = isPomagma }
+
+  open Pomagma pomagma public using (preorder; poset; magma; promagma)
+
+  prosemigroup : Prosemigroup c ℓ₁ ℓ₂
+  prosemigroup = record { isProsemigroup = isProsemigroup }
+
+  open Prosemigroup prosemigroup public using (semigroup)
+
+-- Partially ordered monoids (pomonoids)
+
+record Pomonoid c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier    : Set c
+    _≈_        : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_        : Rel Carrier ℓ₂  -- The partial order.
+    _∙_        : Op₂ Carrier     -- The monoid multiplication.
+    ε          : Carrier         -- The monoid unit.
+    isPomonoid : IsPomonoid _≈_ _≤_ _∙_ ε
+
+  open IsPomonoid isPomonoid public
+
+  posemigroup : Posemigroup c ℓ₁ ℓ₂
+  posemigroup = record { isPosemigroup = isPosemigroup }
+
+  open Posemigroup posemigroup public using
+    ( preorder
+    ; poset
+    ; magma
+    ; promagma
+    ; pomagma
+    ; semigroup
+    ; prosemigroup
+    )
+
+  promonoid : Promonoid c ℓ₁ ℓ₂
+  promonoid = record { isPromonoid = isPromonoid }
+
+  open Promonoid promonoid public using (monoid)
+
+-- Partially ordered commutative monoids (commutative pomonoids)
+
+record CommutativePomonoid c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _∙_
+  field
+    Carrier               : Set c
+    _≈_                   : Rel Carrier ℓ₁  -- The underlying equality.
+    _≤_                   : Rel Carrier ℓ₂  -- The partial order.
+    _∙_                   : Op₂ Carrier     -- The monoid multiplication.
+    ε                     : Carrier         -- The monoid unit.
+    isCommutativePomonoid : IsCommutativePomonoid _≈_ _≤_ _∙_ ε
+
+  open IsCommutativePomonoid isCommutativePomonoid public
+
+  pomonoid : Pomonoid c ℓ₁ ℓ₂
+  pomonoid = record { isPomonoid = isPomonoid }
+
+  open Pomonoid pomonoid public using
+    ( preorder
+    ; poset
+    ; magma
+    ; promagma
+    ; pomagma
+    ; semigroup
+    ; prosemigroup
+    ; posemigroup
+    ; monoid
+    ; promonoid
+    )
+
+  commutativePromonoid : CommutativePromonoid c ℓ₁ ℓ₂
+  commutativePromonoid =
+    record { isCommutativePromonoid = isCommutativePromonoid }
+
+  open CommutativePromonoid commutativePromonoid public
+    using (commutativeSemigroup; commutativeMonoid)
+
+-- Partially ordered semirings (posemirings)
+
+record Posemiring c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _≤_
+  infixl 7 _*_
+  infixl 6 _+_
+  field
+    Carrier      : Set c
+    _≈_          : Rel Carrier ℓ₁
+    _≤_          : Rel Carrier ℓ₂
+    _+_          : Op₂ Carrier
+    _*_          : Op₂ Carrier
+    0#           : Carrier
+    1#           : Carrier
+    isPosemiring : IsPosemiring _≈_ _≤_ _+_ _*_ 0# 1#
+
+  open IsPosemiring isPosemiring public
+
+  +-commutativePomonoid : CommutativePomonoid c ℓ₁ ℓ₂
+  +-commutativePomonoid = record
+    { isCommutativePomonoid = +-isCommutativePomonoid }
+
+  open CommutativePomonoid +-commutativePomonoid public
+    using (preorder)
+    renaming
+    ( magma                to +-magma
+    ; promagma             to +-promagma
+    ; pomagma              to +-pomagma
+    ; semigroup            to +-semigroup
+    ; prosemigroup         to +-prosemigroup
+    ; posemigroup          to +-posemigroup
+    ; monoid               to +-monoid
+    ; promonoid            to +-promonoid
+    ; pomonoid             to +-pomonoid
+    ; commutativeSemigroup to +-commutativeSemigroup
+    ; commutativeMonoid    to +-commutativeMonoid
+    ; commutativePromonoid to +-commutativePromonoid
+    )
+
+  *-pomonoid : Pomonoid c ℓ₁ ℓ₂
+  *-pomonoid = record { isPomonoid = *-isPomonoid }
+
+  open Pomonoid *-pomonoid public
+    using ()
+    renaming
+    ( magma        to *-magma
+    ; promagma     to *-promagma
+    ; pomagma      to *-pomagma
+    ; semigroup    to *-semigroup
+    ; prosemigroup to *-prosemigroup
+    ; posemigroup  to *-posemigroup
+    ; monoid       to *-monoid
+    ; promonoid    to *-promonoid
+    )
+
+  prosemiring : Prosemiring c ℓ₁ ℓ₂
+  prosemiring = record { isProsemiring = isProsemiring }
+
+  open Prosemiring prosemiring public using (semiring)
--- a/src/Algebra/Ordered/Structures.agda
+++ b/src/Algebra/Ordered/Structures.agda
@@ -1,0 +1,323 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Ordered algebraic structures (not packed up with sets, orders,
+-- operations, etc.)
+------------------------------------------------------------------------
+
+-- The contents of this module should be accessed via
+-- `Algebra.Ordered`.
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Relation.Binary.Core using (Rel; _⇒_)
+
+module Algebra.Ordered.Structures
+  {a ℓ₁ ℓ₂} {A : Set a}  -- The underlying set
+  (_≈_ : Rel A ℓ₁)       -- The underlying equality relation
+  (_≤_ : Rel A ℓ₂)       -- The order
+  where
+
+open import Algebra.Core
+open import Algebra.Definitions _≈_
+open import Algebra.Structures _≈_
+open import Data.Product using (proj₁; proj₂)
+open import Function using (flip)
+open import Level using (_⊔_)
+open import Relation.Binary.Definitions using (Transitive; Monotonic₁; Monotonic₂)
+open import Relation.Binary.Structures using (IsPreorder; IsPartialOrder)
+open import Relation.Binary.Consequences using (mono₂⇒cong₂)
+
+------------------------------------------------------------------------
+-- Preordered structures
+
+-- Preordered magmas (promagmas)
+
+record IsPromagma (∙ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+  field
+    isPreorder  : IsPreorder _≈_ _≤_
+    ∙-cong      : Congruent₂ ∙
+    mono        : Monotonic₂ _≤_ _≤_ _≤_ ∙
+
+  open IsPreorder isPreorder public
+
+  mono₁ : ∀ {x} → Monotonic₁ _≤_ _≤_ (flip ∙ x)
+  mono₁ y≈z = mono y≈z refl
+
+  mono₂ : ∀ {x} → Monotonic₁ _≤_ _≤_ (∙ x)
+  mono₂ y≈z = mono refl y≈z
+
+  isMagma : IsMagma ∙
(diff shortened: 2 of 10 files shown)


sstucki:
(BTW, I also have an implementation of residuated pomonoids and lattices, but that will need some more refactoring before they it's ready to become a PR.)

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+open import Level using (_⊔_)
+open import Relation.Binary.Core using (Rel; _⇒_)
+open import Relation.Binary.Definitions using (Transitive)
+open import Relation.Binary.Structures using (IsPreorder; IsPartialOrder)
+
+------------------------------------------------------------------------
+-- Preordered structures
+
+module _ {a ℓ₁ ℓ₂} {A : Set a}
+  (_≈_ : Rel A ℓ₁)  -- The underlying equality.
```
MatthewDaggitt, line 29: Hmm so the equality operator should definitely be made a parameter of the top-level module so you can open it with all the equalities instantiated. I'm trying to figure out whether the order should also be lifted to the top-level as well, but I think it probably shouldn't be as it's more like the algebraic operators whose properties vary between structures.

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+  (_∼_ : Rel A ℓ₂)  -- The preorder.
+  where
+
+  -- Preordered semigroups (prosemigroups)
+
+  record IsProsemigroup (_∙_ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
```
MatthewDaggitt, line 35: I know nothing about this field of algebra so take this with a pinch of salt, but should there also be a `IsPromagma`?

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+module _ {a ℓ₁ ℓ₂} {A : Set a}
+  (_≈_ : Rel A ℓ₁)  -- The underlying equality.
+  (_≤_ : Rel A ℓ₂)  -- The partial.
+  where
+
+  -- Partially ordered semigroups (posemigroups)
+
+  record IsPosemigroup (_∙_ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      isPartialOrder : IsPartialOrder _≈_ _≤_
```
MatthewDaggitt, line 145: Hmm the lack of symmetry between this and `IsProsemigroup` is a bit off-putting. The latter uses the algebraic structure `IsSemigroup` as the base structure, whereas this uses the ordered structure `IsPartialOrder` as the base structure. Is there a good reason for this? If not I think it'd be better for them to be consistent!

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+      assoc          : Associative _≈_ _∙_
+      monotonic      : Monotonic₂ _≤_ _∙_
+
+    open IsPartialOrder isPartialOrder public
+
+    ∙-cong : Congruent₂ _≈_ _∙_
+    ∙-cong x≈y u≈v =
```
MatthewDaggitt, line 152: You can use this pre-existing proof :smile: 

https://github.com/agda/agda-stdlib/blob/650e05f3a5ef06cb938c3b91a4fc8b9b8d7c5ef2/src/Relation/Binary/Consequences.agda#L89-L91

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+      antisym (monotonic (reflexive x≈y) (reflexive u≈v))
+              (monotonic (reflexive (Eq.sym x≈y)) (reflexive (Eq.sym u≈v)))
+
+    isProsemigroup : IsProsemigroup _≈_ _≤_ _∙_
+    isProsemigroup = record
```
MatthewDaggitt, line 157: Case in point. If you fix the lack of symmetry, then you won't have to build this entire structure right?

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,205 @@
...
+    field
+      isPomonoid : IsPomonoid _∙_ ε
+      comm       : Commutative _≈_ _∙_
+
+    open IsPomonoid isPomonoid public
+
+    isCommutativePromonoid : IsCommutativePromonoid _≈_ _≤_ _∙_ ε
+    isCommutativePromonoid = record
+      { isPromonoid = isPromonoid
+      ; comm        = comm }
```
MatthewDaggitt, line 202: End bracket should go on the next line.

src/Relation/Binary/Lattice/Properties/BoundedJoinSemilattice.agda
```diff
@@ -51,3 +54,22 @@ dualBoundedMeetSemilattice = record
   { ⊤                        = ⊥
   ; isBoundedMeetSemilattice = dualIsBoundedMeetSemilattice
   }
+
+-- Every bounded semilattice gives rise to a commutative pomonoid
+
+isPomonoid : CommutativePomonoid c ℓ₁ ℓ₂
```
MatthewDaggitt, line 60: This should be called `commutativePomonoid`? Also the usual way to build these up is by defining the `is` structures one by one so that they are also available publicly in the file.

src/Relation/Binary/Lattice/Properties/BoundedMeetSemilattice.agda
```diff
@@ -14,6 +14,7 @@ module Relation.Binary.Lattice.Properties.BoundedMeetSemilattice
 open BoundedMeetSemilattice M
 
 open import Algebra.Definitions _≈_
+open import Algebra.Ordered.Bundles using (CommutativePomonoid)
```
MatthewDaggitt, line 17: Err this import doesn't seem to be used? Looks like you are missing adding something to this file?

src/Algebra/Ordered/Definitions.agda
```diff
@@ -0,0 +1,41 @@
...
+Monotonic₁ : Op₁ A → Set _
+Monotonic₁ f = f Preserves _≤_ ⟶ _≤_
+
+Antitonic₁ : Op₁ A → Set _
+Antitonic₁ f = f Preserves (flip _≤_) ⟶ _≤_
+
+Monotonic₂ : Op₂ A → Set _
+Monotonic₂ f = f Preserves₂ _≤_ ⟶ _≤_ ⟶ _≤_
+
+MonotonicAntitonic : Op₂ A → Set _
```
MatthewDaggitt, line 31: Out of interest are these standard names or something concocted?

MatthewDaggitt (commented):
Thanks for the PR! Looks pretty great, just a few comments. As I mentioned, I don't have much knowledge in this area of algebra so I'm trusting that the definitions are sensible. If anyone else has more knowledge, then another review would be welcome!

src/Algebra/Ordered/Definitions.agda
```diff
@@ -0,0 +1,41 @@
...
+  {a ℓ} {A : Set a}   -- The underlying set
+  (_≤_ : Rel A ℓ)     -- The underlying order
+  where
+
+open import Algebra.Core
+open import Function.Base using (flip)
+open import Data.Product using (_×_)
+
+Monotonic₁ : Op₁ A → Set _
+Monotonic₁ f = f Preserves _≤_ ⟶ _≤_
```
MatthewDaggitt, line 23: Hmm I'm now wondering if these should go in `Relation.Binary.Definitions` instead of here?

MatthewDaggitt:
> The PR is also missing a CHANGELOG patch, but I need some advice on what to put where there. (Is there a CHANGELOG guide somewhere?)

There isn't a guide I'm afraid. 
 - new modules should be listed under the "[New modules](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG.md#new-modules)" section
 - new additions to existing modules should be added in the appropriate place under the "[Other minor changes](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG.md#new-modules)" section.

src/Algebra/Ordered/Structures.agda
sstucki: You're right. I just had a look at `Relation.Binary.Structures` and there the equality is a module parameter while the orders are not. Then again, there are some structures there that don't have an order part. Maybe it would make sense to parametrize over everything here. One issue is that different notation are used in `Relation.Binary.Structures` for preorders (`_∼_`)  and partial orders (`_≤_`). This seems a purely cosmetic point, but I tried to be consistent here. If you're fine with just using `_≤_` everywhere, I think it would make sense to make that relation a module parameter too.

src/Algebra/Ordered/Structures.agda
sstucki: I'm not an expert either. Honestly, I hadn't heard of magmas before seeing them defined in the stdlib. 😅  As I said, this PR is derived from some old code I had, and in that project I just mechanized those structures I needed. But I'm happy to add promagmans and pomagmas if you think they are useful.

src/Algebra/Ordered/Structures.agda
sstucki: Yea, I see what you mean. The reason is that you get some structure for free once you assume anti-symmetry as well. Notably, you get `∙-cong` for free or really `cong` for any operation that is also monotone (as you noted yourself). There's a tradeoff here between adding redundant fields/assumptions to the records vs. keeping the record structure "regular". I personally prefer not to add redundant fields unless there are good technical reasons to do so (e.g. because it makes some proofs simpler). But I don't feel strongly about this, so I'm happy to discuss alternatives.

src/Algebra/Ordered/Structures.agda
sstucki: Will do, thanks for pointing it out!

src/Algebra/Ordered/Structures.agda
sstucki: Indeed, but see my earlier comment about redundant fields...

src/Relation/Binary/Lattice/Properties/BoundedJoinSemilattice.agda
sstucki: Yep, sorry, will fix.

src/Relation/Binary/Lattice/Properties/BoundedMeetSemilattice.agda
sstucki: Huh, probably redundant, yes. There was an extra proof earlier, but it was a "duplicated" solved via a re-export. Will double-check.

src/Algebra/Ordered/Definitions.agda
sstucki: Well monotonic and antitonic are standard — though I'm thinking of renaming them to "monotone" and "antitone", which are more common, I think. (I don't remember why I went with the *tonic variants in the first place.) The combination of mono+anti isn't "standard" but I don't know a better naming scheme. I'm open to suggestions. What is certain is that there are binary operations (like subtraction, division, residuation) that are antitone in one parameter and monotone in the other (or vice-versa).

src/Algebra/Ordered/Definitions.agda
sstucki: The reason I put them here is that they involve binary/unary operations from `Algebra`. There are analogous `Congruence` definitions in `Algebra.Definitions`:
https://github.com/agda/agda-stdlib/blob/e0e5db48582da275ce5f1fa747505795ee0b714c/src/Algebra/Definitions.agda#L27-L31

I figured that, if these live in the `Algebra` hierarchy (rather than `Relation.Binary`), then so should `Monotonic` and `Antitonic`.

sstucki:
Thanks for the review @MatthewDaggitt. I'll update the PR to address the obvious stuff, but I think some points warrant further discussion. In particular, I'm not an algebraist either. I mechanized this stuff a while back because I needed it for my experiments with graded (co)monads. Maybe @andreasabel, @JacquesCarette or @Taneb, who are part of the discussion in #1694 have some additional comments and ideas?

sstucki:
Oh, and @laMudri too!

Taneb:
I'll take a look this evening!

src/Algebra/Ordered/Structures.agda
sstucki: Oops, that didn't work. The proof of `mono⇒cong` is for unary operations/functions, but `_∙_` is a binary operation. I see three ways to fix this:
 1. Leave the proof as is.

src/Algebra/Ordered/Definitions.agda
sstucki: Update: it seems both `monotone` and `monotonic` are used in different parts of the library. Might make sense to be consistent and pick one or the other. For now, I'm leaving things as they are in the PR.

src/Relation/Binary/Lattice/Properties/BoundedJoinSemilattice.agda
sstucki: Fixed.

src/Algebra/Ordered/Bundles.agda
```diff
@@ -0,0 +1,223 @@
...
+
+-- Preordered semigroups (prosemigroups)
+
+record Prosemigroup c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix  4 _≈_ _∼_
+  infixl 7 _∙_
+  field
+    Carrier        : Set c
+    _≈_            : Rel Carrier ℓ₁  -- The underlying equality.
+    _∼_            : Rel Carrier ℓ₂  -- The preorder.
```
Taneb, line 33: Could we use a symbol more suggestive of ordering for the preorder relation?

Taneb (changes requested):
I'd like to see some proofs in `Data.Nat.Properties` giving examples of these structures, too, but that could be in a later PR

src/Algebra/Ordered/Bundles.agda
sstucki: Yea, I agree. I used `_∼_` to be consistent with the notation used in `Preorder` from `Relation.Binary.Bundles`, but I'd prefer `_≤_`.

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: Hmm I think best to play it safe and add them. If I've learnt anything from maintaining this library is that there's always someone out there who wants full generality.

sstucki:
Adding example properties about `Nat` (as @Taneb suggests) should be easy since all the ingredients are already there. @MatthewDaggitt, would you prefer having these in the current PR or a separate one?

Also: does anyone know if it's possible to create a PR based on another one without the changes from the first PR showing up in the first one? I like to split PRs into manageable chunks but it kinda defeats the point if everything just gets lumped together anyway...

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: Yup, go for it! Can't think of any compelling reason not to.

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: Yup, I see your point and this is exactly the reason we have `Algebra.Structures.Biased` which provide alternate, "easier" ways of constructing structures when you can derive some of the axioms from the others. So I'd recommend that we preserve the symmetry here, and then add this variant to `Algebra.Ordered.Structures.Biased`? (which is also re-exported from `Algebra.Ordered` for easy use).

src/Algebra/Ordered/Definitions.agda
MatthewDaggitt: Hmm so both `monotone` and `monotonic` are used pretty rarely in full. We do however use `mono` and `antimono` to stand for `monotonic` and `antimonotonic` a lot.

Trying to think how best to reconcile this with a consistent naming scheme for `MonotonicAntitonic` etc. ...

src/Algebra/Ordered/Definitions.agda
MatthewDaggitt: So actually `_Preserves _⟶ _` and friends are from the `Relation.Binary` hierarchy. I think the notion of monotonic and antitonic functions make sense for _any_ functions, rather than just those that are used in an algebraic setting right? For example `f x = x ::_` is monotonic over the lexicographic ordering over lists.

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: If we move the definitions to `Relation.Binary` then I would vote for proposal 2.  (with the proofs implemented via proposal 3.!)

MatthewDaggitt:
> Adding example properties about Nat (as @Taneb suggests) should be easy since all the ingredients are already there. @MatthewDaggitt, would you prefer having these in the current PR or a separate one?

Happy to have them added to this one. They're conceptually part of the same addition to the library, and it's always a good sanity check to actually use things after they've been defined to check there aren't any unforeseen wrinkles.

> Also: does anyone know if it's possible to create a PR based on another one without the changes from the first PR showing up in the first one? I like to split PRs into manageable chunks but it kinda defeats the point if everything just gets lumped together anyway...

Not easily. The only way I could think of for not having the changes in the 1st PR show up in the 2nd PR without having the merged the 1st one is making a PR against your original PR rather than the master branch. But I don't think you can do such a thing on this repo, only your own fork?

src/Algebra/Ordered/Structures.agda
sstucki: Alright. Two questions:
 1. Which structure do you prefer: using the algebraic structure as a base and adding what's missing from the orders, or vice-versa? (We could also add both at once, but then `isEquivalence` would be duplicated, which I really don't think is a good idea.)
 2. Is there maybe a better name than `Biased` for this? I don't really see in which sense either version is biased. But I also don't have any good alternative suggestions, TBH...

src/Algebra/Ordered/Definitions.agda
sstucki: That's true, but it's equally true for `Congruent`. I mean the only difference between `Preserves`, `Monotone` and `Congruent` is how they are intended to be used. Maybe adding all these anti/mono/whathever definitions is overkill and we should just use `Preserves` directly? (That's what I did originally, but then I figured I should probably mirror the definition of `Congruent`.)

src/Algebra/Ordered/Structures.agda
sstucki: Upon reflection, the answer to 1. is probably to go with preorders as the base since we're going to mirror the whole algebra hierarchy (starting with magmas) and build them up anyway. For the po* structures, I'll similarly start with partial orders.

src/Algebra/Ordered/Structures.agda
sstucki: Done.

src/Algebra/Ordered/Structures.agda
sstucki: OK, I completely reworked the definition of the various preordered structures using the preorder as a base, adding promagmas (and pomagmas), etc. Surprisingly, it looks like no biased versions are needed after all. 🤷

src/Algebra/Ordered/Structures.agda
sstucki: I've now added a single new proof `mono₂⇒cong₂` to `Relation.Binary.Consequences`. It's independent of `Congruence` and `Monotonic` but it does *not* use `mono⇒cong` as that would require an additional assumption (transitivity) which the direct proof does not.

sstucki:
The updated PR now contains example properties about `Nat` (as suggested by @Taneb). The corresponding definitions/proofs of the ordered structures and bundles replace the earlier (unordered) versions, which are now available as re-exported definitions from the ordered modules (to remain backwards-compatible). This is to avoid code duplication. If, for some reason, you prefer to keep the old hand-coded versions nevertheless, I can add them back.

src/Algebra/Ordered/Structures.agda
sstucki: Added promagmas and pomagmas now.

src/Algebra/Ordered/Bundles.agda
MatthewDaggitt: Nope!

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: You're going to hate me for this (sorry!), but wouldn't it be much cleaner to have the algebraic structures as the base and then add the ordering on top? The algebraic hierarchy is a mess, and the re-exports/re-naming as you've probably found are incredibly painful. Using the orders as the base structures means we have to replicate it in all its gory detail. Whereas using the algebra as the base structures means that we would get it for "free". The preorder/partial order hierarchy is pretty simple, so re-exporting everything should be much easier.

MatthewDaggitt:
Apologies, I didn't realise your changes to `Data.Nat.Properties` were going to involve anything but additions. Could we leave the `Data.Nat.Properties` changes to a separate PR?

MatthewDaggitt:
By the way thank you for your patience with this @sstucki, I realise it's a lot of refactoring. The design decisions are complicated, and definitely best to try and get them right now.

sstucki:
> Apologies, I didn't realise your changes to `Data.Nat.Properties` were going to involve anything but additions. Could we leave the `Data.Nat.Properties` changes to a separate PR?

OK, I'll factor this into a separate PR then.

src/Algebra/Ordered/Structures.agda
sstucki: Oh no! 😰

OK, let's think this through before I embark on another refactoring spree. 😅 

I actually don't think using the algebraic structures as a base is such a good idea. For the pre-ordered structures, it would just mean creating a number of records that would simply contain the base algebra plus the axioms of a preorder (using `IsPreorder` directly is not a good idea because it would duplicate the `IsEquivalence` filed which you really don't want, there shouldn't be two versions of that in the same record). But for the partially ordered version, this would mean a redundant `cong` field, which is not a deal breaker, but then we *would* need a separate collection of `Biased` modules that would just replicate what is in the PR now. I think that would actually decrease maintainability overall, no?

Again, I don't have strong feelings about this. If you think it's cleaner the other way round, @MatthewDaggitt, I'll put in the extra effort, but I'd rather we agreed first that's really what we want.

src/Algebra/Ordered/Definitions.agda
sstucki: You may have noticed I changed the `monotonic` fields to `mono` to be more consistent with `cong` and other parts of the library. But I'm still not sure what to do with the extra definitions (`Monotonic` and friends). I'm also fine with dropping those entirely and use `Respects` directly, if you think it's cleaner.

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,308 @@
...
+
+record IsPomagma (∙ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+  field
+    isPartialOrder : IsPartialOrder _≈_ _≤_
+    mono           : Monotonic₂ ∙
+
+  open IsPartialOrder isPartialOrder public
+
+  ∙-cong : Congruent₂ ∙
+  ∙-cong = mono₂⇒cong₂ _≈_ _≈_ Eq.sym reflexive antisym mono
```
laMudri, line 172: There's always an argument with these derived properties that they should appear in an auxiliary constructor rather than in the main record. In this case, I'm not really sure; I think we often have a proof of `∙-cong` lying around anyway, but maybe we don't actually gain anything from putting it in because we never really have to look at these proof terms. This could come up if we want to take the opposite ordering and have that be a definitional involution, but maybe because `mono` should stay the same, `∙-cong` does too.

src/Algebra/Ordered/Bundles.agda
```diff
@@ -0,0 +1,346 @@
...
+open import Level using (suc; _⊔_)
+open import Relation.Binary using (Rel)
+open import Relation.Binary.Bundles using (Preorder; Poset)
+
+------------------------------------------------------------------------
+-- Bundles of preordered structures
+
+-- Preordered magmas (promagmas)
+
+record Promagma c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
```
laMudri, line 27: I'm mildly worried about the “pro” prefix because I've seen “proset” used to mean a partial order where we don't have/never talk about equality. I don't know whether this is standard, but it's very relevant to the setoid setting.

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,308 @@
...
+
+------------------------------------------------------------------------
+-- Partially ordered structures
+
+-- Partially ordered magmas (pomagmas)
+
+record IsPomagma (∙ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+  field
+    isPartialOrder : IsPartialOrder _≈_ _≤_
+    mono           : Monotonic₂ ∙
```
laMudri, line 167: Maybe this should be `∙-mono` to fit with `∙-cong`, but I can see the argument both ways.

src/Algebra/Ordered/Structures.agda
```diff
@@ -0,0 +1,308 @@
...
+------------------------------------------------------------------------
+-- Preordered structures
+
+-- Preordered magmas (promagmas)
+
+record IsPromagma (∙ : Op₂ A) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
+  field
+    isPreorder  : IsPreorder _≈_ _≤_
+    ∙-cong      : Congruent₂ ∙
+    mono        : Monotonic₂ ∙
```
laMudri, line 40: It'd be nice to have `monoˡ` and `monoʳ`, too, like for `cong`.

laMudri (commented):
Looks good! In the long run, I think we need a better approach to generating these definitions, but this is probably the best we can do for now.

src/Algebra/Ordered/Structures.agda
sstucki: Yes, I was thinking along the same lines. @MatthewDaggitt actually suggested a different design where the algebraic structures (rather than the orders) are used as a base, and then there's a better case for keeping `cong` and adding the type of constructors you mention (see our discussion further up in the thread). I'd be curious to hear which design you'd prefer.

src/Algebra/Ordered/Bundles.agda
sstucki: Hum, indeed. I thought I had seen at least promonoid somewhere, but I might be fooling myself. Staying consistent with the literature is important though, so I'm open to alternative suggestions. Then there's the issue of keeping names short 😅

src/Algebra/Ordered/Structures.agda
sstucki: Yes, I went back and forth on that one as well but settled for the shorter variant as the extra prefix doesn't really add any info when there's just one operation.

src/Algebra/Ordered/Structures.agda
sstucki: Good point, will add those! 👍

sstucki:
Thanks for your input @Taneb and @laMudri!

src/Algebra/Ordered/Structures.agda
sstucki: Uh, now I'm confused. The type of `congˡ` is `LeftCongruent` which actually states that the operation is congruent in its second, i.e. **right** argument. And similarly for `RightCongruent`. Is that a bug? Or some special meaning of "left" and "right" that I'm missing?
https://github.com/agda/agda-stdlib/blob/650e05f3a5ef06cb938c3b91a4fc8b9b8d7c5ef2/src/Algebra/Definitions.agda#L33-L37

Also, I'm not a huge fan of left/right because it doesn't generalize to *n*-ary operations. How about `mono₁` and `mono₂`?

src/Algebra/Ordered/Bundles.agda
sstucki: Fixed.

src/Algebra/Ordered/Structures.agda
laMudri: Hmm, yeah, I'd have thought `congˡ` had type `∀ {x x′ y} → x ≈ x′ → x ∙ y ≈ x′ ∙ y` too. I *can* justify the current naming as saying that *left*-multiplication (by `x`) preserves equivalence, but when I'm using it in a proof, I don't think I'd think that way. IIRC, there's a similar problem with distributivity or zero, where one is the reverse of the other (something like `LeftZero` means `0 * x ≈ x` but `LeftDistributive` means `x * (y + z) ≈ x * y + x * z`, or vice versa).

As for numbers, I guess it side-steps any existing naming conventions, but introduces the problem of needing to choose between being 0-based or 1-based. I think (3+n)-ary operations are very rare in algebra, so I'm not too worried about that potential generalisation.

src/Algebra/Ordered/Definitions.agda
MatthewDaggitt: Changing the fields to `mono` :+1: Let's keep the names for now. The more complicated ones don't appear in any record field names yet, so it can all be sorted out by a bit of judicious deprecation if we need to later.

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: Yup, let's stick with no prefix. I still hope to convince people to get rid of the `∙-` in `∙-cong` some day.

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: > using IsPreorder directly is not a good idea because it would duplicate the IsEquivalence field which you really don't want, there shouldn't be two versions of that in the same record

Agreed!

Hmm so I just wrote 3 paragraphs about why it was better with the algebra hierarchy as a base, but deleted it as I can't distinguish in my head whether I think that because the current algebra hierarchy is so hopelessly broken or whether its actually better that I way. In the end, I've decided at this moment I also don't have the time to work through all the implications (sorry future me!). Let's just keep it the way you've got it now :+1:

src/Algebra/Ordered/Definitions.agda
MatthewDaggitt: That's true about `Congruent`. Perhaps having it in `Algebra.Definitions` instead of `Relation.Binary.Definitions` is a mistake. Nonetheless, I would really like these new monotonic definitions to live in the latter. 

Another argument is that if we ever want to add a new ordered definition that used both the order and the equality relation, then this module's current module-level arguments wouldn't allow that.

src/Algebra/Ordered/Definitions.agda
sstucki: SGTM.

src/Algebra/Ordered/Structures.agda
sstucki: Yea, I can totally see that there are good arguments for and against here. It's hard to know which will work out better in the end.

Out of curiosity: when you say the algebra hierarchy is broken, do you think there would be a reasonably clean fix, or is it just an impossible problem to design hierarchies like this in Agda? I mean, it does feel like we're just implementing OO/multiple inheritance manually here, but maybe my judgment is clouded from having worked in OO languages too much in the past. In any case, I'm curious to hear if you have any ideas on how things could be improved in general. Is there some discussion around this topic somewhere? E.g. in an issue?

src/Algebra/Ordered/Definitions.agda
sstucki: OK, so just to double check before making the changes:

 1. I'll move **all** the definitions from `Algebra.Definitions.Ordered` to `Relation.Binary.Definitions` then. 
 2. I'll drop the use of `Op` from `Algebra.Definitions` (to make the definition generic, and avoid cycles in the module imports).
 3. I'll keep `{Mono,Andt}tonic` even though they will be essentially aliases of `Preserves`.

Agreed?

src/Algebra/Ordered/Structures.agda
sstucki: OK, I can see that someone might have had good reasons to name `{Left,Right}Congruent` the way they did and I don't want to break backwards compatibility, so I'll not "fix" this unless @MatthewDaggitt tells me to. As for the `mono` variants, I'll go with `mono₁` and `mono₂` instead of for `monoˡ` and `monoʳ` for now, mostly because I don't know whether to follow that strange backwards terminology or not. (Again, unless @MatthewDaggitt tells me otherwise.)

sstucki:
> By the way thank you for your patience with this @sstucki, I realise it's a lot of refactoring.

Well, thank you for *your* patience @MatthewDaggitt, and for all your hard work on the Agda stdlib and for the community in general — much appreciated!

src/Algebra/Ordered/Structures.agda
MatthewDaggitt: > do you think there would be a reasonably clean fix, or is it just an impossible problem to design hierarchies like this in Agda?

I think the consensus is that it's the latter. There's no good way to define diamonds compositionally without biasing them in some regards that impacts usability. I hear that Arend's implementation of record types is supposed to have gone a long way to fixing the problem.

In no particular order here are a selection of the issues discussing problems:
- https://github.com/agda/agda-stdlib/issues/239
- https://github.com/agda/agda-stdlib/issues/455
- https://github.com/agda/agda-stdlib/issues/876
- https://github.com/agda/agda-stdlib/issues/1238
- https://github.com/agda/agda-stdlib/issues/1617

sstucki:
@MatthewDaggitt and @Taneb: the changes to `Data.Nat.Properties` are now living in a separate branch: https://github.com/sstucki/agda-stdlib/tree/nat-ordered
I'll submit the corresponding PR once this one has been merged.

src/Algebra/Ordered/Structures.agda
sstucki: Thanks for these pointers. I'll bookmark #1238, which contains a lot of references to relevant related work. (Thanks @JacquesCarette!) If it's not already the case, it might be nice to summarize these discussions in some sort of README.

sstucki:
I think I have now addressed all the open requests for changes. I hope I didn't miss anything, as there have been quite a few discussions going on in parallel. 😅  (Thanks to all  involved!)

sstucki:
@MatthewDaggitt, is there anything more you need me to do here?

MatthewDaggitt:
Sorry for the delay. Merging in!

Thanks again for the great PR @sstucki!

MatthewDaggitt:
I should add that if you are planning to make the changes to `Data.Nat.Properties` in a separate PR, can we make the changes add the new structures only rather than refactor everything to define the new structures first and then extract the old structures later? I know it would be less code, but I think we would lose a significant amount of readability and maintainability.

sstucki:
> Sorry for the delay. Merging in!

Thanks for all the help @MatthewDaggitt!

> I should add that if you are planning to make the changes to `Data.Nat.Properties` in a separate PR, can we make the changes add the new structures only rather than refactor everything to define the new structures first and then extract the old structures later? I know it would be less code, but I think we would lose a significant amount of readability and maintainability.

I have already prepared the changes for the PR in a separate branch (just separated them out from what was in this PR earlier). I'm afraid, it does precisely what you ask me not to do. 😓  If it's OK with you, I'll submit the PR as is anyway, and if the readability/maintainability is indeed worse, I'll refactor it to keep as many of the old proofs as possible. Does that sound OK?

src/Algebra/Ordered/Structures.agda
JacquesCarette: No problem. [Sorry for the late comment, I am just now catching up to a couple of months' worth of agda-stdlib activity!]

PR #1754, opened 2022-04-20
Add some very simple properties about algebraic structures.

sstucki:
Concretely: add proofs that the basic structure of magmas, semigroups and monoids are preserved if we switch the arguments of their binary operation.

### DIFF
--- a/src/Algebra/Construct/Flip/Op.agda
+++ b/src/Algebra/Construct/Flip/Op.agda
@@ -1,0 +1,241 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Flipping the arguments of a binary operation preserves many of its
+-- algebraic properties.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.Flip.Op where
+
+open import Algebra
+import Data.Product as Prod
+import Data.Sum as Sum
+open import Function.Base using (flip)
+open import Level using (Level)
+open import Relation.Binary.Core using (Rel; _Preserves₂_⟶_⟶_)
+open import Relation.Binary.Definitions using (Symmetric)
+
+private
+  variable
+    a ℓ : Level
+    A   : Set a
+    ≈   : Rel A ℓ
+    ε   : A
+    ⁻¹  : Op₁ A
+    ∙   : Op₂ A
+
+------------------------------------------------------------------------
+-- Properties
+
+preserves₂ : (∼ ≈ ≋ : Rel A ℓ) →
+             ∙ Preserves₂ ∼ ⟶ ≈ ⟶ ≋ → (flip ∙) Preserves₂ ≈ ⟶ ∼ ⟶ ≋
+preserves₂ _ _ _ pres = flip pres
+
+module _ (≈ : Rel A ℓ) (∙ : Op₂ A) where
+
+  associative : Symmetric ≈ → Associative ≈ ∙ → Associative ≈ (flip ∙)
+  associative sym assoc x y z = sym (assoc z y x)
+
+  identity : Identity ≈ ε ∙ → Identity ≈ ε (flip ∙)
+  identity id = Prod.swap id
+
+  commutative : Commutative ≈ ∙ → Commutative ≈ (flip ∙)
+  commutative comm = flip comm
+
+  selective : Selective ≈ ∙ → Selective ≈ (flip ∙)
+  selective sel x y = Sum.swap (sel y x)
+
+  idempotent : Idempotent ≈ ∙ → Idempotent ≈ (flip ∙)
+  idempotent idem = idem
+
+  inverse : Inverse ≈ ε ⁻¹ ∙ → Inverse ≈ ε ⁻¹ (flip ∙)
+  inverse inv = Prod.swap inv
+
+------------------------------------------------------------------------
+-- Structures
+
+module _ {≈ : Rel A ℓ} {∙ : Op₂ A} where
+
+  isMagma : IsMagma ≈ ∙ → IsMagma ≈ (flip ∙)
+  isMagma m = record
+    { isEquivalence = isEquivalence
+    ; ∙-cong        = preserves₂ ≈ ≈ ≈ ∙-cong
+    }
+    where open IsMagma m
+
+  isSelectiveMagma : IsSelectiveMagma ≈ ∙ → IsSelectiveMagma ≈ (flip ∙)
+  isSelectiveMagma m = record
+    { isMagma = isMagma m.isMagma
+    ; sel     = selective ≈ ∙ m.sel
+    }
+    where module m = IsSelectiveMagma m
+
+  isCommutativeMagma : IsCommutativeMagma ≈ ∙ → IsCommutativeMagma ≈ (flip ∙)
+  isCommutativeMagma m = record
+    { isMagma = isMagma m.isMagma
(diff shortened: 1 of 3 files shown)


gallais:
Should this go in [`[Relation.Binary.Construct.Flip](http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Flip.html)`](http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Flip.html)?

sstucki:
> Should this go in [`Relation.Binary.Construct.Flip`](http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Flip.html)?

I don't think so because we're flipping the binary operation, not the associated equivalence relation. If anything, it should go into a new module `Algebra.Construct.Flip`, but I figured `Algebra.Consequences.Base` would do for now. 🤷

MatthewDaggitt:
Okay so this should all definitely go in an `Algebra.Construct` module. The question is what under what name.

I think it should be `Algebra.Construct.Converse` not `Algebra.Construct.Flip` to match the semantics of `Relation.Binary.Construct.Converse/Flip`. The former does not flip the underlying equality relation, whereas the latter does. This PR does not flip the equality and therefore `Converse` it is.

MatthewDaggitt:
See https://github.com/agda/agda-stdlib/pull/1751 for how such a module should look like.

sstucki:
> I think it should be `Algebra.Construct.Converse` not `Algebra.Construct.Flip` to match the semantics of `Relation.Binary.Construct.Converse/Flip`. The former does not flip the underlying equality relation, whereas the latter does. This PR does not flip the equality and therefore `Converse` it is.

But `Converse` doesn't make sense if `flip` is applied to a binary operation. When you exchange the arguments of an relation, you get it's converse, but when you exchange the arguments of a binary operation you don't get it's converse, you just get another operation. `Flip` makes a lot more sense IMO.

MatthewDaggitt:
Okay so what are we going to call the module that flips both the binary operation and the underlying equality?

Taneb:
Do we ever want to flip the underlying equality? We know it's symmetric

MatthewDaggitt:
> Do we ever want to flip the underlying equality? We know it's symmetric

So `Binary.Construct.Flip` predated my involvement of the library, so it was needed "first" but I'm unsure what for. I added `Converse` later, because `Flip` was deeply inconvenient. In general though, I've learnt that if something seems a sensible operation then someone will need it somewhere even if I can't come up with a use case immediately. 

> If it's ever necessary to flip the equality in the Algebra hierarchy, we could have submodules Flip.{Op,Eq,OpEq} at least that would make it clear what's going on.

This seems like an excellent suggestion to me. So this module would be called `Algebra.Construct.Flip.Op`? We should be consistent though and refactor the `Relation.Binary.Construct` modules to follow the same pattern. Let's not do it in this PR though.

sstucki:
> This seems like an excellent suggestion to me. So this module would be called `Algebra.Construct.Flip.Op`? We should be consistent though and refactor the `Relation.Binary.Construct` modules to follow the same pattern. Let's not do it in this PR though.

OK, I'll create that module and move the proofs there then. But to be clear: that module will contain the "generic" proofs, the derived proofs involving bundles of algebraic structures should still go into separate modules under `Algebra.Properties`, right?

MatthewDaggitt:
I'm not quite sure what you mean by "generic" proofs. All the proofs that you've added should live in the new module. If you want to (no requirement to though) you could also keep the statements in the places you've got currently, but use the new module to prove them. Does that make sense?

sstucki:
Oh, I see now in [`Relation.Binary.Construct.Flip`](http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Flip.html) that there is a "Structure" and "Bundles" section that implements the per-structure and per-bundle properties. Then the separate modules in `Properties` are no longer needed. 👍

sstucki:
OK, version 2 ready for review. Now everything is in a single module called `Algebra.Construct.Flip.Op` as discussed previously. I threw in proofs for some additional structures for good measure (everything with a single binary operation).

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -0,0 +1,241 @@
...
+  idempotent : Idempotent ≈ ∙ → Idempotent ≈ (flip ∙)
+  idempotent idem = idem
+
+  inverse : Inverse ≈ ε ⁻¹ ∙ → Inverse ≈ ε ⁻¹ (flip ∙)
+  inverse inv = swap inv
+
+------------------------------------------------------------------------
+-- Structures
+
+module _ {≈ : Rel A ℓ} {∙ : Op₂ A} where
```
MatthewDaggitt, line 59: With the variables, we don't need this anonymous module right?

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -0,0 +1,241 @@
...
+open import Function.Base using (flip)
+open import Level using (Level)
+open import Relation.Binary.Core using (Rel; _Preserves₂_⟶_⟶_)
+open import Relation.Binary.Definitions using (Symmetric)
+
+private
+  variable
+    a b p ℓ ℓ₁ ℓ₂ : Level
+    A B : Set a
+    ∼ ≈ ≋ : Rel A ℓ
```
MatthewDaggitt, line 24: `∼` and `≋` are pretty unfamiliar symbols to use in the algebra hierarchy, so I'd prefer if they weren't made variables. Indeed they're only used in the `preserves₂` proof, so perhaps can we just make them implicit parameters of that proof instead?

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -0,0 +1,241 @@
...
+-- Flipping the arguments of a binary operation preserves many of its
+-- algebraic properties.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Construct.Flip.Op where
+
+open import Algebra
+open import Data.Product using (swap)
```
MatthewDaggitt, line 13: Can we be consistent between `Product` and `Sum` and import them both qualified? i.e. `import Data.Product as Prod`

src/Algebra/Construct/Flip/Op.agda
sstucki: The variables help with the occurrences of `≈` and `∙` in signatures, but not in the actual definitions (where they have to be passed explicitly, otherwise the unifier is not happy). For example, the line
https://github.com/agda/agda-stdlib/blob/a123013ec41c8341108058764e6c9eaa3d136697/src/Algebra/Construct/Flip/Op.agda#L69
would have to be changed to
```
isSelectiveMagma {≈ = ≈} {∙ = ∙} m = record
```
and likewise for all the other `isXYZStructure` definitions. I find the latter less readable and prefer the module.

src/Algebra/Construct/Flip/Op.agda
sstucki: Yup, you're right. There was some other garbage in that variable declaration too, sorry. I cleaned it up now.

src/Algebra/Construct/Flip/Op.agda
sstucki: Fixed.

src/Algebra/Construct/Flip/Op.agda
MatthewDaggitt: Good point. Agreed!

PR #1756, opened 2022-04-22
[ add ] some properties connecting ∈ and ++ inherited from Any

laMudri:
These are some properties I need for https://github.com/agda/agda-categories/pull/290 (plus a few more related properties from the same section that I thought I'd add). This PR is entirely boilerplate, so maybe this is not the best or intended way of doing things.


MatthewDaggitt:
I don't know of any better way of doing it so no comments from me! Oh apart from can you add a CHANGELOG entry?

PR #1758, opened 2022-04-30
[ doc ] update installation instructions to 1.7.1

laMudri:
I noticed that the installation instructions still refer to version 1.7, so I updated the versions mentioned to what I think the current ones are.

### DIFF
--- a/notes/installation-guide.md
+++ b/notes/installation-guide.md
@@ -3,7 +3,7 @@ Installation instructions
 
 Note: the full story on installing Agda libraries can be found at [readthedocs](http://agda.readthedocs.io/en/latest/tools/package-system.html).
 
-Use version v1.7.1 of the standard library with Agda 2.6.2.1 or 2.6.2.
+Use version v1.7.1 of the standard library with Agda 2.6.2.2, 2.6.2.1, or 2.6.2.
 
 1. Navigate to a suitable directory `$HERE` (replace appropriately) where
    you would like to install the library.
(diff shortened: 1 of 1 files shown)


notes/installation-guide.md
```diff
@@ -3,19 +3,19 @@ Installation instructions
 
 Note: the full story on installing Agda libraries can be found at [readthedocs](http://agda.readthedocs.io/en/latest/tools/package-system.html).
 
-Use version v1.7 of the standard library with Agda 2.6.2.
+Use version v1.7.1 of the standard library with Agda 2.6.2.1 or 2.6.2.
```
andreasabel, line 6: While you are at it, the lastest release of Agda is 2.6.2.2.

andreasabel (changes requested):
I think switching the docs to v1.7.1 is good, even though the only difference (afaik) is that v1.7.1 builds with GHC 9.2 while v1.7 doesn't.

Change request: if you mention 2.6.2.1, please also mention 2.6.2.2.

notes/installation-guide.md
laMudri: Does https://wiki.portal.chalmers.se/agda/pmwiki.php?n=Libraries.StandardLibrary also need to be updated to mention 2.6.2.2?

andreasabel: Indeed.  Just did that.

andreasabel (approved):

MatthewDaggitt:
Thanks!

PR #1761, opened 2022-05-03
[ fix #1354 ] Refactoring Permutation.Propositional

### LINKED ISSUE #1354: Reimplement `Permutation.Propositional` in terms of `Permutation.Setoid`
Unlike other binary relations over lists, `Data.List.Relation.Binary.Permutation.(Setoid/Propositional)` are implemented separately which results in an awful lot of proof duplication. It also means that other modules (e.g. `Subset`) don't play nice when trying to transfer results from setoid equality to propositional equality. I've added the breaking tag as it remains to be seen if they can be unified in a backwards compatible manner.

gallais:
Still missing: CHANGELOG entries

### DIFF
--- a/src/Data/List/Relation/Binary/BagAndSetEquality.agda
+++ b/src/Data/List/Relation/Binary/BagAndSetEquality.agda
@@ -23,6 +23,7 @@ open import Data.List.Membership.Propositional.Properties
 open import Data.List.Relation.Binary.Subset.Propositional.Properties
   using (⊆-preorder)
 open import Data.List.Relation.Binary.Permutation.Propositional
+  using (_↭_; ↭-refl; ↭-sym; module PermutationReasoning)
 open import Data.List.Relation.Binary.Permutation.Propositional.Properties
 open import Data.Product.Base as Prod hiding (map)
 import Data.Product.Function.Dependent.Propositional as Σ
@@ -53,6 +54,7 @@ private
     A B : Set a
     x y : A
     ws xs ys zs : List A
+
 
 ------------------------------------------------------------------------
 -- Definitions
@@ -142,8 +144,8 @@ module _ {k} {f g : A → B} {xs ys} where
     helper y = mk↔ₛ′
       (λ x≡fy → ≡.trans x≡fy (        f≗g y))
       (λ x≡gy → ≡.trans x≡gy (≡.sym $ f≗g y))
-      (λ { ≡.refl → ≡.trans-symˡ (f≗g y) })
-      λ { ≡.refl → ≡.trans-symʳ (f≗g y) }
+      (λ { refl → ≡.trans-symˡ (f≗g y) })
+      λ { refl → ≡.trans-symʳ (f≗g y) }
 
 ------------------------------------------------------------------------
 -- _++_
@@ -282,7 +284,6 @@ empty-unique {xs = _ ∷ _} ∷∼[] with ⇒→ ∷∼[] (here refl)
                                                 MP.right-distributive xs₁ xs₂ (pure ∘ f)) ⟩
   (fs >>= λ f → (xs₁ >>= pure ∘ f) ++
                 (xs₂ >>= pure ∘ f))       ≈⟨ >>=-left-distributive fs ⟩
-
   (fs >>= λ f → xs₁ >>= pure ∘ f) ++
   (fs >>= λ f → xs₂ >>= pure ∘ f)         ≡⟨ ≡.cong₂ _++_ (Applicative.unfold-⊛ fs xs₁) (Applicative.unfold-⊛ fs xs₂) ⟨
 
@@ -296,7 +297,8 @@ private
 
   ¬-drop-cons : ∀ {x : A} →
     ¬ (∀ {xs ys} → x ∷ xs ∼[ set ] x ∷ ys → xs ∼[ set ] ys)
-  ¬-drop-cons {x = x} drop-cons with Equivalence.to x∼[] (here refl)
+  ¬-drop-cons {x = x} drop-cons
+    with Equivalence.to x∼[] (here refl)
     where
     x,x≈x :  (x ∷ x ∷ []) ∼[ set ] [ x ]
     x,x≈x = ++-idempotent [ x ]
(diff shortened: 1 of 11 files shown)


src/Data/List/Relation/Binary/BagAndSetEquality.agda
```diff
@@ -115,7 +115,7 @@ module ⊆-Reasoning where
 module _ {a k} {A : Set a} {x y : A} {xs ys} where
 
   ∷-cong : x ≡ y → xs ∼[ k ] ys → x ∷ xs ∼[ k ] y ∷ ys
-  ∷-cong refl xs≈ys {y} =
+  ∷-cong P.refl xs≈ys {y} =
```
MatthewDaggitt, line 118: Eugh the newly introduced clash with `refl` from `PropositionalEquality` is an unpleasant long-term problem and from a user perspective, we've managed to make the interface worse.

Perhaps we should consider prepending all the constructors with `↭-` to avoid this. A breaking change, but it does avoid this problem. The one wrinkle with that is that looking `↭-trans` in `Permutation.Propositional` it's actually doing something different from constructor `trans`. So we'd have to rename that as well...

Alternatively, as this isn't really going to bring the user any benefit, maybe we should simply prove an equivalence statement between `Setoid` instantiated with the propositional setoid and the `Propositional` definition and transfer the proofs that way?

Thoughts?

gallais: > Thoughts?

I'm in general against adding prefixes (except for things like bundles where there are multiple
reasonable constructs with the name) because people can achieve the same result by importing
the module qualified.

MatthewDaggitt: > I'm in general against adding prefixes (except for things like bundles where there are multiple
reasonable constructs with the name) because people can achieve the same result by importing
the module qualified.

Fair enough. It's a good point.

> I can swap the import strategy from qualifying PropositionalEquality's refl to
qualifiying Permutation's instead if you want.

I don't think that would change the fact that we're degrading the user experience to make our own life easier as library developers?

MatthewDaggitt:
Closing as we've decided that its better to wait for typed pattern synonyms to finally make an appearance.

src/Data/List/Relation/Binary/BagAndSetEquality.agda
jamesmckinna: See #2321 .
For whatever reason, Agda is (now!) happy for the `refl` of each of `PropositionalEquality` and `Permutation.Propositional` to coexist happily. Moreover the refactoring uncouples the dependency on inductive proofs on the latter relation.

jamesmckinna:
Suggest re-opening this, as I seem to have got it to work with the pattern synonyms... using some ideas from #2317 and #2321 towards further unification, if someone can help me unify the two `Permutation.*.Properties` modules along the lines considered in #2317 ... incoming revised/streamlined/refactored/rebased commits shortly.

src/Data/List/Relation/Binary/Permutation/Propositional.agda
```diff
@@ -10,41 +10,38 @@ module Data.List.Relation.Binary.Permutation.Propositional
   {a} {A : Set a} where
 
 open import Data.List.Base using (List; []; _∷_)
+import Data.List.Relation.Binary.Permutation.Homogeneous as Homogeneous
 open import Relation.Binary
-open import Relation.Binary.PropositionalEquality using (_≡_; refl)
+open import Relation.Binary.PropositionalEquality as Eq using (_≡_)
```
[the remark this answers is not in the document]
jamesmckinna: Resolved now.

src/Data/List/Relation/Binary/BagAndSetEquality.agda
jamesmckinna: Resolved as far as this module is concerned.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -356,20 +356,31 @@ module _ {ℓ} {R : Rel A ℓ} (R? : Decidable R) where
...
     where open PermutationReasoning
 
+------------------------------------------------------------------------
+-- foldr of Commutative Monoid
+
+module _ (commutativeMonoid : CommutativeMonoid a r) where
+
+  open module CM = CommutativeMonoid commutativeMonoid
+
+  foldr-commMonoid : ∀ {xs ys} → xs ↭ ys → List.foldr _∙_ ε xs ≈ List.foldr _∙_ ε ys
```
JacquesCarette, line 370: Shouldn't this be more like foldr respects permutation? From the name, I was expecting a definition of foldr for a commutative monoid.

JacquesCarette (commented):
Other than that one odd name (which seems to actually be something pre-existing), this is looking quite good.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
jamesmckinna: *Very* happy to have a better name for this: I had been planning to deprecate the corresponding one under `Setoid.Properties` because it takes an `IsCommutativeMonoid` instance, and I think such things should take *bundles* not *structures*.

jamesmckinna:
So far all the heavy lifting was done years ago by @gallais so props to him. 

But I think that my recent work on #2317 hopefully points the way to getting the `Properties` unified as well. If that succeeds, then most of the guff in `Propositional` disappears entirely, and with it, all/most of the angst about qualified import of `refl`... I hope. Fingers crossed!

jamesmckinna:
Removing the milestone marker from this for the time being...

JacquesCarette:
So what needs to be done to push this to the finish line? [Other than a likely rather nasty merge!]

JacquesCarette:
I might have a student who could push this to the finishing line, if it was clear what needed to be done - @jamesmckinna ?

jamesmckinna:
> I might have a student who could push this to the finishing line, if it was clear what needed to be done - @jamesmckinna ?

See my comment on #2317

MatthewDaggitt:
My understanding of the argument was that we wanted to eventually deprecate `Permutation` in favour of the two new definitions. Which would mean re-proving all the existing results for both variants. This probably would be a good student project.

jamesmckinna:
> My understanding of the argument was that we wanted to eventually deprecate `Permutation` in favour of the two new definitions. Which would mean re-proving all the existing results for both variants. This probably would be a good student project.

Happy to consult/supervise/review if there is a willing candidate. There may well be more we can usefully do, esp. eg. wrt names of lemmas, as well as the overall structure. The factorisation out of `Homogeneous` properties may still be useful, but less so now than before? I suspect any end result will also end up being v3.0 because `breaking`...

MatthewDaggitt:
> I suspect any end result will also end up being v3.0 because breaking...

Yes, the big switch over will be, but we extend the API for the two new modules in a non-breaking way in 2.4.

JacquesCarette:
@gabriellisboaconegero this might be a more straightforward project than some you are on right now, i.e. reproving many properties of permutations given using different definitions, than some of the things you're currently working on.

jamesmckinna:
We discussed this PR 2026-07-01 and agreed to close, with transfer of ideas to #2317.

PR #1762, opened 2022-05-04
List.Permutation: simplify ++-comm proofs

omelkonian:
@gallais Will these be unified after the change in #1761? If so, I could base my PR onto yours if you wish.


gallais:
For now they won't be. We'd have to (generalise!) & move proofs to `Homogeneous.Properties`
to be able to resume them.

PR #1764, opened 2022-05-05
Try to improve reduction behaviour of rationals using Andreas's suggestion

### LINKED ISSUE #1753: Disastrous expansion of operations in `Data.Rational.Base`
The operations such as addition and multiplication in `Data.Rational.Base` are defined in terms of the numerator and denominator projection operations, e.g.

https://github.com/agda/agda-stdlib/blob/650e05f3a5ef06cb938c3b91a4fc8b9b8d7c5ef2/src/Data/Rational/Base.agda#L206-L207

This is beautiful and elegant, and arguably the right way to define them. However, it does mean that Agda unfolds these definitions automatically. By the time you have an expression `x + y + z + a + b + c` the unfolded expression is over 200 lines long. 

This makes the rational numbers almost impossible to work with in large proofs. In [recent proofs](https://github.com/vehicle-lang/vehicle/blob/dev/examples/windController/agdaProof/AbstractRationals.agda) I've resorted to redefining everything that's needed under the `abstract` keyword.

MatthewDaggitt:
Closes #1753. Haven't addressed the problem with integer multiplication.

### DIFF
--- a/src/Data/Rational/Base.agda
+++ b/src/Data/Rational/Base.agda
@@ -30,6 +30,12 @@ open import Relation.Binary.PropositionalEquality.Core
 -- way to represent every rational number.
 
 record ℚ : Set where
+  -- We add "no-eta-equality; pattern" to the record to stop Agda
+  -- automatically unfolding rationals when arithmetic operations are
+  -- applied to them (see definition of operators below and Issue #1753
+  -- for details).
+  no-eta-equality; pattern
+
   constructor mkℚ
   field
     numerator     : ℤ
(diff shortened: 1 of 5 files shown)


JacquesCarette:
Very nice. I'm a little surprised at just how many times you had to force evaluation (in proofs - all the ones for defining functions makes sense).

I'm guessing none are actually intrinsic but are because of the choice of method of proof? Mostly curious, I don't think it's worth doing anything about it in this PR. But might be fun over time to slowly make these go away, to further decrease coupling from the proofs to the exact implementation.

src/Data/Rational/Base.agda
```diff
@@ -178,41 +184,44 @@ NonNegative p = ℚᵘ.NonNegative (toℚᵘ p)
 ≢-nonZero {mkℚ +0       (suc d) c} p≢0 = contradiction (λ {i} → C.recompute c {i}) ¬0-coprimeTo-2+
 
 >-nonZero : ∀ {p} → p > 0ℚ → NonZero p
->-nonZero {p} (*<* p<q) = ℚᵘ.>-nonZero {toℚᵘ p} (ℚᵘ.*<* p<q)
+>-nonZero {p@(mkℚ _ _ _)} (*<* p<q) = ℚᵘ.>-nonZero {toℚᵘ p} (ℚᵘ.*<* p<q)
```
andreasabel, line 187: Why not the '@record' pattern here as well?

PR #1768, opened 2022-05-09
[ fix #1765 ] Specialise re-exported constructors

### LINKED ISSUE #1765: Unsolved implicits in Data.List.Relation.Ternary.Appending.Propositional
The following code checks with unsolved implicit arguments – in particular a type `A` and its level – in both constructors of `Appending`. This seems to be an unused argument, because I can fill it with any old rubbish. I don't think anything similar affects the `.Setoid` module.

```agda
open import Data.List
open import Data.List.Relation.Ternary.Appending.Propositional
open import Data.List.Relation.Binary.Pointwise hiding (refl)
open import Data.Nat

foo : Appending (3 ∷ []) (2 ∷ []) (3 ∷ 2 ∷ [])
foo = refl ∷ []++ refl ∷ []
```

Agda version 2.6.2.1, stdlib version 1.7.1.

gallais:


src/Data/List/Relation/Ternary/Appending.agda
```diff
@@ -42,7 +38,7 @@ module _ {A : Set a} {B : Set b}
 ------------------------------------------------------------------------
 -- Functions manipulating Appending
 
-_++_ : ∀ {cs₁ cs₂} → Pointwise L as cs₁ → Pointwise R bs cs₂ →
+_++_ : ∀ {cs₁ cs₂ : List C} → Pointwise L as cs₁ → Pointwise R bs cs₂ →
```
gallais, line 41: There was a subtle bug in here due to the `variable` block! :-o

JacquesCarette: Wow - super subtle.

MatthewDaggitt:
Thanks @gallais! Any chance you could add a CHANGELOG entry documenting the fixed bug?

gallais:
Ah but now I have to understand the bug to document it.

MatthewDaggitt:
> Ah but now I have to understand the bug to document it.

:laughing: okay, well maybe just a comment saying fixed a bug with unsolved meta variables or something non-descript?

laMudri:
Is it worth adding my bug report in a `private` block as a unit test, or is it just really unlikely to break in the same way again?

gallais:
I think I'll try to write a certified prefix matcher and put it in the test suite. This way we will also
get Agda to compile the module. :)

src/Data/List/Relation/Ternary/Appending.agda
gallais: My bad: there was no bug. The bug was introduced by moving the sets from the `variable` block
to the telescope of parameters.

PR #1769, opened 2022-05-14
Make the RingSolver work with generic rings (#1116)

felixwellen:
Work on issue #1116 . So far, this only concerns the problem that deBruijn-indices have to be adjusted when using the forall-version of the solver macro.

### DIFF
--- a/src/Reflection/AST/DeBruijn.agda
+++ b/src/Reflection/AST/DeBruijn.agda
@@ -58,7 +58,7 @@ module _ where
 -- Apply Weakening to substitute under lambdas
 
 prependLams : List (String × Visibility) → Term → Term
-prependLams xs t = foldr (λ {(s , v) t → lam v (abs s t)}) t (reverse xs)
+prependLams xs t = foldr (λ (s , v) t → lam v (abs s t)) t (reverse xs)
 
 prependHLams : List String → Term → Term
 prependHLams vs = prependLams (map (_, hidden) vs)
--- a/src/Reflection/AST/DeBruijn.agda
+++ b/src/Reflection/AST/DeBruijn.agda
@@ -10,7 +10,9 @@ module Reflection.AST.DeBruijn where
 
 open import Data.Bool.Base using (Bool; true; false; _∨_; if_then_else_)
 open import Data.Nat.Base as Nat using (ℕ; zero; suc; _+_; _∸_; _<ᵇ_; _≡ᵇ_)
-open import Data.List.Base using (List; []; _∷_; _++_)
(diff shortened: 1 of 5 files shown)


src/Reflection/AST/DeBruijn.agda
```diff
@@ -53,6 +55,16 @@ module _ where
   weakenClauses : (by : ℕ) → Clauses → Clauses
   weakenClauses = weakenFrom′ traverseClauses 0
 
+-- Apply Weakening to substitute under lambdas
+
+prependLams : List (String × Visibility) → Term → Term
+prependLams xs t = foldr (λ {(s , v) t → lam v (abs s t)}) t (reverse xs)
```
gallais, line 61: Note that you don't need a pattern-matching lambda for irrefutable patterns like `(s , v)`

src/Tactic/RingSolver/Examples.agda
```diff
@@ -0,0 +1,71 @@
...
+  ℤAsCRing = record
+               { Carrier = ℤ
+               ; _≈_ = _
+               ; _+_ = _
+               ; _*_ = _
+               ; -_ = _
+               ; 0# = _
+               ; 1# = _
+               ; isCommutativeRing = +-*-isCommutativeRing
+               }
```
gallais, line 34:
```suggestion
               { Carrier = ℤ
               ; isCommutativeRing = +-*-isCommutativeRing
               }
```

src/Reflection/AST/DeBruijn.agda
felixwellen: Thanks! Didn't know that..

src/Tactic/RingSolver/Examples.agda
```diff
@@ -0,0 +1,65 @@
...
+open import Data.Integer.Properties
+open import Data.Maybe
+
+open import Agda.Builtin.Equality using (_≡_ ; refl)
+
+open import Tactic.RingSolver
+open import Tactic.RingSolver.Core.AlmostCommutativeRing
+
+
+module IntegerExamples where
```
MatthewDaggitt, line 23: You could add these examples to the existing [README.Tactic.RingSolver](https://github.com/agda/agda-stdlib/blob/master/README/Tactic/RingSolver.agda)?

MatthewDaggitt:
@felixwellen unless you have any objections in the next week or so I'll close this PR as it's going stale?

felixwellen:
Yes, I think it can be closed. From my vague memories, the status was, that I didn't understand why some expressions do not normalize and I don't think it is worth figuring out.

PR #1779, opened 2022-06-06
Make `Map` public in `Data.Tree.AVL.IndexedMap`

KislyjKisel:


gallais:
Won't this make goals a lot more horrible to read?

KislyjKisel:
Can you elaborate, please? Was this made deliberately? Is it harder to read when using "simplified" view?
When experimenting, I needed to take an indexed Map as an argument to a function, a synonym would be handy. Functions inside this module use it, so as a workaround I used `typeOf Map.empty`.

MatthewDaggitt:
So @gallais's point is that now some `Map`s that appear in the goal type will be reduced to the horrible `Map′ (AVL.MkValue (Value ∘ proj₁) (subst Value ∘′ cong proj₁))`. 

I agree with you @KislyjKisel that it seems desirable to expose the name `Map`. Maybe we should put it under an `abstract` block instead which would get us the best of both worlds. Thoughts @gallais?

KislyjKisel:
As I understand putting `Map` under `abstract` prohibits its use in general `Data.Tree.AVL.Tree` operations. Would it require making conversions and duplication of generic code? Also, when do goals appear partially normalized? With trivial example of `x : Map`, goal for `x` shows `Map` if choosing "simplified" (when `Map` is public), and when choosing "normalized" it shows fully normalized `Data.Tree.AVL.Tree (record { ... }) (Data.Tree.AVL.Value.MkValue ... ...)` (both with public and private `Map`).

PR #1781, opened 2022-06-17
Support for abstract vector spaces.

capn-freako:
This PR introduces `Algebra.Linear.VectorSpace`, an abstract vector space.

The purpose of this new type is to provide the familiar metaphor/interface of "vector/matrix" that people are used to using in their client code, without locking the back-end implementation into arrays (more generally, n-tuples).
In this way, we hope to decouple the _capture of design intent_ from _design implementation_, for architectures based on the concepts of linear algebra, such as machine learning networks.


capn-freako:
Okay, Matthew, I just pushed after merging in upstream/master.

src/Data/Nat/Logarithm.agda
```diff
@@ -8,7 +8,7 @@
 
 module Data.Nat.Logarithm where
 
-open import Data.Nat 
+open import Data.Nat
```
MatthewDaggitt, line 11: I know this annoying, but any chance you could submit these two whitespace changes in a separate PR that I'll merge in immediately?

src/Algebra/Module/Morphism/Properties.agda
```diff
@@ -0,0 +1,125 @@
...
+open import Relation.Nullary          using (¬_)
+open import Relation.Nullary.Negation using (contraposition)
+
+module LinearMapProperties
+  {r ℓr m ℓm n ℓn : Level}
+  {ring           : CommutativeRing r ℓr}
+  {modA           : Module ring m ℓm}
+  {modB           : Module ring n ℓn}
+  (lm             : LinearMap modA modB)
+  where
```
MatthewDaggitt, line 30: This seems to be an old version of [Algebra.Module.Morphism.ModuleHomomorphism](https://github.com/agda/agda-stdlib/blob/master/src/Algebra/Module/Morphism/ModuleHomomorphism.agda)? Can we remove this file from the PR?

src/Data/VectorSpace/Core.agda
```diff
@@ -0,0 +1,94 @@
...
+  field
+    _∙_           : V → V → S
+    basisSet      : List V
+    basisComplete : ∀ {a : V} →
+                    a ≈ᴹ vgen (a ∙_) basisSet
+
+    -- ToDo: Can these be unified, by using one of the
+    -- existing algebraic structures?
+    -- I'm only finding things that are predicated upon: `A → A → A`, or
+    -- `A → B`; nothing for: `A → A → B`.
```
MatthewDaggitt, line 63: See `Algebra.Module.Definitions`

src/Function/Consequences.agda
```diff
@@ -54,3 +55,24 @@ module _
   contraInjective : Injective _≈₁_ _≈₂_ f →
                     ∀ {x y} → ¬ (x ≈₁ y) → ¬ (f x ≈₂ f y)
   contraInjective inj p = contraposition inj p
+
+∘-cong : ∀ {f} {g} {≈₁ : Rel A ℓ₁} {≈₂ : Rel B ℓ₂} →
+          Congruent ≈₁ ≈₁ f → Congruent ≈₁ ≈₂ g → Congruent ≈₁ ≈₂ (g ∘ f)
+∘-cong f-cong g-cong = g-cong ∘ f-cong
+-- ∘-cong = flip _∘_  -- Yields implicit/explicit function type error.
```
MatthewDaggitt, line 62: That's because `_∘_` is the dependent version and therefore you can't flip it. If you use the primed version "∘\prime" then you should be able to flip it.

MatthewDaggitt: This lemma already exists though, in `Function.Construct.Composition`, which is where the `∘-cong₂` should live as well if we keep it.

src/Function/Consequences.agda
```diff
@@ -54,3 +55,24 @@ module _
...
+
+∘-cong₂ : ∀ {f} {g} {≈₁ : Rel A ℓ₁} {≈₂ : Rel B ℓ₂} →
+          Congruent ≈₁ ≈₁ f → Congruent₂ ≈₁ ≈₁ ≈₂ g → Congruent₂ ≈₁ ≈₁ ≈₂ (g ∘ f)
+∘-cong₂ f-cong g-cong₂ = g-cong₂ ∘ f-cong
+
+module _ (To : Setoid b ℓ₂) {≈₁ : Rel A ℓ₁} where
+
+  open Setoid To using () renaming (Carrier to B; _≈_ to ≈₂)
+
+  flip-cong₂ : ∀ {f} → Congruent₂ ≈₁ ≈₁ ≈₂ f → Congruent₂ ≈₁ ≈₁ ≈₂ (flip f)
```
MatthewDaggitt, line 72: Likewise this should live in `Algebra.Construct.Symmetry`

src/Function/Definitions.agda
```diff
@@ -28,6 +28,12 @@ open import Level using (_⊔_)
 Congruent : (A → B) → Set (a ⊔ ℓ₁ ⊔ ℓ₂)
 Congruent f = ∀ {x y} → x ≈₁ y →  f x ≈₂ f y
 
+Congruent₂′ : (A → B → A) → Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂)
+Congruent₂′ f = ∀ {x y w z} → x ≈₁ y → w ≈₂ z → f x w ≈₁ f y z
+module _ {c ℓ₃} {C : Set c} (_≈₃_ : Rel C ℓ₃) where
+  Congruent₂ : (A → B → C) → Set (a ⊔ b ⊔ ℓ₁ ⊔ ℓ₂ ⊔ ℓ₃)
+  Congruent₂ f = ∀ {x y w z} → x ≈₁ y → w ≈₂ z → f x w ≈₃ f y z
```
MatthewDaggitt, line 35: These definitions probably already exist in the `Algebra.Module.Definitions.(Left/Right/Bi)` modules? If not, then they should be added there.

src/Data/VectorSpace/Core.agda
```diff
@@ -0,0 +1,94 @@
...
+    B : Set b
+    C : Set c
+
+record VectorSpace
+  {r ℓr m ℓm : Level}
+  {ring      : CommutativeRing r ℓr}
+  (mod       : Module ring m ℓm)
+  : Set (suc (ℓr ⊔ r ⊔ ℓm ⊔ m)) where
+
+  constructor mkVS
```
MatthewDaggitt, line 41: I refuse to believe that this constructor with 15+ arguments is actually useful :laughing:

src/Data/VectorSpace/Core.agda
```diff
@@ -0,0 +1,94 @@
...
+    ∙-distrib-+ : ∀ {a b c}   → a ∙ (b +ᴹ c)    ≈ (a ∙ b) + (a ∙ c)
+    ∙-comm-*ₗ   : ∀ {s} {a b} → a ∙ (s *ₗ b)    ≈ s * (a ∙ b)
+    ∙-comm-*ᵣ   : ∀ {s} {a b} → a ∙ (b *ᵣ s)    ≈ (a ∙ b) * s  -- Prove.
+    ∙-congˡ     : ∀ {a b c}   → b ≈ᴹ c → a ∙ b ≈ a ∙ c
+    ∙-congʳ     : ∀ {a b c}   → b ≈ᴹ c → b ∙ a ≈ c ∙ a        -- Prove.
+    ∙-idˡ       : ∀ {a}       → 0ᴹ ∙ a ≈ 0#
+    ∙-idʳ       : ∀ {a}       → a ∙ 0ᴹ ≈ 0#                    -- Prove.
+    ∙-homo-⁻¹    : ∀ {a b}     → a ∙ (-ᴹ b) ≈ - (a ∙ b)
+
+  vscale-cong : ∀ f → Congruent _≈ᴹ_ _≈_ f → Congruent _≈ᴹ_ _≈ᴹ_ (vscale f)
```
MatthewDaggitt, line 74: This proof shouldn't live in the definition of the record itself.

src/Data/VectorSpace/Core.agda
```diff
@@ -0,0 +1,94 @@
...
+  V⊸S = LinearMap mod ⟨module⟩
+
+  -- Equivalent vector generator.
+  v : V⊸S → V
+  v lm = vgen (LinearMap.f lm) basisSet
+
+  open Setoid (≈ᴸ-setoid mod ⟨module⟩) using () renaming
+    ( _≈_ to _≈ᴸ_
+    ; _≉_ to _≉ᴸ_
+    ) public
```
MatthewDaggitt, line 93: `public` should go immediately before the `using`.

src/Data/VectorSpace/Core.agda
```diff
@@ -0,0 +1,94 @@
...
+--
+-- Abstract vector spaces.
+--
+-- A "vector space" is a Module with a commutative, homomorphic inner
+-- product and a complete set of building blocks for mapping the space.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.VectorSpace.Core where
```
MatthewDaggitt, line 12: This shouldn't live in `Data`. If I had to make a stab at it, a better place would be to create a new subfolder `Algebra.Linear` for linear algebra and place `VectorSpace` in `Algebra.Linear.Bundles`?

src/Relation/Binary/ExtensionalEquivalence.agda
```diff
@@ -0,0 +1,37 @@
...
+-- Definition of extensional equivalence.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Relation.Binary.Bundles
+open import Relation.Binary.Definitions
+
+module Relation.Binary.ExtensionalEquivalence
```
MatthewDaggitt, line 13: This should probably live in `Function.Relation.Binary.Equality` (a bit of a mouthful I know, there's discussions to improve the name).

We should also generalise it over setoid equality for both the input and the output of the function!

src/Relation/Binary/ExtensionalEquivalence.agda
```diff
@@ -0,0 +1,37 @@
...
+    A : Set a
+
+open Setoid B-setoid renaming (Carrier to B)
+
+infix 4 _≗_
+
+_≗_ : ∀ {a} → {A : Set a} → (f g : A → B) → Set (ℓ ⊔ a)
+f ≗ g = ∀ x → f x ≈ g x
+
+≗-refl : ∀ {a} → {A : Set a} → Reflexive (_≗_ {A = A})
```
MatthewDaggitt, line 30: You can avoid a set of brackets by writing `Reflexive {A = A} _≗_`

MatthewDaggitt (commented):
Thanks for bringing it up to date. I've done an initial review, but be warned this another relatively big and complicated PR so it's going to take a few iterations!

src/Data/VectorSpace/Core.agda
capn-freako: **==> Code moved, as requested, in next push.**

capn-freako (commented):
Partial response, intended to facilitate further discussion, coming in next push.

src/Data/Nat/Logarithm.agda
capn-freako: **==> Done.**

src/Relation/Binary/ExtensionalEquivalence.agda
MatthewDaggitt: Yes, sorry, should have made clearer that it would be a new module!

src/Relation/Binary/ExtensionalEquivalence.agda
MatthewDaggitt: Ah sorry, yes sorry, should be something like `Reflexive {A = A → B} _≗_`?

src/Data/VectorSpace/Core.agda
MatthewDaggitt: Ah sorry, yes, the definitions I was thinking of are here:
https://github.com/agda/agda-stdlib/blob/d5ecb051ffb1965b2db45d09e330a13b3bf64471/src/Algebra/Module/Core.agda#L18-L22

But actually `A -> A -> B` is a bit weird as you're conjuring out an entirely new `B` from 2 `A`s. I'm not aware of any general name for such an operation, but you probably know the literature better than I do? I'd just call that a function... and hence not give it a name.

src/Data/VectorSpace/Core.agda
MatthewDaggitt: No I didn't! I feel like maybe we should be extracting the concept of `IsBasis` out into its own record?

src/Algebra/Linear/Bundles.agda
```diff
@@ -0,0 +1,95 @@
...
+    a b c : Level
+    A : Set a
+    B : Set b
+    C : Set c
+
+-- Abstract vector spaces.
+--
+-- A "vector space" is a Module with a commutative, homomorphic inner
+-- product and a complete set of building blocks for mapping the space.
+record VectorSpace
```
MatthewDaggitt, line 36: Can we follow the standard pattern in `Algebra` and split this out into a `IsVectorSpace` structure in `Algebra.Linear.Structure` and `VectorSpace` bundle in `Algebra.Linear.Bundle`?

src/Algebra/Linear/Properties.agda
```diff
@@ -0,0 +1,180 @@
...
+v∙g[x]+y-cong₂ {g} {v} {x} {w} {y} {z} g-cong x≈w y≈z = begin⟨ setoid ⟩
+  v ∙ g x + y ≈⟨ +-congʳ (∙-congˡ (g-cong x≈w)) ⟩
+  v ∙ g w + y ≈⟨ +-congˡ y≈z ⟩
+  v ∙ g w + z ∎
+
+foldr-cong : ∀ {f g : V → S → S} {d e : S} →
+             (∀ {y z} → y ≈ z → ∀ x → f x y ≈ g x z) → d ≈ e →
+             foldr f d ≗ foldr g e
+foldr-cong f≈g d≈e []       = d≈e
+foldr-cong f≈g d≈e (x ∷ xs) = f≈g (foldr-cong f≈g d≈e xs) x
```
MatthewDaggitt, line 58: I feel this should be a far more general proof in `Data.List.Properties`? (if it doesn't already exist!)

src/Data/VectorSpace/Core.agda
capn-freako: Okay, thanks!

src/Relation/Binary/ExtensionalEquivalence.agda
capn-freako: Okay, I created that new module.

src/Relation/Binary/ExtensionalEquivalence.agda
capn-freako: Yep, that worked; thanks!

src/Algebra/Linear/Bundles.agda
capn-freako: Yep. Arriving w/ next push.

src/Algebra/Linear/Properties.agda
capn-freako: Yeah, I get that. But, the one in `Data.List.Properties` uses _equality_, as opposed to _equivalence_, which I can't use here:

```agda
foldr-cong : ∀ {f g : A → B → B} {d e : B} →
             (∀ x y → f x y ≡ g x y) → d ≡ e →
             foldr f d ≗ foldr g e
```

That's why I had to write my own. :(

src/Data/VectorSpace/Core.agda
capn-freako: Yeah, I think you're right.
Check out the new:

- `Algebra.Linear.Structures`,
- `Algebra.Linear.Bundles`, and
- `Algebra.Linear.Properties.VectorSpace`.

And let me know if that seems a little more consistent w/ the established conventions to you.

Thanks!
:)

src/Algebra/Linear/Properties.agda
Taneb: Would `Data.List.Relation.Binary.Setoid.foldr⁺` be useful here?

capn-freako: > Would `Data.List.Relation.Binary.Setoid.foldr⁺` be useful here?

I don't think so, only because it applies to `Op2 A` while mine applies to `V -> S -> S`.

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,64 @@
...
+  {ring      : CommutativeRing r ℓr}
+  (mod       : Module ring m ℓm)
+  where
+
+  open CommutativeRing ring renaming (Carrier  to S)
+  open Module          mod  renaming (Carrierᴹ to V)
+
+  -- A set of "vectors" forms a basis for a space iff it is complete
+  -- under some inner product.
+  record IsBasis : Set (suc (ℓm ⊔ m ⊔ r)) where
```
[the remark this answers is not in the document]
MatthewDaggitt: Actually maybe a basis is a fundemental enough concept that it (and related proofs) should live in its own module `Algebra.Linear.Basis`?

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,185 @@
...
+  ; from      = λ u  → mkLM (u ∙_) u∙-homo
+  ; to-cong   = vgen-cong basisSet
+  ; from-cong = λ z x → ∙-congʳ z
+  ; inverse   = (λ v → begin⟨ ≈ᴹ-setoid ⟩
+                     vgen (v ∙_) basisSet
+                       ≈⟨ Setoid.sym ≈ᴹ-setoid basisComplete ⟩
+                     v ∎ )
+                , λ lm x → begin⟨ setoid ⟩
+                      v lm ∙ x ≈⟨ sym (f≈v∙ lm) ⟩
+                      LinearMap.f lm x   ∎
```
MatthewDaggitt, line 184: As a first step could you expand these inline proofs out into a `where` clause?

src/Algebra/Linear/Bundles.agda
```diff
@@ -0,0 +1,52 @@
...
+
+  open IsBasis       isBasis       public
+  open IsVectorSpace isVectorSpace public
+
+  -- Linear maps from vectors to scalars.
+  V⊸S = LinearMap mod ⟨module⟩
+
+  -- Equivalent vector generator.
+  v : V⊸S → V
+  v lm = vgen (LinearMap.f lm) basisSet
```
MatthewDaggitt, line 46: I have no idea what the comment means or what it is? "Equivalent" to what? Definitely needs a better name than just `v`!

CHANGELOG.md
```diff
@@ -1851,7 +1871,6 @@ Other minor changes
   ∈-++-comm : ∀ {v} xs ys → v ∈ xs ++ ys → v ∈ ys ++ xs
   ∈-++-comm∘++-comm : ∀ {v} xs {ys} (p : v ∈ xs ++ ys) → ∈-++-comm ys xs (∈-++-comm xs ys p) ≡ p
   ∈-++↔++ : ∀ {v} xs ys → v ∈ xs ++ ys ↔ v ∈ ys ++ xs
-  ```
```
MatthewDaggitt, line 1854: Is there a reason you've deleted this line?

MatthewDaggitt (commented):
Thanks, already looking much better! 

Next set of suggestions incoming :smile:

src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,37 @@
...
+  variable
+    a : Level
+    A : Set a
+
+open Setoid B-setoid renaming (Carrier to B)
+
+infix 4 _≗_
+
+_≗_ : ∀ {a} → {A : Set a} → (f g : A → B) → Set (ℓ ⊔ a)
+f ≗ g = ∀ x → f x ≈ g x
```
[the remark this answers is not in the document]
capn-freako: I think that requires _congruence_, which isn't currently required.

src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,37 @@
...
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Level
+open import Relation.Binary.Bundles
+open import Relation.Binary.Definitions
+
+module Function.Relation.Binary.Equality
+  {b ℓ : Level}
+  (B-setoid : Setoid b ℓ)
```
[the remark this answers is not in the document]
capn-freako: Changed as requested in next push.

src/Function/Relation/Binary/Equality.agda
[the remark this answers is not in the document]
capn-freako: I think that would require adding `Congruence`, which the current definition does _not_ require.  
(Is that correct?)

src/Algebra/Linear/Structures.agda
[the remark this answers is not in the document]
capn-freako: It has only a couple proofs. (The rest are on `VectorSpace`.)
So, I went w/ your second suggestion.
It felt the most correct to me.

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,64 @@
...
+      _∙_           : V → V → S
+      basisSet      : List V                     -- ToDo: List => Foldable Functor
+      basisComplete : ∀ {a : V} →
+                      a ≈ᴹ vgen (a ∙_) basisSet
+
+
+  -- The space mapped (i.e. - represented) by a basis constitutes a
+  -- vector space iff the inner product has certain properties.
+  record IsVectorSpace
+    (isBasis : IsBasis)
```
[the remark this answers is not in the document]
capn-freako: Yep, you're right.
Changed, as requested, in next push.

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,64 @@
...
+-- The Agda standard library
+--
+-- Some linear algebraic structures.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Linear.Structures where
+
+open import Algebra                             using (CommutativeRing)
```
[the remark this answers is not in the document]
capn-freako: Fixed in next push.

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,64 @@
...
+
+  open CommutativeRing ring renaming (Carrier  to S)
+  open Module          mod  renaming (Carrierᴹ to V)
+
+  -- A set of "vectors" forms a basis for a space iff it is complete
+  -- under some inner product.
+  record IsBasis : Set (suc (ℓm ⊔ m ⊔ r)) where
+
+    vscale : (V → S) → V → V
+    vscale f = uncurry _*ₗ_ ∘ < f , id >
```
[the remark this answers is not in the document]
capn-freako: Added in next push.

src/Algebra/Linear/Properties/VectorSpace.agda
capn-freako: Changed as requested in next push.

src/Algebra/Linear/Bundles.agda
capn-freako: Fixed in next push.

CHANGELOG.md
capn-freako: Nope; restored in next push.

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,64 @@
...
+    : Set (suc (ℓm ⊔ ℓr ⊔ m ⊔ r))
+    where
+
+    open IsBasis isBasis
+
+    field
+      ∙-comm      : ∀ {a b}     → a ∙ b ≈ b ∙ a
+      ∙-distrib-+ : ∀ {a b c}   → a ∙ (b +ᴹ c)    ≈ (a ∙ b) + (a ∙ c)
+      ∙-comm-*ₗ   : ∀ {s} {a b} → a ∙ (s *ₗ b)    ≈ s * (a ∙ b)
+      ∙-comm-*ᵣ   : ∀ {s} {a b} → a ∙ (b *ᵣ s)    ≈ (a ∙ b) * s  -- Prove.
```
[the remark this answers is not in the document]
capn-freako: Did you mean, "remove the `Prove` _comment_"?

src/Function/Relation/Binary/Equality.agda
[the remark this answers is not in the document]
capn-freako: Shoot! I think I actually need to keep `A` as a general purpose variable.
I'm getting the following w/ the new code:

```agda
66 foldr-cong : ∀ {f g : V → S → S} {d e : S} →
67              (∀ {y z} → y ≈ z → ∀ x → f x y ≈ g x z) → d ≈ e →
68             foldr f d ≗ foldr g e
69 foldr-cong f≈g d≈e []       = d≈e
70 foldr-cong f≈g d≈e (x ∷ xs) = f≈g (foldr-cong f≈g d≈e xs) x

```

(All code on line 68 highlighted in medium brown.)

```
/home/dbanas/prj/agda-stdlib/src/Algebra/Linear/Properties/VectorSpace.agda:68,14-23
V !=< List _A_210
  List _A_210 → _B_212
matches the expected type
  Setoid.Carrier ≈ᴹ-setoid → Setoid.Carrier setoid
```

I'm guessing that if I fix this for `List` then it will just break again when I move to `Foldable Functor`.
What do you think?

I've temporarily added a new module: `Function.Relation.Binary.Equality2`, which contains the new code (i.e. - modified as per your request).
And it can be selected, via comment toggling, at these locations:

1. `src/Algebra/Linear/Properties/VectorSpace.agda: 39`
2. `src/Algebra/Module/Morphism/Bundles.agda: 50`

I think I'll need some help w/ this one if you want to push towards two module parameterization.

Thanks!

src/Algebra/Linear/Structures.agda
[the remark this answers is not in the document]
MatthewDaggitt: Yes, sorry should have been more clear!

src/Algebra/Linear/Bundles.agda
```diff
@@ -0,0 +1,56 @@
...
+  field
+    _∙_ : V → V → S
+    basisFor      : BasisFor      mod _∙_
+    isVectorSpace : IsVectorSpace mod _∙_
+
+  open BasisFor      basisFor      public
+  open IsVectorSpace isVectorSpace public
+
+  -- Linear maps from vectors to scalars.
+  V⊸S = LinearMap mod ⟨module⟩
```
MatthewDaggitt, line 45: Could we add a type for `V⊸S`?

src/Algebra/Linear.agda
```diff
@@ -0,0 +1,11 @@
...
+-- The Agda standard library
+--
+-- Definitions of linear algebra bundles
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Linear where
+
+open import Algebra.Linear.Bundles public
```
MatthewDaggitt, line 11: Should add `Linear.Structures` here.

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of linear algebraic structures.
```
MatthewDaggitt, line 4: Should be "Properties of vector spaces"

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,58 @@
...
+  -- under some inner product.
+  record BasisFor (_∙_ : V → V → S) : Set (suc (ℓm ⊔ m ⊔ r)) where
+
+    -- Scale a vector according to some reduction function.
+    vscale : (V → S) → V → V
+    vscale f = uncurry _*ₗ_ ∘ < f , id >
+
+    -- Generate a vector from the basis, according to some reduction function.
+    vgen : (V → S) → List V → V
+    vgen f = foldr (_+ᴹ_ ∘ vscale f) 0ᴹ
```
MatthewDaggitt, line 37: Hmm the type of this function looks a bit weird as is, I assume the `List V` that it takes *is* the basis set declared below? Why not move the definition below the fields, and then pass `basisSet` to it explicitly?

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
...
+vgen-cong {f₁} {f₂} []       f₁≗f₂ = Setoid.reflexive ≈ᴹ-setoid Eq.refl
+vgen-cong {f₁} {f₂} (x ∷ xs) f₁≗f₂ = begin⟨ ≈ᴹ-setoid ⟩
+  f₁ x *ₗ x +ᴹ vgen f₁ xs ≈⟨ +ᴹ-congʳ (*ₗ-congʳ (f₁≗f₂ x)) ⟩
+  f₂ x *ₗ x +ᴹ vgen f₁ xs ≈⟨ +ᴹ-congˡ (vgen-cong xs f₁≗f₂) ⟩
+  f₂ x *ₗ x +ᴹ vgen f₂ xs ∎
+
+------------------------------------------------------------------------
+-- Some consequences of `VectorSpace` inner product properties.
+v∙g[x]+y-cong₂ : {g : V → V} {v x w : V} {y z : S} →
+                 Congruent _≈ᴹ_ _≈ᴹ_ g → x ≈ᴹ w → y ≈ z →
```
MatthewDaggitt, line 62: Could the `Congruent` proof go before the `v x w y z` declarations? That allows the function to be partially applied more easily. Similar comment with several others below, e.g. `foldr-homo-∙`

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
...
+      ≈⟨ +-congˡ (foldr-homo g xs) ⟩
+    g x * f x + vred g xs
+      ∎
+    where
+    h = _+ᴹ_ ∘ vscale g
+
+  vSum : List V → V
+  vSum xs = foldr _+ᴹ_ 0ᴹ xs
+
+  fScale : V → V
```
MatthewDaggitt, line 132: What does the `f` stand for in the names here? Again comments on these functions would be welcome!

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
...
+      ≈⟨ sym (foldr-homo (a ∙_) basisSet) ⟩
+    f (foldr (_+ᴹ_ ∘ (uncurry _*ₗ_) ∘ < (a ∙_) , id >) 0ᴹ basisSet)
+      ≈⟨ ⟦⟧-cong (Setoid.sym ≈ᴹ-setoid (basisComplete)) ⟩
+    f a ∎)
+    where
+    v′ = lmToVec lm
+
+  -- Inner product extensional equivalence.
+  x·z≈y·z⇒x≈y : ∀ {x y} → DoubleNegationElimination ℓm →
+                 ∃₂ (λ s z → ((s *ₗ (x +ᴹ -ᴹ y) ≈ᴹ z) × (f z ≉ 0#))) →
```
MatthewDaggitt, line 160: Any reason for declaring this as a dependent pair and not just a standard quantification?

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
...
+    fx≈fy : f x ≈ f y
+    fx≈fy = begin⟨ setoid ⟩
+      f x   ≈⟨ f≈v∙ {x} ⟩
+      v′ ∙ x ≈⟨ ∙-comm ⟩
+      x ∙ v′ ≈⟨ x∙z≈y∙z ⟩
+      y ∙ v′ ≈⟨ ∙-comm ⟩
+      v′ ∙ y ≈⟨ sym (f≈v∙ {y}) ⟩
+      f y   ∎
+
+-- Isomorphism 1: (V ⊸ S) ↔ V
```
MatthewDaggitt, line 174: Again, could you make this comment a little bit more verbose?

src/Algebra/Linear/Properties/VectorSpace.agda
```diff
@@ -0,0 +1,191 @@
...
+  vSum : List V → V
+  vSum xs = foldr _+ᴹ_ 0ᴹ xs
+
+  fScale : V → V
+  fScale = vscale f
+
+  fGen : List V → V
+  fGen = vgen f
+
+  f≈v∙ : ∀ {a} → f a ≈ lmToVec lm ∙ a
```
MatthewDaggitt, line 138: Could we try to come up with a more meaningful name for this proof?

src/Function/Relation/Binary/Equality.agda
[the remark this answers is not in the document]
MatthewDaggitt: Hmm yes good point. We really want equality over `Function` bundles.... I'll give this some more thought and get back to you.

Taneb:
I'm... uncomfortable with this definition of vector spaces. It's really not what literature means by "vector space", like, at all.  If anything I'd call this a free inner-product module, but that's not really a thing, and besides a definition of that ought to take into account complex conjugates somehow.

src/Algebra/Linear/Structures.agda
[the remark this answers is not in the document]
capn-freako: No problem.

**==> Changed, as requested, in next push.**

src/Algebra/Linear/Bundles.agda
```diff
@@ -0,0 +1,56 @@
...
+  constructor mkVS
+
+  open CommutativeRing ring renaming (Carrier  to S)   public
+  open Module          mod  renaming (Carrierᴹ to V)   public
+
+  infix 7 _∙_
+
+  field
+    _∙_ : V → V → S
+    basisFor      : BasisFor      mod _∙_
```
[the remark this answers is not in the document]
capn-freako: Changed as requested in next push.

src/Algebra/Linear/Bundles.agda
capn-freako: Added in next push.

src/Algebra/Linear.agda
capn-freako: Added in next push.

src/Algebra/Linear/Properties/VectorSpace.agda
capn-freako: Fixed in next push.

src/Algebra/Linear/Structures.agda
capn-freako: Because the `basisComplete` field calls it.

src/Algebra/Linear/Properties/VectorSpace.agda
capn-freako: Changed as requested in next push.

src/Algebra/Linear/Properties/VectorSpace.agda
capn-freako: `f` is the linear function contained within the linear map.
I've added a comment to this effect.

I've also made a clean-up pass through this code:

- eliminating any unused code,
- narrowing the scope of helper functions, and
- adding comments about intent/function.

**==> All changes described above arriving with next push.**

src/Algebra/Linear/Properties/VectorSpace.agda
capn-freako: I've used a longer, more explicit alternative, in the next push.
Let me know what you think.

capn-freako:
> I'm... uncomfortable with this definition of vector spaces. It's really not what literature means by "vector space", like, at all. If anything I'd call this a free inner-product module, but that's not really a thing, and besides a definition of that ought to take into account complex conjugates somehow.

Hi @Taneb ,

Thanks for voicing your concern here. :)

Is it _just_ the separation of the _basis set_ and _inner product_, which has evolved in the back and forth between Matthew and me, here, or is there something else that you feel is fundamentally wrong with my definition of _vector space_?

Thanks,
-db

capn-freako:
Anyone know why the _Ubuntu build / test-stdlib (pull_reuest)_ test failed?

src/Algebra/Linear/Properties/VectorSpace.agda
JacquesCarette: The name of this property mentions `z` (so one thinks it's an input), but the output doesn't have the right 'shape' as expected from the name.

I would have expected the type to be
```agda
∀ {x y} → (∀ {z} → x · z ≈ y · z) → x ≈ y
```

guilhermehas:
From Wikipedia (https://en.wikipedia.org/wiki/Vector_space), I think that Vector Space is defined over a Field and not from a Commutative Ring.

MatthewDaggitt:
> Anyone know why the Ubuntu build / test-stdlib (pull_reuest) test failed?

Not your fault (https://github.com/agda/agda-stdlib/issues/1804)

> From Wikipedia (https://en.wikipedia.org/wiki/Vector_space), I think that Vector Space is defined over a Field and not from a Commutative Ring.

Thanks @Taneb and @guilhermehas for the heads up! I'd totally missed that. It's a good point. @capn-freako I think they're right that we should stick to the standard definition. May I ask where you go the current definition from?

capn-freako:
> Thanks @Taneb and @guilhermehas for the heads up! I'd totally missed that. It's a good point. @capn-freako I think they're right that we should stick to the standard definition. May I ask where you go the current definition from?

Admittedly, I "cheated" when I didn't find `Field` available among the other standard algebraic definitions.
Shall I take a whack at defining `Field` and, then, redefining `VectorSpace` to use it, instead of `CommutativeRing`?
I think we'll need _division_ eventually, anyway, when working most generally w/ vector spaces, even abstractly.

JacquesCarette:
You're likely to want to define R-module over a (commutative?) ring first. That's fully constructive. Vector Space over a Field is... hard.

capn-freako:
> You're likely to want to define R-module over a (commutative?) ring first. That's fully constructive. Vector Space over a Field is... hard.

Thanks, Jacques! Have you been down that particular road before?

I like this concise summary of `Field` (from the _Wikipedia_ page):

> Even more summarized: a field is a [commutative ring](https://en.wikipedia.org/wiki/Commutative_ring) where 0 ≠ 1 and all nonzero elements are [invertible](https://en.wikipedia.org/wiki/Unit_(ring_theory)).

capn-freako:
> You're likely to want to define R-module over a (commutative?) ring first. That's fully constructive. Vector Space over a Field is... hard.

Oh, shoot! The new multiplicative inverse function must be _partial_, because `0#` must be excluded from its domain.
Is that why you warned me thusly?

capn-freako:
Hey guys, a question about something in `Algebra.Structures`:
In `IsRingWithoutOne`, the additive inverse is defined thusly:

```
+-isAbelianGroup : IsAbelianGroup + 0# -_
```

while in `IsNearring` it is defined:

```
+-inverse   : Inverse 0# _⁻¹ +
```

What is the significance of this difference?

MatthewDaggitt:
> Oh, shoot! The new multiplicative inverse function must be partial, because 0# must be excluded from its domain.
Is that why you warned me thusly?

Yes we haven't quite worked out how we want to handle the partiality. See https://github.com/agda/agda-stdlib/issues/1175 for discussion.

JacquesCarette:
> Have you been down that particular road before?

Not this exact one, but close enough to know that this way lies much pain.

> Oh, shoot! The new multiplicative inverse function must be partial, because 0# must be excluded from its domain.
Is that why you warned me thusly?

Yes!

Matthew also said:
> In general it's very difficult to come up with a consistent inheritance pattern and is an area of active research how to do it modularly (which we have definitely not achieved in the standard library!)

In my mind, the most general solution requires explicit morphisms between definitions (not just embeddings induced by same-name-same-thing) and a proper notion of pullback/pushout. Agda actually has many of the needed features already, but "closing the gap" still requires quite a lot of work.

capn-freako:
Hey guys,

Please, ignore this next push.
And sorry to trigger your runners unnecessarily.

I need to get something posted in an easily referenced fashion, so it can be discussed in group today.

Thanks,
-db

MatthewDaggitt:
@capn-freako I'm going to close this PR as I think the current approach needs some rethinking and probably some addition scaffolding first. Feel free to open a new PR if you revisit it later on and think you've fixed the problems!

capn-freako:
Oh, snap! I was just putting the finishing touches on a next push, which I think addresses the concerns and is much more in line w/ the literature.
Any chance you could reopen this PR for one more round?

MatthewDaggitt:
Sure!

capn-freako:
Okay, guys, I think that last push is much more in line with the literature and addresses your previous concerns.
Please, take a look and let me know what you think.

Thanks!
-db

src/Algebra/Linear/Structures.agda
```diff
@@ -0,0 +1,351 @@
...
+    vscale : (V → S) → V → V
+    vscale f = uncurry _*ₗ_ ∘ < f , id >
+
+    -- Accumulate a list of vectors, according to some reduction function.
+    vgen : (V → S) → List V → V
+    vgen f = foldr (_+ᴹ_ ∘ vscale f) 0ᴹ
+
+  -- An _Inner Product Space_ is a vector space augmented with a function
+  -- that takes two vectors and returns a scalar.
+  record IsInnerProductSpace (f : V → V → S) : Set (suc (ℓm ⊔ ℓr ⊔ m ⊔ r)) where
```
Taneb, line 106: There really ought to be _some_ laws here.

Taneb (commented):
On the whole I'm really not happy with the approach taken here. We certainly shouldn't merge this before the imminent release of 2.0, and I don't think this approach will lead to something merits inclusion in the standard library.

I welcome @capn-freako to adapt this into a standalone library and continue developing it.

I suggest for the standard library that we work towards linear algebra in small, well-reasoned chunks building on what we've already implemented. I have in my head a rough roadmap for this (and other things!) that I've been working on for a good while.

MatthewDaggitt:
Given @Taneb's feedback and the lack of reply I'm going to close this again for now.

PR #1783, opened 2022-06-25
Parametrized Real numbers and properties

DPDmancul:
Waiting for a constructive type representing real numbers, I wrote this type which postulates real numbers and their principal properties.
Then I wrote the proofs of all other common properties, which can be reused when constructive real number will be available.

When #1582 (Field algebra structure)  will be ready (it is waiting  #1175 for generic Nonzero) some of the properties of real numbers involving multiplicative inverse can be generalized to properties of all fields.

All the introduced postulates:
```agda
postulate
  ℝ     : Set
  0ℝ 1ℝ : ℝ
  1≢0   : 1ℝ ≢ 0ℝ

  _≟_ : DecidableEquality ℝ

  _+_ : ℝ → ℝ → ℝ
  -_  : ℝ → ℝ
  _*_ : ℝ → ℝ → ℝ
  1/_ : (x : ℝ) .⦃ _ : NonZero x ⦄ → ℝ

  +-assoc     : Associative _+_
  +-comm      : Commutative _+_
  +-identityˡ : LeftIdentity 0ℝ _+_
  +-inverseʳ  : RightInverse 0ℝ -_ _+_

  *-assoc         : Associative _*_
  *-comm          : Commutative _*_
  *-identityˡ     : LeftIdentity 1ℝ _*_
  *-distribˡ-+    : _*_ DistributesOverˡ _+_
  *-neg-identityˡ : (x : ℝ) → (- 1ℝ) * x ≡ - x
  *-inverseˡ      : ∀ x .⦃ _ : NonZero x  ⦄ → (1/ x) * x ≡ 1ℝ

  _≤_ : Rel ℝ 0ℓ
  0≤1 : 0ℝ ≤ 1ℝ

  ≤-refl    : Reflexive _≤_
  ≤-antisym : Antisymmetric _≡_ _≤_
  ≤-trans   : Transitive _≤_
  ≤-total   : Total _≤_

  +-mono-≤         : _+_ Preserves₂ _≤_ ⟶ _≤_ ⟶ _≤_
  *-monoʳ-≤-nonNeg : ∀ z .⦃ _ : NonNegative z ⦄ → (_* z) Preserves _≤_ ⟶ _≤_
```


MatthewDaggitt:
Thank you for the proposed PR. So a couple of points:

1. Unfortunately decidability and totality is not provable in Agda itself, as Agda is a constructive system. Therefore you're postulating things that can never be proved, even though they're valid in classical mathematics. The standard library probably wants to stick with constructive interpretations.

2. Nonetheless there would be significant value in adding an "interface" like this for the constructive real numbers. However rather than using postulates (which prevents a user from using the `--safe` option), I think a better approach would be formulate the interface as a record. Users could then either postulate the record in one line, or if they want to stay `--safe` they can parameterise their modules by the record.

Does that sound reasonable to you @DPDmancul ?

DPDmancul:
>     2. Nonetheless there would be significant value in adding an "interface" like this for the constructive real numbers. However rather than using postulates (which prevents a user from using the `--safe` option), I think a better approach would be formulate the interface as a record. Users could then either postulate the record in one line, or if they want to stay `--safe` they can parameterise their modules by the record.

Yes, I thought to write a parametrized module instead of postulating, but then I decided to postulate to avoid passing a lot of parameters each time. The idea of a record as parameter is great and should be the best option: I will change the code this way.

>     1. Unfortunately decidability and totality is not provable in Agda itself, as Agda is a constructive system. Therefore you're postulating things that can never be proved, even though they're valid in classical mathematics. The standard library probably wants to stick with constructive interpretations.

Unfortunately this is the big problem, and I was unsure about doing the PR for this. I do not recall a way to substitute those two postulates with something which can be constructively proven: for example we can prove the corresponding continuation monads, but then they cannot be used in the `with` statement, since Agda cannot split on functions. Maybe the continuation monads can be used in another way in the proofs, but I am not sure.
Have you any suggestions?

MatthewDaggitt:
> Unfortunately this is the big problem, and I was unsure about doing the PR for this. I do not recall a way to substitute those two postulates with something which can be constructively proven: for example we can prove the corresponding continuation monads, but then they cannot be used in the with statement, since Agda cannot split on functions. Maybe the continuation monads can be used in another way in the proofs, but I am not sure.
Have you any suggestions?

I'm afraid I'm not an expert on constructive mathematics, so I'm not sure I can give you any valuable advice. The `max` and `min` functions that you seem to be after here are not constructive though, so a some of the current contents of this PR will need to be removed. Alternatively, if you're after a classical representation of the reals, maybe it would be better to keep this as a standalone library rather than trying to make it fit the constructive ethos of the standard library?

DPDmancul:
Yes a standalone library representing classical reals could be an interesting idea, but if someone with experience in constructive mathematics (unfortunately I'm also new to the field) can tell us which part are not provable, I can remove them and obtain an interface which will fit the standard library spirit.

GenerateEverything.hs
```diff
@@ -42,10 +42,11 @@ unsafeModules = map modToFile
...
   , "Codata.Musical.Stream"
-  , "Data.PostulatedReal"
-  , "Data.PostulatedReal.Base"
-  , "Data.PostulatedReal.Propertis"
-  , "Data.PostulatedReal.Propertis.Core"
+  , "Data.ParametrizedReal"
+  , "Data.ParametrizedReal.Base"
+  , "Data.ParametrizedReal.Interface"
+  , "Data.ParametrizedReal.Propertis"
+  , "Data.ParametrizedReal.Propertis.Core"
```
gallais, line 49:
```suggestion
  , "Data.ParametrizedReal.Properties"
  , "Data.ParametrizedReal.Properties.Core"
```

gallais: Although I guess it should not be in the list of unsafe modules anymore if it's `--safe`

DPDmancul: Yes, you are absolutely right.

MatthewDaggitt:
It doesn't look like any advice is forthcoming unfortunately, so I'm going to close this PR in its current state as I don't think there's an obvious proposal for how to progress it. Feel free to open a new PR @DPDmancul if at some point you have a revised plan. Thank you for the proposal though, we appreciate it, even if it didn't work out.

PR #1784, opened 2022-07-09
Library for (Generalized) Inference Systems

LcicC:
Implementation of a library for (Generalized) Inference Systems

paper    : "Flexible Coinduction in Agda"
authors  : Luca Ciccone, Francesco Dagnino, Elena Zucca
doi      : https://doi.org/10.4230/LIPIcs.ITP.2021.13
examples : https://github.com/LcicC/inference-systems-agda

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The library has been tested using Agda 2.6.2.
 Highlights
 ----------
 
+* Library for (Generalized) Inference Systems in Data
+
 * A golden testing library in `Test.Golden`. This allows you to run a set
   of tests and make sure their output matches an expected `golden` value.
   The test runner has many options: filtering tests by name, dumping the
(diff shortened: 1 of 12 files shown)


LcicC:
Thank you for the answer. We thought that our library would have been a nice addition to std-lib because inference systems (rule-based definitions) are ubiquitous in computer science (type systems, operational semantics, proof systems, ...).  Moreover they are somewhat similar to  indexed (endo)containers. However, they are less used in the Agda community.

Regards, 

Luca Ciccone, Francesco Dagnino

MatthewDaggitt:
Yup, agreed, it's definitely useful! I can imagine it will be popular especially as the Agda community is generally interested in formalising PL theory. It's just that (in my own very subjective opinion) they're not going to be used in a suitably high enough proportion of non-PL projects to justify their inclusion into the standard library.

But thank you for the proposed PR, don't feel discouraged from making further ones!

PR #1786, opened 2022-07-13
Adding useful isomorphisms

guilhermehas:
Isomorphisms such as Unit isomorphism.


guilhermehas:
I proved this isomorphism too, but I don't know where to put it:

```agda
  ↔Fun : A ↔ B → (A → C) ↔ (B → C)
  ↔Fun AB = mk↔′ (_∘ from) (_∘ to)
    (λ f → ext λ _ → cong f (inverseˡ _))
    (λ f → ext λ _ → cong f (inverseʳ _))
    where open Inverse AB
```

JacquesCarette:
Great stuff, but I think you might be better off submitting multiple PRs rather than one large one (at least in the future). That way some things can be merged quickly regardless of whether a specific case is engendering a lot of discussion.

[I haven't wrapped my head around the new function hierarchy yet to suggest where to put the extra one, sorry]

src/Data/Vec/Recursive.agda
```diff
@@ -148,3 +155,18 @@ zip = zipWith _,_
...
+lift↔ : {A B : Set a} → ∀ n → A ↔ B → A ^ n ↔ B ^ n
+lift↔ 0 AB = ↔.refl
+lift↔ 1 AB = AB
+lift↔ (2+ n) AB = ×-cong AB (lift↔ _ AB)
+
+Vec↔^ : ∀ n → Vec A n ↔ (A ^ n)
+Vec↔^ 0 = mk↔ ((λ _ → refl) , λ where Vec.[] → refl)
+Vec↔^ 1 = mk↔′ Vec.head Vec.[_] (λ _ → refl) (λ where (x ∷ Vec.[]) → refl)
+Vec↔^ {A = A} (2+ n) with Vec↔^ {A = A} (suc n)
+... | vec↔ =  mk↔′ (λ where (x ∷ xs) → x , to xs) (λ (x , xs) → x ∷ from xs)
```
MatthewDaggitt, line 170: Do we need to `with`-abstract here? Can we write `where open Inverse (Vec↔^ {A = A} (suc n))`

src/Data/Vec/Recursive.agda
```diff
@@ -148,3 +155,18 @@ zip = zipWith _,_
 
 unzip : ∀ n → (A × B) ^ n → A ^ n × B ^ n
 unzip = unzipWith id
+
+private module ↔ {ℓ} = IsEquivalence (↔-isEquivalence {ℓ})
```
MatthewDaggitt, line 159: Hmm having to reparameterise the module every time we want to use reflexivity/transitivity of inverse is a bit ugly. Instead could we add proofs of `↔-refl`, `↔-trans` etc. directly to `Function.Properties.Inverse` and then just use them directly?

src/Data/Vec/Recursive.agda
```diff
@@ -148,3 +155,18 @@ zip = zipWith _,_
...
+
+lift↔ : {A B : Set a} → ∀ n → A ↔ B → A ^ n ↔ B ^ n
+lift↔ 0 AB = ↔.refl
```
MatthewDaggitt, line 162: Very minor, but I think it would more readable to name `AB` as `A↔B`?

src/Data/Fin/Properties.agda
```diff
@@ -674,6 +682,14 @@ combine-surjective {m} {n} i with remQuot {m} n i | P.inspect (remQuot {m} n) i
   (uncurry remQuot-combine)
   (combine-remQuot {m} n)
 
+private module ↔ {ℓ} = IsEquivalence (↔-isEquivalence {ℓ})
+
+v↔Fin : ∀ m n → Fin (m ^ n) ↔ Fin m ^v n
```
MatthewDaggitt, line 687: Perhaps a slightly more meaningful name would be `Fin[m^n]↔Fin[m]^n`? Also could we import the `Vec` module qualified as `Vec` or something and then write `Fin m Vec.^ n` instead?

MatthewDaggitt (commented):
Also needs a CHANGELOG entry!

MatthewDaggitt:
Thanks for the PR! 

The final isomorphism could live in `Function.Properties`?

src/Data/Vec/Recursive.agda
guilhermehas: Yes. Without `with`-abstract, the termination checker does not work. If you know another solution to this problem, let me know.

gallais: `let open Inverse (Vec↔^ (suc n)) in (...)` is probably enough.

guilhermehas: I removed this part because I saw that it is already in the library.

src/Data/Vec/Recursive.agda
guilhermehas: I put this module in `Function.Properties.Inverse`.

src/Data/Vec/Recursive.agda
```diff
@@ -17,17 +17,24 @@
 module Data.Vec.Recursive where
 
 open import Level using (Level; lift)
+open import Axiom.Extensionality.Propositional using (Extensionality)
```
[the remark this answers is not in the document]
guilhermehas: I removed it.

src/Data/Fin/Properties.agda
guilhermehas: I changed the name and wrote `Vec.^` instead of `^v`.

guilhermehas:
> The final isomorphism could live in `Function.Properties`?

I put it in `Function.Properties.Inverses`, because it looks like a property of Inverses.

src/Function/Properties/Inverse.agda
```diff
@@ -48,6 +49,9 @@ isEquivalence = record
   ; trans = Composition.inverse
   }
 
+open module ↔ {ℓ} = IsEquivalence (↔-isEquivalence {ℓ})
+  renaming (refl to ↔refl; sym to ↔sym; trans to ↔trans) public
```
MatthewDaggitt, line 53: These names should be `↔-refl`, `↔-sym`, `↔-trans`?

src/Function/Properties/Inverse.agda
```diff
@@ -81,3 +85,13 @@ Inverse⇒Equivalence I = record
 
 ↔⇒⇔ : A ↔ B → A ⇔ B
 ↔⇒⇔ = Inverse⇒Equivalence
+
+module _ (ext : ∀ {a b} → Extensionality a b) where
+
+  ↔Fun : A ↔ B → C ↔ D → (A → C) ↔ (B → D)
```
MatthewDaggitt, line 91: Again this should be called `↔-fun`?

src/Data/Vec/Recursive.agda
```diff
@@ -148,3 +154,8 @@ zip = zipWith _,_
 
 unzip : ∀ n → (A × B) ^ n → A ^ n × B ^ n
 unzip = unzipWith id
+
+lift↔ : {A B : Set a} → ∀ n → A ↔ B → A ^ n ↔ B ^ n
```
MatthewDaggitt, line 158: Out of interest why do `A` and `B` need to be at the same level?

src/Data/Vec/Recursive.agda
```diff
@@ -17,17 +17,23 @@
...
+open import Data.Product.Algebra using (×-cong)
 open import Data.Sum.Base as Sum using (_⊎_)
 open import Data.Unit.Base using (tt)
 open import Data.Unit.Polymorphic.Base using (⊤)
 open import Data.Vec.Base as Vec using (Vec; _∷_)
+open import Data.Vec.N-ary using (N-ary)
 open import Function
 open import Relation.Unary
-open import Agda.Builtin.Equality using (_≡_)
+open import Agda.Builtin.Equality using (_≡_; refl)
```
MatthewDaggitt, line 34: This import can be removed in favour of the `Relation.Binary.PropositionalEquality.Core` immediately after?

CHANGELOG.md
```diff
@@ -2029,3 +2029,34 @@ This is a full list of proofs that have changed form to use irrelevant instance
...
+  ```
+
+* Added new isomorphisms to `Data/Vec/Recursive`:
+  ```agda
+  lift↔ : {A B : Set a} → ∀ n → A ↔ B → A ^ n ↔ B ^ n
+  ```
+
+* Added new functions to `Function/Properties/Inverse`:
+  ```agda
+  ↔refl ↔sym ↔trans
```
MatthewDaggitt, line 2056: Please can you update these names and add their types!

CHANGELOG.md
```diff
@@ -2029,3 +2029,34 @@ This is a full list of proofs that have changed form to use irrelevant instance
   Inverse⇒Injection : Inverse S T → Injection S T
   ↔⇒↣ : A ↔ B → A ↣ B
   ```
+
+* Added new isomorphisms to `Data/Fin/Properties`:
```
MatthewDaggitt, line 2033: Should use `.` rather than `/` in the module names here?

MatthewDaggitt (commented):
A few more minor changes, and then should be good to merge in!

src/Function/Properties/Inverse.agda
guilhermehas: I have already updated them.

src/Function/Properties/Inverse.agda
guilhermehas: I updated it.

src/Data/Vec/Recursive.agda
guilhermehas: I removed this line.

CHANGELOG.md
guilhermehas: I updated the definitions in CHANGELOG.

CHANGELOG.md
guilhermehas: I have already changed `/` for `.`

CHANGELOG.md
```diff
@@ -2029,3 +2029,34 @@ This is a full list of proofs that have changed form to use irrelevant instance
...
+  ```
+
+* Added new isomorphisms to `Data.Vec.Recursive`:
+  ```agda
+  lift↔ : ∀ n → A ↔ B → A ^ n ↔ B ^ n
+  ```
+
+* Added new functions to `Function.Properties.Inverse`:
+  ```agda
+  ↔-refl ↔-sym ↔-trans
```
MatthewDaggitt, line 2056: Please can we add types for these?

src/Function/Properties/Inverse.agda
```diff
@@ -48,6 +49,9 @@ isEquivalence = record
   ; trans = Composition.inverse
   }
 
+open module ↔ {ℓ} = IsEquivalence (↔-isEquivalence {ℓ})
```
MatthewDaggitt, line 52: We need to add a `using ()` clause otherwise we're unintentionally exporting a load of other junk from this module as well. Can you also quickly check the module `↔` isn't exposed outside of `Function.Properties.Inverse`? Can't remember if it is or not...

MatthewDaggitt (commented):
Otherwise looks good to go after those last minor changes.

CHANGELOG.md
guilhermehas: I have already added the types.

src/Function/Properties/Inverse.agda
guilhermehas: I have already put `using ()`.

src/Data/Fin/Properties.agda
```diff
@@ -674,6 +682,12 @@ combine-surjective {m} {n} i with remQuot {m} n i | P.inspect (remQuot {m} n) i
   (uncurry remQuot-combine)
   (combine-remQuot {m} n)
 
+Fin[m^n]↔Fin[m]^n : ∀ m n → Fin (m ^ n) ↔ Fin m Vec.^ n
```
MatthewDaggitt, line 685: Sorry, one last comment. I've just noticed that `Data.Vec.Recursive` depends on `Data.Fin.Base`. Therefore `Data.Vec.Properties` should also depend on `Data.Fin.Properties`. However, this PR introduces a dependency the other way around which is going to cause problems in future. To fix this the proof needs moving to `Data.Vec.Recursive`. (don't forget to update the CHANGELOG as well).

MatthewDaggitt:
Thanks for the PR!

PR #1789, opened 2022-07-20
Added new Fin Inequality

guilhermehas:
This new Fin inequality is that:
```agda
data _≤′_ : Rel (Fin n) 0ℓ  where
  ≤′-refl : ∀ {i : Fin n}                             → i ≤′ i
  ≤′-step : ∀ {i} {j : Fin n} (i≤′j : i ≤′ inject₁ j) → i ≤′ suc j
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2075,7 +2075,7 @@ This is a full list of proofs that have changed form to use irrelevant instance 
     í ≤′ inject₁ j → i ≤′ suc j
   _<′_ : Rel (Fin n) 0ℓ
   pattern <′-base          = ≤′-refl
-  pattern <′-step {n} m<′n = ≤′-step {n} m<′n
+  pattern <′-step {j} i<′j = ≤′-step {j} i<′j
   _≥′_ : Rel (Fin n) 0ℓ
   _>′_ : Rel (Fin n) 0ℓ
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2067,3 +2067,15 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   Inverse⇒Injection : Inverse S T → Injection S T
   ↔⇒↣ : A ↔ B → A ↣ B
   ```
+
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
What's the advantage of adding this? In general I'm a bit wary of adding new datatypes that re-encode existing definitions as it means we eventually end up duplicating a lot of proofs.

Also note that we already have an induction principle based on this datatype if it's termination you're worried about?

https://github.com/agda/agda-stdlib/blob/902226a8bdb8a35348de5e167460b354322b66fe/src/Data/Fin/Induction.agda#L51-L54

guilhermehas:
> What's the advantage of adding this? In general I'm a bit wary of adding new datatypes that re-encode existing definitions as it means we eventually end up duplicating a lot of proofs.

One possible advantage is to make this induction:
```agda
finRecursionIneq : ∀ {ℓ} {n : ℕ} {k : Fin (suc n)} (P : ∀ p → k ≤′ p → Set ℓ)
  (pk : P k ≤′-refl)
  (ps : ∀ {m k≤′m} → P (inject₁ m) k≤′m → P (suc m) (≤′-step k≤′m))
  → ∀ {m} k≤′m → P m k≤′m
finRecursionIneq P pk ps ≤′-refl = pk
finRecursionIneq P pk ps (≤′-step k≤′m) = ps (finRecursionIneq P pk ps k≤′m)
```

MatthewDaggitt:
> One possible advantage is to make this induction:

What do you mean by "making" the induction? Getting it to pass the termination checker? It looks to me like its the strong version of the `<-weakInduction` lemma I referenced above? You should be able to derive a similar (although slightly longer) proof for your result from that?

I know it seems like more work to prove it directly instead of using a custom datatype, but adding a new definition means that eventually all of the related lemmas invariably get added to the library as well, not to mention an ever expanding explosion of conversion lemmas. I think its better to stick with one canonical version of each relation.

MatthewDaggitt:
Going to close this, as I think we can get similar results without adding new datatypes. Apologies @guilhermehas!

PR #1790, opened 2022-07-22
Fix issue #1778

### LINKED ISSUE #1778: Expose container combinator conversion functions separately from their correctness proofs
#### Description
Currently `Data.Container.Combinator.Properties` exposes `Inverse` records which provide both conversion functions and their correctness proofs. Proofs for `const` and `_×_` combinators require extensionality and thus disallow usage of corresponding conversions, which don't actually need extensionality.

KislyjKisel:
Exposed container combinator conversion functions from `Data.Container.Combinator.Properties` separate from their correctness proofs in `Data.Container.Combinator`. Functions are named same as combinators prefixed with `to-` (to combinator) and `from-` (from combinator). Explicit arguments (besides the last one - value being converted) correspond to combinators' arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1840,6 +1840,34 @@ Other minor changes
   ∈-++↔++ : ∀ {v} xs ys → v ∈ xs ++ ys ↔ v ∈ ys ++ xs
   ```
 
+* Exposed container combinator conversion functions from `Data.Container.Combinator.Properties` separate from their correctness proofs in `Data.Container.Combinator`:
+  ```
+  to-id      : ∀ {s p x} {X : Set x} → F.id X → ⟦ id {s} {p} ⟧ X
+  from-id    : ∀ {s p x} {X : Set x} → ⟦ id {s} {p} ⟧ X → F.id X
+  to-const   : ∀ {s p y} (X : Set s) {Y : Set y} → X → ⟦ const {p = p} X ⟧ Y
+  from-const : ∀ {s p y} (X : Set s) {Y : Set y} → ⟦ const {p = p} X ⟧ Y → X
+  to-_∘_     : ∀ {s₁ s₂ p₁ p₂} (C₁ : Container s₁ p₁) (C₂ : Container s₂ p₂) →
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
@KislyjKisel apologies for your comments in #1778 slipping off my radar!

To answer them, I think the extracted lemmas should live in `Combinator.Properties` not `Combinators`. Also it's standard library convention to try and avoid infix notation in proofs. So I would suggest `to-id`, `to-∘` etc.

CHANGELOG.md
```diff
@@ -1840,6 +1840,34 @@ Other minor changes
   ∈-++↔++ : ∀ {v} xs ys → v ∈ xs ++ ys ↔ v ∈ ys ++ xs
   ```
 
+* Exposed container combinator conversion functions from `Data.Container.Combinator.Properties` separate from their correctness proofs in `Data.Container.Combinator`:
+  ```
+  to-id      : ∀ {s p x} {X : Set x} → F.id X → ⟦ id {s} {p} ⟧ X
```
MatthewDaggitt, line 1845: Also we tend to omit implicit arguments in the types in the CHANGELOG so that we give only the most important bits of the proof. For example this would become `F.id X → ⟦ id ⟧ X`

MatthewDaggitt (commented):
Otherwise looks great!

src/Data/Container/Combinator.agda
```diff
@@ -29,36 +29,68 @@ module _ {s p : Level} where
   id .Shape    = ⊤
   id .Position = F.const ⊤
 
+  to-id : ∀ {x} {X : Set x} → F.id X → ⟦ id ⟧ X
```
MatthewDaggitt, line 32: Oh sorry, one more thing. By convention we tend to use `a b` and `A : Set a`, `B: Set b` for universe levels and types.

KislyjKisel:
@MatthewDaggitt aren't these definitions analogous to conversions such as `fromℕ`?

MatthewDaggitt:
> @MatthewDaggitt aren't these definitions analogous to conversions such as fromℕ?

Possibly yes... Okay willing to keep them in the main file :+1: 

But I think I stand by my request to remove the use of `infix` notation, as the standard library doesn't tend to use it in this fashion, even for conversion functions!

MatthewDaggitt:
Thanks for the great PR!

PR #1767, opened 2022-05-08
Proofs of some linearity axioms.

capn-freako:
A definition of "linear map" along w/ some proofs of a few axioms of linearity.

PR #1791, opened 2022-07-24
Fixed whitespace errors.

capn-freako:
At Matthew's request.


capn-freako:
@MatthewDaggitt , why is this PR showing so many included commits?
There should only be one.
I pulled from upstream/master, after you accepted/merged PR #1767 ; I would've thought that doing so would prevent this sort of thing. Is there something else I need to do in addition?

gallais:
You probably started your new PR from the branch already containing the commit instead of from `master`.
Pulling the updated master is not enough: you need to to `checkout` to it too.

PR #1793, opened 2022-07-27
Additional DivMod proofs.

uncle-betty:
Hello there.

I would like to explore a contribution to the standard library. I have a few DivMod proofs that I use all the time and that I think others might find useful.

I'm not sure, though, whether my Agda skills and mathematical reasoning are sufficiently mature. So, feel free to just close the pull request, if the proofs don't meet the library's standards.

If the proofs are OK, then I'd like some guidance on two dependency issues, please. Lines 103 and 342 currently contain commented-out code:

  * Line 103 is in the % section of the file, but requires a proof from further down in the / section of the file.
  * Line 342 requires a proof from `Data.Nat.Divisibility`, which leads to a circular dependency.

Are there any standard ways to deal with dependency issues like this?

The proofs type-check with Agda 2.6.2.2 as well as with current `master`.


uncle-betty:
Oh, wait! What was I thinking! The proof from `Data.Nat.Divisibility` is trivial. I fixed line 342 simply by inlining the proof. Which gets rid of the circular dependency. I pushed the change.

Which leaves the ordering issue in line 103, which refers to a proof further down in the file.

uncle-betty:
Hmmm. I'm now having second thoughts about the ordering of the other proofs as well. Maybe it would be wiser to create a new file for these more complex proofs that depend on the simpler proofs from `Data.Nat.DivMod`? Something like `Data.Nat.DivMod.Properties`?

As I'm already asking for guidance, please allow for another question: Where does the standard library draw the line between what's to be included and what isn't?

As an example, I just added aonther potential proof that served me quite well: `[m*n+o]%[p*n]≡[m*n]%[p*n]+o`. (Yes, it could probably use a more concise name.) This particular proof has been useful for me when dealing with base-n numbers represented as fixed-length vectors. It basically says that you can pull out the least significant digit from under the `%` prescribed by the vector length. So, it's a little more niche than the other proofs. Would something like this still be interesting enough to be included in the standard library? Err on the side of including something? Or rather err the other way around?

Sorry for all the questions. I'm just trying to better understand how the standard library structures things and what goes into it and what doesn't.

uncle-betty:
FWIW, I went and checked the Coq standard library, and it does have variants of the last proof I added. I guess it's not just me who found it useful.

See `mul_mod_distr_l` in https://coq.inria.fr/distrib/current/stdlib/Coq.Numbers.NatInt.NZDiv.html .

src/Data/Nat/DivMod.agda
```diff
@@ -67,6 +75,34 @@ m%n%n≡m%n m (suc n-1) = modₕ-idem 0 m n-1
   (m + n)           % n ≡⟨ [m+n]%n≡m%n m n ⟩
   m                 % n ∎
 
+n≤m⇒[m∸n]%n≡m%n : ∀ {m n} → n ≤ m → .⦃ _ : NonZero n ⦄ →
```
MatthewDaggitt, line 78: Can we move all the `NonZero` arguments further forward in the types, i.e. before `n ≤ m` in this case?

MatthewDaggitt: Also by convention, we usually introduce the variables in alphabetical order in the name for predictability, e.g. `m≤n⇒[n∸m]%m≡n%m`

src/Data/Nat/DivMod.agda
```diff
@@ -255,6 +294,78 @@ m*n/m*o≡n/o m@(suc m-1) n o = helper (<-wellFounded n)
     n / o                         ∎
     where n∸o<n = ∸-monoʳ-< (n≢0⇒n>0 (≢-nonZero⁻¹ o)) (≮⇒≥ n≮o)
 
+m<n*o⇒m/o<n : ∀ {m n o} → m < n * o → .⦃ _ : NonZero o ⦄ → m / o < n
+m<n*o⇒m/o<n {m} {suc n} {o} m<n*o with m <? o
+... | yes m<o = begin-strict
+  m / o ≡⟨ m<n⇒m/n≡0 m<o ⟩
+  0     <⟨ s≤s z≤n ⟩
+  suc n ∎
+... | no ¬m<o = begin-strict
```
MatthewDaggitt, line 303: Very minor but can we use the unicode `≮`?

src/Data/Nat/DivMod.agda
uncle-betty: Ah, I see. Yes. Absolutely.

src/Data/Nat/DivMod.agda
```diff
@@ -67,6 +75,34 @@ m%n%n≡m%n m (suc n-1) = modₕ-idem 0 m n-1
...
+  m % n           ∎
+
+n*o≤m⇒[m∸n*o]%o≡m%o : ∀ {m n o} → n * o ≤ m → .⦃ _ : NonZero o ⦄ →
+                      (m ∸ n * o) % o ≡ m % o
+n*o≤m⇒[m∸n*o]%o≡m%o {m} {n} {o} n*o≤m = begin-equality
+  (m ∸ n * o) % o         ≡˘⟨ [m+kn]%n≡m%n (m ∸ n * o) n o ⟩
+  (m ∸ n * o + n * o) % o ≡⟨ cong (_% o) (m∸n+n≡m n*o≤m) ⟩
+  m % o                   ∎
+
+m%[n*o]%o≡m%o : ∀ m n o → .⦃ _ : NonZero o ⦄ → .⦃ _ : NonZero (n * o) ⦄ →
```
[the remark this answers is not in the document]
uncle-betty: Oh, yes. Thanks. We can and we should.

src/Data/Nat/DivMod.agda
uncle-betty: Right. That would be consistent with things like `≮⇒≥`.

src/Data/Nat/DivMod.agda
```diff
@@ -255,6 +294,78 @@ m*n/m*o≡n/o m@(suc m-1) n o = helper (<-wellFounded n)
...
+[m∸n*o]/o≡m/o∸n : ∀ m n o → .⦃ _ : NonZero o ⦄ →
+                  (m ∸ n * o) / o ≡ m / o ∸ n
+[m∸n*o]/o≡m/o∸n m zero    o = refl
+[m∸n*o]/o≡m/o∸n m (suc n) o = begin-equality
+  (m ∸ (o + n * o)) / o ≡˘⟨ /-congˡ (∸-+-assoc m o (n * o)) ⟩
+  (m ∸ o ∸ n * o) / o   ≡⟨ [m∸n*o]/o≡m/o∸n (m ∸ o) n o ⟩
+  (m ∸ o) / o ∸ n       ≡⟨ lem ⟩
+  m / o ∸ suc n         ∎
+  where
+  lem : (m ∸ o) / o ∸ n ≡ m / o ∸ suc n
```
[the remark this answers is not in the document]
uncle-betty: Oh, yes. I took the liberty to modify it a little and turned it into `m*n/o*n≡m/o`, to mirror the existing `m*n/m*o≡n/o`.

src/Data/Nat/DivMod.agda
```diff
@@ -274,6 +385,67 @@ m*n/m!≡n/[m∸1]! : ∀ m n .{{_ : NonZero m}} →
...
+  m % n * o                         ≡⟨ cong (_* o) (m%n≡m∸m/n*n m n) ⟩
+  (m ∸ m / n * n) * o               ≡⟨ *-distribʳ-∸ o m (m / n * n) ⟩
+  m * o ∸ m / n * n * o             ≡⟨ cong (λ # → m * o ∸ # * n * o) lem ⟩
+  m * o ∸ m * o / (n * o) * n * o   ≡⟨ cong (m * o ∸_) (*-assoc (m * o / (n * o)) n o) ⟩
+  m * o ∸ m * o / (n * o) * (n * o) ≡˘⟨ m%n≡m∸m/n*n (m * o) (n * o) ⟩
+  m * o % (n * o)                   ∎
+  where
+  instance o*n≢0 = subst NonZero (*-comm n o) n*o≢0
+
+  lem : m / n ≡ m * o / (n * o)
```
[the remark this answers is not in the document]
uncle-betty: Oh, yes. I took the liberty to modify it a little and turned it into `m*n/o*n≡m/o`, to mirror the existing `m*n/m*o≡n/o`.

src/Data/Nat/DivMod.agda
```diff
@@ -274,6 +385,67 @@ m*n/m!≡n/[m∸1]! : ∀ m n .{{_ : NonZero m}} →
...
+  lem = begin-equality
+    m / n           ≡˘⟨ m*n/m*o≡n/o o m n ⟩
+    o * m / (o * n) ≡⟨ /-congʳ (*-comm o n) ⟩
+    o * m / (n * o) ≡⟨ /-congˡ (*-comm o m) ⟩
+    m * o / (n * o) ∎
+
+[m*n+o]%[p*n]≡[m*n]%[p*n]+o : ∀ m {n o} p → ⦃ _ : NonZero (p * n) ⦄ → o < n →
+                              (m * n + o) % (p * n) ≡ (m * n) % (p * n) + o
+[m*n+o]%[p*n]≡[m*n]%[p*n]+o m {n} {o} (suc p) ⦃ p*n≢0 ⦄ o<n = begin-equality
```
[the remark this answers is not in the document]
uncle-betty: Oh, yes! Clever! That bothered me, too.

uncle-betty:
Thank you for your encouragement and guidance!

I just pushed the changes. Let's see how much progress we made towards a final version.

uncle-betty:
Oh, wait. One more thing, in line 97: `m*n≤o⇒[o∸m*n]%n≡o%n {m = o / pm * p} lem`, where Agda cannot infer the implicit argument `m`.

Is the universe telling me here to make `m` an explicit parameter instead?

uncle-betty:
Oh! I seem to have caused a name clash with `m/n/o≡m/[n*o]` in `Data.Nat.Divisibility`.

I guess I should pick a different name for my proof, right? I assume that this is preferable over changing the proof's name in `Data.Nat.Divisibility`, as this would break existing proofs.

uncle-betty:
Regarding the implicit `m` argument: I looked around and the [Pattern Matching section of Inference in Agda](https://htmlpreview.github.io/?https://github.com/effectfully/inference-in-agda/blob/master/InferenceInAgda.html#pattern-matching) seems to suggest that `m`, in fact, cannot be inferred: `m` only occurs in `m * n`, and `*` pattern-matches on its LHS, i.e., on `m`.

I'll make `m` explicit. Which leaves the naming conflict.

src/Data/Nat/DivMod.agda
```diff
@@ -274,6 +400,58 @@ m*n/m!≡n/[m∸1]! : ∀ m n .{{_ : NonZero m}} →
...
+[m*n+o]%[p*n]≡[m*n]%[p*n]+o m {n} {o} p@(suc p∸1) ⦃ p*n≢0 ⦄ o<n = begin-equality
+  (mn + o) % pn           ≡⟨ %-distribˡ-+ mn o pn ⟩
+  (mn % pn + o % pn) % pn ≡⟨ cong (λ # → (mn % pn + #) % pn) (m<n⇒m%n≡m lem₁) ⟩
+  (mn % pn + o) % pn      ≡⟨ m<n⇒m%n≡m lem₃ ⟩
+  mn % pn + o             ∎
+  where
+  mn = m * n
+  pn = p * n
+
+  lem₁ : o < pn
```
MatthewDaggitt, line 436: `m<n⇒m<o*n` is probably worthy of it's own lemma in `Data.Nat.Properties`?

src/Data/Nat/DivMod.agda
```diff
@@ -255,6 +299,88 @@ m*n/m*o≡n/o m@(suc m-1) n o = helper (<-wellFounded n)
...
+  (m ∸ o) / o ∸ n       ≡⟨ cong (_∸ n) ([m∸n]/n≡m/n∸1 m o) ⟩
+  m / o ∸ 1 ∸ n         ≡⟨ ∸-+-assoc (m / o) 1 n ⟩
+  m / o ∸ suc n         ∎
+
+m/n/o≡m/[n*o] : ∀ m n o .⦃ _ : NonZero n ⦄ .⦃ _ : NonZero o ⦄
+                .⦃ _ : NonZero (n * o) ⦄ → m / n / o ≡ m / (n * o)
+m/n/o≡m/[n*o] m n o = begin-equality
+  m / n / o                             ≡⟨ /-congˡ {o = o} (/-congˡ (m≡m%n+[m/n]*n m n*o)) ⟩
+  (m % n*o + m / n*o * n*o) / n / o     ≡⟨ /-congˡ (+-distrib-/-∣ʳ (m % n*o) lem₁) ⟩
+  (m % n*o / n + m / n*o * n*o / n) / o ≡⟨ cong (λ # → (m % n*o / n + #) / o) lem₂ ⟩
```
MatthewDaggitt, line 356: Just to say I really like the use of `#` :smile:

uncle-betty:
And again it shows that the hardest problem in computer science is... naming things!

Yes, the `m/n/o≡m/[n*o]` from `Data.Nat.Divisibility` is a bit of an imposter and should probably be called `m/n/o≡m/[n*o]-terms-and-conditions-may-apply` because of its divisibility precondition, but I guess that backward compatibility trumps that.

So I renamed mine to `m/n/o≡m/[n*o]-nsa` (no strings attached) for now, just to make CI go through. But then again, having "NSA" in a code base will likely start rumors of conspiracies.

Can anybody else think of a reasonable name?

src/Data/Nat/DivMod.agda
uncle-betty: Thanks! It makes it stick out, doesn't it? Much easier to see where the substitution happens, I think.

MatthewDaggitt:
> I guess I should pick a different name for my proof, right? I assume that this is preferable over changing the proof's name in Data.Nat.Divisibility, as this would break existing proofs that depend on it.

So we're actually in luck! Unless I'm getting confused, it looks like your version is strictly more general than the one in `Data.Nat.Divisibility` right?  In which case we can take advantage of the fact that we're preparing for `v2.0` and break compatibility, by removing the one in `Data.Nat.Divisibility`. As long as we note it in the CHANGELOG, under the minor breaking changes section, it should be fine!

Otherwise, apart from that and the small comment above, it all looks good. All that's left is to add some entries in the CHANGELOG, describing the new functions.

src/Data/Nat/DivMod.agda
uncle-betty: Yes, good point! For addition, we have things like ` ≤-stepsʳ` and friends. I'll add proofs for multiplication.

MatthewDaggitt: Let's not call it `steps` though, I really hate that name!

uncle-betty:
Oh, great! Yes, I also think that the new version is strictly more general than the one in `Data.Nat.Divisibility`.

I'll remove the old version and I'll update the CHANGELOG.

Thanks again for you help!

src/Data/Nat/DivMod.agda
uncle-betty: Yes. `m<n⇒m<o*n` is definitely more descriptive.

uncle-betty:
Alright. I made the changes. I hope that I did the CHANGELOG modifications correctly.

MatthewDaggitt:
CHANGELOG entries look great. Merging in, thanks again for the PR!

uncle-betty:
Woohoo! Awesome! Thank you for the merge. I'm so happy!

PR #1798, opened 2022-08-07
Added a new ≤-Induction

guilhermehas:
I defined my ≤-Induction without needing to create a new inequality for that.


src/Data/Fin/Induction.agda
```diff
@@ -59,6 +59,20 @@ open WF public using (Acc; acc)
...
     where i<i+1 = ℕ<⇒inject₁< (i<1+i i)
 
+≤-Induction : {k : Fin (suc n)} (P : ∀ {p} → k ≤ p → Set ℓ)
+  (pk : P ≤-refl)
+  (ps : ∀ {m k≤m} → P {inject₁ m} k≤m → P (≤-inj→suc k≤m))
+  → ∀ {m} k≤m → P {m} k≤m
+≤-Induction {k = k} P pk ps = induct (<-wellFounded _) (compare' k _)
+  where
+  induct : ∀ {m} → Acc _<_ m → Ordering' k m → ∀ k≤m → P {m} k≤m
+  induct {suc m} (acc rs) (less (s≤s k≤m)) _ = subst P (≤-irrelevant _ _) (ps finB)
```
JacquesCarette, line 69: Since `≤-irrelevant` does hold, there should be an analog to `cast` (for `Fin`) that computes, unlike `subst`. I was able to make various proofs `subst`-free because of that.

guilhermehas: What is your idea to remove `subst`?
I thought about changing `P {m} k≤m` to `Σ[ k P {m} k≤m` to `Σ[ k≤m₂ ∈ k ≤ m ] (P m k≤m₂)`.

JacquesCarette: I was starting to realize that my explanation of how to replace `subst` here was 90% of the work to actually do it, so maybe I should do that instead.

src/Data/Fin/Base.agda
```diff
@@ -310,6 +310,19 @@ compare (suc i) (suc j) with compare i j
 ... | greater greatest least = greater (suc greatest) (suc least)
 ... | equal   i              = equal   (suc i)
 
+data Ordering' (i : Fin n) : Fin n → Set where
```
MatthewDaggitt, line 313: I'm pretty sure this is morally an instantiation of the existing `Trichotomous` type from `Relation.Binary.Definitions` (exported by `Relation.Binary`) and that `compare` is the existing `<-cmp` proof from `Data.Fin.Properties`?

Can we us those instead?

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +59,20 @@ open WF public using (Acc; acc)
   induct {suc i} (acc rec) = Pᵢ⇒Pᵢ₊₁ i (induct (rec (inject₁ i) i<i+1))
     where i<i+1 = ℕ<⇒inject₁< (i<1+i i)
 
+≤-Induction : {k : Fin (suc n)} (P : ∀ {p} → k ≤ p → Set ℓ)
+  (pk : P ≤-refl)
+  (ps : ∀ {m k≤m} → P {inject₁ m} k≤m → P (≤-inj→suc k≤m))
+  → ∀ {m} k≤m → P {m} k≤m
```
MatthewDaggitt, line 65: Could you make sure the code conforms to the [indentation style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) please? Oh and the section on variable names!

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +59,20 @@ open WF public using (Acc; acc)
   induct {suc i} (acc rec) = Pᵢ⇒Pᵢ₊₁ i (induct (rec (inject₁ i) i<i+1))
     where i<i+1 = ℕ<⇒inject₁< (i<1+i i)
 
+≤-Induction : {k : Fin (suc n)} (P : ∀ {p} → k ≤ p → Set ℓ)
```
MatthewDaggitt, line 62: Hmm so I've been trying to work out what this is actually doing! So you start at `P k` and then prove `P (k+1)`, `P (k+2)` etc. etc. right?

It feels like parameterising the predicate `P` by the inequality is overkill. Could you reformulate the property as:

```agda
≤-Induction : (P : Pred (Fin (suc n)) ℓ) →
              ∀ {k : Fin (suc n)} → P k →
              (ps : ∀ {m} → m ≥ k → P (inject₁ m) → P (suc m)) →
              ∀ {m} → m ≥ k → P m
```
which I feel is simpler and more natural?

src/Data/Fin/Properties.agda
```diff
@@ -456,6 +456,10 @@ inject₁ℕ≤ = ℕₚ.<⇒≤ ∘ inject₁ℕ<
 ℕ<⇒inject₁< : ∀ {i : Fin (ℕ.suc n)} {j : Fin n} → j < i → inject₁ j < i
 ℕ<⇒inject₁< {i = suc i} (s≤s j≤i) = ≤̄⇒inject₁< j≤i
 
+≤-inj→suc : i ≤ inject₁ j → i ≤ suc j
```
MatthewDaggitt, line 459: Could we name this `i≤inject₁[j]⇒i≤1+j` please?

MatthewDaggitt (commented):
Thanks for revisiting this! Yes, this approach looks more promising :smile: Added a few comments.

src/Data/Fin/Induction.agda
guilhermehas: I indented and put better variable names.

src/Data/Fin/Induction.agda
guilhermehas: I used your suggestion.

src/Data/Fin/Properties.agda
guilhermehas: I renamed it.

src/Data/Fin/Base.agda
guilhermehas: I changed my code to use them instead.

guilhermehas:
I don't understand why CI is failing. I received this message:
```
Warning: Unknown/unsupported 'ghc' version detected (Cabal 3.6.2.0 supports
'ghc' version < 9.4): /usr/local/.ghcup/bin/ghc is version 9.4.1
```

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +60,20 @@ open WF public using (Acc; acc)
   induct {suc i} (acc rec) = Pᵢ⇒Pᵢ₊₁ i (induct (rec (inject₁ i) i<i+1))
     where i<i+1 = ℕ<⇒inject₁< (i<1+i i)
 
+≤-Induction : ∀ (P : Pred (Fin (suc n)) ℓ) {i} →
```
MatthewDaggitt, line 63: Okay so we should come up with a good name for this lemma! It's weak induction from an arbitrary starting point. The obvious `<-weakInduction-arbitraryStart` is a bit of a mouthful though? Do you have any better ideas?

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +60,20 @@ open WF public using (Acc; acc)
...
+              P i →
+              (∀ {i} → P (inject₁ i) → P (suc i)) →
+              ∀ {j} → j ≥ i → P j
+≤-Induction P {i} Pi Pᵢ⇒Pᵢ₊₁ {j} j≥i = induct (<-wellFounded _) (<-cmp i j) j≥i
```
MatthewDaggitt, line 67: Also we should be able to redefine `<-weakInduction` in terms of this right?

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +60,20 @@ open WF public using (Acc; acc)
...
+  where
+  induct : ∀ {j} → Acc _<_ j → Tri (i < j) (i ≡ j) (i > j) → j ≥ i → P j
+  induct {suc j} (acc rs) (tri< (s≤s i≤j) _ _) _ = Pᵢ⇒Pᵢ₊₁ finB
```
MatthewDaggitt, line 70: Very minor, but can we put this case with the additional `where` clause last? Makes it easier to read I think.

src/Data/Fin/Induction.agda
```diff
@@ -59,6 +60,20 @@ open WF public using (Acc; acc)
...
+    where
+    finB = induct (rs _ (s≤s (subst (ℕ._≤ toℕ j) (sym $ toℕ-inject₁ j) ≤-refl)))
```
MatthewDaggitt, line 72: `finB` isn't a great variable name. Maybe something like `P[1+j]`? Also could we split the term out into some slightly more meaningful named subparts? For example `sym $ toℕ-inject₁ j` is duplicated twice.

MatthewDaggitt:
> I don't understand why CI is failing. I received this message:

Yup, neither do I (see https://github.com/agda/agda-stdlib/issues/1804). It's not your fault!

src/Data/Fin/Induction.agda
guilhermehas: I modified the name, but I didn't think about a better name.

src/Data/Fin/Induction.agda
guilhermehas: I put it in the last case.

src/Data/Fin/Induction.agda
guilhermehas: I modified the name and created a new one for `sym $ toℕ-inject₁ j`.

src/Data/Fin/Induction.agda
guilhermehas: Yes, it is now using it.

src/Data/Fin/Induction.agda
MatthewDaggitt: Okay, I've settled on `<-weakInduction-startingFrom` and have made the change myself.

MatthewDaggitt:
Thanks for the great PR @guilhermehas! Merging in now.

PR #1802, opened 2022-08-19
Add IO.Base.untilRight

### LINKED ISSUE #1801: Loop combinator for IO
agda-stdlib has one combinator for loops, `untilJust`, but it doesn't allow loops to be stateful (without some `IORef`). The following `untilRight` seems a natural enough generalization:

```
untilRight : ∀ {A B : Set} → (A → IO (A ⊎ B)) → A → IO B
untilRight f x = bind (♯ f x) λ where
  (inj₁ x′) → ♯ untilRight f x′
  (inj₂ y) → ♯ return y
```

Does that seem like a good addition to agda-stdlib or did I miss some other existing construct?

(Related is #1065 which is also about recursion in IO but seems much more heavyweight.)

Lysxia:
Close #1801

### DIFF
--- a/src/IO/Base.agda
+++ b/src/IO/Base.agda
@@ -12,6 +12,7 @@ open import Level
 open import Codata.Musical.Notation
 open import Data.Bool.Base using (Bool; true; false; not)
 open import Agda.Builtin.Maybe using (Maybe; nothing; just)
+open import Data.Sum.Base using (_⊎_; inj₁; inj₂)
 import Agda.Builtin.Unit as Unit0
 open import Data.Unit.Polymorphic.Base
 open import Function.Base using (_∘′_; const; flip)
@@ -128,3 +129,8 @@ untilJust : IO (Maybe A) → IO A
 untilJust m = bind (♯ m) λ where
   nothing  → ♯ untilJust m
   (just a) → ♯ return a
+
+untilRight : {A B : Set a} → (A → IO (A ⊎ B)) → A → IO B
+untilRight f x = bind (♯ f x) λ where
+  (inj₁ x′) → ♯ untilRight f x′
+  (inj₂ y) → ♯ return y
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1691,6 +1691,7 @@ Other minor changes
 
   whenJust  : Maybe A → (A → IO ⊤) → IO ⊤
   untilJust : IO (Maybe A) → IO A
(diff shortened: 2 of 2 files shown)


src/IO/Base.agda
```diff
@@ -128,3 +128,8 @@ untilJust : IO (Maybe A) → IO A
 untilJust m = bind (♯ m) λ where
   nothing  → ♯ untilJust m
   (just a) → ♯ return a
+
+untilRight : ∀ {A B : Set} → (A → IO (A ⊎ B)) → A → IO B
```
gallais, line 132:
```suggestion
untilRight : (A → IO (A ⊎ B)) → A → IO B
```

This will:
* rely on the `variable` block instead of quantifying locally
* make the definition level polymorphic

gallais (approved):
You'll need a CHANGELOG entry too

MatthewDaggitt:
Thanks for the PR! Looks good to me. 

Apologies, merging is blocked at the moment by test suite build problems (https://github.com/agda/agda-stdlib/issues/1804).

Lysxia:
Actually there is a universe error that I'm not sure how to solve

```
/home/sam/code/agda/agda-stdlib/src/IO/Base.agda:134,24-29
Set (suc A.a) is not less or equal than Set (suc B.b)
when checking that the expression ♯ f x has type ∞ (IO _B_320)
```

src/IO/Base.agda
```diff
@@ -128,3 +129,8 @@ untilJust : IO (Maybe A) → IO A
 untilJust m = bind (♯ m) λ where
   nothing  → ♯ untilJust m
   (just a) → ♯ return a
+
+untilRight : (A → IO (A ⊎ B)) → A → IO B
```
gallais, line 133: Ah, I guess you can't get away with not having `A B` explicitly bound because then they're introduced
at distinct set levels and so the `IO` in the assumption and the one in the conclusion are not at compatible
levels.

You could either have
```suggestion
untilRight : {A B : Set a} → (A → IO (A ⊎ B)) → A → IO B
```
or
```suggestion
untilRight : {A : Set a} → (A → IO (A ⊎ B)) → A → IO (Lift a B)
```

MatthewDaggitt:
Thanks for the great first PR @Lysxia! It's much appreciated. Merging in now.

(Just a heads up that no need to force push in general, pushing is fine. We squash all the commits down before merging.)

PR #1805, opened 2022-08-23
[ fix #1804 ] always insist on installing cabal & ghc

### LINKED ISSUE #1804: Build failure in test suite due to too advanced GHC
We've started getting the following failure when running the test suite:

> Warning: Unknown/unsupported 'ghc' version detected (Cabal 3.6.2.0 supports
'ghc' version < 9.4): /usr/local/.ghcup/bin/ghc is version 9.4.1
Resolving dependencies...

I'm a bit confused about how version 9.4.1 is sneaking in as in `ubuntu.ci` we have:
https://github.com/agda/agda-stdlib/blob/d6d6087cf4da0261239937e213b7ebe9fcd8796c/.github/workflows/ci-ubuntu.yml#L49

and indeed in the previous step `Install alex & happy` we are using 8.10.7

gallais:


.github/workflows/ci-ubuntu.yml
```diff
@@ -112,8 +112,7 @@ jobs:
 ## INSTALLATION STEPS
 ########################################################################
 
-      - name: Install cabal
-        if: steps.cache-cabal.outputs.cache-hit != 'true'
+      - name: Install ghc & cabal
```
MatthewDaggitt, line 115: Hmm this feels like we're throwing the baby out with the bath water. 99% of the time the cache hit will be good and we'll save a lot of work right? Is the syntax expressive enough that we can write something like:
```
if: steps.cache-cabal.outputs.cache-hit != 'true' or run `ghc --version` != ${{ env.GHC_VERSION }}
```
No worries if not!

MatthewDaggitt (commented):
Ah nice diagnosing!

.github/workflows/ci-ubuntu.yml
gallais: We still keep the cache, we just insist on downloading & installing the specific ghc & cabal no matter what.
We should not be recompiling anything further down the action as long as the cache hits.

MatthewDaggitt: Okay, I was more thinking about the expense of installing GHC and cabal themselves, but if its negligible then :+1:

PR #1810, opened 2022-09-01
Move raw bundles from Data.X.Properties to Data.X.Base

### LINKED ISSUE #1755: Move "raw" bundles to `Data.X.Base`
Currently the raw bundles, e.g. `RawMagma` are defined in `Data.X.Properties` but they should really be defined in `Data.X.Base` as they don't require any properties.

5xy1:
fixes #1755


Taneb:
@5xy1 if you're up to it, could you also move `RawLattice` from `Algebra.Lattice.Bundles` to a new module `Algebra.Lattice.Bundles.Raw`?

5xy1:
@Taneb Of course, I'd be happy to! I might update this PR tonight.

src/Algebra/Bundles.agda
```diff
@@ -11,6 +11,7 @@
 
 module Algebra.Bundles where
 
+open import Algebra.Bundles.Raw public
```
MatthewDaggitt, line 14: We usually move `public` imports out of the main import list and into their own commented section of the file to draw attention to them. e.g.

https://github.com/agda/agda-stdlib/blob/bc9c1b6a117fcb86b321113c0958ffc3b3526b4e/src/Text/Regex.agda#L26-L30

Any chance you could change this and other ones?

src/Algebra/Bundles/Raw.agda
```diff
@@ -0,0 +1,289 @@
...
+--
+-- Definitions of 'raw' bundles
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Bundles.Raw where
+
+open import Algebra.Core
+open import Algebra.Structures
```
MatthewDaggitt, line 12: We shouldn't need `Algebra.Structures`? And can probably get away with `Relation.Binary.Core` as well?

src/Algebra/Bundles/Raw.agda
```diff
@@ -0,0 +1,289 @@
...
+open import Relation.Binary
+open import Function.Base
+import Relation.Nullary as N
```
MatthewDaggitt, line 15: Could just import `¬_` directly, rather than using this qualified?

src/Algebra/Lattice/Bundles/Raw.agda
```diff
@@ -0,0 +1,35 @@
...
+{-# OPTIONS --without-K --safe #-}
+
+module Algebra.Lattice.Bundles.Raw where
+
+open import Algebra.Core
+open import Algebra.Bundles
+open import Algebra.Structures
+open import Algebra.Lattice.Structures
+open import Level using (suc; _⊔_)
+open import Relation.Binary
```
MatthewDaggitt, line 16: Again `Relation.Binary.Core` and we shouldn't need the `Structures`?

CHANGELOG.md
```diff
@@ -75,6 +75,32 @@ Non-backwards compatible changes
...
+#### Moved raw bundles from Data.X.Properties to Data.X.Base
+
+* As mentioned by MatthewDaggitt in Issue #1755, Raw bundles defined in Data.X.Properties
+  should be defined in Data.X.Base as they don't require any properties.
+  * Moved raw bundles From `Data.Nat.Properties` to `Data.Nat.Base`
+  * Moved raw bundles From `Data.Nat.Binary.Properties` to `Data.Nat.Binary.Base`
+  * Moved raw bundles From `Data.Rational.Properties` to `Data.Rational.Base`
+  * Moved raw bundles From `Data.Rational.Unnormalised.Properties` to `Data.Rational.Unnormalised.Base`
+
+#### Moved the definition of RawX from `Algebra.X.Bundles` to `Algebra.X.Bundles.Raw`
```
MatthewDaggitt, line 87: This is not a breaking change right? It can go in the `Major improvements` section :smile:

MatthewDaggitt (commented):
Thanks, looks great! Just a few more small comments and then it'll be ready to merge!

MatthewDaggitt:
Ooh also @5xy1 we'll need to update [README.Design.Heirarchies](https://github.com/agda/agda-stdlib/blob/master/README/Design/Hierarchies.agda) with an entry for `Bundles.Raw`. Do you think you could do so? Happy to do it otherwise if you're not comfortable with doing so.

5xy1:
@MatthewDaggitt I'm afraid of missing something, so I would be grateful if you could do it!

PR #1811, opened 2022-09-06
Remove remaining map-with-∈  names

### LINKED ISSUE #1575: Remove remaining `map-with-∈` names in favour of the new `mapWith∈`
While going through https://github.com/agda/agda-stdlib/pull/1569 I noticed that some of the proofs for `mapWith∈` are still using the old name `map-with-∈`

https://github.com/agda/agda-stdlib/blob/faa5a3cb7fde2f14cdfbd3b75ab8d08bbdc5e047/src/Data/List/Relation/Unary/Any/Properties.agda#L587

To be honest I think `mapWith∈` is a pretty hideous name anyway, even if it no longer uses hyphens. I propose to rename it to `mapMembers` in v2.0. Any thoughts or objections?

5xy1:
Fixes #1575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -723,6 +723,18 @@ Deprecated names
   zipWith-identityʳ  ↦  zipWith-zeroʳ
   ```
 
+* In `Data.List.Relation.Binary.Subset.Propositional.Properties`:
+  ```
+  map-with-∈⁺  ↦  mapWith∈⁺
+  ```
+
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
Thanks @5xy1! Any chance you could:
1. add deprecation warnings to the old names in the same style that you changed in the `Subset.Propositional.Properties` file?
2. add a CHANGELOG entry detailing the deprecations?

MatthewDaggitt:
Perfect. Thank you very much!

PR #1812, opened 2022-09-06
Add `flip′` to `Function.Base`

XiaohuWang0921:
Add a non-dependent version of `flip` to `Function.Base`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1949,6 +1949,12 @@ Other minor changes
   from-Σ     : (I : Set i) (C : I → Container s p) → ⟦ Σ I C ⟧ A → ∃ λ i → ⟦ C i ⟧ A
   ```
 
+* Added a non-dependent version of `Function.Base.flip` due to an issue noted in
+  Pull Request #1812:
+  ```agda
+  flip′ : (A → B → C) → (B → A → C)
+  ```
+
 NonZero/Positive/Negative changes
 ---------------------------------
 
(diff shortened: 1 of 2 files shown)


Taneb:
I'm not opposed to adding this, but do you know of a situation where the dependent version doesn't work?

XiaohuWang0921:
It happened when I tried to `flip` an overloaded constructor in a complex enough context. My original code looks something like this:
```agda
open import Data.Nat.Base
open import Data.Fin.Base
open import Data.Vec.Base hiding (map; alignWith)
open import Data.Maybe.Base
open import Data.These.Base hiding (map; alignWith)
open import Function.Base

data Lazy (n : ℕ) : Set where
  var : Fin n → Lazy n
  abs : Lazy (suc n) → Lazy n
  app : ∀ {k} → Vec (Lazy n) k → Lazy k → Lazy k → Lazy n

data Dummy : Set where
  app : Dummy

private
  variable
    m n k : ℕ

subs : Vec (Lazy n) m → Lazy m → Lazy n
subs = {!!}

allStep : Vec (Lazy n) k → Maybe (Vec (Lazy n) k)
allStep = {!!}

step : Lazy n → Maybe (Lazy n)
step (app shared s @ (var _) t) = alignWith (foldWithDefaults shared t (flip {-flip′-} app s)) (allStep shared) (step t)
step _ = {!!}
```
And Agda complains about unsolved constraints at `flip` (I am using version 2.6.2.2). Remove the `Dummy` type, and the complaint disappears; try `flip app` in a simpler function, and the complaint also disappears.

Taneb:
OK, great! Could you add a note about this to the changelog?

MatthewDaggitt:
Merging in, thanks for the PR @XiaohuWang0921

PR #1813, opened 2022-09-08
Fix doc typo in README.Data.List.Membership

ahubers:
The documentation reads
> For example if we want to reason about membership for `List ℕ`
> then you would use the `DecSetoidMembership` as we use ...

but then the passage later says:
> the module `DecPropMembership` should be opened as follows:

I am guessing that these are meant to both be `DecPropMembership`, as equality of `ℕ` is propositional and decidable.


MatthewDaggitt:
Thanks, nice spot!

PR #1817, opened 2022-09-12
Fix missing fixities in some files (#207)

duinomaker:
## For `infixr 6 _<∣>_` in /Data/Maybe/Base.agda

```
_<∣>_ : Maybe A → Maybe A → Maybe A
just x  <∣> my = just x
nothing <∣> my = my
```

`_<∣>_` should be right associative, similar to `Data.Sum._⊎_`,

```
case₁ : ∀ (x y : Maybe A) → x <∣> y <∣> nothing ≡ x <∣> y
case₁ (just x) y = refl
case₁ nothing (just y) = refl
case₁ nothing nothing = refl
```

and since `_<∣>_` cannot connect `List`'s (e.g. `just 1 ∷ [] <∣> nothing ∷ []`), it is safe to give it a precedence greater than 5 to make the following case possible without parentheses.

```
case₂ : List (Maybe ℕ)
case₂ = just 1 <∣> nothing ∷ []
```

## For `infixl 0 _!|>_` in /Function/Strict.agda

Looks like the first `_!|>′_` is a misspelling of `_!|>_`.

```
infixl 0 _!|>′_ _!|>′_
```


MatthewDaggitt:
Thanks for the PR! Looks great, except that you've added the CHANGELOG entry to v1.7 which has already been released. Please could move the entry to the current `CHANGELOG.md` at the top-level of the directory?

duinomaker:
> except that you've added the CHANGELOG entry to v1.7 which has already been released

Sorry, I didn't notice! I have moved the entry to the current `CHANGELOG.md` in the updated commits.

MatthewDaggitt:
Thanks again!

Issue #1427, opened 2021-02-17
Redefine as many numeric operations in terms of primitive ones as possible


MatthewDaggitt:
Whilst discussing #581 it came up that several operations in `Data.Nat.Base` still evaluate using unary arithmetic. In version 2.0 it may be worth fixing these...
  - `_⊔_`
  - `_⊓_`
  - `∣_-_∣`

andreasabel:
Do you need more builtins on the Agda side for this?  (There are no NATMAX, NATMIN etc. builtins atm.)

gallais:
I think the plan is to use `_<ᵇ_`

PR #1818, opened 2022-09-13
Add primop backed versions of natural-number operations

Taneb:
One possible solution for #1427 

I add versions of min, max, and dist that use primitive operations so they compile to fast code (hopefully - I admit I haven't checked this yet). I also add proofs that they match their recursively defined counterparts, and a proof about distance that I felt was missing.


MatthewDaggitt:
Sorry for taking so long to review this. Looks fine to me. Can you add a CHANGELOG entry?

Taneb:
Sure, I'll add a changelog entry in the next few days

Issue #626, opened 2019-02-19
Use `-local` consistently for properties


gallais:
In `Data.List.Properties` we use

* `map-id₂ :  All (λ x → f x ≡ x) xs → map f xs ≡ xs`
* `map-cong₂ : All (λ x → f x ≡ g x) xs → map f xs ≡ map g xs`
* (we do not have `map-compose₂`)

I think that we should instead use the `-relative` suffix introduced in #510

MatthewDaggitt:
What does `relative` refer to?

gallais:
The fact the equality constraint is relative to a specific domain (in order to be
as tight a constraint as possible) rather than absolute.

MatthewDaggitt:
Hmm looking at this again, I'd probably prefer the prefix -locally (or -local) instead of  -relative, as in _the equality constraints hold locally_.  Thoughts?

JacquesCarette:
To be honest, I don't like either -locally or -relative.  Isn't this rather a 'lifting' property? You're lifting an equality to hold at a list. In some ways, this is a generalization of `ap`.

MatthewDaggitt:
So `map-id-lift`? I like it better than -relative but less than -local. I'd be fine with it though....

MatthewDaggitt:
Going to leave this out of v1.0, as I don't think we've reached consensus yet.

MatthewDaggitt:
@jamesmckinna thinks the suffix should be `at` e.g. `map-id-at`

After discussion we're going with `local`

jamesmckinna:
Is it the case, for `List`s at least, that the 'global' property `map-id` is (definitionally equal to; or merely equivalent to?) `\forall xs \r (map-id-local xs)` (modulo the actual arguments in order to fix the property)?

That is, do we uniformly have that `property` is `All (property-local)` for `property`s of interest? Should we?

jamesmckinna:
@MatthewDaggitt this could also be linked to issue #509 (or at least, to a single PR which will knock them both off). You or me?

jamesmckinna:
I think this bit of PR #1822 is now ready.

jamesmckinna:
I was too hasty. There is a very large raft of `updateAt-*` properties, each with a `-relative` version, in:
- [ ] `Data.Vec.Properties`
- [ ] `Data.Vec.Functional.Properties`
- [ ] `Data.List.Relation.Unary.All.Properties`
Turning to those now.

[...]

Issue #509, opened 2018-10-20
Inconsistency: `map-compose` vs. `map-∘`


gallais:
In [`Data.List.Properties`](https://agda.github.io/agda-stdlib/Data.List.Properties.html#2896) we use `map-compose` whereas in [`Data.Vec.Properties`](https://agda.github.io/agda-stdlib/Data.Vec.Properties.html#5621) we use `map-∘`.

gallais:
And in `Codata.*` we use `map-map-fusion`.

jamesmckinna:
Suggest `map-∘` throughout?

gallais:
:+1:  That is consistent with the convention we use

gallais:
I see that we also have [map-++-commute](https://agda.github.io/agda-stdlib/Data.List.Properties.html#3178) instead of `map-++`.

jamesmckinna:
And similarly [sum-++-commute](https://agda.github.io/agda-stdlib/Data.List.Properties.html#20000). 
I'll open a PR.

jamesmckinna:
Fixed them in `Data.Maybe.Properties`.

JacquesCarette:
I wonder if `map-++-commute` might have been my doing. I like that name... but I'm completely fine with `map-++` if that's the library convention.

jamesmckinna:
@JacquesCarette I take your point about 'small' PRs: but #626 and #509 are about consistency/uniformity of names *according to a systematic library convention* (whose stipulation to a large extent is neither here nor there; consistency is the main consideration here); the case of `map-id(entity)` was, at least on my reading, of the same order of need for consistency, as that of `map-∘` which started this issue for milestone 2.0...

I'm certainly open to the possibility that `Codata.*` properties might better want distinct names, to draw attention to the fact that 
* in the case of `-identity`, the relation is that of bismilarity, not equality
* in the case of `-∘`, that the direction of the equation is opposite to that in `Data.*`
For this reason, I have stopped pushing commits regarding the `Codata` changes.

JacquesCarette:
@jamesmckinna it is indeed a philosophical disagreement. You see this PR as achieving consistency/uniformity as a complete unit, while I'd be quite happy with a sequence of PRs whose union would achieve that. Moving closer to the goal is worthwhile, even if the ultimate goal (for whatever reason) is not ultimately achieve for 2.0. [I use this philosophy quite strongly in my Drasil project.]

[...]

Issue #501, opened 2018-10-17
Deprecate  `pos-distrib-*` in favour of `pos-*-commute`?


gallais:
In `Data.Integer.Properties`,

```agda
pos-distrib-* : ∀ x y → (+ x) * (+ y) ≡ + (x ℕ* y)
```

is the symmetric of the *not yet existing* function

```agda
pos-*-commute : ℕtoℤ.Homomorphic₂ +_ _ℕ*_ _*_
```

where `module ℕtoℤ = Morphism.Definitions ℕ ℤ _≡_`.

Names based on the *existing* function

```agda
abs-*-commute : ℤtoℕ.Homomorphic₂ ∣_∣ _*_ _ℕ*_
```

where `module ℤtoℕ = Morphism.Definitions ℤ ℕ _≡_`

mechvel:
I vote for the suggestion by @gallais. 
Also `pos-*-commute`  do exist in lib-1.5. 
Also even if it is `pos-distrib-*` then at least the equality parts in the signature need to be swapped.

jamesmckinna:
But what about the general injunction against spurious additional `-commute` suffixes, as per issue #509 ?

JacquesCarette:
So `pos-*` and `abs-*` ?

PR #1822, opened 2022-09-15
Resolves issue #509 and resolves issue #626

jamesmckinna:
First things first: resolving issue #626 for `Data.List.Properties`
- [x] changed 'relative' properties to `-local`
- [x] this needs to be noted in `CHANGELOG` 

Secondly, resolving issue #509 going through and making `map-∘` properties etc. have consistent names across the library
- [x] `Data.List.Properties`
- [x] what/where else? see below, esp. as regards `Codata.*`


jamesmckinna:
And what do we do about these?
The relevant property uses the *opposite* direction of the bisimilarity relation, because `map-map-fusion` is meant to simplify/create opportunities for simplification, by pushing the function composition inwards. So do we rename to `map-∘` or keep as is?
- [ ] `Codata.Guarded.Stream.Properties`
- [ ] `Codata.Sized.M.Properties`
- [ ] `Codata.Sized.Delay.Properties`
- [ ] `Codata.Sized.Covec.Properties`
- [ ] `Codata.Sized.Colist.Properties`

That said, for consistency's sake, we could nevertheless rename `map-identity` in each to `map-id`.
How to proceed?

JacquesCarette:
You might want to tackle this more incrementally (i.e. don't do `map-id` changes in this PR).

My impression is that changes like these do show up in CHANGELOG.

Don't know about `Codata`.

jamesmckinna:
Re: incremental changes
as in my comment on the (in my mind, at least) coupled issues, I decided to roll the changes up together in a single PR. 
Changing `map-identity` to `map-id` is arguably a separate change, but theres' no knock-on viscosity in std-lib, so only clients' code need pay attention to the deprecation... and it's still only a deprecation warning; ergo, harmless?

JacquesCarette:
The benefit of small PRs is that if there's a problem with one thing, most of the pieces can still be merged in, while if it's all in one big one, perfection must be attained before progress happens.

jamesmckinna:
Thanks @gallais for the prompt about `Codata.*`. Similarly change `map-identity` to `map-id`? But I think at this stage I would not change the order of the equalities. But it's probably tomorrow or Monday now before I'll tackle these in any case...

jamesmckinna:
I propose marking this as `status:ready`, while noting that resolving #509 might also entail another renaming to resolve #501.

MatthewDaggitt:
Looks great, thanks for the PR @jamesmckinna !

PR #1830, opened 2022-09-20
Cleaning up dependencies on `Data.Fin` and `Data.Nat` ahead of a first attempt to fix Issue1686

jamesmckinna:
Along the lines of the 'worker/wrapper' distinction I made in the discussion of the issue. 

Outstanding task(s): 
- [ ] `CHANGELOG`, as ever; 
- [ ] `Algebra.Properties.CommutativeMonoid.Sum`
- [ ] `Data.Fin.Base`
- [ ] `Data.Fin.Properties`
- [ ] `Data.List.Properties`
- [ ] `Data.Vec.Base`
- [ ] `Data.Vec.Functional`
- [ ] `Data.Vec.Properties`
- [ ] `Data.Vec.Recursive`
- [ ] `Data.Vec.Relation.Unary.All`
- [ ] `Data.Vec.Relation.Unary.Any`
- [ ] `Data.Vec.Relation.Unary.Linked.Properties`

EDITED: ~~These are now both pending, given Matthew's comments below.~~
UPDATED: the above task list shows all (at least) the files needing to be touched by the PR
I'd like to merge the existing changes first before proceeding to working on them.


MatthewDaggitt:
Thanks @jamesmckinna! 

Unfortunately this PR is doing too many different things at once to easily review. It's doing straightforward things like deprecations (great!) and dependency simplification (also great!) and then also has the more controversial non-zero changes which require some significant thought. In order to make this easier to review and avoid the former from being blocked by the latter, please could you separate them out into separate PRs?

You may find the `git cherry-pick` **or** the `git rebase -i` commands useful in doing so...

MatthewDaggitt:
As an aside, the names of the PRs might be improved by referencing what it actually does as well as the issue number.

jamesmckinna:
Matthew, thanks for the comments, which I will attempt to take on board.
The easiest path to progress right now would seem to be to remove the additional commits of `NZ*` material, and recommit as a stable PR which prepares the ground for subsequent investigations, perhaps after further discussion with you, and/or experiments of your own to confirm my suspicions that, while the `{{_ : Nonzero n}}` versions of operations are indeed definable, they cannot be done entirely smoothly as *replacements* to what is there already, because the typechecker will complain that the goal types are ill-formed, until, in the actual function bodies, the pattern `{n = suc n}` ensures everything goes as before.

MatthewDaggitt:
> The easiest path to progress right now would seem to be to remove the additional commits of NZ* material, and recommit as a stable PR which prepares the ground for subsequent investigations, perhaps after further discussion with you, and/or experiments of your own to confirm my suspicions that, while the {{_ : Nonzero n}} versions of operations are indeed definable, they cannot be done entirely smoothly as replacements to what is there already, because the typechecker will complain that the goal types are ill-formed, until, in the actual function bodies, the pattern {n = suc n} ensures everything goes as before.

Sounds like a plan!

jamesmckinna:
Hmmm. I think that this is now OK, but I am slightly worried that the previous commits to `agda/master` which needed to have their merge conflicts resolved, are now part of this PR? Ooops, if so, and I'll try again later. 

Otherwise, this PR is the 'part I' alluded to above: namely groundwork preparation (mostly dependency thinning; and no (?) new functions or deprecations) ahead of the more controversial part II, where `NonZero`-bracketed functions and their properties get propagated throughout. Towards such a move, all the imports of `Data.Nat{.Base}` now explicitly import `NonZero` and `pred`, ahead of their eventual (heavy) use to implement Matthew's proposal. But further discussion of that required, it would seem.

jamesmckinna:
Following f2f discussion with @MatthewDaggitt today, my revised proposal is to:
* retain existing functionality as 'worker' functions, *with their existing names*, reflecting that they are the 'more dependent' (because their types involve *inhomogeneous* telescopes eg `{n : ℕ} (i : Fin (suc n))`)
* in place of the original, ugly, `NZ` prefix to the above, write 'wrapper' functions with the same names, *but primed*, for the corresponding *homogeneous* operations prefixed on a `{{NonZero n}` instance (because 'less dependent' here = homogeneously telescoped)
I'll push those commits another day, once this proposed naming convention has settled (?).

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -551,13 +551,11 @@ applyDownFrom⁺₂ f n Pf = applyDownFrom⁺₁ f n (λ _ → Pf _)
 ------------------------------------------------------------------------
 -- tabulate
 
-tabulate⁺ : ∀ {n} {f : Fin n → A} →
-            (∀ i → P (f i)) → All P (tabulate f)
+tabulate⁺ : ∀ {n} {f} → (∀ i → P (f i)) → All P (tabulate {n = n} f)
```
MatthewDaggitt, line 554: So I appreciate the ideological purity at work here in that you should get the type-checker to infer the maximum amount of information that it can! The problem is that we don't have good UI tools to recover that information for the user. I think these changes significantly increase cognitive load on the user, as there is no mouse-over or one single command to recover what the type of `f` is.

So in summary I would probably prefer these type erasure changes to reverted? Thoughts?

src/Data/Fin/Base.agda
```diff
@@ -234,6 +236,13 @@ _ℕ-ℕ_ : (n : ℕ) → Fin (suc n) → ℕ
 n     ℕ-ℕ zero   = n
 suc n ℕ-ℕ suc i  = n ℕ-ℕ i
 
+-- n ℕ-suc "i" = n ∸ (suc i)
+
+infixl 6 _ℕ-suc_
+
+_ℕ-suc_ : (n : ℕ) → Fin n → ℕ
+suc n ℕ-suc i = n ℕ-ℕ i
```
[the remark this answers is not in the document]
jamesmckinna: Matthew, 
this function is not currently used, but since at various points in the library there are uses of its definitional equivalent `n ∸ suc (toℕ i)` I thought it worth lifting out as something in its own right; but admittedly in obscure corners:
* `Data.Fin.Properties.opposite-prop`, `Data.Fin.Properties.opposite-suc`
* `Data.List.Properties.lookup-applyDownFrom`, `Data.List.Properties.lookup-downFrom`

So very easy to prune this, and its properties, back. But it *is* an example of a 'naturally'-occurring homogenously-telescoped function on `Fin n`, so in the spirit of this PR.

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: Thanks for "ideological purity" ;-)
Agree that there is choice here between redundant recoding of 'types-as-machine-checkable-comments' versus "don't import where you don't have to". And as for UI support, indeed using `C-c C-d` to type an expression is fairly limited in applicability...

The trouble is, `tabulate` shouldn't ever have to say *what* is being tabulated (and so, perhaps, neither should the user ever have to think about it, either; in any mode of use where a lemma such as `tabulate⁺` is being deployed, the user presumably already has an `f` in mind which is being `tabulate`d); as for `tabulate` itself, it (un)fortunately *does* need to know the length `n` of the 'table', so any instinct of mine towards 'purity' indeed led me in such directions.

jamesmckinna: I don't want to make this a hill on which to take an extreme stand, but I will now go onto do precisely that...;-) 

I wonder sometimes about exactly how we make the trade-off between 'cognitive load' (which I do care about, believe me) and 'notational clutter' (which, especially if it can be avoided by avoiding expensive imports, perhaps should be...). 

"Redundant recoding" is usually seen in a positive light in the Green/Blackwell "cognitive dimensions" school of HCI-of-PL, but I do wonder in this instance: why give `f` a type, and not `i`? why go to the trouble here, when in the sections on `applyUpTo` and `applyDownFrom` in this module, there is no such fastidiousness?

jamesmckinna: There are other examples eg `Data.List.Membership.Setoid.Properties`, where a module's `import`s rename a `Setoid`'s  `Carrier` to `A`, only so that it can then be mentioned in the type of a `tabulate`d `f`... etc. 

Maybe `tabulate` *is* special, but I don't really see why...

src/Data/Fin/Base.agda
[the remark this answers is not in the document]
MatthewDaggitt: Okay, in which case could you make use of it in this PR to simplify the code that you mention? 

My deepest apologies if this was previously in this PR before and you have to add it back in. I'm just playing the part of Goldilocks. Previously the PR was too big to review, but now it contains too small in that it contains this partial change that doesn't stand on its own. In general, we tend to like each PR to stand on its own as people (including me!) promise to do things, but don't actually come back and do it! Not implying that you will, but as a general rule these things happen! Therefore we like each PR to be an obvious standalone improvement over the existing code. Adding a function that is never used doesn't quite pass this test.

src/Data/List/Relation/Unary/All/Properties.agda
MatthewDaggitt: > Thanks for "ideological purity" ;-)

Will you believe me if I say that was honestly meant as a positive thing? :sweat_smile: I am always fascinated by the principles raised by you, most of which have never even crossed my mind.... Apologies if it came across otherwise! 

> Agree that there is choice here between redundant recoding of 'types-as-machine-checkable-comments' versus "don't import where you don't have to". 

So this is saving us an import of `Fin` from `Data.Fin.Base` right? The thing is that that doesn't actually help us in any practical way. The dependencies we care about are module level dependencies, i.e. do we import `Data.Fin.Base` at all. The `using` keyword doesn't speed up type-checking or interface loading in Agda. It's purely to avoid naming conflicts and for a certain amount of documentation purposes. The code is still using `Fin` whether we make it implicit or explicit.

> The trouble is, tabulate shouldn't ever have to say what is being tabulated (and so, perhaps, neither should the user ever have to think about it, either;)

I think you may be coming at this from the perspective of an expert user who knows automatically what something called `tabulate` does! A new user, such as myself a few years back, has no idea what the word "tabulate" means and has to piece it together from the implementation and the type.

> but I do wonder in this instance: why give f a type, and not i?

So we have a fixed set of [naming conventions](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#variables-1) in the library (of which `Fin` should be there as `i`, `j`, `k` but isn't for some reason). These conventions are there for exactly the reason that if the user sees an `i` then they should know what type it is. However, there is obviously no convention for `f` beyond it being some sort of function....

> why go to the trouble here, when in the sections on applyUpTo and applyDownFrom in this module

I think that `applyUpTo` and `applyDownFrom` probably _should_ follow this convention, but we've already had bigger fish to fry :cry:

src/Data/Fin/Base.agda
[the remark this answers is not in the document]
MatthewDaggitt: Don't mind, happy either way. Whichever you'd prefer!

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: Re: purity
I had taken it in a positive vein... I just hadn't quite realised that it came across as 'hardline'

jamesmckinna: Re: 'expert user'

hmmm... I think it's fair to say I'm 'expert', but not necessarily expert as an Agda user, or as a std-lib contributor. 

But even 'novice' users are explicitly invited to use mouse-middle-click to examine definitions... so I'd expect them to do so if they wanted to understand tabulate... which they really should perhaps want to do before examinining its properties here...?

jamesmckinna: Re: naming conventions

Again a fair cop, except as above: f is only the thing-tabulated... so to know anything about f is only to want to know how tabulate works...

.. ie naming conventions giving unambiguous cues as to typing will get you only so far; but the very context-sensitivity of typing (as opposed to naming) is what allows/would/should allow     (the types of) names to be unambiguously interpreted/understood... one of the payoffs of uniqueness of typing?

MatthewDaggitt: > Other commits in this PR do achieve the aim of eliminating imports completely, but I agree this one does not (might it? Not sure, those zero/suc constructors seem indispensable alas)

I think impossible. Even if we somehow eliminate it syntactically, the module will still transitively depend on it. 

> But even 'novice' users are explicitly invited to use mouse-middle-click to examine definitions... so I'd expect them to do so if they wanted to understand tabulate... which they really should perhaps want to do before examinining its properties here...?

Yes fair enough. But I think every hint we can give people helps. Getting actual documentation is definitely part of the story as well!

> Again a fair cop, except as above: f is only the thing-tabulated... so to know anything about f is only to want to know how tabulate works...

Yes, but doesn't mean that you can't go the other way either, i.e. remember what tabulate does from what `f` is.

To be honest there's also another reason I'm reluctant to make the changes as well. The `{n = n}` notation is seriously ugly :disappointed:

jamesmckinna: Pushed.

PR #1831, opened 2022-09-21
Be a little tighter with imports

Taneb:
This was prompted by me noticing that building `Data.Nat.Binary.Base` was pulling in `Algebra`. This change reduces its dependency graph by quite a lot.


JacquesCarette:
Oh, this is excellent! I love these kinds of PRs.

Issue #588, opened 2019-01-13
Inconsistent convention: lookup's order of arguments


gallais:
Here is the current situation:

```agda
lookup : (xs : List A) → Fin (length xs) → A
lookup : Fin n → (xs : Vec A n) → A
```

I think we should adopt the list-style `lookup` because of the nice symmetry with
`tabulate : (Fin n -> A) -> Vec A n`

JacquesCarette:
Yesterday, in my Haskell course, I gave my students the following rule of thumb: put the arguments to a function in the order of varies-little to varies-lots, from a call-site perspective.  For `lookup`, one will tend to look up multiple positions from a single container more often than a single position from multiple containers. It's a heuristic that agrees quite a lot with the order in the Prelude.  And it works here too.

[I should note that certain design patterns in Haskell programming go in the opposite direction, in particular the shift from explicitly threading in some 'options' record versus using a Reader monad for same, where a step-wise refactor of code written in least-varying style will first require that this parameter be shift from front to back before being absorbed in the Reader monad].

gallais:
In `Data.Vec.Relation.Unary.All`, the type of `lookup` uses a `Fin` whereas
in `Data.List.Relation.Unary.All`, the type of `lookup` uses a `Membership` proof:

```
lookup : ∀ {n} {xs : Vec A n} (i : Fin n) →
         All P xs → P (Vec.lookup xs i)


lookup : All P xs → (∀ {x} → x ∈ₚ xs → P x)
```

I could see moving the `Vec` one to `.Properties` and calling it `lookup⁺`.

andreasabel:
The version for `Vec` has also a counterpart for `List`:
```agda
lookup : {xs : List A} -> All P xs -> (i : Fin (length xs)) -> P (lookup xs i)
```
I think both the `Fin` version and the `Membership` version are useful.

MatthewDaggitt:
We decided we should go with the list order as standard everywhere, and also follow @gallais's suggestion for `lookup⁺`

MatthewDaggitt:
Reopening as 93f5adb only fixed the naming issue. Did not fix the original argument ordering issue.

gallais:
Oof there are more of them in `Codata` :scream:

PR #1832, opened 2022-09-21
[ re #588 ] More lookup cleanup

gallais:


MatthewDaggitt:
Thanks, looked like a big job :sweat_smile:

Issue #1437, opened 2021-02-24
Fix equality combinator names in `Data.Rational.Unnormalised.Properties.≤-Reasoning` 


MatthewDaggitt:
Currently the combinators use `≈` instead of `≃`

JacquesCarette:
I don't know how to rename syntax declarations?

jamesmckinna:
Ha! Is this an Agda issue, that syntax declarations have 'global' scope once imported/else are not Renamable like other imports... AFAICT, 'pattern' syntax isn't renamable either...?

MatthewDaggitt:
Yes, it is an Agda problem, and it sucks. 

See the last bullet in the [v1.3 CHANGELOG](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.3.md#changes-to-how-equational-reasoning-is-implemented) for how we rename syntax.

PR #1836, opened 2022-09-29
moves forward on issue #1437

JacquesCarette:
Renames the functions, but not the syntax.

### DIFF
--- a/src/Data/Rational/Unnormalised/Properties.agda
+++ b/src/Data/Rational/Unnormalised/Properties.agda
@@ -507,6 +507,7 @@ module ≤-Reasoning where
     <-≤-trans
     ≤-<-trans
     public
+    renaming (step-≈ to step-≃; step-≈˘ to step-≃˘)
 
 ------------------------------------------------------------------------
 -- Properties of ↥_/↧_


MatthewDaggitt:
Sorry, for not replying earlier. I have replied on the issue for how to rename the syntax.

JacquesCarette:
Thanks - hopefully I'll have time (and wifi) on the long train ride tomorrow to deal with this.

jamesmckinna (commented):
i think that following @MatthewDaggitt 's suggestions regarding `syntax` declarations, the following should do the trick instead of your `renaming` directive:
```agda
module ≤-Reasoning where
  import Relation.Binary.Reasoning.Base.Triple
    ≤-isPreorder
    <-trans
    <-resp-≃
    <⇒≤
    <-≤-trans
    ≤-<-trans
    as Triple
  open Triple public hiding (step-≈; step-≈˘)

  infixr 2 step-≃ step-≃˘
  step-≃  = Triple.step-≈
  step-≃˘ = Triple.step-≈˘
  
  syntax step-≃  x y∼z x≃y = x ≃⟨  x≃y ⟩ y∼z
  syntax step-≃˘ x y∼z y≃x = x ≃˘⟨ y≃x ⟩ y∼z
```
with this a model for any subsequent downstream changes which may be required. 

EDITED The 'problem' with a mere `renaming `/why it does not work, is that the syntax declaration *and the infixes* are declared in a different scope, and then agda's parser complains... so one needs to be more belt-and-braces about redclaring the names, their infix status, and their syntax... groan.

jamesmckinna:
superseded now by #1916

PR #1839, opened 2022-09-30
 [ fix #1443 ] Fin & Vec casts and properties

### LINKED ISSUE #1443: Missing `cast` properties
Realised tonight we don't have these (and their equivalents for `Vec`)

```agda
cast-is-id : ∀ {m} (eq : m ≡ m) (k : Fin m) → cast eq k ≡ k
cast-is-id eq Fin.zero    = refl
cast-is-id eq (Fin.suc k) = cong Fin.suc (cast-is-id (suc-injective eq) k)

cast-is-subst : ∀ {m n} (eq : m ≡ n) (k : Fin m) → subst Fin eq k ≡ cast eq k
cast-is-subst refl k = sym (cast-is-id refl k)
```

Note that it would also be nice to prove commutation lemmas for subst &
constructors for the various inductive types we have.

gallais:


jamesmckinna:
Just one query about the type
```
cast-is-subst : cast eq k ≡ subst Fin eq k
```
The name of this term (going from left-to-right) is consistent with the other `cast-is-*` terms. 

But to me, the type seems the wrong way round? 

If the general aim of equations in the std-lib is to 'reduce complex expressions to simpler', then in this instance, and for all the reasons previously discussed for preferring `cast` to `subst`, then shouldn't this equation be oriented from `subst` in favour of `cast`?, ie
```
subst-is-cast : subst Fin eq k ≡ cast eq k
```
If not, I'd love to be have that spelt out. 

Either way, I'm otherwise happy with this PR as it stands.

JacquesCarette:
I agree with @jamesmckinna on the 'wanted' direction of that one lemma, and on this otherwise being ready.

Issue #10, opened 2014-02-28
Generic n-ary programs (zipWith, alignWith, lift, etc.)


mechvel:
I wonder of wherher the library needs some generalized  zipWith-n  functions.
For example, the below function is actually  zipWith3,
but one list is by   List  and others are by  List.All:
```agda
   zip3with? : (triples : List (C × ℕ × C))  →  All.All ^eq triples  →
                    All.All (\e → e > 0) $ map tuple32 triples  →  List Item 
    zip3with? [] _ _  =  []
    zip3with? ((p , e , d) ∷ triples) (d=p^e ∷a eqs) (e>0 ∷a restExps>0) =
                                                      item ∷ (zip3with? triples eqs restExps>0)
                                                     where
                                                     item = record{ ... } 
```

MatthewDaggitt:
So off the top of my head I'm struggling to come up with a generalised `zipWith-n` function which is both sufficiently general and easy to reason about. Do you have such an implementation?

I don't think `zip3with` is common or useful enough to warrant being included in the standard library.

MatthewDaggitt:
Closing as I'm afraid I can't come up with a suitable `zipWith-n` function.

gallais:
I think this ought to be doable using the code in #608

jespercockx:
How about
```agda
open import Level using (Level)
open import Function
open import Function.Nary.NonDependent
open import Data.Nat
open import Data.Product using (_×_; _,_; proj₁; proj₂) renaming (curry to curry₂; uncurry to uncurry₂)
open import Data.List using (List; []; _∷_; map) renaming (zipWith to zipWith₂)

variable
  l : Level
  R : Set l

zipWith : ∀ n {ls} {as : Sets n ls} →
          Arrows n as R → Arrows n (List <$> as) (List R)
zipWith zero f = []
zipWith (suc zero) f = map f
zipWith (suc (suc n)) f xs₀ xs₁ = zipWith (suc n) id (zipWith₂ f xs₀ xs₁)
```
or alternatively
```agda
zipWith : ∀ n {ls} {as : Sets n ls} →
          Arrows n as R → Arrows n (List <$> as) (List R)
zipWith zero f = []
zipWith (suc zero) f = map f
zipWith (suc (suc n)) f xs₀ xs₁ = zipWith (suc n) (uncurry₂ f) (zipWith₂ _,_ xs₀ xs₁)
```

gallais:
[...]

PR #1840, opened 2022-09-30
[ re #10 ] n-ary functions manipulating lists

gallais:


jamesmckinna:
I don't know what the library conventions are about pointfree vs. pointwise function composition, but how about 
```
zipWith (suc n@(suc _)) f (v , vs) (w , ws) =
  f zero v w , zipWith n (f ∘ suc) vs ws
```
versus your
```
zipWith (suc n@(suc _)) f (v , vs) (w , ws) =
  f zero v w , zipWith n (λ k → f (suc k)) vs ws
```

gallais:
I had started with that alternative but it does not typecheck.

jamesmckinna:
Hmmm. I begin to see why: the type of `_∘_` would need to be generalised so that its `Level` arguments `b`,`c` were made dependent (resp. on `(x : A)`, `{x : A} (y : B y)`... and even then the typechecker really gets confused, trying to unify `Levelₙ lsas k ⊔ Levelₙ lsbs k ⊔ Levelₙ lscs k` with `Levelₙ lsas k ⊔ Levelₙ lsbs k ⊔ Levelₙ lscs k`... so consider my query withdrawn... ;-)

... but I can't help wondering what would value dependent levels look like, and how should we expect them to behave?

Issue #952, opened 2019-10-19
Subclassing in the `Category` hierarchy


laMudri:
Currently, the `Category` hierarchy principally contains the definitions `RawFunctor`, `RawApplicative`, and `RawMonad`. The `RawMonad` module contains a definition `rawApplicative`, showing that from a `RawMonad`, one can get a `RawApplicative`. In turn, the `RawApplicative` module contains a definition `rawFunctor`. Furthermore, the members of `rawApplicative` and `rawFunctor` are reexported, meaning that `RawMonad` exports definitions of `_<$>_`, `pure`, and `_⊛_`.

These exported definitions are defined in terms of `return` and `_>>=_`. This means that often, a user will write implementations for all three structures, but their `RawFunctor` will have a definitionally different implementation of `_<$>_` than their `RawMonad` gives rise to.

It is perhaps useful to be able to derive functors and applicatives just from the monad data, but it also causes problems. There needs somewhere to be a definition of `RawMonad` containing a `RawApplicative` field, which itself contains a `RawFunctor` field. This would provide more coherence to the hierarchy.

MatthewDaggitt:
So the non-indexed versions should not be defined in terms of the indexed ones.

gallais:
I've started working on this. I'm tempted to additionally drop the exotic unicode combinators for `Applicative` and
`Alternative` in favour of their haskell counterparts.

gallais:
Hmmm this is really annoying.

In #636 I defined `MonadZero` and `MonadPlus` by combining `Monad` with
`ApplicativeZero` and `Alternative` respectively. But now that `Monad` packs
an `Applicative`, we have a diamond problem with two potentially unrelated
instances of `Applicative` contained in each.

But rolling back #636 to have once more duplicated `empty` and `_<|>_` in
both the applicative & monad hierarchies is not really satisfactory as we'll
really easily get ambiguities.

So what I'm considering is a parallel hierarchy with record packing `empy : F A`
alone first and combined with  `_<|>_ : F A -> F A -> F A` second.

But I need names for these.
Let's say Empty & Sum as placeholder for now

```
Empty = empty
Sum = Empty + _<|>_

ApplicativeZero = Applicative + Empty
Alternative = Applicative + Sum

MonadZero = Monad + Empty
MonadPlus = Applicative + Sum
```

Intuitively I was thinking `Pointed` for `Empty` but from what I can read for functors
[...]

Issue #356, opened 2018-07-04
`Function`'s `case_return_of_` clashes with `RawMonad`'s `return`


gallais:
It is very unfortunate that as soon as you write `open RawMonad m` to be able to write
some monadic code, you get a parse error on `case_return_of_`.

I suggest changing the definition of `RawIMonad` to a `RawIApplicative` + `_>>=_` and
to explicitly deprecate `return` in favour of `pure`.

JacquesCarette:
`return` to `pure` makes perfect sense.  Deduplification of definitions is often a good thing, so the other half of this seems like a good idea as well. But it could reduce usability in non-trivial ways... [nest record structures are a pain. Creating them in a modular way is nice, but the flattening them before offering them to users is a much nicer interface.]

gallais:
> flattening them before offering them to users is a much nicer interface

In the standard library we systematically open the module associated to a nested
record which means that all of its fields are brought in scope whenever someone
opens the enclosing module.

For instance [`IsPartialOrder` opens `IsPreorder`](https://agda.github.io/agda-stdlib/Relation.Binary.html#3681).

Is that not sufficient?

JacquesCarette:
Almost. For access, it is. For creation, AFAIK, you still have to use the nested structure. Or, at least, that's what ctrl-R always gives me.

gallais:
Ah I see. I suppose we could also think about having smart constructors taking a
`pure` and a `bind` and generating the whole `RawIMonad` record given that the rest
can be derived automatically.

MatthewDaggitt:
In https://github.com/agda/agda-stdlib/pull/2157 this was fixed by renaming `case_return_of` to `case_returning_of`.

PR #1841, opened 2022-10-03
[ re #952, #356, #939 ] Overhaul of Functor/Applicative/Monad

gallais:


JacquesCarette:
Wow, that's impressive. That does indeed seem like quite the slog.

gallais:
Oof, finally writing a test case and I'm in level-hell because of `IO`.
I'm experimenting with polymorphising but it is tempting to use an escape hatch to
force `IO : Set a -> Set a`.

gallais:
Alright, I think this is approaching readiness. Sorry about the 3k+ LoC diff. :weary:

MatthewDaggitt:
Looks amazing @gallais. Thanks so much! Had a read through it and can't spot anything major to comment on.

PR #1848, opened 2022-10-18
Fix Float libraries w.r.t. toWord in newer Agda versions

dolio:
Using a recent development Agda, I got errors in the `Float` modules (which are dependencies of e.g. `Reflection`) due to `toWord` now returning a `Maybe`. This fixes up the issues.

I imagine this breaks the same libraries relative to released versions of Agda, so it should only be merged when that is acceptable.


gallais:
I've you're using the dev version of Agda then you should be using the `experimental` branch of the stdlib.

PR #1850, opened 2022-10-20
[resolves #1333] writer transformer

### LINKED ISSUE #1333: Add `WriterT` and `ExceptT` monad transformers
As the title says, these monad transformers are widely used in Haskell so it would be nice to have them available in Agda as well. They should also get corresponding `instance` declarations (see #559 and #1332).

cmcmA20:
I'm aware of `Data.Sum.Effectful.*` but it didn't help me much, I have no idea how to properly organize all the stuff about ExceptT reusing those modules.

I'm also asking for tips about improving the current code. First of all I couldn't come up with a really indexed construction for ExceptT because short-circuiting computations mess with the indices. Exceptions can be modeled with `Effect.Monad.Continuation` but is it a good definition for stdlib? Also there's no standard definition of categories in this library so indexed writer as in Bob Atkey's paper is out of reach.

### DIFF
--- a/src/Effect/Monad/Reader/Transformer.agda
+++ b/src/Effect/Monad/Reader/Transformer.agda
@@ -9,6 +9,7 @@
 
 module Effect.Monad.Reader.Transformer where
 
+open import Algebra using (RawMonoid)
 open import Effect.Choice
 open import Effect.Empty
 open import Effect.Functor
@@ -115,7 +116,20 @@ liftReaderT MRead = record
   ; local = λ f mx → mkReaderT (λ r₂ → local f (runReaderT mx r₂))
   } where open RawMonadReader MRead
 
-open import Data.Product using (_,_)
+open import Data.Product using (_×_; _,_)
+open import Effect.Monad.Writer.Transformer.Base
+
+liftWriterT : (MR : RawMonoid r g) →
+              RawFunctor M →
+              RawMonadReader R M →
+              RawMonadReader R (WriterT MR M)
+liftWriterT MR M MRead = record
+  { reader = λ k → mkWriterT λ w → ((w ,_) <$> reader k)
+  ; local = λ f mx → mkWriterT λ w → (local f (runWriterT mx w))
+  } where open RawMonadReader MRead
+          open RawFunctor M
+          open RawMonoid MR
+
 open import Effect.Monad.State.Transformer.Base
 
 liftStateT : RawFunctor M →
--- a/src/Effect/Monad/Reader/Transformer.agda
+++ b/src/Effect/Monad/Reader/Transformer.agda
@@ -128,7 +128,6 @@ liftWriterT MR M MRead = record
   ; local = λ f mx → mkWriterT λ w → (local f (runWriterT mx w))
   } where open RawMonadReader MRead
           open RawFunctor M
-          open RawMonoid MR
 
 open import Effect.Monad.State.Transformer.Base
 
--- a/src/Effect/Monad/State/Transformer.agda
+++ b/src/Effect/Monad/State/Transformer.agda
@@ -10,6 +10,7 @@ open import Level using (Level; suc; _⊔_)
 
 module Effect.Monad.State.Transformer where
 
+open import Algebra using (RawMonoid)
 open import Data.Product using (_×_; _,_; map₂; proj₁; proj₂)
 open import Data.Unit.Polymorphic.Base
 open import Effect.Choice
@@ -136,3 +137,16 @@ liftReaderT Mon = record
   { gets   = λ f → mkReaderT (const (gets f))
   ; modify = λ f → mkReaderT (const (modify f))
   } where open RawMonadState Mon
+
+open import Effect.Monad.Writer.Transformer.Base
+
+liftWriterT : (MS : RawMonoid s f) →
+              RawFunctor M →
+              RawMonadState S M →
+              RawMonadState S (WriterT MS M)
+liftWriterT MS M Mon = record
+  { gets   = λ f → mkWriterT λ w → (gets ((w ,_) ∘′ f))
+  ; modify = λ f → mkWriterT λ w → (const (w , tt) <$> modify f)
+  } where open RawMonadState Mon
+          open RawFunctor M
+          open RawMonoid MS
--- a/src/Effect/Monad/State/Transformer.agda
+++ b/src/Effect/Monad/State/Transformer.agda
@@ -149,4 +149,3 @@ liftWriterT MS M Mon = record
   ; modify = λ f → mkWriterT λ w → (const (w , tt) <$> modify f)
   } where open RawMonadState Mon
           open RawFunctor M
-          open RawMonoid MS
--- a/src/Effect/Monad/Writer.agda
+++ b/src/Effect/Monad/Writer.agda
@@ -1,0 +1,62 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The writer monad
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Effect.Monad.Writer where
+
+open import Algebra using (RawMonoid)
+open import Data.Product using (_×_)
+open import Effect.Applicative
+open import Effect.Choice
+open import Effect.Empty
+open import Effect.Functor
+open import Effect.Monad
+open import Effect.Monad.Identity as Id using (Identity; runIdentity)
+open import Function.Base using (_∘′_)
+open import Level using (Level)
+
+import Effect.Monad.Writer.Transformer as Trans
+
+private
+  variable
+    w ℓ : Level
+    A : Set w
+    𝕎 : RawMonoid w ℓ
+
+------------------------------------------------------------------------
+-- Re-export the monad writer operations
+
+open Trans public
+  using (RawMonadWriter)
+
+------------------------------------------------------------------------
+-- Writer monad
+
+Writer : (𝕎 : RawMonoid w ℓ) (A : Set w) → Set w
+Writer 𝕎 = Trans.WriterT 𝕎 Identity
+
+functor : RawFunctor (Writer 𝕎)
+functor = Trans.functor Id.functor
+
+module _ {𝕎 : RawMonoid w ℓ} where
+
+  open RawMonoid 𝕎 renaming (Carrier to W)
+
+  runWriter : Writer 𝕎 A → W × A
+  runWriter ma = runIdentity (Trans.runWriterT ma ε)
+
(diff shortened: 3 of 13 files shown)


gallais:
Note that we now have: http://agda.github.io/agda-stdlib/Effect.Monad.Error.Transformer.html

cmcmA20:
Very good! I think your changes subsume my `ExceptT` version, so I'm withdrawing it. Take a look at the updated Writer though.

src/Effect/Monad/Reader/Transformer.agda
```diff
@@ -115,7 +116,22 @@ liftReaderT MRead = record
   ; local = λ f mx → mkReaderT (λ r₂ → local f (runReaderT mx r₂))
   } where open RawMonadReader MRead
 
-open import Data.Product using (_,_)
+open import Data.Product using (_×_; _,_)
+open import Effect.Monad.Writer.Transformer.Base
+
+module _ {MR' : RawMonoid r g} where
```
gallais, line 122: I don't think this should be an implicit argument as there is no chance it will be reconstructed by unification.

src/Effect/Monad/State/Transformer.agda
```diff
@@ -136,3 +137,18 @@ liftReaderT Mon = record
   { gets   = λ f → mkReaderT (const (gets f))
   ; modify = λ f → mkReaderT (const (modify f))
   } where open RawMonadState Mon
+
+open import Effect.Monad.Writer.Transformer.Base
+
+module _ {MS' : RawMonoid s f} where
```
gallais, line 143: Same remark here

src/Effect/Monad/Writer.agda
```diff
@@ -0,0 +1,64 @@
...
+runWriter : Writer W A → W × A
+runWriter = runIdentity ∘′ Trans.runWriterT
+
+------------------------------------------------------------------------
+-- Structure
+
+functor : RawFunctor (Writer W)
+functor = Trans.functor Id.functor
+
+module _ {MW : RawMonoid w w} where
```
gallais, line 50: And here

src/Effect/Monad/Writer/Instances.agda
```diff
@@ -0,0 +1,26 @@
...
+  writerTChoice = λ {s} {S} {f} {M} {{ch}} → choice {s} {S} {f} {M} ch
+  writerTApplicativeZero = λ {s} {S} {f} {M} {{mo}} {{mon}} → applicativeZero {s} {S} {f} {M} {mo} mon
+  writerTAlternative = λ {s} {S} {f} {M} {{mo}} {{mpl}} → alternative {s} {S} {f} {M} {mo} mpl
+  writerTMonad = λ {s} {S} {f} {M} {{mo}} {{mon}} → monad {s} {S} {f} {M} {mo} mon
+  writerTMonadZero = λ {s} {S} {f} {M} {{mo}} {{mz}} → monadZero {s} {S} {f} {M} {mo} mz
+  writerTMonadPlus = λ {s} {S} {f} {M} {{mo}} {{mpl}} → monadPlus {s} {S} {f} {M} {mo} mpl
+  writerTMonadT = λ {s} {S} {f} {M} {{mo}} {{mon}} → monadT {s} {S} {f} {M} {mo} mon
+  writerTMonadWriter = λ {s} {S} {f} {M} {{mon}} → monadWriter {s} {S} {f} {M} mon
+  writerTLiftReaderT = λ {s} {S₁} {S₂} {f} {M} {{fun}} {{mw}} → liftReaderT {s} {S₁} {S₂} {f} {M} fun mw
+  writerTLiftStateT = λ {s} {S₁} {S₂} {f} {M} {{fun}} {{mw}} → liftStateT {s} {S₁} {S₂} {f} {M} fun mw
```
gallais, line 26: Do we need a lift for `ErrorT` too?

src/Effect/Monad/Writer/Transformer/Base.agda
```diff
@@ -0,0 +1,44 @@
...
+
+------------------------------------------------------------------------
+-- Writer monad transformer
+
+record WriterT
+       (M : Set w → Set g)
+       (A : Set w)
+       : Set (w ⊔ g) where
+  constructor mkWriterT
+  field runWriterT : M (W × A)
```
gallais, line 43: Can we please also get
```
evalWriterT : RawFunctor M → WriterT W M A → M A
execWriterT : RawFunctor M → WriterT W M A → M W
```
(and the equivalent variant without `T` in `Effect.Monad.Writer`?

gallais (commented):
Looks good!

My main concern is that this gives a blessed implementation
```agda
record WriterT
       (M : Set w → Set g)
       (A : Set w)
       : Set (w ⊔ g) where
  constructor mkWriterT
  field runWriterT : M (W × A)
 ```
 but there are other valid ones, including the CPS-based (cf. [this thread](https://mail.haskell.org/pipermail/libraries/2013-March/019528.html) for instance)
 ```agda
record WriterT
       (M : Set w → Set g)
       (A : Set w)
       : Set (w ⊔ g) where
  constructor mkWriterT
  field runWriterT : W → M (W × A)
 ```
 
 Should we have two modules, one with each implementation?
 Maybe in `Effect.Monad.Writer.Transformer.FirstOrder` vs. `Effect.Monad.Writer.Transformer.CPS`?

**Comment by gallais** [comment]
Should `WriterT` be the CPS-based one? Because the other one is the left-product transformer
(just like we have it [for `Sum`](http://agda.github.io/agda-stdlib/Data.Sum.Effectful.Left.Transformer.html)).

**Comment by cmcmA20** [comment]
Thanks for the link! Looks like the naive version is not that useful from the performance perspective.

I agree with making `WriterT` default implementation CPS-based, let the other one be in `Data.Product.Effectful`. Maybe you should also add a comment about it being direct but impractical for programming?

File: src/Effect/Monad/Writer/Instances.agda (hunk 4 above)
Comment (cmcmA20) [reply src/Effect/Monad/Writer/Instances.agda:26]: It would be nice to have all lifts for all transformers but writing them by hand is a bitter experience.

What are the prospects of adding an `Effect.Algebraic` module or something? Seems like Agda's dependent types are much more accommodating to effect systems than Haskell which has a zoo of incompatible and inefficient ones.

Comment (gallais) [reply src/Effect/Monad/Writer/Instances.agda:26]: I think it's still an open research question. Edwin Brady had a paper a while ago
about doing something like that in Idris1 but he's come back from it for Idris2 and
I'm not sure we have a good story yet.

**Comment by cmcmA20** [comment]
https://github.com/agda/agda-stdlib/blob/7b7230752e86ba4e4b311fb812ede7cc5818a5e8/src/Effect/Monad/Error/Transformer.agda#L29-L31
There was no chance of using it with `RawMonadIO` which bumps the level.

**Comment by cmcmA20** [comment]
This error shows up in the monad/fibonacci test:
```
———— Errors ————————————————————————————————————————————————
  Resolve instance argument
    _r_79 : RawMonadState Fib (λ v → _M_161 v)
  Candidates
    stateTMonadState :
      {s S : Agda.Primitive.Level} {f : Set s → Set S} {M : Set s}
      ⦃ mon : RawMonad f ⦄ →
      RawMonadState M (Effect.Monad.State.Transformer.Base.StateT M f)
    stateTLiftWriterT :
      {R s S : Agda.Primitive.Level} {f : Set R → Set S} {M : Set R}
      ⦃ fun : Effect.Functor.RawFunctor f ⦄
      ⦃ mo : Algebra.Bundles.Raw.RawMonoid R s ⦄
      ⦃ ms : RawMonadState M f ⦄ →
      RawMonadState M
      (Effect.Monad.Writer.Transformer.Base.WriterT
       (Algebra.Bundles.Raw.RawMonoid.Carrier mo) f)
    (stuck)
```
Bad naming of level variables aside, the error is not illuminating at all. I think it's the same problem as with `stateTLiftStateT` instance. What's your plan on this?

I doubt Writer is useful in this form without the corresponding lifts. Should I abandon this PR or convert it to draft?

**Comment by MatthewDaggitt** [comment]
@gallais do you have any more comments?

File: src/Effect/Monad/Reader/Transformer.agda [hunk 6]
```diff
@@ -115,7 +116,20 @@ liftReaderT MRead = record
...
   } where open RawMonadReader MRead
 
-open import Data.Product using (_,_)
+open import Data.Product using (_×_; _,_)
+open import Effect.Monad.Writer.Transformer.Base
+
+liftWriterT : (MR : RawMonoid r g) →
+              RawFunctor M →
+              RawMonadReader R M →
+              RawMonadReader R (WriterT (RawMonoid.Carrier MR) M)
```
gallais, line 125: This seems really dodgy. There is no chance in hell that Agda will be able to reconstruct
`MR` simply from a `RawMonoid.Carrier MR ~ some_type` constraint e.g. `(0, max)`, `(0, +)`
and `(1, *)` are all valid monoids for the natural numbers.

Can we get rid of the `RawMonoid r g` constraint?
Otherwise we may want to parameterise `WriterT` not by a `Set` but rather a `RawMonoid`?

gallais (commented):
I think I may have found the source of the error.

src/Effect/Monad/Reader/Transformer.agda
cmcmA20: Fair point, I did the latter.

MatthewDaggitt:
@gallais are you happy for this to be merged?

gallais (approved):

MatthewDaggitt:
Ah sorry @cmcmA20, one last thing! Please could add an entry in the CHANGELOG about the new modules you've added?

Then I promise this is merging in :+1:

cmcmA20:
Sorry I forgot about the changelog, fixed now.

MatthewDaggitt:
Great. Merging now. Thanks again for the great first PR!

jamesmckinna:
I think I have to apologise profusely to all involved with this PR. I think my reverted commits in sorting out the parity-bis (now deleted!) branch/PR [#1894](https://github.com/agda/agda/issues/1894) seem to have undone the effect of PR [#1850](https://github.com/agda/agda/issues/1850)... But I honestly don't understand how this has happened, nor how (best) to fix it.

MatthewDaggitt:
No worries, it's my fault for not checking the PR properly before hitting merge. I'll fix it with a `revert`

MatthewDaggitt:
Done. @jamesmckinna could you reopen a clean parity semiring PR please?

jamesmckinna:
OK, checked out a clean branch `parity-clean`(but maybe from the wrong base?), but the associated PR doesn't look right in terms of the diffs... please advise...

PR #1854, opened 2022-10-23
Replaced --without-K with --cubical-compatible

nad:
The standard library should switch from `--without-K` to `--cubical-compatible`, otherwise the code cannot be imported from modules that use `--cubical` or `--erased-cubical`.

Perhaps you would like to add something like the following text somewhere in the changelog:
```
* Modules that used `--without-K` now use `--cubical-compatible`.
```
However, I don't know if you classify this change as being backwards-compatible or not.

I replaced uses of `--without-K` in `OPTIONS` pragmas, in `HACKING.md`, and in `GenerateEverything.hs`. (I did not modify old changelogs.) The flag `--without-K` is mentioned in some comments in Agda files. I looked into changing those, but it seems as if the comments were out of date already before this change, so I leave such changes to you.


MatthewDaggitt:
Thanks!

Issue #1229, opened 2020-06-06
Make some `Dec` operations more easily accessible?


masaeedu:
Hello there. I was doing some stuff with decidable equality and ended up needed the following operations:

```agda
  imap : ∀ {ℓ} {a b : Set ℓ} → (a → b) → (b → a) → Dec a → Dec b
  imap f _ (yes x) = yes (f x)
  imap _ g (no  x) = no $ x ∘ g

  zip : ∀ {ℓ} {a b : Set ℓ} → Dec a → Dec b → Dec (a × b)
  zip (yes x) (yes y) = yes $ (x , y)
  zip (no  x) _       = no $ λ { (x' , _) → x x' }
  zip _       (no  y) = no $ λ { (_ , y') → y y' }

  _∪_ : ∀ {ℓ} {a b : Set ℓ} → Dec a → Dec b → Dec (a ⊎ b)
  _∪_ (yes a) _ = yes (inj₁ a)
  _∪_ (no  a) b = imap inj₂ ([ ⊥-elim ∘ a , id ]′) b
```

Does it make any sense to add these to the `Relation.Nullary` module that `Dec` is exported from?

masaeedu:
`zip` and `_<|>_` respectively have the following units, but doesn't seem to be much of a use for exporting these given that they're very thin aliases for stuff that's already exported under other names:

```agda
  pure : ∀ {ℓ} {a : Set ℓ} → a → Dec a
  pure = yes

  nil : ∀ {ℓ} {a : Set ℓ} → Dec ⊥
  nil = no ⊥-elim
```

gallais:
This is respectively:

* [map′](http://agda.github.io/agda-stdlib/Relation.Nullary.Decidable.Core.html#3636)
* [_×-dec_](http://agda.github.io/agda-stdlib/Relation.Nullary.Product.html)
* [_⊎-dec_](http://agda.github.io/agda-stdlib/Relation.Nullary.Sum.html)

masaeedu:
Doii, should have looked around more. Thanks

gallais:
No bother. I myself find it annoying that you have to import 3 or 4 modules to
start working with `Dec`. I wonder whether we could reorganise this part of the
hierarchy in a backwards-compatible manner that would allow you to import, say
`Relation.Decidability` and have immediately access to a whole host of functions.

JacquesCarette:
[...]

Issue #1635, opened 2021-12-01
Precedence of negation and sigma/exists syntax


MatthewDaggitt:
As pointed out by @jamesmckinna when trying to update PFLA to use v1.7 of the library (analysis below is his), currently you can't use `¬_` from `Relation.Nullary` and `∃[ A ] B` syntax (i.e. `∃-syntax`) from `Data.Product` without bracketing because the precedences are the wrong way around:

https://github.com/agda/agda-stdlib/blob/ed8efad584ccf0e96503371571bc0ed24ded62ee/src/Relation/Nullary.agda#L23

https://github.com/agda/agda-stdlib/blob/ed8efad584ccf0e96503371571bc0ed24ded62ee/src/Data/Product.agda#L78

So you can't write `¬ ∃[ A ] B`. The reason that it broke from v1.3 to v1.7 is that the fixity of the `∃-syntax` was first added by @Taneb in a0b06ba

So it looks like we need to switch the precedences somehow (`Σ-syntax` included). Anyone have any thoughts? We should probably have a look at what else has precedences in at the 2/3/4 level...

jamesmckinna:
@wenkokke You remarked on Monday that PLFA was now up-to-date with `std-lib`. How did you resolve/deal with this particular issue?

wenkokke:
We added hella parentheses.

jamesmckinna:
Just did a quick Friday-afternoon experiment: changing the precedence of negation to
```
infix 2 ¬_
```
incurs only *one* knock-on change, namely in `Data.Fin.Properties.¬∀⟶∃¬-smallest`, with type
```
∀ n {p} (P : Pred (Fin n) p) → Decidable P →
                 ¬ (∀ i → P i) → ∃ λ i → (¬ P i) × ((j : Fin′ i) → P (inject j))
```
(we now need brackets around `(¬ P i)` in addition to the `((j : Fin′ i) → P (inject j))` of the original)
while the example of 
```
example : (A : Set a) → (A → Set b) → Set (a ⊔ b)
example A B = ¬ Σ[ x ∈ A ] B x
```
shows that we have 'what we want' (or: what I/we *think* we might want). 

But it (perhaps) feels a fragile move to have `¬`, `Σ` (and `_×_`) all at the same precedence level. Thoughts?

JacquesCarette:
[...]

PR #1855, opened 2022-10-24
Stage 1 of `Relation.Nullary` refactor

MatthewDaggitt:
First steps in #1229 to getting `Relation.Nullary` to re-export more stuff. 

Moved basic definitions (`¬_`/`Dec` etc.) from `Relation.Nullary` into the appropriate submodules (e.g. `Relation.Nullary.Negation.Core`), and then sort out resulting dependency problems.

Had to create the long overdue `Data.Product.Base` in order to allow the existential view of products to live in `Data.Product`.


jamesmckinna:
Wow! I notice that you have also seized the nettle and given `Σ-syntax` etc. precedence 2, not 3. Does this mean that you will also manage to knock off issue #1635 at the same time?

jamesmckinna:
(deleted comment based on faulty memory of various discussions)

PR #1861, opened 2022-10-24
[fix] added raw bundles to `Data.Nat.Base`

jamesmckinna:
EDITED: Removed duplicate bundles, so should be fixed now. 
DONE: update CHANGELOG to record new bundles added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1739,6 +1739,13 @@ Other minor changes
   _⊓′_ : ℕ → ℕ → ℕ
   ∣_-_∣′ : ℕ → ℕ → ℕ
   _! : ℕ → ℕ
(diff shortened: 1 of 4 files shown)


jamesmckinna:
I had thought I'd tidy up after myself with addressing issue #1865 , but perhaps that can now wait for a subsequent PR? 
What's the preferred workflow?

JacquesCarette:
Oh no, I blinked and missed this... I really don't think it's a good idea to have large `.Base`. `.Base` should just define the operations and nothing else. I hate having `.Base` depend on anything in `Algebra`.  I really think all these bundles should be in `Data.Nat.Bundles` (or `Data.Nat.RawBundles`).

Issue #1844, opened 2022-10-06
`Data.Integer.Properties.+-isAbelianGroup` should be called `+-0-isAbelianGroup`


Taneb:
This is consistent with the other names in the module as well as `Data.Rational.Properties`

jamesmckinna:
Additionally, the raw bundles in `Data.Integer.Properties` should move to `Data.Integer.Base` as per issue #1755 PR #1810, etc.

jamesmckinna:
Oh, and the *Monoids* in ``Rational.*` should have the names of their units added to the bundle names... sigh.

PR #1862, opened 2022-10-24
[fix] issue #1844

jamesmckinna:
Suggest also fixing #1755 for `Data.Integer` as part of this PR.

### DIFF
--- a/src/Data/Integer/Properties.agda
+++ b/src/Data/Integer/Properties.agda
@@ -2387,5 +2387,5 @@ Please use pos-* instead."
 Please use +-0-isAbelianGroup instead."
 #-}
 {- issue1844/issue1755: raw bundles have moved to `Data.X.Base` -}
-open Data.Integer.Base public 
+open Data.Integer.Base public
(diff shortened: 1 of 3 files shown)


src/Data/Integer/Properties.agda
```diff
@@ -944,8 +944,8 @@ distribʳ-⊖-+-neg m n o = begin
   ; ⁻¹-cong  = cong (-_)
   }
 
-+-isAbelianGroup : IsAbelianGroup _+_ +0 (-_)
-+-isAbelianGroup = record
++-0-isAbelianGroup : IsAbelianGroup _+_ +0 (-_)
++-0-isAbelianGroup = record
```
MatthewDaggitt, line 948: We should add a deprecation warning for the old name at the bottom of the file.

jamesmckinna: Aaargh... yes. 🤦

jamesmckinna:
Now fixed issue #1865  for Data.Integer so hopefully that works and this PR is now ready.

PR #1867, opened 2022-10-25
[fix] issue 1865 for all the raw bundles added in #1755 [coda to Issue1858/ PR 1861]

jamesmckinna:


MatthewDaggitt:
While you're here any chance you could do the same thing for the bundles moved in https://github.com/agda/agda-stdlib/commit/578e7a60511c9fcf1f6931e13da957e9e0dbb6ac as well? Then we could close the issue :smile:

jamesmckinna:
Yes! not tonight I think (SPLS all day, plus I'm driving there and back...) ;-)

jamesmckinna:
cf. discussion in PR #1852: I'll do the re-imports, and then rename/deprecate the names. On second thoughts, I won't do the second, and leave the inconsistency to mop up later. But my obsessive compulsion towards consistency is strong...

jamesmckinna:
How do I achieve strikethrough in the name of the PR? to reflect that it is now more general?

MatthewDaggitt:
I'm afraid something has gone wrong in your merging here, and you've picked up a load of extraneous commits to `Relation.Nullary` that make it very difficult to review. Any chance you could try to clean it up?

jamesmckinna:
Aaargh... apologies. I'll try to do this (more) slowly and carefully. 
I shouldn't have sat down to do it in a burst of wild energy after an already long day including 100-mile round trip drive. 
Let me know how I'm getting on.

jamesmckinna:
@MatthewDaggitt I've gone through each of the 13 commits *on my branch* [issue1858](https://github.com/jamesmckinna/agda-stdlib/tree/issue1858) associated with this PR, and they all seem good, so I've commented to that effect on the individual commits.

That leaves commits [167fc7f](167fc7f3964be58d9aa407110d66e0a4d07fff8a) and [fd37e5d](fd37e5d216ab97b88058304a5631abfdd773b059) and I think I don't understand how they came to be part of this PR. And the PR itself records *16* commits, not *15* as I might have expected (my 13 plus 2 spurious). 

Commit `c2302f4` appears in the PR, but not on my branch, it would seem, so I'm now thoroughly confused. 
Ah, I see it was part of #1844/#1862 which was successfully merged in a while back. Sorry!

Each of the commits on the branch seem OK, and of the others, obviously yours and Nathan's are the rogue elements. 

What I don't understand is how to remove `167fc7f` and `fd37e5d` from this PR (or, alternatively, given that GH says that the branch can successfully be merged, how to shield your reviewing eyes form them... :-(). Advice welcome! Hopefully the above detective work is a start, at least.

MatthewDaggitt:
So the easiest way is to run `git rebase -i HEAD~16` and then remove the lines containing the commits that you _don't_ want. Then run `git push -f`.

In general this is not encouraged as it changes the history, so if other people were working on this branch then you would royally screw things up for them. But in this case I think its fine!

jamesmckinna:
Hmmm. I tried this, and expected to see 16 commits, of which 3 could be removed. But in fact the 16 I saw went way further back in the history than I might have expected, and the merge commits which caused all the trouble seem not to appear in the list for potential rebasing... so now I have gone somewhat passed my horizon of clear understanding of what's going on.

jamesmckinna:
OK, don't know what wet wrong this time around: I tried `cherry-pick`ing the commits from the backup branch (but there seemed to be only 12 of them this time, not 13), but now have two merge conflicts arising from (what?)

jamesmckinna:
OK, perhaps that has now restored the missing commits in `Data.Integer.Base` and resolved the conflicts in `CHANGELOG`.

jamesmckinna:
`(n+1)`th time is the charm? Sigh.

CHANGELOG.md
```diff
@@ -1815,8 +1829,6 @@ Other minor changes
   ∣_-_∣′ : ℕ → ℕ → ℕ
   _! : ℕ → ℕ
 
-  parity : ℕ → Parity
```
MatthewDaggitt, line 1818: I think removing this entry is a typo?

MatthewDaggitt (commented):
Otherwise looks great now! Thanks for fixing the commits. Hope it wasn't too difficult in the end!

CHANGELOG.md
jamesmckinna: Ach. You're right. I'll reinstate that

jamesmckinna: Did this last commit fix the problem?

MatthewDaggitt: Yup!

MatthewDaggitt:
Thanks for sticking with this one!

Issue #1726, opened 2022-02-26
Deprecate `_≺_` in `Data.Fin.Base`


MatthewDaggitt:
I think we already have more than enough definitions of this relation (I think this is the 5th?), and I'd prefer to have less. This is a moderately odd definition with virtually no properties about it. 

Anybody object to the deprecating it in `v2.0`? I'll ask on Zulip as well.

jamesmckinna:
Did you ask on Zulip? I was beginning to wonder if the `#stdlib` channel/stream should have a specific `> deprecation` or `< RFC` (or something of that ilk...) channel/stream for discussion of proposed changes, esp. wrt deprecation and/or labelled as `breaking` (cf. issue `#1636`)

MatthewDaggitt:
Nope, not yet. Will do so now.

jamesmckinna:
It looks as though, from the deafening silence here, that this deprecation, at least, will (and should!) go through...

JacquesCarette:
I've never used it, so I won't miss it! 😁

jamesmckinna:
Whatever happened to this? Removing it eventually will yield some more dependency simplification arising from PR#1837...

JacquesCarette:
Looks caught in the lack of available cycles from the maintenance team. I'll assign myself, in a great fit of hope.

jamesmckinna:
It looks as if, perhaps, this was a relic from early/earlier experiments with Data.Fin.Induction... (?), and recent commits to that make no use of this relation.

jamesmckinna:
So I now have [a branch](https://github.com/jamesmckinna/agda-stdlib/tree/issue1726) with this *almost* working: but the main thing in `Data.Fin.Properties` is to show this relation equivalent to `_<_`. Now, if such a relation (and its constructor, coding for a cute inversion principle akin to `Data.Fin.Properties.toℕ<` ) is to be deprecated, how do I go about deprecating *properties* of it? I can't seem to avoid getting the deprecation warnings arising from even mentioning the relation symbol.

jamesmckinna:
I suppose one solution might be to only deprecate the constructor in `Data.Fin.Base`, and then deprecate the relation+properties in  `Data.Fin.Properties`?

jamesmckinna:
@JacquesCarette how have you got on progressing a solution? Discuss f2f on Wednesday?

MatthewDaggitt:
[...]

PR #1868, opened 2022-10-25
[fix] Issue #1726: deprecates `_≺_` in `Data.Fin.Base` plus its properties

jamesmckinna:


src/Data/Fin/Base.agda
```diff
@@ -335,3 +332,18 @@ NB argument order has been flipped:
 the left-hand argument is the Fin m
 the right-hand is the Nat index increment."
 #-}
+------------------------------------------------------------------------
+-- _≺_
```
MatthewDaggitt, line 336: We tend to only use section headings in the deprecation sections to denote versions. Could you remove this one?

src/Data/Fin/Induction.agda
```diff
@@ -5,6 +5,7 @@
 ------------------------------------------------------------------------
 
 {-# OPTIONS --without-K --safe #-}
+{-# OPTIONS --warn=noUserWarning #-} -- for issue #1726
```
MatthewDaggitt, line 8: It would be more useful to say  `for _≺_` because then we know when we can remove it.

src/Data/Fin/Induction.agda
```diff
@@ -159,3 +142,46 @@ module _ {_≈_ : Rel (Fin n) ℓ} where
...
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.0
+
+------------------------------------------------------------------------
+-- Induction over _≺_
```
MatthewDaggitt, line 155: Again can remove the heading.

src/Data/Fin/Properties.agda
```diff
@@ -1133,3 +1122,37 @@ eq? = inj⇒≟
 "Warning: eq? was deprecated in v2.0.
 Please use inj⇒≟ instead."
 #-}
+------------------------------------------------------------------------
+-- deprecated properties of deprecated _≺_
+-- introduce new proofs, actually simplified compared to original
+-- so that Data.Fin.Properties doesn't mention the constructor
```
MatthewDaggitt, line 1128: And here.

MatthewDaggitt (commented):
Otherwise looks great!

src/Data/Fin/Induction.agda
```diff
@@ -94,24 +96,6 @@ private
   ... | no  n≢i = subst P (inject₁-lower₁ i n≢i) (Pᵢ₊₁⇒Pᵢ _ Pᵢ₊₁)
     where Pᵢ₊₁ = induct (rec _ (ℕ.≤-reflexive (cong suc (sym (toℕ-lower₁ i n≢i)))))
 
-------------------------------------------------------------------------
--- Induction over _≺_
-
-≺-Rec : RecStruct ℕ ℓ ℓ
```
MatthewDaggitt, line 100: Deprecation missing from CHANGELOG?

jamesmckinna: I've pushed a `CHANGELOG` entry, but it is not an orthodox '`x` deprecated in favour of `y` message'. Please advise.
Also, starting to panic lest the merge problems on other branches might have polluted this one, too. Ditto re advice!

MatthewDaggitt: Nope, all looks good to me :+1:

Issue #1851, opened 2022-10-20
`Data.Parity` and the `2+_` pattern


jamesmckinna:
@Taneb 's recent work on permutations-as-transposition-lists has drawn attention to the (possible) need for a Parity type (rather than proxying via `Data.Sign`), together with various properties: so far I have a [branch](https://github.com/jamesmckinna/agda-stdlib/tree/parity) with 
UPDATED:
- [x] `Parity` admits a commutative semiring structure (with `0` = `0ℙ`; `1` = `1ℙ` etc.)
- [x] a function `parity` from `Nat` to `Parity` in `Data.Nat.Base`
- [x] a function `toSign` from `Parity` to `Sign`
- [x] a function `fromSign` from `Sign` to `Parity`
- [x] proofs that `(toSign, fromSign)` form an abelian group iso from `(Parity, 0, _+_)` to `(Sign, +, _*_)` 
- [x] miscellaneous tidyings up of `Data.Sign` etc. to reflect new names for algebraic structures/bundles, cf. issue #1844 and in line with review comments on current PR
- [x] `CHANGELOG`

TODO in a subsequent PR:
- [x] proof that `parity` is a commutative semiring homomorphism from `Nat` to `Parity`
- [ ] rationalise that proof; and in particular resolve where it belongs

Additionally, this might have the following as knock-on effects: 
- [x] rationalise the many instances in the library where the `0, 1, 2+ n` pattern analysis of a `Nat` argument takes place (esp. wrt `N{-}ary` operations, as well as the obvious `Binary` and `Nat` examples) see [this branch](https://github.com/jamesmckinna/agda-stdlib/tree/suc-suc) and PR #1870 

EDITED:
~~It raised a question in my mind as well regarding where homomorphism-like properties belong in the library: under the source type, in `*.Properties`, or somewhere else? I currently have `Parity` import `Sign` in order to define the operations, but it strikes me that there's an asymmetry to that decision, which might better be reflected in having subdirectories of the `Algebra.*` hierarchy which record concrete instances of `(Semi)RingHomomorphism`s etc. suggestions welcome!~~ moved discussion to issue #1871

PR #1870, opened 2022-10-26
rationalise use of `suc (suc n)` and `suc n@(suc _)` patterns throughout; remove `2+ _` pattern

jamesmckinna:
This was left as TODO in issue #1851 / PR #1852
NB I have NOT touched `Data.Nat.Logarithm.Core` as being too fiddly to want to mess with. 
UPDATED: less fiddly than I feared!


src/Data/Vec/Recursive/Effectful.agda
```diff
@@ -37,9 +37,9 @@ module _ {f g F} (App : RawApplicative {f} {g} F) where
   open RawApplicative App
 
   sequenceA : ∀ {n A} → F A ^ n → F (A ^ n)
-  sequenceA {0}    _          = pure _
-  sequenceA {1}    fa         = fa
-  sequenceA {2+ n} (fa , fas) = _,_ <$> fa ⊛ sequenceA fas
+  sequenceA {0}    _                   = pure _
+  sequenceA {1}    fa                  = fa
+  sequenceA {suc n@(suc _)} (fa , fas) = _,_ <$> fa ⊛ sequenceA fas
```
MatthewDaggitt, line 42: Don't need to name `n` here?

jamesmckinna:
In fact, it doesn't seem too difficult to 'fix' the patterns in `Data.Nat.Logarithm.Core`, provided that one can accept an additional (definitional!) unfolding of floor and ceiling in the relevant case(s). But this just applies to the first two definition in `Core`, and otherwise everything else works the same (because of definitional equality!)

MatthewDaggitt:
Thanks, looks great!

PR #1872, opened 2022-10-29
Standard algebraic properties of gcd

Lysxia:


src/Data/Integer/GCD.agda
```diff
@@ -44,5 +48,17 @@ gcd[i,j]≡0⇒i≡0 i j eq = ∣i∣≡0⇒i≡0 (ℕ.gcd[m,n]≡0⇒m≡0 (+-i
 gcd[i,j]≡0⇒j≡0 : ∀ {i j} → gcd i j ≡ 0ℤ → j ≡ 0ℤ
 gcd[i,j]≡0⇒j≡0 {i} eq = ∣i∣≡0⇒i≡0 (ℕ.gcd[m,n]≡0⇒n≡0 ∣ i ∣ (+-injective eq))
 
-gcd-comm : ∀ i j → gcd i j ≡ gcd j i
+gcd-comm : Commutative gcd
 gcd-comm i j = cong (+_) (ℕ.gcd-comm ∣ i ∣ ∣ j ∣)
+
```
JacquesCarette, line 53: 1. This is great,
2. Shouldn't all these things be in `Data.Integer.GCD.Properties`?

src/Data/Nat/GCD.agda
```diff
@@ -103,11 +106,41 @@ gcd[m,n]≡0⇒n≡0 : ∀ m {n} → gcd m n ≡ 0 → n ≡ 0
 gcd[m,n]≡0⇒n≡0 m {zero}  eq = P.refl
 gcd[m,n]≡0⇒n≡0 m {suc n} eq = contradiction eq (gcd[m,n]≢0 m (suc n) (inj₂ λ()))
 
-gcd-comm : ∀ m n → gcd m n ≡ gcd n m
+gcd-comm : Commutative gcd
 gcd-comm m n = ∣-antisym
   (gcd-greatest (gcd[m,n]∣n m n) (gcd[m,n]∣m m n))
   (gcd-greatest (gcd[m,n]∣n n m) (gcd[m,n]∣m n m))
 
+gcd-assoc : Associative gcd
```
JacquesCarette, line 114: Same comment (great, but ought to be in `.Properties`.

src/Data/Nat/GCD.agda
```diff
@@ -103,11 +106,41 @@ gcd[m,n]≡0⇒n≡0 : ∀ m {n} → gcd m n ≡ 0 → n ≡ 0
...
+gcd-assoc m n p = ∣-antisym
+  (gcd-greatest {m} {gcd n p} (∣-trans (gcd[m,n]∣m (gcd m n) p) (gcd[m,n]∣m m n))
```
JacquesCarette, line 116: Wouldn't this proof be much more readable if written equationally?

JacquesCarette (changes requested):

src/Data/Integer/GCD.agda
MatthewDaggitt: There is no such `Properties` file. The dependencies for the more "complex" operations are still snarled up. I've been trying to gradually migrate all operations into the `Base` files, leaving e.g. `Data.Integer.GCD` as the de-facto properties file, but I've got stuck on the dependency on the proof that `_<_` is well-founded.

Therefore I don't think @Lysxia should responsible for changing this. I think its fine here at the moment.

src/Data/Integer/GCD.agda
```diff
@@ -44,5 +48,17 @@ gcd[i,j]≡0⇒i≡0 i j eq = ∣i∣≡0⇒i≡0 (ℕ.gcd[m,n]≡0⇒m≡0 (+-i
...
+gcd-assoc : Associative gcd
+gcd-assoc i j k = cong (+_) (ℕ.gcd-assoc ∣ i ∣ ∣ j ∣ (∣ k ∣))
+
+gcd-1ˡ : LeftZero 1ℤ gcd
```
MatthewDaggitt, line 57: These should be called `gcd-zeroˡ` etc.

src/Data/Nat/GCD.agda
```diff
@@ -103,11 +106,41 @@ gcd[m,n]≡0⇒n≡0 : ∀ m {n} → gcd m n ≡ 0 → n ≡ 0
...
+    (gcd-greatest {n} (∣-trans (gcd[m,n]∣m (gcd m n) p) (gcd[m,n]∣n m n))
+                      (gcd[m,n]∣n (gcd m n) p)))
+  (gcd-greatest {gcd m n}
+    (gcd-greatest {m} (gcd[m,n]∣m m (gcd n p))
+                      (∣-trans (gcd[m,n]∣n m (gcd n p)) (gcd[m,n]∣m n p)))
+    (∣-trans (gcd[m,n]∣n m (gcd n p)) (gcd[m,n]∣n n p)))
+
+gcd-0ˡ : LeftIdentity 0 gcd
```
MatthewDaggitt, line 124: Again should be called `gcd-zeroˡ`, ``gcd-identityˡ` etc.

src/Data/Nat/GCD.agda
MatthewDaggitt: Have a look at `∣-Reasoning`

MatthewDaggitt (commented):
Otherwise, as @JacquesCarette says, this looks great! Thank you for the contribution :+1:

src/Data/Nat/GCD.agda
Lysxia: I may be missing something. How would equational reasoning help here? As I understand it, it's another way to write `|-trans`, but it is not being chained here. IMO the "creative" part of the proof is the application of `|-antisym` and `gcd-greatest` at the beginning, so it doesn't seem worth drowning it out by expanding the bits at the leaves of the proof, or is there really a more equational way to phrase this proof?

JacquesCarette: `|-trans` **is** chaining.

JacquesCarette: I thought there were other proofs further down that used `|-trans` for more than 2 steps. Also, I often find that it's worth extracting an inner proof like this into a lemma.

I don't insist that this particular proof use equational reasoning, just that you carefully look at all your proofs and see if some of them could be easier to understand (by future readers!) if they were done equationally.

Lysxia: Good idea. If I put those in lemmas then (I hope) everyone will be happy. (done)

MatthewDaggitt:
Thanks, looks really good @Lysxia. The only thing left to do is to add an entry in `CHANGELOG.md` describing the changes. After you've done that, I'll merge it in  :+1:

MatthewDaggitt:
@JacquesCarette could you accept the changes?

JacquesCarette (approved):

Lysxia:
Should I mention the refactoring with `Commutative`, even though the type remains the same definitionally?

MatthewDaggitt:
> Should I mention the refactoring with Commutative, even though the type remains the same definitionally?

No, that's fine! If its not going to affect the user then we're good to leave it out.

MatthewDaggitt:
Thanks for the great PR :+1:

Issue #1864, opened 2022-10-25
Dependencies of `Reflection.AST.Argument.Information`


MatthewDaggitt:
Type-checking the above module seems to drag in a huge amount of unnecessary stuff. It must be possible to simplify the dependencies. Also check other `Reflection.AST` modules!

jamesmckinna:
The above PR is a fix to this specific issue regarding `import`s. A more serious effort would refactor much of this code, consisting as it does in each module of how to write a decidable equality on a non-dependent product, given decidable equality on the underlying types, and then prove injectivity of such equality, given injectivity on the underlying types etc. There might then be knock-on consequences regarding what should then be exported?

jamesmckinna:
The relevant proofs of `-injective` involved are all of the form `λ where refl → refl , refl` FWIW, but not inlined as such.

Saransh-cpp:
I can take this up! A lot of the `Data.X` and `Relation.Binary.X` imports should ideally go away in the ongoing PRs.

jamesmckinna:
Is this still open, or should it be closed now?

Saransh-cpp:
Hi! All the imports in `Reflection.AST` were narrowed down in #2072.

I am not sure if your comment above:

> A more serious effort would refactor much of this code, consisting as it does in each module of how to write a decidable equality on a non-dependent product, ...

was resolved, but can always open a new issue/discussion for that :)

jamesmckinna:
Thanks @Saransh-cpp I agree that your earlier PR should (probbaly) have closed this one, but for some reason did not.

PR #1876, opened 2022-10-31
[fix] issue #1864

jamesmckinna:


MatthewDaggitt:
So what was the actual problem here?

jamesmckinna:
See the issue #1864. The 'problem' you had identified was the `import`s being too heavy, so i thinned them out. But while I was doing that, I also saw that the structure of what is actually there is a lot of highly repetitive code for conjuring a decidable relation on a product out of decidable relations on its components. And that seems worth thinking about trying to refactor. So the PR as it stands/stood closed the original issue (as I understood it), while leaving the issue open with more work potentially to do... does that make sense?

MatthewDaggitt:
Yes, sorry, that was too terse of me! I meant what was actually the concrete `import` that caused the dependency cycles! But if you didn't identify it and just removed all that you could, that is also fine!

> So the PR as it stands/stood closed the original issue (as I understood it), while leaving the issue open with more work potentially to do... does that make sense?

:+1:

jamesmckinna:
yes, i didn't do the analysis... except by the tried-and-true-but-brainless method of commenting out imports, and only reinstating locally the actual functions needing to be imported. But quite a lot, it seemed, could simply be deleted...

MatthewDaggitt:
> BTW, I never saw dependency cycles as such, so i wonder if that was a problem at your end?

No, I never saw dependency cycles, apologies if I wrote that. Its just that I was type-checking them and it was ending up type-checking half the standard library in order to do so. Which wasn't exactly great for code needed for writing tactics.

PR #1885, opened 2022-12-12
Add proofs to `Data.List.Properties`

XiaohuWang0921:
Some simple proofs that seem to be missing in the stdlib.


src/Data/List/Properties.agda
```diff
@@ -255,6 +255,27 @@ module _ (A : Set a) where
...
+  cartesianProductWith-zeroʳ (x ∷ xs) = cartesianProductWith-zeroʳ xs
+
+  cartesianProductWith-distribʳ-++ : ∀ xs ys zs → cartesianProductWith f (xs ++ ys) zs ≡ cartesianProductWith f xs zs ++ cartesianProductWith f ys zs
+  cartesianProductWith-distribʳ-++ [] ys zs = refl
+  cartesianProductWith-distribʳ-++ (x ∷ xs) ys zs = begin
+    cartesianProductWith f (x ∷ xs ++ ys) zs ≡⟨⟩
+    map (f x) zs ++ cartesianProductWith f (xs ++ ys) zs ≡⟨ cong (map (f x) zs ++_) (cartesianProductWith-distribʳ-++ xs ys zs) ⟩
+    map (f x) zs ++ cartesianProductWith f xs zs ++ cartesianProductWith f ys zs ≡˘⟨ ++-assoc (map (f x) zs) (cartesianProductWith f xs zs) (cartesianProductWith f ys zs) ⟩
+    (map (f x) zs ++ cartesianProductWith f xs zs) ++ cartesianProductWith f ys zs ≡⟨⟩
+    cartesianProductWith f (x ∷ xs) zs ++ cartesianProductWith f ys zs ∎
```
MatthewDaggitt, line 277: Hmm this proof is very hard to read. The long name is definitely getting in the way. Maybe if we introduced a `private prod = cartesianProductWith` declaration at the top of this block it would improve readability? (while keeping the lemma names the same obviously!)

Also could you try and align the arguments as per the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#function-arguments)?

MatthewDaggitt (commented):
Other than that looks like a great contribution, thanks!

Could you also add a CHANGELOG entry?

src/Data/List/Properties.agda
XiaohuWang0921: Hmm, if only this proof is hard to read wouldn't it be better if we introduce a `let` or `where` binding that is only local to this proof?

Also, wouldn't it be better to introduce the binding as `prod = cartesianProductWith f`?

MatthewDaggitt: > Hmm, if only this proof is hard to read wouldn't it be better if we introduce a let or where binding that is only local to this proof?

The difficulty is that you can't make a `let` or `where` binding for which both the type and the body of the proof is in scope....

> Also, wouldn't it be better to introduce the binding as prod = cartesianProductWith f?

Yup, it would!

CHANGELOG.md
```diff
@@ -2514,6 +2514,17 @@ Other minor changes
   flip′ : (A → B → C) → (B → A → C)
   ```
 
+* Added new proofs to `Data.List.Properties`
+  ```agda
+  cartesianProductWith-zeroˡ       : ∀ (f : A → B → C) ys → cartesianProductWith f [] ys ≡ []
```
MatthewDaggitt, line 2519: Thanks for the entries!

We usually only provide the "interesting" part of the type in the CHANGELOG to avoid overwhelming the reader with too much detail, (e.g. `cartesianProductWith f [] ys ≡ []` and `foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs`). Any chance you could tweak them slightly?

src/Data/List/Properties.agda
gallais: You can use a `private` local definition in an anonymous module to have a declaration
span both the type & body of a proof.

MatthewDaggitt: Yes, that was what I was suggesting.

gallais: My bad !

CHANGELOG.md
XiaohuWang0921: Sorry, my bad!

MatthewDaggitt: No worries, no reason why you should have known :smile:

XiaohuWang0921:
I just realised something: in `Data.Vec.Base` there is `_⊛*_ : Vec (A → B) m → Vec A n → Vec B (m * n)`, which is similar to `cartesianProductWith` but has two advantages: it has a much shorter name, and it is defined in terms of `_>>=_` (or `concatMap`), which means its properties can be proven simply as special cases thereof.

So how about we also add this function to `Data.List.Base` as `_⊛*_ : List (A → B) → List A → List B` and redefine `cartesianProductWith` as `cartesianProductWith f xs ys = map f xs ⊛* ys`?

MatthewDaggitt:
> So how about we also add this function to Data.List.Base as _⊛*_ : List (A → B) → List A → List B and redefine cartesianProductWith as cartesianProductWith f xs ys = map f xs ⊛* ys?

That sounds like  a good idea to me. 

I'm going to merge this in now as this stands alone, but a follow up PR would be welcome!

PR #1887, opened 2022-12-15
Define membership relation for Data.Tree.AVL

Lysxia:
- Prove lemmas about singleton, insert, and lookup

### DIFF
--- a/src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
+++ b/src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
@@ -61,13 +61,13 @@ lookup-bounded {t = node kv lk ku bal} (left p)  =
 lookup-bounded {t = node kv lk ku bal} (right p) =
   Prod.map₁ (trans⁺ _ (ordered lk)) (lookup-bounded p)
 
-rebuild : {t : Tree V l u n} (p : Any P t) → Q (Any.lookup p) → Any Q t
-rebuild (here _)  q = here q
-rebuild (left p)  q = left (rebuild p q)
-rebuild (right p) q = right (rebuild p q)
+lookup-rebuild : {t : Tree V l u n} (p : Any P t) → Q (Any.lookup p) → Any Q t
+lookup-rebuild (here _)  q = here q
+lookup-rebuild (left p)  q = left (lookup-rebuild p q)
+lookup-rebuild (right p) q = right (lookup-rebuild p q)
 
-rebuild-accum : {t : Tree V l u n} (p : Any P t) → Q (Any.lookup p) → Any (Q ∩ P) t
-rebuild-accum p q = rebuild p (q , lookup-result p)
+lookup-rebuild-accum : {t : Tree V l u n} (p : Any P t) → Q (Any.lookup p) → Any (Q ∩ P) t
+lookup-rebuild-accum p q = lookup-rebuild p (q , lookup-result p)
 
 joinˡ⁺-here⁺ : ∀ {l u hˡ hʳ h} →
              (kv : K& V) →
@@ -233,15 +233,17 @@ module _ {V : Value v} where
 
   singleton⁺ : {P : Pred (K& V) p} →
                (k : Key) →
-               (v : Val k) → {l<k<u : l < k < u} →
+               (v : Val k) →
+               (l<k<u : l < k < u) →
                P (k , v) → Any P (singleton k v l<k<u)
-  singleton⁺ k v Pkv = here Pkv
+  singleton⁺ k v l<k<u Pkv = here Pkv
 
   singleton⁻ : {P : Pred (K& V) p} →
                (k : Key) →
-               (v : Val k) → {l<k<u : l < k < u} →
+               (v : Val k) →
+               (l<k<u : l < k < u) →
                Any P (singleton k v l<k<u) → P (k , v)
-  singleton⁻ k v (here Pkv) = Pkv
+  singleton⁻ k v l<k<u (here Pkv) = Pkv
 
   ------------------------------------------------------------------------
   -- insert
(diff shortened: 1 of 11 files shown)


src/Data/Tree/AVL/Indexed/Relation/Unary/Any/Properties.agda
```diff
@@ -148,15 +163,89 @@ joinʳ⁺-right⁺ k₂ t₁ (1# , node⁻ k₆ k₄ t₃ t₅ bal t₇) ∼+  (
...
 
 module _ {V : Value v} where
 
   private
     Val  = Value.family V
     Val≈ = Value.respects V
 
+  singleton⁺ : {P : Pred (K& V) p} →
+               (k : Key) →
+               (v : Val k) → {l<k<u : l < k < u} →
```
MatthewDaggitt, line 236: Should `l<k<u` really be implicit? Can it be reliably reconstructed by Agda from `Any P (singleton k v l<k<u)` in general? Same for the function below.

src/Data/Tree/AVL/Map/Membership.agda
```diff
@@ -0,0 +1,159 @@
...
+    m : Map V
+    k k′ : Key
+    x x′ y y′ : V
+    kx : Key × V
+
+infix 4 _∈_
+
+private
+  Is : Rel (Key × V) _
+  Is = _≈_ ×ᴿ _≡_
```
MatthewDaggitt, line 54: So this is propositional membership right? This module should be called `Tree.AVL.Map.Membership.Propositional` so that we have room for the `Setoid` version later.

src/Data/Tree/AVL/Map/Membership.agda
```diff
@@ -0,0 +1,159 @@
...
+  Is-trans : Transitive (Is {V = V})
+  Is-trans {i = i} {k = k} = ×-transitive Eq.trans ≡-trans {i = i} {k = k}
+
+  Is-sym : Symmetric (Is {V = V})
+  Is-sym {x = x} {y = y} = ×-symmetric sym ≡-sym {x} {y}
+
+------------------------------------------------------------------------
+-- Membership: ∈
+
+_∈_ : Key × V → Map V → Set _
```
MatthewDaggitt, line 65: We will also probably want different types of membership, e.g. key-value, key, value etc. Can we distinguish this particular definition by using a subscript `kv`?

src/Data/Tree/AVL/Map/Membership.agda
```diff
@@ -0,0 +1,159 @@
...
+... | just x = contradiction (∈-lookup⁻ eq) (k∉m x)
+
+∈-lookup-nothing⁻ : lookup m k ≡ nothing → (k , x) ∉ m
+∈-lookup-nothing⁻ eq kx∈m with ≡-trans (≡-sym eq) (∈-lookup⁺ kx∈m)
+... | ()
+
+------------------------------------------------------------------------
+-- ∈?
+
+∈-∈? : (k , x) ∈ m → (k ∈? m) ≡ true
```
MatthewDaggitt, line 142: Okay so here we run into a little bit of a naming problem. The name `_∈?_` is usually reserved for the full proof relevant decision procedure, rather than the boolean procedure which it currently is. Obviously this is a problem with existing code, not your additions. Nonetheless it does affect the naming of your new lemmas!

I'd like to make room for (eventually!) adding the proof relevant decision procedure. I think the best option would be for you to deprecate the name ` _∈?_` in the original file and replace it with the name `member`? Would you be willing to do that?

src/Data/Tree/AVL/Sets/Membership.agda
```diff
@@ -0,0 +1,99 @@
...
+open import Data.Tree.AVL.Sets strictTotalOrder
+import Data.Tree.AVL.Map.Membership strictTotalOrder as Map
+open import Data.Tree.AVL.Map.Relation.Unary.Any strictTotalOrder as Mapₚ
+
+private
+  variable
+    x y : A
+    s : ⟨Set⟩
+
+infix 4 _∈_
```
MatthewDaggitt, line 35: Missing infix declaration for `_∉_`

src/Data/Tree/AVL/Sets/Membership.agda
```diff
@@ -0,0 +1,99 @@
...
+∈toMap : x ∈ s → (x , tt) Map.∈ s
+∈toMap = Mapₚ.map (_, refl)
+
+∈fromMap : (x , tt) Map.∈ s → x ∈ s
+∈fromMap = Mapₚ.map proj₁
+
+------------------------------------------------------------------------
+-- empty
+
+∈-empty : x ∉ empty
```
MatthewDaggitt, line 55: Could the properties of membership live in a separate file please? e.g. `Membership.Properties`

src/Data/Tree/AVL/Map/Membership.agda
```diff
@@ -0,0 +1,159 @@
...
+------------------------------------------------------------------------
+-- Membership: ∈
+
+_∈_ : Key × V → Map V → Set _
+kx ∈ m = Any (Is kx) m
+
+_∉_ : Key × V → Map V → Set _
+kx ∉ m = ¬ kx ∈ m
+
+∈-Respectsˡ : _∈_ {V = V} Respectsˡ Is
```
MatthewDaggitt, line 71: Could these properties live in a separate file please? e.g. `Membership.Propositional.Properties`

MatthewDaggitt (commented):
Thanks, this looks like a fantastic PR :+1: 

I've added a couple of comments below. You'll also need to add a suitable entry in the CHANGELOG :+1:

src/Data/Tree/AVL/Map/Membership.agda
Lysxia: The weird thing is that `Key` is a setoid but `V` uses propositional equality. But if it's indeed your intention for `Setoid`/`Propositional` to refer to the structure of `V`, then sure I can do that.

gallais:
The force-push make it hard to see what has changed and force the review to restart
from the very beginning each time. Please don't do that.

Lysxia:
Sorry about that. I won't do that next time.

MatthewDaggitt:
Looked over it and the changes look good! Thanks again for the great PR :+1:

Issue #1871, opened 2022-10-27
Where should homomorphisms live in the `stdlib` hierarchy?


jamesmckinna:
Knocking off the last outstanding task in Issue #1851 re-raised a question in my mind as well regarding where, in general, proofs homomorphism-like properties belong in the library: under the source type, in `*.Properties`, or somewhere else? 

For example, in PR#1852 I currently have `Parity` import `Sign` in order to define the operations, but it strikes me that there's an asymmetry to that decision, which might better be reflected in having subdirectories of the `Algebra.*` hierarchy which record concrete instances of `(Semi)RingHomomorphisms` etc. Similar considerations apply, *mutatis mutandis*, to @Taneb 's recent #1857 / #1863 : *I* don't regard `List.length` as being a homomorphism from `List _` to `Nat` as a *property* of `List _`, rather it's a (consequence of the) *universal* property of the free monoid construction, of which `List _` is merely a very convenient representative for the *underlying carrier type* of that free algebra. 

Moreover the amount of additional work involved in these proofs, and especially also for the `Semiring` homomorphism `parity` from `Nat` to `Parity`, led me eventually to write this last one in a module on its own; I was at the limits of how to disambiguate all the various equational/algebraic reasoning from that already intrinsic to the structure of `Nat` itself, when I first tried to shoehorn this proof into `Data.Nat.Properties`. 

It struck me before I embarked on this, but especially now afterwards, and with a categorical eye on what's going on, that it doesn't make sense to me to regard this homomorphism as a *property* inherent to `Nat`, importing structure from `Parity`: rather that each (through the particular lens of their respective `Semiring` structure) participates in the construction of a new structure/bundle, taking components from the (intrinsic/internal) `Properties` of each type `Nat` and `Parity`... and neither 'side' has ownership of the homomorphism. Rather, if rather prosaically, I've constructed an arrow in the (`Set`-/`Setoid-` based) category of `Semiring`s and their homomorphisms... such that each of the `Properties` modules is responsible for witnessing that `Nat` resp. `Parity` is even permitted to be considered as candidate domain/codomain for such an arrow. But *the arrow itself* lives/should live... somewhere else, associated in the hierarchy with the abstract *algebraic* structure involved. 

Accordingly, I suggest/propose that we have a new hierarchy below `Algebra.Homomorphism|Morphism`, corresponding to each of the (concrete) categories of algebras where the domain/codomain of any such homomorphism should live. 

[...]

PR #1894, opened 2022-12-21
`parity` is a `Semiring` homomorphism

jamesmckinna:
This PR follows up on the earlier #1852, and in the absence of any satisfactory resolution of issue #1871, makes the choice to put this under `Data.Parity`, moreover with an ugly/improvable module name. Suggestions welcome!


jamesmckinna:
Re: `CHANGELOG`
Should I add simply the final lemma, or all the new definitions, or simply the new module (name) to the log?

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
++-homo-+ (suc m) n = begin
+  parity (suc m + n)          ≡⟨ suc-+-homo-⁻¹ m n ⟩
+  (parity m) ⁻¹ +ℙ (parity n) ≡⟨ cong (_+ℙ (parity n)) (suc-homo-⁻¹ (suc m)) ⟩
+  parity (suc m) +ℙ parity n  ∎ where open ≡-Reasoning
+
+-- parity is a _*_ homomorphism
+
+*-homo-*      : ∀ m n → parity (m * n) ≡ (parity m) *ℙ (parity n)
+*-homo-* zero    n = refl
+*-homo-* (suc m) n with parity m in m≡p | parity n in n≡q
```
MatthewDaggitt, line 58: We're not pattern matching on the result of `with` so why not simply define `p = parity m` in the `where` block? Then no need for `m≡p` etc.

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
+... | p | q = begin
+  parity (suc m * n)                  ≡⟨⟩
+  parity (n + m * n)                  ≡⟨ +-homo-+ n (m * n) ⟩
+  parity n +ℙ parity (m * n)          ≡⟨ cong ((parity n) +ℙ_) (*-homo-* m n) ⟩
+  parity n +ℙ (parity m *ℙ parity n)  ≡⟨ cong (_+ℙ (parity m *ℙ parity n)) n≡q ⟩
+  q +ℙ (parity m *ℙ parity n)         ≡⟨ cong (q +ℙ_) (cong₂ _*ℙ_ m≡p n≡q) ⟩
+  q +ℙ (p *ℙ q)                       ≡⟨ lemma p q ⟩
+  ((p ⁻¹) *ℙ q)                       ≡⟨ cong (_*ℙ q) (cong _⁻¹ (sym m≡p)) ⟩
```
MatthewDaggitt, line 66: Unnecessary brackets on this line and the line below.

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Parity.HomomorphismNatParity where
+
+open import Data.Nat.Base
+  using (zero; suc; _+_; _*_; parity
+  ; +-rawMagma; +-0-rawMonoid
+  ; +-*-rawNearSemiring; +-*-rawSemiring)
+open import Data.Parity.Base using (Parity; 0ℙ; 1ℙ; _⁻¹)
```
MatthewDaggitt, line 15: The style we're moving towards in the standard library is to import the module `as ℙ` and then write `ℙ.+` rather than rename to `_+ℙ_`.

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
+-- successor and (_⁻¹)
+
+suc-homo-⁻¹ : ∀ n → (parity (suc n)) ⁻¹ ≡ parity n
+suc-homo-⁻¹ zero    = refl
+suc-homo-⁻¹ (suc n) = ⁻¹-inverts (suc-homo-⁻¹ n)
+
+-- parity is a _+_ homomorphism
+
++-homo-+      : ∀ m n → parity (m + n) ≡ (parity m) +ℙ (parity n)
+suc-+-homo-⁻¹ : ∀ m n → parity (suc m + n) ≡ (parity m) ⁻¹ +ℙ (parity n)
```
MatthewDaggitt, line 39: Why are these proofs mutually dependent? The definitions of  `+` and `⁻¹` are not mutually dependent so this feels very odd to me.

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
+
+-- successor and (_⁻¹)
+
+suc-homo-⁻¹ : ∀ n → (parity (suc n)) ⁻¹ ≡ parity n
+suc-homo-⁻¹ zero    = refl
+suc-homo-⁻¹ (suc n) = ⁻¹-inverts (suc-homo-⁻¹ n)
+
+-- parity is a _+_ homomorphism
+
++-homo-+      : ∀ m n → parity (m + n) ≡ (parity m) +ℙ (parity n)
```
MatthewDaggitt, line 38: Unnecessary brackets round the `(parity m)`?

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
...
+suc-+-homo-⁻¹ : ∀ m n → parity (suc m + n) ≡ (parity m) ⁻¹ +ℙ (parity n)
+
+suc-+-homo-⁻¹ zero    n = sym (suc-homo-⁻¹ (suc n))
+suc-+-homo-⁻¹ (suc m) n = begin
+  parity (suc (suc m) + n)          ≡⟨⟩
+  parity (m + n)                    ≡⟨ +-homo-+ m n ⟩
+  (parity m) +ℙ (parity n)          ≡⟨ cong (_+ℙ (parity n)) (sym (suc-homo-⁻¹ m)) ⟩
```
MatthewDaggitt, line 45: Again unnecessary brackets?

src/Data/Parity/HomomorphismNatParity.agda
```diff
@@ -0,0 +1,103 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Some properties about parities
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.Parity.HomomorphismNatParity where
```
MatthewDaggitt, line 9: While I agree we should do something about https://github.com/agda/agda-stdlib/issues/1871 in the medium to long term, I think it would be better to stick with the current convention used everywhere else in the library and put this in `Data.Parity.Properties`, rather than have two competing conventions.

MatthewDaggitt:
And of course a CHANGELOG entry would be great :+1:

src/Data/Parity/Properties.agda
```diff
@@ -420,58 +417,137 @@ toSign-injective {p} {q} eq = begin
...
+
++-homo-+      : ∀ m n → parity (m ℕ.+ n) ≡ parity m ℙ.+ parity n
++-homo-+ zero    n = refl
++-homo-+ (suc m) n = begin
+  parity (suc m ℕ.+ n)        ≡⟨ suc-+-homo-⁻¹ m n ⟩
+  (parity m) ⁻¹ ℙ.+ parity n   ≡⟨ cong (ℙ._+ parity n) (suc-homo-⁻¹ (suc m)) ⟩
+  parity (suc m) ℙ.+ parity n  ∎
+  where
+    open ≡-Reasoning
+    suc-+-homo-⁻¹ : ∀ m n → parity (suc m ℕ.+ n) ≡ (parity m) ⁻¹ ℙ.+ parity n
```
JacquesCarette, line 495: I would probably export this lemma as well, instead of burying it here?

src/Data/Parity/Properties.agda
```diff
@@ -420,58 +417,137 @@ toSign-injective {p} {q} eq = begin
...
+  q ℙ.+ (p ℙ.* q)             ≡⟨ lemma p q ⟩
+  ((p ⁻¹) ℙ.* q)              ≡⟨⟩
+  ((parity m) ⁻¹ ℙ.* q)       ≡⟨ cong (ℙ._* q) (suc-homo-⁻¹ (suc m)) ⟩
+  parity (suc m) ℙ.* q        ≡⟨⟩
+  parity (suc m) ℙ.* parity n ∎
+  where
+    open ≡-Reasoning
+    p = parity m
+    q = parity n
+    -- this lemma simplifies things a lot
```
JacquesCarette, line 520: Same for this lemma, pull it out. It feels like it ought to be useful in other proofs as well.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: You're absolutely right. Fixed now.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: Thanks! Fixed now.

JacquesCarette (commented):
I'm quite happy with these things being defined where they are. These are things that are quite special to the relationship between Nat and parity.

Any reason for all the `Effect` changes to be here? Some errant merge?

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: Thanks for the style hint. Fixed now, with 'symmetric' imports for `Nat` and `Parity` in that style, and revised import of `Sign` in the same spirit.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: Again, thanks. I had been unsure about the termination checker's willingness to check the recursive call on `+-homo-+` in the scope of the lemma `suc-+-homo-+`  but it turns out I needn't have been worried.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: ... and `(parity n)` it turns out. Fixed now.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: :+1:

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: Done. This module also now removed from the branch.

src/Data/Parity/Properties.agda
jamesmckinna: Here, I disagree, or at least: if I were to do so, it would reinstate the mutual dependency between this lemma (clearly auxiliary to `+-homo-+`) which @MatthewDaggitt objected to above. On balance, I though he was correct. The other lemma, `suc-homo-⁻¹`, on th eother hand, is independently useful, I think.

JacquesCarette: Fair enough.

src/Data/Parity/Properties.agda
jamesmckinna: Perhaps you're right. It didn't 'smell' right to me as a property of independent interest. I think that it follows 'easily' (at least: I believe it does, but *laboriously*) from `p+p≡0ℙ`, `p⁻¹+p≡1ℙ`, `*-identityˡ`, and `*-distribˡ-+` , but the calculational proof here was definitely quicker/slicker...

jamesmckinna:
And yes, an errant merge. Blast. Pause while I roll evrything back again and re-commit. Grrr.

src/Data/Parity/HomomorphismNatParity.agda
jamesmckinna: I think resolved, but see @JacquesCarette 's suggestion contra below.

src/Data/Parity/Properties.agda
JacquesCarette: I've been there before: a property that doesn't seem to naturally arise from the external API, but is somehow key in a proof and general.

My experience has been that such key lemmas show up again later, so I pull them out. [I still have multiple passes to do on agda-categories to refactor multiple proofs to use the patterns that were discovered after the proofs were laboriously hand-written.]

It's a very subjective call.

JacquesCarette:
Things are looking great. Except that you seem to have lost your CHANGELOG changes while fixing the errant merge.

jamesmckinna:
> Things are looking great. Except that you seem to have lost your CHANGELOG changes while fixing the errant merge.

Fixed now?

JacquesCarette:
Yes, everything's looking good now.

jamesmckinna:
OK, checked out a clean branch (but maybe from the wrong base?), but the associated PR doesn't look right...

PR #1895, opened 2022-12-30
Use new style function hierarchy everywhere.

MatthewDaggitt:
Part of #759


MatthewDaggitt:
Gah, this is more interconnected than I'd hoped. Need to push it through `Data.Product.Function` as well...

JacquesCarette:
And right now, build failed with
```
/home/runner/work/agda-stdlib/agda-stdlib/src/Function/Related/TypeIsomorphisms.agda:181,21-29
x ∼[ ⌊ k ⌋ ] y !=< _A_802 ∼[ _k_809 ]′ _B_804
when checking that the expression _⊎-cong_ has type
Related ⌊ k ⌋ x y →
Related ⌊ k ⌋ u v → Related ⌊ k ⌋ (x ⊎ u) (y ⊎ v)
```

JacquesCarette:
Ah, no, I see the monster proof that was `Injection` before, that's the nightmare that I have to reproduce.

MatthewDaggitt:
Okay this is finally ready. All uses of the old function hierarchy have been stripped out and been replaced with the new ones :tada: There are a couple of follow up issues that I'll file in the morning.

JacquesCarette:
This is fantastic. This stuff was ridiculously finicky. No one should be doing this kind of transport wrt propositional equality!

PR #1896, opened 2022-12-30
Add proofs for List mapWith∈

sgodwincs:
This was suggested following a question in the Zulip chat (https://agda.zulipchat.com/#narrow/stream/259644-newcomers/topic/Trying.20to.20prove.20a.20proposition.20about.20lists)

### DIFF
--- a/src/Data/List/Membership/Propositional/Properties.agda
+++ b/src/Data/List/Membership/Propositional/Properties.agda
@@ -80,6 +80,13 @@ mapWith∈≗map : ∀ (f : A → B) xs → mapWith∈ xs (λ {x} _ → f x) ≡
 mapWith∈≗map f xs =
   ≋⇒≡ (Membershipₛ.mapWith∈≗map (P.setoid _) (P.setoid _) f xs)
 
+mapWith∈-id : (xs : List A) → mapWith∈ xs (λ {x} _ → x) ≡ xs
+mapWith∈-id = Membershipₛ.mapWith∈-id (P.setoid _)
+
+map∘mapWith∈ : (xs : List A) (f : ∀ {x} → x ∈ xs → B) (g : B → C) →
+               map g (mapWith∈ xs f) ≡ mapWith∈ xs (g ∘′ f)
+map∘mapWith∈ = Membershipₛ.map∘mapWith∈ (P.setoid _)
+
 ------------------------------------------------------------------------
 -- map
 
--- a/src/Data/List/Membership/Propositional/Properties.agda
+++ b/src/Data/List/Membership/Propositional/Properties.agda
@@ -83,9 +83,9 @@ mapWith∈≗map f xs =
 mapWith∈-id : (xs : List A) → mapWith∈ xs (λ {x} _ → x) ≡ xs
 mapWith∈-id = Membershipₛ.mapWith∈-id (P.setoid _)
 
-map∘mapWith∈ : (xs : List A) (f : ∀ {x} → x ∈ xs → B) (g : B → C) →
+map-mapWith∈ : (xs : List A) (f : ∀ {x} → x ∈ xs → B) (g : B → C) →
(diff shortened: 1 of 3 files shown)


jamesmckinna:
For (my) info, as well as this particular PR: is there any style(-guide) preference for proofs involving `_≡_` to use `rewrite` vs. explicit appeals to `cong`?

JacquesCarette:
I've grown to dislike `rewrite`. It seems fragile (on more complex things) and has broken between versions. I only use it in a classroom setting to show students the various styles, never in my own code. But I don't recall what the style guide says about it. Certainly there are very few uses of it in the library itself.

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -80,6 +80,13 @@ mapWith∈≗map : ∀ (f : A → B) xs → mapWith∈ xs (λ {x} _ → f x) ≡
 mapWith∈≗map f xs =
   ≋⇒≡ (Membershipₛ.mapWith∈≗map (P.setoid _) (P.setoid _) f xs)
 
+mapWith∈-id : (xs : List A) → mapWith∈ xs (λ {x} _ → x) ≡ xs
+mapWith∈-id = Membershipₛ.mapWith∈-id (P.setoid _)
+
+map∘mapWith∈ : (xs : List A) (f : ∀ {x} → x ∈ xs → B) (g : B → C) →
```
MatthewDaggitt, line 86: Could we change this name to the slightly more standard `map-mapWith∈`? Same in `Setoid.Properties`.

src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -122,6 +122,15 @@ module _ (S : Setoid c ℓ) where
                     length (mapWith∈ xs f) ≡ length xs
   length-mapWith∈ []       = P.refl
   length-mapWith∈ (x ∷ xs) = P.cong suc (length-mapWith∈ xs)
+  
+  mapWith∈-id : ∀ xs → mapWith∈ xs (λ {x} _ → x) ≡ xs
+  mapWith∈-id [] = P.refl
+  mapWith∈-id (x ∷ xs) rewrite mapWith∈-id xs = P.refl
+
+  map∘mapWith∈ : ∀ {a b} {A : Set a} {B : Set b} xs (f : ∀ {x} → x ∈ xs → A) (g : A → B) →
```
MatthewDaggitt, line 130: Very minor but could we break this line at `∀ {a b} {A : Set a} {B : Set b}→` and `∀ xs (f : ∀ {x} → x ∈ xs → A) (g : A → B)`?

MatthewDaggitt (commented):
Thanks for the great first PR! I've added a couple of minor comments to fix, but apart from that looks good.

MatthewDaggitt:
I agree with @JacquesCarette comments about the fragility of `rewrite`, however we don't yet have a standard policy not to use it. @sgodwincs if you're happy eliminating it then go for it, otherwise I'm okay with them being left in.

sgodwincs:
I've updated it to not use rewrite, though not sure I did it in the most readable fashion.

jamesmckinna (commented):
Version without `rewrite` looks fine, but as with my comment above, it is perhaps a bit 'hairy' in the case of `map-mapWith∈` compared to the `rewrite`-based proof. 

RHS of the second equation of that definition with 
```agda
P.cong ((g (f (here refl))) ∷_) (map∘mapWith∈ xs (f ∘ there) g)
```
is arguably no better, but perhaps a bit shorter? 

As for `mapWith∈-id`, it's fine (better?) to use a section for the `cong` function, viz.
```agda
P.cong (x ∷_ ) (mapWith∈-id xs)
```

sgodwincs:
Updated per your suggestion

MatthewDaggitt:
Merged in. Thanks again for the great first PR :+1:

Issue #1888, opened 2022-12-16
Should Semimodules and Modules have a rule saying that left- and right-multiplication are the same?


Taneb:
i.e. `∀ x m → x *ₗ m ≈ᴹ m *ᵣ x`.

This can't be stated for `Bisemimodule` and `Bimodule` but for `Semimodule` and `Module` it feels a useful property to have that we can't currently prove for a general module.

MatthewDaggitt:
I have no idea about module theory, so I'm not really equipped to answer this one.

jamesmckinna:
I've been doing some work recently with modules, and I think this property is *not* in general provable. But best place to ask: mathoverflow imho

Taneb:
I've found an example of a module (by our definition) where the two operations are different.

Take G to be the usual commutative of Gaussian integers (complex numbers of the form x + i*y for x, y integers).

Then G acts as a left module over itself with multiplication.
We can also take G to be a right module over itself with the operation `\x y -> x * conjugate y`.

With these two definitions, we can build a module according to stdlib's understanding of the term. See here: https://gist.github.com/Taneb/64cd57ef1896a820b27921131cfab5a6

I claim this is strongly undesirable.

JacquesCarette:
It took me a while to reconstruct what @Taneb found objectionable, so I'll expand it out in this comment: normally a module has a single notion of multiplication, not a left one and a right one. If one builds up module by joining left/right modules, then it's possible to create a module where they are different.

The fundamental issue here is that there is no simple syntactic embedding going on here. The relation between the theories involved are an almost-trivial theorem, but not a trivial theorem (i.e. the proof is not exactly 'refl').

The simplest fix is as the title says: it should be an axiom that left- and right- multiplication amount to the same thing. They are not quite 'the same' in the strong sense of 'the same', but they are trivially equivalent...

jamesmckinna:
But see also the discussion [here](https://math.stackexchange.com/questions/1493609/right-module-vs-left-module), as well as [this](https://mathoverflow.net/questions/64370/simplest-examples-of-rings-that-are-not-isomorphic-to-their-opposites) which links from the first; also [this]()https://math.stackexchange.com/questions/535246/on-similarities-and-differences-between-right-and-left-modules-over-a-ring). 

[...]

PR #1900, opened 2023-01-05
opposite of a `Ring`

jamesmckinna:
See discussion in issue #1888  and associated pr #1898

### DIFF
--- a/src/Algebra/Construct/Flip/Op.agda
+++ b/src/Algebra/Construct/Flip/Op.agda
@@ -264,3 +264,7 @@ group g = record { isGroup = isGroup g.isGroup }
 abelianGroup : AbelianGroup a ℓ → AbelianGroup a ℓ
 abelianGroup g = record { isAbelianGroup = isAbelianGroup g.isAbelianGroup }
   where module g = AbelianGroup g
+
+ring : Ring a ℓ → Ring a ℓ
+ring r = record { isRing = isRing r.isRing }
+  where module r = Ring r
--- a/src/Algebra/Construct/Flip/Op.agda
+++ b/src/Algebra/Construct/Flip/Op.agda
@@ -53,6 +53,14 @@ module _ (≈ : Rel A ℓ) (∙ : Op₂ A) where
   inverse : Inverse ≈ ε ⁻¹ ∙ → Inverse ≈ ε ⁻¹ (flip ∙)
   inverse inv = Prod.swap inv
 
+  zero : Zero ≈ ε ∙ → Zero ≈ ε (flip ∙)
+  zero zer = Prod.swap zer
+
+  module _ (+ : Op₂ A) where
+
+    distributes : (≈ DistributesOver ∙) + → (≈ DistributesOver (flip ∙)) +
+    distributes distrib = Prod.swap distrib
+
 ------------------------------------------------------------------------
 -- Structures
 
@@ -162,6 +170,23 @@ module _ {≈ : Rel A ℓ} {∙ : Op₂ A} where
     ; comm    = commutative ≈ ∙ g.comm
     }
     where module g = IsAbelianGroup g
+
+module _ {≈ : Rel A ℓ} {+ * : Op₂ A} { - : Op₁ A} {0# 1# : A} where
+
+  isRing : IsRing ≈ + * - 0# 1# → IsRing ≈ + (flip *) - 0# 1#
+  isRing r = record
+    { +-isAbelianGroup = r-isAbelianGroup
+    ; *-cong = preserves₂ ≈ ≈ ≈ r.*-cong
+    ; *-assoc = associative ≈ * sym r.*-assoc
+    ; *-identity = identity ≈ * r.*-identity
+    ; distrib = distributes ≈ * + r.distrib
+    ; zero = zero ≈ * r.zero
+    }
+    where
+      module r = IsRing r
+      r-isAbelianGroup = r.+-isAbelianGroup
+      open IsAbelianGroup r-isAbelianGroup using (sym)
+
 
 ------------------------------------------------------------------------
 -- Bundles
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1587,11 +1587,12 @@ Other minor changes
   moufangLoop : MoufangLoop a ℓ₁ → MoufangLoop b ℓ₂ → MoufangLoop (a ⊔ b) (ℓ₁ ⊔ ℓ₂)
  ```
 
-* Added new proofs to `Algebra.Construct.Flip.Op`:
+* Added new functions and proofs to `Algebra.Construct.Flip.Op`:
   ```agda
   zero : Zero ≈ ε ∙ → Zero ≈ ε (flip ∙)
   distributes : (≈ DistributesOver ∙) + → (≈ DistributesOver (flip ∙)) +
   isRing : IsRing ≈ + * - 0# 1# → IsRing ≈ + (flip *) - 0# 1#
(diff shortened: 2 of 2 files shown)


JacquesCarette (approved):
Nice.

But also: the infamous 'Op' meta-functor. Claudio Sacerdoti Coen and I spent two full afternoons in front a blackboard working out the details of what it actually means. And how it ought to be automated. We never got around to writing anything up, unfortunately.

jamesmckinna:
Well, re not-written-up work: I'd love to see notes (or rational reconstruction) of that discussion with Claudio, but I suppose I was wanting to... for once ;-)... not get too caught up with the 'in full generality' situation, in the spirit of your previous comments about writing 'smaller' PRs...

jamesmckinna:
And: should such an 'Op' reverse the equivalence relation? 

I had considered this, with a view (de-strictifying, bicategories vs. 2-categories etc.) to having Op on Setoids being an iso, not a definitional equality... but *not* today...

JacquesCarette:
Looked for my notes, and didn't find them in the notebooks where I expected them to be. Sigh. Ah, but I do have pictures of the boards!

So whether 'Op' should reverse the equivalence relation indeed is mixed up with whether you're looking at things 1- or 2-categorically. Of course, 2-categorically, there are 3 choices of what to flip, so that there should be 3 'flip' operations.

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -53,6 +53,14 @@ module _ (≈ : Rel A ℓ) (∙ : Op₂ A) where
   inverse : Inverse ≈ ε ⁻¹ ∙ → Inverse ≈ ε ⁻¹ (flip ∙)
   inverse inv = Prod.swap inv
 
+  zero : Zero ≈ ε ∙ → Zero ≈ ε (flip ∙)
+  zero zer = Prod.swap zer
+
+  module _ (+ : Op₂ A) where
```
MatthewDaggitt, line 59: I know its not the most efficient, but we don't usually nest anonymous modules like this for readability reasons. They tend to very quickly get out of hand and the code becomes difficult to pass. Instead we just list all the arguments again. It also allows you to rename the `∙` operator to the more standard `*` for rings.

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -163,6 +171,23 @@ module _ {≈ : Rel A ℓ} {∙ : Op₂ A} where
...
+    ; *-cong = preserves₂ ≈ ≈ ≈ r.*-cong
+    ; *-assoc = associative ≈ * sym r.*-assoc
+    ; *-identity = identity ≈ * r.*-identity
+    ; distrib = distributes ≈ * + r.distrib
+    ; zero = zero ≈ * r.zero
+    }
+    where
+      module r = IsRing r
+      r-isAbelianGroup = r.+-isAbelianGroup
+      open IsAbelianGroup r-isAbelianGroup using (sym)
```
MatthewDaggitt, line 188: Why can we not just use `r.sym`?

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -163,6 +171,23 @@ module _ {≈ : Rel A ℓ} {∙ : Op₂ A} where
...
+    { +-isAbelianGroup = r-isAbelianGroup
+    ; *-cong = preserves₂ ≈ ≈ ≈ r.*-cong
+    ; *-assoc = associative ≈ * sym r.*-assoc
+    ; *-identity = identity ≈ * r.*-identity
+    ; distrib = distributes ≈ * + r.distrib
+    ; zero = zero ≈ * r.zero
+    }
+    where
+      module r = IsRing r
+      r-isAbelianGroup = r.+-isAbelianGroup
```
MatthewDaggitt, line 187: I don't think this redirection is needed?

src/Algebra/Construct/Flip/Op.agda
```diff
@@ -53,6 +53,14 @@ module _ (≈ : Rel A ℓ) (∙ : Op₂ A) where
   inverse : Inverse ≈ ε ⁻¹ ∙ → Inverse ≈ ε ⁻¹ (flip ∙)
   inverse inv = Prod.swap inv
 
+  zero : Zero ≈ ε ∙ → Zero ≈ ε (flip ∙)
+  zero zer = Prod.swap zer
+
+  module _ (+ : Op₂ A) where
+
+    distributes : (≈ DistributesOver ∙) + → (≈ DistributesOver (flip ∙)) +
```
MatthewDaggitt, line 61: The other thing we could do to make this is a little nicer is to import `Algebra.Definitions` and `Algebra.Structures` locally in each anonymous module parameterised by the equality. Which would then allow us to avoid messing up the infix notation here?

MatthewDaggitt:
> So whether 'Op' should reverse the equivalence relation indeed is mixed up with whether you're looking at things 1- or 2-categorically. Of course, 2-categorically, there are 3 choices of what to flip, so that there should be 3 'flip' operations.

Or 8 operations? For `Relation.Binary.Construct.Flip` I've used 3 out of the 4 possible permutations in my own work.

src/Algebra/Construct/Flip/Op.agda
jamesmckinna: It seems that we can. This shows my continued failure to understand the intricacies of how each structure/bundle systematically re-exports all its substructures/bundles as 'flat' definitions (always assuming that that is indeed what is going on...)

src/Algebra/Construct/Flip/Op.agda
jamesmckinna: Indeed! As above...

src/Algebra/Construct/Flip/Op.agda
jamesmckinna: Ah... will now turn to this. 

And it seems that this applies throughout, even to the original module contents: so I've now named each previously anonymous module, and opened it locally in place with the appropriate arguments. is this better now? See most recent commit.

jamesmckinna:
I won't comment on the multiplicities of `Flip` and `Op`, after Matthew's successful resolution of the `sym` problem which I had previously had. But starting to scratch my head about the use of only 3 out of 4 possibilities in his work...

JacquesCarette:
Isn't the 4th possibility 'do nothing', which is why it's hard to see it? We're talking about a group with 4 elements, which is the product of the group with 2 elements, so it's rather natural that only 3 of them show up!

jamesmckinna:
D'oh!

jamesmckinna:
Oh man, it's too late in the evening to be doing this. I've let yet more bogus commits enter the PR. :facepalm: And it clearly goes back to before the most recent commit/revert. @#$%

jamesmckinna:
OK, I've screwed this one up as well. Converting to draft and then sitting on my hands for a few days.

jamesmckinna:
Nope... those reversions didn't work either.

MatthewDaggitt:
> Isn't the 4th possibility 'do nothing', which is why it's hard to see it? We're talking about a group with 4 elements, which is the product of the group with 2 elements, so it's rather natural that only 3 of them show up!

Haha yes. Whoops, sorry for throwing out that confusing statement :sweat_smile:

jamesmckinna:
To follow up on my comments form a week ago (or more). I'll start a fresh branch and roll up the commits form this one, and then open a fresh PR. So this one moved to 'Draft' while I thought about it some more, and of course lots has changed in the meantime, but this one should be a very 'local' change.

jamesmckinna:
superseded by #1910

PR #1901, opened 2023-01-05
'Sufficient' lists

jamesmckinna:
A structurally inductive view of lists

### DIFF
--- a/src/Data/List/Sufficient.agda
+++ b/src/Data/List/Sufficient.agda
@@ -18,11 +18,6 @@ module Data.List.Sufficient {a} {A : Set a} where
 open import Data.List.Base using (List; []; _∷_; _++_)
 open import Relation.Binary.PropositionalEquality using (_≡_; refl)
 
-private
-  variable
-    x : A
-    xs : List A
-
 
 ------------------------------------------------------------------------
 -- Sufficient view
@@ -30,27 +25,32 @@ private
 data Sufficient : (xs : List A) → Set a where
 
   suff-acc : ∀ {xs} →
-             (suff_ih : ∀ {a} {pre suff} → xs ≡ a ∷ (pre ++ suff) →
-                        Sufficient suff) →
+             (suff_ih : ∀ {a} {prefix suffix} → xs ≡ a ∷ (prefix ++ suffix) →
+                        Sufficient suffix) →
              Sufficient xs
 
 
 ------------------------------------------------------------------------
--- Sufficient properties: constructors
+-- Sufficient properties: constructors (typically not for export)
 
-[]⁺ : Sufficient []
-[]⁺ = suff-acc λ ()
+module Constructors where
 
-∷⁺ : Sufficient xs → Sufficient (x ∷ xs)
-∷⁺ {xs = xs} suff-xs@(suff-acc acc) = suff-acc λ { refl → suf _ refl }
-  where
-    suf : ∀ pre {suff} → xs ≡ pre ++ suff → Sufficient suff
-    suf []        refl = suff-xs
-    suf (a ∷ pre) eq   = acc eq
+  []⁺ : Sufficient []
+  []⁺ = suff-acc λ ()
+
+  ∷⁺ : ∀ {x} {xs} → Sufficient xs → Sufficient (x ∷ xs)
+  ∷⁺ {xs = xs} suff-xs@(suff-acc acc) = suff-acc λ { refl → suf _ refl }
+    where suf : ∀ prefix {suffix} → xs ≡ prefix ++ suffix → Sufficient suffix
+          suf []           refl = suff-xs
+          suf (a ∷ prefix) eq   = acc eq
 
 ------------------------------------------------------------------------
 -- Sufficient view covering property
 
-sufficient : (xs : List A) → Sufficient xs
-sufficient []       = []⁺
-sufficient (x ∷ xs) = ∷⁺ (sufficient xs)
+module _ where
+
+  open Constructors
+
+  sufficient : (xs : List A) → Sufficient xs
+  sufficient []       = []⁺
+  sufficient (x ∷ xs) = ∷⁺ (sufficient xs)
--- a/src/Data/List/Sufficient.agda
+++ b/src/Data/List/Sufficient.agda
@@ -16,34 +16,51 @@
 
 module Data.List.Sufficient {a} {A : Set a} where
 
-open import Data.List.Base using (List; []; _∷_; _++_)
+open import Level using (_⊔_)
+open import Data.List.Base using (List; []; _∷_; [_]; _++_)
 open import Relation.Binary.PropositionalEquality using (_≡_; refl)
 
+------------------------------------------------------------------------
+-- Sufficient builder
+
+suffAcc : ∀ {b} (B :  List A → Set b) (xs : List A) → Set (a ⊔ b)
+suffAcc B xs = ∀ {a} {prefix} suffix → xs ≡ a ∷ prefix ++ suffix → B suffix
 
 ------------------------------------------------------------------------
 -- Sufficient view
 
 data Sufficient : (xs : List A) → Set a where
 
-  suff-acc : ∀ {xs} →
-             (suff_ih : ∀ {a} {prefix suffix} → xs ≡ a ∷ prefix ++ suffix →
-                        Sufficient suffix) →
-             Sufficient xs
+  acc : ∀ {xs} (ih : suffAcc Sufficient xs) → Sufficient xs
 
 
 ------------------------------------------------------------------------
--- Sufficient properties: constructors (typically not for export)
+-- Sufficient properties
+
+-- constructors (typically not for export)
 
 module Constructors where
 
   []⁺ : Sufficient []
-  []⁺ = suff-acc λ ()
+  []⁺ = acc λ _ ()
 
   ∷⁺ : ∀ {x} {xs} → Sufficient xs → Sufficient (x ∷ xs)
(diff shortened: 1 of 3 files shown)


src/Data/List/Sufficient.agda
```diff
@@ -0,0 +1,57 @@
...
+data Sufficient : (xs : List A) → Set a where
+
+  suff-acc : ∀ {xs} →
+             (suff_ih : ∀ {a} {prefix suffix} → xs ≡ a ∷ prefix ++ suffix →
+                        Sufficient suffix) →
+             Sufficient xs
+
+
+------------------------------------------------------------------------
+-- Sufficient properties: constructors (typically not for export)
```
JacquesCarette, line 35: So make them `private`?

src/Data/List/Sufficient.agda
```diff
@@ -0,0 +1,57 @@
...
+
+------------------------------------------------------------------------
+-- Sufficient view covering property
+
+module _ where
+
+  open Constructors
+
+  sufficient : (xs : List A) → Sufficient xs
+  sufficient []       = []⁺
```
JacquesCarette, line 56: This worries me. It feels like a time bomb. Shouldn't it be possible to prove that the empty is never `Sufficient` ? I'm really not understanding something about `suff-acc`, it shouldn't be happy to be given an impossible function!

My guess is that I'm misreading this somehow, and yet...

JacquesCarette (commented):
Comments/questions only.

This does seem nice, modulo caveat.

src/Data/List/Sufficient.agda
jamesmckinna: So... rather like the discussion if other recent work, I supposed there might be cases where clients would want access to them; so I guess thus represents a... hesitation on my part.

src/Data/List/Sufficient.agda
jamesmckinna: Good question. 
I think the analogy with the base cases of WF induction is appropriate here: if we have an IH installed which for all x gives P y for y < x, then certainly P 0, because it has no <-predecessors....

src/Data/List/Sufficient.agda
jamesmckinna: Further to that: I realised that I didn't know how to mark the constructors off as being '*likely* to be private/hidden during `import`', but without committing to that choice, and then finding that I needed immediately (!)  to use them in the scope of `sufficient`... so perhaps I don't need the explicit `module _ where` around *that* definition? 

is it the case that the `open Constructors` *won't*  export them from the module (whereas `open ... public` *would*)? I confess I still trip up over this in Agda... :-(

jamesmckinna:
On `suff-acc`: perhaps it would be better (? for a suitable definition of 'better'...?) to overload the constructor name `acc` from `induction.*`? I think names for things are important; but on a first cut of a branch/PR, I can be a bit cavalier about such things.

jamesmckinna:
Latest commit represents a considerable extension/refactoring, mostly to lift out the `suffAcc` predicate builder, and write `suffRec` to encapsulate appeals to the view. 

Not clear to me if it's an improvement?

jamesmckinna:
Grrr. Might need to rethink the implicit/explicit arguments in this refactoring. Apologies.

src/Data/List/Sufficient.agda
jamesmckinna: Added commit demonstrating recursion on the view.

jamesmckinna:
It (now) occurs to me that this (perhaps) properly belongs in `Data.List.Relation.Unary.Sufficient`, not as now at the top-level under `Data.List`. Advice please!

src/Data/List/Sufficient.agda
JacquesCarette: Thanks, I think this answers my questions.

src/Data/List/Sufficient.agda
```diff
@@ -0,0 +1,87 @@
...
+
+-- destructors
+
+module Destructors where
+
+  ++⁻ : ∀ xs {ys} → Sufficient (xs ++ ys) → Sufficient ys
+  ++⁻ []            suff-ys   = suff-ys
+  ++⁻ (x ∷ xs) {ys} (acc hyp) = hyp ys refl
+
+  ∷⁻ : ∀ {x} {xs} → Sufficient (x ∷ xs) → Sufficient xs
```
MatthewDaggitt, line 62: Minor but we can write `{x xs}` although I would be tempted to turn them into `variable`s

src/Data/List/Sufficient.agda
```diff
@@ -0,0 +1,87 @@
...
+
+-- constructors (typically not for export)
+
+module Constructors where
+
+  []⁺ : Sufficient []
+  []⁺ = acc λ _ ()
+
+  ∷⁺ : ∀ {x} {xs} → Sufficient xs → Sufficient (x ∷ xs)
+  ∷⁺ {xs = xs} suff-xs@(acc hyp) = acc λ { _ refl → suf _ refl }
```
MatthewDaggitt, line 48: Can we avoid using `-` in variable names?

MatthewDaggitt:
> It (now) occurs to me that this (perhaps) properly belongs in Data.List.Relation.Unary.Sufficient, not as now at the top-level under Data.List. Advice please!

Yes it absolutely does :+1:

MatthewDaggitt (commented):
Sorry for taking so long for the review!

src/Data/List/Sufficient.agda
jamesmckinna: Sure! I had no idea that A Thing...

jamesmckinna:
Regarding the visibility of exports from this module, one reason in the first instance for defining the `Constructors` was to establish admissibility ('sufficiency', sic) of the view covering function `sufficient`; I'm not sure that it's (the predicate even) independently interesting to export properties of the predicate when *really* what you want this for is to do recursion on the view, now pre-packaged as `suffRec`. But I will look again at `Induction.WellFounded` and friends for (suitable) inspiration as to what seems the 'good interface to export. 

I like the library style of encapsulation in modules things which *need not, but could* be exported: the main outstanding issue (for me, at least) is a reasoning principle which can be used to propositionally unfold `suffRec`, so that one doesn't, once again, have to delve into the view itself in order to support such reasoning (I find this to be the bane of `WF` arguments, cf. Ulf's remarks about dealing directly with recursion on 'accessibility'). But maybe I should save that for a follow-up PR?

src/Data/List/Sufficient.agda
```diff
@@ -0,0 +1,87 @@
...
+-- which provably consume a non-empty (but otherwise arbitrary) prefix
+-- *without* having to resort to ancillary WF induction on length etc.
+-- e.g. lexers, parsers etc.
+--
+-- Credited by Conor McBride as originally due to James McKinna
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+module Data.List.Sufficient {a} {A : Set a} where
```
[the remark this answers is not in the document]
jamesmckinna: So I tried this... and the very first definition of `suffAcc` breaks:
```agda
private
  variable

    a : Level
    A : Set a
    x : A
    xs : List A

-- Sufficient builder

suffAcc : ∀ {b} (B :  List A → Set b) (xs : List A) → Set (a ⊔ b)
suffAcc B xs = ∀ {x} {prefix} suffix → xs ≡ x ∷ prefix ++ suffix → B suffix
```
yielding
```agda
Set (A.a ⊔ b) != Set (a ⊔ b)
∀ {x} {prefix} suffix → xs ≡ x ∷ prefix ++ suffix → B suffix has
type Set (a ⊔ b)
```
and I realise that I *hate* these kinds of errors (which arise in my experience  precisely, and only, from using `variable` declarations... :-() because I just don't understand how I am supposed to fix them (because they make no sense: surely `suffAcc` is defined within their scope, so that `a`, `A`etc. are also in scope ...

... so some Agda hand-holding would be welcome here!

src/Data/List/Sufficient.agda
jamesmckinna: Done

src/Data/List/Sufficient.agda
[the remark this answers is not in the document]
Taneb: In that particular case you'd need to explicitly have `{A : Set a}`. The hidden `a` generated by `A` is distinct from the `a` you explicitly wrote, I guess so that you can alpha-rename things without them breaking mysteriously. https://agda.readthedocs.io/en/v2.6.2.2/language/generalization-of-declared-variables.html explains them in detail.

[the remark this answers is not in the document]
jamesmckinna: Thanks @Taneb I had just re-read the explanation in detail, and I still don't have an adequate internal mental model of what it's telling me... and in particular *why* I need the additional outermost parametric declaration... but commit `a53300d` seems to do the trick (but with some additional knock-on effects, sigh)

src/Data/List/Sufficient.agda
jamesmckinna: OK, but as below (comments maybe got put in wrong place), I'd rather leave `Constructors` as *solely* relevant to constructing the view covering function `sufficient`.

src/Data/List/Sufficient.agda
jamesmckinna: There's one more use to snag... I'll commit today if I can.

jamesmckinna: Done.

jamesmckinna:
I think this is now ready to go!

Issue #1828, opened 2022-09-20
Should Modules have raw bundles?


Taneb:
Currently, for each bundle defined in `Algebra.Bundles`, there is a corresponding raw bundle, consisting of the operations but without the laws. The same is not true for `Algebra.Module.Bundles`. Should they be added?

Taneb:
Further question: should there be a distinct `RawSemimodule`/`RawModule`? This works out the same as `RawBisemimodule`/`RawBimodule` with its ring argument repeated

MatthewDaggitt:
So in general, if two structures have the same signature but different laws then we only have a single raw bundle for them.

jamesmckinna:
I, too would feel bad, especially for the interaction with #1898 ... and the fact that `BimoduleMorphisms` have *two* fields for the homomorphism property wrt scalar multiplication. I really think that `ModuleMorphisms` should just have the one... and of course give rise in the usual way by duplication to the associated `BimoduleMorphisms` etc structure... 

... but perhaps I am missing the force both of @MatthewDaggitt 's comment, and yours. But in practical terms, if I were working with modules, I shouldn't want to be dropping down into substructures for the sake of stdlib style conventions. Not being able to rename record fields is a pain, so commitments to any distinguished choice of those taking priority over others... seems like something requiring further discussion?

jamesmckinna:
Any progress on this @Taneb ?

Taneb:
I've just independently re-motivated myself for this. Going to give it another push, hopefully won't take long.

PR #1902, opened 2023-01-07
initial+terminal algebras

jamesmckinna:
Refactored `Algebra.Construct.Zero` into
- [x] `Algebra.Construct.Initial`
- [x] `Algebra.Construct.Terminal`

and fixed up the public (re-)imports of the second into `Algebra.Construct.Zero` (for backwards compatibility), while deprecating the occurrences of the non-initial terminal algebras in favour of their counterparts in `Algebra.Construct.Terminal`. 

Also: amended `Algebra.Module.Construct.Zero` with commented-out hooks for the various `Raw` bundles associated with module-like algebras, pending the resolution of issue #1828 

One wrinkle: I wanted to abstract out all the common structure, so i did that in each module via an additional private module, that nevertheless is `open`ed into order to make the definitions (transparently) visible. But I may have got my `private`/`public` modifiers a bit mixed up. My concern is whether by making the internal modules `private`, I somehow fail to transparently export the definitions inside the publicly exported (raw) bundles, so that no actually reasoning is possible with them. But barring this wrinkle (fixable by removing the `private` modifiers), I think this one is now ready to go.

### DIFF
--- a/src/Algebra/Construct/Initial.agda
+++ b/src/Algebra/Construct/Initial.agda
@@ -26,13 +26,13 @@ open import Relation.Binary using (Rel; Reflexive; Symmetric; Transitive; IsEqui
 
 
 ------------------------------------------------------------------------
--- re-export those algebras which are also terminal
+-- Re-export those algebras which are also terminal
 
 open import Algebra.Construct.Terminal {c} {ℓ} public
   hiding (rawMagma; magma; semigroup; band)
 
 ------------------------------------------------------------------------
--- gather all the functionality in one place
+-- Gather all the functionality in one place
 
 private module ℤero where
 
--- a/src/Algebra/Construct/Initial.agda
+++ b/src/Algebra/Construct/Initial.agda
@@ -34,7 +34,7 @@ open import Algebra.Construct.Terminal {c} {ℓ} public
 ------------------------------------------------------------------------
 -- Gather all the functionality in one place
 
-private module ℤero where
+module ℤero where
 
   infixl 7 _∙_
   infix  4 _≈_
@@ -66,13 +66,13 @@ open ℤero
 -- Raw bundles
 
 rawMagma : RawMagma c ℓ
-rawMagma = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ }
+rawMagma = record { ℤero }
 
 ------------------------------------------------------------------------
 -- Structures
 
 isEquivalence : IsEquivalence _≈_
-isEquivalence = record { refl = refl ; sym = sym ; trans = trans }
+isEquivalence = record { ℤero }
 
 isMagma : IsMagma _≈_ _∙_
 isMagma = record { isEquivalence = isEquivalence ; ∙-cong = ∙-cong }
--- a/src/Algebra/Construct/Initial.agda
+++ b/src/Algebra/Construct/Initial.agda
@@ -1,0 +1,92 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Instances of algebraic structures where the carrier is ⊥.
+-- In mathematics, this is usually called 0.
+--
+-- From monoids up, these are zero-objects – i.e, the initial
+-- object is the terminal object in the relevant category.
+------------------------------------------------------------------------
+
+{-# OPTIONS --without-K --safe #-}
+
+open import Level using (Level)
+
+module Algebra.Construct.Initial {c ℓ : Level} where
+
+open import Algebra.Bundles
+  using (Magma; Semigroup; Band)
+open import Algebra.Bundles.Raw
+  using (RawMagma)
+open import Algebra.Core using (Op₂)
+open import Algebra.Definitions using (Congruent₂)
+open import Algebra.Structures using (IsMagma; IsSemigroup; IsBand)
+open import Data.Empty.Polymorphic
+open import Relation.Binary using (Rel; Reflexive; Symmetric; Transitive; IsEquivalence)
+
+
+------------------------------------------------------------------------
+-- re-export those algebras which are also terminal
+
+open import Algebra.Construct.Terminal {c} {ℓ} public
+  hiding (rawMagma; magma; semigroup; band)
+
+------------------------------------------------------------------------
+-- gather all the functionality in one place
+
+private module ℤero where
+
+  infixl 7 _∙_
+  infix  4 _≈_
+  Carrier : Set c
+  _≈_     : Rel Carrier ℓ
+  _∙_     : Op₂ Carrier
+
+  Carrier = ⊥
+  _≈_ ()
+  _∙_ ()
+
+  refl : Reflexive _≈_
+  refl {x = ()}
+
+  sym : Symmetric _≈_
+  sym {x = ()}
+
+  trans : Transitive _≈_
+  trans {i = ()}
+
+  isEquivalence : IsEquivalence _≈_
+  isEquivalence = record { refl = refl ; sym = sym ; trans = trans }
+
+  ∙-cong : Congruent₂ _≈_ _∙_
+  ∙-cong {x = ()}
+
+  isMagma : IsMagma _≈_ _∙_
+  isMagma = record { isEquivalence = isEquivalence ; ∙-cong = ∙-cong }
+
+  isSemigroup : IsSemigroup _≈_ _∙_
+  isSemigroup = record { isMagma = isMagma ; assoc = λ () }
+
+  isBand : IsBand _≈_ _∙_
+  isBand = record { isSemigroup = isSemigroup ; idem = λ () }
+
+open ℤero
+
+------------------------------------------------------------------------
+-- Raw bundles
+
+rawMagma : RawMagma c ℓ
+rawMagma = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ }
+
+------------------------------------------------------------------------
+-- Bundles
+
+magma : Magma c ℓ
+magma = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ ; isMagma = isMagma }
+
+semigroup : Semigroup c ℓ
+semigroup = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ ; isSemigroup = isSemigroup }
+
+band : Band c ℓ
+band = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ ; isBand = isBand }
+
--- a/src/Algebra/Construct/Initial.agda
+++ b/src/Algebra/Construct/Initial.agda
@@ -38,12 +38,14 @@ private module ℤero where
 
   infixl 7 _∙_
   infix  4 _≈_
-  Carrier : Set c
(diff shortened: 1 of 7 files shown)


JacquesCarette (approved):

JacquesCarette:
Very nice!

A few remarks:
- instead of the `private/public` thing you do, in `Construct.Terminal` you can define 
`rawMagma = record { Carrier = ⊤ ; _≈_ = \ _ _ -> ⊤}` and then everywhere else `rawX = record { rawMagma }` and you'll get the same thing.
- for `Construct.Initial`, there are other 'weird' / 'exotic' equivalence relations on `⊥` other than the one you give, and I think all of them will still "work". That's likely overkill though.

I don't think you need to change anything for this PR. You can if you think the result will be improved, but it's really unclear if that would indeed be the case.

jamesmckinna:
Many thanks for the review and the suggestions regarding modularisation. I *really* wanted a DRY solution to all the repetition of the basic record structure, but I find the concrete syntax and semantics of these things hard to get my head round. Ideally I would have liked to reify the private modules as records, and then 'paste them in-place', but I couldn't work out how to do that. Might tweak the refs with your suggestion though. 

As for defns of initial algebras on the empty carrier, I hope the present ones suffice. I wasn't really sure I could think of other ones which short-cut so much (though that's as nothing to the gain from eta-expansion of the unit type in the terminal case...)

src/Algebra/Construct/Initial.agda
```diff
@@ -0,0 +1,92 @@
...
+  hiding (rawMagma; magma; semigroup; band)
+
+------------------------------------------------------------------------
+-- gather all the functionality in one place
+
+private module ℤero where
+
+  infixl 7 _∙_
+  infix  4 _≈_
+  Carrier : Set c
```
MatthewDaggitt, line 41: Is there a reason why you have separated the declarations from the definitions? They're not mutually recursive right? And I found out recently that doing this actually changes how type-checking works (see https://github.com/agda/agda/issues/6417)! I don't think it affects things in this case, but I think we'd best avoid it unless necessary :+1:  Ditto other modules.

src/Algebra/Construct/Initial.agda
```diff
@@ -0,0 +1,92 @@
...
+  using (RawMagma)
+open import Algebra.Core using (Op₂)
+open import Algebra.Definitions using (Congruent₂)
+open import Algebra.Structures using (IsMagma; IsSemigroup; IsBand)
+open import Data.Empty.Polymorphic
+open import Relation.Binary using (Rel; Reflexive; Symmetric; Transitive; IsEquivalence)
+
+
+------------------------------------------------------------------------
+-- re-export those algebras which are also terminal
```
MatthewDaggitt, line 29: Very minor but can we capitalise the first letter of comments? Ditto other modules.

src/Algebra/Construct/Initial.agda
```diff
@@ -0,0 +1,92 @@
...
+  refl : Reflexive _≈_
+  refl {x = ()}
+
+  sym : Symmetric _≈_
+  sym {x = ()}
+
+  trans : Transitive _≈_
+  trans {i = ()}
+
+  isEquivalence : IsEquivalence _≈_
```
MatthewDaggitt, line 58: It's often quite useful to have the structures as well as the bundles exported publicly. Can we move the structures out of the private module?

src/Algebra/Construct/Initial.agda
```diff
@@ -0,0 +1,92 @@
...
+-- Raw bundles
+
+rawMagma : RawMagma c ℓ
+rawMagma = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ }
+
+------------------------------------------------------------------------
+-- Bundles
+
+magma : Magma c ℓ
+magma = record { Carrier = Carrier ; _≈_ = _≈_ ; _∙_ = _∙_ ; isMagma = isMagma }
```
MatthewDaggitt, line 85: You only need `magma = record { isMagma = isMagma }` here and Agda will infer the rest. Ditto elsewhere.

src/Algebra/Construct/Zero.agda
```diff
@@ -17,47 +19,51 @@ open import Level using (Level)
...
+  using (Magma; Semigroup; Band)
 
 ------------------------------------------------------------------------
--- Raw bundles
+-- re-export those algebras which are both initial and terminal
 
-rawMagma : RawMagma c ℓ
-rawMagma = record { Carrier = ⊤ ; _≈_ = λ _ _ → ⊤ }
+open import Algebra.Construct.Terminal public
+  hiding (rawMagma; magma; semigroup; band)
```
MatthewDaggitt, line 31: I'm a little bit confused about what the aim of this deprecation is. 

You're hiding the definitions `rawMagma` etc, but then using them qualified and then marking them deprecated. To me this looks like the worst of all worlds :sweat_smile: backwards compatibility is broken and they're still being flagged as deprecated!

What is the desired final state of this module is in the future after any deprecations have eventually been removed? Should it have any of its own declarations or should it just re-export the contents of `Terminal`? If so do we want to deprecate the module itself?

src/Algebra/Construct/Initial.agda
jamesmckinna: OK, will do. 
I think I still hanker after a 'declare the whole interface before defining the implementation' mentality when writing.

src/Algebra/Construct/Initial.agda
jamesmckinna: Oh, OK...

src/Algebra/Construct/Initial.agda
jamesmckinna: For some reason (as with Jacques' comments) I had tried this, but not managed to get it to work. I'll try again.

src/Algebra/Construct/Zero.agda
jamesmckinna: Thanks for seeking clarification. let me try to express what I had been trying to achieve... and how precisely to (be able to) do so in Agda may be where I am coming unstuck. 

`Zero` defines structures/bundles which are `Terminal`, for those algebras where the two are identified. 
But it *also* defines, implicitly, by dint of the module name, the `Terminal` `rawMagma`, `magma` and `semigroup` *as if* they were `Zero` objects: they're not. 

But the `Terminal` ones which *are* zero objects should be re-exported by this module (there will, even if for backwards compatibility only, be reasons to view those terminal objects as zero objects, and only want to import them from this module, not from `Terminal`), precisely because there are terminal pbjects which are zero ones.

The 'others' which this module used to define, are included for backwards compatibility, but deprecated, because `Zero.magma` etc. aren't correct mathematically. So I provided definitions for them, and then deprecated them. 

In the future, I don't see this module as being deprecated altogether, even though it will consist purely of re-exporting a *subset* of the definitions in `Terminal`. That's what I had hoped to achieve by the `hiding` directive, but perhaps I don't understand the semantics of it properly to achieve that effect

It makes perfect sense (to me, at least) for `Zero` to re-export those `Terminal` algebras which are in fact also `Zero` objects. 

`Initial` is then the place to define the actual initial objects for `Magma` and `Semigroup`. 

I hope this makes sense!

src/Algebra/Construct/Initial.agda
jamesmckinna: Just in the middle of trying to do this, and thinking: 
could I achieve the same effect by suitably modifying the `open ℤero` declaration?

src/Algebra/Construct/Initial.agda
jamesmckinna: For `Initial`, this seemed to work. But I don't understand what to do for `Terminal`. What I really wanted to achieve was to define a single record `one` packaging the `Carrier` and ` _≈_` in module `𝕆ne` for which all the definitions in this file would consist of writing `... = one` or `... = record { one }`, but this didn't seem to work. But the previous version of these definitions, with repetitions of `Carrier = ⊤ ; _≈_  = λ _ _ → ⊤` seemed even worse!

jamesmckinna:
Re; rationale for the PR. 

I guess that I could, perhaps should, have raised an initial issue of the form "`Algebra.Construct.Zero` defines zero objects which are not zero objects, only terminal ones. Correspondingly, true zero objects should also be re-exported as *initial* objects. Suggest refactor to put things 'in the right place': 
* define `Terminal` to introduce all the terminal algebras; 
* re-export the subset of those which are zero objects also in `Initial` and `Zero`; 
* define the true initial objects for `Magma`, `Semigroup` in `Initial`"

Perhaps that would have made the intentions a bit clearer.

src/Algebra/Construct/Zero.agda
jamesmckinna: Don't know if this resolves your concerns. There may be something I'm still not understanding...

jamesmckinna:
Modulo discussion of @MatthewDaggitt 's concerns about deprecation/backwards compatibility, I think this is ready to go. But I'd like to clarify those issues before starting work on a PR towards fixing #1907

src/Algebra/Construct/Initial.agda
JacquesCarette: I think this will indeed only work for `Initial`. What's going on is probably quite subtle, and having to do with Agda being quite happy to eagerly unify things when it's clear there's only one, but not nearly as keen to fill in `bottom` when it should be equally clear that something is impossible.

Which is actually sensible: a priori, there's no reason that witnesses of impossibility are unique. In fact, being too eager to match on `bottom` would have made it so that `Set` would not be a rig category because the coherence law between the two ways of doing `0*0=0` would not hold!!

MatthewDaggitt: No I think it's much simpler than that. It's that Agda can simply infer `_≈_` and `_∙_` (and hence `Carrier`) are from something of type `IsMagma _≈_ _∙_`

src/Algebra/Construct/Initial.agda
MatthewDaggitt: This conversation was marked as resolved, but I don't see any changes?

src/Algebra/Construct/Zero.agda
MatthewDaggitt: Thank you for the detailed explanation. I now understand :smile:

src/Algebra/Construct/Initial.agda
MatthewDaggitt: Yes you could. By duplicating it and having one with a `public using (X)` and one with a `hiding (X)`.

MatthewDaggitt:
Okay once the two outstanding conversations have been resolved, I'm happy to merge this.

src/Algebra/Construct/Initial.agda
jamesmckinna: Sorry, I evidently forgot to push the relevant commit!

src/Algebra/Construct/Initial.agda
jamesmckinna: Hmm... (I think) I see. Have left thinsg as they were: definitions inside the (private) module; structures/bundles outside. Hope that's (still) OK?!

jamesmckinna:
Final tweaks to remove `private` modifiers on the internal modules, but then I was able to use them to construct all the other stuff as originally intended.

PR #1903, opened 2023-01-10
[fixes #1894] clean commit of parity homomorphism properties

jamesmckinna:
@Matthew, can you check that this is what you were after?
I've made one change, to remove the renaming of the constructors of `Sign`, but otherwise identical. But the file-change comparison is still showing the diffs relative to the corrupted master? Fingers crossed?!


jamesmckinna:
superseded by #1904

PR #1908, opened 2023-01-12
Reduce Data.Nat.Base's import dependency graph

Taneb:
I noticed that it was indirectly importing `Algebra.Bundles` so I trimmed it down a bunch.


JacquesCarette (approved):
These are the kinds of changes I like to see.

Issue #1544, opened 2021-07-11
Get rid of inconsistent `∙-cong` name in algebra hierarchy?


MatthewDaggitt:
Every other property in the algebraic (not to mention in `Relation.Binary` etc.) structures are simply named `assoc`/`comm` etc. unless the structure contains two operators in which case they are prepended with the operator name(s).

The exception to this rule is the congruence proof which is in**cong**ruously named `∙-cong`. 
https://github.com/agda/agda-stdlib/blob/b09525ccb0b0d0993ecbd3123eae1e72046b8263/src/Algebra/Structures.agda#L36

We should have a discussion about whether we should rename this to simply `cong` in version 2.0.

As I see it:

**Pros**
 - More consistent
 - Easier to type (frequently get confused about whether it's `\.` `\cdot` or `\bub` in certain fonts)

**Cons**
 - May clash with `cong` from `Relation.Binary.PropositionalEquality` a lot.

JacquesCarette:
That con could be a really big one. So I headed over to `Algebra.Properties` to find out how bad it would be... and it would be fine. They are not used together there!

So I started writing this comment being quite against it (because of the con), and when I went digging for data... didn't find it.  This could be worse in code bases "out there", I don't know. But internally to the library itself, this doesn't look like a big deal.

MatthewDaggitt:
> So I started writing this comment being quite against it (because of the con), and when I went digging for data... didn't find it. This could be worse in code bases "out there", I don't know. But internally to the library itself, this doesn't look like a big deal.

Anecdotally, I also rarely use them together. As I'm very keen on having small modules that do one thing, either I'm working with setoid based equality from some algebraic structure or I'm working with propositional equality. However, it may have a greater impact on people who tend to throw everything together in a big file...

jamesmckinna:
`v2.0`?

jamesmckinna:
For whatever reason (but see numbers below), I'm going to press on my proposal to rename to `cong₂` (and `cong₁` where appropriate, along with @MatthewDaggitt 's left-/right- names). The definitions in `Algebra.Definitions` have no compunction about the subscripting, and relative to `PropositionalEquality`, we (appear to) have 
* 100+ uses of `∙-cong`, relative to 250 uses of `cong₂`, compared to 2500+ uses of `cong`

[...]

PR #1914, opened 2023-01-24
`SelfInverse` operations on `Algebra`s and their properties

jamesmckinna:
Reviewing my contribution of `Data.Parity` and friends, I found myself annoyed about repeated proofs of properties of the self-inverse functions `_⁻¹` and corresponding `opposite` on `Data.Sign`...
... and was surprised that being `SelfInverse` didn't seem to be in either the `Function` or `Algebra` hierarchies. 

Now, with this PR, I've chosen to put these things in the `Algebra` hierarchy, because that is where `Involutive` is defined, and self-inverse operations are involutive in the presence of reflexivity (one of the properties I needed/used/defined). 

But it seems that I could have put these definitions under `Function`, and certainly some of the `Consequences.Setoid` for self-inverse operations are really about their behaviour as `Function`s... so some advice on where these properties should properly go would be welcome. 

An argument I am open to is that, *qua* functions, they belong under `Function`; but as operations which are *automorphisms* of the corresponding `Algebra` structure, then they belong under `Algebra`. 

It's not obvious to me that we systematically observe such distinctions. One thing that does jump out, cf. issue #1544, is that `Algebra.Definitions` defines `Congruent₁` and `Congruent₂`, while `Function.Definitions` defines `Congruent`, so (t)here's another failure of systematisation: ... I prove that self-inverse operations are `Congruent`, rather than `Congruent₁`... but which should I have done?

~~TODO~~DONE: refactor to use the new definitions/proofs
- [x] `Data.Sign.Properties`
- [x] `Data.Parity.Properties`

### DIFF
--- a/src/Algebra/Consequences/Base.agda
+++ b/src/Algebra/Consequences/Base.agda
@@ -2,7 +2,7 @@
 -- The Agda standard library
 --
 -- Lemmas relating algebraic definitions (such as associativity and
--- commutativity) that don't the equality relation to be a setoid.
+-- commutativity) that don't require the equality relation to be a setoid.
 ------------------------------------------------------------------------
 
 {-# OPTIONS --without-K --safe #-}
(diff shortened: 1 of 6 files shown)


jamesmckinna:
PR #1156 (and perhaps an ounce more work) would streamline the (construction and proof of the) iso between `Sign` and `Parity`, currently done by hand in `Data.Parity.Properties`; it might also thereby streamline entirely (?) the treatment of self-inverse operations undertaken in the PR at hand.

src/Algebra/Consequences/Base.agda
```diff
@@ -14,9 +14,19 @@ open import Algebra.Core
...
+module _ {ℓ} {_•_ : Op₂ A} (_≈_ : Rel A ℓ) where
+
+  sel⇒idem : Selective _≈_ _•_ → Idempotent _≈_ _•_
+  sel⇒idem sel x with sel x x
+  ... | inj₁ x•x≈x = x•x≈x
+  ... | inj₂ x•x≈x = x•x≈x
+
+module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) where
+
+  reflexive+selfinverse⇒involutive : Reflexive _≈_ →
```
MatthewDaggitt, line 28: In names it should be referred to as `selfInverse` everywhere rather than the current mix of `selfinverse` and `self-inverse`

MatthewDaggitt (commented):
Other than that it looks good to me. I have no strong feelings about the `Function` vs `Algebra` distinction. I think it's fine here for the moment.

PR #1916, opened 2023-01-25
fixes #1437, supersedes #1836

### LINKED ISSUE #1437: Fix equality combinator names in `Data.Rational.Unnormalised.Properties.≤-Reasoning`

jamesmckinna:
Fixes https://github.com/agda/agda-stdlib/issues/1437

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -344,7 +344,7 @@ Non-backwards compatible changes
 ### Change in reduction behaviour of rationals
 
 * Currently arithmetic expressions involving rationals (both normalised and
-  unnormalised) undergo disastorous exponential normalisation. For example,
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
Thanks! Looks good apart from a CHANGELOG entry needed!

jamesmckinna:
Ah... I think as I wasn't changing any actual proofs, only their internal names, that no entry was required. Which section do such things belong in (there is no deprecation, as such, and I've certainly added no warnings to that effect)

MatthewDaggitt:
> Maybe the CHANGELOG should also include a "Breaking changes" section?

Everything in the first section is a breaking change :shrug:

PR #1922, opened 2023-02-09
two whitespace violations

JacquesCarette:
I think these must have been introduced in 1.7.2?


jamesmckinna:
I'd noticed these as well, but worked around them... but there's a separate issue; did the `make test` script used to check for whitespace violations in `CHANGELOG` prior to `v.1.7.2`? This behaviour seemed new, and strange...

PR #1923, opened 2023-02-09
a recursive view of `Fin n`

jamesmckinna:
Another view-focused PR.

This one is one (independently isolable) part of my reworking of PR #1287 which focuses on the structure of `Fin n` and in particular the definition and properties of the `inject₁` function used heavily there, and its (partial) inverse `lower₁`.

Those (re-)definitions are given in a `README` module ~~`Examples`~~, but could serve as future definitions in a refactoring of `Data.Fin.Base` and `Data.Fin.Properties`... but that's not the main purpose of this PR. Discussion welcome!


src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,257 @@
...
+open import Relation.Binary.PropositionalEquality
+
+private
+  variable
+    n : ℕ
+
+------------------------------------------------------------------------
+-- The View, considered as a unary relation on Fin (suc n)
+
+-- First, a lemma not in `Data.Fin.Properties`,
```
JacquesCarette, line 31: The lemma really ought to be moved to its proper home...

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,257 @@
...
+  view-top-toℕ : ∀ i → .{{IsTop (view i)}} → toℕ i ≡ n
+  view-top-toℕ i with top ← view i = toℕ-fromℕ n
+
+  view-inj-toℕ< : ∀ i → .{{IsInj (view i)}} → toℕ i < n
+  view-inj-toℕ< i with inj j ← view i = inject₁ℕ< j
+
+------------------------------------------------------------------------
+-- Examples
+
+private module Examples {n} where
```
JacquesCarette, line 142: I **think** that people have put such examples under the `README.` hierarchy. This is valuable stuff and shouldn't be lost, but I dislike polluting the main source file with this. [To be clear: I would rather have it remain here than it be deleted.]

JacquesCarette (changes requested):
Fascinating stuff. What really ought to happen is that, once it's in stdlib, a bit of a hunt should be done in the rest of the library to see where it should be used instead of whatever sub-optimal schemes are currently in place.

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: [Ooops. Hit the wrong button]

I agree! But as with your subsequent comment, it's really a case of "the library used to define things a certain way, and that meant that this property appears only modulo `toℕ ` --- but that's injective, so... meh".

But I did like putting it here for didactic reasons: it's the lemma which tells the user (if not,alas, the typechecker) that the view is a *disjoint* cover of `Fin (suc n)`. So I am open to suggestions as to renaming, or moving. But moving goes with a  more general weeding out of stuff from `Data.Fin` I think.

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: OK, I haven't really experimented with `README` files before... and this was really the 'discussion' part of the PR: using views, we can recreate all of what has gone before, and with more or less (mostly less!  I guess YMMV) work. 

Elevating it to `README` feels like trying to have it both ways: retaining the 'old' way, but showing how to use the new? But happy to be nudged either way... and I think some more discussion with others would be helpful.

jamesmckinna:
Regarding the general point about moving/refactoring code.. but probably the topic of a followup PR:

Moving code from `Examples` back into `Data.Fin.Base/Properties` might/would/should go with a  more general weeding out of stuff from `Data.Fin` I think. 

But I'd be... hesitant as to the deprecation-hell which might ensue, unless one went all the way, and deprecated eg `lower₁` *entirely*, in favour of working directly with `Top.view`... solvable perhaps by renaming the new version to `inject₁⁻¹`?

There's some delicacy about the import/dependency graph (a recurrent issue): `Top` needs `inject₁` to be definable, and in turn defines a partial inverse to it. So this module might really be a 'little theory of `inject₁` and its inverse' and as such, does it easily factorise into `Base` and `Properties` without breaking a lot (more) of things up?

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: I'll do two things here:
* move to `Data.Fin.Properties`
* rename to `fromℕ≢inject₁` (murder your darlings)

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: Will work on this later this weekend.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,129 @@
...
+    ... | inj j = inj j
+
+open Instances public
+
+------------------------------------------------------------------------
+-- As a corollary, we obtain two useful properties, which are
+-- witnessed by, but can also serve as proxy replacements for,
+-- the corresponding properties in `Data.Fin.Properties`
+
+module _ {n} where
```
JacquesCarette, line 122: you have a `variable` declaration for `n` up top already, so it this `module _` needed?

src/Data/Fin/Relation/Unary/Top.agda
JacquesCarette: I'm personally of the opinion that `Instances` should not be `public`, but it's @MatthewDaggitt 's call as to what the convention is in the overall library.

In a `README` file, since the intent is pedagogical instead of being "library standard", I think there should be a lot more leeway. So there, I'd go with whatever you think illustrates the idea most clearly.

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: Good catch! For the cost of an additional `{n = n}` argument prefix in the first definition.

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: Good, thanks. I think I agree with making `Instances` *not* public.

As for `opposite`, I think the name duplication does make the point forcefully. 
For `inject₁⁻¹`, even more so, as we no longer have to think what the relationship is between `lower₁` and `inject₁`.

jamesmckinna:
Only other place in `stdlib` where `lower₁` is used is in `Data.Fin.Induction.>-weakInduction`. 
And unsurprisingly, it falls to the new view-based analysis; `README...` updated accordingly.

JacquesCarette (approved):
Very nice indeed.

The main thing I don't like is the "open Instances". I'd leave that out. In fact, I think library standard would have those be in a separate file. But that's for @MatthewDaggitt to rule on - the approval is on "the rest".

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+
+-- NB `Data.Fin.Properties.fromℕ≢inject₁` establishes that the following
+-- inductively defined family on `Fin (suc n)` has constructors which
+-- target *disjoint* instances of the family; and hence the interpretations
+-- of the View constructors will also be disjoint
+
+data View : (i : Fin (suc n)) → Set where
+
+  top :                View (fromℕ n)
+  inj : (j : Fin n) → View (inject₁ j)
```
gallais, line 36: Should this come with a further `View j`? We're already traversing the whole `Fin`
so may as well hand users the ability to repeatedly use the view without redoing
the work?

gallais (approved):

MatthewDaggitt (commented):
Eek sorry, I did a review of this and then never actually hit "submit". My bad.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+------------------------------------------------------------------------
+-- The View, considered as a unary relation on Fin (suc n)
+
+-- NB `Data.Fin.Properties.fromℕ≢inject₁` establishes that the following
+-- inductively defined family on `Fin (suc n)` has constructors which
+-- target *disjoint* instances of the family; and hence the interpretations
+-- of the View constructors will also be disjoint
+
+data View : (i : Fin (suc n)) → Set where
+
```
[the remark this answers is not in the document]
jamesmckinna: OK. Many of these views were written with an avowedly 'didactic' intention, and I tend to be pretty liberal with whitespace as part of that. But elsewhere, @JacquesCarette has commented on that not being in line with the style guide.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+  top :                View (fromℕ n)
```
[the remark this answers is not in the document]
jamesmckinna: Good spot, thanks, and yes, as with comments about  whitespace above.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+  inj : (j : Fin n) → View (inject₁ j)
+
+-- The view covering function, witnessing soundness of the view
+
+view : ∀ {n} i → View {n} i
```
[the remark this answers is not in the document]
jamesmckinna: I'll look again at this, but I have found Agda to be pretty picky about `variable` scope (and I'm trying to avoid giving an explicit type to `i`). I think, but will re-check, that it complained when I wrote
```agda
view : ∀ i → View {n} i
```
and then wrote the definition based on *pattern matching* on `n`.

My belief is that `{n}` is correct, because in any *use* of `view`, `n` will already be known/inferrable. But I will go through it again and see.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+view {n = zero}  zero    = top
+view {n = suc _} zero    = inj _
+view {n = suc n} (suc i) with view {n} i
```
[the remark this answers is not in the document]
jamesmckinna: Well, you do for the base case `i = zero`, where the split on `n` determines which of `top` or `inj` branch you take. But otherwise, no, you don't. As as above, you don't ever at any call site of `view` (not even , again, IIRC, in the above recursive call).

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+view-complete top     = refl
+view-complete (inj j) = refl
+
+-- 'Computational' behaviour of the covering function
+
+view-top : ∀ n → view {n} (fromℕ n) ≡ top
+view-top zero    = refl
+view-top (suc n) rewrite view-top n = refl
+
+view-inj : ∀ j → view {n} (inject₁ j) ≡ inj j
```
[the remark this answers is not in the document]
jamesmckinna: Ah, that's a convention left over from the development for which I developed this view, where it is convenient always to distinguish `i : Fin (suc n)` from the `j : Fin n` such that `i = inject₁ j`. If `j` were declared as a `variable`, I could enforce this convention better, but at the expense of the wrong choice of implicit/explicit binding. I would happily be shown a better way as to how to persuade Agda to observe the domain-specific conventions I have hand-coded.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+--
+-- But such assertions can only ever have a unique (irrelevant) proof
+-- so we introduce instances to witness them, themselves given in
+-- terms of the functions `view-top` and `view-inj` defined above
+
+module Instances {n} where
+
+  private
+
+    lemma : n ≡ toℕ (inject₁ (fromℕ n))
```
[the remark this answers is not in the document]
Taneb: As far as I can see it's used in only one place. Perhaps it should be inlined there?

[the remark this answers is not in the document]
jamesmckinna: Hmmm. It's the composition of two lemmas, and I thought that we didn't, as a rule, support defining such 'simple' derivable compositions in the library.

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: I tried removing the spaces, and found it hard for me to read the code I had written! :-(

 I think I have a harder time reading Agda than either you or Nathan!

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: Well, I have now introduced the `variable` declarations to enforce the convention, but I'm not able to profit from it here.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+
+    inj⁺ : ∀ {j} → IsInj (view (inject₁ j))
+    inj⁺ {j} rewrite view-inj j = inj j
+
+    inject₁≡⁺ : ∀ {i} {j} {eq : inject₁ i ≡ j} → IsInj (view j)
+    inject₁≡⁺ {eq = refl} = inj⁺
+
+    inject₁≢n⁺ : ∀ {i} {n≢i : n ≢ toℕ (inject₁ i)} → IsInj (view {n} i)
+    inject₁≢n⁺ {i} {n≢i} with view i
+    ... | top with () ← n≢i lemma
```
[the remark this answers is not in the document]
jamesmckinna: Done, and that lets me inline the lemma as a `where` clause, instead of the `private` block. 

[Comments I had added earlier didn't seem to have propagated properly while connection to github was so bad... I'm tethered to my phone now for data]

Separately, though, I disagree with library convention; in particular, I think the use of irrefutable `with () ← ...`, as with other uses of irrefutable `with`, is a *very* convenient idiomatic usage. Here, it was rendered horrible by not being able to scope the `where` block over the `with`... a separate Agda issue [#5883](https://github.com/agda/agda/issues/5883) that I have raised before.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,135 @@
...
+  data IsInj : ∀ {i} → View {n} i → Set where
+
+    inj : ∀ j → IsInj (inj j)
+
+  instance
+
+    inj⁺ : ∀ {j} → IsInj (view (inject₁ j))
+    inj⁺ {j} rewrite view-inj j = inj j
+
+    inject₁≡⁺ : ∀ {i} {j} {eq : inject₁ i ≡ j} → IsInj (view j)
```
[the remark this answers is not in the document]
jamesmckinna: [Another comment that got lost in the network]

This `instance` is used only once in a lemma, in the associated `README`, where indeed the lemma requires the equality explicitly, but passes it implicitly to the instance... Agda complains that `instance`s with explicit arguments will be ignored, so I've declared them all with implicit arguments... :-(

jamesmckinna:
> Very nice indeed.
> 
@JacquesCarette looping back to your (much) earlier comment
> The main thing I don't like is the "open Instances". I'd leave that out. In fact, I think library standard would have those be in a separate file. But that's for @MatthewDaggitt to rule on - the approval is on "the rest".

Well, now that I don't `open Instances` publicly, this is OK, right? And in any case, I *do* need the `Instances` for the final two lemmas
```agda
view-top-toℕ : ∀ i → .{{IsTop (view i)}} → toℕ i ≡ n
view-top-toℕ {n = n} i with top ← view i = toℕ-fromℕ n

view-inj-toℕ< : ∀ i → .{{IsInj (view i)}} → toℕ i < n
view-inj-toℕ< i with inj j ← view i = inject₁ℕ< j
```
so I'd otherwise simply end up creating an anonymous module for these two, then `open`ing in the scope of that module, and proving the two lemmas... which is equivalent to what I have already, yes?

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: So the correct thing to do perhaps is to remove this instance, and in-line the analysis in `README....`

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: But the `{n}` in the type signature is indeed fixed by the `variable` declaration.

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: And yes, now by dint of using `contradiction`, it is possible to inline/localise it.

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: Also, my instinct is always to provide `{n}` as a nudge to the typechecker, in place of the verbosity of quantifying over `(i : Fin (suc n))`, cf. other PRs where we have differed as to explicit quantification vs. supplying enough implicits to satisfy the typechecker.

JacquesCarette:
The "local" uses of `open Instances` are fine. I'm happy with this whole PR now.

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: Hmmm... making the proposed change mangles the index arithmetic... with one outstanding proof in `Instances` failing to through... without my thinking a bit harder.

jamesmckinna: Hmmm... this may in fact be a simplification... more to folloe

jamesmckinna: Yes, I think I agree with you, so will push commits which:
* make the view recursive
* add a 'non-recursive' smart constructor/pattern synonym which projects out the `j : Fin n` component
* simplify with types + examples

And which... now makes this a view of `Fin n` (!)

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,139 @@
...
+-- NB `Data.Fin.Properties.fromℕ≢inject₁` establishes that the following
+-- inductively defined family on `Fin n` has constructors which target
+-- *disjoint* instances of the family (at index {suc n}); and hence the
+-- interpretations of the View constructors will also be disjoint
+
+data View : ∀ {n} (j : Fin n) → Set where
+  top : View (fromℕ n)
+  inj : {j : Fin n} → View j → View (inject₁ j)
+
+pattern inj₁ {n} j = inj {n = n} {j = j} _
```
gallais, line 40: Does that cause trouble if you're importing both this module and `Data.Sum.Base`?
I think pattern desugaring is not type-directed so the overloading may be problematic.

gallais: How about `fromℕ̌` and `inject₁̌ `  for the view's constructors?

jamesmckinna: oh... hadn't considered such possibility (how do I write the backtick, much less quote it in Markdown?)...

... but (despite naming conventions elsewhere), I personally loathe `fromℕ` for the function (I've always found `top` to be more mnemonic...), and the choices of `inject₁` etc. in `Data.Fin.Base` are certainly... suboptimal in my view, so this addition was an opportunity to break with those choices. 

Your 'quoted' versions drag me back to them, but maybe you're correct to do so ... (but I find it *so* annoying that they are not the same length symbols, though... despite their particular arities)

jamesmckinna: OK... I might be persuadable, subject to the qualification that I'd make ``inject₁` the derived form, and retain ``inj₁` for the view constructor? That way, the occurrence of a pattern of the form ``inject₁ j` really does correlate with a viewed expression of value `inject₁ j`...

... will sleep on it and commit tomorrow.

jamesmckinna: That only leaves the question now of what to call the module itself: `Top` or `Fromℕ`, for consistency's sake?

JacquesCarette:
I would be quite wary of having a non-ASCII top-level module name. That's a can of worms that I think is unnecessary to open!

src/Data/Fin/Relation/Unary/Top.agda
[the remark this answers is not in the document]
jamesmckinna: Final revision to the type of the `view` function, as per your suggestion

jamesmckinna:
OK, then I think it is now 'ready'!?

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,84 @@
...
+
+view : (j : Fin n) → View j
+view zero = view-zero _
+view (suc i) with view i
+... | ‵fromℕ  = ‵fromℕ
+... | ‵inject₁ j = ‵inj₁ (view (suc j))
+
+-- Interpretation of the view constructors
+
+⟦_⟧ : View {n} j → Fin n
```
MatthewDaggitt, line 56: I'm a bit worried about how often the `{n}` argument in `View` appears. It seems to be needed in every single proof below, which suggests to me that it should be an explicit argument instead?

src/Data/Fin/Relation/Unary/Top/Instances.agda
```diff
@@ -0,0 +1,89 @@
...
+
+data IsInject₁ : View {suc n} i → Set where
+  ‵inj₁ : ∀ {j} (v : View {n} j) → IsInject₁ (‵inj₁ v)
+
+instance
+
+  inject₁⁺ : IsInject₁ (view (inject₁ j))
+  inject₁⁺ {j = j} rewrite view-inject₁ j = ‵inj₁ _
+
+  inject₁≡⁺ : {eq : inject₁ j ≡ i} → IsInject₁ (view i)
```
MatthewDaggitt, line 64: I'm a little dubious about this instance in particular. How is Agda ever going to find `{eq : inject₁ j ≡ i}` automatically via instance search? Similarly how it going to find `{n≢i : n ≢ toℕ (inject₁ i)}` in `inject₁≢n⁺` below?

Are there any places in your code that actually use these instances automatically?

Indeed, I think my suspicions are backed up by it looking like instance search fails in the proof `inject₁≡⇒inject₁⁻¹≡`?

src/Data/Fin/Relation/Unary/Top/Instances.agda
```diff
@@ -0,0 +1,89 @@
...
+--
+-- This is an example of a view of (elements of) a datatype,
+-- here i : Fin (suc n), which exhibits every such i as
+-- * either, i = fromℕ n
+-- * or, i = inject₁ j for a unique j : Fin n
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Fin.Relation.Unary.Top.Instances where
```
MatthewDaggitt, line 14: Hmm, I'm not sure this is really a good match for instance search. Can you explain what the advantages of using instance search here are over simply applying the proofs manually? And why they're worth the decrease in readability and ability for the reader to comprehend proofs?

MatthewDaggitt (commented):
Sorry for taking so long to get to this. I have a few comments (/doubts) on the presence of instance search....

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: So... I think we have been swirling around this issue for a while, and I very much appreciate you continuing to hold my feet to the fire over it ;-)

What I (and I think, you, as well ;) would like is for the `variable` declarations to do all the heavy lifting for me in terms of (telescopic, but that's not how it works in Agda... :-() typing and scoping of both `j` *and* `n`, so that when I mention `j` in the type of a definition, I get the `n` index in its type 'for free', and in particular, its *actual name/identifier* 'n'... but that's not how variable declaration generalisation works (see the [recent discussion on the Agda Zulip](https://agda.zulipchat.com/#narrow/stream/238741-general/topic/Referring.20to.20variables.20used.20by.20variables.3F) about `Set _` level inference... ).

So the 'correct' solution is probably to bite the bullet and write
```
⟦_⟧ : {j : Fin n} → View j → Fin n
```
which my pig-headedness otherwise prevents me from doing, in favour of *mentioning* the (correctly implicitly-bound: when considering `j`, the typechecker will never *not* know its index) name 'n' in order to be able to refer to it in the return type...

Alternatively (but even more opaquely!), to write
```
⟦_⟧ : View j → Fin _
```
(NB I need to check whether this works).

So what you see, and perhaps rightly complain about, is my tricksy-but-hamfisted attempt to workaround/futile protest against what I see as the 'problem' with Agda's approach to `varaiable` generalisation.

src/Data/Fin/Relation/Unary/Top/Instances.agda
jamesmckinna: I think you're quite correct. You might see this module (originally marked 'Experimental' in the module, before its subsequent lifting out as a separate part of the PR) as part of an ongoing experiment/essay in trying to explore the space of (implicit) argument vs. instance inference. And not being entirely willing to follow through and admit that this experiment may, in fact, have failed...

src/Data/Fin/Relation/Unary/Top/Instances.agda
jamesmckinna: See above: not sure that I can!

The original *rationale* (sic) for the use of instances was again, to try to workaround (based on what I observed from the error messages ;-) the limitations of the typechecker arising from working `--without-K`, viz. the inability to unify type indices involving non-constructor-but-nevertheless-injective defined symbols (an important bug/feature of the otherwise ostensible use of 'green slime' in `View` types), and before I had even first heard of `injectivity` hints in `lossy-unification` (which I still don't quite understand how to use/exploit in such situations). 

The `View` type constructors here mention 'green slime' in their indices, but nevertheless (as `view` witnesses!), define a provably-disjoint-and-total covering of the telescope of indices in the type... my deployment of `instance`s was an idiomatic/idiosyncratic attempt to negotiate turning that 'provably-disjoint-and-total covering' into something the typechecker knows and can exploit.

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,84 @@
...
+⟦ ‵fromℕ ⟧     = fromℕ _
+⟦ ‵inject₁ j ⟧ = inject₁ j
+
+-- Completeness of the view
+
+view-complete : (v : View {n} j) → ⟦ v ⟧ ≡ j
```
jamesmckinna, line 62: Here, `{n}` may even be removed? Will investigate... and if so, does/would/might that assuage your concern(s) above?

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,84 @@
...
+view-complete ‵fromℕ    = refl
+view-complete (‵inj₁ _) = refl
+
+-- 'Computational' behaviour of the covering function
+
+view-fromℕ : ∀ n → view {suc n} (fromℕ n) ≡ ‵fromℕ
```
jamesmckinna, line 68: Ditto.?

jamesmckinna: I definitely, as a developer, experience a tension/tensions between: 
* wanting the typechecker to work as hard as possible for me, in terms of implicit argument inference (and instances, but here I am perhaps on shakier ground)
* wanting to (redundantly!) document certain nevertheless-implicit index/argument positions, as a hint to the reader/client of such developments

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,84 @@
...
+view-fromℕ zero                         = refl
+view-fromℕ (suc n) rewrite view-fromℕ n = refl
+
+view-inject₁ : ∀ j → view {suc n} (inject₁ j) ≡ ‵inj₁ (view {n} j)
```
jamesmckinna, line 72: Ditto.?

src/Data/Fin/Relation/Unary/Top.agda
```diff
@@ -0,0 +1,84 @@
...
+view-inject₁ zero                           = refl
+view-inject₁ (suc j) rewrite view-inject₁ j = refl
+
+-- Uniqueness of the view
+
+view-unique : (v : View {n} j) → view j ≡ v
```
jamesmckinna, line 78: Ditto.

src/Data/Fin/Relation/Unary/Top/Instances.agda
```diff
@@ -0,0 +1,89 @@
...
+-- are extensionally equivalent to the assertions
+-- `view {n} i ≡ ‵fromℕ` and `view {n} i ≡ ‵inject₁ j`
+--
+-- But such assertions can only ever have a unique (irrelevant) proof
+-- so we introduce instances to witness them, themselves given in
+-- terms of the functions `view-fromℕ` and `view-inject₁` defined in
+-- `Data.Fin.Relation.Unary.Top`
+
+data IsFromℕ : View {suc n} i → Set where
+  ‵fromℕ : IsFromℕ (‵fromℕ {n})
```
jamesmckinna, line 49: Are the indices/implicits also eliminable here, too?

src/Data/Fin/Relation/Unary/Top/Instances.agda
```diff
@@ -0,0 +1,89 @@
...
+
+instance
+
+  fromℕ⁺ : IsFromℕ (view (fromℕ n))
+  fromℕ⁺ {n = n} rewrite view-fromℕ n = ‵fromℕ
+
+data IsInject₁ : View {suc n} i → Set where
+  ‵inj₁ : ∀ {j} (v : View {n} j) → IsInject₁ (‵inj₁ v)
```
jamesmckinna, line 57: Ditto.?

src/Data/Fin/Properties.agda
```diff
@@ -443,6 +443,9 @@ toℕ-inject {i = suc i} (suc j) = cong suc (toℕ-inject j)
 -- inject₁
 ------------------------------------------------------------------------
 
+fromℕ≢inject₁ : (j : Fin n) → fromℕ n ≢ inject₁ j
```
jamesmckinna, line 446: Can/should `j` be made implicit here?

jamesmckinna: With the definition
```agda
fromℕ≢inject₁ : {j : Fin n} → fromℕ n ≢ inject₁ j
fromℕ≢inject₁ {j = suc j} eq = fromℕ≢inject₁ {j = j} (suc-injective eq)
```
being (less) preferable to read/write, but easier/more efficient to *use*?

src/Data/Fin/Relation/Unary/Top/Instances.agda
jamesmckinna: Against that, I think that the example `inject₁⁻¹` of inverting  `inject₁` in `README.DATA.Fin.Relation.Unary.Top` *does* mostly shine... only the example 
```agda
inject₁≡⇒inject₁⁻¹≡ : (eq : inject₁ {n} j ≡ i) →
                       inject₁⁻¹ i {{inject₁≡⁺ {eq = eq}}} ≡ j
```
*really* fails in that regard?

jamesmckinna:
OK, I think I have 'improved' version of the main development which address most of the above issues you identified. The question about `Instances` may be a trickier one, but I guess I would still maintain that a function taking an irrelevant argument to witness its domain predicate, one which moreover will/should only ever be called in a proof context in which that domain predicate can easily/automatically be satisfied by that context, *ought* at leats to be considered a candidate for `instance` resolution... or am I really still not getting it?

src/Data/Fin/Relation/Unary/Top/Instances.agda
jamesmckinna: See most recent commits: I think I can now get more value out of (automatic) instance inference... but maybe that shows I still have much to learn about that! Specifically,
```agda
inject₁⁻¹-inject₁ : (j : Fin n) → inject₁⁻¹ (inject₁ j) ≡ j
inject₁⁻¹-inject₁ j rewrite view-inject₁ j = refl

inject₁≡⇒inject₁⁻¹≡ : (eq : inject₁ {n} j ≡ i) → let instance _ = inject₁≡⁺ eq in inject₁⁻¹ i ≡ j
inject₁≡⇒inject₁⁻¹≡ refl = inject₁⁻¹-inject₁ _
```

jamesmckinna: Re: your original question
I think it is now the case that the proofs of the properties of the inverse function now show off the use of instance inference as (originally) intended. I suppose your question might remain as to whether, in practice, a hypothetical user of these example programs might have trouble producing witnesses when they want to appeal to the inverse function. But actually, I don't think that would be the case, precisely because the ambient proof context should contain enough witnessing information (eg in the scope of a `with` based on a call to the `view`, or else through other explicit mention of `inject₁` in the types of things) to know that it is safe to appeal to the inverse function without further ado?

src/Data/Fin/Relation/Unary/Top.agda
jamesmckinna: Hopefully my most recent commits have now resolved this discussion?

src/Data/Fin/Relation/Unary/Top/Instances.agda
jamesmckinna: But see most recent commits?

jamesmckinna:
Thanks for the tweak! When the PR started, I think it did make sense to maintain a uniform distinction between the `variable`s depending on their index, but as time has gone on the interfaces have become more homogeneous. 
I'll make the corresponding commit to `CHANGELOG` and `Data.Fin.Properties` now.

jamesmckinna:
[Sorry I hadn't seen these comments sooner! Too preoccupied with the other things I'd been doing...]

Thanks very much for the time and detailed attention you have put in to this, and for the effort in coming up with a good compromise solution; I'd been coming to a similar conclusion... so thanks again for helping get there, and past my stubbornness!

jamesmckinna:
Drastic `import` simplifications now possible: in particular, uncoupling the dependency on `Data.Fin.Properties`, which could now, therefore, import the view in ... subsequent refactorings ;-) Commit incoming...

PR #1926, opened 2023-02-13
Binomial combinatorics

jamesmckinna:
This is a stand-alone contribution, which proves the hard combinatorial part of the Binomial Theorem (cf #1287 ), viz. 
```agda
nCk+nC[k+1]≡[n+1]C[k+1] : n C k + n C (suc k) ≡ suc n C suc k
```
together with some miscellaneous additional supporting lemmas.

### DIFF
--- a/src/Data/Nat/Combinatorics.agda
+++ b/src/Data/Nat/Combinatorics.agda
@@ -36,7 +36,7 @@ open Base public
 -- Properties of _P_
 
 open Specification public
-  using (nPk≡n!/[n∸k]!; k>n⇒nPk≡0; [n∸k]!k!∣n!; k![n∸k]!∣n!)
+  using (nPk≡n!/[n∸k]!; k>n⇒nPk≡0; [n∸k]!k!∣n!)
 
 nPn≡n! : ∀ n → n P n ≡ n !
 nPn≡n! n = begin-equality
@@ -108,6 +108,9 @@ nC1≡n n@(suc n-1) = begin-equality
 
 ------------------------------------------------------------------------
 -- Arithmetic of (n C k)
+
+k![n∸k]!∣n! : ∀ {n k} → k ≤ n →  k ! * (n ∸ k) ! ∣ n !
+k![n∸k]!∣n! {n} {k} k≤n = subst (_∣ n !) (*-comm ((n ∸ k) !) (k !)) ([n∸k]!k!∣n! k≤n)
 
 nCk+nC[k+1]≡[n+1]C[k+1] : ∀ n k → n C k + n C (suc k) ≡ suc n C suc k
 nCk+nC[k+1]≡[n+1]C[k+1] n k with <-cmp k n
--- a/src/Data/Nat/Combinatorics.agda
+++ b/src/Data/Nat/Combinatorics.agda
@@ -12,11 +12,13 @@ open import Data.Nat.Base
 open import Data.Nat.DivMod
 open import Data.Nat.Divisibility
 open import Data.Nat.Properties
+open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality
   using (_≡_; refl; sym; cong; subst)
 
 import Data.Nat.Combinatorics.Base          as Base
 import Data.Nat.Combinatorics.Specification as Specification
+import Algebra.Properties.CommutativeSemigroup as CommSemigroupProperties
 
 open ≤-Reasoning
 
@@ -34,7 +36,7 @@ open Base public
 -- Properties of _P_
 
 open Specification public
-  using (nPk≡n!/[n∸k]!; k>n⇒nPk≡0; [n∸k]!k!∣n!)
+  using (nPk≡n!/[n∸k]!; k>n⇒nPk≡0; [n∸k]!k!∣n!; k![n∸k]!∣n!)
 
 nPn≡n! : ∀ n → n P n ≡ n !
 nPn≡n! n = begin-equality
@@ -45,6 +47,15 @@ nPn≡n! n = begin-equality
   n !             ∎
   where instance
     _ = (n ∸ n) !≢0
+
+nP1≡n : ∀ n → n P 1 ≡ n
+nP1≡n zero    = k>n⇒nPk≡0 (z<s {n = zero})
+nP1≡n n@(suc n-1) = begin-equality
+  n P 1        ≡⟨ nPk≡n!/[n∸k]! (s≤s (z≤n {n-1})) ⟩
+  n ! / n-1 !  ≡⟨ m*n/n≡m n (n-1 !) ⟩
+  n            ∎
+  where instance
+    _ = n-1 !≢0
 
 ------------------------------------------------------------------------
 -- Properties of _C_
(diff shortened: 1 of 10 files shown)


src/Data/Nat/Properties.agda
```diff
@@ -121,6 +121,10 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 <ᵇ-reflects-< : ∀ m n → Reflects (m < n) (m <ᵇ n)
 <ᵇ-reflects-< m n = fromEquivalence (<ᵇ⇒< m n) <⇒<ᵇ
 
+n<ᵇ1+n : ∀ n → (n <ᵇ suc n) ≡ true
```
MatthewDaggitt, line 124: I would really like to avoid ending up copying across all proofs for the ordering relations to their boolean counterparts. This doesn't seem to be used in this PR so could it be removed?

src/Data/Nat/Combinatorics/Specification.agda
```diff
@@ -98,6 +98,10 @@ k!∣nP′k n@{suc n-1} k@{suc k-1} k≤n@(s≤s k-1≤n-1) with k-1 ≟ n-1
   (subst (k ! ∣_) (nP′k≡n!/[n∸k]! k≤n) (k!∣nP′k k≤n))
   where instance _ = (n ∸ k) !≢0
 
+k![n∸k]!∣n! : ∀ {n k} → k ≤ n →  k ! * (n ∸ k) ! ∣ n !
+k![n∸k]!∣n! {n} {k} k≤n = subst (_∣ n !) (*-comm ((n ∸ k) !) (k !)) ([n∸k]!k!∣n! k≤n)
```
MatthewDaggitt, line 102: This proof isn't part of the specification so shouldn't it go in the module above?

src/Data/Nat/Properties.agda
```diff
@@ -1558,6 +1562,9 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
   (m + n) ∸ o          ≡⟨ +-∸-assoc m o≤n ⟩
   m + (n ∸ o)          ∎
 
+[m-n-1]+1≡[m-n] : ∀ {m n} (n<m : n < m) → suc (m ∸ suc n) ≡ m ∸ n
```
MatthewDaggitt, line 1565: The name of this proof should match its type, i.e. `1+[m∸[1+n]]≡m∸n`. Also it's basically free to make it more general, i.e. make `1` to some natural `o` instead. 

However, looking at it, is it really worth adding it as a standalone item? The proof isn't exactly long?

src/Data/Nat/Combinatorics.agda
```diff
@@ -85,3 +96,115 @@ nCn≡1 n = begin-equality
...
+    [n-k-1]+1≡[n-k]         : suc [n-k-1] ≡ [n-k]
+    [n-k-1]+1≡[n-k]         = [m-n-1]+1≡[m-n] k<n
+
+    n!                      = n !
+    k!                      = k !
+    [k+1]!                  = (suc k) !
+    [n-k]!                  = [n-k] !
+    [n-k-1]!                = [n-k-1] !
+
+    [n-k]*[n-k-1]!≡[n-k]!   : [n-k] * [n-k-1]! ≡ [n-k]!
```
MatthewDaggitt, line 175: I feel like the two non-trivial sub-lemmas should be lifted to the top level?

MatthewDaggitt (commented):
Otherwise looks great! Thank you so much!

jamesmckinna:
All good suggestions... and I've a long train journey tomorrow...

src/Data/Nat/Properties.agda
jamesmckinna: Apologies... this commit ended up in this part of the overall binomial collection of linked PRs, instead of PR#1928 where it is actually used. 

But you are of course correct that it use should be avoided in favour of `Reflects`... I had been cutting corners to avoid having to grapple  with that. Will remedy!

src/Data/Nat/Properties.agda
jamesmckinna: Re: name: good, will do
Re: generality: again, good, but I had been finding it a common case, so wanted to put it on its  own footing.

The proof isn't long, but I found that I kept using it in this formulation (rather than anything more general;l but crucially, IIRC, across multiple PRs associated with `Binomial`), and that the lemma it derives from has a *completely* unmemorbale name (to me at least). 

But I will clean it up, if necessary by having a `private` lemma elsewhere.

src/Data/Nat/Combinatorics.agda
jamesmckinna: OK, that's probably fair. Any subsequent use of this module might indeed find such things useful!

jamesmckinna:
Aaargh. More branch merge nonsense. Will revert. Blast!

src/Data/Nat/Combinatorics.agda
jamesmckinna: I've tried this; one argument against doing so is that too much let-bound state is shared between the lemmas and their use in the third branch of the main result. But I can see the argument for their being independenlty interesting.

src/Data/Nat/Properties.agda
jamesmckinna: Yes, I think it's avoidable here.

MatthewDaggitt:
Yup it looks like the merge has gone wrong again. It seems to be a similar problem each time. What is the series of commands you're using to merge?

jamesmckinna:
Matthew, it seems to be a combination of things:
* first fatal mistake: doing too many things at once, and doing them while too tired to concentrate; 
* having two copies of the same branch on two different local machines, each pushing to a single remote branch on `jamesmckinna/agda-stdlib`;
* the two copies having out-of-sync versions of the `agda:master` branch from which they were derived;
* `push`ing from one local to the remote, and then `pull`ing from there to the other
* generating an explicit `merge-commit` cycle, 9at which point everything is in fact screwed up)
* compounding the error by pushing to the remote
* now, the third, 'unbroken' local branch can't be pushed to the remote because of the push/pull merge conflict this has generated

There are then second-order effect arising from (still!) not getting right how to revert such changes far enough back to undo the mess, and then re-committing the 'good' commits---I seem always to be off-by-one in one direction of time or the other. 

I'm snowed under with other thinsg right now, so it qwill be next week now before I dig my way back out of this one. One copy I have is still (relatively) clean, so I may simply push commits from there to a fresh branch. Sorry that this will screw up the reviewing, but it seems more reliable than trying to fix up a broken tree on the remote branch. But I will also investigate that as well. 

I should by now have learnt that being prompted to record a merge commit message is sign of doing something wrong; if I I `fetch` before attempting to `merge` (ie don't `pull`, only `fetch`), can I somehow preview what will happen and forestall this kind of screw-up again?

jamesmckinna:
Thanks @MatthewDaggitt but I think you should have squashed those commits... they each seem to have contributed to my tally on the contributors page, which is the wrong arithmetic, i think.

PR #1927, opened 2023-02-13
binomial lemmas for *-commuting elements in a `Semiring`

jamesmckinna:
... on the way to a full proof of the Binomial Theorem PR #1287 / #1928 

Plus a proof of the `sumₜ-init` lemma of the old PR, updated, renamed, and moved to `Algebra.Properties.Monoid.Sum`. 

Plus the proofs of the lemmas concerning the `n ×_` operation in `Algebra.Properties.Semiring.Mult`. 

Plus proof of some useful cosmetic congruence lemmas in `Algebra.Properties.Monoid.Mult` and `Algebra.Properties.Semiring.Exp`

### DIFF
--- a/src/Algebra/Properties/Monoid/Sum.agda
+++ b/src/Algebra/Properties/Monoid/Sum.agda
@@ -31,6 +31,7 @@ open Monoid M
 open import Data.Vec.Functional.Relation.Binary.Equality.Setoid setoid
 open import Algebra.Properties.Monoid.Mult M
 open import Algebra.Definitions _≈_
+open import Relation.Binary.Reasoning.Setoid setoid
 
 ------------------------------------------------------------------------
 -- Definition
@@ -70,3 +71,21 @@ sum-replicate-idem idem n = trans (sum-replicate n) (×-idem idem n)
 sum-replicate-zero : ∀ n → sum {n} (replicate 0#) ≈ 0#
 sum-replicate-zero zero    = refl
 sum-replicate-zero (suc n) = sum-replicate-idem (+-identityˡ 0#) (suc n)
+
+------------------------------------------------------------------------
+-- When summing over a `Vector`, we can pull out last element
+
+sum-init : ∀ {n} (t : Vector Carrier (suc n)) → sum t ≈ sum (init t) + last t
+sum-init {zero} t  = begin
+  t₀ + 0# ≈⟨ +-identityʳ t₀ ⟩
+  t₀      ≈˘⟨ +-identityˡ t₀ ⟩
+  0# + t₀ ∎ where t₀ = t zero
+sum-init {suc n} t = begin
+  t₀ + ∑t             ≈⟨ +-congˡ (sum-init (tail t)) ⟩
+  t₀ + (∑t′ + tₗ)     ≈˘⟨ +-assoc _ _ _ ⟩
+  (t₀ + ∑t′) + tₗ     ∎
+  where
+    t₀ = head t
+    tₗ = last t
+    ∑t = sum (tail t)
+    ∑t′ = sum (tail (init t))
--- a/src/Algebra/Properties/Monoid/Sum.agda
+++ b/src/Algebra/Properties/Monoid/Sum.agda
@@ -72,16 +72,15 @@ sum-replicate-zero : ∀ n → sum {n} (replicate 0#) ≈ 0#
 sum-replicate-zero zero    = refl
 sum-replicate-zero (suc n) = sum-replicate-idem (+-identityˡ 0#) (suc n)
 
-------------------------------------------------------------------------
 -- When summing over a `Vector`, we can pull out last element
 
-sum-init : ∀ {n} (t : Vector Carrier (suc n)) → sum t ≈ sum (init t) + last t
-sum-init {zero} t  = begin
+sum-init-last : ∀ {n} (t : Vector Carrier (suc n)) → sum t ≈ sum (init t) + last t
+sum-init-last {zero} t  = begin
   t₀ + 0# ≈⟨ +-identityʳ t₀ ⟩
   t₀      ≈˘⟨ +-identityˡ t₀ ⟩
   0# + t₀ ∎ where t₀ = t zero
-sum-init {suc n} t = begin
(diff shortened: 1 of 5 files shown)


src/Algebra/Properties/Semiring/Exp.agda
```diff
@@ -34,10 +34,37 @@ open import Algebra.Definitions.RawSemiring rawSemiring public
...
 ^-homo-* = Mult.×-homo-+
 
 -- (xᵐ)ⁿ≈xᵐ*ⁿ
 ^-assocʳ : ∀ x m n → (x ^ m) ^ n ≈ x ^ (m ℕ.* n)
 ^-assocʳ x m n rewrite ℕ.*-comm m n = Mult.×-assocˡ x n m
+
+------------------------------------------------------------------------
+-- A lemma using commutativity, needed for the Binomial Theorem
+
+module _ {x} {y} (x*y≈y*x : x * y ≈ y * x) where
```
MatthewDaggitt, line 51: I think it's best just to stick these parameters in the type rather than having a module. We usually use modules to generalise over assumptions about operators that apply to multiple proofs. It's slightly weird to have one which refers to properties of specific elements.

Speaking of which is there a reason why you're not simply assuming `_*_` is `Commutative`?

src/Algebra/Properties/Semiring/Exp.agda
```diff
@@ -34,10 +34,37 @@ open import Algebra.Definitions.RawSemiring rawSemiring public
...
+    y * (1# * y ^ n)           ≈⟨ *-congˡ (*-identityˡ (y ^ n)) ⟩
+    y * (y ^ n)                ≡⟨⟩
+    y ^ (suc n)                ≈˘⟨ *-identityˡ (y ^ suc n) ⟩
+    1# * y ^ (suc n)           ≡⟨⟩
+    x ^ ℕ.zero * y ^ (suc n)  ∎
+  y*x^m*y^n≈x^m*y^[n+1] (suc m) n = begin
+    y * (x ^ suc m * y ^ n)    ≡⟨⟩
+    y * ((x * x ^ m) * y ^ n)  ≈⟨ *-congˡ (*-assoc x (x ^ m) (y ^ n)) ⟩
+    y * (x * (x ^ m * y ^ n))  ≈˘⟨ *-assoc y x (x ^ m * y ^ n) ⟩
+    y * x * (x ^ m * y ^ n)    ≈˘⟨ *-congʳ {- here -} x*y≈y*x ⟩
```
MatthewDaggitt, line 65: Remove the comment?

src/Algebra/Properties/Monoid/Sum.agda
```diff
@@ -70,3 +71,21 @@ sum-replicate-idem idem n = trans (sum-replicate n) (×-idem idem n)
 sum-replicate-zero : ∀ n → sum {n} (replicate 0#) ≈ 0#
 sum-replicate-zero zero    = refl
 sum-replicate-zero (suc n) = sum-replicate-idem (+-identityˡ 0#) (suc n)
+
+------------------------------------------------------------------------
+-- When summing over a `Vector`, we can pull out last element
+
+sum-init : ∀ {n} (t : Vector Carrier (suc n)) → sum t ≈ sum (init t) + last t
```
MatthewDaggitt, line 78: A better name might be `sum-init-last`?

src/Algebra/Properties/Monoid/Sum.agda
```diff
@@ -70,3 +71,21 @@ sum-replicate-idem idem n = trans (sum-replicate n) (×-idem idem n)
 sum-replicate-zero : ∀ n → sum {n} (replicate 0#) ≈ 0#
 sum-replicate-zero zero    = refl
 sum-replicate-zero (suc n) = sum-replicate-idem (+-identityˡ 0#) (suc n)
+
+------------------------------------------------------------------------
+-- When summing over a `Vector`, we can pull out last element
```
MatthewDaggitt, line 76: I think using a section heading for this overkill. I would just attach the comment to the proof. i.e. delete the empty line and remove the long line of `------`s

src/Algebra/Properties/Semiring/Exp.agda
jamesmckinna: Regarding explicit `module` parametrisation over putting them in the type; these were 'ambient' assumptions across a number of parts of the development. But for sure that's not a deal-breaker. 

Regarding the assumption itself, for sure it means that the lemmas (and the eventual theorem) are an easy corollary for `CommutativeSemiring`. But I was reasonably persuaded by the [Wikipedia article](https://en.wikipedia.org/wiki/Binomial_theorem#In_abstract_algebra) of the greater utility in having it stated for individual `x`, `y` (in a possibly non-commutative (semi)ring, eg of matrices) which happen to commute, than insistign on commutativity up-front.

src/Algebra/Properties/Semiring/Exp.agda
jamesmckinna: OK

src/Algebra/Properties/Monoid/Sum.agda
jamesmckinna: Yes, I had been taking the name directly from #1287 . Suggestion is better!

src/Algebra/Properties/Monoid/Sum.agda
jamesmckinna: Again, the original had this as one of a number of lemmas about `sum`, each with its own section heading, and I had just carried that one over to its new home. Will clean up.

jamesmckinna:
Aaargh. I've screwed up my branch naming, and a spurious push to one of them has broken things. Grrr.

src/Algebra/Properties/Monoid/Sum.agda
jamesmckinna: Done, except that I left the blank line in, to lay off the lemma from the preceding ones. Looked better!

src/Algebra/Properties/Semiring/Exp.agda
MatthewDaggitt: Fair enough :+1:

PR #1932, opened 2023-02-22
agda-stdlib-utils: drop GHC < 8.6, modernise .cabal file, update CI

andreasabel:
With Agda-2.6.4 dropping GHC < 8.6, we can follow suit here. This encourages us to also bump the `cabal-version`, which allows us to modernise the `.cabal` file.

Update the Haskell-CI to the latest versions.


jamesmckinna:
Does this change permit continued support for 2.6.2.2 until the stabilisation of the v2.0 milestone?

andreasabel:
Yes.  It is only about the GHC version used to set up the Agda standard library, i.e., to generate `Everything.agda` and the like.

PR #1938, opened 2023-03-20
Add monotonicity laws for `Nat._^_`

HarrisonGrodin:


src/Data/Nat/Properties.agda
```diff
@@ -1046,6 +1046,24 @@ m^n≡1⇒n≡0∨m≡1 m (suc n) eq = inj₂ (m*n≡1⇒m≡1 m (m ^ n) eq)
 m^n≢0 : ∀ m n .{{_ : NonZero m}} → NonZero (m ^ n)
 m^n≢0 m n = ≢-nonZero (≢-nonZero⁻¹ m ∘′ m^n≡0⇒m≡0 m n)
 
+^-monoˡ-≤ : ∀ n → (_^ n) Preserves _≤_ ⟶ _≤_
+^-monoˡ-≤ zero m≤o = s≤s z≤n
+^-monoˡ-≤ (suc n) {m} {o} m≤o = *-mono-≤ m≤o (^-monoˡ-≤ n m≤o)
```
jamesmckinna, line 1051: I don't think that L.1051 needs to mention the implicits `{m}` and `{o}`?

HarrisonGrodin:
Thanks for the comments! I just made these changes.

src/Data/Nat/Properties.agda
```diff
@@ -1046,6 +1046,24 @@ m^n≡1⇒n≡0∨m≡1 m (suc n) eq = inj₂ (m*n≡1⇒m≡1 m (m ^ n) eq)
...
+^-monoˡ-< : ∀ n .{{_ : NonZero n}} → (_^ n) Preserves _<_ ⟶ _<_
+^-monoˡ-< (suc zero) m<o = *-monoˡ-< 1 m<o
+^-monoˡ-< (suc n@(suc _)) m<o = *-mono-< m<o (^-monoˡ-< n m<o)
+
+^-monoʳ-< : ∀ m → 1 < m → (m ^_) Preserves _<_ ⟶ _<_
+^-monoʳ-< m 1<m {.zero} {suc o} z<s =
+  *-mono-≤ 1<m (n≢0⇒n>0 (≢-nonZero⁻¹ (m ^ o) {{(m^n≢0 m o {{>-nonZero (<-trans z<s 1<m)}})}}))
+^-monoʳ-< m 1<m {suc _} (s<s n≤o) =
+  *-monoʳ-< m {{ >-nonZero (<-trans z<s 1<m) }} (^-monoʳ-< m 1<m n≤o)
+
```
jamesmckinna, line 1066: So... lines L.1062--1065 got me thinking, as follows:
* the repetition of the instance argument `{{>-nonZero (<-trans z<s 1<m)}}` has to be a code smell, given that there are other (less convoluted!) ways to witness `0 < m`; 
* the logic of the argument is otherwise clear  (but makes subtle use of the hypothesis that `1 < m` in the form `2 ≤ m` !); 
* the immediately succeeding lines L.1067--1069 witness a proof which can, and should be generalised, and then used in your proof.

So I came up with this instead, for your consideration:
```agda
m^n>0 : ∀ m .{{_ : NonZero m}} n → m ^ n > 0
m^n>0 m n = >-nonZero⁻¹ (m ^ n) {{m^n≢0 m n}}

^-monoʳ-< : ∀ m → 1 < m → (m ^_) Preserves _<_ ⟶ _<_
^-monoʳ-< m@(suc _) 1<m {zero}  {suc o} z<s       = *-mono-≤ 1<m (m^n>0 m o)
^-monoʳ-< m@(suc _) 1<m {suc n} {suc o} (s<s n<o) = *-monoʳ-< m (^-monoʳ-< m 1<m n<o)

2^n>0 : ∀ (n : ℕ) → 2 ^ n > 0
2^n>0 = m^n>0 2
```

The proof now features that 
* `m > 0` is finessed by pattern matching on `m` as `suc _`, without further ado;
* the proof that `m ^ n > 0` which you had conjured up, is now itself made systematic as a lemma, as an example of (inversion of) instance inference; 
* the basic logical structure is retained; 
* the succeeding lemma is now a trivial consequence of the new lemma.

Moral: never be afraid to touch/refactor other parts of a module on the way to making a clean new addition (within reason! maintainers may have other views on the matter... but in this case, the additional changes are tight, local, and in the spirit of the argument you gave).

jamesmckinna:
Otherwise, looks a good first PR to me! 
(@MatthewDaggitt I'll leave it to you to approve the test/ci workflow for this, unless you're happy for me to do so).

src/Data/Nat/Properties.agda
HarrisonGrodin: Your points make a lot of sense. Thanks for the idea - I just made this change!

HarrisonGrodin:
Upon further inspection, `Data.Nat.Logarithm.Core` is the only module using the new-in-2.0 `2^n>0`, and it uses  `>-nonZero (2^n>0 n)`, which could instead simply be `m^n≢0 2 n`. Thus, to avoid making `2` a special case, it might be nicer to replace `2^n>0` with this more general `m^n>0` here?

MatthewDaggitt:
Thanks for the great PR! Nice contributions and a great refactoring. As soon as this has passed CI I'll merge it in.

MatthewDaggitt:
> Otherwise, looks a good first PR to me!
(@MatthewDaggitt I'll leave it to you to approve the test/ci workflow for this, unless you're happy for me to do so).

@jamesmckinna yup of course more than happy for you to do this in future!

jamesmckinna:
@HarrisonGrodin sorry not to have managed to get back to this PR before @MatthewDaggitt merged it in. It all looks/looked great! 
And you even managed to avoid having to introduce a deprecation warning for `2^n>0` ;-)

HarrisonGrodin:
@jamesmckinna No worries - thank you for the helpful review comments!

PR #1940, opened 2023-04-11
Added Field structure

guilhermehas:
Field is a commutative ring with inverse operation.


guilhermehas:
> Thanks for the PR! However, fields are surprisingly complicated in constructive settings such as Agda, there's multiple inequivalent ways to define them, and they're not "strictly algebraic". We already have one definition of fields in `Algebra.Apartness`.

I haven't found it because it was not in `Algebra.Bundles` when I looked for it. And I don't know what Apartness is to look at this directory.

PR #1941, opened 2023-04-13
Added VectorSpace

guilhermehas:
Vector space is defined in this way:

```agda
record isVectorSpace (VS : Pred Carrierᴹ ℓ) : Set (m ⊔ r ⊔ ℓ) where
  field
    +-closed     : x ∈ VS → y ∈ VS → x +ᴹ y ∈ VS
    0m-closed    : 0ᴹ ∈ VS
    ·-closedLeft : ∀ r → x ∈ VS → r *ₗ x ∈ VS
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3028,7 +3028,7 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   <-weakInduction-startingFrom : P i →  (∀ j → P (inject₁ j) → P (suc j)) → ∀ {j} → j ≥ i → P j
   ```
 
-* Added new data structure `VectorSpace` in `Algebra.VectorSpace`
+* Added new data structure `VectorSpace` in `Algebra.VectorSpace.Base`
 ```agda
 record isVectorSpace (VS : Pred Carrierᴹ ℓ) : Set (m ⊔ r ⊔ ℓ)
   +-closed     : (x∈VS : x ∈ VS) (y∈VS : y ∈ VS) → x +ᴹ y ∈ VS
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3027,3 +3027,13 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   ```agda
   <-weakInduction-startingFrom : P i →  (∀ j → P (inject₁ j) → P (suc j)) → ∀ {j} → j ≥ i → P j
   ```
+
+* Added new data structure `VectorSpace` in `Algebra.VectorSpace`
+```agda
+record isVectorSpace (VS : Pred Carrierᴹ ℓ) : Set (m ⊔ r ⊔ ℓ)
+  +-closed     : (x∈VS : x ∈ VS) (y∈VS : y ∈ VS) → x +ᴹ y ∈ VS
+  0m-closed    : 0ᴹ ∈ VS
+  ·-closedLeft : ∀ r (x∈VS : x ∈ VS) → r *ₗ x ∈ VS
+
+VectorSpace = Σ[ VS ∈ Pred Carrierᴹ ℓ ] isVectorSpace VS
+```
--- a/src/Algebra/VectorSpace.agda
+++ b/src/Algebra/VectorSpace.agda
@@ -1,4 +1,0 @@
-{-# OPTIONS --cubical-compatible --safe #-}
-module Algebra.VectorSpace where
-
-open import Algebra.VectorSpace.Base public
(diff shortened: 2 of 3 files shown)


jamesmckinna:
Agree with @Taneb : to do vector spaces 'properly', we would need to agree a sensible constructive definition of 'Field'... but this is a minefield, repeatedly explored in previous issues and PRs (I can't track them all down just now).

Much of the theory of vector spaces is easily generalised to modules M over a ring R; but if you really need rescaling *by reciprocals*, suggest you pass from a module M to its isomorphic copy Hom(R, M), (or even Hom(List R, M) for the multiplicative structure) and cache the scaling factors modulo suitable preprocessing of the lists to normalise expressions over R, in particular to cancel reciprocals. 

And those can be handled by passing to a formal construction akin to the rationals derived from integers by taking non-zero denominators in formal 'fractions'... (whence the constructive concerns about 'non-zero' in general). 

But TL:DR, please don't add vector spaces as such, at least not without exploring the design space (and the associated linear/commutative algebra), and previous attempts at this, a bit more deeply first. 

In any case... the Module.* sub hierarchy of Algebra also could use some attention/development effort instead ;-)

Taneb:
@guilhermehas Note that the lean definition is called `submodule`, not `vectorspace` or anything similar. This library doesn't currently have any notion of substructures to my knowledge. This would be one way to define substructures, but not the only one.

Taneb:
> > I'm not convinced this adds anything much over `LeftModule`. If you want to work on a subset this way, it's possible to use `Data.Product` to make the type and subset into a type.
> 
> I didn't understand. Can you give me a simple example?

@guilhermehas

If we have this definition:

```agda
record ℤ³ : Set where
  constructor 𝕫³
  field
    x y z : ℤ
```

We can define a version that enforces the `z` parameter to be 0 by the following:

```agda
ℤ² : Set
ℤ² = Σ[ a ∈ ℤ³ ] z a ≡ 0ℤ
```

We can then make a module definition for that type too

guilhermehas:
I was trying to prove that if `x ≈ᴹ y`, so `x ∈ VS → y ∈ VS`. However, it is impossible to prove using `VS : Pred Carrierᴹ ℓ`.
So now, I really understand why we can't define it in this way. Does anyone have a better suggestion for the type of VS?

Taneb:
> I was trying to prove that if `x ≈ᴹ y`, so `x ∈ VS → y ∈ VS`. However, it is impossible to prove using `VS : Pred Carrierᴹ ℓ`. So now, I really understand why we can't define it in this way. Does anyone have a better suggestion for the type of VS?

I'm confused about what you're asking here. Could you rephrase or elaborate?

guilhermehas:
> > I was trying to prove that if `x ≈ᴹ y`, so `x ∈ VS → y ∈ VS`. However, it is impossible to prove using `VS : Pred Carrierᴹ ℓ`. So now, I really understand why we can't define it in this way. Does anyone have a better suggestion for the type of VS?
> 
> I'm confused about what you're asking here. Could you rephrase or elaborate?

I made a code in Agda that can explain this problem. It should be possible to prove this theorem, but it is not:
```agda
open import Level
open import Relation.Unary
open import Algebra.Bundles
open import Algebra.Module


module _ {r} {ℓr} {m} {ℓm}
  {ring : Ring r ℓr}
  (leftModule : LeftModule ring m ℓm)
  where

  open LeftModule leftModule

  Cᴹ : (ℓ : Level) → Set (m ⊔ suc ℓ)
  Cᴹ ℓ = Pred Carrierᴹ ℓ

  variable
    ℓ : Level
    x y : Carrierᴹ
    VS : Cᴹ ℓ

  theo : x ≈ᴹ y → x ∈ VS → y ∈ VS
  theo x≈y x∈VS = {!!}
```

I have to define Cᴹ in a way that this theorem could be possible to solve.

Taneb:
Ah, I see the problem. Your `theo` is not provable. The `_≈ᴹ_` could be an equivalence relation other than equality, like for example equivalence modulo 6 for the integers. Then if your predicate `VS` is `_< 8`, you'll have `3 ≈ᴹ 9` and `3 ∈ VS`, but not `9 ∈ VS`. You need to have your `theo` as an additional law, not something you can assume for every equivalence relation and predicate.

guilhermehas:
I created this record, but I would like to know if there is something similar in the library:
```agda
open import Level
open import Relation.Binary

record CRel a b c d : Set (suc (a ⊔ b ⊔ c ⊔ d)) where
  field
    Carrier₁ : Set a
    Carrier₂ : Set b
    _≈₁_ : Rel Carrier₁ c
    _∈_ : REL Carrier₁ Carrier₂ d
    cong : ∀ {x y : Carrier₁} {xs : Carrier₂} → x ≈₁ y → x ∈ xs → y ∈ xs
```

PR #1945, opened 2023-04-17
Functional vector module

guilhermehas:
Added structures and bundles about functional vectors and modules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3027,3 +3027,54 @@ This is a full list of proofs that have changed form to use irrelevant instance 
   ```agda
   <-weakInduction-startingFrom : P i →  (∀ j → P (inject₁ j) → P (suc j)) → ∀ {j} → j ≥ i → P j
   ```
+
+* In `Data.Vec.Functional.Algebra.Base`
+```agda
+  _≈ᴹ_ : Rel (VC n) ℓ
+  _+ᴹ_ : Op₂ $ VC n
+  0ᴹ : VC n
+  -ᴹ_ : Op₁ $ VC n
+  _*ₗ_ : Opₗ Carrier (VC n)
+```
+
+* Added algebraic properties in `Data.Vec.Functional.Algebra.Properties`
+```agda
+  +ᴹ-cong : Congruent₂ (_+ᴹ_ {n})
+  +ᴹ-assoc : Associative (_+ᴹ_ {n})
+  +ᴹ-comm : Commutative (_+ᴹ_ {n})
+  +ᴹ-identityˡ : LeftIdentity (0ᴹ {n}) _+ᴹ_
+  +ᴹ-identityʳ : RightIdentity (0ᴹ {n}) _+ᴹ_
+  +ᴹ-identity : Identity (0ᴹ {n}) _+ᴹ_
+  -ᴹ‿cong : Congruent₁ (-ᴹ_ {n})
+  -ᴹ‿inverseˡ : AD'.LeftInverse (0ᴹ {n}) -ᴹ_ _+ᴹ_
+  -ᴹ‿inverseʳ : AD'.RightInverse (0ᴹ {n}) -ᴹ_ _+ᴹ_
+  -ᴹ‿inverse : AD'.Inverse (0ᴹ {n}) -ᴹ_ _+ᴹ_
+  *ₗ-cong : Congruent SR._≈_ (_*ₗ_ {n})
+  *ₗ-zeroˡ : LD.LeftZero SR.0# (0ᴹ {n}) _*ₗ_
+  *ₗ-distribʳ : _*ₗ_ LD.DistributesOverʳ SR._+_ ⟶ (_+ᴹ_ {n})
+  *ₗ-identityˡ : LD.LeftIdentity SR.1# (_*ₗ_ {n})
+  *ₗ-assoc : LD.Associative SR._*_ (_*ₗ_ {n})
+  *ₗ-zeroʳ : LD.RightZero (0ᴹ {n}) _*ₗ_
+  *ₗ-distribˡ : _*ₗ_ LD.DistributesOverˡ (_+ᴹ_ {n})
+```
+
+* Added structures in `Data.Vec.Functional.Algebra.Properties`
+```agda
+  isMagma : IsMagma (_+ᴹ_ {n})
+  isSemigroup : IsSemigroup (_+ᴹ_ {n})
+  isMonoid : IsMonoid (_+ᴹ_ {n}) 0ᴹ
+  isCommutativeMonoid : IsCommutativeMonoid (_+ᴹ_ {n}) 0ᴹ
+  isPreleftSemimodule : IsPreleftSemimodule semiring (_≈ᴹ_ {n}) _+ᴹ_ 0ᴹ _*ₗ_
+  isLeftSemimodule : IsLeftSemimodule semiring (_≈ᴹ_ {n}) _+ᴹ_ 0ᴹ _*ₗ_
+  isLeftModule : IsLeftModule ring (_≈ᴹ_ {n}) _+ᴹ_ 0ᴹ -ᴹ_ _*ₗ_
+```
+
+* Added bundles in `Data.Vec.Functional.Algebra.Properties`
+```agda
+  magma : ℕ → Magma _ _
+  semiGroup : ℕ → Semigroup _ _
+  monoid : ℕ → Monoid _ _
+  commutativeMonoid : ℕ → CommutativeMonoid _ _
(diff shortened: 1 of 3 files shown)


src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,173 @@
...
+open module LD {n} = LeftDefs Carrier (_≈ᴹ_ {n}) using (Congruent)
+
+------------------------------------------------------------------------
+-- Algebraic properties of _+ᴹ_ -ᴹ_ _*ₗ_
+
++ᴹ-cong : Congruent₂ (_+ᴹ_ {n})
++ᴹ-cong x≈y u≈v _ = +-cong (x≈y _) (u≈v _)
+
++ᴹ-assoc : Associative (_+ᴹ_ {n})
++ᴹ-assoc _ _ _ _ = +-assoc _ _ _
```
Taneb, line 44: I think it would be clearer to explicitly write out the arguments here and in other definitions, and not use wildcards and inference.

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,173 @@
...
+
+------------------------------------------------------------------------
+-- Bundles
+
+magma : ℕ → Magma _ _
+magma n = record
+  { isMagma = isMagma {n}
+  }
+
+semiGroup : ℕ → Semigroup _ _
```
Taneb, line 150: Should be `semigroup`, not `semiGroup`, for consistency with the rest of library

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,46 @@
...
+module Data.Vec.Functional.Algebra.Base
+  {c ℓ} (ring : Ring c ℓ) where
+
+private variable
+  n : ℕ
+
+open Ring ring
+open VecSetoid setoid
+
+VC = Vector Carrier
```
Taneb, line 31: I don't think we should have this definition. It adds another layer of indirection without adding any clarity.

Taneb (commented):
I like this addition. There's a lot of algebraic structures that you could define but don't, especially:

* `IsCommutativeMagma`
* `IsCommutativeSemigroup`
* `IsGroup`
* `IsAbelianGroup`
* `IsPrerightSemimodule`
* `IsRightSemimodule`
* `IsBisemimodule`
* `IsRightModule`
* `IsBimodule`
* `IsModule`

And with element-wise multiplication:
*  `IsNearSemiring`
*  `IsSemiringWithoutOne`
*  `IsCommutativeSemiringWithoutOne`
*  `IsSemiringWithoutAnnihilatingZero`
*  `IsSemiring`
*  `IsCommutativeSemiring`
*  `IsRingWithoutOne`
*  `IsRing`
*  `IsCommutativeRing`

Plus the corresponding bundles.

src/Data/Vec/Functional/Algebra/Properties.agda
guilhermehas: solved

src/Data/Vec/Functional/Algebra/Properties.agda
guilhermehas: solved

src/Data/Vec/Functional/Algebra/Base.agda
guilhermehas: removed it

guilhermehas:
I added all of them.

CHANGELOG.md
```diff
@@ -3027,3 +3027,113 @@ This is a full list of proofs that have changed form to use irrelevant instance
   ```agda
   <-weakInduction-startingFrom : P i →  (∀ j → P (inject₁ j) → P (suc j)) → ∀ {j} → j ≥ i → P j
   ```
+
```
gallais, line 3030: In these cases we typically only mention we have added entirely new modules rather than listing their content.
Cf. https://github.com/agda/agda-stdlib/blob/master/CHANGELOG.md#new-modules

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,53 @@
...
+open import Algebra.Core
+open import Algebra.Bundles
+open import Algebra.Module
+open import Relation.Binary
+import Data.Vec.Functional.Relation.Binary.Equality.Setoid as VecSetoid
+import Algebra.Definitions as AD
+import Algebra.Structures as AS
+
+module Data.Vec.Functional.Algebra.Base
+  {c ℓ} (cring : CommutativeRing c ℓ) where
```
Taneb, line 23: I'm not happy about this whole module being parameterized on `CommutativeRing`, as many of the definitions work with smaller algebraic structures, but I'm not sure what to do about it. Maybe other people have some ideas?

guilhermehas:
I have changed most of my code and it is not done yet. It is still with a lot of comments.
However, I would like some feedback if I am going in the right direction.

I tried to make the code work with more cases, but I don't know if it is helpful in that way.

guilhermehas:
I can remove the code that is commented on if they are not necessary.

I also would like some suggestions if something is not in the right way.

jamesmckinna:
Sorry not to have had time for a thorough review of this draft PR yet!
It'll be at least the week after next before I can look at it properly.

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,73 @@
...
+open import Algebra.Module
+open import Relation.Binary using (Rel)
+open import Data.Vec.Functional.Relation.Binary.Pointwise using (Pointwise)
+
+private variable
+  a ℓ : Level
+  A : Set ℓ
+  n : ℕ
+
+module EqualityVecFunc (_≈_ : Rel A ℓ) where
```
JacquesCarette, line 27: Is there a reason everything in here is in sub-modules? That is not a common `stdlib` pattern.

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,73 @@
...
+  _≈ᴹ_ : Rel (Vector A n) ℓ
+  _≈ᴹ_ = Pointwise _≈_
+
+module VecAddition (_∙_ : Op₂ A) where
+  _∙ᴹ_ : Op₂ $ Vector A n
```
JacquesCarette, line 32: Why not call this `_+ᴹ_`  ?

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,73 @@
...
+  _∙ᴹ_ = zipWith _∙_
+
+module VecMagma (rawMagma : RawMagma a ℓ) where
```
JacquesCarette, line 35: I expected the 'result' here to define an instance the *structure* `Magma` instead of all the exploded pieces of one.
 (Same for `Monoid`, `Group`, etc)

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,638 @@
...
+  ∙ᴹ-cong ∙-cong x≈y u≈v i = ∙-cong (x≈y i) (u≈v i)
+
+  ∙ᴹ-assoc : ≈.Associative _∙_ → ≈ᴹ.Associative (_∙ᴹ_ {n})
+  ∙ᴹ-assoc ∙-assoc xs ys zs i = ∙-assoc (xs i) (ys i) (zs i)
+
+  ∙ᴹ-comm : ≈.Commutative _∙_ → ≈ᴹ.Commutative (_∙ᴹ_ {n})
+  ∙ᴹ-comm ∙-comm xs ys i = ∙-comm (xs i) (ys i)
+
+
+  -- isMagma : IsMagma _≈_ _∙_ → IsMagma _≈ᴹ_ (_∙ᴹ_ {n})
```
JacquesCarette, line 54: This commented out code is good code! It fits the rest of the `stdlib` style.

guilhermehas:
> This seems like good additions to `stdlib`, if it were made to fit the style more. Can anyone jump in and do so? (i.e. I'm asking permission)

Yes, anyone can do it.

jamesmckinna:
In place of a now hopelessly overdue review, I'm instead going to suggest that the revised version #2817 of this PR, and the now-extensive discussion there, means that we should close this now as obsolete/duplicate, especially in view of `Algebra.Construct.Pointwise` #2381 / #2555 generalising beyond the `Fin n` case considered here.

That said, we still need corresponding work on a hypothetical `Algebra.Module.Construct.Pointwise` on the model of pointwise algebra (where the scalar multiplication would be given by the action of pointwise scalars on pointwise module elements...).

Issue #1935, opened 2023-03-14
No deprecation for `Data.List.Categorical` in v2.0?


andreasabel:
It seems `Data.List.Categorical` simply vanishes when going from v1.7 to v2.0.
Will there be any deprecation module for backwards compatible?  (I thought this was the general policy of the std-lib.)

jamesmckinna:
See the v2.0 CHANGELOG... we put out an RFC here on the Zulip, and the mailing list, for issue #1636 / the associated PR #1735, which  was sufficiently gross a global change that I pushed for it to be strongly breaking, without deprecation warnings...

andreasabel:
@jamesmckinna
Ah, seems I have to pay the price now for unsubscribing from general notifications for this repo.

I just had the impression that the standard-library always had strong emphasis on backwards-compatibility and explicit deprecation, and this tradition seems discontinued now.
With Agda lacking CPP pragmas like `#if MIN_VERSION_standard_library(2,0,0)` (hello Haskell!), I see no way to support both standard-library 1.7 and 2.0 in my Agda development, other than having branches in my repo.  I would prefer not having to maintain branches.
It would be nice if one could use both the latest release of the library and the current master.  Say once 2.0 is out, I could address all the deprecations, drop 1.7, and then live with 2.0 and the new master.

In principle, if module A is renamed to module B, would it be hard to keep A as deprecated module that simply reexports all of B?
```agda
module Data.List.Categorical where
{-# WARNING_ON_IMPORT "Deprecated: Import Data.List.Effectful instead" #-}
open import Data.List.Effectful public
```

jamesmckinna:
(Sorry not to have replied to this sooner, as the originator of PR #1735, but not the original issue #1636) I can see the argument for Data.*.Effectful being shadowed by deprecated Data.*.Categorical modules, but I think the price I/we did not wish to pay was that of maintaining an entire deprecated hierarchy below Category.* as well (which would have delayed the whole point of the change by at least one version bump).

PR #1946, opened 2023-04-19
Restore deleted `*.Categorical.*` as deprecated modules

jamesmckinna:
Fixes (partially) @andreasabel 's issue #1935, by restoring client modules of the renamed `Effect.*` hierarchy deleted under #1735 . 

NB: suggest these modules to be deleted as soon as possible after the release of v2.0!


jamesmckinna:
BUT: I'm getting errors from `make test`
```
src/Codata/Sized/Colist/Categorical.agda is not declared unsafe but not using the --safe option either.
```
which doesn't seem to be the case  when checking the module in interactive mode under Emacs?

Has the semantics of `GenerateEverything` changed?

andreasabel (approved):
Excellent, thank you!

jamesmckinna:
@MatthewDaggitt : `status: ready`?

Issue #1937, opened 2023-03-19
Redundant uses of `with`


jamesmckinna:
I have recently undertaken a study of all the uses of `with` in `stdlib` (there are 650-700 of them, for the sake of a ballpark figure), and discovered some surprising things as a result. Two things (at least) leap out: 
* the use of `with` on *boolean* subexpressions, in place of `if_then_else` (there are over 100 of these alone, sometimes but not always in the context of more complex pattern-matching analysis); 
* the use of `with` to introduce a name for a recursive call on a function, not subsequently analysed by pattern matching. 

The purpose of this issue is to draw attention to such uses, and ask whether they should be replaced/refactored away in favour of 'simpler' definitions. The first seems now to be 'idiomatic' Agda: I'm certainly not arguing *per se* for uses of `if_then_else`, but it is striking that we (seem to) have adopted a style self-consciously *not* like 'ordinary' functional programming, even when that is available. 

A specimen example of the second case is given the following, in `Data.List.Properties`:
```agda
splitAt-defn : ∀ n → splitAt {A = A} n ≗ < take n , drop n >
splitAt-defn zero    xs       = refl
splitAt-defn (suc n) []       = refl
splitAt-defn (suc n) (x ∷ xs) with splitAt n xs | splitAt-defn n xs
... | (ys , zs) | ih = cong (Prod.map (x ∷_) id) ih
```
whose final defining clause could, much more straightforwardly, be replaced by
```agda
splitAt-defn (suc n) (x ∷ xs) = cong (Prod.map (x ∷_) id) (splitAt-defn n xs)
```

JacquesCarette:
I think there was an era of doing Agda development where some people were over-enthusiastic users of `with`, before its fragility was fully known/realized. I think there might have been a bit of a "shiny new thing" effect -- `if_then_else` is so *passé*, especially for those who know about *boolean blindness*.

My opinion:
- if `if_then_else` would do the job just fine, use it. 
- don't abuse `with` to do simple pattern-matching. **Do** use `with` when it causes rather non-trivial new information to get exposed. (Personally, I avoid it in my own code even for that, but that's personal style.)

For the specimen you give, I prefer the `cong` version. And I'm pretty sure there's a function in `Data.Product` to avoid the `id`, while you're at it!

gallais:
`with` gets stuck until you abstract over & match on the expression it mentions.
A function with an `if_then_else_`-based RHS will unfold just fine, potentially making
the goal noisier.

[...]

PR #1947, opened 2023-04-20
removed redundant use of `with`

jamesmckinna:
cf. issue #1937

UPDATED: these are several of the more egregious 'gratuitous' uses of `with` in stdlib. I haven't (and won't, on this iteration) attempted to be systematic in replacing use of `with` on a Boolean argument in favour of `if_then_else_`, ~~nor those where they are used essentially as `let`-bindings,~~ but now I have where (at the bare minimum) an irrefutable pattern with a variable would be more (cognitively) efficient. 

The one 'noteworthy' instance here concerns `scanr`, where a reordering of the (nested) with patterns permitted elimination of the dead branch completely in favour of an irrefutable `_∷_` pattern. 

No `CHANGELOG`: only the proofs of existing lemmas have been modified.

### DIFF
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -130,15 +130,15 @@ module _ (f : A → Maybe B) where
 
   mapMaybe-concatMap : mapMaybe f ≗ concatMap (fromMaybe ∘ f)
   mapMaybe-concatMap [] = refl
-  mapMaybe-concatMap (x ∷ xs) with f x
-  ... | just y  = cong (y ∷_) (mapMaybe-concatMap xs)
-  ... | nothing = mapMaybe-concatMap xs
+  mapMaybe-concatMap (x ∷ xs) with ih ← mapMaybe-concatMap xs | f x
+  ... | just y  = cong (y ∷_) ih
+  ... | nothing = ih
 
   length-mapMaybe : ∀ xs → length (mapMaybe f xs) ≤ length xs
   length-mapMaybe []       = z≤n
-  length-mapMaybe (x ∷ xs) with f x
-  ... | just y  = s≤s (length-mapMaybe xs)
-  ... | nothing = m≤n⇒m≤1+n (length-mapMaybe xs)
+  length-mapMaybe (x ∷ xs) with ih ← length-mapMaybe xs | f x
+  ... | just y  = s≤s ih
+  ... | nothing = m≤n⇒m≤1+n ih
 
 ------------------------------------------------------------------------
 -- _++_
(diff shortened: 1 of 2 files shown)


src/Data/List/Properties.agda
```diff
@@ -912,9 +910,9 @@ module _ {P : Pred A p} (P? : Decidable P) where
   length-partition : ∀ xs → (let (ys , zs) = partition P? xs) →
                      length ys ≤ length xs × length zs ≤ length xs
   length-partition []       = z≤n , z≤n
-  length-partition (x ∷ xs) with does (P? x) | length-partition xs
-  ...  | true  | rec = Prod.map s≤s m≤n⇒m≤1+n rec
-  ...  | false | rec = Prod.map m≤n⇒m≤1+n s≤s rec
+  length-partition (x ∷ xs) with does (P? x)
+  ...  | true  = Prod.map s≤s m≤n⇒m≤1+n (length-partition xs)
+  ...  | false = Prod.map m≤n⇒m≤1+n s≤s (length-partition xs)
```
MatthewDaggitt, line 915: So this one is kind of interesting, when I originally wrote this code I put the second `with` in in order to reduce code duplication. Can we somehow maintain it so we only have to write `length-partition xs` once without the `with`?

MatthewDaggitt (commented):
Otherwise no complaints from me!

src/Data/List/Properties.agda
jamesmckinna: I think that is/would be possible with an irrefutable with pattern. In the first instance I had been trying to avoid those, but they are at least more notationally economical than the previous versions. What about the computational behaviour at typechecking time though?

jamesmckinna: But note: this kind of idiomatic use if with is very common... and I wasn't aiming for a systematic trawl through all of them, at this stage at least!

jamesmckinna:
Oh! I renamed the branch (source of my original confusion re #1948/#1919) and GH closed the PR... blast!

src/Data/List/Properties.agda
jamesmckinna: So: I see (at least) two possible solutions:
* a `where clause`, but I think they don't (necessarily) scope correctly over all the branches of a `with`, so I don't think works in this case (but see below)
* an irrefutable ` with rec ← <recursive call>` idiom, which has to *precede* the (Boolean test)
* rewrite to the Boolean with `if_then_else_`, and then use a `where`/`let` clause for the recursive (or not at all, if not actually needed; some of these could simply be inlined)

I think the request for discussion is that I don;t know which of these alternatives is actually preferable (I raised the `if_then_else_` issue on issue #1937 itself). Which do you prefer?

MatthewDaggitt: Yes, I'm interested in how we resolve this problem in general. Current solution seems okay.

JacquesCarette: We should document our "use cases" and post them on the main Agda issue tracker. I think this particular situation doesn't have a nice solution in now-agda and may well be a simple extension.

Issue #1919, opened 2023-02-01
`Data.Nat.Base._≤″_` and `Algebra.Definitions.RawMagma._∣_`


jamesmckinna:
So... 
... part of issue #1726 was a concern about the proliferation of (allegedly) useful variants on the usual ordering relation on `Nat`s. Yet another one of these, `_≤″_`, is nothing more or less than the (abstract, algebraic) divisibility ordering induced by the (raw) `_+_, _≡_`-monoid (magma) structure, while (usual) divisibility is... the same for the (raw) `_*_, _≡_`-m... structure... 

... three distinct representations of the 'same' algebraic concept, differing even up to the direction of the underlying equality, `record`s vs. `∃` for packaging up the data, different field names, ... etc. 

Now that `Data.X.Base` modules #1810 (for better or worse) expose their underlying `Raw` structure/bundle(s), ... Is it time to reconsider whether the above (non-)duplication is A Good Thing?

UPDATED: Aaargh! I've just discovered that this would be a reversion (and then some) of #192 / #196 ... groan. 
But I think the original reasons for that issue/PR no longer hold, not because the semantics of pattern-matching (and `let` in particular wrt records) has got so much more sophisticated since 2017... and the change proposed here, and embarked upon in  #2013 is 'only' to replace a custom record with a Σ-type, so I don't *think* it should cause a problem?

... still: :facepalm: those who forget the (lessons/mistakes of the) past are condemned to repeat them...

[deleted: attached to the wrong issue]

jamesmckinna:
Oh... and the underlying adjunction structure across the ordering to define monus, quotient, etc.

JacquesCarette:
So I'm quite pro having alternate representation of things, but I'm not fond of stuffing them all in `.Base`. I would prefer seeing a `Data.Nat.Extra` (I'm not wed to that name) where this is defined. I strongly dislike having `.Base` modules filled with seldom-used things.

So, for me:
- alternate representations are good and belong in the library
- they don't belong in the core

jamesmckinna:
Jacques, I certainly agree as regards alternative representations, where the alternatives are actually mathematically distinct, so indeed it's ok by me for there to be 5 distinct representations if _<_ on Nat (Matthew I think found 5 obe ir two too many)

[...]

PR #1948, opened 2023-04-22
Redefines `Data.Nat.Base._≤″_`

jamesmckinna:
As per issue #1919 (cf. #1726 )

One puzzle is that I (ended up thinking that I had to) introduce a definition equivalent to the former field name projection `proof`, only to then deprecate it immediately. Couldn't quite see how else to trigger a warning about the redefinition of the relation (otherwise, with the single proof modification in `Data.Nat.Properties`, the redefinition actually affects nothing in the library, thanks to the additional pattern synonym!)

Also fixes a spurious whitespace violation from PR #1926 (not sure how this happened).


CHANGELOG.md
```diff
@@ -427,6 +427,32 @@ Non-backwards compatible changes
...
+    field
+      {k}   : ℕ
+      proof : m + k ≡ n
+  ```
+  which introduced a spurious additional definition, when this is in fact, modulo
+  field names and implicit/explicit qualifiers, equivalent to the definition of left-
+  divisibility, `_∣ˡ_` for the `RawMagma` structure of `_+_`. Since the addition of
+  raw bundles to `Data.X.Base`, this definition can now be made directly. Knock-on
+  consequences include the need to retain the old constructor name, now introduced
+  as a pattern synonym, and deprecation of (a function equiavlent to) the former
```
JacquesCarette, line 445: typo: equivalent

src/Data/Nat/Base.agda
```diff
@@ -314,51 +362,17 @@ compare (suc m) (suc n) with compare m n
...
-  { Carrier = _
-  ; _≈_ = _≡_
-  ; _+_ = _+_
-  ; _*_ = _*_
-  ; 0# = 0
-  ; 1# = 1
-  }
+-- Version 2.0
 
+proof : ∀ {m n} (le : m ≤″ n) → let less-than-or-equal {k} _ = le in m + k ≡ n
```
JacquesCarette, line 374: So I really don't like exposing this. It wasn't there before, as it really was `_≤″_.proof` which doesn't "pollute" the name space nearly as much. I would seriously consider not defining it at all?

src/Data/Nat/Properties.agda
```diff
@@ -2098,8 +2098,8 @@ _>″?_ = flip _<″?_
 
 ≤‴⇒≤″ : ∀{m n} → m ≤‴ n → m ≤″ n
 ≤‴⇒≤″ {m = m} ≤‴-refl     = less-than-or-equal {k = 0} (+-identityʳ m)
-≤‴⇒≤″ {m = m} (≤‴-step x) = less-than-or-equal (trans (+-suc m _) (_≤″_.proof ind)) where
-  ind = ≤‴⇒≤″ x
+≤‴⇒≤″ {m = m} (≤‴-step x) with less-than-or-equal proof ← ≤‴⇒≤″ x
+  = less-than-or-equal (trans (+-suc m _) proof)
```
JacquesCarette, line 2102: can't you use proj_2 instead of `with` here?

JacquesCarette (approved):
It's a tough call, whether to do everything "by hand" using the least possible assumptions, or to reveal a bit more of the algebraic structure that "make things work".

In a sense, I think what I'd want to do, in some hypothetical Agda 5.0, would be to use meta-programming to have the library definition by as general as possible, but the user-facing one be as concrete as possible.

In the meantime, I think this experiment is worthwhile.

src/Data/Nat/Properties.agda
jamesmckinna: Ah!... nice question: if you do use `proj₂`, then you are exposing the (new) representation as a Σ-type; whereas, by pattern matching on the synonym, you preserve the original abstraction. This also arises (I realised later) in the very definition of the function `proof` that I introduce for the sake of having a name to deprecate...

src/Data/Nat/Base.agda
jamesmckinna: I agree, but I wasn't sure how to handle deprecation of a field name... even though there was only a single use of it elsewhere in the library, and which is now removed in any case in favour of the abstraction-preserving appeal to the `less-than-or-equal` pattern synonym.

jamesmckinna:
Aaaargh! Committed to wrong branch! Will revert last two commits...

jamesmckinna:
So... I think this is ready to go once we have resolved the issue of whether it is worth the trouble of introducing a functional replacement for the former field projection, solely for the sake of the deprecation warning. It *is* a wart, to be sure.

CHANGELOG.md
gallais, line 445:
```suggestion
  as a pattern synonym, and deprecation of (a function equivalent to) the former
```

gallais (approved):

jamesmckinna:
The only outstanding issue for me is whether to include the function-instead-of-record-field `proof`, in order to then deprecate it. Any opinions on this, @MatthewDaggitt ? Otherwise, i think it should be `status:ready`...

jamesmckinna:
Can we do this for v2.0?

jamesmckinna:
Aaaargh! That's a horrible (and entirely unwittingly introduced) bug, sorry!
Is it this particular PR, or an Agda/`emacs` main issue?

PR #1950, opened 2023-04-24
New property merge-is-superlist

andreasabel:
Proofs that both lists involved in a merge are sublists of the result of the merge.

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
@@ -13,7 +13,6 @@ module Data.List.Relation.Binary.Sublist.Setoid.Properties
 
 open import Data.List.Base hiding (_∷ʳ_)
 open import Data.List.Relation.Unary.Any using (Any)
-open import Data.List.Relation.Binary.Sublist.Heterogeneous using () renaming (minimum to []⊆_)
 import Data.Maybe.Relation.Unary.All as Maybe
 open import Data.Nat.Base using (_≤_; _≥_)
 import Data.Nat.Properties as ℕₚ
@@ -21,6 +20,7 @@ open import Data.Product using (∃; _,_; proj₂)
 open import Function.Base
 open import Function.Bundles using (_⇔_; _⤖_)
 open import Level
+open import Relation.Binary using () renaming (Decidable to Decidable₂)
 open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong)
 open import Relation.Binary.Structures using (IsDecTotalOrder)
 open import Relation.Unary using (Pred; Decidable; Irrelevant)
@@ -208,25 +208,23 @@ module _ {as bs : List A} where
 ------------------------------------------------------------------------
 -- merge
 
-module Merge {ℓ′} (_≤_ : Rel A ℓ′) (dto : IsDecTotalOrder _≈_ _≤_)  where
+module _ {ℓ′} {_≤_ : Rel A ℓ′} (_≤?_ : Decidable₂ _≤_) where
 
-  open IsDecTotalOrder dto using (_≤?_)
-
-  merge-is-superlistˡ : ∀ xs ys → xs ⊆ merge _≤?_ xs ys
-  merge-is-superlistˡ []       ys = []⊆ ys
-  merge-is-superlistˡ (x ∷ xs) [] = ⊆-refl
-  merge-is-superlistˡ (x ∷ xs) (y ∷ ys)
-   with x ≤? y  | merge-is-superlistˡ xs (y ∷ ys)
-                      | merge-is-superlistˡ (x ∷ xs) ys
+  ⊆-mergeˡ : ∀ xs ys → xs ⊆ merge _≤?_ xs ys
+  ⊆-mergeˡ []       ys = minimum ys
+  ⊆-mergeˡ (x ∷ xs) [] = ⊆-refl
+  ⊆-mergeˡ (x ∷ xs) (y ∷ ys)
+   with x ≤? y  | ⊆-mergeˡ xs (y ∷ ys)
+                      | ⊆-mergeˡ (x ∷ xs) ys
   ... | yes x≤y | rec | _   = ≈-refl ∷ rec
   ... | no  x≰y | _   | rec = y ∷ʳ rec
 
-  merge-is-superlistʳ : ∀ xs ys → ys ⊆ merge _≤?_ xs ys
-  merge-is-superlistʳ [] ys =  ⊆-refl
-  merge-is-superlistʳ (x ∷ xs) [] = []⊆ (merge _≤?_ (x ∷ xs) [])
-  merge-is-superlistʳ (x ∷ xs) (y ∷ ys)
-   with x ≤? y  | merge-is-superlistʳ xs (y ∷ ys)
-                      | merge-is-superlistʳ (x ∷ xs) ys
+  ⊆-mergeʳ : ∀ xs ys → ys ⊆ merge _≤?_ xs ys
+  ⊆-mergeʳ [] ys =  ⊆-refl
+  ⊆-mergeʳ (x ∷ xs) [] = minimum (merge _≤?_ (x ∷ xs) [])
+  ⊆-mergeʳ (x ∷ xs) (y ∷ ys)
+   with x ≤? y  | ⊆-mergeʳ xs (y ∷ ys)
+                      | ⊆-mergeʳ (x ∷ xs) ys
   ... | yes x≤y | rec | _   = x ∷ʳ rec
   ... | no  x≰y | _   | rec = ≈-refl ∷ rec
 
--- a/src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
@@ -208,25 +208,25 @@ module _ {as bs : List A} where
 ------------------------------------------------------------------------
 -- merge
 
-module _ {ℓ′} (_≤_ : Rel A ℓ′) (dto : IsDecTotalOrder _≈_ _≤_)  where
+module Merge {ℓ′} (_≤_ : Rel A ℓ′) (dto : IsDecTotalOrder _≈_ _≤_)  where
 
   open IsDecTotalOrder dto using (_≤?_)
 
-  merge-is-sublistˡ : ∀ xs ys → xs ⊆ merge _≤?_ xs ys
-  merge-is-sublistˡ []       ys = []⊆ ys
-  merge-is-sublistˡ (x ∷ xs) [] = ⊆-refl
-  merge-is-sublistˡ (x ∷ xs) (y ∷ ys)
-   with x ≤? y  | merge-is-sublistˡ xs (y ∷ ys)
-                      | merge-is-sublistˡ (x ∷ xs) ys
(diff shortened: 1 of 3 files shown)


src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
```diff
@@ -36,6 +37,15 @@ open SetoidEquality S using (_≋_; ≋-refl)
...
 
+------------------------------------------------------------------------
+-- Least element
+
+module _ where
+
+  []⊆ : ∀ xs → [] ⊆ xs
+  []⊆ [] = []
+  []⊆ (x ∷ xs) = x ∷ʳ []⊆ xs
+
```
jamesmckinna, line 48: Is this operation not [already exported/exportable](https://github.com/andreasabel/agda-stdlib/blob/merge-is-sublist/src/Data/List/Relation/Binary/Sublist/Heterogeneous.agda#L37-L39) from
`Data.List.Relation.Binary.Sublist.Heterogeneous` as `minimum`, so a simple renaming-on-import would suffice?

Taneb:
I think these functions are incorrectly named; `merge` is the superlist, not the sublist. But on the other hand I don't think we ever use the term `superlist`, so I'm not sure what to call them

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
andreasabel: Thanks!  I hadn't found the `minimum` function.

src/Data/List/Relation/Unary/Sorted/TotalOrder/Properties.agda
```diff
@@ -109,22 +117,43 @@ module _ (DO : DecTotalOrder a ℓ₁ ℓ₂) where
...
+  merge-is-sublistʳ {xs = []}     {ys} rxs rys =  ⊆-refl
+  merge-is-sublistʳ {xs = x ∷ xs} {[]} rxs rys = []⊆ (merge _≤?_ (x ∷ xs) [])
+  merge-is-sublistʳ {xs = x ∷ xs} {y ∷ ys} rxs rys
+   with x ≤? y   | merge-is-sublistʳ (Linked.tail rxs) rys
+                       | merge-is-sublistʳ rxs (Linked.tail rys)
+  ... | yes x≤y  | rec | _   = x ∷ʳ rec
+  ... | no  x≰y  | _   | rec = ≡.refl ∷ rec
+
+
+
```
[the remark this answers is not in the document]
andreasabel: Indeed, after some plumbing!

It is been a while I wrote the code, hadn't originally taken it to a PR...

The comments greatly help!

andreasabel:
@Taneb wrote:
> I think these functions are incorrectly named; `merge` is the superlist, not the sublist. But on the other hand I don't think we ever use the term `superlist`, so I'm not sure what to call them

Thanks for the comment, now I am using `superlist`.

jamesmckinna:
Nice! Thanks for the response to reviews... I fully sympathise with the difficulty of finding names, here where they are imported from deep below the hierarchy (and `Heterogeneous` doesn't immediately suggest that its contents will be relevant to the homogeneous case)

jamesmckinna:
Regarding `superlist`: do we have a universal property of `merge` as a supremum of the two sublists?

Taneb:
> Regarding `superlist`: do we have a universal property of `merge` as a supremum of the two sublists?

@jamesmckinna I couldn't find it. Could you make an issue for it, because I think it's important to have and also doesn't need to be done in this PR (I think it might need stronger preconditions on the lists)

andreasabel:
> > Regarding `superlist`: do we have a universal property of `merge` as a supremum of the two sublists?
> 
> @jamesmckinna I couldn't find it. Could you make an issue for it, because I think it's important to have and also doesn't need to be done in this PR (I think it might need stronger preconditions on the lists)

Yeah,`merge` is only a supremum for totally ordered lists.
On arbitrary lists, `merge` is only a supremum up to permutation and I am not sure this is worth proving, as it is typically used on ordered lists.

jamesmckinna:
Re: 'up to permutation' Is that not what is being captured by `_⊆_`?
Re: the case of totally (linearly) order lists, is that not what is being addressed in the (now re-exported) version in `Data.List.Relation.Unary.Sorted.TotalOrder.Properties`? In this respect, pace @Taneb 's suggestion to raise it as a separate issue, how hard would it be to prove here, agaisnt eh background assumptions of the lists being `Sorted`?

andreasabel:
@jamesmckinna wrote:
> Re: 'up to permutation' Is that _not_ what is being captured by `_⊆_`? 

No, sublists are order-preserving, so the bigger list just inflates the smaller list by new elements, but the relative order of the old elements stays the same.  (Up to permutation would be the "subset" concept for lists.)

> Re: the case of totally (linearly) order lists, is that not what is being addressed in the (now re-exported) version in `Data.List.Relation.Unary.Sorted.TotalOrder.Properties`? In this respect, _pace_ @Taneb 's suggestion to raise it as a separate issue, how hard would it be to prove here, against the background assumptions of the lists being `Sorted` 

I think if the PR is meaningful as such, it should be accepted as such (avoiding _feature creep_ that kills PRs).
More can always be done later.
I don't know how hard it is to prove that the merge is the lub, maybe it is just an inductive proof over the two lists being merged...

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
```diff
@@ -203,6 +205,31 @@ module _ {as bs : List A} where
...
   reverse⁻ = HeteroProperties.reverse⁻
 
+------------------------------------------------------------------------
+-- merge
+
+module Merge {ℓ′} (_≤_ : Rel A ℓ′) (dto : IsDecTotalOrder _≈_ _≤_)  where
+
+  open IsDecTotalOrder dto using (_≤?_)
+
+  merge-is-superlistˡ : ∀ xs ys → xs ⊆ merge _≤?_ xs ys
```
MatthewDaggitt, line 215: I think the standard naming convention would be dictate either `⊆-mergeˡ` or `xs⊆merge[xs,ys]`?

src/Data/List/Relation/Unary/Sorted/TotalOrder/Properties.agda
```diff
@@ -109,22 +113,30 @@ module _ (DO : DecTotalOrder a ℓ₁ ℓ₂) where
...
   ... | no  x≰y | _   | rec = merge-con (just (≰⇒≥ x≰y)) (head′ rys) ∷′ rec
 
+  -- Reexport merge-is-superlistˡʳ
+
+  open
+    SublistProperties.Merge
+      (Preorder.Eq.setoid (DecTotalOrder.preorder DO))
+       _≤_
+      (DecTotalOrder.isDecTotalOrder DO)
+    public using (merge-is-superlistˡ; merge-is-superlistʳ)
```
MatthewDaggitt, line 138: Hmm although this style is tempting, we try and avoid writing it where possible as it makes it very confusing for users to work out what the final type actually is. I think it's better to write out the final type, and then apply the more general lemma directly (e.g. see `Sorted⇒AllPairs` above)

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
```diff
@@ -203,6 +205,31 @@ module _ {as bs : List A} where
   reverse⁻ : reverse as ⊆ reverse bs → as ⊆ bs
   reverse⁻ = HeteroProperties.reverse⁻
 
+------------------------------------------------------------------------
+-- merge
+
+module Merge {ℓ′} (_≤_ : Rel A ℓ′) (dto : IsDecTotalOrder _≈_ _≤_)  where
```
MatthewDaggitt, line 211: Making the change described below would then allow us to get rid of this named module. These names should be exported publicly from this module anyway, not hidden away in a submodule.

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
```diff
@@ -6,13 +6,14 @@
...
 
-open import Relation.Binary using (Setoid; _⇒_; _Preserves_⟶_)
+open import Relation.Binary using (Rel; Setoid; _⇒_; _Preserves_⟶_)
 
 module Data.List.Relation.Binary.Sublist.Setoid.Properties
   {c ℓ} (S : Setoid c ℓ) where
 
 open import Data.List.Base hiding (_∷ʳ_)
 open import Data.List.Relation.Unary.Any using (Any)
+open import Data.List.Relation.Binary.Sublist.Heterogeneous using () renaming (minimum to []⊆_)
```
MatthewDaggitt, line 16: Hmm best to not rename things unnecessarily? If you wanted to you could add an extra lemma named `⊆-minimum` here?

MatthewDaggitt:
(I've deleted the `Squash me` label, as we should always be squashing. For some reason Github's default changed recently, but I've enabled the `Linear history` option in the settings that should prevent us from not squashing them).

andreasabel:
> (I've deleted the `Squash me` label, as we should always be squashing. For some reason Github's default changed recently, but I've enabled the `Linear history` option in the settings that should prevent us from not squashing them).

I thought "linear history" still leaves two options: "rebase" and "squash", the first for well-engineered individual commits and the second for "commit followed by fixups".  (So I put up a flag that "rebase" would be a bad option here...)

In my experience, Github's default is whatever you used the last time...

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
andreasabel: OK, I'll change, even though it makes the proof less readable, because `[]⊆ xs` is unambiguous for `[] ⊆ xs`, while `minimum xs` would (in my intuitition) mean to extract the minimal element from list `xs` (see Haskell's `Foldable`s).

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
andreasabel: Thanks, I'll fix this.

src/Data/List/Relation/Unary/Sorted/TotalOrder/Properties.agda
andreasabel: Ok, fixed this.

gallais:
Looks good to me. We'll need a CHANGELOG entry though.

MatthewDaggitt:
@andreasabel any chance you could add a CHANGELOG entry as @gallais requests and then I'll merge this in?

andreasabel:
@MatthewDaggitt : Sorry, I missed the last comments.
Now there is a CHANGELOG entry.

Btw, why is CHANGELOG.md excluded from fix-whitespace?

andreasabel:
Fine to merge now, @MatthewDaggitt ?

MatthewDaggitt:
> Btw, why is CHANGELOG.md excluded from fix-whitespace?

Unsure... I guess because it's not code?

andreasabel:
> > Btw, why is CHANGELOG.md excluded from fix-whitespace?
> 
> Unsure... I guess because it's not code?

Technically, markdown has this obscure (and imo ill-designed) feature that two trailing spaces force a line break (a `<br/>` tag). https://www.oasys.net/fragments/markdown-line-breaks/
I think one shouldn't make use of this feature, and then enforcing the usual whitespace discipline would be fine.

PR #1954, opened 2023-04-27
The free `Magma` on a `Set`, resp. `Setoid`

jamesmckinna:
This is a specimen instance of a general hierarchy, intended to be placed under `Algebra.Bundles.Free.X`, and to be reexported publicly by `Algebra.Bundles.Free`. 
Currently lacking: the 'free' homomorphism guaranteed by initiality.
For discussion: is this the right way to proceed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1325,6 +1325,12 @@ New modules
   Algebra.Construct.Flip.Op
   ```
 
+* Algebraic structures obtained as the free thing (for their signature):
+  ```
+  Algebra.Bundles.Free
+  Algebra.Bundles.Free.Magma
+  ```
+
 * Morphisms between module-like algebraic structures:
   ```
   Algebra.Module.Morphism.Construct.Composition
--- a/src/Algebra/Bundles/Free/Magma.agda
+++ b/src/Algebra/Bundles/Free/Magma.agda
@@ -1,0 +1,129 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definitions of 'pre-free' and 'free' magma
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Algebra.Bundles.Free.Magma where
+
+open import Algebra.Core
+import Algebra.Definitions as Defs using (Congruent₂)
+import Algebra.Structures as Strs using (IsMagma)
+open import Algebra.Bundles using (Magma)
+open import Algebra.Bundles.Raw using (RawMagma)
+open import Relation.Binary.Core using (Rel)
+open import Level using (Level; _⊔_)
+open import Relation.Nullary.Negation.Core using (¬_)
+open import Relation.Binary using (Setoid; IsEquivalence; Reflexive; Symmetric; Transitive)
+open import Relation.Binary.PropositionalEquality
+  using (_≡_; cong₂) renaming (refl to ≡refl; isEquivalence to ≡isEquivalence)
+
+------------------------------------------------------------------------
+-- 'pre'-free algebra
+
+infixl 7 _∙_
+
+data PreFreeMagma {c} (A : Set c) : Set c where
+
+  var : A → PreFreeMagma A
+  _∙_ : Op₂ (PreFreeMagma A)
+
+------------------------------------------------------------------------
+-- parametrised 'equational' theory over the 'pre'-free algebra
+
+module PreFreeTheory {c ℓ} {A : Set c} (R : Rel A ℓ) where
+
+  data PreFreeMagmaTheory : Rel (PreFreeMagma A) (c ⊔ ℓ)
+
+  open Defs PreFreeMagmaTheory
+
+  data PreFreeMagmaTheory where
+
+    var : ∀ {a b} → (R a b) → PreFreeMagmaTheory (var a) (var b)
+    _∙_ : Congruent₂ _∙_
+
+PreFreeTheorySyntax : ∀ {c ℓ} {A : Set c} (R : Rel A ℓ) → Rel (PreFreeMagma A) (c ⊔ ℓ)
+PreFreeTheorySyntax R = PreFreeMagmaTheory where open PreFreeTheory R
+
+syntax PreFreeTheorySyntax R m n = m ≈[ R ] n
+
+module PreservesEquivalence {c ℓ} {A : Set c} (R : Rel A ℓ) where
+
+  open PreFreeTheory R
+
+  _≈R_ = λ m n → m ≈[ R ] n
+
+  refl : (Reflexive R) → Reflexive _≈R_
+  refl r {var _} = var r
+  refl r {_ ∙ _} = (refl r) ∙ (refl r)
+
+  sym : (Symmetric R) → Symmetric _≈R_
+  sym s (var r)   = var (s r)
+  sym s (r₁ ∙ r₂) = sym s r₁ ∙ sym s r₂
+
+  trans : (Transitive R) → Transitive _≈R_
+  trans t (var r)   (var s)   = var (t r s)
+  trans t (r₁ ∙ r₂) (s₁ ∙ s₂) = trans t r₁ s₁ ∙ trans t r₂ s₂
+
+  preservesEquivalence : IsEquivalence R → IsEquivalence _≈R_
+  preservesEquivalence eq = record { refl = refl r ; sym = sym s ; trans = trans t }
+    where open IsEquivalence eq renaming (refl to r; sym to s; trans to t) 
+
+------------------------------------------------------------------------
+-- Free algebra on a Setoid
+
+module FreeMagmaOn {c ℓ} (S : Setoid c ℓ) where
+
+  open Setoid S renaming (Carrier to A; isEquivalence to isEq)
+
+  open PreFreeTheory _≈_
+
+  open PreservesEquivalence _≈_
+
+  open Strs _≈R_
+
+  isMagma : IsMagma  _∙_
+  isMagma = record { isEquivalence = preservesEquivalence isEq ; ∙-cong = _∙_ }
+
+  freeMagma : Magma c (c ⊔ ℓ)
+  freeMagma = record { Carrier = PreFreeMagma A; _≈_ = _≈R_ ; _∙_ = _∙_ ; isMagma = isMagma } 
+
+{- in the propositional case, we can immediately define the following
+   but how best to organise this under the Algebra.Bundles.Free hierarchy? -}
+
+------------------------------------------------------------------------
+-- Free algebra on a Set
+
+module FreeMagma {c} (A : Set c) where
+
+  private Carrier = PreFreeMagma A
+
+  _≈_ : Rel Carrier c
+  m ≈ n = m ≈[ _≡_ ] n
+
+  open PreFreeTheory {A = A} _≡_
+
+  ≈⇒≡ : ∀ {m n} → m ≈ n → m ≡ n
+  ≈⇒≡ (var ≡refl) = ≡refl
+  ≈⇒≡ (eq₁ ∙ eq₂) = cong₂ _∙_ (≈⇒≡ eq₁) (≈⇒≡ eq₂)
+
+  refl : Reflexive _≈_
+  refl {var _} = var ≡refl
+  refl {_ ∙ _} = refl ∙ refl
+
+  ≡⇒≈ : ∀ {m n} → m ≡ n → m ≈ n
+  ≡⇒≈ ≡refl = refl
+
+  rawFreeMagma : RawMagma c c
+  rawFreeMagma = record { Carrier = Carrier ; _≈_ = _≡_ ; _∙_ = _∙_ }
+
+  open Strs {A = Carrier} _≡_
+
+  isMagma : IsMagma _∙_
+  isMagma = record { isEquivalence = ≡isEquivalence ; ∙-cong = cong₂ _∙_ }
+  
+  freeMagma : Magma c c
+  freeMagma = record { RawMagma rawFreeMagma ; isMagma = isMagma }
+
--- a/src/Algebra/Bundles/Free.agda
+++ b/src/Algebra/Bundles/Free.agda
@@ -1,0 +1,11 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definitions of 'pre-free' and 'free' bundles
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Algebra.Bundles.Free where
+
+open import Algebra.Bundles.Free.Magma public


JacquesCarette:
Hmm, a lot of this can be made generic. [All of it can be generated, but that's a different story for another day.] I do have some code in the [Theories and Data Structures](https://github.com/JacquesCarette/TheoriesAndDataStructures) repository that should be resurrected and cleaned up for that purpose.

Even if we agree to write a lot of this "by hand", there is tons to discuss regarding the details of the design. Certainly `Magma` is a good place to start; I'd suggest `Pointed` as the 'next' one to do, to see what patterns emerge.

I'll add doing a thorough commentary on this to my "to do" list.

jamesmckinna:
Thanks for the prompt to think about automation, and I look forward to your detailed comments! 

As to patterns present in the design, thinking a bit harder about being systematic with naming would be helpful... as would a bit more `Setoid` yoga to state when a function is actually a `Setoid` morphism... and other such niceties.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+
+  data PreFreeMagmaTheory where
+
+    var : ∀ {a b} → (R a b) → PreFreeMagmaTheory (var a) (var b)
+    _∙_ : Congruent₂ _∙_
+
+PreFreeTheorySyntax : ∀ {c ℓ} {A : Set c} (R : Rel A ℓ) → Rel (PreFreeMagma A) (c ⊔ ℓ)
+PreFreeTheorySyntax R = PreFreeMagmaTheory where open PreFreeTheory R
+
+syntax PreFreeTheorySyntax R m n = m ≈[ R ] n
```
JacquesCarette, line 75: Nice! (meant for several lines above)

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+
+
+------------------------------------------------------------------------
+-- Free algebra on a Set
+{-
+   in the propositional case, we can immediately define the following
+   but how best to organise this under the Algebra.Bundles.Free hierarchy?
+   e.g. should we distinguish Free.Magma.Setoid from Free.Magma.Propositional?
```
JacquesCarette, line 83: Unfortunately, I think the `Setoid` and propositional cases do need to be separated. Reason: in Agda, some free algebras don't exist in the propositional case. Example: free Commutative Monoid exists over `Setoid` but not over `Set`.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+  private Carrier = PreFreeMagma A
+
+  _≈_ : Rel Carrier c
+  m ≈ n = m ≈[ _≡_ ] n
+
+  open PreFreeTheory {A = A} _≡_
+
+  ≈⇒≡ : ∀ {m n} → m ≈ n → m ≡ n
+  ≈⇒≡ (var ≡-refl) = ≡-refl
+  ≈⇒≡ (eq₁ ∙ eq₂) = cong₂ _∙_ (≈⇒≡ eq₁) (≈⇒≡ eq₂)
```
JacquesCarette, line 97: Notice how the same pattern is used in this proof as well?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+  refl {var _} = var ≡-refl
+  refl {_ ∙ _} = refl ∙ refl
+
+  ≡⇒≈ : ∀ {m n} → m ≡ n → m ≈ n
+  ≡⇒≈ ≡-refl = refl
+
+  rawFreeMagma : RawMagma c c
+  rawFreeMagma = record { Carrier = Carrier ; _≈_ = _≡_ ; _∙_ = _∙_ }
+
+  open Strs {A = Carrier} _≡_
```
JacquesCarette, line 109: Took me a bit. I would use `Structs` as `Strs` makes me think of 'String'.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+  sym s (var r)   = var (s r)
+  sym s (r₁ ∙ r₂) = sym s r₁ ∙ sym s r₂
+
+  trans : Transitive R → Transitive _≈R_
+  trans t (var r)   (var s)   = var (t r s)
+  trans t (r₁ ∙ r₂) (s₁ ∙ s₂) = trans t r₁ s₁ ∙ trans t r₂ s₂
+
+  preservesEquivalence : IsEquivalence R → IsEquivalence _≈R_
+  preservesEquivalence isEq = record { refl = refl r ; sym = sym s ; trans = trans t }
+    where open IsEquivalence isEq renaming (refl to r; sym to s; trans to t)
```
JacquesCarette, line 140: `where module REq = IsEquivalence isEq` and then use `REq.refl` in the record (etc)?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+
+    isRelHomomorphism : IsRelHomomorphism _≈_ _≈ᴹ_ ⟦_⟧ᴹ
+    isRelHomomorphism = record { cong = cong }
+
+    isMagmaHomomorphism : IsMagmaHomomorphism rawMagmaᴬ rawMagmaᴹ ⟦_⟧ᴹ
+    isMagmaHomomorphism = record { isRelHomomorphism = isRelHomomorphism
+                                 ; homo = λ _ _ → reflᴹ
+                                 }
+
+    unfold-⟦_⟧ᴹ : ∀ t → ⟦ t ⟧ᴹ ≈ᴹ algᴹ (map η t)
```
JacquesCarette, line 227: Nice! Though I wonder if this is also known as 'zig' in the full proof of adjointness?

JacquesCarette (commented):
See lots of inline comments.

JacquesCarette:
I think a lot of my comments would make even more sense if you did a second one - maybe even `Free Ring` would be good, to really see all the duplication. But `Pointed` and `Unar` would be good exercises as well.

For the 'impossibility' stuff, `Free Idempotent Magma` will already show that some properties are quite different.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+  using (_≡_; cong₂) renaming (refl to ≡-refl; isEquivalence to ≡-isEquivalence)
+import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
+
+
+------------------------------------------------------------------------
+-- 'pre'-free algebra
+
+infixl 7 _∙_
+
+data PreFreeMagma {a} (A : Set a) : Set a where
```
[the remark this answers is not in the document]
jamesmckinna: Perhaps: I was mulling this over before your comments, and I thought that better, in the interests of genericity, and to reflect the character of what I'm doing better, it should called `Syntax`... and then `Var` makes perfect sense.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+
+  var : A → PreFreeMagma A
```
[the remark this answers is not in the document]
jamesmckinna: See above

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+
+module _ {a b} {A : Set a} {B : Set b} where
+
+  map : (A → B) → PreFreeMagma A → PreFreeMagma B
+  map f (var a) = var (f a)
+  map f (s ∙ t) = (map f s) ∙ (map f t)
+
+module _ {a b c} {A : Set a} {B : Set b} {C : Set c} where
+
+  map-id : ∀ (t : PreFreeMagma A) → map id t ≡ t
```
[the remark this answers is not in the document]
jamesmckinna: Yes, I trip over this repeatedly (the semantics of these internal sectioning/scoping mechanisms was/us different in other systems I have used).

And yes, `variable` declarations are the way to go, but interacts badly later with the idiom of opening a `Setoid` or other algebra to expose its underlying `Carrier` renamed to one of the ore-existing variable names. Is this a bug?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,262 @@
...
+  map-id (var a) = ≡-refl
+  map-id (s ∙ t) = cong₂ _∙_ (map-id s) (map-id t)
+
+  map-∘ : (g : A → B) → (f : B → C) → ∀ t → map (f ∘ g) t ≡ (map f ∘ map g) t
+  map-∘ g f (var a) = ≡-refl
+  map-∘ g f (s ∙ t) = cong₂ _∙_ (map-∘ g f s) (map-∘ g f t)
```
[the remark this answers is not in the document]
jamesmckinna: Do you can induction as a higher order function, and then apply it to many individual functions, or present directly via pattern matching? As with other discussions, I think it's partly a matter of taste. And for didactic purposes, pattern matching seems always preferable?

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: But still, just as the functor should perhaps better be called `Syntax`, this should have a better name. But the mutual declare/import definitions/define trick us part of what makes Agda beautiful to work with

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Good point. I think you will have detected my reticence about `Set` structure only being what I dubbed 'pre-free', because raw syntax doesn't admit the right equations once the algebra gets interesting. I might now call this `RawFree` or `FreeRaw`. But in the case, it ok because `Magma` had no equations?

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Sure, but notice that it's a different structural induction principle here, because I've tried hard to overload the constructors 'sensibly'

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Yes... stdlib convention seems typically to use `Defs` and `Str(s)`, but there's no reason not to be more explicit. But `String`s? Not my first association with `Strs`...

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Interesting idiom. Yes, perhaps that is better. In any case, all this should now, I think, be folded into the module defining the equations theory.

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Yes, perhaps. I'm never quite sure how people think about/talk about these special cases, but I guess I had `foldMap` at the back if my mind, but wanted a more neutral/descriptive name.

jamesmckinna:
Major refactoring coming up...

jamesmckinna:
Re: `Pointed` ... it doesn't unfortunately exist under the `Relation.Binary` hierarchy alongside `Setoid`.

I guess one of us should open an issue for this?

CHANGELOG.md
```diff
@@ -1325,6 +1325,12 @@ New modules
   Algebra.Construct.Flip.Op
   ```
 
+* Algebraic structures obtained as the free thing (for their signature):
+  ```
+  Algebra.Bundles.Free
```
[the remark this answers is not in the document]
jamesmckinna: See the radical refactoring... I'm fairly clear that `FreeMagma` lifts from `SetoidHomomorphism` to `MagmaHomomorphism` (that's defined now, but not its functoriality), so yes, eventually, a full left adjoint. But I'd like the current 'design' to stabilise before tackling that ;-)

JacquesCarette:
`Pointed` qua theory induces `Maybe`; I'm sure you know that, but I'm just making sure we're talking about the same thing.

So I was picturing `Pointed` as belonging to `Algebra` not `Relation.Binary`.

jamesmckinna:
Re; `Pointed` yes as to `Maybe` (but I would perversely insist on writing `Syntax` again in this instance). 

But `Setoid`-with-a-point: is it an `Algebra`, or a special kind of `Setoid`, and hence should live with those things under `Relation.Binary`? (this may be a library design pain point regarding `Setoid`: is that an `Algebra`, albeit one with no operations?)

jamesmckinna:
This last commit revealed another 'missing' item from the library: there seems to be no `Bundle`d version of `MagmaHomomorphism` etc., with `Algebra.Morphism.Bundles` as the (obvious?) candidate place for such a thing to live?

jamesmckinna:
OK... after functoriality of the `FreeMagmaFunctor`, I think it's time to stop. 

One outstanding question: besides resolving the above review conversations, where should such a contribution live, assuming that we want it anywhere?

I chose `Algebra.Bundles.Free.Magma`, but perhaps `Algebra.Free.Magma` would be better?

CHANGELOG.md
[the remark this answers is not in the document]
jamesmckinna: UPDATED: it is now (modulo equality on `SetoidHomomorphism`) a full left adjoint to the forgetful functor (I think!)

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
jamesmckinna: I side-stepped the 'bug' by introducing slightly better names, in particular `UA` etc. for the `Carrier` of the (underlying) `Setoid` (of the `Magma` instance)

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: If what you want is to functorise the signatures, and make all of this machinery, syntax plus theory-relator, key off such a signature functor, then I'd perhaps/probably agree. But the library as is seems quite at ease with very 'concrete' presentations of `Algebra`s...?

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: IIRC, `algᴹ` is the (left) mate across the adjunction of the identity on the underlying `Setoid`. The `varSetoidHomomoprhism` is the `η` of the adjunction (I've overloaded `η`, perhaps unfortunately, because I want to use it for 'injection of generators' as well in the defintion of `Eval`), and there'll be (yet) more work to do to show that their compositions do the right thing...

jamesmckinna:
Blast, I've screwed up revertign the bogus merge commit again... sigh.

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
JacquesCarette: I agree that for didactic purposes, pattern matching will always be clearer.

More fodder for stdlib-5.0.

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
JacquesCarette: It might be a bug. Is it worth finding a MWE?

src/Algebra/Bundles/Free/Magma.agda
JacquesCarette: It's ok for all theories that have no equations; you can have some equations too, but these have to be 'linear' and 'rigid' - but that's another matter. So it will work for all theories X where the term algebra is the free X for sure.

src/Algebra/Bundles/Free/Magma.agda
JacquesCarette: I'm still trying to figure out what I want to functorise and what I want to generate. And by 'generate', I mean human-friendly generation, at least for now.

Though having a good `deriving` for Agda sure would be nice.

Issue #1957, opened 2023-04-29
`Pointed`: `Setoid` with a distinguished point


jamesmckinna:
This concept doesn't exist in the library (for `Set`, `Data.Maybe` does the trick of course), so we should add it.

But where should it belong? A `Setoid`-with-a-point: is it an `Algebra` (with a very boring signature, consisting of a single nullary operation), or a special kind of `Setoid`, and hence should live with those things under `Relation.Binary`? 

(this may be a separate library design pain point regarding `Setoid`: is that an `Algebra` too, albeit one with no operations?)

UPDATED: grrr. `Pointed` introduced in two ways already. Once to add a distinguished point; this apparently lives under `Relation.Nullary.Construct.Add.Point` where
```agda
open import Data.Maybe.Base public
  using () renaming (Maybe to Pointed; nothing to ∙; just to [_])
```
and once to add a distinguished point which is moreover an identity for a binary operation. And overloading the shared common use of `Maybe` as the representing `Free` functor. This seems a bit wart-y how to proceed? 

Either way, we might then want a `Free` such thing, (cf. #1962 / #1954 ) and know that its algebra is that of the usual adjoint/monadic situation arising from `Maybe` and the forgetful operation of throwing away the distinguished point.

Taneb:
I'm not at a computer this weekend but `Relation.Binary.Construct.Add.Point` is relevant

JacquesCarette:
I picture `Pointed` as a *theory*, so a record. What `Relation.Nullary.Construct.Add.Point` does is build the **free** `Pointed` on a `Set`, as @jamesmckinna says.

`Maybe` isn't a theory, it's a data-structure -- which is a standard implementation of the Free Pointed Set.

jamesmckinna:
Disentangling the existing design to separate out the various use cases seems like:
"Another one to punt until after v2.0, I think ;-)"

JacquesCarette:
I still see it as something that belongs in `Algebra`. It could be called `Pointed` but also `PointedCarrier`, `PointedSet`, `PointedSetoid`, I'm ok with all of those.

jamesmckinna:
I'm going to close this for now:
* difficulty figuring out how to refactor the existing (overlapping) uses of `Maybe` into a more coherent design
* failure to achieve consensus on how to tackle `Free` constructions generally #1962 and much other discussion on- and off-line about this.

I'll re-open as and when we make progress on the second? Hopefully the first will follow (more) straightforwardly after that?

PR #1958, opened 2023-04-29
added `Pointed` algebraic structure, (raw) bundle and homomorphism

jamesmckinna:
See issue #1957 

TODOs:
- [ ] add a `RawPointed` instance to `Data.Maybe.Base`
- [ ] `FreePointed` functor, as in #1962 
- [ ] `Data.Nat.Base`? `List`? what else?
- [ ] forgetful functor from `Pointed` back to `Setoid`
- [ ] what else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -835,6 +835,8 @@ Major improvements
   * `RawRing`
   * `RawQuasigroup`
   * `RawLoop`
+* A new definition is also introduced: 
+  *  `RawPointed`
 * A new module `Algebra.Lattice.Bundles.Raw` is also introduced.
   * `RawLattice` has been moved from `Algebra.Lattice.Bundles` to this new module.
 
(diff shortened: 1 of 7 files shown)


jamesmckinna:
Hmmm... the library disagrees with @JacquesCarette 's analysis that `Pointed` belongs under `Algebra`:
```agda
module Relation.Nullary.Construct.Add.Point where

open import Data.Maybe.Base public
  using () renaming (Maybe to Pointed; nothing to ∙; just to [_])
```
Maybe this PR isn't required after all?

JacquesCarette:
See my comments on #1957 .

There's a difference between the theory and the free algebra associated to it. I meant the theory, which belongs in `Algebra`. The data structure belong in `Data`. That the construction lives under `Relation` is fine, but `Relation.Nullary` is an aberration. What's the use of a relation on nothing?

JacquesCarette (approved):

jamesmckinna:
As with #1963 I'm going to close this for now, pending (the design of) a resolution of #1957, at which point this PR might be the (best?) place to implement such a solution.

Issue #1960, opened 2023-05-01
Bundles for `Algebra.*Homomorphism`, and their algebra: `Hom`-'sets' for `Algebra`


jamesmckinna:
[cf. #1962 where there's the beginnings of a possible design for this, so that I could properly define a left adjoint to the forgetful functor `Magma -> Setoid`, witnessed on objects by `Magma.setoid`, but not really properly on morphisms. ]

`Algebra.Morphism.Structures` defines all the *properties* eg `isMagmaHomomorphism` of being a `MagmaHomomorphism`, but there's nowhere (?) in `stdlib` where those properties are bundled with their morphisms between the underlying `Setoids`.

Correspondingly, identity and composition for such things are distributed between `Relation.Binary.Morphism.Construct` and `Algebra.Morphism.Construct`, and which again operate only on the underlying proofs of properties. 

Proposal: we add actual bundled homomorphisms for all the algebraic `Structures` in the library, on a 'one structure, one module' basis, to which we also package the `Identity` and `Composition` structures. UPDATED: bundled homomorphisms now under `Algebra.Morphism.Bundles`, with their `Identity` and `Composition` now handled under `Algebra.Morphism.Construct`

Now, this may already all be done (nicely, better, ...?) in `agda-categories`, and if so, then a plea to merge in this stuff.

Taneb:
I support this proposal. It's not currently done better in `agda-categories`, or indeed at all. I'm working on that on the agda-categories side but was just planning to use Data.Product

Taneb:
I'm not convinced we should have one module per structure but not strongly against it. I'd be interested to hear your justification of that over one file with all the bundles in it, which matches closer what we do elsewhere in the algebra hierarchy

jamesmckinna:
> but was just planning to use `Data.Product`
Is this because all the bundling/unbundling gets to painful?

Taneb:
> > but was just planning to use `Data.Product`
> Is this because all the bundling/unbundling gets to painful?

No, it's simply because the bundles didn't exist. It works a little better with some subcategory machinery in `agda-categories` but it's still not a perfect match

jamesmckinna:
Re: one module per one algebra vs. how it is currently done now:
I think I managed to get confused as to whether all the various instances of `identity` and `compose` would share the same namespace happily? But if so, then indeed, consistency with current practice demands such an approach.

Taneb:
I'd certainly be happy to see some experimentation with the design here

[...]

PR #1962, opened 2023-05-01
The free `Magma` on a `Set`, resp. `Setoid` [bis]

jamesmckinna:
Clean version of #1954


src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+
+    isRelHomomorphismᴹ : IsRelHomomorphism _≈_ _≈ᴹ_ ⟦_⟧ᴹ
+    isRelHomomorphismᴹ = record { cong = cong-⟦_⟧ }
+
+    setoidHomomorphismᴹ : SetoidHomomorphism FA setoidᴹ
+    setoidHomomorphismᴹ = record { ⟦_⟧ = ⟦_⟧ᴹ ; isRelHomomorphism = isRelHomomorphismᴹ }
+
+    isMagmaHomomorphismᴹ : IsMagmaHomomorphism rawMagma rawMagmaᴹ ⟦_⟧ᴹ
+    isMagmaHomomorphismᴹ = record { isRelHomomorphism = isRelHomomorphismᴹ
+                                  ; homo = λ s t → begin ⟦ s ⟧ᴹ ∙ᴹ ⟦ t ⟧ᴹ ∎ }
```
JacquesCarette, line 283: is the begin and end really needed?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+
+  module Corollary (𝓗 𝓚 : η-MagmaHomomorphism)
+    where
+      open η-MagmaHomomorphism 𝓗 renaming (⟦_⟧ to ⟦_⟧ᴴ)
+      open η-MagmaHomomorphism 𝓚 renaming (⟦_⟧ to ⟦_⟧ᴷ)
+      open Uniqueness 𝓗 renaming (isUnique⟦_⟧ᴹ to isUnique⟦_⟧ᴴ)
+      open Uniqueness 𝓚 renaming (isUnique⟦_⟧ᴹ to isUnique⟦_⟧ᴷ)
+      
+      isUnique⟦_⟧ :  ∀ t → ⟦ t ⟧ᴴ ≈ᴹ ⟦ t ⟧ᴷ
+      isUnique⟦ t ⟧ = begin ⟦ t ⟧ᴴ ≈⟨ isUnique⟦ t ⟧ᴴ ⟩ ⟦ t ⟧ᴹ ≈˘⟨ isUnique⟦ t ⟧ᴷ ⟩ ⟦ t ⟧ᴷ ∎
```
JacquesCarette, line 320: maybe do a proper vertical layout for this equational reasoning?

JacquesCarette (commented):
Looking good!

I can definitely see how to generate the vast majority of this (given just a few user-provided names).

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+
+------------------------------------------------------------------------
+-- Morphisms between Magmas: belongs in its own place
+-- Algebra.Morphism.Bundles
+-- open import Algebra.Morphism.Bundles using (MagmaHomomorphism)
+------------------------------------------------------------------------
+
+record MagmaHomomorphism (𝓐 : Magma a ℓa) (𝓑 : Magma b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
+
+  open Magma 𝓐 renaming (rawMagma to rawMagmaᴬ; setoid to setoidᴬ; Carrier to UA)
```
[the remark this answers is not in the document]
jamesmckinna: Hmmm, what does your proposal/idiom look like in practice? I'm less familiar with introducing local private modules...

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+private
+  variable
+    ℓ a ℓa b ℓb c ℓc m ℓm n ℓn : Level
+    A : Set a
+    B : Set b
+    C : Set c
+
+
+------------------------------------------------------------------------
+-- Morphisms between Magmas: belongs in its own place
```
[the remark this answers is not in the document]
jamesmckinna: Yes, as per #1958.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+  setoidHomomorphism = record { ⟦_⟧ = ⟦_⟧ ; isRelHomomorphism = isRelHomomorphism }
+ 
+------------------------------------------------------------------------
+-- Syntax: 'pre'-free algebra
+
+module Syntax where
+
+  infixl 7 _∙_
+
+  data Syntax (A : Set a) : Set a where
```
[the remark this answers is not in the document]
jamesmckinna: It's a good question, and before any too-hasty merging, I think it might be good to discuss and agree the naming scheme here:
* I proposed `Syntax` because this is the datatype definition of the abstract syntax of terms over the signature given by the single binary operation; but I didn't abstract out the (datatype defining) the *functorial* version of that, with `Syntax` (or: whatever) then being the the free monad construction over such a functor. There's not a lot in it either way for this example, but there might be down the road... (it's also the case that the 'usual' definition of the free monad is only up-to-isomorphism, and other choices are available, IIRC, so again there might be choices to make); 
* I wrote `EquationalTheory` for the setoid structure, but again, other choices are available (I had considered simply writing `Equations` or `Theory` before settling on the slightly more verbose version); eg, in `Isabelle/HOL`, this relation transformer is usually just called the  `relator`, and perhaps that would be better if we call `Syntax` `Functor` (but I'm reluctant to do that because of the clash with `Effect.Functor`...);
* where there is, irritatingly, more source of variation, is in how the `...Theory` is presented: for this example, the relator preserves `isEquivalence`, but for more complicated non-free theories, you typically have to build equivalence in at the bottom, alongside congruence; a general account of how best to present congruence closures in the presence of non-free axioms would be helpful, but I don't know the literature that well;

The subsequent naming choices are also potentially up-for-grabs.

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+------------------------------------------------------------------------
+-- Free algebra on a Set
+{-
+   in the propositional case, we can immediately define the following
+   but how best to organise this under the Algebra.Bundles.Free hierarchy?
+   e.g. should we distinguish Free.Magma.Setoid from Free.Magma.Propositional?
+-}
+
+module FreeRawMagma (A : Set a) where
+
```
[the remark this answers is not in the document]
jamesmckinna: (Vertical) Whitespace helps me concentrate/focus better, but if `stdlib` suggests suppressing it, happy to go either way.

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Interestingly it did seem to be, but I will look again. I was torn between and anonymous appeal to `refl` (and hence having to open a structure enough to expose that field, again with a sufficiently standardised-apart name) or else to make a trivial bit of 'equational reasoning' explicit so that the term is exhibited. The typechecker was grumbling about insufficiently specified field projection names at this point, and this was one way out of the bottle. But it's a bit of a wart, so I'll rethink/reconsider.

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Again, here, because it really is an equation of the form `trans eq1 (sym eq2)` I though it easier to inline it. But maybe producing a custom combinator for this case would be clearer?. It doesn't feel elaborate enough to rquire a full Feijen-style layout, though.

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
MatthewDaggitt:
````
module A = Magma 𝓐
module B = Magma 𝓑

field
    ⟦_⟧ : A.Carrier → B.Carrier
```

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+    C : Set c
+
+
+------------------------------------------------------------------------
+-- Morphisms between Magmas: belongs in its own place
+-- Algebra.Morphism.Bundles
+-- open import Algebra.Morphism.Bundles using (MagmaHomomorphism)
+------------------------------------------------------------------------
+
+record MagmaHomomorphism (𝓐 : Magma a ℓa) (𝓑 : Magma b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
```
MatthewDaggitt, line 49: This should live in a new module `Algebra.Morphism.Bundles`?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+
+module EquationalTheory {A : Set a} (_≈ᴬ_ : Rel A ℓ) where
+
+  open Syntax
+
+  infix 4 _≈_
+
+  data _≈_ : Rel (Syntax A) (a ⊔ ℓ)
+
+  open Definitions _≈_
```
MatthewDaggitt, line 114: Whoa, didn't know you could do this! Having said that, I think it's unnecessary and a bit confusing, all to avoid writing `Congruent₂ _≈_ _∙_` instead of `Congruent₂ _∙_`...

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+
+  data _≈_ where
+
```
MatthewDaggitt, line 117: Can we remove the empty line?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+  sym s (var s₀)  = var (s s₀)
+  sym s (s₁ ∙ s₂) = sym s s₁ ∙ sym s s₂
+
+  trans : Transitive _≈ᴬ_ → Transitive _≈_
+  trans t (var r₀)  (var s₀)  = var (t r₀ s₀)
+  trans t (r₁ ∙ r₂) (s₁ ∙ s₂) = trans t r₁ s₁ ∙ trans t r₂ s₂
+
+  preservesEquivalence : IsEquivalence _≈ᴬ_ → IsEquivalence _≈_
+  preservesEquivalence isEq = record
+    { refl = refl Eq.refl ; sym = sym Eq.sym ; trans = trans Eq.trans }
```
MatthewDaggitt, line 135: Can we layout all the records with more than one field, with one field per line?

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+  open EquationalTheory _≈ᴬ_ public
+    renaming (_≈_ to _≈ᵀ_) hiding (refl; sym; trans)
+
+  open Structures _≈ᵀ_
+
+  isMagma : IsMagma  _∙_
+  isMagma = record { isEquivalence = preservesEquivalence isEqᴬ ; ∙-cong = _∙_ }
+
+  freeMagma : Magma a (a ⊔ ℓa)
+  freeMagma = record { Carrier = Syntax Carrier; _≈_ = _≈ᵀ_ ; _∙_ = _∙_ ; isMagma = isMagma }
```
MatthewDaggitt, line 195: Can get away with just `freeMagma = record { isMagma = isMagma }`

src/Algebra/Bundles/Free/Magma.agda
```diff
@@ -0,0 +1,443 @@
...
+  open Setoid 𝓐 renaming (isEquivalence to isEqᴬ; _≈_ to _≈ᴬ_)
+
+  open Syntax
+
+  open EquationalTheory _≈ᴬ_ public
+    renaming (_≈_ to _≈ᵀ_) hiding (refl; sym; trans)
+
+  open Structures _≈ᵀ_
+
+  isMagma : IsMagma  _∙_
```
MatthewDaggitt, line 191: Double empty space.

src/Algebra/Bundles/Free/Magma.agda
MatthewDaggitt: But without that layout it's very difficult to tell it's actually a `trans ... (sym ...)`!

MatthewDaggitt (commented):
Hmm I'm a little bit unsure of all these complicated named and unnamed submodules. Generally when they appear it means we haven't got the organisation quite right. It also makes it harder for the user to understand what's in scope at any given point in time.

I wonder how we can simplify it? While doing so we should consider how the organisation will be extended when encoding the free variants of more complicated structures.

>  I'm now starting to worry about where they should live:
Algebra.Bundles.Free.X or Algebra.Free.X? or somewhere else?

Definitely prefer `Algebra.Free.X`

src/Algebra/Bundles/Free/Magma.agda
Taneb: If you're concerned about vertical space I'd consider writing it explicitly using `trans`

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: yes! issue #1960

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Well, I made the change, and so now it reads `Definitions.Congruent₂ _≈_ _∙_` .
Again, the (no longer especially) new, more flexible syntax for separating `data` declarations from their definitions makes this possible, even if (perhaps?) not desirable here.

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: I've done this for some of them, but not all, and it doesn't really solve the problem when we get loooong definitions involving composition(s) of `IsMagmaHomomorphism` proofs... :-(

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
jamesmckinna: Largely done now.

src/Algebra/Bundles/Free/Magma.agda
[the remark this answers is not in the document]
jamesmckinna: Well, I'm less happy than I at first thought I would be, so I'm sure I will need to do another pass to weed out more of this (to me, helpful) 'layout'

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: Now replaced by appeal to `refl` instead.

src/Algebra/Bundles/Free/Magma.agda
jamesmckinna: I've decided to leave it as it is.

src/Data/List/Properties.agda
```diff
@@ -872,7 +872,7 @@ module _ {R : Rel A p} (R? : B.Decidable R) where
   length-derun : ∀ xs → length (derun R? xs) ≤ length xs
   length-derun [] = ≤-refl
   length-derun (x ∷ []) = ≤-refl
-  length-derun (x ∷ y ∷ xs) with ih ← length-derun (y ∷ xs) | does (R? x y) 
+  length-derun (x ∷ y ∷ xs) with ih ← length-derun (y ∷ xs) | does (R? x y)
```
jamesmckinna, line 875: @gallais This change should have happened in master, not in this PR!

src/Data/Nat/Combinatorics.agda
```diff
@@ -118,7 +118,7 @@ module _ {n k} (k<n : k < n) where
 
     [n-k]!           = [n-k] !
     [n-k-1]!         = [n-k-1] !
-  
+
```
jamesmckinna, line 121: Ditto. @gallais This change should have happened in master, not in this PR!

jamesmckinna:
Many thanks for all the very helpful and constructive review suggestions, esp. wrt use of private modules.

jamesmckinna:
@MatthewDaggitt 

> Definitely prefer `Algebra.Free.X`

I agree, and moved things accordingly.

jamesmckinna:
> Hmm I'm a little bit unsure of all these complicated named and unnamed submodules. Generally when they appear it means we haven't got the organisation quite right. It also makes it harder for the user to understand what's in scope at any given point in time.
> 
> I wonder how we can simplify it? While doing so we should consider how the organisation will be extended when encoding the free variants of more complicated structures.

So... I agree. I think (and hope!) that the use of private modules, and reverting to a more standard naming scheme, now means that the scoping is (much) easier to grasp. 

But as for the use of nested modules in general, I personally do find it easier to use them to help localise/focus my thinking. And since the constructions are, categorically, unique 'up to', I've definitely found it helpful to keep the various appeals to `Existence` and `Uniqueness` separate, and local.

src/Data/Nat/Combinatorics.agda
jamesmckinna: A `rebase` seems to have removed these spurious commits.

src/Data/List/Properties.agda
jamesmckinna: A `rebase` seems to have removed these spurious commits.

src/Algebra/Free/Magma.agda
```diff
@@ -0,0 +1,441 @@
...
+import Relation.Binary.Morphism.Construct.Composition as Compose
+open import Relation.Binary.PropositionalEquality
+  using (_≡_; cong₂; _≗_)
+  renaming (refl to ≡-refl; isEquivalence to ≡-isEquivalence)
+import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
+
+private
+  variable
+    ℓ a ℓa b ℓb c ℓc m ℓm n ℓn : Level
+    A : Set a
```
JacquesCarette, line 40: Note that variables are scoped / generative (or whatever the right name is). What this means is that you can do
```agda
     A B C : Set ℓ
 ```
 and at use site, each will have different `Level` automatically. The same idea can help you shorten the `Setoid` and `Magma` declarations to be 1-liners too.

Comment (jamesmckinna) [reply src/Algebra/Free/Magma.agda:40]: Interesting! Thanks for the nudge. 
I'll (try to remember to) implement the idea in subsequent versions/variations, but not now, I think.

File: src/Algebra/Free/Magma.agda [hunk 16]
```diff
@@ -0,0 +1,441 @@
...
+
+  map-id : map {A = A} id ≗ id
+  map-id (var a) = ≡-refl
+  map-id (s ∙ t) = (map-id s) ∙-cong (map-id t)
+
+  map-∘ : (g : A → B) → (f : B → C) → map (f ∘ g) ≗ (map f ∘ map g)
+  map-∘ g f (var a) = ≡-refl
+  map-∘ g f (s ∙ t) = (map-∘ g f s) ∙-cong (map-∘ g f t)
+
+  syntaxRawFunctor : RawFunctor (Syntax {a})
```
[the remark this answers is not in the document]
jamesmckinna: Yes, I think I agree as well, on both points. But I'll try to confine any (more) edits/commits to cosmetic ones such as this. But next week...

src/Algebra/Free/Magma.agda
```diff
@@ -0,0 +1,441 @@
...
+------------------------------------------------------------------------
+-- parametrised 'equational' theory over the 'pre'-free algebra
+
+module EquationalTheory {A : Set a} (_≈ᴬ_ : Rel A ℓ) where
+
+  open Syntax
+
+  infix 4 _≈_
+
+  data _≈_ : Rel (Syntax A) (a ⊔ ℓ) where
```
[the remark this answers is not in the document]
jamesmckinna: I'll have to think about this some more. So... also next week (at the earliest). Maybe discuss at SPLS?
Not clear to me at the moment, without a lot more thought/discussion, what the extra generality buys you, *for this particular* collection of constructions. 

In particular, when considering 'non-free' (quotiented) theories such as eg that of `Monoid` or `Ring`, the modularity of these proofs of lifting of reflexivity etc. start to get fairly hairy fairly quickly without building them in from the ground up. But perhaps there are insights from frex which might be relevant here?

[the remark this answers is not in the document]
jamesmckinna: Sorry that I haven't had any time or brainspace in the meantime to think harder about the question. 

That said, and in the spirit of "don't let PRs bitrot through mission creep"/perfect-is-the-enemy-of-the-good(-enough), I'm tempted to say: by all means, but in a downstream PR that refactors this one.
After all, it is in some sense only a prototype for something which can/should be done generically, for which I really didn't have the brainspace to work through the details... sigh.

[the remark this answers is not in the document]
JacquesCarette: I agree that waiting for perfection is a bad idea and that such improvements should wait for a later PR.

jamesmckinna:
Regarding other instances of `LeftAdjoint`/`Free` constructions (I agree that we don't have good naming conventions for *functorial* constructions /constructions defined on functors in the library), I've also been experimenting with Church-style, level-raising versions. Now I'm undecided as to the best way to proceed: (more or less...)
```agda
FreeMagma (X : Setoid) = (M : Magma {c} {l}) (open Magma M) -> (X ⇒ setoid) -> setoid
```
etc. compared to explicitly constructing the syntax and equational theory presentation. Thoughts?

MatthewDaggitt:
Looks like the merge has gone wrong somewhere.

jamesmckinna:
Re: merge going wrong. yes, but how? This was fine, for a good while... and hasn't changed, so what has?

jamesmckinna:
Converting to `DRAFT` as a way of avoiding having to think about the `CHANGELOG` merge conflicts for the time being, until after v2.0 finally gets merged/released.

jamesmckinna:
Can we (now) merge this, please?

JacquesCarette:
As stand-alone code, this is great.

But this also represents a non-trivial new design, and there's various things I am not so sure about. I don't want to merge this in right now, just to have a backwards compatibility problem in 3 months from now.

In places, this code seems really quite verbose (and that might be inevitable). And it might benefit quite a lot from the vapourware of a merge of parts of agda-categories into stdlib. And shrink too.

So I don't feel like pushing the 'merge' button. I wouldn't be upset of someone else did - I just don't want to be the one to commit this design to the library!

jamesmckinna:
I'll convert back to `DRAFT` until the discussion concludes on whether to merge at all, and then fix up whatever `CHANGELOG` conflicts there are then.

jamesmckinna:
OK, a year has gone by, and even the one positive review is... unclear whether this should be merged.

I think it's worth pulling out the bundled `Homomorphism` commits, and trying to get *those* merged, ahead of returning to this one... but when? May 2025? :-(

JacquesCarette:
I agree that splitting off the immediately mergable stuff into a separate PR is a good idea. As to the rest, you'll need to explicitly prod some not-me reviewers into action.

jamesmckinna:
Marking as `blocked-by-issue` until #2383 (and possibly others!?) get completed and merged...

jamesmckinna:
Converting back to DRAFT:
* the `CHANGELOG` is too bit-rotted (pre-v2.0!)
* the dependency on (the fruits of) #2383 should be made explicit/refactored after any eventual merge of that PR
* the design should be further discussed (somewhere... where!?) before proceeding, but so far only @JacquesCarette has engaged with that, and then to express scepticism about the particular choices made here

This PR was only ever intended as a prototype/template for adding all manner of `Free` algebraic constructions, but if we can't agree on `Magma`, then perhaps it's only worth revisiting at some later point in time (despite eg #2407 and its antecedents/descendants being based on various implementations of `FreeMonoid`... one `Fin`-based, as here, the other in `Tactic.MonoidSolver` in 'functorial' style).

JacquesCarette:
So I think we should implement a few more `FreeX` to shake out the design space. Have a student who could, given a list of things to try.

jamesmckinna:
> So I think we should implement a few more `FreeX` to shake out the design space. Have a student who could, given a list of things to try.

@JacquesCarette that would be great:
* `FreeGroup` (and, in proof-relevant guise, to normalising/`Solver` architecture for compiling equational proofs #2629 )
* `FreeAbelianGroup` (we're missing `Solver` for `Integer`!)
* `Free(Commutative)Ring`

In each case, there's (morally) also a connection with the associated `Solver` architecture(s), and more generally, the choice between 'Free `X`' and 'Free *extension* of `X`'... so plenty of space for a student to explore!

JacquesCarette:
One could also envisage
- `FreeSemigroup` (non-empty lists)
- `FreeLeftZeroSemigroup` (aka 'First')
- `FreeRightZerosemigroup` (aka `Last`)
- `FreeMonoid` (!)

since they all have normal forms.

`FreeGroup` is tricky as it is quite different from `FreeMagma` and `FreeMonoid`: the latter have normal forms, but `FreeGroup` doesn't. So it feels like a completely different kettle of fish for which we don't really have a prototype.

jamesmckinna:
What's the relationship between `Free(Abelian)Group` and `FreeZModule` (suitably understood in the noncommutative case)?

JacquesCarette:
> What's the relationship between `Free(Abelian)Group` and `FreeZModule` (suitably understood in the noncommutative case)?

Good question - I don't know! Of course, Z-Modules and Abelian Groups are 'the same thing', but no idea how that lifts to `Free`.

JacquesCarette:
Apparently [they are the same](https://en.wikipedia.org/wiki/Free_abelian_group).

jamesmckinna:
See also https://github.com/agda/agda-stdlib/issues/2695#issuecomment-2798386074 and perhaps we should move the discussion back to #2539 to record the various choices, compared to any particular PR such as this.

PR #1963, opened 2023-05-01
DRAFT: `FreePointed` for comparison

jamesmckinna:
cf. #1954 / #1962 and @JacquesCarette 's request for comparison. See also issue #1957 for blocking dependencies. 

Besides library design issues, the only real difference is that of the arity (0, rather than 2) of the `pt` constructor, and hence also of the `cong`/`homo` properties of the associated `Is*Homomorphism` records. Otherwise, it was pure cut-and-paste, modulo the additional infrastructure for `Identity` and `Composition` for `IsPointedHomomorphism`.

### DIFF
--- a/src/Algebra/Bundles/Free/Pointed.agda
+++ b/src/Algebra/Bundles/Free/Pointed.agda
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Definitions of 'pre-free' and 'free' magma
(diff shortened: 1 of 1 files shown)


JacquesCarette:
The reason for the exercise was to show just how deeply the cut-and-paste goes.

jamesmckinna:
I'll close this for now, but potentially re-open it once:
- [ ] `Algebra.Free.Magma` stabilises and gets merged
- [ ] we work out the best way to resolve issue #1957

Issue #1917, opened 2023-01-30
`Algebra.Bundles.SemiringWithoutOne` fails to export `setoid`?


jamesmckinna:
More specifically, `Algebra.Bundles.Semiring` does indeed export all of its corresponding substructures (down to `setoid` derivable via `Algebra.Structures`), while `Algebra.Bundles.SemiringWithoutOne` appears not to. Not quite sure how to fix this, but will update if/when I do... with a minimum (non-)working example. 

EDITED: See the two `Test` files at `Algebra.Properties.Semiring{WithoutOne}` on [this branch](https://github.com/jamesmckinna/agda-stdlib/tree/SemiringWithoutOne) now minimised. 

UPDATE: the problem seems similar to that described/solved in issue #1159 / PR #1174, and I think I have a solution, but I'm conscious of being out of my depth as to the arcane define/public re-export structure of the `Algebra.*` hierarchy, so reluctant to issue a PR about this... Which raises a separate question: how 'high' (=fewest operations, axiomatic properties) in the `Structures` (and hence `Properties.BundleX.OperationY`) hierarchy do we want certain things to live: eg distributivity of `_*_` over `sum`, defined in #1462, placed it as a property of `Semiring`s; @uzytkownik 's original contribution (from a year earlier) under #1287 placed it under `SemiringWithoutOne`...  and AFAICT, it doesn't even need that much structure, only that of the theory pullback (which has no name) of `SemiringWithoutOne` and `SemiringWithoutAnnihilatingZero`. In pursuit of #1287, I'd be happy for everything to live under properties of `Semiring`, and leave the appropriate weakening of structure to subsequent refactoring... but I know that there are strong views about the ambient guiding principle of 'be no less general than you absolutely can', and I thought that this was worth flagging.

jamesmckinna:
Will see if I can knock this off along the lines of @Taneb 's #1967 ... later.

jamesmckinna:
Aaargh, this issue is really two: 
- [ ] exporting `setoid`
- [ ] adding more diamonds to find a home for eg distributivity of `_*_` over `sum`

The first might be easy (but I feel faint thinking about it), but the second?

jamesmckinna:
Re: the first. In my 'success' test, viz.
```agda
open import Algebra.Bundles using (Semiring)

module Algebra.Properties.Semiring.Test {c} {ℓ} (R : Semiring c ℓ) where

open Semiring R
{- succeeds for Semiring -}
[...]

PR #1967, opened 2023-05-11
Make sure RawRing defines rawRingWithoutOne

Taneb:


gallais (commented):
Can we please get a CHANGELOG entry?

Taneb:
I think tomorrow I'm going to do a more thorough survey for what inherited definitions might be missing

jamesmckinna:
See also #1917 and the discussion there about 'missing' exports...

jamesmckinna:
Did your investigations @Taneb yield any more items to add to this PR, or should we wrap this up now?

Taneb:
I drew a big graph of how everything in `Algebra.Structures` relates, and then... never did anything with it. This can be merged as-is, I'll make another PR if I do a more exhaustive check

Taneb:
I don't think a changelog entry is necessary, `RawRingWithoutOne` was only added in this release

Issue #1711, opened 2022-02-10
Add properties of the quotient to Data.Nat.Divisibility


Taneb:
I'd like there to be certain properties of the quotient of a proof of divisibility, for example:

```agda
foo : ∀ {m n} → (m∣n : m ∣ n) → quotient m∣n ∣ n
bar : ∀ {m n} (m∣n : m ∣ n) → m > 1 → quotient m∣n < n
```

I can't think of good names for these though

jamesmckinna:
How about: `quotient-|` and `quotient-<`? or else `*-divides`, `*-less-than`?

MatthewDaggitt:
I like `quotient-|` and `quotient-<`, better to include the name of the functions involved.

jamesmckinna:
Did we do this?

MatthewDaggitt:
Nope. Still open.

jamesmckinna:
In the spirit of #1969 you probably also want
```agda
baz : ∀ {m n} (m∣n : m ∣ n) → m > 1 → n > 1 → quotient m∣n > 1
```
as, say, `quotient->1`?

jamesmckinna:
And as a last thought: with the addition of `quotient≢0 : .{{NonZero n}} → NonZero quotient` to the `_|_` API, a number of arguments about involving possible-zero quotients might (hopefully) be automatically simplifiable by the typechecker, eg. eliminating [this case from `Data.Nat.Primality.euclidsLemma`](https://github.com/agda/agda-stdlib/blob/3a1bda24f1949c6fe0ba85d686745dcf246513a8/src/Data/Nat/Primality.agda#L107-L109)

PR #1969, opened 2023-05-13
Add prime factorization and its properties

Taneb:
This PR proves that every non-zero natural number has a unique (up to permutation) prime factorization.

To this end, it also adds a definition of "rough" numbers (numbers which don't have prime factors below a bound)

This is something that I think is important to have a proof for, but I think it needs considerable improvement before it meets the quality criterion for this library, and I can't see how to improve it, so I'd appreciate feedback.

The factorization algorithm is trial division. It does not currently short-circuit when it checks a factor greater than the square root of the number being factorized, which would significantly improve its speed in some cases, but makes the code a bit more complicated.


src/Data/Nat/Divisibility.agda
```diff
@@ -300,3 +300,9 @@ m≤n⇒m!∣n! m≤n = help (≤⇒≤′ m≤n)
   help : ∀ {m n} → m ≤′ n → m ! ∣ n !
   help {m} {n}     ≤′-refl        = ∣-refl
   help {m} {suc n} (≤′-step m≤′n) = ∣n⇒∣m*n (suc n) (help m≤′n)
+
+------------------------------------------------------------------------
+-- Properties of quotient
+
+quotient∣n : ∀ {m n} (m∣n : m ∣ n) → quotient m∣n ∣ n
+quotient∣n {m = m} m∣n = divides m (trans (_∣_.equality m∣n) (*-comm (quotient m∣n) m))
```
gallais, line 308: Any chance we can have
```
quotient∣n {m = m} m∣n = divides m $ begin
  lhs ≡⟨ _∣_.equality m∣n ⟩
  mhs  ≡⟨ *-comm (quotient m∣n) m ⟩
  rhs ∎
```
for added clarity?

src/Data/Nat/Primality/Factorization.agda
```diff
@@ -0,0 +1,125 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Prime factorization of naturla numbers and its properties
```
gallais, line 4:
```suggestion
-- Prime factorization of natural numbers and its properties
```

src/Data/Nat/Primality/Factorization.agda
```diff
@@ -0,0 +1,125 @@
...
+open import Data.Sum.Base
+open import Function.Base
+open import Relation.Nullary.Decidable
+open import Relation.Nullary.Negation
+open import Relation.Binary.PropositionalEquality as ≡
+
+-- Core definition
+------------------------------------------------------------------------
+
+record Factorization (n : ℕ) : Set where
```
gallais, line 32: Grepping for `zation` vs. `sation`, we seem to be using the British spelling more often (although not systematically).
I would be in favour of making it the official spelling.

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,62 @@
...
+-- any number is 2-rough because all primes are greater than or equal to 2
+2-rough-n : ∀ n → 2 Rough n
+2-rough-n n {1} (s≤s ()) 1<2
+2-rough-n n {suc (suc d)} 1<d (s≤s (s≤s ()))
+
+-- if a number is k-rough, and it's not a multiple of k, then it's (k+1)-rough
+extend-∤ : ∀ {k n} → k Rough n → k ∤ n → suc k Rough n
+extend-∤ {k = k} k-rough-n k∤n {suc d} 1<d (s≤s d<k) with suc d ≟ k
+... | yes refl = k∤n
+... | no  d≢k  = k-rough-n 1<d (≤∧≢⇒< d<k d≢k)
```
gallais, line 34: Feels like this pattern could be abstracted over with an auxiliary function in `Data.Nat.Properties`:

```
k<sn-inversion : k < suc n -> k == suc n + k < n
```

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,62 @@
...
+reduce-∣ k-rough-n n∣m d<k d-prime d∣n = k-rough-n d<k d-prime (∣-trans d∣n n∣m)
+
+-- if a number is k-rough, then no number 2 ≤ d < k will divide it, whether prime or not
+rough⇒∤ : ∀ {d k n} → k Rough n → 2 ≤ d → d < k → d ∤ n
+rough⇒∤ {d} {k} {n} k-rough-n = <-rec (λ d′ → 2 ≤ d′ → d′ < k → d′ ∤ n) (rough⇒∤Rec k-rough-n) d
+  where
+    rough⇒∤Rec : ∀ {n k} → k Rough n → ∀ d → <-Rec (λ d′ → 2 ≤ d′ → d′ < k → d′ ∤ n) d → 2 ≤ d → d < k → d ∤ n
+    rough⇒∤Rec {n} {k} k-rough-n (suc (suc d)) rec (s≤s (s≤s z≤n)) d<k with composite? (2 + d)
+    ... | yes (d′ , 2≤d′ , d′<d , d′∣d) = rec d′ d′<d 2≤d′ (<-trans d′<d d<k) ∘ ∣-trans d′∣d
+    ... | no ¬d-composite = k-rough-n {2 + d} (s≤s (s≤s z≤n)) d<k
```
gallais, line 52: Shouldn't this be a trivial consequence of `1 < 2` instead of a proof by strong induction?

gallais (commented):
I haven't looked at the big proof yet.

src/Data/Nat/Divisibility.agda
Taneb: Yes, that's reasonable

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -354,3 +359,17 @@ module _ {ℓ} {R : Rel A ℓ} (R? : Decidable R) where
     (x ∷ xs) ++ y ∷ ys       ≡˘⟨ Lₚ.++-assoc [ x ] xs (y ∷ ys) ⟩
     x ∷ xs ++ y ∷ ys         ∎
     where open PermutationReasoning
+
+------------------------------------------------------------------------
+-- product
+
+productPreserves↭⇒≡ : product Preserves _↭_ ⟶ _≡_
```
MatthewDaggitt, line 366: The standard name is `product-pres-↭-≡` but I think it's safe to call this `product-↭`?

src/Data/Nat/Primality.agda
```diff
@@ -141,3 +141,18 @@ private
...
+
+Prime⇒NonZero : Prime n → NonZero n
+Prime⇒NonZero {suc _} _ = _
+
+Composite⇒NonZero : Composite n → NonZero n
+Composite⇒NonZero {suc _} _ = _
+
+-- If m is a factor of prime p, then it is equal to either 1 or p itself
+∣p⇒≡1∨≡p : ∀ m {p} → Prime p → m ∣ p → m ≡ 1 ⊎ m ≡ p
+∣p⇒≡1∨≡p 0 {p} p-prime m∣p = contradiction (0∣⇒≡0 m∣p) λ p≡0 → subst Prime p≡0 p-prime
```
MatthewDaggitt, line 156: Feels like we want a "prime implies not zero" lemma?

src/Data/Nat/Primality.agda
```diff
@@ -141,3 +141,18 @@ private
...
+∣p⇒≡1∨≡p 1 {p} p-prime m∣p = inj₁ refl
+∣p⇒≡1∨≡p (suc (suc m)) {suc (suc p)} p-prime m∣p = inj₂ (≤∧≮⇒≡ (∣⇒≤ m∣p) λ m<p → p-prime (s≤s (s≤s z≤n)) m<p m∣p)
```
MatthewDaggitt, line 158: Similarly there might be another useful lemma hiding here?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,211 @@
...
+
+      prop : (2 + k) * product (factors res) ≡ 2 + n
+      prop = begin
+        (2 + k) * product (factors res)
+          ≡⟨ cong ((2 + k) *_) (isFactorisation res) ⟩
+        (2 + k) * q
+          ≡⟨ *-comm (2 + k) q ⟩
+        q * (2 + k)
+          ≡˘⟨ _∣_.equality k∣n ⟩
+        2 + n ∎ where open ≡-Reasoning
```
MatthewDaggitt, line 132: Could just open `≤-Reasoning` once and use `begin-equality` here?

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,53 @@
...
+open import Data.Sum.Base using (inj₁; inj₂)
+open import Function.Base using (_∘_; flip)
+open import Relation.Nullary.Decidable.Core using (yes; no)
+open import Relation.Binary.PropositionalEquality.Core using (refl)
+
+-- A number is k-rough if all its prime factors are greater than or equal to k
+_Rough_ : ℕ → ℕ → Set
+k Rough n = ∀ {d} → 1 < d → d < k → d ∤ n
+
+-- any number is 2-rough because all primes are greater than or equal to 2
```
MatthewDaggitt, line 26: Can we have consistent capitalisation of comments? I think a starting capital is good :+1:

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,211 @@
...
+open import Data.Sum.Base using (inj₁; inj₂)
+open import Function.Base using (_$_; _∘_; _|>_; flip)
+open import Relation.Nullary.Decidable using (yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong; module ≡-Reasoning)
+
+------------------------------------------------------------------------
+-- Core definition
+
+record Factorisation (n : ℕ) : Set where
```
MatthewDaggitt, line 33: Can we call this `PrimeFactorisation`, I imagine we'll want non-prime factorisations in future?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,211 @@
...
+        (2 + k) * q
+          ≡⟨ *-comm (2 + k) q ⟩
+        q * (2 + k)
+          ≡˘⟨ _∣_.equality k∣n ⟩
+        2 + n ∎ where open ≡-Reasoning
+
+------------------------------------------------------------------------
+-- Properties of a factorisation
+
+factorisation≥1 : ∀ {as} → All Prime as → product as ≥ 1
```
MatthewDaggitt, line 137: This feels like it should live in `Data.Nat.Primality`? All this is saying is that the product of prime numbers is great than 1. It's not about a factorisation per se.

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,211 @@
...
+  step {as′} as↭p∷as′ = begin
+    a ∷ as      ↭⟨ prep a as↭p∷as′ ⟩
+    a ∷ p ∷ as′ ↭⟨ swap a p refl ⟩
+    p ∷ a ∷ as′ ∎ where open PermutationReasoning
+
+... | inj₁ p∣a with ∣p⇒≡1∨≡p p aPrime p∣a
+...   | inj₁ refl = ⊥-elim pPrime
+...   | inj₂ refl = as , ↭-refl
+
+factorisationUnique′ : (as bs : List ℕ) → product as ≡ product bs → All Prime as → All Prime bs → as ↭ bs
```
MatthewDaggitt, line 160: I feel like this should be private, maybe a sublemma of `factorisationUnique`?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,211 @@
...
+    ; factorsPrime = roughn∧∣n⇒prime k-rough-n 2≤2+n k∣n ∷ factorsPrime res
+    }
+    where
+
+      q : ℕ
+      q = quotient k∣n
+
+      -- we know that k < n, so if q is 0 or 1 then q * k < n
+      2≤q : 2 ≤ q
+      2≤q = ≮⇒≥ (λ q<2 → contradiction (_∣_.equality k∣n) (>⇒≢ (prf (≤-pred q<2)))) where
```
MatthewDaggitt, line 96: Can we stick all the `where`s on the line below rather than the end of the lines? More consistent with the rest of the library.

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,53 @@
...
+2-rough-n n {1} (s≤s ()) 1<2
+2-rough-n n {suc (suc d)} 1<d (s≤s (s≤s ()))
+
+extend-∤ : ∀ {k n} → k Rough n → k ∤ n → suc k Rough n
+extend-∤ k-rough-n k∤n 1<d d<suc[k] with m<1+n⇒m<n∨m≡n d<suc[k]
+... | inj₁ suc[d]≤k = k-rough-n 1<d suc[d]≤k
+... | inj₂ refl = k∤n
+
+-- 1 is always rough
+b-rough-1 : ∀ k → k Rough 1
```
MatthewDaggitt, line 37: Do you mean `k` rather than `b` in the name?

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,53 @@
...
+-- A number is k-rough if all its prime factors are greater than or equal to k
+_Rough_ : ℕ → ℕ → Set
+k Rough n = ∀ {d} → 1 < d → d < k → d ∤ n
+
+-- any number is 2-rough because all primes are greater than or equal to 2
+2-rough-n : ∀ n → 2 Rough n
+2-rough-n n {1} (s≤s ()) 1<2
+2-rough-n n {suc (suc d)} 1<d (s≤s (s≤s ()))
+
+extend-∤ : ∀ {k n} → k Rough n → k ∤ n → suc k Rough n
```
MatthewDaggitt, line 31: Could add a few variables for `k m n`?

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,53 @@
...
+-- 1 is always rough
+b-rough-1 : ∀ k → k Rough 1
+b-rough-1 k {d} 1<d d<k d∣1 with ∣1⇒≡1 d∣1
+b-rough-1 k {.1} (s≤s ()) d<k d∣1 | refl
+
+-- if a number is k-rough, then so are all of its divisors
+reduce-∣ : ∀ {k m n} → k Rough m → n ∣ m → k Rough n
+reduce-∣ k-rough-n n∣m d<k d-prime d∣n = k-rough-n d<k d-prime (∣-trans d∣n n∣m)
+
+-- if a number is k-rough, and k divides it, then it must be prime
```
MatthewDaggitt, line 45: This comment doesn't quite make sense. I think the `it` refers to `n` but it should refer to `k`?

MatthewDaggitt: Likewise the name makes it sound as if `n` will be prime? Is there a better name?

src/Data/Nat/Divisibility.agda
```diff
@@ -300,3 +300,13 @@ m≤n⇒m!∣n! m≤n = help (≤⇒≤′ m≤n)
   help : ∀ {m n} → m ≤′ n → m ! ∣ n !
   help {m} {n}     ≤′-refl        = ∣-refl
   help {m} {suc n} (≤′-step m≤′n) = ∣n⇒∣m*n (suc n) (help m≤′n)
+
+------------------------------------------------------------------------
+-- Properties of quotient
+
+quotient∣n : ∀ {m n} (m∣n : m ∣ n) → quotient m∣n ∣ n
+quotient∣n {m = m} {n = n} m∣n = divides m $ begin-equality
```
MatthewDaggitt, line 308: No idea how we didn't have this one yet. It should definitely go right at the top of the file?

MatthewDaggitt:
Also needs a CHANGELOG entry

src/Data/Nat/Rough.agda
```diff
@@ -0,0 +1,53 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Natural numbers whose prime factors are all above a bound
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Nat.Rough where
```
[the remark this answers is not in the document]
Taneb: I initially had it there, but moved it when I noted that its definition didn't actually require a notion of prime number. I can move it back

src/Data/Nat/Divisibility.agda
jamesmckinna: It is/was issue #1711 ...

src/Data/Nat/Primality.agda
jamesmckinna: `Prime⇒Irreducible`?

src/Data/Nat/Primality.agda
Taneb: I had defined one, I just hadn't used it

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,218 @@
...
+open import Data.Sum.Base using (inj₁; inj₂)
+open import Function.Base using (_$_; _∘_; _|>_; flip)
+open import Relation.Nullary.Decidable using (yes; no)
+open import Relation.Nullary.Negation using (contradiction)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong; module ≡-Reasoning)
+
+------------------------------------------------------------------------
+-- Core definition
+
+record PrimeFactorisation (n : ℕ) : Set where
```
JacquesCarette, line 33: The 'traditional' notion of PrimeFactorisation uses a list of (multiplicity, prime) pairs and has the extra invariant that all primes are distinct.

I am not suggesting you change this definition - but you should probably comment on your design choice, so that this aspect doesn't puzzle a maintainer years down the road.

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,218 @@
...
+  field
+    factors : List ℕ
+    isFactorisation : product factors ≡ n
+    factorsPrime : All Prime factors
+
+open PrimeFactorisation public using (factors)
+open PrimeFactorisation
+
+------------------------------------------------------------------------
+-- Finding a factorisation
```
JacquesCarette, line 43: beautiful comment!

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,218 @@
...
+        2 + n ∎
+
+------------------------------------------------------------------------
+-- Properties of a factorisation
+
+factorisation≥1 : ∀ {as} → All Prime as → product as ≥ 1
+factorisation≥1 {[]} [] = s≤s z≤n
+factorisation≥1 {suc a ∷ as} (pa ∷ asPrime) = *-mono-≤ {1} {1 + a} (s≤s z≤n) (factorisation≥1 asPrime)
+
+factorisationPullToFront : ∀ {as} {p} → Prime p → p ∣ product as → All Prime as → ∃[ as′ ] as ↭ (p ∷ as′)
```
JacquesCarette, line 142: Wouldn't it be 'simpler' if this took a `PrimeFactorisation` as input instead of the pieces it contains?

Also it feels like this could be proven more simply:
1. prove p is a member of the list as
2. then there's a permutation that puts it at the front.

This would be straightforward if you already knew it was a PrimeFactorisation.

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,218 @@
...
+      a ∷ p ∷ as′ ↭⟨ swap a p refl ⟩
+      p ∷ a ∷ as′ ∎
+      where open PermutationReasoning
+
+... | inj₁ p∣a with ∣p⇒≡1∨≡p p aPrime p∣a
+...   | inj₁ refl = ⊥-elim pPrime
+...   | inj₂ refl = as , ↭-refl
+
+factorisationUnique′ : (as bs : List ℕ) → product as ≡ product bs → All Prime as → All Prime bs → as ↭ bs
+factorisationUnique′ [] [] Πas≡Πbs asPrime bsPrime = refl
```
JacquesCarette, line 163: This lemma would simplify a lot if you made your `PrimeFactorisation` a *sorted* list of naturals.

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,53 @@
...
+open import Data.Nat.Induction using (<-rec; <-Rec)
+open import Data.Nat.Primality using (Prime; composite?)
+open import Data.Nat.Properties using (_≟_; <-trans; ≤∧≢⇒<; m<1+n⇒m<n∨m≡n)
+open import Data.Product.Base using (_,_)
+open import Data.Sum.Base using (inj₁; inj₂)
+open import Function.Base using (_∘_; flip)
+open import Relation.Nullary.Decidable.Core using (yes; no)
+open import Relation.Binary.PropositionalEquality.Core using (refl)
+
+-- A number is k-rough if all its prime factors are greater than or equal to k
```
JacquesCarette, line 22: Is this a standard notion? A reference in a comment would be good, if so.

JacquesCarette (commented):
Very nice piece of work.

src/Data/Nat/Primality/Rough.agda
Taneb: It's not something that I made up, at least. The notion has a Wikipedia page, but seems to originate with OEIS contributions. I'm not sure if there's a good reference I can link.

src/Data/Nat/Divisibility.agda
Taneb: I've moved this to the top of the file. I don't want to close #1711 with just this, there are other relevant properties

src/Data/Nat/Primality/Factorisation.agda
Taneb: I've simplified it a bit. I don't think I do want it to take a `PrimeFactorisation`, because it recurses on the list, which would be a little more awkward. But maybe I'm wrong here, I'll think about it more

src/Data/Nat/Primality/Factorisation.agda
Taneb: I didn't want to make the lists sorted, because I wanted to leave things open for other factorisation algorithms, which don't necessarily naturally make a sorted list. And I'm not at all convinced that it would make things simpler enough for it to be worth it

src/Data/Nat/Primality.agda
jamesmckinna: That is, can we please have (somewhere) the definition
```agda
Irreducible p = ∀ {m} → m ∣ p → m ≡ 1 ⊎ m ≡ p
```
and then `∣p⇒≡1∨≡p` gets renamed to `prime⇒irreducible`, possibly even of *type* `Prime ⇒ Irreducible` for `Unary` predicate implication? (there's some explicit/implicit choreography here, but otherwise...)

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,210 @@
...
+-- * a witness that this limit is getting closer to the number of interest, in a
+--   way that helps the termination checker (the k≤n parameter)
+-- * a proof that we can factorise any smaller number, which is useful when we
+--   encounter a factor, as we can then divide by that factor and continue from
+--   there without termination issues
+
+private
+  pattern 2+ n = suc (suc n)
+  pattern 2≤2+n = s≤s (s≤s z≤n)
+  pattern 1<2+n = 2≤2+n
```
jamesmckinna, line 61: Apologies for insisting on having removed such things from earlier versions of the library. Do we need to reinstate them full-bloodedly in `Data.Nat.Base`?

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,57 @@
...
+    k m n : ℕ
+
+-- A number is k-rough if all its prime factors are greater than or equal to k
+_Rough_ : ℕ → ℕ → Set
+k Rough n = ∀ {d} → 1 < d → d < k → d ∤ n
+
+-- any number is 2-rough because all primes are greater than or equal to 2
+2-rough-n : ∀ n → 2 Rough n
+2-rough-n n {1} (s≤s ()) 1<2
+2-rough-n n {suc (suc d)} 1<d (s≤s (s≤s ()))
```
jamesmckinna, line 33: Suggested refactoring, at the cost of a(nother, independently useful?) lemma, and a renaming:
```agda
¬0<n<1 : ∀ {n} → 0 < n → n < 1 → ⊥
¬0<n<1 z<s (s<s ())

2-rough : ∀ n → 2 Rough n
2-rough n (s<s 0<d) (s<s d<1) with () ← ¬0<n<1 0<d d<1
```

src/Data/Nat/Primality.agda
Taneb: I _think_ (please correct me if I'm wrong) that that's the ring-theoretic definition of `Prime`, not `Irreducible`. Renaming `Prime` to `Irreducible` and introducing a corresponding `Prime` is an option but it's not without its costs.

src/Data/Nat/Primality/Rough.agda
Taneb: I'd actually be surprised if there isn't a general form of that lemma in Data.Nat.Properties!

src/Data/Nat/Primality/Factorisation.agda
Taneb: I think these were snuck in by @gallais here so I'm going to refer this to him

src/Data/Nat/Primality/Rough.agda
Taneb: I've made this change, and noticed that I could easily make the first argument implicit while I was at it

jamesmckinna:
This all looks great! 
Two further comments, of different character: 
* the big proof makes several appeals to decomposing proofs/a proof of `m ∣ n` to expose both the quotient `q` and its defining equation. Some of those proofs (and the supporting `where` clauses for all the several lemmas involved) might usefully have those record projections exposed at the beginning, and then shared throughout.., (this is a cosmetic refactoring, but still worth doing, I think)
* can you/does the library, show anywhere that the divisibility relation is well-founded? And with such a proof, would the ambient wf-recursion which drives the big proof be simplified/simplifiable in any way?

Well foundedness of divisibility is independently interesting as part of showing various things, including I think the original proofs of the Fundamental Theorem. Might be interesting to compare!?

src/Data/Nat/Divisibility.agda
```diff
@@ -34,6 +34,16 @@ import Relation.Binary.PropositionalEquality.Properties as PropEq
...
 
+------------------------------------------------------------------------
+-- Properties of quotient
+
+quotient∣n : ∀ {m n} (m∣n : m ∣ n) → quotient m∣n ∣ n
+quotient∣n {m = m} {n = n} m∣n = divides m $ begin-equality
+  n                ≡⟨ _∣_.equality m∣n ⟩
+  quotient m∣n * m ≡⟨ *-comm (quotient m∣n) m ⟩
+  m * quotient m∣n ∎
+  where open ≤-Reasoning
```
jamesmckinna, line 45: Is the following proof not a little simpler?
```agda
quotient∣n {m = m} {n = n} m∣n@(divides q eq) = divides m (begin
  n     ≡⟨ eq ⟩
  q * m ≡⟨ *-comm q m ⟩
  m * q ∎)
  where open ≡-Reasoning
```

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,56 @@
...
+k-rough-1 k {.1} (s≤s ()) d<k d∣1 | refl
+
+-- if a number is k-rough, then so are all of its divisors
+reduce-∣ : k Rough m → n ∣ m → k Rough n
+reduce-∣ k-rough-n n∣m d<k d-prime d∣n = k-rough-n d<k d-prime (∣-trans d∣n n∣m)
+
+-- if a number is k-rough, and k divides it, then k must be prime
+roughn∧∣n⇒prime : k Rough n → 2 ≤ k → k ∣ n → Prime k
+roughn∧∣n⇒prime {k = suc (suc k)} {n = n} k-rough-n (s≤s (s≤s z≤n)) k∣n
+  = <-rec (λ d → 2 ≤ d → d < 2 + k → d ∤ 2 + k) (roughn∧∣n⇒primeRec k-rough-n k∣n) _
```
jamesmckinna, line 51: I suspect that these 'local' predicates which define `Prime` as their universal closure should be lifted out in `Data.Nat.Primality`, so they can be referred to here.

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,56 @@
...
+  where
+  roughn∧∣n⇒primeRec : ∀ {k} → k Rough n → k ∣ n → ∀ d → <-Rec (λ d′ → 2 ≤ d′ → d′ < k → d′ ∤ k) d → 2 ≤ d → d < k → d ∤ k
```
jamesmckinna, line 53: Ditto.

jamesmckinna (commented):
Some leftover review comments from before. Sorry I never finished the review until now!

src/Data/Nat/Primality.agda
jamesmckinna: See #2180

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,56 @@
...
+... | inj₂ refl = k∤n
+
+-- 1 is always rough
+k-rough-1 : ∀ k → k Rough 1
+k-rough-1 k {d} 1<d d<k d∣1 with ∣1⇒≡1 d∣1
+k-rough-1 k {.1} (s≤s ()) d<k d∣1 | refl
+
+-- if a number is k-rough, then so are all of its divisors
+reduce-∣ : k Rough m → n ∣ m → k Rough n
+reduce-∣ k-rough-n n∣m d<k d-prime d∣n = k-rough-n d<k d-prime (∣-trans d∣n n∣m)
```
jamesmckinna, line 46: `k-rough-m` instead of `k-rough-n`?

src/Data/Nat/Primality/Rough.agda
jamesmckinna: But in fact this lemma can be reduced to a simple(r) corollary to your `reduce-∣`... see #2181

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,210 @@
...
+open import Relation.Nullary.Negation using (contradiction)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong; module ≡-Reasoning)
+
+------------------------------------------------------------------------
+-- Core definition
+
+record PrimeFactorisation (n : ℕ) : Set where
+  field
+    factors : List ℕ
+    isFactorisation : product factors ≡ n
```
jamesmckinna, line 39: I think this equality should be the other way round, so that any particular (complex expression denoting) `n` is then transformed LHS to RHS as a product of `Prime` factors. It won't materially affect the development as it stands, but may well be more 'efficient' in deployment?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,210 @@
...
+factorise (2+ n) = <-rec P factoriseRec (2 + n) {2} 2≤2+n (≤⇒≤‴ 2≤2+n) 2-rough
+  where
+
+  P : ℕ → Set
+  P n′ = ∀ {k} → 2 ≤ n′ → k ≤‴ n′ → k Rough n′ → PrimeFactorisation n′
+
+  factoriseRec : ∀ n → <-Rec P n → P n
+  factoriseRec (2+ n) rec (s≤s (s≤s n≤z)) ≤‴-refl k-rough-n = record
+    { factors = 2 + n ∷ []
+    ; isFactorisation = *-identityʳ (2 + n)
```
jamesmckinna, line 78: Well, perhaps I am mistaken here?

jamesmckinna:
Re: the proof in #2181 that every number is `1 Rough`
This permits the main complete induction proof `factorise` to (mostly!) avoid having to relitigate the `0`, `1`, `2+_` pattern analysis in favour of a simpler induction on `NonZero n`, and thereby stripping out a great deal of `1<2+n` to-ing and fro-ing, as follows:
```agda
factorise : ∀ n → .{{NonZero n}} → PrimeFactorisation n
factorise n@(suc _) = <-rec P facRec n (≤⇒≤‴ z<s) 1-rough
  where

  P : ℕ → Set
  P m = .{{NonZero m}} → ∀ {k} → k ≤‴ m → k Rough m → PrimeFactorisation m

  facRec : ∀ n → <-Rec P n → P n
  facRec 1 rec k<1 rough = record
    { factors = []
    ; factorsPrime = []
    ; isFactorisation = refl
    }
  facRec n@(2+ _) rec = h
    where
    h : P n
    h {0} (≤‴-step (≤‴-step k<n)) rough = h k<n 2-rough
    h {1} (≤‴-step k<n)           rough = h k<n 2-rough
    h {k@(2+ _)} ≤‴-refl          rough = record
      { factors = n ∷ []
      ; factorsPrime = prime rough ∷ []
      ; isFactorisation = ≡.sym (*-identityʳ n)
      }
    h {k@(2+ _)} (≤‴-step k<n)    rough with k ∣? n
    ... | no  k∤n = h k<n (∤⇒rough-suc k∤n rough)
    ... | yes k∣n with divides q eq ← k∣n = record
      { factors = k ∷ ps
      ; factorsPrime = rough⇒∣⇒prime 1<k rough k∣n ∷ primes
      ; isFactorisation = n≡k*Πps
      }
      where
      open ≤-Reasoning

      1<k : 1 < k
      1<k = 1<2+n

      n≡k*q : n ≡ k * q
      n≡k*q = begin-equality
        n     ≡⟨ eq ⟩
        q * k ≡⟨ *-comm q k ⟩
        k * q  ∎

      1<q : 1 < q
      1<q = ≰⇒> λ q≤1 → n≮n n $ begin-strict
        n     ≡⟨ eq ⟩
        q * k ≤⟨ *-monoˡ-≤ k q≤1 ⟩
        1 * k ≡⟨ *-identityˡ k ⟩
        k     <⟨ ≤‴⇒≤ k<n ⟩
        n     ∎

      0<q : 0 < q
      0<q = begin-strict
        0 <⟨ z<s ⟩
        1 <⟨ 1<q ⟩
        q ∎

      instance _ = >-nonZero 0<q

      q<n : q < n
      q<n = begin-strict
        q      <⟨ m<m*n q k 1<k ⟩
        q * k ≡⟨ eq ⟨
        n      ∎

      q∣n : q ∣ n
      q∣n = divides k n≡k*q

      k≤q : k ≤ q
      k≤q = ≮⇒≥ λ q<k → rough (boundedComposite 1<q q<k q∣n)

      factorisation[q] : PrimeFactorisation q
      factorisation[q] = rec q<n (≤⇒≤‴ k≤q) (rough⇒∣⇒rough rough q∣n)

      ps = factors factorisation[q]

      primes : All Prime ps
      primes = factorsPrime factorisation[q]

      q≡Πps : q ≡ product ps
      q≡Πps = isFactorisation factorisation[q]

      n≡k*Πps : n ≡ k * product ps
      n≡k*Πps = begin-equality
        n              ≡⟨ n≡k*q ⟩
        k * q          ≡⟨ cong (k *_) q≡Πps ⟩
        k * product ps ∎
```

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,56 @@
...
+  variable
+    k m n : ℕ
+
+-- A number is k-rough if all its prime factors are greater than or equal to k
+_Rough_ : ℕ → ℕ → Set
+k Rough n = ∀ {d} → 1 < d → d < k → d ∤ n
+
+-- any number is 2-rough because all primes are greater than or equal to 2
+2-rough : 2 Rough n
+2-rough 1<d 2>d with () ← ≤⇒≯ 1<d 2>d
```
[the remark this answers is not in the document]
jamesmckinna: See also, for your definition:
```agda
1-rough : 1 Rough n
1-rough 1<d d<1 _ = <-asym 1<d d<1
```

src/Data/Nat/Divisibility.agda
jamesmckinna: See  #2182 .

src/Data/Nat/Divisibility.agda
jamesmckinna: See also: #2182

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -29,11 +30,9 @@ open import Level using (Level)
 open import Relation.Unary using (Pred)
 open import Relation.Binary
 open import Relation.Binary.PropositionalEquality as ≡
-  using (_≡_ ; refl ; cong; cong₂; _≢_; inspect)
+  using (_≡_ ; refl ; cong; cong₂; _≢_; inspect; module ≡-Reasoning)
```
jamesmckinna: Why is `inspect` still on the `import` list?

jamesmckinna:
> The factorization algorithm is trial division. It does not currently short-circuit when it checks a factor greater than the square root of the number being factorized, which would significantly improve its speed in some cases, but makes the code a bit more complicated.

Returning to your initial discussion, how hard is it to include the analysis in terms of only considering trial factors of `n` up to (a suitable approximation to) square root of `n`... esp. in terms of characterisation via `Rough` ness?

jamesmckinna:
Another followup:
* there's a wealth of additional material on theorem proving about primes from 20+ years ago in a [great paper](https://inria.hal.science/inria-00071985v1/file/RR-4600.pdf) from Laurent Thery

which might offer opportunities for interesting downstream extensions to your work. 

And apologies for all the diversions around #2181 and #2182 getting in the way of you getting this merged, but hopefully they help streamline it!

src/Data/Nat/Primality/Rough.agda
```diff
@@ -0,0 +1,56 @@
...
+
+extend-∤ : k Rough n → k ∤ n → suc k Rough n
```
jamesmckinna, line 34: Can this lemma (since #2181 as `Data.Nat.Primality.∤⇒rough-suc`) be sharpened so that for `k > 2` it is possible to count up in increments of `2+ ` rather than merely `suc `?

Taneb:
After a fair amount of refactoring, I've successfully made it short-circut when the candidate factor is greater than the square root of the number being factorised! This makes it _so much quicker_ (my very scientific test of "factorise a decent sized prime from stdin in a compiled program" went from 40s to instant for 433494437)

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,192 @@
...
+  -- Case 1: m * m > n, ∴ Prime n
+  facRec (n , zero) _ rough eq = record
+    { factors = n ∷ []
+    ; isFactorisation = *-identityʳ n
+    ; factorsPrime = rough∧>square⇒prime rough (m∸n≡0⇒m≤n eq) ∷ []
+    }
+  facRec (n@(2+ _) , suc k) (recFactor , recQuotient) {m@(2+ _)} rough eq with m ∣? n
+  -- Case 2: m ∤ n, try larger m, reducing k accordingly
+  ... | no m∤n = recFactor (≤-<-trans (m∸n≤m k (2 * m)) (n<1+n k)) {suc m} (∤⇒rough-suc m∤n rough) $ begin
+    suc n ∸ (suc m + m * suc m)   ≡⟨ cong (λ # → suc n ∸ (suc m + #)) (*-suc m m) ⟩
```
Taneb, line 85: Can this equality reasoning be simplified or made more clear?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,192 @@
...
+  -- Case 3: m ∣ n, record m and recurse on the quotient
+  ... | yes m∣n = record
+    { factors = m ∷ ps
+    ; isFactorisation = m*Πps≡n
+    ; factorsPrime = rough∧∣⇒prime rough m∣n ∷ primes
+    }
+    where
+      m<n : m < n
+      m<n = begin-strict
+        m            <⟨ s≤s (≤-trans (m≤n+m m _) (+-monoʳ-≤ _ (m≤m+n m _))) ⟩
```
Taneb, line 102: Can this step of the inequality reasoning be made more legible?

src/Data/Nat/Primality/Factorisation.agda
jamesmckinna: Yes, probably, and I will look at this in my (eventual) review. Glad that I held off until you made the sqrt breakthrough ;-)

src/Data/Nat/Primality/Factorisation.agda
jamesmckinna: Yes, I'd say *very* probably... but I'll investigate!

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,192 @@
...
+        pred (m * m) <⟨ s<s⁻¹ (m∸n≢0⇒n<m λ eq′ → 0≢1+n (trans (sym eq′) eq)) ⟩
+        n            ∎
+        where open ≤-Reasoning
+      q = quotient m∣n
+      instance _  = n>1⇒nonTrivial (quotient>1 m∣n m<n)
+      factorisation[q] : PrimeFactorisation q
+      factorisation[q] = recQuotient (quotient-< m∣n) (suc q ∸ m * m) (rough∧∣⇒rough rough (quotient-∣ m∣n)) refl
```
jamesmckinna, line 109: I'll also look at the call structure overall, in case this can be simplified too (eg wrt the `suc n ∸ m * m ≡ k` variant relation...)

jamesmckinna (commented):
Fantastic work as an enhancement over the original inefficient trial division process. 
Suggest making the low-level changes as already advertised, and please give me a moment to see if the major call structure can also be improved.
Otherwise, this is great!

src/Data/Nat/Primality.agda
```diff
@@ -294,6 +294,17 @@ prime⇒rough (prime pr) = pr
 rough∧∣⇒prime : .{{NonTrivial p}} → p Rough n → p ∣ n → Prime p
 rough∧∣⇒prime r p∣n = prime (rough∧∣⇒rough r p∣n)
 
+-- If a number n is m-rough, and m * m > n, then n must be prime.
+rough∧>square⇒prime : .{{NonTrivial n}} → m Rough n → m * m > n → Prime n
+rough∧>square⇒prime rough m*m>n = prime ¬composite
```
jamesmckinna, line 299: Re: name of this lemma: `rough∧<square⇒prime`? 
In any case, probably room for a (slightly) improved name?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,191 @@
...
+
+open PrimeFactorisation public using (factors)
+open PrimeFactorisation
+
+------------------------------------------------------------------------
+-- Finding a factorisation
+
+-- this builds up three important things:
+-- * a proof that every number we've gotten to so far has increasingly higher
+--   possible least prime factor, so we don't have to repeat smaller factores
```
jamesmckinna, line 51: Typo: `factors`

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,191 @@
...
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Nat.Primality.Factorisation where
+
+open import Data.Empty using (⊥-elim)
+open import Data.Nat.Base
+open import Data.Nat.Divisibility using (_∣_; _∣?_; quotient; ∣1⇒≡1; divides; quotient-<; m|n⇒n≡m*quotient; hasNonTrivialDivisor; quotient-∣; quotient>1)
+open import Data.Nat.Properties
+open import Data.Nat.Induction using (<-Rec; <-rec; <-recBuilder)
+open import Data.Nat.Primality using (Prime; prime; euclidsLemma; prime⇒irreducible; prime⇒nonZero; _Rough_; 2-rough; ∤⇒rough-suc; rough∧∣⇒prime; ¬prime[1]; rough∧∣⇒rough; rough∧>square⇒prime)
```
jamesmckinna, line 16: I'd either drop the explicit `using` list, or else add `rough⇒≤` to it...

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,191 @@
...
+  facRec : ∀ n×k → (<-Rec ⊗ <-Rec) P n×k → P n×k
+  -- Case 1: m * m > n, ∴ Prime n
+  facRec (n , zero) _ rough eq = record
+    { factors = n ∷ []
+    ; isFactorisation = *-identityʳ n
+    ; factorsPrime = rough∧>square⇒prime rough (m∸n≡0⇒m≤n eq) ∷ []
+    }
+  facRec (n@(2+ _) , suc k) (recFactor , recQuotient) {m@(2+ _)} rough eq with m ∣? n
+  -- Case 2: m ∤ n, try larger m, reducing k accordingly
+  ... | no m∤n = recFactor (≤-<-trans (m∸n≤m k (m + m)) (n<1+n k)) {suc m} (∤⇒rough-suc m∤n rough) $ begin
```
jamesmckinna, line 84: Regarding the arguments to `recFactor`:
* The first argument can be simplified to: `s≤s (m∸n≤m k (2 * m))`; debatable whether it is an improvement, although shorter, because it exposes the `_<_` abtraction, where your proof-term respects it; better, I might think would be to have a lemma ` m∸n<m` in `Data.Nat.Properties` which has `m ∸ n < m` for `.{{NonZero n}}`?
* The second, visible-but-implicit, binding `{suc m}` can be omitted; again debatable whether it improves readability, but it is in any case governed by the appeal to `∤⇒rough-suc m∤n rough`, so the diligent reader could discover this for themselves...

jamesmckinna (changes requested):
Sorry for the additional nitpicky comments, but it think after tidying up the auxiliary lemmas, this should be good to go.

Very nice addition to the library!

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,191 @@
...
+factorisationUnique′ [] (suc (suc b) ∷ bs) Πas≡Πbs asPrime (bPrime ∷ bsPrime) =
+  contradiction Πas≡Πbs (<⇒≢ Πas<Πbs)
+  where
+    Πas<Πbs : product [] < product (2+ b ∷ bs)
+    Πas<Πbs = begin-strict
+      1                    ≡⟨⟩
+      1       * 1          <⟨ *-monoˡ-< 1 {1} {2 + b} 1<2+n ⟩
+      (2 + b) * 1          ≤⟨ *-monoʳ-≤ (2 + b) (factorisation≥1 bsPrime) ⟩
+      2+ b    * product bs ≡⟨⟩
+      product (2+ b ∷ bs)  ∎
```
jamesmckinna, line 145: More DRY/`@`-pattern opportunity here:
```agda
factorisationUnique′ [] (b@(2+ _) ∷ bs) Πas≡Πbs asPrime (bPrime ∷ bsPrime) =
  contradiction Πas≡Πbs (<⇒≢ Πas<Πbs)
  where
    Πas<Πbs : product [] < product (b ∷ bs)
    Πas<Πbs = begin-strict
      1                 ≡⟨⟩
      1 * 1             <⟨ *-monoˡ-< 1 {1} {b} sz<ss ⟩
      b * 1             ≤⟨ *-monoʳ-≤ b (factorisation≥1 bsPrime) ⟩
      b * product bs    ≡⟨⟩
      product (b ∷ bs)  ∎
```

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -0,0 +1,191 @@
...
+
+------------------------------------------------------------------------
+-- Core definition
+
+record PrimeFactorisation (n : ℕ) : Set where
+  field
+    factors : List ℕ
+    isFactorisation : product factors ≡ n
+    factorsPrime : All Prime factors
+
```
jamesmckinna, line 42: Consider also adding some 'smart' constructors/example instances:
```agda
primeFactorisation[1] : PrimeFactorisation 1
primeFactorisation[1] = record
  { factors = []
  ; isFactorisation = refl
  ; factorsPrime = []
  }

primeFactorisation[p] : ∀ {p} → Prime p → PrimeFactorisation p
primeFactorisation[p] {p} pr = record
    { factors = p ∷ []
    ; isFactorisation = sym (*-identityʳ p)
    ; factorsPrime = pr ∷ []
    }
```

src/Data/Nat/Primality.agda
```diff
@@ -294,6 +294,17 @@ prime⇒rough (prime pr) = pr
 rough∧∣⇒prime : .{{NonTrivial p}} → p Rough n → p ∣ n → Prime p
 rough∧∣⇒prime r p∣n = prime (rough∧∣⇒rough r p∣n)
 
+-- If a number n is m-rough, and m * m > n, then n must be prime.
+rough∧square>⇒prime : .{{NonTrivial n}} → m Rough n → m * m > n → Prime n
+rough∧square>⇒prime rough m*m>n = prime ¬composite
+  where
+    ¬composite : ¬ Composite _
+    ¬composite (hasNonTrivialDivisor d<n d∣n) = contradiction (m∣n⇒n≡quotient*m d∣n)
```
jamesmckinna, line 302: `composite` is a (very!) shallow pattern synonym for `hasNonTrivialDivisor` here, but I think it's helpful to have it as a constructor of `Composite`, just as `prime` is the constructor of `Prime`...

jamesmckinna:
@Taneb you wrote on #2300 

> The reason I asked about commutative structures is that the original reason I wanted `foldMap` was because it felt like the right way to say that `sum` respects permutation - I've defined something similar in #1969 for `product` [here](https://github.com/agda/agda-stdlib/blob/578718d5d930e6976d3ac192300ac24fd3acdbcf/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda#L366)

Following up on this here... why not factor this proof as one for a commutative monoid operation, and then instantiate to `product` being commutative?

MatthewDaggitt:
Addressed some of the feedback. @Taneb if you're willing to write something brief about @JacquesCarette's comment on the unique multiplicity representation I'm happy to merge this in.

MatthewDaggitt:
@jamesmckinna any chance you could turn your thumbs up into an "approve"? :smile:

jamesmckinna (approved):

jamesmckinna:
> @jamesmckinna any chance you could turn your thumbs up into an "approve"? 😄

Apologies for holding things up through inattention and too much nit-picking! 

Now DONE, and hitting 'merge'...

PR #1975, opened 2023-06-12
Add a vector type with logarithmic-time lookup

Taneb:
I haven't benchmarked this yet. I also want to add something corresponding to the standard `Vec`'s `_[_]=_`, and properties relating `head`, `tail`, and `lookup`.

### DIFF
--- a/src/Data/Fin/Binary/Base.agda
+++ b/src/Data/Fin/Binary/Base.agda
@@ -1,0 +1,23 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- A binary representation of finite sets
+------------------------------------------------------------------------
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Data.Fin.Binary.Base where
+
+open import Data.Nat.Binary.Base
+
+private
+  variable
+    n : ℕᵇ
+
+data Finᵇ : ℕᵇ → Set where
+  zeroᵒ   : Finᵇ 1+[2 n ]
+  zeroᵉ   : Finᵇ 2[1+ n ]
+  oneᵉ    : Finᵇ 2[1+ n ]
+  1+[2_]ᵒ : Finᵇ n → Finᵇ 1+[2 n ]
+  2[1+_]ᵒ : Finᵇ n → Finᵇ 1+[2 n ]
+  2[1+_]ᵉ : Finᵇ n → Finᵇ 2[1+ n ]
+  3+[2_]ᵒ : Finᵇ n → Finᵇ 2[1+ n ]
--- a/src/Data/Fin/Binary/Base.agda
+++ b/src/Data/Fin/Binary/Base.agda
@@ -20,4 +20,4 @@ data Finᵇ : ℕᵇ → Set where
   1+[2_]ᵒ : Finᵇ n → Finᵇ 1+[2 n ]
   2[1+_]ᵒ : Finᵇ n → Finᵇ 1+[2 n ]
   2[1+_]ᵉ : Finᵇ n → Finᵇ 2[1+ n ]
-  3+[2_]ᵒ : Finᵇ n → Finᵇ 2[1+ n ]
+  3+[2_]ᵉ : Finᵇ n → Finᵇ 2[1+ n ]
--- a/src/Data/Fin/Binary/Base.agda
+++ b/src/Data/Fin/Binary/Base.agda
@@ -7,7 +7,7 @@
 
 module Data.Fin.Binary.Base where
 
-open import Data.Nat.Binary.Base
+open import Data.Nat.Binary.Base as ℕᵇ hiding (zero; suc)
 
 private
   variable
@@ -21,3 +21,17 @@ data Finᵇ : ℕᵇ → Set where
   2[1+_]ᵒ : Finᵇ n → Finᵇ 1+[2 n ]
   2[1+_]ᵉ : Finᵇ n → Finᵇ 2[1+ n ]
   3+[2_]ᵉ : Finᵇ n → Finᵇ 2[1+ n ]
+
+zero : Finᵇ (ℕᵇ.suc n)
+zero {n = ℕᵇ.zero} = zeroᵒ
+zero {n = 2[1+ n ]} = zeroᵒ
+zero {n = 1+[2 n ]} = zeroᵉ
+
+suc : Finᵇ n → Finᵇ (ℕᵇ.suc n)
+suc zeroᵒ = oneᵉ
+suc zeroᵉ = 1+[2 zero ]ᵒ
+suc oneᵉ = 2[1+ zero ]ᵒ
+suc 1+[2 i ]ᵒ = 2[1+ i ]ᵉ
+suc 2[1+ i ]ᵒ = 3+[2 i ]ᵉ
+suc 2[1+ i ]ᵉ = 1+[2 suc i ]ᵒ
+suc 3+[2 i ]ᵉ = 2[1+ suc i ]ᵒ
--- a/src/Data/Vec/Binary/Base.agda
+++ b/src/Data/Vec/Binary/Base.agda
@@ -8,6 +8,7 @@
 
 module Data.Vec.Binary.Base where
 
+open import Data.Fin.Binary.Base
 open import Data.Nat.Binary.Base
 open import Level using (Level)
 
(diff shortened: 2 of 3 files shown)


src/Data/Fin/Binary/Base.agda
```diff
@@ -0,0 +1,23 @@
...
+
+module Data.Fin.Binary.Base where
+
+open import Data.Nat.Binary.Base
+
+private
+  variable
+    n : ℕᵇ
+
+data Finᵇ : ℕᵇ → Set where
```
JacquesCarette, line 16: Can you add some comments regarding how we're supposed to interpret these? All the interpretations that I can come up with make me want to report problems... so I am quite sure someone is misunderstanding something!

src/Data/Vec/Binary/Base.agda
```diff
@@ -0,0 +1,95 @@
...
+-- Core type
+------------------------------------------------------------------------
+
+data Vecᵇ (A : Set a) : ℕᵇ → Set a where
+  -- The empty vector
+  []        :                               Vecᵇ A zero
+  -- A vector with an odd number of elements
+  _∷⟨_/_⟩   : A     → Vecᵇ A n → Vecᵇ A n → Vecᵇ A 1+[2 n ]
+  -- A vector with a non-zero even number of elements
+  _×_∷⟨_/_⟩ : A → A → Vecᵇ A n → Vecᵇ A n → Vecᵇ A 2[1+ n ]
```
JacquesCarette, line 32: I would be tempted to use two `Vecᵇ A (suc n)` instead?

JacquesCarette (commented):
You should probably mark this PR as draft?

src/Data/Vec/Binary/Base.agda
Taneb: That feels like it would complicate a lot of definitions for not much improvement. Why does it tempt you?

JacquesCarette: Probably because I don't have a global view of the design and this thought was from looking just at the definition here on its own.

So this probably calls for (more) comments rather than an actual change.

Taneb: It's difficult for me to explain when I don't know what's not being understood! 

One thing that might be making this more confusing is that `Vecᵇ` is indexed by a `ℕᵇ`, not a more familiar `ℕ`. This means that `suc` is a function, not a constructor, and `suc n` isn't structurally smaller than `2[1+ n ]`.

src/Data/Fin/Binary/Base.agda
Taneb: It feels pretty easy to interpret to me, but that's me suffering the curse of knowledge, I can't put myself into the position of someone who doesn't understand this construction. I can't think of a way to explain it that doesn't feel like an unhelpful, patronizing, spelling out of the constructors, apart from maybe "`Finᵇ n` corresponds directly to a path to an element of a `Vecᵇ n`."

I'd be interested to hear the interpretations you came up with, and their problems

src/Data/Vec/Binary/Base.agda
JacquesCarette: I think you are correct that this indexing is definitely part of the reason I'm not quite understanding. Maybe if more of the semantics of things were explained, it would all make more sense.

src/Data/Fin/Binary/Base.agda
JacquesCarette: That 'maybe' explanation is a good start. I guess I just don't understand `ℕᵇ` well enough (its structure, the usefulness of that structure, etc). I think `ℕᵇ` is sufficiently "unknown" by most that I doubt expanding out the point of the definition won't be patronizing to most.

src/Data/Vec/Binary/Base.agda
```diff
@@ -0,0 +1,104 @@
...
+private
+  variable
+    a b c : Level
+    A : Set a
+    B : Set b
+    C : Set c
+    n : ℕᵇ
+
+-- Core type
+------------------------------------------------------------------------
```
MatthewDaggitt, line 25: The long lines usually come before headings.

src/Data/Vec/Binary/Base.agda
```diff
@@ -0,0 +1,104 @@
...
+
+data Vecᵇ (A : Set a) : ℕᵇ → Set a where
+  -- The empty vector
+  []        :                               Vecᵇ A zero
```
MatthewDaggitt, line 29: We usually avoid excessive alignment like this. It's a bit difficult to maintain in general.

src/Data/Vec/Binary/Base.agda
```diff
@@ -0,0 +1,104 @@
...
+  -- A vector with an odd number of elements
+  _∷⟨_/_⟩   : A     → Vecᵇ A n → Vecᵇ A n → Vecᵇ A 1+[2 n ]
+  -- A vector with a non-zero even number of elements
+  _×_∷⟨_/_⟩ : A → A → Vecᵇ A n → Vecᵇ A n → Vecᵇ A 2[1+ n ]
```
MatthewDaggitt, line 33: The `×` in `_×_∷⟨_/_⟩` looks a bit weird to me, because `×` is usually used at the type level. I would argue for `_∷_∷⟨_/_⟩` instead?

src/Data/Vec/Binary/Base.agda
```diff
@@ -0,0 +1,104 @@
...
+    a b c : Level
+    A : Set a
+    B : Set b
+    C : Set c
+    n : ℕᵇ
+
+-- Core type
+------------------------------------------------------------------------
+
+data Vecᵇ (A : Set a) : ℕᵇ → Set a where
```
MatthewDaggitt, line 27: To me this looks more like some sort of balanced binary tree in disguise rather than a vector encoding? I'm a bit reluctant to add it under the `Vec` hierarchy...

MatthewDaggitt (commented):
If the reason for adding this is performance, then we really do need to add some benchmarking tests.

src/Data/Vec/Binary/Base.agda
Taneb: It's a vector implemented as a binary tree. It's based, roughly, on zeroless binary random access lists from section 9.2.2 of Okasaki's purely functional data structures. It has operations for manipulating the head of the vector, and none for inserting or deleting from arbitrary positions (those are hard to write). It feels a lot more vector-like than tree-like to me, in terms of interface. Is there anything I could define that would convince you it's truly a vector?

MatthewDaggitt: Sorry that was poorly phrased on my part. It's not that I don't believe there's an interface that is vector-like (although the way you describe it with only being able to manipulate the head, it sounds a little bit more like a list). It's that maybe there's a (slightly more general?) tree data-structure that is lurking underneath here that we should be able to extract out?

Taneb: I really don't think there is a useful tree structure we can extract here! The shape of the internal tree is guided carefully by the `ℕᵇ` index so that there's only one possible shape for each size.

I think you may have slightly misunderstood me when I said that we can't insert or delete from arbitrary positions. We can in fact efficiently perform arbitrary lookups (that the whole point of this type, really), and we can do `updateAt` quickly.

jamesmckinna:
Only just looking at this now. Three quick remarks: 
* all for having more Okasaki data structures in the library;
* I confess I found the development hard going in spite of the preceding discussion; 

* I couldn't help but think that for some people it might be easier were it to be restructured, or at least reexpressed with the additional observations, as a series of *view*s: 
-`ℕᵇ` as a view of `ℕ`;
- `Finᵇ` as a view of `Fin`;
- `Vecᵇ` as a view of `Vec`;
where in each case, the views from the preceding layer are being correlated in the types of subsequent views...


... so that what you end up with is a collection of functions for efficient decomposition of `Vec`, going via the `Vecᵇ` view...


That might make the representation overly coupled, but at least establishing the correspondence with the underlying 'unary' semantics might help a lot in understanding what's going on...?

MatthewDaggitt:
There hasn't been any movement on this for 9 months, in terms of evaluating whether the performance of this is any better than the standard definition. There were also some discussion in the library developers' meeting of whether this belonged better in a separate library. 

For these reasons, I'm going to close this PR. However, @Taneb feel free to reopen if you believe it should live in the standard library and you have some data to suggest it does indeed perform better.

PR #1976, opened 2023-06-12
add take-[] and drop-[] to stdl Data/List/Properties.agda

Sofia-Insa:
Add lemmas `take-[]` and `drop-[]`  to std library. In Data.List.Properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2784,23 +2784,12 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-
-* Added 'take-[]' to 'Data.List.Proppreties'
-  ```agda
-  -- Taking and dropping from an empty list do nothing.
+  
+ * Added new functions in `Data.List.Proppreties`:
+  ```
   take-[] : ∀ m → take {A = A} m [] ≡ []
-  take-[] ℕ.zero = refl
-  take-[] (suc m) = refl
-  ```
-
-* Added 'drop-[]' to 'Data.List.Proppreties'
-  ```agda
-  -- Taking and dropping from an empty list do nothing.
   drop-[] : ∀ m → drop {A = A} m [] ≡ []
-  drop-[] ℕ.zero = refl
-  drop-[] (suc m) = refl
   ```
- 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2784,6 +2784,12 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
+  
+ * Added new lemmas in `Data.List.Proppreties`:
+  ```
+  take-[] : ∀ m → take {A = A} m [] ≡ []
```
JacquesCarette, line 2790: Don't need the `{A = A}` in the CHANGELOG.

JacquesCarette (changes requested):

JacquesCarette (approved):

src/Data/List/Properties.agda
```diff
@@ -758,6 +758,11 @@ length-take zero    xs       = refl
 length-take (suc n) []       = refl
 length-take (suc n) (x ∷ xs) = cong suc (length-take n xs)
 
+-- Taking and dropping from an empty list do nothing.
```
MatthewDaggitt, line 761: This comment should only reference `taking`, likewise the one below only `dropping`.

src/Data/List/Properties.agda
```diff
@@ -758,6 +758,11 @@ length-take zero    xs       = refl
...
+take-[] : ∀ m → take {A = A} m [] ≡ []
+take-[] ℕ.zero = refl
```
MatthewDaggitt, line 763: Can get rid of the `ℕ.` here and below as well.

MatthewDaggitt (commented):
Otherwise looks great! Thanks for the first PR.

agda-stdlib
gallais, line 1: ugh?

Sofia-Insa: Sorry, I got mixed up when I tried to do a new PR changes end up here. Didn't manage to delete a commit so I commit once more to delete the changes. 
There is only the add of take-[] and drop-[] in Data.List.Propreties and change in CHANGELOG.md that should appear now

MatthewDaggitt:
Err so it looks like now you've added the library as a git submodule of itself and then made the changes  in the submodule rather than the main library?!

PR #1981, opened 2023-06-13
Add lemmas take drop take-drop to Data.List.Propreties

Sofia-Insa:
Add 3 lemmas to the sdtl Data.List.Propreties 

take-[]
drop-[] 
were already in PR #1976 but they are needed for the new lemma 
take-drop-[]

The utility of take-drop-[] is up to debate ( it is a very simple combination of the previous two )


MatthewDaggitt:
In general I think we would consider `take-drop-[]` superfluous. As you say it can easily be derived, and if we started adding properties of this form, the list would quickly grow unmanageable. I'm therefore going to close this in favour of #1976.

PR #1983, opened 2023-06-14
added new function `take-all` to stdl 

Sofia-Insa:
Added new function take all to stdl Data.List.Properties and declare the change in CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2784,7 +2784,11 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-
+  
+* Added new function to `Data.List.Properties`
+  ```agda
+  take-all : {a : Level} {A : Set a} (k : ℕ) (l : List A) → (k ≥ length l) → take k l ≡ l
+  ```
 NonZero/Positive/Negative changes
 ---------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2053,6 +2053,8 @@ Other minor changes
 
   length-isMagmaHomomorphism : (A : Set a) → IsMagmaHomomorphism (++-rawMagma A) +-rawMagma length
   length-isMonoidHomomorphism : (A : Set a) → IsMonoidHomomorphism (++-[]-rawMonoid A) +-0-rawMonoid length
+  
+  take-all : n ≥ length l → take n xs ≡ xs     
   ```
 
 * Added new patterns and definitions to `Data.Nat.Base`:
@@ -2784,11 +2786,7 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-  
-* Added new function to `Data.List.Properties`
-  ```agda
-  take-all : {a : Level} {A : Set a} (k : ℕ) (l : List A) → (k ≥ length l) → take k l ≡ l
-  ```
+
 NonZero/Positive/Negative changes
 ---------------------------------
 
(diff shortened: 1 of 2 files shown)


src/Data/List/Properties.agda
```diff
@@ -758,6 +758,13 @@ length-take zero    xs       = refl
 length-take (suc n) []       = refl
 length-take (suc n) (x ∷ xs) = cong suc (length-take n xs)
 
+-- If you take at least as many elements from a list as it has, you get the whole list.
+take-all : {a : Level} {A : Set a} (k : ℕ) (l : List A) → (k ≥ length l) → take k l ≡ l
```
MatthewDaggitt, line 762: So we've already declared `a` and `A` as variables at the top of the file so we don't need to include them here. Also if you check the `Naming conventions` of the [style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md), you'll see that we usually use `xs` for list variables and `n` for natural numbers. (It would be a good idea to read the whole style guide anyway!).

src/Data/List/Properties.agda
```diff
@@ -758,6 +758,13 @@ length-take zero    xs       = refl
...
+take-all ℕ.zero [] _ = refl
```
MatthewDaggitt, line 763: `ℕ.` not needed again

src/Data/List/Properties.agda
MatthewDaggitt: Also don't need brackets around `(k ≥ length l)`

CHANGELOG.md
```diff
@@ -2784,7 +2784,11 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-
+  
+* Added new function to `Data.List.Properties`
```
MatthewDaggitt, line 2788: Could you add this under the existing entry for `Data.List.Properties`?

MatthewDaggitt (commented):
Thanks! A few comments.

CHANGELOG.md
```diff
@@ -2784,7 +2784,11 @@ Other minor changes
...
+  ```agda
+  take-all : {a : Level} {A : Set a} (k : ℕ) (l : List A) → (k ≥ length l) → take k l ≡ l
```
MatthewDaggitt, line 2790: Also for the CHANGELOG entries we typically only provide the "useful" part of the type. It's a little bit subjective, but generally variable declarations aren't necessary. So for example this one would become 
```
n ≥ length xs → take n xs ≡ xs`
```

jamesmckinna:
Hmmm... as an alternative, consider the following 'Mathematics Made Difficult' style argument: for fixed `l`
* for any `n`, `take n l` is a `Prefix` of `l`
* moreover, this is *functorial* (monotone) from the usual ordering on `n` to the `Prefix` extension/inclusion ordering (maybe this is more fiddly to prove than it looks)

Each of these seems a worthwhile property to have in the library?

Now, by entirely analogous, but simpler (the cases are in lockstep) reasoning to the lemma here
* `take (length l) l = l`

(Ditto. Though it obviously follows from the lemma proved in this PR)

Conclude, via the partial order structure of `Prefix`?

CHANGELOG.md
```diff
@@ -2053,6 +2053,8 @@ Other minor changes
 
   length-isMagmaHomomorphism : (A : Set a) → IsMagmaHomomorphism (++-rawMagma A) +-rawMagma length
   length-isMonoidHomomorphism : (A : Set a) → IsMonoidHomomorphism (++-[]-rawMonoid A) +-0-rawMonoid length
+  
+  take-all : n ≥ length l → take n xs ≡ xs
```
jamesmckinna, line 2057: Not `l` here, but `xs`!?

JacquesCarette:
So I had a bunch of lemmas (still do) using the ordering structure. But because of bugs in Agda (Cubical leaking) and the weird reduction properties of `fromN<`, I gave up on that path. It sure makes sense, but since I was trying to get specific things done, it wasn't worth pushing through the pain.

PR #1984, opened 2023-06-14
New functions `take-drop-1` and `take-one-more` added to Data.List.Properties

Sofia-Insa:
functions 
   take-drop-1 
   take-one-more 
added to Data.List.Properties and declared in CHANGELOG.md

Creation of a section take-drop in Data.List.Properties. 
Moved  `take++drop` from drop section to take-drop section.  

The name of the two new functions are up to debate


MatthewDaggitt (commented):
Please see comments in https://github.com/agda/agda-stdlib/pull/1983 for similar things that need addressing.

src/Data/List/Properties.agda
```diff
@@ -758,6 +758,10 @@ length-take zero    xs       = refl
 length-take (suc n) []       = refl
 length-take (suc n) (x ∷ xs) = cong suc (length-take n xs)
 
+take-one-more : {ℓ : Level} {X : Set ℓ} {m : ℕ} (x : Fin m) (f : Fin m → X) →
+   take (toℕ x) (tabulate f) ∷ʳ f x ≡ take (suc (toℕ x)) (tabulate f)
```
MatthewDaggitt, line 762: I think this equality should the other way around. We usually have the function application on the left and the reduction behaviour on the right.

src/Data/List/Properties.agda
```diff
@@ -766,11 +770,20 @@ length-drop zero    xs       = refl
 length-drop (suc n) []       = refl
 length-drop (suc n) (x ∷ xs) = length-drop n xs
 
+------------------------------------------------------------------------
+-- take-drop
```
MatthewDaggitt, line 774: No need for this separate section. It's fine to have them under `drop` as it comes after the `take` section.

src/Data/List/Properties.agda
```diff
@@ -758,6 +758,10 @@ length-take zero    xs       = refl
 length-take (suc n) []       = refl
 length-take (suc n) (x ∷ xs) = cong suc (length-take n xs)
 
+take-one-more : {ℓ : Level} {X : Set ℓ} {m : ℕ} (x : Fin m) (f : Fin m → X) →
```
MatthewDaggitt, line 761: This should definitely have `tabulate` in the name somewhere. How about `take-suc-tabulate`?

src/Data/List/Properties.agda
```diff
@@ -766,11 +770,20 @@ length-drop zero    xs       = refl
...
 
+------------------------------------------------------------------------
+-- take-drop
+
 take++drop : ∀ n (xs : List A) → take n xs ++ drop n xs ≡ xs
 take++drop zero    xs       = refl
 take++drop (suc n) []       = refl
 take++drop (suc n) (x ∷ xs) = cong (x ∷_) (take++drop n xs)
 
+take-drop-1 : {x : Level} {X : Set x} {k : ℕ} (f : Fin k → X) (n : Fin k) →
```
MatthewDaggitt, line 781: Again should have `tabulate` in the name somewhere. How about `drop-take-tabulate-1`?

jamesmckinna:
Not sure that the `tabulate` really belongs in the lemma statement. What you seem to have us a lemma which combines:
* interaction of `take` and `suc` on the one hand;
* semantics of `tabulate` on the other.

Suggest splitting out the first one. That may also uncouple the dependency on `Fin` (which would otherwise be needed for `tabulate`)?

JacquesCarette:
So I had tried the lemmas without `tabulate` and succeeded. However, the results were basically unusable.

For example, with `take-one-more`, the natural statement involves `lookup` and then you get stuff that involves `length l` in the type and it ends up being just horrible. Trying to then prove the combined lemma with `tabulate` (which is the one I really needed) got me deep into `subst` hell. [I think that `lookup`'s type involves "green slime" if I've got that concept right.]

So yes, both of these should have `tabulate` in their name.  Also, I don't know how to factor them into *useful* pieces, even though there indeed **are** theorems where they are factored.

src/Data/List/Properties.agda
jamesmckinna: The meta-heuristic being
```
complex => simple(r)
```

jamesmckinna:
Hmm. @JacquesCarette i would look at these proofs... but have no spare time to allocate right now. My first thought (don't be the first to groan) is: clearly try to avoid `subst` hell, but if you have green slime in a type, ... then abstract over it!?

This is what `with`/views are actually for. 

So I'm sorry not to have the time, if I could then try to make this a good didactic example on non-trivial with programming...

Separately: I further can't help thinking that proofs of lemmas about `tabulate` on *lists* should 'really' go via the corresponding statements on `Data.Vec.Functional`, where the `Fin n -> A` abstraction is the basic notion...?

jamesmckinna:
OK... I may have to revise my comments above (but I still need to think some more). But in the meantime:
* `cast` hell is definitely not as bad as `subst` hell, and we only need `cast` here...
* ... thus enabling me to refactor the proof of `take-one-more` as follows
```agda
take-suc : (xs : List A) (i : Fin (length xs)) → let m = toℕ i in
           take (suc m) xs ≡ take m xs ∷ʳ lookup xs i
take-suc (x ∷ xs) zero    = refl
take-suc (x ∷ xs) (suc i) = cong (x ∷_) (take-suc xs i)

take-suc-tabulate : ∀ {n} (i : Fin n) (f : Fin n → A) → let m = toℕ i in
                    take (suc m) (tabulate f) ≡ take m (tabulate f) ∷ʳ f i
take-suc-tabulate i f
  rewrite sym (toℕ-cast (sym (length-tabulate f)) i) = begin
  take (suc (toℕ (cast _ i))) (tabulate f)
    ≡⟨ take-suc (tabulate f) (cast _ i) ⟩
  take (toℕ (cast _ i)) (tabulate f) ∷ʳ lookup (tabulate f) (cast _ i)
    ≡⟨ cong (take (toℕ (cast _ i)) (tabulate f) ∷ʳ_) (lookup-tabulate f i) ⟩
  take (toℕ (cast _ i)) (tabulate f) ∷ʳ f i ∎
  where open P.≡-Reasoning
```

Conclusion (partisan/partial, at least on my side): the refactoring is worth doing, and not as bad as you claim.

jamesmckinna:
Likewise:
```agda
drop-take-suc : (xs : List A) (i : Fin (length xs)) → let m = toℕ i in
           drop m (take (suc m) xs) ≡ [ lookup xs i ]
drop-take-suc (x ∷ xs) zero    = refl
drop-take-suc (x ∷ xs) (suc i) = drop-take-suc xs i

drop-take-suc-tabulate : ∀ {n} (f : Fin n → A) (i : Fin n) → let m = toℕ i in
                  drop m (take (suc m) (tabulate f)) ≡ [ f i ]
drop-take-suc-tabulate f i rewrite sym (toℕ-cast (sym (length-tabulate f)) i) | sym (lookup-tabulate f i)
  = drop-take-suc (tabulate f) (cast _ i)
```

MatthewDaggitt:
Apart from that, and the merge conflicts, this now looks good to go. Thanks for the help @jamesmckinna!

jamesmckinna:
Of course, the multiple appeals to `sym` during the various `rewrite`s are a very ugly wart (because they adhere to the `complex => simple(r)` design metaheuristic... sigh).

MatthewDaggitt (commented):
Sorry forgot to post his review. Once this is addressed, and the merge conflicts are resolved, I'll merge this in :+1:

src/Data/List/Properties.agda
```diff
@@ -758,6 +759,15 @@ length-take zero    xs       = refl
 length-take (suc n) []       = refl
 length-take (suc n) (x ∷ xs) = cong suc (length-take n xs)
 
+take-suc : (xs : List A) (o : Fin (length xs)) → let m = toℕ o in
```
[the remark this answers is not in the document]
jamesmckinna: Good pickup! I'd meant to make the same observation myself (having done it in my own versions... ;-))

PR #1985, opened 2023-06-14
Fixities for `Data` and `Codata`

Saransh-cpp:
Related to #207 

Added fixities for mixfir operators in `Codata` and `Data`.


MatthewDaggitt:
Thanks, had a look through them and they look reasonable to me. I might leave this PR for a few days or so so that other people can have a look at them and comment, as I'm not a big user of fixities myself.

PR #1994, opened 2023-06-16
Get rid of redundant deps in `System.Console.ANSI`

### LINKED ISSUE #1993: Dependencies of `System.Console.ANSI`
When loading `System.Console.ANSI` we're somehow typechecking
the following modules which seems ridiculous:

```agda
Function.HalfAdjointEquivalence
Data.Product.Function.Dependent.Propositional
```

Saransh-cpp:
Closes #1993

See https://github.com/agda/agda-stdlib/issues/1993#issuecomment-1595352224 and https://github.com/agda/agda-stdlib/issues/1993#issuecomment-1595356776

- redundant deps : `Function.HalfAdjointEquivalence`, `Data.Product.Function.Dependent.Propositional`
- replaced `Data.String` import with `Data.String.Base` in `Data.Nat.Show`

### DIFF
--- a/src/Algebra/Definitions.agda
+++ b/src/Algebra/Definitions.agda
@@ -109,7 +109,7 @@ _DistributesOver_ : Op₂ A → Op₂ A → Set _
 * DistributesOver + = (* DistributesOverˡ +) × (* DistributesOverʳ +)
 
 _MiddleFourExchange_ : Op₂ A → Op₂ A → Set _
-_*_ MiddleFourExchange _+_ = 
+_*_ MiddleFourExchange _+_ =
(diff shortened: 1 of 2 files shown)


jamesmckinna:
What about the import of `True` in `Data.Nat.Show` from `Relation.Nullary.Decidable`...
... should this be narrowed to `Relation.Nullary.Decidable.Core`?

(A genuine question for the library experts as well: how narrow should we make things on import?)

MatthewDaggitt:
Generally if you're in a module that focuses on definitions, then usually you want to be strict in your imports. If you're in a module that focuses on proofs then there's often not much to be gained by being stricter.  The exception to the former is if  your data types include proofs within them.

MatthewDaggitt:
Thanks @Saransh-cpp!

Issue #1998, opened 2023-06-20
Irrelevant instance arguments for `_<_`


jamesmckinna:
So: (DRAFT)
* issue #1988 
* the discussion around #1951 
* ditto. regarding #1709 and friends
* @MatthewDaggitt 's heroic work on #1581 #1538 etc.
* etc.

lead me to conclude that we should add (!?) an irrelevant-instance version of `LessThan` to `Data.Nat.Base`, along the analogy with `NonZero`, and forever after use this relation for 'computational' appeals to `_<_`, such as in the (old/refactored; *mea culpa*) definition of `fromN<`...

... the only question being whether this can already be achieved by back-porting `Positive` etc. to `Data.Nat`, as any instance of `LessThan m n` (sic) defined in terms of `_<ᵇ_`would eventually reduce to one of `Positive k` for some `k`? or else defining `Positive` in terms of `LessThan`?

Taneb:
It's possible to define `LessThan` in terms of `NonZero` and `_∸_`, as well, which avoids having to implement new instances

jamesmckinna:
Ah!? ...

JacquesCarette:
I found that working with things that were in terms of `_<ᵇ_` to be quite difficult to work with, as they made it super unclear what I needed to 'nudge' to get things to reduce further. Basically because they had already reduced to things that were so incredibly far from the original "question" (involving `fromN<`) that I just gave up.

For my examples, I suspect that a version using `_∸_` would have made 'semantic sense' to me.

In the end, I refactored absolutely everything to not use `_<_` at all. [I was formalizing a notion of 'interval' in an `Enumerable` set. There are *so so* many potential design choices, it's quite ridiculous.]

jamesmckinna:
Hmmmph, yes, I share your frustrations. If we can prove (we do!) that vanilla `_<_` is `Irrelevant`, then why should we worry about whether arguments of that form are used strictly or non-strictly? But it seems we must, because back-end erasure / run-time efficiency is not the only consideration, we want also to improve compile-time type-checking efficiency. And it would seem that there are regression tests out there to detect my introduction of overly aggressive/strict approach to matching on such proofs. 

And so the further I go with #2000 , the more I end up thinking that it can all be reduced to having moved the former definitions in now-deprecated `Data.Nat.Properties.Core` (introduced to break a dependency cycle) back into `Data.Nat.Base`. And reverting the former changes to `Data.Fin.*`... sigh. 

[...]

Issue #1988, opened 2023-06-15
Possible Regression at fromℕ<


dMaggot:
The following module works in `v1.7.2` but does not work in `master`:

```agda
module bug where

open import Data.Maybe using (just)
open import Data.Fin using (fromℕ<)
open import Data.List using (List; lookup; length; head; _∷_; [])
open import Data.Nat using (_<_; ℕ)
open import Data.Nat.Properties using (n<1⇒n≡0; ≤-reflexive)
open import Relation.Binary.PropositionalEquality using (_≡_; refl; sym)

bug : ∀ {l : List ℕ} → (len≡1 : length l ≡ 1) → lookup l (fromℕ< (≤-reflexive (sym len≡1))) < (length l) → head l ≡ just 0
bug {x ∷ []} len≡1 headBounded rewrite n<1⇒n≡0 headBounded = refl
```
According to `git bisect ` the culprit is 82c1b9df0769d3df4bbb33e46d535e07c8e12e60 so I suppose the problem is `fromℕ<`.

gallais:
Can you please provide a more detailed report than "does not work"?

dMaggot:
The error message is

```
bug.agda:11,62-66
x != ℕ.zero of type ℕ
when checking that the expression refl has type just x ≡ just 0
```
but I do not think that is very revealing without familiarizing oneself with the example.

Here `≤-reflexive (sym len≡1)` is simply `0 < length l` so `fromℕ<` should give you `0` and `headBounded` should be `lookup (x ∷ []) 0 < (length (x ∷ []))` and the rewriting should be on `x ≡ 0`. This all works as expected in `v1.7.2` but the rewriting does not happen in `master`. If the culprit is indeed `fromℕ<` then there may be a simpler minimal example but I have not had time to look into that yet.

jamesmckinna:
```agda
module unbug where

open import Data.Maybe using (just)
open import Data.Fin using (fromℕ<)
open import Data.List using (List; lookup; length; head; _∷_; [])
open import Data.Nat using (_<_; ℕ; z<s)
open import Data.Nat.Properties using (n<1⇒n≡0; ≤-reflexive)
open import Relation.Binary.PropositionalEquality using (_≡_; refl; sym)

unbug : ∀ {l : List ℕ} → (len≡1 : length l ≡ 1) → lookup l (fromℕ< (≤-reflexive (sym len≡1))) < (length l) → head l ≡ just 0
[...]

PR #2000, opened 2023-06-21
Refactoring (inversion) properties of `_<_` on `Nat`, plus consequences

jamesmckinna:
Addresses #1998: REVISED PROPOSAL
* [DONE/KEEP] Adds (a case of *reculer pour mieux sauter*?) the two inversions previously deprecated under #1709 together with an additional pattern synonym. This fixes regressions introduced by that earlier PR, which turn on the strictness of pattern matching against proofs of `_≤_`, resp. `_<_`.
* [TODO/POSTPONE] Defines `LessThan` in terms of `_<ᵇ_`, (against @Taneb 's suggestion in terms of `NonZero` and `_∸_`), and proves it irrelevantly equivalent to `_<_` (using the inversion `s<s⁻¹`)

DONE:
- [x] refactor `Data.Nat.Properties` 
- [x] deprecate `≤-pred` and `<-pred`, as copies of `s≤s⁻¹` and `s<s⁻¹`
- [x] refactor `Data.Fin.Base` and `Data.Fin.Properties`  to reinstate the irrelevant/less strict semantics of `fromN<` etc.
- [x] knock on changes in `Data.Integer.Properties`
- [x] ditto. `Data.Nat.Binary.Properties`
- [x] ditto. `Data.Vec.Bounded.Base`
- [x] `CHANGELOG`

 TODOs for subsequent PRs:
- [ ] well/better-thought-out design of `LessThan` and its behaviour/properties;
- [ ] *naming*: `LessThan` (on the model of `NonZero`), or yet another variation on `_<_`?
- [ ] refactor/deprecate the various `fromN*` definitions in favour of ones using `LessThan` in an irrelevant way;
- [ ] in particular, does this simplify the analysis of the `--with-K` equational reasoning in `Data.Nat.DivMod.WithK`, because `fromN<` can now be defined to have an equational property which doesn't depend on the proof of `_<_`? ie we only need apply the `≡-erase` analysis directly to that proof, instead of detour via `fromℕ<″` and `≤″-erase`?
- [ ] what else?


src/Data/Nat/Base.agda
```diff
@@ -120,11 +129,29 @@ instance
 >-nonZero⁻¹ (suc n) = z<s
 
 ------------------------------------------------------------------------
--- Arithmetic
+-- Another useful (proof-irrelevant) ordering relation
 
 open import Agda.Builtin.Nat public
```
Taneb, line 134: These should definitely be opened under the `Arithmetic` header. Maybe it's worth moving them (and `div-helper` and `mod-helper`) before the ordering relations?

jamesmckinna: Hmm... I wondered about this at the time: I need monus for `LessThan`, but not (yet) the other arithmetic. Maybe a separate `import` (but I worried about subsequent `imports` shadowing the renaming of monus...?)

jamesmckinna: Or: restrict the import before `LessThan` to only consist of monus, and move the other arithmetic operations to the subsequent import?

JacquesCarette:
Certainly what's in this PR as of [80a6386](https://github.com/agda/agda-stdlib/pull/2000/commits/80a63862c1bd0f7ba9aa7717708376ce89d14167) seems good to me. The evolution of the PR will be interesting as well. 

Also: you (@jamesmckinna ) should probably mark it as Draft.

jamesmckinna:
I'm (once again) reconsidering the wisdom of this somewhat Quixotic experiment. See the discussion on #1998 .

jamesmckinna:
The regression test of issue #1988 now passes:
```agda
module Regression1988 where

open import Data.Maybe using (just)
open import Data.Fin.Base using (fromℕ<)
open import Data.List using (List; lookup; length; head; _∷_; [])
open import Data.Nat using (ℕ; zero; _<_; z<s)
open import Data.Nat.Properties using (n<1⇒n≡0; ≤-reflexive)
open import Relation.Binary.PropositionalEquality using (_≡_; refl; sym)

lazy strict : ∀ {l : List ℕ} →
              (len≡1 : length l ≡ 1) →
              lookup l (fromℕ< (≤-reflexive (sym len≡1))) < (length l) →
              head l ≡ just 0

strict {zero ∷ []} refl z<s = refl

lazy   {_    ∷ []} len≡1 headBounded rewrite n<1⇒n≡0 headBounded = refl
```

src/Data/Nat/Base.agda
```diff
@@ -233,7 +242,40 @@ zero  ! = 1
...
+------------------------------------------------------------------------
+-- Another useful (proof-irrelevant) definition of _<_
+
+LessThan : Rel ℕ 0ℓ
+LessThan m n = T (m <ᵇ n)
+
+-- instance
+
+instance
+  lessThan : ∀ {n} → LessThan 0 (suc n)
```
Taneb, line 256: Does adding a second instance `{m n} {{LessThan m n}} -> LessThan (suc m) (suc n)` work?

jamesmckinna: Good question, but my most recent commits seem to take a lot longer to typecheck with the instance inference, so i am reluctant to add extra sources of possible useless searching to the mix. But I will see what happens when I do!

jamesmckinna: So... making the instance argument irrelevant fails without patttern matching on `m` and `n`. Making it relevant makes this basically a no-op. not clear what the trade-offs are between the two solutions. What do you advise?

jamesmckinna: But I do wonder what the actual effect of defining
```agda
lessThanSucSuc : ∀ {m n} → ⦃ LessThan m n ⦄ → LessThan (suc m) (suc n)
lessThanSucSuc ⦃ lt ⦄ = lt
```
might be... a no-op, basically, or something more interesting?

jamesmckinna:
For review/comments:
- [ ] names of things such as `LessThan`
- [ ] ditto. fixities
- [ ] instances vs. implicit arguments
- [ ] deprecations of `≤-pred` and `<-pred` in `Data.Nat.Properties`
- [ ] reinstatement of the above in `Data.Nat`, but with types that mention `pred`?

src/Data/Nat/Base.agda
jamesmckinna: Compared to 
```agda
lessThanSucSuc {zero}  {suc _} = _
lessThanSucSuc {suc m} {suc n} = lessThanSucSuc {m} {n}
```

But also, I am still actually unclear about the trade-offs involved in writing such functions as 'constructors' of the relation, but using them in local `instance` declarations in subsequent proofs, or else declaring them up-front as `instances` in the module. 

Plus: the warnings about ignoring instance declarations taking explicit arguments seem to stymie one or two subsequent arguments...

MatthewDaggitt:
Sorry, I promise to have a proper look at this next week!

jamesmckinna:
My attempts at resolving the conflicts so if done were to walk me through that, I'd be grateful... (pull from master? Rebase the branch from master...? Feel slightly out if my froth with git... again)

JacquesCarette (approved):
Just eyeballing the code, I only see improvements. But I have not tried it out in practice.

jamesmckinna:
Let me know how/when your experiments with `Enumerable` go...

jamesmckinna:
I resolved the merge conflicts... which introduced some new ones (sigh). Hopefully now fixed. But there are still unresolved renaming/redundancy/deprecation issues as above, so maybe should convert to DRAFT?

src/Data/Nat/Base.agda
jamesmckinna: Now that `LessThan` is defined in terms of `_<ᵇ_`, on further reflection it seems as though `*SucSuc`-like lemmas, or instances, will just be definitional no-ops, so only serve to pollute the name/instance search space... so I think they should *not* be added? let me know if I'm misunderstanding something here?

jamesmckinna:
Converted to `DRAFT`: when/if #2056 is merged, there will be a knock-on effect on this PR (about the implementation of `inject≤` in `Data.Fin.Base` and property `lookup-inject≤-takeinject≤` in `Vec.Properties`), so waiting for that to merge before proceeding to update this PR for review.

jamesmckinna:
A mess!  I'd like to salvage a minimal version of this for v2.0... advice welcome!

jamesmckinna:
Merge conflict will I think fail on the `Relation.Binary` stuff. Will look more closely now at that.

jamesmckinna:
Regression from issue #1988 now successfully passes again.

jamesmckinna:
Note to self/selves: commit ee823acc8f610a34e2c770210d40587dd7703b5b represents a local maximum of the drfat PR.
Now turning to shrinking this down for review...

Taneb:
Did you have a use-case in mind for this? Does the implementation satisfy that use-case?

jamesmckinna:
@Taneb , does your question regarding use cases concern any such for
* `LessThan`, or some future version/deployment of it
* or the PR (as it currently stands)?

Regarding the first, which is now removed from the current state of the PR, I'll allude to my thinking above about: 
* wanting irrelevant instance version of the ordering relation: finding an *automatically* computationally irrelevant version of the propositional forms, but one which is actually usable; 
* candidate use-case: currently the library has *two* versions of `Data.Nat.DivMod`, ostensibly because one version is an 'abstract' one based on `_<_`, while the other permits a `--with-K`-erasable version of `DivMod` based on `_≤″`... and I think that there is/may be a way forward to unite both those strands, and not to need the extensive/expensive duplication. Other downstream consequences may also be available, but it seemed wise to step back from the chaos of trying to sort all that out, from the more conservative refactoring now being proposed.

Regarding the second, viz. that conservative refactoring, it's clear from issues such as #1988 that earlier (less mature/insightful!) work of mine on `Nat` and `Fin` had actually degraded the behaviour and performance of the library. The current state of the PR is an attempt to restore a behaviourally equivalent version of functionality we had before, but streamlined in the light of many things that have changed or that I/we have learnt in the last two years, towards making a better foundation for future exploration post-v2.0 (last chance to break things for a while, as you yourself have observed...)

jamesmckinna:
Merge conflict looks a little delicate, some reordering required.

src/Data/Nat/Properties.agda
```diff
@@ -2099,8 +2088,8 @@ _>″?_ = flip _<″?_
 ≤″-irrelevant : Irrelevant _≤″_
 ≤″-irrelevant {m} (less-than-or-equal eq₁)
                   (less-than-or-equal eq₂)
-  with +-cancelˡ-≡ m _ _ (trans eq₁ (sym eq₂))
-... | refl = cong less-than-or-equal (≡-irrelevant eq₁ eq₂)
+  with refl ← +-cancelˡ-≡ m _ _ (trans eq₁ (sym eq₂))
+  = cong less-than-or-equal (≡-irrelevant eq₁ eq₂)
```
jamesmckinna, line 2092: NB (cf. our discussion of v2.0 this morning)
*This* instance of the `with refl ←...` idiom *does not* typecheck if expressed in terms of `rewrite...` (!?)
Of course, the original with `... | refl = ` still works, but I'm going to leave this instance of using irrefutable `with` as it stands.

src/Data/Nat.agda
```diff
@@ -42,5 +42,14 @@ open import Data.Nat.Properties public
...
 open import Data.Nat.Properties public
-  using (≤-pred)
+  hiding (≤-pred; <-pred)
+
+≤-pred : ∀ {m n} → (m ≤ n) → pred m ≤ pred n
+≤-pred {m = zero}              _   = z≤n
+≤-pred {m = suc _} {n = suc _} m≤n = s≤s⁻¹ m≤n
+
+<-pred : ∀ {m n} → .⦃ _ : NonZero m ⦄ → (m < n) → pred m < pred n
+<-pred {m = suc _} {n = suc _} = s<s⁻¹
```
jamesmckinna, line 55: Re: our discussion this morning.

* previously, these lemmas (but with different types) were being exported here from `Data.Nat.Properties`...
* now, those lemmas are (still) defined in `Data.Nat.Properties`, but now deprecated there, so if they are used on import from that module, then they will trigger a deprecation warning, as they would have done by being imported and used from here
* the change now proposed is to continue to deprecate them in `Data.Nat.Properties`, but to allow a non-deprecated export of lemmas with those names from `Data.Nat`, but now with different types, which reflect the 'monotone/congruent' formulation of the lemmas.

So, I've probably made a mistake giving them the same name as the former lemmas which are now deprecated. Perhaps it would be better to have *new* names for these (and then find a home for them, either in this module, or indeed in `Data.Nat.Properties`), while continuing to export the old names, but now deprecated, from `Data.Nat.Properties`.

Apologies for causing so much confusion with this change!

So it seems that there are two options:
* don't bother to have these lemmas at all
* rename them, and then they will be distinct from the deprecated names, and then find a home for them somewhere

What do others think?

jamesmckinna: My own view is to retain the new lemmas, but rename them as
* `pred-≤-monotone` (or `...-mono`)
* `pred-<-monotone` (or `...-mono`)

and push them into `Data.Nat.Properties`, but then re-export them here.

jamesmckinna: Ha, there already exists `pred-mono` for the first one! Sigh... for consistency, I'd want to deprecate *that* name as well. Oh dear.

jamesmckinna:
Provisional commits incoming to try to resolve the deprecation mess. But not yet definitive.
Then attempt to resolve merge conflict.
Then sleep on things!

jamesmckinna:
Merge commit defe03a30a1a78b208800ac7883b347a61049571 seemed (?) to have (re-)introduced some bugs, as well as one new occurrence of `≤-pred`. Hopefully fixed all these up now.

src/Data/Fin/Base.agda
```diff
@@ -68,16 +67,15 @@ fromℕ (suc n) = suc (fromℕ n)
 
 -- fromℕ< {m} _ = "m".
 
-fromℕ< : m ℕ.< n → Fin n
-fromℕ< {zero}  {suc n} z<s = zero
-fromℕ< {suc m} {suc n} (s<s m<n) = suc (fromℕ< m<n)
+fromℕ< : ∀ {m n} → .(m ℕ.< n) → Fin n
```
MatthewDaggitt, line 70: I agree with making the argument irrelevant. However is there a reason why we have re-introduced the implicit arguments that are already variables?

src/Data/Fin/Base.agda
```diff
@@ -95,9 +93,9 @@ zero    ↑ʳ i = i
 
 -- reduce≥ "m + i" _ = "i".
 
-reduce≥ : ∀ (i : Fin (m ℕ.+ n)) (i≥m : toℕ i ℕ.≥ m) → Fin n
-reduce≥ {zero}  i       i≥m       = i
-reduce≥ {suc m} (suc i) (s≤s i≥m) = reduce≥ i i≥m
+reduce≥ : ∀ (i : Fin (m ℕ.+ n)) → .(m ℕ.≤ (toℕ i)) → Fin n
```
MatthewDaggitt, line 96: Don't need the brackets around `toℕ i`?

src/Data/Fin/Base.agda
```diff
@@ -95,9 +93,9 @@ zero    ↑ʳ i = i
...
+reduce≥ {zero}  i       _   = i
+reduce≥ {suc m} (suc i) m≤i = reduce≥ {m} i (ℕ.s≤s⁻¹ m≤i)
```
MatthewDaggitt, line 98: Is the application of `{m}` really necessary here? A bit worried if it is...

src/Data/Fin/Properties.agda
```diff
@@ -532,34 +534,34 @@ inject₁≡⇒lower₁≡ n≢j i≡j = inject₁-injective (trans (inject₁-l
...
+inject≤-injective : ∀ .(m≤n m≤n′ : m ℕ.≤ n) i j →
                     inject≤ i m≤n ≡ inject≤ j m≤n′ → i ≡ j
-inject≤-injective (s≤s p) (s≤s q) zero    zero    eq = refl
-inject≤-injective (s≤s p) (s≤s q) (suc i) (suc j) eq =
-  cong suc (inject≤-injective p q i j (suc-injective eq))
+inject≤-injective {n = suc _} p q zero    zero    eq = refl
+inject≤-injective {n = suc _} p q (suc i) (suc j) eq =
+  cong suc (inject≤-injective (ℕ.s≤s⁻¹ p) (ℕ.s≤s⁻¹ q) i j (suc-injective eq))
 
 inject≤-irrelevant : ∀ (m≤n m≤n′ : m ℕ.≤ n) i →
```
MatthewDaggitt, line 562: Given that you're marking them irrelevant everywhere else?

src/Data/Fin/Properties.agda
```diff
@@ -891,10 +898,10 @@ pinch-surjective zero    (suc j) = suc (suc j) , λ { refl → refl }
 pinch-surjective (suc i) (suc j) = map suc (λ {f refl → cong suc (f refl)}) (pinch-surjective i j)
 
 pinch-mono-≤ : ∀ (i : Fin n) → (pinch i) Preserves _≤_ ⟶ _≤_
-pinch-mono-≤ 0F      {0F}    {k}     0≤n       = z≤n
-pinch-mono-≤ 0F      {suc j} {suc k} (s≤s j≤k) = j≤k
-pinch-mono-≤ (suc i) {0F}    {k}     0≤n       = z≤n
-pinch-mono-≤ (suc i) {suc j} {suc k} (s≤s j≤k) = s≤s (pinch-mono-≤ i j≤k)
+pinch-mono-≤ 0F      {0F}    {k}     0≤n = z≤n
+pinch-mono-≤ 0F      {suc j} {suc k} j≤k = (ℕ.s≤s⁻¹ j≤k)
```
MatthewDaggitt, line 902: No need for brackets?

MatthewDaggitt (commented):
Okay so my view is that:
- We should keep `pred-mono-≤ ` and `pred-mono-<` as is in `Data.Nat.Properties` as they're consistent with proofs for other operators.
- We should actually keep `≤-pred` and `<-pred` in `Data.Nat.Properties` with their old type _undeprecated_ and simply implement them as `s≤s⁻¹` and `s<s⁻¹`. My reasoning for this that they are two different views, the former are destructors of the datatype and the other are properties of the relation.

Also heads up that there's a merge conflict...

src/Data/Fin/Properties.agda
jamesmckinna: Good catch!

src/Data/Fin/Properties.agda
jamesmckinna: Indeed!? Don't know what happened here.

src/Data/Fin/Base.agda
jamesmckinna: Probably not. More 'note-to-self/others' internal documentation... but we've disagreed before about my tendency to overwork the implicit argument inference, so sometimes I put things back which aren't perhaps necessary? Anyway, I didn't write the original definition, so it's probably legacy! Will fix...

src/Data/Fin/Base.agda
jamesmckinna: No. Legacy I think.

src/Data/Fin/Base.agda
jamesmckinna: Hmm, good question. Legacy I think, in terms of restoring the module to its state pre my original investogtaions. But should tidy this up!

src/Data/Nat/Properties.agda
```diff
@@ -168,9 +167,7 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 ≤-total : Total _≤_
 ≤-total zero    _       = inj₁ z≤n
 ≤-total _       zero    = inj₂ z≤n
-≤-total (suc m) (suc n) with ≤-total m n
-... | inj₁ m≤n = inj₁ (s≤s m≤n)
-... | inj₂ n≤m = inj₂ (s≤s n≤m)
+≤-total (suc m) (suc n) = [ inj₁ ∘ s≤s , inj₂ ∘ s≤s ]′ (≤-total m n)
```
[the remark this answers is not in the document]
jamesmckinna: Ah! Thanks!

src/Data/Nat/Base.agda
```diff
@@ -326,6 +338,16 @@ m ≥″ n = n ≤″ m
...
 m >″ n = n <″ m
 
+-- Smart constructors of _≤″_ and _<″_
+
+pattern ≤″-offset k = less-than-or-equal {k = k} refl
+pattern <″-offset k = ≤″-offset k
+
+-- Smart destructor of _<″_
+
+s<″s⁻¹ : ∀ {m n} → suc m <″ suc n → m <″ n
```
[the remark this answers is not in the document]
jamesmckinna: Yes, but I never used it anywhere. This lemma gets one use, and as you know, I'm on a kick to get rid of this relation altogether anyway, so I wasn't in a hurry to start adding properties to it. But tidier with, I suppose.

src/Data/Nat.agda
```diff
@@ -42,5 +42,8 @@ open import Data.Nat.Properties public
 
 -- Version 0.17
 
+-- Version 2.0
+
+-- solely for the sake of exposing the deprecation warnings for these two names
 open import Data.Nat.Properties public
-  using (≤-pred)
+  using (≤-pred; <-pred)
```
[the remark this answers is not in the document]
jamesmckinna: Exporting ti for the sake of deprecation. But yes, I'm not even sure it existed before v2.0 as a self-standing thing. It woul dbe the first time someone deprecated something they themselves introduced between 1.7.x and 2.0... it's sometimes hard to keep track of these things. How do you do it?

src/Data/Nat/Base.agda
```diff
@@ -283,7 +292,10 @@ zero  ! = 1
 suc n ! = suc n * n !
 
 ------------------------------------------------------------------------
--- Alternative definition of _≤_
+-- Alternative definitions of _≤_/_<_
+
+------------------------------------------------------------------------
+-- An alternative definition of _≤_
```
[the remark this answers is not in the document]
jamesmckinna: Probably some unfinished edits/uncommitted edits/unpushed commits somewhere. Good spot!

jamesmckinna:
Interesting comment about *un*deprecating the `pred` properties. I'll ponder (I'm once again awake in the middle of the night, and should consider things when I am more awake later)
Oh, the merge conflcits! Yes, another knock-on consequence of keeping PRs alive for too long. Will look at, similarly, later!

src/Data/Nat/Base.agda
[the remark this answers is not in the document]
MatthewDaggitt: If the plan is to get rid of it altogether then in that case, don't worry about it.

src/Data/Nat.agda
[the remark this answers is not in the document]
MatthewDaggitt: If I'm in doubt, I use the Github web interface to go back in time to `v1.7.2` then navigate to the module in question and have a look.

src/Data/Nat/Base.agda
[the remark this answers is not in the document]
jamesmckinna: Well, eventually. I'll introduce here, but I can pretty much guarantee that it won't (shouldn't!) be used...
OTOH, maybe it *is* useful, to somebody, one day?

src/Data/Fin/Base.agda
jamesmckinna: Done.

src/Data/Nat/Base.agda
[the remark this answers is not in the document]
jamesmckinna: So I re-read the whole section, and redrafted all the headers. 
Hope this is an improvement!

src/Data/Fin/Properties.agda
jamesmckinna: And throughout. Let me know if I haven't caught them all...

jamesmckinna:
You wrote: 
> Okay so my view is that:
> 
>     * We should keep `pred-mono-≤ ` and `pred-mono-<` as is in `Data.Nat.Properties` as they're consistent with proofs for other operators.

NB. Introduced by in this PR! And the first is (a simplified version of) `pred-mono` (which has a different proof). So we should perhaps decide on which one gets saved, and which gets deleted/deprecated. 

>     * We should actually keep `≤-pred` and `<-pred` in `Data.Nat.Properties` with their old type _undeprecated_ and simply implement them as `s≤s⁻¹` and `s<s⁻¹`. My reasoning for this that they are two different views, the former are destructors of the datatype and the other are properties of the relation.

Interesting. Happy to oblige!

> Also heads up that there's a merge conflict...

Inevitably! :-(

src/Data/Nat.agda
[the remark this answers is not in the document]
jamesmckinna: Your suggestion below regarding *un*deprecation now seems to resolve this. But thanks for the pro tips about `git` ...

PR #2002, opened 2023-06-22
Eliminate many propositional equality imports

Taneb:
Many of the remaining imports of `Relation.Binary.PropositionalEquality` use either `_≗_` or `→-to-⟶`. Moving `_≗_` to `Relation.Binary.PropositionalEquality.Core` feels like a clear win, except it's defined in a little bit of an indirect way that pulls in some more imports


src/Data/Bool/Properties.agda
```diff
@@ -22,9 +22,12 @@ open import Function.Equivalence
 open import Induction.WellFounded using (WellFounded; Acc; acc)
 open import Level using (Level; 0ℓ)
 open import Relation.Binary hiding (_⇔_)
-open import Relation.Binary.PropositionalEquality hiding ([_])
+open import Relation.Binary.PropositionalEquality.Core
+open import Relation.Binary.PropositionalEquality.Properties
 open import Relation.Nullary.Reflects using (ofʸ; ofⁿ)
 open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no)
+open import Relation.Nullary using (ofʸ; ofⁿ; does; proof; yes; no)
+open import Relation.Nullary.Decidable using (True)
```
gallais, line 30:
```suggestion
open import Relation.Nullary.Reflects using (ofʸ; ofⁿ)
open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no)
```

gallais (commented):
I think all the alignment changes in `Reflection.AST.*`  suggests we should just do away with lining up the `using` keywords.

gallais:
Also please don't force push the changes if you make any, it'll be a *lot* easier to
review an additional couple commits than have to go through the whole 200 modules
one by one once again.

gallais (approved):

PR #2003, opened 2023-06-22
Simplify `Data.Product` imports

Saransh-cpp:
Some straightforward simplifications of `Data.Product` to `Data.Product.Base`. 

This PR does not get rid of all the `Data.Product` imports. I'll be creating PRs for the same.


gallais:
Some of these are already covered by #1995. We should maybe wait for that PR
to be merged to avoid duplicating too much work.

Saransh-cpp:
The majority of the `Data.Product` imports were indeed only using `Data.Product.Base`.  I've simplified all of the `Data.Product` imports (excluding the ones importing `∃`, `∃-syntax`, ...) in my fork and will create PRs iteratively to avoid confusion.

The imports that I could not simplify involved importing `∃`, `∃-syntax`, ... from `Data.Product`. Moving `∃`, `∃-syntax`, ... to `Data.Product.Base` will simplify every `Data.Product` import in the library to `Data.Product.Base`, but I'm not sure if that should be done.

Saransh-cpp:
Ah, #1995 already moves the problematic variables and operators to `Data.Product.Base`.

MatthewDaggitt:
https://github.com/agda/agda-stdlib/pull/1995 is now merged, so this can be rebased.

gallais (approved):

Issue #496, opened 2018-10-16
Include laws for Functor, Applicative, Monad


alex-mckenna:
Currently, the modules for Functor, Applicative and Monad include records for the Raw types similar to the algebraic structures, but do not have corresponding types including the laws. I think this could be a nice issue for someone like myself who knows the laws, and wants to take a gentle first step into contributing to the standard library.

From what I can tell this would involve:

  - [ ] Creating a `Category.Structures` module for definitions of `IsFunctor`, `IsApplicative` and `IsMonad`
  - [ ] Creating the extended records which include proof the laws hold

alex-mckenna:
@gallais Any comments / gotchas I should know about with this?

gallais:
> Any comments / gotchas I should know about with this?

The right notion of equality to use in these is a hard question. E.g. you'll probably need
functional extensionality to prove the laws for `State` if you pick `PropositionalEquality`
as your notion of equality.

HuStmpHrrr:
i think this is a good issue. I was wondering the same thing too.

jespercockx:
It would be great if someone could take a look at this! It is kind of embarrassing to claim one of the big advantages of Agda over Haskell is the ability to specify lawful typeclasses, but then the standard library does not actually do this.

JacquesCarette:
I was hoping to start migrating the basic parts of `agda-categories` to `agda-stdlib`, so that `Category`, `Functor` and friends could be 'right'.  I think that things in `agda-categories` have settled, so that once stdlib-1.3 ships, that could start.

jamesmckinna:
... fast-forward to v2.x and year 2024...

JacquesCarette:
Right, this serves as a nice reminder for me to start a design discussion on just that.

PR #2004, opened 2023-06-22
Add new module `Effect.Functor.Naperian`

Sofia-Insa:
New module `Effect.Functor.Naperian` added to stdlib. 
With two records defining Naperian functors: 
* `record RawNaperian`
* `record Naperian`.


jamesmckinna:
I think, but I'm not sure (so looking at related examples in the `Algebra` hierarchy might clarify) that the modern way is to make compound records 'flat', and then in the more-decorated version (here:`Naperian`) to re-export an instance of the less-decorated (here:`RawNaperian`). But maybe the `Effect.*` sub-hierarchy does it differently?

jamesmckinna:
[Comment deleted because based on a misunderstanding in my part]

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
...
+-- APLicative Programming with Naperian Functors. Jeremy Gibbons.
+
+-- RawNaperian contains just the functions, not the proofs
+record RawNaperian {F : Set b → Set c} (d : Level) (RF : RawFunctor F) : Set (suc (b ⊔ d) ⊔ c) where
+  field
+    Log : Set d
+    index : ∀ {A} → F A → (Log → A)
+    tabulate : ∀ {A} → (Log → A) → F A
+
+-- Full Naperian has the coherence conditions too. Propositional version (hard to work with).
```
jamesmckinna, line 36: In general, we have had problems in the past with trying to work out how best to tackle equality on `Functor`s (and hence also [formalise the laws](https://github.com/agda/agda-stdlib/issues/496) for them, as well as `Applicative, `Monad, etc.) *in general*, but here, precisely because of the `RawNaperian` structure, we can lift `Setoid` structure `_≈_` on `A` to (the `Pointwise`) one on `Log → A`... and thus induce a `Setoid` structure on `F A` ? (there may well be details to check/fix here!)

... where it's worth noting that `Log` itself should probably be considered only `Propositional`ly, that is, indexing should be considered *intensionally* with respect to the underlying `Set` (perhaps Joyal's theory of species requires more complex `Setoid` (or:other!) indexing, but that's out-of-scope for this PR!)

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
...
+record Naperian {F : Set b → Set c} (d : Level) (RF : RawFunctor F) : Set (suc (b ⊔ d) ⊔ c) where
+  field
+    rawNaperian : RawNaperian d RF
+  open RawNaperian rawNaperian public
+  field
+    tabulate-index : ∀ {A} → (fa : F A) → tabulate (index fa) ≡ fa
```
jamesmckinna, line 42: So: here the induced `Setoid` relation on `F A`?

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
...
+    index-tabulate : ∀ {A} → (f : Log → A) → ((l : Log) → index (tabulate f) l ≡ f l)
```
jamesmckinna, line 43: And here: that on `A`?

jamesmckinna (changes requested):
A very nice initial go at a contribution in this area, and hopefully one which can be taken further as suggested, without too much difficulty.

JacquesCarette:
I should take this PR over. I'm not entirely sure I know how?

MatthewDaggitt:
Given the comment of @JacquesCarette above, @jamesmckinna would you be willing to take over this? Otherwise, I close it.

jamesmckinna:
@MatthewDaggitt @JacquesCarette happy to take over on this one, but it'll be a while before I return attention to it, if that's OK?

CHANGELOG.md
```diff
@@ -2879,7 +2884,18 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-
+  
+ * Added new lemmas in `Data.List.Propreties`:
+  ```
+  take-[] : ∀ m → take  m [] ≡ []
+  drop-[] : ∀ m → drop  m [] ≡ []
+  ```
```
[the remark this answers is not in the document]
jamesmckinna: Done (I hope).

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
...
+
+module Effect.Functor.Naperian where 
+
+open import Effect.Functor using (RawFunctor)
+open import Level using (Level; suc; _⊔_)
+open import Relation.Binary.PropositionalEquality.Core using (_≡_)
+
+private
+  variable
+    b c : Level
```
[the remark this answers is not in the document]
jamesmckinna: Done.

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
...
+-- then p behaves a little like a logarithm of f
+-- in particular, if f and g are both Naperian, then Log(f×g)≃Logf+Logg and Log(f.g) ≃ Log f × Log g"
+-- APLicative Programming with Naperian Functors. Jeremy Gibbons.
+
+-- RawNaperian contains just the functions, not the proofs
+record RawNaperian {F : Set b → Set c} (d : Level) (RF : RawFunctor F) : Set (suc (b ⊔ d) ⊔ c) where
+  field
+    Log : Set d
+    index : ∀ {A} → F A → (Log → A)
+    tabulate : ∀ {A} → (Log → A) → F A
```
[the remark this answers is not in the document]
jamesmckinna: Done.

jamesmckinna:
Minimal cleanup for the time being.

JacquesCarette:
Thanks a lot! I've been wanting to come back to these, but I've barely found enough time to do a few reviews.

src/Effect/Functor/Naperian.agda
jamesmckinna: I think this perhaps entails changing the parametrisation of 'full' `Naperian` to build in the `Setoid` structure (otherwise we're back at `Set\omega`), and then back-derive its underlying `RawNaperian` functor. So more thought required!

JacquesCarette:
This is now back onto my plate to finish. And I even know how to do so!

src/Effect/Functor/Naperian.agda
```diff
@@ -0,0 +1,43 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Naperian functor
+--
+-- This file contains definitions of Naperian introduced by Jeremy Gibbons 
+-- in the book APLicative Programming with Naperian Functors.
+-- https://link.springer.com/chapter/10.1007/978-3-662-54434-1_21
```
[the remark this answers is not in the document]
JacquesCarette: Have clarified the text (but not given the full history, as that really ought to go elsewhere)

jamesmckinna:
> This is now back onto my plate to finish. And I even know how to do so!

Fantastic! I'm happy to hand the baton back. But sign me up as a reviewer come the right time...

src/Effect/Functor/Naperian.agda
JacquesCarette: I've done an attempt at this - could you check it out please?

jamesmckinna (changes requested):
Besides the notational nitpicking (which I think important, but advisory here), I think the design needs some more thinking as to the dependency/parametrisation in the `Setoid` case.

jamesmckinna:
The PR is still DRAFT, so I'll revisit/re-review when it's 'ready', but at that point I'll also revisit the early comment above about proving the 'logarithm' properties.

jamesmckinna:
Possible low(er)-hanging fruit simpler warm-up task (@JacquesCarette or your students?):
* for `D : DecSetoid c ℓ`...
* and `d : D.Carrier`
* define `D \ { d } : DecSetoid c ℓ` as 'the obvious thing'?

together with the 'obvious' adjunction (suitably understood ;-)) between adding a point (as an operation on `DecSetoid`s extending `...Construct.Add.Point`) and subtracting it as above?

JacquesCarette:
This is superceded by #2815 which leverages all the work done here.

Issue #1881, opened 2022-11-15
Add a notion of finite algebraic structures


Taneb:
e.g. a group is finite if there exists `order : ℕ` with `isFinite : Bijection setoid (≡-setoid (Fin order))`.

There's also the constructive notion of subfinite structures, which only have an injection to some finite set, for example arbitrary subgroups of finite groups.

I was imagining this going somewhere like `{Relation.Binary,Algebra}.Finite.{Structures,Bundles}`

gallais:
Is this #863?

Taneb:
> Is this #863?

I hadn't seen that issue! And it's definitely related. But this issue goes further, asking for finite groups and not just finite setoids. The discussion in #863 is relevant, though.

Issue #1770, opened 2022-05-23
Algebraic structures with stable or decidable equality


Taneb:
In light of #1767, it might be useful to have copies of the algebraic hierarchy with _stable_ equality, and with _decidable_ equality

Here's a very rough sketch of what this could look like

```agda
record IsStableMagma {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) (∙ : Op₂ A) : Set (a ⊔ ℓ) where
  field
    isMagma : IsMagma _≈_ ∙
    isStable : ∀ x y → Stable (x ≈ y)

record IsDecidableMagma {a ℓ} {A : Set a} (_≈_ : Rel A ℓ) (∙ : Op₂ A) : Set (a ⊔ ℓ) where
  field
    isMagma : IsMagma _≈_ ∙
    isDecidable : ∀ x y → Dec (x ≈ y)

  isStableMagma : IsStableMagma _≈_ ∙
  isStableMagma = record
    { isMagma = isMagma
    ; isStable = λ x y → decidable-stable (isDecidable x y)
    }
```

I've heard that many classical theorems, when you try to prove them constructively, depend on decidablility of equality.

I'm not sure how this interacts with the recently added Apartness things. I _suspect_ that Apartness is weaker than Stable.

Taneb:
I've been thinking about this recently, and I think there's a hierarchy that goes something like `Stable` -> `Decidable` -> `Enumerable`  (not actually sure how to define this)-> `Finite` (plus maybe `Subfinite` but I don't know where that goes). (c.f. #1881, I hadn't realised how they interacted when I opened it).

Given that this is getting to be quite a few duplicates of the algebra hierarchy, which is itself already very big, we should maybe consider doing something cleverer. @JacquesCarette I'd be interested in your thoughts here

JacquesCarette:
Sorry the summoning took so long to actually work, life interfered a lot.

I've definitely thought long and hard about just this problem, and have done some prototype implementations. In theory, this is some kind of "mixin" situation.

Unfortunately, it's a mixin "at the wrong level", i.e. at a part of Agda that is very far from first-class: record *declarations*. The duplication occurs not at the value level, but at the level of extending the environment with new definitions derived from old ones.

In an internalized representation (think of doing Universal Algebra in Agda), this is all relatively straightforward. The problem is that this results in something highly unusable - the levels of 'coding' is unbearable.

[...]

Issue #863, opened 2019-08-18
finite setoid


HuStmpHrrr:
I am considering a definition of finite setoids, which is a setoid with a bijection to a `Fin`. then there are a few useful properties:

1. all finite setoids have decidable equality.
2. all finite setoids can be finitely enumerated.
3. there are other things like finite setoids are closed under finite products and sum.

JacquesCarette:
There is a downside to that: a bijection to `Fin` also gives an ordering. Which means that all finite setoids inherit an ordering by transporting it along the bijection. The problem here is that you don't want your bijection to be proof-relevant! If you want to see way more details about this, see [Brent Yorgey's Ph.D. thesis](https://github.com/byorgey/thesis) (for the full repo with all the versions).

I'm not saying don't do this!  I'm saying that what you really want is a "Finite" API that may contain a bijection as an implementation, but the API doesn't actually given access to the full details.

JacquesCarette:
I guess what I'm saying is that maybe this should be called `StronglyFinite` rather than `Finite`. Then such a type would be quite useful indeed.

MatthewDaggitt:
Sounds like a great idea to me. I'd definitely use it! I would suggest maybe putting `StronglyFinite` under `Relation.Nullary.Finite.Setoid` with a corresponding module `Relation.Nullary.Finite.Propositional`?

MatthewDaggitt:
Actually speaking of this, @HuStmpHrrr don't suppose you have a proof that transitive closure is decidable if the underlying relation is decidable over a strongly finite setoid do you? That's what I really need this for and I'm dreading doing it :disappointed: Interestingly I think this property may require strong finiteness rather than @JacquesCarette's alternative notion of finiteness.

HuStmpHrrr:
@MatthewDaggitt I believe that's true. imagine a relation represented as a graph, then the graph is finite and this can be achieved by DFS or BFS. either case, however, seems difficult to implement in type theory.

WolframKahl:
For “strongly finite”, any concrete (proven) implementation of transitive closure on `Fin n` will serve to “decide transitive closure”; I have section 5.5 “Simple Recursive Star Implementation on `Fin n`” in:
````
@InProceedings{Kahl-2012_AgdaGT1,
  author = 	 {Wolfram Kahl},
  title = {Towards Certifiable Implementation of Graph Transformation via Relation Categories},
  pages = 	 {82--97},
[...]

PR #2005, opened 2023-06-22
Upload file `Relation.Unary.Enumerable` to stdlib 

Sofia-Insa:
Upload file `Relation.Unary.Enumerable` to stdlib 
with one `record IsEnumerable`

changes declared in CHANGELOG.md


jamesmckinna:
So I think of `Unary`, `Binary` as qualifiers to `Relation` as defining unary etc. *predicates on a given set*. 

Here you've defined a property of sets *in general*. 

So, somewhat unfortunately (I continue to find it a very awkward naming convention), I think I am right in saying that such properties belong instead under `Relation.Nullary`. (Along the analogy with `Decidable`... and indeed, an `Enumerable` of size 2 had better be equivalent to such a thing?)

Can one of the 'true' experts please confirm this?

Taneb:
I don't qualify as a "true" expert if @jamesmckinna doesn't, but I agree: this should go under `Relation.Nullary`.

Also, this seems to be only _finitely_ enumerable. Maybe `Finite` would be a better name for it? Although I'd like to see an actual `Enumerable` as well.

jamesmckinna:
Alternatively/additionally, you might wish to factor out as an `IsP`/`HasP` record the property of being 'of size `n`', with `Enumerable` then being the 'bundled' form which packages the size `n` and the 'of size `n`' property...?

jamesmckinna:
And as for `Finite` vs. `Enumerable', I'd be inclined to agree with @Taneb , were the many intuitionistically distinct notions of 'finite'ness not already a minefield. 

Suggest instead, somewhat reluctantly given the verboseness: `FinitelyEnumerable` (groan)

src/Relation/Unary/Enumerable.agda
```diff
@@ -0,0 +1,20 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Relation.Unary.Enumerable where
+
+  open import Data.Fin using (Fin)
+  open import Data.Nat using (ℕ)
+  open import Level using (Level)
+  open import Function.Bundles using (_↔_)
```
gallais, line 14:
```suggestion
  open import Data.Fin.Base using (Fin)
  open import Data.Nat.Base using (ℕ)
  open import Level using (Level)
  open import Function.Bundles using (_↔_)
```

gallais (changes requested):
Let's minimise the import footprint.

It would also be good to have more properties e.g.

```agda
enumerate : IsEnumerable A -> List A
enumerated : (enum : IsEnumerable A) -> (x : A) -> x ∈ enumerate enum
```

And closure under sum, product

JacquesCarette:
Just to be clear, I've asked @Sofia-Insa to "port" some things from a library of mine to stdlib. That library was developed with a slightly different context in mind, which is what is creating some of these issues. All of which do need to be fixed. [I'm mostly saying this so that the blame goes to me.]

On tightening imports: absolutely. What made it to this PR was already better than the starting point...

I completely despise `Relation.Nullary`. Awful naming scheme. However, I currently have nothing better to offer, and agree that that is where current conventions say this should go.

I also agree that `IsEnumerable` should be the predicate with `n` as a parameter and `Enumerable` should be the Bundled version. 

I don't like `Finite` for the name as there are too many constructive definitions of `Finite` which end up being different. #863 calls this `StronglyFinite`. I kind of prefer that over `FinitelyEnumerable` ? Not strongly (ha!) though. There are relevant discussions in #1881 and #1770.

I agree about having more properties (and I already have some of these, so they'd be incoming as well). I think it's better to first settle on the many issues brought up here, then submit a new PR with additional properties.

jamesmckinna:
I think @gallais 's suggested additions would make this a 'better' PR, and now that you (@Sofia-Insa) have cut your teeth on some smaller PRs, working up to a slightly bigger one seems like a worthwhile progression?

Regarding closure properties (but I think there may also be prior work on the `(Near)Semiring` structure on `Set`? Can't remember where/when/who though?)
- [ ] closure under Sum should be straightforward
- [ ] closure under Product: row order or column order? I guess the former makes more sense for the generalisation to dependent product?

Regarding naming (for all authors/reviewers to consider), some alternative suggestions spring to mind, but these can all (?) be fixed put at the end, once the module contents stabilise:
- [ ] should `enumerate` be called `toList` for uniformity/consistency with other aspects of the library?
- [ ] ditto. `enumerated`: `toList∈`?
- [ ] regarding the `*Finite` designations, I guess I'd be OK with `StronglyFinite`, except that I don't personally find `Strongly` an especially helpful qualifier, as it's not entirely clear to me what the (linear?) ordering by 'strength' of these things is supposed to be...

Taneb:
I'd be a lot happier with `StronglyFinite` if we also had a `WeaklyFinite` with the appropriate implication

gallais:
Also: should this be generalised to an arbitrary setoid?
The use of `_↔_` forces propositional equality but `A` could have a more interesting structure than that!

jamesmckinna:
Also (though this is perhaps marginal?):
- [ ] `toList` or `toVec`, given that the `size` is statically known?
- [ ] if generalised to `Setoid`, do we end up needing to proliferate the number, and types, of enumerations in the presence of eg `DecSetoid` to ensure no duplications in the enumeration?

gallais:
Can we not prove the no duplication just from the bijection assumption?

JacquesCarette:
Yes, we should absolutely generalize this to an arbitrary setoid. And indeed, the bijection is enough to prove no duplication.

Note that since `Fin n` has decidable equality, setoid `A` will automatically have a `DecSetoid` structure.

The discussion does bring to mind a philosophical question of `stdlib` development: should it be allowed to be small-step incremental, or only big complete chunks are wanted? [Personally, given the current slow growth, I would vote for small-step; once we get to a dozen  PR/day (I think Lean gets even more), obviously that can be revisited.]

gallais:
I am not sure I agree. It is common to get definitions subtly wrong and
we often only discover that when trying to prove results. In which case
adding a single definition means that we may need to make backwards
incompatible changes in the future to patch the issue. Which will be a pain.

E.g. the fact we want a `Setoid` arises when we start thinking of proving
that if `A` and `B` are finite then so is `A -> B`.

JacquesCarette:
That's an excellent point. Ok, I have to agree, as I've witnessed "subtly wrong" sufficiently many times. Sigh.

JacquesCarette:
I've converted this PR to Draft, so that it can be augmented to be more full-featured regarding (at least) finite enumerability.

jamesmckinna:
Thanks @JacquesCarette and due apologies to @Sofia-Insa for getting caught up in a longer-running, but important, debate about library design, and with it, the role and size of PRs

My own thinking, for what it's worth, is that I (could be persuaded to) prefer 'small' PRs when there is an 'easy'/label:low-hanging-fruit issue at stake. 

But a lot of label:addition PRs represent ... well, choices where things may, indeed, be "subtly wrong", so I hope that everyone can entertain the debate in a constructive fashion. And that involves *both* the mathematics, as here, and its implementation in type theory...

jamesmckinna:
Apologies for not understanding bijections between `Set`s and `DecSetoid`s... my brain too full of other stuff to have thought things through properly. But then, all the more so: flatten to a `List`, or to a (`Functional`) `Vec`?

gallais:
> But then, all the more so: flatten to a List, or to a (Functional) Vec?

Probably every single one of them using the `toList`, `toVec`, (`toFresh`?) naming scheme you suggested earlier.

MatthewDaggitt:
Sorry, I've missed most of this conversation. I agree with all the conclusions, i.e. `Relation.Nullary`, extending it to `Setoid`, and in this case when introducing new definitions they should have at least a small selection of proofs about them to show the definition is reasonable.

JacquesCarette:
Another one that I should take over. I most definitely want to continue pushing on this.

MatthewDaggitt:
@JacquesCarette any chance you plan to pick this up in the next couple of weeks? Otherwise, I'll post a link to this discussion in the original issue and then close this as there's not very much code here.

JacquesCarette:
Next couple of weeks? Unfortunately not so likely, sorry.

PR #2009, opened 2023-06-27
Simplify some `Relation.Binary` imports

Saransh-cpp:
This PR does not get rid of all the `Relation.Binary` imports. I'll be creating PRs for the same.


src/Relation/Binary/Indexed/Homogeneous/Structures.agda
```diff
@@ -20,7 +20,10 @@ module Relation.Binary.Indexed.Homogeneous.Structures
 open import Data.Product using (_,_)
 open import Function.Base using (_⟨_⟩_)
 open import Level using (Level; _⊔_; suc)
-open import Relation.Binary as B using (_⇒_)
+open import Relation.Binary.Core using (_⇒_)
+import Relation.Binary.Definitions as Definitions
+import Relation.Binary.Bundles as Bundles
+import Relation.Binary.Structures as Structures
```
Saransh-cpp, line 26: I am not very sure about the naming schemes here

gallais: Can't you import them all as `B`?

Saransh-cpp: That worked, thanks @gallais!

JacquesCarette (approved):

src/Text/Regex/Properties.agda
```diff
@@ -23,7 +23,8 @@ open import Relation.Nullary.Negation
   using (¬_; contradiction)
 
 import Relation.Unary  as U
-open import Relation.Binary using (Decidable)
+open import Relation.Binary.Definitions using (Decidable)
+import Relation.Binary.PropositionalEquality as P
```
Saransh-cpp, line 27:
```suggestion
```

Taneb:
Does anyone have any objection to merging this and the similar PRs? I think they've sat open long enough

JacquesCarette:
I'd like some truly independent person to do a final review (perhaps @gallais ?) before doing so.

gallais (approved):

PR #2016, opened 2023-07-07
simplified `String` import in stdl

Sofia-Insa:
simplified `String` import in stdl
In `src.Text.Tabular.Vec` and `src.Relection.AST.Show` still need to import `Data.String` to use `rectangle` and `parensIfSpace`


src/Reflection/AST/Show.agda
```diff
@@ -16,8 +16,8 @@ import Data.Float as Float using (show)
 open import Data.List.Base hiding (_++_; intersperse)
 import Data.Nat.Show as ℕ using (show)
 open import Data.Product   using (_×_; _,_)
-open import Data.String as String
-  using (String; _++_; intersperse; braces; parens; parensIfSpace; _<+>_)
+open import Data.String.Base as String using (String; _++_; intersperse; braces; parens; _<+>_)
```
JacquesCarette, line 19: While I like the document aspect, that is independent of a PR that is mean to simplify dependencies, so this change should be reverted.

src/Text/Tabular/Vec.agda
```diff
@@ -10,7 +10,8 @@ module Text.Tabular.Vec where
 
 open import Data.List.Base using (List)
 open import Data.Product as Prod using (uncurry)
-open import Data.String using (String; rectangle; fromAlignment)
+open import Data.String.Base using (String; fromAlignment)
+open import Data.String using (rectangle)
```
JacquesCarette, line 14: this one should definitely be reverted.

JacquesCarette (changes requested):

src/Reflection/AST/Show.agda
JacquesCarette: My mistake, I somehow didn't see the `.Base` here, I must have been reading too fast.

jamesmckinna: Not sure the change is an improvement *per se* to trade one qualified import using 5-6 names, in favour of two qualified imports using all-but-one of the previous names, plus a second import using one of them? It's greater precision, to be sure, but not obviously greater clarity, and I'm not sure how it adds up in terms of build times?

Taneb: @jamesmckinna `Data.String` has a whole heap of module dependencies that neither `Data.String.Base` nor `Data.String.Properties` have, such as `Data.Nat.Properties` and `Data.Vec.Base`

gallais (approved):

JacquesCarette (approved):
I think this set of changes is worth pulling in.

PR #2017, opened 2023-07-09
[ new ] notions of finiteness

laMudri:
This PR contains a rough draft of some definitions of constructively well behaved notions of finiteness for setoids, based on [this math.se answer](https://math.stackexchange.com/questions/4285348/is-there-a-definition-of-finite-sets-such-that-it-can-be-used-to-constructively/4285367#4285367). Included are the proofs that the two definitions of *subfinitely enumerable* are logically equivalent (`lemma→` and `lemma←`). It is a fairly direct transcription, perhaps meaning that we miss out on idiomatic uses of lists rather than `Fin`.

Clear problems to be resolved before this is merged:
- The whole development is in a single file.
- [ ] Some of the earlier definitions should be dispersed throughout the library.
- [ ] A place should be found in the stdlib hierarchy for a `Finite` module.
- Some of the names are bad/questionable.
- [ ] `SubFinitelyEnumerable` and `SubfinitelyEnumerable` differ only in case (analogous to the presence/absence of a hyphen in the reference answer).
- [ ] `lemma→`, `lemma←`
- [ ] Maybe `inj` and `srj` fields should not have abbreviated names.
- [ ] Direct image, `_[_]`, has a name which might cause clashes. I don't use it here, but I thought I'd include it together with the inverse image `_[_]⁻¹`. These two should be named as a pair.
- [ ] Setoid quotient, `_/_`, might also clash.
- General code quality problems:
- [ ] Some of the supporting definitions may have duplicates in the library. I didn't check carefully.
- [ ] `Subset` and `EquivalenceRelation` are conceptually related, but are named and implemented completely differently. We should think carefully about how we want to deal in general with predicates/relations which respect setoid equality.
- [ ] Some of the proofs are probably too inlined (particularly extended λ-abstractions).
- [ ] The definition of the equivalence relation `R` is a bit gnarly, with its 5-place iterated Σ-type. I'm following the reference answer here, but maybe it (together with the resulting quotient) could be defined in a more symmetric way, reflecting the symmetry of the pushout construction being performed.
- [ ] I'm not making use of any particulars of `Fin`, so maybe the whole development could/should be generalised to arbitrary cardinalities.

Related PR: #2005 
cc @JacquesCarette @Sofia-Insa

### DIFF
--- a/src/Function/Bundles.agda
+++ b/src/Function/Bundles.agda
@@ -94,6 +94,12 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
       cong       : to Preserves _≈₁_ ⟶ _≈₂_
       surjective : Surjective to
 
+    function : Func
+    function = record
+      { to   = to
+      ; cong = cong
+      }
+
     to⁻ : B → A
     to⁻ = proj₁ ∘ surjective
 
--- a/src/Finite.agda
+++ b/src/Finite.agda
@@ -1,3 +1,11 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Notions of finiteness for setoids
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
 module Finite where
 
 open import Data.Fin.Base using (Fin)
--- a/src/Finite.agda
+++ b/src/Finite.agda
@@ -3,8 +3,7 @@ module Finite where
 open import Data.Fin.Base using (Fin)
 open import Data.Nat.Base using (ℕ)
 open import Data.Product.Base as ×
-open import Data.Product.Relation.Binary.Pointwise.NonDependent using (×-setoid)
-open import Data.Sum as ⊎ using (_⊎_; inj₁; inj₂)
+open import Data.Sum.Base as ⊎ using (_⊎_; inj₁; inj₂)
 open import Data.Unit using (⊤; tt)
 open import Function
 open import Level renaming (suc to lsuc)
@@ -33,9 +32,6 @@ Subset X p = Func X (Ω p)
 
 FullSubset : {X : Setoid c ℓ} → Subset X 0ℓ
 FullSubset = record { to = λ _ → ⊤ }
-
-Relation : Setoid c ℓ → (r : Level) → Set (c ⊔ ℓ ⊔ lsuc r)
-Relation X r = Func (×-setoid X X) (Ω r)
 
 record EquivalenceRelation (X : Setoid c ℓ) r : Set (c ⊔ ℓ ⊔ lsuc r) where
   infix 4 _∼_
@@ -124,25 +120,25 @@ record FinitelyEnumerable (X : Setoid c ℓ) : Set (c ⊔ ℓ) where
 record SubFinitelyEnumerable (X : Setoid c ℓ) c′ ℓ′
        : Set (c ⊔ ℓ ⊔ lsuc (c′ ⊔ ℓ′)) where
   field
-    F : Setoid c′ ℓ′
-    finitelyEnumerable : FinitelyEnumerable F
-    inj : Injection X F
+    Apex : Setoid c′ ℓ′
+    finitelyEnumerable : FinitelyEnumerable Apex
+    inj : Injection X Apex
 
   open FinitelyEnumerable finitelyEnumerable public
 
 record SubfinitelyEnumerable (X : Setoid c ℓ) c′ ℓ′
        : Set (c ⊔ ℓ ⊔ lsuc (c′ ⊔ ℓ′)) where
   field
-    F : Setoid c′ ℓ′
-    subfinite : Subfinite F
-    srj : Surjection F X
+    Apex : Setoid c′ ℓ′
+    subfinite : Subfinite Apex
+    srj : Surjection Apex X
 
   open Subfinite subfinite public
 
 lemma→ : {X : Setoid c ℓ} →
   SubFinitelyEnumerable X c′ ℓ′ → SubfinitelyEnumerable X (c ⊔ ℓ′) 0ℓ
 lemma→ {X = X} sfe = record
-  { F = e.function ⁻¹[ ↣⇒⊆ inj ]
+  { Apex = e.function ⁻¹[ ↣⇒⊆ inj ]
   ; subfinite = record
     { size = size
     ; inj = record
@@ -154,27 +150,27 @@ lemma→ {X = X} sfe = record
   ; srj = record
     { to = λ (i , x , q) → x
     ; cong = λ {(i , x , q)} {(i′ , x′ , q′)} p →
-      let open SetR F in m.injective $ begin
+      let open SetR Apex in m.injective $ begin
       m.to x   ≈⟨ q ⟩
       e.to i   ≡⟨ ≡.cong e.to p ⟩
       e.to i′  ≈˘⟨ q′ ⟩
       m.to x′  ∎
     ; surjective = λ x →
-      (e.to⁻ (m.to x) , x , F.sym (e.surjective (m.to x) .proj₂)) , X.refl
+      (e.to⁻ (m.to x) , x , A.sym (e.surjective (m.to x) .proj₂)) , X.refl
     }
   }
   where
-  -- X --m--> F <--e-- Fin size
+  -- X --m--> Apex <--e-- Fin size
   open SubFinitelyEnumerable sfe
   module X = Setoid X
-  module F = Setoid F
+  module A = Setoid Apex
   module m = Injection inj
   module e = Surjection srj
 
 lemma← : {X : Setoid c ℓ} →
   SubfinitelyEnumerable X c′ ℓ′ → SubFinitelyEnumerable X 0ℓ (ℓ ⊔ c′)
 lemma← {X = X} se = record
-  { F = ≡.setoid (Fin size) / R
+  { Apex = ≡.setoid (Fin size) / R
   ; finitelyEnumerable = record
     { size = size
     ; srj = include-/ R
@@ -203,17 +199,17 @@ lemma← {X = X} se = record
     }
   }
   where
-  -- X <--e-- F --m--> Fin size
+  -- X <--e-- Apex --m--> Fin size
   open SubfinitelyEnumerable se
   module X = Setoid X
-  module F = Setoid F
+  module A = Setoid Apex
   module m = Injection inj
   module e = Surjection srj
 
   R : EquivalenceRelation (≡.setoid (Fin size)) _
   R = record
     { _∼_ = λ i j → i ≡ j ⊎
-      ∃ \ ((f , g) : F.Carrier × F.Carrier) →
+      ∃ \ ((f , g) : A.Carrier × A.Carrier) →
         m.to f ≡ i × m.to g ≡ j × e.to f X.≈ e.to g
     ; ∼-resp-≈ = λ { ≡.refl ≡.refl → id }
     ; ∼-isEquivalence = record
--- a/src/Finite.agda
+++ b/src/Finite.agda
@@ -1,0 +1,233 @@
+module Finite where
+
+open import Data.Fin.Base using (Fin)
+open import Data.Nat.Base using (ℕ)
+open import Data.Product.Base as ×
+open import Data.Product.Relation.Binary.Pointwise.NonDependent using (×-setoid)
+open import Data.Sum as ⊎ using (_⊎_; inj₁; inj₂)
+open import Data.Unit using (⊤; tt)
+open import Function
+open import Level renaming (suc to lsuc)
+open import Relation.Binary using (Rel; Setoid; IsEquivalence)
+import Relation.Binary.Reasoning.Setoid as SetR
+import Relation.Binary.Construct.On as On
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
+
+private
+  variable
+    c ℓ c′ ℓ′ : Level
+
+Ω : ∀ p → Setoid (lsuc p) p
+Ω p = record
+  { Carrier = Set p
+  ; _≈_ = λ P Q → (P → Q) × (Q → P)
+  ; isEquivalence = record
+    { refl = id , id
+    ; sym = swap
+    ; trans = λ (f , g) (f′ , g′) → f′ ∘ f , g ∘ g′
+    }
+  }
+
+Subset : Setoid c ℓ → (p : Level) → Set (c ⊔ ℓ ⊔ lsuc p)
+Subset X p = Func X (Ω p)
+
+FullSubset : {X : Setoid c ℓ} → Subset X 0ℓ
+FullSubset = record { to = λ _ → ⊤ }
+
+Relation : Setoid c ℓ → (r : Level) → Set (c ⊔ ℓ ⊔ lsuc r)
+Relation X r = Func (×-setoid X X) (Ω r)
+
+record EquivalenceRelation (X : Setoid c ℓ) r : Set (c ⊔ ℓ ⊔ lsuc r) where
+  infix 4 _∼_
+  open Setoid X
+  field
+    _∼_ : Rel Carrier r
+    ∼-resp-≈ : ∀ {x x′ y y′} → x ≈ x′ → y ≈ y′ → x ∼ y → x′ ∼ y′
+    ∼-isEquivalence : IsEquivalence _∼_
+
+  open IsEquivalence ∼-isEquivalence public renaming
+    ( refl                 to ∼-refl
+    ; trans                to ∼-trans
+    ; sym                  to ∼-sym
+    ; reflexive            to ∼-reflexive
+    ; isPartialEquivalence to ∼-isPartialEquivalence
+    )
+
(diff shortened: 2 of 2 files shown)


Saransh-cpp:
Some thoughts (or queries) -
- #863 recommends using the name `StronglyFinite`. Should it be named `StronglyFinite` or `StrictlyFinite`?
- #2005 recommends creating `IsStronglyFinite` as the predicate with n as a parameter and `StronglyFinite` as the Bundled version. Should I still do that?
- #2005 and #863 recommend the file should go in `Relation.Nullary.Finite.Setoid` with a corresponding `Propositional` module. Should we have the `Finite` and `Enumerable` records in `Relation.Nullary.Finite.Setoid` and the lemmas in `Propositional`?

More work here (that I will be carrying out)
- Closure over sum and product
- toList, toVec, toVector functions

I am supposed to take over this and #2005, but I cannot directly push to this branch. How should I proceed ahead with this? Should I create a new PR with @laMudri's code and then work on it with reviews? I think I will start with the checklist mentioned above and ask for help on the way 🙂 

cc: @JacquesCarette

laMudri:
Hi! You must be Jacques' student. I'm happy to give you push access to this branch, in case that's the easiest solution. As for your points:

1. Yeah, now you mention it, I like `StronglyFinite`. I think it is indeed the strongest notion of finiteness, while “strict” can mean other things (e.g, I think a “strict setoid” is a setoid whose equivalence relation is propositional equality).
2. It fits a common pattern to have these unbundled versions, but given that `IsStronglyFinite X n = Inverse X (≡.setoid (Fin n))`, are we gaining anything by introducing such a definition? Particularly, in view of my last to-do item (if it works out), we just end up with aliases of `Inverse`, `Injection`, and `Surjection`. Perhaps these would be beneficial for the subfinitely enumerable definitions, though. Going the other way, it might be worth creating *more* bundled definitions, where `X` too is a field.
3. Yeah, looking around the previous work, there seems to be a consensus around the `Relation.Nullary` hierarchy. I don't fully understand why, but it seems plausible. I guess that's where `Dec` lives. I thought anyway there was promise that these `Nullary`/`Unary`/`Binary`/c hierarchies would be done away with soon.
4. As for a `Propositional` module, the earlier definitions would adapt fine to `Set`s (just precompose with `≡.setoid`), but the two subfinitely enumerable definitions would have problems. Both mention the existence of some `Apex` *setoid*, so even if you start out with a set and finish with a `Fin n`, you have to go through setoids. This use of setoids is crucial for the equivalence proofs, because we need quotients to produce the apex in both directions. Probably I would omit the subfinitely enumerable definitions from a `Propositional` module, because they don't behave as intended.
5. For the last question, do you mean `Properties` at the end? Personally, I think modules like `Data.List.Relation.Unary.All.Properties` are filled with things that aren't properties of `All`, and instead are just the basic constructions of `All` (despite being property-like for `List`). This leaves properties of `All` (e.g, equations between expressions of `All` type) with nowhere to go except this really big `Properties` module. If I can get any sympathy for this view, then I think the equivalences and constructions should go either in the same module as the definitions, or in some sort of `Construct` module.
6. The extra work all sounds like worthwhile things to do. I wonder whether `toList` gives rise to equivalence proofs with alternative definitions of finiteness in terms of lists.
7. I guess some low-hanging fruit is to provide coercions from stronger notions to weaker notions, if you can think of some reasonable names.

Saransh-cpp:
Hi, yes! Thanks for the detailed discussion and the invite to the repository! I've started working on this locally.

> It fits a common pattern to have these unbundled versions, but given that IsStronglyFinite X n = Inverse X (≡.setoid (Fin n)), are we gaining anything by introducing such a definition? Particularly, in view of my last to-do item (if it works out), we just end up with aliases of Inverse, Injection, and Surjection. Perhaps these would be beneficial for the subfinitely enumerable definitions, though. Going the other way, it might be worth creating more bundled definitions, where X too is a field.

I don't really have an opinion here. I'll leave this up to you and other maintainers 🙂 

> I thought anyway there was promise that these Nullary/Unary/Binary/c hierarchies would be done away with soon.

Oh, I see. A lot of API changes before 2.0.0

> As for a Propositional module, the earlier definitions would adapt fine to Sets (just precompose with ≡.setoid), but the two subfinitely enumerable definitions would have problems. Both mention the existence of some Apex setoid, so even if you start out with a set and finish with a Fin n, you have to go through setoids. This use of setoids is crucial for the equivalence proofs, because we need quotients to produce the apex in both directions. Probably I would omit the subfinitely enumerable definitions from a Propositional module, because they don't behave as intended.

What exactly is supposed to in the Propositional module? Is it supposed to have a different version of `StronglyFinite` which takes a `Set` instead of a `Setoid`? I'm not very sure I understood this very well.

> For the last question, do you mean Properties at the end?

Yes, my bad, I meant `Properties`. I can maybe start working on them in the `Properties` and see how it goes with the reviews.

JacquesCarette:
- Re `IsStronglyFinite`: we actually gain quite a lot of type inference by making it a 1-constructor `data` (less sure about 1-field `record` ?). The extra indirection makes operating over it a bit more annoying. I think the original suggestion was to make these uniform with respect to the other algebraic structures in Agda. 
- Not a big fan of `Nullary` either, but that is how 2.0 is being done, so we go along.
- `Propositional` is short for `PropositionalEquality` and is the version that hard-wires `≡.setoid`.

I will also comment on the current code regarding a number of other items.

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,148 @@
...
+import Relation.Binary.PropositionalEquality.Properties as ≡
+
+private
+  variable
+    c ℓ c′ ℓ′ : Level
+
+Ω : ∀ p → Setoid (lsuc p) p
+Ω p = record
+  { Carrier = Set p
+  ; _≈_ = λ P Q → (P → Q) × (Q → P)
```
JacquesCarette, line 31: This function is [_⇔_](https://agda.github.io/agda-stdlib/Function.Bundles.html#8934). And I just notice a bug: this very same definition is repeated as [_⇔_](https://agda.github.io/agda-stdlib/Function.Equivalence.html#1213) (same name, entirely different file).

In the latter, we also find [⇔-setoid](https://agda.github.io/agda-stdlib/Function.Equivalence.html#2586) which is Omega.

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,148 @@
...
+  ; isEquivalence = record
+    { refl = id , id
+    ; sym = swap
+    ; trans = λ (f , g) (f′ , g′) → f′ ∘ f , g ∘ g′
+    }
+  }
+
+Subset : Setoid c ℓ → (p : Level) → Set (c ⊔ ℓ ⊔ lsuc p)
```
JacquesCarette, line 39: Surely something like these two (`Subset` and `FullSubset`) already exist in stdlib elsewhere? Or should, independently.

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,148 @@
...
+Subset X p = Func X (Ω p)
+
+FullSubset : {X : Setoid c ℓ} → Subset X 0ℓ
+FullSubset = record { to = λ _ → ⊤ }
+
+record EquivalenceRelation (X : Setoid c ℓ) r : Set (c ⊔ ℓ ⊔ lsuc r) where
```
JacquesCarette, line 45: I'm quite sure that all the definitions from `EquivalenceRelation` to `include-/` do not belong in this file (or even this PR). Some of them likely already exist while others are non-trivial new items that should be in a separate PR.

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,148 @@
...
+include-/ {X = X} R = record
+  { to = id
+  ; cong = ≈⇒∼
+  ; surjective = λ y → y , ∼-refl
+  }
+  where
+  open Setoid X
+  open EquivalenceRelation R
+
+record IsStronglyFinite (X : Setoid c ℓ) (n : ℕ) : Set (c ⊔ ℓ) where
```
JacquesCarette, line 114: Above these, there should be comments as to the origin of the names (with links).

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,148 @@
...
+  field
+    size : ℕ
+    inj : Injection X (≡.setoid (Fin size))
+
+record FinitelyEnumerable (X : Setoid c ℓ) : Set (c ⊔ ℓ) where
+  field
+    size : ℕ
+    srj : Surjection (≡.setoid (Fin size)) X
+
+record SubFinitelyEnumerable (X : Setoid c ℓ) c′ ℓ′
```
JacquesCarette, line 132: The 'source literature' here is silly with its names, and this is copied here.

`InjectsIntoFinitelyEnumerable` ? And then `SurjectionFromFinitelyEnumerable` ?

JacquesCarette (changes requested):

src/Relation/Nullary/Finite/Setoid.agda
Saransh-cpp: I think `Function.Equivalence` is marked for deprecation in v2.0.0; hence, the repetition?

src/Function/Bundles.agda
```diff
@@ -94,6 +94,12 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
       cong       : to Preserves _≈₁_ ⟶ _≈₂_
       surjective : Surjective to
 
+    function : Func
```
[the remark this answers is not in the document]
laMudri: Yeah, I'd be happy to have a separate PR for this. It seemed like a bug-fix, given that all of the surrounding records have a similar definition.

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: Indeed, I missed that part - well spotted!

laMudri:
> * Re `IsStronglyFinite`: we actually gain quite a lot of type inference by making it a 1-constructor `data` (less sure about 1-field `record` ?). The extra indirection makes operating over it a bit more annoying. I think the original suggestion was to make these uniform with respect to the other algebraic structures in Agda.

In this case, `≡.setoid (Fin _)` is already definitionally injective (see the snippet below), so there is no gain when it comes to unification.

```agda
open import Data.Fin

test : ∀ n → ≡.setoid (Fin n) ≡ ≡.setoid (Fin _)
test n = refl
```

src/Relation/Nullary/Finite/Setoid.agda
Saransh-cpp: These are used in `lemma→` and `lemma←`. Should these also go in a different PR?

JacquesCarette: I think so. It might well turn out that you'll need a preliminary PR or two to set up the groundwork for this one.

Saransh-cpp:
I think it would be better to keep this PR intact and create small subset PRs. I have created #2022 to add the relevant finiteness records by picking up the first commit.

src/Relation/Nullary/Finite/Setoid.agda
laMudri: Yeah, `Function.Equivalence`, `Function.Injection`, &c have been pre-deprecated for a long time (since the inception of `Function.Bundles`, probably). Both definitions of `_⇔_` are unfortunate in that the congruence fields become entirely redundant, which is a (post hoc) justification for this much simpler `(P → Q) × (Q → P)` type. As for `⇔-setoid`, maybe I'm biased by the particular use of it here, but it seems like a terrible name. It's not just some random setoid made out of the notion of equivalence; it's the setoid of propositions.

src/Relation/Nullary/Finite/Setoid.agda
Saransh-cpp: I could not find an origin for the name `StronglyFinite`. There are a few references that define finite sets or a notion of finiteness but I could not see any of them using `StronglyFinite` in particular -

- [Dependently typed programming with finite sets ](https://dl.acm.org/citation.cfm?id=2808102)
- [Brent's Yorgey's PhD thesis](https://github.com/byorgey/thesis)
- [Towards Certifiable Implementation of Graph Transformation via Relation Categories](https://link.springer.com/chapter/10.1007/978-3-642-33314-9_6)

Which one should I cite? Or should it be the GitHub comment where `StronglyFinite` popped up for the first time?

src/Relation/Nullary/Finite/Setoid.agda
Saransh-cpp: There is `Data.Fin.Subset` but the definition is very different.

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: Probably `FinitelyEnumerable` is closer to what's used out there.

Saransh-cpp: Loking at the references, [Dependently typed programming with finite sets](https://dl.acm.org/citation.cfm?id=2808102) introduces "finitely enumerable" for the first time. It classifies the notion of finiteness using some other references -

> Intuitionistic frameworks give rise to a rich variety of notions of finiteness that collapse classically. In [[8]](https://inria.hal.science/inria-00503917/file/Coquand_Spiwack_-_2010_-_Constructively_Finite.pdf), [[5]](https://arxiv.org/pdf/1210.1200.pdf) and [[14]](https://www.irif.fr/~letouzey/types2014/slides-39.pdf), the author describe various concepts of finiteness and their interrrelation. According to their classification, this paper focuses on the strongest notion of finiteness, namely, finitely enumerable (listable) sets.

[[8]](https://inria.hal.science/inria-00503917/file/Coquand_Spiwack_-_2010_-_Constructively_Finite.pdf) talks about "Enumerated" sets which I found to be the closest here.

> A set A will be said to be enumerated, or simply finite, noted A ε F when there is a list of all its elements. We call such a list an enumeration of A.

I also found https://github.com/agda/agda-finite-prover, which looks somewhat relevant here.

JacquesCarette: `Enumerated` is another excellent name.

laMudri: Spiwack and Coqand unfortunately use rather vague wording in our definition of interest:

> **Definition**
> A set *A* will be said to be *enumerated*, or simply finite, noted *A* *ε* 𝓕 when there
> is a list of all its elements. We call such a list an *enumeration* of *A*.

I *think* this means ∃*xs* : List *A*. ∀*x* : *A*. *x* ∈ *xs*, which allows for duplicates in the list. This is equivalent to `FinitelyEnumerable` from this PR:

**Proof**
(enumerated → `FinitelyEnumerable`): Let `size` be the length of the list *xs*. The function is given by `lookup`. This is surjective by the fact that for every *x*, we have an *e* : *x* ∈ *xs*, and we can take the index of *e*, which `lookup` maps to *x*.
(`FinitelyEnumerable` → enumerated): The list is given by the tabulation of the surjective function. Given an *x* : *A*, it is in the tabulation by surjectivity.

I think this discussion, if anything, motivates the name `StronglyFinite`. It is stronger than all of these other commonly studied notions of finiteness, and is perhaps the strongest notion we could reasonably call “finiteness”. Definitely, if we had univalence, then any property stronger than “∃ `n`. it's equivalent to `Fin n`” precludes the type `Fin n` for some `n`. If `StronglyFinite` is strictly stronger than the other notions of finiteness being discussed, we could ask whether it's worth having in the library at all. I think it is. For example, when people talk about categories whose objects are finite sets, I think they nearly always mean strongly finite sets. Also, I can imagine that strong finiteness is too trivial to mention in pen-and-paper mathematics, because you could just say Fin *n* instead, but I think people genuinely do want to talk about sets which are strongly finite in Agda.

Anyway, it's important that we get these things right, so thanks for doing the research. I don't have a good enough intuition on the three later (weaker) definitions from the same paper, but hopefully one of them is equivalent to being subfinitely enumerable (if I had to put money on it, I'd go with bounded size, because it seems to have similar properties). I guess it would be good to have all of these list-based definitions in the library too, connected with the appropriate equivalence/inclusion proofs to the definitions already in this PR. It would help get rid of some definitions without proofs.

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: I agree, but I don't want this to explode any more than it already has!

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: The names might be quite different.

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: I certainly very happy to stick to `StronglyFinite` !

jamesmckinna:
I think it was right to turn #2022 back to DRAFT, but probably a mistake on my part to leave comments there, rather than a more substantial review of the material here, esp. eg the discussion of `Apex` etc.

JacquesCarette:
We're also closing this one, for now, in favour of project issue #2511  as it's been hanging idle for too long and still requires a lot of work.

PR #2020, opened 2023-07-13
Additional lemmas about lists and vectors

amblafont:


src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -62,6 +63,16 @@ module _ (S : Setoid c ℓ) where
   ∉-resp-≋ : ∀ {x} → (x ∉_) Respects _≋_
   ∉-resp-≋ xs≋ys v∉xs v∈ys = v∉xs (∈-resp-≋ (≋-sym xs≋ys) v∈ys)
 
+  -- index is injective in its first argument.
+
+  index-injective
+    : ∀ {x₁ x₂ xs} (x₁∈xs : x₁ ∈ xs) (x₂∈xs : x₂ ∈ xs) →
+      Any.index x₁∈xs ≡ Any.index x₂∈xs → x₁ ≈ x₂
```
gallais, line 70: This does not respect the style guide, cf.:
https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md#indentation

gallais (commented):
This looks nice. The only issue AFAICT is the indentation.

src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -62,6 +63,16 @@ module _ (S : Setoid c ℓ) where
...
+  index-injective (here x₁≈x)   (here x₂≈x)   _  = trans x₁≈x (sym x₂≈x)
+  index-injective (there x₁∈xs) (there x₂∈xs) eq = index-injective x₁∈xs x₂∈xs (suc-injective eq)
+  index-injective (here _)  (there _) ()
+  index-injective (there _) (here _)  ()
```
MatthewDaggitt, line 74: If you remove these two lines, is Agda smart enough to recognise that they are not possible?

src/Data/Vec/Membership/Propositional/Properties.agda
```diff
@@ -13,7 +13,7 @@ open import Data.Product as Prod using (_,_; ∃; _×_; -,_)
 open import Data.Vec hiding (here; there)
 open import Data.Vec.Relation.Unary.Any using (here; there)
 open import Data.List.Base using ([]; _∷_)
-open import Data.List.Relation.Unary.Any using (here; there)
+open import Data.List.Relation.Unary.Any using (here; there) renaming (index to indexₗ)
```
MatthewDaggitt, line 16: Could we use module imports rather than renaming? i.e. `open import Data.List.Relation.Unary.Any as ListAny using (here; there)` instead?

src/Data/Vec/Properties.agda
```diff
@@ -509,6 +520,11 @@ lookup-splitAt (suc m) (x ∷ xs) ys (suc i) = trans
   (lookup-splitAt m xs ys i)
   (sym ([,]-map (Fin.splitAt m i)))
 
+toList-++ : (xs : Vec A n)(ys : Vec A m) →
```
MatthewDaggitt, line 523: Very minor, but here and in the proofs below, can you add spaces in between the arguments, i.e. `(xs : Vec A n) (ys : Vec A m)` rather than `(xs : Vec A n)(ys : Vec A m)`

MatthewDaggitt (changes requested):
Other than those comments, this looks great. Thanks for the awesome first PR :+1:

shhyou:
Since `toList-map`, `toList-++` and `toList-replicate` are about the relationship between `toList` and other functions, can they be placed together after `toList∘fromList`? I have a set of lemmas about `fromList-` and other functions in #2045, so it'd be nice for them to stay close to each other.

MatthewDaggitt:
> Since toList-map, toList-++ and toList-replicate are about the relationship between toList and other functions, can they be placed together after toList∘fromList? I have a set of lemmas about fromList- and other functions in https://github.com/agda/agda-stdlib/pull/2045, so it'd be nice for them to stay close to each other.

The problem is that the opposite argument could also apply, e.g. all the `map` related lemmas should stay together. I think I'd prefer not to mess about with the ordering of the file unless there's a compelling reason and we do it across the library.

MatthewDaggitt (approved):

MatthewDaggitt:
Thanks once again for the PR

PR #2022, opened 2023-07-13
[ new ] notions of finiteness

Saransh-cpp:
A subset of #2017 containing minimal changes related to finiteness

### DIFF
--- a/src/Relation/Nullary/Finite/Setoid.agda
+++ b/src/Relation/Nullary/Finite/Setoid.agda
@@ -19,10 +19,13 @@ private
   variable
     c ℓ c′ ℓ′ : Level
 
-record StrictlyFinite (X : Setoid c ℓ) : Set (c ⊔ ℓ) where
+record IsStronglyFinite (X : Setoid c ℓ) (n : ℕ) : Set (c ⊔ ℓ) where
    field
-     size : ℕ
-     inv : Inverse X (setoid (Fin size))
+     inv : Inverse X (setoid (Fin n))
+
+record StronglyFinite (X : Setoid c ℓ) (n : ℕ) : Set (c ⊔ ℓ) where
+   field
+     isFinite : IsStronglyFinite X n
 
 record Subfinite (X : Setoid c ℓ) : Set (c ⊔ ℓ) where
   field
--- a/src/Relation/Nullary/Finite/Setoid.agda
+++ b/src/Relation/Nullary/Finite/Setoid.agda
@@ -1,0 +1,58 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Notions of finiteness for setoids
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Relation.Nullary.Finite.Setoid where
+
+open import Data.Fin.Base using (Fin)
+open import Data.Nat.Base using (ℕ)
+open import Data.Product.Base as ×
+open import Data.Sum.Base as ⊎ using (_⊎_; inj₁; inj₂)
+open import Data.Unit using (⊤; tt)
+open import Function
+open import Level renaming (suc to lsuc)
+open import Relation.Binary using (Rel; Setoid; IsEquivalence)
+import Relation.Binary.Reasoning.Setoid as SetR
+import Relation.Binary.Construct.On as On
+open import Relation.Binary.PropositionalEquality as ≡ using (_≡_)
+
+private
+  variable
+    c ℓ c′ ℓ′ : Level
+
+record StrictlyFinite (X : Setoid c ℓ) : Set (c ⊔ ℓ) where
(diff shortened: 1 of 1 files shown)


laMudri:
This PR on its own fails the “definitions should come with proofs” test, but maybe it's permissible with the promise of proofs in a later PR. We'll have to let the maintainers decide.

Saransh-cpp:
Yes, that is very true. I just don't how to include the proofs without bringing in a lot more changes (which should be in separate PRs according to the discussions in #2017).

jamesmckinna:
(Overlooked in yesterday's meeting)
Suggest at the very least that we bump this to v2.2, or even close until/unless someone comes along able to pick up the slack?

JacquesCarette:
I put it in v2.2 as I really would like to push this through.

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,56 @@
...
+  field
+    size : ℕ
+    srj : Surjection (setoid (Fin size)) X
+
+record InjectsIntoFinitelyEnumerable (X : Setoid c ℓ) c′ ℓ′
+       : Set (c ⊔ ℓ ⊔ lsuc (c′ ⊔ ℓ′)) where
+  field
+    Apex : Setoid c′ ℓ′
+    finitelyEnumerable : FinitelyEnumerable Apex
+    inj : Injection X Apex
```
jamesmckinna, line 45: @JacquesCarette I can *just about* take `Apex` as the name of the 'covering object' `A` in an epic situation `A → X` (L52-L54 below), but it seems less defensible here in the dual situation `X → A`. Is it 'standard' terminology? Or is there something better from eg the categorical literature?

src/Relation/Nullary/Finite/Setoid.agda
```diff
@@ -0,0 +1,56 @@
...
+
+  open FinitelyEnumerable finitelyEnumerable public
+
+record SurjectionFromFinitelyEnumerable (X : Setoid c ℓ) c′ ℓ′
+       : Set (c ⊔ ℓ ⊔ lsuc (c′ ⊔ ℓ′)) where
+  field
+    Apex : Setoid c′ ℓ′
+    subfinite : Subfinite Apex
+    srj : Surjection Apex X
```
jamesmckinna, line 54: See above. I'd prefer a 'better' name, even though I appreciate it's useful to have a field name that is clearly 'standardised apart' from any potential clashes...

... ob: which, will `Apex` ever be mentioned explicitly in use of these records, or could it be declared implicit, as it's governed by the more-informative subsequent field name in each case?

jamesmckinna:
> I put it in v2.2 as I really would like to push this through.

Comments on notation. I'd like to see this too, but as a fresh addition, we should try to get a 'convenient'/usable/memorable syntax/naming convention fixed before committing...

src/Relation/Nullary/Finite/Setoid.agda
laMudri: I suppose being able to answer the latter question is why PRs normally come with proofs. They'd give some idea as to what the relevant idioms are.

For what it's worth, I would guess that `Apex` often does have to be given and referred to explicitly. At least in basic cases, I expect that a lot of the work will go into finding the right setoid, and then backward-chaining all of the required maps.

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: Yes, this is probably too influenced by other code / ideas, and no longer really fits here. Suggestions?

src/Relation/Nullary/Finite/Setoid.agda
JacquesCarette: I agree that PRs that introduce new ideas should not be so minimalistic. (My preference for small PRs was pushed too far here.)

src/Relation/Nullary/Finite/Setoid.agda
jamesmckinna: I hunted around, and unfortunately the literature uses 'generic' names such as `X`, `Y`/`A`, `B` when discussing things as morphisms. For embedding/extension, the domain of an injection is sometimes called `I`, and of a projection as `P`, but these don't seem quite satisfactory, either?

In each case, though, the relevant object is (suitably) enumerable, so could it be called `Enum`, or even `Enumeration`?

JacquesCarette: `Enumeration` > `Enum` > `E`.

MatthewDaggitt:
I've converted this to a draft. @JacquesCarette when it's ready to review properly, turn it back again :+1:

JacquesCarette:
We've decided to close this for now, in favour of the "project issue #2511 " as it's a bigger deal.

PR #2023, opened 2023-07-17
[ new ] associativity of Appending

laMudri:
This PR adds a few new lemmas about the `Appending` relation on lists. The main results are the two associativity implications, `assoc→` and `assoc←`. Potential problems:

- Bad names: `unzip`, `through{←,→}`.
- - `unzip` is a kind of inverse to `transitive` for `Pointwise`.
- - The `through` lemmas just turn up in the base cases of the `assoc` lemmas.
- There is a repeated pattern `∃[ y ] P y × Q y` throughout. In my experience, this pattern is pervasive in relational programming. It is tempting to factor these in terms of relation composition (because that's what they are, fundamentally), but it's notationally a hassle to turn a ternary relation like `Appending` into the correct binary relation.
- `trans⁻¹` makes an arbitrary choice to use witness `z` rather than `x`. With `Appending` being asymmetric, it feels like one choice should be better than the other, but I haven't thought about which one that is.


laMudri:
I was having a look through some related properties, and it seems that the situation with `_++[]` (one direction of the right identity law) is unideal. `_++[]` is only defined in `...Propositional`, despite the fact that it has a simple analogue in the heterogeneous case, listed below:

```agda
_++[] : Pointwise L as cs → Appending L R as [] cs
[] ++[] = []++ []
(l ∷ ls) ++[] = l ∷ (ls ++[])
```

Also, back in the propositional case, we have the anomaly that `[]++_ : Pointwise _≡_ bs cs → Appending [] bs cs` (that's a constructor of `Appending` inherited from the heterogeneous and setoid cases), but `_++[] : ∀ as → Appending as [] as`.

It seems to me that the point of the `Propositional` section is to replace `Pointwise _≡_` by `_≡_` (otherwise, I guess you could just directly use the `Setoid` section instantiated to the ≡-setoid). But it's impossible to do this for a constructor, so we end up stuck with compromises.

jamesmckinna:
Agree regarding relational composition: where that's all you need, please use it (and save on `syntax` imports...). Fir ternary relations like `Appending`, I agree it's annoying, although I'd love it if skilful use of sections might solve the problem? (Probably not, but please consider...)

MatthewDaggitt:
Gone through and refactored, using composition where appropriate. Like @laMudri I can't think of a good way to squeeze the `Appending` relation into a general shape that fits composition so left as is. Propose to merge if someone else is willing to approve?

src/Relation/Binary/Properties/Setoid.agda
```diff
@@ -77,6 +79,15 @@ preorder = record
...
 
+------------------------------------------------------------------------
+-- Equality is closed under composition
+
+≈;≈⇒≈ : _≈_ ; _≈_ ⇒ _≈_
+≈;≈⇒≈ (_ , p , q) = trans p q
+
+≈⇒≈;≈ : _≈_ ⇒ _≈_ ; _≈_
+≈⇒≈;≈ q = _ , q , refl
+
```
jamesmckinna, line 90: I would be _strongly_ tempted to refactor these proofs, either in this PR, or subsequently, in terms of properties of general binary relations:
* any reflexive relation `R` (of the correct type) satisfies `S ⇒ R ; S` and `T ⇒ T ; R`;
* any transitive relation `T` satisfies `T ; T ⇒ T`;
* `Setoid`/`IsEquivalence`, `Preorder/`IsPreorder` satisfy reflexivity and transitivity: QED

Specialising to `Setoid` here seems ... a missed opportunity.

jamesmckinna:
Haven't done a full review, but have a left a 'review'-like comment.

src/Relation/Binary/Properties/Setoid.agda
MatthewDaggitt: My aim was to avoid having to come up with names for these :sweat_smile:

jamesmckinna: Ouch!

JacquesCarette: Seems a shame to hold up this PR for this refactor - though it sure would be nice to do the refactor before the release so that no deprecations are needed.

src/Data/List/Relation/Binary/Pointwise/Base.agda
```diff
@@ -8,16 +8,16 @@
 
 module Data.List.Relation.Binary.Pointwise.Base where
 
-open import Data.Product.Base using (_×_; <_,_>)
+open import Data.Product.Base as Σ using (_×_; _,_; <_,_>; ∃-syntax)
```
jamesmckinna, line 11: Do we leave this, or bring it into line with #2280 / #2284 ?

src/Data/List/Relation/Ternary/Appending/Properties.agda
```diff
@@ -10,48 +10,89 @@ module Data.List.Relation.Ternary.Appending.Properties where
 
 open import Data.List.Base using (List; [])
 open import Data.List.Relation.Ternary.Appending
+open import Data.List.Relation.Binary.Pointwise as Pw using (Pointwise; []; _∷_)
+open import Data.Product.Base as Σ using (∃-syntax; _×_; _,_)
```
jamesmckinna, line 14: Ditto.

src/Relation/Binary/Properties/Setoid.agda
jamesmckinna: Ok, let's merge, and work on the refactor later.

jamesmckinna (approved):
Happy with the discussion of potential subsequent refactoring. Let's merge!

src/Data/List/Relation/Binary/Pointwise/Base.agda
MatthewDaggitt: Fixed both.

Issue #569, opened 2018-12-21
Show proposal


mechvel:
Standard library needs the Show class,
as in Haskell (only it is desirable a slightly better design, with ShowOptions).
Because  `show'  for arbitrary ``A and B : Set _``  is easily lifted to `show'  for 
A × B, List A, Maybe A, Fraction A, ...

Consider something like

```
lChar          = List Char
pageWidth = 72    -- edit it

Verbosity : Set 
Verbosity = ℕ       -- Verbosity level in messages:
                            -- v < 1 is the least verbose,  1 is more verbose ...

data ParensOrNot : Set where  Parens    : ParensOrNot
                                                 ParensNot : ParensOrNot

record ShowOptions : Set where    -- something of this sort
       field
          verbosity        : Verbosity 
          listSeparator   : LChar
          fieldSeparator : LChar
          parInShow      : ParensOrNot

-- Options for printing to a string -- related to the Show class.
-- fieldSeparator  is for data fields except List.
-- parInShow  effects only the case of a data being a Tuple or a List;
--            is defines whether to unparenth the members of this data
--            print at the top level.
--   Examples: 1) a polynomial list is printed as  "[x^2+1, x-1]",
--                under  parInShow = ParensNot,  and it is printed as
--               "[(x^2+1), (x-1)]"  under Parens.
--             2) A pair list is printed as  "[(2,1),(1,2)]",
--                under Parens,  and it is printed as  "[2,1,1,2]"
--                under ParensNot.

record Show {α} (A : Set α) :  Set α
  where
  constructor show′ 

  field  shows :  ShowOptions → A → LChar → LChar     

  show : ShowOptions → A → LChar
  show opt a =  shows opt a [] 

  showsn : ℕ → A → LChar → LChar
  showsn n =  shows (shOpt n)

  shown : ℕ → A → LChar 
  shown verb =  show $ shOpt verb

  showS : ShowOptions → A → String
  showS opt =  Str.fromList ∘ show opt 

  shownS : ℕ → A → String
  shownS verb =  Str.fromList ∘ shown verb

  showsList : ShowOptions → List A → LChar → LChar 
  showsList opt xs =   id
```
--------------------------------------------------------------------------------------

See  the Show instances in DoCon-A.

What people think of this?

gallais:
related: #431

MatthewDaggitt:
See https://github.com/agda/agda-stdlib/pull/2024 for a good discussion of the design, which needs to include `Precedence` as well.

PR #2024, opened 2023-07-17
move `parensIfSpace` from `Data.String` to `Data.String.Base`

Sofia-Insa:
In [PR#2016](https://github.com/agda/agda-stdlib/pull/2016), in `src/Reflection/AST/Show.agda`, `parensIfSpace` was the only function used from `Data.String`, while the other were from Data.String.Base. 

In the stdl `parensIfSpace` is only called once : in `src/Reflection/AST/Show.agda`.

### DIFF
--- a/src/Data/String.agda
+++ b/src/Data/String.agda
@@ -9,27 +9,14 @@
 module Data.String where
 
 open import Data.Bool using (true; false; T?)
-open import Data.Char as Char using (Char)
 open import Function.Base
 open import Data.Nat.Base as ℕ using (ℕ; _∸_; ⌊_/2⌋; ⌈_/2⌉)
 import Data.Nat.Properties as ℕₚ
 open import Data.List.Base as List using (List; _∷_; []; [_])
 open import Data.List.NonEmpty as NE using (List⁺)
 open import Data.List.Extrema ℕₚ.≤-totalOrder
-open import Data.List.Relation.Binary.Pointwise using (Pointwise)
-open import Data.List.Relation.Binary.Lex.Strict using (Lex-<; Lex-≤)
 open import Data.Vec.Base as Vec using (Vec)
 open import Data.Char.Base as Char using (Char)
-import Data.Char.Properties as Char using (_≟_)
-open import Function
-open import Relation.Binary using (Rel)
-open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
-open import Relation.Nullary.Decidable using (does)
-open import Relation.Unary using (Pred; Decidable)
-
-open import Data.List.Membership.DecPropositional Char._≟_
-
-
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Taneb:
I don't think we should do this. Importing `Data.Char.Properties` from `Data.String.Base` hugely increases the latter module's dependency footprint.

JacquesCarette:
Oops, my bad - I didn't look deeply enough into `Data.Char.Properties` when I asked @Sofia-Insa to try for this version.

I agree that the `showPred` solution is much better.

Saransh-cpp:
I and @Sofia-Insa spent some time on this and went through (surprisingly limited) resources on the internet, but we are not sure how to implement this. We think the type signature should be -

```agda
showsPrec : Precedence → String → String
-- or
showsPrec : ℕ → String → String
```

but we are not sure how the fixity would factor in while adding brackets (also, fractional fixity is a possibility - #1999). Should it be that if `showsPrec` gets a precedence of 6, it will add brackets around all the operations performed by an operator with fixity 6?

https://stackoverflow.com/a/24207467/14746647 and https://stackoverflow.com/a/61160682/14746647 discuss implementing (or overriding?) Haskell's `showsPrec` but they hardcode it, listing out all the possible cases with the operators they are using. The definition in https://hackage.haskell.org/package/base-4.7.0.0/docs/src/GHC-Show.html does generalize it, but it also has a lot of defined cases. How should we proceed ahead with this?

JacquesCarette:
It is indeed tricky to find good documentation for this. I eventually found some in Haskell's [Text.Show documentation](https://hackage.haskell.org/package/base-4.18.0.0/docs/Text-Show.html).

Certainly the precedence should not be `ℕ` as that's not what Agda itself uses. Digging into the depths of the [Reflection API](https://agda.github.io/agda-stdlib/Agda.Builtin.Reflection.html) I see that `Float` is what is used (though actually, the `Precedence` type has 2 constructors). To be maximally compatible with Agda itself, you should use `Precedence`.

As the Haskell doc indicates, precedence gives a total order, so the comparison to use is `<`.

It is indeed rather odd that stdlib currently does not seem to use `Precedence` at all!

gallais:
> It is indeed rather odd that stdlib currently does not seem to use Precedence at all!

We don't really have show instances for anything beyond base types.
I'm think the show for the reflection stuff was hacked in a hurry for
debugging purposes rather than properly designed.

JacquesCarette:
So we shouldn't use the current `Precedence` type? Or it's just the `show` that is a hack?

gallais:
I think we should definitely use `Precedence`.

MatthewDaggitt:
I agree with all the discussion in here so far. Going to close this PR as I think we've agreed this is not the approach we want to follow. Further discussion can continue in https://github.com/agda/agda-stdlib/issues/569.

PR #2028, opened 2023-07-21
Put indexed data types in the right universes

plt-amy:
To fix agda/agda#6654, we've decided that large indices will no longer be allowed by default. There is an infective flag `--large-indices` to bring them back, but none of the uses of large indices in the standard library were essential: to avoid complicated mutually-recursive PRs across repos, I adjusted the levels to check with `--no-large-indices` instead of adding the flag to the modules that used them.


andreasabel:
Branch this pr from `master` not from outdated `experimental`:
- #2030

PR #2029, opened 2023-07-23
Add `Data.Nat.Binary.Base.divMod`

XiaohuWang0921:
Long division for `Data.Nat.Binary` as well as some related functions. This is still a draft. I will probably add some properties later.


XiaohuWang0921:
Oops, did not see that `_∸_` was already defined in `Data.Nat.Binary.Subtraction`. I simply removed my definition because it was not necessary to define long division anyway.
Though should I move my definition of `_-_` also to `Data.Nat.Binary.Subtraction`? And perhaps also move `divMod` and related functions into a separate module? Frankly, I do not understand why there is a separate module just for subtraction and its properties in the first place.

Taneb:
Thank you for this PR, this is definitely a worthwhile addition to the library. Have you managed to do any form of benchmarking?

XiaohuWang0921:
@Taneb No, unfortunately I'm not very familiar with benchmarking.

andreasabel:
CI reports failure: https://github.com/agda/agda-stdlib/actions/runs/5636154398/job/15267920707#step:10:654
```
/home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Binary/Properties.agda:149,53-54
Ambiguous name _%_. It could refer to any one of
  Data.Nat.Binary.Base._%_ bound at
    /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Binary/Base.agda:146,1-4
  ℕ._%_ bound at
    /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Base.agda:226,1-4
_%_ is in scope as
  * a defined name Data.Nat.Binary.Base._%_ brought into scope by
    - the opening of Data.Nat.Binary.Base at /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Binary/Properties.agda:17,13-33
    - its definition at /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Binary/Base.agda:146,1-4
  * a defined name Data.Nat.Base._%_ brought into scope by
    - the opening of Data.Nat.DivMod at /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Binary/Properties.agda:19,13-28
    - the opening of Data.Nat.Base at
    - its definition at /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Nat/Base.agda:226,1-4
when scope checking n % 2
```

XiaohuWang0921:
@andreasabel Yep, it was nothing serious, just some naming conflicts in `Data.Nat.Binary.Properties` since it also imports from `Data.Nat.DivMod`. My latest commit fixed it.

src/Data/Nat/Binary/Base.agda
```diff
@@ -106,6 +113,38 @@ _        * zero     =  zero
...
+                             r₂    = double r
+                         in maybe′ (1+[2 q ] ,_) (q₂ , r₂) (r₂ - n >>= predᴹ)
+
+_div_ = proj₁ ∘₂′ _divMod_
+
+_mod_ = proj₂ ∘₂′ _divMod_
+
+_/_ = _div_
+
+_%_ = _mod_
```
MatthewDaggitt, line 146: Can we have types and precedence declarations for these operators?

src/Data/Nat/Binary/Base.agda
```diff
@@ -106,6 +113,38 @@ _        * zero     =  zero
...
+  helper 2[1+ m ] n    = let q , r = helper m n
+                             q₂    = double q
+                             r₂    = double r
+                         in maybe′ (1+[2 q ] ,_) (q₂ , 1+[2 r ]) (r₂ - n)
+  helper 1+[2 m ] n    = let q , r = helper m n
+                             q₂    = double q
+                             r₂    = double r
+                         in maybe′ (1+[2 q ] ,_) (q₂ , r₂) (r₂ - n >>= predᴹ)
+
+_div_ = proj₁ ∘₂′ _divMod_
```
MatthewDaggitt, line 140: We should follow the same conventions as in `Data.Nat.DivMod` and should reserve the names `_div_` and `_mod_` and indeed `_divMod_` operators for the variants that return `Fin`. I suggest removing `_div_` and `_mod_` entirely and rename `_divMod_` to `_/%_`?

src/Data/Nat/Binary/Base.agda
```diff
@@ -106,6 +113,38 @@ _        * zero     =  zero
 1+[2 x ] * 2[1+ y ] =  2[1+ y + x * 2[1+ y ] ]
 1+[2 x ] * 1+[2 y ] =  1+[2 x + y * 1+[2 x ] ]
 
+_-_ : ℕᵇ → ℕᵇ → Maybe ℕᵇ
```
MatthewDaggitt, line 116: There's some inconsistency here between the names. We have the `ᴹ` in `predᴹ` signifying that it returns a `Maybe` but `_-_` returns a `Maybe` but has no `ᴹ`. To be honest I think the `Maybe` versions should be the canonical ones (i.e. have the simple name) and the non-`Maybe` versions should have a derived name.

What are other people's thoughts?

src/Data/Nat/Binary/Base.agda
```diff
@@ -106,6 +113,38 @@ _        * zero     =  zero
...
+
+_divMod_ : ℕᵇ → ℕᵇ → ℕᵇ × ℕᵇ
+m divMod n = maybe′ (flip′ helper (pred n)) (zero , zero) (predᴹ m)
+  module divMod where
+  -- helper m n computes (suc m / suc n, suc m % suc n)
+  helper : ℕᵇ → ℕᵇ → ℕᵇ × ℕᵇ
+  helper zero     zero = 1+[2 zero ] , zero
+  helper zero     _    = zero        , 1+[2 zero ]
+  helper 2[1+ m ] n    = let q , r = helper m n
+                             q₂    = double q
```
MatthewDaggitt, line 132: Can we rename these `2q` instead of `q₂`. A little bit more meaningful.

MatthewDaggitt (changes requested):

src/Data/Nat/Binary/Base.agda
XiaohuWang0921: I actually agree with you on that the `Maybe` versions should be the canonical ones, but if we rename `predᴹ` into `pred` and the original `pred` into something else, we would break the convention followed by `Data.Nat.Base` as there `pred` is defined without `Maybe`.

src/Data/Nat/Binary/Base.agda
XiaohuWang0921: I also renamed `r₂` to `2r` in my latest commit, in line with your suggestion.

src/Data/Nat/Binary/Base.agda
MatthewDaggitt: That's a good point. Okay, let's leave it lie for now.

jamesmckinna:
What happened here?

XiaohuWang0921:
@jamesmckinna I forgot this hasn't been merged and discarded my commits because there seemed to be conflicts with the upstream. Didn't know this happened. Sorry.

PR #2030, opened 2023-07-24
Put indexed data types in the right universes

### LINKED ISSUE #490: Missing properties in Data.Integer.Properties

### LINKED ISSUE #588: Inconsistent convention: lookup's order of arguments

### LINKED ISSUE #645: Redefine Bottom to make use of Prop?

### LINKED ISSUE #693: Change the definition of the free monad on a container?
Currently the definition is:

```agda
_⋆C_ : ∀ {x s p} → Container s p → Set x → Container (s ⊔ x) p
C ⋆C X = const X ⊎ C

_⋆_ : ∀ {x s p} → Container s p → Set x → Set (x ⊔ s ⊔ p)
C ⋆ X = μ (C ⋆C X)
```

However Agda can't see that the `Set x` argument is used in a strictly
positive manner (it would need to partially evaluate the specialised type
of `μ`'s constructor (what a mouthful) to see so).

This means that if we try to write the kind of translations we can see in
@ichistmeinname's [Monad to prove them all](https://github.com/ichistmeinname/free-proving) we get a strict positivity
error when defining:

```agda
data List {s p a} (C : Container s p) (A : Set a) : Set (a ⊔ s ⊔ p) where
  []  : List C A
  _∷_ : C ⋆ A → C ⋆ (List C A) → List C A
```

However if we decide to define the free monad directly:

```agda
data _⋆_ {s p a} (C : Container s p) (A : Set a) : Set (s ⊔ p ⊔ a) where
  pure   : A → C ⋆ A
  impure : ⟦ C ⟧ (C ⋆ A) → C ⋆ A
```

Then the fact it's using `A` in a strictly positive manner is self-evident
and our effectful-list is accepted.

We could / should prove that there is a bijection between the old and the new
definitions.

### LINKED ISSUE #942: Port Andreas' work on reverseAcc to vectors too

### LINKED ISSUE #1289: three way comparison of natural number?
I just realize the library does not have comparison returning `<`, `=`, or `>`. is it actually the case or I just over looked?

### LINKED ISSUE #1333: Add `WriterT` and `ExceptT` monad transformers

### LINKED ISSUE #1443: Missing `cast` properties

### LINKED ISSUE #1535: lib-1.7 : Reexport from Definitions.RawMagma is not full
`Algebra.Properties.Magma.Divisibility` reexports several items from
`Algebra.Definitions.RawMagma`, 
but not all of them.
For example, if one needs  `_∣ˡ_`  (which is likely to happen) one is forced to add the import 
```
import Algebra.Definitions.RawMagma as OfRawMagma
...
open OfRawMagma rawMagma ...
```
So, it is better for `Algebra.Properties.Magma.Divisibility`  to reexport all items from  `Algebra.Definitions.RawMagma`.

### LINKED ISSUE #1646: Make the destructors for signed numeric predicates use instance arguments
For example in `Data.Rational.Unnormalised.Properties` we have:
```
nonPositive⁻¹ : ∀ {q} → .(NonPositive q) → q ≤ 0ℚᵘ
```
whereas it should be:
```
nonPositive⁻¹ : ∀ q → .{{NonPositive q}} → q ≤ 0ℚᵘ
```

I missed these in my initial instance argument sweep.

### LINKED ISSUE #1694: Ordered algebraic structures?

### LINKED ISSUE #1765: Unsolved implicits in Data.List.Relation.Ternary.Appending.Propositional

### LINKED ISSUE #1829: Support GHC 9.4
The bounds for `base` in `agda-stdlib-utils.cabal` should include 4.17 so that GHC 9.4 is supported.

### LINKED ISSUE #1844: `Data.Integer.Properties.+-isAbelianGroup` should be called `+-0-isAbelianGroup`

andreasabel:
Base #2028 on `master` instead of on outdated `experimental`.

@plt-amy writes:
To fix https://github.com/agda/agda/issues/6654, we've decided that large indices will no longer be allowed by default. 
There is an infective flag `--large-indices` to bring them back, but none of the uses of large indices in the standard library were essential: to avoid complicated mutually-recursive PRs across repos, I adjusted the levels to check with `--no-large-indices` instead of adding the flag to the modules that used them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -733,6 +733,13 @@ Non-backwards compatible changes
 * In accordance with changes to the flags in Agda 2.6.3, all modules that previously used
   the `--without-K` flag now use the `--cubical-compatible` flag instead.
 
+* To avoid _large indices_ that are by default no longer allowed in Agda 2.6.4,
+  universe levels have been increased in the following definitions:
+  - `Data.Star.Decoration.DecoratedWith`
+  - `Data.Star.Pointer.Pointer`
+  - `Reflection.AnnotatedAST.Typeₐ`
+  - `Reflection.AnnotatedAST.AnnotationFun`
+
 * The first two arguments of `m≡n⇒m-n≡0` (now `i≡j⇒i-j≡0`) in `Data.Integer.Base`
   have been made implicit.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2165,9 +2165,9 @@ Other minor changes
   take-suc-tabulate : (f : Fin n → A) (o : Fin n) → let m = toℕ o in take (suc m) (tabulate f) ≡ take m (tabulate f) ∷ʳ f o
   drop-take-suc : (o : Fin (length xs)) → let m = toℕ o in drop m (take (suc m) xs) ≡ [ lookup xs o ]
   drop-take-suc-tabulate : (f : Fin n → A) (o : Fin n) → let m = toℕ o in drop m (take (suc m) (tabulate f)) ≡ [ f o ]
-  
-  take-all : n ≥ length xs → take n xs ≡ xs     
-  
+
+  take-all : n ≥ length xs → take n xs ≡ xs
+
   take-[] : ∀ m → take  m [] ≡ []
   drop-[] : ∀ m → drop  m [] ≡ []
   ```
@@ -2901,7 +2901,7 @@ Other minor changes
   foldr-map : foldr f x (map g xs) ≡ foldr (g -⟨ f ∣) x xs
   foldl-map : foldl f x (map g xs) ≡ foldl (∣ f ⟩- g) x xs
   ```
-  
+
 NonZero/Positive/Negative changes
 ---------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -371,42 +371,42 @@ Non-backwards compatible changes
   `m / n` without having to explicitly provide a proof, as instance search will fill it in
   for you. The full list of such operations changed is as follows:
     - In `Data.Nat.DivMod`: `_/_`, `_%_`, `_div_`, `_mod_`
-	- In `Data.Nat.Pseudorandom.LCG`: `Generator`
-	- In `Data.Integer.DivMod`: `_divℕ_`, `_div_`, `_modℕ_`, `_mod_`
-	- In `Data.Rational`: `mkℚ+`, `normalize`, `_/_`, `1/_`
-	- In `Data.Rational.Unnormalised`: `_/_`, `1/_`, `_÷_`
+    - In `Data.Nat.Pseudorandom.LCG`: `Generator`
+    - In `Data.Integer.DivMod`: `_divℕ_`, `_div_`, `_modℕ_`, `_mod_`
+    - In `Data.Rational`: `mkℚ+`, `normalize`, `_/_`, `1/_`
+    - In `Data.Rational.Unnormalised`: `_/_`, `1/_`, `_÷_`
 
 * At the moment, there are 4 different ways such instance arguments can be provided,
   listed in order of convenience and clarity:
     1. *Automatic basic instances* - the standard library provides instances based on the constructors of each
-	   numeric type in `Data.X.Base`. For example, `Data.Nat.Base` constains an instance of `NonZero (suc n)` for any `n`
-	   and `Data.Integer.Base` contains an instance of `NonNegative (+ n)` for any `n`. Consequently,
-	   if the argument is of the required form, these instances will always be filled in by instance search
-	   automatically, e.g.
-	   ```
-	   0/n≡0 : 0 / suc n ≡ 0
-	   ```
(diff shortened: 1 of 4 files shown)


jespercockx (approved):

andreasabel:
Added a CHANGELOG entry (and fixed whitespace violations there).

andreasabel:
I suppose the stdlib tsars are on vacation/busy atm.  To clear the road for an Agda 2.6.4 RC, I am merging this now.  These are anyhow mild changes...

JacquesCarette:
It would be nice indeed to not have a bottleneck of a single person who merges PRs. I've discussed a bit with @jamesmckinna who has also talked w/ @MatthewDaggitt about it, but I don't think it has gone beyond that. [I don't really want that power myself! I just would like the process to be a little faster without putting undue pressure onto a single person.]

jespercockx:
Well there *is* more than one person with the power to merge PRs here (as Andreas just demonstrated). It is just a power that's rarely exercised.

JacquesCarette:
I think I'm going to wait until the most active people seem to be back online - but yes, good idea.

PR #2035, opened 2023-07-28
Add more properties for `xor`

Saransh-cpp:
Taken directly from `Idris2` - https://github.com/idris-lang/Idris2/blob/4fcb0fb4a796650f5df101526a923c0aa522d1ad/libs/base/Data/Bool/Xor.idr#L13-L64

The type checker is happy, but I am not sure if everything is right, because the `Idris` version of some properties look comparatively more complicated.

Also, please let me know if I should link the original `Idris` properties in the comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1989,14 +1989,23 @@ Other minor changes
   ∧-conicalˡ : LeftConical true _∧_
   ∧-conicalʳ : RightConical true _∧_
   ∧-conical  : Conical true _∧_
-  xor-same-false    : ∀ x → x xor x ≡ false
-  xor-false-neutral : ∀ x → false xor x ≡ x
-  xor-true-not      : ∀ x → true xor x ≡ not x
-  xor-not-true      : ∀ x → x xor (not x) ≡ true
-  not-xor           : ∀ x y → not (x xor y) ≡ (not x) xor y
-  not-xor-cancel    : ∀ x y → (not x) xor (not y) ≡ x xor y
-  xor-commutative   : ∀ x y → x xor y ≡ y xor x
-  xor-associative   : ∀ x y z → x xor (y xor z) ≡ (x xor y) xor z
+
+  true-xor            : true xor x ≡ not x
+  xor-same            : x xor x ≡ false
+  not-distribˡ-xor    : not (x xor y) ≡ (not x) xor y
+  not-distribʳ-xor    : not (x xor y) ≡ x xor (not y)
+  xor-assoc           : Associative _xor_
+  xor-comm            : Commutative _xor_
+  xor-identityˡ       : LeftIdentity false _xor_
+  xor-identityʳ       : RightIdentity false _xor_
+  xor-identity        : Identity false _xor_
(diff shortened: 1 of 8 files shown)


src/Data/Bool/Properties.agda
```diff
@@ -627,20 +627,7 @@ true  <? _     = no  (λ())
...
-
-xor-∧-commutativeRing : CommutativeRing 0ℓ 0ℓ
-xor-∧-commutativeRing = ⊕-∧-commutativeRing
-  where
-  open BooleanAlgebraProperties ∨-∧-booleanAlgebra
-  open XorRing _xor_ xor-is-ok
-
-------------------------------------------------------------------------
--- Miscellaneous other properties
+-- Propertied of not
```
Saransh-cpp, line 630:
```suggestion
-- Properties of not
```

gallais:
Dashed names typically only describe the LHS (e.g. `map-replicate`, not `map-replicate-replicate`) so:
`xor-same` rather than `xor-same-true`, `xor-true` rather than `xor-true-not`.

src/Data/Bool/Properties.agda
```diff
@@ -660,6 +647,56 @@ not-¬ {false} refl ()
...
+
+xor-is-ok : ∀ x y → x xor y ≡ (x ∨ y) ∧ not (x ∧ y)
+xor-is-ok true  y = refl
+xor-is-ok false y = sym (∧-identityʳ _)
+
+xor-same : ∀ x → x xor x ≡ false
+xor-same false = refl
+xor-same true  = refl
+
+xor-false : ∀ x → false xor x ≡ x
```
MatthewDaggitt, line 661: In the same way we do for `and` and `or` can we use the pre-defined types in `Algebra.Definitions`  e.g. identity, comm etc, and change the names to match?

JacquesCarette:
I'll work with @Saransh-cpp to get soem more stdlib-sounding names in this and the next batch of PRs. And also to use the stdlib types.

CHANGELOG.md
```diff
@@ -1985,10 +1985,18 @@ Other minor changes
...
-  ∧-conical : Conical true _∧_
+  ∧-conical  : Conical true _∧_
+  xor-same-false    : ∀ x → x xor x ≡ false
+  xor-false-neutral : ∀ x → false xor x ≡ x
+  xor-true-not      : ∀ x → true xor x ≡ not x
+  xor-not-true      : ∀ x → x xor (not x) ≡ true
+  not-xor           : ∀ x y → not (x xor y) ≡ (not x) xor y
+  not-xor-cancel    : ∀ x y → (not x) xor (not y) ≡ x xor y
+  xor-commutative   : ∀ x y → x xor y ≡ y xor x
+  xor-associative   : ∀ x y z → x xor (y xor z) ≡ (x xor y) xor z
```
Saransh-cpp, line 1999: Will update once the property names and types and final.

MatthewDaggitt: Thanks. Once the CHANGELOG is updated, this looks good to me.

Saransh-cpp: Thanks for the review, updated!

PR #2036, opened 2023-07-28
Simplify more `Data.Product` imports to `Data.Product.Base`

Saransh-cpp:


src/Reflection/AST/Abstraction.agda
```diff
@@ -8,7 +8,7 @@
 
 module Reflection.AST.Abstraction where
 
-open import Data.Product                               using (_×_; <_,_>; uncurry)
+open import Data.Product.Base                     using (_×_; <_,_>; uncurry)
 open import Data.String as String                      using (String)
```
Saransh-cpp, line 12:
```suggestion
open import Data.Product.Base                         using (_×_; <_,_>; uncurry)
open import Data.String as String                      using (String)
```

gallais (approved):

PR #2041, opened 2023-08-02
Add some `_∷ʳ_` properties to `Data.Vec.Properties`

shhyou:
This PR adds the obvious equations concerning `Data.Vec._∷ʳ_` (snoc), `cast`, `++`, `init` and `last`. Unfortunately, many of them are wrapped inside `cast`.

Besides this PR, I have two additional sets of properties that I would like to add as well. The first set of equations involve `Data.Vec.reverse` and `fromList`, so all of them are again wrapped inside a `cast`. The second set of equations are about the `head`, `tail`, `init` and `last` of vectors, e.g. the `init` after `tail` is equivalent to `tail` after `init`, and that `head` is equivalent to `head` after `init`, etc. I'm less certain about either the form or the usefulness of these two set of equations, so I hope to open a separate PR.

p.s. @amblafont  Thank you for adding `zipwith-++` in #2020! I found it missing while working on vectors.


src/Data/Vec/Properties.agda
```diff
@@ -860,12 +866,39 @@ foldr-∷ʳ : ∀ (B : ℕ → Set b) (f : FoldrOp A B) {e} y (ys : Vec A n) →
 foldr-∷ʳ B f y []       = refl
 foldr-∷ʳ B f y (x ∷ xs) = cong (f x) (foldr-∷ʳ B f y xs)
 
+-- init, last and _∷ʳ_
+
+init-∷ʳ : ∀ x (xs : Vec A n) → init (xs ∷ʳ x) ≡ xs
+init-∷ʳ x xs with xs ∷ʳ x in eq
```
JacquesCarette, line 872: Are you sure this can't be done by induction on `xs` ?

src/Data/Vec/Properties.agda
```diff
@@ -860,12 +866,39 @@ foldr-∷ʳ : ∀ (B : ℕ → Set b) (f : FoldrOp A B) {e} y (ys : Vec A n) →
...
+init-∷ʳ x xs | xs* with initLast xs*
+init-∷ʳ x xs | .(xs′ ∷ʳ x′) | xs′ , x′ , refl = sym (proj₁ (∷ʳ-injective xs xs′ eq))
+
+last-∷ʳ : ∀ x (xs : Vec A n) → last (xs ∷ʳ x) ≡ x
```
JacquesCarette, line 876: same here?

JacquesCarette (commented):
Those casts seems rather inevitable. Nice additions.

src/Data/Vec/Properties.agda
shhyou: `init` is itself defined using `initLast`, so I think somehow I need to get `initLast (xs ∷ʳ x)` computing. I am not sure what to do with this goal:

```
init-∷ʳ x (y ∷ xs) = {!!}

Goal: (init (y ∷ (xs ∷ʳ x))
       | (initLast (y ∷ (xs ∷ʳ x)) | initLast (xs ∷ʳ x)))
      ≡ y ∷ xs
```

https://github.com/agda/agda-stdlib/blob/bfa4f8bc41da10bffaf69f5786301e587ae69c83/src/Data/Vec/Base.agda#L342-L349

JacquesCarette: `init` should be defined as `take n xs`. You'll still need a `with` because of `splitAt`.  

These combinators that return proofs along with other things feel oddly out of place in `Data.Vec.Base`. I expected things here to all be pure computations.

shhyou: Do you suggest that I change the existing definition of `init` in `Data.Vec.Base`? Could that break existing code in the wild?

shhyou: If the definition of `init` and `last` were to change, how about I directly write an inductive definition?

JacquesCarette: Yes, that is what I am suggesting. And "code" should not break, but proofs that rely on the exact definition might. But I doubt many people have such proofs.

Yes, a direct inductive definition would be better!

shhyou: Done. I like the straightforward definition better too. Thank you!

JacquesCarette (approved):
This is indeed much nicer. I'm glad you persevered.

shhyou:
Actually, is the name of `++-∷ʳ` correct, or should it be `∷ʳ-++`? In list there's this lemma whose naming order seems to be the other way around:

https://github.com/agda/agda-stdlib/blob/9a240ee51ac9b2b541ceab6effcc325eb11c96ab/src/Data/List/Properties.agda#L984-L988

@jamesmckinna Do you have any suggestion on the name?

src/Data/Vec/Base.agda
```diff
@@ -345,12 +345,12 @@ initLast {n = suc n} (x ∷ xs) with initLast xs
 ... | (ys , y , refl) = (x ∷ ys , y , refl)
 
 init : Vec A (1 + n) → Vec A n
-init xs with initLast xs
-... | (ys , y , refl) = ys
+init {n = zero}  (x ∷ []) = []
+init {n = suc n} (x ∷ xs) = x ∷ init xs
```
MatthewDaggitt, line 349: Do we actually need to match on the implicit `n` here? Can we get away with?
```
init (x ∷ []) = []
init (x ∷ y :: xs) = x ∷ init (y :: xs)
```
Same with `last` below?

MatthewDaggitt (commented):
Thanks, these proofs look really nice! Can we get a CHANGELOG entry to go with them? You'll need to add an entry for the breaking change to `init` and `last` as well.

src/Data/Vec/Base.agda
shhyou: It turns out that matching on `xs` instead of matching on `n` blocks expressions like `init (y ∷ (xs ∷ʳ x))` from computing. Instead, I can change `init` to signify that `xs` is not actually matched:

```agda
init : Vec A (1 + n) → Vec A n
init {n = zero}  (x ∷ xs) = []
init {n = suc n} (x ∷ xs) = x ∷ init xs
```

The same problem applies to `last` and `last-∷ʳ`.

shhyou:
Thanks a lot for the review! I added the items to the CHANGELOG.

src/Data/Vec/Base.agda
shhyou: Okay, I edited both `init` and `last` to emphasis that `xs` in the tail is not matched.

jamesmckinna:
Good question about naming (and operator order when doing so: there does seem to be an anomaly in the examples you quote above...)! Not sure how best to resolve, but I suppose doing it here would be better than waiting for a subsequent weeding-out of inconsistencies. Trouble is, it *may* generate complex deprecation cycles, except that I think all these proofs have been added since the revisions for v2.0 started (?), so perhaps that's not an issue. The one in `List` is a bit weird though... more later

CHANGELOG.md
```diff
@@ -858,6 +858,8 @@ Non-backwards compatible changes
 
 * In `Data.Sum.Base` the definitions `fromDec` and `toDec` have been moved to `Data.Sum`.
 
+* In `Data.Vec.Base`: the definitions `init` and `last` have been replaced by an inductive definition.
```
jamesmckinna, line 861: Suggest replrasing:
"... replaced by a direct recursive definition in place of the `InitLast` view-derived definitions"

jamesmckinna:
Regarding naming: I believe that we observe left-to-right order (Euro-/Anglo-centric, I realise), rather than outside-in precedence order, so 
* `∷ʳ-++ : ∀ (xs : List A) (a : A) (ys : List A) → xs ∷ʳ a ++ ys ≡ xs ++ a ∷ ys` is OK
* `ʳ++-++ : ∀ (xs {ys zs} : List A) → (xs ++ ys) ʳ++ zs ≡ ys ʳ++ xs ʳ++ zs` is NOT, I think, and should be `++-ʳ++` instead (shout-out to @MatthewDaggitt to confirm?)

Fortunately, that proof only occurs in this one file, so I think it's OK simply to rename it, define the old identifier equal to this new one, and add  a deprecation warning (and update the `CHANGELOG`  as well). Fiddly, but doable. Ask if you have problems!

src/Data/Vec/Base.agda
```diff
@@ -345,12 +345,12 @@ initLast {n = suc n} (x ∷ xs) with initLast xs
...
 init : Vec A (1 + n) → Vec A n
-init xs with initLast xs
-... | (ys , y , refl) = ys
+init {n = zero}  (x ∷ xs) = []
+init {n = suc n} (x ∷ xs) = x ∷ init xs
 
 last : Vec A (1 + n) → A
-last xs with initLast xs
-... | (ys , y , refl) = y
+last {n = zero}  (x ∷ xs) = x
```
jamesmckinna, line 352: I still think in this clause that we should have
```agda
last {n = zero} (x ∷ []) = x
```
It's more (perhaps needlessly?) strict, but documents the situation enforced by the index being 1+0...

jamesmckinna: Alternatively, use an underscore...

src/Data/Vec/Base.agda
```diff
@@ -345,12 +345,12 @@ initLast {n = suc n} (x ∷ xs) with initLast xs
...
+last {n = suc n} (x ∷ xs) = last xs
```
jamesmckinna, line 353: ... and correspondingly here:
```agda
last {n = suc n} (_ ∷ xs) = last xs
```

shhyou: Done.

src/Data/Vec/Base.agda
shhyou: Sure, I changed `xs` in the first clause to `_`.

CHANGELOG.md
```diff
@@ -2610,6 +2612,12 @@ Other minor changes
   ∷ʳ-injectiveˡ : xs ∷ʳ x ≡ ys ∷ʳ y → xs ≡ ys
   ∷ʳ-injectiveʳ : xs ∷ʳ x ≡ ys ∷ʳ y → x ≡ y
 
+  unfold-∷ʳ : cast eq (xs ++ x ∷ []) ≡ xs ∷ʳ x
```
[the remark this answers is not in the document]
shhyou: Ah, thanks for catching this! Fixed now.

So far I use `cast` in the name if the subject of lemma is about commuting or simplifying `cast` itself, e.g. `cast-∷ʳ`, `cast-reverse`, `cast-++ˡ` and `cast-++ʳ` etc.

CHANGELOG.md
shhyou: Done. Thanks!

shhyou:
> **[jamesmckinna said:](https://github.com/agda/agda-stdlib/pull/2041#issuecomment-1674480093)**
>
> Regarding naming: I believe that we observe left-to-right order (Euro-/Anglo-centric, I realise), rather than outside-in precedence order, so
>   * `∷ʳ-++ : ∀ (xs : List A) (a : A) (ys : List A) → xs ∷ʳ a ++ ys ≡ xs ++ a ∷ ys` is OK
>   * `ʳ++-++ : ∀ (xs {ys zs} : List A) → (xs ++ ys) ʳ++ zs ≡ ys ʳ++ xs ʳ++ zs` is NOT, I think, and should be `++-ʳ++` instead (shout-out to @MatthewDaggitt to confirm?)
>
> Fortunately, that proof only occurs in this one file, so I think it's OK simply to rename it, define the old identifier equal to this new one, and add a deprecation warning (and update the `CHANGELOG` as well). Fiddly, but straightforward. Ask if you have problems!

Sounds good. I'll be using the left-to-right naming convention, then. And fortunately `ʳ++-++` hasn't been in `Data.Vec.Properties` yet and that `ʳ++-ʳ++` has type `(xs ʳ++ ys) ʳ++ zs ≡ ys ʳ++ xs ++ zs` rather than the other way around, so there's no naming conflict 😃.

jamesmckinna:
`status:ready` now?

shhyou:
Thanks!

PR #2042, opened 2023-08-03
Add `find`, `findIndex`, and `findIndices` for `Data.List`

Saransh-cpp:
Porting from Idris2 - https://github.com/idris-lang/Idris2/blob/b7bda5e96df62e8aee0af8c5cfb2444db29774b4/libs/base/Data/List.idr#L147-L170

- `find`: first element that satisfies the predicate
- `findIndex`: index of the first element that satisfies the predicate
- `findIndices`:  list of indices of all the elements that satisfies the predicate


CHANGELOG.md
```diff
@@ -2145,6 +2145,9 @@ Other minor changes
   wordsByᵇ     : (A → Bool) → List A → List (List A)
   derunᵇ       : (A → A → Bool) → List A → List A
   deduplicateᵇ : (A → A → Bool) → List A → List A
+  find         : (A → Bool) → List A -> Maybe A
```
gallais, line 2148: As you can see in the previous entries in the CHANGELOG, boolean
versions have a name with the `ᵇ` suffix and the others instead use
a `Dec`-like interface.

src/Data/List/Base.agda
```diff
@@ -329,6 +330,24 @@ splitAt zero    xs       = ([] , xs)
...
+find p [] = nothing
+find p (x ∷ xs) = if p x then just x else find p xs
+
+findIndex : (A → Bool) → (x : List A) → Maybe $ Fin (length x)
+findIndex p [] = nothing
+findIndex p (x ∷ xs) = if p x
+  then just zero
+  else Maybe.map suc (findIndex p xs)
+
+findIndices : (A → Bool) → List A → List ℕ
```
gallais, line 343: Is this `mapMaybe` in disguise?

Taneb:
I think that the changes to `Function.Base` and `Function.Indexed.Relation.Binary.Equality` should belong to different PRs?

CHANGELOG.md
Saransh-cpp: Oh yes, I should also add the corresponding non-`ᵇ` functions. Thanks!

jamesmckinna:
As with issue #2048, please consider the trade-offs with implementing these functions directly, or by specialising existing functionality in `Any`. And where that functionality doesn't yet exist, please consider adding it there... eg `findIndices`

jamesmckinna:
In your API extension, 
`findIndex` returns a `Maybe (Fin ...)`
whereas
`findIndices` returns a `List ℕ`

I don't think it should be like this: `ℕ` is *not* the associated index type for `List`, or `Vec` (so please also add suitable functionality there, too!), it's `Fin ...`, for maximum compile-time (pattern-matching during typechecking) precision...

Lists of `ℕ`s, if actually required, can always be recovered via `List.map toℕ`...


(There's a separate discussion issue to have with the `idris` team about some of their choices of library API design... ;-))

(There's a further, again separate, discussion to be had about whether `List` deserves a custom indexing type, proxied here by `Maybe (Fin ...)`, which is a *view* of a list `xs` an element `x` and a `n : ℕ` such that
* either `x` occurs in `xs`, at index `i : Fin (length cs)`, and moreover `n = toℕ i` (definitionally!)
* or `x` does not occur (and we can compute a proof of that!)

But view-based approaches still seem... ahead of their time ;-), some 20 years later :-( ...)

gallais:
> There's a separate discussion issue to have with the idris team about some of their choices of library API design... ;-)

I don't there is any design *per se*, it's more of an accumulation of things that
have been moved around from different parts of the library based on people's need.
That's why I don't think it's necessarily a good idea to blindly port "missing" features
from the idris base lib to the agda stdlib: it results in opening a lot of PR, each of
which incur a lot of design work for the reviewers here.

jamesmckinna:
Under a lazy semantics, can we (yes!) obtain `findᵇ` by taking a suitably `Maybe`-fied composition of `filter p` and `List.head`...? I.e. is it sufficiently interesting/useful to have that composition fused away with an additional primitive definition? 

Probably yes, but advocates for a more minimal API might disagree? Or else, please consider adding as a property to `List.Properties` that it does, in fact, satisfy such an equational specification?

JacquesCarette:
@gallais I did do a pre-filtering of the Idris "missing features" (quite a number of things were skipped as not being very Agda-like) before settling on a few that seemed they could be genuinely useful. The easiest way to hash out a design seems to be on top of a very concrete proposal. It's also quite ok to decide "no, don't want these" and close the PR.

`find` and friends as per this PR do seem to occur frequently enough in programming to consider seriously. I was certainly surprised to not find them, even though there are huge swaths of way-more-advance code available (around `All` and `Any`) that satisfy a rather different need.

jamesmckinna:
Re: Jacques' "way more advanced" definitions

Some thoughts that I could move to a separate 'discussion' issue. 

There has (at least to my mind since my PhD (1988-1992)) been a tension between 
* "simply-typed function definitions plus dependently-typed proofs of their correctness" (which has always been the `Coq` methodology, and was that of my PhD thesis, until I got tired of the repeated effort, first to program, then to prove; the methods I developed to avoid such duplication didn't seem really to scale past small examples)
* *bona fide* dependently-typed programming, already (eg (NPS, 1990)) argued for well before Epigram and then Agda2, and now idris2 made the style of DTFP distinctive as 'programming' (as opposed to theorem proving in Martin-Lof type theory with program extraction); the methods developed by Conor and myself for programming with views being one particular (and still, seemingly, poorly understood/accepted) approach.

For me, the comment does raise questions about how much the stdlib should support the first style of development (which is reflected, to a degree, in the `X.Base` vs. `X.Properties` distinction), or be a showcase for the second. It can sometimes feel as though working in the first style inevitably leads to the need to prove properties which amount to showing the functions equivalent to those we might as well have developed in the second to begin with. I'm with Conor in wanting to avoid 'engineered coincidences', in favour of 'correct-by-construction' programming. 

Eg. what properties do we want the functions `find`, `findIndex` and `findIndices` to have, and how hard (easy?) will it be to establish them? This PR doesn't (yet) tackle such questions. As with other recent PRs, I think we do need properties of any additional functions we add to the library, if only as a litmus-test of the 'reasonableness' of their definitions, if not full correctness.

JacquesCarette:
I agree that there is a real tension between simply-typed function w/ proofs and *bona fide* dependently-typed programs. I've definitely 'seen the light' of doing proper DTP.

I'm actually quite happy with non-trivial redundancy; for example, right now, the library does have routines that use both decidable predicates and boolean-values ones (and that wasn't my doing!). I like that.

So yes, maybe `find`, `findIndex` and `findIndices` should have quite different types in Agda. And also exist as 'programming' facilities with simple types. And perhaps neither set belong in `Data.List.base` !!

jamesmckinna:
Re: things in `X.Base`: we're going have to have to take that question outside, for the time being ;-)

For the record, I support this kind of 'simple'/'programmer/programming-oriented' functionality being added to the stdlib, even at the `Base` level (refactoring at the level/granularity of modules can, I think, wait until it starts to take ~60mins to typecheck the library, rather than the current 15 or so). That's how the library will grow, and subsequent downstream PRs may make the whole edifice better, leaner, meaner, etc.

But I think it shouldn't be done without at least half an eye to, and serious discussion of, the initial design choices. And as @gallais points out, reviewer resource can be a/the bottleneck here.

jamesmckinna:
Hmmm... your tail-recursive definition seems definitely more efficient in the simply-typed case; the best I could do on a first cut of my preferred typing is the following (the type indices get in the way :-():
```agda
findIndicesᵇ : (A → Bool) → (xs : List A) → List $ Fin (length xs)
findIndicesᵇ p [] = []
findIndicesᵇ p (x ∷ xs) = if p x then zero ∷ indices else indices
  where indices = map Fin.suc (findIndicesᵇ p xs)
```
Perhaps you, or our colleagues, can find a good recursive->tail-recursive transformation/optimisation that respects the typing, and achieves what you had before... ?

jamesmckinna:
Dependently-typed abstraction fans may also note the following common (not-tail-recursive) generalisations:
```agda
findGeneralᵇ : (A → Bool) → (B : List A → Set b) →
               B [] →
               ({x : A} {xs : List A} (b : Bool) (r : B xs) → B (x ∷ xs)) →
               (xs : List A) → B xs
findGeneralᵇ {A = A} p B n c xs = h xs where
  h : (xs : List A) → B xs
  h [] = n
  h (x ∷ xs) = c (p x) (h xs)

findᵇ : (A → Bool) → List A → Maybe A
findᵇ {A = A} p = findGeneralᵇ p (const (Maybe A)) nothing
  λ {x = x} b _ → if b then nothing else just x

findIndexᵇ : (A → Bool) → (xs : List A) → Maybe $ Fin (length xs)
findIndexᵇ p = findGeneralᵇ p (Maybe ∘ Fin ∘ length) nothing
  λ b r → if b then nothing else Maybe.map suc r


findIndicesᵇ : (A → Bool) → (xs : List A) → List $ Fin (length xs)
findIndicesᵇ p = findGeneralᵇ p (List ∘ Fin ∘ length) []
  λ b r → let indices = map suc r in if b then indices else zero ∷ indices
```
(NB. NOT that I am necessarily suggesting this is the right way to proceed... ;-))

JacquesCarette:
I had a slightly different idea for making `findIndicies` tail recursive: go via a generalized version that takes in a separate Fin-to-Fin function to apply to the index (i.e. here: piles of `suc`).

Saransh-cpp:
Thanks for the review! Took me a lot of time to go through the conversation above, and I am not going to pretend I got all of that 🙂, but I think, in the end, the only modification required in this PR at the moment is the type of `findIndicesᵇ`, right? I will experiment with the following type -
```agda
findIndicesᵇ : (A → Bool) → (xs : List A) → List $ Fin (length xs)
```
as suggested by @jamesmckinna.

jamesmckinna:
Very nice now (I'm grateful for your responsiveness to my suggestions). Looks good, and ready, to me now.

Saransh-cpp:
Thanks for the review!

> any Properties to be proved about these functions?

I would be more than happy to add properties in downstream PRs!

src/Data/List/Base.agda
```diff
@@ -395,6 +396,23 @@ deduplicateᵇ : (A → A → Bool) → List A → List A
 deduplicateᵇ r [] = []
 deduplicateᵇ r (x ∷ xs) = x ∷ filterᵇ (not ∘ r x) (deduplicateᵇ r xs)
```
jamesmckinna, line 398: I think these functions below deserve an opening documentation comment to the effect that the first two are about the *first* occurrence in the list, while the third is about *all* occurrences, rather than having to go back to this originating PR... 

... alternatively, and against existing practice in FP libraries elsewhere, they could instead be named
* `findFirst`
* `findFirstIndex`
* `findAllIndices`

But you (and others here) may find those names a bit too verbose... ;-)?

jamesmckinna: Hmm, this comment (intended to be attached to your revised module L398) now seems to govern the original version... hope this is visible!

jamesmckinna:
I'll merge this PR once you've added some comments to `Data.List.base` specifying the first/all intended behaviour.

Saransh-cpp:
Thanks for the review! Added a few comments.

src/Data/List/Base.agda
jamesmckinna: Sorry to nitpick! The style guide asks for sentence capitalisation of comments, rather than all-lower-case. Can you fix this please?

Saransh-cpp: Missed that. Thanks!

jamesmckinna:
In some discussion offstage with @JacquesCarette we had pondered the trade-offs between documentary comments and the (possibly more) informative function names I'd suggested above. But let's leave this for another time!

PR #2043, opened 2023-08-03
Add `drop-drop` in `Data.List.Properties`

Saransh-cpp:
Dropping m elements and then n elements is same as dropping n+m elements

Reference: `dropFusion` in `Idris` - https://github.com/idris-lang/Idris2/blob/b7bda5e96df62e8aee0af8c5cfb2444db29774b4/libs/base/Data/List.idr#L1020C1-L1026C53


src/Data/List/Properties.agda
```diff
@@ -824,6 +824,13 @@ drop-take-suc-tabulate : ∀ {n} (f : Fin n → A) (i : Fin n) → let m = toℕ
 drop-take-suc-tabulate f i rewrite sym (toℕ-cast (sym (length-tabulate f)) i) | sym (lookup-tabulate f i)
   = drop-take-suc (tabulate f) (cast _ i)
 
+-- Dropping m elements and then n elements is same as dropping n+m elements
+drop-drop : (n m : ℕ) → (x : List A) → drop n (drop m x) ≡ drop (n + m) x
```
gallais, line 828: The proof is probably a lot easier if you reformulate it as `drop n (drop m xs)  ≡ drop (m + n) xs`.
(Also can we please stick to plural names for lists?)

Saransh-cpp: Thanks for the review! By a lot easier, do you mean less lines or less complexity? The following variants work -

```agda
drop-fusion' : (n m : ℕ) → (xs : List A) → drop n (drop m xs) ≡ drop (m + n) xs
drop-fusion' ℕ.zero m xs rewrite +-identityʳ m = refl
drop-fusion' (suc n) ℕ.zero [] = refl
drop-fusion' (suc n) (suc m) [] = refl
drop-fusion' (suc n) ℕ.zero (x ∷ xs) = refl
drop-fusion' (suc n) (suc m) (x ∷ xs) = drop-fusion' (suc n) m xs
```

```agda
drop-fusion' : (n m : ℕ) → (xs : List A) → drop n (drop m xs) ≡ drop (m + n) xs
drop-fusion' ℕ.zero m xs rewrite +-identityʳ m = refl
drop-fusion' (suc n) ℕ.zero xs = refl
drop-fusion' (suc n) (suc m) [] = refl
drop-fusion' (suc n) (suc m) (x ∷ xs) = drop-fusion' (suc n) m xs
```

```agda
drop-fusion' : (n m : ℕ) → (xs : List A) → drop n (drop m xs) ≡ drop (m + n) xs
drop-fusion' n ℕ.zero [] = refl
drop-fusion' n (suc m) [] = drop-[] n
drop-fusion' ℕ.zero m (x ∷ xs) rewrite +-identityʳ m = refl
drop-fusion' (suc n) ℕ.zero (x ∷ xs) = refl
drop-fusion' (suc n) (suc m) (x ∷ xs) = drop-fusion' (suc n) m xs
```

Out of curiosity, how is one supposed to choose between multiple correct definitions for the same proof?

gallais: I mean that you don't need to split on `n` because both `_+_` will compute just
enough for the inner drop to step. Cf.

```
drop-drop : (m n : ℕ) → (xs : List A) → drop n (drop m xs) ≡ drop (m + n) xs
drop-drop zero n xs = refl
drop-drop (suc m) n [] = drop-[] n
drop-drop (suc m) n (x ∷ xs) = drop-drop m n xs
```

Saransh-cpp: Ah, thanks for the clarification! I'll get this in, that was very helpful!

Saransh-cpp: Okay now I know where I went wrong - I swapped `n` and `m` _only_ in `drop (n + m) xs` and the type checker refused to simplify the proof. Spent a lot of time in confusion but everything works now! 😬

src/Data/List/Properties.agda
```diff
@@ -824,6 +824,12 @@ drop-take-suc-tabulate : ∀ {n} (f : Fin n → A) (i : Fin n) → let m = toℕ
 drop-take-suc-tabulate f i rewrite sym (toℕ-cast (sym (length-tabulate f)) i) | sym (lookup-tabulate f i)
   = drop-take-suc (tabulate f) (cast _ i)
 
+-- Dropping m elements and then n elements is same as dropping n+m elements
```
Saransh-cpp, line 827:
```suggestion
-- Dropping m elements and then n elements is same as dropping m+n elements
```

Saransh-cpp:
I found a `drop-fusion` and a `drop-drop-fusion` in agda-stdlib -

https://github.com/agda/agda-stdlib/blob/9a240ee51ac9b2b541ceab6effcc325eb11c96ab/src/Codata/Guarded/Stream/Properties.agda#L271-L273

https://github.com/agda/agda-stdlib/blob/9a240ee51ac9b2b541ceab6effcc325eb11c96ab/src/Codata/Sized/Colist/Properties.agda#L200-L204

also `map-fusion` and a few `map-map-fusion` were renamed as `map-∘` -

https://github.com/agda/agda-stdlib/blob/9a240ee51ac9b2b541ceab6effcc325eb11c96ab/src/Codata/Guarded/Stream/Properties.agda#L329

https://github.com/agda/agda-stdlib/blob/9a240ee51ac9b2b541ceab6effcc325eb11c96ab/src/Codata/Sized/Colist/Properties.agda#L349

Should these names be uniform?

jamesmckinna:
Good question re names. Always a tricky problem picking good ones, and the style-guide heuristics are just that, heuristics. 

`fusion` is probably too ambiguous (hmmm?), and there's no composition *per se*, and `drop-∘` isn't going to work...

... perhaps `drop-+` or else `drop-drop-+`? The conventions identified in your comments suggest the former, perhaps? More verbose would be something like `drop∘drop=drop-+`, but that's too heavy, I think. 

As for the names in `Codata.*`, I'm not sure that standardising those belongs in this PR (but I probably missed/ignored it in a previous one), but if you come up with a compelling enough name (perhaps `drop∘drop` is the best, after all?) then perhaps it is worth doing?

gallais:
Names typically only include the shape of the LHS. So `drop-drop`.

Saransh-cpp:
Great! Should I then update the names of `drop-fusion` and `drop-drop-fusion` in `Codata` or leave them as it is (or leave them for another PR)?

JacquesCarette:
It would make sense to me to update the names of `drop-fusion` in `Codata` in this PR as well.

PR #2045, opened 2023-08-04
Add properties of vector selectors, `Vec.reverse`, `Vec._++_` and `Vec.fromList`

shhyou:
~~(This PR depends on the change in #2041.)~~ EDIT: merged now

This PR adds three sets of properties to `Data.Vec.Properties`. ~~The first set of properties enable reasoning about functions that look like this (simplified version):~~

```
zipMap : Vec A m → (A → Vec B m) → Vec B m
zipMap []       fs = []
zipMap (x ∷ xs) fs = head (fs x) ∷ zipMap xs (tail ∘ fs)

zipMap-reverse : reverse (zipMap xs fs) ≡ zipMap (reverse xs) (reverse ∘ fs)
```

~~In the code, `zipMap` uses its second argument `fs : A → Vec B m` as if it were a vector of functions (i.e. `Vec (A → B) m`) and apply each function to an element from `xs` in parallel. The property `zipMap-reverse` states that `reverse` after `zipMap` equals reversing the input list and composing `fs` with `reverse`.~~

~~Since `zipMap` uses `head` and `tail` to destruct the result of `fs`, the proof of `zipMap-reverse` needs to reason about `zipMap` and `_∷ʳ_` (snoc), which in turn can be described using `init` and `last`. Therefore the first set of properties prove the commutativity / equivalence of the selectors `head`, `tail`, `init` and `last`.~~

1.  ~~head-singleton : head ≗ last {A = A} {n = zero}  
    head-init : head ≗ head ∘ init  
    last-tail : last ≗ last ∘ tail  
    init-tail : init ∘ tail ≗ tail ∘ init~~
    ```
    init-reverse : init ∘ reverse ≗ reverse ∘ tail  
    last-reverse : last ∘ reverse ≗ head
    ```

2. The second set of properties concerns `reverse` and `_++_`, analogous to that of lists'.
    ```
    ++-assoc : ∀ .(eq : (n + m) + o ≡ n + (m + o)) xs ys zs →
               cast eq ((xs ++ ys) ++ zs) ≡ xs ++ (ys ++ zs)
    reverse-++ : ∀ .(eq : n + m ≡ m + n) xs ys →
                 cast eq (reverse (xs ++ ys)) ≡ reverse ys ++ reverse xs
    cast-reverse : ∀ .(eq : n ≡ m) → cast eq ∘ reverse {A = A} {n = n} ≗ reverse ∘ cast eq
    ```

3. The third set of properties transports the equivalence of lists to the equivalence of vectors produced by `fromList`. I am not certain if these are appropriate for the standard library.
    ```
    cong-fromList : ∀ {xs ys : List A} (eq : xs ≡ ys) →
                    cast (cong List.length eq) (fromList xs) ≡ fromList ys
    fromList-map : ∀ (f : A → B) (xs : List A) →
                   cast (Listₚ.length-map f xs) (fromList (List.map f xs)) ≡ map f (fromList xs)

    fromList-++ : ∀ (xs : List A) {ys : List A} →
                  cast (Listₚ.length-++ xs) (fromList (xs List.++ ys)) ≡ fromList xs ++ fromList ys
    ```

I am open to changing the properties to be added or dropping some of them. While I think these properties are “basic” and general, I am not certain if they are really useful or sufficiently non-trivial to be included in the standard library. For example, I have one use of each of the `fromList` properties in the third set, but that's a result of a rather unfortunate choice of definition when I started my project.


src/Data/Vec/Properties.agda
```diff
@@ -477,6 +488,11 @@ map-⊛ f g (x ∷ xs) = cong (f x (g x) ∷_) (map-⊛ f g xs)
 ++-injective ws xs eq =
   (++-injectiveˡ ws xs eq , ++-injectiveʳ ws xs eq)
 
+++-assoc : ∀ .(eq : (n + m) + o ≡ n + (m + o)) (xs : Vec A n) (ys : Vec A m) (zs : Vec A o) →
```
jamesmckinna, line 491: I think it would be better for the order of length indices to be `m` `n` `o`, ie.
`++-assoc : ∀ .(eq : (m + n) + o ≡ m + (n + o)) (xs : Vec A m) (ys : Vec A n) (zs : Vec A o) →`
(But check the style guide as well)

src/Data/Vec/Properties.agda
```diff
@@ -927,6 +957,38 @@ map-reverse f (x ∷ xs) = begin
   reverse (f x ∷ map f xs)  ≡⟨⟩
   reverse (map f (x ∷ xs))  ∎
 
+-- append and reverse
+
+reverse-++ : ∀ .(eq : n + m ≡ m + n) (xs : Vec A n) (ys : Vec A m) →
```
jamesmckinna, line 962: Ditto. here

src/Data/Vec/Properties.agda
```diff
@@ -1100,6 +1162,31 @@ toList∘fromList : (xs : List A) → toList (fromList xs) ≡ xs
 toList∘fromList List.[]       = refl
 toList∘fromList (x List.∷ xs) = cong (x List.∷_) (toList∘fromList xs)
 
+toList-cast : ∀ .(eq : m ≡ n) (xs : Vec A m) → toList (cast eq xs) ≡ toList xs
+toList-cast {n = zero}  eq []       = refl
+toList-cast {n = suc _} eq (x ∷ xs) =
+  cong (x List.∷_) (toList-cast (cong pred eq) xs)
+
+cast-fromList : ∀ {xs ys : List A} (eq : xs ≡ ys) →
```
jamesmckinna, line 1170: I don't think this lemma is needed: one way to view it is as the composition of `cong fromList` and `cast-is-id`, or am I missing something?

jamesmckinna:
The definition of `zipMap` (thankfully not part of the PR) doesn't really make sense to me, but perhaps the actual usage you have in mind does make the other lemmas make sense. But given that the proofs of all the lemmas in 1) are given by `refl` after a suitable irrefutable pattern-match, I'm not completely clear what they add to the library... ie. in any client module, could they (or their consequence) not already be provable by matching?

Groups 2) and 3) make sense, not least because `fromList` does seem to necessitate the use of `cast`, except, I think, in the first lemma in group 2), as reviewed above.

src/Data/Vec/Properties.agda
shhyou: Yes, it is sort of `cong`ing `fromList` onto the equality, but the dependency in the types makes it to be not quite just adding `fromList` after `cast-is-id`. That is, the `refl` in your implementation substituted two types and rather than just one. My best attempt at rewriting it using the standard library is:

```agda
cast-fromList′ : ∀ {a} {A : Set a} {xs ys : List A} →
                 (eq : xs ≡ ys) → cast (cong List.length eq) (fromList xs) ≡ fromList ys
cast-fromList′ {xs = xs} eq =
  dsubst₂ (λ ys eq → cast eq (fromList xs) ≡ fromList ys) {y₁ = refl} eq refl (cast-is-id _ (fromList xs))
```
which I think is rather difficult to write in an inline style (for usages in equational reasoning). Therefore I think it is worth having a separate lemma.

There's a minor issue on how the lemma should compute. The implementation in the PR computes on the list, whereas the `cong`-like implementation computes on `eq`. I don't have strong opinions on which should be the case.

src/Data/Vec/Properties.agda
shhyou: Done! Thanks for noticing this.

src/Data/Vec/Properties.agda
shhyou: Done.

shhyou:
> **[jamesmckinna said:](https://github.com/agda/agda-stdlib/pull/2045#issuecomment-1674535026)**
>
> The definition of `zipMap` (thankfully not part of the PR) doesn't really make sense to me, but perhaps the actual usage you have in mind does make the other lemmas make sense.

The actual function I am using is the substitution of a vector of terms which is defined by applying the single-term substitution function `tmsubst` to each term in the vector:

```agda
substs : ∀ {m Δ Δ′ τ} {σty : Δ → Ty Δ′} →
  Vec (Term Δ τ) m →
  (σtms : (x : Δ) → Vec (Term Δ′ (σty x)) m) →
  Vec (Term Δ′ (tysubst τ σty)) m
substs []         σtms = []
substs (tm ∷ tms) σtms = tmsubst tm (head ∘ σtms) ∷ substs tms (tail ∘ σtms)
```

So the `fs` in `zipMap` is sort of the environment `σtms` plus `tmsubst`. But I admit that the choice of the setup isn't ideal in the first place. Perhaps I should try `Data.Vec.Functional`.

> But given that the proofs of all the lemmas in 1) are given by `refl` after a suitable irrefutable pattern-match, I'm not completely clear what they add to the library... ie. in any client module, could they (or their consequence) not already be provable by matching?

To prove that applying `substs` to equivalent environments (`σtms`) produce equal results, there is no way to pattern match on the vector (the result of calling `σtms`) because `tmsubst` decides how to use the environment. Specifically, I have to resort to equational reasoning:

```agda
substs-≗ : ∀ {m Δ Δ′ τ} {σty : Δ → Ty Δ′} →
  {σtms σtms′ : (x : Δ) → Vec (Term Δ′ (σty x)) m} →
  (terms : Vec (Term Δ τ) m) →
  (∀ x → σtms x ≡ σtms′ x) → -- this, the equational reasoning part
  substs terms σtms ≡ substs terms σtms′

tmsubst-≗ : ∀ {Δ Δ′ τ} {σty : Δ → Ty Δ′} →
  (term : Term Δ τ) →
  {σtm σtm′ : (x : Δ) → Term Δ′ (σty x)} →
  (∀ x → σtm x ≡ σtm′ x) → -- needs equational reasoning
  tmsubst term σtm ≡ tmsubst term σtm′
```

To show that `substs` commute with `reverse`, I have to construct (extensional) equalities of type `∀ a → σtms a ≡ σtms′ a`. Therefore functions in the first group such as `init-tail` become handy.

jamesmckinna:
Thanks for the thoughtful response to my critique, and for the actual intended mode-of-use involving `Vec`torised term substitution...

... about which I also have an amount of experience/knowledge (see recent co-authored papers of @gallais  about renaming and substitution in the setting of generic syntax, as well as ancient (!) ones of mine from the 1990s), the TL;DR of which is probably:
* *don't* define parallel (vectorised) substitution in terms of single-variable substitution
* *do* define single-variable substitution as a specialised instance of parallel substitution

Independent corroboration of these ideas (search term "context morphisms") comes from work by the Autosubst(2) group, whose members were based originally in Saarbruecken and now also elsewhere (including my own department), eg [at CPP 2019](https://dl.acm.org/doi/10.1145/3293880.3294101)

jamesmckinna:
As for 'matching on data' vs. 'matching on equations', indeed, if you want the latter arguments to be marked as irrelevant, then my solution won't straightforwardly work. But will you ever use `cong-fromList` in a computationally relevant context, or only opaquely in the course of establishing properties of your functions? 

Still, I fully accept the argument about pragmatics of laying out Feijen-style equational reasoning proofs... so the definition can stand! The proof I gave was definitely a fast(er)-path to it, however...

jamesmckinna:
Oh, and for the `head`/`tail` programming pattern, indeed it may very well be that functional vectors prove ot be a (much) better fit for such modes of use, modulo function extensionality...

shhyou:
Thanks a lot for the review and the pointers to the papers!

> Oh, and for the `head`/`tail` programming pattern, indeed it may very well be that functional vectors prove to be a (much) better fit for such modes of use, modulo function extensionality...

I removed the properties in group 1 (`head`, `tail`, `init` and `last` thingy) from this pull request while still keeping `cast-fromList` in. After checking my original code, I figured that lemmas in group 1 were less trivial (at least not just pattern matching) because `init` and `last` were defined using views, not recursion. But since that's going to be fixed in #2041 and #2056 I think it'd be acceptable to omit `init`/`last` properties from the PR.

> As for 'matching on data' vs. 'matching on equations', indeed, if you want the latter arguments to be marked as irrelevant, then my solution won't straightforwardly work. But will you ever use `cong-fromList` in a computationally relevant context, or only opaquely in the course of establishing properties of your functions?
> 
> Still, I fully accept the argument about pragmatics of laying out Feijen-style equational reasoning proofs... so the definition can stand! The proof I gave was definitely a fast(er)-path to it, however...

No, I've never noticed the computational content of lemmas like `cast-fromList`. The only reason is that the other lemmas about `cast` seems to have similar computation characteristic. The equality argument actually can't be marked as irrelevant because `Listₚ.∷-injective` uses it in the proof. I am happy to change the proof if matching on `eq` seems better!

> Thanks for the thoughtful response to my critique, and for the actual intended mode-of-use involving `Vec`torised term substitution...

Thanks for the pointers! I'll look into the JFP paper more closely. The CPP'19 paper also looks very cool.

Yes, I am using parallel substitution that replaces all variables at once. The function `substs` on the other hand is computing `[e_1[x_11↦e'_11,...,x_1m↦e'_1m], e_2[x_21↦e'_21,...,x_2m↦e'_2m], ..., e_n[x_n1↦e'_n1,...,x_nm↦e'_nm]]`, that's why an extra vector dimension appears.

src/Data/Vec/Properties.agda
```diff
@@ -1130,6 +1182,31 @@ toList∘fromList : (xs : List A) → toList (fromList xs) ≡ xs
...
+                cast (cong List.length eq) (fromList xs) ≡ fromList ys
+cast-fromList {xs = List.[]}     {ys = List.[]}     eq = refl
+cast-fromList {xs = x List.∷ xs} {ys = y List.∷ ys} eq = begin
+  x ∷ cast (cong (pred ∘ List.length) eq) (fromList xs) ≡⟨ cong (_ ∷_) (cast-fromList xs-equals-ys) ⟩
+  x ∷ fromList ys                                       ≡⟨ cong (_∷ _) x-equals-y ⟩
+  y ∷ fromList ys                                       ∎
+  where
+  x-equals-y = proj₁ (Listₚ.∷-injective eq)
```
MatthewDaggitt, line 1198: The names `x≡y` and `xs≡ys` would be shorter and more traditional. Also could we get away with `(x≡y, xs≡ys) = Listₚ.∷-injective eq`?

MatthewDaggitt (commented):
Otherwise this looks like a great first PR, thank you! Once @jamesmckinna is happy, I'll merge this in.

src/Data/Vec/Properties.agda
shhyou: Thanks for the suggestion! I changed them to irrefutable `let`.

jamesmckinna:
Regarding merging, I think that this PR is blocking on merging of #2041 (which should be a bit more straightforward).

jamesmckinna:
Regarding `zipMap`, please also see, and think about the relationship with, `Data.Vec.Base.diagonal` and `Data.Vec.Base.DiagonalBind._>>=_` which I believe already give you the functionality you were looking for under 1) above?

jamesmckinna:
Does this now pass the tests after the merge of #2041? Another `rebase` required?

src/Data/Vec/Properties.agda
```diff
@@ -1130,6 +1182,29 @@ toList∘fromList : (xs : List A) → toList (fromList xs) ≡ xs
 toList∘fromList List.[]       = refl
 toList∘fromList (x List.∷ xs) = cong (x List.∷_) (toList∘fromList xs)
 
+toList-cast : ∀ .(eq : m ≡ n) (xs : Vec A m) → toList (cast eq xs) ≡ toList xs
```
jamesmckinna, line 1185: Same remark as for `fromList-cast`: but this one, I think, can be analysed in terms of `cong toList`? or not here, either?

Taneb: I don't think it can.  `cast eq xs` and `xs` have types `Vec A m` and `Vec A n` where `m` and `n` are only propositionally, not definitionally equal, so `cong toList` won't type-check here.

jamesmckinna: D'oh! Suggest perhaps relativising instead to:
```agda
toList-cast : ∀ .(eq : m ≡ n) (xs : Vec A m) {ys} → cast eq xs ≡ ys toList (cast eq xs) ≡ toList ys
```
and similarly for `fromList-cast`? or does this suffer from the same problem (you'd think I could think this through myself, but... brain-fog/code-blindness)?

jamesmckinna: :man_facepalming:

jamesmckinna:
I'm happy now, or will be, once the tests pass... ;-) But if #2056 gets merged first, which I think is desirable, then various other things here might simplify further? or not?

src/Data/Vec/Properties.agda
shhyou: FWIW, `toList-cast` was introduced in #1839. I only moved the definition down so that `fromList` properties and `toList` properties are grouped together.

shhyou:
> Regarding `zipMap`, please also see, and think about the relationship with, `Data.Vec.Base.diagonal` and `Data.Vec.Base.DiagonalBind._>>=_` which I believe already give you the functionality you were looking for under 1) above?

The properties in group 1 (that are not included in the PR anymore) are intended for the proof that `reverse` and an `zipMap`-ish function commute, therefore `diagonal` doesn't address my use case.

jamesmckinna:
My question was really: what's the relationship between your `zipMap` and `DiagonalBind._>>=_`?

shhyou:
They look the same to me, ha!

MatthewDaggitt:
Merging in then. Thanks again for the PR @shhyou!

shhyou:
Thank you and thanks to @jamesmckinna for the review!

Issue #1996, opened 2023-06-19
[Discussion]: Moving `setoid`, `≡-setoid`, and `flip` away from `Function.Equality`


Saransh-cpp:
Related to #759

Now that agda-stdlib has a `Function.Indexed` directory, I think it would be nice to migrate `setoid`, `≡-setoid`, and `flip` to the new hierarchy.

Along with these, I am unsure if other functions in `Function.Equality` must also be migrated somewhere. The issue mentions - "Missing an indexed function variant. Should go in a new hierarchy `Function.Dependent.(Structures/Bundles)`" - but I'm not sure if I understand that very well. Should the operators and functions not involving `IndexedSetoid` have a new indexed version of themselves in a different module, or should the indexed functions be moved to `Function.Dependent.(Structures/Bundles)`?

Moreover, should I start by moving `_⇨_`, `Π`, `_⟶_`, `setoid`, and `≡-setoid` to `Function.Indexed.Relation.Binary.Equality`, given that all of them have `IndexedSetoid` involved within their definitions?

Thanks!

MatthewDaggitt:
The remaining unported functions in `Function.Equality` are  `setoid`, `_⇨_`, `≡-setoid`, `flip` and `→-to-⟶`. Once again, they're slightly weird constructs so I think I'm going to omit them, and see if we get any complaints.

JacquesCarette:
The first two (`setoid`, `_⇨_`) ,were needed, so there's now PRs for them.

PR #2047, opened 2023-08-08
Move `≡-setoid` to `Function.Indexed.Relation.Binary.Equality`

Saransh-cpp:
Related to #759 #1996

I could not get rid of the `_⟨$⟩_` imports from `Function.Equality`. Removing the occurrence of `_⟨$⟩_` in the files was not as straightforward as I initially thought 🙂


src/Function/Equality.agda
```diff
@@ -11,6 +11,7 @@
 -- `Bijection`. The alternative definitions found in this file will
 -- eventually be deprecated.
 
+-- check if modules are actually using Function.Equality and port them
```
MatthewDaggitt, line 14: Could we get rid of these temporary notes before merging?

MatthewDaggitt (commented):
Otherwise looks good! Thanks.

src/Function/Equality.agda
Saransh-cpp: Oh oops, those are my notes, pushed them by mistake. Thanks!

Saransh-cpp:
Thanks for the review. @MatthewDaggitt! Should this be documented in the CHANGELOG somewhere?

MatthewDaggitt:
> Should this be documented in the CHANGELOG somewhere?

No that's okay.

Issue #2048, opened 2023-08-08
Association lists and their functions


Saransh-cpp:
I am currently looking to port [`lookup`](https://github.com/idris-lang/Idris2/blob/badf1e98c8812e8b54258754cf47782f61d8fb15/libs/base/Data/List.idr#L184-L185), [`lookupBy`](https://github.com/idris-lang/Idris2/blob/badf1e98c8812e8b54258754cf47782f61d8fb15/libs/base/Data/List.idr#L174-L180), [`insertAt`](https://github.com/idris-lang/Idris2/blob/badf1e98c8812e8b54258754cf47782f61d8fb15/libs/base/Data/List.idr#L228-L230), and [`deleteAt`](https://github.com/idris-lang/Idris2/blob/badf1e98c8812e8b54258754cf47782f61d8fb15/libs/base/Data/List.idr#L241-L243) from  Idris2, but I am not sure if they are exactly required in Agda. It does look like that they do not exist in Agda. If they should be added in, should the API of `lookup` and `lookupBy` stay the same, or should it be changed (given that I could not find any such pair of functions in Agda)? Thanks!

jamesmckinna:
`deleteAt` appears to be the `_─_` operation? 

No answer to why there isn't (?) the companion `insertAt` (though there is in `Data.Vec.*` etc. I think ... this may be yet another knock-on consequence of `List` and `Vec` not being mutually consistent)

As for association lists themselves, I'd be tempted to organise them under `Data.List.Association.*` rather than try to shoehorn them into `Data.List.Base`? They're a distinguished mode of use of lists, satisfying the finite Map API, and that little theory seems worth abstracting out on its own? Just my opinion, though!

jamesmckinna:
As to the question about API design, as a precursor to eventual implementation, should we consider more deliberately discussing those as separate issues/PRs/threads on the Zulip (under the #stdlib channel?)

jamesmckinna:
And now returning to the main question: how much of `idris2`'s (or that of any classical/orthodox Boolean-valued/simply-typed approach) behaviour on association lists can be realised in terms of the `...Relation.Unary.Any` machinery, suitably specialised? Question prompted by threading back through the definitions in `Data.List.Membership` etc.

if it is possible to 'encode everything' that way, that doesn't stop it being independently interesting to consider versions specialised to lists of pairs, where the 'unary' predicate on pairs is given by ... a binary, Boolean-valued function.

[...]

PR #2049, opened 2023-08-09
Sync insert, remove, and update functionalities for `List` and `Vec`

Saransh-cpp:
See https://github.com/agda/agda-stdlib/issues/2048#issuecomment-1671584606

This PR -

Adds a missing function
- a new `insert` function for `List`, similar to the one available for `Vec`

Makes the insert, remove, and update functionalities of `Vec` and `List` uniform -
- a new `updateAt` function for `List` and the existing `_[_]%=_` operator defined using `updateAt`, just how it it is done for `Vec` (same functionality, different names)
- a new `_─_` operator for `Vec` to remove an element, defined in terms of the existing `Vec.Base.remove` (removing an element is achieved using `_─_` in List and `remove` in `Vec`, but given that one can update an element of `Vec` using both a function and a mixfix operator, I think the remove functionality should also be treated in the same way)
- a new `remove` function for `List` to remove an element and `Data.List._─_` defined in terms of `Data.List.remove` (the point above)

I have not updated the CHANGELOG as I wanted to discuss these changes before. Lastly, there is no corresponding mixfix operator for the `insert` functions in `Vec` and `List`, and I am not sure they are needed.


jamesmckinna:
At a first look, this seems to be a useful reconciliation of the hierarchies under the two datatypes `List` and `Vec`. A much more ambitious approach would be to develop the appropriate ornamentation gadgetry to permit this to be automated/automatable, but that is work for another time, I think! 

I suspect the original divergence was a simple matter of evolution-by-need/demand, but I'd be interested in comments from some of the 'old-timers' about the relative state of things, then and now. (`Vector` may also be worth looking at)

Otherwise, I'd suggest adding the clerical work to `CHANGELOG` now, to save effort (or forgotten entries) later?

Saransh-cpp:
Vector does have `insert`, `remove`, and `updateAt` defined, but not the mixfix operators 🙂

gallais:
`insert` is not a faithful port. The `Vec`-based one correctly identifies that there
are `1 + n` positions at which to insert in a vector of size `n`.

gallais (changes requested):
`insert` is not a faithful port. The `Vec`-based one correctly identifies that there
are `1 + n` positions at which to insert in a vector of size `n`.

jamesmckinna:
Plus `insert`-delete commutation properties (downstream PR?)

Saransh-cpp:
Thanks for the reviews!

I can see that the `insert-remove` and `remove-insert` properties would be nice along with maybe (?) `insert-lookup`, but does "downstream PR" means adding them in another repository that depends on `agda-stdlib`?

I am also not quite sure if I understood what proofs @gallais meant by "proofs that insert and others commute with toList", could you please link a similar existing proof or maybe a type signature that helps me? Thank you!

JacquesCarette:
"downstream PR" means that it might get done in a subsequent PR rather than necessarily this one.

The "proofs that insert commute with toList" mean to check that Vec.insert and List.insert do the same thing, up to change to representation. i.e. starting with a vector, if you insert into it then convert to a list should be identical to converting to a list then inserting.

Saransh-cpp:
Thank you for the help! I'll start adding them.

Saransh-cpp:
> The Vec-based one correctly identifies that there
are 1 + n positions at which to insert in a vector of size n.

Oh yes, I think that should be the case for List as well? -
```agda
insert : (xs : List A) → Fin (suc (length xs)) → A → List A
insert []       zero    v = [ v ]
insert (x ∷ xs) zero    v = v ∷ x ∷ xs
insert (x ∷ xs) (suc i) v = v ∷ insert xs i v
```

JacquesCarette:
Typo: last line should have `x` in the head.

Saransh-cpp:
Oh yes, thanks!

src/Data/List/Properties.agda
```diff
@@ -1212,3 +1236,15 @@ zipWith-identityʳ = zipWith-zeroʳ
...
+{-# WARNING_ON_USAGE length-─
+"Warning: length-─ was deprecated in v2.0.
+Please use length-remove instead."
+#-}
+
+map-─ = map-remove
+{-# WARNING_ON_USAGE map-─
+"Warning: map-─ was deprecated in v2.0.
+Please use map-remove instead."
+#-}
```
Saransh-cpp, line 1250: I am not sure if this should be done. did it to maintain uniformity, but please let me know if this should be reversed.

src/Data/Vec/Properties.agda
```diff
@@ -1099,6 +1099,20 @@ insert-punchIn xs       zero     v j       = refl
 insert-punchIn (x ∷ xs) (suc i)  v zero    = refl
 insert-punchIn (x ∷ xs) (suc i)  v (suc j) = insert-punchIn xs i v j
 
+-- where should this go? Or can I write toList-insert without this? Or does
+-- something similar already exist?
+length-length : (xs : Vec A n) → length xs ≡ List.length (toList xs)
+length-length []       = refl
+length-length (x ∷ xs) = cong suc (length-length xs)
```
Saransh-cpp, line 1106: Not sure what exactly should I do with this 🙂

src/Data/List/Properties.agda
```diff
@@ -746,17 +746,41 @@ map-∷= (x ∷ xs) zero    v f = refl
...
+remove-insert []       zero    v = refl
+remove-insert (x ∷ xs) zero    v = refl
+remove-insert (x ∷ xs) (suc i) v = cong (_ ∷_) (remove-insert xs i v)
+
+insert-remove : (xs : List A) (i : Fin (length xs)) → insert (remove xs i) (cast (sym (length-remove xs i)) i) (lookup xs i) ≡ xs
+insert-remove (x ∷ xs) zero = h xs x
+  where
+  h : ∀ (xs : List A) v → insert xs zero v ≡ v ∷ xs
+  h []       v = refl
+  h (x ∷ xs) v = refl
```
Saransh-cpp, line 782: I don't know why agda does not like -
```agda
insert-remove (x ∷ xs) zero = refl
```

src/Data/List/Properties.agda
```diff
@@ -746,17 +746,41 @@ map-∷= (x ∷ xs) zero    v f = refl
...
 
-map-─ : ∀ xs k (f : A → B) →
-        let eq = sym (length-map f xs) in
-        map f (xs ─ k) ≡ map f xs ─ cast eq k
-map-─ (x ∷ xs) zero    f = refl
-map-─ (x ∷ xs) (suc k) f = cong (f x ∷_) (map-─ xs k f)
+------------------------------------------------------------------------
+-- remove
+
+length-remove : ∀ (xs : List A) k → suc (length (remove xs k)) ≡ length xs
```
Saransh-cpp, line 759: Along with the name, I have also changed the type here (which was helpful in writing `insert-remove`). Should that also be documented somewhere in the CHANGELOG?

Saransh-cpp (commented):
This should be ready to review again!

MatthewDaggitt:
Okay, so having looked at this I agree consistency is good! But I don't think pursuing consistency by having two different names for the same operations is a good idea. Instead we should follow the standard path and deprecate one (or both!) of the names in favour of a consistent one.

- For `remove`/`_-_`. I don't actually like either of these names! Both of them look like you're removing an element rather than deleting the value at an index. I would suggest we deprecate both of them and instead call the operation `deleteAt` which is much more meaningful, and ties nicely with `updateAt`.

- For `updateAt` the first thing I notice is that the order of arguments in the `List` and `Vec` versions differ. I really think the order should be ` (A → A) → Vec A n → Fin n →Vec A n` to mimic `map` and to ensure compatibility with the list version. I think I'm willing to make a backwards incompatible change to fix the ordering of the arguments.

- Similarly I think `insert` should really be called `insertAt`...

I guess the alternative infix notation for `updateAt` isn't too bad, so I guess we could keep it...

src/Data/Vec/Properties.agda
MatthewDaggitt: I think this should be called to `length-toList` and should have the type `List.length (toList xs) ≡ length xs`

src/Data/List/Properties.agda
MatthewDaggitt: Yes should definitely be documented in the CHANGELOG :+1:

src/Data/List/Properties.agda
MatthewDaggitt: Hmm I assume somehow `cast` is blocking the reduction behaviour. @jamesmckinna this might be the sort of thing you're interested in working out?

src/Data/List/Properties.agda
MatthewDaggitt: If we accept my proposal to change the names to `removeAt` then yes this should be done.

src/Data/List/Properties.agda
jamesmckinna: I will certainly take a look, but I think similar things came up for @shhyou in #2045 [see discussion here](https://github.com/agda/agda-stdlib/pull/2045#discussion_r1292577189) so you might like to look at that discussion first. In theory, `cast` is great, because it avoids having to match on `refl` in proofs (`--without-K` hovering off-stage, plus the desire for 'only computational proof normalisation on *data*') but in practice, once it appears in a non-trivial goal involving a list `xs` and a function (such as: `length`!) defined on the *same* list, then such coupling gives rise to 'wicked' complexity. 

Possible 'uncoupling' solutions (might) include:
* introducing additional equalities on lists, to standardise apart otherwise coupled occurrences;
* doing `with ... in ... eq` abstraction over the list `length`, in order to obtain a fresh identifier for that length on which you can do pattern-matching/induction
* reconsidering the definition of `insert`, in this instance (!): I agree that you want the lemma witnessed by `h` to be a *definitional* property...

That's my 2centimes' worth of thought (without really reading the code in detail ... yet) to be going on with, but I will look at it more closely now... thanks @MatthewDaggitt for the nudge.

jamesmckinna: What happens if you telescope the first two lines of the definition of `insertAt` into the following?
```agda
insertAt xs zero v = v ∷ xs
```
Type/termination-checker both seem happy? And then your `h` would, indeed, be definitional!?

src/Data/List/Properties.agda
jamesmckinna: I'm a bit surprised that the previous type,using `pred`, didn't work, because any instance of `removeAt` *is* going to be applied to a non-empty list, and hence its length will be `NonZero`, and hence such a length will be available for `suc-pred` to apply... but again, that's me thinking out loud without reading ;-)

jamesmckinna: So, firstly, I get:
```agda
length-removeAt : ∀ (xs : List A) k → suc (length (removeAt xs k)) ≡ length xs
length-removeAt (x ∷ xs) zero    = refl
length-removeAt (x ∷ xs) (suc k) = cong suc (length-removeAt xs k)
```
as a simplification of your proof. Suggesting that perhaps the list `xs` could be made implicit? Only `k` is doing any work here...

Secondly, if I try the 'old' type, using `pred`, and exploiting your refined pattern analysis of `(length-)removeAt` this time, I obtain:
```agda
length-removeAt : ∀ (xs : List A) k → length (removeAt xs k) ≡ pred (length xs)
length-removeAt (x ∷ xs) zero    = refl
length-removeAt (x ∷ xs@(_ ∷ _)) (suc k) = begin
  length (x ∷ removeAt xs k)
    ≡⟨⟩
  suc (length (removeAt xs k))
    ≡⟨ cong suc (length-removeAt xs k) ⟩
  suc (pred (length xs))
    ≡⟨ suc-pred (length xs) ⟩
  length xs
    ≡⟨⟩
  pred (length (x ∷ xs)) ∎
```
which can of course be shortened by removing the 'obvious' `≡⟨⟩` steps...

jamesmckinna: So: for the cost of one more step of equational reasoning, involving one perhaps-non-trivial appeal to `NonZero` which is nevertheless available on observing the `xs@(_ ∷ _)` pattern, then we're good to go. As hinted at above ;-)

jamesmckinna: I'm more inclined to agree with you as regards `insertAt-removeAt` justifying the type change, though ;-)

Mind you, it's then 'easy' enough to obtain
```agda
length-removeAt′ : ∀ (xs : List A) k → length xs ≡ suc (length (removeAt xs k))
length-removeA′ xs@(_ ∷ _) k rewrite length-removeAt xs k = refl
```
as a corollary to my definition above... and then proceed as before:
```agda
insertAt-removeAt′ : (xs : List A) (i : Fin (length xs)) →
  insertAt (removeAt xs i) (cast (length-removeAt′ xs i) i) (lookup xs i) ≡ xs
insertAt-removeAt′ (x ∷ xs) zero    = refl
insertAt-removeAt′ (x ∷ xs) (suc i) = cong (x ∷_) (insertAt-removeAt′ xs i)
```

src/Data/List/Base.agda
```diff
@@ -195,6 +195,15 @@ tails : List A → List (List A)
 tails []       = [] ∷ []
 tails (x ∷ xs) = (x ∷ xs) ∷ tails xs
 
+insertAt : (xs : List A) → Fin (suc (length xs)) → A → List A
+insertAt []       zero    v = [ v ]
+insertAt (x ∷ xs) zero    v = v ∷ x ∷ xs
```
jamesmckinna, line 200: The single clause `insertAt xs zero v = v ∷ xs` solves many of the problems you have been having.

src/Data/List/Relation/Unary/Any.agda
```diff
@@ -69,7 +69,7 @@ _∷=_ {xs = xs} x∈xs v = xs List.[ index x∈xs ]∷= v
 
 infixl 4 _─_
 _─_ : {P : Pred A p} → ∀ xs → Any P xs → List A
-xs ─ x∈xs = xs List.─ index x∈xs
+xs ─ x∈xs = removeAt xs (index x∈xs)
```
jamesmckinna, line 72: Should this operation `_─_` not also be renamed to `removeAt`... if so, you'll need to remove that name from the `import` list for `List`, and use `List.removeAt` on the RHS, ...

src/Data/Vec/Base.agda
```diff
@@ -64,29 +64,27 @@ iterate : (A → A) → A → ∀ {n} → Vec A n
 iterate s z {zero}  = []
 iterate s z {suc n} = z ∷ iterate s (s z)
 
-insert : Vec A n → Fin (suc n) → A → Vec A (suc n)
-insert xs       zero     v = v ∷ xs
-insert (x ∷ xs) (suc i)  v = x ∷ insert xs i v
+insertAt : Vec A n → Fin (suc n) → A → Vec A (suc n)
+insertAt xs       zero     v = v ∷ xs
```
jamesmckinna, line 68: For `Vec`, the first line of the definition is evident. So, why not in `List` as well?

src/Data/Vec/Properties.agda
```diff
@@ -1108,20 +1123,20 @@ remove-punchOut (x ∷ y ∷ xs) {suc i} {suc j} i≢j =
...
-remove-insert : ∀ (xs : Vec A n) (i : Fin (suc n)) (v : A) →
-                remove (insert xs i v) i ≡ xs
-remove-insert xs           zero           v = refl
-remove-insert (x ∷ xs)     (suc zero)     v = refl
-remove-insert (x ∷ y ∷ xs) (suc (suc i))  v =
-  cong (x ∷_) (remove-insert (y ∷ xs) (suc i) v)
+removeAt-insertAt : ∀ (xs : Vec A n) (i : Fin (suc n)) (v : A) →
+                    removeAt (insertAt xs i v) i ≡ xs
+removeAt-insertAt xs           zero           v = refl
+removeAt-insertAt (x ∷ xs)     (suc zero)     v = refl
```
jamesmckinna, line 1131: And here as well...

src/Data/Vec/Properties.agda
```diff
@@ -1108,20 +1123,20 @@ remove-punchOut (x ∷ y ∷ xs) {suc i} {suc j} i≢j =
...
 
-insert-remove : ∀ (xs : Vec A (suc n)) (i : Fin (suc n)) →
-                insert (remove xs i) i (lookup xs i) ≡ xs
-insert-remove (x ∷ xs)     zero     = refl
-insert-remove (x ∷ y ∷ xs) (suc i)  =
-  cong (x ∷_) (insert-remove (y ∷ xs) i)
+insertAt-removeAt : ∀ (xs : Vec A (suc n)) (i : Fin (suc n)) →
+                    insertAt (removeAt xs i) i (lookup xs i) ≡ xs
+insertAt-removeAt (x ∷ xs)     zero     = refl
+insertAt-removeAt (x ∷ y ∷ xs) (suc i)  =
```
jamesmckinna, line 1138: Use an `@`-pattern: `xs@(_ ∷ _)` and then you don't need the redundant repetition of the cons on the RHS; simply `xs` suffices (think about the compiler, and the poor allocator in the run-time, if this were a vanilla functional language: you've destructed a cons cell, only to then put it back together again... ;-))

src/Data/List/Properties.agda
Saransh-cpp: Thanks for digging so deep into this, @jamesmckinna! The following -

```agda
length-removeAt′ : ∀ (xs : List A) k → length xs ≡ suc (length (removeAt xs k))
length-removeAt′ xs@(_ ∷ _) k rewrite length-removeAt xs k = refl
```
gives me -
```
Termination checking failed for the following functions:
  length-removeAt
Problematic calls:
  length-removeAt (x ∷ xs) k
```

src/Data/List/Properties.agda
Saransh-cpp: Thanks for looking at this @JacquesCarette and @MatthewDaggitt! 

The following -
```agda
insertAt xs zero v = v ∷ xs
```
does solve a lot of problems!

Saransh-cpp:
Thanks for the detailed review and information, @jamesmckinna and @MatthewDaggitt. Was very helpful! I've incorporated most of the suggestions, but I could not incorporate a few of them. These suggestions are marked as not resolved. Please let me know how I proceed with them.

src/Data/List/Properties.agda
jamesmckinna: ??? wow ??? `agda --version` ???
I tested this on `v2.6.3`, from a fairly recent fresh `cabal install Agda-2.6.3` on Ubuntu20.04.1LTS...

jamesmckinna:
Hmmm... sorry if some of the advice was bad! More head-scratching required!?

src/Data/List/Properties.agda
Saransh-cpp: Ah, that worked! Thanks!

```agda
length-removeAt : ∀ (xs : List A) k → length (removeAt xs k) ≡ pred (length xs)
length-removeAt (x ∷ xs) zero        = refl
length-removeAt (x ∷ y ∷ xs) (suc k) = cong suc (length-removeAt (y ∷ xs) k)

length-removeAt′ : ∀ (xs : List A) k → length xs ≡ suc (length (removeAt xs k))
length-removeAt′ xs@(_ ∷ _) k rewrite length-removeAt xs k = refl

 -- insertAt and removeAt

removeAt-insertAt : ∀ (xs : List A) (i : Fin (suc (length xs))) v →
  removeAt (insertAt xs i v) ((cast (sym (length-insertAt xs i v)) i)) ≡ xs
removeAt-insertAt xs       zero    v = refl
removeAt-insertAt (x ∷ xs) (suc i) v = cong (_ ∷_) (removeAt-insertAt xs i v)

insertAt-removeAt : (xs : List A) (i : Fin (length xs)) →
  insertAt (removeAt xs i) (cast (length-removeAt′ xs i) i) (lookup xs i) ≡ xs
insertAt-removeAt (x ∷ xs) zero    = refl
insertAt-removeAt (x ∷ xs) (suc i) = cong (x ∷_) (insertAt-removeAt xs i)
```

Saransh-cpp:
Re:

> For updateAt the first thing I notice is that the order of arguments in the List and Vec versions differ. I really think the order should be  (A → A) → Vec A n → Fin n →Vec A n to mimic map and to ensure compatibility with the list version. I think I'm willing to make a backwards incompatible change to fix the ordering of the arguments

Should I also reorder the arguments of the proofs involving `updateAt`, keeping in mind https://github.com/agda/agda-stdlib/pull/2051#discussion_r1325908915?

Saransh-cpp:
> Hmmm... sorry if some of the advice was bad! More head-scratching required!?

Not at all, all the comments were extremely helpful! The only proofs I could not simplify today are `Vec.insertAt-removeAt` and `Vec.removeAt-insertAt`. I'll work on them this weekend.

README/Nary.agda
```diff
@@ -14,7 +14,7 @@ open import Data.Nat.Base
 open import Data.Nat.Properties
 open import Data.Fin using (Fin; fromℕ; #_; inject₁)
 open import Data.List
-open import Data.List.Properties
+open import Data.List.Properties hiding (removeAt-insertAt)
```
MatthewDaggitt, line 17: No longer need to hide this, as it doesn't clash.

src/Data/List/Countdown.agda
```diff
@@ -20,7 +20,7 @@ open import Function.Base
 open import Function.Bundles
   using (Injection; module Injection)
 open import Data.Bool.Base using (true; false)
-open import Data.List.Base hiding (lookup)
+open import Data.List.Base hiding (lookup; insertAt)
```
MatthewDaggitt, line 23: Again, think this can be removed

src/Data/List/Relation/Unary/Any.agda
```diff
@@ -67,9 +67,8 @@ infixr 5 _∷=_
 _∷=_ : {P : Pred A p} → Any P xs → A → List A
 _∷=_ {xs = xs} x∈xs v = xs List.[ index x∈xs ]∷= v
 
-infixl 4 _─_
-_─_ : {P : Pred A p} → ∀ xs → Any P xs → List A
-xs ─ x∈xs = xs List.─ index x∈xs
```
MatthewDaggitt, line 72: This is a slightly different operation, it doesn't take an index, so I don't think `removeAt` is maybe the best name for this one.

I think for the sake of not letting this PR grow out of control, let's keep this function as is in this PR!

src/Data/List/Properties.agda
MatthewDaggitt: Yes it would be great to preserve the original type :+1:

src/Data/List/Base.agda
```diff
@@ -196,6 +196,14 @@ tails : List A → List (List A)
 tails []       = [] ∷ []
 tails (x ∷ xs) = (x ∷ xs) ∷ tails xs
 
+insertAt : (xs : List A) → Fin (suc (length xs)) → A → List A
```
MatthewDaggitt, line 199: `insertAt` should come after `updateAt` and be implemented in terms of it as is done in `Vec.Base`

MatthewDaggitt (commented):
> Should I also reorder the arguments of the proofs involving updateAt, keeping in mind https://github.com/agda/agda-stdlib/pull/2051#discussion_r1325908915?

I wouldn't do it at the cost of breaking backwards compatability, but if they are new proofs that's fine. 

I've also slightly changed my mind about the ordering of `updateAt`. I think the `Vec.Base` one should follow the same order as the `List.Base` definition, i.e. container, index, function.

src/Data/List/Relation/Unary/Any.agda
jamesmckinna: Yes: to keeping it out of the PR :+1:
No: (at least with my dependently typed spectacles on) it *does* take an (abstract!) 'index', `x∈xs`, but one which is more interestingly typed, as `Any P xs`. 
But let's take it off the table for the time being ;-)

src/Data/List/Base.agda
jamesmckinna: And such appeal to DRY would have avoided the detour via the previous troublesome 3-clause definition of `insertAt`.

src/Data/List/Properties.agda
```diff
@@ -746,17 +746,39 @@ map-∷= (x ∷ xs) zero    v f = refl
...
--- _─_
+-- insertAt
 
-length-─ : ∀ (xs : List A) k → length (xs ─ k) ≡ pred (length xs)
-length-─ (x ∷ xs) zero        = refl
-length-─ (x ∷ y ∷ xs) (suc k) = cong suc (length-─ (y ∷ xs) k)
+length-insertAt : ∀ (xs : List A) (i : Fin (suc (length xs))) v →
+                  length (insertAt xs i v) ≡ suc (length xs)
+length-insertAt xs       zero    v = refl
+length-insertAt (x ∷ xs) (suc i) v = cong suc (length-insertAt xs i v)
```
jamesmckinna, line 755: And so... to be systematic, does that mean there should also be a `length-updateAt`, and then `length-insertAt` as a corollary...?

src/Data/List/Base.agda
jamesmckinna: Wait: `updateAt` doesn't disturb the underlying *structure* of the `Vec`, it simply modifies elements. `insertAt` *does* change the structure!

src/Data/List/Properties.agda
jamesmckinna: But again, please, in this clause
```agda
length-removeAt (x ∷ y ∷ xs) (suc k) = cong suc (length-removeAt (y ∷ xs) k)
```
consider instead
```agda
length-removeAt (x ∷ xs@(_ ∷ _)) (suc k) = cong suc (length-removeAt xs k)
```

src/Data/List/Base.agda
MatthewDaggitt: Yes, good spot. This was a nonsense comment. I was getting confused between `setAt` and `insertAt`

Saransh-cpp:
>  I think the Vec.Base one should follow the same order as the List.Base definition, i.e. container, index, function.

This is how changing the order of args look like (the commented proofs are the ones I could not migrate) -

```diff
diff --git a/src/Data/Vec/Base.agda b/src/Data/Vec/Base.agda
index d4c0d3b58..2aeb17619 100644
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -72,16 +72,16 @@ removeAt : Vec A (suc n) → Fin (suc n) → Vec A n
 removeAt (_ ∷ xs)     zero     = xs
 removeAt (x ∷ y ∷ xs) (suc i)  = x ∷ removeAt (y ∷ xs) i
 
-updateAt : Fin n → (A → A) → Vec A n → Vec A n
-updateAt zero    f (x ∷ xs) = f x ∷ xs
-updateAt (suc i) f (x ∷ xs) = x   ∷ updateAt i f xs
+updateAt : Vec A n → Fin n → (A → A) → Vec A n
+updateAt (x ∷ xs) zero    f = f x ∷ xs
+updateAt (x ∷ xs) (suc i) f = x   ∷ updateAt xs i f
 
 -- xs [ i ]%= f  modifies the i-th element of xs according to f
 
 infixl 6 _[_]%=_ _[_]≔_
 
 _[_]%=_ : Vec A n → Fin n → (A → A) → Vec A n
-xs [ i ]%= f = updateAt i f xs
+xs [ i ]%= f = updateAt xs i f
 
 -- xs [ i ]≔ y  overwrites the i-th element of xs with y
 
diff --git a/src/Data/Vec/Properties.agda b/src/Data/Vec/Properties.agda
index eaea4d12e..eea2fe3d9 100644
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -229,14 +229,14 @@ lookup-inject≤-take (suc (suc m)) (s≤s (s≤s m≤m+n)) (suc (suc i)) (x ∷
 -- (+) updateAt i actually updates the element at index i.
 
 updateAt-updates : ∀ (i : Fin n) {f : A → A} (xs : Vec A n) →
-                   xs [ i ]= x → (updateAt i f xs) [ i ]= f x
+                   xs [ i ]= x → (updateAt xs i f) [ i ]= f x
 updateAt-updates zero    (x ∷ xs) here        = here
 updateAt-updates (suc i) (x ∷ xs) (there loc) = there (updateAt-updates i xs loc)
 
 -- (-) updateAt i does not touch the elements at other indices.
 
 updateAt-minimal : ∀ (i j : Fin n) {f : A → A} (xs : Vec A n) →
-                   i ≢ j → xs [ i ]= x → (updateAt j f xs) [ i ]= x
+                   i ≢ j → xs [ i ]= x → (updateAt xs j f) [ i ]= x
 updateAt-minimal zero    zero    (x ∷ xs) 0≢0 here        = contradiction refl 0≢0
 updateAt-minimal zero    (suc j) (x ∷ xs) _   here        = here
 updateAt-minimal (suc i) zero    (x ∷ xs) _   (there loc) = there loc
@@ -259,13 +259,13 @@ updateAt-minimal (suc i) (suc j) (x ∷ xs) i≢j (there loc) =
 
 updateAt-id-local : ∀ (i : Fin n) {f : A → A} (xs : Vec A n) →
                     f (lookup xs i) ≡ lookup xs i →
-                    updateAt i f xs ≡ xs
+                    updateAt xs i f ≡ xs
 updateAt-id-local zero    (x ∷ xs) eq = cong (_∷ xs) eq
 updateAt-id-local (suc i) (x ∷ xs) eq = cong (x ∷_) (updateAt-id-local i xs eq)
 
 -- 1b. identity:  updateAt i id ≗ id
 
-updateAt-id : ∀ (i : Fin n) (xs : Vec A n) → updateAt i id xs ≡ xs
+updateAt-id : ∀ (i : Fin n) (xs : Vec A n) → updateAt xs i id ≡ xs
 updateAt-id i xs = updateAt-id-local i xs refl
 
 -- 2a. local composition:  f ∘ g = h ↾ (lookup xs i)
@@ -273,15 +273,15 @@ updateAt-id i xs = updateAt-id-local i xs refl
 
 updateAt-∘-local : ∀ (i : Fin n) {f g h : A → A} (xs : Vec A n) →
                          f (g (lookup xs i)) ≡ h (lookup xs i) →
-                         updateAt i f (updateAt i g xs) ≡ updateAt i h xs
+                         updateAt (updateAt xs i g) i f ≡ updateAt xs i h
 updateAt-∘-local zero    (x ∷ xs) fg=h = cong (_∷ xs) fg=h
 updateAt-∘-local (suc i) (x ∷ xs) fg=h = cong (x ∷_) (updateAt-∘-local i xs fg=h)
 
 -- 2b. composition:  updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
 
-updateAt-∘ : ∀ (i : Fin n) {f g : A → A} →
-                   updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
-updateAt-∘ i xs = updateAt-∘-local i xs refl
+-- updateAt-∘ : ∀ (i : Fin n) {f g : A → A} →
+--                    updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
+-- updateAt-∘ i xs = updateAt-∘-local i xs refl
 
 -- 3. congruence:  updateAt i  is a congruence wrt. extensional equality.
 
@@ -290,28 +290,28 @@ updateAt-∘ i xs = updateAt-∘-local i xs refl
 
 updateAt-cong-local : ∀ (i : Fin n) {f g : A → A} (xs : Vec A n) →
                       f (lookup xs i) ≡ g (lookup xs i) →
-                      updateAt i f xs ≡ updateAt i g xs
+                      updateAt xs i f ≡ updateAt xs i g
 updateAt-cong-local zero    (x ∷ xs) f=g = cong (_∷ xs) f=g
 updateAt-cong-local (suc i) (x ∷ xs) f=g = cong (x ∷_) (updateAt-cong-local i xs f=g)
 
 -- 3b. congruence:  f ≗ g → updateAt i f ≗ updateAt i g
 
-updateAt-cong : ∀ (i : Fin n) {f g : A → A} →
-                f ≗ g → updateAt i f ≗ updateAt i g
-updateAt-cong i f≗g xs = updateAt-cong-local i xs (f≗g (lookup xs i))
+-- updateAt-cong : ∀ (i : Fin n) {f g : A → A} →
+--                 f ≗ g → updateAt i f ≗ updateAt i g
+-- updateAt-cong i f≗g xs = updateAt-cong-local i xs (f≗g (lookup xs i))
 
 -- The order of updates at different indices i ≢ j does not matter.
 
 -- This a consequence of updateAt-updates and updateAt-minimal
 -- but easier to prove inductively.
 
-updateAt-commutes : ∀ (i j : Fin n) {f g : A → A} → i ≢ j →
-                    updateAt i f ∘ updateAt j g ≗ updateAt j g ∘ updateAt i f
-updateAt-commutes zero    zero    0≢0 (x ∷ xs) = contradiction refl 0≢0
-updateAt-commutes zero    (suc j) i≢j (x ∷ xs) = refl
-updateAt-commutes (suc i) zero    i≢j (x ∷ xs) = refl
-updateAt-commutes (suc i) (suc j) i≢j (x ∷ xs) =
-  cong (x ∷_) (updateAt-commutes i j (i≢j ∘ cong suc) xs)
+-- updateAt-commutes : ∀ (i j : Fin n) {f g : A → A} → i ≢ j →
+--                     updateAt i f ∘ updateAt j g ≗ updateAt j g ∘ updateAt i f
+-- updateAt-commutes zero    zero    0≢0 (x ∷ xs) = contradiction refl 0≢0
+-- updateAt-commutes zero    (suc j) i≢j (x ∷ xs) = refl
+-- updateAt-commutes (suc i) zero    i≢j (x ∷ xs) = refl
+-- updateAt-commutes (suc i) (suc j) i≢j (x ∷ xs) =
+--   cong (x ∷_) (updateAt-commutes i j (i≢j ∘ cong suc) xs)
 
 -- lookup after updateAt reduces.
 
@@ -319,14 +319,14 @@ updateAt-commutes (suc i) (suc j) i≢j (x ∷ xs) =
 -- using []=↔lookup.
 
 lookup∘updateAt : ∀ (i : Fin n) {f : A → A} xs →
-                  lookup (updateAt i f xs) i ≡ f (lookup xs i)
+                  lookup (updateAt xs i f) i ≡ f (lookup xs i)
 lookup∘updateAt i xs =
   []=⇒lookup (updateAt-updates i xs (lookup⇒[]= i _ refl))
 
 -- For different indices it easily follows from updateAt-minimal.
 
 lookup∘updateAt′ : ∀ (i j : Fin n) {f : A → A} → i ≢ j → ∀ xs →
-                   lookup (updateAt j f xs) i ≡ lookup xs i
+                   lookup (updateAt xs j f) i ≡ lookup xs i
 lookup∘updateAt′ i j xs i≢j =
   []=⇒lookup (updateAt-minimal i j i≢j xs (lookup⇒[]= i _ refl))
 
@@ -335,11 +335,11 @@ lookup∘updateAt′ i j xs i≢j =
 []%=-id : ∀ (xs : Vec A n) (i : Fin n) → xs [ i ]%= id ≡ xs
 []%=-id xs i = updateAt-id i xs
 
-[]%=-∘ : ∀ (xs : Vec A n) (i : Fin n) {f g : A → A} →
-     xs [ i ]%= f
-        [ i ]%= g
-   ≡ xs [ i ]%= g ∘ f
-[]%=-∘ xs i = updateAt-∘ i xs
+-- []%=-∘ : ∀ (xs : Vec A n) (i : Fin n) {f g : A → A} →
+--      xs [ i ]%= f
+--         [ i ]%= g
+--    ≡ xs [ i ]%= g ∘ f
+-- []%=-∘ xs i = updateAt-∘ i xs
 
 
 ------------------------------------------------------------------------
@@ -348,13 +348,13 @@ lookup∘updateAt′ i j xs i≢j =
 -- _[_]≔_ is defined in terms of updateAt, and all of its properties
 -- are special cases of the ones for updateAt.
 
-[]≔-idempotent : ∀ (xs : Vec A n) (i : Fin n) →
-                 (xs [ i ]≔ x) [ i ]≔ y ≡ xs [ i ]≔ y
-[]≔-idempotent xs i = updateAt-∘ i xs
+-- []≔-idempotent : ∀ (xs : Vec A n) (i : Fin n) →
+--                  (xs [ i ]≔ x) [ i ]≔ y ≡ xs [ i ]≔ y
+-- []≔-idempotent xs i = updateAt-∘ i xs
 
-[]≔-commutes : ∀ (xs : Vec A n) (i j : Fin n) → i ≢ j →
-               (xs [ i ]≔ x) [ j ]≔ y ≡ (xs [ j ]≔ y) [ i ]≔ x
-[]≔-commutes xs i j i≢j = updateAt-commutes j i (i≢j ∘ sym) xs
+-- []≔-commutes : ∀ (xs : Vec A n) (i j : Fin n) → i ≢ j →
+--                (xs [ i ]≔ x) [ j ]≔ y ≡ (xs [ j ]≔ y) [ i ]≔ x
+-- []≔-commutes xs i j i≢j = updateAt-commutes j i (i≢j ∘ sym) xs
 
 []≔-updates : ∀ (xs : Vec A n) (i : Fin n) → (xs [ i ]≔ x) [ i ]= x
 []≔-updates xs i = updateAt-updates i xs (lookup⇒[]= i xs refl)
@@ -440,7 +440,7 @@ lookup-map (suc i) f (x ∷ xs) = lookup-map i f xs
 map-updateAt : ∀ {f : A → B} {g : A → A} {h : B → B}
                (xs : Vec A n) (i : Fin n) →
                f (g (lookup xs i)) ≡ h (f (lookup xs i)) →
-               map f (updateAt i g xs) ≡ updateAt i h (map f xs)
+               map f (updateAt xs i g) ≡ updateAt (map f xs) i h
 map-updateAt (x ∷ xs) zero    eq = cong (_∷ _) eq
 map-updateAt (x ∷ xs) (suc i) eq = cong (_ ∷_) (map-updateAt xs i eq)
 
@@ -1272,11 +1272,11 @@ updateAt-compose-relative = updateAt-∘-local
 Please use updateAt-∘-local instead."
 #-}
 
-updateAt-compose = updateAt-∘
-{-# WARNING_ON_USAGE updateAt-compose
-"Warning: updateAt-compose was deprecated in v2.0.
-Please use updateAt-∘ instead."
-#-}
+-- updateAt-compose = updateAt-∘
+-- {-# WARNING_ON_USAGE updateAt-compose
+-- "Warning: updateAt-compose was deprecated in v2.0.
+-- Please use updateAt-∘ instead."
+-- #-}
 
 updateAt-cong-relative = updateAt-cong-local
 {-# WARNING_ON_USAGE updateAt-cong-relative
@@ -1284,11 +1284,11 @@ updateAt-cong-relative = updateAt-cong-local
 Please use updateAt-cong-local instead."
 #-}
 
-[]%=-compose = []%=-∘
-{-# WARNING_ON_USAGE []%=-compose
-"Warning: []%=-compose was deprecated in v2.0.
-Please use []%=-∘ instead."
-#-}
+-- []%=-compose = []%=-∘
+-- {-# WARNING_ON_USAGE []%=-compose
+-- "Warning: []%=-compose was deprecated in v2.0.
+-- Please use []%=-∘ instead."
+-- #-}
 
 []≔-++-inject+ : ∀ {m n x} (xs : Vec A m) (ys : Vec A n) i →
                  (xs ++ ys) [ i ↑ˡ n ]≔ x ≡ (xs [ i ]≔ x) ++ ys
```

Should I push this or is there a way to do this without changing the API of the proofs?

MatthewDaggitt:
> Should I push this or is there a way to do this without changing the API of the proofs?

Yup, push it. There's no way to avoid changing the API unfortunately.

Saransh-cpp (commented):
I could not change the arguments of the proofs below.

For example, updating `updateAt-∘` -
```diff
diff --git a/src/Data/Vec/Properties.agda b/src/Data/Vec/Properties.agda
index f0fdc3a24..427e58407 100644
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -232,9 +232,9 @@ updateAt-∘-local (suc i) (x ∷ xs) fg=h = cong (x ∷_) (updateAt-∘-local i
 
 -- 2b. composition:  updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
 
--- updateAt-∘ : ∀ (i : Fin n) {f g : A → A} →
---                    updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
--- updateAt-∘ i xs = updateAt-∘-local i xs refl
+updateAt-∘ : ∀ (i : Fin n) {f g : A → A} (xs : Vec A n) →
+                   updateAt xs i f ∘ updateAt xs i g ≗ updateAt xs i (f ∘ g)
+updateAt-∘ i xs = updateAt-∘-local i xs refl
```

gives me the following error -
```
saransh@192 agda-stdlib % agda -i. src/Data/Vec/Properties.agda
Checking Data.Vec.Properties (/Users/saransh/Code/Mitacs/agda-stdlib/src/Data/Vec/Properties.agda).
/Users/saransh/Code/Mitacs/agda-stdlib/src/Data/Vec/Properties.agda:236,20-35
Vec _A_1227 _n_1228 !=< ((y : _B_1224 x) → _C_1225 y)
  Vec _A_1227 _n_1228
matches the expected type
  (y : _B_1224 x) → _C_1225 y
```

The other proofs throw a similar error as well.

Could you please let me know what I am missing here? Thanks!

src/Data/Vec/Base.agda
```diff
@@ -64,29 +64,27 @@ iterate : (A → A) → A → ∀ {n} → Vec A n
...
+insertAt (x ∷ xs) (suc i)  v = x ∷ insertAt xs i v
 
-remove : Vec A (suc n) → Fin (suc n) → Vec A n
-remove (_ ∷ xs)     zero     = xs
-remove (x ∷ y ∷ xs) (suc i)  = x ∷ remove (y ∷ xs) i
+removeAt : Vec A (suc n) → Fin (suc n) → Vec A n
+removeAt (_ ∷ xs)     zero     = xs
+removeAt (x ∷ y ∷ xs) (suc i)  = x ∷ removeAt (y ∷ xs) i
```
jamesmckinna, line 73:
```agda
removeAt (x ∷ xs@(_ ∷ _)) (suc i) = x ∷ removeAt xs i
```

src/Data/Vec/Properties.agda
```diff
@@ -212,29 +212,29 @@ updateAt-minimal (suc i) (suc j) (x ∷ xs) i≢j (there loc) =
...
 updateAt-∘-local (suc i) (x ∷ xs) fg=h = cong (x ∷_) (updateAt-∘-local i xs fg=h)
 
 -- 2b. composition:  updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
 
-updateAt-∘ : ∀ (i : Fin n) {f g : A → A} →
-                   updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
-updateAt-∘ i xs = updateAt-∘-local i xs refl
+-- updateAt-∘ : ∀ (i : Fin n) {f g : A → A} →
+--                    updateAt i f ∘ updateAt i g ≗ updateAt i (f ∘ g)
+-- updateAt-∘ i xs = updateAt-∘-local i xs refl
```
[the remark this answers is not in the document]
jamesmckinna: Try adding an explicit `{h = f ∘ g}` binding to the call to `updateAt-∘-local`?

src/Data/Vec/Properties.agda
```diff
@@ -243,43 +243,43 @@ updateAt-∘ i xs = updateAt-∘-local i xs refl
...
 updateAt-cong-local (suc i) (x ∷ xs) f=g = cong (x ∷_) (updateAt-cong-local i xs f=g)
 
 -- 3b. congruence:  f ≗ g → updateAt i f ≗ updateAt i g
 
-updateAt-cong : ∀ (i : Fin n) {f g : A → A} →
-                f ≗ g → updateAt i f ≗ updateAt i g
-updateAt-cong i f≗g xs = updateAt-cong-local i xs (f≗g (lookup xs i))
+-- updateAt-cong : ∀ (i : Fin n) {f g : A → A} →
+--                 f ≗ g → updateAt i f ≗ updateAt i g
+-- updateAt-cong i f≗g xs = updateAt-cong-local i xs (f≗g (lookup xs i))
```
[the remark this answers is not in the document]
jamesmckinna: This is a parametrisation thing? Previously, the `xs` argument to `updateAt i f` came at the end, so it was easy to appeal to eta in the proof. But here, it's now the first (?) argument to `updateAt`, so `updateAt i f ≗ updateAt i g` isn't well=formed as a goal...

src/Data/Vec/Properties.agda
```diff
@@ -243,43 +243,43 @@ updateAt-∘ i xs = updateAt-∘-local i xs refl
...
-updateAt-commutes (suc i) zero    i≢j (x ∷ xs) = refl
-updateAt-commutes (suc i) (suc j) i≢j (x ∷ xs) =
-  cong (x ∷_) (updateAt-commutes i j (i≢j ∘ cong suc) xs)
+-- updateAt-commutes : ∀ (i j : Fin n) {f g : A → A} → i ≢ j →
+--                     updateAt i f ∘ updateAt j g ≗ updateAt j g ∘ updateAt i f
+-- updateAt-commutes zero    zero    0≢0 (x ∷ xs) = contradiction refl 0≢0
+-- updateAt-commutes zero    (suc j) i≢j (x ∷ xs) = refl
+-- updateAt-commutes (suc i) zero    i≢j (x ∷ xs) = refl
+-- updateAt-commutes (suc i) (suc j) i≢j (x ∷ xs) =
+--   cong (x ∷_) (updateAt-commutes i j (i≢j ∘ cong suc) xs)
```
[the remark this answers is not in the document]
jamesmckinna: Hmm. This seems to be the same as in 2. above: check the parametrisation, and then maybe re-consider with @MatthewDaggitt which has the correct order, `List`, or `Vec` (which as we have seen, is more fiddly!)

src/Data/Vec/Properties.agda
```diff
@@ -288,11 +288,11 @@ lookup∘updateAt′ i j xs i≢j =
...
-[]%=-∘ : ∀ (xs : Vec A n) (i : Fin n) {f g : A → A} →
-     xs [ i ]%= f
-        [ i ]%= g
-   ≡ xs [ i ]%= g ∘ f
-[]%=-∘ xs i = updateAt-∘ i xs
+-- []%=-∘ : ∀ (xs : Vec A n) (i : Fin n) {f g : A → A} →
+--      xs [ i ]%= f
+--         [ i ]%= g
+--    ≡ xs [ i ]%= g ∘ f
+-- []%=-∘ xs i = updateAt-∘ i xs
```
jamesmckinna, line 295: Ditto.

src/Data/Vec/Properties.agda
```diff
@@ -301,13 +301,13 @@ lookup∘updateAt′ i j xs i≢j =
 -- _[_]≔_ is defined in terms of updateAt, and all of its properties
 -- are special cases of the ones for updateAt.
 
-[]≔-idempotent : ∀ (xs : Vec A n) (i : Fin n) →
-                 (xs [ i ]≔ x) [ i ]≔ y ≡ xs [ i ]≔ y
-[]≔-idempotent xs i = updateAt-∘ i xs
+-- []≔-idempotent : ∀ (xs : Vec A n) (i : Fin n) →
+--                  (xs [ i ]≔ x) [ i ]≔ y ≡ xs [ i ]≔ y
+-- []≔-idempotent xs i = updateAt-∘ i xs
```
jamesmckinna, line 306: Etc.

src/Data/Vec/Properties.agda
```diff
@@ -301,13 +301,13 @@ lookup∘updateAt′ i j xs i≢j =
...
 
-[]≔-commutes : ∀ (xs : Vec A n) (i j : Fin n) → i ≢ j →
-               (xs [ i ]≔ x) [ j ]≔ y ≡ (xs [ j ]≔ y) [ i ]≔ x
-[]≔-commutes xs i j i≢j = updateAt-commutes j i (i≢j ∘ sym) xs
+-- []≔-commutes : ∀ (xs : Vec A n) (i j : Fin n) → i ≢ j →
+--                (xs [ i ]≔ x) [ j ]≔ y ≡ (xs [ j ]≔ y) [ i ]≔ x
+-- []≔-commutes xs i j i≢j = updateAt-commutes j i (i≢j ∘ sym) xs
```
jamesmckinna, line 310: Etc.

src/Data/Vec/Base.agda
Taneb: Do we even need to match against `(_ ∷ _)` here? Would `removeAt (x ∷ xs) (suc i) = x ∷ removeAt xs i` work?

jamesmckinna: All the better if so!

jamesmckinna:
@MatthewDaggitt I think that the order of arguments for the `updateAt-global` lemmas, in terms of the `updateAt-local` make sense for a particular argument order, in the case of `Vec`. Don't know how/why the case for `List` is somehow more flexible, but I would take the order as in `Vec` over that in `List` (something has to break...) to avoid morepain in the `*-global` proofs... but YMMV?

src/Data/Vec/Base.agda
gallais: You probably need to match for Agda to see that the recursive call is legal (it needs
to be performed on a `Fin (suc _)`).

Saransh-cpp: Thanks!

MatthewDaggitt:
> I think that the order of arguments for the updateAt-global lemmas, in terms of the updateAt-local make sense for a particular argument order, in the case of Vec. Don't know how/why the case for List is somehow more flexible, but I would take the order as in Vec over that in List (something has to break...) to avoid morepain in the *-global proofs... but YMMV?

The problem is that you actually can't take the current `Vec` order and port it over to the `List` setting, the size of the index has to be known in advance so the container argument has to come before the index. I know it's unfortunate that we can't use pointwise syntax anymore, but I still think the `List` order is morally the right one.

MatthewDaggitt:
I'm pushing this through now.

MatthewDaggitt:
Unless anyone has any objections over the next couple of days I'm going to merge this.

jamesmckinna:
ill try to give it one last look today.

jamesmckinna:
Seem to have picked up some extraneous commits from some recently merged PRs of mine...

Saransh-cpp:
Thanks for working on this! 🙂

PR #2051, opened 2023-08-09
Sync `iterate` and `replicate` for `List` and `Vec`

Saransh-cpp:
Corresponding `Vec.Base.iterate` -

https://github.com/agda/agda-stdlib/blob/bfa4f8bc41da10bffaf69f5786301e587ae69c83/src/Data/Vec/Base.agda#L63-L65

There is also a similar function in Idris2 - https://github.com/idris-lang/Idris2/blob/badf1e98c8812e8b54258754cf47782f61d8fb15/libs/base/Data/List.idr#L99-L102 (not exaclty the same, but similar)


src/Data/List/Base.agda
```diff
@@ -187,6 +187,10 @@ replicate : ℕ → A → List A
 replicate zero    x = []
 replicate (suc n) x = x ∷ replicate n x
 
+iterate : (A → A) → A → ℕ → List A
+iterate f k zero    = []
```
jamesmckinna, line 191: I wouldn't give the argument of type `A` the name `k`... I think that stdlib convention has that as either `x` (pros and cons either way), or else `a` (but often reserved as the level of `A`... so perhaps, sadly, best avoided). 

To me, `k` is always a `Nat`, maybe a `Fin`, or even an `Integer`...

jamesmckinna:
Question for the stdlib developers: why is the `n` index in the `Vec`-version marked as *implicit*? 'Surely' it should be an explicit argument? (Or are the only call-sites for `iterate` ones in which the index is already inferrable from the ambient context?

src/Data/List/Base.agda
jamesmckinna: Or, for that matter, take the `s` and `z` names from `Vec.iterate`? These are, after all, operations derived from the canonical NNO structure on`Nat`...

Saransh-cpp: Thanks for the review! I've updated the argument names to match `Vec.Base.iterate`.

Saransh-cpp:
Also, looking at the discussion about adding proofs in #2049. Should this PR also have some proof to ensure everything is working fine? Maybe something like (if I wrote it right) -
```agda
iterate-is-replicate-if-id : ∀ (n : ℕ) (x : A) → iterate id x n ≡ replicate n x
iterate-is-replicate-if-id  ℕ.zero x = refl
iterate-is-replicate-if-id  (suc n) x = cong (_ ∷_) (iterate-is-replicate-if-id n (id x))
```

JacquesCarette:
Yes, adding some properties at the same time as the function is a very good idea. It should have "the same" properties as those for `Vec.iterate` as well as commuting with `toList`.

Taneb:
> Also, looking at the discussion about adding proofs in #2049. Should this PR also have some proof to ensure everything is working fine? Maybe something like (if I wrote it right) -
> 
> ```agda
> iterate-is-replicate-if-id : ∀ (n : ℕ) (x : A) → iterate id x n ≡ replicate n x
> iterate-is-replicate-if-id  ℕ.zero x = refl
> iterate-is-replicate-if-id  (suc n) x = cong (_ ∷_) (iterate-is-replicate-if-id n (id x))
> ```

I feel like this would be better the other way around, like `replicate-is-iterate` (then you don't need to mention `id` in the name)

jamesmckinna:
@Taneb @Saransh-cpp : why not `iterate-id-is-replicate`, as it stands?

Saransh-cpp:
This could also be `iterate-id` (similar to `map-id`), considering the - use only LHS for naming -convention.

Saransh-cpp:
(will rebase once #2056 is merged)

src/Data/List/Properties.agda
```diff
@@ -624,6 +626,47 @@ length-replicate : ∀ n {x : A} → length (replicate n x) ≡ n
 length-replicate zero    = refl
 length-replicate (suc n) = cong suc (length-replicate n)
 
+lookup-replicate : ∀ n (x : A) (i : Fin (length (replicate n x))) → lookup (replicate n x) i ≡ x
```
jamesmckinna, line 629: Consider `let` binding the duplicated expression `xs = replicate n x` for the sake of legibility:
```agda
lookup-replicate : ∀ n (x : A) → let xs = replicate n x in (i : Fin (length xs)) → lookup xs i ≡ x
```
(here, and in `CHANGELOG`)

src/Data/List/Properties.agda
```diff
@@ -624,6 +626,47 @@ length-replicate : ∀ n {x : A} → length (replicate n x) ≡ n
...
+
+toList-iterate : ∀ n {f} {x : A} → toList (Vec.iterate f x {n = n}) ≡ iterate f x n
+toList-iterate zero    = refl
+toList-iterate (suc n) = cong (_ ∷_) (toList-iterate n)
+
+iterate-id : ∀ n {x : A} → iterate id x n ≡ replicate n x
+iterate-id zero    = refl
+iterate-id (suc n) = cong (_ ∷_) (iterate-id n)
+
+take-iterate : ∀ n {f} {x : A} → take n (iterate f x n) ≡ iterate f x n
```
jamesmckinna, line 658: Is this not a (fairly immediate?) consequence of `length-iterate` and `take-all`?

Even if it is worth retaining as a lemma in its own right, please consider bnding `let xs = iterate f x n in ...` in the type, as in L629 above

src/Data/List/Properties.agda
```diff
@@ -624,6 +626,47 @@ length-replicate : ∀ n {x : A} → length (replicate n x) ≡ n
...
+take-iterate zero    = refl
+take-iterate (suc n) = cong (_ ∷_) (take-iterate n)
+
+drop-iterate : ∀ n {f} {x : A} → drop n (iterate f x n) ≡ []
```
jamesmckinna, line 662: I'm surprised to discover that this is *not* similarly a consequence of `length-iterate` and a lemma of the form
```agda
drop-all : (n : ℕ) (xs : List A) → n ≥ length xs → drop n xs ≡ []
drop-all n       []       _ = drop-[] n
drop-all (suc n) (x ∷ xs) p = drop-all n xs (≤-pred p)
```
because such a lemma doesn't exist! Please consider adding it as part of this PR?

src/Data/List/Properties.agda
```diff
@@ -624,6 +626,47 @@ length-replicate : ∀ n {x : A} → length (replicate n x) ≡ n
...
+drop-iterate zero    = refl
+drop-iterate (suc n) = drop-iterate n
+
+lookup-iterate : ∀ f n (x : A) (i : Fin (length (iterate f x n))) → lookup (iterate f x n) i ≡ ℕ.iterate f x (toℕ i)
```
jamesmckinna, line 666: Argument `n` may be bound implicit here: it's determined uniquely by the type of `i`

CHANGELOG.md
```diff
@@ -2642,6 +2652,12 @@ Other minor changes
   lookup-cast   : lookup (cast eq xs) (Fin.cast eq i) ≡ lookup xs i
   lookup-cast₁  : lookup (cast eq xs) i ≡ lookup xs (Fin.cast (sym eq) i)
   lookup-cast₂  : lookup xs (Fin.cast eq i) ≡ lookup (cast (sym eq) xs) i
+
+  length-iterate : length (iterate f x {n}) ≡ n
+  iterate-id     : iterate id x {n} ≡ replicate x
+  take-iterate   : take n (iterate f x {n + m}) ≡ iterate f x
```
jamesmckinna, line 2658: For the sake of documentation, consider the redundant recoding in the type
```agda
take-iterate : take n (iterate f x {n + m}) ≡ iterate f x {n}
```

src/Data/List/Properties.agda
jamesmckinna: Better, however: you know that `length (replicate n x)` is equal to `n`, so no need to consider `i : Fin (length (replicate n x))` when `i : Fin n` will suffice?

src/Data/Vec/Properties.agda
```diff
@@ -1002,6 +999,29 @@ zipWith-replicate₂ _⊕_ []       y = refl
 zipWith-replicate₂ _⊕_ (x ∷ xs) y =
   cong (x ⊕ y ∷_) (zipWith-replicate₂ _⊕_ xs y)
 
+------------------------------------------------------------------------
+-- iterate
+
+length-iterate : ∀ n f (x : A) → length (iterate f x {n}) ≡ n
```
jamesmckinna, line 1005: No case analysis required; actually this lemma is not required... its proof is `refl` *definitionally* by virtue of the definition of `length` for vectors. Consider omitting.

src/Data/Vec/Properties.agda
```diff
@@ -1002,6 +999,29 @@ zipWith-replicate₂ _⊕_ []       y = refl
...
+length-iterate zero    f x = refl
+length-iterate (suc n) f x = cong suc refl
+
+iterate-id : ∀ n (x : A) → iterate id x {n} ≡ replicate x
+iterate-id zero x    = refl
+iterate-id (suc n) x = cong (_ ∷_) (iterate-id n (id x))
+
+take-iterate : ∀ n f (x : A) → take n (iterate f x {n + m}) ≡ iterate f x
```
jamesmckinna, line 1013: As in `CHANGELOG` above, consider adding a redundant `{n = n}` to the RHS...

jamesmckinna: Is there an 'easy' analogue of `List.Properties.take-all` which you can derive this from?

src/Data/Vec/Properties.agda
```diff
@@ -1002,6 +999,29 @@ zipWith-replicate₂ _⊕_ []       y = refl
...
+take-iterate zero    f x = refl
+take-iterate (suc n) f x = cong (_ ∷_) (take-iterate n f (f x))
+
+drop-iterate : ∀ n f (x : A) → drop n (iterate f x) ≡ []
```
jamesmckinna, line 1017: Is there an 'easy' analogue of `List.Properties.drop-all` which you can derive this from?

src/Data/List/Properties.agda
jamesmckinna: And better, again, if you have shown that `length (iterate f x n)` is equal to `n`, then simplify the typing of `i` to `i : Fin n`...

src/Data/List/Properties.agda
JacquesCarette: The lemma is probably worth retaining, but I agree that it would be nice to have the proof use those results instead of being done "by hand", even though the resulting proof is quite simple.

src/Data/List/Properties.agda
Saransh-cpp: Replacing `i : Fin (length (replicate n x))` with `i: Fin n` gives me this -
```
n != (foldr (const suc) 0 (replicate n x)) of type ℕ
when checking that the expression i has type
Fin (length (replicate n x))
```

JacquesCarette: I think you'd need
```agda
lookup-replicate : ∀ n (x : A) (i : Fin n) → lookup (replicate n x) (cast (sym (length-replicate n x)) i) ≡ x
```
which does not really feel like a large improvement?

[The problem really stems from the type of `lookup` which has a computation in it that doesn't reduce, and infects everything else.]

Saransh-cpp: Thanks! I'll update this and `lookup-iterate`.

src/Data/List/Properties.agda
Saransh-cpp: Should it be written like this instead? -
```agda
drop-iterate : ∀ n f (x : A) → drop n (iterate f x n) ≡ []
drop-iterate n f x = drop-all n (iterate f x n) {!(length-iterate n f x)!}
```

Saransh-cpp:
Thanks for the reviews, @jamesmckinna and @JacquesCarette!

Also, I found 2 similar proofs named very differently while going through the files, should both of them have the same name? -
```agda
take-drop-id : ∀ (m : ℕ) (x : Vec A (m + n)) → take m x ++ drop m x ≡ x
take++drop   : ∀ n (xs : List A) → take n xs ++ drop n xs ≡ xs
```

jamesmckinna:
My feeling is that they should have the same name. What I'm not clear about us which one or even a new one, hybridising the two, such as: 
`take++drop=id`...

I'll think about it, but also throw it out to @MatthewDaggitt @taneb @gallais for their opinion(s)...

src/Data/Vec/Base.agda
```diff
@@ -263,16 +263,13 @@ splitAt : ∀ m {n} (xs : Vec A (m + n)) →
           ∃₂ λ (ys : Vec A m) (zs : Vec A n) → xs ≡ ys ++ zs
 splitAt zero    xs                = ([] , xs , refl)
 splitAt (suc m) (x ∷ xs)          with splitAt m xs
-splitAt (suc m) (x ∷ .(ys ++ zs)) | (ys , zs , refl) =
-  ((x ∷ ys) , zs , refl)
+splitAt (suc m) (x ∷ xs) | (ys , zs , p) = ((x ∷ ys) , zs , cong (x ∷_) p)
```
MatthewDaggitt, line 266: Don't need the outermost brackets on the LHS?

MatthewDaggitt (commented):
> Question for the stdlib developers: why is the n index in the Vec-version marked as implicit? 'Surely' it should be an explicit argument? Or are the only call-sites for iterate ones in which the index is already inferrable from the ambient context?

I agree. The argument to `Vec.iterate` should be explicit, not implicit. @Saransh-cpp any chance you could change it as part of this PR and add a suitable CHANGELOG entry?

> Also, I found 2 similar proofs named very differently while going through the files, should both of them have the same name? -

Yup, agreed. I dislike `take-drop-id` which is very misleading. I like @jamesmckinna's suggestion of `take++drop≡id`

src/Data/Vec/Base.agda
Saransh-cpp: This is coming from #2056. I will rebase once that is merged. (Some of the proofs in this PR depend on #2056.)

Saransh-cpp:
> I agree. The argument to Vec.iterate should be explicit, not implicit

Should that be the case with `Vec.replicate` too?

MatthewDaggitt:
> Should that be the case with Vec.replicate too?

Yes, nice spot!

Saransh-cpp:
Re: the naming of `iterate-id` (`iterate-is-replicate`, `iterate-id-replicate`, ...) - I found -

https://github.com/agda/agda-stdlib/blob/104125c62d88ebe91274ebde521b428bad71f26e/src/Data/Fin/Properties.agda#L263-L268

which looks like a similar situation.

CHANGELOG.md
```diff
@@ -2693,6 +2711,13 @@ Other minor changes
   lookup-cast₁  : lookup (cast eq xs) i ≡ lookup xs (Fin.cast (sym eq) i)
   lookup-cast₂  : lookup xs (Fin.cast eq i) ≡ lookup (cast (sym eq) xs) i
 
+  length-iterate : length (iterate f x {n}) ≡ n
+  iterate-id     : iterate id x {n} ≡ replicate x
+  take-iterate   : take n (iterate f x {n + m}) ≡ iterate f x {n}
+  drop-iterate   : drop n (iterate f x) ≡ []
+  lookup-iterate : lookup (iterate f x) i ≡ ℕ.iterate f x (toℕ i)
+  toList-iterate : toList (iterate f x {n}) ≡ List.iterate f x n
+
```
gallais, line 2720: Given the breaking change, I assume these `n` should now be explicit?

```suggestion

  length-iterate : length (iterate f x n) ≡ n
  iterate-id     : iterate id x n ≡ replicate x
  take-iterate   : take n (iterate f x (n + m)) ≡ iterate f x n
  drop-iterate   : drop n (iterate f x n) ≡ []
  lookup-iterate : lookup (iterate f x n) i ≡ ℕ.iterate f x (toℕ i)
  toList-iterate : toList (iterate f x n) ≡ List.iterate f x n

```

src/Data/Vec/Base.agda
jamesmckinna: Nor on the RHS... ;-)

CHANGELOG.md
Saransh-cpp: Yes, thanks!

src/Data/List/Properties.agda
jamesmckinna:
```agda
take-iterate : ∀ n {f} {x : A} → let xs = iterate f x n in take n xs ≡ xs
take-iterate n {f} {x} = take-all n xs n≥length[xs]
  where
  xs = iterate f x n
  n≥length[xs] : n ≥ length xs
  n≥length[xs] rewrite length-iterate n {f} {x} = ≤-refl
```
NB You are going to have a hard time inferring `f` in these proofs (Agda can't unify functions easily), so, as here, the arguments will (probably: CHECK!) need to be passed explicitly, so nothing to be gained by keeping them implicit here?

jamesmckinna (changes requested):
```agda
take-iterate : ∀ n {f} {x : A} → let xs = iterate f x n in take n xs ≡ xs
take-iterate n {f} {x} = take-all n xs n≥length[xs]
  where
  xs = iterate f x n
  n≥length[xs] : n ≥ length xs
  n≥length[xs] rewrite length-iterate n {f} {x} = ≤-refl
```
NB You are going to have a hard time inferring `f` in these proofs (Agda can't unify functions easily), so, as here, the arguments will (probably: CHECK!) need to be passed explicitly, so nothing to be gained by keeping them implicit here?

Similarly, it's annoying to have to repeat the `let` binding in the type and then the proof, so better to use an anonymous `module`:
```agda
module _ n f x where
  private
    xs = iterate f x n
    n≥length[xs] : n ≥ length xs
    n≥length[xs] rewrite length-iterate n {f} {x} = ≤-refl
    
  take-iterate : take n xs ≡ xs
  take-iterate = take-all n xs n≥length[xs]
```
(and then the other proofs can also be in the scope of that module...?)

src/Data/List/Properties.agda
jamesmckinna: See my review comment, requesting changes, below.

src/Data/List/Properties.agda
```diff
@@ -872,6 +828,55 @@ drop-drop zero n xs = refl
...
+length-iterate : ∀ n {f} {x : A} → length (iterate f x n) ≡ n
+length-iterate zero    = refl
+length-iterate (suc n) = cong suc (length-iterate n)
+
+iterate-id : ∀ n {x : A} → iterate id x n ≡ replicate n x
+iterate-id zero    = refl
+iterate-id (suc n) = cong (_ ∷_) (iterate-id n)
+
+module _ n f x where
+  open import Data.List.Base
```
jamesmckinna, line 864: Apologies if I am the source of this `import` here! But I don't think it's necessary (there's already an import of `Base` at the top of the `Properties` module)... and may be causing the failing test?

src/Data/List/Properties.agda
```diff
@@ -828,12 +818,65 @@ drop-take-suc-tabulate : ∀ {n} (f : Fin n → A) (i : Fin n) → let m = toℕ
...
+    n≥length[xs] : n ≥ length xs
+    n≥length[xs] rewrite length-iterate n {f} {x} = ≤-refl
+    
+  take-iterate : take n xs ≡ xs
+  take-iterate = take-all n xs n≥length[xs]
+
+  drop-iterate : drop n xs ≡ []
+  drop-iterate = drop-all n xs n≥length[xs]
+
+lookup-iterate : ∀ n f (x : A) (i : Fin n) → lookup (iterate f x n) (cast (sym (length-iterate n)) i) ≡ ℕ.iterate f x (toℕ i)
```
jamesmckinna, line 875: I think it's a good heuristic to have the parameters to a function proceed from left-to-right from 'what stays fixed' through to 'what varies throughout' (I learnt this from @JacquesCarette ... ;-)), moreover in a way which correlates with the corresponding argument order in the function invocation `iterate f x n`. So apologies are in order for not reflecting that in the module parameter list above... where `f` and `x` should come before `n`.

Indeed, within the scope of a given `f` and `x`, the module can/should consider *all* the proofs about `iterate f x` for various arguments `n`, which may vary internally (or else bind/be bound by the types of arguments like `i`), or which may be fixed as in `take-iterate` etc. 

So a bit of gentle refactoring is in order here, I think, to get the maximum value out of working in a module scope where `f` and `x` are 'fixed'...

src/Data/Vec/Base.agda
```diff
@@ -9,10 +9,10 @@
 module Data.Vec.Base where
 
 open import Data.Bool.Base using (Bool; true; false; if_then_else_)
-open import Data.Nat.Base
+open import Data.Nat.Base as N
```
jamesmckinna, line 12: Do you need the qualified import? If you do, convention/style dictates it should be imported as either `Nat` or `ℕ`.

src/Data/Vec/Properties.agda
```diff
@@ -1063,6 +1059,29 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+
+take-iterate : ∀ n f (x : A) → take n (iterate f x (n + m)) ≡ iterate f x n
+take-iterate zero    f x = refl
+take-iterate (suc n) f x = cong (_ ∷_) (take-iterate n f (f x))
+
+drop-iterate : ∀ n f (x : A) → drop n (iterate f x (n + zero)) ≡ []
+drop-iterate zero    f x = refl
+drop-iterate (suc n) f x = drop-iterate n f (f x)
+
+lookup-iterate :  ∀ f (i : Fin n) (x : A) → lookup (iterate f x n) i ≡ ℕ.iterate f x (toℕ i)
```
jamesmckinna, line 1077: `x` before `i` in the parameter telescope...

src/Data/List/Properties.agda
Saransh-cpp: Thanks @jamesmckinna! I'll try to refactor all the iterate proofs into this module. I've rearranged the arguments in all the proofs.

MatthewDaggitt:
@Saransh-cpp this is no longer blocked on https://github.com/agda/agda-stdlib/pull/2056 :+1:

Saransh-cpp:
Thanks for the ping, @MatthewDaggitt!

src/Data/List/Properties.agda
```diff
@@ -834,6 +824,60 @@ drop-drop zero n xs = refl
...
+    xs = iterate f x n
+    n≥length[xs] : n ≥ length xs
+    n≥length[xs] rewrite length-iterate f x n = ≤-refl
+
+  take-iterate : take n xs ≡ xs
+  take-iterate = take-all n xs n≥length[xs]
+
+  drop-iterate : drop n xs ≡ []
+  drop-iterate = drop-all n xs n≥length[xs]
+
```
jamesmckinna, line 875: Looks fine now. A possible future refactoring would be to try to squeeze the maximum out of the module abstraction across all, or most of, the lemmas. But `iterate` changes its base `x` (and counter `n`!) on each iteration, so perhaps that harder than it first seems. Ah well!

jamesmckinna (approved):

src/Data/List/Properties.agda
Saransh-cpp: Thanks, @jamesmckinna!

MatthewDaggitt:
In the interests of getting this merged in, I've removed `take-iterate` and `drop-iterate` as I think they're probably derivative enough space cases that they don't make the bar to get included. Will merge once tests pass.

As always, thanks for your work @Saransh-cpp !

PR #2053, opened 2023-08-11
More `Relation.Binary` simplifications

Saransh-cpp:
Just one more PR after this 🙂


jamesmckinna:
The `low-hanging-fruit` may seem harsh: this is a mammoth effort!

Taneb:
> The `low-hanging-fruit` may seem harsh: this is a mammoth effort!

Sometimes there's enough low hanging fruit to fill several trucks

Issue #1678, opened 2022-01-07
Surjections should be clarified as being split surjections


mietek:
Could we have some clarifying comments added to the stdlib about the definition of surjection actually being the definition of split surjection?  It would be helpful to people such as myself that continue to have trouble with the overloading of meaning in mathematical discourse.  Example:
http://web.archive.org/web/20211026191457/https://twitter.com/EscardoMartin/status/1453039001783439370

> (Andrej Bauer) Your definition of surjective is incorrect. Surjectivity is not definable in pure MLTT. Martín Escardó's [notes](https://www.cs.bham.ac.uk/~mhe/HoTT-UF-in-Agda-Lecture-Notes/HoTT-UF-Agda.html#images-and-surjections) elaborate on this in a nice way.

> (Martín Escardó) We can say a bit more: your definition gives the notion of *split* surjection, that is, a surjection f with a right inverse g (meaning that f after g is the identity function).

Such clarifying comments can be found in @nad’s work:
http://www.cse.chalmers.se/~nad/listings/equality/Surjection.html

As a side note, I love how the linked code clearly states the following definition.  I challenge you to locate this definition in the stdlib.
```agda
Split-surjective :
  ∀ {a b} {A : Type a} {B : Type b} → (A → B) → Type (a ⊔ b)
Split-surjective f = ∀ y → ∃ λ x → f x ≡ y
```

gallais:
> I challenge you to locate this definition in the stdlib.

I looked at `Function/`, opened `Function.Definitions` and found that it
was re-exporting `Surjective` from `Function.Definitions.Core2`.
Granted the definition is parametrised over the notion of equality for the
domain / codomain but it's fine to provide a generalisation IMO.

mietek:
> I looked at `Function/`, opened `Function.Definitions` and found that it was re-exporting `Surjective` from `Function.Definitions.Core2`.

Ah!  Unfortunately, I looked at `Function.Surjection` and found myself in a twisty maze of setoids.

laMudri:
> This raises an important question that I am not qualified to answer. Is the setoid-based definition of `Surjective` named accurately, or not?

I think not. In standard mathematical notation (i.e, ∃ means *mere* existence), we have:

- Surjective f = ∀y. ∃x. f x = y
- SplitSurjective f = ∃g. ∀y. f(g y) = y

[...]

PR #2054, opened 2023-08-13
Connect `LeftInverse` with (`Split`)`Surjection`

laMudri:
This PR partially addresses #1678. I have not decided what to do with `Function.Definitions` (potential confusion over its definition of `Surjective`) or `Epi` (because I'm not confident on which way to address the size issue is right).


src/Function/Structures.agda
```diff
@@ -105,6 +105,22 @@ record IsLeftInverse (to : A → B) (from : B → A) : Set (a ⊔ b ⊔ ℓ₁ 
   strictlyInverseˡ : StrictlyInverseˡ _≈₂_ to from
   strictlyInverseˡ x = inverseˡ Eq₁.refl
 
+  isSurjection : IsSurjection to
+  isSurjection = record
+    { isCongruent = isCongruent
+    ; surjective = λ y → from y , inverseˡ
+    }
```
laMudri, line 112: One could argue for this to be called `to-isSurjection`. I don't know which name convention suggests.

CHANGELOG.md
```diff
@@ -2658,7 +2658,7 @@ Other minor changes
                 ∀ {m n} → _Respectsʳ_ (_<_ {m} {n}) _≋_
   <-wellFounded : Symmetric _≈_ →  Transitive _≈_ → _≺_ Respectsʳ _≈_ → WellFounded _≺_ →
                   ∀ {n} → WellFounded (_<_ {n})
-```
+  ```
```
laMudri, line 2661: This is a drive-by fix, but important because the formatting of the entire rest of the file is broken without it.

src/Function/Bundles.agda
```diff
@@ -188,14 +188,54 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
...
+    isSplitSurjection = record { from = from ; isLeftInverse = isLeftInverse }
+
+    surjection : Surjection
+    surjection = record
+      { to = to
+      ; cong = to-cong
+      ; surjective = λ y → from y , inverseˡ
+      }
+
+  -- A left inverse is also known as a “split surjection”.
```
laMudri, line 209: I don't know whether it's right to call an inhabitant of `LeftInverse` “a left inverse”.

laMudri:
I forget how notifications work here, so maybe I should cc @mietek.

CHANGELOG.md
jamesmckinna: Is this fixed now?

MatthewDaggitt: If it's not showing up as a conflict then it's probably not...

jamesmckinna: Should it be lifted out as a separate 'PR' then (or directly edited on `master`)?

jamesmckinna: Fixed in #2154

src/Function/Bundles.agda
jamesmckinna: No, but such usage seems to have been present since the old hierarchy (in `Function.Inverse`)...

src/Function/Structures.agda
MatthewDaggitt: I think it's fine as it is.

PR #1869, opened 2022-10-26
Stage 2 of refactoring of `Relation.Nullary`

MatthewDaggitt:
Deprecates and cleans up the imports
  ```
  Relation.Nullary.Product
  Relation.Nullary.Sum
  Relation.Nullary.Implication
  ```

PR #2055, opened 2023-08-13
minor additions to `Decidable` and `Reflects`

jamesmckinna:
This a largely cosmetic PR, which does the following:
- [x] tidies up deprecation warning from #2026 (adds boilerplate to the module, moves the text of the deprecation in `CHANGELOG`) (UPDATED: this seems dealt with under #2110?)
- [x] adds two examples from `README.Design.Decidability` to `Relation.Nullary.Reflects` and `Relation.Nullary.Decidable.Core`
- [ ] adds a (trivial) lifting function for proofs of `Dec` from `Bool` to `A -> Bool`
- [x] ~~hundreds~~ a 'truckload' of trivial (and not-so-trivial) `import` simplifications/refinements arising from #1869 (UPDATED: now seemingly broken by recent commits to `master`.)

The first addition, in `Relation.Nullary.Reflects`, by analogy with other smart constructors in stdlib, is named
* `T⁺ : ∀ b → Reflects (T b) b`  (this is example `README.Design.Decidability.ex₂`)

The others, defined in `Relation.Nullary.Decidable.Core`, have perhaps less satisfactory names, so I'm open to alternative suggestions if this PR is accepted:
* RENAMED: `T? : ∀ b → Dec (T b)`   (this one is example `README.Design.Decidability.ex₃`)
* SUPPRESSED: `predᵇ⁺ : (p : A → Bool) (x : A) → Dec (T (p x))`
~~and, indeed, to suppressing this second defn on grounds of being 'too trivial'~~

~~Otherwise, good to go, I think.~~

UPDATED: hmmm, didn't quite see through to all the consequences of this, eg even in `Data.Bool.Properties`... will convert to draft, weed out as much as possible, re-commit and then re-open. 

UPDATED: Good to go now! modulo the above naming/usefulness considerations, and my comments about names in these modules in #2058, towards possible additions to this PR, or a subsequent one. 

UPDATED: Comments on the above naming/placing
* ~~`bool⁺`~~ `T?` above, both now is used to define `T?` in `Data.Bool.Properties`, and I think morally should be ~~called this name *in the module* `Relation.Nullary.Decidable.Core`, and then~~ re-exported via `Data.Bool` (which publicly re-exports a lot, but could easily do so from this module instead...)
* ~~`predᵇ⁺` above, arguably does *not* belong in `Relation.Nullary.Decidable.Core`, being a lifting of `T?` (sic) to `Relation.Unary`, so maybe doesn't belong here at all~~
* (not considered so far in this PR): a `Relation.Binary` analogue of `predᵇ⁺`, `relᵇ⁺` say, etc.

~~Suggested additional refactorings (for this PR!)~~

Suggested additional refactorings (as follow up PR? or not at all?)
- [ ] move `T?` (plus deprecation warning) from `Data.Bool.Properties` to `Relation.Nullary.Decidable.Core`, plus its friends `T-irrelevant` and `T?-diag`
- [ ] clean up `import`s from `Data.Bool.Properties`, especially where those *only* import `using (T?)` (baby steps starting with #2061 and #2062)
- [ ] introduce 'lazier' constructors/pattern synonyms than `yes` and `no`, corresponding to `true  because_` and `false because_` (and if so: what should they be called?)
- [ ] move/rename `predᵇ⁺` as `T?` in `Relation.Unary`?
- [ ] define `relᵇ⁺`  as `T?` in `Relation.Binary.?` etc.?


src/Relation/Nullary/Decidable.agda
```diff
@@ -60,8 +60,8 @@ module _ {a₁ a₂ b₁ b₂} {A : Setoid a₁ a₂} {B : Setoid b₁ b₂}
 -- A lemma relating True and Dec
 
 True-↔ : (dec : Dec P) → Irrelevant P → True dec ↔ P
-True-↔ (true  because  [p]) irr = mk↔′ (λ _ → invert [p]) _ (irr (invert [p])) cong′
-True-↔ (false because ofⁿ ¬p) _ = mk↔′ (λ ()) (invert (ofⁿ ¬p)) (⊥-elim ∘ ¬p) λ ()
+True-↔ (yes p) irr = mk↔′ (λ _ → p) _ (irr p) cong′
```
jamesmckinna, line 63: is this definition now too strict?

I confess I find the use of `Reflects.invert` hard to understand, because it only ever seem to occur in the scope of pattern-matching context, of the form `true because ...` or else `false because ...` in which the relevant clause of `invert` can be statically known...

In other words, if we were instead to define
```agda
pattern lazy-yes [p] = true because [p]
pattern lazy-no [¬p] = false because [¬p]
```
then is there any distinction between pattern-matches of the form
```agda
f (lazy-yes [p]) = ... invert [p] ...
f (lazy-no [¬p] = ... invert [¬p] ...
```
and the corresponding
```agda
f (yes p) = ... p ...
f (no ¬p = ... ¬p ...
```
Apologies if this is 'obvious', but I'm confused...

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -332,10 +331,10 @@ Related-cong {A = A} {B} {C} {D} A≈B C≈D =
...
-True↔ ( true because  [p]) irr =
-  inverse (λ _ → invert [p]) (λ _ → _) (λ _ → P.refl) (irr _)
-True↔ (false because ofⁿ ¬p) _ =
-  inverse (λ()) (invert (ofⁿ ¬p)) (λ ()) (⊥-elim ∘ ¬p)
+True↔ (yes p) irr =
+  inverse (λ _ → p) (λ _ → _) (λ _ → P.refl) (irr _)
+True↔ (no ¬p) _ =
+  inverse (λ()) ¬p (λ ()) (⊥-elim ∘ ¬p)
 
 ------------------------------------------------------------------------
```
jamesmckinna, line 339: Same comment about strictness here as in commit f6ff90205b3b7176978de590b533e59226440fd2, but here it's more egregious, as the second pattern really is equivalent to `no ¬p` (only the `true`/`yes`/`lazy-yes` branch is potentially lazier...)

src/Relation/Nullary/Decidable.agda
```diff
@@ -60,8 +60,8 @@ module _ {a₁ a₂ b₁ b₂} {A : Setoid a₁ a₂} {B : Setoid b₁ b₂}
...
+True-↔ (no ¬p) _   = mk↔′ (λ ()) ¬p (⊥-elim ∘ ¬p) λ ()
```
jamesmckinna, line 64: Here it's more egregious, as the second pattern really is equivalent to `no ¬p` (only the `true`/`yes`/`lazy-yes` branch is potentially lazier...)

src/Relation/Nullary/Decidable.agda
jamesmckinna: Well, I've now restored (a version of) the less strict definition, so resolving this now.

src/Relation/Nullary/Decidable.agda
jamesmckinna: Ditto. here, but can actually be made less strict with `invert` in the same way as the `true  because...` clause

src/Function/Related/TypeIsomorphisms.agda
jamesmckinna: And... resolved in favour of a lazy/lazier version in the same way, using a `let`-bound `invert`...

README/Design/Decidability.agda
```diff
@@ -37,22 +37,20 @@ ex₁ : (n : ℕ) → Reflects (zero ≡ suc n) false
 ex₁ n = ofⁿ λ ()
 
 ex₂ : (b : Bool) → Reflects (T b) b
-ex₂ false = ofⁿ id
-ex₂ true  = ofʸ tt
+ex₂ = T⁺
```
MatthewDaggitt, line 40: I'm not keen about these changes to the README file. The whole point was to demonstrate the record fields `does` and `proof` (especially in the second example) but they no longer exist here.

src/Data/Fin/Subset/Properties.agda
```diff
@@ -220,9 +220,9 @@ x∉⁅y⁆⇒x≢y x∉⁅x⁆ refl = x∉⁅x⁆ (x∈⁅x⁆ _)
 infix 4 _⊆?_
 _⊆?_ : B.Decidable {A = Subset n} _⊆_
 []          ⊆? []          = yes id
-outside ∷ p ⊆?       y ∷ q = Dec.map out⊆-⇔ (p ⊆? q)
+outside ∷ p ⊆?       y ∷ q = map′ out⊆ drop-∷-⊆ (p ⊆? q)
```
MatthewDaggitt, line 223: Is there a good reason for all of these changes from the higher-level `map` interface to the lower-level `map′ ` interface? Does it reduce the dependencies somehow? Otherwise I'm not quite sure what the value is? It makes some of the proofs much longer, e.g. `anySubset?` below.

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -52,6 +51,15 @@ open Dec public
 pattern yes p =  true because ofʸ  p
 pattern no ¬p = false because ofⁿ ¬p
 
+------------------------------------------------------------------------
+-- decision procedures are Decidable
```
MatthewDaggitt, line 55: I feel these should live in `Data.Bool.Properties`?

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -52,6 +51,15 @@ open Dec public
...
+
+bool⁺ : ∀ b → Dec (T b)
```
MatthewDaggitt, line 57: Name suggestion: `T?`

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -52,6 +51,15 @@ open Dec public
...
+bool⁺ b = b because (T⁺ b)
+
+predᵇ⁺ : ∀ {a} {A : Set a} (p : A → Bool) (x : A) → Dec (T (p x))
+predᵇ⁺ p x = bool⁺ (p x)
```
MatthewDaggitt, line 61: I'm unsure about how useful this is? Isn't much clearer to write `T? (p x)` rather than `predᵇ⁺ p x`? The former is shorter as well.

MatthewDaggitt (commented):
Otherwise looks good!

src/Data/Fin/Subset/Properties.agda
jamesmckinna: Yes, it does remove the dependency on `Relation.Nullary.Decidable` in favour of the already-being-imported `Relation.Nullary.Decidable.Core`... and when reaidng the proofs, I was realising that within a single module (in each case), an `Equivalence`/`Inverse` bundle was being created only to then be destructed by `Dec.map`, when the pieces of each bundle were already defined. But indeed, perhaps it is too clumsy a refactoring?

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Well I suppose that is part of the question here. My thoughts had been towards moving `Dec`/`Decidable` questions out of `Data.Bool.*` in favour of `Relation.Nullary.Decidable.*` (which is already importing `Bool`...), which further simplifies the imports of `Data.Bool.Properties` (the dependency graph does get a bit hairy back-and-forth between the two, and moreover the `Other properties` section at the end of `Data.Bool.Properties` feels a bit of an orphan, unless it was placeholder for further development?

README/Design/Decidability.agda
jamesmckinna: Yes, I see. Those changes aren't difficult to revert, and would only amount to redundant repetition. But I found myself wanting to 'promote' the definitions from (mere) 'examples' under `README.*` to *bona fide* definitions under `src.*`... if you see what I mean?

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Well, quite, and then we are precisely (redundantly) duplicating the definition between this module and `Data.Bool.Properties`, and as I've argued above (perhaps not altogether successfully), I think they belong better *here*, and not *there*.

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Yes, I was going back on myself about this definition in any case. My (overly-fetishing-uniformity/consistency) brain was thinking that `T?` (as suggested above) from `Relation.Nullary.*` should be being systematically/successively lifted in some way to `Relation.Unary` and `Relation.Binary`... and this was a clumsy way of stumbling on that thought. Happy to suppress it!

jamesmckinna:
Oh! The merge conflicts seem new since I first committed. Will investigate and try to tidy all this up. 
One feature/bug of my thoughts/actions regarding this PR, is that it started *very* small, and then ... got me thinking too much about dependencies... and then got me reconsidering other dependency issues between `Data.Bool.Properties` and `Relation.Nullary.Decidable.Core`... and the whole thing became a bit of a big soup...

README/Design/Decidability.agda
jamesmckinna: Now reverted.

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Have now removed `predᵇ⁺` (but with a thought to replicate it later, under the name `T?`, in `Relation.Unary`).

src/Relation/Nullary/Decidable.agda
```diff
@@ -60,8 +59,10 @@ module _ {a₁ a₂ b₁ b₂} {A : Setoid a₁ a₂} {B : Setoid b₁ b₂}
 -- A lemma relating True and Dec
 
 True-↔ : (dec : Dec P) → Irrelevant P → True dec ↔ P
```
gallais, line 61: This looks like a duplicate of `True↔` from src/Function/Related/TypeIsomorphisms.agda !

JacquesCarette: Isn't the latter deprecated (or about to be)?

jamesmckinna: Re: duplication
I had wondered about this, but there is no new content here from me. Actually, I wasn't at all sure if the arrow was the same... and in the proofs, one uses `mk<funny arrow>`, while the other uses `inverse`... IIRC

Re: deprecation
Again I stand in ignorance. I was blindly chasing dependencies, rather than looking up at the bigger picture to 'see what is going on'. 

Either way, I'd suggest merging my commits, and then figuring out the downsteam deprecation/dependencies, but if you want me to look harder, I guess I can, but not before the weekend...

jamesmckinna: Re: possible deprecation
Can you link the/a relevant issue/PR please?

gallais: None of the two modules involved seem to be deprecated. I agree it's mostly orthogonal to this PR,
it's just something I noticed because the PR was hitting both definitions.

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -52,6 +51,12 @@ open Dec public
 pattern yes p =  true because ofʸ  p
 pattern no ¬p = false because ofⁿ ¬p
 
+------------------------------------------------------------------------
+-- Booleans (decision procedures) are Decidable
+
+T? : ∀ b → Dec (T b)
+T? b = b because (T⁺ b)
```
jamesmckinna, line 58: This addition does, I think, belong in this PR.

src/Relation/Nullary/Reflects.agda
```diff
@@ -37,6 +36,10 @@ data Reflects {p} (P : Set p) : Bool → Set p where
 ------------------------------------------------------------------------
 -- Constructors and destructors
 
+T⁺ : ∀ b → Reflects (T b) b
+T⁺ true  = ofʸ _
+T⁺ false = ofⁿ id
```
jamesmckinna, line 41: This addition does, I think, belong in this PR.

src/Data/Bool/Properties.agda
```diff
@@ -761,9 +760,7 @@ T-irrelevant : U.Irrelevant T
 T-irrelevant {true}  _  _  = refl
 
 T? : U.Decidable T
-does  (T? b) = b
-proof (T? true ) = ofʸ _
-proof (T? false) = ofⁿ λ()
+T? = Dec.T?
```
jamesmckinna, line 763: I'm inclining towards reverting this, in favour of
```agda
T? b = b because (T⁺ b)
```
in order to decouple this change (and the need to reinstate the import of `Relation.Nullary.Reflects`) from any future possible refactoring/dependency simplification of `Data.Bool.Properties`?

jamesmckinna:
Like #2000 i think there's a salvageable core to this.

src/Relation/Nullary/Decidable.agda
jamesmckinna: Indeed, and I think it would be better to  redefine this lemma to be its counterpart. Reason? complex `import` dependencies wrt relevant/irrelevant elimination of \bot...

... but I'm abandoning this PR, because whatever has happened recently, this is now too broken to want to fix, especially for v2.0.

jamesmckinna:
Aaargh. Latest round of commits to `master` seem to have torpedoed this PR. 
Scratching my head... but also abandoning this for now. Not for `v2.0`, by any stretch of the imagination.

MatthewDaggitt:
Okay closing for now then.

PR #2060, opened 2023-08-16
fixes #2046

### LINKED ISSUE #2046: Renaming `Data.Bool.Properties.push-function-into-if`
Following existing library naming conventions regarding commutation properties, and (perhaps too generously) extending the notion of `map` to the degenerate case of function application, I would be very strongly tempted to rename
```agda
push-function-into-if : ∀ (f : A → B) x {y z} →
                        f (if x then y else z) ≡ (if x then f y else f z)
```
to the simpler, more memorable, `map-if`. Failing that, `apply-if`?

jamesmckinna:
(plus tidies up some typos in `CHANGELOG`)

### DIFF
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -762,7 +762,21 @@ proof (T? false) = ofⁿ λ()
 T?-diag : ∀ b → T b → True (T? b)
 T?-diag true  _ = _
 
-push-function-into-if : ∀ (f : A → B) x {y z} →
-                        f (if x then y else z) ≡ (if x then f y else f z)
-push-function-into-if _ true  = refl
-push-function-into-if _ false = refl
+if-float : ∀ (f : A → B) b {xᵗ xᶠ} →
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

src/Data/Bool/Properties.agda
```diff
@@ -762,7 +762,21 @@ proof (T? false) = ofⁿ λ()
 T?-diag : ∀ b → T b → True (T? b)
 T?-diag true  _ = _
 
-push-function-into-if : ∀ (f : A → B) x {y z} →
-                        f (if x then y else z) ≡ (if x then f y else f z)
-push-function-into-if _ true  = refl
-push-function-into-if _ false = refl
+if-float : ∀ (f : A → B) b {xᵗ xᶠ} →
```
MatthewDaggitt, line 765: I'm not a huge fan of the change of `x` and `y` for `xᵗ` and `xᶠ`, as its against the style-guide. Can we switch it back?

MatthewDaggitt (commented):
Otherwise looks good!

PR #2066, opened 2023-08-19
Added functional vector permutation

guilhermehas:
Added permutation of functional vectors using finite sets permutation.


MatthewDaggitt:
Thanks for the PR @guilhermehas!

The general consensus between the maintainers at the moment, is that when adding new definitions we prefer to have at least a few lemmas showing that the definition is correct and relatively easy to work with. Would it be possible to add a few?

guilhermehas:
I have added a few theorems.

jamesmckinna:
So, under the`*.Relation.Binary.*` hierarchy for each `Data`/functorial container type, the library distinguishes between `Properties.Propositional`, where the underlying relation on the carrier type is`PropositionalEquality`, and `Properties.Setoid` for an arbitrary `Setoid`/`isEquivalence`relation on the carrier... with the `Propositional` version derived from the `Setoid` by specialising to `PropositionalEquality`...

If I understand this development properly, what you have is the `Propositional` version. How hard would it be to extend your development to an arbitrary `Setoid`? 

Secondly, it might also be nice to know more than just that `Propositional` lists-modulo-permutation is, indeed, an equivalence relation... such as some *concrete* properties regarding eg. `reverse`, `snoc`, flipped `append`... etc

guilhermehas:
I didn't develop for setoids and relations, not because it is hard, but because it does not make sense.
When dealing with permutations using Vectors, we care a lot more about their indices than the content of the vectors. In addition, it is better to prove `xs ++ (ys ++ zs) = (xs ++ ys) ++ zs` than  `xs ++ (ys ++ zs) ~ (xs ++ ys) ++ zs` because the second theorem can be derived by the first one. So in most cases, propositional equality will be better than setoids when using permutations using finite sets.

gallais:
I think james was suggesting an alternative definition that is not based on permutations on fins.
e.g. http://agda.github.io/agda-stdlib/Data.List.Relation.Binary.Permutation.Homogeneous.html#905

guilhermehas:
I think that this definition can be more intuitive than this propositional definition.
I can prove later that both definitions are the same.

src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
```diff
@@ -0,0 +1,44 @@
...
+open ≡-Reasoning
+
+private
+  variable
+    ℓ : Level
+    A : Set ℓ
+    n : ℕ
+    xs ys zs : Vector A n
+
+↭-refl : xs ↭ xs
```
Taneb, line 29: Could you use the definitions in `Relation.Binary.Indexed.Heterogeneous.Definitions` in this type and the types of the other definitions in this module?

src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
```diff
@@ -0,0 +1,44 @@
...
+-- Properties of permutation
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Vec.Functional.Relation.Binary.Permutation.Properties where
+
+open import Level using (Level)
+open import Data.Product using (_,_; proj₁; proj₂)
+open import Data.Nat using (ℕ)
```
Taneb, line 13: Tighten imports

```suggestion
open import Data.Product.Base using (_,_; proj₁; proj₂)
open import Data.Nat.Base using (ℕ)
```

src/Data/Vec/Functional/Relation/Binary/Permutation.agda
```diff
@@ -0,0 +1,26 @@
...
+--
+-- Permutation relations over Vector
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Vec.Functional.Relation.Binary.Permutation where
+
+open import Level using (Level)
+open import Data.Product using (Σ-syntax)
```
Taneb, line 12: Tighten imports
```suggestion
open import Data.Product.Base using (Σ-syntax)
```

src/Data/Vec/Functional/Relation/Binary/Permutation.agda
```diff
@@ -0,0 +1,26 @@
...
+open import Data.Fin.Permutation using (Permutation; _⟨$⟩ʳ_)
+open import Data.Vec.Functional using (Vector)
+open import Relation.Binary.Indexed.Heterogeneous.Core using (IRel)
+open import Relation.Binary.PropositionalEquality using (_≡_)
```
Taneb, line 16: Tighten imports
```suggestion
open import Relation.Binary.PropositionalEquality.Core using (_≡_)
```

Taneb (changes requested):
I've made some minor suggestions. I think this is a good addition to the library.

src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
guilhermehas: I added these definitions.

Taneb: Could you change the changelog to use them too?

Taneb (approved):

JacquesCarette (approved):

gallais (approved):

PR #2067, opened 2023-08-20
Add (propositional) equational reasoning combinators for vectors

shhyou:
This pull request adds a set of combinators for equational reasoning about vector equality of the form `cast eq xs ≡ ys` and a few properties demonstrating its application. In particular, the new combinators deal with the threading of index equality proofs, greatly reducing the boilerplate that manipulates `cast`s to make writing equational proofs about vectors more or less feasible.

~~(Sorry for the piles of PR and the mess of the commit history. I promise this is the last one I have in hand. The first commit is the content of #2041 and #2045. After these two PRs get in, I'll rebase to remove the first commit. To see the code of this PR, please check <https://github.com/agda/agda-stdlib/commit/f8971193dce2620cdabf4c7f1d8319c2aa8fb0ed>.)~~

## Motivation

Index arithmetic is notably troublesome when it comes to proving vector properties (#942). Existing solutions include stating and proving `cast`-equality by induction, proving [pointwise equality](https://github.com/agda/agda-stdlib/blob/104125c62d88ebe91274ebde521b428bad71f26e/src/Data/Vec/Relation/Binary/Equality/Setoid.agda) (#1668, #1621), or turning to [heterogeneous equality with K](https://github.com/agda/agda-stdlib/blob/104125c62d88ebe91274ebde521b428bad71f26e/src/Data/Vec/Properties/WithK.agda). However, sometimes even an inductive proof can benefit from equational reasoning as seen in `reverse-++`, and that one really wants to convert back to propositional equality at the end to facilitate type casting. Therefore, the goal of this PR is to design a set of combinators that **interoperates with `_≡_` while reducing explicit index castings in the proofs to an tolerable level**.

@JacquesCarette pointed me to @Taneb's nice PR on displayed categories and equational reasoning over some base equality in agda/agda-categories#382. It seems to me that the core combinators in this PR can be seen as a instance of displayed setoid, but I am not sure if I understand it correctly.

In this PR, the new combinators directly operate on equality of the form `cast eq xs ≡ ys`. Fortunately, this type of equality is 'reflexive', 'symmetric' and 'transitive', therefore `cast eq xs ≡ ys` admits the standard equational reasoning combinators. The combinators can hide `cast`s in the terms from the user to reduce redundancy. Index equality is inferred from the proof in each step and automatically threaded throughout the process. In addition, since the underlying type is just `cast eq xs ≡ ys`, the combinators automatically work with `_≡_`.

## Comparison

In principle, `cast` proofs can be written directly without using the combinators in this PR. Sadly, the chaining of the proofs of indices quickly become tedious to the degree of unmanageable. Here are two examples comparing proofs using the new combinators versus the proofs in #2045. It would be even annoying to prove `fromList-reverse`,
```agda
fromList-reverse : cast ... (fromList (List.reverse xs)) ≡ reverse (fromList xs)
```
since the proof necessarily involves `List` equational reasoning that ended up changing the index at every step.

(https://github.com/agda/agda-stdlib/commit/f8971193dce2620cdabf4c7f1d8319c2aa8fb0ed#diff-8e6f618f4d6093cefbe779df9d3964ca20710e5dc33da8596c22da33d761eccdR1033-R1043)
```diff
 cast-reverse : ∀ .(eq : m ≡ n) → cast eq ∘ reverse {A = A} {n = m} ≗ reverse ∘ cast eq
 cast-reverse {n = zero}  eq []       = refl
-cast-reverse {n = suc n} eq (x ∷ xs) = begin
-  cast eq (reverse (x ∷ xs))              ≡⟨ cong (cast eq) (reverse-∷ x xs) ⟩
-  cast eq (reverse xs ∷ʳ x)               ≡⟨ cast-∷ʳ eq x (reverse xs) ⟩
-  (cast (cong pred eq) (reverse xs)) ∷ʳ x ≡⟨ cong (_∷ʳ x) (cast-reverse (cong pred eq) xs) ⟩
-  (reverse (cast (cong pred eq) xs)) ∷ʳ x ≡˘⟨ reverse-∷ x (cast (cong pred eq) xs) ⟩
-  reverse (x ∷ cast (cong pred eq) xs)    ≡⟨⟩
-  reverse (cast eq (x ∷ xs))              ∎
+cast-reverse {n = suc n} eq (x ∷ xs) = begin′
+  reverse (x ∷ xs)           ≂⟨ reverse-∷ x xs ⟩
+  reverse xs ∷ʳ x            ≈⟨ cast-∷ʳ eq x (reverse xs)
+                                ≈cong[ (_∷ʳ x) ] cast-reverse (cong pred eq) xs ⟩
+  reverse (cast _ xs) ∷ʳ x   ≂˘⟨ reverse-∷ x (cast (cong pred eq) xs) ⟩
+  reverse (x ∷ cast _ xs)    ≈⟨⟩
+  reverse (cast eq (x ∷ xs)) ∎′
```

```diff
 reverse-++ : ∀ .(eq : m + n ≡ n + m) (xs : Vec A m) (ys : Vec A n) →
              cast eq (reverse (xs ++ ys)) ≡ reverse ys ++ reverse xs
-reverse-++ {m = zero}  {n = n} eq []       ys = begin
-  cast _ (reverse ys)                ≡˘⟨ cong (cast eq) (++-identityʳ (sym eq) (reverse ys)) ⟩
-  cast _ (cast _ (reverse ys ++ [])) ≡⟨ cast-trans (sym eq) eq (reverse ys ++ []) ⟩
-  cast _ (reverse ys ++ [])          ≡⟨ cast-is-id (trans (sym eq) eq) (reverse ys ++ []) ⟩
-  reverse ys ++ []                   ≡⟨⟩
-  reverse ys ++ reverse []           ∎
+reverse-++ {m = zero}  {n = n} eq []       ys = ≈-sym (++-identityʳ (sym eq) (reverse ys))
-reverse-++ {m = suc m} {n = n} eq (x ∷ xs) ys = begin
-  cast eq (reverse (x ∷ xs ++ ys))                ≡⟨ cong (cast eq) (reverse-∷ x (xs ++ ys)) ⟩
-  cast eq (reverse (xs ++ ys) ∷ʳ x)               ≡˘⟨ cast-trans eq₂ eq₁ (reverse (xs ++ ys) ∷ʳ x) ⟩
-  (cast eq₁ ∘ cast eq₂) (reverse (xs ++ ys) ∷ʳ x)  ≡⟨ cong (cast eq₁) (cast-∷ʳ _ x (reverse (xs ++ ys))) ⟩
-  cast eq₁ ((cast eq₃ (reverse (xs ++ ys))) ∷ʳ x)  ≡⟨ cong (cast eq₁) (cong (_∷ʳ x) (reverse-++ _ xs ys)) ⟩
-  cast eq₁ ((reverse ys ++ reverse xs)      ∷ʳ x) ≡⟨ ++-∷ʳ _ x (reverse ys) (reverse xs) ⟩
-  reverse ys ++ (reverse xs ∷ʳ x)                 ≡˘⟨ cong (reverse ys ++_) (reverse-∷ x xs) ⟩
-  reverse ys ++ (reverse (x ∷ xs))                ∎
-  where
-  eq₁ = sym (+-suc n m)
-  eq₂ = cong suc (+-comm m n)
-  eq₃ = cong pred eq₂
+reverse-++ {m = suc m} {n = n} eq (x ∷ xs) ys = begin′
+  reverse (x ∷ xs ++ ys)              ≂⟨ reverse-∷ x (xs ++ ys) ⟩
+  reverse (xs ++ ys) ∷ʳ x             ≈⟨ cast-∷ʳ (cong suc (+-comm m n)) x (reverse (xs ++ ys))
+                                         ≈cong[ (_∷ʳ x) ] reverse-++ _ xs ys ⟩
+  (reverse ys ++ reverse xs) ∷ʳ x     ≈⟨ ++-∷ʳ (sym (+-suc n m)) x (reverse ys) (reverse xs) ⟩
+  reverse ys ++ (reverse xs ∷ʳ x)     ≂˘⟨ cong (reverse ys ++_) (reverse-∷ x xs) ⟩
+  reverse ys ++ (reverse (x ∷ xs))    ∎′
```

## Limitation

Here are some limitations I found while proving properties in this PR using the proposed combinators.

1. The combinators heavily expects other lemmas to be properly shaped in the form of `case eq xs ≡ ys`. This is not surprising, but requires some sort of convention.
2. The combinators don't help with more primitive lemmas that have to be proved directly using fold fusion. In other words, without the more fundamentals lemmas like `reverse-∷`, `unfold-ʳ++` etc, the combinators probably can't prove anything. Thanks to @jamesmckinna for porting the `reverse` lemmas for vectors!
3. An equational reasoning step `xs ≈⟨ xs≈ys ⟩ ys≈zs` obtains the underlying `eq` of `cast` from the result type and from the proof `xs≈ys`. Therefore, in the middle of an equational proof, the underlying `eq` will become an unsolved metavariable.

    ```agda
    pf : cast eq xs ≡ ys
    pf = begin
      xs ≈⟨ xs≈zs ⟩
      zs ≈⟨ {! !} ⟩ -- this step introduces an unsolved metavariable until it's filled in
      ys ∎
    ```

### DIFF
--- a/README/Data/Vec/Relation/Binary/Equality/Cast.agda
+++ b/README/Data/Vec/Relation/Binary/Equality/Cast.agda
@@ -226,22 +226,22 @@ example5-fromList-++-++′ : {xs ys zs : List A} →
                                   L.length xs + (L.length ys + L.length zs)) →
                            cast eq (fromList (xs L.++ ys L.++ zs)) ≡
                                    fromList xs ++ fromList ys ++ fromList zs
-example5-fromList-++-++′ {xs = xs} {ys} {zs} eq = begin′
+example5-fromList-++-++′ {xs = xs} {ys} {zs} eq = begin
   fromList (xs L.++ ys L.++ zs)                 ≈⟨ fromList-++ xs ⟩
   fromList xs ++ fromList (ys L.++ zs)          ≃⟨ cast-++ʳ (Lₚ.length-++ ys) (fromList xs) ⟩
   fromList xs ++ cast _ (fromList (ys L.++ zs)) ≡⟨ cong (fromList xs ++_) (fromList-++ ys) ⟩
-  fromList xs ++ fromList ys ++ fromList zs     ∎
-  where open ≡-Reasoning; open CastReasoning renaming (begin_ to begin′_; _∎ to _∎′)
+  fromList xs ++ fromList ys ++ fromList zs     ≡-∎
+  where open CastReasoning
 
 -- Of course, it is possible to start with the reasoning system of `_≡_`
 -- and then switch to the reasoning system of `_≈[_]_`.
 example6a-reverse-∷ʳ : ∀ x (xs : Vec A n) → reverse (xs ∷ʳ x) ≡ x ∷ reverse xs
-example6a-reverse-∷ʳ {n = n} x xs = begin
+example6a-reverse-∷ʳ {n = n} x xs = begin-≡
   reverse (xs ∷ʳ x)     ≡˘⟨ ≈-reflexive refl ⟩
   reverse (xs ∷ʳ x)     ≈⟨ ≈-cong reverse (cast-reverse _ _) (unfold-∷ʳ (+-comm 1 n) x xs) ⟩
   reverse (xs ++ [ x ]) ≈⟨ reverse-++ (+-comm n 1) xs [ x ] ⟩
-  x ∷ reverse xs        ∎′
-  where open ≡-Reasoning; open CastReasoning renaming (begin_ to begin′_; _∎ to _∎′)
+  x ∷ reverse xs        ∎
+  where open CastReasoning
 
 example6b-reverse-∷ʳ-by-induction : ∀ x (xs : Vec A n) → reverse (xs ∷ʳ x) ≡ x ∷ reverse xs
 example6b-reverse-∷ʳ-by-induction x []       = refl
--- a/README/Data/Vec/Relation/Binary/Equality/Cast.agda
+++ b/README/Data/Vec/Relation/Binary/Equality/Cast.agda
@@ -1,0 +1,259 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- An equational reasoning library for propositional equality over
+-- vectors of different indices using cast.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module README.Data.Vec.Relation.Binary.Equality.Cast where
+
+open import Agda.Primitive
+open import Data.List.Base as L using (List)
+import Data.List.Properties as Lₚ
+open import Data.Nat.Base
+open import Data.Nat.Properties
+open import Data.Vec.Base
+open import Data.Vec.Properties
+import Data.Vec.Relation.Binary.Equality.Cast as CastReasoning
+open import Relation.Binary.PropositionalEquality
+  using (_≡_; refl; trans; sym; cong; subst; module ≡-Reasoning)
+
+private variable
+  a : Level
+  A : Set a
+  l m n o : ℕ
+  xs ys zs ws : Vec A n
+
+
+-- To see example usages of this library, scroll to the combinators
+-- section.
+
+
+------------------------------------------------------------------------
+--
+-- Motivation
+--
+-- The `cast` function is the computational variant of `subst` for
+-- vectors. Since `cast` computes under vector constructors, it
+-- enables reasoning about vectors with non-definitionally equal indices
+-- by induction. See, e.g., Jacques Carette's comment in issue #1668.
+-- <https://github.com/agda/agda-stdlib/pull/1668#issuecomment-1003449509>
+--
+-- Suppose we want to prove that ‘xs ++ [] ≡ xs’. Because `xs ++ []`
+-- has type `Vec A (n + 0)` while `xs` has type `Vec A n`, they cannot
+-- be directly related by homogeneous equality.
+-- To resolve the issue, `++-right-identity` uses `cast` to recast
+-- `xs ++ []` as a vector in `Vec A n`.
+--
+++-right-identity : ∀ .(eq : n + 0 ≡ n) (xs : Vec A n) → cast eq (xs ++ []) ≡ xs
+++-right-identity eq []       = refl
+++-right-identity eq (x ∷ xs) = cong (x ∷_) (++-right-identity (cong pred eq) xs)
+--
+-- When the input is `x ∷ xs`, because `cast eq (x ∷ _)` equals
+-- `x ∷ cast (cong pred eq) _`, the proof obligation
+--     cast eq (x ∷ xs ++ []) ≡ x ∷ xs
+-- simplifies to
+--     x :: cast (cong pred eq) (xs ++ []) ≡ x ∷ xs
+
+
+-- Although `cast` makes it possible to prove vector identities by ind-
+-- uction, the explicit type-casting nature poses a significant barrier
+-- to code reuse in larger proofs. For example, consider the identity
+-- ‘fromList (xs L.∷ʳ x) ≡ (fromList xs) ∷ʳ x’ where `L._∷ʳ_` is the
+-- snoc function of lists. We have
+--
+--     fromList (xs L.∷ʳ x)            : Vec A (L.length (xs L.∷ʳ x))
+--   =   {- by definition -}
+--     fromList (xs L.++ L.[ x ])      : Vec A (L.length (xs L.++ L.[ x ]))
+--   =   {- by fromList-++ -}
+--     fromList xs ++ fromList L.[ x ] : Vec A (L.length xs + L.length [ x ])
+--   =   {- by definition -}
+--     fromList xs ++ [ x ]            : Vec A (L.length xs + 1)
+--   =   {- by unfold-∷ʳ -}
+--     fromList xs ∷ʳ x                : Vec A (suc (L.length xs))
+-- where
+--     fromList-++ : cast _ (fromList (xs L.++ ys)) ≡ fromList xs ++ fromList ys
+--     unfold-∷ʳ   : cast _ (xs ∷ʳ x) ≡ xs ++ [ x ]
+--
+-- Although the identity itself is simple, the reasoning process changes
+-- the index in the type twice. Consequently, its Agda translation must
+-- insert two `cast`s in the proof. Moreover, the proof first has to
+-- rearrange (the Agda version of) the identity into one with two
+-- `cast`s, resulting in lots of boilerplate code as demonstrated by
+-- `example1a-fromList-∷ʳ`.
+example1a-fromList-∷ʳ : ∀ (x : A) xs .(eq : L.length (xs L.∷ʳ x) ≡ suc (L.length xs)) →
+                        cast eq (fromList (xs L.∷ʳ x)) ≡ fromList xs ∷ʳ x
+example1a-fromList-∷ʳ x xs eq = begin
+  cast eq (fromList (xs L.∷ʳ x))                   ≡⟨⟩
+  cast eq (fromList (xs L.++ L.[ x ]))             ≡˘⟨ cast-trans eq₁ eq₂ (fromList (xs L.++ L.[ x ])) ⟩
+  cast eq₂ (cast eq₁ (fromList (xs L.++ L.[ x ]))) ≡⟨ cong (cast eq₂) (fromList-++ xs) ⟩
+  cast eq₂ (fromList xs ++ [ x ])                  ≡⟨ CastReasoning.≈-sym (unfold-∷ʳ (sym eq₂) x (fromList xs)) ⟩
+  fromList xs ∷ʳ x                                 ∎
+  where
+  open ≡-Reasoning
+  eq₁ = Lₚ.length-++ xs {L.[ x ]}
+  eq₂ = +-comm (L.length xs) 1
+
+-- The `cast`s are irrelevant to core of the proof. At the same time,
+-- they can be inferred from the lemmas used during the reasoning steps
+-- (e.g. `fromList-++` and `unfold-∷ʳ`). To eliminate the boilerplate,
+-- this library provides a set of equational reasoning combinators for
+-- equality of the form `cast eq xs ≡ ys`.
+example1b-fromList-∷ʳ : ∀ (x : A) xs .(eq : L.length (xs L.∷ʳ x) ≡ suc (L.length xs)) →
+                        cast eq (fromList (xs L.∷ʳ x)) ≡ fromList xs ∷ʳ x
+example1b-fromList-∷ʳ x xs eq = begin
+  fromList (xs L.∷ʳ x)       ≈⟨⟩
+  fromList (xs L.++ L.[ x ]) ≈⟨ fromList-++ xs ⟩
+  fromList xs ++ [ x ]       ≈˘⟨ unfold-∷ʳ (+-comm 1 (L.length xs)) x (fromList xs) ⟩
+  fromList xs ∷ʳ x           ∎
+  where open CastReasoning
+
+
+
+
+------------------------------------------------------------------------
+--
+-- Combinators
+--
+-- Let `xs ≈[ m≡n ] ys` denote `cast m≡n xs ≡ ys`. We have reflexivity,
+-- symmetry and transitivity:
+--     ≈-reflexive : xs ≈[ refl ] xs
+--     ≈-sym       : xs ≈[ m≡n ] ys → ys ≈[ sym m≡n ] xs
+--     ≈-trans     : xs ≈[ m≡n ] ys → ys ≈[ n≡o ] zs → xs ≈[ trans m≡n n≡o ] zs
+-- Accordingly, `_≈[_]_` admits the standard set of equational reasoning
+-- combinators. Suppose `≈-eqn : xs ≈[ m≡n ] ys`,
+--     xs ≈⟨ ≈-eqn  ⟩   -- `_≈⟨_⟩_` takes a `_≈[_]_` step, adjusting
+--     ys               -- the index at the same time
+--
+--     ys ≈˘⟨ ≈-eqn ⟩   -- `_≈˘⟨_⟩_` takes a symmetric `_≈[_]_` step
+--     xs
+example2a : ∀ .(eq : suc m + n ≡ m + suc n) (xs : Vec A m) a ys →
+            cast eq ((reverse xs ∷ʳ a) ++ ys) ≡ reverse xs ++ (a ∷ ys)
+example2a eq xs a ys = begin
+  (reverse xs ∷ʳ a) ++ ys ≈⟨ ∷ʳ-++ eq a (reverse xs) ⟩ -- index: suc m + n
+  reverse xs ++ (a ∷ ys)  ∎                            -- index: m + suc n
+  where open CastReasoning
+
+-- To interoperate with `_≡_`, this library provides `_≂⟨_⟩_` (\-~) for
+-- taking a `_≡_` step during equational reasoning.
+-- Let `≡-eqn : xs ≡ ys`, then
+--     xs ≂⟨ ≡-eqn  ⟩    -- Takes a `_≡_` step; no change to the index
+--     ys
+--
+--     ys ≂˘⟨ ≡-eqn ⟩    -- Takes a symmetric `_≡_` step
+--     xs
+-- Equivalently, `≈-reflexive` injects `_≡_` into `_≈[_]_`. That is,
+-- `xs ≂⟨ ≡-eqn ⟩ ys` is the same as `xs ≈⟨ ≈-reflexive ≡-eqn ⟩ ys`.
+-- Extending `example2a`, we have:
+example2b : ∀ .(eq : suc m + n ≡ m + suc n) (xs : Vec A m) a ys →
+            cast eq ((a ∷ xs) ʳ++ ys) ≡ xs ʳ++ (a ∷ ys)
+example2b eq xs a ys = begin
+  (a ∷ xs) ʳ++ ys         ≂⟨ unfold-ʳ++ (a ∷ xs) ys ⟩          -- index: suc m + n
+  reverse (a ∷ xs) ++ ys  ≂⟨ cong (_++ ys) (reverse-∷ a xs) ⟩  -- index: suc m + n
+  (reverse xs ∷ʳ a) ++ ys ≈⟨ ∷ʳ-++ eq a (reverse xs) ⟩         -- index: suc m + n
+  reverse xs ++ (a ∷ ys)  ≂˘⟨ unfold-ʳ++ xs (a ∷ ys) ⟩         -- index: m + suc n
+  xs ʳ++ (a ∷ ys)         ∎                                    -- index: m + suc n
+  where open CastReasoning
+
+-- Oftentimes index-changing identities apply to only part of the proof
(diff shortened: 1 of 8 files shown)


jamesmckinna:
Very nice account in the documentation above. 
Hope to have time to actually play with the proofs a bit before offering review comments.

src/Data/Vec/Properties.agda
```diff
@@ -12,14 +12,16 @@ open import Algebra.Definitions
...
 open import Data.Nat.Base
 open import Data.Nat.Properties
-  using (+-assoc; m≤n⇒m≤1+n; ≤-refl; ≤-trans; suc-injective)
+  using (+-assoc; m≤n⇒m≤1+n; ≤-refl; ≤-trans; suc-injective; +-comm; +-suc)
 open import Data.Product.Base as Prod
   using (_×_; _,_; proj₁; proj₂; <_,_>; uncurry)
 open import Data.Sum.Base using ([_,_]′)
 open import Data.Sum.Properties using ([,]-map)
 open import Data.Vec.Base
+open import Data.Vec.Relation.Binary.Reasoning.Propositional as VecReasoning renaming (begin_ to begin′_; _∎ to _∎′)
```
jamesmckinna, line 24: This is the one thing I can see straightaway which looks sub-optimal. You shouldn't need to rename the combinators you've worked hard to develop. There is an issue about overlapping reasoning system coexisting within a given development,so that is why [the style guide recommends](https://github.com/agda/agda-stdlib/blob/fd541d090bd2cb18f7eedd4c74d33fff2f953828/notes/style-guide.md?plain=1#L245-L254) that appeals to such things be localised to a proof, and that the tools themselves be packaged as a module for selective `open`ing.

Here, I would make the `open import` into an `import`, and then use `open VecReasoning` when you need the enhanced combinators... and that way, no renaming will be required!

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+open import Relation.Binary.Definitions using (Sym; Trans)
+open import Relation.Binary.PropositionalEquality.Core
+  using (_≡_; refl; trans; sym; cong; module ≡-Reasoning)
+
+private
+  variable
+    l m n o : ℕ
+    xs ys zs : Vec A n
+
+-- Duplicate definition of cast-is-id and cast-trans to avoid circular dependency
```
jamesmckinna, line 24: What is the circular dependency here?

If you want to import `Data.Vec.Properties`, but then use this reasoning module for subsequent lemmas in the rest of that file, then the 'obvious', but perhaps painful, solution, is to make the contents of your `Data.Vec.Relation.Binary.Reasoning.Propositional` a *sub*module of `Data.Vec.Properties`. 

The alternative is to make the parametrisation of this module more complex, and then re-import it in `Data.Vec.Properties` once the `cast` lemmas have been proved. Compare [how this is done in `Data.Integer.Properties`](https://github.com/agda/agda-stdlib/blob/fd541d090bd2cb18f7eedd4c74d33fff2f953828/src/Data/Integer/Properties.agda#L361-L375) for example


Suggest we poll @MatthewDaggitt on the best way to proceed here.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+example : ∀ .(eq : (m + 1) + n ≡ n + suc m) a (xs : Vec A m) (ys : Vec A n) →
+          cast eq (reverse ((xs ++ [ a ]) ++ ys)) ≡ ys ʳ++ reverse (xs ∷ʳ a)
+example {m = m} {n} eq a xs ys = begin
+  reverse ((xs ++ [ a ]) ++ ys)       ≈˘⟨ cast-reverse (cong (_+ n) (ℕₚ.+-comm 1 m)) ((xs ∷ʳ a) ++ ys)
+                                          ≈cong[ reverse ] cast-++ˡ (ℕₚ.+-comm 1 m) (xs ∷ʳ a)
+                                                           ≈cong[ (_++ ys) ] unfold-∷ʳ _ a xs ⟩
+  reverse ((xs ∷ʳ a) ++ ys)           ≈⟨ reverse-++ (ℕₚ.+-comm (suc m) n) (xs ∷ʳ a) ys ⟩
+  reverse ys ++ reverse (xs ∷ʳ a)     ≂˘⟨ unfold-ʳ++ ys (reverse (xs ∷ʳ a)) ⟩
+  ys ʳ++ reverse (xs ∷ʳ a)            ∎
+-}
```
jamesmckinna, line 139: Suggest that in addition to your documentation on the PR, that you lift out this material to the `README` sub-hierarchy, at eg `README.Data.Vec.Reasoning` (again, worth asking for suggestions/comments as to the best place to put such a module).
That way, we don't need to read PR documentation, or not be able to check code because it is commented out. Much better for all potential clients/users!

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+  cast-trans {m = suc _} {n = suc _} {o = suc _} eq₁ eq₂ (x ∷ xs) =
+    cong (x ∷_) (cast-trans (suc-injective eq₁) (suc-injective eq₂) xs)
+
+
+
+≈-by : .(eq : n ≡ m) → REL (Vec A n) (Vec A m) _
+≈-by eq xs ys = (cast eq xs ≡ ys)
+
+infix 3 ≈-by
+syntax ≈-by n≡m xs ys = xs ≈[ n≡m ] ys
```
MatthewDaggitt, line 42: How about we rename this `_≡[_]_` directly?

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+------------------------------------------------------------------------
+-- Reasoning combinators
+
+begin_ : ∀ .{m≡n : m ≡ n} {xs : Vec A m} {ys} → xs ≈[ m≡n ] ys → cast m≡n xs ≡ ys
+begin xs≈ys = xs≈ys
+
+_∎ : (xs : Vec A n) → cast refl xs ≡ xs
+_∎ xs = ≈-reflexive refl
+
+_≈⟨⟩_ : ∀ .{m≡n : m ≡ n} (xs : Vec A m) {ys} → (xs ≈[ m≡n ] ys) → (xs ≈[ m≡n ] ys)
```
MatthewDaggitt, line 75: A lot of these brackets around the cast equalities are redundant, e.g. `(xs ≈[ m≡n ] ys)` can just be `xs ≈[ m≡n ] ys`. Please could you change everywhere?

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
MatthewDaggitt: I suggest we simply re-export this version of the proofs in `Data.Vec.Properties`

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The basic code for equational reasoning with displayed propositional equality over vectors
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Vec.Relation.Binary.Reasoning.Propositional {a} {A : Set a} where
```
MatthewDaggitt, line 9: How about renaming this `Data.Vec.Relation.Binary.Equality.Cast`?

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The basic code for equational reasoning with displayed propositional equality over vectors
```
MatthewDaggitt, line 4: Can you make sure the comments wrap properly?

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+
+
+
+≈-by : .(eq : n ≡ m) → REL (Vec A n) (Vec A m) _
+≈-by eq xs ys = (cast eq xs ≡ ys)
+
+infix 3 ≈-by
+syntax ≈-by n≡m xs ys = xs ≈[ n≡m ] ys
+
+----------------------------------------------------------------
```
MatthewDaggitt, line 44: Can you make sure the section headings stay the same standard length everywhere please!

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
```diff
@@ -0,0 +1,139 @@
...
+step-≃ xs ys≡zs xs≈ys = ≈-trans xs≈ys (≈-reflexive ys≡zs)
+
+-- composition of the equality type on the left-hand side of _≈[_]_
+step-≂ : ∀ .{m≡n : m ≡ n} (xs : Vec A m) {ys zs} → (ys ≈[ m≡n ] zs) → (xs ≡ ys) → (xs ≈[ m≡n ] zs)
+step-≂ xs ys≈zs xs≡ys = ≈-trans (≈-reflexive xs≡ys) ys≈zs
+
+-- `cong` after a `_≈[_]_` step that exposes the `cast` to the `cong` operation
+≈-cong : ∀ .{l≡o : l ≡ o} .{m≡n : m ≡ n} {xs : Vec A m} {ys zs} (f : Vec A o → Vec A n) →
+         (ys ≈[ l≡o ] zs) → (xs ≈[ m≡n ] f (cast l≡o ys)) → (xs ≈[ m≡n ] f zs)
+≈-cong f ys≈zs xs≈fys = trans xs≈fys (cong f ys≈zs)
```
MatthewDaggitt, line 94: I don't think this really should be part of the reasoning interface (e.g. with syntax declarations etc.). Should just be a normal lemma or fully expanded.

src/Data/Vec/Properties.agda
shhyou: I moved the `open` statement to local `where` blocks. However, this library is intended to coexist with the standard propositional equational reasoning combinators, so `begin_` and QED still have conflicts as in <https://github.com/agda/agda-stdlib/pull/2067/commits/da40fb82556b0aa23cb61e0692223f475795b0df>.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Done!

To avoid name collisions, I keep `CastReasoning.cast-is-id` & `CastReasoning.cast-trans` private and re-implemented `Data.Vec.Properties.cast-{is-id,trans}` in terms of `≈-reflexive` & `≈-trans`.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Good idea! I'm working on this.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Done. I name it `_≈[_]_`  so it's coherent with `≈-{refl,sym,trans}` and the equational reasoning combinators; not sure if you intended to change `≈` to `≡`.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Fixed.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Done, thanks!

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Fixed -- sorry I didn't notice the width requirement in the style guide.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Fixed.

src/Data/Vec/Relation/Binary/Reasoning/Propositional.agda
shhyou: Here `≈-cong` is indeed an ordinary lemma. The syntax declaration `_≈cong[_]_` adjust the argument order so it reads more naturally in the context of equational reasoning. Could you expand more on 'normal lemma', 'fully expanded' or 'reasoning interface'?

In some sense, `≈-cong` is the analogy of `cong` for `≈` because plain `cong` doesn't work for `_≈[_]_`. If it is helpful, I'm happy to document how `≈-cong` is useful, why `cong` doesn't work and what boilerplate `≈-cong` has saved. The alternatives I've tried necessarily duplicate code.

shhyou:
README added!

@jamesmckinna  -- Following https://github.com/agda/agda-stdlib/pull/2067#discussion_r1326781475, I moved the `open` statement to local `where` blocks. However, this library is intended to coexist with the standard propositional equational reasoning combinators, so `begin_` and QED still have conflicts as in <https://github.com/agda/agda-stdlib/pull/2067/commits/da40fb82556b0aa23cb61e0692223f475795b0df>.

@MatthewDaggitt -- After https://github.com/agda/agda-stdlib/pull/2067#discussion_r1328039322, I deleted the syntax declaration `_≈cong[_]_` and instead use the ordinary lemma `≈-cong` directly. Please let me know if that's undesired. `≈-cong` is used in 7 out of 8 lemmas added alongside this PR so I think it's useful.

jamesmckinna:
More on 'conflicting' systems of `Reasoning`: the main examples of this seems to be in `Dta.Vec.Properties`, which begins with a *global* opening of  `≡-Reasoning`, so removing that would then show all the places the various reasoning modules need to be opened *locally*.

A slightly refined version of that would be to enclose lemmas which only require `≡-Reasoning` to be inside an anonymous `module _` scope, within which you `open ≡-Reasoning`, and correspondingly (ege `reverse and append`) a module in which you locally open `CastReasoning`. 

Either way, i think that all the `renaming` directives can (still) be avoided.

jamesmckinna:
A variant of this approach is in the the `Reasoning` module for `Strict` order relations, where a distinction is made (and syntax appropriately introduced for) between  `begin-strict` reasoning, which combines equality steps and ordered steps, and 'pure' equational proofs introduced with `begin-equality`... see `Data.Integer.Properties` for examples of this.

That might make an alternative design, but I think at the moment that simply finding (suitable) ways to localise appeals to reasoning modules should suffice.

HTH!

shhyou:
Sure. Done!

I opened `≡-Reasoning` locally instead of creating an anonymous module because there will be too many lemmas to move in that case.

src/Data/Vec/Relation/Binary/Equality/Cast.agda
```diff
@@ -0,0 +1,124 @@
...
+  cast (trans m≡n n≡o) xs ≡˘⟨ cast-trans m≡n n≡o xs ⟩
+  cast n≡o (cast m≡n xs)  ≡⟨ cong (cast n≡o) xs≈ys ⟩
+  cast n≡o ys             ≡⟨ ys≈zs ⟩
+  zs                      ∎
+  where open ≡-Reasoning
+
+------------------------------------------------------------------------
+-- Reasoning combinators
+
+begin_ : ∀ .{m≡n : m ≡ n} {xs : Vec A m} {ys} → xs ≈[ m≡n ] ys → cast m≡n xs ≡ ys
```
MatthewDaggitt, line 70: The standard approach in the library is to put everything below this point in a named module called `CastReasoning`. Then you can open this top-level module without polluting the namespace with all the reasoning combinators. Any chance we could do this?

MatthewDaggitt (commented):
Thanks, fantastic documentation! A couple more minor comments and then it should be ready to go!

README/Data/Vec/Relation/Binary/Equality/Cast.agda
```diff
@@ -0,0 +1,259 @@
...
+  l m n o : ℕ
+  xs ys zs ws : Vec A n
+
+
+-- To see example usages of this library, scroll to the combinators
+-- section.
+
+
+------------------------------------------------------------------------
+--
```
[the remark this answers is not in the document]
shhyou: Fixed.

src/Data/Vec/Properties.agda
```diff
@@ -365,8 +364,7 @@ lookup∘update′ {i = i} {j} i≢j xs y = lookup∘updateAt′ i j i≢j xs
 -- cast
 
 cast-is-id : .(eq : m ≡ m) (xs : Vec A m) → cast eq xs ≡ xs
-cast-is-id eq []       = refl
-cast-is-id eq (x ∷ xs) = cong (x ∷_) (cast-is-id (suc-injective eq) xs)
+cast-is-id _ _ = CastReasoning.≈-reflexive refl
```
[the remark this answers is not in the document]
shhyou: Done! I was worried that there may be name conflicts but actually re-exporting the same definition solves the issue.

src/Data/Vec/Relation/Binary/Equality/Cast.agda
```diff
@@ -0,0 +1,124 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- An equational reasoning library for propositional equality over
+-- vectors of different indices using cast.
+--
+-- See README.Data.Vec.Relation.Binary.Equality.Cast for
```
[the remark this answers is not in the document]
shhyou: Done. Thanks!

src/Data/Vec/Relation/Binary/Equality/Cast.agda
shhyou: Sounds good. Done!

I saw modules in `Relation.Binary.Reasoning.*` put all combinators at the top-level and thought that it was more preferable.

shhyou:
Thanks again for the review effort!

README/Data/Vec/Relation/Binary/Equality/Cast.agda
```diff
@@ -0,0 +1,259 @@
...
+  reverse ((xs ∷ʳ a) ++ ys)       ≈⟨ reverse-++ (+-comm (suc m) n) (xs ∷ʳ a) ys ⟩
+  reverse ys ++ reverse (xs ∷ʳ a) ≂˘⟨ unfold-ʳ++ ys (reverse (xs ∷ʳ a)) ⟩
+  ys ʳ++ reverse (xs ∷ʳ a)        ∎
+  where open CastReasoning
+
+------------------------------------------------------------------------
+--
+-- Interoperation between `_≈[_]_` and `_≡_`
+--
+-- This library is designed to interoperate with `_≡_`. Examples in the
```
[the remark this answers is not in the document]
shhyou: Done as in 2be93c951a. I ended up re-exporting everything from `≡-Reasoning` while renaming its `begin_` to `begin-≡_` and `_∎` to `_≡-∎`.

The two examples in this section is non-standard in that they switch from one reasoning system to another in the middle.

jamesmckinna:
Once @MatthewDaggitt is happy, I'm happy! 
This is a very nice piece of work that (should!) finally lay #942 (and the loose ends of #1668) to rest.

MatthewDaggitt:
Thanks looks good! Merging in now.

Issue #2070, opened 2023-08-21
Orders without equivalence relations


laMudri:
Sometimes we come across types which naturally have an order, but for which the appropriate equivalence relation is, intensionally, the intersection of the order and its inverse. For example, if we have types about which we only care about inhabitedness (propositions), then the order is given by `→`, while the equivalence relation is given by bi-implication. Giving this definition using `Poset` from `Relation.Binary` is annoying because we have to prove reflexivity and transitivity for implication, then prove them twice more for equivalence, and then give (simple, but slightly tedious) proofs of symmetry and antisymmetry.

It would be nice if there were some general code to take a type with a reflexive transitive relation and produce the corresponding poset. Could the structure be slotted into the `Relation.Binary` hierarchy, perhaps even before `Setoid`, together with the construction suggested above? I've occasionally heard the term “proset” used for a partially ordered set without an equivalence relation, though nLab doesn't distinguish this from partially ordered sets in the normal sense.

gallais:
I suppose we could have something like:

http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Closure.Equivalence.html

MatthewDaggitt:
Yup adding a `Construct` module seems to be the right way to go about it.

laMudri:
See #2071.

jamesmckinna:
'Inverse' or 'converse' relation? 
One reason I'm less happy on the PR with 'core groupoid' as justification for the terminology 'core' is that you're not actually inverting anything here, are you, in the sense of having left- and right- inverses to an arrow? (Or are you, if I squint hard enough?)

Calling the opposite/converse of a relation 'inverse' seems... perverse, to me at least.

jamesmckinna:
As for the lexicon shift from `core` to `interior` (as a dual to `closure`, in the context of topology originally, AFAIK): 

Just as the `P`-closure of a relation `R` for some property `P` is the *least* relation `S` containing `R` such that `P S`, the `P`-*interior* is the *dual* notion, vs. the *greatest* relation `S` *contained in* `R` such that `P S`, for any suitable `P`. 

This issue/PR addresses the case `P = Symmetric`. 

[...]

PR #2071, opened 2023-08-21
[ new ] symmetric core of a binary relation

laMudri:
Addresses #2070. The name of the central definition is based on that of the [core groupoid](https://ncatlab.org/nlab/show/core+groupoid). It is distinct from the symmetric closure; the symmetric core is a smaller relation than the original, while the symmetric closure is larger.

I put `Proset` in the same file as everything else, but maybe it should be in the main `Relation.Binary.Bundles` file.

I intended to use `Relation.Binary.Construct.Intersection`, but found that the unification behaviour of `_∩_` in conjunction with `Reflexive`, `Symmetric`, &c was poor. I don't like the names of the fields in `SymCore`, and maybe they could be something better, or even just `proj₁` and `proj₂`. It would be nice if `_∩_` were defined as a record type, but that would be a somewhat breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ Deprecated names
 New modules
 -----------
 
+* Symmetric interior of a binary relation
+  ```
+  Relation.Binary.Construct.Interior.Symmetric
+  ```
+
 Additions to existing modules
 -----------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,3 +90,8 @@ Additions to existing modules
   ```agda
   map : (Char → Char) → String → String
   ```
+
+* Added new definitions in `Relation.Binary.Definitions`:
+  ```agda
+  Empty _∼_ = ∀ {x y} → x ∼ y → ⊥
+  ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,20 @@ Highlights
 Bug-fixes
 ---------
 
+* Fix statement of `Data.Vec.Properties.toList-replicate`, where `replicate`
+  was mistakenly applied to the level of the type `A` instead of the
+  variable `x` of type `A`.
+
 Non-backwards compatible changes
 --------------------------------
+
+* The modules and morphisms in `Algebra.Module.Morphism.Structures` are now
+  parametrized by _raw_ bundles rather than lawful bundles, in line with other
+  modules defining morphism structures.
+* The definitions in `Algebra.Module.Morphism.Construct.Composition` are now
+  parametrized by _raw_ bundles, and as such take a proof of transitivity.
+* The definitions in `Algebra.Module.Morphism.Construct.Identity` are now
+  parametrized by _raw_ bundles, and as such take a proof of reflexivity.
 
 Other major improvements
 ------------------------
@@ -20,6 +32,11 @@ Deprecated modules
 
 Deprecated names
 ----------------
+
+* In `Algebra.Properties.Semiring.Mult`:
+  ```agda
+  1×-identityʳ  ↦  ×-homo-1
+  ```
 
 * In `Data.Nat.Divisibility.Core`:
   ```agda
@@ -34,8 +51,73 @@ New modules
   Relation.Binary.Construct.Interior.Symmetric
   ```
 
+* `Algebra.Module.Bundles.Raw`: raw bundles for module-like algebraic structures
+
+* Nagata's construction of the "idealization of a module":
+  ```agda
+  Algebra.Module.Construct.Idealization
+  ```
+
 Additions to existing modules
 -----------------------------
+
+* Exporting more `Raw` substructures from `Algebra.Bundles.Ring`:
+  ```agda
+  rawNearSemiring   : RawNearSemiring _ _
+  rawRingWithoutOne : RawRingWithoutOne _ _
+  +-rawGroup        : RawGroup _ _
+  ```
+
+* In `Algebra.Module.Bundles`, raw bundles are re-exported and the bundles expose their raw counterparts.
+
+* In `Algebra.Module.Construct.DirectProduct`:
+  ```agda
+  rawLeftSemimodule  : RawLeftSemimodule R m ℓm → RawLeftSemimodule m′ ℓm′ → RawLeftSemimodule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawLeftModule      : RawLeftModule R m ℓm → RawLeftModule m′ ℓm′ → RawLeftModule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawRightSemimodule : RawRightSemimodule R m ℓm → RawRightSemimodule m′ ℓm′ → RawRightSemimodule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawRightModule     : RawRightModule R m ℓm → RawRightModule m′ ℓm′ → RawRightModule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawBisemimodule    : RawBisemimodule R m ℓm → RawBisemimodule m′ ℓm′ → RawBisemimodule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawBimodule        : RawBimodule R m ℓm → RawBimodule m′ ℓm′ → RawBimodule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawSemimodule      : RawSemimodule R m ℓm → RawSemimodule m′ ℓm′ → RawSemimodule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  rawModule          : RawModule R m ℓm → RawModule m′ ℓm′ → RawModule R (m ⊔ m′) (ℓm ⊔ ℓm′)
+  ```
+
+* In `Algebra.Module.Construct.TensorUnit`:
+  ```agda
+  rawLeftSemimodule  : RawLeftSemimodule _ c ℓ
+  rawLeftModule      : RawLeftModule _ c ℓ
+  rawRightSemimodule : RawRightSemimodule _ c ℓ
+  rawRightModule     : RawRightModule _ c ℓ
+  rawBisemimodule    : RawBisemimodule _ _ c ℓ
+  rawBimodule        : RawBimodule _ _ c ℓ
+  rawSemimodule      : RawSemimodule _ c ℓ
+  rawModule          : RawModule _ c ℓ
+  ```
+
+* In `Algebra.Module.Construct.Zero`:
+  ```agda
+  rawLeftSemimodule  : RawLeftSemimodule R c ℓ
+  rawLeftModule      : RawLeftModule R c ℓ
+  rawRightSemimodule : RawRightSemimodule R c ℓ
+  rawRightModule     : RawRightModule R c ℓ
+  rawBisemimodule    : RawBisemimodule R c ℓ
+  rawBimodule        : RawBimodule R c ℓ
+  rawSemimodule      : RawSemimodule R c ℓ
+  rawModule          : RawModule R c ℓ
+  ```
+
+* In `Algebra.Properties.Monoid.Mult`:
+  ```agda
+  ×-homo-0 : ∀ x → 0 × x ≈ 0#
+  ×-homo-1 : ∀ x → 1 × x ≈ x
+  ```
+
+* In `Algebra.Properties.Semiring.Mult`:
+  ```agda
+  ×-homo-0#     : ∀ x → 0 × x ≈ 0# * x
+  ×-homo-1#     : ∀ x → 1 × x ≈ 1# * x
+  idem-×-homo-* : (_*_ IdempotentOn x) → (m × x) * (n × x) ≈ (m ℕ.* n) × x
+  ```
 
 * In `Data.Fin.Properties`:
   ```agda
@@ -91,7 +173,42 @@ Additions to existing modules
   map : (Char → Char) → String → String
   ```
 
-* Added new definitions in `Relation.Binary.Definitions`:
+* In `Function.Bundles`, added `_⟶ₛ_` as a synonym for `Func` that can
+  be used infix.
+
+* Added new definitions in `Relation.Binary.Definitions`
+  ```
+  Stable _∼_ = ∀ x y → Nullary.Stable (x ∼ y)
+  Empty  _∼_ = ∀ {x y} → x ∼ y → ⊥
+  ```
+
+* Added new definitions in `Relation.Nullary`
+  ```
+  Recomputable    : Set _
+  WeaklyDecidable : Set _
+  ```
+
+* Added new definitions in `Relation.Unary`
+  ```
+  Stable          : Pred A ℓ → Set _
+  WeaklyDecidable : Pred A ℓ → Set _
+  ```
+
+* Added new proof in `Relation.Nullary.Decidable`:
   ```agda
-  Empty _∼_ = ∀ {x y} → x ∼ y → ⊥
+  ⌊⌋-map′ : (a? : Dec A) → ⌊ map′ t f a? ⌋ ≡ ⌊ a? ⌋
+  ```
+
+* Added module `Data.Vec.Functional.Relation.Binary.Permutation`:
+  ```agda
+  _↭_ : IRel (Vector A) _
+  ```
+
+* Added new file `Data.Vec.Functional.Relation.Binary.Permutation.Properties`:
+  ```agda
+  ↭-refl      : Reflexive (Vector A) _↭_
+  ↭-reflexive : xs ≡ ys → xs ↭ ys
+  ↭-sym       : Symmetric (Vector A) _↭_
+  ↭-trans     : Transitive (Vector A) _↭_
+>>>>>>> master
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -210,5 +210,4 @@ Additions to existing modules
   ↭-reflexive : xs ≡ ys → xs ↭ ys
   ↭-sym       : Symmetric (Vector A) _↭_
   ↭-trans     : Transitive (Vector A) _↭_
->>>>>>> master
   ```
--- a/src/Relation/Binary/Construct/Interior/Symmetric.agda
+++ b/src/Relation/Binary/Construct/Interior/Symmetric.agda
@@ -16,9 +16,7 @@ private
   variable
     a b c ℓ r s t : Level
     A : Set a
-    R : Rel A r
-    S : Rel A s
-    T : Rel A t
+    R S T : Rel A r
 
 ------------------------------------------------------------------------
 -- Definition
@@ -28,6 +26,7 @@ record SymInterior (R : Rel A ℓ) (x y : A) : Set ℓ where
   field
     lhs≤rhs : R x y
     rhs≤lhs : R y x
+
 open SymInterior public
 
 ------------------------------------------------------------------------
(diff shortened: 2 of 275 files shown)


jamesmckinna:
Pity about the category theoretic terminology trumping the more classical (in the theory of relational algebra/preorders) counterpart to 'closure', viz. 'interior'... with the guarantee that while a closure is the 'least thing containing the thing satisfying ...', the interior is the 'greatest thing contained in the thing satisfying...'

Does your development explicitly prove such a characterisation? If not, could you please add such a universal property?

laMudri:
I'm happy to change to `SymInterior` now I know the word. I'll also give the characterisation proofs a go.

jamesmckinna:
Oh: and (again, personal taste) truly perverse field names in the record. Must we? I think `sdloh` (trying to pronounce it) would try the patience of a saint... `snoc` it most certainly is *not*...

Elsewhere, the library standardises on `to`/`from`, or variations thereon (`xRy` and `yRx` perhaps?) but 'surely' not this!?

jamesmckinna:
How about:
* `holds` $\mapsto$ `LHS≤RHS`
* `op-holds` $\mapsto$ `RHS≤LHS`

being:
* unambiguous, moreover rendering the sense (in a way that `holds` does not? cf `proof` in #1948/#1919)
* symbolic/abstract (we intend a proset/partial order `_≤_` but are otherwise uncommitted; we intend two terms, but are otherwise uncommitted to anything other than their relative positions/occurrences wrt `_≤_`)
* standardised apart from any other field names (potentially) frequently occurring elsewhere

laMudri:
Okay, yeah, that seems like an improvement. I went for lowercase names, because that's the convention, and I think it slightly improves legibility. Lack of legibility seems like the biggest argument against these names.

CHANGELOG.md
```diff
@@ -1749,6 +1749,11 @@ New modules
   Function.Indexed.Bundles
   ```
 
+* Symmetric core of a binary relation
+  ```
+  Relation.Binary.Construct.SymmetricCore
```
jamesmckinna, line 1754: `Core` $\mapsto$ `Interior`... ;-)

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+-- Properties
+
+-- The symmetric interior is symmetric.
+symmetric : Symmetric (SymInterior R)
+symmetric (r , r′) = r′ , r
+
+-- The symmetric interior of R is greater than (or equal to) any other symmetric
+-- relation contained by R.
+unfold : Symmetric S → S ⇒ R → S ⇒ SymInterior R
+unfold sym f s = f s , f (sym s)
```
jamesmckinna, line 43: `unfold` is usually reserved for a definition/function/proof which witnesses the equational unfolding/rewrite property of a... function

.. as such, it seems sub-optimal here.

Why not `interior-greatest` or `sym$\subseteq$interior` or even `symmetric⇒interior`...?

jamesmckinna: Or even `interior⁺` (it's a 'constructor'/introduction rule along the axis of relational implication...)?

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+
+-- SymInterior preserves various properties.
+reflexive : Reflexive R → Reflexive (SymInterior R)
+reflexive refl = refl , refl
+
+transitive′ : Trans R S T → Trans S R T →
+  Trans (SymInterior R) (SymInterior S) (SymInterior T)
+transitive′ trans trans′ (r , r′) (s , s′) = trans r s , trans′ s′ r′
```
jamesmckinna, line 51: I'd be tempted to have the defn be called `trans′` ("\'" throughout, not "'" as the suffix, per library guidelines), so you'd need to change the names of the arguments: but `trs` and `tsr` would do here, and are reasonably mnemonic...

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+record Proset c ℓ : Set (suc (c ⊔ ℓ)) where
+  infix 4 _≤_
+  field
+    Carrier : Set c
+    _≤_ : Rel Carrier ℓ
+    isProset : IsProset _≤_
+
+IsProset⇒IsPartialOrder : ∀ {a ℓ} {A : Set a} {≤ : Rel A ℓ} →
+  IsProset ≤ → IsPartialOrder (SymInterior ≤) ≤
+IsProset⇒IsPartialOrder {≤ = ≤} isProset = record
```
jamesmckinna, line 76: As with other defns inthe `Construct` hierarchy, I think that this would conventionally simply be called `isPartialOrder`, being the proof (in lower-camel-case) of the corresponding `Structure` property (in upper-camel-case`, with the ambient `IsProset⇒` being given implicitly by module-level scoping...

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+transitive′ : Trans R S T → Trans S R T →
+  Trans (SymInterior R) (SymInterior S) (SymInterior T)
+transitive′ trans trans′ (r , r′) (s , s′) = trans r s , trans′ s′ r′
+
+transitive : Transitive R → Transitive (SymInterior R)
+transitive {R = R} tr = transitive′ {R = R} tr tr
+
+-- The symmetric interior of a strict relation is empty.
+Empty-SymInterior : Asymmetric R → Empty (SymInterior R)
+Empty-SymInterior asym (r , r′) = asym r r′
```
jamesmckinna, line 58: `asymmetric⇒empty`?

Also, I think it's not always straightforward as to whther you announce a proposition/predicate instance/relation instance as being empty, or whether you prove it implies/is contained in, the corresponding bottom element `Empty` of the ordering... and here I do confess I'm not clear either what policy the library has, and if it does, what names are recommended here...

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+
+-- A reflexive transitive relation _≤_ gives rise to a poset in which the
+-- equivalence relation is SymInterior _≤_.
+record IsProset {a ℓ} {A : Set a} (≤ : Rel A ℓ) : Set (a ⊔ ℓ) where
+  field
+    refl : Reflexive ≤
+    trans : Transitive ≤
```
jamesmckinna, line 65: Conventionally, given what you subsequent define as the bundled contents, better to declare the relation argument here also as `_≤_`?

... and even in L26-L30 above?

jamesmckinna:
Have left some comments, mostly about names. But otherwise, I welcome this addition as a very natural, naturally occurring construct. Thanks!

src/Relation/Binary/Construct/Interior/Symmetric.agda
laMudri: The name is inspired by [`fold` in `Relation.Binary.Construct.Closure.Symmetric`](http://agda.github.io/agda-stdlib/Relation.Binary.Construct.Closure.Symmetric.html#1365). Better name vs consistency.

src/Relation/Binary/Construct/Interior/Symmetric.agda
laMudri: I was remembering the convention from `Algebra.Structures`, which is different to that of `Relation.Binary.Structures`. I find `_≤_` more readable, to be honest, so I'll switch to it here. I don't want to use it in `SymInterior` because it might not be an order relation (not being reflexive or transitive in general).

src/Relation/Binary/Construct/Interior/Symmetric.agda
laMudri: This looks like another case of the module name necessarily being part of the name (because asymmetric relations are not necessarily empty). Somehow it doesn't feel like the right thing to do in this case, when it's just a single property.

As for `Empty`, this is a new definition in this PR, so there's no convention *to* use it. I hadn't thought of `_⇒ Never`, but I think it's definitionally equal anyway. `Never` also appears to be unused in the library. It feels similar to the difference between `P → ⊥` and `¬ P`.

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: Ah yes, an inductive/least  notion yields`fold` (rather than `elim`) while a conductive/greatest yields `unfold` (rather than `intro` etc...)...

... another round in the battle for hearts and minds between an Oxford/haskell lexicon, and a Stockholm/Göteborg/MLTT one...

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: Agree that `ProSet`-related definitions belong 'somewhere else', except that, as you say, best if they don't... for now ;-)

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: apologies re primes etc. !

jamesmckinna:
Modulo the possibility of revisiting these names again sometime in the future (and waiting for colleagues to comment on them in the meantime), I'd be happy to proceed with this.

src/Relation/Binary/Definitions.agda
```diff
@@ -229,6 +230,11 @@ Recomputable _∼_ = ∀ {x y} → .(x ∼ y) → x ∼ y
 Universal : REL A B ℓ → Set _
 Universal _∼_ = ∀ x y → x ∼ y
 
+-- Empty - no elements are related
+
+Empty : REL A B ℓ → Set _
+Empty _∼_ = ∀ {x y} → x ∼ y → ⊥
```
MatthewDaggitt, line 236: Missing from CHANGELOG.

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+    trans : Transitive _≤_
+
+record Proset c ℓ : Set (suc (c ⊔ ℓ)) where
+  infix 4 _≤_
+  field
+    Carrier : Set c
+    _≤_ : Rel Carrier ℓ
+    isProset : IsProset _≤_
+
+IsProset⇒IsPartialOrder : ∀ {a ℓ} {A : Set a} {≤ : Rel A ℓ} →
```
MatthewDaggitt, line 74: The names `IsProset` and `IsPartialOrder` are usually lower case. Furthermore, unless there are multiple different ways of proving that the resultant relation is a partial order, we usually just omit the LHS of the implication, i.e. just `isPartialOrder`. (Otherwise, the proofs above would be named e.g. `transitive=>transitive`)

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+transitive : Transitive R → Transitive (SymInterior R)
+transitive {R = R} tr = trans′ {R = R} tr tr
+
+-- The symmetric interior of a strict relation is empty.
+Empty-SymInterior : Asymmetric R → Empty (SymInterior R)
+Empty-SymInterior asym (r , r′) = asym r r′
+
+-- A reflexive transitive relation _≤_ gives rise to a poset in which the
+-- equivalence relation is SymInterior _≤_.
+record IsProset {a ℓ} {A : Set a} (_≤_ : Rel A ℓ) : Set (a ⊔ ℓ) where
```
MatthewDaggitt, line 62: I'm very dubious about introducing a new record here. Either this record should go in `Relation.Binary.Structures` or it shouldn't exist. I'm leaning towards the latter. It's only two arguments `Reflexive` and `Transitive` after all.

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+  field
+    Carrier : Set c
+    _≤_ : Rel Carrier ℓ
+    isProset : IsProset _≤_
+
+IsProset⇒IsPartialOrder : ∀ {a ℓ} {A : Set a} {≤ : Rel A ℓ} →
+  IsProset ≤ → IsPartialOrder (SymInterior ≤) ≤
+IsProset⇒IsPartialOrder {≤ = ≤} isProset = record
+  { isPreorder = record
+    { isEquivalence = record
```
MatthewDaggitt, line 78: We should probably split this proof out into `IsEquivalence (SymInterior ≤)`?

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Symmetric interior of a binary relation
```
MatthewDaggitt, line 4: Could we have a little bit more description about what an `interior` actually is?

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+-- SymInterior preserves various properties.
+reflexive : Reflexive R → Reflexive (SymInterior R)
+reflexive refl = refl , refl
+
+trans′ : Trans R S T → Trans S R T →
+  Trans (SymInterior R) (SymInterior S) (SymInterior T)
+trans′ trans-rs trans-sr (r , r′) (s , s′) = trans-rs r s , trans-sr s′ r′
+
+transitive : Transitive R → Transitive (SymInterior R)
+transitive {R = R} tr = trans′ {R = R} tr tr
```
MatthewDaggitt, line 54: Do this suggest that maybe `trans'` should have an explicit rather than implicit parameter for at least one of the relations?

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,91 @@
...
+-- The symmetric interior of R is greater than (or equal to) any other symmetric
+-- relation contained by R.
+unfold : Symmetric S → S ⇒ R → S ⇒ SymInterior R
+unfold sym f s = f s , f (sym s)
+
+-- SymInterior preserves various properties.
+reflexive : Reflexive R → Reflexive (SymInterior R)
+reflexive refl = refl , refl
+
+trans′ : Trans R S T → Trans S R T →
```
MatthewDaggitt, line 49: Why not call this `trans` instead of `trans'`?

src/Relation/Binary/Construct/Interior/Symmetric.agda
laMudri: It turns out that `R` doesn't need to be given explicitly here. I guess I left it in from a previous definition of `SymInterior`, which required more implicit arguments to be given.

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: Try (and edit as you see fit, @laMudri ): 

"Just as the `P`-closure of a relation `R` for some property `P` is the *least* relation `S` containing `R` such that `P S`, the `P`-*interior* is the *dual* notion, vs. the *greatest* relation `S` *contained in* `R` such that `P S`, for any suitable `P`. Here, we take `P` to be `Symmetric`."

Or some such...

For a citation, you could look at the treatment of 'intersection-closed' and 'union-closed' families of predicates/relations defined in [Paul M. Cohn's "Universal Algebra" (1965)](https://link.springer.com/book/10.1007/978-94-009-8399-1) and their application to defining closures and interiors, and inductive/coinductive definitions generally (see also Peter Aczel's article on inductive definitions in the Handbook of Mathematical Logic (1977) (ISBN: 9780720422856)). NB impredicative definitions! 

Being `Symmetric` is, IIRC, an example of a property which is both intersection- and union-closed, and hence has a corresponding closure and interior, as here.

src/Relation/Binary/Construct/Interior/Symmetric.agda
MatthewDaggitt: To be honest I think we should ditch the record entirely, and pass `Reflexive` and `Transitive` as arguments directly. I say this because a) that's what we do elsewhere in the library and b) I can find no hits on the internet when searching for variations of `Proset`.

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: On second thoughts, maybe I should (have) move(d) this comment to the issue, rather than pollute the specific PR addressing it? Will do that, and then you can cherry-pick from there...

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: cf. `Relation.Binary.Properties.Preorder.InducedEquivalence`.
I should love to see a downstream reconciliation of all these things.

src/Relation/Binary/Definitions.agda
MatthewDaggitt: Still missing

src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -0,0 +1,95 @@
...
+
+------------------------------------------------------------------------
+-- Properties
+
+-- A reflexive transitive relation _≤_ gives rise to a poset in which the
+-- equivalence relation is SymInterior _≤_.
+record IsProset {a ℓ} {A : Set a} (_≤_ : Rel A ℓ) : Set (a ⊔ ℓ) where
+  field
+    refl : Reflexive _≤_
+    trans : Transitive _≤_
```
MatthewDaggitt, line 41: Did you see my comments about the existence of `IsProset`/`Proset`?

src/Relation/Binary/Construct/Interior/Symmetric.agda
MatthewDaggitt: I strongly agree with @jamesmckinna on this one. None of the other results in this file (or similar files) have the name of the relation in the proof.

MatthewDaggitt (changes requested):

src/Relation/Binary/Construct/Interior/Symmetric.agda
laMudri: So, `IsProset` can't be moved to `Relation.Binary.Structures` because of the file-level module parameters (particularly `_≈_`). Perhaps a compromise is to expand it out in the definition of `Proset` and move that to `Relation.Binary.Bundles`. This way, the definition `poset` is retained.

MatthewDaggitt (approved):

MatthewDaggitt:
I've eliminated `Proset` as well, as I think if want to add those sort of structures in we need to have a convincing story for the whole hierarchy rather a single instance. 

@laMudri if that's okay I propose to merge in?

CHANGELOG.md
```diff
@@ -166,7 +173,42 @@ Additions to existing modules
...
+  ```
+
+* Added new file `Data.Vec.Functional.Relation.Binary.Permutation.Properties`:
   ```agda
-  Empty _∼_ = ∀ {x y} → x ∼ y → ⊥
+  ↭-refl      : Reflexive (Vector A) _↭_
+  ↭-reflexive : xs ≡ ys → xs ↭ ys
+  ↭-sym       : Symmetric (Vector A) _↭_
+  ↭-trans     : Transitive (Vector A) _↭_
+>>>>>>> master
```
jamesmckinna, line 213: Ugh! What happened here?

jamesmckinna (changes requested):
Happy to approve, subject to fixing up the `CHANGELOG` ;-)

laMudri:
Okay, yeah, modulo the issues James has just mentioned.

jamesmckinna (approved):
Tick!

Issue #2065, opened 2023-08-19
monadic `join`


jamesmckinna:
Apologies if I have missed something down the years, but why does `Effect.Monad.RawMonad` not also define monadic join
```agda
μ : F (F A) → F A
μ = _>>= id
```
(also: `join` or `μ` as preferred identifier for this?)

gallais:
Remember that the functor is not necessarily an endofunctor level-wise:
`(F : Set f → Set g)`
We can only have `join` if `f = g`.

jamesmckinna:
Oh dear... I see. And would `g ≤ f` work also?
But I suppose we could define `RawMonadWithJoin` for the special case...?

jamesmckinna:
Revised design at #2079... via a `module` parametrised on a suitable instance of `RawMonad` ...

jamesmckinna:
[Moved from the discussion of PR#2079]
> Do we need a whole new record or is it sufficient to have a module parametrised by a monad and deriving join from bind?

Good question! ... you're probably right, something along the lines of : in module `Effect.Monad`
```agda
module Join {F : Set f → Set f} {{M : RawMonad F}} where
  join : F (F A) → F A
  join = _>>= id where open RawMonad M
```
... and then for downstream `RawMonad` instances `monad`:
```agda
join = Join.join where instance _ = monad
```

PR #2079, opened 2023-09-01
Monadic join

jamesmckinna:
In the style of #2065, which this would close. 

UPDATED: @gallais 's suggestion leads to a better design, with a uniform naming scheme, and no pesky additional `record` definition...

[Moved questions about design to the issue as well]

### DIFF
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -189,9 +189,6 @@ module DiagonalBind where
   _>>=_ : Vec A n → (A → Vec B n) → Vec B n
   xs >>= f = diagonal (map f xs)
 
-  join : Vec (Vec A n) n → Vec A n
-  join = diagonal
-
 
 ------------------------------------------------------------------------
 -- Operations for reducing vectors
@@ -368,8 +365,3 @@ transpose (as ∷ ass) = replicate _∷_ ⊛ as ⊛ transpose ass
 -- not guaranteed.
 
 -- Version 2.0
-
-{-# WARNING_ON_USAGE DiagonalBind.join
-"Warning: DiagonalBind.join was deprecated in v2.0
-Please use Data.Vec.Effectful.join instead."
-#-}
--- a/src/Data/Vec/Base.agda
+++ b/src/Data/Vec/Base.agda
@@ -189,9 +189,6 @@ module DiagonalBind where
   _>>=_ : Vec A n → (A → Vec B n) → Vec B n
   xs >>= f = diagonal (map f xs)
 
-  join : Vec (Vec A n) n → Vec A n
-  join = _>>= id
-
 ------------------------------------------------------------------------
 -- Operations for reducing vectors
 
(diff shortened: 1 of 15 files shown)


gallais:
Do we need a whole new record or is it sufficient to have a module parametrised by a monad and deriving join from bind?

jamesmckinna:
Good question! Wished you'd asked it on the issue, or that I had been less impatient ;-)

jamesmckinna:
Blast! Too hasty with the last commit.

jamesmckinna:
Revised design as per #2079... via a `module` parametrised on a suitable instance of `RawMonad` ...

src/Data/Vec/Base.agda
```diff
@@ -358,3 +355,20 @@ last xs with initLast xs
...
+
+-- Version 2.0
+
+join : Vec (Vec A n) n → Vec A n
+join = _>>= id where open DiagonalBind
+
+{-# WARNING_ON_USAGE join
+"Warning: join was deprecated in v2.0
+Please use Data.Vec.Effectful.join instead."
+#-}
```
jamesmckinna, line 374: Is this deprecation necessary?
That is, was `join` only introduced in v2.0, in which case it is harmless (is it?) to simply move it?

jamesmckinna: Actually, it seems as though `join` was only introduced in commit `7b7230752e86ba4e4b311fb812ede7cc5818a5e8`, so we can avoid the deprecation... just as well, as once again, we are deprecating/trying to deprecate a name local to the module `DiagonalBind`, not at the top-level...

src/Data/Vec/Effectful/Transformer.agda
```diff
@@ -50,7 +48,7 @@ monad {f} {g} M = record
   ; _>>=_ = λ ma k → mkVecT $ do
                       a ← runVecT ma
                       bs ← mapM {a = f} (runVecT ∘′ k) a
-                      pure (Vec.DiagonalBind.join bs)
+                      pure (diagonal bs)
```
jamesmckinna, line 51: See issue #2080

src/Data/Vec/Base.agda
```diff
@@ -358,3 +356,12 @@ last xs with initLast xs
...
 transpose (as ∷ ass) = replicate _∷_ ⊛ as ⊛ transpose ass
+
+
+------------------------------------------------------------------------
+-- DEPRECATED
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.0
```
MatthewDaggitt, line 367: This deprecated section is empty? Was it supposed to contain the `join` that was removed above?

src/Effect/Monad.agda
```diff
@@ -58,7 +58,16 @@ record RawMonad (F : Set f → Set g) : Set (suc f ⊔ g) where
   unless : Bool → F ⊤ → F ⊤
   unless = when ∘′ not
 
+-- When level g=f, a join/μ operator is definable
+
+module Join {F : Set f → Set f} ⦃ M : RawMonad F ⦄ where
```
MatthewDaggitt, line 63: I'm a little bit hesitant about making this an instance argument. As the rest of the PR shows, the current monad hierarchy isn't exactly setup for using instances. 

If we are going to introduce instances, I'd rather have a concrete plan in mind for the whole hierarchy?

Could we make this explicit instead?

src/Data/Vec/Base.agda
jamesmckinna: Hmmm. It previously did do so, until I realised that that definition had only been introduced recently enough for it not to require deprecation... (see the discussion on the issues above). But then I lost my nerve, and left this section in, as a placeholder... 'just in case'. My belief is that it, too, could be removed. Should it be?

src/Effect/Monad.agda
jamesmckinna: Ah, I see... and I certainly do not have a concrete plan for `instance`s for the `Effect` hierarchy (and in any case, that was *not* the point of this PR). Will proceed as suggested.

src/Data/Vec/Base.agda
MatthewDaggitt: Yup, that's fine. If it was introduced but never released, it can be zapped without a warning.

PR #2082, opened 2023-09-10
`WellFounded` proofs for transitive closure

jamesmckinna:
This PR adds the following new proofs [UPDATED]:
- [x] `accessible⁻ : ∀ {x} → Acc _∼⁺_ x → Acc _∼_ x`
- [x] `wellFounded⁻ : WellFounded _∼⁺_ → WellFounded _∼_`

which are instances of the corresponding proofs for ` module ∼⊆∼⁺ = Subrelation {_<₂_ = _∼⁺_} [_]`, together with
- [x] `accessible : ∀ {x} → Acc _∼_ x → Acc _∼⁺_ x`

which then permits refactoring the old proof `Relation.Binary.Construct.Closure.Transitive.wellFounded`, to streamline it using `accessible` (and removing unnecessary detours via the shadow reimplementation `downwardsClosed` of `Induction.WellFounded.acc-inverse`).

~~Possible candidate deprecation opportunity: to deprecate the old name in favour of `wellFounded⁺`, for the sake of uniformity of the above naming scheme, which is, of course, up for discussion/debate (an alternative uniformity might be to call `accessible⁺` simply `accessible`, for example)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2933,14 +2933,6 @@ Other minor changes
   ```
   accessible⁻ : ∀ {x} → Acc _∼⁺_ x → Acc _∼_ x
   wellFounded⁻ : WellFounded _∼⁺_ → WellFounded _∼_
-  ```
-  as instances of the corresponding proofs in `Induction.WellFounded.Subrelation`,
-  together with a refactoring of the former proof
-  ```
-  wellFounded : WellFounded _∼_ → WellFounded _∼⁺_
-  ```
-  in terms of a new proof:
-  ```
   accessible : ∀ {x} → Acc _∼_ x → Acc _∼⁺_ x
   ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2929,6 +2929,18 @@ Other minor changes
                     f Preserves₂ ≈₁ ⟶ ≈₁ ⟶ ≈₂
   ```
 
+* Added new proofs to `Relation.Binary.Construct.Closure.Transitive`:
+  ```
+  accessible⁺ : ∀ {x} → Acc _∼_ x → Acc _∼⁺_ x
+  accessible⁻ : ∀ {x} → Acc _∼⁺_ x → Acc _∼_ x
+  wellFounded⁻ : WellFounded _∼⁺_ → WellFounded _∼_
+  ```
+  together with a refactoring of the former proof
+  ```
+  wellFounded : WellFounded _∼_ → WellFounded _∼⁺_
+  ```
+  along these lines. 
+
 * Added new operations in `Relation.Binary.PropositionalEquality.Properties`:
   ```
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

jamesmckinna:
Any thoughts on the names, @JacquesCarette ?

JacquesCarette:
I have yet to integrate the supposed meaning of superscript + and -, so right now, not really.

jamesmckinna:
Elsewhere in `stdlib`, the modifiers superscript + and - are intended to designate 'introduction' and 'elimination' of the concept at hand; here, I borrowed that cue to (try to) indicate 'superscript + passes from (a property of) a relation to (that of) its transitive closure', while 'superscript - passes from (a property of) the transitive closure of a relation to (that of) its underlying relation'... HTH!

jamesmckinna:
Note to self: (potential) merge conflict with #2083 whichever gets merged first...

src/Relation/Binary/Construct/Closure/Transitive.agda
```diff
@@ -69,17 +70,21 @@ module _ (_∼_ : Rel A ℓ) where
   transitive : Transitive _∼⁺_
   transitive = _++_
 
-  wellFounded : WellFounded _∼_ → WellFounded _∼⁺_
-  wellFounded wf = λ x → acc (accessible′ (wf x))
-    where
-    downwardsClosed : ∀ {x y} → Acc _∼⁺_ y → x ∼ y → Acc _∼⁺_ x
-    downwardsClosed (acc rec) x∼y = acc (λ z z∼x → rec z (z∼x ∷ʳ x∼y))
+  accessible⁻ : ∀ {x} → Acc _∼⁺_ x → Acc _∼_ x
+  accessible⁻ = accessible where open ∼⊆∼⁺
```
MatthewDaggitt, line 74: Maybe `∼⊆∼⁺.accessible` is both shorter and more clear what's going on? Same below?

CHANGELOG.md
```diff
@@ -2929,6 +2929,21 @@ Other minor changes
                     f Preserves₂ ≈₁ ⟶ ≈₁ ⟶ ≈₂
   ```
 
+* Added new proofs to `Relation.Binary.Construct.Closure.Transitive`:
+  ```
+  accessible⁻ : ∀ {x} → Acc _∼⁺_ x → Acc _∼_ x
+  wellFounded⁻ : WellFounded _∼⁺_ → WellFounded _∼_
+  ```
+  as instances of the corresponding proofs in `Induction.WellFounded.Subrelation`,
+  together with a refactoring of the former proof
```
MatthewDaggitt, line 2938: I don't think the refactoring of the proof needs to be mentioned, nor the implementation of the new proofs. Otherwise the CHANGELOG will become very long indeed!

MatthewDaggitt (commented):
Otherwise looks good :+1:

Issue #2083, opened 2023-09-10
Definition of `Induction.WellFounded.WfRec`


jamesmckinna:
Currently, the definition:
```agda
WfRec : Rel A r → ∀ {ℓ} → RecStruct A ℓ _
WfRec _<_ P x = ∀ y → y < x → P y
```
suffers from two (unfortunate) defects:
* the use of the name `RecStruct`, when this type is nothing more or less than that of `PredicateTransformer` (or: what you will...) which only really takes on the role of being a structure-for-recursion once a given such `T : PredicateTransformer`  is accompanied by a *proof* that `T P ⊆′ P` for suitable `P`...;
* the *explicit* quantification over `y` in `WfRec`, which leads, infectively, to a profusion of (almost-?)always-inferrable arguments when working with subsequent `Acc`essibility and `WellFounded`ness proofs. ADDRESSED by #2084 

The first of these is, indeed, unfortunate but ignorable.

The second is, frankly, a complete pain, but likewise (almost) completely avoidable. 

PROPOSAL: to make the *breaking* change for v2.0 that the definition be changed to:
```agda
WfRec _<_ P x = ∀ {y} → y < x → P y
```
together with all of its downstream consequences... including the in-progress #2077 #2082 etc.

jamesmckinna:
A separate issue exposed by this proposal: in `Data.Vec.Relation.Binary.Lex.Strict`, the following lemma proposed in #602 and added in #1648 and its immediate corollary
```agda
  xs≮[] : ∀ {n} (xs : Vec A n) → ¬ xs < []
  xs≮[] _ (base ())

  ¬[]<[] : ¬ [] < []
  ¬[]<[] = xs≮[] []
```
should, likewise, make the `xs` argument implicit.

JacquesCarette:
Both of these explicit proposals make sense to me.

MatthewDaggitt:
> the explicit quantification over y in WfRec, which leads, infectively, to a profusion of (almost-?)always-inferrable arguments when working with subsequent Accessibility and WellFoundedness proofs. FIXED by https://github.com/agda/agda-stdlib/pull/2084

I'm not sure I agree with this. What happens when `A` is a record type and `<` is a strict _partial_ order over it that only inspects some of the fields? Then you're not going to be able to infer the values of the missing fields right?

jamesmckinna:
[...]

PR #2085, opened 2023-09-10
making (more) arguments implicit: #602, #1648, #1672 revisited

jamesmckinna:
See the [discussion comment](https://github.com/agda/agda-stdlib/issues/2083#issuecomment-1712751837) on issue #2083. 

Clients of the earlier PRs (specifically: @mechvel ?) might note that this would be a breaking change for v2.0, but hopefully a 'harmless' one. 

Would that I had had the courage of my convictions in those earlier PRs!


gallais (approved):

src/Data/List/Relation/Binary/Lex/Strict.agda
```diff
@@ -45,11 +45,11 @@ module _ {a ℓ₁ ℓ₂} {A : Set a} where
       _≋_ = Pointwise _≈_
       _<_ = Lex-< _≈_ _≺_
 
-    xs≮[] : ∀ xs → ¬ xs < []
-    xs≮[] _ (base ())
+    xs≮[] : ∀ {xs} → ¬ xs < []
+    xs≮[] (base ())
```
MatthewDaggitt, line 49: Was this added in `v2.0`? If not, it probably needs a CHANGELOG entry?

jamesmckinna: So, I added the `List` version in #1648 and the `Vec` version in #1672 each of which made it into 1.7*, I think, so yes, I suppose so.

PR #2093, opened 2023-09-14
Adds `setoid` export to `Algebra.Bundles.SemiringWithoutOne` (possible solution to #1917)

jamesmckinna:
What I don't understand is why/how this solution is to be preferred over others (such as?), or whether this may break anything subsequently... nor how this problem came about in the first place... nor what, if any, `CHANGELOG` entry might be suitable to account for the fix.

### DIFF
--- a/src/Algebra/Bundles.agda
+++ b/src/Algebra/Bundles.agda
@@ -504,7 +504,7 @@ record SemiringWithoutOne c ℓ : Set (suc (c ⊔ ℓ)) where
 
   open NearSemiring nearSemiring public
     using
-    ( _≉_; +-rawMagma; +-magma; +-unitalMagma; +-semigroup
+    ( +-rawMagma; +-magma; +-unitalMagma; +-semigroup
     ; +-rawMonoid; +-monoid
     ; *-rawMagma; *-magma; *-semigroup
     ; rawNearSemiring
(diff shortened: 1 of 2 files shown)


src/Algebra/Bundles.agda
```diff
@@ -504,7 +504,7 @@ record SemiringWithoutOne c ℓ : Set (suc (c ⊔ ℓ)) where
 
   open NearSemiring nearSemiring public
     using
-    ( _≉_; +-rawMagma; +-magma; +-unitalMagma; +-semigroup
```
Taneb, line 507: Why are you dropping the reexport of `_≉_`?

jamesmckinna: Thanks for the speedy eyes: it's dropped because Agda complained about ambiguous imports, presumably arising from the `open Setoid setoid public` addition, which brings these things into scope already (I think?). Removing them made the typechecker happy, so... (see previous about my finding this corner of the library *very* arcane)... I thought that I should be happy too...

Taneb: OK, this looks correct, then.

Taneb (approved):

Taneb:
As an aside I've been wondering if there's a good way to either test that this sort of thing is done right, or rewrite the library so all the re-exports are there by construction.

jamesmckinna:
Well, hence my mention of #1917 on #1967... and my hope that you will publish your big diagram at some point, as the source for exploring your "good way to test"/"rewrite the library"... but I'm glad it's not just me who feels this kind of things is too difficult to reliably debug 'by hand'!

jamesmckinna:
`status:ready` modulo `CHANGELOG` of course...

PR #2095, opened 2023-09-15
[fixes #1214] Add negated ordering relation symbols systematically to `Relation.Binary.*`

### LINKED ISSUE #1214: Add `_≮_` and `_≰_` to bundles in the binary relation hierarchy.
In lib-1.3 I see  ``_≉_`` in Setoid,
but I do not find  ``≰, ≮``
(I expect they to be defined via ``¬_``).
I grepped for it in Relation.Binary,  Relation.Binary.Construct, and do not find it.
Probably it needs to be in Preorder or in PartialOrder.
?

jamesmckinna:
Choosing precedence level = 4...


jamesmckinna:
This was/is surprisingly subtle, and now I'm still not entirely clear whether all the (definitions of the) relations have been successfully propagated... so once ths is stable, a fresh pair of eyes would be *most* gratefully received.

jamesmckinna:
More incoming commits... the deprecation strategy for this PR made my head spin!

jamesmckinna:
There's one leftover puzzle: when `import`ing and re-exporting the various `Bundles`, and renaming their fields, we are nevertheless left with sub-bundles defined with *the un-renamed symbol(s)*. This appears to be because, while `module`s may have their fields arbitrarily renamed, and `record`s can have their `field` names renamed, it isn't possible to rename manifest fields of `record`s when creating new instances of the record type. or if it is, I don't know how to do it. 

So, properly speaking, there should be a deprecation warning every time a symbol is renamed in a module export for the associated manifest field which should now not be referred to. I'll try, if I can, to produce a MWE to illustrate, but for now, I will chalk it up to (very) 'advanced features', and will not pursue further!

src/Relation/Binary/Bundles.agda
```diff
@@ -107,7 +110,7 @@ record TotalPreorder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
   preorder = record { isPreorder = isPreorder }
 
   open Preorder preorder public
-    using (module Eq)
+    using (module Eq) renaming (_≁_ to _⋦_)
```
jamesmckinna, line 113: I think it may also be necessary to hide `_∼_` here, so that it is not  re-exported?

src/Relation/Binary/Bundles.agda
```diff
@@ -131,7 +134,7 @@ record Poset c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
     }
 
   open Preorder preorder public
-    using (module Eq)
+    using (module Eq) renaming (_≁_ to _≰_)
```
jamesmckinna, line 137: Ditto. here?

src/Relation/Binary/Bundles.agda
```diff
@@ -152,7 +155,7 @@ record DecPoset c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
     }
 
   open Poset poset public
-    using (preorder)
+    using (preorder; _≰_)
```
jamesmckinna, line 158: NB we can do nothing about the fact that `preorder` will have (redundant!) fields names not congruent with those renamed in `Poset`... is that a problem in practice? In theory?

src/Relation/Binary/Bundles.agda
```diff
@@ -231,7 +240,7 @@ record TotalOrder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
     }
 
   open Poset poset public
-    using (module Eq; preorder)
+    using (module Eq; preorder; _≰_)
```
jamesmckinna, line 243: Again: `hiding` anything here?

src/Relation/Binary/Bundles.agda
```diff
@@ -256,7 +265,7 @@ record DecTotalOrder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
     { isTotalOrder = isTotalOrder
     }
 
-  open TotalOrder totalOrder public using (poset; preorder)
+  open TotalOrder totalOrder public using (poset; preorder; _≰_)
```
jamesmckinna, line 268: Ditto.?

src/Relation/Binary/Bundles.agda
Taneb: I don't think you ever need both `using` and `hiding`?

jamesmckinna: Yes, you're right. I was scratching my head trying think how to suppress spurious manifest fields in a `record`. Maybe it's all ok after all, but I do worry that in `Preorder`screamed as substructures if some other ordering type, the fields if that record are going to have the wrong names wrt the ambient enclosing ordering. But no obvious bugs so far...???

src/Relation/Binary/Bundles.agda
jamesmckinna: Here's how I understand it:
In the substructure `preorder` of `Poset`  (each inside `DecPoset`), *even though the base relation* has been passed in with one symbol `_≤_`, this export of the substructure will expose its field name with the `Preorder` symbol `_≲_`, even though references to that symbol, *qua* field name, will nevertheless be definitionally indistinguishable from its associated binding `_≤_`... suggesting at the very least (?) that the reference `_≲_` should be deprecated? Or is everything *really*/'really' OK after all (not least because `Preorder`-specific operations may even expect a field called `_≲_``...)

jamesmckinna:
Fixed merge conflict.

jamesmckinna:
`status:ready` modulo `CHANGELOG` changes/merge conflicts which may arise from #2099 ...

jamesmckinna:
This would be the second refactoring of `Relation.Binary.Bundles` to systematically incorporate appropriately correlated notations for each of the negated/flipped versions of each of the 'order' relations, and as such, I think should be bumped up to v2.0, following the successful merger of #2099, and hopefully shortly, one also addressing #2098 (the final one!?)

src/Relation/Binary/Bundles.agda
MatthewDaggitt: I mean we're not opening `preorder` here so we don't get the `_≲_` relation added to the scope. If the user really wants to open the `preorder` bundle themselves, then yes, they will get `_≲_` exposed, but I don't see why anyone would do so, as opening `DecPoset` will bring everything into scope with the correct names.

So not a problem in practice I think :+1:

src/Relation/Binary/Properties/Poset.agda
```diff
@@ -62,15 +63,10 @@ open Poset ≥-poset public
 ------------------------------------------------------------------------
 -- Negated order
 
-infix 4 _≰_
-
-_≰_ : Rel A p₃
-x ≰ y = ¬ (x ≤ y)
-
-≰-respˡ-≈ : _≰_ Respectsˡ _≈_
+≰-respˡ-≈ : _≰A_ Respectsˡ _≈_
```
MatthewDaggitt, line 66: I'm getting a very weird rendering of the symbol here: `_≰A_`, in particular the `≰` symbol has an `A` after it? Weirdly it doesn't appear in `Bundles` though. Any ideas what is causing it?

src/Relation/Binary/Properties/TotalOrder.agda
```diff
@@ -13,7 +13,9 @@ open import Relation.Binary.Structures using (IsTotalOrder)
 module Relation.Binary.Properties.TotalOrder
   {t₁ t₂ t₃} (T : TotalOrder t₁ t₂ t₃) where
 
-open TotalOrder T
+-- issue #1214: locally to this module, we rename the relation _≰_, so
+-- that we can deprecate it here, yet re-export it elsewhere
+open TotalOrder T renaming (_≰_ to _≰A_)
```
MatthewDaggitt, line 18: Same problem here with the `A`. Why are we renaming it at all?

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
-* NB (issues #1214 #2098) the corresponding situation regarding the `flip`ped
-  relation symbols `_≥_`, `_>_` (and their negated versions!) has not (yet)
-  been addressed.
+* NB (issue #2098) the corresponding situation regarding the `flip`ped
+  relation symbols `_≥_`, `_>_` has not (yet) been addressed.
+
+### (Issue #1214) Reorganisation of the introduction of negated relation symbols under `Relation.Binary`
+
+* Previously, negated relation symbols `_≰_` (for `Poset`) and `_≮_` (`TotalOrder`)
+  were introduced in the corresponding `Relation.Binary.Properties` modules, for re-export.
```
MatthewDaggitt, line 844: Why are we not mentioning `_≮_` here?

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+
+* Now they are introduced as definitions in the corresponding `Relation.Binary.Bundles`,
+  together with, for uniformity's sake, an additional negated symbol `_⋦_` for `Preorder`.
+  NB `_≮_` is now introduced *earlier* in the hierarchy, at `StrictPartialOrder`, which
+  incurs some rather complicated deprecation gymnastics...
```
MatthewDaggitt, line 849: I don't think this NB is needed. The user doesn't care about "complicated deprecation gymnastics", and I don't there's anything special about `_≮_`

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+
+* Accordingly, in fact we now have `_≰_` (in `Poset`) as a renamed `public` re-export
+  from `_⋦_` in `Preorder`, with `TotalPreorder` inheriting `_⋦_` from `Preorder`.
```
MatthewDaggitt, line 852: I would omit this fact. It's a private implementation detail, not relevant to the user facing API.

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+
+* As knock-on changes, public re-exports of `Preorder`, `Poset`, `StrictPartialOrder`
+  in derived bundles also now need to re-export the various new symbols accordingly.
```
MatthewDaggitt, line 855: Again would omit. Implementation detail.

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+
+* Backwards compatibility has been maintained, with deprecated definitions in the
+  corresponding `Relation.Binary.Properties` modules, and the corresponding client
+  client module `import`s being adjusted accordingly.
+
+* Elsewhere under `Relation.Binary.Properties` etc. the use of *explicitly* negated
+  relation symbols have now been replaced by their definitionally equal counterparts
+  using the new symbols accordingly.
```
MatthewDaggitt, line 863: Again would omit.

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+
+* NB modules such as `Relation.Binary.Construct.NonStrictToStrict` which operate
+  only on the the underlying `Structures` such as `IsPartialOrder` etc., are unable
+  to make use of the new symbols.
+
+* NB (issue #2098) the corresponding situation regarding the `flip`ped relation
+  symbols `_≥_`, `_>_` (and their negated versions!) has not (yet) been addressed.
+  Ditto. the strict ordering relation `_<_` defined in `Relation.Binary.Properties.Poset`...
```
MatthewDaggitt, line 871: Not relevant I think. Best to leave out.

CHANGELOG.md
```diff
@@ -836,9 +835,40 @@ Non-backwards compatible changes
...
+  corresponding `Relation.Binary.Properties` modules, and the corresponding client
+  client module `import`s being adjusted accordingly.
+
+* Elsewhere under `Relation.Binary.Properties` etc. the use of *explicitly* negated
+  relation symbols have now been replaced by their definitionally equal counterparts
+  using the new symbols accordingly.
+
+* NB modules such as `Relation.Binary.Construct.NonStrictToStrict` which operate
+  only on the the underlying `Structures` such as `IsPartialOrder` etc., are unable
+  to make use of the new symbols.
```
MatthewDaggitt, line 867: Again, I don't think this is relevant. The API of these modules haven't changed in anyway.

CHANGELOG.md
```diff
@@ -3083,6 +3123,20 @@ Additions to existing modules
   prependVLams : List String → Term → Term
   ```
 
+* Added new definitions to `Relation.Binary.Bundles`:
```
MatthewDaggitt, line 3126: This entry isn't needed as the full details should be included in the main `breaking` change entry above. Anything that you think should be preserved in this entry should be moved to the above entry.

MatthewDaggitt:
Otherwise, looks good, thanks for pushing this through!

src/Relation/Binary/Properties/Poset.agda
jamesmckinna: The `A` is there to avoid a renaming conflict; it's a locally introduced symbol at the top of the module, renaming apart from an existing symbol. See my comment in the `CHANGELOG` about "deprecation gymnastics"... or maybe there's a better way to proceed that you know?

src/Relation/Binary/Properties/TotalOrder.agda
jamesmckinna: This module used to export this symbol `_≰_`. And this behaviour, likewise for `StrictPartialOrder` with `\<n`, was an anomalous special case, which this PR is now systematising. So... to maintain backwards compatibility but with deprecation to force users to rewrite their imports, this seemed the only way to achieve that end. Otherwise the symbol we eventually want to have ends up being deprecated. Or, as above: is there a better way to achieve this?

So library clients who try to import that symbol from here should be given a deprecation warning if they do, and directed to `import` it form the correct module (or even, not at all, as it should simply come for free from `Bundles`)

CHANGELOG.md
jamesmckinna: Do you mean `\<~n`? Because it was never used in the library before #2099 introduced it as the symbol for `Preorder`. But I can do so if you think it adds additional force?

CHANGELOG.md
jamesmckinna: Re: gymnastics, see my comment above. Because the symbol has moved, its deprecation can't simply be achived locally inside one module.
As for the user, the warning should tell them corretcly where to import form , but his note was additional reinforcement.
As for the author of the PR, I was trying to document as clearly as possible what was quite a complicated thought process while I was doing this, for the sake of subsequent reviewers ;-) and myself or anyone else trying to tackle #2098...

CHANGELOG.md
jamesmckinna: OK. But see above.

CHANGELOG.md
jamesmckinna: Well, it was really a forward pointer to a PR I had hoped to already have written. Now the above are useful notes for me 9or any one else) to help guide them through the next one. Happy to keep them private though.

CHANGELOG.md
jamesmckinna: OK

CHANGELOG.md
jamesmckinna: OK

jamesmckinna:
Global comment: it's clear my `CHANGELOG` style is unnecessarily verbose. Yet as a developer, it often feels hard to learn how/why things were done in a certain way in the past, in order to navigate into the future. Reading PR discussion threads is pretty exhausting, and the `CHANGELOG` does seem to be a shared maintained resource which documents not only what, but also why and how. If it is only supposed to support 'what', then where should the 'why' and 'how' be documented?

MatthewDaggitt:
So in general, in large developments (e.g. the Agda code base or in industry settings that I've worked in), these sort of comments generally live in the commit messages themselves. Therefore when you want to work out why something was defined that way, you can look at the commit history that touched that definition, and then read any messages left by the developer.

I have to confess I'm very bad at doing it myself, but I'm happy for commit messages to start including more detail!

src/Relation/Binary/Properties/Poset.agda
MatthewDaggitt: Ooh weird. Let me have a closer look at the renaming going on. Obviously didn't appreciate what was going on properly in the first review.

jamesmckinna: BTW, something else weird is going on re: renaming/deprecation, on #2099: which I'll re-open temporarily to document.

jamesmckinna:
Re: commit messages. Yes I can see that such an approach (I had no real prior experience of `git`-hosted projects before this, with apologies to my former PhD student Craig McLaughlin, who tried to school me in these matters) is the 'most localised' in both history/space for finding (possibly contingent) 'reasons why'/'methods how'. But as I don't think we necessarily, even in a single PR, have perfect 'linearised' history, such an approach is vulnerable to ... retconning, or any other reason (screw-ups on my part) which may mean that such fragments of history get lost...

I have instead been tending to record those things (to the extent tolerated by yourself) in `CHANGELOG`, as 'the record of how we got to where we are now' (Certainly, the 'Major changes' section seems to function in that way...?)

But good to have an alternative/authoritative view.

jamesmckinna:
Yuk. Don't quit understand how the merge conflict appeared, esp. after fast-forwarding on the remote branch. Will fix when next I have the correctly aligned brain for such things.

CHANGELOG.md
jamesmckinna: Just to check: I was confused by your original question, because I do mention `_≮_` here... so I (perhaps mistakenly) presumed you had meant to type an their symbol. Could you restate the question please, so that I can better see the drift of it? Sorry to be obtuse!

CHANGELOG.md
jamesmckinna: DONE

CHANGELOG.md
jamesmckinna: DONE

jamesmckinna:
OK, I think that's nearly/mostly there now. Any update on (your analysis/review of) the deprecation "gymnastics"?

JacquesCarette:
I am a big fan of having rationale information in easy to search places (like `CHANGELOG`) instead of buried in commit messages. [I have, at times, written `Design.md` files as an alternative.] Rationale is the true gold of software design and should be as visible as possible, in my opinion.

MatthewDaggitt:
> I am a big fan of having rationale information in easy to search places (like CHANGELOG) instead of buried in commit messages. [I have, at times, written Design.md files as an alternative.] Rationale is the true gold of software design and should be as visible as possible, in my opinion.

I think the CHANGELOG is fundamentally the wrong place. It's a user facing document designed to help people migrate their code as quickly as possible. Design considerations should only be mentioned there if they either a) help the user to understand why the change has been made or b) help them migrate their code.

The advantages of putting messages in the commits is that they don't bit rot in the same why as `Design.md` file. A commit message is explicitly localised in time in a way that a `Design.md` file is not.

MatthewDaggitt:
> OK, I think that's nearly/mostly there now. Any update on (your analysis/review of) the deprecation "gymnastics"?

So if we were going to go for your approach, the better method in my opinion would to use it qualified everywhere instead of renaming.

However, in this instance, I think the cure is worse than the disease. When you write the usual pattern
```
open Poset poset
open import Relation.Binary.Properties.Poset poset


test : a ≰ a
test = ?
```
you don't get a nice deprecation warning message, you get a hard ambiguous name *error*, which breaks the user code.

In this case, I don't think Agda's warning machinary is flexible enough. We need the ability introduced in one of the most recent release of GHC, to localise warnings so that they only fire when you use the name from a particular module, rather than simply when you use the name.

MatthewDaggitt:
To that end I think we actually get a better user experience, in the average case, by simply removing the definition from the `Properties` file. Have done so in the latest commit.

jamesmckinna:
Thanks for the careful thoughtful analysis, which definitely extends my understanding of some of the dynamics of import... or will, once I've understood it...

src/Relation/Binary/Bundles.agda
```diff
@@ -338,18 +357,3 @@ record ApartnessRelation c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) w
...
--- DEPRECATED
-------------------------------------------------------------------------
--- Please use the new names as continuing support for the old names is
--- not guaranteed.
-
--- Version 2.0
-
-{-# WARNING_ON_USAGE Preorder._∼_
-"Warning: Preorder._∼_ was deprecated in v2.0. Please use Preorder._≲_ instead. "
-#-}
```
jamesmckinna, line 355: Not sure I quite understand why this particular deprecation warning has been removed. Is this removal part of this PR, or a knock-on consequence of (a refactoring of) #2099 ?

jamesmckinna: Oh, OK, I see the warning has now moved to L105-L108. Fine!

JacquesCarette:
Sophisticated users might really want to know the rationale behind some changes. You are quite right to point out that a `Design.md` will inevitably bit-rot if it is supposed to be maintained by hand.

On this point (of where to put rationale information), let's just agree to disagree; and as you @MatthewDaggitt are 'principal maintainer', we'll go with your method.

MatthewDaggitt:
> Sophisticated users might really want to know the rationale behind some changes. 

I agree _for some changes_, in particular those that affect user code.  As I said above, this falls under my reason a) above:

>  a) help the user to understand why the change has been made

In that case, yes we should absolutely include the discussion in the CHANGELOG above. However, if we start including all design decisions for individual cases that have no effect on user code, the CHANGELOG is quickly going to become impenetrable to the average user. 

I don't know of any major public facing project that includes such information in their change logs. Can you imagine if e.g. Tensorflow did the same thing?

src/Relation/Binary/Bundles.agda
MatthewDaggitt: Ah sorry yes. I should have commented, and I guess add to the style guide, that deprecation warnings should be attached to the definition of the name. Ideally at the end of the file, but in the case of a record declaration at the end of the record.

The problem with putting it at the end, is that the warning wouldn't trigger if the deprecated definition was used between the declaration and the end of the file!

jamesmckinna:
@MatthewDaggitt I'm not sure if the \`CHANGELOG\` was/is eventually brought up-to-date to reflect your recent commits?

MatthewDaggitt:
Looking at it, I think the CHANGELOG entry is fine. No worries about running out of steam on it, I'll take a look at #2098.

Issue #2096, opened 2023-09-15
`Relation.Binary.Bundles.Preorder` declares the 'wrong' symbol for the ordering relation!!!


jamesmckinna:
`Preorder`s are rarely `Symmetric`, unless they arise already from instances of `Setoid`/`IsEquivalence`.

So... why in the name of all that is holy do we declare them with a field name `_∼_` which is (notationally/syntactically/formally) *symmetric*. 

Consequences: 
* the converse (treated in `Relation.Binary.Properties.Preorder`) can't have a 'flipped' notation, so we have to make do with the (semantically) `flip`ped definition of the converse relation;
* the (underlying relation of the) converse of a `Poset` is declared in `Relation.Binary.Properties.Poset`, when that could simply be a renaming of the converse from `Relation.Binary.Properties.Preorder`, were that to be able to have a notation (*qua* definition, or else as `syntax` declaration);
* my head exploded while working on #1214 / #2095 as a result...

Refactoring PROPOSAL: that we rename (and appropriately deprecate, groan) the infix symbol/definition `_∼_` to be `_≲_`, which becomes `_≤_` in `Poset`, etc... and whose `Relation.Binary.Properties.Preorder.InducedEquivalence` be given by @laMudri 's `Relation.Binary.Construct.Interior.Symmetric` from #2071 applied to `_≲_`...

Just sayin'.

laMudri:
I agree that `_∼_` is a bad name and plain old `_≤_` would be much better. Unfortunately, I think this is inevitably a breaking change because of all the `record { _∼_ = Leq }`-like code which exists. At least the error message you get out of that should be somewhat instructive.

Taneb:
I agree that it's a breaking change. And I think that means if this is something we want to do, we ought to do it _now_. We're not going to have another major version increment any time soon

jamesmckinna:
Two/three things to say to this:
* yes, as a major breaking change, I think it's important to do it sooner rather than later (thanks @Taneb !);
* I'm nearly done with the corresponding one for negated symbols, #2095 which exposed to me the larger problem concerning `flip`ped symbols as a whole; out of sheer rage last night I started work on it, but rolled back before I went completely berserk, but a follow-up PR in the spirit of my current one would be... very desirable I think ;-)
[...]

PR #2099, opened 2023-09-16
De-symmetrising `Relation.Binary.Bundles.Preorder._∼_`

jamesmckinna:
See #2096

### DIFF
--- a/src/Relation/Binary/Properties/Preorder.agda
+++ b/src/Relation/Binary/Properties/Preorder.agda
@@ -18,10 +18,14 @@ import Relation.Binary.Construct.Flip.EqAndOrd as EqAndOrd
 
 open Preorder P
 
-------------------------------------------------------------------------
--- The inverse relation is also a preorder.
+private
+  infix 4 _∼ᵒ_
+  _∼ᵒ_ = flip _∼_
 
-converse-isPreorder : IsPreorder _≈_ (flip _∼_)
+------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jamesmckinna:
First (non-trivial!) problem: `Effect.Partiality` makes use of a very 'tricky' construction which takes the converse of the ordering it intends to define, and the generic machinery being developed here... falls over. Grrr.

jamesmckinna:
That seems `status:ready` (for v2.0) now.

MatthewDaggitt:
I went to review this expecting to say that I thought this change was too big a breaking change for too little gain, but the PR and the reasoning in the CHANGELOG has convinced me.

jamesmckinna:
OK, I don't want to revert this PR, but I do want to add the following comment which may be a (more serious) issue:
in `Relation.Unary.Closure.Preorder`, we have the following:
```agda
module Relation.Unary.Closure.Preorder {a r e} (P : Preorder a e r) where

open Preorder P
open import Relation.Unary using (Pred)

-- Specialising the results proven generically in `Base`.
import Relation.Unary.Closure.Base _∼_ as Base
```
which triggers *no* deprecation warning for the use of `_~_`. 
So I tried localising the warning to the exact place where the deprecated symbol/definition is introduced; still no warning. 

@MatthewDaggitt any ideas?

JacquesCarette:
Thanks for the review request - sorry I couldn't get to it in a timely manner. Very nice change.

Issue #2057, opened 2023-08-13
`Data.Vec.Base.splitAt`, `take` and `drop` green slime


JacquesCarette:
The "improved" implementation of PR #2056 still is flawed. For ease of comparison, it is
```agda
splitAt : ∀ m {n} (xs : Vec A (m + n)) →
          ∃₂ λ (ys : Vec A m) (zs : Vec A n) → xs ≡ ys ++ zs
splitAt zero    xs                = ([] , xs , refl)
splitAt (suc m) (x ∷ xs)          with splitAt m xs
splitAt (suc m) (x ∷ .(ys ++ zs)) | (ys , zs , refl) =
  ((x ∷ ys) , zs , refl)
```

But this says: if you give me (`splitAt`) a `Vec` `xs` that you built in such a way that you know that its length is of the form `m + n`, then I can split it into two pieces which will concatenate to the original, with the pieces being of the 'right' length.

It's pretty much sure that the vector that you want to split wasn't actually built that way! To me, a better interface (and code) should be:
```agda
splitAt : ∀ m {n o} → (o≡m+n : o ≡ m + n) → (xs : Vec A o) →
          ∃₂ λ (ys : Vec A m) (zs : Vec A n) → cast o≡m+n xs ≡ ys ++ zs
splitAt zero    p xs                = ([] , cast p xs , refl)
splitAt (suc m) p (x ∷ xs)          with splitAt m (suc-injective p) xs
splitAt (suc m) p (x ∷ xs) | (ys , zs , p′) = ((x ∷ ys) , zs , cong (x ∷_) p′)
```

But that is quite a large interface change. It does get rid of the green slime. [And I'm annoyed there's a need for `cast` but that might be inevitable.]

Taneb:
It might be more consistent with elsewhere to express the equalities the other way around, so we transform the more complicated expression into the simpler one (and expect the equality that way around for the addition equality)

gallais:
I don't think it's appropriate to call it 'green slime': it's a view so its whole purpose
is to invert a computation.

The proposed changes give you back a `cast`ed vector and so you will need additional
manipulations using a `cast-refl` lemma to 'unappend' a vector. Additionally that new
behaviour can already be obtained by passing the `cast`ed vector to the original `splitAt`
so I think this is a strict regression.

JacquesCarette:
[...]

PR #2101, opened 2023-09-19
Refactor some lookup and truncation lemmas

jamesmckinna:
~~Most of the way there, but I would like to introduce a definition of~~
Refactored to take advantage of:
```agda
truncate : (m≤n : m ≤ n) (xs : Vec A n) → Vec A m
```
and add: ~~a version of (itself refactored by this PR... ;-))~~
```agda
lookup-take-inject≤ : lookup (take m xs) i ≡ lookup xs (Fin.inject≤ i (m≤m+n m n))
```
~~which then makes use of it. But currently in `cast` hell... cf. #2057~~
UPDATED: Ugh! RTFL... The follies of ignorance... and there being rather scant properties of the function `truncate` already defined in `Data.Vec.Base`. 

Moving to draft status while I sort out the signal from the noise... NOW DONE. And with no plans to add to this PR further.

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -547,11 +547,20 @@ inject≤-idempotent {_} {suc n} {suc o} zero    _   _   _ = refl
 inject≤-idempotent {_} {suc n} {suc o} (suc i) (s≤s m≤n) (s≤s n≤o) (s≤s m≤o) =
   cong suc (inject≤-idempotent i m≤n n≤o m≤o)
 
+inject≤-trans : ∀ (i : Fin m) (m≤n : m ℕ.≤ n) (n≤o : n ℕ.≤ o) →
+                inject≤ (inject≤ i m≤n) n≤o ≡ inject≤ i (ℕₚ.≤-trans m≤n n≤o)
+inject≤-trans i m≤n n≤o = inject≤-idempotent i m≤n n≤o _             
+
 inject≤-injective : ∀ (m≤n m≤n′ : m ℕ.≤ n) i j →
                     inject≤ i m≤n ≡ inject≤ j m≤n′ → i ≡ j
 inject≤-injective (s≤s p) (s≤s q) zero    zero    eq = refl
 inject≤-injective (s≤s p) (s≤s q) (suc i) (suc j) eq =
   cong suc (inject≤-injective p q i j (suc-injective eq))
+
+inject≤-irrelevant : ∀ (m≤n m≤n′ : m ℕ.≤ n) i →
+                    inject≤ i m≤n ≡ inject≤ i m≤n′
+inject≤-irrelevant (s≤s p) (s≤s q) zero    = refl
+inject≤-irrelevant (s≤s p) (s≤s q) (suc i) = cong suc (inject≤-irrelevant p q i)
 
 ------------------------------------------------------------------------
 -- pred
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -549,7 +549,7 @@ inject≤-idempotent {_} {suc n} {suc o} (suc i) (s≤s m≤n) (s≤s n≤o) (s�
 
 inject≤-trans : ∀ (i : Fin m) (m≤n : m ℕ.≤ n) (n≤o : n ℕ.≤ o) →
                 inject≤ (inject≤ i m≤n) n≤o ≡ inject≤ i (ℕₚ.≤-trans m≤n n≤o)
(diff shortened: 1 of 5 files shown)


jamesmckinna:
Ugh! I've run out of steam with this one, but it's a start.

shhyou:
(FWIW, I have `cast-sym` as `CastReasoning.≈-sym` in #2067)

jamesmckinna:
I introduced `cast-sym` thinking that I could use it to unfold `take≤`, but as it turned out, I couldn't push it all the way to the end. So it's (probably) removable. In any case, with all the PRs flying around at the moment, a period of sober reflection and post hoc rationalisation is probably in order...

src/Data/Vec/Properties.agda
```diff
@@ -348,6 +365,13 @@ cast-is-id eq (x ∷ xs) = cong (x ∷_) (cast-is-id (suc-injective eq) xs)
 subst-is-cast : (eq : m ≡ n) (xs : Vec A m) → subst (Vec A) eq xs ≡ cast eq xs
 subst-is-cast refl xs = sym (cast-is-id refl xs)
 
+cast-sym : (eq : m ≡ n) {xs : Vec A m} {ys : Vec A n} →
```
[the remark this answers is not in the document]
jamesmckinna: Re: irrelevance. Thanks for the hint! Commit incoming.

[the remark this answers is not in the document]
jamesmckinna: Re: orientation of `cast-sym`... also a good suggestion. Thanks! 
Haven't tried your alternative (and clever!) proof strategy (yet)...

src/Data/Fin/Properties.agda
```diff
@@ -547,12 +547,21 @@ inject≤-idempotent {_} {suc n} {suc o} zero    _   _   _ = refl
...
 inject≤-injective : ∀ (m≤n m≤n′ : m ℕ.≤ n) i j →
                     inject≤ i m≤n ≡ inject≤ j m≤n′ → i ≡ j
 inject≤-injective (s≤s p) (s≤s q) zero    zero    eq = refl
 inject≤-injective (s≤s p) (s≤s q) (suc i) (suc j) eq =
   cong suc (inject≤-injective p q i j (suc-injective eq))
 
+inject≤-irrelevant : ∀ (m≤n m≤n′ : m ℕ.≤ n) i →
+                    inject≤ i m≤n ≡ inject≤ i m≤n′
+inject≤-irrelevant (s≤s p) (s≤s q) zero    = refl
+inject≤-irrelevant (s≤s p) (s≤s q) (suc i) = cong suc (inject≤-irrelevant p q i)
```
MatthewDaggitt, line 563: Why are we proving this by induction? Can't we do something like `cong (inject≤ i) (≤-irrelevant p1 p2)`

https://github.com/agda/agda-stdlib/blob/85829575904d9d7afc7a5ecc2b6902365f9cd564/src/Data/Nat/Properties.agda#L175

src/Data/Vec/Properties.agda
```diff
@@ -122,6 +122,20 @@ truncate-trans : ∀ {p} (m≤n : m ≤ n) (n≤p : n ≤ p) (xs : Vec A p) →
 truncate-trans z≤n       n≤p       xs = refl
 truncate-trans (s≤s m≤n) (s≤s n≤p) (x ∷ xs) = cong (x ∷_) (truncate-trans m≤n n≤p xs)
 
+truncate-irrelevant : (m≤n₁ m≤n₂ : m ≤ n) → truncate {A = A} m≤n₁ ≗ truncate m≤n₂
+truncate-irrelevant {m = zero}  m≤n₁ m≤n₂ _        = refl
+truncate-irrelevant (s≤s m≤n₁) (s≤s m≤n₂) (x ∷ xs) = cong (x ∷_) (truncate-irrelevant m≤n₁ m≤n₂ xs)
```
MatthewDaggitt, line 127: Again shouldn't need to recurse here.

MatthewDaggitt (commented):
Otherwise looks good.

src/Data/Fin/Properties.agda
jamesmckinna: Of course! Oh dear... I think I probably need to go back to school about the whole issue of `Irrelevant` propositions in the library, and the relation between *proving* irrelevance, and the typechecker being able to *observe* it.

src/Data/Vec/Properties.agda
jamesmckinna: Indeed...

jamesmckinna:
That merge conflict resolution should do the trick?

Issue #1580, opened 2021-08-19
Deprecate `inspect` function in favour of new `with ... in` construct


MatthewDaggitt:
@gallais added the inspect idiom as a first class language construct in Agda 2.6.2. We should consider using it in the library to simplify our code, and possibly deprecating the old `inspect` function to encourage other people to migrate as well.

jamesmckinna:
Suggest closing this now, after the merge of #1930, notwithstanding the issues of strictness of `inspect` vs. laziness of `with ... in ...` raised by 
[agda#5833](https://github.com/agda/agda/issues/5833)

jamesmckinna:
The saga continues,unfortunately: https://github.com/agda/agda/issues/6841

jamesmckinna:
For v2.0: suggest removing the deprecation warning?

MatthewDaggitt:
Opened a PR undeprecating it, and have moved this to `Agda-future`

ncfavier:
With https://github.com/agda/agda/issues/5833 and https://github.com/agda/agda/issues/6841 both closed, maybe it's time to revisit this?

jamesmckinna:
Thanks to @ncfavier for the heroic work on https://github.com/agda/agda/pull/8318 which seems, at first sight, just to 'do The Right Thing', so yes, ... except that it think it would at present have to be on the `experimental` branch, because, IIUC, the new behaviour will only ship with v2.9.0, and our current release point v2.4 is (morally) tied to v2.8.0.

So tagging this for the v3.0 milestone (anticipating a v2.9.0 release before we even get near v3.0 for `stdlib` ;-)), and getting there via the `experimental` branch would seem the way to go?

jamesmckinna:
Independently of any deprecation of `Reveal_·_is_`/`inspect`, I nevertheless would welcome that the definition be retained in some form (but with a better name!) in order to capture the *graph* relation associated with a given function `f` (dependently-typed or not), namely as `λ x y → y ≈ f x`, even if only restricted, in the dependently-typed case, to `λ x y → y ≡ f x`...?

jamesmckinna:
@ncfavier what's the intended release date for v2.9.0? Before or after the next AIM in April?

ncfavier:
I'm not a maintainer but the [2.9.0 milestone](https://github.com/agda/agda/milestone/52) is "due by March 1, 2026".

JacquesCarette:
[...]

PR #2107, opened 2023-09-27
Un-deprecate inspect idiom

MatthewDaggitt:
As per https://github.com/agda/agda-stdlib/issues/1580, we're not deprecating it for v2.0 as `with ... in` is not a perfect replacement.

### DIFF
--- a/src/Relation/Binary/PropositionalEquality.agda
+++ b/src/Relation/Binary/PropositionalEquality.agda
@@ -104,15 +104,17 @@ module _ (_≟_ : DecidableEquality A) {x y : A} where
   ≢-≟-identity = dec-no (x ≟ y)
 
 
-
-
 ------------------------------------------------------------------------
--- DEPRECATED NAMES
-------------------------------------------------------------------------
--- Please use the new names as continuing support for the old names is
--- not guaranteed.
+-- Inspect
 
--- Version 2.0
+-- Inspect can be used when you want to pattern match on the result r
+-- of some expression e, and you also need to "remember" that r ≡ e.
+
+-- See README.Inspect for an explanation of how/why to use this.
+
+-- Normally (but not always) the new `with ... in` syntax described at
+-- https://agda.readthedocs.io/en/v2.6.3/language/with-abstraction.html#with-abstraction-equality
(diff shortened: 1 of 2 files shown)


jamesmckinna:
But sad all the same. I hope that a longer term discussion with @gallais and others can mail this one down, but it dies seem a wicked problem.

src/Relation/Binary/PropositionalEquality.agda
```diff
@@ -104,15 +104,17 @@ module _ (_≟_ : DecidableEquality A) {x y : A} where
   ≢-≟-identity = dec-no (x ≟ y)
 
 
+------------------------------------------------------------------------
+-- Inspect
 
+-- Inspect can be used when you want to pattern match on the result r
```
jamesmckinna, line 110: 'result' or 'value'?
(I realise this is the previous text, but I wonder if this is more precise usage?)

src/Relation/Binary/PropositionalEquality.agda
```diff
@@ -104,15 +104,17 @@ module _ (_≟_ : DecidableEquality A) {x y : A} where
...
+-- of some expression e, and you also need to "remember" that r ≡ e.
```
jamesmckinna, line 111: One of the (many? subtle!)  peculiarities of `inspect` us that it does rely of `e` being of the form `f a` and requires you to identify those pieces separately. That is, there seems no way to fold together `with e` with an invocation of `inspect e` (the latter makes no sense *per se*) in a single statement form. 

Unless one perhaps writes directly 
```agda
with inspect {.(f a)} f a
... | eq = ...
```
but even then the explicit dot-patterned term is not bound to a(n independent) variable subsequently available to the matcher. 

So even the pragmatic of this notation are both annoying, and hard to explain well...

jamesmckinna:
Suggest that we merge as is, but my 'review' comments are some attempt for me (and perhaps me alone, else they belong in the ussue instead) to try to pin down what I continue to find unsatisfactory about all this :-( and for which `with x <- e in eq` seemed so much more satisfactory, modulo the still-breaking regressions that @omelkonian and others have laid bare.

Issue #453, opened 2018-09-05
Misnomer: `Relation.Binary.Core.Total`


WolframKahl:
I find that the current `Total` for binary relations is a misnomer: Currently there is:
```agde
Total : ∀ {a ℓ} {A : Set a} → Rel A ℓ → Set _
Total _∼_ = ∀ x y → (x ∼ y) ⊎ (y ∼ x)
```
But a binary relation `R` is normally called total iff `Reflexive (R relCompose (converse R))`,
so ripping this name out of the context of a poor naming of linear orders is, in my opinion, very unfortunate.

HuStmpHrrr:
Aren't they the same relation though? I do see this definition in other reference: https://ncatlab.org/nlab/show/total+relation

WolframKahl:
@HuStmpHrrr , ncatlab is a somewhat surprising source for relation-algebraic terminology.
A more common starting point, which I am also using in [RATH-Agda](http://RelMiCS.McMaster.ca/RATHAgda/), is the following textbook:
```bibtex
@Book{Schmidt-Stroehlein-1993,
  author = {Gunther Schmidt and Thomas Str{\"o}hlein},
  title = {Relations and Graphs, Discrete Mathematics for Computer Scientists},
  year = 1993,
  series = {EATCS-Mono\-graphs on Theoret.\null{} Comput.\null{} Sci.},
  publisher = Springer,
  DOI = {10.1007/978-3-642-77968-8},
  abstract = {Relational methods can be found at various places in
               computer science, notably in data base theory,
               relational semantics of concurrency, relational type
               theory, analysis of rewriting systems, and modern
               programming language design. In addition, they appear in
               algorithms analysis and in the bulk of discrete
               mathematics taught to computer scientists. This book
               devoted to the background of these methods. It is the
               first to explain how to use relational and graph-
               theoretic methods systematically in computer science.
               The powerful calculus of relation algebra is developed
               with respect to applications to a diverse range of
               problem areas. Results are first motivated by practical
               examples, often visualized by both Boolean 0-1-matrices
               and graphs, and then derived algebraically.}
  ISBN = {3-540-56254-0, 0-387-56254-0}
}
```

MatthewDaggitt:
Hypothetically what would you suggest as a replacement name?

Unfortunately this can't be deprecated and is a reasonably large breaking change as it would require renaming the record field in `TotalOrder`.

JacquesCarette:
[...]

PR #2111, opened 2023-09-28
Ob #453: added `Dense` relations and `DenseLinearOrder`

jamesmckinna:
Tackles the last point of issue #453 : NB extends `StrictTotalOrder` to `DenseLinearOrder`, so lays the ground work for subsequent renaming/refactoring in the spirit of #453... and proposes it be part of `v2.0` because of that issue.

~~OUTSTANDING problem: trying to open the sub-bundle built out of the `IsStrictTotalOrder` field causes a name clash, for some reason. Don't really understand why, so `Algebra.*` experts please advise!~~ FIXED: thanks @MatthewDaggitt !

### DIFF
--- a/src/Data/Rational/Properties.agda
+++ b/src/Data/Rational/Properties.agda
@@ -51,11 +51,11 @@ open import Function.Definitions using (Injective)
 open import Level using (0ℓ)
 open import Relation.Binary.Core using (_⇒_; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
 open import Relation.Binary.Bundles
-  using (Setoid; DecSetoid; TotalPreorder; DecTotalOrder; StrictPartialOrder; StrictTotalOrder)
+  using (Setoid; DecSetoid; TotalPreorder; DecTotalOrder; StrictPartialOrder; StrictTotalOrder; DenseLinearOrder)
 open import Relation.Binary.Structures
-  using (IsPreorder; IsTotalOrder; IsTotalPreorder; IsPartialOrder; IsDecTotalOrder; IsStrictPartialOrder; IsStrictTotalOrder)
+  using (IsPreorder; IsTotalOrder; IsTotalPreorder; IsPartialOrder; IsDecTotalOrder; IsStrictPartialOrder; IsStrictTotalOrder; IsDenseLinearOrder)
 open import Relation.Binary.Definitions
-  using (DecidableEquality; Reflexive; Transitive; Antisymmetric; Total; Decidable; Irrelevant; Irreflexive; Asymmetric; Trans; Trichotomous; tri<; tri>; tri≈; _Respectsʳ_; _Respectsˡ_; _Respects₂_)
+  using (DecidableEquality; Reflexive; Transitive; Antisymmetric; Total; Decidable; Irrelevant; Irreflexive; Asymmetric; Dense; Trans; Trichotomous; tri<; tri>; tri≈; _Respectsʳ_; _Respectsˡ_; _Respects₂_)
 open import Relation.Binary.PropositionalEquality
 open import Relation.Binary.Morphism.Structures
 import Relation.Binary.Morphism.OrderMonomorphism as OrderMonomorphisms
@@ -614,6 +614,15 @@ toℚᵘ-isOrderMonomorphism-< = record
 <-asym : Asymmetric _<_
 <-asym (*<* p<q) (*<* q<p) = ℤ.<-asym p<q q<p
 
+<-dense : Dense _<_
+<-dense {p} {q} p<q with m , p<ᵘm , m<ᵘq ← ℚᵘ.<-dense (toℚᵘ-mono-< p<q)
+        = fromℚᵘ m , p<m , m<q
+  where
+  m≃m : m ≃ᵘ toℚᵘ (fromℚᵘ m)
+  m≃m = ℚᵘ.≃-sym (toℚᵘ-fromℚᵘ m)
+  p<m = toℚᵘ-cancel-< (ℚᵘ.<-respʳ-≃ m≃m p<ᵘm)
+  m<q = toℚᵘ-cancel-< (ℚᵘ.<-respˡ-≃ m≃m m<ᵘq)
+
 <-≤-trans : Trans _<_ _≤_ _<_
 <-≤-trans {p} {q} {r} (*<* p<q) (*≤* q≤r) = *<*
   (ℤ.*-cancelʳ-<-nonNeg _ (begin-strict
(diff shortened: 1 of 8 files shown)


jamesmckinna:
Errors arising from #2112 . Will fix locally for this case, but general issue remains.

Taneb:
As part of this PR can you show that Data.Rational._<_ satisfies this?

jamesmckinna:
Indeed: showing that the rationals are DLO (without endpoints!) is an important milestone in itself. Will try to do so, while trying not to either take on too much additional development just before v2.0, nor drive my brain further towards/over the edge.

jamesmckinna:
Grrr... typo in last commit message: should be "fixed #2112 for this PR"...

src/Relation/Binary/Bundles.agda
```diff
@@ -299,6 +299,23 @@ record StrictTotalOrder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) wh
...
+    isDenseLinearOrder : IsDenseLinearOrder _≈_ _<_
+
+  open IsDenseLinearOrder isDenseLinearOrder public
+
+  strictTotalOrder : StrictTotalOrder c ℓ₁ ℓ₂
+  strictTotalOrder = record
+    { isStrictTotalOrder = isStrictTotalOrder
+    }
+
+  --open StrictTotalOrder strictTotalOrder public
```
MatthewDaggitt, line 317: So to resolve this problem, you have to open first thing A (either the structure or the bundle), and then open the other one exporting only the things missing in A. 

Here would recommend opening `StrictTotalOrder` as you'll get all the bundles and structures, and only have to import `dense` from `IsStrictTotalOrder`, i.e. `open IsDenseLinearOrder isDenseLinearOrder using (dense)`

MatthewDaggitt (commented):
Otherwise looks good to me.

src/Relation/Binary/Bundles.agda
jamesmckinna: ... and , in fact, `using (isStrictTotalOrder; dense)` in order to be able to create the `StrictTotalOrder` sub-bundle in the first place.

src/Data/Rational/Properties.agda
```diff
@@ -614,6 +614,15 @@ toℚᵘ-isOrderMonomorphism-< = record
 <-asym : Asymmetric _<_
 <-asym (*<* p<q) (*<* q<p) = ℤ.<-asym p<q q<p
 
+<-dense : Dense _<_
+<-dense {p} {q} p<q with m , p<ᵘm , m<ᵘq ← ℚᵘ.<-dense (toℚᵘ-mono-< p<q)
```
MatthewDaggitt, line 618: Should this not just be a let statement? Do we really need `with`?

src/Data/Rational/Properties.agda
```diff
@@ -614,6 +614,15 @@ toℚᵘ-isOrderMonomorphism-< = record
...
+        = fromℚᵘ m , p<m , m<q
+  where
+  m≃m : m ≃ᵘ toℚᵘ (fromℚᵘ m)
+  m≃m = ℚᵘ.≃-sym (toℚᵘ-fromℚᵘ m)
```
MatthewDaggitt, line 622: Minor but can we have an empty line between proofs.

MatthewDaggitt (commented):
Otherwise looks good to me.

src/Data/Rational/Properties.agda
jamesmckinna: You're absolutely correct (and I've been a demon for insisting on it elsewhere, ... :facepalm:)

src/Data/Rational/Properties.agda
jamesmckinna: A-ha! not a style recommendation that I had (yet) internalised.

src/Data/Rational/Properties.agda
jamesmckinna: Blast! No, actually: If I (try to) use `let`, the the let-bound names don't scope correctly over the subsequent `where` clauses... which is very annoying. Or am I missing something?

Ideally, I would make a simultaneous definition in the `where` block` and destruct it in place, as for example
```agda
<-dense {p} {q} p<q = fromℚᵘ m , p<m , m<q
  where
  d : ∃[ m ] p < m × m < q
  d@(m , p<ᵘm , m<ᵘq) = ℚᵘ.<-dense (toℚᵘ-mono-< p<q)
  
  m≃m : m ≃ᵘ toℚᵘ (fromℚᵘ m)
  m≃m = ℚᵘ.≃-sym (toℚᵘ-fromℚᵘ m)
  
  p<m = toℚᵘ-cancel-< (ℚᵘ.<-respʳ-≃ m≃m p<ᵘm)
  
  m<q = toℚᵘ-cancel-< (ℚᵘ.<-respˡ-≃ m≃m m<ᵘq)
```
but that doesn't seem to be legal, either. Using/Having to use projections would be an abomination.

jamesmckinna: If you know a better way, I'd love to see it!

jamesmckinna (commented):
Must have hot wrong button. Not intended to be review comments; only intended to be replies to you!

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -412,6 +412,37 @@ drop-*<* (*<* pq<qp) = pq<qp
...
 
+<-dense : Dense _<_
+<-dense {p} {q} (*<* p<q) = mid , p<r , r<q
+  where
+  open ℤ.≤-Reasoning
+  mid : ℚᵘ
+  mid = mkℚᵘ ((↥ p) ℤ.+ (↥ q)) (pred ((↧ₙ p) ℕ.+ (↧ₙ q)))
+  p<r : p < mid
+  p<r = *<* (begin-strict
+    (↥ p) ℤ.* (↧ mid)
```
Taneb, line 423: I don't think parentheses are needed around `↥ p` and similar.

src/Data/Rational/Properties.agda
MatthewDaggitt: Hmm weird. You're right I can't see any good way of working around that. I agree that what you've written with the `where` block above would be the "correct" way of doing it in my opinion but I can't get that to work either. So I guess we just fall back on good old `let`:
```
<-dense : Dense _<_
<-dense {p} {q} p<q =
  let
    m , p<ᵘm , m<ᵘq = ℚᵘ.<-dense (toℚᵘ-mono-< p<q)

    m≃m : m ≃ᵘ toℚᵘ (fromℚᵘ m)
    m≃m = ℚᵘ.≃-sym (toℚᵘ-fromℚᵘ m)

    p<m : p < fromℚᵘ m
    p<m = toℚᵘ-cancel-< (ℚᵘ.<-respʳ-≃ m≃m p<ᵘm)

    m<q : fromℚᵘ m < q
    m<q = toℚᵘ-cancel-< (ℚᵘ.<-respˡ-≃ m≃m m<ᵘq)
  in fromℚᵘ m , p<m , m<q
```

MatthewDaggitt: I think this is a little clearer, so I've pushed the change.

Issue #1579, opened 2021-08-12
Inconsistent naming between `Data.(Nat/Integer).Properties`


MatthewDaggitt:
We have our left and rights switched round in `*-monoʳ-≤`, `*-monoʳ-≤-pos`. It would be kind of good if we had definitions for `Monotonic`/`LeftMonotonic`/`RightMonotonic` etc. in `Relation.Binary.Definitions` to help keep us on the straight and narrow.

MatthewDaggitt:
Also `Antimonotonic` etc.

jamesmckinna:
See also #2087 / #2089

jamesmckinna:
Regarding `Anti`, I have the (slight? vague?) feeling that as with `Cancellative`, it might be better to parametrise these properties on 'source' and 'target' relations... cf. `Preserves`?

jamesmckinna:
> Also `Antimonotonic` etc.

Are we sure we need this? Monotonicity is/should be usually stated wrt two *distinct* orderings, and these can then be varied as:
* same, same: 'usual' monotonicity within a given ordered structure
* same, flipped: antitone
* etc.

jamesmckinna:
See also #2532 for the wider issue of the (need to agree the) semantics of 'left' and 'right'...

jamesmckinna:
Aaaargh! The various definitions *do* exist in `Relation.Binary.Definitions`, but aren't then specialised in `Algebra.Definitions` (and perhaps that wouldn't make sense as such to do so...?).

JacquesCarette:
My guess why it's this way: history. Things were developed "by hand" and only slowly did patterns emerge. You could say that those patterns should be well-known - and indeed they are to the people whose educational path included enough of the right kind of Algebra. But that doesn't tend to be the same people as those who develop ITPs! [I've seen the same thing happen in CASes, but there it's mostly about not knowing enough analysis.]

jamesmckinna:
> [#2580](https://github.com/agda/agda-stdlib/pull/2580) shouldn't close this issue right?

No, you're correct, but it might do once that DRAFT PR is finished? 
In any case, will tweak the preamble to it to remove the `fixes` cue for the time being...

jamesmckinna:
Progress so far on #2580 suggests that at least the various `Data.X.Properties` module are *mutually* consistent, in naming lemmas vs. the 'opposite' handedness in the lemma statements... so we should only consider a `breaking` change *after* resolving the discussion on #2532 ?

Issue #1562, opened 2021-07-26
AlmostCancellative in Nat.Properties ?


mechvel:
lib-1.7 has a general notion AlmostLeft(Right)Cancellative,
and it is used in `*-cancelˡ-nonZero` in `IsCancellativeCommutativeSemiring`. 

On the other hand  `Nat.Properties` has the definitions of kind.
```
*-cancelʳ-≡ :  ∀ m n {o} → m * suc o ≡ n * suc o → m ≡ n  
```
It has sense to add there  ``*-cancelˡ-nonZero,  *-cancelʳ-nonZero``
having  types AlmostLeft(Right)Cancellative,
and to add the instances of `CancellativeCommutativeSemiring` for Nat and Integer.
?

Taneb:
I think this is a good idea. It should probably also be done for Integers and Rationals, if it hasn't already.

mechvel:
The matter is also in that  `*-cancel...nonZero`  does rely on `suc`.

jamesmckinna:
I think a useful reconciliation of this proposal with the instances arising for the concrete numerical types might be to systematically add
`NonZero num <-> num ≢ 0#` lemmas to their `Properties`, and work outwards from there? Eg towards also fixing #1436...?

Issue #1436, opened 2021-02-23
Standardise implicit arguments of cancellation properties.


negiizhao:
In `Algebra.Definitions`

```agda
LeftCancellative _•_ = ∀ x {y z} → (x • y) ≈ (x • z) → y ≈ z
RightCancellative _•_ = ∀ {x} y z → (y • x) ≈ (z • x) → y ≈ z
AlmostLeftCancellative e _•_ = ∀ {x} y z → ¬ x ≈ e → (x • y) ≈ (x • z) → y ≈ z
AlmostRightCancellative e _•_ = ∀ {x} y z → ¬ x ≈ e → (y • x) ≈ (z • x) → y ≈ z
```

In `Data.Rational.Unnormalised.Properties`
```agda
+-cancelˡ : ∀ {r p q} → r + p ≃ r + q → p ≃ q
+-cancelʳ : ∀ {r p q} → p + r ≃ q + r → p ≃ q
```

Perhaps we should make them uniform.

MatthewDaggitt:
Hey, thanks for the spot! Yup, so for starters the rational proofs should be using the `LeftCancellative` and `RightCancellative` types. 

As for the `AlmostX` and `X` definitions, the reason for the differences in the implicits is Agda's ability to infer them. `LeftCancellative` and `RightCancellative` were introduced early on in the library's life and the implicit arguments are slightly broken, see the code below. To fix this I think none of the arguments to those definitions should be implicit. The `Almost` versions were introduced very recently and I think have the right implicits (`x` is inferrable from the equality).

```agda
open import Data.Rational.Base
open import Algebra.Definitions
open import Relation.Binary.PropositionalEquality.Core

postulate
  rc : RightCancellative _≡_ _+_
  arc : AlmostRightCancellative _≡_ 0ℚ _+_
  x y z : ℚ
  x≢0 : x ≢ 0ℚ
  eq : y + x ≡ z + x

-- Inference error
test : y ≡ z
test = rc y z eq

-- No inference error
test2 : y ≡ z
test2 = arc y z x≢0 eq
```

Therefore I think the correct definitions are:
```agda
LeftCancellative _•_ = ∀ x y z → (x • y) ≈ (x • z) → y ≈ z
RightCancellative _•_ = ∀ x y z → (y • x) ≈ (z • x) → y ≈ z
AlmostLeftCancellative e _•_ = ∀ {x} y z → ¬ x ≈ e → (x • y) ≈ (x • z) → y ≈ z
[...]

PR #2117, opened 2023-09-30
Numeric non-zero equivalents/Redesign of `Almost*` properties in `Algebra.Definitions`

jamesmckinna:
DRAFT: for discussion. 

See my recent comments on/regarding #1175 #1436 #1562 #1579...

TODO: plenty!


jamesmckinna:
Oops, pushed before `make test` had finished...

src/Algebra/Definitions.agda
```diff
@@ -26,21 +25,122 @@ module Algebra.Definitions
...
+------------------------------------------------------------------------
+-- Properties of operations
+
+------------------------------------------------------------------------
+-- A generalisation: because not everything can be defined in terms of
+-- a single relation _≈_, AND because the corresponding definitions in
+-- `Relation.Binary.Core` use *implicit* quantification... so there is
+-- plenty of potential for later/downstream refactoring of this design
+
+module Monotonicity {ℓ₁ ℓ₂} (_≤₁_ : Rel A ℓ₁) (_≤₂_ : Rel A ℓ₂) where
```
Taneb, line 41: You've probably realized this, but so it doesn't get forgotten: these definitions don't use `_≈_` but are passed it anyway. We should have files `Algebra.Definitions.Monotonicity` and `Algebra.Definitions.Cancellativity` to avoid this if we keep these

src/Algebra/Definitions.agda
```diff
@@ -26,21 +25,122 @@ module Algebra.Definitions
...
+
+  Monotone₁ : Op₁ A → Set _
+  Monotone₁ f = ∀ x y → x ≤₁ y → (_≤₂_ on f) x y
+
+  Monotone₂ : Op₂ A → Set _
+  Monotone₂ _∙_ = ∀ x y u v → x ≤₁ y → u ≤₁ v → (x ∙ u) ≤₂ (y ∙ v)
```
Taneb, line 47: This looks very similar to `Congruent₂` to me. I wonder if there's something there.

src/Algebra/Definitions.agda
jamesmckinna: Good point, and I had indeed, overlooked this aspect of the design, in favour of trying to work out in eg `Data.Rational.Properties` how to selctively instantiate `Definitions` at various points with relations *other than* equality... on the analogy with `*-Reasoning` modules...

src/Algebra/Definitions.agda
jamesmckinna: Well, you're correct, of course, but this is precisely because the `Congruent*`/`Relation.Binary.Preserves*` properties use *implicit* quantification (because the relation instances can be inferred by the typechecker), distinct from the design decision made in issue #1436 to make the `Cancellative` properties take *explicit* arguments. I chose `Monotone` etc. because at one stage I was trying to have a single design in which both the `Monotone` and `Cancellative` deifnitions could be derived from a single family. But it required a fair amount of `_on_` gymnastics, plus redundant uses of `id`, and I decided to inline everything by hand. 

But the general point: we almost never use `Congruent*`, but definitely *do* use `Monotone` a lot, still stands. 

So I would pleased if this specimen design prompts a wider rethink. There are a number of outstanding issues even in the v2.0 milestone which I think could be tackled with a design along these lines, if we could only think about it enough and agree... so this contribution is offered up towards such an end.

src/Data/Integer/Base.agda
```diff
@@ -183,6 +183,12 @@ nonNegative : ∀ {i} → i ≥ 0ℤ → NonNegative i
 nonNegative {+0}       _ = _
 nonNegative {+[1+ n ]} _ = _
 
+-- Destructors -- mostly see `Data.Integer.Properties`
+
+≢-nonZero⁻¹ : ∀ i → .{{NonZero i}} → i ≢ 0ℤ
```
jamesmckinna, line 188: Should `i` be implicit here? (ditto. all the subsequent lemmas of related form)

src/Data/Rational/Base.agda
```diff
@@ -202,6 +202,11 @@ nonPositive {p@(mkℚ _ _ _)} (*≤* p≤q) = ℚᵘ.nonPositive {toℚᵘ p} (
 nonNegative : ∀ {p} → p ≥ 0ℚ → NonNegative p
 nonNegative {p@(mkℚ _ _ _)} (*≤* p≤q) = ℚᵘ.nonNegative {toℚᵘ p} (ℚᵘ.*≤* p≤q)
 
+-- Destructor -- mostly see `Data.Rational.Properties`
+
+≢-nonZero⁻¹ : ∀ p → .{{NonZero p}} → p ≢ 0ℚ
+≢-nonZero⁻¹ _ ⦃ () ⦄ refl
```
jamesmckinna, line 208: This is a good example of why/how `p` could/should be implicit?

src/Data/Rational/Properties.agda
```diff
@@ -165,6 +165,13 @@ mkℚ+-pos (suc n) (suc d) = _
   ... | refl with ℤ.*-cancelʳ-≡ n₁ n₂ (+ suc d₁) eq
   ...   | refl = refl
 
+
+------------------------------------------------------------------------
+-- Properties of NonZero: destructor
+
+≄-nonZero⁻¹ : ∀ p → .{{NonZero p}} → ¬ (p ≃ 0ℚ)
+≄-nonZero⁻¹ p eq = ≢-nonZero⁻¹ p (≃⇒≡ eq)
```
jamesmckinna, line 173: Infectively so.

src/Data/Rational/Unnormalised/Base.agda
```diff
@@ -190,6 +190,11 @@ nonNegative : ∀ {p} → p ≥ 0ℚᵘ → NonNegative p
 nonNegative {mkℚᵘ +0       _} (*≤* _) = _
 nonNegative {mkℚᵘ +[1+ n ] _} (*≤* _) = _
 
+-- Destructors -- mostly see `Data.Rational.Properties`
+
+≢-nonZero⁻¹ : ∀ p → .{{NonZero p}} → p ≢ 0ℚᵘ
+≢-nonZero⁻¹ _ ⦃ () ⦄ refl
```
jamesmckinna, line 196: Ditto.

src/Data/Integer/Base.agda
```diff
@@ -183,6 +183,12 @@ nonNegative : ∀ {i} → i ≥ 0ℤ → NonNegative i
...
+≢-nonZero⁻¹ +0 ⦃ () ⦄
+≢-nonZero⁻¹ +[1+ n ] ()
```
jamesmckinna, line 190: This being the only instance where there is more than one clause to consider...

CHANGELOG.md
```diff
@@ -523,14 +526,6 @@ Non-backwards compatible changes
     - From: `∀ {x} y z → (y • x) ≈ (z • x) → y ≈ z`
     - To: `∀ x y z → (y • x) ≈ (z • x) → y ≈ z`
```
jamesmckinna, line 528: I've removed the existing text regarding `Alomost*Cancellative`, in the absence of a stable story emerging about how it should be replaced. But this comment is a placeholder TODO...

jamesmckinna:
Closing this as superseded by #2573 , although much of the work here on `Monotonicity` should/could/might perhaps be revived later...

PR #2084, opened 2023-09-10
Changes explicit argument `y` to implicit in `Induction.WellFounded.WfRec`

jamesmckinna:
See the discussion on issue #2083 . 
"Almost-always-inferrable" turns out to cover ~~86~~ *all*  instances, but with ~~10~~ 1 instance ~~s retained, or~~ inserted for typechecking consistency. (UPDATED:  think the arithmetic is correct here!)

A further change in the types of `SubsetRecursor` and `SubsetRecursorBuilder` to make their first argument implicit as well might make certain other uses eliminable... but not for this PR, I think.

PR #2119, opened 2023-10-02
Added Irreflexivity and Asymmetry of WellFounded Relations

jmarkakis:


src/Induction/WellFounded.agda
```diff
@@ -112,6 +113,24 @@ module FixPoint
...
+  Wf-asymm wf x<y y<x = Acc-asymm _ (wf _) _ x<y y<x
+
+  Acc-irrefl : {_≈_ : Rel A ℓ} → Symmetric _≈_ → _<_ Respects₂ _≈_ → 
+               (x : A) → Acc _<_ x →  (y : A) → x ≈ y → ¬ (x < y)
+  Acc-irrefl s r x p y x≈y x<y = 
+    Acc-asymm x p y x<y (proj₂ r x≈y (proj₁ r (s x≈y) x<y))
+
+  Wf-irrefl : WellFounded _<_ → {_≈_ : Rel A ℓ} → Symmetric _≈_ → 
+              _<_ Respects₂ _≈_ → Irreflexive _≈_ _<_
+  Wf-irrefl wf s r = Acc-irrefl s r _ (wf _) _
```
jamesmckinna, line 133: These proofs are (redundant) consequences of the preceding two, coupled with `Relation.Binary.Consequences.asym⇒irr`. Indeed, they don't as such depend on `Acc` or `WellFounded`, so it's not obvious that they belong here? 

Or if you think they do, then please refactor to use the above lemma.

src/Induction/WellFounded.agda
```diff
@@ -112,6 +113,24 @@ module FixPoint
   unfold-wfRec : ∀ {x} → wfRec P f x ≡ f x (λ y _ → wfRec P f y)
   unfold-wfRec {x} = f-ext x wfRecBuilder-wfRec
 
+------------------------------------------------------------------------
+-- Well-founded relations are asymmetric and irreflexive.
+
+module _ {_<_ : Rel A r} where
+  Acc-asymm : (x : A) → (Acc _<_ x) → (y : A) → x < y → ¬ (y < x)
+  Acc-asymm = Some.wfRec _ λ x hx y x<y y<x → hx y y<x x y<x x<y
```
jamesmckinna, line 121: Firstly: naming
I think that the ambient style of the library calls `Asymmetric` proofs `asym`, so I'd probably call this `Acc-asym`?

Secondly: implicit/explicit quantification
At the cost of some clunkiness on the RHS of this definition, I would be tempted to refactor the type of `Acc-asym(m)` to be
```agda
∀ {x y} → (Acc _<_ x) → x < y → ¬ (y < x)
```

src/Induction/WellFounded.agda
```diff
@@ -112,6 +113,24 @@ module FixPoint
...
+
+  Wf-asymm : WellFounded _<_ → Asymmetric _<_
+  Wf-asymm wf x<y y<x = Acc-asymm _ (wf _) _ x<y y<x
```
jamesmckinna, line 124: With the knock-on consequence that this lemma then becomes the rather cleaner
```agda
Wf-asym : WellFounded _<_ → Asymmetric _<_
Wf-asym wf wf = Acc-asym (wf _)
```

src/Induction/WellFounded.agda
MatthewDaggitt: `acc-asym` is good. Also don't need brackets around `Acc _<_ x`.

jmarkakis:
Thank you for the comments. I made the suggested changes, renaming the lemmas and making some arguments implicit.

jamesmckinna:
You'll need to update the `CHANGELOG` as well to reflect the changes, and to omit the implicit quantifications...

src/Induction/WellFounded.agda
jamesmckinna: So, I spent some time playing with this, and came to the following compromise position, for your consideration:
* do you need `acc-irrefl` as a stand-alone result? if so, then there's nothing further to do besides fix the `CHANGELOG`
* if not, then `acc-asym` and `wf-asym` can work as they stand, and `wf-irrefl` can simply delegate to `wf-asym` and `asym⇒irr`

I prefer the second choice, but you may want to retain the more 'local' property as well.

jamesmckinna:
The use of the `Some` submodule of ``Induction.WellFounded` in this PR draws attention to the *restriction*/'local'isation of properties to the `Acc`-essible subtype of a given set `A`. What `Relation.Binary.Consequences` expects/provides, are proofs for the 'global'/*unrestricted* properties. So we might need to consider a downstream reorganisation there to make these things a bit more streamlined. I'll consider raising a separate issue after v2.0.

jamesmckinna:
Thanks for a nice first PR!

src/Induction/WellFounded.agda
jmarkakis: I fixed the changelog, and simplified the proof of `wf-irrefl` as you suggested removing `acc-irrefl`. Thank you for the comments again.

CHANGELOG.md
```diff
@@ -3514,6 +3514,12 @@ This is a full list of proofs that have changed form to use irrelevant instance
 * Added new proof to `Induction.WellFounded`
   ```agda
   Acc-resp-flip-≈ : _<_ Respectsʳ (flip _≈_) → (Acc _<_) Respects _≈_
+
+  acc-asym : ∀ {x y} → Acc _<_ x → x < y → ¬ (y < x)
```
jamesmckinna, line 3518: Style guide advises removing the initial `∀ {x y} →` prefix here, as it can be inferred by the reader.

CHANGELOG.md
```diff
@@ -3514,6 +3514,12 @@ This is a full list of proofs that have changed form to use irrelevant instance
...
+  wf-asym : WellFounded _<_ → Asymmetric _<_
+    
+  wf-irrefl : {_≈_ : Rel A ℓ} → _<_ Respects₂ _≈_ →
```
jamesmckinna, line 3521: Ditto with `{_≈_ : Rel A ℓ} → ` here.

jamesmckinna:
Otherwise great! Thanks!
@MatthewDaggitt could we bring this forward to v2.0? It's pretty much there...

CHANGELOG.md
jmarkakis: Fixed as suggested

CHANGELOG.md
jmarkakis: Fixed as suggested

src/Induction/WellFounded.agda
```diff
@@ -116,6 +119,21 @@ module FixPoint
   unfold-wfRec : ∀ {x} → wfRec P f x ≡ f x λ _ → wfRec P f _
   unfold-wfRec {x} = f-ext x wfRecBuilder-wfRec
 
+------------------------------------------------------------------------
+-- Well-founded relations are asymmetric and irreflexive.
+
+module _ {_<_ : Rel A r} where
+  acc-asym : ∀ {x y} → Acc _<_ x → x < y → ¬ (y < x)
```
MatthewDaggitt, line 126: For [these kind of properties](https://github.com/agda/agda-stdlib/blob/84f7366c3a038f8bcfc4ab58eaba43cb59e02d16/src/Relation/Binary/Consequences.agda#L75) the standard name would probably use `⇒` rather than `-` in the name, e.g. `acc⇒asym`, `wf⇒asym`, `wf⇒irrefl`.

MatthewDaggitt (commented):
Other than that looks good! Heads up that there may however be implicit merge conflicts with https://github.com/agda/agda-stdlib/pull/2084 that just got merged in.

src/Induction/WellFounded.agda
```diff
@@ -116,6 +119,21 @@ module FixPoint
...
+  acc-asym {x} {y} hx = 
+    (Some.wfRec _ λ x hx y x<y y<x → hx y y<x x y<x x<y) x hx y
```
jamesmckinna, line 128: I calculate the following to repair things after the merge conflict:
```agda
acc-asym {x} hx = Some.wfRec (λ x → ∀ {y} → x < y → ¬ (y < x)) (λ x hx x<y y<x → hx y<x y<x x<y) x hx
```
where now the explicit predicate argument needs to be supplied to `Some.wfRec`, and the implicit arguments `y` are suppressed.

HTH.

src/Induction/WellFounded.agda
jmarkakis: I changed the names as suggested

src/Induction/WellFounded.agda
jmarkakis: Thank you for solving this problem, and sorry for the late response. I wasn't sure what's the correct way to get the new commits into my branch - I'm still a bit new to git.

MatthewDaggitt:
Thanks, merging in. Thanks again for the great first PR!

PR #2120, opened 2023-10-05
Make size parameter on 'replicate' explicit

### LINKED ISSUE #2097: `n` in `Vec.replicate` should be explicit
> > > > I agree. The argument to Vec.iterate should be explicit, not implicit
> > >
> > > Should that be the case with `Vec.replicate` too?
> >
> >Yes, nice spot!
>
> Turns out, Vec.replicate is comparatively a lot more involved within the codebase. Would it be better if I create a downstream PR for this (given that I have updated 15+ files till now, and there still are files left with Vec.replicate occurrences)?

_Originally posted in https://github.com/agda/agda-stdlib/issues/2051#issuecomment-1685420148_

(creating an issue to ensure that this isn't lost between PRs and reviews)

MatthewDaggitt:
Fixes #2097. Lots of implicits being passed explicitly removed, so it looks like it was the right idea :tada:

### DIFF
--- a/src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
+++ b/src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
@@ -21,7 +21,7 @@ open import Relation.Unary
 
 private
   variable
-    a b c p q r ℓ : Level
+    a p ℓ : Level
     A B C : Set a
     P Q R : Pred A p
     m n : ℕ
--- a/src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
+++ b/src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
@@ -8,6 +8,7 @@
 
 module Data.Vec.Functional.Relation.Unary.All.Properties where
 
+open import Data.Nat.Base using (ℕ)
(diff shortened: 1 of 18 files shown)


jamesmckinna:
I'll try to review tomorrow.

src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
```diff
@@ -21,79 +22,72 @@ open import Relation.Unary
 private
   variable
     a b c p q r ℓ : Level
-    A : Set a
-    B : Set b
-    C : Set c
+    A B C : Set a
```
jamesmckinna, line 25: In view of your refactoring of the declarations for `A B C`, do we need so many `Levels` declared as well? Or would `a p ℓ : Level` suffice?

jamesmckinna:
The extensive `variable` refactoring is nice, but see my comment.
Agree that the issue was a good one to pursue, given the number of occurrences!

src/Data/Vec/Functional/Relation/Unary/All/Properties.agda
MatthewDaggitt: Yup, nice spot!

Issue #2112, opened 2023-09-28
*Top-level* Record definitions need accompanying type signatures


jamesmckinna:
Working on #2111 exposes a previously uncaught error: under `Data.Tree.AVL`, a number of modules (attempt to) declare 'local' instances of `StrictTotalOrder` *without declaring a suitable type signature for them*, which worked perfectly fine *as long as this record type was a leaf in the `Algebra.Bundles` hierarchy*...

... but adding `(Is)DenseLinearOrder` as a sub(structure)/bundle of `(Is)StrictTotalOrder` has now broken that. 

Remedy: identify all `record` definitions without a type signature, and supply it appropriately. 
UPDATED: issue only applies to *top-level* definitions; within the scope of an enclosing record, the type of the field being defined as a sub-record ensures that a suitable type is being supplied to the typechecker. 
UPDATED: I've attempted to identify any other instances in the library, but it seems just to have been these 'historical' examples. Suggest closing the issue now/for the time being.

MatthewDaggitt:
Ooh interesting. Haven't come across that sort of bug before!

jamesmckinna:
So the bug expressed itself in terms of unsolved `Level` meta variables... my diagnosis in terms of leaves in the hierarchy is post hoc inference, and there may be other possibilities, but inserting the type deck certainly solves/d the instances that came up. 

And yes, a weird one!

PR #2125, opened 2023-10-06
Backward compatibility with v1.7.1/2 for the IO Monad

### LINKED ISSUE #2124: Export `return` from `IO.Primitive`
Trying to check my development _Type-preserving compilation via dependently typed syntax_ ([SPLS March 2023](https://scottish-pl-institute.github.io/spls/meetings/2023/february/)) with the current standard library, I get a failure for a trivial reason:
```
The module IO.Primitive doesn't export the following:
  return
  open import IO.Primitive public using (return; _>>=_)
```
In this state, I see no easy way to make my development work with both std-lib 1.7.1/2 and 2.0.

An easy fix would be to restore `return` to this module with the alias `return = pure` which could maybe be coming with a deprecation warning.  
(Though in my view there is no consensus in the FP community that `return` is a bad word and `pure` the good word. Esp. when the `return` statement is the last of my function, it makes perfectly sense to say `return result` rather than `pure result`).

Ref: https://github.com/agda/agda-stdlib/blame/master/src/IO/Primitive.agda

andreasabel:
- Fix #2124: `IO.Primitive.return` (alias for `pure`) for backwards compat
- Restore Category/* tree of v1.7 as deprected module aliases

These patches allow me to port an Agda development using v1.7.1 to v2.0 (current `master`).  

The code I want to have working with both version uses `do` for the `IO` monad and `mapM`.
```agda
foo = do
   ...
   _ <- mapM f xs
   ...
   return _
   where
   open IOMonad
   open TraversableM record{ IOMonad }
```
So I am using the `module` based technique to bring `>>=` into scope for the `do` notation.

Previously (v1.7), I defined `IOMonad` like this:
```agda
module IOMonad where
  open import IO.Primitive public using (return; _>>=_)

  infixl 1 _>>_

  _>>_  : ∀ {b} {B : Set b} → IO ⊤ → IO B → IO B
  _>>_ = λ m m' → m >>= λ _ → m'

  infixr 1 _=<<_

  _=<<_  : ∀ {a b} {A : Set a} {B : Set b} → (A → IO B) → IO A → IO B
  k =<< m = m >>= k
```
This definition was good enough to form `record {IOMonad}` for the instantiation of `TraversableM`.

With v2.0, forming a `RawMonad` also needs embedded instances of `RawFunctor` and `RawApplicative`.
I managed to make the definition of `IOMonad` compatible with v2.0 by changing it to be the following:
```agda
open import Category.Functor     using (RawFunctor)
open import Category.Applicative using (RawApplicative)

module IOFunctor where
  open import IO.Primitive public using (return; _>>=_)

  infixl 1 _>>_

  _>>_  : ∀ {b} {B : Set b} → IO ⊤ → IO B → IO B
  _>>_ = λ m m' → m >>= λ _ → m'

  infixr 1 _=<<_

  _=<<_  : ∀ {a b} {A : Set a} {B : Set b} → (A → IO B) → IO A → IO B
  k =<< m = m >>= k

  -- The following definition is needed to create the RawFunctor instance.

  infixl 4 _<$>_

  _<$>_ :  ∀ {a b} {A : Set a} {B : Set b} → (A → B) → IO A → IO B
  f <$> m = m >>= λ a -> return (f a)

-- This module is needed to create the RawApplicative instance
-- in a way compatible with both std-lib v1.7.1/2 and v2.0.

module IOApplicative where
  open IOFunctor public

  rawFunctor : ∀{ℓ} → RawFunctor (IO {a = ℓ})
  rawFunctor = record { IOFunctor }

  pure : ∀{a} {A : Set a} → A → IO A
  pure = return

  infixl 4 _<*>_ _⊛_

  _<*>_ : ∀ {a b} {A : Set a} {B : Set b} → IO (A → B) → IO A → IO B
  mf <*> ma = mf >>= λ f → ma >>= λ a → return (f a)

  _⊛_ : ∀ {a b} {A : Set a} {B : Set b} → IO (A → B) → IO A → IO B
  _⊛_ = _<*>_

module IOMonad where
  open IOApplicative public

  -- Field rawApplicative is part of the RawMonad of std-lib v2.0
  -- Thus we have to construct the Applicative implementation
  -- even though we do not use it.
  --
  rawApplicative : ∀{ℓ} → RawApplicative (IO {a = ℓ})
  rawApplicative = record { IOApplicative }
```
This works with 1.7.1/2.  To have it work with 2.0 as well, the present PR is needed, because:
- I need to be able to import `return` from `IO.Primitive`.
- I need to be able to import `Raw*` from `Category.*` (for `*` in {`Functor`, `Applicative`}).

### DIFF
--- a/src/Category/Applicative.agda
+++ b/src/Category/Applicative.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Applicative` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Applicative where
+
+open import Effect.Applicative public
+
+{-# WARNING_ON_IMPORT
+"Category.Applicative was deprecated in v2.0.
+Use Effect.Applicative instead."
+#-}
--- a/src/Category/Applicative/Indexed.agda
+++ b/src/Category/Applicative/Indexed.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Applicative.Indexed` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Applicative.Indexed where
+
+open import Effect.Applicative.Indexed public
+
+{-# WARNING_ON_IMPORT
+"Category.Applicative.Indexed was deprecated in v2.0.
+Use Effect.Applicative.Indexed instead."
+#-}
--- a/src/Category/Applicative/Predicate.agda
+++ b/src/Category/Applicative/Predicate.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Applicative.Predicate` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Applicative.Predicate where
+
+open import Effect.Applicative.Predicate public
+
+{-# WARNING_ON_IMPORT
+"Category.Applicative.Predicate was deprecated in v2.0.
+Use Effect.Applicative.Predicate instead."
+#-}
--- a/src/Category/Comonad.agda
+++ b/src/Category/Comonad.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Comonad` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Comonad where
+
+open import Effect.Comonad public
+
+{-# WARNING_ON_IMPORT
+"Category.Comonad was deprecated in v2.0.
+Use Effect.Comonad instead."
+#-}
--- a/src/Category/Functor.agda
+++ b/src/Category/Functor.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Functor` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Functor where
+
+open import Effect.Functor public
+
+{-# WARNING_ON_IMPORT
+"Category.Functor was deprecated in v2.0.
+Use Effect.Functor instead."
+#-}
--- a/src/Category/Functor/Indexed.agda
+++ b/src/Category/Functor/Indexed.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Functor.Indexed` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Functor.Indexed where
+
+open import Effect.Functor.Indexed public
+
+{-# WARNING_ON_IMPORT
+"Category.Functor.Indexed was deprecated in v2.0.
+Use Effect.Functor.Indexed instead."
+#-}
--- a/src/Category/Functor/Predicate.agda
+++ b/src/Category/Functor/Predicate.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Functor.Predicate` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Functor.Predicate where
+
+open import Effect.Functor.Predicate public
+
+{-# WARNING_ON_IMPORT
+"Category.Functor.Predicate was deprecated in v2.0.
+Use Effect.Functor.Predicate instead."
+#-}
--- a/src/Category/Monad.agda
+++ b/src/Category/Monad.agda
@@ -1,0 +1,17 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- This module is DEPRECATED. Please use
+-- `Effect.Monad` instead.
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Category.Monad where
+
+open import Effect.Monad public
+
+{-# WARNING_ON_IMPORT
+"Category.Monad was deprecated in v2.0.
+Use Effect.Monad instead."
+#-}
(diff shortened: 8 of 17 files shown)


gallais:
I thought that the point of v2.0 was that we were happy to break backwards compatibility more aggressively?
Otherwise we may as well call it v1.8.

Also, if this does get in then `IO.Primitive`'s `return`  should definitely be deprecated
because of the name clash I mention in the related issue which was the motivation
behind getting rid of it in the first place.

jamesmckinna:
And... the whole point of the breaking change to remove the top-level namespace entry `Category.*` altogether. The subsidiary `*.Categorical` modules were (eventually ;-)) retained for backwards compatibility, but this change would condemns us to another cycle of versions which can't make use of anything `agda-categories` to populate `Catgeory.*`. 

So I would be strongly against this merge, FWIW.

jamesmckinna:
Can the definitions of `rawFunctor` and `rawApplicative` be inlined (without associated typedeclarations)in the definition of `record{ IOMonad }`?

andreasabel:
@jamesmckinna wrote:
> Can the definitions of `rawFunctor` and `rawApplicative` (not?) be inlined (without associated type declarations) in the definition of `record{ IOMonad }`?

The problem is that v1.7 does not expect such fields here in a `RawMonad`.  The trick I am using here is that `record{ M }`  will pick from `M` the fields it needs and ignores the rest of `M`.  

@jamesmckinna wrote:
> but this change would condemns us to another cycle of versions which can't make use of anything `agda-categories` to populate `Catgeory.*`.

Ah, sorry, I was not aware of such plans.  Currently, `Category.*` is unused, and `agda-categories` uses `Categories.*`.

@gallais wrote:
> Also, if this does get in then IO.Primitive's return should definitely be deprecated

I added a deprecation warning.

jamesmckinna:
Issue #1636 / PR #1735

jamesmckinna:
@andreasabel wrote:
> @jamesmckinna wrote:
> 
> > Can the definitions of `rawFunctor` and `rawApplicative` (not?) be inlined (without associated type declarations) in the definition of `record{ IOMonad }`?
> 
> The problem is that v1.7 does not expect such fields here in a `RawMonad`. The trick I am using here is that `record{ M }` will pick from `M` the fields it needs and ignores the rest of `M`.

Yes, I see. But I think the suggestion might (?) still work, even with the `module IOMonad` if the definitions `rawFunctor` and `rawApplicative` are given *without* type signatures. There are examples of this in stdlib, with local naming of `Algebra.Bundles.*` objects without explicitly giving their signatures. It does actually cause problems #2112 in some cases, but here? I wonder...

MatthewDaggitt:
> This works with 1.7.1/2. To have it work with 2.0 as well, the present PR is needed, because:

Yes, as @gallais and @jamesmckinna have identified, this is the crux of the problem. We've taken the decision not to preserve backwards compatibility with the 2.0 release.

What reason do you need for working both with `1.7.2` and the unreleased `2.0` at the same time? Could you simply upgrade when `2.0` is released?

andreasabel:
@jamesmckinna wrote:
> I think the suggestion might (?) still work, even with the `module IOMonad` if the definitions `rawFunctor` and `rawApplicative` are given _without_ type signatures.

Thanks for the hint.  I tried it out.  I got unsolved metas because these defs are polymorphic. But even if I could solve this problem, there would still be the problem that an alias `foo = record{Bar}` does not work as the elaboration of `record{Bar}` needs to know which fields you want to pick from module `Bar`.  So you need the record type here.

andreasabel:
On a more general note, my goal is to develop such that:
1. My code works with the latest release of `std-lib`.
2. My code also works with the `master` of `std-lib`.

Unfortunately, Agda does not natively understand CPP; that's the technique used in GHC Haskell to support different API versions.  I also lacks good support for dependency versioning (like Cabal's `MIN_VERSION_package` macros).

So far, I have been relying on the backwards compatibility of the std-lib. Things have not been removed from one release to the next, but just deprecated.  Then, _after_ the release, one can fix the deprecation warnings to again get something that works with the latest release and the master.

In the current gap between v1.7 and master I have to either use CPP or fork my project.

I think you must have discussed this scenario; do you have some advise how to proceed here?

In the present situation, because Agda 2.6.4 also works with 1.7.1/2, I could postpone updating my project to the next std-lib, but in general, there could be a problem...

andreasabel:
@jamesmckinna wrote:
> And... the whole point of the breaking change to remove the top-level namespace entry `Category.*` altogether. The subsidiary `*.Categorical` modules were (eventually ;-)) retained for backwards compatibility, but this change would condemns us to another cycle of versions which can't make use of anything `agda-categories` to populate `Catgeory.*`.

It seems like if v2.0 does not aim to populate the `Category.*` module tree, it should be possible to remove the aliases I propose here directly after the release of v2.0 and so they would be free for use in v2.1.  Or maybe there is a policy of having names unused for one release before rededication?  (Like phone numbers should rest for a couple of years before recycling...)

jamesmckinna:
Have you tried this suggestion https://github.com/agda/agda-stdlib/issues/2124#issuecomment-1750318453?

andreasabel:
> Have you tried the suggestion proposed in [#2124 (comment)](https://github.com/agda/agda-stdlib/issues/2124#issuecomment-1750318453)?

Making 1.7 forward compatible is of course another alternative, that would mean adding `Effect.*` modules in there reflecting the new organization in 2.0.  
Forward compatibility releases would of course be a new path we have not walked so far.

Btw, I just released Agda 2.6.4!

jamesmckinna:
> > Have you tried the suggestion proposed in [#2124 (comment)](https://github.com/agda/agda-stdlib/issues/2124#issuecomment-1750318453)?
> 
> Making 1.7 forward compatible is of course another alternative, that would mean adding `Effect.*` modules in there reflecting the new organization in 2.0. Forward compatibility releases would of course be a new path we have not walked so far.
> 
> Btw, I just released Agda 2.6.4!

Ah... there really seems to be no good solution without, in some way, interpolating @gallais 's hypothetical "v1.8" or "v1.7.3". Maddening! Or else, craving your indulgence to accept a *breaking* change, this one time. I understand the desire to maintain a 'smooth'/'progressive' upgrade/deprecation path (and v2.0 incorporates *many* such detours where it has been possible), but v2.0 has *so* many changes over v1.7.* that I think we would like users to accept... the break. 

Shout-outs to @gallais and @MatthewDaggitt ... I am a still `newcomer`, albeit 'relatively' such.

jamesmckinna:
Of course, we haven't consider forward compatibility before, I'm a drowning man clutching at anything that floats by, while trying to defend breaking changes that required quite some effort to achieve (and which I have already rolled back on once already, again on request from you). 

If we were to consider it as policy, then I could imagine that would/might open a *world* of pain in respect of any particular development which some user or other wishes to maintain. Your situation in that regard is somewhat different... ;-)

I am not, nor I think could be, any (final) arbiter in this.

shhyou:
> So far, I have been relying on the backwards compatibility of the std-lib. Things have not been removed from one release to the next, but just deprecated. Then, _after_ the release, one can fix the deprecation warnings to again get something that works with the latest release and the master.

FWIW, stdlib 2.0 is backwards-incompatible in many different ways in addition to declaring some definitions as deprecated. If I understand correctly, the hierarchy of quite some modules is reorganized with exports being moved around; some definitions change implicit arguments to explicit; some definitions have different implementation so their proofs have to change, etc.

MatthewDaggitt:
Closing as superseded by https://github.com/agda/agda-stdlib/pull/2141

PR #2126, opened 2023-10-07
'No infinite descent' for (`Acc`essible elements of) `WellFounded` relations

jamesmckinna:
I've added this, because it didn't obviously seem part of the existing library.
Currently in the form of a stand-alone module, but could/should be merged into `Induction.WellFounded`?

Includes an additional lemma for `TransClosure` which I first thought I needed (and then realised I could do without), but which is also itself a useful, but 'missing', lemma. 

Refactoring opportunities, in addition to the above: rewrite in terms of `Relation.Unary.Closure.Base`, but if so, then arguably perhaps all of `Induction.WellFounded` should be, as well.


src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,81 @@
...
+------------------------------------------------------------------------
+-- Definitions
+
+module InfiniteDescent (_<_ : Rel A r) (P : Pred A ℓ)  where
+
+    DescentAt : Pred A _
+    DescentAt x = P x → ∃[ y ] y < x × P y
+
+    InfiniteDescent : Set _
+    InfiniteDescent = ∀ {x} → DescentAt x
```
JacquesCarette, line 35: How is this 'infinite' ? Who tells me that `A` actually has infinitely many inhabitants?

It seems to me that this predicate is true basically because `DescentAt x` is always false if you pick `x` to be a 'top' element.

I expected some kind of map involving Nat that traced an infinitely long path.

jamesmckinna: You make a very fair point. 

Originally, I had called the `InfiniteDescent` definition `SmallerCounterexample`, and the overall module `NoSmallestCounterexample`, because that's the to-me-more-familiar name for the principle. But Archimedes' "method of infinite descent" (from "The Sand Reckoner", IIRC), as (re)discovered by Fermat in the 17th century seems to be the more accepted term in (classical) mathematics (where of course `P` is a *negated* proposition, corresponding to "counterexamples to P", so the conclusion is `All P` from the impossibility of `¬ P`... but I digress).

In the 20th century, I suppose we might/should call this "Noetherian induction", and the property of being 'Noetherian' being also usually called "no infinite descending chains", because *in the presence of countable dependent choice*, one may iterate the `InfiniteDescent` condition precisely to generate a Nat-indexed descending chain in the ordering. But perhaps we *do* have enough from the existential to iterate the process? I'll try.

JacquesCarette (approved):
This seems to match what I'd expect (from the names) much more closely.

jamesmckinna:
Hope that `rebase`/`merge` conflict resolution has fixed things now.

jamesmckinna:
OK, so something went *horribly* wrong with the merge conflict resolution there. Yuk. Assistance required!

jamesmckinna:
In addition to the evident `CHANGELOG` merge conflict, I realise that there was never an entry there for the new module, while I dithered about making it separate, or rolling it into `Induction.WellFounded`...

jamesmckinna:
It would be nice to merge this before it gets clobbered by another `CHANGELOG` merge conflict...

src/Relation/Unary.agda
```diff
@@ -162,6 +162,11 @@ IUniversal P = ∀ {x} → x ∈ P
 
 syntax IUniversal P = ∀[ P ]
 
+-- Stability - instances of P are stable wrt double negation
+
+Stable : Pred A ℓ → Set _
+Stable P = ∀ x → Nullary.Stable (P x)
```
MatthewDaggitt, line 168: Missing CHANGELOG entry?

src/Relation/Binary/Construct/Closure/Transitive.agda
```diff
@@ -70,6 +70,13 @@ module _ (_∼_ : Rel A ℓ) where
   transitive : Transitive _∼⁺_
   transitive = _++_
 
+  transitive⁻ : Transitive _∼_ → _∼⁺_ ⇒ _∼_
+  transitive⁻ trans = ∼⁺⊆∼
+    where
+    ∼⁺⊆∼ : _∼⁺_ ⇒ _∼_
+    ∼⁺⊆∼ [ x∼y ]      = x∼y
+    ∼⁺⊆∼ (x∼y ∷ x∼⁺y) = trans x∼y (∼⁺⊆∼ x∼⁺y)
```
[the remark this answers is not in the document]
MatthewDaggitt: Missing CHANGELOG entry as well?

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+    DescentAt x = P x → ∃[ y ] y < x × P y
+
+    Acc⇒Descent : Set _
+    Acc⇒Descent = ∀[ Acc _<_ ⇒ DescentAt ]
+
+    Descent : Set _
+    Descent = ∀[ DescentAt ]
+
+    InfiniteDescendingSequence : (f : ℕ → A) → Set _
+    InfiniteDescendingSequence f = ∀ n → f (suc n) < f n
```
MatthewDaggitt, line 47: This is unnecessarily parameterised on `P`?

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+
+    InfiniteSequence_From_ : (f : ℕ → A) → Pred A _
+    InfiniteSequence f From x = f zero ≡ x × InfiniteDescendingSequence f
```
MatthewDaggitt, line 50: As is this

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+    module ID⁺ = InfiniteDescent _<⁺_ P
+
+    DescentAt⁺   = ID⁺.DescentAt
+    Descent⁺     = ID⁺.Descent
+
+  Acc⇒Descent⁺ : Set _
+  Acc⇒Descent⁺ = ∀[ Acc _<_ ⇒ DescentAt⁺ ]
+
+  InfiniteDescendingSequence⁺ : (f : ℕ → A) → Set _
+  InfiniteDescendingSequence⁺ f = ∀ {m n} → m ℕ.< n → f n <⁺ f m
```
MatthewDaggitt, line 155: Again independent of P etc.

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+------------------------------------------------------------------------
+-- Finally:  the (classical) 'no smallest counterexample' principle itself
+
+module _ {_<_ : Rel A r} {P : Pred A ℓ} (stable : Stable P) where
+
+  open FurtherCorollaries {_<_ = _<_} (∁ P)
+    using (acc⇒noInfiniteDescent⁺; wf⇒noInfiniteDescent⁺)
+    renaming (Acc⇒Descent⁺ to NoSmallestCounterExample)
+
+  acc⇒noSmallestCounterExample : NoSmallestCounterExample → ∀[ Acc _<_ ⇒ P ]
```
MatthewDaggitt, line 210: It's weird to have lemmas exported from this module that refer to the name `NoSmallestCounterExample` and for that name to not be exported from this module?

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+          where
+          f : ℕ → A
+          f zero = y
+          f (suc n) = g n
+          f0≡y : f zero ≡ y
+          f0≡y = refl
+          f<P : ∀ n → f (suc n) < f n
+          f<P zero rewrite g0≡z = z<y
+          f<P (suc n)           = g<P n
+          Π[P∘f] : Π[ P ∘ f ]
```
MatthewDaggitt, line 83: Please can we have empty lines between of these lemmas?

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+          Π[P∘f] zero rewrite g0≡z = py
+          Π[P∘f] (suc n)           = Π[P∘g] n
+
+      wf⇒infiniteDescent : WellFounded _<_ → InfiniteDescent
+      wf⇒infiniteDescent wf = acc⇒infiniteDescent (wf _)
+
+      acc⇒noInfiniteDescent : ∀[ Acc _<_ ⇒ ¬_ ∘ P ]
```
MatthewDaggitt, line 90: The point-free style for the type ` ∀[ Acc _<_ ⇒ ¬_ ∘ P ]` is really _really_ arcane. You have to have super high levels of knowledge of precedence and types even to be able to parse it. Is the couple of extra characters needed for `∀ x → Acc _<_ x → ¬ (P x)` really so bad?

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,221 @@
...
+        rec : _
+        rec y rec[y] py = let z , z<y , pz = descent py in rec[y] z<y pz
+
+      wf⇒noInfiniteDescent : WellFounded _<_ → ∀[ ¬_ ∘ P ]
+      wf⇒noInfiniteDescent wf = acc⇒noInfiniteDescent (wf _)
+
+------------------------------------------------------------------------
+-- Corollaries: assume descent only for Acc _<_ elements
+
+module _ {_<_ : Rel A r} (P : Pred A ℓ) where
```
MatthewDaggitt, line 102: Given the amount of module indirection, it is very difficult to work out whether `_<_` can truly be implicit for all the lemmas in this module... Have you checked them all?

MatthewDaggitt (changes requested):
Left some preliminary comments, but my overall one is that the nested module structure of this module has lead to at least one major bug, and as some of my comments mention, makes it hard to check for others. 

What are your thoughts about adding to the style guide that unless in exceptional circumstances, more than two levels of nested modules should be banned?

src/Induction/NoInfiniteDescent.agda
jamesmckinna: You're right... and this PR dates from the time (only a few months' back) when I was super-invested in that pointfree approach to `Unary` predicates, which you criticised previously in #2181 ... and I accept the criticism ;-)

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Er, ... no, I can't honestly say I have. I think I've been working on gut instinct that Agda can infer `_<_` from instances of `Descent`/`DescentAt` but I may very well be mistaken in those instincts. Good catch!

src/Relation/Unary.agda
jamesmckinna: Oops!

src/Relation/Binary/Construct/Closure/Transitive.agda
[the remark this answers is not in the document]
jamesmckinna: Re: overkill
Perhaps you're right. But I think sometimes we don't pay enough attention to observing when certain lemmas really are 'arrow/preorder-theoretic' for structures other than the usual one of implication... so there's a balance to strike between 'minimality' and 'redundant recoding' in my mind as regards how certain things get phrased. I'm not clear in my own mind how much we want to strip down stdlib entries, or have them as exemplary 'staged' versions of how to lay out proofs (for newcomers to follow, and for our students, hopefully, to emulate...)?

In this instance, I think I had had a more articulated version of `_∼⁺_ ⇒ _∼_` and `_∼_ ⇒ _∼⁺_` in mind in order to set up instances of `WellFounded.Subrelation` until I realised that I didn't in fact need them. And this may be a fossil record of that earlier line of thinking.

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Aaargh, yes, you get me every time on this. I'm still stuck in the 1990s, when the LEGO and Coq sectioning mechanisms did 'auto-strengthening' of definitions to remove unused hypotheses. And of course, Agda modules don't work like that, and I clearly subconsciously still wish that they did!

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Sigh...

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Yup.

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Ha! Good catch!

jamesmckinna:
> Left some preliminary comments, but my overall one is that the nested module structure of this module has lead to at least one major bug, and as some of my comments mention, makes it hard to check for others.
> 
> What are your thoughts about adding to the style guide that unless in exceptional circumstances, more than two levels of nested modules should be banned?

Two thoughts:
* yes, as with previous PRs, I am perhaps too enthusiastic (to the point of being careless) in my use of (nested) modules;
* I can see, again from this and previous experience, that at least as a reviewer (where it makes sense in terms of cognitive load; but correspondingly, as a developer, I find them useful, not least in trying to impose uniformity on dependencies, which otherwise might become ad hoc on a per lemma/definition basis; but you've already observed that I make mistakes with that :-()), but perhaps generally cf #2182 , you do have _strong_ views about the use of internal modules in the library!

It seems as though, regarding your proposal about the style guide, that our recent experiences with v2.0 are each inclining us to want to mandate more stringent guidelines along various axes, according to experience, taste and perhaps pet peeves. So I'd be worried about being too authoritarian... 

... that said, I do think it's a good `discussion` issue, either for the stdlib-admin group in Zulip, or for the issue tracker, so that others can join in?

JacquesCarette:
When I find myself wanting to use "too many" levels of nested modules, I try to use that as a signal to sit back and think some more. It's a signal that maybe the current organization can be improved, or at least that it should be reconsidered.

jamesmckinna:
Wise words indeed @JacquesCarette !

src/Relation/Unary.agda
MatthewDaggitt: No worries! That's why we have a review process!

I'm happy to stick with the definition. It's consistent with us lifting up `Decidable`/`Irrelevant` etc, to unary and binary relations.

MatthewDaggitt:
Okay, I'll wait to see what you come up with refactoring-wise to resolve the problem of the extra dependencies, and then maybe we could make this PR a topic of the next meeting, and get everyone's opinion on use of modules?

src/Relation/Unary.agda
jamesmckinna: Would you be happy with the definition
```agda
Stable = Universal (Nullary.Stable)
```
or is your/srtdlib's preference to write out the quantifiers in full? It seems as though, as more and more of such definitions get added, that we strive for some uniformity via higher-order combinators (but I understand about not exploiting the provided `syntax` declarations for `Universal` and `IUniversal`)...

src/Relation/Binary/Construct/Closure/Transitive.agda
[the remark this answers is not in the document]
jamesmckinna: Local recursive definitions vs. (for me!) harder-to-read inlined recursions... a `style-guide` question?
I've gone with your suggestion, but I think it's worth revisiting another time.

src/Induction/NoInfiniteDescent.agda
```diff
@@ -0,0 +1,235 @@
...
+
+    private
+      descent : ID⁺.Acc⇒Descent
+      descent = descent⁺  ∘ (accessible⁻ _)
+
+      module Corollaries⁺ = Corollaries _ _ descent
+
+    acc⇒infiniteDescent⁺ : Acc⇒InfiniteDescent⁺
+    acc⇒infiniteDescent⁺ acc[x] px
+      with f , (f0≡x , sequence[f]) , Π[P∘f] ← Corollaries⁺.acc⇒infiniteDescent (accessible _ acc[x]) px
```
jamesmckinna, line 202: I have/had no idea how best to break this too-long line!?

src/Induction/NoInfiniteDescent.agda
jamesmckinna: Is the name exported correctly now?

src/Induction/NoInfiniteDescent.agda
jamesmckinna: So... I adopted the solution that all appeals to `_<_` would be explicit, at the potential cost of having to supply `_` arguments as module/definition parameters... hope this fits the bill!

jamesmckinna:
Regarding the (global) review comment about module nesting, now the maximum nesting depth is 2...

jamesmckinna:
Sorry I haven't returned to this for a while. I thought I might have time/brain-space to remove the nested module structure and rethink my approach generally, but then... that never happened. Thanks @MatthewDaggitt for the merge conflict resolution!

MatthewDaggitt:
I've had a hack at it to reduce the depth of nested modules to only 1 deep. As a result the file is now about ~25 lines shorter... @jamesmckinna what do you think?

jamesmckinna:
Suggested cosmetic renamings:
* `n` for `z` for quantifications over Nat in infinite sequences;
* `satisfiable`/`unsatisfiable` for `hold`/`notHold` (so the identifier correlates with the named definition in `Relation.*`)
* rectify the lemma names to follow #2168 / #2206 

Commit incoming...

MatthewDaggitt (approved):

Issue #2129, opened 2023-10-09
1.7.2 errors with Agda 2.6.4


juhp:
This may be my fault, but  I tried to rebuild the Fedora agda-stdlib 1.7.2 [package](https://src.fedoraproject.org/rpms/Agda-stdlib/blob/rawhide/f/Agda-stdlib.spec) with Agda 2.6.4 and it seems to fail:

```
Reason: It relies on injectivity of the data constructor +_, which
        is not yet supported

when checking the definition of neg-antimono-≤
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.2/src/Data/Rational/Properties.agda:726,1-732,97
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor -[1+_],
        which is not yet supported

when checking the definition of neg-antimono-≤
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.2/src/Data/Rational/Properties.agda:805,1-824,80
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor false,
        which is not yet supported

when checking the definition of Data.Rational.Properties.with-2764
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.2/src/Data/Rational/Properties.agda:805,1-824,80
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor true,
        which is not yet supported

when checking the definition of Data.Rational.Properties.with-2764
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.2/src/Data/Rational/Properties.agda:1029,1-1047,80
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor false,
        which is not yet supported

when checking the definition of Data.Rational.Properties.with-3110
[...]

PR #2132, opened 2023-10-10
release 1.7.3

andreasabel:
Branched off tag `v1.7.2` (branch `release-1.7.3-base`) to gather commits for `v1.7.3`.

TODO:
- [ ] Get CI to run here. Status: 
        * CI runs at https://github.com/andreasabel/agda-stdlib/actions
        * Deploys to: https://andreasabel.github.io/agda-stdlib/v1.7.3/
- [x] Implement missing points described in the [release guide](https://github.com/agda/agda-stdlib/blob/78e11bdee2eaaf7652c7be32d890821aeebe8781/notes/release-guide.txt)
- [x] Consider additions for forward-compatibility with `v2.0`: https://github.com/agda/agda-stdlib/issues/2129#issuecomment-1754700700
- [ ] `CHANGELOG/` is missing changelogs for `v1.7` and `v1.7.1`

Commits:
- Put indexed data types in the right universe
- Prepare for v1.7.3: move CHANGELOG to CHANGELOG/v1.7.2
- v1.7.3 CHANGELOG entry for `--large-indices` (#2030)
- Bump agda-stdlib-utils to 1.7.3; update its dependencies and CI
- Update CI for v1.7.3: Use Agda 2.6.4; deploy HTML to v1.7.3 subdirectory


.github/workflows/ci-ubuntu.yml
```diff
@@ -71,22 +65,10 @@ jobs:
...
-               || '${{ github.base_ref }}' == 'experimental' ]]; then
-            # Pick Agda version for experimental
-            echo "AGDA_COMMIT=9047e32a1b0cba98a299ed439a08d35bc4846f99" >> $GITHUB_ENV;
-            echo "AGDA_HTML_DIR=html/experimental" >> $GITHUB_ENV
-          fi
-
-          if [[ '${{ github.ref }}' == 'refs/heads/master' \
-             || '${{ github.ref }}' == 'refs/heads/experimental' ]]; then
-             echo "AGDA_DEPLOY=true" >> $GITHUB_ENV
-          fi
```
MatthewDaggitt, line 89: So I imagine that we'd be merging this branch back into `master` at some point. While a lot of the changes to this CI file are enhancements (e.g. bumping versions), can we avoid removing functionality like this? 

In particular it's important that if we're on the experimental branch, we install a different version of Agda!

.github/workflows/haskell-ci.yml
```diff
@@ -6,79 +6,144 @@
 #
 #   haskell-ci regenerate
 #
-# For more information, see https://github.com/haskell-CI/haskell-ci
+# For more information, see https://github.com/andreasabel/haskell-ci
```
MatthewDaggitt, line 9: Could you explain why we're switching to your own fork? Have to confess I'm a bit reluctant to, unless strictly necessarily. If it necessary, then documentation as to the reason should be added here.

MatthewDaggitt (commented):
Thanks for getting this off the ground!

>  Get CI to run: CI runs at https://github.com/andreasabel/agda-stdlib/actions

I'm afraid my expertise at CI wrangling pales in comparison to yours so I'm maybe you have better ideas why its failing than me? I have to confess I've never dug deeply into the CI code myself.

> Implement missing points described in the [release guide](https://github.com/agda/agda-stdlib/blob/78e11bdee2eaaf7652c7be32d890821aeebe8781/notes/release-guide.txt)

Happy to do this once we have the content thrashed out.

> Consider additions for forward-compatibility with v2.0: https://github.com/agda/agda-stdlib/issues/2129#issuecomment-1754700700

Could you make a separate PR onto this branch with your proposed changes?

> CHANGELOG/ is missing changelogs for v1.7 and v1.7.1

Hmm so they're present on `master`. Not quite sure why they never made it to this branch. I'll copy them from there.

MatthewDaggitt:
> > CHANGELOG/ is missing changelogs for v1.7 and v1.7.1
>
> Hmm so they're present on master. Not quite sure why they never made it to this branch. I'll copy them from there.

Actually I'm a bit confused about the setup for this whole PR. When I try to checkout `andreasabel/release-1.7.3` I end up in a detached head state, but it looks just like an ordinary branch...

.github/workflows/haskell-ci.yml
andreasabel: My fork uses the `ubuntu-20.04` image rather than the `ubuntu-18.04` image which allows it to use the latest version of `action/checkout`.  Otherwise, it is interchangeable, and you could as well regenerate the CI with the original `haskell-ci`.

andreasabel:
> Actually I'm a bit confused about the setup for this whole PR. When I try to checkout `andreasabel/release-1.7.3` I end up in a detached head state, but it looks just like an ordinary branch...

What do you get with `gh pr checkout 2132`?  (You might have to install the `gh` tool first.)

.github/workflows/ci-ubuntu.yml
MatthewDaggitt: Well on principle, I don't think we should be making any changes that we _don't_ want to merge back across. Furthermore you're making various improvements to the CI for example right? Bumping versions etc. it would nice if those were added back across. I guess we can cherry-pick them...

.github/workflows/haskell-ci.yml
MatthewDaggitt: And why do we need the latest version of `action/checkout`?

MatthewDaggitt:
> What do you get with gh pr checkout 2132? (You might have to install the gh tool first.)

Unfortunately I don't have `gh` installed. Do you have a non-`gh` based command?

MatthewDaggitt:
@andreasabel would you be free to briefly join the standard library dev ((https://meet.google.com/trh-fpbj-kfi)) meeting at https://time.is/1200_12_Oct_2023_in_UTC today?

.github/workflows/haskell-ci.yml
andreasabel: Not needed.  
But you might get harassed by dependabot if you don't use it...

If it matters to you, I can install the original haskell-ci and recreate this workflow.  But this is anyway just customary CI for 1.7.3 which won't end up in master.

andreasabel:
> @andreasabel would you be free to briefly join the standard library dev ((https://meet.google.com/trh-fpbj-kfi)) meeting at https://time.is/1200_12_Oct_2023_in_UTC today?

Yes, thanks for the invitation!  I will join, but have only like 50min because of another meeting.

.github/workflows/haskell-ci.yml
MatthewDaggitt: Yes, I guess that's my real question! Why are we trying to make a custom CI pass for v1.7.3?

andreasabel:
I added a commit with aliases for forward compatibility as proposed by @jamesmckinna  in:
- https://github.com/agda/agda-stdlib/issues/2124#issuecomment-1750318453

@MatthewDaggitt: I'll make a new PR that does not touch CI.

andreasabel:
Superseded by #2141 which does not make an attempt at running it through CI.

.github/workflows/ci-ubuntu.yml
andreasabel: The versions are already bumped in `master`, the last CI bumps are in:
- https://github.com/agda/agda-stdlib/pull/2142

.github/workflows/haskell-ci.yml
andreasabel: I don't want to release without having run CI.  But I now I am passing the bucket to you, so I make #2141 without CI changes.

Issue #2135, opened 2023-10-12
Make `Relation.Binary.Structures.IsStrictTotalOrder` consistent with the hierarchy


MatthewDaggitt:
While trying to polish off #2098 I've run into the issue that the structure `IsStrictTotalOrder` doesn't inherit directly from `IsStrictPartialOrder`. Instead it has flattened fields that try to avoid having to have the user re-prove irreflexivity.

This has various irritating effects including multiple non-identical copies of certain proofs such as `isStrictPartialOrder` in `StrictTotalOrder` bundle.

We did have exactly the same problem with lattices in https://github.com/agda/agda-stdlib/pull/1108, which we fixed by adding biased versions of the records that replicated the old structure. We should take advantage of `v2.0` and implement the same solution for `IsStrictTotalOrder`

jamesmckinna:
Interesting! 
I'm even more glad you took charge of #2098 , as I wouldn't have had the (historical) insights to raise, never mind tackle, this issue ;-)

PR #2137, opened 2023-10-12
Change definition of `IsStrictTotalOrder`

MatthewDaggitt:
Fixes https://github.com/agda/agda-stdlib/issues/2135 by defining `IsStrictTotalOrder` in terms of `IsStrictPartialOrder`

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -32,12 +32,7 @@ open import Level using (0ℓ)
 open import Relation.Unary as U using (Pred)
 open import Relation.Binary.Core
   using (_⇒_; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
-open import Relation.Binary.Bundles
-  using (DecSetoid; Preorder; TotalPreorder; Poset; TotalOrder; DecTotalOrder; StrictPartialOrder; StrictTotalOrder)
-open import Relation.Binary.Structures
-  using (IsDecEquivalence; IsPreorder; IsTotalPreorder; IsPartialOrder; IsTotalOrder; IsDecTotalOrder; IsStrictPartialOrder; IsStrictTotalOrder)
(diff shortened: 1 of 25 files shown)


src/Data/Nat/Properties.agda
```diff
@@ -32,12 +32,7 @@ open import Level using (0ℓ)
 open import Relation.Unary as U using (Pred)
 open import Relation.Binary.Core
   using (_⇒_; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
-open import Relation.Binary.Bundles
-  using (DecSetoid; Preorder; TotalPreorder; Poset; TotalOrder; DecTotalOrder; StrictPartialOrder; StrictTotalOrder)
-open import Relation.Binary.Structures
-  using (IsDecEquivalence; IsPreorder; IsTotalPreorder; IsPartialOrder; IsTotalOrder; IsDecTotalOrder; IsStrictPartialOrder; IsStrictTotalOrder)
-open import Relation.Binary.Definitions
-  using (DecidableEquality; Irrelevant; Reflexive; Antisymmetric; Transitive; Total; Decidable; Connex; Irreflexive; Asymmetric; LeftTrans; RightTrans; Trichotomous; tri≈; tri>; tri<; _Respects₂_)
+open import Relation.Binary
```
MatthewDaggitt, line 35: Just a heads up @JacquesCarette I think the dependency refactoring went a little too far here. I'm all for light dependencies, but I think there are times like these when it really makes sense to not be explicit about all imports and use the bundling modules. Do you agree? This isn't introducing any new dependencies apart from the collating module `Relation.Binary` itself.

jamesmckinna:
Apart from the above discussion point, I think this has the appearance, at least, of a very smooth transition, in particular in respect of the simplifications of the concrete client 'numeric' modules... so :+1: from me.

Issue #1736, opened 2022-03-11
Add `_<$>_` to new function hierarchy.


MatthewDaggitt:
Follow up from [#1727](https://github.com/agda/agda-stdlib/pull/1727) and https://github.com/agda/agda-stdlib/issues/1493

PR #2144, opened 2023-10-13
Add `_⟨$⟩_` operator for Function bundle

MatthewDaggitt:
Fixes https://github.com/agda/agda-stdlib/issues/1736

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3114,6 +3114,12 @@ Additions to existing modules
   execState : State s a → s → s
   ```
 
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

jamesmckinna:
Given that this seems in the end to boil down to delegation to a (deeply-nested) field name, I'm surprised it wasn't/isn't 'easy' to have it percolate up (down?) the whole `Function` hierarchy. But see previous discussions: not really my territory ;-)

PR #2152, opened 2023-10-15
Make reasoning modular by adding new `Reasoning.Syntax` module

### LINKED ISSUE #2146: Make reasoning syntax actually modular
While trying to fix #2145 and #1138 I've realised quite how much duplication has crept into the many many `Reasoning` modules we have in the library. Trying to change the syntax is incredibly painful as you have to do it all over the library.

I think a modular approach along the lines of:
```agda
module Relation.Binary.Reasoning.Base.Syntax
  {a ℓ} {A : Set a} (_R_ : Rel A ℓ) where

-- Propositional equality syntax

module ≡-step-syntax where

  -- Step with a non-trivial propositional equality
  step-≡ : ∀ x {y z} → y R z → x ≡ y → x R z
  step-≡ _ x∼z refl = x∼z

  -- Step with a definition identity
  step-≡-refl : ∀ (x : A) {y} → x R y → x R y
  step-≡-refl x xRy = xRy

  -- Step with a flipped non-trivial propositional equality
  step-≡-sym : ∀ x {y z} → y R z → y ≡ x → x R z
  step-≡-sym _ x∼z refl = x∼z

  infixr 2 step-≡ step-≡-refl step-≡-sym

  syntax step-≡-refl x xRy     = x ≡⟨⟩ xRy
  syntax step-≡-sym  x yRz y≡x = x ≡⟨ y≡x ⟨ yRz
  syntax step-≡      x yRz x≡y = x ≡⟨ x≡y ⟩ yRz

-- Setoid equality syntax

module ≈-syntax
  {ℓ₂} {_≈_ : Rel A ℓ₂}
  (sym : Symmetric _≈_)
  (step : Trans _≈_ _R_ _R_)
  where

  -- Step with the setoid equality
  step-≈ : ∀ (x : A) {y z} → y R z → x ≈ y → x R z
  step-≈ x yRz x≈y = step {x} x≈y yRz

  -- Flipped step with the setoid equality
  step-≈-sym : ∀ x {y z} → y R z → y ≈ x → x R z
  step-≈-sym x y∼z x≈y = step-≈ x y∼z (sym x≈y)

  syntax step-≈     x y∼z x≈y = x ≈⟨ x≈y ⟩ y∼z
  syntax step-≈-sym x y∼z y≈x = x ≈⟨ y≈x ⟨ y∼z
```
where we can then use these modules to mix and match whatever syntax we want for a particular `Reasoning` module.

I'm going to try and squeeze this in for `v2.0` as this will make the two issues above almost trivial to do.

MatthewDaggitt:
Fixes #2146. The main contribution of this to move as many reasoning syntax declarations as possible into the new module `Relation.Binary.Reasoning.Syntax`, whose parts are then picked as desired by each individual `Reasoning` module.

This has several advantages:
 1. Much easier to rename or add new syntax to when building custom `Reasoning` modules
 2. Centralises all the syntax declarations and documentation so that it is much easier to maintain.

See the list at the top of the `Syntax` module for the short list of `Reasoning` modules that for a variety of reasons aren't immediately amenable to this approach.

This change is backwards compatible as evidenced by the fact that no reasoning proofs have had to be changed. The interface for the `Reasoning` modules themselves also remains the same.

### DIFF
--- a/src/Relation/Binary/Reasoning/Syntax.agda
+++ b/src/Relation/Binary/Reasoning/Syntax.agda
@@ -1,0 +1,366 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The basic code for equational reasoning with a single relation
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Level using (Level; _⊔_; suc)
+open import Relation.Nullary.Decidable.Core
+  using (Dec; True; toWitness)
+open import Relation.Nullary.Negation using (contradiction)
+open import Relation.Binary.Core using (Rel; REL; _⇒_)
+open import Relation.Binary.Definitions
+open import Relation.Binary.PropositionalEquality.Core as P
+  using (_≡_)
+
+-- List of `Reasoning` modules that do not use this framework and so
+-- need to be updated manually if the syntax changes.
+--
+--   Data/Vec/Relation/Binary/Equality/Cast
+--   Relation/Binary/HeterogeneousEquality
+--   Effect/Monad/Partiality
+--   Effect/Monad/Partiality/All
+--   Relation/Binary/PropositionalEquality/Core
+--   Codata/Guarded/Stream/Relation/Binary/Pointwise
+--   Codata/Sized/Stream/Bisimilarity
+--   Function/Reasoning
+ 
+module Relation.Binary.Reasoning.Syntax where
+
+private
+  variable
+    a ℓ₁ ℓ₂ ℓ₃ ℓ₄ : Level
+    A B C : Set a
+    x y z : A
+
+------------------------------------------------------------------------
+-- Syntax for beginning a reasoning chain
+------------------------------------------------------------------------
+
+------------------------------------------------------------------------
+-- Basic begin syntax
+
+module begin-syntax
+  (R : REL A B ℓ₁)
+  {S : REL A B ℓ₂}
+  (reflexive : R ⇒ S)
+  where
+
+  infix 1 begin_
+
+  begin_ : R x y → S x y
+  begin_ = reflexive
+
+------------------------------------------------------------------------
+-- Begin subrelation syntax
+
+-- Sometimes we want to support sub-relations with the
+-- same reasoning operators as the main relations (e.g. perform equality
+-- proofs with non-strict reasoning operators). This record bundles all
+-- the parts needed to extract the sub-relation proofs.
+record SubRelation {A : Set a} (R : Rel A ℓ₁) ℓ₂ ℓ₃ : Set (a ⊔ ℓ₁ ⊔ suc ℓ₂ ⊔ suc ℓ₃) where
+  field
+    S : Rel A ℓ₂
+    IsS : R x y → Set ℓ₃
+    IsS? : ∀ (xRy : R x y) → Dec (IsS xRy)
+    extract : ∀ {xRy : R x y} → IsS xRy → S x y
+
+module begin-subrelation-syntax
+  (R : Rel A ℓ₁)
+  (sub : SubRelation R ℓ₂ ℓ₃)
+  where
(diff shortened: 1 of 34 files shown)


jamesmckinna:
Bravo! (modulo the wicked-looking typechecking failures... is this behaviour new to 2.6.4?)

src/Relation/Binary/Reasoning/Syntax.agda
```diff
@@ -0,0 +1,366 @@
...
+      syntax step-⤖ x yRz x⤖y = x ⤖⟨ x⤖y ⟩ yRz
+      syntax step-⬻ x yRz y⤖x = x ⬻⟨ y⤖x ⟩ yRz
+
+
+    -- Inverse syntax
+    module ↔-syntax where
+      infixr 2 step-↔ step-↔-sym
+      step-↔ = forward
+      step-↔-sym = backward
+      syntax step-↔ x yRz x⤖y = x ↔⟨ x⤖y ⟩ yRz
```
JacquesCarette, line 305: Perhaps a bit too much cut-and-paste, here and below, in the name of the 3rd argument? If this ever leaks out (say in the names that Agda suggests), it might make things harder for the users to understand.

JacquesCarette (changes requested):
This is a superb change! Worth sneaking it in under the wire. I spotted one tiny thing that should be polished though.

src/Relation/Binary/Reasoning/Syntax.agda
MatthewDaggitt: Nice spot, fixed!

MatthewDaggitt:
> And what is weird is that the 'obvious' instantiations of the _ terms supplied to ℤ.*-cancelʳ-≡ on line 855 (identified as the source of the error) of Data.Rational.Properties, namely ↥ᵘ (toℚᵘ (p + q)) and ↧+ᵘ p q are not enough to nudge the typechecker into accepting the whole proof. Very weird indeed...

Yup, so the chain of events was:
1. The definition of the `≃` combinators in `Data.Rational.Properties.≤-Reasoning` were buggy as they actually accepted proofs of type `≡`.
2. Fixing this required proving that `≃` is `Symmetric`
3. Which revealed that the definition of `≃` suffers the same eager reducing performance problems, previously fixed, for the rational operations.
4. Which fixing caused the unsolved metas.

So thankfully this isn't actually a side effect of the reasoning refactor, it was just a knock on consequence that the definition of `≃` needed to be fixed :relieved:

jamesmckinna:
Wow!? Great summary of a wicked problem. Very interesting, and thanks for taking the time to hunt it down and document the solution. Fascinating.

src/Relation/Binary/Reasoning/Syntax.agda
```diff
@@ -0,0 +1,366 @@
...
+    backward : ∀ x {y z} → S y z → U y x → T x z
+    backward x yRz x≈y = forward x yRz (sym x≈y)
+
+    -- Setoid equality syntax
+    module ≈-syntax where
+      infixr 2 step-≈ step-≈˘
+      step-≈ = forward
+      step-≈˘ = backward
+      syntax step-≈  x yRz x≈y = x ≈⟨ x≈y ⟩ yRz
+      syntax step-≈˘ x yRz y≈x = x ≈˘⟨ y≈x ⟩ yRz
```
jamesmckinna, line 261: Coming, as ever, late to this: in #2268 I've had to use explicit `x ≈⟨ refl ⟩...` steps, because there is no abbreviation `≈⟨⟩` declared, whereas `≡⟨⟩` *is* available as an alternative. Was this what you intended, or should there be a bug-fixing issue/PR raised to add the additional syntax?

MatthewDaggitt: Yes, the intent is that people should use `≡⟨⟩` not `≈⟨⟩`

Issue #1138, opened 2020-03-31
Fix the syntax for symmetric equality reasoning combinator


ajrouvoet:
Quoting @MatthewDaggitt 

> I too would like to do something about the ˘ syntax, even though I introduced it 
> I really hate it: it's difficult to type, it ruins the alignment of proofs and doesn't clearly convey symmetry to me. 

Here are a few proposals to get the discussion going:

1. `_↑≡⟨_⟩_` and `_↓≡⟨_⟩_`, stolen from the categories library (at least they used to be there when I used it way back). The arrow indicates the direction in which we are applying the equality.
2. `_↑⟨_⟩_` and `_↓⟨_⟩_` as shorter alternatives.

JacquesCarette:
I like the shorter alternatives. If we have to put in the relation wrt which we're working, I'm wondering about putting it in the other order, i.e. `_≡↑⟨_⟩_`.

I think having these as aliases (or the current ones as aliases) could be an interesting experiment. Let a couple of versions go by, and see what people tend to use the most, then deprecate the other.

dylanede:
+1 for putting the direction indicator after the relation symbol.

Regarding the direction indicator, would horizontal arrows (of some style) make more sense or be more likely to be confused with the many other uses of horizontal arrows?

ajrouvoet:
Equality reasoning proves are usually code-set vertically. :man_shrugging:

MatthewDaggitt:
I like the idea of vertical arrows. My only comment is that I also like having the angle brackets aligned on the page, which is one of things that I find very distressing about the current sym notation. This proposal would partially fix it as the sym and non-sym-ed versions would now align, but it still wouldn't align with other operators, e.g. inequalities:
```agda
begin
  (m / n) * n          ≤⟨ ... ⟩
  (m / n) * n + m % n  ≡↑⟨ ... ⟩
  m % n + (m / n) * n  ≡↓⟨ ... ⟩
  m                    ∎
```

Would people hate it if we moved the arrows inside the brackets? i.e.  `_≡⟨↓_⟩_` and `_≡⟨↑_⟩_`
```agda
begin
  (m / n) * n          ≤⟨ ... ⟩
  (m / n) * n + m % n  ≡⟨↑ ... ⟩
  m % n + (m / n) * n  ≡⟨↓ ... ⟩
  m                    ∎
```

WolframKahl:
Yes.        :wink:

[...]

PR #2159, opened 2023-10-17
Move ≡-Reasoning from Core to Properties and implement using syntax

MatthewDaggitt:
This follows up #2152 and implements `≡-Reasoning` using the new reasoning combinators. In order to do this it was necessary to move it from `Core` to `Properties` as the `Syntax` module imports `Core`.

This is a reasonable change to make as:
 1. all other `Reasoning` modules live in `Properties` not `Base/Core`
 2. users are told not to import `Core` modules anyway (have added the `breaking` tag anyway)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1238,6 +1238,12 @@ Major improvements
   pre-existing `Reasoning` module in just a couple of lines
   (e.g. see `∣-Reasoning` in `Data.Nat.Divisibility`)
 
+* One pre-requisite for that is that `≡-Reasoning` has been moved from
+  `Relation.Binary.PropositionalEquality.Core` (which shouldn't be
+  imported anyway as it's a `Core` module) to 
+  `Relation.Binary.PropositionalEquality.Properties`.
+  It is still exported by `Relation.Binary.PropositionalEquality`.
+  
 Deprecated modules
 ------------------
 
--- a/src/Codata/Guarded/Stream/Properties.agda
+++ b/src/Codata/Guarded/Stream/Properties.agda
@@ -21,6 +21,8 @@ open import Data.Vec.Base as Vec using (Vec; _∷_)
 open import Function.Base using (const; flip; id; _∘′_; _$′_; _⟨_⟩_; _∘₂′_)
 open import Level using (Level)
 open import Relation.Binary.PropositionalEquality.Core as P using (_≡_; cong; cong₂)
+open import Relation.Binary.PropositionalEquality.Properties
+  using (module ≡-Reasoning)
 
 private
   variable
@@ -218,7 +220,7 @@ lookup-transpose n (as ∷ ass) = begin
   lookup as n ∷ lookup (transpose ass) n     ≡⟨ cong (lookup as n ∷_) (lookup-transpose n ass) ⟩
   lookup as n ∷ List.map (flip lookup n) ass ≡⟨⟩
   List.map (flip lookup n) (as ∷ ass)        ∎
-  where open P.≡-Reasoning
+  where open ≡-Reasoning
 
 lookup-transpose⁺ : ∀ n (ass : List⁺ (Stream A)) →
                     lookup (transpose⁺ ass) n ≡ List⁺.map (flip lookup n) ass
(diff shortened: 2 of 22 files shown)


jamesmckinna:
Re:2 Users may be told this, but recent refactorings have been fairly aggressive in tightening imports, so there could be a bunch of donkey work to do. But definitely worth it for the uniformity you're in the process of achieving.

MatthewDaggitt:
But user's haven't been tightening their code? Only modules in the standard library, so yes, it did require a bit of work (hence touching 22 files), but it wasn't ridiculous.

jamesmckinna:
> This follows up #2152 and implements `≡-Reasoning` using the new reasoning combinators. In order to do this it was necessary to move it from `Core` to `Properties` as the `Syntax` module imports `Core`.

Just a quick thought: in similar situations elsewhere in the library (eg in the #2149 refactoring of `Relation.Binary.Reflects`, emulating usage elsewhere - see below), one can replace appeals to 
```agda
open import Relation.Binary.PropositionalEquality.Core using (_≡_)
```
with (suitably qualified/renaming) import of `Agda.Builtin.Equality`, as of writing in the following modules:
```
Data/Float/Base.agda: open import Agda.Builtin.Equality
Data/List/Relation/Binary/Sublist/DecPropositional.agda: open import Agda.Builtin.Equality us
ing (_≡_)
Data/Sum.agda: open import Agda.Builtin.Equality
Data/Sum/Effectful/Examples.agda: open import Agda.Builtin.Equality
Data/W.agda: open import Agda.Builtin.Equality using (_≡_; refl)
Data/W/WithK.agda: open import Agda.Builtin.Equality
Debug/Trace.agda: open import Agda.Builtin.Equality
Function/Endomorphism/Setoid.agda: open import Agda.Builtin.Equality
Function/Strict.agda: open import Agda.Builtin.Equality using (_≡_)
Relation/Binary/Definitions.agda: open import Agda.Builtin.Equality using (_≡_)
Relation/Binary/PropositionalEquality/TrustMe.agda: open import Agda.Builtin.TrustMe
Relation/Binary/Rewriting.agda:open import Agda.Builtin.Equality using (_≡_ ; refl)
Relation/Nullary.agda:open import Agda.Builtin.Equality
Relation/Nullary/Reflects.agda:open import Agda.Builtin.Equality
Text/Tabular/Base.agda:open import Agda.Builtin.Equality
```

jamesmckinna:
So you could (consider a) refactor of #2152 and this PR to exploit such a technique?

MatthewDaggitt:
> So you could (consider a) refactor of https://github.com/agda/agda-stdlib/pull/2152 and this PR to exploit such a technique?

Yup it's a great thought and I did consider it as well! Unfortunately the propositional equational reasoning combinators require a proof `Symmetry` for propositional equality which definition belongs in `Core` :cry: 

I thought about duplicating or inlining it, but then when I thought about all the other `Reasoning` modules in `Properties` files I thought that maybe this was the more principled and consistent method of getting rid of the dependency cyle.

jamesmckinna:
Ah well... it 's was worth a try.
On which... somewhat/particularly bikeshedding/rabbit-hole point... it occurs to me that most, if not all the machinery you have developed could simply be parametrised on `sym` and `trans`, because to have an *non*-trivial equation `x ≈ y` of whatever character, provided `x` or `y` participate in any equation whatsoever, then we have `x ≈ x` and `y ≈ y`, because PERs are reflexive on their domains. That is, all the *actual* proofs of `refl` in any Feijen-style derivation could be computed, at least at `$\qed$` time. Leaving `_≡⟨⟩_` for actual *definitional* steps?

jamesmckinna:
Re: the second point
No, I wasn't suggesting that `Preorder`-based reasoning use symmetry (which won't *a priori* hold, indeed if it does, we're already in a (partial) equivalence relation)...
... rather I was 'thinking out loud' (cf. the parenthesisation discussion on #1138 ) about 
* `begin <exp>` being one sort of opening parenthesis, with
* `<exp> \qed` being its closing parenthesis (so I think this is strictly a breaking change from the current `syntax` directive for $\qed$ being a proxy for `refl`

... more to follow, phone needs charging! ;-)

MatthewDaggitt:
> ... more to follow, phone needs charging! ;-)

Okay sounds like it's ideas about the general approach! For now, I'll merge this in.

PR #2160, opened 2023-10-17
Rename symmetric reasoning combinators. Minimal set of changes

MatthewDaggitt:
This is the minimal set of changes needed to fix https://github.com/agda/agda-stdlib/issues/1138

Because of https://github.com/agda/agda/issues/5617 I haven't had to rename all the uses of the old symmetric combinators. I've only renamed those, that I haven't had to deprecate because they were introduced post `v1.7`. 

The renaming should be done before merging. I've put this up now, because these changes are the "interesting" bit.


MatthewDaggitt:
Okay pushed the changes and the regex needed to do them. One tiny wrinkle was that the new names interacted badly with `_⟨_⟩_` from `Function.Base` which was used in a single module in `Algebra.Lattice.BooleanAlgebra.Properties`. However, I think that's an acceptable price to pay.

jamesmckinna:
Agree re the symbol clash with `Function.Base`... possible deprecation opportunity there? I always have to lookup what those custom symbols are supposed to denote whenever I encounter them.

MatthewDaggitt:
> Agree re the symbol clash with Function.Base... possible deprecation opportunity there?

Deprecate the `_⟨_⟩_` symbol? Do you have a suitable replacement name?

jamesmckinna:
> > Agree re the symbol clash with Function.Base... possible deprecation opportunity there?
> 
> Deprecate the `_⟨_⟩_` symbol? Do you have a suitable replacement name in mind?

Er, no. Purely irresponsible out-loud 'thinking'/daydreaming... sorry!

gallais:
> Out of interest, why do the new names interact badly with that symbol, where the previous ones did not?

With the previous notation `⟨`  would never appear alone because there would always be an equality character before it. Now it can show up as the last parens of a symmetric step.

jamesmckinna:
D'oh!

PR #2162, opened 2023-10-18
Added remaining flipped and negated relations to binary relation bundles

### LINKED ISSUE #2098: Add `_≥_`, `_>_`, `_≱_` and `_≯_` systematically to the `Relation.Binary.Bundles` hierarchy
See #1214  #2095 #2096 and #2099 for further discussion.

There is a design issue at stake here, however: do we use
* defined symbols (the solution adopted in #2095 , #2099)
* `syntax` declarations?

pro the first is that we can be more selective about `import`s, `renaming`, `hiding` etc. (needed in those PRs)
pro the second is that it is 'more lightweight', yet somehow more irrevocable.

My current thinking is: defined symbols

UPDATED: ~~one~~ two reasons I may have hesitated to pursue this already to a PR might have been:
* concern about the reception for #2099 (merged) and #2095 (not yet) before proceeding
* more serious: given the proliferation of hand-rolled flipped relations throughout the library, the amount of weeding out that might be required, in order to avoid clashes between the hand-rolled and the machine-made, (still!) seems a bit daunting...

MatthewDaggitt:
Fixes #2098.

Also went through and systematically fixed all the missing import gaps in the `Bundles` hierarchy.


CHANGELOG.md
```diff
@@ -849,18 +849,24 @@ Non-backwards compatible changes
...
-  has been achieved by redeclaring a deprecated version of the old name in the record.
-  Therefore, only _declarations_ of `PartialOrder` records will need their field names
-  updating.
+### (Issue #2096) Introduction of flipped and negated relation symbols to bundles in `Relation.Binary.Bundles`
+
+* Previously, bundles such as `Preorder`, `Poset`, `TotalOrder` etc. did not have the flipped 
+  and negated versions of the operators exposed. In some cases they could obtained by opening the
+  relevant `Relation.Binary.Properties.X` file but usually they had to be redefined every time.
+  
+* To fix this, these bundles now all export all 4 versons of the operator: normal, converse, negated,
```
jamesmckinna, line 858: 'version'

jamesmckinna:
What consequence are there downstream for, eg `Data.Nat.Base` where converse relations in particular are systematically introduced?

MatthewDaggitt:
> What consequence are there downstream for, eg Data.Nat.Base where converse relations in particular are systematically introduced?

Nothing in the library modules. We don't open the bundles publicly anywhere, just construct them. In general you only open bundles when they're abstract, i.e. you don't know what type they are. Where it is going to break user code is where they defined those abstract flipped operators themselves. But that's the point, they can then delete their own definitions.

MatthewDaggitt:
> It all looks good, but I think I need to take on trust the various modifications to the public exports from substructures, which seem to have shifted from explicit naming of what gets exported, in favour of explicit naming of what gets hidden.

Yup, so actually we now do both. I think I've actually come up with a general algorithm for doing so which I'll add to the documentation. 

So imagine we have the following general scenario where bundle `A` is a direct refinement of  bundle `C` but is also morally a refinement of bundles `B` and `D`.
```
  Structures               Bundles
  ==========               =======
      IsA                     A
   /  ||   \               /  ||  \
IsB   IsC   IsD          B    C    D
```
The procedure for re-exports in the bundles is as follows:
1. `open IsA isA public using (IsC, M)` where `M` is everything exported by `IsA` that is not exported by `IsC`.
2. construct `c : C` via the `isC` obtained in step 1.
3. `open C c public hiding (N)` where `N` is the list of fields shared by both `A` and `C`.
4. construct `b : B` via the `isB` obtained in step 1.
5. `open B b public using (O)` where `O` is everything exported by `B` but not exported by `IsA`
6. construct `d : D` via the `isC` obtained in step 1.
7. `open D d public using (P)` where `P` is everything exported by `D` but not exported by `IsA`

jamesmckinna:
Bravo! Thanks for taking the time to develop this documentation. Very illuminating.

PR #2164, opened 2023-10-18
Restore 'return' as an alias for 'pure'

MatthewDaggitt:
Fixes https://github.com/agda/agda-stdlib/issues/2124. 

@gallais is there anywhere else you're aware of it needs to go?


MatthewDaggitt:
Okay this addresses Andreas' concerns for now, so will merge in in the interests of getting a release candidate out the door. Feel free to comment if I've missed something @gallais

Issue #2180, opened 2023-10-23
(re-)Definition of `Prime` (and `Irreducible`) in `Data.Nat.Primality`


jamesmckinna:
Reflecting on @Taneb 's PR #1969 , and wanting to add the definition of `Irreducible` to `Data.Nat.Primality`, with 
```agda
IrreducibleAt : ℕ → Pred ℕ _
IrreducibleAt n m = m ∣ n → m ≡ 1 ⊎ m ≡ n

Irreducible : ℕ → Set
Irreducible n = ∀[ IrreducibleAt n ]
```
I found myself wanting to redefine `Composite` and `Prime` to eliminate the special cases for `0` and `1` which otherwise pollute the definitions and properties given there.

This leads to the following definitions [REVISED in PR #2181 but not updated here]:
[DELETED: please see #2181 for heavily revised/redacted versions]
~~with this last pattern synonym essentially permitting proofs formerly involving `Prime` to access the second functional component without difficulty, and moreover to enforce that the argument to `Prime` *must* be of the form `suc (suc _)`, and thus eliminate some now-redundant pattern matching against patterns of that form.~~ with a constructor `prime` introducing witnesses to `Prime p` imposing that `p` must be `NonTrivial` ie `> 1` (and much else besides ;-))

I have a revised version of `Data.Nat.Primality` based on these re-definitions, towards a _breaking_ PR. So I'm cheekily tagging this as `v2.0` in order to try to sneak it into the upcoming release.

jamesmckinna:
Above revised in the light of comments on #2181 as a potential refactoring of the `Rough` component of #1969 .

jamesmckinna:
On the PR #2181 I wrote:
> The main issue as I see it ... is the `label:breaking` change to the definition of `Prime`, but I think in view of @Taneb 's #1969, plus the fact that `Prime` has already once been redefined for v2.0, that this is a last chance for a while to sort this out... and obvs. IMHO, better than before. But it is a slightly 'fiddly' definition, so I have taken care to introduced enough pattern synonyms to (hopefully!) smooth that path.

Suggest we continue the discussion aspects of this issue/PR here, with implementation details/review comments on the code on the PR.

Taneb:
Hmm, I wonder if we could go even more (semi)-ring theoretical.

* instead of using `1 < d`, we could write something like `NonZero d × NonInvertible d`
* We could make a definition of "strict" divisibility, which includes a not-equal. This implies `_<_`.

[...]

Issue #2092, opened 2023-09-14
Irrelevant instance predicate `NonNull` for `List`, `Vec`, ...


jamesmckinna:
Proposal: we already have in `Data.List.Base`
```agda
null : List A → Bool
null [] = true
null _ = false
```
so is it time to add (cf. #1538 etc.)
```agda
NonNull : List A → Set _
NonNull = T ∘ not ∘ null
```
etc. 
with eg
```agda
instance
  nonNull⇒nonZero[length] : {{ _ : NonNull xs }} → NonZero (length xs)
  nonNull⇒nonZero[length] {xs = _ ∷ _} = _
```
etc.
together with some structure proxying for the type class for having a `null` method and a `NonNull` predicate... etc.

jamesmckinna:
Re: #1538 and the choice of a `record`-based definition to avoid explicit import of `Data.Unit`:
the only reason to explicitly import would be to have the name `tt` in scope...
... but AFAIU, Agda will happily do (forced) argument inference on any sub-term of type `⊤` without ever having to explicitly mention either `⊤` or `tt`, because of `eta` for records... so `_` ALWAYS suffices?

I think that's why I used vanilla `T` in #2000, and in the above fragments, because instances will likewise ALWAYS be given by `_`... so no `import`s necessary, and the error messages don't really seem to be better for having the `record` wrapper (or are they?). Is this understanding correct?

jamesmckinna:
See [the `list-refinement` branch](https://github.com/jamesmckinna/agda-stdlib/tree/list-refinement) for an initial sketch. Not quite the panacea I'd hoped, even for `scanr`...

jamesmckinna:
Six months on, and having closed #2260 already, I'll close this for lack of any positive take-up.

PR #2181, opened 2023-10-23
definition of `Irreducible` and `Rough`; refactoring of `Prime` and `Composite` cf. #2180

jamesmckinna:
DONE:
- [x] (better!) names of things?
- [x] `CHANGELOG`
- [x] refactoring to demand all the `1 < n` etc. preconditions be given by instance arguments/explicit `2+_` patterns (see [my response](https://github.com/agda/agda-stdlib/issues/2180#issuecomment-1783777598) to @Taneb 's [earlier comment on the issue](https://github.com/agda/agda-stdlib/issues/2180#issuecomment-1776737729)
- [x] move the relevant 'prelude' definitions `NonTrivial` etc. to `Data.Nat.Base`...
- [x] proof `irreducible?` of `Decidable Irreducible`
- [x] reconciliation of `PrimeAt` and @Taneb 's [definition of `_Rough_`](https://github.com/agda/agda-stdlib/blob/7482c463b5295eed435e1cc99367f162aff1a012/src/Data/Nat/Primality/Rough.agda#L28)
- [x] proof `1-rough` that every number is `1 Rough`
- [x] proof `0-rough` that every number is `0 Rough`
- [x] `Rough` is a *bounded* predicate `.{{ NonTrivial n }} → k Rough n → k ≤ n`

Plus knock-on consequences for:
- [x]  `Coprimality`: UPDATED removed the coupling on any actual representation of `Prime`, in favour of using the new proof `prime⇒irreducible`

Also:
- [x] fixities for these things? #207 ~~(suggestions welcome!) I decided in the interests of simplicity to remove all the previous infix definitions of the various relations and their pointwise instances in favour of a flatter more streamlined structure. perhaps some `syntax` might be introduced to restore the infix style, but I think it's not really necessary any more...?~~

Discussion (on the issue  #2180 ) /review (of the code, here) very welcome!

The main issue as I see it  ... is the `label:breaking` change to the definition of `Prime`, but I think in view of @Taneb 's #1969, plus the fact that `Prime` has already once been redefined for v2.0, that this is a last chance for a while to sort this out... and obvs. IMHO, better than before. But it is a slightly 'fiddly' definition, so I have taken care to introduced enough pattern synonyms to (hopefully!) smooth that path.


src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,140 @@
...
+    n p : ℕ
+
 
 ------------------------------------------------------------------------
 -- Definitions
 
 -- Definition of compositeness
 
+CompositeAt : ℕ → Pred ℕ _
+CompositeAt n d = 1 < d × d < n × d ∣ n
```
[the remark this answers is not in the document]
JacquesCarette: I strongly second this suggestion.

Tuples are the assembly-level of dependently-typed programming.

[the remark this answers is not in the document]
jamesmckinna: Good question! I suppose I was doing the minimum necessary to reflect the existing definitions of `Composite` and `Prime`, with its iterated-product design, rather than using a custom record... so AFAIAC, it's up for discussion ;-)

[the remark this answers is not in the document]
jamesmckinna: Aaaargh... it turns out to be really quite painful to refactor along those lines... persevering, though :-(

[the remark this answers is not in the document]
jamesmckinna: With an eventual simplification of @Taneb 's lemma relating `Rough` and `Prime`...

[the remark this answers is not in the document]
jamesmckinna: It homes in on the *complement* of `Rough` as driving the whole development, so possible refactorings include:
* what name to give to this concept: `Smooth` is idiotic, except as the complement of `Rough`!?
* a lot of the section on implications between the concepts consists almost entirely of formalistic intuitionistically-valid deMorgan-dual relations between `Satisfiable` and `(I)Universal`, so could/should properly move to an as-yet underdeveloped ecosystem for `Unary` predicates;
* ...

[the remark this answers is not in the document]
Taneb: `Smooth` is a term of the art, and a much more common one than `Rough` too, indicating a number all of whose prime factors are below a bound. Does this match your complement of `Rough`?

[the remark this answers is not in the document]
jamesmckinna: Oh? I had had no idea... perhaps the name should stay as it is. I thought that `BoundedComposite` was a bit more 'thinkable' (to use Andreas' phrase), and NB I say nothing about the possible factors `d` in that definition being *prime*... that a downstream (possibly implicit?) strengthening which you exploit in #1969 ?

src/Data/Nat/Coprimality.agda
```diff
@@ -139,11 +138,14 @@ coprime-factors c (divides q₁ eq₁ , divides q₂ eq₂) with coprime-Bézout
...
+
 prime⇒coprime : ∀ m → Prime m →
                 ∀ n → 0 < n → n < m → Coprime m n
-prime⇒coprime (suc (suc _)) p _ _ _ {0} (0∣m , _) =
-  contradiction (0∣⇒≡0 0∣m) λ()
-prime⇒coprime (suc (suc _)) _ _ _ _ {1} _         = refl
-prime⇒coprime (suc (suc _)) p (suc _) _ n<m {(suc (suc _))} (d∣m , d∣n) =
-  contradiction d∣m (p 2≤d d<m)
-  where 2≤d = s≤s (s≤s z≤n); d<m = <-≤-trans (s≤s (∣⇒≤ d∣n)) n<m
+prime⇒coprime _ p _ z<s = prime⇒coprime≢0 p
```
jamesmckinna, line 150: Now, there's a possible deprecation opportunity?

CHANGELOG.md
```diff
@@ -2753,6 +2753,11 @@ Additions to existing modules
   <-asym : Asymmetric _<_
   ```
 
+* Added a new definition to `Data.Nat.Coprimality`:
+  ```agda
+  prime⇒coprime≢0 : Prime p → .{{_ : NonZero n}} → n < p → Coprime p n
```
jamesmckinna, line 2758: There's an/the obvious question about what best to call this lemma?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+
+pattern 2+ n = suc (suc n)
+
+trivial : ℕ → Bool
+trivial 0      = true
+trivial 1      = true
+trivial (2+ _) = false
+
+Unit NonUnit NonTrivial : Pred ℕ _
+Unit       = T ∘ (_≡ᵇ 1)
```
jamesmckinna, line 41: I'm in two minds as to whether this belongs here, but in a sense it is leaving a future hook for the general ring-theoretic definition of being a `Unit`... but perhaps that should wait for a subsequent PR?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+nonUnit⇒≢1 ⦃()⦄ refl
+instance
+  nonTrivial : NonTrivial (2+ n)
+  nonTrivial = _
+
+nonTrivial⇒≢1 : .{{NonTrivial n}} → n ≢ 1
+nonTrivial⇒≢1 ⦃()⦄ refl
+
+nonTrivial⇒nonZero : .{{NonTrivial n}} → NonZero n
+nonTrivial⇒nonZero {n = 2+ k} = _
```
jamesmckinna, line 62: A thing I don't understand is that my instinct woud have been to make this an `instance` declaration... but that appears to cause some subsequent proofs to *fail* (I think; but running these things interactively doesn't always give accurate feedback, again seemingly because of hidden cached state in the editor/proof engine; but maybe that's faulty rationalisation on my part?)

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+
+pattern 1<2+n {n} = s<s (z<s {n})
+
+nonTrivial⇒n>1 : .{{NonTrivial n}} → 1 < n
+nonTrivial⇒n>1 {n = 2+ _} = 1<2+n
+
+n>1⇒nonTrivial : 1 < n → NonTrivial n
+n>1⇒nonTrivial 1<2+n = _
```
jamesmckinna, line 70: Again: naming. The model of `≢-nonZero{⁻¹}` doesn't quite seem to fit here?

jamesmckinna: Also: can we make the assumption `1 < n` in the second lemma into an instance? Should we?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+    d<k : d < k
+    d∣n : d ∣ n
+
+-- smart constructors
+
+boundedComposite≢ : .{{NonZero n}} → {{NonTrivial d}} →
+                    d ≢ n → d ∣ n → BoundedComposite n n d
+boundedComposite≢ d≢n d∣n = boundedComposite (≤∧≢⇒< (∣⇒≤ d∣n) d≢n) d∣n
+
+boundedComposite>1 : 1 < d → d < n → d ∣ n → BoundedComposite n n d
```
jamesmckinna, line 91: Ditto. re `instance`s

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
-Composite 1 = ⊥
-Composite n = ∃ λ d → 2 ≤ d × d < n × d ∣ n
+Composite : Pred ℕ _
+Composite n = ∃⟨ BoundedComposite n n ⟩
+
+-- smart constructor
+
+composite : .{{NonZero n}} → {{NonTrivial d}} →
+            d ≢ n → d ∣ n → Composite n
+composite {d = d} d≢n d∣n = d , boundedComposite≢ d≢n d∣n
```
jamesmckinna, line 103: Irritatingly, the example proof below of `Composite 4` seems to suggest that `d` here should be explicitly, not implicitly, quantified?

src/Data/Nat/Primality.agda
```diff
@@ -106,7 +317,8 @@ euclidsLemma m n {p@(suc (suc _))} p-prime p∣m*n = result
   result with Bézout.lemma m p
   -- if the GCD of m and p is zero then p must be zero, which is
   -- impossible as p is a prime
-  ... | Bézout.result 0 g _ = contradiction (0∣⇒≡0 (GCD.gcd∣n g)) λ()
+  ... | Bézout.result 0 g _ = contradiction (0∣⇒≡0 (GCD.gcd∣n g)) (≢-nonZero⁻¹ _)
+  -- this should be a typechecker-rejectable case!?
```
jamesmckinna, line 321: Two observations:
* the `contradiction` argument basically reprises that on L285 above, so can we isolate a common abstraction?
* I couldn't (easily) find a way for the in stance on L307 above to rule out this case from even occurring, by unification failure in the typechecker; but I think it *ought* to be possible?

src/Data/Nat/Primality.agda
```diff
@@ -141,3 +355,4 @@ private
   -- Example: 6 is composite
   6-is-composite : Composite 6
   6-is-composite = from-yes (composite? 6)
+
```
jamesmckinna, line 358: A spurious blank line; should probably be removed?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
 
 module Data.Nat.Primality where
 
-open import Data.Empty using (⊥)
+open import Data.Bool.Base using (Bool; true; false; not; T)
```
jamesmckinna, line 11: Provisional; until the 'prelude' definitions of `{Non}Unit`, `NonTrivial` get moved to `Data.Nat.Base`...

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
-composite? 0               = no λ()
-composite? 1               = no λ()
-composite? n@(suc (suc _)) = Dec.map′
-  (map₂ λ { (a , b , c) → (b , a , c)})
-  (map₂ λ { (a , b , c) → (b , a , c)})
-  (anyUpTo? (λ d → 2 ≤? d ×-dec d ∣? n) n)
+composite? n = Dec.map′
+  (map₂ λ (d<n , 1<d , d∣n) → boundedComposite>1 1<d d<n d∣n)
+  (map₂ λ (boundedComposite d<n d∣n) → d<n , nonTrivial⇒n>1 , d∣n)
+  (anyUpTo? (λ d → 1 <? d ×-dec d ∣? n) n)
```
MatthewDaggitt, line 242: Might be more natural using a new `anyBetween?` proof? Then you could write it as `anyBetween _∣?_ 2 n`?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+-- NonZero
+
+composite⇒nonZero : Composite n → NonZero n
+composite⇒nonZero {suc _} _ = _
+
+prime⇒nonZero : Prime p → NonZero p
+prime⇒nonZero (prime _) = nonTrivial⇒nonZero
+
+irreducible⇒nonZero : Irreducible n → NonZero n
+irreducible⇒nonZero {zero} = flip contradiction ¬irreducible[0]
```
MatthewDaggitt, line 231: Same comment as `composite⇒nonTrivial`

src/Data/Nat/Primality.agda
```diff
@@ -91,9 +301,10 @@ prime⇒¬composite {n@(suc (suc _))} n-prime (d , 2≤d , d<n , d∣n) =
 -- the ring theoretic definition of a prime element of the semiring ℕ.
 -- This is useful for proving many other theorems involving prime numbers.
 euclidsLemma : ∀ m n {p} → Prime p → p ∣ m * n → p ∣ m ⊎ p ∣ n
-euclidsLemma m n {p@(suc (suc _))} p-prime p∣m*n = result
+euclidsLemma m n {p} (prime pr) p∣m*n = result
   where
   open ∣-Reasoning
+  instance _ = nonTrivial⇒nonZero -- for NonZero p
```
MatthewDaggitt, line 307: `nonTrivial⇒nonZero` should be an instance?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+irreducible[2] {suc _} d∣2 with ∣⇒≤ d∣2
+... | z<s     = inj₁ refl
+... | s<s z<s = inj₂ refl
+
+
+------------------------------------------------------------------------
+-- NonTrivial
+
+composite⇒nonTrivial : Composite n → NonTrivial n
+composite⇒nonTrivial {1}      = flip contradiction ¬composite[1]
```
MatthewDaggitt, line 215: Given that you're introducing the argument in the other clause, I think clearer to introduce it in the first clause as well and leave the contradiction un-flip-ed

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+-- If a number is k-rough, then so are all of its divisors
+rough⇒∣⇒rough : Rough k m → n ∣ m → Rough k n
+rough⇒∣⇒rough r n∣m (boundedComposite d<k d∣n)
+  = r (boundedComposite d<k (∣-trans d∣n n∣m))
+
+------------------------------------------------------------------------
+-- Corollary: relationship between roughness and primality
+
+-- If a number n is p-rough, and p > 1 divides n, then p must be prime
+rough⇒∣⇒prime : ⦃ NonTrivial p ⦄ → Rough p n → p ∣ n → Prime p
```
MatthewDaggitt, line 168: First argument should be irrelevant?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+
+1-rough : Rough 1 n
+1-rough (boundedComposite ⦃()⦄ z<s _)
+
+2-rough : Rough 2 n
+2-rough (boundedComposite ⦃()⦄ (s<s z<s) _)
 
--- Definition of primality.
+-- If a number n > 1 is k-rough, then k ≤ n
+rough⇒≤ : ⦃ NonTrivial n ⦄ → Rough k n → k ≤ n
```
MatthewDaggitt, line 150: First argument should be irrelevant?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
 prime⇒¬composite : Prime n → ¬ Composite n
-prime⇒¬composite {n@(suc (suc _))} n-prime (d , 2≤d , d<n , d∣n) =
-  n-prime 2≤d d<n d∣n
+prime⇒¬composite (prime p) (d , composite[d]) = p composite[d]
 
 -- note that this has to recompute the factor!
-¬prime⇒composite : 2 ≤ n → ¬ Prime n → Composite n
-¬prime⇒composite {n} 2≤n ¬n-prime =
-  decidable-stable (composite? n) (¬n-prime ∘′ ¬composite⇒prime 2≤n)
+¬prime⇒composite : ⦃ NonTrivial n ⦄ → ¬ Prime n → Composite n
```
MatthewDaggitt, line 279: First argument should be irrelevant?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
 
 composite⇒¬prime : Composite n → ¬ Prime n
-composite⇒¬prime {n@(suc (suc _))} (d , 2≤d , d<n , d∣n) n-prime =
-  n-prime 2≤d d<n d∣n
+composite⇒¬prime (d , composite[d]) (prime p) = p composite[d]
 
-¬composite⇒prime : 2 ≤ n → ¬ Composite n → Prime n
-¬composite⇒prime (s≤s (s≤s _)) ¬n-composite {d} 2≤d d<n d∣n =
-  ¬n-composite (d , 2≤d , d<n , d∣n)
+¬composite⇒prime : ⦃ NonTrivial n ⦄ → ¬ Composite n → Prime n
```
MatthewDaggitt, line 271: First argument should be irrelevant?

src/Data/Nat/Primality.agda
```diff
@@ -141,3 +355,4 @@ private
   -- Example: 6 is composite
   6-is-composite : Composite 6
   6-is-composite = from-yes (composite? 6)
```
MatthewDaggitt, line 357: Probably need to expand the comment in this section, showing how you can use `Decidable` to produce proofs of primality automatically? As opposed to above where you do it manually?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+  (allUpTo? (λ d → 1 <? d →-dec ¬? (d ∣? n)) n)
+
+irreducible? : Decidable Irreducible
+irreducible? zero      = no ¬irreducible[0]
+irreducible? n@(suc _) = Dec.map′ bounded-irr⇒irr irr⇒bounded-irr
+  (allUpTo? (λ m → (m ∣? n) →-dec ((m ≟ 1) ⊎-dec m ≟ n)) n)
+  where
+  BoundedIrreducible : Pred ℕ _
+  BoundedIrreducible n = ∀ {m} → m < n → m ∣ n → m ≡ 1 ⊎ m ≡ n
+  bounded-irr⇒irr : BoundedIrreducible n → Irreducible n
```
MatthewDaggitt, line 259: Even though we're in a `where` block can we separate these with spaces?

MatthewDaggitt (changes requested):
Sorry big review!

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+n>1⇒nonTrivial : 1 < n → NonTrivial n
+n>1⇒nonTrivial 1<2+n = _
+
 
 ------------------------------------------------------------------------
 -- Definitions
 
+-- Definition of having a non-trivial divisor below a given bound
+
+record BoundedComposite (k n d : ℕ) : Set where
```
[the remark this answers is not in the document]
jamesmckinna: Hmmm. It seems important to me to establish `Composite` as  `∃⟨ ... ⟩` and `Prime` as `∀[ \neg ... ]` for suitable `...`, but I see that perhaps a better name is called for here. 

But I don't understand your reasoning as to why `d` should be a field?

src/Data/Nat/Primality.agda
```diff
@@ -127,9 +339,11 @@ euclidsLemma m n {p@(suc (suc _))} p-prime p∣m*n = result
 
   -- if the GCD of m and p is greater than one, then it must be p and hence p ∣ m.
   ... | Bézout.result d@(suc (suc _)) g _ with d ≟ p
-  ...   | yes refl = inj₁ (GCD.gcd∣m g)
-  ...   | no  d≢p  = contradiction (GCD.gcd∣n g) (p-prime 2≤d d<p)
-    where 2≤d = s≤s (s≤s z≤n); d<p = flip ≤∧≢⇒< d≢p (∣⇒≤ (GCD.gcd∣n g))
+  ...   | yes d≡p@refl = inj₁ (GCD.gcd∣m g)
+  ...   | no  d≢p = contradiction d∣p λ d∣p → pr (boundedComposite≢ d≢p d∣p)
```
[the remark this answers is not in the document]
jamesmckinna: It's how it was proved before... which I simply cleaned up a bit. But I agree, this style is annoying!

src/Data/Nat/Primality.agda
jamesmckinna: Those examples were inherited from the previous version... but yes, perhaps these examples better belong in a `README.Primality`?

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
jamesmckinna: Modulo @MatthewDaggitt 's review comments, as there is now an explicit `record` definition, I'm resolving this bit of discussion.

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
MatthewDaggitt: If we follow my suggestion, then `Composite` becomes `BoundedComposite n n` and `Rough k n` as `¬ (BoundedComposite k n)` which is significantly clearer to my mind?

As for the name, either `HasNonTrivialDivisorBelow` or `HasDivisorBelow` seems pretty descriptive?

The `d` is a witness to the fact that it has a non-trivial divisor. In the formalisation so far, there's no need for you to actually access the witness at the type-level. Therefore, I think it should be a field not a parameter.

src/Data/Nat/Primality.agda
jamesmckinna: So... the (anonymous, but relevant) instance *is* passed from the input to the constructor in the output, so it appears to be 'relevant', and indeed the constructor turns yellow if the input is marked irrelevant... But your suggestion about the divisor `d` being a record field might change that.

src/Data/Nat/Primality.agda
jamesmckinna: Ditto.

src/Data/Nat/Primality.agda
jamesmckinna: So the coda to all these comments is that, having moved `NonTrivial` to `Data.Nat.Base`, a lot of explicit instances need to be put back in... already suggesting that the `{{nt}}` fields of `BoundedComposite` and `Prime` need to be made irrelevant, and then... that might be the way to tackle your comments in the whole?

src/Data/Nat/Primality.agda
jamesmckinna: Moreover, the structure of this `where` block, demonstrating the equivalence of one notion with its *bounded* version, via a 'decidable' reflection of the predicate, might ought to be lifted out for each of the three predicates, for which the `all? ..` version in independently interesting (?), and the types of the functions witnessing the equivalence might be worth explicitly stating...?

jamesmckinna:
Rounds of changes in response to @MatthewDaggitt 's review: .
- [x] First: move `NonTrivial` definitions to `Data.Nat.Base`, and knock-on consequences; 
- [ ] Next: try to make all the instances of `NonTrivial` irrelevant;
- [ ] Final: tackle moving divisor `d` from record parameter to record field.

src/Data/Nat/Primality.agda
jamesmckinna: DONE!

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
jamesmckinna: Re: `_∣_` Indeed `quotient` is a record field there!
Re: the definitions at hand. I guess I had been (overly) invested in defining `Prime` and `Composite` as instances of `Relation.Unary.(I)Universal` and `Relation.Unary.Satisfiable`, and of course the latter, being defined via an existential, lends itself to a `record` definition with a field not occurring in the type... so I guess you've convinced me... ;-)

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
jamesmckinna: NB the 'introduction' is a deliberate pun: we need a proof of `¬ (d ∣ p)` for the contradiction, so we introduce a hypothesis of the same name, and then use `pr (boundedComposite≢ d≢p d∣p)` to yield a proof of `⊥`. Of course, the proof could directly appeal to `⊥-elim` or `with () ← ...` but elsewhere in the history of v2.0, you have emphasised using `contradiction` over the latter form, and I find the `⊥-elim` proofs introduce redundant (syntactic) `import`s. Matter of taste, or one for the style guide to rule (less or more dogmatically) upon?

src/Data/Nat/Primality.agda
jamesmckinna: Added some text around the examples, towards eventually (but not now!) writing a `README.Primality`...

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
--- Definition of primality.
+-- If a number n > 1 is k-rough, then k ≤ n
+rough⇒≤ : ⦃ NonTrivial n ⦄ → Rough k n → k ≤ n
+rough⇒≤ rough = ≮⇒≥ λ k>n → rough (boundedComposite k>n ∣-refl)
+
+-- If a number n is k-rough, and k ∤ n, then n is (suc k)-rough
+∤⇒rough-suc : k ∤ n → Rough k n → Rough (suc k) n
+∤⇒rough-suc k∤n r (boundedComposite d<1+k d∣n) with m<1+n⇒m<n∨m≡n d<1+k
+... | inj₁ d<k      = r (boundedComposite d<k d∣n)
+... | inj₂ d≡k@refl = contradiction d∣n k∤n
```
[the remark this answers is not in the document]
jamesmckinna: I suppose so. It's idiomatic to me: ostensibly the two propositions are not in contradiction *unless* `d≡k`, but if we inline the pattern, and just put `refl`, then the thing that is witnessed by it is lost. So for "redundant recoding"'s sake (in the 'cognitive dimensions' language of Green/Petrie/Blackwell et al.), I've taken to leaving these explicit bindings in, by way of 'documentation'. Can remove if, but only if, you *really* disapprove...

src/Data/Nat/Primality.agda
jamesmckinna: Now refactored to reflect this.

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
Taneb: I think it can be shuffled around a bit to be simpler. Something like `contradiction (boundedComposite≢ d≢p d∣p) pr`. I think it was written like this as a consequence of how I was thinking about this bit of the proof - "I know this case is impossible, because d divides p but if d isn't p then d can't divide p". Clean-up there is very welcome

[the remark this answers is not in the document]
jamesmckinna: Double-plus :+1:!

src/Data/Nat/Primality.agda
jamesmckinna: Yes, ... but that causes (non-local!) problems, and I find this kind of thing such a headache to get right. Any ideas?

src/Data/Nat/Primality.agda
```diff
@@ -126,18 +379,9 @@ euclidsLemma m n {p@(suc (suc _))} p-prime p∣m*n = result
...
-  2-is-prime : Prime 2
-  2-is-prime = from-yes (prime? 2)
-
-
-  -- Example: 6 is composite
-  6-is-composite : Composite 6
-  6-is-composite = from-yes (composite? 6)
+  ... | Bézout.result d@(2+ _) g _ with d ≟ p
+  ...   | yes d≡p@refl = inj₁ (GCD.gcd∣m g)
+  ...   | no  d≢p = contradiction (hasBoundedNonTrivialDivisor≢ d≢p d∣p) prp
```
jamesmckinna, line 384: @Taneb 's nudge was very helpful!

src/Data/Nat/Properties.agda
```diff
@@ -902,6 +911,12 @@ m*n≡0⇒m≡0∨n≡0 (suc m) {zero}  eq = inj₂ refl
 m*n≢0 : ∀ m n → .{{_ : NonZero m}} .{{_ : NonZero n}} → NonZero (m * n)
 m*n≢0 (suc m) (suc n) = _
 
+m*n≢0⇒m≢0 : ∀ m {n} → .{{NonZero (m * n)}} → NonZero m
+m*n≢0⇒m≢0 (suc _) = _
+
+m*n≢0⇒n≢0 : ∀ m {n} → .{{NonZero (m * n)}} → NonZero n
+m*n≢0⇒n≢0 m {n} rewrite *-comm m n = m*n≢0⇒m≢0 n {m}
```
jamesmckinna, line 918: So these are also part of the commit trail in #2182 ... which ideally might have been merged before this PR, but is non-breaking, so... all aboard the `rebase` bus!

jamesmckinna:
OK, so there are a number of loose ends here, but the *mathematical*/semantic design, and its consequences, are stable now. 

Suggest we merge as soon as I've fixed whatever else needs fixing from the review comments, and mop up the residue in downstream PRs?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
+  using (_≡_; _≢_; refl; cong)
 
 private
   variable
-    n : ℕ
+    d k m n p : ℕ
+
+  instance
+    nt[2] : NonTrivial 2
+    nt[2] = nonTrivial {0}
```
jamesmckinna, line 31: So it feels strange to have to explicitly declare this... why is this not caught under the generic `nonTrivial` instance?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+
+------------------------------------------------------------------------
+-- Corollary: relationship between roughness and primality
+
+-- If a number n is p-rough, and p > 1 divides n, then p must be prime
+rough⇒∣⇒prime : ⦃ NonTrivial p ⦄ → Rough p n → p ∣ n → Prime p
+rough⇒∣⇒prime r p∣n = prime (rough⇒∣⇒rough r p∣n)
+
+------------------------------------------------------------------------
+-- Basic (non-)instances of Composite
```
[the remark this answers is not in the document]
jamesmckinna: Now changed, but commit not yet pushed.

src/Data/Nat/Primality.agda
jamesmckinna: Yes... but the typechecker gets fouled up in the implicit vs. explicit quantification, and I just don't have the patience to figure those things out!

jamesmckinna:
> Sorry big review!

No worries! The overall is *immeasurably* better for all the feedback... or at least; I hope so!?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
 >-nonZero⁻¹ (suc n) = z<s
 
+-- The property of being a non-zero, non-unit
+
+pattern 2+ n = suc (suc n)
+
+trivial : ℕ → Bool
+trivial 0      = true
+trivial 1      = true
+trivial (2+ _) = false
```
MatthewDaggitt, line 142: It's kind of weird to define predicates using Bool first, as we try and avoid using raw Bool wherever possible. We certainly wouldn't encourage the user to use such a function. Usually we define such predicates as records, e.g. `NonZero`?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
 >-nonZero⁻¹ : ∀ n → .{{NonZero n}} → n > 0
 >-nonZero⁻¹ (suc n) = z<s
 
+-- The property of being a non-zero, non-unit
+
+pattern 2+ n = suc (suc n)
```
MatthewDaggitt, line 137: Not sure this should live under the predicates subheading? Probably way up around line 32 would be better? But open to other suggestions!

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+trivial (2+ _) = false
+
+NonTrivial : Pred ℕ 0ℓ
+NonTrivial = T ∘ not ∘ trivial
+
+instance
+  nonTrivial : ∀ {n} → NonTrivial (2+ n)
+  nonTrivial = _
+
+pattern 1<2+n {n} = s<s (z<s {n})
```
MatthewDaggitt, line 151: Again should go up with the definition of `<` I think?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+
+-- Constructors
+
+n>1⇒nonTrivial : ∀ {n} → 1 < n → NonTrivial n
+n>1⇒nonTrivial 1<2+n = _
+
+nonZero⇒≢1⇒nonTrivial : ∀ {n} → .⦃ NonZero n ⦄ → n ≢ 1 → NonTrivial n
+nonZero⇒≢1⇒nonTrivial {1}    = contradiction refl
+nonZero⇒≢1⇒nonTrivial {2+ _} = _
```
MatthewDaggitt, line 160: This is kind of a weird one. Is this actually needed for this development? If not I suggest we remove it?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+
+instance
+  nonTrivial : ∀ {n} → NonTrivial (2+ n)
+  nonTrivial = _
+
+pattern 1<2+n {n} = s<s (z<s {n})
+
+-- Constructors
+
+n>1⇒nonTrivial : ∀ {n} → 1 < n → NonTrivial n
```
MatthewDaggitt, line 155: Minor, but maybe the inequality in the type should match the name?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+n>1⇒nonTrivial 1<2+n = _
+
+nonZero⇒≢1⇒nonTrivial : ∀ {n} → .⦃ NonZero n ⦄ → n ≢ 1 → NonTrivial n
+nonZero⇒≢1⇒nonTrivial {1}    = contradiction refl
+nonZero⇒≢1⇒nonTrivial {2+ _} = _
+
+recompute-nonTrivial : ∀ {n} → .⦃ NonTrivial n ⦄ → NonTrivial n
+recompute-nonTrivial {2+ _} = _
```
MatthewDaggitt, line 163: I don't think we should prove this lemma here. Instead it should be a defined in terms of `Relation.Nullary.recompute` if needed in this development?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+  nonTrivial = _
+
+pattern 1<2+n {n} = s<s (z<s {n})
+
+-- Constructors
+
+n>1⇒nonTrivial : ∀ {n} → 1 < n → NonTrivial n
+n>1⇒nonTrivial 1<2+n = _
+
+nonZero⇒≢1⇒nonTrivial : ∀ {n} → .⦃ NonZero n ⦄ → n ≢ 1 → NonTrivial n
```
MatthewDaggitt, line 158: Everywhere else in the library I have used the non-unicode version of the instance brackets. Two advantages, shorter to type, and more obvious how to type. Could we be consistent and stick with the double braces everywhere? (Should be added to the style guide)

src/Data/Nat/Primality.agda
MatthewDaggitt: I'm sorry, you'll have to give an idea of what the error messages are when you make it an instance?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+nonZero⇒≢1⇒nonTrivial {1}    = contradiction refl
+nonZero⇒≢1⇒nonTrivial {2+ _} = _
+
+recompute-nonTrivial : ∀ {n} → .⦃ NonTrivial n ⦄ → NonTrivial n
+recompute-nonTrivial {2+ _} = _
+
+-- Destructors
+
+nonTrivial⇒nonZero : ∀ {n} → .⦃ NonTrivial n ⦄ → NonZero n
```
MatthewDaggitt, line 167: If we *don't* make this an instance (see discussion above), then the `n` should definitely be explicit. I'd prefer it to be an instance though.

src/Data/Nat/Base.agda
```diff
@@ -130,6 +132,47 @@ instance
...
+nonTrivial⇒nonZero {n = 2+ k} = _
+
+nonTrivial⇒n>1 : ∀ n → .⦃ NonTrivial n ⦄ → 1 < n
+nonTrivial⇒n>1 (2+ _) = 1<2+n
+
+nonTrivial⇒≢1 : ∀ {n} → .⦃ NonTrivial n ⦄ → n ≢ 1
```
MatthewDaggitt, line 173: This `n` should be explicit!

src/Data/Nat/Coprimality.agda
```diff
@@ -87,63 +89,63 @@ coprime? i j with mkGCD i j
 -- Nothing except for 1 is coprime to 0.
 
 0-coprimeTo-m⇒m≡1 : ∀ {m} → Coprime 0 m → m ≡ 1
```
MatthewDaggitt, line 91: `m` is now a variable and be removed? Same for `{n}` below?

src/Data/Nat/Coprimality.agda
```diff
@@ -87,63 +89,63 @@ coprime? i j with mkGCD i j
...
-0-coprimeTo-m⇒m≡1 {m} c = c (m ∣0 , ∣-refl)
+0-coprimeTo-m⇒m≡1 {m} coprime = coprime (m ∣0 , ∣-refl)
 
-¬0-coprimeTo-2+ : ∀ {n} → ¬ Coprime 0 (2 + n)
+¬0-coprimeTo-2+ : ∀ {n} → ¬ Coprime 0 (2+ n)
```
MatthewDaggitt, line 94: This should use `.{{ NonTrivial n }}` instead of `2+ n`?

src/Data/Nat/Coprimality.agda
MatthewDaggitt: Elsewhere in the library (e.g. divisibility proofs) I've just taken advantage of v2.0 to change the type of the proof entirely when introducing instance arguments. So the new proof above would assume the name `prime=>comprime`. You would need to add an entry in the very bottom of the CHANGELOG where we list all proofs that have been changed in such a way.

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
-open import Relation.Unary using (Decidable)
+open import Relation.Unary using (Pred; Decidable; IUniversal; Satisfiable; _∩_; _⇒_)
 open import Relation.Binary.PropositionalEquality
-  using (refl; cong)
+  using (_≡_; _≢_; refl; cong)
 
 private
   variable
-    n : ℕ
+    d k m n p : ℕ
```
MatthewDaggitt, line 27: Again same comment about `k`

src/Data/Nat/Primality.agda
MatthewDaggitt: I'm not sure off the top of my head, but it does indicate a bug. If you can't spot the problem, let me know and I'll try and take a look.

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
+    d<k     : divisor < k
+    d∣n     : divisor ∣ n
+
+-- smart constructors
+
+hasBoundedNonTrivialDivisor≢ : .{{NonTrivial d}} → .{{NonZero n}} →
+                    d ≢ n → d ∣ n → HasBoundedNonTrivialDivisor n n
+hasBoundedNonTrivialDivisor≢ d≢n d∣n = hasBoundedNonTrivialDivisor (≤∧≢⇒< (∣⇒≤ d∣n) d≢n) d∣n
+
+hasBoundedNonTrivialDivisor>1 : 1 < d → d < n → d ∣ n → HasBoundedNonTrivialDivisor n n
```
MatthewDaggitt, line 52: First argument should be `.{{NonTrival d}}` which means the instance declaration below is no longer needed?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
 
 ------------------------------------------------------------------------
 -- Definitions
 
+-- Definition of having a non-trivial divisor below a given bound
+
+record HasBoundedNonTrivialDivisor (k n : ℕ) : Set where
+  constructor hasBoundedNonTrivialDivisor
+  field
+    {divisor} : ℕ
```
MatthewDaggitt, line 41: Looking at the code below, I think it would improve readability to make this explicit? We are constructing a witness object after all?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
-Composite : ℕ → Set
-Composite 0 = ⊥
-Composite 1 = ⊥
-Composite n = ∃ λ d → 2 ≤ d × d < n × d ∣ n
+Composite : Pred ℕ _
+Composite n = HasBoundedNonTrivialDivisor n n
+
+-- bounded equivalent
+CompositeUpTo : Pred ℕ _
+CompositeUpTo n = ∃⟨ (_< n) ∩ HasNonTrivialDivisor ⟩
```
MatthewDaggitt, line 68: I'm a bit confused about the purpose of this definition? Is it equivalent to `Composite`? Should it be private in this module?


I also have to confess I find this point free style of writing the predicates, (e.g. see `PrimeUpTo` as well) very hard to parse. Would you be against just writing out the full lambda under the exists instead?

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
+  where
+  HasNonTrivialDivisor : Pred ℕ _
+  HasNonTrivialDivisor d = NonTrivial d × d ∣ n
+
+-- smart constructors
+
+composite : .{{NonTrivial d}} → d < n → d ∣ n → Composite n
+composite {d = d} = hasBoundedNonTrivialDivisor {divisor = d}
+
+composite≢ : ∀ d → .{{NonTrivial d}} → .{{NonZero n}} → d ≢ n → d ∣ n → Composite n
```
MatthewDaggitt, line 78: Name of this and the constructor below should have a `-` in between composite and the symbol

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
+-- NonTrivial p is p-Rough, and thereby enforces that
+-- p is a fortiori NonZero and NonUnit
+
+record Prime (p : ℕ) : Set where
+  constructor prime
+  field
+    .{{nt}}  : NonTrivial p
+    isPrime : Rough p p
+
+-- bounded equivalent
```
MatthewDaggitt, line 105: Again what's the purpose of it? Should it be private? Regardless could you add a bit more documentation above it?

src/Data/Nat/Primality.agda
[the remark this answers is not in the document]
MatthewDaggitt: Okay, that's fine!

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,331 @@
...
+PrimeUpTo : Pred ℕ _
+PrimeUpTo n = ∀[ (_< n) ⇒ HasNoNonTrivialDivisor ]
+  where
+  HasNoNonTrivialDivisor : Pred ℕ _
+  HasNoNonTrivialDivisor d = NonTrivial d → d ∤ n
+
+-- Definition of irreducibility
+
+module _ (n : ℕ) where
```
MatthewDaggitt, line 114: I think the interoperability cost of adding a module here isn't really worth it? The two main ones should be `Pred` after all, and the private definition should be a `Rel`? To write otherwise is a little confusing!

MatthewDaggitt (commented):
Sorry a load more comments!

src/Data/Nat/Base.agda
jamesmckinna: Point taken regarding not letting users have access to this.

Regarding the use of `record`s to encode such predicates, I've argued [on #2092](https://github.com/agda/agda-stdlib/issues/2092#issuecomment-1719250454)  that the additional indirection given by an explicit top-level record is overkill for what are in the end, instances of `T _`, and hence *either* trivially inhabited by `_` (avoiding bothering to mention `tt` explicitly), *else* trivially uninhabited and matched by `()`. (See also my comments on reducing all `Decidable` properties to `T?` via `Reflects.fromEquivalence`, and its weaponsation in #2149)

So I've been developing a style of first introducing the `Bool`-valued indicator/characteristic function, and then the predicate via `T ∘_`... but clearly the/one way to satisfy us both is to make `trivial` local to the definition of `NonTrivial`?

src/Data/Nat/Base.agda
jamesmckinna: OK, I think it isn't needed, and indeed seems a bit heavy. I'd been anticipating a ring-theoretic characterisation of `NonTrivial` as `NonZero, NonUnit`, but... YAGNI, at least, not yet ;-)

src/Data/Nat/Base.agda
jamesmckinna: One question which has been in my mind: should its argument be explicit or implicit, and hence, what even should this be called? As it stands, or else
```agda
pattern 1<2+ n = s<s (z<s {n})
```

src/Data/Nat/Primality.agda
jamesmckinna: Both good points. I was on a hobby horse about bounded predicates and PRA...

Regarding 'point-free' predicate definitions, happy to let this go (though I'd always prefer the `∃-syntax` notational style, which this syntax for `Relation.Unary.Satisfiable` affords...). 

Regarding each of the `UpTo` definitions:
* they all follow the same pattern, so it seemed worthwhile emphasising that, even though the instance for `Composite` is ... pointless, inelegant, variation on the definition;
* the `Decidable` property of the public definitions is proved by delegation to the decidability of the `UpTo` versions, and indeed, could be finessed via `Decidable.map` plus an explicit equivalence proof, because they are each equivalent to their `UpTo` counterparts;
* whose proof of `Decidable` can be *automatically* read off the definition by translating `forall` to `all?`, `exists` to `any?`, and so on for the propositional connectives, because the boundedness supplies the requisite bound for those combinators... (a more ambitious refactoring might try to obtain these things by `Reflection`, but that's for another time!);

TL;DR: there's a longer, different, story to tell about these definitions than there is space to develop in this PR... but I had wnated to leave enough breadcrumbs for me (or someone else!?) to be able to follow up on such a story later...

src/Data/Nat/Primality.agda
jamesmckinna: See above, but yes I will write an opening paragraph of commentary/documentation at the top of the `Definitions` section, and make the `UpTo` definitions `private`...

src/Data/Nat/Primality.agda
jamesmckinna: OK! (But further evidence of how my strange design-mind operates... so good to have that checked/litmus-tested)

jamesmckinna:
Thanks again for the additional comments, and apologies for making this such an ordeal!
What's left:
* naming (I've tweaked the field names, but the main record and its constructor name are unchanged; also the various `pattern` synonyms...)
* `CHANGELOG` (I think once we're happy, I'll just chop out the one definition of `Prime` there and put in the new definitions?)

src/Data/Nat/Base.agda
jamesmckinna: Missed this one earlier!

src/Data/Nat/Base.agda
jamesmckinna: Ah... missed this one, as well. It *is* needed, in one place.

src/Data/Nat/Base.agda
jamesmckinna: Yes, I'll see what I can do about hunting this one down... but I do remember it being a pain...

src/Data/Nat/Base.agda
jamesmckinna: Surely not? `n ≢ 1` is a negated proposition, so any assumption of the form `n ≡ 1` towards a contradiction will have instantiated `n`? See also eg. `Data.Nat.Base.≢-nonZero⁻¹` which never, AFAICT, needs its argument explicit except in the definition...?

src/Data/Nat/Coprimality.agda
jamesmckinna: Good spot!

src/Data/Nat/Coprimality.agda
jamesmckinna: So.. should also be renamed/deprecated?

CHANGELOG.md
jamesmckinna: Resolved by most recent review.

src/Data/Nat/Primality.agda
jamesmckinna: Closing this for now. I think it needs some refactoring of the `NonZero` instance reasoning in `GCD`...

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
+  field
+    {{nt}}  : NonTrivial p
+    isPrime : Rough p p
+
+-- Definition of irreducibility
+
+Irreducible : Pred ℕ _
+Irreducible n = ∀[ irreducible n ]
+  where
+  irreducible : ℕ → Pred ℕ _
```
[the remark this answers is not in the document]
jamesmckinna: Now refactored.

src/Data/Nat/Primality.agda
```diff
@@ -8,80 +8,290 @@
...
 
 composite? : Decidable Composite
-composite? 0               = no λ()
-composite? 1               = no λ()
-composite? n@(suc (suc _)) = Dec.map′
-  (map₂ λ { (a , b , c) → (b , a , c)})
-  (map₂ λ { (a , b , c) → (b , a , c)})
-  (anyUpTo? (λ d → 2 ≤? d ×-dec d ∣? n) n)
+composite? n = Dec.map′
+  (map₂ λ (d<n , 1<d , d∣n) → boundedComposite>1 1<d d<n d∣n)
```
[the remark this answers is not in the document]
jamesmckinna: Refactored via `Dec.map` and explicit proofs of `Equivalence`...

src/Data/Nat/Primality.agda
jamesmckinna: Ah, just saw this (GitHub hides too many conversations! and I don't always go back and re-open them... sorry!)
Interesting question! Can we make this a downstream refactoring PR, in the spirit of my comments in general about bounded predicates on Nat?

src/Data/Nat/Primality.agda
jamesmckinna: No, all it required was to make the instance field in the record irrelevant...

src/Data/Nat/Coprimality.agda
jamesmckinna: Interestingly, making this change caused `make test` to fail in `Data.Rational.Base`, with the solution involving some rather horrible explicit `instance` reasoning. One to debug... later?

src/Data/Nat/Primality.agda
jamesmckinna: Here's an example:
```agda
/home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/src/Data/Nat/DivMod.agda:309,22-40
No instance of type NonTrivial (n * o) was found in scope.
when checking that (*-comm o n) is a valid argument to a function
of type
{m = m₁ : ℕ} .⦃ _ = z : NonZero (o * n) ⦄
.⦃ _ = z₁ : NonZero (n * o) ⦄ →
o * n ≡ n * o → m₁ / (o * n) ≡ m₁ / (n * o)```

Comment (jamesmckinna) [reply src/Data/Nat/Primality.agda]: The step at hand:
```agda
  n * m / (o * n) ≡⟨ /-congʳ {o = n * o} (*-comm o n) ⟩
  n * m / (n * o)
```
make *no* appeal to `NonTrivial`, and yet the instance means that the typechecker goes looking for one... very annoying! Is it a bug?

jamesmckinna: Even when rewritten to make the instances more atomic, as:
```agda
m*n/o*n≡m/o : ∀ m n o .⦃ _ : NonZero n ⦄ ⦃ _ : NonZero o ⦄ →
              let instance _ = m*n≢0 o n in 
              m * n / (o * n) ≡ m / o
m*n/o*n≡m/o m n o = begin-equality
  m * n / (o * n) ≡⟨ /-congˡ (*-comm m n) ⟩
  n * m / (o * n) ≡⟨ /-congʳ {o = n * o} (*-comm o n) ⟩
  n * m / (n * o) ≡⟨ m*n/m*o≡n/o n m o ⟩
  m / o           ∎
  where instance _ = m*n≢0 n o; instance _ = m*n≢0 o n
```

src/Data/Nat/Primality.agda
jamesmckinna: Commented out, it generates unsolved metas for Nats at:
* L225
* L228
* L241
* L298
* L359
* L413

all seemingly arising from constructor applications...

src/Data/Nat/Primality.agda
jamesmckinna: Having now pushed to the bitter end of #2182 in its simplification of `NonZero` instance reasoning in `Data.Nat.DivMod` , making the above `nonTrivial⇒nonZero` into an `instance` now breaks the lemma `m*n/m*o≡n/o` instead ... so something is going on, but I'm struggling to be able to precisely nail it down.

src/Data/Nat/Primality.agda
jamesmckinna: So assistance with the `instance` typechecking weirdness would be *very* helpful!

src/Data/Nat/Base.agda
jamesmckinna: See also #2197 , where I have folded this in as a commit to `style-guide`...

src/Data/Nat/Base.agda
jamesmckinna: `Data.Nat.DivMod` L309:
```agda
/home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/src/Data/Nat/DivMod.agda:309,22-29
No instance of type NonTrivial (n * o) was found in scope.
when checking that (*-comm o n) is a valid argument to a function
of type
{n = n₁ : ℕ} {o = o₁ : ℕ} {m = m₁ : ℕ} .⦃ _ = z : NonZero n₁ ⦄
.⦃ _ = z₁ : NonZero o₁ ⦄ →
n₁ ≡ o₁ → m₁ / n₁ ≡ m₁ / o₁
```
trying to typecheck
```agda
m*n/o*n≡m/o : ∀ m n o .⦃ _ : NonZero o ⦄ ⦃ _ : NonZero (o * n) ⦄ →
              m * n / (o * n) ≡ m / o
m*n/o*n≡m/o m n o ⦃ _ ⦄ ⦃ o*n≢0 ⦄ = begin-equality
  m * n / (o * n) ≡⟨ /-congˡ (*-comm m n) ⟩
  n * m / (o * n) ≡⟨ /-congʳ (*-comm o n) ⟩
  n * m / (n * o) ≡⟨ m*n/m*o≡n/o n m o ⟩
  m / o           ∎
  where instance n*o≢0 = subst NonZero (*-comm o n) o*n≢0
```

src/Data/Nat/Primality.agda
jamesmckinna: Cf. [Agda issue #6976](https://github.com/agda/agda/issues/6976)?

src/Data/Nat/Base.agda
MatthewDaggitt: Okay, so I've remembered the real reason they're all defined as records. As @gallais reminds us again and again, Agda behaves far far better when reconstructing meta-variables when the original definition is a record rather than a plain alias. 

I'm afraid I don't have good intuition why, but it's definitely true. In particular when I change the definition to a record, the bug where you need to redefine instances at the top of the `Primality` file goes away again. See the latest commit. 

So I'm pushing the change, and adding some documentation to `Data.Nat.Base`.

src/Data/Nat/Base.agda
MatthewDaggitt: To be consistent I'd advocate for `z<ss` and leave it implicit?

src/Data/Nat/Primality.agda
MatthewDaggitt: Okay thanks for the explanation :+1:

MatthewDaggitt:
Okay, I've had a run through and done a few more refactorings here and there. Have a look through and let me know what you think. The open issues are:
- Updating CHANGELOG
- The naming of the pattern `1<2+ n`
- The name of the `isPrime` field of `Prime` (see comment below)

src/Data/Nat/Primality.agda
```diff
@@ -8,92 +8,186 @@
...
+-- Prime as the diagonal of Rough (and hence the complement of Composite
+-- as defined below). The constructor `prime` takes a proof `isPrime`
+-- that NonTrivial p is p-Rough and thereby enforces that:
+-- * p is a fortiori NonZero and NonUnit
+-- * any non-trivial divisor of p must be at least p, i.e. p itself
+record Prime (p : ℕ) : Set where
+  constructor prime
+  field
+    .{{nontrivial}} : NonTrivial p
+    isPrime         : Rough p p
```
MatthewDaggitt, line 71: I'm a bit worried about the name of this field. It's shouldn't be `isPrime` as otherwise it would be the only field in the record. I'm at a  bit of a loss about what to call it though. Maybe `noDivisors`?

jamesmckinna: Hmm, good thought. I'll ponder that for a bit. Todays a bad day for me, so probably not before tomorrow or the weekend.

jamesmckinna: `isSelfRough`? `isRough`?

MatthewDaggitt: I could get on board with `isRough`. Or how about `notComposite`?

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -40,10 +40,23 @@ open _∣_ using (quotient) public
...
+-- Smart constructor
 
 pattern divides-refl q = divides q refl
 
+------------------------------------------------------------------------
+-- Restricted divisor relation
+
+-- Relation for having a non-trivial divisor below a given bound.
+-- Useful when reasoning about primality.
+record _HasNonTrivialDivisorLessThan_ (m n : ℕ) : Set where
```
jamesmckinna, line 52: Should `m` and `n` be exchanged in order to be congruent with the meaning intended by the name of this type?

Should we (now) make this the `Bundle`, for which a `Structure` `IsBoundedStrictDivisor` is definable with the 'obvious' meaning?

src/Data/Nat/Primality.agda
jamesmckinna: Mond you, if we take this seriously, should we not put the definition of `Compiste` first, then define the type of the field name `noComposite` simply to be `¬ (Composite p)`, with a rider that `Rough p p` is, indeed, definitionally equal to such a property...?

src/Data/Nat/Primality.agda
```diff
@@ -8,92 +8,186 @@
...
+-- Prime as the complement of Composite (and hence the diagonal of Rough
+-- as defined above). The constructor `prime` takes a proof `notComposite`
+-- that NonTrivial p is not composite and thereby enforces that:
+-- * p is a fortiori NonZero and NonUnit
+-- * p is p-Rough, i.e. any proper divisor must be at least p, i.e. p itself
+record Prime (p : ℕ) : Set where
+  constructor prime
+  field
+    .{{nontrivial}} : NonTrivial p
+    notComposite    : ¬ Composite p
```
jamesmckinna, line 78: So that's the definition I finally adopted, lifting the `Composite` definition over that of `Prime`. But I didn't move the code for the properties of `Composite` after that... easy to do, though!

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: One outstanding issue: fixity declaration for this parametrised `record` definition?

jamesmckinna:
OK, so what's now left (?):
- [ ] fixity (and name of?) `BoundedNonTrivialDivisor`
- [ ] moving the code for properties of `Composite` above those for `Prime`, to reflect that this is now the order in which they are defined.

And then we're done!?

src/Data/Nat/Base.agda
```diff
@@ -130,6 +141,34 @@ instance
...
+
+-- Constructors
+
+n>1⇒nonTrivial : ∀ {n} → n > 1 → NonTrivial n
+n>1⇒nonTrivial 1<2+n = _
+
+-- Destructors
+
+nonTrivial⇒nonZero : ∀ n → .{{NonTrivial n}} → NonZero n
+nonTrivial⇒nonZero (2+ _) = _
```
jamesmckinna, line 164: Making this an `instance` still seems to break things in `Data.Nat.DivMod`, but will revisit in #2182 ...

src/Data/Nat/Primality.agda
jamesmckinna: Now updated to reflect this.

src/Data/Nat/Primality.agda
jamesmckinna: Now done.

jamesmckinna:
More incoming...

MatthewDaggitt:
Thanks. I think it all looks good to go apart from your renaming of  `_HasNonTrivialDivisorLessThan_` to `_BoundedNonTrivialDivisor_`. The latter's infix notation no longer makes any sense to me when I read it, i.e. what does `m BoundedNonTrivialDivisor o` actually mean? Which is the divisor? What is it bounded by? 

Compare this to `m HasNonTrivialDivisorLessThan o` which, at least in my opinion, is crystal clear what it's about.

Would you be willing for me to change this back?

MatthewDaggitt:
> UNLESS... anyone thinks that _Rough_ should now join its complement in Data.Nat.Divisibility.Core, along with its 'basic' properties...?

Err, maybe... but at another time?

jamesmckinna:
Re: naming. By all means change back,  but then please make sure that the argument order is correct: the second argument is the thing that has a nontrivial divisor bounded by the first argument, so it should be `n HasBoundedNonTrivialDivisor m`... and then it *that* doesn't quite read correctly? Or have I misunderstood the original suggestion?

Alternatively, my revised name should be modified to `m BoundedNonTrivialDivisorOf n`... which I had experimented with, but found myself irritated by the `Of`...

MatthewDaggitt:
> Re: naming. By all means change back, but then please make sure that the argument order is correct: the second argument is the thing that has a nontrivial divisor bounded by the first argument, so it should be n HasBoundedNonTrivialDivisor m... and then it that doesn't quite read correctly? Or have I misunderstood the original suggestion?

Ah right, that was my mistake. Thanks for picking it up! Okay then I think we should swap the argument order.

MatthewDaggitt (approved):
If you're happy with the renaming @jamesmckinna I'll let you do the honours and hit merge!

Issue #2115, opened 2023-09-29
Reconciling `Data.Nat.Divisibility.Core._∣_` and `Algebra.Definitions.RawMagma._∣_`


jamesmckinna:
The substance of the issue (for the *additive* fragment of Nat) is already discussed in #1919, but that issue can (should!) be closed by merging #1948. So this is a placeholder to work on this specific problem in the *multiplicative* fragment of Nat. 

Carrying over the summary from #2013:
~~Two~~ Three parts to this:
- [x] low-hanging fruit: added pattern synonym and consequences/uses of it to `Data.Nat.*` (merged in #2013)
- [ ] higher-hanging fruit: reconcile the two definitions of `_∣_` for the `Data.Nat.Base.*-rawMagma` instance
- [ ] knock-on changes for `Data.Integer.Divisibility.*` and  `Data.Rational.*` only if the preceding step succeeds

jamesmckinna:
Picking this up again on the back of [#2604](https://github.com/agda/agda-stdlib/pull/2604).

There seem to be two directions to go in (once again: `flip` symmetry!)
* redefine `Data.Nat.Divisibility.Core._∣_` in terms of `Algebra.Definitions.RawMagma *-rawMagma`, renaming `_,_` to `divides`, and then having to fix up all the uses in `Data.Nat`/`Data.Integer`/`Data.Rational` uses with appropriate sprinkles of symmetry
* redefine `Algebra.Definitions.RawMagma._∣_` to flip the direction of the `equality` fields, etc.

Personally, I think that the latter is the better course of action, but may cause problems for external clients of `stdlib` such as @mechvel 's libraries.

However, i think it is the better choice, because of the following refinement of the `complicated = simple` heuristic which has emerged in writing equations in the library, as discussed elsewhere (I can't remember off the top of my head, but I was the one making the point, about *mode* of an equation... how do we search GitHub comments for these things?) namely:
* `complicated` might only appear as a *variable* name in the stated equation, but it will be *used* in a context where that variable sis instantiated with an arbitrarily complicated expression... whereas
* `simple` might be a superficially *more* complicated RHS expression than the LHS, but is appearing in a 'standard' form. moreover one which instantiates its sub-terms upon application of the relevant lemma.

[...]

Issue #2059, opened 2023-08-15
dependencies of things (vol. 94): `Relation.Nullary.Decidable`


jamesmckinna:
Working on PR #2055 drew my attention to the following dependency design 'indecision', probably arising from previous large-scale refactorings:
* ~~`Relation.Nullary.Decidable` imports `Data.Sum.Base` but does not use it at all (anymore)~~ UPDATED this non-dependency now seems to have been fixed
* `Data.Sum` imports various ('the usual') things from `Relation.Nullary.*`, but all/only in the service of defining
```agda
-- Conversion back and forth with Dec

fromDec : Dec A → A ⊎ ¬ A
fromDec ( true because  [p]) = inj₁ (invert  [p])
fromDec (false because [¬p]) = inj₂ (invert [¬p])

toDec : A ⊎ ¬ A → Dec A
toDec (inj₁ p)  = yes p
toDec (inj₂ ¬p) = no ¬p
```

So a question naturally arises (usual asymmetry between which module in a `to*`/`from*` scenario takes ownership of the functions) whether 
* (NOCHANGE) `Data.Sum` should continue to import the relevant pieces of `Relation.Nullary.*`
* (CHANGE) else `Relation.Nullary.Decidable` (~~NOT~~ or `.Core`? not sure about the potential dependency cycles here) should ~~continue to~~ import `Data.Sum.Base`, but also define `fromDec` and `toDec`, drastically simplifying the import list for `Data.Sum`

PROPOSAL: adopt the second 'CHANGE' option, on the basis that `Sum` is conceptually/computationally prior to `Decidable` (so should have 'lighter' dependencies), moreover these functions only exercise specialised instances of `_⊎_`

A followup question would/might then be, cf. #2058 :
* possible candidate renamings of these functions
```agda
fromDec ↦  dec⁻
toDec ↦  dec⁺
```
* or else
```agda
fromDec ↦  toSum
toDec ↦  fromSum
```

jamesmckinna:
Similarly `Data.Maybe.Base.decToMaybe`?

gallais:
```agda
isYes : Dec A → Maybe A
isNo : Dec A → Maybe (¬ A)
```

jamesmckinna:
Sorry to be obtuse, but not sure what your point is here @gallais ? Can you explain for the sclerotic-brained?

jamesmckinna:
[...]

PR #2182, opened 2023-10-25
[fixes #1711] Refactoring `Data.Nat.Divisibility` and `Data.Nat.DivMod`

### LINKED ISSUE #1711: Add properties of the quotient to Data.Nat.Divisibility

jamesmckinna:
Inspired by the discussion of #1969 , a candidate solution to #1711 . Possible issues:
* naming ~~(I've slightly adapted the names suggested/discussed on the issue)~~; 
* choice of `public` exports from `Data.Nat.Divisibility{.Core}`; 
* knock-on consequences for client modules (esp. wrt visible exports?)


src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,37 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
+
+  quotient≡0⇒n≡0 : quotient ≡ 0 → n ≡ 0
+  quotient≡0⇒n≡0 q≡0 = subst (λ q → n ≡ q * m) q≡0 equality
+
+  instance
+    quotient≢0 : .{{NonZero n}} → NonZero quotient
```
Taneb, line 43: Doesn't this instance run into the problems discussed in https://github.com/agda/agda-stdlib/issues/1565

jamesmckinna: Hmmm... not sure!?
I confess I haven't really understood much of that discussion or the examples.
In the one intended candidate specimen deployment I've looked at (your `factorise` function), the instance is only brought into scope by an explicit `open`... so perhaps the thing to do is restrict the export from the module?

jamesmckinna: Latest commit takes the second approach...

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,15 +35,55 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
 
+  quotient≡0⇒n≡0 : quotient ≡ 0 → n ≡ 0
+  quotient≡0⇒n≡0 q≡0 = subst (λ q → n ≡ q * m) q≡0 equality
```
JacquesCarette, line 40: Why don't you match on `refl` and let arithmetic just give you the proof? Or use an equational proof? `subst` seems like such a ill-suited hammer for this.

jamesmckinna: Good idea! but *in the scope of the record*, matching on `refl` fails...
```agda
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  quotient ≟ 0
when checking that the pattern refl has type quotient ≡ 0
```

JacquesCarette: Then I might be tempted to prove a lemma in `Data.Nat.Properties` so that you can use it here?

But I wonder if the matching failure is an Agda bug.

jamesmckinna: > Then I might be tempted to prove a lemma in `Data.Nat.Properties` so that you can use it here?

Re: lemma? I'm happy with the new proofs... are you?

> But I wonder if the matching failure is an Agda bug.

Re: bug? Not sure, but if you can reproduce it with an MWE, do please post an issue on the main `agda` tracker!
(I'm not sure if I can be bothered at this point... sorry!)

jamesmckinna: When you can't match on `refl`, try `rewrite`...: which do you prefer?
```agda
  quotient≡0⇒n≡0 q≡0 rewrite equality | q≡0 = refl
```
etc.

JacquesCarette: I guess I prefer `rewrite` over `subst`, but prefer explicit proofs over both. So weird that `rewrite` would work!

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,15 +35,52 @@ record _∣_ (m n : ℕ) : Set where
...
+  quotient≢0 = ≢-nonZero (contraposition quotient≡0⇒n≡0 (≢-nonZero⁻¹ n))
+
+  n≡m*quotient : n ≡ m * quotient
+  n≡m*quotient rewrite *-comm m quotient = equality
+
+  module _ (1<m : 1 < m) where
+
+    open ≤-Reasoning
+
+    quotient>1 :  (m<n : m < n) → 1 < quotient
```
jamesmckinna, line 52: Does this need the assumption `1 < m`? if not... time to refactor!

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: As I understand it (but not very well... you'd think!?), `rewrite` *is* in fact encapsulating the abstraction over the proof of the equation, and then returning the `refl` instance of that type as the next subgoal/hole... which is morally what the previous `subst` proof was doing 'by hand'... 

... now, just like its grown-up friend `with`, that abstraction step might not be well-typed (hence your reasonable lament about the non-robustness of `rewrite`), but when it does work, I think FWIW that it does save a lot of too-much/too-unnecessary cognitive load on the user.

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,30 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
+
+  quotient≡0⇒n≡0 : quotient ≡ 0 → n ≡ 0
+  quotient≡0⇒n≡0 q≡0 rewrite equality | q≡0 = refl
+
+  quotient≢0 : .{{NonZero n}} → NonZero quotient
+  quotient≢0 = ≢-nonZero (contraposition quotient≡0⇒n≡0 (≢-nonZero⁻¹ n))
```
jamesmckinna, line 43: This can't be the only place in the library when this kind of encapsulated reasoning about `NonZero`ness based on an implication might be useful?

A subsequent PR might ought to refactor this to `Data.Nat.*` somewhere?

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: @JacquesCarette my apologies, I think you are quite correct about preferring a proof in `Data.Nat.Properties`, not least because it reduces the module interface here. Incoming commit...

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: See latest commit...

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,27 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
+
+  quotient≢0 : .{{NonZero n}} → NonZero quotient
+  quotient≢0 rewrite equality = m*n≢0⇒m≢0 quotient
+
+  n≡m*quotient : n ≡ m * quotient
+  n≡m*quotient rewrite equality = *-comm quotient m
```
MatthewDaggitt, line 43: I think all these definitely belong in the `Basic properties` subsection rather the definition? Given that it would probably make more sense to define these outside of the record definition? Especially as we're exporting them publicly in the end?

Actually another question! Presumably these don't need to go in the `Core` file at all, but can go in `Data.Nat.Divisibility` instead?

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,27 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
+
+  quotient≢0 : .{{NonZero n}} → NonZero quotient
```
MatthewDaggitt, line 39: This could be declared as an instance to avoid having to use it manually?

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,27 @@ record _∣_ (m n : ℕ) : Set where
...
+  quotient≢0 rewrite equality = m*n≢0⇒m≢0 quotient
+
+  n≡m*quotient : n ≡ m * quotient
+  n≡m*quotient rewrite equality = *-comm quotient m
+
+  quotient>1 :  (m<n : m < n) → 1 < quotient
```
MatthewDaggitt, line 45: Don't need to name the first argument in the type?

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,27 @@ record _∣_ (m n : ℕ) : Set where
...
+  quotient>1 m<n = ≰⇒> λ q≤1 → n≮n n (begin-strict
```
MatthewDaggitt, line 46: There has got to be a cleaner way to prove this than the double negation used here. Sorry, I know that's not a very helpful comment!

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,7 +35,27 @@ record _∣_ (m n : ℕ) : Set where
...
+      n            ≡⟨ equality ⟩
+      quotient * m ≤⟨ *-monoˡ-≤ m q≤1 ⟩
+      1 * m        ≡⟨ *-identityˡ m ⟩
+      m            <⟨ m<n ⟩
+      n            ∎) where open ≤-Reasoning
+
+  quotient< : (1<m : 1 < m) .{{_ : NonZero n}} → quotient < n
```
MatthewDaggitt, line 53: Likewise no need to name the first argument here.

MatthewDaggitt (changes requested):

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: > I think all these definitely belong in the `Basic properties` subsection rather the definition? Given that it would probably make more sense to define these outside of the record definition? Especially as we're exporting them publicly in the end?

So... that's an interesting design point, for which I confess I made the opposite choice ('wide' `record` rather than 'narrow' definition plus top module-level properties) based on the experience of two things:
* the 'fat' interfaces to `Structures` and `Bundles` in the `Algebra` hierarchy: you want these things to be always-in-scope the moment you have an instance (here, of the relation);
* the discussion of 'symmetric instances for equalities' between myself and @JacquesCarette on issue #2115 (where there's a separate/underlying issue regarding the use of `Σ`-types in design: they are, almost by definition/design, a *non*-extensible `record`);
* the sense that being a 'property' (basic or otherwise) is somehow something which is (at least potentially, if not actually), *abstract* wrt the implementation/representation choices of the underlying type, whereas these 'properties' are somehow 'more' basic still, viz. they are bits of *structure* which should be *automatic* by virtue of the representation (so this might be something about wide or narrow APIs to ADTs, which we've discussed before... ;-))... with a 'logician''s hat on, I'd like to distinguish between 'admissible' (properties) vs. 'derivable' (structure), and this was one attempt to explore that space...

But maybe the 'narrow' version is preferable? I confess I still don't have a good feel for the spectrum of trade-offs here, and it's not something I have seen much discussed in the library, style-guide, etc. or the literature... 

> Actually another question! Presumably these don't need to go in the `Core` file at all, but can go in `Data.Nat.Divisibility` instead?

Well again, my perception/instinct was that that these were `Core` properties, again because of a (possibly-faulty!?) understanding that `Core` things don't exist merely to uncouple complex dependency cycles, but also to reflect the semantic admissible (abstract)/derivable (concrete) distinction above...?

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: Well, that's how I had it originally, but @Taneb [pointed out](https://github.com/agda/agda-stdlib/pull/2182#discussion_r1371486070) this might give rise to the problems discussed in #1565 !? But you're the expert on `instance`s form this perspective, so I'd take your cue. But if we *do* want it as an instance, then shouldn't it be automatically exported by any instance of `_|_`, ie it should be part of the record API, not the 'properties'?

Again, I don't feel entirely on safe ground here as regards my intuitions, so greater experience/wisdom on such points is welcome!

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: I know! (re: 'better' proofs) All of this came about in the context of #1969 #2180 #2181 where I had indeed found myself doing too much 'negative' reasoning (and rightly had that drawn to my attention in @Taneb 's review comments). 

As I had tagged this as v2.1, I thought that this could probably sit a while, but after our discussion of the v2.0 release/transition to v2.1, I realise that perhaps I should stop daydreaming. I definitely thought that another set of eyes might help with refactoring though ;-) and I share yours and Nathan's frustration about this style of reasoning... which also partly comes from not wanting to make the API of `Data.Nat.Properties` any wider, so trying to confine myself to using only what's already there...

jamesmckinna:
Thanks for the extremely helpful review comments! :heart: I'll do the cosmetic ones now, but I'd rather wait for the others until some further consideration of the discussion responses above...

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: But I'll refrain from any refactoring of `Data.Nat.Divisibility` here...!
(So there's another separate design issue about this PR: keeping its footprint small... ;-))

src/Data/Nat/Divisibility/Core.agda
MatthewDaggitt: Hmm it's a good point. The case when the same issue would trigger is when `quotient` would compute to something of the form `suc n`... However, how likely is that? I think you'd have to have a concrete proof of divisibility for a concrete pair of numbers, e.g. you open an actual computed proof of `8 | 2` and you then try and pass `quotient = 4` as an argument to something that needs instance search.

Regardless, instances always trigger. They don't need to be in records or out of records. They just need to be in scope.

src/Data/Nat/Divisibility/Core.agda
MatthewDaggitt: Ah your discussion has highlighted to me that I think this might be another instance of "implicit" knowledge that I have that's being missed here. There's a crucial difference between this type of record and the ones in algebra. 

**Namely**: (ideally) abstract records should be thin, non-abstract records can be fatter.

The algebra records are clearly designed to be non-abstract. Indeed if they were abstract, they would be pretty useless in general. However, the divisibility record, as you touch upon above, in an ideal world should be abstract! Therefore all properties should live externally to the record, as we shouldn't be encouraging the user to inspect the implementation of the record?

Does that make sense?

jamesmckinna: Ah... will need to think, but I (start to) see what you might be saying...

jamesmckinna:
Ready to implement the significant changes... but then, in `Divisibility.Core` or `Divisibility`?

Taneb:
In `Divisibility` unless you expect there to be an import cycle, I think

src/Data/Nat/Divisibility/Core.agda
JacquesCarette: Oh, that's a really good point, and I agree: abstract records should be thin, non-abstract ones can be fatter.

If I try to put my thoughts into words: abstract records define new concepts where the outside world has no business knowing how everything is actually represented. Non-abstract ones are a convenient bundling of information, which indeed can for further convenience bundle up some 'conservative extensions' that are frequently used.

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: So... there seems to be a problem (dark grey background to the definition) if we keep it as an `instance` and define these things in `Core` with the `open _∣_ using (quotient) public` declaration...?

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: So... now in the midst for refactoring, and I'm finding it is still (slightly) more efficient to 'cache' the flipped equality in the record... so that `open m∣n` saves some equational reasoning. So... suggest keeping it, but renaming it to `equalityᵒ`?

jamesmckinna: See the recent commits...

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -48,9 +49,14 @@ pattern divides-refl q = divides q refl
...
 
 *-pres-∣ : ∀ {m n o p} → o ∣ m → p ∣ n → o * p ∣ m * n
-*-pres-∣ {m} {n} {o} {p} (divides c m≡c*o) (divides d n≡d*p) =
+*-pres-∣ {m@.(c * o)} {n@.(d * p)} {o} {p} (divides-refl c) (divides-refl d) =
   divides (c * d) (begin
-    m * n             ≡⟨ cong₂ _*_ m≡c*o n≡d*p ⟩
+    m * n             ≡⟨⟩
     (c * o) * (d * p) ≡⟨ [m*n]*[o*p]≡[m*o]*[n*p] c o d p ⟩
     (c * d) * (o * p) ∎)
   where open ≡-Reasoning
```
jamesmckinna, line 57: The introduction of the pattern synonym makes this lemma largely trivial, and it has only a single usage in `Data.Nat.DivMod./-*-interchange`, where moreover the interchange employed in the proof of the lemma is then undone by that use, so it seems as though this could even be dispensed with altogether? (and moved to `Data.Nat.Divisibility`... with suitable deprecation?)

jamesmckinna:
This last commit proves the defining relationship between `_∣_` and `_/_`, which could live in `Divisibility.Core`, or in `DivMod` or in `Divisibility`, currently landing in the last one. But I'm wondering if, in subsequent PRs, it could be used to simplify certain properties in `DivMod`, and hence perhaps properly belongs in `Divisibility.Core`... but for now that might be the best place to stop?

CHANGELOG.md
```diff
@@ -2753,9 +2753,21 @@ Additions to existing modules
...
+  ```
+
+* Added  a new instance and new definitions to `Data.Nat.Divisibility`:
+  ```agda
+  instance quotient≢0 : .{{NonZero n}} → NonZero quotient
+
+  quotient∣  : q ∣ n
+  quotient>1 : m < n → 1 < quotient
+  quotient<  : 1 < m → .{{NonZero n}} → quotient < n
+  n/m≡quotient : .{{ _ : NonZero m }} → n / m ≡ quotient
```
jamesmckinna, line 2770: Missing: the `divides` hypotheses. Commit incoming

src/Data/Nat/Divisibility.agda
```diff
@@ -26,43 +24,87 @@ open import Relation.Binary.Definitions
...
+  n/m≡quotient : .⦃ _ : NonZero m ⦄ → n / m ≡ quotient
+  n/m≡quotient = *-cancelʳ-≡ _ _ m $ begin-equality
+    n / m * m    ≡⟨ m/n*n≡m m∣n ⟩
+    n            ≡⟨ equality ⟩
+    quotient * m ∎
+
+------------------------------------------------------------------------
+-- Exports
+
+open _∣_ using (quotient) public
```
jamesmckinna, line 77: Modulo deciding how much to export as `public` here, I think that this PR is now ready.

jamesmckinna:
I now think that everything is (back) where it belongs, together with a new round of simplifications arising from the earlier PR introducing `divides-refl` into `Core`...

jamesmckinna:
... Aaaaargh! Didn't do enough testing for the knock-on consequences...

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: All new properties now in `Data.Nat.Divisibility`, so resolving this conversation. But very helpful discussion as regards 'tacit' knowledge, and criteria for 'narrow' vs. 'wide' APIs, esp. in respect of how 'abstract they may, or may not, be...!

src/Data/Nat/Divisibility.agda
jamesmckinna: This export now follows an initial prelude introducing all the new lemmas, and I think, as before, suffices.

jamesmckinna:
Time to stop noodling!

jamesmckinna:
OK, that's actually ready (for review) now?

CHANGELOG.md
```diff
@@ -2777,6 +2782,17 @@ Additions to existing modules
   pattern divides-refl q = divides q refl
   ```
 
+* Added  a new instance and new definitions to `Data.Nat.Divisibility`:
+  ```agda
+  instance quotient≢0 : m ∣ n → .{{NonZero n}} → NonZero quotient
+
+  equalityᵒ    : m ∣ n → n ≡ m * quotient
+  quotient-∣   : m ∣ n → quotient ∣ n
+  quotient>1   : m ∣ n → m < n → 1 < quotient
```
jamesmckinna, line 2791: After #2181 suggest that this be redefined in terms of `NonTrivial` defined there.

CHANGELOG.md
```diff
@@ -2777,6 +2782,17 @@ Additions to existing modules
...
+  quotient-<   : m ∣ n → 1 < m → .{{NonZero n}} → quotient < n
```
jamesmckinna, line 2792: Ditto.

jamesmckinna:
Good to go now, I think!

jamesmckinna:
Yes, definitely now!

jamesmckinna:
OK, so this has grown beyond its original boundaries... ;-) 
But hopefully in the last few commits, a useful collection of increments to `Data.Nat.Properties` and their use in simplifying of `Data.Nat.DivMod`...

src/Data/Digit.agda
```diff
@@ -55,11 +55,8 @@ toNatDigits base@(suc (suc _)) n = aux (<-wellFounded-fast n) []
   aux : {n : ℕ} → Acc _<_ n → List ℕ → List ℕ
   aux {zero}        _      xs =  (0 ∷ xs)
   aux {n@(suc _)} (acc wf) xs with does (0 <? n / base)
-  ... | false = (n % base) ∷ xs
-  ... | true  = aux (wf q<n) ((n % base) ∷ xs)
-    where
-    q<n : n / base < n
-    q<n = m/n<m n base (s<s z<s)
+  ... | false = (n % base) ∷ xs -- n here is already sufficient?
```
jamesmckinna, line 58: Suggests that refactoring this module be raised as a separate issue/PR... not least because the type of the function
```agda
toNatDigits : (base : ℕ) {base≤16 : True (1 ≤? base)} → ℕ → List ℕ
```
of which this is part, betrays some evolutionary... cruft.

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -14,11 +14,14 @@ module Data.Nat.Divisibility.Core where
 
 open import Data.Nat.Base using (ℕ; _*_; _<_; NonTrivial)
 open import Data.Nat.Properties
-open import Level using (0ℓ)
 open import Relation.Nullary.Negation using (¬_)
 open import Relation.Binary.Core using (Rel)
-open import Relation.Binary.PropositionalEquality
-  using (_≡_; refl; sym; cong₂; module ≡-Reasoning)
+open import Relation.Binary.PropositionalEquality.Core
+  using (_≡_; refl)
```
jamesmckinna, line 20: I suppose that this might even be simplified further in favour of the `Builtin` version...?

MatthewDaggitt: I think it better we try and avoid depending on `Builtin` modules unless we absolutely have to. Otherwise, we become more vulnerable to changes to the Agda API.

src/Data/Digit.agda
MatthewDaggitt: Yup seems sensible.

jamesmckinna:
See also #2221.

src/Data/Digit.agda
MatthewDaggitt, line 58: What does "sufficient" in this comment mean?

src/Data/Fin/Properties.agda
```diff
@@ -65,6 +65,9 @@ private
 ¬Fin0 : ¬ Fin 0
 ¬Fin0 ()
 
+nonZeroIndex : Fin n → ℕ.NonZero n
```
MatthewDaggitt, line 68: Missing in CHANGELOG?

src/Data/Nat/Coprimality.agda
```diff
@@ -65,18 +67,8 @@ coprime-/gcd m n = GCD≡1⇒coprime (GCD-/gcd m n)
...
-
-  2+≢1 : ∀ {n} → 2+ n ≢ 1
-  2+≢1 ()
-
 coprime? : Decidable Coprime
-coprime? m n with mkGCD m n
-... | (0    , g) = no  (0≢1  ∘ GCD.unique g ∘ coprime⇒GCD≡1)
-... | (1    , g) = yes (GCD≡1⇒coprime g)
-... | (2+ _ , g) = no  (2+≢1 ∘ GCD.unique g ∘ coprime⇒GCD≡1)
+coprime? m n = Dec.map′ gcd≡1⇒coprime coprime⇒gcd≡1 (gcd m n ≟ 1)
```
MatthewDaggitt, line 71: Nice change!

src/Data/Nat/Coprimality.agda
```diff
@@ -21,7 +21,9 @@ open import Function.Base using (_∘_)
 open import Level using (0ℓ)
 open import Relation.Binary.PropositionalEquality.Core as P
   using (_≡_; _≢_; refl; trans; cong; subst)
-open import Relation.Nullary as Nullary using (¬_; contradiction; yes ; no)
+import Relation.Nullary.Decidable as Dec
+open import Relation.Nullary.Negation using (¬_; contradiction)
+import Relation.Nullary as Nullary
```
MatthewDaggitt, line 26: After the refactoring in v2.0, it should be possible to export all of this via `Relation.Nullary`. I'd only do this fine grained importing if you can eliminate the `Relation.Nullary` imports altogether?

src/Data/Nat/DivMod.agda
```diff
@@ -222,11 +203,13 @@ m/n≤m m n = *-cancelʳ-≤ (m / n) m n (begin
   m           ≤⟨ m≤m*n m n ⟩
   m * n       ∎)
 
-m/n<m : ∀ m n .{{_ : NonZero m}} .{{_ : NonZero n}} → n ≥ 2 → m / n < m
-m/n<m m n n≥2 = *-cancelʳ-< _ (m / n) m (begin-strict
-  (m / n) * n ≤⟨ m/n*n≤m m n ⟩
-  m           <⟨ m<m*n m n n≥2 ⟩
-  m * n       ∎)
+m/n<m : ∀ m n .{{_ : NonZero m}} .{{_ : NonTrivial n}} →
+        let instance _ = Nat.nonTrivial⇒nonZero n in m / n < m
```
MatthewDaggitt, line 207: So unfortunately this is a breaking change to the lemma, so we can't make it v2.1. Will have to wait for v3.0 I'm afraid.

src/Data/Nat/DivMod.agda
```diff
@@ -285,64 +279,76 @@ m/n≡1+[m∸n]/n {m@(suc m-1)} {n@(suc n-1)} m≥n = begin-equality
...
+  pm = p * m
+
+  lem : o / pm * p * m ≤ o
+  lem = begin
+    o / pm * p * m       ≡⟨ *-assoc (o / pm) p m ⟩
+    o / pm * pm          ≤⟨ m/n*n≤m o pm ⟩
+    o                    ∎
+
+m*n/m*o≡n/o : ∀ m n o .{{_ : NonZero m}} .{{_ : NonZero o}} →
+              let instance _ = m*n≢0 m o in
```
MatthewDaggitt, line 312: Again a breaking change...

src/Data/Nat/DivMod.agda
```diff
@@ -285,64 +279,76 @@ m/n≡1+[m∸n]/n {m@(suc m-1)} {n@(suc n-1)} m≥n = begin-equality
...
+    (m * n) / (m * o)             ≡⟨ m/n≡1+[m∸n]/n (*-monoʳ-≤ m n≥o) ⟩
+    1 + (m * n ∸ m * o) / (m * o) ≡⟨ cong (suc ∘ (_/ (m * o))) (*-distribˡ-∸ m n o) ⟨
+    1 + (m * (n ∸ o)) / (m * o)   ≡⟨ cong suc (helper (rec n∸o<n)) ⟩
+    1 + (n ∸ o) / o               ≡⟨ m/n≡1+[m∸n]/n n≥o ⟨
     n / o                         ∎
-    where n∸o<n = ∸-monoʳ-< (n≢0⇒n>0 (≢-nonZero⁻¹ o)) (≮⇒≥ n≮o)
+    where n∸o<n = ∸-monoʳ-< (n≢0⇒n>0 (≢-nonZero⁻¹ o)) n≥o
 
-m*n/o*n≡m/o : ∀ m n o .⦃ _ : NonZero o ⦄ ⦃ _ : NonZero (o * n) ⦄ →
+m*n/o*n≡m/o : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
```
MatthewDaggitt, line 328: Breaking change.

src/Data/Nat/DivMod.agda
```diff
@@ -351,8 +357,9 @@ m<n*o⇒m/o<n {m} {suc n} {o} m<n*o with m <? o
   m / o ∸ 1 ∸ n         ≡⟨ ∸-+-assoc (m / o) 1 n ⟩
   m / o ∸ suc n         ∎
 
-m/n/o≡m/[n*o] : ∀ m n o .⦃ _ : NonZero n ⦄ .⦃ _ : NonZero o ⦄
-                .⦃ _ : NonZero (n * o) ⦄ → m / n / o ≡ m / (n * o)
+m/n/o≡m/[n*o] : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
```
MatthewDaggitt, line 360: Breaking change

src/Data/Nat/DivMod.agda
```diff
@@ -385,56 +394,66 @@ m/n/o≡m/[n*o] m n o = begin-equality
...
   n / d + m * (n / d) ∎
 
-/-*-interchange : ∀ {m n o p} .{{_ : NonZero o}} .{{_ : NonZero p}} .{{_ : NonZero (o * p)}} →
-                  o ∣ m → p ∣ n → (m * n) / (o * p) ≡ (m / o) * (n / p)
-/-*-interchange {m} {n} {o@(suc _)} {p@(suc _)} o∣m p∣n = *-cancelˡ-≡ _ _ (o * p) (begin-equality
-  (o * p) * ((m * n) / (o * p)) ≡⟨  m*[n/m]≡n (*-pres-∣ o∣m p∣n) ⟩
-  m * n                         ≡⟨ cong₂ _*_ (m*[n/m]≡n o∣m) (m*[n/m]≡n p∣n) ⟨
-  (o * (m / o)) * (p * (n / p)) ≡⟨ [m*n]*[o*p]≡[m*o]*[n*p] o (m / o) p (n / p) ⟩
-  (o * p) * ((m / o) * (n / p)) ∎)
+/-*-interchange : .{{_ : NonZero o}} .{{_ : NonZero p}} →
```
MatthewDaggitt, line 403: Breaking change.

src/Data/Nat/DivMod.agda
```diff
@@ -385,56 +394,66 @@ m/n/o≡m/[n*o] m n o = begin-equality
...
+  (m / o) * (n / p)         ∎
 
 m*n/m!≡n/[m∸1]! : ∀ m n .{{_ : NonZero m}} →
-                 (m * n / m !) {{m !≢0}}  ≡ (n / (pred m) !) {{pred m !≢0}}
-m*n/m!≡n/[m∸1]! (suc m) n = m*n/m*o≡n/o (suc m) n (m !) {{m !≢0}} {{suc m !≢0}}
-
-m%[n*o]/o≡m/o%n : ∀ m n o .⦃ _ : NonZero n ⦄ .⦃ _ : NonZero o ⦄
-                  ⦃ _ : NonZero (n * o) ⦄ → m % (n * o) / o ≡ m / o % n
-m%[n*o]/o≡m/o%n m n o ⦃ _ ⦄ ⦃ _ ⦄ ⦃ n*o≢0 ⦄ = begin-equality
+                  let instance _ = m !≢0 ; instance _ = (pred m) !≢0 in
```
MatthewDaggitt, line 416: Breaking change

src/Data/Nat/DivMod.agda
```diff
@@ -385,56 +394,66 @@ m/n/o≡m/[n*o] m n o = begin-equality
...
+                  (m * n / m !) ≡ (n / (pred m) !)
+m*n/m!≡n/[m∸1]! m′@(suc m) n = let instance _ = m !≢0 in m*n/m*o≡n/o m′ n (m !)
+
+m%[n*o]/o≡m/o%n : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
+                  let instance _ = m*n≢0 n o in
```
MatthewDaggitt, line 421: Breaking change

src/Data/Nat/DivMod.agda
```diff
@@ -385,56 +394,66 @@ m/n/o≡m/[n*o] m n o = begin-equality
...
   m / o ∸ m / (n * o) * n           ≡⟨ cong (λ # → m / o ∸ # * n) (/-congʳ (*-comm n o)) ⟩
   m / o ∸ m / (o * n) * n           ≡⟨ cong (λ # → m / o ∸ # * n) (m/n/o≡m/[n*o] m o n ) ⟨
   m / o ∸ m / o / n * n             ≡⟨ m%n≡m∸m/n*n (m / o) n ⟨
   m / o % n                         ∎
-  where instance o*n≢0 = subst NonZero (*-comm n o) n*o≢0
+  where instance _ = m*n≢0 n o; instance _ = m*n≢0 o n
 
-m%n*o≡m*o%[n*o] : ∀ m n o .⦃ _ : NonZero n ⦄ ⦃ _ : NonZero (n * o) ⦄ →
+m%n*o≡m*o%[n*o] : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
+                  let instance _ = m*n≢0 n o in
```
MatthewDaggitt, line 434: Breaking change

src/Data/Nat/DivMod.agda
```diff
@@ -385,56 +394,66 @@ m/n/o≡m/[n*o] m n o = begin-equality
...
   (m ∸ m / n * n) * o               ≡⟨ *-distribʳ-∸ o m (m / n * n) ⟩
   m * o ∸ m / n * n * o             ≡⟨ cong (λ # → m * o ∸ # * n * o) (m*n/o*n≡m/o m o n) ⟨
   m * o ∸ m * o / (n * o) * n * o   ≡⟨ cong (m * o ∸_) (*-assoc (m * o / (n * o)) n o) ⟩
   m * o ∸ m * o / (n * o) * (n * o) ≡⟨ m%n≡m∸m/n*n (m * o) (n * o) ⟨
   m * o % (n * o)                   ∎
+  where instance _ = m*n≢0 n o
 
-[m*n+o]%[p*n]≡[m*n]%[p*n]+o : ∀ m {n o} p ⦃ _ : NonZero (p * n) ⦄ → o < n →
+[m*n+o]%[p*n]≡[m*n]%[p*n]+o : ∀ m {n o} p .{{_ : NonZero n}} .{{_ : NonZero p}} →
+                              let instance _ = m*n≢0 p n in o < n →
```
MatthewDaggitt, line 446: Breaking change

src/Data/Nat/DivMod.agda
```diff
@@ -243,39 +226,50 @@ m/n<m m n n≥2 = *-cancelʳ-< _ (m / n) m (begin-strict
...
   m / m ≤⟨ /-monoʳ-≤ m m≥n ⟩
   m / n ∎
 
+m/n≡0⇒m<n : ∀ {m n} .{{_ : NonZero n}} → m / n ≡ 0 → m < n
+m/n≡0⇒m<n {m} {n@(suc _)} m/n≡0  with <-≤-connex m n
+... | inj₁ m<n = m<n
+... | inj₂ n≤m = contradiction m/n≡0 (≢-nonZero⁻¹ _)
+  where instance _ =  >-nonZero (m≥n⇒m/n>0 n≤m)
+
+m/n≢0⇒n≤m : ∀ {m n} .{{_ : NonZero n}} → .{{NonZero (m / n)}} → n ≤ m
```
MatthewDaggitt, line 248: Eek I'm not sure about this one. I can't help but feel it should be `m / n ≢ 0` rather than `.{{NonZero (m / n)}}` and I'm trying to work out why. Possible reasons:
- Symmetry with the lemma above
- We're never generating proofs of inequalities via instance search so no need to make it an instance.

If you're keen to still have it in this form then we probably need to make `m` and `n` explicit...

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,68 @@ open import Relation.Binary.Definitions
...
+    n            ∎
+
+-- Exports
+
+open _∣_ using (quotient) public
+
 
 ------------------------------------------------------------------------
--- Definition
+-- Defining equation for _/_
```
MatthewDaggitt, line 76: "Specification" instead of "defining equation" maybe? And isn't it a specification for `_∣_` rather than `_/_`?

src/Data/Nat/Divisibility.agda
```diff
@@ -79,20 +119,20 @@ m%n≡0⇔n∣m m n = mk⇔ (m%n≡0⇒n∣m m n) (n∣m⇒m%n≡0 m n)
...
+∣-antisym {zero}  {n}     p∣q  _  = sym (equalityᵒ p∣q)
+∣-antisym {suc m} {suc n} p∣q q∣p = ≤-antisym (∣⇒≤ p∣q) (∣⇒≤ q∣p)
 
 infix 4 _∣?_
 
 _∣?_ : Decidable _∣_
-zero  ∣? zero   = yes (divides-refl 0)
-zero  ∣? suc m  = no ((λ()) ∘′ ∣-antisym (divides-refl 0))
-suc n ∣? m      = Dec.map (m%n≡0⇔n∣m m (suc n)) (m % suc n ≟ 0)
+zero  ∣? zero   = Dec.yes (divides-refl 0)
```
MatthewDaggitt, line 129: Is there a reason we're starting to qualify `yes` and `no`?

src/Data/Nat/Divisibility.agda
```diff
@@ -125,42 +165,43 @@ module ∣-Reasoning where
 
   open ∣-syntax _IsRelatedTo_ _IsRelatedTo_ ∣-go public
 
+open ∣-Reasoning
```
MatthewDaggitt, line 168: I know this is tempting, but bitter experience has always shown me that after you make this change, it'll only be a matter of weeks before someone tries to add something that needs a different form of reasoning and you'll have to revert the change again.

That nugget of wisdom should probably be added to the style guide...

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,68 @@ open import Relation.Binary.Definitions
...
+open import Data.Nat.Divisibility.Core public hiding (*-pres-∣)
+
+module _ (m∣n : m ∣ n) where
+
+  open _∣_ m∣n
+  open ≤-Reasoning
+
+  instance
+    quotient≢0 : .{{NonZero n}} → NonZero quotient
+    quotient≢0 rewrite equality = m*n≢0⇒m≢0 quotient
```
MatthewDaggitt, line 49: Making this an instance is pointless as it has an explicit argument (which would be clearer if the explicit parameter wasn't lifted up to anonymous module)

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,68 @@ open import Relation.Binary.Definitions
...
+
+  equalityᵒ : n ≡ m * quotient
```
MatthewDaggitt, line 51: What does the `ᵒ` in the name stand for? I'm not a huge fan of this name to be honest. Maybe `quotient-≡` would be more in keeping with the other names?

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,68 @@ open import Relation.Binary.Definitions
...
+
+  quotient-< : .{{NonTrivial m}} → .{{NonZero n}} → quotient < n
+  quotient-<     = begin-strict
+    quotient     <⟨ m<m*n quotient m (nonTrivial⇒n>1 m) ⟩
+    quotient * m ≡⟨ equality ⟨
+    n            ∎
+
+-- Exports
+
+open _∣_ using (quotient) public
```
MatthewDaggitt, line 72: This export doesn't do anything as opening `Divisibility.Core` above already exports `quotient`

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,68 @@ open import Relation.Binary.Definitions
...
+private
+  variable d m n o p : ℕ
+
+
+------------------------------------------------------------------------
+-- Definition and derived properties
+
+open import Data.Nat.Divisibility.Core public hiding (*-pres-∣)
+
+module _ (m∣n : m ∣ n) where
```
MatthewDaggitt, line 42: I kind of feel that lifting this up to a module parameter makes the proofs below harder to interpret. If you're looking at line 64 in isolation for example, it's really not clear what the relationship between `m` and `n` and are, and scrolling up 24 lines to find the answer is disruptive.

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -35,9 +38,8 @@ record _∣_ (m n : ℕ) : Set where
   constructor divides
   field quotient : ℕ
         equality : n ≡ quotient * m
-open _∣_ using (quotient) public
```
MatthewDaggitt, line 38: Ah so why have you removed the public export here? This is not a full-blown breaking change as we don't officially encourage the use of `Core` modules, but the nonetheless it's big enough that we should have a good reason for doing so!

src/Data/Nat/Divisibility/Core.agda
```diff
@@ -60,12 +62,18 @@ record _HasNonTrivialDivisorLessThan_ (m n : ℕ) : Set where
...
-    (c * d) * (o * p) ∎)
-  where open ≡-Reasoning
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.1
+
+*-pres-∣ : o ∣ m → p ∣ n → o * p ∣ m * n
```
MatthewDaggitt, line 72: This is nice! I wonder what happened that resolved this dependency cycle for us. Was it something that you did in this PR?

src/Data/Nat/Properties.agda
```diff
@@ -1728,6 +1736,25 @@ pred-mono-≤ {suc _} {suc _} m≤n = s≤s⁻¹ m≤n
...
+pred-cancel-≤ {m = zero}  {n = zero}  _  = inj₂ z≤n
+pred-cancel-≤ {m = zero}  {n = suc _} _  = inj₂ z≤n
+pred-cancel-≤ {m = suc _} {n = zero} z≤n = inj₁ (refl , refl)
+pred-cancel-≤ {m = suc _} {n = suc _} le = inj₂ (s≤s le)
+
+pred-cancel-< : pred m < pred n → m < n
+pred-cancel-< {m = zero}  {n = suc _} _ = z<s
+pred-cancel-< {m = suc _} {n = suc _}   = s<s
+
+pred-injective : .{{NonZero m}} → .{{NonZero n}} → pred m ≡ pred n → m ≡ n
```
MatthewDaggitt, line 1749: No `→` between `NonZero` proofs needed

src/Data/Nat/Properties.agda
```diff
@@ -1598,11 +1596,21 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
   (m + n) ∸ o          ≡⟨ +-∸-assoc m o≤n ⟩
   m + (n ∸ o)          ∎
 
+m≤n+o⇒m∸n≤o : ∀ m n {o} → m ≤ n + o → m ∸ n ≤ o
```
MatthewDaggitt, line 1599: We are really inconsistent with which arguments are explicit/implicit in this and the three proofs immediately below...

MatthewDaggitt (changes requested):
Some great refactorings here. Sorry as usual a few comments here and there.

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: No, in fact it's a knock on consequence of introducing `divides-refl`, albeit one I hadn't noticed, or planned for, at the time of ... whichever PR that was ;-)

src/Data/Nat/Properties.agda
jamesmckinna: Yes, some if that us historical. Some if it may be carelessness on my part, some if it may be because what's inferrable does actually vary ... will investigate.

src/Data/Nat/DivMod.agda
```diff
@@ -285,64 +279,76 @@ m/n≡1+[m∸n]/n {m@(suc m-1)} {n@(suc n-1)} m≥n = begin-equality
...
+    where n∸o<n = ∸-monoʳ-< (n≢0⇒n>0 (≢-nonZero⁻¹ o)) n≥o
 
-m*n/o*n≡m/o : ∀ m n o .⦃ _ : NonZero o ⦄ ⦃ _ : NonZero (o * n) ⦄ →
+m*n/o*n≡m/o : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
+              let instance _ = m*n≢0 o n in
               m * n / (o * n) ≡ m / o
-m*n/o*n≡m/o m n o ⦃ _ ⦄ ⦃ o*n≢0 ⦄ = begin-equality
-  m * n / (o * n) ≡⟨ /-congˡ (*-comm m n) ⟩
+m*n/o*n≡m/o m n o = begin-equality
+  m * n / (o * n) ≡⟨ /-congˡ {o = o * n} (*-comm m n) ⟩
```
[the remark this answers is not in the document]
jamesmckinna: Good question! But I seemed to need to... and I'm not sure how the 2.6.3 to 2.6.4 transition may have affected this...

src/Data/Digit.agda
jamesmckinna: That in this case branch, IIUC, `n % base` is actually equal (propositionally) to `n`. But I'd be happier for a second pair of eyes to confirm that!

src/Data/Nat/Coprimality.agda
jamesmckinna: OK... IIRC I'd abandoned #2055 partly over such questions. I agree that importing from top-level `Relation.Nullary` might be more robust, but I think I'd been trying to get my head round the precise dependencies (plus using qualified imports in order to do so; eg `Dec.map` (primed or otherwise) seems more informative than finding a bare occurrence of `map`?)

src/Data/Nat/Divisibility.agda
jamesmckinna: Hmm.

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: Two reasons:
* It now gets publicly exported from `Divisibility` instead (and so, in theory, should be 'better' than relying on a `Core` module)
* If it's exported here, then the lemmas above in the anonymous module as phrased end up with ambiguous resolution of the projection name `quotient`... and thus was my 'solution' to that...

Happy to take a steer on how best to handle this, but some of these things arise, I think,  precisely from your earlier suggestion to take all of this out of `Core`, where I had previously had put the #1711 lemmas as manifest fields of the record...

src/Data/Digit.agda
jamesmckinna: It's partly a question of how these things get implemented later, I think, in the compiled haskell. If you can compute division with remainder in one go, then there's no point in relitigating the real tip ships between the various numerical values and their references. But on the face of it, it's an expensive computation of `n / base`, then a test for non-zeroness, and then an appeal to the value of the remainder `n % base`. 

A simpler analysis might be to test `does (n <? base)` and only compute `n % base` in the `false` branch of such a test. How much if this reflects the complex history of this module and it's friends, I don't really know, which perhaps meant I'd been (too?) conservative as to this refactoring. From the library point of view, I think the current definition is bonkers. But without really being clear about how this interacts with compilation via `MAlonzo` (I don't find it easy to discover anything about that pipeline), it's perhaps one to leave alone. But I (perhaps mistakenly) decided to leave the comment there for 'the future'...?

CHANGELOG.md
```diff
@@ -2524,6 +2529,8 @@ Additions to existing modules
 
 * Added new proofs in `Data.Fin.Properties`:
   ```
+  nonZeroIndex       : Fin n → ℕ.NonZero n
```
jamesmckinna, line 2532: So, despite your comment below, here it is?

src/Data/Nat/Properties.agda
jamesmckinna: So there are (at least) two instances where it seems useful to have the first (left-hand) arguments explicit, namely to each of `_+_` and `_*_`, because Agda doesn't always seem able to infer those, even if they are subsequently governed by  a predicate... the 'usual' function inversion problem. And as elsewhere in the library, custom seems to dictate using explicit arguments when doing an explicit pattern match. That said, this is an example where the witness to the the relation could drive the proof (it's not irrelevant/erasable), so maybe a better version is possible.

src/Data/Nat/Properties.agda
jamesmckinna: Actually, it's funny. 
If you remove the arrow, Agda throws an error:
```agda
/opt/AGDA/RAMDISK/Agda-stdlib-dev-v2.0-rc1/src/Data/Nat/Properties.agda:1749,19-32
⦃ NonZero m ⦄ cannot appear by itself. It needs to be the argument
to a function expecting an instance argument.
when scope checking ⦃ NonZero m ⦄
```
If, instead, you supply a leading forall quantifier, then you get:
```agda
/opt/AGDA/RAMDISK/Agda-stdlib-dev-v2.0-rc1/src/Data/Nat/Properties.agda:1749,57-58
Variable m is declared irrelevant, so it cannot be used here
when checking that the expression m has type ℕ
```
So the alternative might to introduce the redundant `.{{_ : ...}}` notation, but since the subsequent operations don't depend on such irrelevant proofs, I find it... clutter.

src/Data/Nat/Coprimality.agda
jamesmckinna: As for importing `Relation.Nullary` itself, that partly rests on unresolved issues about where `Recomputable` should live as a definition, together with its inherited pointwise versions in `Unary` and `Binary`. But happy to postpone such niceties, against later refactoring.

src/Data/Nat/Properties.agda
jamesmckinna: Or is this a bug?

src/Data/Nat/Coprimality.agda
jamesmckinna: Done.

src/Data/Nat/DivMod.agda
jamesmckinna: Ah! Happy to revert this change, and return to it later, if that means the rest can go through?

src/Data/Nat/DivMod.agda
jamesmckinna: Ah... I see (now). I was too eager!

src/Data/Nat/Divisibility.agda
jamesmckinna: Oh... I think I see what you mean. In any case, I was surprised not to see it in the library, and I suppose I had been taking `_∣_` as conceptually prior to `_/_`. Maybe the neutral middle ground would be `Relating _/_ and _∣_.quotient`?

src/Data/Nat/Divisibility.agda
jamesmckinna: Good question. It may be pure idiosyncrasy on my part. I'll look again.

src/Data/Nat/Divisibility.agda
jamesmckinna: Ha! Sigh... I suspect you are correct. Indeed I think the style guide does already say something about making appeals to `Reasoning` modules local to any particular lemma... and I thought I could get away with it otherwise. Oh dear!

src/Data/Nat/Divisibility.agda
jamesmckinna: I think in earlier versions of this PR, we went round in circles about making this an instance or not. Previously, when it was manifest field, I think you argued for it being an instance (against Nathan); when I moved it up to a lemma in the anonymous module, I didn't rethink the question. Will remove the `instance` ...

src/Data/Nat/Divisibility.agda
jamesmckinna: Ah, but see elsewhere. It doesn't now...

src/Data/Nat/DivMod.agda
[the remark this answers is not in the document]
jamesmckinna: So, of course, the original formulation, with the two instance arguments made explicit, solved the problem a different way, but one that is worse, I think?

jamesmckinna:
Thanks for the further review comments! I've pushed commits addressing some of the 'easy'/easier' ones, and I'll worry away at the others. 

I think it will take a while for me to revert all the breaking changes (which I think is worth doing; just as *they* would be, downstream; but reverting the breaking ones may then have knock-on changes in `Rational.Properties`, so I'll take it slowly) and to scratch my head over some of the others, esp. naming ones. 

Still hopefully in line for v2.1, but clearly it's *still* `not-ready`... sigh.

src/Data/Nat/Properties.agda
jamesmckinna: I think I have a tidying up for this now, but the still there seems to be some variation.

src/Data/Nat/DivMod.agda
jamesmckinna: Yes, this is an odd one. Better to revert it for now, I guess.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

src/Data/Nat/DivMod.agda
jamesmckinna: Reverted now.

jamesmckinna:
OK, so I think I have now successfully reverted all the breaking changes (while kicking myself that I hadn't understood them to be such... otherwise I could have pushed harder to have had this in v2.0!), so it's now down to the others, plus the ... knock-on `CHANGELOG` explosion.

src/Data/Nat/DivMod.agda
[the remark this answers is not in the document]
jamesmckinna: But reverting the breaking change restores the *status quo ante*...

src/Data/Fin/Properties.agda
jamesmckinna: Now: after resolving merge conflict, it's [L37 in `CHANGELOG`](https://github.com/agda/agda-stdlib/blob/fac8e6c5f7cafddba653b5a3d150ed170e758f9a/CHANGELOG.md?plain=1#L37)

src/Data/Nat/Divisibility.agda
jamesmckinna: I suspect I had been thinking "the proof of `_∣?_ : Decidable _∣_` is the one place where decidability is appealed to, so let's localise use of `Relation.Nullary.Decidable` to this one place via a qualified import.

Easy enough to revert (but there is a secret agenda, about deprecating `yes` and `no` downstream, because they are too strict as *patterns*, while harmless as *definitions*/'smart constructors', so I guess I was laying a trail of breadcrumbs towards that cf issues #2058 #2059 and my various experiments around resolving those (plus #2055 and also #2199 ...)

src/Data/Nat/Divisibility.agda
jamesmckinna: Commit 9708ca8a6 now tries it your way, and I'm forced to say: I find this *much* harder to read, precisely because all the unnecessary parametrisation now has to be made explicit.

src/Data/Nat/Divisibility/Core.agda
jamesmckinna: Reverted by commit 9708ca8a6

src/Data/Nat/Divisibility.agda
```diff
@@ -25,24 +25,59 @@ open import Relation.Binary.Definitions
...
+  n                ∎
+  where open ≤-Reasoning; instance _ = quotient≢0 m∣n
 
 ------------------------------------------------------------------------
--- Definition
+-- Relating _/_ and quotient
 
-open import Data.Nat.Divisibility.Core public
+n/m≡quotient : (m∣n : m ∣ n) .{{_ : NonZero m}} → n / m ≡ quotient m∣n
+n/m≡quotient {m = m} (divides-refl q) = m*n/n≡m q m
```
jamesmckinna, line 70: The more I think about it, the more I think this lemma belongs in `DivMod` or even `Divisibility.Core`, for then the [two lemmas in `DivMod`](https://github.com/agda/agda-stdlib/blob/9708ca8a6b528180d59025892aa8df941b9e3141/src/Data/Nat/DivMod.agda#L187-L191) follow almost immediately; indeed would do so if moreover `equalityᵒ` (however we end up naming this) went with them... what do you think?

jamesmckinna:
State of play:
* need to agree a better name for `equalityᵒ`; you don't like this choice, and I'm not thrilled by the proposed alternative...
* I'm still unhappy about commit 9708ca8

But otherwise I think that's ready to reconsider for merging?

src/Data/Nat/Divisibility.agda
MatthewDaggitt: Sure sounds good :+1:

src/Data/Nat/Divisibility.agda
MatthewDaggitt: Okay, well that's a good point about their eventual deprecation. However, I think the trail is a little obtuse, and this change won't pro-actively help when and if those changes land. So can we revert it for now?

src/Data/Nat/Divisibility.agda
MatthewDaggitt: Thanks, sorry!

src/Data/Nat/DivMod.agda
MatthewDaggitt: Just a heads up that the incoming commit hasn't seemed to arrived?

src/Data/Nat/Divisibility.agda
MatthewDaggitt: Okay I think the actual problem is that `equality` is a frankly terrible name. How about we define a more meaningful alias name for `equality` in the record `m|n=>n≡ quotient*m` and then this can be ``m|n=>n≡ m*quotient`?

src/Data/Nat/Properties.agda
MatthewDaggitt: Okay great thanks!

src/Data/Nat/Properties.agda
MatthewDaggitt: The leading quantifier error message certainly looks like a bug to me as `m` is not declared irrelevant. You could report it to Agda? However, I was wrong in my initial comment - you do need the arrow :sweat_smile: Sorry!

jamesmckinna: Thanks! ;-)
I'll try to rig up a MWE to post on Agda, but... later.

src/Data/Nat/DivMod.agda
jamesmckinna: Oh?

src/Data/Nat/Divisibility.agda
MatthewDaggitt: > Commit https://github.com/agda/agda-stdlib/commit/9708ca8a6b528180d59025892aa8df941b9e3141 now tries it your way, and I'm forced to say: I find this much harder to read, precisely because all the unnecessary parametrisation now has to be made explicit.

Eek, I'm sorry to hear that. I much prefer the new version :sweat_smile:. In terms of preferences I would order them as:

1. Current situation
2. A collection of manifest fields
3. Previous situation

I understand the DRY argument, but I don't think it outweighs the enhanced readability. Under the previous situation, you actually had to know the internals of the `_|_` record really well to understand what each lemma was saying when you looked at it in isolation. Now in my opinion it's clear in the types what's going on in each lemma even to someone who doesn't know what the internals of `_|_` are...

Anybody else care to arbitrate? @JacquesCarette, @gallais, @Taneb ?

src/Data/Digit.agda
MatthewDaggitt: Ah right in which case might a clearer comment be something along the lines of `-- Could this simply be `n :: xs`?`

src/Data/Nat/Divisibility.agda
jamesmckinna: Pushed.

src/Data/Digit.agda
MatthewDaggitt: To be honest, I've never once used what comes out of Malonzo as a consideration for whether to make a refactoring or not. I'm not saying it's a bad thing to consider, but just a heads up that not considering it hasn't come to bite us at any point _yet_...

src/Data/Nat/DivMod.agda
MatthewDaggitt: Ah, yup. Here now. Thanks!

src/Data/Nat/Divisibility.agda
jamesmckinna: Agree that `equality` in the record is a terrible name!
I think I didn't understand: where should the proposed alias live? As a manifest field in the record?

MatthewDaggitt:
> But otherwise I think that's ready to reconsider for merging?

Yes, there are three unresolved conversations above, but once they're resolved, happy to merge!

src/Data/Nat/Divisibility.agda
MatthewDaggitt: I mean I would personally put it in the section with all the other stuff you moved out of the record... but I realise that may make your DRY objection worse, so happy to wait for the resolution of that conversation before resolving this one!

src/Data/Digit.agda
jamesmckinna: Comment revised. 
As for MAlonzo my attempts to query how it behaves on the main issue tracker didn't yield any useful answer, so yes, happy not ever to have to think about it... ;-)

src/Data/Nat/Divisibility.agda
jamesmckinna: Well, in a way the issue is orthogonal to the DRY one, so I've pushed the alias definition as is.

src/Data/Nat/Divisibility.agda
jamesmckinna: Ahead of any arbitration from people who are allowed to be asleep at the moment (I wish I were!), I suggest we park this.
After all, subsequently reverting to the anonymous module version would be a no-CHANGELOG low-hanging fruit commit, so could happen at any time!

As for the discussion of 'internals', yes this came up before as an argument against having these lemmas as manifest fields. I agree it's undesirable to expose such details unless absolutely necessary, but here, `quotient` is already exported publicly as a field name, and all these lemmas talk about is that name, and its relationship to the visible arguments `m`, `n`, and `m|n`... 

... so to my, admittedly partisan, eyes, if as a reader you want to be shielded from any abstraction violation, just look at the types: then there's nothing exposed that isn't intended to be. If you want to understand _why_ those lemmas are provable, it seems a big ask to imagine such proofs _not_ exposing the internals, surely?

And, at the risk of wanting to have the last word ;-), that might be my argument for having them as manifest fields in the first place: internal to the record is precisely the place (and perhaps even: _nowhere_ else) where we should be allowed to exploit the representation, so long as the public API for the record (now including the types of the manifest fields) doesn't expose such details?

What's maybe undesirable about this is that Agda doesn't smoothly support the separate specification of that publicly-facing API, except as another `record` type (one which now may hide certain details). But the marshalling and unmarshalling between the concrete `record` defined in `Core`, and its witnessing of an 'abstract' one, if taken seriously, would probably drive everyone beyond reason... (cf `DivMod` itself as a specification!). And that's an issue we can't resolve in this PR (but I did raise it on the [Simple Agda issue](https://github.com/agda/agda/issues/7006) regarding datatypes...).

jamesmckinna: Oh, and: when/if we ever return to the hard parts of #2013 , that may be the moment to take a deep breath, and revisit this discussion!?

jamesmckinna:
> > But otherwise I think that's ready to reconsider for merging?
> 
> Yes, there are three unresolved conversations above, but once they're resolved, happy to merge!

I hope that's now done enough to resolve those three?

src/Data/Nat/Divisibility.agda
MatthewDaggitt: Yup I think I agree with your analysis here, that maybe the underlying problem here is the lack of abstraction capabilities in Agda. Okay, will mark this as resolved.

MatthewDaggitt (approved):

src/Data/Nat/Divisibility.agda
JacquesCarette: I know it's "too late", but I do agree that it seems that the real problem here is a lack of abstraction features in Agda.

PR #2183, opened 2023-10-26
[fixes #2175] Documentation misc. typos etc. for RC1 

jamesmckinna:

PR #2184, opened 2023-10-26
[ fix #1743 ] move README to `doc/` directory

### LINKED ISSUE #1743: top-level README.agda (and Everything.agda)
Over in agda-categories we have https://github.com/agda/agda-categories/issues/342 which is about `Everything.agda` that's in the repo but doesn't typecheck.

The situation is different in stdlib: `Everything.agda` is generated for the doc, but not in the repo. `README.agda` is, however, in. It has the same problem, in that it doesn't typecheck where it is (AFAIK).

I'm not quite sure what to do about it. But likely the same 'solution' is likely to apply to both libraries - thus this issue.

gallais:


MatthewDaggitt (approved):
Looks great. Thank you for all the extra clean-up. @jamesmckinna do you want to check this branch out and confirm that it doesn't disrupt your workflow?

MatthewDaggitt:
Going to see if this branch gets through the merge queue.

shhyou:
From the GH docs https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions, I guess the CI https://github.com/agda/agda-stdlib/blob/3a1bda24f1949c6fe0ba85d686745dcf246513a8/.github/workflows/ci-ubuntu.yml#L2 needs to add `merge_group:` to `on:`.

MatthewDaggitt:
Yes, I'm trying to fix the CI on this branch: https://github.com/agda/agda-stdlib/pull/2183

dev/experimental.agda-lib
```diff
@@ -0,0 +1,4 @@
+name: standard-library-dev
+include: . ../src
+flags:
+  --warning=noUnsupportedIndexedMatch
```
jamesmckinna, line 4: This makes perfect sense as an incarnation of the orthodox library `import` convention.

But for end users/new stdlib developers, what, if anything, should they add to their `~/.agda/libraries` to pick up the `development` library rather than the `release-version` library during their own development? Obviously, this can't be part of the PR, but can it be documented in `HACKING.md` for example?

HACKING.md
```diff
@@ -244,23 +244,14 @@ you are never committing anything with a whitespace violation:
...
-  in the `standard-library.agda-lib` file.
-
-* Please do not include this change in your pull request.
-
+* By default the README files are not exported in the
+  `standard-library.agda-lib` file in order to avoid
+  clashing with other people's README files.
+
+* If you wish to type-check a README file, then you will
+  need to change the present working directory to `doc/`
```
jamesmckinna, line 252: So, I've been thinking a lot (if 'hard') about this, and I don't know if it makes more, or less, sense, to have the `README.*` modules in a `doc` directory, or in good-old `README` as a top-level directory, and simply bump all the modules names 'up' one level, by having what would then be `README.agda` import from `Axiom` rather than `README.Axiom` etc.

Alternatively, leave the *entire* existing `README/` directory alone, but simply have *its* `blah.agda-lib` point at... the right directories (and here is where I get confused about what those should be... `. ../src` as already, or `. ./README ../src` (I guess this latter is subsumed by the former?)

In other words, the change could be less invasive still simply by moving toplevel `README.agda` to `README/`, and updating the `GNUmakefile` and `.github/tooling` scripts?

jamesmckinna (changes requested):

jamesmckinna:
Apologies for choosing 'request changes' option, but I did at least want to debate the `doc/README/*` vs. `README/*` choice, but maybe there's a good reason for choosing the former over the latter? Bu the latter seems a less obviously 'invasive'/obtrusive change as I have termed it in my comments. Or do we envisage a time when `doc/` might contain *more* than simply `README.*` modules?

As for the `dev/experimental.agda-lib` approach, yes it seems to work fine, as of course does my existing hand-hacked solution ;-) So in that sense I am OK either way with how this issue gets resolved in any given release...

dev/experimental.agda-lib
gallais: Isn't that just the standard agda docs about managing libraries?
There's nothing special about the dev version.

jamesmckinna:
Oh... my run of `make test` (run from the development root directory) fell over with
```agda
/home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/doc/README.agda:277,1-18
Failed to find source of module Everything in any of the following
locations:
  /home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/doc/Everything.agda
  /home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/doc/Everything.lagda
  /home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/doc/../src/Everything.agda
  /home/james/RESEARCH/Agda-stdlib-dev-v2.0-rc1/doc/../src/Everything.lagda
  /opt/CABAL/store/ghc-8.10.7/Agda-2.6.3-054ad03dbee6d0c2006b869b3262bbffa6b1d3fb97d573e861153c9b3fb59d3c/share/lib/prim/Everything.agda
  /opt/CABAL/store/ghc-8.10.7/Agda-2.6.3-054ad03dbee6d0c2006b869b3262bbffa6b1d3fb97d573e861153c9b3fb59d3c/share/lib/prim/Everything.lagda
  import Everything
```
but *that* may be an artefact of my local setup :-( ... After all, the GitHub test run passed ... ;-)

gallais:
Don't know, for me it fails to build because of badly named primitives in the Reflection modules.
Which is quite puzzling given that I though `master` was supposed to build with 2.6.4

Anyways, the CI does not use `make test` so you could very well see issues there.
TBH I have no interest in maintaining the Makefile that has already been bit rotting for ages...

gallais:
Nevermind, I got the same error after correctly setting `AGDA_EXEC`.

jamesmckinna:
Good to learn that the `GNUmakefile` might be obsolete, but if so, perhaps we *should* positively remove it and document that somewhere?

HACKING.md
shhyou: Many modules in `README` have identical paths as the modules from the standard library (e.g. `README/Function/Reasoning.agda` and `src/Function/Reasoning.agda`, `README/Data/List.agda` and `src/Data/List.agda`, etc.). If the `README.` prefix in the module paths is removed, I wonder how Agda resolves the module?

jamesmckinna: I think because [the `include` directive line](https://github.com/agda/agda-stdlib/blob/561d92420ce6f318b17d431530e1ccdaa6a04263/doc/standard-library-doc.agda-lib#L2)
ensures that those in `doc/README` will get searched for before trying to look for them in `src`?

shhyou: The `include` in this PR looks good to me. What I don't understand is how your alternative approach works. (this one:
> ... in good-old README as a top-level directory, and simply bump all the modules names 'up' one level, by having what would then be README.agda import from Axiom rather than README.Axiom etc.
> 
> Alternatively, leave the entire existing README/ directory alone, but simply have its blah.agda-lib point at... the right directories (and here is where I get confused about what those should be... . ../src as already, or . ./README ../src (I guess this latter is subsumed by the former?)

I might be wrong, but including `../README` directory before `../src` in `include:` sounds like modules in `README` will shadow those that have the same name in `src`. Without the `README.` prefix in the module paths, how could files like `README.Data.List.Relation.Binary.Equality` that import `Data.Nat` continue to work? Wouldn't `Data.Nat` incorrectly refers to `README.Data.Nat`?

jamesmckinna: Ah... ok, perhaps you're right. 

In any case, I would prefer simply to have the `doc` directory be called `README` instead. Or are you saying that such a solution also would not work?

shhyou: I think that should work.

MatthewDaggitt:
> Apologies for choosing request changes option, but I did at least want to debate the doc/README/* vs. README/* choice, but maybe there's a good reason for choosing the former over the latter? 

I definitely prefer the latter. Much simpler, if we can get to work. I'll give it a go.

> Good to learn that the GNUmakefile might be obsolete, but if so, perhaps we should positively remove it and document that somewhere?

I still use `make test` locally and it works well. It also looks like it's used to publish listings whatever that does. I guess the question is how would we recommend people test the whole library locally without it? But yes we can open a separate issue to discuss.

MatthewDaggitt:
Hmm okay three options:

- `doc/README/*` - what we have now, can keep `README` in the module names, but `doc` is annoying as we don't have other plans to put other documentation in there.
- `README/*` - not feasible unless we drop `README` from all the module names
- `README/README/*` - maybe the best option?

What are people's preferences?

jamesmckinna:
In the interests of merging sooner rather than laterI'd go with the third option, although the second would be more attractive (to me) if it could be made to work, but I'm not clear about the cost/benefit. 

Happy to remove my 'changes requested' embargo ahead of any eventual decision...

jamesmckinna (approved):

JacquesCarette:
I prefer `doc/README` over `README/README`. The latter screams 'hack'. Opening up a `doc/` hierarchy may be a psychologically worthy move that might spur people to start putting more things here.

jamesmckinna:
Well, I guess we could move all the `notes/` documents there, to start with...?

JacquesCarette:
Indeed!

MatthewDaggitt:
Okay moved all `notes/` to `doc/` as suggested. @jamesmckinna would you be accept the changes?

MatthewDaggitt:
Ooh you already have!

MatthewDaggitt:
Thank you @gallais for pushing this through!

gasche:
The toplevel README of the stdlib now includes a dead link:

https://github.com/agda/agda-stdlib/blob/ff1dc85bcbdea1d83edfa40c5ed251ac0ab5b0e8/README.md?plain=1#L23

gallais:
Thanks Gabriel, fix in #2202

Issue #2123, opened 2023-10-05
Refactoring uses of `with`


jamesmckinna:
Cf. #1937 

As well as removing 'redundant' uses of `with`, it seems worth flagging as a separate issue the need/desire to exploit the (idiomatics of the) extended capabilities of `with` notation throughout the library. Specifically:
- [ ] use of *nested* `with` esp. in conjunction with the next idiom, irrefutable `with`, building up a cascade of dependent inversions until the final RHS proof term may be assembled from the resulting pieces;
- [ ] use of 'irrefutable' `with`: `with <pat> ← <exp> = ...` which avoids having to uncouple the sub-expression `<exp>` from the single matching clause `... | <pat> = ...` with which the definition then proceeds; there are three distinguished special cases of this, according to the form of `<pat>`:
* `()` or more generally, `<constructor> (<constructor>... ()...)`, as a (convenient?) drop-in replacement for more explicit appeals to `⊥-elim` or `Relation.Nullary.Negation.Core.contradiction`; 
* `refl`: such uses seem strictly to extend the capability of `rewrite <exp>` for an `<exp>` of the form `s ≡ t`; 
* ` <pat1> , <pat2> , ... , <patn>` (and `record` constructors more generally) for extended/nested `Σ`-types; the advantages being that parentheses around the pattern are frequently redundant clutter, but also that such uses may often also permit simplification in the form of replacing the use of an irrefutable `with` pattern with a `let <pat1> , <pat2> , ... , <patn> = <exp> in ...` binding instead;
- [ ] the use of `with <exp> in <eq>`: this is perhaps more controversial, given the ongoing saga of this notation not *precisely* offering a drop-in replacement for the hoped-to-be-deprecated `inspect` idiom; nevertheless, all uses of `inspect` have been successfully removed from the library, so the use of this extended form is encouraged where possible;

Other features/idioms as they arise... with the proposal that at some point (v2.1? v3.0?) we systematically weed out the redundant uses of `with` and refactor to take advantage of the above.

PR #2186, opened 2023-10-30
Make decidable versions of sublist functions the default

### LINKED ISSUE #2172: Change in definition of `Data.List.Base.filter` etc. result in new unsolved metas in proofs
The definition of `filter` etc got changed to be defined in terms of the boolean test rather than a decidable predicate. Several problems have resulted:

1. Firstly this is not documented in the CHANGELOG
2. This has lead to many of the proofs not being able to infer their list argument. In particular `filter-accept` and `filter-reject` in `Data.List.Properties` now need their `xs` parameter to be passed explicitly. 

We should check all functions which have been redefined similarly.

MatthewDaggitt:
Fixes #2172 but as discussed in issue, unfortunately introduces an undesirable new dependency from a `Base` file on a `Properties` file.

Maybe the answer is to move `T?` to `Data.Nullary.Decidable.Core`?

### DIFF
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -405,9 +405,9 @@ partitionᵇ p = partition (T? ∘ p)
 
 span : ∀ {P : Pred A p} → Decidable P → List A → (List A × List A)
 span P? []       = ([] , [])
-span P? (x ∷ xs) with does (P? x)
+span P? ys@(x ∷ xs) with does (P? x)
 ... | true  = Prod.map (x ∷_) id (span P? xs)
-... | false = ([] , x ∷ xs)
+... | false = ([] , ys)
 
 
 spanᵇ : (A → Bool) → List A → List A × List A
@@ -458,7 +458,7 @@ wordsByᵇ p = wordsBy (T? ∘ p)
 derun : ∀ {R : Rel A p} → B.Decidable R → List A → List A
 derun R? [] = []
 derun R? (x ∷ []) = x ∷ []
-derun R? (x ∷ y ∷ xs) with does (R? x y) | derun R? (y ∷ xs)
+derun R? (x ∷ xs@(y ∷ _)) with does (R? x y) | derun R? xs
 ... | true  | ys = ys
 ... | false | ys = x ∷ ys
 
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -14,7 +14,6 @@ module Data.List.Base where
 open import Algebra.Bundles.Raw using (RawMagma; RawMonoid)
 open import Data.Bool.Base as Bool
   using (Bool; false; true; not; _∧_; _∨_; if_then_else_)
-open import Data.Bool.Properties using (T?)
 open import Data.Fin.Base using (Fin; zero; suc)
 open import Data.Maybe.Base as Maybe using (Maybe; nothing; just; maybe′)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; _+_; _*_ ; _≤_ ; s≤s)
@@ -24,11 +23,11 @@ open import Data.These.Base as These using (These; this; that; these)
 open import Function.Base
   using (id; _∘_ ; _∘′_; _∘₂_; _$_; const; flip)
 open import Level using (Level)
-open import Relation.Nullary.Decidable using (does; ¬?)
 open import Relation.Unary using (Pred; Decidable)
 open import Relation.Binary.Core using (Rel)
 import Relation.Binary.Definitions as B
 open import Relation.Binary.PropositionalEquality.Core using (_≡_)
+open import Relation.Nullary.Decidable.Core using (T?; does; ¬?)
 
 private
   variable
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -14,6 +14,7 @@ module Data.List.Base where
 open import Algebra.Bundles.Raw using (RawMagma; RawMonoid)
 open import Data.Bool.Base as Bool
   using (Bool; false; true; not; _∧_; _∨_; if_then_else_)
+open import Data.Bool.Properties using (T?)
 open import Data.Fin.Base using (Fin; zero; suc)
 open import Data.Maybe.Base as Maybe using (Maybe; nothing; just; maybe′)
 open import Data.Nat.Base as ℕ using (ℕ; zero; suc; _+_; _*_ ; _≤_ ; s≤s)
@@ -23,7 +24,7 @@ open import Data.These.Base as These using (These; this; that; these)
 open import Function.Base
   using (id; _∘_ ; _∘′_; _∘₂_; _$_; const; flip)
 open import Level using (Level)
-open import Relation.Nullary.Decidable.Core using (does; ¬?)
+open import Relation.Nullary.Decidable using (does; ¬?)
 open import Relation.Unary using (Pred; Decidable)
 open import Relation.Binary.Core using (Rel)
 import Relation.Binary.Definitions as B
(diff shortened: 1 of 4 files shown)


jamesmckinna:
And along with `T?`, all its (slightly rag-tag collection) properties... so double :+1: to the move!
(but if you do, then please also add the [companion lemma](https://github.com/jamesmckinna/agda-stdlib/blob/f938d1bb6b7e578f66ab9688a2fc564cf6b8f90d/src/Relation/Nullary/Reflects.agda#L39-L41)
```agda
T⁺ : ∀ b → Reflects (T b) b
T⁺ true  = of _
T⁺ false = of id
```
to `Relation.Nullary.Reflects`...)

jamesmckinna:
Oh, and if you do go down the road if updating `Reflects` and `Decidable.Core`, then I'd push (back, again) for #2149 to be merged into v2.0 ...

JacquesCarette:
:+1: on moving to `Data.Nullary.Decidable.Core` (though its property should not be in that file too - that should presumably be in `Data.Nullary.Decidable.Properties`).

In general, I think it is a good idea to move definitions of *predicates* out of `*.Properties` files into `*.Predicates` (or the like) files.

jamesmckinna:
Well, I think the property can now be removed entirely, so no need for a distinct `Data.Nullary.Decidable.Properties` in this instance... the only consideration might be how to deprecate `T?-diag` ... but it's used nowhere in the library, so...?

MatthewDaggitt:
> In general, I think it is a good idea to move definitions of predicates out of *.Properties files into *.Predicates (or the like) files.

I'm not aware of any predicate defined in `Properties` files. Only proofs that they are decidable.

jamesmckinna:
Looping back to last weeks discussion: have you experimented with replacing calls to `does (P?)` with `isYes P?`? There's a comment in `Decidable.Core` to the effect that the use of `isYes` helps the type checker infer `P` from a proof of `isYes P?`? If this is *not* the case, we should clean up the comment, and if so, it might be a simpler solution to your issue?

MatthewDaggitt:
> Looping back to last weeks discussion: have you experimented with replacing calls to does (P?) with isYes P?? There's a comment in Decidable.Core to the effect that the use of isYes helps the type checker infer P from a proof of isYes P?? If this is not the case, we should clean up the comment, and if so, it might be a simpler solution to your issue?

I haven't. It would definitely be a worthwhile experiment, although regardless of whether it works or not, I have to confess I'm still happier with this solution though?

jamesmckinna:
Global comment ahead of a full review: now is the moment to deprecate the style of doing `with` on a `Bool`-valued expression, in favour of vanilla `if_then_else_`, surely? #1937 / #2123 or did we decide not to do this? It seems especially glaring here, given that all the functions are being redefined anyway...

MatthewDaggitt:
> Global comment ahead of a full review: now is the moment to deprecate the style of doing with on a Bool-valued expression, in favour of vanilla if_then_else_, surely? https://github.com/agda/agda-stdlib/issues/1937 / https://github.com/agda/agda-stdlib/issues/2123 or did we decide not to do this? It seems especially glaring here, given that all the functions are being redefined anyway...

I did think about that, but I wasn't sure I wanted to risk accidentally introducing any more meta-inference changes...

jamesmckinna:
> > Global comment ahead of a full review: now is the moment to deprecate the style of doing with on a Bool-valued expression, in favour of vanilla if_then_else_, surely? #1937 / #2123 or did we decide not to do this? It seems especially glaring here, given that all the functions are being redefined anyway...
> 
> I did think about that, but I wasn't sure I wanted to risk accidentally introducing any more meta-inference changes...

Fair point! Correspondingly, I can be too impatient/imagine I can reliably see further than is necessarily the case. Small increments might have to be my watchword...

src/Data/Vec/Base.agda
```diff
@@ -17,7 +17,7 @@ open import Data.These.Base as These using (These; this; that; these)
 open import Function.Base using (const; _∘′_; id; _∘_)
 open import Level using (Level)
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl; trans; cong)
-open import Relation.Nullary.Decidable.Core using (does)
+open import Relation.Nullary.Decidable.Core using (does; T?)
```
jamesmckinna, line 20: Actually, no need for this import of `T?` here: it's not used anywhere!

src/Data/List/Base.agda
```diff
@@ -346,141 +346,160 @@ removeAt : (xs : List A) → Fin (length xs) → List A
...
   ... | true  = cons acc (go [] cs)
   ... | false = go (c ∷ acc) cs
 
+wordsByᵇ : (A → Bool) → List A → List (List A)
+wordsByᵇ p = wordsBy (T? ∘ p)
+
+derun : ∀ {R : Rel A p} → B.Decidable R → List A → List A
+derun R? [] = []
+derun R? (x ∷ []) = x ∷ []
+derun R? (x ∷ y ∷ xs) with does (R? x y) | derun R? (y ∷ xs)
```
jamesmckinna, line 461: Recommend (`style-guide`?) using an `@`-pattern here:
```agda
derun R? (x ∷ xs@(y ∷ _)) with does (R? x y) | derun R? xs
```

src/Data/List/Base.agda
```diff
@@ -346,141 +346,160 @@ removeAt : (xs : List A) → Fin (length xs) → List A
...
 partitionᵇ : (A → Bool) → List A → List A × List A
-partitionᵇ p []       = ([] , [])
-partitionᵇ p (x ∷ xs) = (if p x then Prod.map₁ else Prod.map₂′) (x ∷_) (partitionᵇ p xs)
+partitionᵇ p = partition (T? ∘ p)
+
+span : ∀ {P : Pred A p} → Decidable P → List A → (List A × List A)
+span P? []       = ([] , [])
+span P? (x ∷ xs) with does (P? x)
+... | true  = Prod.map (x ∷_) id (span P? xs)
+... | false = ([] , x ∷ xs)
```
jamesmckinna, line 410: Margin call, but I'd be tempted to use an `@`-pattern here, for the sake of the `false` branch.

jamesmckinna (approved):
Two global remarks:
* I'm no longer clear about the convoluted history here, but does any of this need a `CHANGELOG` entry?
* this PR seems only/largely to target *definitions*, rather than also their *properties*, and that's how/why you can get away with not investigating the `if_then_else_` forms rather than the `with` forms you have here. But what about the properties of the (new; former?) definitions? I guess my instincts would have been to investigate, but as with your earlier comment, such investigation can come later, I suppose, *but I think that it should*.

src/Data/Vec/Base.agda
MatthewDaggitt: Used in `countᵇ p = count (T? ∘ p)`

src/Data/List/Base.agda
MatthewDaggitt: Done

src/Data/List/Base.agda
MatthewDaggitt: Done

MatthewDaggitt:
> I'm no longer clear about the convoluted history here, but does any of this need a CHANGELOG entry?

So the old change should have had an entry. But this refactoring now means that the implementation is actually the same as it was before, so no entry needed!

> this PR seems only/largely to target definitions, rather than also their properties, and that's how/why you can get away with not investigating the if_then_else_ forms rather than the with forms you have here. But what about the properties of the (new; former?) definitions? I guess my instincts would have been to investigate, but as with your earlier comment, such investigation can come later, I suppose, but I think that it should.

The new definitions (the boolean versions) don't have any properties associated with them, and they were introduced in v2.0 so we don't need to worry about altering their behaviour :+1:

src/Data/Vec/Base.agda
jamesmckinna: 🤦

PR #2189, opened 2023-10-31
Move `T?` to `Relation.Nullary.Decidable.Core`

MatthewDaggitt:
The only thing I haven't addressed is @jamesmckinna's comment:

> And along with T?, please consider (re-)moving its property T?-diag as well (for the sake of keeping the import dependencies of Data.Bool.Properties as small as possible)

I didn't quite understand the reasoning here? Moving `T?-diag` to somewhere in `Relation.Nullary.Decidable.Core` wouldn't make the import dependencies any smaller, because `Data.Bool.Properties`, as expected, heavily already depends on `Relation.Nullary.Decidable.Core`....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -838,6 +838,9 @@ Non-backwards compatible changes
 
   4. The modules `Relation.Nullary.(Product/Sum/Implication)` have been deprecated
          and their contents moved to `Relation.Nullary.(Negation/Reflects/Decidable)`.
+		 
+  5. The proof `T?` has been moved from `Data.Bool.Properties` to `Relation.Nullary.Decidable.Core`
+	 (but is still re-exported by the former).
 
   as well as the following breaking changes:
 
@@ -3561,6 +3564,11 @@ Additions to existing modules
 
   isPartialOrder : IsPartialOrder _≡_ _≡_
   poset          : Set a → Poset _ _ _
+  ```
+
+* Added new proof in `Relation.Nullary.Reflects`:
+  ```agda
+  T-reflects : Reflects (T b) b
   ```
 
 * Added new operations in `System.Exit`:
--- a/notes/style-guide.md
+++ b/notes/style-guide.md
@@ -402,7 +402,7 @@ word within a compound word is capitalized except for the first word.
 
 * Rational variables are named `p`, `q`, `r`, ... (default `p`)
 
-* All other variables tend to be named `x`, `y`, `z`.
+* All other variables, including `Bool`, should be named `x`, `y`, `z`.
 
 * Collections of elements are usually indicated by appending an `s`
   (e.g. if you are naming your variables `x` and `y` then lists
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -29,7 +29,7 @@ open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality.Core
 open import Relation.Binary.PropositionalEquality.Properties
 open import Relation.Nullary.Reflects using (ofʸ; ofⁿ)
-open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no)
+open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no; fromWitness)
 import Relation.Unary as U
 
 open import Algebra.Definitions {A = Bool} _≡_
@@ -726,15 +726,17 @@ xor-∧-commutativeRing = ⊕-∧-commutativeRing
   open XorRing _xor_ xor-is-ok
 
 ------------------------------------------------------------------------
--- Miscellaneous other properties
+-- Properties of if_then_else_
 
-⇔→≡ : {x y z : Bool} → x ≡ z ⇔ y ≡ z → x ≡ y
-⇔→≡ {true } {true }         hyp = refl
-⇔→≡ {true } {false} {true } hyp = sym (Equivalence.to hyp refl)
-⇔→≡ {true } {false} {false} hyp = Equivalence.from hyp refl
-⇔→≡ {false} {true } {true } hyp = Equivalence.from hyp refl
-⇔→≡ {false} {true } {false} hyp = sym (Equivalence.to hyp refl)
-⇔→≡ {false} {false}         hyp = refl
+if-float : ∀ (f : A → B) b {x y} →
+           f (if b then x else y) ≡ (if b then f x else f y)
+if-float _ true  = refl
+if-float _ false = refl
+
+------------------------------------------------------------------------
+-- Properties of T
+
+open Relation.Nullary.Decidable.Core public using (T?)
 
 T-≡ : ∀ {x} → T x ⇔ x ≡ true
 T-≡ {false} = mk⇔ (λ ())       (λ ())
@@ -757,18 +759,20 @@ T-∨ {false} {false} = mk⇔ inj₁ [ id , id ]
 T-irrelevant : U.Irrelevant T
 T-irrelevant {true}  _  _  = refl
 
-T? : U.Decidable T
-does  (T? b) = b
-proof (T? true ) = ofʸ _
-proof (T? false) = ofⁿ λ()
-
 T?-diag : ∀ b → T b → True (T? b)
-T?-diag true  _ = _
+T?-diag b = fromWitness
 
-if-float : ∀ (f : A → B) b {x y} →
-           f (if b then x else y) ≡ (if b then f x else f y)
-if-float _ true  = refl
-if-float _ false = refl
+------------------------------------------------------------------------
+-- Miscellaneous other properties
+
+⇔→≡ : {x y z : Bool} → x ≡ z ⇔ y ≡ z → x ≡ y
+⇔→≡ {true } {true }         hyp = refl
+⇔→≡ {true } {false} {true } hyp = sym (Equivalence.to hyp refl)
+⇔→≡ {true } {false} {false} hyp = Equivalence.from hyp refl
+⇔→≡ {false} {true } {true } hyp = Equivalence.from hyp refl
+⇔→≡ {false} {true } {false} hyp = sym (Equivalence.to hyp refl)
+⇔→≡ {false} {false}         hyp = refl
+
 
 ------------------------------------------------------------------------
 -- DEPRECATED NAMES
--- a/src/Relation/Nullary.agda
+++ b/src/Relation/Nullary.agda
@@ -15,20 +15,9 @@ open import Agda.Builtin.Equality
 ------------------------------------------------------------------------
 -- Re-exports
 
-open import Relation.Nullary.Negation.Core public using
-  ( ¬_; _¬-⊎_
-  ; contradiction; contradiction₂; contraposition
-  )
-
-open import Relation.Nullary.Reflects public using
-  ( Reflects; ofʸ; ofⁿ
-  ; _×-reflects_; _⊎-reflects_; _→-reflects_
-  )
-
-open import Relation.Nullary.Decidable.Core public using
-  ( Dec; does; proof; yes; no; _because_; recompute
-  ; ¬?; _×-dec_; _⊎-dec_; _→-dec_
-  )
+open import Relation.Nullary.Negation.Core public
+open import Relation.Nullary.Reflects public
+open import Relation.Nullary.Decidable.Core public
 
 ------------------------------------------------------------------------
 -- Irrelevant types
--- a/src/Relation/Nullary/Decidable/Core.agda
+++ b/src/Relation/Nullary/Decidable/Core.agda
@@ -12,7 +12,7 @@
 module Relation.Nullary.Decidable.Core where
 
 open import Level using (Level; Lift)
-open import Data.Bool.Base using (Bool; false; true; not; T; _∧_; _∨_)
+open import Data.Bool.Base using (Bool; T; false; true; not; _∧_; _∨_)
 open import Data.Unit.Base using (⊤)
 open import Data.Empty using (⊥)
 open import Data.Empty.Irrelevant using (⊥-elim)
@@ -78,6 +78,10 @@ recompute (no ¬a) a = ⊥-elim (¬a a)
 
 infixr 1 _⊎-dec_
 infixr 2 _×-dec_ _→-dec_
+
+T? : ∀ b → Dec (T b)
+does  (T? b) = b
+proof (T? b) = T-reflects b
 
 ¬? : Dec A → Dec (¬ A)
 does  (¬? a?) = not (does a?)
--- a/src/Relation/Nullary/Decidable/Core.agda
+++ b/src/Relation/Nullary/Decidable/Core.agda
@@ -79,9 +79,9 @@ recompute (no ¬a) a = ⊥-elim (¬a a)
 infixr 1 _⊎-dec_
 infixr 2 _×-dec_ _→-dec_
 
-T? : ∀ b → Dec (T b)
-does  (T? b) = b
-proof (T? b) = T-reflects b
+T? : ∀ x → Dec (T x)
+does  (T? x) = x
+proof (T? x) = T-reflects x
 
 ¬? : Dec A → Dec (¬ A)
 does  (¬? a?) = not (does a?)
--- a/src/Relation/Nullary/Reflects.agda
+++ b/src/Relation/Nullary/Reflects.agda
@@ -11,11 +11,12 @@ module Relation.Nullary.Reflects where
 open import Agda.Builtin.Equality
 
 open import Data.Bool.Base
+open import Data.Unit.Base using (⊤)
 open import Data.Empty
 open import Data.Sum.Base using (_⊎_; inj₁; inj₂)
 open import Data.Product.Base using (_×_; _,_; proj₁; proj₂)
 open import Level using (Level)
-open import Function.Base using (_$_; _∘_; const)
+open import Function.Base using (_$_; _∘_; const; id)
 
 open import Relation.Nullary.Negation.Core
 
@@ -23,6 +24,7 @@ private
   variable
     a : Level
     A B : Set a
+    x y : Bool
 
 ------------------------------------------------------------------------
 -- `Reflects` idiom.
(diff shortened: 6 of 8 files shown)


src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -79,6 +79,10 @@ recompute (no ¬a) a = ⊥-elim (¬a a)
 infixr 1 _⊎-dec_
 infixr 2 _×-dec_ _→-dec_
 
+T? : ∀ x → Dec (T x)
+does  (T? x) = x
+proof (T? x) = T-reflects x
+
```
jamesmckinna, line 85: I think this is more pungently expressed in terms of `_because_`?
```agda
T? x = x because (T-reflects x)
```

src/Relation/Nullary/Reflects.agda
```diff
@@ -43,54 +45,54 @@ data Reflects (A : Set a) : Bool → Set a where
...
+
+T-reflects : ∀ x → Reflects (T x) x
+T-reflects true  = of _
+T-reflects false = of id
+
 -- If we can decide A, then we can decide its negation.
-¬-reflects : ∀ {b} → Reflects A b → Reflects (¬ A) (not b)
+¬-reflects : Reflects A x → Reflects (¬ A) (not x)
 ¬-reflects (ofʸ  a) = ofⁿ (_$ a)
 ¬-reflects (ofⁿ ¬a) = ofʸ ¬a
```
jamesmckinna, line 69: See also #2149 which naturally could be merged after this, or as part of it: something to discuss?

src/Relation/Nullary/Reflects.agda
```diff
@@ -43,54 +45,54 @@ data Reflects (A : Set a) : Bool → Set a where
...
+fromEquivalence : (T x → A) → (A → T x) → Reflects A x
+fromEquivalence {x = true}  sound complete = ofʸ (sound _)
+fromEquivalence {x = false} sound complete = ofⁿ complete
 
 -- `Reflects` is deterministic.
-det : ∀ {b b′} → Reflects A b → Reflects A b′ → b ≡ b′
+det : Reflects A x → Reflects A y → x ≡ y
 det (ofʸ  a) (ofʸ  _) = refl
 det (ofʸ  a) (ofⁿ ¬a) = contradiction a ¬a
 det (ofⁿ ¬a) (ofʸ  a) = contradiction a ¬a
```
jamesmckinna, line 98: As with my comments on the related PR #2186 I suggest that you add the corollary (what to call it?)
```agda
T⁻ : ∀ {x y} → Reflects (T x) y → y ≡ x
T⁻ {x} r = det r (T-reflects x)
```

CHANGELOG.md
```diff
@@ -838,6 +838,9 @@ Non-backwards compatible changes
 
   4. The modules `Relation.Nullary.(Product/Sum/Implication)` have been deprecated
          and their contents moved to `Relation.Nullary.(Negation/Reflects/Decidable)`.
+		 
+  5. The proof `T?` has been moved from `Data.Bool.Properties` to `Relation.Nullary.Decidable.Core`
+	 (but is still re-exported by the former).
```
jamesmckinna, line 843: Is the re-export, as opposed to an explicit deprecation here, th best choice? Eg `Data.Nat.Properties` imports *only* `T?` from `Data.Nat.Properties` (and presumably also some of the downstream uses of `Decidable` relations in eg your work on #2186 

I guess my point in all my comments on this topic is that, if you are going to the trouble of importing `Relation.*.Decidable`, then it's an unnecessary detour to re-import from `Data.Bool.Properties` structure which would (after the present PR is merged) be already present...
... and this seems to be for the sake of ... well I'm not sure exactly?

CHANGELOG.md
```diff
@@ -3563,6 +3566,11 @@ Additions to existing modules
   poset          : Set a → Poset _ _ _
   ```
 
+* Added new proof in `Relation.Nullary.Reflects`:
+  ```agda
+  T-reflects : Reflects (T b) b
+  ```
```
jamesmckinna, line 3572: So, I chose the name `T⁺` in #2055 on the model of the style-guide for "pre- and post-conditions". I don't disagree with the new name per se, but then what would you call `T⁻` defined below? In any case, I don't think these names should ever be exported  outside `Relation.Nullary.Decidable.Core`...?

notes/style-guide.md
```diff
@@ -402,7 +402,7 @@ word within a compound word is capitalized except for the first word.
 
 * Rational variables are named `p`, `q`, `r`, ... (default `p`)
 
-* All other variables tend to be named `x`, `y`, `z`.
+* All other variables, including `Bool`, should be named `x`, `y`, `z`.
```
jamesmckinna, line 405: Interesting choice. Why make it here? (Or: why are the numeric types special? isn't `Bool` also special?)

Does it belong as part of this PR? (It's clearly had lots of knock-on consequences throughout the PR)

src/Data/Bool/Properties.agda
```diff
@@ -29,7 +29,7 @@ open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality.Core
 open import Relation.Binary.PropositionalEquality.Properties
 open import Relation.Nullary.Reflects using (ofʸ; ofⁿ)
-open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no)
+open import Relation.Nullary.Decidable.Core using (True; does; proof; yes; no; fromWitness)
```
jamesmckinna, line 32: Similarly, `does; proof` in the import list here.
And as I have observed elsewhere, it's not clear to me, given the redefinition of `T?-diag` in terms of `fromWitness` that `T?-diag` as a name shouldn't also be deprecated in favour of the latter. Or even removed: is it used anywhere by anyone?

src/Data/Bool/Properties.agda
```diff
@@ -29,7 +29,7 @@ open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality.Core
 open import Relation.Binary.PropositionalEquality.Properties
 open import Relation.Nullary.Reflects using (ofʸ; ofⁿ)
```
jamesmckinna, line 31: This import can be removed entirely; it is only required by the definition of `T?`.

src/Data/Bool/Properties.agda
```diff
@@ -726,15 +726,17 @@ xor-∧-commutativeRing = ⊕-∧-commutativeRing
...
-⇔→≡ {false} {false}         hyp = refl
+if-float : ∀ (f : A → B) b {x y} →
+           f (if b then x else y) ≡ (if b then f x else f y)
+if-float _ true  = refl
+if-float _ false = refl
+
+------------------------------------------------------------------------
+-- Properties of T
+
+open Relation.Nullary.Decidable.Core public using (T?)
```
jamesmckinna, line 739: I see: the import/re-export happens here. I would add `True` to the `using` list here, and remove it from the module prelude imports on L32.

src/Relation/Nullary.agda
```diff
@@ -15,20 +15,9 @@ open import Agda.Builtin.Equality
...
-  ( Reflects; ofʸ; ofⁿ
-  ; _×-reflects_; _⊎-reflects_; _→-reflects_
-  )
-
-open import Relation.Nullary.Decidable.Core public using
-  ( Dec; does; proof; yes; no; _because_; recompute
-  ; ¬?; _×-dec_; _⊎-dec_; _→-dec_
-  )
+open import Relation.Nullary.Negation.Core public
+open import Relation.Nullary.Reflects public
```
jamesmckinna, line 19: Fine. Working on #2149 encouraged me to think about subsequent tightening of these export lists (esp. concerning `Reflects`), but that can follow later.

jamesmckinna (changes requested):
The PR is fine as it stands, but if we are going to merge this, I think the comments indicate plenty of places where it can be tightened up, or even made more drastic wrt deprecation/removal of clutter such as `T?-diag`, whose definition/use could eg be moved to a `README` module such as `README.Bool` or `README.Design.Decidability`. I don't personally think it belongs in the library. 

Likewise (see #2149), I think we should avoid using *patterns* `yes` and `no` (they are needlessly strict, and expose too much of `Relation.Nullary.Reflects`) from `Relation.Nullary.Decidable.Core` in favour of *definitions* as `true because (of _)` and `false because (of _)`, but that's a larger breaking change for another time ;-)

src/Relation/Nullary.agda
```diff
@@ -15,20 +15,9 @@ open import Agda.Builtin.Equality
...
+open import Relation.Nullary.Decidable.Core public
```
jamesmckinna, line 20: Test failure suggest this import needs to (at least) hide `⌊_⌋`, which is a renamed clone of `isYes`. Or else remove the former entirely.

src/Data/Bool/Properties.agda
jamesmckinna: Added: against the possibility of moving the whole import to the deprecation section...

jamesmckinna:
> The only thing I haven't addressed is @jamesmckinna's comment:
> 
> > And along with T?, please consider (re-)moving its property T?-diag as well (for the sake of keeping the import dependencies of Data.Bool.Properties as small as possible)
> 
> I didn't quite understand the reasoning here? Moving `T?-diag` to somewhere in `Relation.Nullary.Decidable.Core` wouldn't make the import dependencies any smaller, because `Data.Bool.Properties`, as expected, heavily already depends on `Relation.Nullary.Decidable.Core`....

Well, as in my review comments, the import list would lose that of `Relation.Nullary.Reflects` altogether.
And the only uses of `Relation.Nullary.Decidable.Core` are then in three proofs of `Decidable _` for various properties (which could, eventually, be delegated to instances of `T?`, I think), in the proof `T?` itself (but that's now gone), as well as `T?-diag`, which I think should be removed. So not quite so 'heavy', I think ;-)

src/Relation/Nullary/Decidable/Core.agda
MatthewDaggitt: Sure!

src/Relation/Nullary/Reflects.agda
MatthewDaggitt: I think let's keep that as a separate PR?

notes/style-guide.md
MatthewDaggitt: The numeric types are special, because they have special letters associated with them. I did think that `Bool` should use `a`, `b`, `c` etc, but then I remembered that we can't use `variable`s then because `a` those letters are already reserved for `Level`s. So I thought best to document this explicitly in the style-guide.

As for whether it belongs in the PR, it was a consequence of trying to use `variable`s in `Decidable.Core`, and then running into the above issue. I'm happy to separate it out if you would prefer?

src/Data/Bool/Properties.agda
MatthewDaggitt: Nice spot!

src/Data/Bool/Properties.agda
MatthewDaggitt: Err I don't think we should be exporting `True` publicly from `Data.Bool.Properties`?

src/Relation/Nullary.agda
MatthewDaggitt: I'm not sure we don't want to tighten them. I think in general we just want to export everything via `Relation.Nullary`. Otherwise, we give ourselves a massive maintenance headache of remembering to update `Relation.Nullary` every time we update one of the submodules?

src/Data/Bool/Properties.agda
MatthewDaggitt: Removed them.

CHANGELOG.md
MatthewDaggitt: I chose this name because it was consistent with the other properties. I don't think this falls into `pre- and post- conditions` convention in the style guide. 

> When defining a new relation P over a datatype X in a Data.X.Relation module, it is often common to define how to introduce and eliminate that relation with respect to various functions.

Firstly, we're not in a `Data.X.Relation` module. Secondly, I guess you could argue that `P` is `Reflects` but then what is the datatype `X` and what is the function that eliminates or introduces it?

I'll have a think about a suitable name.

MatthewDaggitt:
Sorry, haven't had time to answer all your review comments. Will come back to them later today.

notes/style-guide.md
jamesmckinna: I think it would be better to legislate on such style points as a separate activity, so yes if it's easy to revert those parts of the PR, then please do...

As for `a` ... for `Level`s, I think that can be too dogmatic, especially when `\ell` is a (very) palatable alternative. And some modules seem much too promiscuous in reserving such names, and then not really using them, when Agda's level polymorphism in fact seems to permit 'overloaded' use of such variables.

And yes, I'd prefer `b` for `Bool`eans, but I think the balance between rigid style rules, consistency, and cognitive efficiency/comfort is a tricky negotiation.  In among all the swirl of change in v2.0, I think this is worth putting out for separate discussion/design/implementation?

src/Data/Bool/Properties.agda
jamesmckinna: My mistake! I'd forgotten (or: chosen to wish away) the `public` qualifier!

src/Relation/Nullary.agda
jamesmckinna: Makes sense, modulo the need to hide certain things (as with the failing test). Again I personally find the trade off between narrowing imports via `using` (a kind of closed world assumption) vs maximally permissive via `hiding` (open world) again, ... a tricky negotiation. For this particular corner of the library, given how much change there has already been to the various representations, exercising some control over visibility (esp. of constructors/patterns, which shouldn't necessarily be exposed) seems a good idea. But happy to be open world for now!

CHANGELOG.md
jamesmckinna: I think that's a fair response (modulo my regarding `Reflects` as the datatype, or else as a `Set`-valued view of `Bool`, but in any case, as an inductive family; and then the 'function' is... `T` itself; but I admit those are the possibly-unconventional spectacles through which I look at the world), and given the rest of `Reflects`, the choice of name here seems fine... up to what its dual/inverse should be called. But I do think the principle that, just as with introduction/elimination, these things 'come in pairs', should be adhered to wherever possible. The `T⁺`/`T⁻` scheme is then a convenient, and minimum-ink, reification of that principle. But this probably belongs in a separate `style-guide` collection of issues... ;-)

src/Relation/Nullary/Reflects.agda
```diff
@@ -43,54 +45,54 @@ data Reflects (A : Set a) : Bool → Set a where
...
+infixr 1 _⊎-reflects_
+infixr 2 _×-reflects_ _→-reflects_
+
+T-reflects : ∀ x → Reflects (T x) x
+T-reflects true  = of _
+T-reflects false = of id
+
 -- If we can decide A, then we can decide its negation.
-¬-reflects : ∀ {b} → Reflects A b → Reflects (¬ A) (not b)
+¬-reflects : Reflects A x → Reflects (¬ A) (not x)
```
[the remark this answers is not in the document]
jamesmckinna: As above, happy to let this one go... for now ;-)

jamesmckinna:
I've self-resolved a number of the `naming` issues, for the sake of documentation, and the possibility to return to them another day. 

As for `T?`, please think hard about whether you do want it re-exported from `Data.Bool.Properties` (bearing in mind what modules do actually import it, namely the sole `Data.Nat.Properties`...), and I'll go with your eventual decision. So happy to lift the review merge embargo, pending the resolution of the failing test...

jamesmckinna (approved):

notes/style-guide.md
MatthewDaggitt: Okay, I removed the changes.

src/Relation/Nullary.agda
MatthewDaggitt: > Again I personally find the trade off between narrowing imports via using (a kind of closed world assumption) vs maximally permissive via hiding (open world) again, ... a tricky negotiation. 

I think my mindset is that there are two ways to publicly have module A re-export from module B. Either a) it exports a small fixed closed set of names which are not envisaged to grow as new things are added to B, or b) module B should be designed so that module A should be able to re-export all of it. Anything else is asking for huge maintenance problems in my opinion.

src/Relation/Nullary.agda
MatthewDaggitt: I've hidden it in `Data.List.Relation.Binary.BagAndSetEquality` as that module should have never have been opening the full `Relation.Nullary` in the first place.

MatthewDaggitt:
> Well, as in my review comments, the import list would lose that of Relation.Nullary.Reflects altogether.
And the only uses of Relation.Nullary.Decidable.Core are then in three proofs of Decidable _ for various properties (which could, eventually, be delegated to instances of T?, I think, on the model of Data.Nat.Properties._≟_ etc.), in the proof T? itself (but that's now gone), as well as T?-diag, which I think should be removed. So not quite so 'heavy', I think ;-)

I think we may be talking at cross purposes here. Sure you can _syntactically_ remove the direct dependency on `Relation.Nullary.Decidable.Core`, but you're never going to be able to remove the _semantic_ dependency. After all we want to prove decidability of propositional equality over `Bool` so we're going to have to import `Relation.Nullary.Decidable.Core`  (and to be honest `Relation.Nullary.Decidable`) at some point in it's dependency graph.

src/Relation/Nullary/Reflects.agda
MatthewDaggitt: Added as `T-reflects-elim` for now. As discussed below, we can open a separate issue if we want to start renaming all the lemmas in this module.

CHANGELOG.md
MatthewDaggitt: I think my view is that `T` is defined in `Data.Bool.Base` and therefore we should export the property that it is decidable from `Data.Bool.Properties`. The fact that dependency cycles force us to actually prove it earlier in the dependency graph outside of the `Data.Bool` folder is largely irrelevant.

jamesmckinna: OK!

src/Relation/Nullary.agda
jamesmckinna: You make a convincing case. In this instance, "module B" (`Relation.Nullary.Decidable.Core`) has been a fast-moving target, suggesting option b) is to be preferred.

jamesmckinna:
> > Well, as in my review comments, the import list would lose that of Relation.Nullary.Reflects altogether.
> > And the only uses of Relation.Nullary.Decidable.Core are then in three proofs of Decidable _ for various properties (which could, eventually, be delegated to instances of T?, I think, on the model of Data.Nat.Properties._≟_ etc.), in the proof T? itself (but that's now gone), as well as T?-diag, which I think should be removed. So not quite so 'heavy', I think ;-)
> 
> I think we may be talking at cross purposes here. Sure you can _syntactically_ remove the direct dependency on `Relation.Nullary.Decidable.Core`, but you're never going to be able to remove the _semantic_ dependency. After all we want to prove decidability of propositional equality over `Bool` so we're going to have to import `Relation.Nullary.Decidable.Core` (and to be honest `Relation.Nullary.Decidable`) at some point in it's dependency graph.

Agreed! (both to the semantic point, and to the crossed purposes...)

That said, consider the following characterisation result for `Dec`, that *everything* reduces to `T?`:
```agda
dec-T∘does⁺ : ∀ (A? : Dec A) → A → T (does A?)
dec-T∘does⁺ (true because _) _ = _
dec-T∘does⁺ (false because [¬a]) a = contradiction a (invert [¬a])

dec-T∘does⁻ : ∀ (A? : Dec A) → T (does A?) → A
dec-T∘does⁻ (true because [a]) _ = invert [a]
dec-T∘does⁻ (false because _) ()

T?⇒Dec : ∀ b → (T b → A) → (A → T b) → Dec A
T?⇒Dec b to from = map′ from to (T? b)
```
(the first two are lazier versions of `toWitness` and `fromWitness`, IIUC), while the final lemma is the essence of how `DecidableEquality` relations get defined (eg in `Data.Nat.Properties`; other `Decidable` relations there as well!), so it feels as though we might as well use `T?⇒Dec` *throughout the library* to yield proofs of `Decidable` (and eg perhaps specialise it in `Relation.Binary.Properties.DecidableEquality` for example), given suitable `Bool`-valued functions). Then `T?` per se, apart from its (legitimate, as a property of `Data.Bool.Base.T`: you've convinced me!) re-export in `Data.Bool.Properties` never after needs to be explicitly appealed to, when it can always be delegated to in the form of suitable `to` and `from` functions supplied to `T?⇒Dec` (naming?)...

jamesmckinna:
I'll do a final readthough of the commits, then hit "Merge". Thanks for your willingness to debate the issues!

CHANGELOG.md
```diff
@@ -3563,6 +3566,12 @@ Additions to existing modules
   poset          : Set a → Poset _ _ _
   ```
 
+* Added new proof in `Relation.Nullary.Reflects`:
+  ```agda
+  T-reflects : Reflects (T b) b
+  T-reflects-elim : Reflects (T a) b → b ≡ a
```
jamesmckinna, line 3572: :+1:

src/Data/List/Relation/Binary/BagAndSetEquality.agda
```diff
@@ -44,7 +44,7 @@ import Relation.Binary.Reasoning.Preorder as PreorderReasoning
 open import Relation.Binary.PropositionalEquality as P
   using (_≡_; _≢_; _≗_; refl)
 open import Relation.Binary.Reasoning.Syntax
-open import Relation.Nullary
+open import Relation.Nullary.Negation using (¬_)
```
jamesmckinna, line 47: A better fix than my `hiding` suggestion!

PR #2190, opened 2023-10-31
Added tabulate+<

guilhermehas:
`tabulate⁺<` is more general than `tabulate⁺`


src/Data/List/Relation/Unary/AllPairs/Properties.agda
```diff
@@ -115,12 +115,16 @@ module _ {R : Rel A ℓ} where
 
 module _ {R : Rel A ℓ} where
 
+  tabulate⁺< : ∀ {n} {f : Fin n → A} → (∀ {i j} → i F.< j → R (f i) (f j)) →
```
MatthewDaggitt, line 118: Could we call this `tabulate⁺-<`?

src/Data/List/Relation/Unary/AllPairs/Properties.agda
```diff
@@ -115,12 +115,16 @@ module _ {R : Rel A ℓ} where
...
+              AllPairs R (tabulate f)
+  tabulate⁺< {zero}  fᵢ~fⱼ = AllPairs.[]
```
MatthewDaggitt, line 120: I don't think we need the `AllPairs.` qualifiers anywhere in this proof?

MatthewDaggitt (changes requested):
Otherwise looks good!

src/Data/List/Relation/Unary/AllPairs/Properties.agda
guilhermehas: I renamed it.

src/Data/List/Relation/Unary/AllPairs/Properties.agda
guilhermehas: I removed it.

MatthewDaggitt (approved):

PR #2191, opened 2023-11-02
Add `Function.Consequences.Setoid`

MatthewDaggitt:
Makes transforming between different function properties far far easier, when both the domain and codomain relations are setoids.

This in preparation for the change I discussed in the meeting of switching back to the old definitions in the `Function` structures and bundles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2013,6 +2013,7 @@ New modules
   ```
   Function.Consequences
   Function.Consequences.Setoid
+  Function.Consequences.Propositional
   Function.Properties.Bijection
   Function.Properties.RightInverse
   Function.Properties.Surjection
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -752,7 +752,7 @@ Non-backwards compatible changes
          of these under the names `strictlySurjective`, `strictlyInverseˡ` and
          `strictlyInverseʳ`,
    - Conversion functions have been added in both directions to
-         `Function.Consequences(.Propositional)`.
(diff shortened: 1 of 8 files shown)


MatthewDaggitt:
Finally found the problem. The bug was that `contraInjection` in `Function.Consequences` needs the second equality to be explicit. As it was added in v2.0 it's a non-breaking change.

jamesmckinna:
@MatthewDaggitt wrote: 

> This in preparation for the change I discussed in the meeting of switching back to the old definitions in the `Function` structures and bundles.

For the sake of the GitHub record, can you summarise that discussion here? I've already forgotten the issue or the proposed solution... :-(

MatthewDaggitt:
The discussion was about the strict vs non-strict definitions of `Surjective`/`LeftInverse` etc. We switched to the new definitions, including in the fields of the `Function.Structures` and `Function.Bundles`, but given that they're equivalent when you're working with setoids, it makes more sense to keep the old definitions which are usually easier to work with, and then simply prove the new definitions.

MatthewDaggitt:
Regardless, I think this PR stands on it's own, as making the `Function` hierarchy easier to work with and so should be merged into v2.0.

jamesmckinna:
Not a review alas, as I still don't consider myself competent to rule on the subtleties of any potential downstream consequences of changes to `Function.*`. But seeing as most of the material is additional, or else a delegation to existing functionality in the presence of specialisation to `Setoid`s given by `Propositional` equality, I'm going to presume it's all OK... what could go wrong?

MatthewDaggitt:
Could I get a review on this from someone?

PR #2204, opened 2023-11-16
fix Connex comment

Jesin:
Connex allows both relations to hold, so the old comment was wrong.


MatthewDaggitt (approved):
Nice spot, thanks!

PR #2206, opened 2023-11-17
[fixes #2168] Change names in `Algebra.Consequences.*` to reflect `style-guide` conventions

### LINKED ISSUE #2168: Choice of pre/post-condition combinators in proof names
CHANGELOG.md  in  lib-2.rc1  writes  
```
reflexive+selfInverse⇒involutive  : ...
```
In other places (in lib-1.7.2),  I recall,  `∧`  is used  in similar situations in the name, probably, in Data.Nat.Properties  ...
It is also reasonable to use comma.
And `+` presumes a different meaning in  the names like  ``m≤n+n``.
Anyway it has sense to choose a uniform denotation for the library.

jamesmckinna:
Pure bureaucracy!
Only outstanding issue seems to be whether I have, in fact, got the `CHANGELOG` and deprecation warnings correct.

### DIFF
--- a/src/Algebra/Consequences/Propositional.agda
+++ b/src/Algebra/Consequences/Propositional.agda
@@ -39,6 +39,17 @@ open Base public
   ; subst∧comm⇒sym
   ; wlog
   ; sel⇒idem
+-- plus all the deprecated versions of the above
+  ; comm+assoc⇒middleFour
+  ; identity+middleFour⇒assoc
+  ; identity+middleFour⇒comm
+  ; assoc+distribʳ+idʳ+invʳ⇒zeˡ
+  ; assoc+distribˡ+idʳ+invʳ⇒zeʳ
+  ; assoc+id+invʳ⇒invˡ-unique
+  ; assoc+id+invˡ⇒invʳ-unique
+  ; comm+distrˡ⇒distrʳ
+  ; comm+distrʳ⇒distrˡ
+  ; subst+comm⇒sym
   )
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


src/Algebra/Consequences/Propositional.agda
```diff
@@ -26,17 +26,17 @@ import Algebra.Consequences.Setoid (setoid A) as Base
...
+  ; assoc∧id∧invʳ⇒invˡ-unique
+  ; assoc∧id∧invˡ⇒invʳ-unique
+  ; comm∧distrˡ⇒distrʳ
+  ; comm∧distrʳ⇒distrˡ
   ; comm⇒sym[distribˡ]
-  ; subst+comm⇒sym
+  ; subst∧comm⇒sym
   ; wlog
   ; sel⇒idem
   )
```
MatthewDaggitt, line 42: We should still be hiding the old deprecated names here. Otherwise, they suddenly appear in scope!

src/Algebra/Consequences/Propositional.agda
```diff
@@ -100,31 +100,41 @@ module _ {_∙_ : Op₂ A} where
...
+
+
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.0
+-- See all the names deprecated in Algebra.Consequences.{Base|Setoid}
```
MatthewDaggitt, line 140: We still need to create deprecated versions of the old names here, as they have different arguments to the ones exported by `Algebra.Consequences.Setoid`.

MatthewDaggitt (changes requested):
Otherwise looks good. Thank you for this huge effort!

src/Algebra/Consequences/Propositional.agda
jamesmckinna: Aaargh... I dithered over this, and realised I wasn't quite sure of myself. Thanks for picking this up.

src/Algebra/Consequences/Propositional.agda
jamesmckinna: Ah. For some reason, I thought that it was enough for the use of deprecated *names* to trigger the warnings, so that that behaviour could 'delegate' to `...Setoid`. More confusion/ignorance/uncertainty on my part. I'll fix when I'm awake. I'm not really awake now... it's my middle of the night insomnia phase.

src/Algebra/Consequences/Propositional.agda
jamesmckinna: I opted to do this in a block, rather than pairing line-by-line. Hope that's OK!

src/Algebra/Consequences/Propositional.agda
jamesmckinna: Done now, again I hope successfully.

MatthewDaggitt: Thanks, that looks great!

src/Algebra/Consequences/Propositional.agda
MatthewDaggitt: A single `hiding` directive is fine!

MatthewDaggitt (approved):

PR #2208, opened 2023-11-22
Add `map` to `Data.String.Base`

### LINKED ISSUE #2207: `Data.String` missing `map`
Hi I was using Data.String and I could not find a way to map overs strings. I end up with:
```agda
map : (Char → Char) → String → String
map f = fromList ∘ List.map f ∘ toList
```

would it make sense to add to [Data/String/Base](https://github.com/agda/agda-stdlib/blob/master/src/Data/String/Base.agda)?

I would love to make such contribution 😄

lemastero:
Add map to Data.String.Base with unit tests propose in #2207.


MatthewDaggitt (changes requested):
Looks great! Can you add an entry to the CHANGELOG?

lemastero:
@MatthewDaggitt thank you for review!  CHANGELOG [now](https://github.com/agda/agda-stdlib/pull/2208/commits/d1e303e3f637b0b94a4d3343e2ff43df290647bd) has map in list of new functions for Data.String.Base.

MatthewDaggitt (approved):

MatthewDaggitt:
Thanks, looks great now. I'm afraid this won't make it into `v2.0` but it'll be in `v2.1`. Will merge it in, in a bit when we sort out the branches for the next version.

lemastero:
[CI build was failing](https://github.com/agda/agda-stdlib/actions/runs/6981162499/job/19013445223) - conflict on [map in Text/Printf/Generic](https://github.com/agda/agda-stdlib/blob/master/src/Text/Printf/Generic.agda#L59-L61):
```text
/home/runner/work/agda-stdlib/agda-stdlib/src/Text/Printf/Generic.agda:59,3-6
Multiple definitions of map. Previous definition at
/home/runner/work/agda-stdlib/agda-stdlib/src/Data/String/Base.agda:189,1-4
  map : ∀ pr → (A → B) → Printf pr A → Printf pr B
```
so I [pushed 1 more commit](https://github.com/agda/agda-stdlib/pull/2208/commits/704ef159442d4c4530aa11f6696d789ec10a7b31) that limits imports there.

Thank you very much for helping me with my first PR 🙇‍♂️ 
Contributions to widely used, open source libraries, feels like I am doing something useful & good in the world :)

MatthewDaggitt:
> Thank you very much for helping me with my first PR 🙇‍♂️
Contributions to widely used, open source libraries, feels like I am doing something useful & good in the world :)

Haha, you're welcome, thank for contributing. Careful, as @jamesmckinna might be able to attest, it can get a bit addictive!

jamesmckinna:
I for one, am *very* happy for the contribution, and the idealism you express/that it expresses. 
Please keep them coming! (I see you already have... ;-))
But please also be careful/mindful to manage your time/energy/resources on what is still a _volunteer_ project ;-)

MatthewDaggitt:
Hmm @lemastero we're still getting test failures unfortunately. You can run the tests locally with the command `make test`.

Taneb:
The imports of `Data.String.Base` in `IO` and `Data.Tree.Rose.Show` need adjusting. The former could be removed entirely, the module doesn't actually use anything from there directly.

lemastero:
I will check. BTW what is a good way to build things locally? Perhaps I have missed section in documentation about it.

A lot of things are happening in  [.github/workflows/haskell-ci.yml](https://github.com/agda/agda-stdlib/blob/master/.github/workflows/haskell-ci.yml), perhaps there is shorter list of commands to run 😅

Taneb:
`make test` is what I do. You'll need to merge in master, though, the commit you branched off had whitespace issues

jamesmckinna:
Re: `make test` for local testing of PRs ahead of committing to your remote branch
* no need to invoke the CI machinery; the `GNUmakefile` has all you need;
* there are several phases: `fix-whitespace`, which you probably anyway can/should run separately before any `push` to the remote branch; `setup` which regenerates the `Everything` and `EverythingSafe` top-level modules in `doc/` via `GenerateEverything`; and finally the main invocation of `agda` on `doc/README`.

So a short(er) cut to testing, once you've run `GenerateEverything` is simply to run the final step. It's definitely compute-, memory-, and disk- intensive (writing build artefacts to the `_build` subdirectory), so I've taken to streamlining it a bit further by tweaking the haskell RTS arguments to my setup (YMMV), running `agda` additionally with `--no-default-libraries`, and after a couple of years now, also with `--verbose=0` to avoid any output other than failing errors... HTH!

PR #2209, opened 2023-11-24
Fix #2195 by removing redundant zero from IsRing

### LINKED ISSUE #2195: Redundant `zero` in `IsRing`
When defining `IsRing`, `zero` is provided in the definition. Instead, we can prove `zero`  as its property.

Here is proof of right zero

```
RightZero : ∀ x → x * 0# ≈ 0# 
RightZero x = begin 
  x * 0#                    ≈⟨ sym(+-identityʳ _) ⟩ 
  x * 0# + 0#               ≈⟨ +-congˡ (sym(-‿inverseʳ  x)) ⟩ 
  x * 0# + (x - x)          ≈⟨ sym(+-assoc _ _ _) ⟩ 
  (x * 0# + x) - x          ≈⟨ +-congʳ (+-congˡ ( sym(*-identityʳ x))) ⟩ 
  ((x * 0#) + (x * 1#)) - x ≈⟨ +-congʳ (sym(distribˡ _ _ _ )) ⟩ 
  (x * (0# + 1#)) - x       ≈⟨ +-congʳ (*-congˡ ( +-identityˡ 1# )) ⟩ 
  (x * (1#)) - x            ≈⟨ +-congʳ (*-identityʳ x) ⟩ 
  x - x                     ≈⟨ -‿inverseʳ  x ⟩ 
  0#                        ∎  
```

Similarly, we can prove left zero. This proof should be used in `-‿distribˡ-* : ∀ x y → - (x * y) ≈ - x * y` and `-‿distribʳ-* : ∀ x y → - (x * y) ≈ x * - y`
Also in the literature, I don't see `zero` given in the definition. Was this part of some design decision?

MatthewDaggitt:


src/Algebra/Structures.agda
```diff
@@ -862,10 +861,15 @@ record IsRing (+ * : Op₂ A) (-_ : Op₁ A) (0# 1# : A) : Set (a ⊔ ℓ) where
...
 
   zeroˡ : LeftZero 0# *
-  zeroˡ = proj₁ zero
+  zeroˡ = Consequences.assoc+distribʳ+idʳ+invʳ⇒zeˡ setoid
+    +-cong *-cong +-assoc (proj₂ distrib) +-identityʳ -‿inverseʳ
 
   zeroʳ : RightZero 0# *
-  zeroʳ = proj₂ zero
+  zeroʳ = Consequences.assoc+distribˡ+idʳ+invʳ⇒zeʳ setoid
+    +-cong *-cong +-assoc (proj₁ distrib) +-identityʳ -‿inverseʳ
```
jamesmckinna, line 869: Two comments on this, the substantive part of the issue/PR:
* it would good to merge #2206 and use the corrected/renamed versions of these lemmas from `Consequences` (we might debate the order of merging, but each PR along the lines of the present one adds to the knock-on viscosity associated with the renaming...)
* have you checked the 'most general' point in the `Structures`/`Bundles` hierarchy that this lemma can be placed? AFAICT, it can already be proved at `IsNearring`, but possibly more general still?

src/Algebra/Structures.agda
```diff
@@ -862,10 +861,15 @@ record IsRing (+ * : Op₂ A) (-_ : Op₁ A) (0# 1# : A) : Set (a ⊔ ℓ) where
     )
 
   zeroˡ : LeftZero 0# *
-  zeroˡ = proj₁ zero
+  zeroˡ = Consequences.assoc+distribʳ+idʳ+invʳ⇒zeˡ setoid
+    +-cong *-cong +-assoc (proj₂ distrib) +-identityʳ -‿inverseʳ
```
jamesmckinna, line 865: I think that there should be a definable `distribʳ` already in scope here, so you don't need to explicit projection?

src/Algebra/Structures.agda
jamesmckinna, line 869: Ditto `distribˡ`?

MatthewDaggitt: Good point. I've moved the proof to `IsRingWithoutOne`. And yes the other PR will be merged as well.

src/Algebra/Structures.agda
MatthewDaggitt: Yup moving the proof up the hierarchy has solved this problem.

jamesmckinna:
So I think for the sake of completeness, I would add/revise the definition of `IsNearring` to the following, as I don't see how to reconcile it with `IsRingWithoutOne`. Let me know what you think:
```agda
record IsNearring (+ * : Op₂ A) (0# 1# : A) (_⁻¹ : Op₁ A) : Set (a ⊔ ℓ) where
  field
    isQuasiring : IsQuasiring + * 0# 1#
    +-inverse   : Inverse 0# _⁻¹ +
    ⁻¹-cong     : Congruent₁ _⁻¹

  open IsQuasiring isQuasiring public hiding (zeroˡ; zeroʳ; zero)

  +-inverseˡ : LeftInverse 0# _⁻¹ +
  +-inverseˡ = proj₁ +-inverse

  +-inverseʳ : RightInverse 0# _⁻¹ +
  +-inverseʳ = proj₂ +-inverse

  zeroˡ : LeftZero 0# *
  zeroˡ = Consequences.assoc+distribʳ+idʳ+invʳ⇒zeˡ setoid
    +-cong *-cong +-assoc distribʳ +-identityʳ +-inverseʳ

  zeroʳ : RightZero 0# *
  zeroʳ = Consequences.assoc+distribˡ+idʳ+invʳ⇒zeʳ setoid
    +-cong *-cong +-assoc distribˡ +-identityʳ +-inverseʳ

  zero : Zero 0# *
  zero = zeroˡ , zeroʳ
```

CHANGELOG.md
```diff
@@ -33,6 +33,12 @@ Highlights
 Bug-fixes
 ---------
 
+* In `Algebra.Structures` the records `IsRing` and `IsRingWithoutOne` contained an unnecessary field
```
jamesmckinna, line 36: ... "and `IsNearring`"...

jamesmckinna (approved):

jamesmckinna:
Looks good to me now, modulo `IsNearring`.

Akshobhya1234:
Should we open `IsRingWithoutOne` in `Properties.Ring` so we don't duplicate proofs for `-‿distribˡ-* ` and `-‿distribʳ-*` ?

MatthewDaggitt:
> So I think for the sake of completeness, I would add/revise the definition of IsNearring to the following, as I don't see how to reconcile it with IsRingWithoutOne.

I don't see the advantage of re-proving the `zero` properties. With your definition, the user still has to provide them when constructing the `IsQuasiring` record.

I think the right solution is to rejig the whole local inheritance graph to avoid `IsNearring` inheriting directly from `IsQuasiring`. However, that's a much bigger change than I want to make right now. So I vote for leaving `IsNearring` as is for the moment.

MatthewDaggitt:
> Should we open RingWithoutOne in Properties.Ring so we don't duplicate proofs for -‿distribˡ-*  and -‿distribʳ-* ?

Yup done. No CHANGELOG entry needed as it was added in v2.0.

jamesmckinna:
@MatthewDaggitt once again thanks for the insightful diagnosis of the problem(s) with `(Is`)`Nearring`, which in any case does feel like a different problem of its own (myself, I haven't looked in detail at them, but even Wikipedia suggest that the standard literature distinguishes them according to left- and right-multiplication/distributivity, suggesting that we might need to revisit/refactor the definitions we have in a more concerted way another day). 

So I was happy before, and especially now there is the addtoinal export of `...WithoutOne`, so let's merge.

jamesmckinna:
So... will you fix the deprecation failure? (I know it arises from my PR... sigh!)

MatthewDaggitt:
Ah yup of course. I'll fix it! Good to know the system is working. That would have got merged before :cry:

jamesmckinna:
Towards rolling up some `Ring`-related issues arising from this, does this issue/PR now mean that `RingWithoutAnnihiliatingZero` is not even a thing... (I confess I'd never heard of such a structure until recently, because of `stdlib` activity... ;-)). And so should be removed?

MatthewDaggitt:
> Towards rolling up some Ring-related issues arising from this, does this issue/PR now mean that RingWithoutAnnihiliatingZero is not even a thing... (I confess I'd never heard of such a structure until recently, because of stdlib activity... ;-)). And so should be removed? Ah, you have deprecated it, sorry!

There is no `RingWithoutAnnihilatingZero` anyway that I'm aware of! `IsRingWithoutAnnhilatingZero` was one of our "biased" structures that are used to provide record syntax for alternative ways of constructing structures. I deprecated it because it was being used to get around this very problem, i.e. how to construct a ring without having to prove the zero law.

jamesmckinna:
Ah... ;-)

src/Algebra/Structures/Biased.agda
```diff
@@ -192,6 +192,33 @@ open IsCommutativeSemiringʳ public
 ------------------------------------------------------------------------
 -- IsRing
 
+record IsRing* (+ * : Op₂ A) (-_ : Op₁ A) (0# 1# : A) : Set (a ⊔ ℓ) where
+  field
+    +-isAbelianGroup : IsAbelianGroup + 0# -_
+    *-isMonoid       : IsMonoid * 1#
+    distrib          : * DistributesOver +
+    zero             : Zero 0# *
```
jamesmckinna, line 200: Too late! But is this field also now redundant as a consequence of this issue/PR?

PR #2210, opened 2023-11-24
Add biased versions of Function structures

MatthewDaggitt:
Eases the conversion to the new strict versions. Also fixed a naming bug introduced in v2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -751,8 +751,10 @@ Non-backwards compatible changes
    - The records in  `Function.Structures` and `Function.Bundles` export proofs
          of these under the names `strictlySurjective`, `strictlyInverseˡ` and
          `strictlyInverseʳ`,
-   - Conversion functions have been added in both directions to
-         `Function.Consequences(.Propositional/Setoid)`.
+   - Conversion functions for the definitions have been added in both directions 
+		 to `Function.Consequences(.Propositional/Setoid)`.
+   - Conversion functions for structures have been added in 
+		 `Function.Structures.Biased`.
 
 ### New `Function.Strict`
 
--- a/src/Algebra/Structures/Biased.agda
+++ b/src/Algebra/Structures/Biased.agda
@@ -2,7 +2,7 @@
 -- The Agda standard library
 --
 -- Ways to give instances of certain structures where some fields can
--- be given in terms of others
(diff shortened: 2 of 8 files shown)


jamesmckinna:
I _think_ I see the usability/ux gain from this kind of PR, but I would be curious to learn how much/how many times this kind of functionality _actually_ gets used downstream... But let's merge!

CHANGELOG.md
```diff
@@ -751,8 +751,10 @@ Non-backwards compatible changes
    - The records in  `Function.Structures` and `Function.Bundles` export proofs
          of these under the names `strictlySurjective`, `strictlyInverseˡ` and
          `strictlyInverseʳ`,
-   - Conversion functions have been added in both directions to
-         `Function.Consequences(.Propositional/Setoid)`.
+   - Conversion functions for the definitions have been added in both directions 
+		 to `Function.Consequences(.Propositional/Setoid)`.
+   - Conversion functions for structures have been added in 
+		 `Function.Structures.Biased`.
```
jamesmckinna, line 757: So, it's good to see these things documented in the `CHANGELOG` (which by its very nature, is 'permanent' only to the extent that diligent users/developers can face wading back through the GitHub history; otherwise, we should regard such commentary as 'essentially ephemeral', much like this review comment ;-)), but at *some* point, we might want such observations/ux hacks to be added to a `user-guide` to the library, which we don't really have (yet)?

src/Algebra/Structures/Biased.agda
```diff
@@ -2,7 +2,7 @@
 -- The Agda standard library
 --
 -- Ways to give instances of certain structures where some fields can
--- be given in terms of others
+-- be given in terms of others. Re-exported via `Algebra`.
```
jamesmckinna, line 5: Ditto. In some sense...

jamesmckinna (approved):

MatthewDaggitt:
> I think I see the usability/ux gain from this kind of PR, but I would be curious to learn how much/how many times this kind of functionality actually gets used downstream... But let's merge!

Only a single data-point, but I'm adding it after trying to port my own library to release candidate 1 and running into horrible UX problems. The main advantage is that the bundles already have all the required setoid and congruence proofs  necessary to perform the switch from one definition to another. Without this you're forced to provide them all manually.

Issue #2033, opened 2023-07-25
A more declarative auto-cong machinery


gallais:
I have written a little bit of declarative congruence magic:

https://github.com/gallais/potpourri/blob/b0b734fb8e768f280bc36519048b977b994d1a5e/agda/cong/Syntax/Focus.agda

Unlike the content of [`Tactic.Cong`](http://agda.github.io/agda-stdlib/Tactic.Cong.html), it is not trying to guess what the
context should be but instead reads of the goal type where users can mark the
focus of their action by using the `⟪_⟫` focus.

JacquesCarette:
Very nice.

A lot of proofs in agda-categories could be made more digestible for humans with such machinery, as quite a lot of work is spent on dealing with cong and re-associating to put a part "in focus". For efficiency, I'd probably want proof-emitting macros rather than using them directly in library code. But that seems most doable. At the level of, say, monoidal categories, such proof-emitting macros might even be a papers' worth of work.

uncle-betty:
Hello there.

I just opened #2310 without realizing that #2033 existed, which looks more elegant than my hack on top of `cong!`.

What's the plan going forward with this? Happy to withdraw my proposal, if this is still being worked on.

PR #2215, opened 2023-11-26
WIP Add Rational RingSolver

### LINKED ISSUE #1879: Add Data.Rational.Tactic.RingSolver
By analogy to `Data.Nat.Tactic.RingSolver` and `Data.Integer.Tactic.RingSolver`

lemastero:
## Changes
- [x] add Data.Rational.Tactic.RingSolver
- [ ] example or Rational.RingSolver
- [x] add Data.Rational.Unnormalised.Tactic.RingSolver
- [ ] example or Rational.Unnormalised.RingSolver
- [x] update CHANGELOG

Fix #1879

### DIFF
--- a/src/Data/Rational/Tactic/RingSolver.agda
+++ b/src/Data/Rational/Tactic/RingSolver.agda
@@ -1,0 +1,46 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Automatic solvers for equations over rationals
+------------------------------------------------------------------------
+
+-- See README.Integer for examples of how to use this solver
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Rational.Tactic.RingSolver where
+
+open import Agda.Builtin.Reflection using (Term; TC)
+
+open import Agda.Builtin.Int using (Int; negsuc; pos)
+open import Data.Nat.Base using (zero; suc)
+open import Data.Maybe.Base using (Maybe; just; nothing)
+open import Data.Rational.Base using (ℚ; 0ℚ; mkℚ)
+open import Data.Rational.Properties using (+-*-commutativeRing)
+open import Level using (0ℓ)
+open import Data.Unit using (⊤)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl)
+
+import Tactic.RingSolver as Solver using (solve-macro; solve-∀-macro)
+import Tactic.RingSolver.Core.AlmostCommutativeRing as ACR
+
+------------------------------------------------------------------------
+-- A module for automatically solving propositional equivalences
+-- containing _+_ and _*_
+
+ring : ACR.AlmostCommutativeRing 0ℓ 0ℓ
+ring = ACR.fromCommutativeRing +-*-commutativeRing f
+  where
+   f : (x : ℚ) → Maybe (0ℚ ≡ x)
+   f (mkℚ (pos 0)       0       _) = just refl
+   f (mkℚ (pos 0)       (suc _) _) = nothing
+   f (mkℚ (pos (suc _)) _       _) = nothing
+   f (mkℚ (negsuc _)    _       _) = nothing
+
+macro
+  solve-∀ : Term → TC ⊤
+  solve-∀ = Solver.solve-∀-macro (quote ring)
+
+macro
+  solve : Term → Term → TC ⊤
+  solve n = Solver.solve-macro n (quote ring)
--- a/src/Data/Rational/Unnormalised/Tactic/RingSolver.agda
+++ b/src/Data/Rational/Unnormalised/Tactic/RingSolver.agda
@@ -1,0 +1,45 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Automatic solvers for equations over unnormalised rationals
+------------------------------------------------------------------------
+
+-- See README.Integer for examples of how to use this solver
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Rational.Unnormalised.Tactic.RingSolver where
+
+open import Agda.Builtin.Reflection
+
+open import Agda.Builtin.Int using (Int; negsuc; pos)
+open import Data.Nat.Base using (zero; suc)
+open import Data.Maybe.Base using (Maybe; just; nothing)
+open import Data.Rational.Unnormalised.Base using (ℚᵘ; 0ℚᵘ; _≃_; mkℚᵘ; *≡*)
+open import Data.Rational.Unnormalised.Properties using (+-*-commutativeRing)
+open import Level using (0ℓ)
+open import Data.Unit using (⊤)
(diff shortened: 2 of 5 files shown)


MatthewDaggitt:
Thanks, looks great! As you say if you could add some examples in `README.Data.Rational(.Unnormalised)` that would be awesome.

src/Data/Rational/Tactic/RingSolver.agda
```diff
@@ -0,0 +1,46 @@
...
+-- containing _+_ and _*_
+
+ring : ACR.AlmostCommutativeRing 0ℓ 0ℓ
+ring = ACR.fromCommutativeRing +-*-commutativeRing f
+  where
+   f : (x : ℚ) → Maybe (0ℚ ≡ x)
+   f (mkℚ (pos 0)       0       _) = just refl
+   f (mkℚ (pos 0)       (suc _) _) = nothing
+   f (mkℚ (pos (suc _)) _       _) = nothing
+   f (mkℚ (negsuc _)    _       _) = nothing
```
jamesmckinna, line 38: I'd be strongly tempted to lift this out as a definition in `Data.Rational.Base`, or else as a property (cf. `Relation.Binary.WeaklyDecidable`, for which we currently lack analogues at arity 0, 1) in `Data.Rational.Properties`.

Suggested name for such a refactored `f`: `isZero-weakly-decidable`?

That way, the `import`s for your solver modules then become very much simplified, in favour of those that are already currently used by the `Rational.*` modules...

... and there's (potentially) a downstream benefit in being able to reuse the definition of the corresponding `Unnormalised` function in the definition of this one.

src/Data/Rational/Unnormalised/Tactic/RingSolver.agda
```diff
@@ -0,0 +1,45 @@
...
+-- A module for automatically solving propositional equivalences
+-- containing _+_ and _*_
+
+ring : ACR.AlmostCommutativeRing 0ℓ 0ℓ
+ring = ACR.fromCommutativeRing +-*-commutativeRing f
+  where
+   f : (x : ℚᵘ) → Maybe (0ℚᵘ ≃ x)
+   f (mkℚᵘ (pos zero)    _) = just (*≡* refl)
+   f (mkℚᵘ (pos (suc _)) _) = nothing
+   f (mkℚᵘ (negsuc _)    _) = nothing
```
jamesmckinna, line 37: Ditto.!

jamesmckinna:
Re: refactoring suggestion

I think that it's possible that a 'better' approach is available via leveraging the construction in the (old) `Algebra.RingSolver.Simple` module: namely to turn a `Decidable` equality into a `WeaklyDecidable` one, and thereby define a 'smart' constructor for `AlmostCommutativeRing`s. 

Now, equality on `Rational(.Unnormalised)` is already proven `Decidable`, so there might (even) be _no need_ to define a separate `f`/`isZero-weakly-decidable` function at all.

jamesmckinna:
Re: refactoring (meta-comment)

I think there's a tension, especially as a new developer, between contributing a PR with as 'small' footprint as possible (which can have as a consequence a 'wide' collection of `import`s, and (lots of) local definitions), and refactoring it to distribute the various pieces over the library (as I suggested above). But I do think the latter is eventually the way to go, for the sake of leveraging the power already present in the library. 

For the PR at hand: The definitions of the `Solver`s for `Nat` and `Integer` might be instructive in this regard, where the import interfaces/implementations are just about as minimal as you can get...

lemastero:
> if you could add some examples in `README.Data.Rational(.Unnormalised)` that would be awesome.

There is clearly something wrong with RingSolvers in this PR. I defined very simple example [here](https://github.com/lemastero/agda-stdlib/blob/rational-ringsolver/doc/README/Data/Rational.agda#L47-L52):
```
open import Data.Integer using (+_)
open import Data.Rational
open import Data.Rational.Properties

1/4 : ℚ
1/4 = + 1 / 4

3/4 : ℚ
3/4 = + 3 / 4

open import Relation.Binary.PropositionalEquality -- using (_≡_; refl)
open import Data.Rational.Tactic.RingSolver

lemma : ∀ x y → x + y + 1/4 + ½ ≃ 3/4 + y + x
lemma = {! solve-∀  !}
```
but this fails with:
```Agda
Malformed call to solve.Expected target type to be like: ∀ x y → x + y ≈ y + x.Instead: _19
when checking that the expression unquote solve-∀ has type _19
```
I think RingSolver does not like PropositionalEquality (but it does not like [equality define in Rational.Properties](https://github.com/lemastero/agda-stdlib/blob/rational-ringsolver/src/Data/Rational/Base.agda#L69-L70) either). 

I will try to see if I can create example with existing SemiringSolver for begin with and do refactoring proposed, by @jamesmckinna.

gallais:
Given that the goal looks like a meta variable, it may be that the tactic is being
run too early. You may want to block on the meta so that you can wait for it to
be solved and then call the tactic once again.

Cf. the `blockOnMeta` call in [this chunk of tactical](https://github.com/gallais/potpourri/blob/1279c8d1ccc4cc121d65294f4a405cb815d4f4b9/agda/cong/Syntax/Focus.agda#L108-L110) I developed for #2033
and where I was facing similar issues.

lemastero:
I've found some resources on semigroup/ring solvers in Agda:
* [Automatic Ring Solving](https://reasonablypolymorphic.com/blog/ring-solving/index.html)
* The solver for equations in commutative rings in agda/cubical - Felix Cherubini ([video](https://www.youtube.com/watch?v=6wDQlkR-V_M))
* [Automatically and Effiently Illustrating Polynomial Equalities in Agda](https://doisinkidney.com/pdfs/bsc-thesis.pdf) - Donnacha Oisín Kidney
* [Evidence-providing problem solvers in Agda](https://umazalakain.info/static/Z18/bsc-thesis.pdf) - Uma Zalakain

I will try to traverse the subject and fold it into better understanding - what is the problem here.

src/Data/Rational/Tactic/RingSolver.agda
dancewithheart: I have revived this work in #2965 and following this remark - the helper weak zero decidable were
moved into the corresponding Properties.

jamesmckinna:
Any likelihood of you being able to get back to this PR again after 2,5 years @lemastero ?
If not, I think we may want to close this unmerged, and try again another day.

dancewithheart:
The work started here was refreshed in #2965 and merged. (I did applied your suggestions there @jamesmckinna :) Thank you!)

So this draft can be closed. I can't do it myself as I no longer have access to lemastero account (ssh keys expired).

jamesmckinna:
> The work started here was refreshed in #2965 and merged. (I did applied your suggestions there @jamesmckinna :) Thank you!)
> 
> So this draft can be closed. I can't do it myself as I no longer have access to lemastero account (ssh keys expired).

Thanks Piotr! Closing now...

PR #2217, opened 2023-11-28
Fix #2216 by making divisibility definitions records

### LINKED ISSUE #2216: Display of (redundant/deprecated) pattern synonyms
> An annoyance that seems to be due to this: the `emacs` tooling seems to prefer `less-than-or-equal` to `_,_` when I do C-c C-c to pattern match on a product when both are in scope, regardless on whether the thing I'm matching has anything to do with natural numbers

_Originally posted by @Taneb in https://github.com/agda/agda-stdlib/issues/1948#issuecomment-1827536125_

MatthewDaggitt:
Fixes #2216. I've left the record names the same, so if we ever unify with the definition in `Data.Nat.Divisibility.Core` the change is less breaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1048,6 +1048,10 @@ Non-backwards compatible changes
 
 * In `Algebra.Core` the operations `Opₗ` and `Opᵣ` have moved to `Algebra.Module.Core`.
 
+* In `Algebra.Definitions.RawMagma.Divisibility` the definitions for `_∣ˡ_` and `_∣ʳ_`
+  have been changed from being defined as raw products to being defined as records. However,
+  the record constructors are called `_,_` so the changes required are minimal.
+
 * In `Codata.Guarded.Stream` the following functions have been modified to have simpler definitions:
   * `cycle`
   * `interleave⁺`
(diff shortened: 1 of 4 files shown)


jamesmckinna:
Comments on the issue, but at some point this may need revisiting to address #2115 but perhaps that can be discussed/implemented for v3.0...

jamesmckinna:
Litmus test ahead of any review/merge: 
@Taneb does this PR fix the issue for you?

Taneb:
Hmm, no it doesn't, the issue still exists.

Here's what I tried:

```agda
module NatTest where

open import Data.Nat.Base
open import Data.Product.Base

foo : ℕ × ℕ → ℕ
foo x = {!!}
```

Typing `C-c C-c x <enter>` in the goal replaces x with `(less-than-or-equal {fst} snd)`. I'm sure I'm using the correct branch.

jamesmckinna:
Hmmm. I get the correct behaviour:
```agda
module Data.Nat.Test where

open import Data.Nat.Base
open import Data.Product.Base

test : ℕ × ℕ → ℕ
test p = {!p!}
```
`C-c C-c` in the goal produces `test (fst , snd) = {!!}`...

jamesmckinna (approved):
OK, once I had got to the bottom of @Taneb 's issue (I think I understand why, but I'm still scratching my head about how fragile things can seem), I was very happy to see such a minimally-invasive solution. Wish I had understood how to achieve this! I've learnt something.

MatthewDaggitt:
That is a little scary. @Taneb are you okay with this solution or are you still having problems?

Taneb:
Yes, this seems to fix the problem for me.

jamesmckinna:
Phew!

Issue #2063, opened 2023-08-17
Release v2.0


MatthewDaggitt:
Okay, so I realise I've been somewhat absent recently. My current job ends on September 12th and my next one doesn't start until late November, so I plan to get v2.0 out the door during that period of time. I promise I will wade through the back log of PRs that have built up :+1:

jamesmckinna:
Congratulations on the new job! 🎉

jamesmckinna:
As to managing the backlog, as both a contributor to making the pile grow, as well as more recently, helping to whittle it down (at least a little bit), I was interested to read that our colleagues on the main Agda issue tracker are correspondingly [concerned about their own policies to manage their PR backlog](https://github.com/agda/agda/issues/6482). 

It might be worth a separate issue, but rather than generate more clutter, I'm linking it here while our efforts are directed towards the v2.0 backlog... once that's out, suggest we (re)start the conversation about how to manage the way forward.

MatthewDaggitt:
Okay, with the merge of https://github.com/agda/agda-stdlib/pull/1895 I'm going to get serious about making a release. 

As of now, all new PRs will get a v2.1 milestone unless they explicitly address an issue that needs to be fixed before v2.0. 

Existing open PRs will be triaged depending on how easy they are to merge.

When would interested people be available for a meeting next week to discuss outstanding PRs? If you are, please state your availability in GMT below:

Matthew
- Monday 2nd - 7am - 4pm 
- Tuesday 3rd - 7am - 11am, 2pm-4pm
- Wednesday 4th - 7am - 4pm
- Thursday 5th - 7am - 4pm
- Friday 6th - 7am - 11am

Taneb:
My availabilities next week:

* Monday 6-7am, 7:30-9am, 10:30am-late
* Tuesday 6-7am, 8am-12pm, 2pm-late
* Wednesday 6-7am, 7:30am-late
* Thursday 6-7am, 7:30am-12pm, 2pm-late
* Friday 6-7am, 7:30am-12pm, 1:30pm-late

jamesmckinna:
This is great news! 

As for a meeting next week, I'm a bit all over the place at the moment, as evidenced by the manic PR/issue/commentary explosion. Will try to identify times to meet soon! 

[...]

PR #2224, opened 2023-12-05
Fix deprecated modules

alexarice:
I realise these modules are deprecated but it probably best that they still build. After these changes I can build an `Everything.agda` built with `runhaskell GenerateEverything.hs --include-deprecated` .

Not sure if I fixed them in the "correct" way.

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -153,8 +153,8 @@ jobs:
         run: |
           # Including deprecated modules purely for testing
           cabal run GenerateEverything -- --include-deprecated
-          ${{ env.AGDA }} --safe EverythingSafe.agda
-          ${{ env.AGDA }} Everything.agda
+          ${{ env.AGDA }} -WnoUserWarning --safe EverythingSafe.agda
+          ${{ env.AGDA }} -WnoUserWarning Everything.agda
 
       - name: Prepare HTML index
         run: |
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -151,10 +151,19 @@ jobs:
 
       - name: Test stdlib
         run: |
+          # Including deprecated modules purely for testing
+          cabal run GenerateEverything -- --include-deprecated
+          ${{ env.AGDA }} --safe EverythingSafe.agda
+          ${{ env.AGDA }} Everything.agda
+
+      - name: Prepare HTML index
+        run: |
+          # Regenerating the Everything files without the deprecated modules
           cabal run GenerateEverything
           cp .github/tooling/* .
           ./index.sh
           ${{ env.AGDA }} --safe EverythingSafe.agda
+          ${{ env.AGDA }} Everything.agda
           ${{ env.AGDA }} index.agda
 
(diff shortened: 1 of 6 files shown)


alexarice:
Seems ci fails on seeing the warning for a deprecated module

gallais:
Seems like there is no way around checking the lib twice: once with the deprecated modules
(and with `-WnoUserWarning`) and once without. Hope this won't inflate the build times too
much. It may be worth parallelising the workflow if that is the case.

jamesmckinna:
I'm not clear about the "seems better if they build" rationale. Part of the point of deprecation, as I understand it,  is precisely to let certain modules/definitions degrade, so that:
* we encourage/oblige/force users to migrate;
* we free ourselves from having to support previous editions of code.

This PR seems to increase the maintenance burden in respect of the second point, and as a developer/maintainer, my first thought: that is extra work that none of us need? 

Now it seems as though it would become not enough to deprecate a module, but that we also are obliged to ensure it continue to typecheck? So that who can (continue to) use it? (Against the first point)

So while I *might* be grateful to someone (else!) for doing such work, I think it sets up... unreasonable expectations going forward...? Or am I missing something?

jamesmckinna:
Maybe to rephrase this in slightly more sober terms: what issue is this addressing/solving? 

We have indeed spent a *lot* of time retrofitting various things against the forthcoming breaking v2.0, and/but I'm not sure I am/any of us are in a hurry to make that part of our SOP?

alexarice:
The concrete issue building the module solves is that it'll actually display the deprecation message instead of some other cryptic message. If deprecated code doesn't even typecheck and we don't want to maintain it then surely it should just be deleted?

jamesmckinna:
Yes... ok. I raised the issue of deprecation policy on the v2.0 release issue #2063 so I think a shout-out to @MatthewDaggitt is in order...

MatthewDaggitt:
Yes, I agree that we want deprecated modules to still check correctly. As @alexrice says, not much point if we have the deprecation messages, but they don't display because of some other error. This needs to get in before v2.0.

jamesmckinna:
As frequently, I stand corrected! Apologies for the intemperate intervention...

MatthewDaggitt:
> As frequently, I stand corrected! Apologies for the intemperate intervention...

Absolutely no worries! It's always good to have long term maintainer's assumptions sanity checked by newer members!

jamesmckinna:
Maybe I need to back up then, and wonder/check  how it is that `make test` had not exposed these errors for me on recent runs...

jamesmckinna:
And then the new failing version of `make test` sadly reveals the painful hidden cost of earlier deprecations... the revenge of #1735 #1946 etc. sigh

gallais:
> ... or is it a bad interaction between the --include-deprecated flag to GenerateEverything and the subsequent division into Everything and EverythingSafe?

Yes. The `--include-deprecated` was added later as an additional `Status` but
different from `Safe`/`Unsafe`. However that's actually an orthogonal dimension.

I've pushed a fix :crossed_fingers:

gallais (approved):

JacquesCarette (approved):

PR #2226, opened 2023-12-12
Fix typo in raise deprecation message

alexarice:


jamesmckinna:
Sigh... thanks for the eagle-eyed spot!

alexarice:
Only spotted it because I blindly followed the advice in the deprecation message and it didn't work

jamesmckinna (approved):
Two things (for maintainers) to note for the future:
* Agda didn't catch this when committed back in 2021 (!) (my first PR, or near enough, so I'm still blushing)
* We need a more robust way of dealing with deprecations? cf. #2224

jamesmckinna:
Thanks again for this and your earlier deprecation bug. Please keep them coming! We've opened a milestone for v2.0-release bug fixes... ;-)

PR #2230, opened 2023-12-14
finer-grained analysis of `NonZero` arguments

jamesmckinna:
[Breaking changes so held back from merge of #2182 into v2.1]

Essentially a placeholder `DRAFT` PR in which to accumulate all such things, towards an eventual v3.0 merge.

NB: currently, the statements/proofs follow a rather clunky
* `let instance _ =...` in the statements;
* `where instance _ = ...` in the proofs.

Suggest subsequent refactorings to make all such things anonymous modules which set up all the appropriate scope... esp. in the cases where there are auxiliary lemmas. 

UPDATED: @MatthewDaggitt 's review comment gives me pause... So I'll refrain from committing further for the time being!

### DIFF
--- a/src/Data/Nat/DivMod.agda
+++ b/src/Data/Nat/DivMod.agda
@@ -203,12 +203,13 @@ m/n≤m m n = *-cancelʳ-≤ (m / n) m n (begin
   m           ≤⟨ m≤m*n m n ⟩
   m * n       ∎)
 
-m/n<m : ∀ m n .{{_ : NonZero m}} .{{_ : NonZero n}} →
-        1 < n → m / n < m
-m/n<m m n 1<n = *-cancelʳ-< _ (m / n) m $ begin-strict
+m/n<m : ∀ m n .{{_ : NonZero m}} .{{_ : NonTrivial n}} →
+        let instance _ = nonTrivial⇒nonZero n in m / n < m
+m/n<m m n     = *-cancelʳ-< _ (m / n) m $ begin-strict
   m / n * n ≤⟨ m/n*n≤m m n ⟩
-  m         <⟨ m<m*n m n 1<n ⟩
+  m         <⟨ m<m*n m n (nonTrivial⇒n>1 n) ⟩
   m * n     ∎
+  where instance _ = nonTrivial⇒nonZero n
 
 /-mono-≤ : .{{_ : NonZero o}} .{{_ : NonZero p}} →
            m ≤ n → o ≥ p → m / o ≤ n / p
@@ -245,7 +246,7 @@ m/n≡0⇒m<n {m} {n@(suc _)} m/n≡0  with <-≤-connex m n
   where instance _ =  >-nonZero (m≥n⇒m/n>0 n≤m)
 
 m/n≢0⇒n≤m : ∀ {m n} .{{_ : NonZero n}} → m / n ≢ 0 → n ≤ m
-m/n≢0⇒n≤m {m} {n@(suc _)} m/n≢0 with <-≤-connex m n
+m/n≢0⇒n≤m {m} {n} m/n≢0 with <-≤-connex m n
 ... | inj₁ m<n = contradiction (m<n⇒m/n≡0 m<n) m/n≢0
 ... | inj₂ n≤m = n≤m
 
@@ -307,7 +308,8 @@ m∣n⇒o%n%m≡o%m m n@.(p * m) o (divides-refl p) = begin-equality
     o / pm * pm          ≤⟨ m/n*n≤m o pm ⟩
     o                    ∎
 
-m*n/m*o≡n/o : ∀ m n o .{{_ : NonZero o}} .{{_ : NonZero (m * o)}} →
+m*n/m*o≡n/o : ∀ m n o .{{_ : NonZero m}} .{{_ : NonZero o}} →
+              let instance _ = m*n≢0 m o in
               (m * n) / (m * o) ≡ n / o
 m*n/m*o≡n/o m n o = helper (<-wellFounded n)
   where
@@ -315,7 +317,6 @@ m*n/m*o≡n/o m n o = helper (<-wellFounded n)
   helper : ∀ {n} → Acc _<_ n → (m * n) / (m * o) ≡ n / o
   helper {n} (acc rec) with <-≤-connex n o
   ... | inj₁ n<o = trans (m<n⇒m/n≡0 (*-monoʳ-< m n<o)) (sym (m<n⇒m/n≡0 n<o))
-    where instance _ = m*n≢0⇒m≢0 m
   ... | inj₂ n≥o = begin-equality
     (m * n) / (m * o)             ≡⟨ m/n≡1+[m∸n]/n (*-monoʳ-≤ m n≥o) ⟩
     1 + (m * n ∸ m * o) / (m * o) ≡⟨ cong (suc ∘ (_/ (m * o))) (*-distribˡ-∸ m n o) ⟨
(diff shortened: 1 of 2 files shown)


src/Data/Nat/DivMod.agda
```diff
@@ -307,15 +308,15 @@ m∣n⇒o%n%m≡o%m m n@.(p * m) o (divides-refl p) = begin-equality
     o / pm * pm          ≤⟨ m/n*n≤m o pm ⟩
     o                    ∎
 
-m*n/m*o≡n/o : ∀ m n o .{{_ : NonZero o}} .{{_ : NonZero (m * o)}} →
+m*n/m*o≡n/o : ∀ m n o .{{_ : NonZero m}} .{{_ : NonZero o}} →
+              let instance _ = m*n≢0 m o in
```
MatthewDaggitt, line 312: So I am very very wary of this change, and similar ones below. Here we are enforcing exactly the form that the proof of `NonZero (m * o)` must be, i.e. it must `m*n≢0 m o`, whereas before we could accept any proof of it. Now it may be the case that because we're only using it in irrelevant places, it doesn't matter, but we should think carefully about it. What is the _advantage_ of this change?

jamesmckinna: Ah... interesting thought!

I had (mistakenly, perhaps) imagined that since we were precisely dealing with _irrelevant_ instances, that it would be OK. 

As to the advantage: clients of these proofs do/would not have to re-manufacture these irrelevant instances at call-sites. My _assumption_ being that the simpler, 'atomic' facts about `NonZero`ness of the arguments would be more likely to be in scope than the 'composite' ones being dealt with here?  Again, possibly a mistaken assumption!

jamesmckinna: Hmmm... it doesn't invalidate your criticism/note of caution, but consider that we now have in `Data.Nat.Properties` all of the (pieces of the) proof that `NonZero (m * n)` iff `NonZero m` and `NonZero n`. So I suppose, following your line of thought, there might be pathological cases where we know `NonZero (m * n)` and `NonZero m` but not (yet) that `NonZero n`, except by a subsidiary appeal to the relevant 'projection' lemma... but I can't quite imagine such examples?

And since, AFAICT, *all* instances of `NonZero n` eventual reduce to the *ur*-instance `nonZero` (and hence to the discrimination of an occurrence of `suc` in the whnf of `n`)... I suppose that the typechecker would always (be able to) find such instance of `NonZero n`, 'just' by wh-normalisation...? Maybe that might prove more expensive than instance search? Can't quite see how, but that's your over-arching point: It's hard to predict how haywire instance search can get!

src/Data/Nat/DivMod.agda
```diff
@@ -239,13 +240,13 @@ m≥n⇒m/n>0 {m@(suc _)} {n@(suc _)} m≥n = begin
   m / n ∎
 
 m/n≡0⇒m<n : ∀ {m n} .{{_ : NonZero n}} → m / n ≡ 0 → m < n
-m/n≡0⇒m<n {m} {n@(suc _)} m/n≡0  with <-≤-connex m n
+m/n≡0⇒m<n {m} {n} m/n≡0  with <-≤-connex m n
 ... | inj₁ m<n = m<n
 ... | inj₂ n≤m = contradiction m/n≡0 (≢-nonZero⁻¹ _)
   where instance _ =  >-nonZero (m≥n⇒m/n>0 n≤m)
 
 m/n≢0⇒n≤m : ∀ {m n} .{{_ : NonZero n}} → m / n ≢ 0 → n ≤ m
```
JacquesCarette, line 248: I seem to be missing something here. 1 / 2 is not zero, so 2 ≤ 1 ???

src/Data/Nat/DivMod.agda
JacquesCarette: I have misgivings about this as well. The whole point of instance arguments is that we let Agda figure them out (globally). Giving them locally is silly: if we're hand-constructing proofs, we may as well pass them in as arguments!

src/Data/Nat/DivMod.agda
```diff
@@ -324,17 +325,17 @@ m*n/m*o≡n/o m n o = helper (<-wellFounded n)
     n / o                         ∎
     where n∸o<n = ∸-monoʳ-< (n≢0⇒n>0 (≢-nonZero⁻¹ o)) n≥o
 
-m*n/o*n≡m/o : ∀ m n o .{{_ : NonZero o}} .{{_ : NonZero (o * n)}} →
+m*n/o*n≡m/o : ∀ m n o .{{_ : NonZero n}} .{{_ : NonZero o}} →
+              let instance _ = m*n≢0 o n in
               m * n / (o * n) ≡ m / o
 m*n/o*n≡m/o m n o = begin-equality
-  m * n / (o * n) ≡⟨ /-congˡ (*-comm m n) ⟩
+  m * n / (o * n) ≡⟨ /-congˡ {{o*n≢0}} (*-comm m n) ⟩
```
JacquesCarette, line 332: whereas here you indeed do what I 'expected', i.e. pass it in (though I don't know why you bother naming it)

src/Data/Nat/DivMod.agda
Taneb: This is division of naturals, rounded towards zero. By this, `1 / 2` is in fact 0.

src/Data/Nat/DivMod.agda
jamesmckinna: Well... I don't know about "silly", but thanks for the feedback ;-)
As you might imagine, I take a different view: fully proof search is inherently unpredictable, and potentially very ill-behaved, so providing 'hint's to the inference, in the interests of narrowing the search space, seems a *perfectly* reasonable thing to do. Or at least, so I thought, until I read your comments...

src/Data/Nat/DivMod.agda
jamesmckinna: Well, this one is *unbelievably* annoying. The old proof had no problem directly using the in-scope, but irrelevant/unmentioned, instance of `{{NonZero (o * n)}}` to ensure that both the term was well-formed, and that `/-congˡ (*-comm m n)` was well-typed. But the change I made has the typechecker unable to infer the sub-term `(o * n)`, leaving two choices: supply it as a named instantiation of the implicit parameter to that proof, or else (re-)supply the instance as a proxy for that. And this despite the instance required being irrelevant! So maybe it's an Agda bug, maybe it points to the whole venture being misguided, but it's *not* what (I believe) you expected.

jamesmckinna: As for "hand-constructing proofs": no, I disagree with that analysis, too. What I'm doing is choosing an inference rule to preferentially apply, towards Agda constructing the eventual complete proof, based on the search that it will *still* have to do to find suitable instances of the 'atomic' `{{NonZero n}}` and `{{NonZero o}}` hypotheses...

jamesmckinna:
Given the negative tone of the comments, and finding myself getting combative/defensive in response to them, maybe the best thing is to close this as `status:abandoned`. Also for the second night in a row, I have had less than 3hrs' sleep, so my resources are severely stretched for thinking about this, much less trying to convince the two of you. Oh well.

src/Data/Nat/DivMod.agda
JacquesCarette: Thanks, I was mis-understanding that `/` was.

src/Data/Nat/DivMod.agda
JacquesCarette: Indeed, this may well be an Agda weakness / bug.

JacquesCarette:
"Convincing" me of finicky stuff with instance arguments is perhaps too high a bar. I really don't like instance arguments (but reluctantly admit that for `NonZero` they are generally a win). Since it has been agreed that `NonZero` is ok in the library, the only 'argument' left is on the level of acceptable gymnastics that is reasonable locally. And not forcing a proof to have a particular shape.

src/Data/Nat/DivMod.agda
JacquesCarette: As I say below: take my comments about instances with a huge grain of salt.

Yes, giving Agda hints is quite sensible. I wish it was syntactically clearer that this is what was going on. (Giving an instance globally is not a hint, it is a statement of fact that gets loaded into a database of facts; giving it locally is indeed a hint, which perhaps should have a different syntax)

PR #2233, opened 2023-12-15
[fixes #2232]

### LINKED ISSUE #2232: Post-release version bump in `standard-library.agda-lib`?
Currently: still showing as 2.0 on `master`
Should this get bumped to the 'next' milestone after a release? If so, this should also be part of the `release-notes`...

jamesmckinna:


MatthewDaggitt:
Thanks! Sorry to be pedantic about minutae, but I think we should be bumping it to 2.1 rather than 2.0.1. The latter is purely a bug fix release, that will branch directly off of the v2.0.0 commit. The `master` branch is for 2.1, as we already have commits in `master` that shouldn't go in 2.0.1.

jamesmckinna:
Thanks for the correction! I thought it might be the case, but hadn't been sure ... (of it, or of myself...)

MatthewDaggitt (approved):

jamesmckinna:
Which milestone does this issue belong to, though?

MatthewDaggitt:
2.1 as this commit won't appear in the history of v2.0.1

PR #2239, opened 2023-12-27
Bring back a convenient short-cut for infix Func

JacquesCarette:
If this is accepted, I can augment this PR with some uses of this inside stdlib in appropriate places as well.

### DIFF
--- a/src/Function/Bundles.agda
+++ b/src/Function/Bundles.agda
@@ -401,6 +401,17 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
     LeftInverse splitSurjection
 
 ------------------------------------------------------------------------
+-- Infix abbreviations for oft-used items
+------------------------------------------------------------------------
+
+-- Same naming convention as used for propositional equality below, with
+-- appended ₛ (for 'S'etoid).
+
+infixr 0 _⟶ₛ_
+_⟶ₛ_ : Setoid a ℓ₁ → Setoid b ℓ₂ → Set _
+_⟶ₛ_ = Func
+
+------------------------------------------------------------------------
 -- Bundles specialised for propositional equality
 ------------------------------------------------------------------------
 
@@ -525,16 +536,6 @@ module _ {A : Set a} {B : Set b} where
     ( strictlyInverseˡ⇒inverseˡ to invˡ
     , strictlyInverseʳ⇒inverseʳ to invʳ
     )
-
-------------------------------------------------------------------------
--- Inline bbreviations for oft-used items
---  (but one define the ones that recur)
-------------------------------------------------------------------------
-
--- Same convention as above, with appended ₛ (for 'S'etoid)
-infixr 0 _⟶ₛ_
-_⟶ₛ_ : Setoid a ℓ₁ → Setoid b ℓ₂ → Set _
-_⟶ₛ_ = Func
 
 ------------------------------------------------------------------------
 -- Other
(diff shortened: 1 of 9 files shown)


src/Function/Bundles.agda
```diff
@@ -400,6 +400,15 @@ module _ (From : Setoid a ℓ₁) (To : Setoid b ℓ₂) where
...
     LeftInverse splitSurjection
 
+------------------------------------------------------------------------
+-- Inline bbreviations for oft-used items
+------------------------------------------------------------------------
+
+-- Since ⟶ is taken below, use a thick arrow instead
+infixr 0 _➙_
+_➙_ : Setoid a ℓ₁ → Setoid b ℓ₂ → Set _
+_➙_ = Func
```
MatthewDaggitt, line 410: So it would be kind of nice if whatever notation we chose for this generalised to the other bundles in this file? It would also be nice if it harmonised with the notation directly below. Would you consider `⟶_s` (subscript)? That would would harmonise with the subscript `s` used to indicate setoid in e.g. https://github.com/agda/agda-stdlib/blob/660d9834670cd8a9afa9d88f3fdf9ab05ee58195/src/Data/Product/Relation/Binary/Pointwise/NonDependent.agda#L183

JacquesCarette: Yes, I would be fine with that.

JacquesCarette: Done.

src/Function/Bundles.agda
```diff
@@ -526,6 +526,16 @@ module _ {A : Set a} {B : Set b} where
     , strictlyInverseʳ⇒inverseʳ to invʳ
     )
 
+------------------------------------------------------------------------
+-- Inline bbreviations for oft-used items
+--  (but one define the ones that recur)
```
MatthewDaggitt, line 531: Typo, for `abreviations`, and the comment in brackets doesn't make sense to me. Also double space at the beginning. Should also read `Infix` not `Inline`.

src/Function/Bundles.agda
```diff
@@ -526,6 +526,16 @@ module _ {A : Set a} {B : Set b} where
...
+------------------------------------------------------------------------
+
+-- Same convention as above, with appended ₛ (for 'S'etoid)
```
MatthewDaggitt, line 534: Very minor but it would make more sense to me for those to live the above the propositional definitions, as they're more general.

MatthewDaggitt (commented):
To be honest I'm not that keen on the changes of some of the existing uses of `Func` to the new name. My reasons are:
- it's inconsistently applied as not all uses of `Func` are changed, even in the files that have been edited.
- it's inconsistent with the names of the other bundles such as `Injection`, `Surjection` in the other places in those files.
 
If we were to introduce that notation consistently everywhere for all types of function bundles I might be convinced, but I think that would be another issue that would need to be discussed separately.

JacquesCarette:
I will remove the inconsistent changes. I knew they were -- I changed the ones that seemed to improve things but not others. I'm fine with prioritizing consistency over local aesthetics.

MatthewDaggitt:
Thanks! Okay, all that's needed in a CHANGELOG entry and I'll merge it in.

JacquesCarette:
Changelog was already done.

MatthewDaggitt:
Ah sorry, I'd folded it up already and didn't spot it!

MatthewDaggitt (approved):

jamesmckinna:
`CHANGELOG` may already have been done, but does not seem to have been updated to the new symbol...?

JacquesCarette:
Oops! Will fix.

PR #2240, opened 2023-12-29
Function setoid is back.

### LINKED ISSUE #1467: Why does equality of Function Setoids have cong baked in?
Right now, we define function setoids as follows (taken from [Function.Equality](https://github.com/agda/agda-stdlib/blob/master/src/Function/Equality.agda#L79)):
```agda
setoid : ∀ {f₁ f₂ t₁ t₂}
         (From : Setoid f₁ f₂) →
         IndexedSetoid (Setoid.Carrier From) t₁ t₂ →
         Setoid _ _
setoid From To = record
  { Carrier       = Π From To
  ; _≈_           = λ f g → ∀ {x y} → x ≈₁ y → f ⟨$⟩ x ≈₂ g ⟨$⟩ y
  ; isEquivalence = record
    { refl  = λ {f} → cong f
    ; sym   = λ f∼g x∼y → To.sym (f∼g (From.sym x∼y))
    ; trans = λ f∼g g∼h x∼y → To.trans (f∼g From.refl) (g∼h x∼y)
    }
  }
  where
  open module From = Setoid From using () renaming (_≈_ to _≈₁_)
  open module To = IndexedSetoid To   using () renaming (_≈_ to _≈₂_)
```

If we look at the definition of `_≈_`, we can see that it has `cong` baked into it. This makes working with these equalities awkward, as you essentially perform a proof that `f ⟨$⟩ x ≈₂ g ⟨$⟩ x`, then perform a `cong` at the very end to get that `g ⟨$⟩ x ≈₂ g ⟨$⟩ y`. It seems like the following equivalent definition would be a bit easier to work with:
```agda
setoid : ∀ {f₁ f₂ t₁ t₂}
         (From : Setoid f₁ f₂) →
         IndexedSetoid (Setoid.Carrier From) t₁ t₂ →
         Setoid _ _
setoid From To = record
  { Carrier       = Π From To
  ; _≈_           = λ f g → ∀ x → f ⟨$⟩ x ≈₂ g ⟨$⟩ x
  ; isEquivalence = record
    { refl  = λ {f} _ → To.refl
    ; sym   = λ f∼g x → To.sym (f∼g x)
    ; trans = λ f∼g g∼h x → To.trans (f∼g x) (g∼h x)
    }
  }
  where
  open module From = Setoid From using () renaming (_≈_ to _≈₁_)
  open module To = IndexedSetoid To   using () renaming (_≈_ to _≈₂_)
```
 
 Thoughts?

### LINKED ISSUE #2228: Equality of functions in the new function hierarchy
As a counterpart to the now-deprecated `Function.Equality.setoid`.

Briefly discussed previously in https://github.com/agda/agda-stdlib/issues/1467#issuecomment-816737487

JacquesCarette:
Fixes #2228 . Fixes #1467 .


src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,35 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Function Equality setoid
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Function.Relation.Binary.Equality where
```
MatthewDaggitt, line 9: So it would be great if we could structure this better than the old hierarchy. In particular this should be `Equality.Setoid` so we have room for the propositional version as well.

src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,35 @@
...
+
+  open import Function.Bundles using (Func; _⟨$⟩_)
+  open import Level using (Level)
+  open import Relation.Binary.Bundles using (Setoid)
+
+  private
+    variable
+      a₁ a₂ b₁ b₂ c₁ c₂ : Level
+      
+  setoid : Setoid a₁ a₂ → Setoid b₁ b₂ → Setoid _ _
```
MatthewDaggitt, line 19: Can we parameterise the whole module by the two setoids?

src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,35 @@
...
+  setoid From To = record
+    { Carrier = Func From To
+    ; _≈_ = λ f g → ∀ {x} → f ⟨$⟩ x To.≈ g ⟨$⟩ x
```
MatthewDaggitt, line 22: Given that we're parameterising the whole module, we should then be able to define a nice infix name for the raw equality relation.

src/Function/Relation/Binary/Equality.agda
```diff
@@ -0,0 +1,35 @@
...
+    ; isEquivalence = record
+      { refl = To.refl
```
MatthewDaggitt, line 24: And the proofs of reflexivity, symmetry, transitivity can be lifted out as their own lemmas.

MatthewDaggitt (changes requested):

src/Function/Relation/Binary/Equality.agda
JacquesCarette: LOL, I named this according to your own suggestion in one of the issues! But sure, will do.

src/Function/Relation/Binary/Equality.agda
JacquesCarette: All done - though it's not quite as nice since I have to eta-expand.

src/Function/Relation/Binary/Setoid/Equality.agda
```diff
@@ -0,0 +1,50 @@
...
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Level using (Level; _⊔_)
+open import Relation.Binary.Bundles using (Setoid)
+
+module Function.Relation.Binary.Setoid.Equality {a₁ a₂ b₁ b₂ : Level}
+  (From : Setoid a₁ a₂) (To : Setoid b₁ b₂) where
+
+  open import Function.Bundles using (Func; _⟨$⟩_)
```
MatthewDaggitt, line 15: We don't usually indent after top level modules?

src/Function/Relation/Binary/Setoid/Equality.agda
```diff
@@ -0,0 +1,50 @@
...
+  open import Relation.Binary.Definitions
+    using (Reflexive; Symmetric; Transitive)
+    
+  private
+    module To = Setoid To
+    module From = Setoid From
+
+  infix 4 _≈_
+  _≈_ : (f g : Func From To) → Set (a₁ ⊔ b₂)
+  f ≈ g = {x : From.Carrier} → f ⟨$⟩ x To.≈ g ⟨$⟩ x
```
MatthewDaggitt, line 25: We could get away with removing thistype annotation if we used a forall?

MatthewDaggitt (commented):
Other than that looks good! Apologies for being inconsistent with my past self!

src/Function/Relation/Binary/Setoid/Equality.agda
JacquesCarette: Indeed - that's one small difference between stdlib and categories. Can fix.

src/Function/Relation/Binary/Setoid/Equality.agda
JacquesCarette: We could, but I rather think that this one serves as useful documentation.

MatthewDaggitt: Okay sure.

jamesmckinna: Any reason to favour the 'diagonal' version of the relation, rather than the PER-like
```agda
f ≈ g = {x y : From.Carrier} → x From.≈ y → f ⟨$⟩ x To.≈ g ⟨$⟩ y
```
(which, incidentally/moreover, would also permit use of a forall...)

JacquesCarette: I forget the exact places where the discussion was had, but this is an explicit choice. The seeming extra generality doesn't buy you anything as these two are inter-derivable and it seems that in practice, it is this version which is more often needed.

Certainly we went back-and-forth on this in `agda-categories` and the version here definitely makes things simpler, at no loss of generality.

MatthewDaggitt (commented):
Okay once the indentation is fixed, happy to merge this in.

JacquesCarette:
Ok @MatthewDaggitt this should now be ready to go.

src/Function/Relation/Binary/Setoid/Equality.agda
jamesmckinna: Yes, I now see the discussion on #1467 and while I may *personally* disagree, ... whatever!?

Taneb (approved):

jamesmckinna:
I'd be grateful for the equivalence of the `cong`/PER-style definition and the proposed one, but happy to follow this PR up with another one, perhaps as part of tackling #2299 so I've no wish to hold this one up. But the `CHANGELOG` merge conflict is managing that on its own... happy to fix that if you don't have time @JacquesCarette ...

JacquesCarette:
`CHANGELOG` conflict fixed and all checks have passed.

JacquesCarette:
@MatthewDaggitt if you could please review? Your 'request changes' is holding this PR back from being merged.

jamesmckinna:
Suggest dismissing the earlier review and approving the revised version instead...
@MatthewDaggitt would you be happy with that?

jamesmckinna:
I don't think this PR, given its size and uncontroversial nature (I'm happy with the design rationale), needs either a second reviewer, nor to wait on unlocking the old stale review changes, so adding to the merge queue now in the interests of unblocking the `CHANGELOG` hell it has already been subject to.

Issue #2091, opened 2023-09-14
Inconsistencies between `Relation.Binary.Definitions` and `Relation.Unary` and `Relation.Nullary`


jamesmckinna:
Maybe it's worth considering reconciling all the relational hierarchy at each arity wrt definitions of: 
- [x] `Irrelevant`
- [x] `Recomputable`
- [x] `Stable`
- [x] `WeaklyDecidable`
- [x] `Decidable`
- [ ] `Satisfiable`
- [ ] `Empty`
- [ ] `Universal`
- [ ] etc. (the list may grow...)

because otherwise, we have a lot of redundant repetition on the one hand, but with 'gaps' on the other. There's some overlap/degeneracy at arity 0 for eg `Satisfiable`/`Universal` (the identity) and `Empty` (negation), but I think that is harmless...

Maybe there are good reasons to do with dependencies (eg. `Relation.Nullary.Decidable.Core.recompute` can't be typed with a `Nullary` definition of `Recomputable`... unless we bite the bullet and lift it out into a `Relation.Nullary.Definitions` etc. But maybe that's not such a bad idea, either? See also #2243 for my current approach to this particular predicate...

UPDATED: partial fix in #2259 (and largely orthogonal to #2243 )

jamesmckinna:
Reconsidering whether to keep this open, I think on balance after the merge of #2259 that this one can be considered closed by that PR... the other cases being a little too non-uniform to be worth pursuing?

PR #2243, opened 2024-01-02
Adds `Relation.Nullary.Recomputable` plus consequences

### LINKED ISSUE #2199: `⊥` is `Recomputable`
The current definition of `Data.Empty`, and its derivative sub-module `Data.Empty.Irrelevant` are the basis of the current (re-)definition of `Relation.Nullary.Negation` and ... everything else involving `⊥` and negation `¬ A` etc. 

What's at present missing, and responsible for the artificial distinction between 
* `Data.Empty.Irrelevant.⊥-elim : ∀ {w} {Whatever : Set w} → .⊥ → Whatever`
* `Data.Empty.⊥-elim : ∀ {w} {Whatever : Set w} → ⊥ → Whatever`
 
is the equivalence of `.⊥` and `⊥`... ie that `⊥` is `Recomputable`.

NB `Recomputable` seems to be the modern-dress version of the very old idea that certain types (negated formulae, among others) have no computational content, that is, we can't make essential use (eg by pattern-matching) of their inhabitants in order to define/prove... other things. 

Well, the use of pattern-matching against the empty pattern `()` ensures that `⊥` itself is `Recomputable`:
```agda
Recomputable : ∀ {a} (A : Set a) → Set a
Recomputable A = .A → A

-- ⊥ is Recomputable

⊥-recompute : Recomputable ⊥
⊥-recompute ()
```
... and this is already provable in the current formulation of the library... and hence that the above two elimination principles are interderivable (so: do we need both, or does one suffice? UPDATED: does the asymmetry in the use of irrelevant marks in types between assumptions and conclusions mean that this cannot precisely be the case?). See [branch](https://github.com/jamesmckinna/agda-stdlib/tree/recomputable-empty), with no (seemingly!?) wider knock-on consequences for the rest of the library. 

So: we should reconcile these things, and their many consequences; esp. wrt proving `Recomputable` for all kinds of types, notably product types (below), negated types, decidable types, ... in general at least (conjecture!?) the (hereditarily) *Harrop formulas*, and more generally, any 'Glivenko class' of formulas (ditto.). 

Question/Issue: Why haven't we done this before/already? cf. #645 / #652 / #762

jamesmckinna:
Fixes #2199 , and the `Recomputable` and `Irrelevant` instances of #2091 .
Cherry-picks some of the lower-hanging fruit from #2149 . 

Outstanding issues:
- [x] names of things: `recompute` vs. `recomputable` vs. `...⇒recomputable` in all the various places; also the various additions to `Relation.Nullary.Negation.Core`; UPDATED sticking with `recompute` for now
- [x] deprecations: module `Data.Empty.Irrelevant` could now be deprecated entirely, but I have not (yet) done so; similarly, there may be knock-on deprecations arising from name choices above; UPDATED no deprecations!
- [x] where things should live/be (re-)exported from. UPDATED see @JacquesCarette 's comments and my revised commits

(Definite~~Probab~~ly) Won't do in relation to issues raised as a result of #2199 :
* "induction on the negative fragment": a separate (relativised!) `Data.Universe` construction for the 'Harrop' universe, as `Relation.Nullary.Recomputable.Universe`, with a view to generalising Harrop's results to ones of the form "if everything in `𝕌 = (U, El)` is `Recomputable`/`Irrelevant`/`Decidable`... then so too is everything in `Harrop 𝕌` etc. (UPDATED: downstream PR, but only once I've figured it out)
* relationship with the 'Irrelevance Axiom' or @nad 's developments mentioned there. (Ditto.)


src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -69,9 +68,8 @@ module _ {A : Set a} where
 
 -- Given an irrelevant proof of a decidable type, a proof can
 -- be recomputed and subsequently used in relevant contexts.
-recompute : Dec A → .A → A
-recompute (yes a) _ = a
-recompute (no ¬a) a = ⊥-elim (¬a a)
+recompute : Dec A → Recomputable A
+recompute = Reflects.recompute ∘ proof
```
jamesmckinna, line 72: Should this be renamed to `dec⇒recomputable` on the model of `Relation.Unary.Consequences`?

src/Relation/Nullary/Reflects.agda
```diff
@@ -52,6 +50,16 @@ invert : ∀ {b} → Reflects A b → if b then A else ¬ A
...
+------------------------------------------------------------------------
+-- recompute
+
+-- Given an irrelevant proof of a reflected type, a proof can
+-- be recomputed and subsequently used in relevant contexts.
+
+recompute : ∀ {b} → Reflects A b → Recomputable A
+recompute (ofʸ  a) _ = a
+recompute (ofⁿ ¬a) a = weak-contradiction a ¬a
+
```
jamesmckinna, line 62: Should this be renamed to `reflects⇒recomputable` on the model of `Relation.Unary.Consequences` and `dec⇒recomputable`  in `Relation.Nullary.Decidable.Core` above?

JacquesCarette: I must admit that I rather like 'recompute' for this one. But it's not a strong opinion.

src/Relation/Nullary/Decidable/Core.agda
gallais: Depends whether we see `Dec` as a predicate or a datatype
e.g. we use the name `toVec` rather than `list⇒vec`.

src/Relation/Nullary/Recomputable.agda
```diff
@@ -0,0 +1,39 @@
...
+(rA ×-recompute rB) p = rA (p .proj₁) , rB (p .proj₂)
+
+_→-recompute_ : (A : Set a) → Recomputable B → Recomputable (A → B)
+(A →-recompute rB) f a = rB (f a)
+
+Π-recompute : (B : A → Set b) → (∀ x → Recomputable (B x)) → Recomputable (∀ x → B x)
+Π-recompute B rB f a = rB a (f a)
+
+∀-recompute : (B : A → Set b) → (∀ {x} → Recomputable (B x)) → Recomputable (∀ {x} → B x)
+∀-recompute B rB f = rB f
```
gallais, line 39: Can we add a proof that sigma is *not* recomputable?
Or is that a meta-theoretical result?

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Nice analysis! (as with our other discussion on #2059 regarding `toSum`/`toDec` etc.)

I guess this means that, here at least, I have an inflection towards 'property' over 'datatype' (hence also my raising of #2091 )... but whether that's an 'innate' choice, or conditioned by/contingent on all of this happening under the aegis of `Nullary`, I'm not sure.

Cf. `Maybe` as a datatype constructor; `WeaklyDecidable` as its `Nullary` property incarnation...

src/Relation/Nullary/Recomputable.agda
jamesmckinna: Oh... great question (there's even actually something funny about the `_×_` case, in that it requires use of the projections, where I might have thought in the presence of eta that pattern-matching would suffice).

Not sure of the answer: the 'classical' picture was/is that what I've started calling the 'Harrop universe' excludes `∃` and `_⊎_`, towards Harrop's theorem about EP and DP being provable even under hypotheses in CL, provided all the assumptions are Harrop...

Danel Ahman remarked to me that all of this was defining a 'funny' universe of containers, so I suppose it's reasonable to ask what other container(functor)s enjoy the Harrop property, but that wasn't my original motivation, which was to understand the notion of being `Recomputable` as a counterpart to the orthodox realisability notion of 'computationally irrelevant' (as distinct from the mere `.` annotation, or more recent efforts on [Definitional Proof-Irrelevance Without K](https://inria.hal.science/hal-01859964v2/document)...).

Long term, what I want is a tactic which can take suitable (and easy to search for) _classical_ proofs of implications, and turn them into intuitionistically/type-theoretically acceptable ones (hence my question on the Zulip about Aczel/Smith-style slash interpretations, which do I think extend the fragment beyond Harrop, but ... It's All A Long Time Ago), by working with the Harrop metatheorem instantiated to the `Decidable`/`Stable` case.

But these things are (to me, at least) interesting research avenues, and (hence?) for the time being, out of scope for the PR ;-). But if you know/can say more about `Σ`, I'd be very interested!

jamesmckinna:
Converting to DRAFT for the time being as I haven't had the brain space to revisit the questions above.

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -51,13 +52,22 @@ _¬-⊎_ = [_,_]
...
 
 contraposition : (A → B) → ¬ B → ¬ A
 contraposition f ¬b a = contradiction (f a) ¬b
 
+weak-contradiction : .A → ¬ A → Whatever
+weak-contradiction a ¬a = ⊥-elim-irr (¬a a)
+
+irr-contradiction : .A → .(¬ A) → Whatever
+irr-contradiction a ¬a = ⊥-elim-irr (¬a a)
+
```
jamesmckinna, line 70: Can be derived from `weak-contradiction` and `¬-recompute`, now so omitted.

src/Relation/Nullary/Reflects.agda
jamesmckinna: Sticking with `recompute` for now...

jamesmckinna:
Incoming commits tidy this up, and then ready for review.

src/Relation/Nullary/Decidable/Core.agda
jamesmckinna: Sticking with `recompute` (for the time being)...

jamesmckinna:
I further conjecture on #2329 that all uses of `⊥-elim` can be replaced by their irrelevant version `⊥-elim-irr`... so question for the maintainers, are we prepared to embrace such a (breaking?) change to the type of `⊥-elim`?

MatthewDaggitt:
> so question for the maintainers, are we prepared to embrace such a (breaking?) change to the type of ⊥-elim?

Yes, but in v3.0! I suggest we make an issue with the 3.0 milestone?

jamesmckinna:
@MatthewDaggitt fair point, but my belief (Modulo the merge conflict) is that everything here is v2.1 compatible so I won't pursue further changes this time around... ;-)

jamesmckinna:
@JacquesCarette after the merge of #2329 could you review this one please?

CHANGELOG.md
```diff
@@ -394,6 +408,13 @@ Additions to existing modules
   WeaklyDecidable : Set _
   ```
 
+* Added new definitions in `Relation.Nullary.Negation.Core`:
+  ```agda
+  weak-contradiction : .A → ¬ A → Whatever
+  contradictionᵒ     : ¬ A → A → Whatever
```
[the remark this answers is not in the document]
jamesmckinna: Fair point... but once I found myself using `(flip contradiction)` more than 3 times in a single module, I decided I wanted/needed the abbreviation. 

Interestingly, even that form alone is not quite enough, especially when it interacts with additional implicit arguments and/or functions not being fully eta-expanded. But indeed, I'm happy to take the pushback...

src/Relation/Nullary/Negation.agda
```diff
@@ -73,7 +72,7 @@ open import Relation.Nullary.Negation.Core public
 -- ⊥).
 
 call/cc : ((A → Whatever) → DoubleNegation A) → DoubleNegation A
-call/cc hyp ¬a = hyp (λ a → ⊥-elim (¬a a)) ¬a
+call/cc hyp ¬a = hyp (contradictionᵒ ¬a) ¬a
```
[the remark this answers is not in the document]
jamesmckinna: Well, this is one where the implicits (I think) screw things up. I'll recheck, and revert if you're correct.

src/Relation/Nullary/Reflects.agda
```diff
@@ -11,14 +11,12 @@ module Relation.Nullary.Reflects where
...
 
 open import Data.Bool.Base
-open import Data.Unit.Base using (⊤)
-open import Data.Empty
 open import Data.Sum.Base using (_⊎_; inj₁; inj₂)
 open import Data.Product.Base using (_×_; _,_; proj₁; proj₂)
 open import Level using (Level)
-open import Function.Base using (_$_; _∘_; const; id)
-
+open import Function.Base using (_$_; _∘_; const; id; flip)
```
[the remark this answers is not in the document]
jamesmckinna: A case of 'flip-flopping' (sic) and failing to on a final design choice. Good catch!

src/Data/Empty.agda
```diff
@@ -9,6 +9,7 @@
 module Data.Empty where
 
 open import Data.Irrelevant using (Irrelevant)
+open import Relation.Nullary.Recomputable using (Recomputable)
```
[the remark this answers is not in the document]
jamesmckinna: Here's a third, which I think is at once less invasive, and more congruent with the spirit of the proposal, which would be to move this 'atomic' instance to `Recomputable` instead? I think that still works without creating cycles, but I'll investigate.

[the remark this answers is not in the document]
jamesmckinna: And yes, I agree with your 'conceptual' analysis (esp. given my recent comments on `decToMaybe` on #2330 ... I'd let my thinking on this PR get stale in the meantime and was trying to the minimum possible/necessary to push it over the line. Thanks for the nudge to reconsider in less haste...

src/Relation/Nullary/Recomputable.agda
```diff
@@ -0,0 +1,39 @@
...
+open import Level using (Level)
+
+private
+  variable
+    a b : Level
+    A : Set a
+    B : Set b
+
+------------------------------------------------------------------------
+-- Definition
```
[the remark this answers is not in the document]
jamesmckinna: Perhaps you're right, but I personally am not that fond of modules containing at most one definition, given that the whole thing is only 40 lines long (I think you are even more extreme than me about 'small modules' and dependency thrashing...).

src/Relation/Nullary/Negation.agda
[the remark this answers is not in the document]
jamesmckinna: This one is fine: reverted.

src/Relation/Nullary/Decidable.agda
```diff
@@ -52,8 +51,8 @@ via-injection inj _≟_ x y = map′ injective cong (to x ≟ to y)
 -- A lemma relating True and Dec
 
 True-↔ : (a? : Dec A) → Irrelevant A → True a? ↔ A
-True-↔ (true  because [a]) irr = mk↔ₛ′ (λ _ → invert [a]) _ (irr (invert [a])) cong′
-True-↔ (false because ofⁿ ¬a) _ = mk↔ₛ′ (λ ()) (invert (ofⁿ ¬a)) (⊥-elim ∘ ¬a) λ ()
+True-↔ (true  because [a]) irr = let a = invert [a] in mk↔ₛ′ (λ _ → a) _ (irr a) cong′
+True-↔ (false because [¬a]) _  = let ¬a = invert [¬a] in mk↔ₛ′ (λ ()) ¬a (λ a → contradiction a ¬a) λ ()
```
jamesmckinna, line 55: The implicits break putting just `(flip contradiction ¬a) here:
```agda
/home/james/RESEARCH/AGDA/Agda-stdlib-dev-v2.1/src/Relation/Nullary/Decidable.agda:55,75-96
Cannot instantiate the metavariable _C_128 to solution
(λ ()) (invert [¬a] x) ≡ x since it contains the variable x
which is not in scope of the metavariable
when checking that the expression flip contradiction ¬a has type
Function.Definitions.StrictlyInverseˡ _≡_ (λ ()) ¬a
```
and I'm stubborn enough to want to not see `⊥-elim` *anywhere* outside `Relation.Nullary.Negation.Core` or `Data.Empty`... (part of the point of this PR)

src/Data/Empty.agda
```diff
@@ -35,3 +35,6 @@ private
 
 ⊥-elim : ∀ {w} {Whatever : Set w} → ⊥ → Whatever
 ⊥-elim ()
+
+⊥-elim-irr : ∀ {w} {Whatever : Set w} → .⊥ → Whatever
+⊥-elim-irr ()
```
jamesmckinna, line 40: Does anyone have a better name for this?

src/Relation/Nullary/Recomputable.agda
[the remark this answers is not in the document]
jamesmckinna: So for now, I've left things as they are, plus moving the two other instances to here from `Data.Empty` and `Relation.Nullary.Negation.Core`, which may rejig the dependency graph more to your liking.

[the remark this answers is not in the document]
jamesmckinna: And at least now it could easily be broken up into `Base` and `Properties` if you insisted on it... but that feels like an easy downstream refactoring...

JacquesCarette (approved):

src/Data/Empty.agda
JacquesCarette: To be explicit: I do not.

src/Relation/Nullary.agda
```diff
@@ -23,8 +23,9 @@ private
 ------------------------------------------------------------------------
 -- Re-exports
 
+open import Relation.Nullary.Recomputable public using (Recomputable)
 open import Relation.Nullary.Negation.Core public
-open import Relation.Nullary.Reflects public
+open import Relation.Nullary.Reflects public hiding (recompute)
```
MatthewDaggitt, line 28: Unless I'm missing something, this is a breaking change and therefore we probably shouldn't do it in v2.1?

src/Relation/Nullary/Recomputable.agda
```diff
@@ -0,0 +1,53 @@
...
+
+Recomputable : (A : Set a) → Set a
+Recomputable A = .A → A
+
+------------------------------------------------------------------------
+-- Constructions
```
MatthewDaggitt, line 29: Empty section heading?

src/Relation/Nullary/Recomputable.agda
```diff
@@ -0,0 +1,53 @@
...
+  variable
+    a b : Level
+    A : Set a
+    B : Set b
+
+------------------------------------------------------------------------
+-- Definition
+
+Recomputable : (A : Set a) → Set a
+Recomputable A = .A → A
```
MatthewDaggitt, line 26: Maybe a small comment (one or two sentences) just explaining what `Recomputable` means in terms of irrelevance? For those new-comers to the library who may not know what irrelevance is in much detail?

src/Relation/Nullary/Recomputable.agda
```diff
@@ -0,0 +1,53 @@
...
+
+------------------------------------------------------------------------
+-- Constructions
+
+------------------------------------------------------------------------
+-- Recomputable types are Harrop
```
MatthewDaggitt, line 32: I feel that this section heading raises more questions than it answers for those not well versed in constructive mathematics. It's fine to mention Harrop constructions in comments, but maybe keep it out of section headings?

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -48,23 +47,26 @@ _¬-⊎_ = [_,_]
 ------------------------------------------------------------------------
 -- Uses of negation
 
+weak-contradiction : .A → ¬ A → Whatever
+weak-contradiction a ¬a = ⊥-elim-irr (¬a a)
```
MatthewDaggitt, line 51: I think the name would be `weakContradiction`?

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -48,23 +47,26 @@ _¬-⊎_ = [_,_]
...
+
 contradiction : A → ¬ A → Whatever
-contradiction a ¬a = ⊥-elim (¬a a)
+contradiction a = weak-contradiction a
 
 contradiction₂ : A ⊎ B → ¬ A → ¬ B → Whatever
-contradiction₂ (inj₁ a) ¬a ¬b = contradiction a ¬a
-contradiction₂ (inj₂ b) ¬a ¬b = contradiction b ¬b
+contradiction₂ (inj₁ a) ¬a ¬b = weak-contradiction a ¬a
+contradiction₂ (inj₂ b) ¬a ¬b = weak-contradiction b ¬b
```
MatthewDaggitt, line 58: Are these refactorings really necessary? It's not like they rely on the contradiction being weak? I would advocate for keeping them using `contradiction` except where necessary? Not to mention it's just plain shorter.

src/Relation/Nullary/Reflects.agda
```diff
@@ -64,40 +72,40 @@ T-reflects false = of id
...
-fromEquivalence {b = false} sound complete = ofⁿ complete
+fromEquivalence {b = true}  sound complete = of (sound _)
+fromEquivalence {b = false} sound complete = of complete
 
 -- `Reflects` is deterministic.
 det : ∀ {b b′} → Reflects A b → Reflects A b′ → b ≡ b′
 det (ofʸ  a) (ofʸ  _) = refl
-det (ofʸ  a) (ofⁿ ¬a) = contradiction a ¬a
-det (ofⁿ ¬a) (ofʸ  a) = contradiction a ¬a
+det (ofʸ  a) (ofⁿ ¬a) = weak-contradiction a ¬a
```
MatthewDaggitt, line 107: Again I wouldn't change unless necessary?

MatthewDaggitt (changes requested):

src/Relation/Nullary.agda
jamesmckinna: Don't think so. 
Previously, `recompute` was defined in `Decidable.Core`, and not at all in `Reflects`.
Now it's defined in `Reflects` for `Reflects A b` instances, then used to re-define `recompute` in `Decidable.Core`, so it's simulatantouesly been newly introduced in `Reflects` and then hidden from clashing with the 'old' definition in `Decidable.Core` in this PR. So I think that's *not* a breaking change? Maybe the `CHANGELOG` needs fixing, but I think it's OK...?

src/Relation/Nullary/Negation/Core.agda
jamesmckinna: `contradiction-irr` is at least congruent with `⊥-elim-irr` so I'll make that change for the time being

src/Relation/Nullary/Negation/Core.agda
jamesmckinna: Have reverted.

src/Relation/Nullary/Recomputable.agda
jamesmckinna: Done in incoming commit.

src/Relation/Nullary/Recomputable.agda
jamesmckinna: Have telescoped/collapsed into the next section...

src/Relation/Nullary.agda
jamesmckinna: Do let me know if the present `CHANGELOG` correctly captures the refactoring (or not!)

jamesmckinna:
Now : proofs computed using `recompute` are *irrelevant*, so UIP follows immediately from `DecidableEquality`.
I've added these in as further `low-hanging-fruit` cherry-picked from #2149

jamesmckinna:
@MatthewDaggitt are you happy with this not being a `breaking` change now?

jamesmckinna:
As I understand things, there are (now!) two outstanding issues:
* is this PR `breaking` or not? I claim it is not, because specifically: in the top-level `stdlib` API, the `public` export of `recompute` (and now, additionally, `recompute-constant}`) from `Relation.Nullary` is still that from `Relation.Nullary.Decidable` (and hence from  `Relation.Nullary.Decidable.Core`), and *not* that of the newly introduced `Relation.Nullary.Reflects.recompute{-constant}`.
* does https://github.com/agda/agda/issues/7193 affect the 'safety' of this PR? From my (limited) understanding of @dolio 's reply to my question there, the answer is also that it does not.  

Hope this helps! (And removes any further obstruction to merging...)

JacquesCarette (approved):
After review of the latest stuff: it continues to look good and, this time, does look backwards compatible as well.

andreasabel (approved):
(Reviewing the irrelevance parts):  To me, these look like mainstream uses of irrelevance, not affected by agda/agda#7193

JacquesCarette (approved):

JacquesCarette:
Just need @gallais to re-review.

jamesmckinna:
> Just need ... to re-review.

And @MatthewDaggitt ?

MatthewDaggitt (approved):

PR #2244, opened 2024-01-03
Nagata's "idealization of a module" construction

jamesmckinna:
This addition is the non-commutative analogue of Nagata Masayoshi's construction of the "idealization of a module", but only as far as the fundamental lemma that the construction does, indeed, define a `Ring`. 

In the absence of a developed theory of ideals of rings, and sub-modules of a module, the order isomorphism between the (lattice of) submodules of `M` and ideals of `R ⋉ M` will have to wait for another time... ;-)

Ditto. all the variations on the construction which weaken the underlying `Ring` and `Module` structures to `Semiring`, `Semimodule` etc. as actually used in the [recent paper on Automatic Differentiation](https://www.sciencedirect.com/science/article/pii/S0167642323000928) with my co-authors Birthe van den Berg, Tom Schrijvers and Alexander Vandenbroucke.


src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,171 @@
...
+-- The Fundamental Lemma, due to Nagata
+
+  isRingᴺ : IsRing _≈ᴺ_ _+ᴺ_ _*ᴺ_ -ᴺ_ 0ᴺ  1ᴺ
+  isRingᴺ = record
+    { +-isAbelianGroup = +ᴺ-isAbelianGroup
+    ; *-cong = *ᴺ-cong
+    ; *-assoc = *ᴺ-assoc
+    ; *-identity = *ᴺ-identity
+    ; distrib = *ᴺ-distrib-+ᴺ
+    --; zero = zeroᴺ
```
jamesmckinna, line 160: Oops, this slipped through.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The non-commutative analogue of Nagata's construction
+-- of the "idealization of a module", defined here
+-- on R-R-*bi*modules M over a ring R.
```
JacquesCarette, line 6: Link to paper?

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+--
+-- Elements of R ⋉ M are pairs |R| × |M|
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Algebra.Module.Construct.Nagata where
+
+open import Algebra.Bundles
```
JacquesCarette, line 15: I remain a big fan of putting in as many `using` clauses as possible. Once you use 8+ things from something, sure, that gets excessive, but otherwise, it serves as helpful documentation as well as forward-proofing the code.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+  _*ᴺ_ : Op₂ N
+  (r₁ , m₁) *ᴺ (r₂ , m₂) = r₁ *ᴿ r₂ , r₁ *ₗ m₂ +ᴹ m₁ *ᵣ r₂
+
+  *ᴺ-identity : Identity 1ᴺ _*ᴺ_
+  *ᴺ-identity = lᴺ , rᴺ
+    where
+      open ≈-Reasoning setoidᴹ
+      lᴺ : LeftIdentity 1ᴺ _*ᴺ_
+      lᴺ (r , m) = (*ᴿ-identityˡ r) , (begin
+        (1ᴿ *ₗ m +ᴹ 0ᴹ *ᵣ r) ≈⟨ +ᴹ-cong (*ₗ-identityˡ m) (*ᵣ-zeroˡ r) ⟩
```
JacquesCarette, line 102: I think it is worthwhile doing the vertical alignment of the equalities (and qed) even here.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+      rᴺ (r₁ , m₁) (r₂ , m₂) (r₃ , m₃) = *ᴿ-distribʳ-+ᴿ r₁ r₂ r₃ , (begin
+        (r₂ +ᴿ r₃) *ₗ m₁ +ᴹ (m₂ +ᴹ m₃) *ᵣ r₁
+          ≈⟨ +ᴹ-cong (*ₗ-distribʳ m₁ r₂ r₃) (*ᵣ-distribʳ r₁ m₂ m₃) ⟩
+        (r₂ *ₗ m₁ +ᴹ r₃ *ₗ m₁) +ᴹ (m₂ *ᵣ r₁ +ᴹ m₃ *ᵣ r₁)
+          ≈⟨ +ᴹ-middleFour (r₂ *ₗ m₁) (r₃ *ₗ m₁) (m₂ *ᵣ r₁) (m₃ *ᵣ r₁) ⟩
+        (r₂ *ₗ m₁ +ᴹ m₂ *ᵣ r₁) +ᴹ (r₃ *ₗ m₁ +ᴹ m₃ *ᵣ r₁) ∎)
+
+
+------------------------------------------------------------------------
+-- The Fundamental Lemma, due to Nagata
```
JacquesCarette, line 151: I was hoping to find a few more comments, likely higher up in the file, that describe what is new and what is old in the following right structure. As far as I can tell, the additive structure of the bimodule is kept as is, and it is the multiplicative structure which is defined in a novel way.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+      open Consequences setoidᴹ
+      +ᴹ-middleFour = comm∧assoc⇒middleFour +ᴹ-cong +ᴹ-comm +ᴹ-assoc
+      lᴺ : _*ᴺ_ DistributesOverˡ _+ᴺ_
+      lᴺ (r₁ , m₁) (r₂ , m₂) (r₃ , m₃) = *ᴿ-distribˡ-+ᴿ r₁ r₂ r₃ , (begin
+        r₁ *ₗ (m₂ +ᴹ m₃) +ᴹ m₁ *ᵣ (r₂ +ᴿ r₃)
+          ≈⟨ +ᴹ-cong (*ₗ-distribˡ r₁ m₂ m₃) (*ᵣ-distribˡ m₁ r₂ r₃) ⟩
+        (r₁ *ₗ m₂ +ᴹ r₁ *ₗ m₃) +ᴹ (m₁ *ᵣ r₂ +ᴹ m₁ *ᵣ r₃)
+          ≈⟨ +ᴹ-middleFour (r₁ *ₗ m₂) (r₁ *ₗ m₃) (m₁ *ᵣ r₂) (m₁ *ᵣ r₃) ⟩
+        (r₁ *ₗ m₂ +ᴹ m₁ *ᵣ r₂) +ᴹ (r₁ *ₗ m₃ +ᴹ m₁ *ᵣ r₃) ∎)
+      rᴺ : _*ᴺ_ DistributesOverʳ _+ᴺ_
```
JacquesCarette, line 141: given how much you like extra white space (way more than I do!), I am rather surprise to see none here. The definitions here do blur into each other, so I would have appreciated the use of space to make that clearer).

Have you considered using `a,b,c,d` and `m,n,o,p` instead of subscripts?  If I were doing this, I'd be tempted to use font to distinguish between Ring and Module variables (i.e. bold for Module).

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+  ιᴿ : R → N
+  ιᴿ r = r , 0ᴹ
+
+  ιᴹ : M → N
+  ιᴹ m = 0ᴿ , m
+
+  1ᴺ : N
+  1ᴺ = ιᴿ 1ᴿ
+
+  open AbelianGroup +ᴺ-abelianGroup
```
JacquesCarette, line 86: Personally, I prefer all my `open` together and not interleave them with definitions.

JacquesCarette (changes requested):
Nice. Funny that that's essentially 90 lines of 'prelude' before anything real starts happening...

jamesmckinna:
Great feedback, many thanks. 
Longer response, and changes, to follow in due course.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
+
+private
+  variable
+    m ℓm r ℓr : Level
+
+------------------------------------------------------------------------
+-- Definitions
+
+module Nagata (ring : Ring r ℓr) (bimodule : Bimodule ring ring m ℓm) where
```
MatthewDaggitt, line 38: We can lift these to top-level module parameters (see comment about the infix notation below for what to do with it).

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+    ; *-identity = *ᴺ-identity
+    ; distrib = *ᴺ-distrib-+ᴺ
+    }
+
+------------------------------------------------------------------------
+-- Bundle
+
+infixl 4 _⋉_
+
+_⋉_ : (R : Ring r ℓr) (M : Bimodule R R m ℓm) → Ring (r ⊔ m) (ℓr ⊔ ℓm)
```
MatthewDaggitt, line 167: So with the top-level module being parameterised, you can still have this definition. For starters I would rename it to `ringᴺ : Ring (r ⊔ m) (ℓr ⊔ ℓm)` as the main definition so that it can be used when the top-level module is opened applied. Then you can define a simple alias `_⋉_ = ringᴺ` so that it can be used when the the top-level module is opened unapplied.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+private
+  variable
+    m ℓm r ℓr : Level
+
+------------------------------------------------------------------------
+-- Definitions
+
+module Nagata (ring : Ring r ℓr) (bimodule : Bimodule ring ring m ℓm) where
+
+  open Ring ring
```
MatthewDaggitt, line 40: Instead of all this renaming I think a nicer pattern everywhere would be
```agda
open private module R = Ring ring 
  renaming (Carrier to R)
```
then you can simply use `R._+_` rather than `_+ᴿ_`  (which is easier to type anyway). This is consistent with how we resolve clashes with overloaded operators in `Data.Rational` etc. This point should probably be added to the style guide....

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+  infixl 7 _*ᴺ_
+
+  _*ᴺ_ : Op₂ N
+  (r₁ , m₁) *ᴺ (r₂ , m₂) = r₁ *ᴿ r₂ , r₁ *ₗ m₂ +ᴹ m₁ *ᵣ r₂
+
+  *ᴺ-identity : Identity 1ᴺ _*ᴺ_
+  *ᴺ-identity = lᴺ , rᴺ
+    where
+      open ≈-Reasoning setoidᴹ
+      lᴺ : LeftIdentity 1ᴺ _*ᴺ_
```
MatthewDaggitt, line 100: Can we lift the left and right proofs (here and in distributivity) up to the top-level please? They're often useful to have individually.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,170 @@
...
+      lᴺ (r , m) = (*ᴿ-identityˡ r) , (begin
```
MatthewDaggitt, line 101: I think the brackets around at least the LHS of the product are unnecessary. Maybe the RHS as well? (less sure about that)

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Do you mean: ourpaper, where I think this non-commutative generalisation first appears, or Nagata's first use in the late 1950s and his 1962 book "Local rings"?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Done, plus pruning out redundant imports altogether.

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Done, please check!

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Hopefully the preamble comments and additional ones later are (now) more helpful, but any additional feedback welcome!

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Re: whitespace, yes, on reflection (this development is now maybe 18months old, so needs considerable updating!) will do so!
Re: variables. Yes, this is perhaps more annoying, and we don't (yet) have conventions for these things. I agree subscripts can be awful, and I hadn't thought of using font/weight distinctions for this... will consider further.

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Yes, OK, have (hopefully, now) done so. Some `open`s do require intermediate defns though, such as that of the `+-bimodule` structure...

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: So this is an interesting one.
As it stands, the construction appears indeed to only take top-level parameters, so could be parametrised at the top-level as such. But I imagined a time, after later refactoring, when all the (semi-)etc. variants of this construction might all be brought under one roof: mind you, my instincts are for smaller modules, so I recoil in horror somewhat at the thought of it (plus being uncertain exactly what axiomatisation yields what; this is already a modest generalisation of what I could find in the literature, which typically focuses on case when R is commutative).

Will consider refactoring when I've done the lower-hanging stuff in yours, and @JacquesCarette 's reviews.

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Ah, that's nice. I confess I have tended to stumble over module-level scope and infixes before, so end up working as I have done here. But I'll try this!

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Yes, this does (as with other review comments you've made about my use of modules in eg #1962 ; sorry to be such a slow learner, but this Nagata code is now quite old, and I hadn't really brought it up to date for this PR, and should have before committing... )

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Ah!? OK, but I assume they would be available as projections once the `IsRing` and `Ring` instances were defined (and opened suitably... ;-)), ...

... but will do. What about naming though, if they are to be publicly visible?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: LHS: :+1:
RHS: :disappointed: 
But as with @JacquesCarette 's comment, have paid more attention to the layout now.

jamesmckinna:
@JacquesCarette  wrote:

> Nice. Funny that that's essentially 90 lines of 'prelude' before anything real starts happening...

Looping back to this now. Comment very much appreciated!

I guess this partly reflects my relative inexperience, back when I started this, of how to structure a 'large' algebraic construction, and I'm still thinking about how best to do this, from the point of view of development, exposition, and pragmatics/didactics, as well as @MatthewDaggitt 's comment on `style-guide` recommendations...

On the exposition/didactics points, it definitely has the flavour of a textbook 'given ..., we define ..., and thus have shown that we have ... defined a ...' embedded-proof-as-definition-style; perhaps reminiscent of some (analogous) patterns of reasoning in Isabelle/Isar (definition followed by tactic to establish termination proof, for example), or of Coq's `Program`/`Equations` style of function definition with embedded verification conditions...

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: > Instead of all this renaming I think a nicer pattern everywhere would be
> 
> ```agda
> open private module R = Ring ring 
>   renaming (Carrier to R)
> ```

Hmmm... this doesn't seem quite to work. What I can get working is 
```agda
private module R = Ring ring
open R renaming (...)
```
it doesn't seem (v2.6.4) to like collapsing the `private` definition with the `open` declaration?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: So... attempting to standardise names so that the `(Is)Ring` structure/bundle for `N` used the 'usual' names, I got a strange constraint error when trying to `open Definitions _≈_` after rename `_≈ᴹ_` to `_≈_`, so I left that one as-is.

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: More or less just this one to do, but still wondering whether it's a good idea?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: On further thoughts, how do I get boldface in the Unicode/Emacs-agda2-mode input encoding? 
I'd like to try it, but don't quite understand how!?

src/Algebra/Module/Construct/Nagata.agda
JacquesCarette: Whatever you think would be most helpful to the future code reader / maintainer to figure out your intent. And it's also nice to record the proper history (i.e. why is it called Nagata) here instead of forcing future readers to reconstruct.

src/Algebra/Module/Construct/Nagata.agda
```diff
@@ -0,0 +1,209 @@
...
+
+    open module N = Bimodule (DirectProduct.bimodule TensorUnit.bimodule bimodule)
+      using ()
+      renaming (Carrierᴹ to N
+               ; _≈ᴹ_ to _≈ᴺ_
+               ; _+ᴹ_ to _+_
+               ; 0ᴹ to 0#
+               ; -ᴹ_ to -_
+               ; +ᴹ-isAbelianGroup to +-isAbelianGroup
+               )
```
jamesmckinna, line 81: So I'm still doing this bit of renaming, so that later the types of the various `(Is)Ring` operations/properties are standard for that signature (rather than a lot of duplication of eg. `N._+ᴹ_` in place of `_+_`... etc.)

jamesmckinna:
What's left TODO:
- [x] some cosmetic naming/typo things (local commits not yet pushed);
- [ ] top-level parametrisation (still thinking about this).

JacquesCarette (approved):
This code is now a thing of beauty (modulo our differences in styles).

src/Algebra/Module/Construct/Nagata.agda
JacquesCarette: This is a much more 'controlled' but of renaming that seems to make sense.

src/Algebra/Module/Construct/Nagata.agda
JacquesCarette: You can't quite get 'bold' per se, but for m, you can try `\Mim`, `\Mcm`, `\MCm`, `\MIm`, `Mfm` to get a variety of weights and styles.  Here `\MIm` appeals most to my sense of style.

src/Algebra/Module/Construct/Nagata.agda
JacquesCarette: They are wonderful, thank you.

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Now done, but not sure if I did it correctly the way you suggested. And what about the potential for "unsolved metas" in using the `_⋉_` notation for the module being opened unapplied?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Thanks for the hints! I'll try this on other PRs I think, and not try to fiddle *too* much more with this one!

jamesmckinna:
Current status: suggest pick one of the following commits to merge:
* (@JacquesCarette 's last-reviewed version) unparametrised top-level module, defining parametrised construction: https://github.com/agda/agda-stdlib/pull/2244/commits/afd70a41af999f7f1c71000b709da59d74af404b
* (most recent, following @MatthewDaggitt ) parametrised top-level module, blocking subsequent extension with variants: https://github.com/agda/agda-stdlib/pull/2244/commits/c859f58037c4fc3ad31c3d870756df90f7102a01

src/Algebra/Module/Construct/Idealization.agda
```diff
@@ -0,0 +1,205 @@
...
+private
+  open module R = Ring ring
+    using ()
+    renaming (Carrier to R)
+
+  open module M = Bimodule bimodule
+    renaming (Carrierᴹ to M)
+
+  open AbelianGroup M.+ᴹ-abelianGroup
+    hiding (_≈_)
```
MatthewDaggitt, line 63: Very minor but:
```
open AbelianGroup M.+ᴹ-abelianGroup
open ≈-Reasoning ≈ᴹ-setoid
open Definitions _≈_
```
all don't need to be private and could go below the private block?

src/Algebra/Module/Construct/Nagata.agda
MatthewDaggitt: Why would using an infix name as opposed to a standard name result in unsolved metas that weren't there before?

src/Algebra/Module/Construct/Idealization.agda
```diff
@@ -0,0 +1,205 @@
...
+-- Properties: because we work in the direct sum, every proof has
+-- * an 'R'-component, which inherits directly from R, and
+-- * an 'M'-component, where the work happens
+
+*-cong : Congruent₂ _*_
+*-cong (r₁ , m₁) (r₂ , m₂) = R.*-cong r₁ r₂ , +ᴹ-cong (*ₗ-cong r₁ m₂) (*ᵣ-cong m₁ r₂)
+
+*-identityˡ : LeftIdentity 1# _*_
+*-identityˡ (r , m) = R.*-identityˡ r , (begin
+
```
MatthewDaggitt, line 110: Could we remove the empty lines between the `begin` and the reasoning everywhere? If you'd like to suggest a global library wide style change then we should open a new issue to discuss?

src/Algebra/Module/Construct/Idealization.agda
```diff
@@ -0,0 +1,205 @@
...
+  R.0# *ₗ m₂ +ᴹ m₁ *ᵣ R.0# ≈⟨ +ᴹ-cong (*ₗ-zeroˡ m₂) (*ᵣ-zeroʳ m₁) ⟩
+  0ᴹ +ᴹ 0ᴹ                 ≈⟨ +ᴹ-identityˡ 0ᴹ ⟩
+  0ᴹ                       ∎)
+
+m*m≈0 : (m : M) → ιᴹ m * ιᴹ m ≈ 0#
+m*m≈0 m = *-annihilates-ιᴹ m m
+
+------------------------------------------------------------------------
+-- Export notation
```
MatthewDaggitt, line 200: Maybe "Infix notation for when opening the module unparameterised"?

src/Algebra/Module/Construct/Nagata.agda
jamesmckinna: Fair point!

jamesmckinna:
Thanks @MatthewDaggitt for the (hopefully final!) round of review comments.
I've had time to reflect on the choices between parametrised/unparametrised, and have gone for the former (most recent commit), and can leave the generalisation(s) to `Semi*` structure(s) for another time.

Issue #2247, opened 2024-01-04
 Names and definitions in the `+-IsAbelianGroup` instance of `IsRing`


jamesmckinna:
Hi Three ~~Two~~ issues arising around `(Is)Ring` (and friends):

* #2195 do we really want the names
```agda
    ; inverse                 to -‿inverse
    ; inverseˡ                to -‿inverseˡ
    ; inverseʳ                to -‿inverseʳ
``` 
in the `+-IsAbelianGroup` instance of `IsRing`. Why not more simply `+-inverse` etc.?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2195#issuecomment-1825142036_

UPDATED: so far, the use of these things is almost completely confined to 'internal' definitions in `Algebra.{Module.}Structures` and `Algebra.{Module.}Construct`, so I'll leave well alone for now...

* should we enrich the manifest signature of `IsAbelianGroup` with an explicit subtraction operator? 
UPDATED: this already happens in `IsGroup`, though I might have expected the `Group` operation to admit two forms, and only collapse to 'subtraction' in the `IsAbelianGroup` case... (?):
```agda
infixl 6 _-_
_-_ : Op₂ A
x - y = x ∙ (y ⁻¹)
```
UPDATED though if I were to, I'd define, in `IsGroup` see #2251 (and UPDATED to reflect the discussion there): 
```agda
infixl 6 _//_
infixr 6 _\\_
_//_ : Op₂ A
x // y = x ∙ (y ⁻¹)
_\\_ : Op₂ A
x \\  y = (x ⁻¹) ∙ y
```
and rename the first into `_-_` in `IsAbelianGroup` and prove the two operations extensionally equal in `Algebra.Properties.AbelianGroup` (or somewhere else?)). Not sure about the `infix`/associativity declarations, though!
Tempting  instead, perhaps though, to define the [Mal'cev operation](https://ncatlab.org/nlab/show/Malcev+variety#examples) and specialise it appropriately...

* this is a separate issue; which might also take in the `trans (sym ...) ...` analogous operation  (not defined anywhere, AFAIK?) for the `IsEquivalence` _`(Is)Groupoid`_ operations (with `(Is)Groupoid` not defined anywhere, either AFAIK?) cf. #2249

JacquesCarette:
I agree that `-‿inverse` feels like overkill and that `+-inverse` would be more appropriate.

[...]

Issue #669, opened 2019-03-26
An alternative to generalised implication in Relation.Binary.Core?


m0davis:
I find the following definition useful.
```agda
_⇒[_]=_ : ∀ {a b ℓ₁ ℓ₂} {A : Set a} {B : Set b} →
          Rel A ℓ₁ → (B → A) → Rel B ℓ₂ → Set _
P ⇒[ f ]= Q = (P on f) ⇒ Q
```
It looks quite similar to generalised implication (or `_Preserves_⟶_`) from `Relation.Binary.Core` but I don't find the above definition there.

My current use-case is `_×_ ⇒[ MyDatatype ]= _≡_`. If there is a more clever way to write this using the standard library as-is, I'd be happy if someone pointed it out to me.

Would it make sense for me to make a pull request to add it? I'm somewhat unsure about the name. The `⇒` looks a bit strange to me sitting on the left-hand side. If it were to be added, I would figure we would need synonyms similar to `Preserves` et al. But again, I'm unsure how to name them.

MatthewDaggitt:
Quite possibly though I'm a little confused about your use case. Could you expand it for a concrete datatype? I've tried letting "MyDatatype"=List and I'm not coming up with something useful, but maybe I'm just not seeing it.

m0davis:
Here's a MWE of what I'm trying to do:
```agda
-- ** common stuff

open import Function
open import Data.Product
open import Relation.Binary.PropositionalEquality
open import Relation.Binary.Core

_⇒[_]=_ : ∀ {a b ℓ₁ ℓ₂} {A : Set a} {B : Set b} →
          Rel A ℓ₁ → (B → A) → Rel B ℓ₂ → Set _
P ⇒[ f ]= Q = (P on f) ⇒ Q

-- ** Nat stuff

open import Data.Nat

variable
  m n s : ℕ

-- ** particular datatype stuff here

-- my datatype (an encoding of summation)
data Add : ℕ → ℕ → ℕ → Set where
  zero : Add zero m m
  suc : Add n m s → Add (suc n) m (suc s)

-- a proof that Add is functional
≡-from-Add : _×_ ⇒[ Add n m ]= _≡_
≡-from-Add (zero , zero) = refl
≡-from-Add (suc add1 , suc add2) rewrite ≡-from-Add (add1 , add2) = refl

add = _+_ -- just a renaming here for brevity, but you can imagine I might have written a new function

-- creating `Add`s on-the-fly
[...]

PR #2249, opened 2024-01-05
Add derived (pre-)groupoid operations to `Relation.Binary.Is(Partial)Equivalence`

jamesmckinna:
This is a partial resolution of the second part of #2247 .

Currently, the design (wrt notation/naming, infix, exports, etc.) is speculative and up for discussion, ahead of considering propagating the use of the new combinators to the many places in the library where they might be used for more streamlined equational reasoning:
- [ ] `Algebra.Construct.LexProduct.Inner` (4 occurrences)
- [ ] `Algebra.Construct.NaturalChoice.MinOp` (5 occurrences)
- [ ] `Algebra.Properties.CommutativeSemigroup` (11 occurrences)
- [ ] `Algebra.Properties.Semiring.Divisibility` (2 occurrences)
- [ ] `Axiom.UniquenessOfIdentityProofs` (4 occurrences)
- [ ] `Codata.Guarded.Stream.Relation.Binary.Pointwise` (8 occurrences)
- [ ] `Data.Container.Relation.Unary.Any.Properties` (2 occurrences)
- [ ] `Data.Integer.Properties` (12 occurrences)
- [ ] `Data.List.Countdown` (7 occurrences)
- [ ] `Data.List.Fresh.Membership.Setoid.Properties` (2 occurrences)
- [ ] `Data.List.Membership.Propositional.Properties` (2 occurrences)
- [ ] `Data.List.Membership.Setoid.Properties.agda` (4 occurrences)
- [ ] `Data.List.Relation.Binary.BagAndSetEquality` (3 occurrences)
- [ ] `Data.List.Relation.Binary.Permutation.Propositional.Properties` (5 occurrences)
- [ ] `Data.List.Relation.Binary.Permutation.Setoid.Properties` (12 occurrences)
- [ ] `Data.List.Relation.Unary.Unique.Setoid.Properties` (2 occurrences)
- [ ] `Data.Vec.Recursive.agda` (2 occurrences)
- [ ] `Data.Vec.Relation.Binary.Equality.Cast` (3 occurrences)
- [ ] `Data.Vec.Relation.Unary.Any.Properties` (2 occurrences)
- [ ] `Relation.Binary.Construct.Closure.SymmetricTransitive` (3 occurrences)
- [ ] `Relation.Binary.Properties.Setoid` (2 occurrences)
- [ ] `Relation.Binary.PropositionalEquality` (8 occurrences)
- [ ] `Relation.Binary.PropositionalEquality.Properties` (2 occurrences)

plus a further 20+ modules with a single occurrence of one of the combinations.

UPDATED: I'm conscious of the discussion [elsewhere](https://github.com/agda/agda-stdlib/issues/669#issuecomment-486258671) of the [Fairbairn threshhold](https://wiki.haskell.org/Fairbairn_threshold), and wondering if this  proposal meets it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,3 +88,15 @@ Additions to existing modules
 
 * In `Function.Bundles`, added `_⟶ₛ_` as a synonym for `Func` that can
   be used infix.
+
+* In `Relation.Binary.Structures.IsPartialEquivalence`, added `_⊗_` as
+  an infix synonym for (the groupoid operation) `trans`, together with
+  commonly occurring derived combinations of `trans` and `sym`:
+  ```agda
+  infixr 5 _⊗_ _ᵒ⊗_ _⊗ᵒ_
+  transᵒ : RightTrans _≈_ (flip _≈_)
+  _⊗ᵒ_ = transᵒ
+  ᵒtrans : Trans (flip _≈_) _≈_  _≈_
+  _ᵒ⊗_ = ᵒtrans
+  ```
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,11 +89,10 @@ Additions to existing modules
 * In `Function.Bundles`, added `_⟶ₛ_` as a synonym for `Func` that can
   be used infix.
 
-* In `Relation.Binary.Structures.IsPartialEquivalence`, added `_⊗_` as
-  an infix synonym for (the groupoid operation) `trans`, together with
+* In `Relation.Binary.Structures.IsPartialEquivalence`, added
   commonly occurring derived combinations of `trans` and `sym`:
   ```agda
-  infixr 5 _⊗_ _ᵒ⊗_ _⊗ᵒ_
+  infixr 5 _ᵒ⊗_ _⊗ᵒ_
   transᵒ : RightTrans _≈_ (flip _≈_)
   _⊗ᵒ_ = transᵒ
   ᵒtrans : Trans (flip _≈_) _≈_  _≈_
--- a/src/Relation/Binary/Structures.agda
+++ b/src/Relation/Binary/Structures.agda
@@ -16,6 +16,7 @@ module Relation.Binary.Structures
   where
 
 open import Data.Product.Base using (proj₁; proj₂; _,_)
+open import Function.Base using (flip)
 open import Level using (Level; _⊔_)
 open import Relation.Nullary.Negation.Core using (¬_)
 open import Relation.Binary.PropositionalEquality.Core as P using (_≡_)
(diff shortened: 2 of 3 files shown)


MatthewDaggitt:
> UPDATED: I'm conscious of the discussion https://github.com/agda/agda-stdlib/issues/669#issuecomment-486258671 of the [Fairbairn threshhold](https://wiki.haskell.org/Fairbairn_threshold), and wondering if this proposal meets it?

Again, my personal view is that it doesn't....

I'm also not that keen on the name for two reasons. Firstly it is not consistent with the naming style for other combinators in the `Setoid` hierarchy. Secondly, it's kind of a common name, so I'm kind of reluctant to pollute every namespace that ever opens a `Setoid` with it. I suspect this would break a lot of user code.... (it certainly breaks some of my personal library)

jamesmckinna:
Thanks Matthew for the feedback!

On syntax per se, I agree that the proposed additions differ/diverge from existing stdlib convention but that's partly, again in my opinion, that the library isn't that good at supporting infix *proof* combinators, for whatever reason. But the particular `trans ... (sym ...)` pattern seems sufficiently common (witness the task list above) to merit the addition, although without agreeing in principle to the idea, I wasn't about to systematically apply it throughout... 

On the actual *choice*s of name, which name(s) clash for you, or all of them? I can/could/might see that `_⊗_` might be a deal-breaker, but that's the one I'd find easiest to jettison, other than it's being the 'root' syntax for the others...

JacquesCarette:
Hate the name too.

However, I've defined (and used) a combinator for this **a lot** in `agda-categories`. See the bottom of [Category.Core](https://agda.github.io/agda-categories/Categories.Category.Core.html) for the details.

While it completely fails the threshold, the increase in readability of some code very strongly makes up for it.

Personally, I'd rather use some kind of 'fat semi colon' instead of a variant on a composition operator. And my 'flip' operation is too big.

jamesmckinna:
OK, @JacquesCarette so "hate" is strong, but perhaps grabbing `_⊗_` is too much of a 'provocation', except that from [my perspective](https://github.com/agda/agda-stdlib/issues/2247), transitivity (and its name) do derive from the underlying bicategory/groupoid structure of relational/equational proofs, so it seemed 'ok' to deploy the symbol here... but I'm happy for the pushback.

I'll look at your usage in `agda-categories`... and think some more. I'm personally happy to stick to the new prefix names, if they were accepted, and only the infix symbolic names for those. But it seems as though this might need more discussion!

JacquesCarette:
There are times where strong language is the simplest way to communicate one's strong feelings, especially in a written medium like this. I wanted to ambiguity.

I think I see things differently here: transitivity does not derive from the structure of equational proof, but is part of the defining structure of being an "equational proof structure", i.e. structure vs property. You prove that things satisfy the API of being a groupoid, you don't start with an arbitrary groupoid.

I'm fine with picking something other than a semicolon; I do think that since `trans` is fundamentally ordered, we should not pick a symmetric symbol for it. So `agda-categories`'s choice there is downright bad too!

MatthewDaggitt:
I'm still not super keen on this idea, but if you are keen to push for it @jamesmckinna then my naming suggestions would be `trans-symˡ` and `trans-symʳ`

MatthewDaggitt:
Ah drat, that name is already taken the proof about it... https://github.com/search?q=repo%3Aagda%2Fagda-stdlib%20trans-sym%CB%A1&type=code

JacquesCarette:
On revisit, I think I'm siding with @MatthewDaggitt on this (i.e. not keen). If one were to use equational reasoning in these cases, all the needed kit is already in place. So this only saves a little bit when doing equational proofs inline.

I think it's actually a good idea to not make those 'convenient' !

jamesmckinna:
I take the consensus view: the 'Noes' have it!

PR #2251, opened 2024-01-11
Add new operations (cf. `RawQuasigroup`) to `IsGroup`

jamesmckinna:
Tackles second item of #2247 .

Issues:
* deprecation of `_-_` in `IsGroup`?
* adding operations to `Algebra.Bundles.Raw.RawGroup`?

Knock-on changes (potentially):

* Instances of `_-_` defined in, and hence would clash with any opening of the `RawGroup`|`IsGroup` instance:
- [ ] `Data.Integer.Base|Properties`
- [ ] `Data.Rational.Base|Properties`
- [ ] `Data.Rational.Unnormalised.Base|Properties`

* Instances of `_-_` not defined in, but would be by opening the `RawGroup`|`IsGroup` instance:
- [ ] `Data.Parity.Base|Properties`
- [ ] `Data.Sign.Base|Properties`

### DIFF
--- a/src/Algebra/Properties/AbelianGroup.agda
+++ b/src/Algebra/Properties/AbelianGroup.agda
@@ -36,3 +36,13 @@ xyx⁻¹≈y x y = begin
   x ⁻¹ ∙ y ⁻¹ ≈⟨ ⁻¹-anti-homo-∙ y x ⟨
   (y ∙ x) ⁻¹  ≈⟨ ⁻¹-cong $ comm y x ⟩
   (x ∙ y) ⁻¹  ∎
+
+module _ where
+  open IsGroup isGroup using (_//_; _\\_)
+
+  \\≗//ᵒ : ∀ x y → x \\ y ≈ y // x
+  \\≗//ᵒ x y = begin
+    x \\ y    ≈⟨ refl ⟩
+    x ⁻¹ ∙ y  ≈⟨ comm _ _ ⟩
+    y ∙ x ⁻¹  ≈⟨ refl ⟩
+    y // x    ∎
--- a/src/Algebra/Properties/AbelianGroup.agda
+++ b/src/Algebra/Properties/AbelianGroup.agda
@@ -36,13 +36,3 @@ xyx⁻¹≈y x y = begin
   x ⁻¹ ∙ y ⁻¹ ≈⟨ ⁻¹-anti-homo-∙ y x ⟨
   (y ∙ x) ⁻¹  ≈⟨ ⁻¹-cong $ comm y x ⟩
   (x ∙ y) ⁻¹  ∎
-
-module _ where
-  open IsGroup isGroup using (_//_; _\\_)
-
-  \\≗//ᵒ : ∀ x y → x \\ y ≈ y // x
-  \\≗//ᵒ x y = begin
-    x \\ y    ≈⟨ refl ⟩
-    x ⁻¹ ∙ y  ≈⟨ comm _ _ ⟩
-    y ∙ x ⁻¹  ≈⟨ refl ⟩
-    y // x    ∎
--- a/src/Algebra/Properties/Group.agda
+++ b/src/Algebra/Properties/Group.agda
@@ -17,6 +17,12 @@ open import Relation.Binary.Reasoning.Setoid setoid
 open import Function.Base using (_$_; _⟨_⟩_)
 open import Data.Product.Base using (_,_; proj₂)
 
+\\-cong₂ : Congruent₂ _\\_
+\\-cong₂ x≈y u≈v = ∙-cong (⁻¹-cong x≈y) u≈v
+
+//-cong₂ : Congruent₂ _//_
+//-cong₂ x≈y u≈v = ∙-cong x≈y (⁻¹-cong u≈v)
+
 leftDividesˡ : ∀ x y → x ∙ (x \\ y) ≈ y
 leftDividesˡ x y = begin
   x  ∙ (x \\ y)  ≈⟨ assoc x (x ⁻¹) y ⟨
@@ -47,12 +53,12 @@ rightDividesʳ x y = begin
 
 isQuasigroup : IsQuasigroup _∙_ _\\_ _//_
 isQuasigroup = record
-                 { isMagma = isMagma
-                 ; \\-cong = λ x≈y u≈v → ∙-cong (⁻¹-cong x≈y) u≈v
-                 ; //-cong = λ x≈y u≈v → ∙-cong x≈y (⁻¹-cong u≈v)
-                 ; leftDivides = leftDividesˡ , leftDividesʳ
-                 ; rightDivides = rightDividesˡ , rightDividesʳ
-                 }
+  { isMagma = isMagma
+  ; \\-cong = \\-cong₂
+  ; //-cong = //-cong₂
+  ; leftDivides = leftDividesˡ , leftDividesʳ
+  ; rightDivides = rightDividesˡ , rightDividesʳ
+  }
 
 isLoop : IsLoop _∙_ _\\_ _//_ ε
 isLoop = record { isQuasigroup = isQuasigroup ; identity = identity }
--- a/src/Algebra/Properties/Group.agda
+++ b/src/Algebra/Properties/Group.agda
@@ -134,11 +134,9 @@ inverseʳ-unique x y eq = begin
 
 \\≗//ᵒ⇒comm : (∀ x y → x \\ y ≈ y // x) → Commutative _∙_
 \\≗//ᵒ⇒comm \\≗//ᵒ x y = begin
-  x ∙ y                ≈⟨ identityʳ _ ⟨
-  (x ∙ y) ∙ ε          ≈⟨ ∙-congˡ (inverseˡ x) ⟨
-  (x ∙ y) ∙ (x ⁻¹ ∙ x) ≈⟨ assoc (x ∙ y) (x ⁻¹) x ⟨
-  (x ∙ y) ∙ x ⁻¹ ∙ x   ≈⟨ ∙-congʳ (assoc x y (x ⁻¹)) ⟩
-  x ∙ (y // x) ∙ x     ≈⟨ ∙-congʳ (∙-congˡ (\\≗//ᵒ x y)) ⟨
+  x ∙ y                ≈⟨ ∙-congˡ (rightDividesˡ x y) ⟨
+  x ∙ ((y // x) ∙ x)   ≈⟨ ∙-congˡ (∙-congʳ (\\≗//ᵒ x y)) ⟨
+  x ∙ ((x \\ y) ∙ x)   ≈⟨ assoc x (x \\ y) x ⟨
   x ∙ (x \\ y) ∙ x     ≈⟨ ∙-congʳ (leftDividesˡ x y) ⟩
   y ∙ x                ∎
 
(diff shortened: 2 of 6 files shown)


MatthewDaggitt:
Hmm so a few questions.

1. Firstly, I think this definitely falls foul of the [Fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold)? I was going to suggest `_∙⁻¹_` should be `_∙_⁻¹` but then why not just write out the full thing?
2. Why do you not export them for `AbelianGroup`?

jamesmckinna:
Matthew, thanks for the attentive close reading...

>     1. Firstly, I think this definitely falls foul of the [Fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold)? I was going to suggest `_∙⁻¹_` should be `_∙_⁻¹` but then why not just write out the full thing?

So, indeed that's a fair question: one reason to introduce a distinct symbol is to avoid ambiguous parsing, which would otherwise definitely occur with your suggestion. Here the situation is slightly different from #2249 (though clearly cognate, and arising from the same root issue #2247), not least because we don't have so much group theory formalised yet to see the recurrence of the pattern... other than as the derived form `_-_`, but that doesn't belong in `Group`... on a 'conventional' reading of the symbol

>     2. Why do you not export them for `AbelianGroup`?

Again, this was a 'design' choice, underlying the whole PR: I don't think `_-_` is a sensible choice of operator symbol for `Group` *at all*, while it surely is for (our customary informal intuitions/practice around) `AbelianGroup`; and once you *are* commutative, then the two otherwise-distinct `Group` operations become extensionally equal... (the alternative design would be to add that lemma to `Algebra.Consequences.Group`), so it struck me that the less invasive thing to do was to retain `_-_` for `AbelianGroup` (again: that could be done by renaming the `Group` operation, but I wasn't sure about the deprecation consequences for trying to shift the operation symbol to the 'right' place), but not pollute its namespace with all the now-equivalent variants.

JacquesCarette:
It does make a lot of sense to seriously consider adding some such symbol - basically because 200 years of Group Theory has continued to do so. So they find it adds something useful to the domain, and we would need a solid reason to decide otherwise. There seems to be a strong usability / readability case to be made.

Having said that, neither `_∙⁻¹_` nor `_∙_⁻¹` seem to actually achieve that! If we can't use some kind of division-like symbol, then the increase in readability is not enough as the suggested symbols are essentially the same as writing it out in full.

It does make sense to call this `_-_` in `AbelianGroup`.

jamesmckinna:
Thanks for the prompts to rethink, but I would definitely welcome suggestions as to a 'good' notation for these: maybe the Lambek-style left- and right- 'divides' symbols are the way to go? Or does that have the wrong associations wrt `Preorder`-based symbology?

JacquesCarette:
I think `quasigroup` and `loop` use \\ and // .

JacquesCarette:
I don't have a good suggestion, as I don't really know the actual effect the refactoring would have. It might range from harmless to completely unacceptable.

jamesmckinna:
... So maybe the correct thing to do is add them in-place on their own terms, and downstream establish a lemma of the form: `IsGroup` yields `IsQuasigroup`/`IsLoop`...?

JacquesCarette:
Such lemmas would definitely be guaranteed low-disruption. Thus definitely a viable path.

jamesmckinna:
Hopefully the addition of the `isQuasigroup` and `isLoop` structural lemmas now gives more 'weight' to this PR.

Outstanding refactoring opportunities, emerging in the course of doing this:
* duplication between proofs of `Cancellative` properties for `Algebra.Properties.Quasigroup` and `Algebra.Properties.Group`;
* do those proofs belong instead in `Algebra.Consequences.Setoid`?
* simplify the new proofs added here to exploit the lemmas in `Algebra.Properties.Quasigroup`?

src/Algebra/Properties/Group.agda
```diff
@@ -96,13 +121,24 @@ identity-unique {x} id = identityˡ-unique x x (proj₂ id x)
...
+  y ⁻¹ ⁻¹ ≈⟨ ⁻¹-cong (inverseˡ-unique x y eq) ⟨
   x ⁻¹    ∎
+
+\\≗//ᵒ⇒comm : (∀ x y → x \\ y ≈ y // x) → Commutative _∙_
+\\≗//ᵒ⇒comm \\≗//ᵒ x y = begin
+  x ∙ y                ≈⟨ identityʳ _ ⟨
+  (x ∙ y) ∙ ε          ≈⟨ ∙-congˡ (inverseˡ x) ⟨
+  (x ∙ y) ∙ (x ⁻¹ ∙ x) ≈⟨ assoc (x ∙ y) (x ⁻¹) x ⟨
+  (x ∙ y) ∙ x ⁻¹ ∙ x   ≈⟨ ∙-congʳ (assoc x y (x ⁻¹)) ⟩
+  x ∙ (y // x) ∙ x     ≈⟨ ∙-congʳ (∙-congˡ (\\≗//ᵒ x y)) ⟨
```
JacquesCarette, line 141: Isn't this the proof `rightDivides` run backwards?

JacquesCarette (changes requested):
This is looking really nice. There's one minor nitpick, but otherwise, I'm very happy with this.

src/Algebra/Properties/Group.agda
jamesmckinna: Ha! Entirely possible, but for having banged my head at each of these fur a while trying to eliminate redundancy, until I could no longer see the wood for the trees... will try to look with fresh eyes later or tomorrow.

jamesmckinna: Beginning to be convinced!

src/Algebra/Properties/Group.agda
```diff
@@ -96,13 +121,22 @@ identity-unique {x} id = identityˡ-unique x x (proj₂ id x)
...
 
 inverseʳ-unique : ∀ x y → x ∙ y ≈ ε → y ≈ x ⁻¹
 inverseʳ-unique x y eq = begin
-  y       ≈⟨ sym (⁻¹-involutive y) ⟩
-  y ⁻¹ ⁻¹ ≈⟨ ⁻¹-cong (sym (inverseˡ-unique x y eq)) ⟩
+  y       ≈⟨ ⁻¹-involutive y ⟨
+  y ⁻¹ ⁻¹ ≈⟨ ⁻¹-cong (inverseˡ-unique x y eq) ⟨
   x ⁻¹    ∎
+
+\\≗//ᵒ⇒comm : (∀ x y → x \\ y ≈ y // x) → Commutative _∙_
```
MatthewDaggitt, line 135: Why is there an `ᵒ` in the name? I think it would work without it?

src/Algebra/Properties/AbelianGroup.agda
```diff
@@ -36,3 +36,13 @@ xyx⁻¹≈y x y = begin
   x ⁻¹ ∙ y ⁻¹ ≈⟨ ⁻¹-anti-homo-∙ y x ⟨
   (y ∙ x) ⁻¹  ≈⟨ ⁻¹-cong $ comm y x ⟩
   (x ∙ y) ⁻¹  ∎
+
+module _ where
+  open IsGroup isGroup using (_//_; _\\_)
+
+  \\≗//ᵒ : ∀ x y → x \\ y ≈ y // x
```
MatthewDaggitt, line 43: If we have `\\≗//⇒comm`, I feel like this should be `comm⇒\\≗//` in `Properties.Group` and then appeal to that here? Also the unexplained `ᵒ` in the name again could be dropped?

MatthewDaggitt (commented):
Otherwise I'm happy!

src/Algebra/Structures.agda
```diff
@@ -253,9 +253,20 @@ record IsGroup (_∙_ : Op₂ A) (ε : A) (_⁻¹ : Op₁ A) : Set (a ⊔ ℓ) w
 
   open IsMonoid isMonoid public
 
+  infixl 6 _//_
```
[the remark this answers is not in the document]
jamesmckinna: OK... uncommitted either way!

src/Algebra/Properties/Group.agda
jamesmckinna: Hmmm... well, formally, it's not the case that the two operations are extensionally equal; there is a 'flip' which should/might (?) be accounted for... but for sure I'm maybe too wedded to the `ᵒ` modifier to connote `flip`... again, I'm not especially committed to this, but equally, I think it's hard to find a good mnemonic/memorable name for things like this in the first place, anyway...

JacquesCarette: Right - I think spelling out 'flip' here would be best.

jamesmckinna: Done.

jamesmckinna:
Only issue outstanding for me: the current design factors between
* `Structures.IsGroup` for the additional derivable operations, and
* `Properties.Group`, which takes a `Bundle`d instance to then expose the underlying `Structure`...

Does the operation material belong under `Structures` or should it all belong in the `Bundle` (leaving out the question of *which* bundle, `Raw` or otherwise... ;-))?

jamesmckinna:
Note on reflexive reasoning (comment also left on #2152 already merged into v2.0):

              ... I've had to use explicit `x ≈⟨ refl ⟩...` steps, because there is no abbreviation `≈⟨⟩` declared, whereas `≡⟨⟩` *is* available as an alternative. Was this what you intended, or should there be a bug-fixing issue/PR raised to add the additional syntax?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/pull/2152#discussion_r1472742455_

src/Algebra/Properties/Group.agda
```diff
@@ -12,44 +12,69 @@ module Algebra.Properties.Group {g₁ g₂} (G : Group g₁ g₂) where
 
 open Group G
 open import Algebra.Definitions _≈_
+open import Algebra.Structures _≈_ using (IsLoop; IsQuasigroup)
 open import Relation.Binary.Reasoning.Setoid setoid
 open import Function.Base using (_$_; _⟨_⟩_)
 open import Data.Product.Base using (_,_; proj₂)
 
+leftDividesˡ : ∀ x y → x ∙ (x \\ y) ≈ y
```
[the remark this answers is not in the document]
jamesmckinna: Are you suggesting a rewrite to exploit the 'group-as-category' intuition directly?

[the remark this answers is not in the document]
JacquesCarette: No. I'm suggesting that we port `Categories.Morphism.Reasoning` over to stdlib appropriately for use in `Algebra`.

JacquesCarette (approved):

jamesmckinna:
On the latest refactorings: they basically consist of moving lemmas from `Algebra.Properties.Group` to their 'true' homes in `Algebra.Properties.Loop`/`Algebra.Properties.Quasigroup` , and publicly (re-)exporting them from `...Group`, together with some refactoring to derive existing `Group` properties from 'simpler' (or: now already proved) lemmas, such as `SelfInverse _⁻¹` etc.

But that's it now.

src/Algebra/Properties/Group.agda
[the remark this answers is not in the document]
jamesmckinna: Great! 'We' should make a detailed plan for how you see such porting (and presumably other pieces) working out, and start chipping away at it... open a new issue to lay out the design?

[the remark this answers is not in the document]
jamesmckinna: On the lemmas themselves, they are, IIRC correctly, the triangular identities for the Lambek operations being appropriate (right-) adjoints to (left- and right-) multiplication, so not surprised that they should fall under a wider umbrella of categorical reasoning...

jamesmckinna:
You beat me to it to resolve the `merge` conflict... Thanks!

Issue #1688, opened 2022-01-13
Ring fails to re-export various raw structures


Taneb:
I haven't investigated to see how big this problem is, but `Algebra.Bundles.Ring` is missing at least `+-rawGroup` and `rawNearSemiring`

jamesmckinna:
So... Algebra.Bundles.Raw.Ring since #1755/#1810?

Taneb:
@jamesmckinna I don't think that's what I meant. I meant that the `Ring` bundle, with the laws, does not expose anything called `+-rawGroup : RawGroup c l` directly. You have to go through either `+-group` or `rawRing` to get it. It's been a long while so I can't remember why this came up, but note that some things (for example homomorphisms) are defined over the raw bundles so it's reasonable to want to get at any of the raw sub-bundles easily, and these two felt like obvious omissions.

jamesmckinna:
OK. I think I had myself only meant that even these definitions had moved. But yes, if `Ring` doesn't export them, then they should be added. But PR#1867 is still in review, and I'm loathe to add more to it now, especially if it is blocking on my having failed to clean up the commits properly. 

That said, the `public` exports in `Algebra.Bundles.Raw` from `rawRing` do export the raw bundles you want, so... doesn't a `open import Algebra.Bundles.Raw` do the trick? Apologies if I am being obtuse (I find the `Algebra` hierarchy, and the (very!) complex re-exporting discipline very hard to get my head round).

jamesmckinna:
I think I had intended to remove the 'fixes' linkage to #2254 to keep this open?
Do you want to (re-)open this @Taneb , or open a fresh issue on eg. "`Algebra.Bundles.*` should re-export all `Raw` sub-structures"?

PR #2254, opened 2024-01-12
(Partial) fix for issue #1688

jamesmckinna:
What else is missing from your shopping list, @Taneb ?
And: should the additional export behaviour be implemented in `Algebra.Bundles`, `Algebra.Bundles.Raw` (and then re-exported in the corresponding place(s) in `Algebra.Bundles`), or both? (This last seems like a bad idea from a DRY point-of-view)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,13 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* Exporting more `Raw` substructures from `Algebra.Bundles.Ring`:
+  ```agda
+  rawNearSemiring   : RawNearSemiring _ _
+  rawRingWithoutOne : RawRingWithoutOne _ _
+  +-rawGroup        : RawGroup _ _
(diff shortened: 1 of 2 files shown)


Taneb:
This is good as is, but something else I noticed recently is that `SemiringWithoutOne` has messed up re-exports (it has two `setoid`s for example, and doesn't reexport a lot of the monoid laws on + at all)

jamesmckinna:
> This is good as is, but something else I noticed recently is that `SemiringWithoutOne` has messed up re-exports (it has two `setoid`s for example, and doesn't reexport a lot of the monoid laws on + at all)

See also the discussion on #1917 / #2093 : did I/we get this one wrong?

JacquesCarette (approved):
Personally, I'd pull this in as an incremental improvement, and a new PR can improve matters even more. (Never know how long the new one will take, may as well take the improvements we have now.)

Issue #2250, opened 2024-01-09
pattern synonym in `Nat.Base` in the way when working with products


flupe:
I'm currently working in a development importing both `Data.Product.Base` and `Data.Nat.Base`.

Every time I have a hole of type `Σ A B` and I refine said hole, the following pattern gets introduced: `less-than-or-equal {?} ?` instead of `? , ?`. This stems from a pattern-synonym recently introduced in `Nat.Base`: https://github.com/agda/agda-stdlib/blame/3515c22f3a2c1bd2eeb5fdd504072f67aab3de4d/src/Data/Nat/Base.agda#L366
 
Am I doing something wrong? Or maybe this pattern-synonym should not stay in `Nat.Base`?

To reproduce, just try to refine the hole in the following code sample:

```agda
open import Data.Unit.Base
open import Data.Nat.Base
open import Data.Product.Base

t : ⊤ × ⊤
t = {!!}
```

flupe:
My bad, duplicate of #2216, known issue. Closing.

jamesmckinna:
The takeaway is that the issue should even have been *fixed* now in v2.0, so please consider yourself encouraged to migrate to that release as soon as possible. But if you still have a problem with v2.0, please let us know, either by re-opening this issue, or by opening a fresh one (better)

PR #2256, opened 2024-01-14
Remove all external use of `less-than-or-equal` beyond `Data.Nat.*`

jamesmckinna:
Following up on #2216 / #2217 (and #2250 ) with a 'pure' refactoring PR (so: no `CHANGELOG` entry required at this stage) which localises all uses of the retained-for-backwards-compatibility `pattern` synonym `less-than-or-equal` to `Data.Nat.Base`, `Data.Nat.Properties`, `Data.Nat.Properties/WithK`.

Outstanding issues (maybe):
* deprecation of the pattern synonym (in `Data.Nat`? or elsewhere?), which would require a `CHANGELOG` entry;
* the fix-up of the various `Vec` lengths in `padRight`, `padBoth` etc. at present use `rewrite`, but it might be cleaner to use `Vec.cast` with the appropriate equality (shoutout to @JacquesCarette for this);
* the ubiquity (now) of the 'pattern' of definition
```agda
f : ... (as : Vec≤ A n) ... 
f ... as@record { length = m ; vec = vs ; bound = m≤n } ...
  with ≤″-offset k ← recompute (_ ℕ.≤″? _) (ℕ.≤⇒≤″ m≤n) = ...
```
suggests that the design of the datatype might be revisited, to optimise for computing the offset `k`?

### DIFF
--- a/src/Data/Vec/Bounded/Base.agda
+++ b/src/Data/Vec/Bounded/Base.agda
@@ -31,6 +31,7 @@ private
     A : Set a
     B : Set b
     C : Set c
+    m n : ℕ
 
 ------------------------------------------------------------------------
 -- Types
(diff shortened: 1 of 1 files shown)


MatthewDaggitt (approved):

jamesmckinna:
Re: revising the design

On reflection, I think the proof pattern alluded to above could be improved off-the-bat by:
* adding to `Data.Nat.Properties`, replacing [these proofs](https://github.com/agda/agda-stdlib/blob/3515c22f3a2c1bd2eeb5fdd504072f67aab3de4d/src/Data/Nat/Properties.agda#L2121-L2128), with
```agda
≤⇒≤″ : _≤_ ⇒ _≤″_
≤⇒≤″ = less-than-or-equal ∘ m+[n∸m]≡n

-- corollary

≤-proof : ∀ {m n} (le : m ≤ n) → ∃[ k ] m + k ≡ n
≤-proof le = _ , ≤″-proof (≤⇒≤″ le)

pattern ≤-offset k = k , refl
pattern <-offset k = ≤-offset k

-- equivalence to _≤_

≤″⇒≤ : _≤″_ ⇒ _≤_
≤″⇒≤ (≤″-offset k) = m≤m+n _ k
```
* and then rewriting the above proof pattern in `Data.Vec.Bounded.Base` as
```agda
  with ℕₚ.≤-offset k ← ℕₚ.≤-proof (recompute (_ ℕₚ.≤? _) m≤n)
```
thereby encapsulating all mention even of `_≤″_` itself... not merely that of its smart constructor...

JacquesCarette (approved):

JacquesCarette:
I would be a definite fan of doing this in two stages, i.e. first pull this in, and then your further improvements later, when they land.

jamesmckinna:
OK, so let's merge this as it stands, and then I'll file a followup? Are you happy for me to hit 'merge'?

JacquesCarette:
Well, I still think it shouldn't be the author who merges - but I'm happy to do it.

jamesmckinna:
I agree re: self-merging, so I hadn't done so... ;-)

Issue #1363, opened 2020-11-23
Literals for any ring?


gallais:
Should we add an `Algebra.Structures.Literals` module declaring a `Number` instance
for any `SemiringWithoutAnnihilatingZero` interpreting `n` as `1 + ... + 1`?

JacquesCarette:
Is `SemiringWithoutAnhilatingZero` the smallest theory in the library that has the natural numbers as initial (and free theory on 0 generators)?

jamesmckinna:
Hmmm. Seems as though the free semiring on no generators is still Nat, surely?
And having Number instances for nilpotent (semi-)rings seems ok too... so that eg integers-mod-n would be able to work 'automatically' eg 3+5 reduces to 1 mod 7, or am I missing something?

jamesmckinna:
So, I think that this issue is better tackled with `Algebra.Definitions.RawMonoid._×_` (suitably renamed, if necessary), with properties and their proofs similarly in `Algebra.Properties.Monoid.Mult`. There's no need for distributivity from `Semiring{WithoutAnnihilatingZero}`, AFAICT...

Outstanding issues, though, concerning the `Number` instance: 
* do we take the `Constraint` field to be constant `⊤` as in `Data.Nat.Literals`;
* for `fromNat`, should we try to develop a theory of `Group`/`Monoid` *order*, or `*Ring` *characteristic*, as additional structure to then normalise numerals to the order/characteristic (needs `DecidableEquality`?)

gallais:
I have a bunch of results

```
  fromℕ′ : ℕ → Carrier
  fromℤ′ : ℤ → Carrier

  fromℕ′-+ : ∀ m n → fromℕ′ (m ℕ.+ n) ≈ fromℕ′ m + fromℕ′ n
  fromℕ′-∸ : ∀ {m n} → n ℕ.≤ m → fromℕ′ (m ℕ.∸ n) ≈ fromℕ′ m - fromℕ′ n
  fromℤ′‿-‿homo : ∀ i → fromℤ′ (ℤ.- i) ≈ - (fromℤ′ i)
  fromℤ′-⊖ : ∀ m n → fromℤ′ (m ℤ.⊖ n) ≈ fromℕ′ m - fromℕ′ n
  fromℤ′-+ : ∀ i j → fromℤ′ (i ℤ.+ j) ≈ fromℤ′ i + fromℤ′ j
```

Happy to open a PR if folks think that'd be useful.

[...]

Issue #581, opened 2019-01-02
Modular arithmetic


WorldSEnder:
I wanted to come up with a better implementation for primality. Basic modular arithmetic should probably be tackled first but they might go hand in hand.

MatthewDaggitt:
What would your suggestions be?

WorldSEnder:
My biggest suggestion would be implementing finite field operations for `Fin` - multiplication, addition, subtraction and division - as binary operations on `Fin`. If those exist, I'm sorry for overlooking them. 

After that, it might be beneficial to be less dependent on the representation of `Nat`. As numbers get bigger, modular arithmetic under some power of 2, say `2^1024` is - in my mind - better formulated in terms of `Bin` rather than `Nat`.

MatthewDaggitt:
> My biggest suggestion would be implementing finite field operations for Fin - multiplication, addition, subtraction and division alongside with the usual theorems - as binary operations on Fin. If those exist, I'm sorry for overlooking them.

Nope, you haven't missed them. If you'd be interested in coming up with an efficient implementation that'd be great!

> After that, it might be beneficial to be less dependent on the representation of Nat. As numbers get bigger, modular arithmetic under some power of 2, say 2^1024 is - in my mind - better formulated in terms of Bin rather than Nat.

Yup, I think @mechvel has suggested something similar in #65. It's quite a big project though as `Data.Bin` isn't very developed and probably needs to be slightly rethought at some point.

oisdk:
[I tried to do this a little in the past](https://doisinkidney.com/posts/2018-11-20-fast-verified-structures.html). I found that `Fin` didn't really work well as a modular arithmetic type. I had better luck with the following:

```agda
infix 4 _≥_
data _≥_ (m : ℕ) : ℕ → Set where
  m≥m : m ≥ m
  m≥p : ∀ {n} → m ≥ suc n → m ≥ n

Mod : ℕ → Set
Mod = ∃ ∘ _≥_
```

(The `Mod` is kind of a misnomer: numbers in base-10, for instance, would be "mod-9")

Addition can be implemented with the same efficiency as on N:

```agda
infixl 6 _+_
_+_ : ∀ {n} → Mod n → Mod n → Mod n
(_ , x) + (ys , y) = go x ys y
  where
  go : ∀ {n m} → n ≥ m → ∀ ys → n ≥ ys → Mod n
  go m≥m ys y = ys , y
  go (m≥p x) zero y = _ , x
[...]

PR #2257, opened 2024-01-15
Add `Data.Nat.Bounded`

jamesmckinna:
Prompted by a [recent discussion on the Agda Zulip](https://agda.zulipchat.com/#narrow/stream/238741-general/topic/Numeric.20proofs.20on.20fixed.20bit.20.22unsigned.20integers.22/near/412916777), and realising that we don't actually have the details of this spelt out in `Data.Fin` as such...

... and I decided to post it as a PR directly, rather than go via a separate issue first. 

Outstanding issues:
* names? ~~(eg. why can't I call the `value` field `⟦_⟧` directly?)~~
* where should this live? It even occurred to me that this might belong purely as a `README` contribution?
* does this even stand on its own feet, or should it simply be refactored as a `README.Data.Refinement` example of [`Data.Refinement`](https://github.com/agda/agda-stdlib/blob/master/src/Data/Refinement.agda)?
* I tried writing `isBounded` as a manifest field, but Agda complains that it needs the unsafe `--irrelevant-projections` flag to do this, whereas the pattern-matching version is fine? issue raised as https://github.com/agda/agda/issues/7063
* ~~the version of `fromℕ<` here as a possible improvement on [`Data.Fin.Base.fromℕ<`](https://github.com/agda/agda-stdlib/blob/3515c22f3a2c1bd2eeb5fdd504072f67aab3de4d/src/Data/Fin/Base.agda#L68-L72);~~
* DONE: proof of the isomorphism witnessed by `toFin`/`fromFin`; 
* DONE: proof of the retraction witnessed by `toℕ`/`fromℕ`;
* DONE: added support for `Number` literals; cf. #1363 , ahead of any `Structures`/`{Raw}Bundles` instances...
* DONE: arithmetic modulo n #581 , based on this design, ~~at: https://github.com/jamesmckinna/agda-stdlib/blob/modular-arithmetic/src/Data/Integer/Modulo.agda (will raise separate PR when/if this gets approval)~~ see #2292 ;
* other things that may occur to you ;-)

If approved, then I would refactor to reflect the customary `Data.X.Base|Properties|Literals|Show|...` hierarchical decomposition... DONE

Comments welcome!


src/Data/Nat/Bounded.agda
```diff
@@ -0,0 +1,69 @@
...
+    m n : ℕ
+
+------------------------------------------------------------------------
+-- Definition
+
+record BoundedNat (n : ℕ) : Set where
+  constructor ⟦_⟧<
+  field
+    value : ℕ
+    .{{bounded}} : value < n
```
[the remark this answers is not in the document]
jamesmckinna: Well, that had been my thinking underlying #1998 and (the reverted parts of) #2000 ... but indeed reimplementing this via `Data.Refinement` wouldn't introduce that extra twist. I had the ... even 'gut feeling' seems a bit strong ... sense that in suitable deployment contexts (fixed-width arithmetic; positional notation for numerals; ...) it might be the case the ambient instances would be available from the context ... but perhaps you're correct...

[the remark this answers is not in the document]
JacquesCarette: I like the idea of adding `BoundedNat` as an equivalent for `Fin`, but I think the 'experiment' of using an instance for `bounded` is premature.

Supporting evidence: in code that Conor and I were playing around with, we had exactly `BoundedNat` in our extensional universe as our definition of `Fin`, and we did have to supply the proof a non-trivial number of times. (That the proof was irrelevant without needed Agda's irrelevance machinery was what we were really after.)

src/Data/Nat/Bounded.agda
```diff
@@ -0,0 +1,69 @@
...
+-- Equality on values is propositional equality
+
+bounded≡⇒⟦⟧≡⟦⟧ : _≡_ {A = BoundedNat n} ⇒ (_≡_ on ⟦_⟧)
+bounded≡⇒⟦⟧≡⟦⟧ = ≡.cong ⟦_⟧
+
+⟦⟧≡⟦⟧⇒bounded≡ :  (_≡_ on ⟦_⟧) ⇒ _≡_ {A = BoundedNat n}
+⟦⟧≡⟦⟧⇒bounded≡ refl = refl
+
+------------------------------------------------------------------------
+-- Conversion to and from `Fin n`
```
JacquesCarette, line 58: Surprised you didn't prove the full iso?

src/Data/Nat/Bounded.agda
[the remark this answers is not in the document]
jamesmckinna: OK, so that's two of you... so either:
* I remove the `instance` bracketing (easy);
* I go all-out for `T (value <ᵇ n)` as above (more involved, but most of the machinery is already in #2000 )

I guess it's the first choice, then, with the second as a possible downstream refactoring?

src/Data/Nat/Bounded.agda
jamesmckinna: I can be very lazy... and I haven't had much experience in `stdlib` of proving isomorphisms (esp. at the `Set`/`Setoid` level) as really adding value...

... so, I'd rather leave that (to somebody else?) as a supplementary PR?

src/Data/Nat/Bounded.agda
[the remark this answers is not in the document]
jamesmckinna: NB making `bounded` an 'ordinary' (even if `irrelevant`) field vitiates somewhat my choice of `⟦_⟧<` as a constructor, which could be used 'bare', without bracketing, as a pattern constructor, in favour of `⟦_⟧<_` with the explicit bound argument, and the `value` argument position now mostly redundant... oh well. The redundancy is harmless, unless someone else has a compelling improvement on my notation?

src/Data/Nat/Bounded.agda
```diff
@@ -0,0 +1,81 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Bounded natural numbers, and their equivalence to `Fin`
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Nat.Bounded where
```
MatthewDaggitt, line 9: So to touch on your question of whether this belongs in the library or a readme file, I guess my question is what circumstances is this easier to work with than `Fin`? In general we try to avoid duplicate but equivalent definitions unless there is clear reason that sometimes one representation is easier to work with than the other...

jamesmckinna: Thanks, Matthew! I can think of two or three, some already in the initial preamble (which was really about helping someone solve a related problem of representation, where `Fin` didn't seem to fit well):
* fixed-width arithmetic (or even: its extension to variable-widths, but in a way which retains predictable/computable bounds);
* `Digit` bases for arithmetic, as a specialised instance/mode of use of the above;
* revising the 'specification' of `Data.Nat.DivMod`, to avoid the marshalling/unmarshalling required by the use of `Fin` to ensure that `mod` is suitably bounded.

Now, none of these could be said to be a deal-breaker, but I increasingly wonder whether `Fin` is better suited to its use as an *indexing* type (for `Vec`, `List`, de Bruijn representations, `Reflection`/deeply embedded syntaxes more generally, etc.), with `ℕ<` as an *arithmetic* type/`Number` instance...

Perhaps other pros (and cons!) might be lurking off-stage...

jamesmckinna: Previously...

> ... UPDATED:
> 
>     * re: 3rd point above. This might (eventually) lead to the deprecation of `Data.Nat.DivMod.WithK` (no longer any need to consider erasing proofs of `_≤″_`...?) and hence perhaps to the deprecation of `_≤″_` _altogether_, cf. [Remove all external use of `_≤″_` beyond `Data.Nat.*` #2262](https://github.com/agda/agda-stdlib/pull/2262)

This cashes out as the following `breaking` (because type-signature-changing... hence not committed to this PR) simplifications to `Data.Nat.DivMod`  and `Data.Nat.DivMod.WithK`:
```agda
record DivMod (dividend divisor : ℕ) : Set where
  constructor result
  field
    quotient  : ℕ
    remainder : ℕ< divisor
    property  : dividend ≡ ⟦ remainder ⟧ + quotient * divisor

  nonZeroDivisor : NonZero divisor
  nonZeroDivisor = nonZeroIndex remainder


infixl 7 _div_ _mod_ _divMod_

_div_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} → ℕ
_div_ = _/_

_mod_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} → ℕ< divisor
m mod n = ⟦ m % n ⟧< (m%n<n m n)

_divMod_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} →
           DivMod dividend divisor
m divMod n = result (m / n) (m mod n) $ m≡m%n+[m/n]*n m n
```
and (`WithK`)
```agda
_divMod_ : (dividend divisor : ℕ) → .{{ NonZero divisor }} →
           DivMod dividend divisor
m divMod n = result (m / n) (m mod n) $ ≡-erase $ m≡m%n+[m/n]*n m n
```

Suggestions on how to proceed, if, like me, you find these compelling.

jamesmckinna:
Previously...

> ... UPDATED:
> 
>     * re: 3rd point above. This might (eventually) lead to the deprecation of `Data.Nat.DivMod.WithK` (no longer any need to consider erasing proofs of `_≤″_`...?) and hence perhaps to the deprecation of `_≤″_` _altogether_, cf. [Remove all external use of `_≤″_` beyond `Data.Nat.*` #2262](https://github.com/agda/agda-stdlib/pull/2262)

This cashes out as the following `breaking` (because type-signature-changing... hence not committed to this PR) simplifications to `Data.Nat.DivMod`  and `Data.Nat.DivMod.WithK`:
```agda
record DivMod (dividend divisor : ℕ) : Set where
  constructor result
  field
    quotient  : ℕ
    remainder : ℕ< divisor
    property  : dividend ≡ ⟦ remainder ⟧ + quotient * divisor

  nonZeroDivisor : NonZero divisor
  nonZeroDivisor = nonZeroIndex remainder


infixl 7 _div_ _mod_ _divMod_

_div_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} → ℕ
_div_ = _/_

_mod_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} → ℕ< divisor
m mod n = ⟦ m % n ⟧< (m%n<n m n)

_divMod_ : (dividend divisor : ℕ) .{{_ : NonZero divisor}} →
           DivMod dividend divisor
m divMod n = result (m / n) (m mod n) $ m≡m%n+[m/n]*n m n
```
and (`WithK`)
```agda
_divMod_ : (dividend divisor : ℕ) → .{{ NonZero divisor }} →
           DivMod dividend divisor
m divMod n = result (m / n) (m mod n) $ ≡-erase $ m≡m%n+[m/n]*n m n
```

Suggestions on how to proceed, if, like me, you find these compelling.

src/Data/Nat/Bounded.agda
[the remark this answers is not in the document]
jamesmckinna: The addition of a `Number` instance with `Constraint m = T (m <ᵇ n)` biases towards the second alternative...
... but `stdlib` style suggests using `True (m <? n)` instead... cf. `Data.Fin.Literals`

Any comment on the trade-offs between the two, given that the latter is much more indirect than the former...?

src/Data/Nat/Bounded.agda
jamesmckinna: And to the last point: worth raising a PR to add something to `Function.Consequences`?

JacquesCarette: I don't go through `Bijective` myself, I tend to go for `Equivalence`. The latter is more categorical while the former is more set-theoretical.

src/Data/Nat/Bounded.agda
```diff
@@ -0,0 +1,112 @@
...
+
+⟦0⟧< ⟦-1⟧< : .{{ NonZero n }} → ℕ< n
+⟦0⟧<              = ⟦ 0 ⟧< >-nonZero⁻¹ _
+⟦-1⟧< {n = suc m} = ⟦ m ⟧< ℕ.n<1+n m
+
+⟦1⟧< : .{{ NonZero n }} → ℕ< n
+⟦1⟧< {n = 1}    = ⟦0⟧<
+⟦1⟧< {n = 2+ _} = ⟦ 1 ⟧< nonTrivial⇒n>1 _
+
+-- Projection from ℕ
```
JacquesCarette, line 53: This is not really a 'projection' in the type-theoretic sense of the word, which at first confused me (I guess my brain's been infected in that way). Does anyone else use `π` for this? I'm really not liking this name.

src/Data/Nat/Bounded.agda
JacquesCarette: And now that I look at the code... you actually prove `Equivalence` and don't write it down explicitly, and then proceed to "throw it away" to get `Bijective` !!! Wow.

JacquesCarette (changes requested):

src/Data/Nat/Bounded.agda
jamesmckinna: What do we call the canonical map from a thing to a quotient-thing? `q`?
We do usually call the canonical map from a sub-thing to a thing $$\iota$$, and $$\iota$$ for 'injection' and $$\pi$$ for 'projection' did make some sort of sense to me, at least...

... but for sure, not a 'projection' in the type-theoretic sense (a battle which ever rages in me between 'math' and 'type theory'... ;-))

jamesmckinna:
I think I proved `Bijective` believing that to proxy for `Isomorphism` (of `Propositional` setoids), and that to be stronger than `Equivalence`. But... suitably chided by "wow", perhaps you should school me in what I should have proved instead...

src/Data/Nat/Bounded.agda
jamesmckinna: Oh, I see now...

src/Data/Nat/Bounded.agda
JacquesCarette: Much better.

JacquesCarette (approved):

jamesmckinna:
Am now in a (better) position to prove the two lemmas relating `ℕ` and `ℕ / ~, at which point this can merge, from my point of view.

JacquesCarette (approved):
I remain happy with this.

jamesmckinna:
In the event that #2292 is *not* merged, then the `Semiring` structure of `ℕ<` should be developed as a followup PR to this one, by analogy with the `Ring` development there, so the additional arithmetic lemmas are stubs towards such eventual downstream development. But I'll *definitely* stop here now.

src/Data/Nat/Bounded/Properties.agda
```diff
@@ -23,15 +24,16 @@ open import Relation.Binary.Structures using (IsPartialEquivalence; IsEquivalenc
 open import Relation.Binary.Core using (_⇒_)
 open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality
-  using (_≡_; refl; erefl; sym; trans; cong; subst; _≗_)
+  hiding (isEquivalence; setoid)
```
JacquesCarette, line 27: going from 'using' to 'hiding', you've now pulled in a lot more of the namespace and it might come back to haunt you later.

jamesmckinna: Yes, you're right. Reverted, with the addition of `module ≡-Reasoning`.

JacquesCarette (approved):

jamesmckinna:
Aaargh... `CHANGELOG` hell. (Partly self-inflicted; hopefully this will converge soon)

src/Data/Nat/Bounded/Base.agda
```diff
@@ -0,0 +1,96 @@
...
+    m n o : ℕ
+
+------------------------------------------------------------------------
+-- Definition
+
+record ℕ< (n : ℕ) : Set where
+  constructor ⟦_⟧<_
+  field
+    ⟦_⟧ : ℕ
+    .bounded : ⟦_⟧ < n
```
gallais, line 32: I think we should use `Data.Irrelevant` because we don't have irrelevant projections
and so manipulating these records is annoying in my experience.

jamesmckinna: Well, in that case, it's probably better simply to define this as an instance of `Data.Refinement`? 
That being so, I'll reconvert to `DRAFT` and take it off the v2.2 table for the time being...

src/Data/Nat/Bounded.agda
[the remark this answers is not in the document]
JacquesCarette: I've found dealing with stuff like `T (m <ᵇ n)` to be pure evil in an open context, i.e. where one (or both) of `m` and `n` are not values. I think that design is too biased towards "is a natural number" and makes "will be a natural number" horrible.

Turns out Agda's used more as a proof assistant (unlike say, Idris), so that computation-oriented designs seem to hinder.

JacquesCarette:
So what's needed to finish this? Could this be v.2.4?

jamesmckinna:
> So what's needed to finish this? Could this be v.2.4?

I think that it could.

But there was quite a lot of discussion, IIRC, which seemed to question whether we actually *want* this in the library. Technically, it's uncoupled from any proposal for `mod n` arithmetic, but it's only *real* use would be to refactor/break the API for `DivMod`, which we haven't properly discussed at all!

I guess one thing we could do woul dbe badge it as v2.4, re-review, merge if successful, and then end up deciding later that we don't actually want it? Worse things happen... ;-)

JacquesCarette:
If it's going to be immediately orphaned, I'm much less keen to pull it in. Feels like we may as well wait for v3.0 to make sure this is the right design, including for `DivMod`.

MatthewDaggitt:
> But there was quite a lot of discussion, IIRC, which seemed to question whether we actually want this in the library. Technically, it's uncoupled from any proposal for mod n arithmetic, but it's only real use would be to refactor/break the API for DivMod, which we haven't properly discussed at all!

For the record, I'm still in this camp. We need a compelling use case to justify adding in another representation.

jamesmckinna:
@MatthewDaggitt thanks for the fedback. For me, changing the abstract interface `DivMod` is the compelling use-case.

Previously, you wrote
> So to touch on your question of whether this belongs in the library or a readme file, I guess my question is what circumstances is this easier to work with than `Fin`? In general we try to avoid duplicate but equivalent definitions unless there is clear reason that sometimes one representation is easier to work with than the other...

I think that your recent comments re: `Data.Fin.Base.inject` on #2744 , as well as your introduction of `Data.Fin.Base.lower` in #2748 in preference to using `lower₁` demonstrate how, in practice, `Fin`-based computations are unnecessarily fiddly, introduce too much choice (which specialised version of what 'easy' operation on Nat do I need here?), compared to 'vanilla' Nat-based computations using `_<_` (irrelevantly marked, or otherwise). 

Currently the `DivMod` API is 'polluted' by the introduction of `Fin`, where otherwise it would be entirely self-contained as governing/governed by, `Nat`, and so introduces unwelcome/unnecessary *concrete* coupling to what otherwise might be *abstract* wrt (any implementation of) `Nat`?

We do already have 'duplicate but equivalent' representations of `_< n` for `n : Nat`, one as a predicate, and one as the type family `Fin`. Each clearly has a role to play... and we rightly don't force `stdlib` users to have to proxy the former with the latter.

MatthewDaggitt:
While I agree with you that the interface of `DivMod` should be changed away from `Fin`, why do we need to bundle up the remainder and the proof of the remainder is less than the dividend into one data type (complete with operations etc.?). Why not just have the proof as another field in the record?

jamesmckinna:
@MatthewDaggitt Ah!... OK, fair enough. 

Looking back at `Properties`, it's clearly strongly skewed towards establishing well-behavedness of the quotient map, ie. as a precursor to #2292 ... so closing both for now. But we *should* agree a design for `mod n` arithmetic, and I hope therefore that #2729 manages to land sooner rather than later.

PR #2258, opened 2024-01-16
a better `scanr`, via `Data.List.NonEmpty`

jamesmckinna:
Avoids the [`with`-based definition](https://stackoverflow.com/questions/77821802/agda-how-to-prove-basic-properties-of-scanr-in-agda-involving-branch-reasoning) in favour of a simpler definition using (pattern-matching ;-)) `let`, with no dead-code branch... cf. also #1937 , #2123

### DIFF
--- a/src/Data/List.agda
+++ b/src/Data/List.agda
@@ -12,9 +12,7 @@
 module Data.List where
 
 import Data.List.NonEmpty.Base as List⁺
-import Data.List.Properties as List
 open import Function.Base using (_∘_)
-open import Relation.Binary.PropositionalEquality using (refl; cong₂; _≗_)
 
 ------------------------------------------------------------------------
 -- Types and basic operations
@@ -26,20 +24,17 @@ open import Data.List.Base public hiding (scanr)
 
 module _ {a b} {A : Set a} {B : Set b} (f : A → B → B) (e : B) where
 
-  open List⁺ using (List⁺; _∷_; toList)
+  open List⁺ using (toList; scanr⁺)
 
 -- definition
 
   scanr : List A → List B
-  scanr = toList ∘ go where
-    go : List A → List⁺ B
-    go []       = e ∷ []
-    go (x ∷ xs) = let y ∷ ys = go xs in f x y ∷ y ∷ ys
+  scanr = toList ∘ scanr⁺ f e
 
--- property
+{-
+-- which satisfies the following property:
 
   scanr-defn : scanr ≗ map (foldr f e) ∘ tails
-  scanr-defn []                 = refl
-  scanr-defn (x ∷ [])           = refl
-  scanr-defn (x ∷ y∷ys@(_ ∷ _)) = let eq = scanr-defn y∷ys in
-    cong₂ (λ z → f x z ∷_) (List.∷-injectiveˡ eq) eq
+  scanr-defn = Data.List.NonEmpty.Properties.toList-scanr⁺
+-}
+
--- a/src/Data/List.agda
+++ b/src/Data/List.agda
@@ -36,13 +36,11 @@ module _ {a b} {A : Set a} {B : Set b} where
     go []       = e ∷ []
     go (x ∷ xs) = let y ∷ ys = go xs in f x y ∷ y ∷ ys
 
-
 -- property
 
   scanr-defn : ∀ (f : A → B → B) (e : B) →
                scanr f e ≗ map (foldr f e) ∘ tails
   scanr-defn f e []                 = refl
   scanr-defn f e (x ∷ [])           = refl
-  scanr-defn f e (x ∷ y∷xs@(_ ∷ _)) = let eq = scanr-defn f e y∷xs in
+  scanr-defn f e (x ∷ y∷ys@(_ ∷ _)) = let eq = scanr-defn f e y∷ys in
     cong₂ (λ z → f x z ∷_) (List.∷-injectiveˡ eq) eq
-
--- a/src/Data/List.agda
+++ b/src/Data/List.agda
@@ -11,7 +11,32 @@
 
 module Data.List where
 
+import Data.List.NonEmpty as List⁺
+import Data.List.Properties as List
+open import Function.Base using (_∘_)
+open import Relation.Binary.PropositionalEquality using (refl; cong₂; _≗_)
+
 ------------------------------------------------------------------------
 -- Types and basic operations
 
-open import Data.List.Base public
+open import Data.List.Base public hiding (scanr)
+
+------------------------------------------------------------------------
+-- scanr
+
+module _ {a b} {A : Set a} {B : Set b} where
+
+-- definition
+
+  scanr  : (A → B → B) → B → List A → List B
+  scanr f e = List⁺.toList ∘ List⁺.scanr⁺ f e
+
+-- property
+
+  scanr-defn : ∀ (f : A → B → B) (e : B) →
+               scanr f e ≗ map (foldr f e) ∘ tails
+  scanr-defn f e []                 = refl
+  scanr-defn f e (x ∷ [])           = refl
+  scanr-defn f e (x ∷ y∷xs@(_ ∷ _)) = let eq = scanr-defn f e y∷xs in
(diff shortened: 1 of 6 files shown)


gallais:
Shouldn't the `scanr` in `NonEmpty`  consume a `NonEmpty`  list?
I think the improved `scanr` should probably go in `Data.List`.

jamesmckinna:
@gallais  wrote:

> Shouldn't the `scanr` in `NonEmpty` consume a `NonEmpty` list? 

I don't think so, although perhaps you could argue either way. I think the main point is that `scanr` transforms lists into non-empty lists. So maybe `scanr⁺` should take a non-empty list, but then you would get (still) get a redundant repetition of `f x y ∷ _` in both the top-level clause, and the local defn of the corresponding `go` function... I think?

> I think the improved `scanr` should probably go in `Data.List`.

That's more plausible, but then where does its defining property `scanr-defn` belong? Also there? I'd be fine with that, but it seems to go against ambient convention? I'll make that as a final change at the end, for comparison's sake.

src/Data/List.agda
```diff
@@ -11,7 +11,35 @@
...
+
+  scanr : List A → List B
+  scanr = toList ∘ go where
+    go : List A → List⁺ B
+    go []       = e ∷ []
+    go (x ∷ xs) = let y ∷ ys = go xs in f x y ∷ y ∷ ys
+
+-- property
+
+  scanr-defn : scanr ≗ map (foldr f e) ∘ tails
```
JacquesCarette, line 41: We shouldn't have properties in definition files, especially for lists.

src/Data/List.agda
```diff
@@ -11,7 +11,35 @@
...
+-- scanr
+
+module _ {a b} {A : Set a} {B : Set b} (f : A → B → B) (e : B) where
+
+  open List⁺ using (List⁺; _∷_; toList)
+
+-- definition
+
+  scanr : List A → List B
+  scanr = toList ∘ go where
```
JacquesCarette, line 34: love how this definition makes it clear that `scanr` is forgetful, i.e. drops information it learned along the way!

JacquesCarette (changes requested):

src/Data/List.agda
jamesmckinna: So this is a very irritating `dependencies` problem: 
* I took @gallais 's suggestion to lift this out of `Data.List.NonEmpty.Base` in favour of `Data.List`;
* but the 'property' then gets orphaned, and yet depends on both `NonEmpty` and `List`;
* ... and `Data.List.Properties` is both already taken, and prior to  `Data.List.NonEmpty.Properties`, and the property at hand...

I could revert various things, but not really sure how best to proceed... partly because `scanr` is already an irritating wart, but I don't want to grant it a module on its own for the sake of breaking the dependency cycle.

src/Data/List.agda
jamesmckinna: ... indeed ;-)

But to refactor it (once again), so that `go` as `scanr⁺` it (properly?!) belongs in `Data.List.NonEmpty`, and hence the `tails` property could be expressed in `Data.List.NonEmpty.Properties`, then means we *also* need to refactor `tails` (as `tails⁺`, for example)... etc. lots of knock-on ... bureaucracy. 

Suggestions as to how to "find the way out of the bottle"?

jamesmckinna:
Thanks @JacquesCarette for the feedback. Any suggestions as to how to "find the way out of the bottle"?

src/Data/List.agda
jamesmckinna: Have attempted a principled refactoring, but with the above redundancy in commentary/documentation.

JacquesCarette: I very much like the refactoring.

JacquesCarette (approved):

JacquesCarette (approved):

src/Data/List.agda
```diff
@@ -11,7 +11,30 @@
...
+-- scanr
+
+module _ {a b} {A : Set a} {B : Set b} (f : A → B → B) (e : B) where
+
+  open List⁺ using (toList; scanr⁺)
+
+-- definition
+
+  scanr : List A → List B
+  scanr = toList ∘ scanr⁺ f e
```
MatthewDaggitt, line 32: This is screaming to me that we're doing something wrong here with the dependencies. Can we discuss this evening at the stdlib implementor's meeting?

jamesmckinna: Well, this *is* the motor for this PR: AFAICT, the `Data.List.Base` defn of `scanr` arises in the (historical) form it takes *precisely* to ensure that it can exist in `Data.List.Base`, without any of the other analysis of its typing. 

But, as other commenters elsewhere note, the 'official' library definition has terrible properties (or better: terrible proofs of its properties...) because of its 'wrinkly' definition, arising from its combination of `with` *and* dead-code. 

The other operations considered in #2267 / #2269 are OK, but `scanr` seems 'special'.

jamesmckinna:
So I think we've (now) agreed to find a better home for this (and #2269 ) under `Data.List.Something`, and not in `Data.List`. Suggestions welcome!

jamesmckinna:
Aaargh: `Data.List.Kleene.*`!

JacquesCarette:
So I think that the stuff in #2269 doesn't really need to move? At least, I don't immediately see the need. So that leaves just `scanr`?

Or do `tails` and `inits` belong so strongly with `scanr` that they really ought to move at the same time?

jamesmckinna:
> So I think that the stuff in #2269 doesn't really need to move? At least, I don't immediately see the need. So that leaves just `scanr`?
> 
> Or do `tails` and `inits` belong so strongly with `scanr` that they really ought to move at the same time?

Well, good question...
* `scanr` and `tails` belong together: `Data.List.Properties.scanr-defn`
* `scanl` and `inits` belong together, ditto.

So, I could imagine that #2269 is 'redundant'/'invalid', but for the fact that taken all together, the four functions belong with one another. It's a judgment call (and one which `Data.List.Kleene` regularises by being able to define them all together via the mutuality of the datatype definitions...), so not really sure what the right thing to do is.

jamesmckinna:
An alternative, but for the cost of a horrible allocation overhead (no multiple return values on the stack!) against a benefit of being observably `NonEmpty` without going via `List.NonEmpty`, might be to use an auxiliary intermediate  `Product`:
```agda
scanr : (A → B → B) → B → List A → List B
scanr {A = A} {B = B} f e xs = let h , t = go xs in h ∷ t
  where
  go : List A → B × List B
  go [] = e , []
  go (x ∷ xs) with y , ys ← go xs = f x y , y ∷ ys
```

JacquesCarette:
That version of `scanr` would be a massive pain to prove anything about. Not only does it use a non-public helper function (whose properties are crucial), it also uses ``with` (so another depth of helper function). This feels like a Haskell version of `scanr`, not an Agda version, if you get my drift.

Taneb:
I'd really like to see this merged, but putting the implementation in `Data.List` really isn't great (as we've discussed before) - would a new module `Data.List.Scans` be an option? Along with the definitions discussed in #2269 and #2267

jamesmckinna:
Thanks @Taneb ! Sorry my recent preoccupations have meant this PR got ignored/abandoned for a while. If there were general assent towards `Data.List.Scans`, perhaps that *is* (would be) the way to go. It feels weird to have to factor them out as 'special', though!? But the definitions indeed expose... anomalies...

JacquesCarette:
I'm a huge believer in "follow the elegant path". Not always (there's lots of room for solid engineering and usability to disrupt some elegant paths), but whenever realistic. I sometimes say it as "follow the math", but by this I usually mean "the simple definitions, without unexplained / unexplainable detours". Those anomalies are a signal, and they're the kinds of signals that I love to pursue.  `scanr` has hit that quite strongly!

jamesmckinna:
If we agree to move this to `data/List.Scans`, is there a way to merge the two sets of commits from here and #2269 into a single one? I realise I am out my depth here!

jamesmckinna:
Now see #2395

PR #2103, opened 2023-09-22
Refinement types plus `Null`/`NonNull`

jamesmckinna:
EXPERIMENTAL/DRAFT: See #2092.

`Refinement` types, and their syntax, are (at present) neither here nor there, but the rest of the PR is a moderately extended exploration of one mode of use: defining the `Refinement` associated with being `non-null` for a 'typeclass' supporting a `null`/`non-null` field... including `Nat`, `List`, ...

Possible target(s): 
- [x] safe `head` function on `List'; UPDATED superseded/specialised by #2260 ?
- [ ] reimplement `Data.List.NonEmpty`?
- [ ] what else?

PR #2260, opened 2024-01-18
Add `NonNull`  to `Data.List.Relation.Unary.NonNull` on the model of `Data.Nat.Base.NonZero`

jamesmckinna:
This is a cherry-picked/revised version of #2092 / #2103 specialised to `List` only, prompted by the discussion in
[this stackoverflow question](https://stackoverflow.com/questions/77821802/how-to-prove-basic-properties-of-scanr-in-agda-involving-branch-reasoning)

Outstanding issues:
* only the specimen applications to `head` and `tail` are defined; (so also: naming)
* UPDATED ~~no properties added to `Data.List.Properties`; but~~ perhaps the constructors/destructors should move ~~there~~ to `Data.List.Relation.Unary.NonNull.Properties`;
* no (additional) examples added to `README.*`;
* no refactoring/deprecation of (any, or all) of `Data.List.NonEmpty`;
* ...

Yields the following solution to the above `stackoverflow` problem:
```agda
module ScanrNonNull (f : A → B → B) (e : B) where

  scanr-nonNull : ∀ xs → NonNull (scanr f e xs)
  scanr-nonNull [] = _
  scanr-nonNull (x ∷ xs)
    with _ ← scanr-nonNull xs
    with y ∷ ys ← scanr f e xs = _

  scanr-head : (xs : List A) → B
  scanr-head xs = nonNull-head (scanr f e xs)
    where instance _ = scanr-nonNull xs

  scanr-head-unfold : ∀ x xs → scanr-head (x ∷ xs) ≡ f x (scanr-head xs)
  scanr-head-unfold x xs
    with _ ← scanr-nonNull xs
    with y ∷ ys ← scanr f e xs = refl
```

Coupled with the redefinition of `scanr` in #2258 , this solution then simplifies further to:
```agda
  scanr-nonNull : ∀ xs → NonNull (scanr xs)
  scanr-nonNull xs = _

  scanr-head : (xs : List A) → B
  scanr-head xs = nonNull-head (scanr xs)

  scanr-head-unfold : ∀ x xs → scanr-head (x ∷ xs) ≡ f x (scanr-head xs)
  scanr-head-unfold x xs
    with _ ← scanr-nonNull xs
    with y ∷ ys ← scanr xs = refl
```


JacquesCarette (commented):
I am strongly against this addition to `Data.List.Base`. This feels like a 'bad smell' coming from non-dependently-typed programming where you don't want to say that you know your list is non-null but want to defer this to proof search instead. 

Can `NonNull` exist in `stdlib` in some module that I personally will never import? Sure!

jamesmckinna:
Ouch! OK...

... easy enough, I guess, to punt this to `Data.List.Relation.Unary.NonNull`? 

But as the PR makes clear, this began as a simple-minded emulation of the compromise we already make in `Data.Nat.Base` to admit `NonZero` and `NonTrivial` as unary predicates, and the (binary) ordering relations as well. So a 'principled' solution might be... quite drastic in its knock-on consequences, were we also to take them seriously for `Nat`. 

Here, not so much, because we don't have the same commitment to `NonNull` reasoning on `List` as we do to `NonZero`/`NonTrivial` reasoning on`Nat`, but IMHO that's a purely contingent state of affairs, ... cf. the [`scanr` PR](https://github.com/agda/agda-stdlib/pull/2258), exposing that we might want to take `NonNull` reasoning more seriously in general. And as that PR demonstrates, the type-based analysis trading off between `List` and `List⁺` is not entirely without wrinkles itself...

JacquesCarette:
That punt definitely would make me happy.

I agree that this is analogical (in design) to `NonZero` and `NonTrivial`. And indeed, we had a lot of equivalent-to-NonZero reasoning already in place. (And I agree, those two paragraphs should be duplicated in the issue.)

I feel like we have not explore the full design space of principled solutions for nonnull reasoning. To me, the conclusion that I draw from the `scanr` experiment is that its output type should be `List⁺` !

jamesmckinna:
Have resolved the merge conflict wrt `CHANGELOG`, but there doesn't seem to be sufficient enthusiasm for this, so closing. Will maybe re-open at a later date if there were to be further (positive!?) discussion of #2092 ...

PR #2261, opened 2024-01-19
fix toList-replicate's statement about vectors

amblafont:
Mea culpa! I am the author of this typo.


JacquesCarette (approved):
So weird that this worked at all!

MatthewDaggitt:
> So weird that this worked at all! 

You would have got an unsolved meta for `x` any time you use it but apart from it all type-checks okay.

PR #2013, opened 2023-07-02
Reconciling `Data.Nat.Divisibility.Core._∣_` and `Algebra.Definitions.RawMagma._∣_`

jamesmckinna:
See the discussion on issue #1919. 

~~Two~~ Three parts to this:
- [x] low-hanging fruit: added pattern synonym and consequences/uses of it to `Data.Nat.*`
- [ ] higher-hanging fruit: reconcile the two definitions of `_∣_` for the `Data.Nat.Base.*-rawMagma` instance
- [ ] knock-on changes for `Data.Integer.Divisibility.*` and  `Data.Rational.*` only if the preceding step succeeds

Proposal: merge the low-hanging fruit one now, and keep this branch, and the above checkboxes, open for a subsequent independent PR. In the meantime, a chance to think about the pro/contra proposals regarding how to reconcile the various definitions...

PR #2262, opened 2024-01-21
Remove (almost!) all external use of `_≤″_` beyond `Data.Nat.*`

jamesmckinna:
Followup to #2256 , based on https://github.com/agda/agda-stdlib/pull/2256#issuecomment-1900611675 .

Refactored `Data.Nat.Properties`:
* introduce proof `≤-proof` by analogy with `≤″-proof` (naming!);
* ~~introduce pattern synonym `≤-offset k` by analogy with `≤″-offset k`.~~
* use of `recompute` of an irrelevant proof of `m ≤ n` to define `m≤n⇒∃[o]m+o≡n` and a guarded version of monus

Added to `Data.Vec.Bounded.Base`:
* ~~`recompute` of the irrelevant proof of boundedness, in an attempt to encapsulate that (but see below);~~
* `toVec` explicitly as a projection.

Outstanding wrinkles:
* `rewrite` (changes type, no-op on underlying data) vs. `cast` (`O(n)` on underlying data?), as per second item of https://github.com/agda/agda-stdlib/pull/2256#issue-2080841774
* (possibly affected by the above) definitions somewhat sensitive to pattern-matching on the whole `record` vs. use of the (new) projections... probably a `with` abstraction problem (won't try to fix)

Outstanding occurrence of `_<″_`:
* `Data.Fin.Base.fromℕ<″` and its equivalence to `Data.Fin.Base.fromℕ<` in `Data.Fin.Base|Properties` cf. #2257


JacquesCarette (approved):

jamesmckinna:
Latest (and final!) commits remove all mention of `≤″-offset` and `<″-offset` in favour solely of `s<″s⁻¹` (in `Data.Fin.*`).

JacquesCarette (approved):

src/Data/Nat/Properties.agda
```diff
@@ -2101,31 +2101,43 @@ n≤′m+n (suc m) n = ≤′-step (n≤′m+n m n)
...
-≤″⇒≤ : _≤″_ ⇒ _≤_
-≤″⇒≤ {zero}  (≤″-offset k) = z≤n {k}
-≤″⇒≤ {suc m} (≤″-offset k) = s≤s (≤″⇒≤ (≤″-offset k))
+≤-proof : m ≤ n → ∃ λ k → m + k ≡ n
+≤-proof le = _ , ≤″-proof (≤⇒≤″ le)
 
-≤⇒≤″ : _≤_ ⇒ _≤″_
-≤⇒≤″ = less-than-or-equal ∘ m+[n∸m]≡n
+pattern ≤-offset k = k , refl
+pattern <-offset k = ≤-offset k
```
MatthewDaggitt, line 2126: We don't traditionally introduce pattern synonyms in the `Properties` files. If we want to keep them, they should live in `Data.Nat.Base`... 

However, I might vote for removing them. The presence of `refl` in a proof is usually quite indicative of what's going on so we'd be losing that. Also without any type annotations on the pattern synonyms, even if you look at them, it's not quite clear what they are constructing....

src/Data/Nat/Properties.agda
```diff
@@ -2101,31 +2101,43 @@ n≤′m+n (suc m) n = ≤′-step (n≤′m+n m n)
...
+≤″-proof : (le : m ≤″ n) → let less-than-or-equal {k} _ = le in m + k ≡ n
 ≤″-proof (less-than-or-equal prf) = prf
 
--- equivalence to _≤_
+-- yielding corresponding proof and `pattern` aliases for _≤″_/_<_
 
-≤″⇒≤ : _≤″_ ⇒ _≤_
-≤″⇒≤ {zero}  (≤″-offset k) = z≤n {k}
-≤″⇒≤ {suc m} (≤″-offset k) = s≤s (≤″⇒≤ (≤″-offset k))
+≤-proof : m ≤ n → ∃ λ k → m + k ≡ n
```
MatthewDaggitt, line 2122: The name for `≤″-proof` was chosen because it extracted the proof component from the record. However, that's not the case here. Can we come up with a better/more standard name for this lemma?  Perhaps `m≤n⇒m+o≡n`?

MatthewDaggitt (changes requested):
Looks great! Two minor comments...

src/Data/Nat/Properties.agda
jamesmckinna: OK, two or three things:
* `style-guide` addition: pattern synonyms should be in `Base` not `Properties
* naming, as with your comment below on `≤-proof` below; some of the changes here are in some sense knock-on changes for the work previously done on trying to constrain/eliminate use of `_≤″_`, including that relation's pattern synonyms
* the use of `refl` in specialising a constructor was a sufficiently common idiom that it seemed worthwhile to reify it as a synonym. But you're right in that perhaps that's not necessary... in which case perhaps also `divides-refl` of `Data.Nat.Divisibility.Core` (introduced in #2013 ) should be removed?

src/Data/Nat/Properties.agda
jamesmckinna: Agree, it's a better name. does it need an `∃` in it? Eg as `m≤n⇒∃[o]m+o≡n`?

src/Data/Nat/Properties.agda
jamesmckinna: > However, I might vote for removing them. The presence of `refl` in a proof is usually quite indicative of what's going on so we'd be losing that. Also without any type annotations on the pattern synonyms, even if you look at them, it's not quite clear what they are constructing....

Most recent commit removes the new synonyms, but not (yet!)  the `Data.Nat.Base.≤″-offset` one?

src/Data/Nat/Properties.agda
MatthewDaggitt: Yup, good idea.

src/Data/Nat/Properties.agda
MatthewDaggitt: > But you're right in that perhaps that's not necessary... in which case perhaps also divides-refl of Data.Nat.Divisibility.Core (introduced in https://github.com/agda/agda-stdlib/pull/2013 ) should be removed?

I think `divides-refl` is slightly less problematic, because at least it has `refl` in the name...

> Most recent commit removes the new synonyms, but not (yet!) the Data.Nat.Base.≤″-offset one?

Hmm yes, they should probably be deprecated as well, (as one of the main points of this PR was to remove them right?)

jamesmckinna: The *main* point is to (almost!) reach the end of a line where any *use* of `_≤″_` is removed entirely... while leaving the definition, and its equivalence with `_≤_` in `Data.Nat.*` as a legacy fossil relic.

But I couldn't quite get rid of it entirely: `Data.Fin.Base.from≤″` uses it (but I hope that a merge of #2257 and then the `breaking` change to `Data.Nat.DivMod{.WithK}` discussed there would deliver the final final blow...)

jamesmckinna: Actually,  I think I was too hasty in responding to your call to remove the new pattern synonyms, precisely *because* they encapsulate the ability to compute the offset `k` from a proof of `m ≤ n` *without having to invoke* either `_≤″_` *or* its own pattern synonyms `less-than-or-equal` or `≤″-offset` outside `Data.Nat.Base` or `Data.Nat.Properties`... with the latter only being necessary on the model of the debate about whether these (proxy) inversions belong in `base` or `Properties`. 

More thought required!

src/Data/Nat/Properties.agda
Taneb: Where would we now use this over `m+[n∸m]≡n : m ≤ n → m + (n ∸ m) ≡ n`, possibly in combination with `-,_`?

jamesmckinna: Fair question. But I think the uses in `Data.Vec.Bounded.Base` still need to match on the equality being `refl`... but perhaps those, too are eliminable. Currently, I'm weeding out *all* the pattern synonyms I introduced for `_≤″_`... they really are all redundant now (and `less-than-or-equal` had only been retained previously for backwards compatibility, but now it can go, I think?)

src/Data/Nat/Properties.agda
```diff
@@ -2177,8 +2186,8 @@ _>″?_ = flip _<″?_
 ------------------------------------------------------------------------
 
 ≤‴⇒≤″ : ∀{m n} → m ≤‴ n → m ≤″ n
-≤‴⇒≤″ {m = m} ≤‴-refl       = less-than-or-equal {k = 0} (+-identityʳ m)
-≤‴⇒≤″ {m = m} (≤‴-step m≤n) = less-than-or-equal (trans (+-suc m _) (≤″-proof (≤‴⇒≤″ m≤n)))
+≤‴⇒≤″ {m = m} ≤‴-refl       = 0 , (+-identityʳ m)
+≤‴⇒≤″ {m = m} (≤‴-step m≤n) = _ , trans (+-suc m _) (≤″-proof (≤‴⇒≤″ m≤n))
```
Taneb, line 2190: The parentheses on the RHS of these lines are now redundant

src/Data/Nat/Properties.agda
jamesmckinna: Smilarly to @Taneb 's comment below, now almost all uses of the lemma removed in favour of direct pattern matching, a 'guarded' definition of monus, and indeed finally along with deprecation of the pattern synonyms... and inlining the proof of `s<″s⁻¹` in the definition of `Data.Fin.Base.from≤″`.

jamesmckinna:
This has ended up going a bit further than planned, but for the better I think!

src/Data/Nat/Properties.agda
jamesmckinna: Inlining wasn't possible, because of the need to prove lemmas about `from≤″` in `Data.Fin.Properties` but we're nearly all the way there!

src/Data/Nat/Properties.agda
jamesmckinna: I think that you might be able to refactor `Data.Vec.Bounded.Base.{padRight|padLeft|padBoth}` further, but I think let's in any case stop here for the time being?

jamesmckinna:
@MatthewDaggitt I think that if you're happy with the latest round(s) of revisions, I'd be happy to stop here ;-)

JacquesCarette (approved):
Reviewed after the many changes since I had last reviewed - still looks good!

@MatthewDaggitt still needs to re-review as he had explicit things he wanted changed (but I do think they have been).

jamesmckinna:
I'll look at the merge conflicts today...

jamesmckinna:
Think I hit 'close' by mistake!?

src/Data/Nat/Base.agda
```diff
@@ -429,5 +419,26 @@ compare (suc m) (suc n) with compare m n
...
 -- not guaranteed.
 
--- Version 2.0
+-- Version 2.1
+
+-- Smart constructors of _≤″_ and _<″_
+pattern less-than-or-equal {k} eq = k , eq
+{-# WARNING_ON_USAGE less-than-or-equal
+"Warning: less-than-or-equal was deprecated in v2.1. Please match directly on proofs of ≤″ using constructor Algebra.RawMagma._∣ˡ_._,_ instead. "
+#-}
```
MatthewDaggitt, line 428: Should refer to `Algebra.Definitions.Magma....`?

src/Data/Nat/Properties.agda
```diff
@@ -2106,31 +2107,48 @@ n≤′m+n (suc m) n = ≤′-step (n≤′m+n m n)
...
-≤″⇒≤ {suc m} (≤″-offset k) = s≤s (≤″⇒≤ (≤″-offset k))
+m≤n⇒∃[o]m+o≡n : .(m ≤ n) → ∃ λ k → m + k ≡ n
+m≤n⇒∃[o]m+o≡n m≤n = _ , m+[n∸m]≡n (recompute (_ ≤? _) m≤n)
 
-≤⇒≤″ : _≤_ ⇒ _≤″_
-≤⇒≤″ = less-than-or-equal ∘ m+[n∸m]≡n
+-- and a 'guarded' version of monus
+
+guarded-∸ : ∀ n m → .(m ≤ n) → ℕ
+guarded-∸ n m m≤n = let k , _ = m≤n⇒∃[o]m+o≡n m≤n in k
```
MatthewDaggitt, line 2134: I dislike having this in `Properties` as an exposed definition, as it's not a property. Can we inline it in the proof below as that's the only place where it is used?

MatthewDaggitt (changes requested):

src/Data/Nat/Properties.agda
jamesmckinna: OK, fine!

jamesmckinna:
UPDATE following our meeting:
* fixed deprecation messages
* inlined guarded monus operation

Separately: will lift `DivMod` record definition out of `Data.Nat.DivMod` as a separate v2.2 issue/PR, ahead of eventual deprecation of this relation altogether (TBC/TBD)

CHANGELOG.md
```diff
@@ -543,14 +552,18 @@ Additions to existing modules
   nonZeroDivisor : DivMod dividend divisor → NonZero divisor
   ```
 
-* Added new proofs in `Data.Nat.Properties`:
+* Added new proofs and 'guarded' version of `_∸_` in `Data.Nat.Properties`:
```
MatthewDaggitt, line 555: I think this comment both here and in the main file can be removed?

MatthewDaggitt (approved):
Subject to altering those comments, this looks good.

CHANGELOG.md
jamesmckinna: Or even (Fairbairn again?) remove altogether?
If we don't have it as a definition, and the proof of this property is definitionally `refl`, maybe it's actually redundant?

Issue #2255, opened 2024-01-13
DRY when defining concrete instances of `RawX` bundles


jamesmckinna:
This is an offshoot of thinking about both #2252 and #2254 (and perhaps requires precise resolution of the latter):
why, in eg. [`Data.Nat.Base`](https://github.com/agda/agda-stdlib/blob/3515c22f3a2c1bd2eeb5fdd504072f67aab3de4d/src/Data/Nat/Base.agda#L172-L220), do we write
```agda
-- Raw bundles

+-rawMagma : RawMagma 0ℓ 0ℓ
+-rawMagma = record
  { _≈_ = _≡_
  ; _∙_ = _+_
  }

+-0-rawMonoid : RawMonoid 0ℓ 0ℓ
+-0-rawMonoid = record
  { _≈_ = _≡_
  ; _∙_ = _+_
  ; ε   = 0
  }

*-rawMagma : RawMagma 0ℓ 0ℓ
*-rawMagma = record
  { _≈_ = _≡_
  ; _∙_ = _*_
  }

*-1-rawMonoid : RawMonoid 0ℓ 0ℓ
*-1-rawMonoid = record
  { _≈_ = _≡_
  ; _∙_ = _*_
  ; ε = 1
  }

+-*-rawNearSemiring : RawNearSemiring 0ℓ 0ℓ
+-*-rawNearSemiring = record
  { Carrier = _
  ; _≈_ = _≡_
  ; _+_ = _+_
  ; _*_ = _*_
  ; 0# = 0
  }

+-*-rawSemiring : RawSemiring 0ℓ 0ℓ
+-*-rawSemiring = record
  { Carrier = _
  ; _≈_ = _≡_
  ; _+_ = _+_
  ; _*_ = _*_
  ; 0# = 0
  ; 1# = 1
  }
```
when it might be simpler/better/DRY to write instead:
```agda
+-*-rawSemiring : RawSemiring 0ℓ 0ℓ
+-*-rawSemiring = record
  { Carrier = _
  ; _≈_ = _≡_
  ; _+_ = _+_
  ; _*_ = _*_
  ; 0# = 0
  ; 1# = 1
  }

open RawSemiring +-*-rawSemiring public
   renaming (+-rawMonoid to +-0-rawMonoid; ... ; rawNearSemiring to +-*-rawNearSemiring)
```

MatthewDaggitt:
[...]

PR #2263, opened 2024-01-23
Raw modules bundles

### LINKED ISSUE #1828: Should Modules have raw bundles?

Taneb:
Closes #1828

Turns out that the raw module bundles only need to be indexed on `Set` rather than `Ring` or `RawRing`!


src/Algebra/Module/Bundles.agda
```diff
@@ -80,8 +77,17 @@ record LeftSemimodule (semiring : Semiring r ℓr) m ℓm
...
+    ; _≉_ to _≉ᴹ_
     )
 
+  rawLeftSemimodule : RawLeftSemimodule Carrier m ℓm
+  rawLeftSemimodule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; 0ᴹ = 0ᴹ
+    }
```
jamesmckinna, line 89: I would love it if such 'substructure' definitions were available without having to be written out in full, cf. #2255 ... eg here, the `+ᴹ-rawMonoid` structure is already available, only the additional `_*ₗ_` field needs defining, but Agda makes such `record` re-combination a bit crunchier than I would like:
```agda
rawLeftSemimodule = record { M ; _*ₗ_ = _*ₗ_ } where module M = RawMonoid +ᴹ-rawMonoid
```
is a bit of a mouthful, but still more 'informative' than the 'flattened' version?

src/Algebra/Module/Bundles.agda
```diff
@@ -106,14 +112,23 @@ record LeftModule (ring : Ring r ℓr) m ℓm : Set (r ⊔ ℓr ⊔ suc (m ⊔ 
...
     using () renaming (group to +ᴹ-group; rawGroup to +ᴹ-rawGroup)
 
+  rawLeftModule : RawLeftModule Carrier m ℓm
+  rawLeftModule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; 0ᴹ = 0ᴹ
+    ; -ᴹ_ = -ᴹ_
+    }
```
jamesmckinna, line 130: Ditto., et. seq.

src/Algebra/Module/Bundles.agda
```diff
@@ -152,8 +163,17 @@ record RightSemimodule (semiring : Semiring r ℓr) m ℓm
...
     ; rawMonoid to +ᴹ-rawMonoid
+    ; _≉_ to _≉ᴹ_
     )
 
+  rawRightSemimodule : RawRightSemimodule Carrier m ℓm
+  rawRightSemimodule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ᵣ_ = _*ᵣ_
+    ; 0ᴹ = 0ᴹ
```
jamesmckinna, line 174: Ditto.

src/Algebra/Module/Bundles.agda
```diff
@@ -178,14 +198,23 @@ record RightModule (ring : Ring r ℓr) m ℓm : Set (r ⊔ ℓr ⊔ suc (m ⊔
...
   open AbelianGroup +ᴹ-abelianGroup public
     using () renaming (group to +ᴹ-group; rawGroup to +ᴹ-rawGroup)
 
+  rawRightModule : RawRightModule Carrier m ℓm
+  rawRightModule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ᵣ_ = _*ᵣ_
+    ; 0ᴹ = 0ᴹ
+    ; -ᴹ_ = -ᴹ_
```
jamesmckinna, line 215: Ditto.

src/Algebra/Module/Bundles.agda
```diff
@@ -220,7 +249,20 @@ record Bisemimodule (R-semiring : Semiring r ℓr) (S-semiring : Semiring s ℓs
...
     using ( +ᴹ-commutativeMonoid; +ᴹ-monoid; +ᴹ-semigroup; +ᴹ-magma; +ᴹ-rawMagma
-          ; +ᴹ-rawMonoid; _≉ᴹ_)
+          ; +ᴹ-rawMonoid; rawLeftSemimodule; _≉ᴹ_)
+
+  open RightSemimodule rightSemimodule public
+    using ( rawRightSemimodule )
+
+  rawBisemimodule : RawBisemimodule R.Carrier S.Carrier m ℓm
+  rawBisemimodule = record
+    { Carrierᴹ = Carrierᴹ
```
jamesmckinna, line 259: Is this field not redundant?

src/Algebra/Module/Bundles.agda
```diff
@@ -220,7 +249,20 @@ record Bisemimodule (R-semiring : Semiring r ℓr) (S-semiring : Semiring s ℓs
...
+    ; _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; _*ᵣ_ = _*ᵣ_
+    ; 0ᴹ = 0ᴹ
```
jamesmckinna, line 264: See above... ditto.

src/Algebra/Module/Bundles.agda
```diff
@@ -254,13 +296,26 @@ record Bimodule (R-ring : Ring r ℓr) (S-ring : Ring s ℓs) m ℓm
...
+    using (leftSemimodule; rightSemimodule; rawBisemimodule)
+
+  rawBimodule : RawBimodule R.Carrier S.Carrier m ℓm
+  rawBimodule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; _*ᵣ_ = _*ᵣ_
+    ; 0ᴹ = 0ᴹ
+    ; -ᴹ_ = -ᴹ_
```
jamesmckinna, line 317: ...

src/Algebra/Module/Bundles/Raw.agda
```diff
@@ -0,0 +1,228 @@
...
+    _+ᴹ_ : Op₂ Carrierᴹ
+    _*ₗ_ : Opₗ R Carrierᴹ
+    0ᴹ : Carrierᴹ
+
+  +ᴹ-rawMonoid : RawMonoid m ℓm
+  +ᴹ-rawMonoid = record
+    { _≈_ = _≈ᴹ_
+    ; _∙_ = _+ᴹ_
+    ; ε = 0ᴹ
+    }
```
jamesmckinna, line 43: But this is unavoidable?

src/Algebra/Module/Bundles/Raw.agda
```diff
@@ -0,0 +1,228 @@
...
+    0ᴹ : Carrierᴹ
+    -ᴹ_ : Op₁ Carrierᴹ
+
+  rawLeftSemimodule : RawLeftSemimodule R m ℓm
+  rawLeftSemimodule = record
+    { _≈ᴹ_ = _≈ᴹ_
+    ; _+ᴹ_ = _+ᴹ_
+    ; _*ₗ_ = _*ₗ_
+    ; 0ᴹ = 0ᴹ
+    }
```
jamesmckinna, line 69: Etc.

jamesmckinna (approved):
This all looks great as-is, but if you can find a nice way to streamline the definition of the `Raw*` substructures, that would be great!

Similarly, long ago in advance of this issue being closed/a PR such as this being raised, I had left [a placeholder](https://github.com/agda/agda-stdlib/blob/bde655f8aa742ea82aa2c33e732353509d7dd397/src/Algebra/Module/Construct/Zero.agda#L40-L46) (commit https://github.com/agda/agda-stdlib/pull/1902/commits/5fbceb2151c566f7ad5f52ff7cda43f9c3f22543) in `Algebra.Module.Construct.Zero` for defining the `Raw*` substructures in-place for the `Zero` instance of `*Module`... 

... can this be tidied up now, or would you rather I did that after this is merged?

src/Algebra/Module/Bundles.agda
Taneb: Would that work in this case? `RawMonoid` has different field names to `RawLeftSemimodule`...

src/Algebra/Module/Bundles.agda
Taneb: Yes. My development process is write the type signature, write "name = ?", load file, press C-c C-r in the hole to get a record skeleton, fill in all the holes, delete the redundant fields. Sometimes I forget the last step.

src/Algebra/Module/Bundles.agda
jamesmckinna: Ach! I think you're right. Maddening! 
Having to add the `renaming` directives would vitiate the whole suggestion. Grrr.

src/Algebra/Module/Bundles/Raw.agda
jamesmckinna: Sigh..

src/Algebra/Module/Construct/Zero.agda
```diff
@@ -38,11 +38,32 @@ module ℤero where
 open ℤero
```
jamesmckinna, line 38: Is this directive now also redundant?

Taneb: I'm not sure it wasn't redundant to begin with, but either way I'll remove it

src/Algebra/Module/Construct/TensorUnit.agda
```diff
@@ -19,6 +19,62 @@ private
...
+rawBimodule {R = R} = record
+  { RawBisemimodule (rawBisemimodule {R = rawSemiring})
+  ; -ᴹ_ = -_
+  } where open RawRing R
+
+
+rawSemimodule : {R : RawSemiring c ℓ} → RawSemimodule _ c ℓ
+rawSemimodule {R = R} = rawBisemimodule {R = R}
+
+rawModule : {R : RawRing c ℓ} → RawModule _ c ℓ
```
jamesmckinna, line 72: Any particular reason (here, and throughout above) to omit `R` as the domain of coefficients in the return type? Could that simplify the subsequent definition body (probably not!?)

Taneb: It's because the parameter isn't `R` but `R.Carrier`, and that's a bother to extract

jamesmckinna:
Looking even better now. Many thanks for all the additional commits!
(And: still a shame about all the repetition. A possible refactoring for another time?)

src/Algebra/Module/Construct/Zero.agda
jamesmckinna: (and I agree: it was redundant before! My mistake, then as now!)

src/Algebra/Module/Construct/TensorUnit.agda
jamesmckinna: D'oh! Apologies once again...

CHANGELOG.md
```diff
@@ -33,9 +33,49 @@ Deprecated names
...
 -----------
 
+* `Algebra.Module.Bundles.Raw`: raw bundles for module-like algebraic structures
+
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Module.Bundles`, raw bundles are re-exported and the bundles expose their raw counterparts.
+
+* In `Algebra.Module.Cosntruct.DirectProduct`:
```
jamesmckinna, line 43: typo: `Construct`

Taneb: Good catch!

jamesmckinna:
:tada: If there are no further comments by CoB tomorrow, I'll hit `merge`...

JacquesCarette (approved):
This is both great and massively screams for automation. No human should have had to type all of that.

jamesmckinna:
Sorry for not merging when I had said I would. 
*very* glad that @JacquesCarette was able to offer a second pair of eyes, though`!
(and looking forward to the 'smart(er)' automation PR downstream... ;-))

Issue #2252, opened 2024-01-11
Reconciling the indices of `IsX` with those of the corresponding `RawX`


jamesmckinna:
Looking again at `README.Design.Hierarchies`, `Algebra.{Module.}Structures` and `Algebra.{Module.}Bundles.Raw`, I can't help wondering at the mismatch in indices vs. fieldnames to the various `record`s, with `RawX` (although this may simply be an artefact of parametrisation of the `Algebra.Structures` module), but my main question is:
* why are the `IsX` records not indexed over a *single* underlying `RawX` parameter? (this may be historical...)
* should they be? (ie. should we at least reconsider revisiting this design...)
Apologies if my ignorance of the history of, and discussion on, the associated issues/PRs obscures what otherwise may be glaringly obvious to those more expert than me!

JacquesCarette:
This is yet another form of the eternal bundled / unbundled question. The correct answer is: don't choose, they are equivalent. Pragmatically, that remains a non-answer because, in Agda, this equivalence can only be witnessed tediously on a case-by-case basis.

Unfortunately, I have no idea what the actual ergonomics of each choice is. This would need a great big (costly) experiment, and it doesn't seem to me that we can decide this just as a thought experiment. 

In theory, I personally like to group things together more. But that's at the "gut feel" level rather than at the "let's make this the design NOW" level.

jamesmckinna:
Well put, and so I guess I defer to your wise insight. Sigh...

jamesmckinna:
That said, if there were an agency willing to fund the costly experiment, ...

MatthewDaggitt:
I've also wondered about this from time to time, and I think it's mainly a swings and roundabouts issues. As you say, one of advantages is that you can import `Structures` while fixing the equality. This a feature that we use relatively widely in the library. 

I agree with @JacquesCarette's points as well that this may be a language design question.

If no one objects, then I might add this to the `hypothetical-rewrite` milestone.

JacquesCarette:
Well look at that, a `hypothetical-rewrite` milestone - very nice.

Hmm, I'd be tempted to start a page on the Wiki also collecting language changes that we library writers would really like to see.

jamesmckinna:
[...]

PR #2264, opened 2024-01-24
[fixes #1363] Add `Algebra.Literals`

### LINKED ISSUE #1363: Literals for any ring?

jamesmckinna:
Possible fix for #1363 .

The design, at present, is pretty convoluted (I felt as though I were writing SML `functor`s again!):
* based on #2252 , defines `IsPointedMonoid` as an extension of `RawMonoid` by an abstract point, in `Algebra.Structures.Pointed`;
* defines abstract `Literals` for such structures in-place, using `Algebra.Definitions.RawMonoid._×_`;
* defines their associated `Number` instance in `Algebra.Structures.Literals`;
* defines `PointedMonoid` bundles, in `Algebra.Bundles.Pointed`, and their associated `Number` instance in `Algebra.Bundles.Literals`;
* constructs an instance from a `SemiringWithoutAnnihilatingZero` using `1#` as the point;
* re-exports the `Number` instance for such from `Algebra.Literals`.

I would really welcome a review, with hints as to 'sensible' refactoring, if any.

### DIFF
--- a/src/Algebra/Bundles/Literals.agda
+++ b/src/Algebra/Bundles/Literals.agda
@@ -1,0 +1,19 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Algebra Literals, from a PointedMonoid bundle
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles.Pointed
+
+module Algebra.Bundles.Literals
+  {c ℓ} (pointedMonoid : PointedMonoid c ℓ)
+  where
+
+open PointedMonoid pointedMonoid
+
+-- Re-export `Number` instance defined in Algebra.Structures.Literals
+
+open import Algebra.Structures.Literals isPointedMonoid public using (number)
--- a/src/Algebra/Bundles/Pointed.agda
+++ b/src/Algebra/Bundles/Pointed.agda
@@ -1,0 +1,50 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- `Pointed` intermediate between `Monoid` and `SemiringWithoutAnnihilatingZero`
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles
+open import Algebra.Bundles.Raw
+open import Algebra.Core
+open import Algebra.Structures.Pointed as Pointed using (IsPointedMonoid)
+import Algebra.Properties.Monoid.Mult as Mult
+open import Data.Nat.Base as ℕ using (ℕ)
+open import Data.Unit.Base
+open import Level using (Level; suc; _⊔_; Lift)
+open import Relation.Binary.Core using (Rel)
+
+module Algebra.Bundles.Pointed where
+
+private
+  variable
+    c ℓ : Level
+
+------------------------------------------------------------------------
+-- Bundles with 1 binary operation & 2 elements
+------------------------------------------------------------------------
+
+record PointedMonoid c ℓ : Set (suc (c ⊔ ℓ)) where
+  field
+    rawMonoid  : RawMonoid c ℓ
+  open RawMonoid rawMonoid using (Carrier)
+  field
+    •      : Carrier
+    isPointedMonoid : IsPointedMonoid rawMonoid •
+
+  open IsPointedMonoid isPointedMonoid public
+
+------------------------------------------------------------------------
+-- instance from any SemiringWithoutAnnihilatingZero
+
+pointedMonoid : SemiringWithoutAnnihilatingZero c ℓ → PointedMonoid c ℓ
+pointedMonoid semiringWithoutAnnihilatingZero
+  = record { isPointedMonoid = isPointedMonoid }
+  where
+  open SemiringWithoutAnnihilatingZero semiringWithoutAnnihilatingZero
+    using (1#; +-rawMonoid; +-isMonoid)
+  isPointedMonoid : IsPointedMonoid +-rawMonoid 1#
+  isPointedMonoid = record { isMonoid = +-isMonoid }
+
--- a/src/Algebra/Structures/Pointed.agda
+++ b/src/Algebra/Structures/Pointed.agda
@@ -1,0 +1,49 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Defines `IsPointedMonoid`
+-- intermediate between `Monoid` and `SemiringWithoutAnnihilatingZero`
+--
+-- By contrast with the rest of `Algebra.Structures`, this is modelled
+-- on an underlying `RawMonoid`, rather than a 'flattened' such signature
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles.Raw
+import Algebra.Structures as Structures
+open import Data.Nat.Base as ℕ using (ℕ)
+open import Data.Unit.Base
+open import Level using (Level; _⊔_; Lift)
+
+import Algebra.Definitions.RawMonoid as Definitions
+
+module Algebra.Structures.Pointed where
+
+private
+  variable
+    c ℓ : Level
+
+
+------------------------------------------------------------------------
+-- Structures with 1 binary operation & 2 elements
+------------------------------------------------------------------------
+
+record IsPointedMonoid
+  (rawMonoid : RawMonoid c ℓ)
+  (• : RawMonoid.Carrier rawMonoid) : Set (c ⊔ ℓ)
+  where
+  open RawMonoid rawMonoid public
+  open Structures _≈_
+
+  field
+    isMonoid : IsMonoid _∙_ ε
+
+  _×• : ℕ → Carrier
+  n ×• = n × • where open Definitions rawMonoid
+
+  module Literals where
+    Constraint : ℕ → Set c
+    Constraint _ = Lift c ⊤
+    fromNat : ∀ n → {{Constraint n}} → Carrier
+    fromNat n = n ×•
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,17 @@ Deprecated names
 New modules
 -----------
 
+* Adding a distinguished point `•` (\bu2) to `Monoid`, and using it to define
+  abstract 'literals' for any `PointedMonoid`, with the intended mode-of-use
+  being to instantiate the point with `1#` from any (semi)ring-like structure:
+  ```agda
+  module Algebra.Bundles.Literals
+  module Algebra.Bundles.Pointed
+  module Algebra.Structures.Literals
+  module Algebra.Structures.Pointed
+  module Algebra.Literals
+  ```
+
 Additions to existing modules
 -----------------------------
 
--- a/src/Algebra/Literals.agda
+++ b/src/Algebra/Literals.agda
@@ -1,0 +1,20 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Algebra Literals, from a SemiringWithoutAnnihilatingZero
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (SemiringWithoutAnnihilatingZero)
+
+module Algebra.Literals {c ℓ}
+  (semiringWithoutAnnihilatingZero : SemiringWithoutAnnihilatingZero c ℓ)
+  where
+
+open import Algebra.Bundles.Pointed
+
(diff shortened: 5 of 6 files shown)


src/Algebra/Bundles/Pointed.agda
```diff
@@ -0,0 +1,50 @@
...
+------------------------------------------------------------------------
+-- Bundles with 1 binary operation & 2 elements
+------------------------------------------------------------------------
+
+record PointedMonoid c ℓ : Set (suc (c ⊔ ℓ)) where
+  field
+    rawMonoid  : RawMonoid c ℓ
+  open RawMonoid rawMonoid using (Carrier)
+  field
+    •      : Carrier
```
JacquesCarette, line 34: The comment "1 binary operation & 2 elements" is correct, but the rest of the names (`Pointed`, etc) give the wrong impression of having a single distinguished element. `BiPointed` ?

JacquesCarette (changes requested):
I only understood the point of this once I read through all the backlog, including #1363.

So I think the smallest algebra that has `Number` in it is "pointed unar" (single point, single unary operation). BiPointedMonoid is a really weird thing! (I know it is called PointedMonoid, but to me, Pointed means 1 point, not 2).

jamesmckinna:
@JacquesCarette you wrote

> I only understood the point of this once I read through all the backlog, including #1363.

Apologies for not providing enough context! As one if the commenters in the original issue, I'd imagined (magical thinking) that you'd still be across the issue. 

It might have been easier to follow the original suggestion and simply use `Semiring...`, and structure the development around having the literal numerals arise as a `Semiring...` homomorphism, but I'd also been trying to follow the library style of introducing operations as minimally-committed as possible, and then building properties on top of that (that gloss may reveal that I have misunderstood ...). 

I'd reached the present PR in that spirit: `1#` isn't special as a point, until you start to try build up towards the homomorphism property (most of the pieces of which are in fact already present in various places), and for numerals, that additional structure isn't needed. 

Identifying `(Bi)PointedMonoid` is the surprising outcome of that journey.

src/Algebra/Bundles/Pointed.agda
jamesmckinna: Much to think about here. Ever since my abortive `Pointed` draft PR, I've been thinking about `PointedX' (and how to formalise the map from `X` in a 'signature polymorphic' way) in general...

The two points in `PointedMonoid` play distinct roles, so I'm unhappy to identify them as '2 points', because one is distinguished as an identity, while the other is an unconstrained extension of the signature... 

... `BiPointedSet`/`BiPointedMagma` do make sense, especially as `Raw` bundles, but I've always been resistant to universal algebra ideas in this regard precisely because not all 'freely constructed' such signatures make sense in the absence of additional structure...? 

So from my POV, I'd want to keep things more cleanly separated...

JacquesCarette:
Facing the fact that there's only so much that I can keep in active memory all the time...

My current feeling is indeed that `PointedUnary` (whose initial algebra is the carrier for`Nat`) is the originator of much of the structure, which does get pushed around via `MonoidAction` / `NNO`.

src/Algebra/Bundles/Pointed.agda
JacquesCarette: Nitpick: they do "make sense", but they are not necessarily *sufficiently useful*. Phrased that way, I'd say that our current state of knowledge makes this true. [Usefulness is a temporal quality, not absolute.]

JacquesCarette: The next question is: that extra point, is it meant to be preserved by homomorphisms or not? If not, then rather than `Pointed` what you're describing is `Inhabited` (I'm writing that up separately, i.e. the observation that the category whose objects are pairs of a type and a point of that type, but whose homomorphisms don't preserve that point, are exactly the inhabited types, without needing to have a notion of truncation).

Of course, a `Monoid` is always inhabited -- the right name here would be `NonTrivialMonoid` although that's tricky as the one-point monoid is a model as you don't actually want to assert that the extra point is not unit!  Best you might be able to do is to ask that the obvious homomorphism induced by 'swap' not be homotopic to the identity.

jamesmckinna: Also an excellent observation (although I'm still insufficently homotopically-inclined to arbitrate on the last point). I think that points should be preserved (esp. in this setting, because `_+ pt` is the unary operation at hand, and we want it as (part of) an NNO-instance to be preserved?

jamesmckinna: And `NonTrivialMonoid`...: does that exclude the case `0# = 1#`? I wouldn't want to insist on that...?

jamesmckinna:
`PointedUnary`, or `NNO` as the 'official' name for this? (I lean towards the latter, for historical/mathematical/categorical reasons; others may prefer the former for consistency/uniformity reasons with the rest fo the library...)

JacquesCarette:
Is it actually an `NNO` though? Weak `NNO` even? The issue is that this is just that data part of `NNO` and none of the properties.

So it's `rawNNO`, in a sense. Which is rather bleh. (I'm not a big fan of `PointedUnary` btw, but at least you don't expect any laws to hold.)

jamesmckinna:
> Is it actually an `NNO` though? Weak `NNO` even? The issue is that this is just that data part of `NNO` and none of the properties.
> 
> So it's `rawNNO`, in a sense. Which is rather bleh. (I'm not a big fan of `PointedUnary` btw, but at least you don't expect any laws to hold.)

Good point. But I think I'd be happier (given the intended mode-of-use of this PR) to call it `RawNNO`?

JacquesCarette:
I fully agree that if it is defined locally in this file, then `RawNNO` makes sense. But is it, or is it going to fit in the full `Algebra` hierarchy?

src/Algebra/Structures/Pointed.agda
```diff
@@ -0,0 +1,49 @@
...
+    c ℓ : Level
+
+
+------------------------------------------------------------------------
+-- Structures with 1 binary operation & 2 elements
+------------------------------------------------------------------------
+
+record IsPointedMonoid
+  (rawMonoid : RawMonoid c ℓ)
+  (• : RawMonoid.Carrier rawMonoid) : Set (c ⊔ ℓ)
```
MatthewDaggitt, line 34: I'm confused. You don't have a not equal to assumption between `•` and `ε` so there's not necessarily two elements...

src/Algebra/Structures/Pointed.agda
```diff
@@ -0,0 +1,49 @@
...
+  open RawMonoid rawMonoid public
+  open Structures _≈_
+
+  field
+    isMonoid : IsMonoid _∙_ ε
+
+  _×• : ℕ → Carrier
+  n ×• = n × • where open Definitions rawMonoid
+
+  module Literals where
```
MatthewDaggitt, line 45: Why define this as a named module, and not just a `Number` record directly?

MatthewDaggitt (changes requested):
> The design, at present, is pretty convoluted

Agreed. This PR needs some _serious_ thinking about the design. 

1. it needs a `README` file to show it actually works...
2. Is `Pointed` supposed to be whole new subhierarchy? If so it should go under `Algebra.Pointed.(Structures/Bundles)` rather than `Algebra.(Structures/Bundles).Pointed`
3. Having said that why do we need such generality? I feel we could get this working in about 5 lines of code in `Algebra.Properties.SemiringWithoutAnnihilatingZero` without needing to go through add 5 new modules, an entire new algebraic subhierarchy, and an entire new way of indexing algebraic structures. I think we need a pretty convincing argument for the need for the extra generality to justify all of this.

jamesmckinna:
So, thanks to each of you for the feedback on what was an 'essay'/attempt to attack the problem, while balancing a lot of ... competing tensions in how/where the library might be extended, esp. in respect of `Pointed` structure.

I'd be very happy to retire this one (or move it to `DRAFT`) while we revisit the originating issue to discuss how to tackle an acceptable solution. But like many 'leftover' so-called `low-hanging-fruit` issue, I was surprised where this one took me, once I started to take the idea more seriously in the spirit of 'put things at the place of maximum generality/reusability'...

(Meta: I maintain a tab of the list of `low-hanging-fruit` issues since 2020, and periodically pick one  up to play with... but I'm conscious that some of those should be closed, but weren't at the time that a PR was posted which did fix them, and some of them, like this, end up more intricate than at first sight appears. But grateful to @MatthewDaggitt for the hint, sometimes at least, to not over-think things... ;-))

src/Algebra/Structures/Pointed.agda
jamesmckinna: Absolutely! Degenerate instances are/should always be possible: we don't insist that `0#` not be equal to `1#` in `(Semi)ring*`, for example.
So perhaps the comment is bad, but this was mindless/lazy cut-and-paste from `Algebra.*`...

src/Algebra/Bundles/Literals.agda
```diff
@@ -0,0 +1,19 @@
...
+
+open import Algebra.Bundles.Pointed
+
+module Algebra.Bundles.Literals
+  {c ℓ} (pointedMonoid : PointedMonoid c ℓ)
+  where
+
+open PointedMonoid pointedMonoid
+
+-- Re-export `Number` instance defined in Algebra.Structures.Literals
```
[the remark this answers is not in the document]
jamesmckinna: No, but this seems to be the model in eg `Data.Nat.Literals` which I had been mindlessly following...

src/Algebra/Structures/Pointed.agda
jamesmckinna: * I had stymied myself with imports from `Agda.Builtin.FromNat`, perhaps in the mistaken belief that that was how `Number` instances get introduced...
* I had (for some reason) wanted to distinguish between the declaration of a module defining the implementation, from its export via a `number : Number ...` definition elsewhere...

jamesmckinna:
> > The design, at present, is pretty convoluted
> 
@MatthewDaggitt wrote:
>  This PR needs some _serious_ thinking about the design.

Agreed! My approach may be falwed, but it seems as though this hasn't been picked up since late 2020 when it was first proposed. I'd been picking at it on and off since 2021, but this was my first 'concentrated' effort since then.

>     1. it needs a `README` file to show it actually works...

OK.

>     2. Is `Pointed` supposed to be whole new subhierarchy? If so it should go under `Algebra.Pointed.(Structures/Bundles)` rather than `Algebra.(Structures/Bundles).Pointed`

That's a very good question! I had been following my nose a bit, given earlier attempts with the simplest cases of such a hierarchy, and wondering, but committing to, the thought that perhaps, indeed, we did?

>     3. Having said that why do we need such generality? I feel we could get this working in about 5 lines of code in `Algebra.Properties.SemiringWithoutAnnihilatingZero` without needing to go through add 5 new modules, an entire new algebraic subhierarchy, and an entire new way of indexing algebraic structures. I think we need a pretty convincing argument for the need for the extra generality to justify all of this.

All fair points. And on that basis, I'd be happy to withdraw this, in favour of someone coming up with just such a simpler solution, save for the following:
* we will want, eventually, to have something equivalent to `{Raw}NNO` in the library, won't we? And this does seem to be the actually minimal node in the hierarchy where the `n`th iterate (starting at `0#`) of `_+ 1#` belongs, or not? It doesn't need `Semiring...` until and unless you want to prove that it defines a `Semiring...` homomorphism from Nat to its image in the given algebra; but even then, that actual image itself is in fact *already* a `Semiring` by virtue of the various homomorphism properties proved for the `n ×_` operation in `Algebra.Properties.Monoid.Mult`... (I think!)
* happy to give way on the `Pointed` hierarchy, but it draws attention to how the existing approach (not indexing over a `Raw` bundle, but over its flattening into a telescope) actually makes it *harder* to entertain such extensions, or to systematically adding derived operations cf. #2268 ... IMHO. 

These points might be moot for the time being, but I can't help thinking they will return as issues... eventually. So I had started to explore the consequences of thinking about how they might affect this specific PR... hope that goes some way to explaining, if not justifying, the avenues I'd taken here.

jamesmckinna:
On naming: 
* `NNO`/`NaturalNumber(s)Object`? ... the Lawvere/categorical way;

JacquesCarette:
I am happy to entertain any of `PointedUnary`, `RawNNO` and `SuccessorSet`. I feel that a no-laws `NNO` seems like squatting over a good name prematurely.

jamesmckinna:
Will revisit as and when #2277 gets merged...

jamesmckinna:
Suggest closing this in favour of a fresh start based on #2277 and #2272 .

PR #2265, opened 2024-01-25
Parametrize module morphisms by raw bundles

Taneb:
Follow-on from #2263. These changes are breaking, although I think they're still important and worth it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
 
+* The modules and morphisms in `Algebra.Module.Morphism.Structures` are now
+  parametrized by _raw_ bundles rather than lawful bundles, in line with other
+  modules defining morphism structures.
+
 Other major improvements
 ------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Non-backwards compatible changes
 * The modules and morphisms in `Algebra.Module.Morphism.Structures` are now
   parametrized by _raw_ bundles rather than lawful bundles, in line with other
   modules defining morphism structures.
+* The definitions in `Algebra.Module.Morphism.Construct.Composition` are now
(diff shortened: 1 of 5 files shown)


jamesmckinna:
Obligations to discuss this, but I think the fact that this turn to indexing on `RawX` is already established (though the underdevelopedness of many homomorphisms under `Algebra.*` suggests that we have yet to really stress-test our designs?) suggests that this is a worthwhile breaking change.

Might (eventually) be worth polling as to how much the existing `Algebra.Module.*` hierarchy is already used in anger... or not.

Taneb:
The very scientific method of typing `language:Agda "Algebra.Module.Morphism"` into the GitHub search bar only came up with various forks of stdlib, so it doesn't look like there's many consumers of this.

JacquesCarette (approved):
This is clearly trickier, but is also code that should succumb to 'smart' automation.

Taneb:
Ugh, `Algebra.Module.Morphism.ModuleHomomorphism` broke. @jamesmckinna went over that module with me at the AIM in Delft lass year and I remember us working out that everything in there was either trivial or impossible to use

jamesmckinna:
> Ugh, `Algebra.Module.Morphism.ModuleHomomorphism` broke. @jamesmckinna went over that module with me at the AIM in Delft lass year and I remember us working out that everything in there was either trivial or impossible to use

Yes! And what we did not do then (alas!) was raise an issue about it there-and-then, if only to be able to return to it now...
Propose that you do not that now, and we tackle it as a separate PR, uncoupled from this one... (to the extent that that is possible; are they completely orthogonal, or would this need to be merged first beforeproceeding?)

Taneb:
It can be a separate PR in either order with no problem.

MatthewDaggitt:
> The very scientific method of typing language:Agda "Algebra.Module.Morphism" into the GitHub search bar only came up with various forks of stdlib, so it doesn't look like there's many consumers of this.

This is a peripheral and new enough module, that I think it's acceptable to make a breaking change.

Taneb:
Can this be merged?

PR #2269, opened 2024-01-29
Revise definitions of `Data.List.Base.scanl|inits`; add `Data.List.NonEmpty.Base.scanl⁺|inits⁺`

### LINKED ISSUE #2267: A 'better' (less strict) `Data.List.Base.tails`?
Working on #2258 suggests the following:
```agda
{- compared to the original
tails []       = [] ∷ []
-}
tails xs = xs ∷ go xs
  where
  go : List A → List (List A)
  go [] = []
  go (_ ∷ xs) = xs ∷ go xs
```
which has the advantage that `tails xs` observably produces a non-empty list (moreover, one whose head is the argument `xs`), *without* having to do a subsidiary pattern-match on `xs`. 

Similar remarks apply to eg `inits` and `scanl`, and as with moving `scanr` to `Data.List.NonEmpty`, perhaps these functions should, too.

jamesmckinna:
Fixes #2267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,6 @@ Highlights
 Bug-fixes
 ---------
 
-* Fix statement of `Data.Vec.Properties.toList-replicate`, where `replicate`
-  was mistakenly applied to the level of the type `A` instead of the
-  variable `x` of type `A`.
-
 Non-backwards compatible changes
 --------------------------------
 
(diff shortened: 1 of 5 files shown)


jamesmckinna:
Hmmm, looks as though I've screwed up the `CHANGELOG`.

JacquesCarette:
Yes, other than that, this looks really good!

jamesmckinna:
Hopefully that's sorted out now.

jamesmckinna:
So I think we've (now) agreed to find a better home for this (and #2258 ) under `Data.List.Something`. Suggestions welcome!

jamesmckinna:
Aaargh: `Data.List.Kleene.*`!

andreasabel (commented):
Thanks for the improvements!

In general, some list operations always return non-empty lists, so this should be expressed in types; and then one can choose to forget the extra information of non-emptiness using a conversion to plain lists.

jamesmckinna:
Now see #2395

PR #1928, opened 2023-02-14
Proofs of the Binomial Theorem for `(Commutative)Semiring` [supersedes #1287]

jamesmckinna:
This PR is ~~draft, for discussion~~ ready for review. A major overhaul of @uzytkownik 's #1287, but preserving the basic structure of his argument. 

It depends on the following ancillary PRs for completion (I've tried to keep the branch/commit structure clean):
- [x] #1923, to establish the `top` view of `Fin (suc n)`; UPDATED ~~have now removed dependency on this PR/can be~~ has now been reconfigured to use the ~~revised~~ merged version of it; 
- [x] #1926 for the required arithmetic of `_C_`; 
- [x] #1927 for the ancillary algebraic lemmas (phrased for `Semiring`s but could be weakened; see also issue #1917)

PR #2272, opened 2024-02-02
lemmas about semiring structure induced by `_× x`

jamesmckinna:
This is a largely cosmetic PR, to lay the groundwork for more extensive additions related to #1363 / #2264 .

It adds:
* 'homomorphism' properties for  `0`/`0#` (I'd dithered over this one, because it holds by definition, but I've included it for the sake of uniformity, and fixed the resulting name-clash... ;-)) and `1`/`1#` to `Algebra.Properties.Monoid.Mult`;
* of which the property for `1` was previously, erroneously, placed in `Algebra.Properties.Semiring.Mult` as part of #1928 , so a deprecation is also issued for that;
* repairs the single use of that lemma in  `Algebra.Properties.Semiring.Binomial` in favour of the simpler `Monoid` property;
* the companion property `×-homo-0#` for `0` which *does* require `Semiring` structure (specifically: the `*-zero` property, `AnnihilatingZero`) now has a new (hopefully more uniform) name in `Algebra.Properties.Semiring.Mult`;
* `CHANGELOG` entries.

Outstanding: refactoring this so that all the lemmas except for `×-homo-0#` move to `Algebra.Properties.SemiringWithoutAnnihilatingZero.Mult` which currently doesn't exist, and re-exporting its content under `Algebra.Properties.Semiring.Mult` along with `×-homo-0#`...

Once merged, I plan to follow this up with a more ambitious development of: (UPDATED :man_facepalming: )
* `SuccessorSet` #2273 , and 
* (TODO) the proof that it induces (via freeness/initiality from Nat, as the weakly initial such thing) a sub-*`Semiring`* structure on the carrier of a *`Monoid`* via `_× x`...

### DIFF
--- a/src/Algebra/Properties/Semiring/Mult.agda
+++ b/src/Algebra/Properties/Semiring/Mult.agda
@@ -82,5 +82,5 @@ idem-×-homo-* m n {x} idem = begin
 1×-identityʳ = ×-homo-1
 {-# WARNING_ON_USAGE 1×-identityʳ
 "Warning: 1×-identityʳ was deprecated in v2.1.
-Please use Algebra.Properties.Monoid.Mult.×-homo-1 instead. "
+Please use ×-homo-1 instead. "
 #-}
--- a/src/Algebra/Properties/Semiring/Mult.agda
+++ b/src/Algebra/Properties/Semiring/Mult.agda
@@ -6,7 +6,7 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra
+open import Algebra.Bundles using (Semiring)
 open import Data.Nat.Base as ℕ using (zero; suc)
 
 module Algebra.Properties.Semiring.Mult
@@ -14,6 +14,7 @@ module Algebra.Properties.Semiring.Mult
 
 open Semiring S renaming (zero to *-zero)
 open import Relation.Binary.Reasoning.Setoid setoid
+open import Algebra.Definitions _≈_ using (_IdempotentOn_)
 
 ------------------------------------------------------------------------
 -- Re-export definition from the monoid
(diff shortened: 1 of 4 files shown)


jamesmckinna:
Thanks @JacquesCarette !
I think it's ready, but felt that I should raise for discussion the potential additional refactoring 'opportunity' it represents. I dimly recall that during #1928  the question came up around "where do/should things live?" in respect of (at least: aiming for) maximum generality/minimum commitment as to algebraic structure/hypotheses. That issue has come up in spades in the context of #2264 , and while this PR contributes to that effort, I'm not clear really what is gained by moving stuff to `SemiringWithoutAnnihilatingZero`. So my hesitation was out of a desire to solicit feedback on whether that additional work is desirable/wanted/needed. And if so, happy to incorporate it it into this PR (but you seem often to seek to avoid such 'mission creep')... So, I'd be happy to mark as draft until that discussion takes place, but perhaps better, instead to say: no this is ready, but we should have that discussion downstream as a follow-on design issue/PR (just as this arises as a 'knock-on' from the discussion on those earlier PRs). 

What still feel provisional, and in need of review, are the naming choices, as with so many PRs!

src/Algebra/Properties/Semiring/Mult.agda
```diff
@@ -60,3 +55,32 @@ open import Algebra.Properties.Monoid.Mult +-monoid public
...
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.1
+
+1×-identityʳ = ×-homo-1
+{-# WARNING_ON_USAGE 1×-identityʳ
+"Warning: 1×-identityʳ was deprecated in v2.1.
+Please use Algebra.Properties.Monoid.Mult.×-homo-1 instead. "
```
MatthewDaggitt, line 85: Given that `×-homo-1` is already publicly exported by this module, I don't think the name should be qualified in the warning.

MatthewDaggitt (commented):
Otherwise happy to merge.

PR #2275, opened 2024-02-03
 [fix #2273] Add `NNO`-related modules 

### LINKED ISSUE #2273: Add `Algebra.Bundles.SuccessorSet` and related records
Previously, on #2264 I raised the following discussion point: 
* ~~`NNO`/`NaturalNumber(s)Object`? ... the Lawvere/categorical way;~~ may he rest in peace, I've been completely ignoring initiality...

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2264#issuecomment-1923771199_
            

So this issue to is to offer the following specimen design/implementation, for discussion (and is tantamount to a fully articulated PR, moreover purely mechanically derivable by analogy with all the others). I'd be happy to change any of the names, in the spirit of the above original comment. ~~Against @JacquesCarette I'd hold out for `NNO`, but~~ 

I'll go for `SuccessorSet` as a reasonable, if still verbose, first choice...

(deleted: previous content, invalid thanks to my stupidity...)

jamesmckinna:


jamesmckinna:
`status:invalid`: I temporarily lost my grip on this by ignoring initiality, will close and reframe. :man_facepalming:

PR #2277, opened 2024-02-04
[fixes #2273] Add `SuccessorSet` and associated boilerplate

### LINKED ISSUE #2273: Add `Algebra.Bundles.SuccessorSet` and related records

jamesmckinna:


src/Algebra/Bundles.agda
```diff
@@ -23,11 +23,32 @@ open import Level
...
+-- Bundles with 1 unary operation & 1 element
+------------------------------------------------------------------------
+
+record SuccessorSet c ℓ : Set (suc (c ⊔ ℓ)) where
+  infix  8 _+1#
+  infix  4 _≈_
+  field
+    Carrier          : Set c
+    _≈_              : Rel Carrier ℓ
+    _+1#             : Op₁ Carrier
```
JacquesCarette, line 41: Nitpick: should this really be `_+1#` instead of the simpler `suc` ? That would be more coherent with the structure's name.

JacquesCarette (changes requested):

src/Algebra/Bundles.agda
jamesmckinna: See also the issue #2273 : I'd rather have *one* of 
* `zero#`/`suc#` (by analogy with `Data.Nat`) *or* 
* `0#`/`_+1#` (by analogy with `Semiring` etc., given the intended use case for this as part of #1363 / #2264)
for the sake of 'notational' consistency. 

I decided *against* using backquote syntax, since such usage (probably!?) better belongs as reserved to the addition of `Free` things, where it would directly correspond to introducing constructors-as-syntax/`Free`-syntax-as-constructors...

JacquesCarette: I vote for `zero#` / `suc#` given that it is called `SuccessorSet`.

jamesmckinna: DONE!

JacquesCarette (approved):

MatthewDaggitt (approved):

PR #2279, opened 2024-02-06
Fix import in `README.Data.Fin.Substitution.UntypedLambda`

### LINKED ISSUE #1711: Add properties of the quotient to Data.Nat.Divisibility

### LINKED ISSUE #1743: top-level README.agda (and Everything.agda)

### LINKED ISSUE #2086: Generated Agda standard library HTML has no version information
@jwaldmann writes

> NB - I found it (mildly) confusing that there is no version information (in the url) at
https://agda.github.io/agda-stdlib/Data.Product.Relation.Binary.Lex.Strict.html

I wonder how we can add version information to the generated HTML? Ideally we don't really want the URL to change every time we make a release as that would make linking a nightmare....

### LINKED ISSUE #2153: Bug in re-exporting `_∷=_` and `_─_` from `Data.List.Membership.Setoid`
Looking back through my huge backlog of standard library related emails, I found this comment by @gallais earlier this year

https://github.com/agda/agda-stdlib/commit/abbd3e7371d87d2141d23080a67bf8b8483030a9#r108952994

I assume this bug is still active.

### LINKED ISSUE #2166: Problems with change of `Semilattice` operation name
First issue with RC 1. Changelog doesn't mention that the name of the operation in `Semilattice` changed from `_∧_` to `_∙_`.

More problematically the names in `IsSemilattice` have not been changed :cry: 

https://github.com/agda/agda-stdlib/blob/a7d23029ac94ba446d2cf2a6f4bf3859d7aa9506/src/Algebra/Lattice/Structures.agda#L39

### LINKED ISSUE #2168: Choice of pre/post-condition combinators in proof names

### LINKED ISSUE #2175: Typos in documentation/`README`/`README.*`
#2176 fixes:
- [x] `README`: Agda version number should be 2.6.4

See branch [documentation](https://github.com/jamesmckinna/agda-stdlib/tree/documentation) where I'm collecting these towards an eventual PR for rc1/rc2...:
- [x] `README`: Missing comma L10 at "Bradley Hardy Joachim Breitner,"
- [x] `Relation.Nullary.Decidable.Core`: corrected reference to `README.Design.Decidability`
- [x] `installation-guide` (thanks @Saransh-cpp !)
- [x] instance arguments discussion in `CHANGELOG`
- [ ] ...
- [ ] documenting the 'how to typecheck `README.agda` question/problem... UPDATED: now going to be handled in a separate PR by @gallais 

Now leaving this issue open, to accumulate any of the other documentation things we might find...?

### LINKED ISSUE #2178: Consistent use of symbol `•`: in `•-cong  : Congruent₂ _•_` in `Algebra.Consequences.Setoid`
Is this the *same* symbol as discussed in #1544 ? If not, why not? And do we need to fix this? 

Anywhere else that we might have 'identical' renderings of distinct symbols? cf #1544 on the potential for confusion...

### LINKED ISSUE #2195: Redundant `zero` in `IsRing`

### LINKED ISSUE #2216: Display of (redundant/deprecated) pattern synonyms

### LINKED ISSUE #2232: Post-release version bump in `standard-library.agda-lib`?

### LINKED ISSUE #2234: updating `README.md`
Two (at least!) potential opportunities/obligations to update:
* discussion of the `experimental` branch vs. `master`: what's the current situation, and our current policy?
* discussion of the `--without-k` flag: should this become, or have added, a note about `--cubical-compatible`?

Also: in the instructions on how to obtain old versions of the library, do we (still) need to go as far back as v0.17?

### LINKED ISSUE #2278: Build failure for `README.Data.Fin.Substitution.UntypedLambda`
```
doc/README/Data/Fin/Substitution/UntypedLambda.agda:22,6-17
No module ≡-Reasoning in scope
  open ≡-Reasoning
```

The examples in `doc/README` do not seem to be covered by CI.
Which inevitably will mean that they bit-rot...  
Maybe related to the latter point:
- #1380

andreasabel:
Fix import in `README.Data.Fin.Substitution.UntypedLambda`, add this module to `doc/README.agda`.

Closes #2278.

### DIFF
--- a/doc/README.agda
+++ b/doc/README.agda
@@ -222,6 +222,10 @@ import Text.Printf
 
 import README.Case
 
+-- Showcasing the framework for well-scoped substitutions
+
+import README.Data.Fin.Substitution.UntypedLambda
+
 -- Some examples showing how combinators can be used to emulate
 -- "functional reasoning"
 
--- a/doc/README/Data/Fin/Substitution/UntypedLambda.agda
+++ b/doc/README/Data/Fin/Substitution/UntypedLambda.agda
@@ -14,7 +14,7 @@ open import Data.Fin.Substitution.Lemmas
 open import Data.Nat.Base hiding (_/_)
 open import Data.Fin.Base using (Fin)
 open import Data.Vec.Base
-open import Relation.Binary.PropositionalEquality.Core
+open import Relation.Binary.PropositionalEquality
   using (_≡_; refl; sym; cong; cong₂; module ≡-Reasoning)
 open import Relation.Binary.Construct.Closure.ReflexiveTransitive
   using (Star; ε; _◅_)


JacquesCarette:
Why does you PR have tons of lines of description when it is a 1-liner change?

andreasabel:
I suppose this `gh pr create --fill` went wrong...

JacquesCarette:
I've edited the commit message to be just what it is about. I guess your branch might have been behind `master` enough that it did a 'merge' of all of that, thus the noise?

andreasabel:
> I've edited the commit message to be just what it is about. I guess your branch might have been behind `master` enough that it did a 'merge' of all of that, thus the noise?

Thanks.  Strangely enough, no, I don't really understand what happened, but I have seen `gh` malfunction before.

jamesmckinna:
Do we want to also add the relevant link(s) to `README.Data.Fin.*` to `doc/README` as part of this PR?
(But not wanting mission-creep for the sake of it... ;-))

andreasabel:
> Do we want to also add the relevant link(s) to `README.Data.Fin.*` to `doc/README` as part of this PR?

Good point.  I added this module.  But now that you say it, there might be more... hang on...

jamesmckinna (approved):

jamesmckinna:
@MatthewDaggitt is this a v2.0.1 for the `bug`, or a regular v2.1?

Issue #2280, opened 2024-02-06
Standardise names for qualified module imports


jamesmckinna:
Task list to address #2201  , itemised by *what* needs to change (with proposed changes), rather than *where*:
 - [ ] `*.Bundles` as `Bundles`
 - [ ] `*.Bundles.Raw` as `Raw`
 - [ ] `*.Consequences{.*}` as `Consequences`
 - [ ] `*.Construct.Composition` as `Composition`
 - [ ] `*.Construct.Identity` as `Identity`
 - [ ] `*.Definitions` as `Definitions`
 - [ ] `*.Effectful.Transformer` as `Transformer`
 - [ ] `*.Structures` as `Structures`
 - [ ] `Data.Nat{.*}` as `ℕ`
 - [ ] `Data.Product.Base` as `Product` (or `Σ`?)
 - [ ] `Data.Sum.Base` as `Sum` (or `⊎`?)
 - [ ] `Data.X.Properties` as `X`, rather than `Xₚ` with `X` the datatype name, not the ASCII module name (as in `ℕ` above)
 - [ ] `Function.*` as `Function`
 - [ ] `Relation.Binary.PropositionalEquality{.*}` as `≡`
 - [ ] `Relation.Binary.Reasoning.Setoid` as `≈-Reasoning` by analogy with `≡-Reasoning`
 - [ ] `Relation.Binary.*` as `Binary`
 - [ ] `Relation.Nullary.*` as `Nullary` (exceptions for `Decidable` and `Negation`?)
 - [ ] `Relation.Unary.*` as `Unary`

Things to leave alone:
 - [ ] `System.*`
 - [ ] ...

jamesmckinna:
On #2284 @JacquesCarette wrote:
 > Sigh - `Product` is non-trivially longer than `Prod`. But this is the right thing to do.

So... I've been reconsidering this, and wondered if, instead, and in line with the 'datatype name as root module name' heuristic, that we standardise on `import Data.Product.Base as Σ`?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2284#issuecomment-1937722328_

jamesmckinna:
(I've touched the opening paragraph many times this morning, and now GitHub seems to have lost all my edits... will wait and see what happens eg. if it's a caching glitch...)

jamesmckinna:
[...]

PR #2281, opened 2024-02-06
Qualified import of `Data.Nat` fixing #2280

jamesmckinna:
One or two additional renamings where it was easiest to do in-place.
Outstanding issue: `Data.Integer.Divisibility` seems to need to disambiguate the module imports, but perhaps this can be fixed by a `pattern` synonym or a renaming?

### DIFF
--- a/src/Algebra/Properties/Semiring/Binomial.agda
+++ b/src/Algebra/Properties/Semiring/Binomial.agda
@@ -10,10 +10,10 @@
 
 open import Algebra.Bundles using (Semiring)
 open import Data.Bool.Base using (true)
-open import Data.Nat.Base as Nat hiding (_+_; _*_; _^_)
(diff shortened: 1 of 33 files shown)


JacquesCarette (approved):

JacquesCarette:
I think this is worth pulling in as is, the outstanding issue shouldn't hold up this nice clean-up. Perhaps an annotation on #2280 might be warranted?

jamesmckinna:
So I've started making PRs on a tickbox-by-tickbox basis, and updating #2280 to reflect (incremental) progress, but let me know if you thin more annotation (and of what kind...) might be helpful here.

jamesmckinna:
I've added some more commentary on the problem: the `import`s occur in `Data.Integer.Divisibility`... move further discussion to the issue #2280

PR #1930, opened 2023-02-14
Another attempt at deprecation of `inspect`

jamesmckinna:
REVISED: Supersedes #1630 . 

Deprecations added to
- [x] `Relation.Binary.PropositionalEquality` 
- [ ] `Relation.Binary.HeterogeneousEquality` (suggest that this be the topic of a separate PR: as well as having its own versions of `Reveal` and `inspect`, it uses the now-deprecated [equational reasoning combinators from before v1.3](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG/v1.3.md#changes-to-how-equational-reasoning-is-implemented)) 

as well as everything else now under #1630 (which this PR updates and extends) towards eliminating uses of `inspect` in favour of `with ... in ...` (issue agda/agda-stdlib#1580)

FIXED by moving to Agda-2.6.3

UPDATED:
- [ ] `README.WithIn` in place of `README.Inspect` NOW: reverted this change; deprecated latter module;
- [x] two additional inversion properties concerning `splitAt` in `Data.Fin.Properties` 

~~UPDATED: for the cost of two new inversion lemmas, the old proof is eliminable in favour of... something no more comprehensible...; `combine` is defined by `with` on `splitAt`, so I took the proofs back to that decomposition... another 'view of a function' deployment, albeit in disguise...~~

~~Otherwise, a frictionless migration.~~

~~EDITED: No deprecations as yet; no `CHANGELOG`. Hence: "towards"...~~

~~Suggest cherry-picking [`57b90e8912d8f43462ac059972ea31cbf8a34fb4`](https://github.com/agda/agda-stdlib/pull/1930/commits/57b90e8912d8f43462ac059972ea31cbf8a34fb4) as below~~

~~Hmm that won't quite cut it, unfortunately. I'll try to produce a cleaned up version in due course. Perhaps also with an updated `README.Inspect`... perhaps better `README.WithIn`?~~

PR #2282, opened 2024-02-06
Qualified import of reasoning modules fixing #2280

jamesmckinna:


src/Data/Integer/Divisibility/Signed.agda
```diff
@@ -50,38 +49,38 @@ open _∣_ using (quotient) public
...
     ikq*∣k∣≢0 = *-nonZero ∣ ikq ∣ ∣ k ∣ {{ikq≢0}} {{≢-nonZero neq}}
 
   sign-eq : sign i ≡ sign (ikq * k)
   sign-eq = sym $ begin
-    sign (ikq * k)                  ≡⟨ sign-◃ (sign ikq S.* sign k) (∣ ikq ∣ ℕ.* ∣ k ∣) ⟩
-    sign ikq S.* sign k             ≡⟨ cong (S._* sign k) (sign-◃ (sign i S.* sign k) q) ⟩
-    (sign i S.* sign k) S.* sign k  ≡⟨ SProp.*-assoc (sign i) (sign k) (sign k) ⟩
-    sign i S.* (sign k S.* sign k)  ≡⟨ cong (sign i S.*_) (SProp.s*s≡+ (sign k)) ⟩
-    sign i S.* S.+                  ≡⟨ SProp.*-identityʳ (sign i) ⟩
+    sign (ikq * k)                  ≡⟨ sign-◃ (sign ikq Sign.* sign k) (∣ ikq ∣ ℕ.* ∣ k ∣) ⟩
```
JacquesCarette, line 72: picky: the alignment seems to need tweaked.

JacquesCarette (changes requested):
A single thing needs tweaking. But also: #2270 doesn't document the names of various reasoning modules; you chose some that are very reasonable, but which do use unicode. That is one doubt that I have, i.e. did we agree on that? I don't remember! And neither the issue nor the change to the style guide bring clarity.

jamesmckinna:
> A single thing needs tweaking. But also: #2270 doesn't document the names of various reasoning modules; you chose some that are very reasonable, but which do use unicode. That is one doubt that I have, i.e. did we agree on that? I don't remember! And neither the issue nor the change to the style guide bring clarity.

On the use of Unicode: we *had* agreed to the use of `ℕ` as the qualified module name for the (most-?) common cases of importing qualified from `Data.Nat.*`, and equality as `≡` from `Relation.Binary.PropositionalEquality.*`.

On the names chosen to standardise the various `Reasoning` modules, I needed to make *a* choice, moreover one which transferred across the imports as `≡-Reasoning`, `SetoidReasoning`, `PreorderReasoning`, `PartialOrderReasoning` (other, shorter names are also used). Following the various PRs to regularise the 'standard' infix notations for the various `Relation.Binary.Bundles`, I decided (unilaterally!) to use those symbols as the prefix to `-Reasoning`...

So yes: a design decision I 'sneaked' into the PR. I'll summarise the above on the main issue #2280 .

src/Data/Integer/Divisibility/Signed.agda
jamesmckinna: Revised first commit fixes the alignment, but I wonder if this is a case where interleaving the terms and the reasoning line-by-line might be a better layout for the sake of avoiding long lines...?

jamesmckinna: The alignment in `Data.List.Relation.Binary.BagAndSetEquality` is also (historically) terrible, and I don't necessarily think this worth fixing here; against #1630 / #1930 , to my surprise this module also makes extensive use of `inspect`... so might want a separate consideration outside this 'boilerplate' PR...

jamesmckinna: Indeed: a global `cosmetic` issue on 'fixing alignment in reasoning' might be in order...

MatthewDaggitt (approved):

JacquesCarette (approved):

JacquesCarette:
Right - 2 approvals for this design, so 3 people like it, seems good enough. Merging. It should be documented in the style guide.

PR #2284, opened 2024-02-07
Qualified import of `Data.Product.Base` fixing #2280

jamesmckinna:
Outstanding issues:
* disambiguation in `Data.Container.*` wrt product structure on containers vs. that on types (both at the level of the type constructor, and at that of (internal) module naming)
* ...

(And with this being the third PR tackling #2280 , starting to feel diminishing returns without a bit more thought about how to deliver on the `style-guide` recommendations)

### DIFF
--- a/src/Algebra/Construct/Flip/Op.agda
+++ b/src/Algebra/Construct/Flip/Op.agda
@@ -10,7 +10,7 @@
 module Algebra.Construct.Flip.Op where
 
 open import Algebra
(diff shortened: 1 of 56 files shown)


MatthewDaggitt (approved):

JacquesCarette (approved):
Sigh - `Product` is non-trivially longer than `Prod`. But this is the right thing to do.

JacquesCarette:
That would be reasonable too.

Issue #2236, opened 2023-12-21
CI needs updating to more recent GHC version


JacquesCarette:
Looks like the CI uses GHC 8.10.7 (see `.github/workflows/ci-ubuntu.yml`) and cabal 3.6.2.0. Agda itself now seems to use stack and GHC 9.8.1.  That's a very large gap!

jamesmckinna:
And now... more upgrading required:
```
Warning: As of 2023-09-09, haskell/action/setup is no longer maintained, please switch to haskell-actions/setup (note: dash for slash).
***************************************************************************
**                                                                       **
**              This action is DEPRECATED.                               **
**                                                                       **
**              Please use haskell-actions/setup instead.                **
**                                                                       **
**              (Note the dash instead of the slash.)                    **
**                                                                       **
***************************************************************************
```

PR #2286, opened 2024-02-08
missing code endquote

jamesmckinna:


Taneb:
I wonder if it would be worth adding a markdown validator to CI for the changelog. It feels like it's subtly broken more often than not

jamesmckinna:
Two things:
* I agree about improving the CI validation, although, as I've argued before, I think that much of that validation should take place *post-merge*, rather than 'continuously' on each commit, especially as we squash commits, and don't necessarily expect to enforce, neither on nor pre- merge, that every commit be valid...
* for my browser at least, the 'bug' fixed in this PR doesn't actually break the rendering of markdown, but I also agree that (especially, ironically, that fixing merge conflicts in `CHANGELOG` makes it too easy, for me at least, to break the markdown... :-() `CHANGELOG` is a weak link in our CI chain. 

But how to improve this state of affairs? And should such improvement be part of the CI on GitHub, or would it be done better as a pre-commit hook locally (as is sometimes suggested for `fix-whitespace`)? Consider opening a recurrent/pinned task-list issue, under "CI improvement" (we already have "CI needs updating..." #2236 )?

JacquesCarette (approved):

JacquesCarette:
I would want such checks at "merge validation" time. Not sure what the proper name for that is?

I definitely would want it to be part of the CI.

Taneb (approved):

PR #2290, opened 2024-02-11
Qualified import of `Data.Sum.Base` fixing #2280

jamesmckinna:
Follows the proposal in the issue: `as Sum`, but the 'datatype name as root module name' heuristic suggests `⊎` instead, cf. my comment on #2280 / #2284 concerning `Product`...

Also: rectifies `Data.Container.Combinator.Properties` to 'standardise' on `as S`, `as P` for disambiguation wrt `Sum` and `Product`, and also`PropositionalEquality as ≡` #2282

### DIFF
--- a/src/Codata/Sized/Delay.agda
+++ b/src/Codata/Sized/Delay.agda
@@ -16,9 +16,9 @@ open import Data.Empty
 open import Relation.Nullary
 open import Data.Nat.Base
 open import Data.Maybe.Base hiding (map ; fromMaybe ; zipWith ; alignWith ; zip ; align)
-open import Data.Product.Base as P hiding (map ; zip)
-open import Data.Sum.Base as S hiding (map)
-open import Data.These.Base as T using (These; this; that; these)
+open import Data.Product.Base hiding (map ; zip)
+open import Data.Sum.Base hiding (map)
+open import Data.These.Base using (These; this; that; these)
 open import Function.Base using (id)
 
 ------------------------------------------------------------------------
--- a/src/Data/Container/Combinator.agda
+++ b/src/Data/Container/Combinator.agda
@@ -10,8 +10,8 @@ module Data.Container.Combinator where
 
 open import Level using (Level; _⊔_; lower)
 open import Data.Empty.Polymorphic using (⊥; ⊥-elim)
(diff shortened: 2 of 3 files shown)


JacquesCarette (commented):
I feel like putting `Data.Container.Combinator.Properties` in this odd intermediate state where a non-standard name is used doesn't seem like a modification worth pulling in?

jamesmckinna:
Well, perhaps you're right, but at least the name of `PropositionalEquality` is corrected (which should stay), but that then permits `Data.Product.Base as P` harmonised with `S` for `Sum`?

jamesmckinna:
And maybe this 'sore thumb' is what started me rethinking the choice of \Sigma... but I'm not thrilled by \uplus ... gah!

JacquesCarette:
Should `Data.Product.Base` be `Product` wrt the current scheme?

jamesmckinna:
Well that's a thing: the 'datatype heuristic' suggest we use the single character Unicode symbol, because... that's the datatype name. Despite all our discussion on #2280 I am also drawn to the 'ASCII module root name as qualified name' heuristic. So it may, indeed, be that I have adopted 'mixed' heuristics in the various PRs I've posted...

JacquesCarette:
(I was more complaining that you used `S` where neither of those heuristics would have meant that was the name to use.)

jamesmckinna:
Sure, it is a ⟪calque⟫, and only 'mere' alphabetisation could be claimed to justify it...

MatthewDaggitt (approved):
Approving on the basis that it is an improvement over the current state and once we've decided on Product and Sum we can come back and fix.

jamesmckinna:
Thanks @MatthewDaggitt I agree with each of you and @JacquesCarette that this is still not-yet what we finally want, but one for further discussion... downstream?

MatthewDaggitt:
Eek merge conflicts don't look great. @jamesmckinna could you take a look?

jamesmckinna:
I think I see the problem: I made different choices in #2293 to the ones in this PR, which were a more straightforward transcription of the original, which seemed to go to some lengths to systematically maintain qualified imports for each of the `Container` and `Set` hierarchies. But I think the 'flattening' in #2293 is 'cleaner'. Will resolve the conflict in favour of `master` unless anyone shouts!

PR #2294, opened 2024-02-14
Qualified imports in `Data.Integer.Divisibility` fixing #2280

jamesmckinna:
This PR (re-)visits the corner case of #2280 , to hopefully avoid the wrinkles:
* makes `Data.Nat.Divisibility` a 'standard' qualified import;
* refactors the appeals to its contents via the qualified name;
* avoids the previous `public` re-export of the constructor `divides` in favour of a pattern synonym;
* knock-on changes to `Data.Integer.*` to reflect the refactoring.

The pattern synonym is perhaps the only awkward bit wrt any potential clients (`stdlib` seems OK), so comments welcome. 

UPDATED: added the `breaking` label to reflect the discussion below.

### DIFF
--- a/src/Data/Integer/Divisibility.agda
+++ b/src/Data/Integer/Divisibility.agda
@@ -14,10 +14,10 @@ open import Function.Base using (_on_; _$_)
 open import Data.Integer.Base
 open import Data.Integer.Properties
 import Data.Nat.Base as ℕ
-import Data.Nat.Divisibility as ℕᵈ
+import Data.Nat.Divisibility as ℕ
 open import Level
 open import Relation.Binary.Core using (Rel; _Preserves_⟶_)
-open import Relation.Binary.PropositionalEquality
+
 
 ------------------------------------------------------------------------
 -- Divisibility
@@ -25,9 +25,7 @@ open import Relation.Binary.PropositionalEquality
 infix 4 _∣_
 
 _∣_ : Rel ℤ 0ℓ
-_∣_ = ℕᵈ._∣_ on ∣_∣
-
-open ℕᵈ public using (divides)
+_∣_ = ℕ._∣_ on ∣_∣
 
 ------------------------------------------------------------------------
 -- Properties of divisibility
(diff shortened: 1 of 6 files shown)


JacquesCarette (approved):
Looks good to me.

src/Data/Integer/Divisibility.agda
```diff
@@ -14,42 +14,42 @@ open import Function.Base using (_on_; _$_)
...
 -- Divisibility
 
 infix 4 _∣_
 
 _∣_ : Rel ℤ 0ℓ
-_∣_ = ℕᵈ._∣_ on ∣_∣
+_∣_ = ℕ._∣_ on ∣_∣
 
-open ℕᵈ public using (divides)
+pattern divides k eq  = ℕ.divides k eq
```
MatthewDaggitt, line 30: Are we sure that this doesn't cause clashes if people have both `Data.Nat.Divisibility` and `Data.Integer.Divisibility` in scope?

jamesmckinna: Actually, in some sense, I think it is worse than that, in that it *will* cause clashes (unless there's something I don't understand), so this is perhaps Not The Right Answer. 

That said, I would worry about any context in which each of `Data.Nat.Divisibility` and `Data.Integer.Divisibility` were in scope *unqualified*, because this would cause a separate (dis-)ambiguation problem wrt the relation symbol `_|_` itself. 

In fact, `Data.Integer.Divisibility.Signed` *does* import both modules, but *qualified*, in the service of defining, and proving properties of, `_|_` on the signed representation of `ℤ`, which *also* introduces `divides` as a constructor...! And that *does* work fine, because each module is imported qualified, and indeed, the imported constructor is used only in one place, in the definition of `∣ᵤ⇒∣` ;-) My PR currently re-uses `ℕ.divides` in that definition, but (of course!?) using `Unsigned.divides` works as well, and perhaps better 'respects the abstraction barrier'.

I think the existing (legacy) design is a hack to 'overload'/'borrow' the constructor for one definition, that of `_|_` on `ℕ`, and turn it into a constructor for `_|_` on `ℤ`. This is fine (I don't like it, but meh) on the current model of untyped patterns/pattern synonyms. But I think that if we ever move to a *typed* model of pattern matching, then this shouldn't be OK? (or could it be?)

The problem, as I understand it, is to find a way to 'export' the constructor, for the (unsigned) `Integer.Divisibility` context, in such a way that it is type-correct for the definition of `_|_` on `ℤ`. My solution, which I admit I perhaps did not think hard enough about, nor long enough, was to make this 'shallow' copy. But I'm not sure I know how to do better than the original, in my view hacky, design. Advice welcome!

jamesmckinna: Committed the 'better' disambiguation, and a note in `CHANGELOG`.

But that doesn't resolve the question of whether this is the 'right' way to solve the problem... client modules (none in `stdlib`) which exploit the pun on import will break, I guess, so I'll add the `breaking` label?

jamesmckinna:
Implied `style-guide` recommendation: avoid puns!?

jamesmckinna:
Separate note: the proof of `∣ᵤ⇒∣` could in any case benefit from refactoring to reflect new lemmas about `NonZero` in `Data.Nat.Properties`, and perhaps to lift some local definitions out as properties of the signed representation in their own right, such as `◃-nonZero : ∀ s n .{{_ : ℕ.NonZero n}} → NonZero (s ◃ n)`...

src/Data/Integer/Divisibility.agda
jamesmckinna: Half-way house: put a `{-# WARNING_ON_USAGE -}` for `divides` `(*not*` a deprecation _per se_) in favour of `ℕ.divides` for v2.1, and then make the breaking change in v3.0?

PR #2295, opened 2024-02-14
Fix `sized-types` error in `README.Data.Tree.Binary`

jamesmckinna:
Corrects another orphan module under `doc/README`: use of `--sized-types` is no longer `--safe` ... cf. #2278 #1380


JacquesCarette (approved):
Good catch. I guess this doesn't affect `GenerateEverything.hs` as `README/*` is not included.

MatthewDaggitt (approved):

Issue #1906, opened 2023-01-11
`Zero` objects in the `Algebra.*` hierarchy


jamesmckinna:
`Algebra.Construct.Zero` defines some zero objects (`magma`, `semiring`, `band`) which are *not* zero objects, only terminal ones. Correspondingly, true zero objects defined there should also be re-exported as *initial* objects. Suggest a refactor to put things 'in the right place': 
* define `Terminal` to introduce all the terminal algebras; 
* re-export the subset of those which *are* zero objects also in `Initial` and `Zero`; 
* define the true initial objects for `Magma`, `Semigroup`, `Band` in `Initial`

See PR #1902

Taneb:
I think we should (not necessarily in the PR that closes this, but maybe) include the defining properties of these initial/terminal/zero objects, so for example for initial objects, that there is a unique homomorphism from them to any other object. Maybe this can go somewhere like `Algebra.Construct.Initial.Properties`?

JacquesCarette:
1. Yes!
2. Next PR. Let's not achieve paralysis by drowning the great by requiring perfection.

PR #2296, opened 2024-02-14
Add unique morphisms from/to `Initial` and `Terminal` algebras

jamesmckinna:
Following up on #1906 , but with different location for the constructions.
- [x] `Monoid`
- [x] `Group`
- [x] `NearSemiring`
- [x] `Semiring`
- [x] `Ring`

PR does *not* introduce `Algebra.Morphism.Construct.Zero`, by contrast with `Algebra.Construct.Zero`, because the initial/terminal *morphisms* are different, even while the zero *objects* are identical...


src/Algebra/Morphism/Construct/Terminal.agda
```diff
@@ -0,0 +1,98 @@
...
+
+private
+  variable
+    a ℓa : Level
+
+
+------------------------------------------------------------------------
+-- The underlying data of the morphism
+
+module 𝕆neMorphism (M : RawMonoid a ℓa) where
```
JacquesCarette, line 40: That module name is perhaps overly cute?

JacquesCarette (commented):
Looking good.

src/Algebra/Morphism/Construct/Terminal.agda
jamesmckinna: I'd wanted to overload `𝕆ne` from `Algebra.Construct.Terminal`, but resolving the disambiguation was horrible. 
As for 'cute', sure, but when writing `Algebra.Construct.Terminal` and `Algebra.Construct.Initial`, i had been trying to write a *single* internal module definition such that all the relevant definitions could be written as `record { 𝕆ne }` etc. (it nearly worked, but not quite). 

I *do* like blackboard-bold identifiers for 'canonical' constructions, but `\b1` doesn't display in my Emacs... :-(

jamesmckinna: Alternative: `The{Unique}Morphism`?

JacquesCarette:
I think I liked `OneMorphism` and `ZeroMorphism` (without the cuteness) more than `The`. In some ways, if these files are meant to be used qualified, then just `Morphism` might work, so that they would be `Terminal.Morphism` and `Initial.Morphism`.

jamesmckinna:
> I think I liked `OneMorphism` and `ZeroMorphism` (without the cuteness) more than `The`. In some ways, if these files are meant to be used qualified, then just `Morphism` might work, so that they would be `Terminal.Morphism` and `Initial.Morphism`.

I don't want to make this a hill on which either of us die defending... I'd be happy/ier with plain `Morphism` than `One` or `Zero` (I agree with removing the cuteness), but I liked the rhetorical force (not: cuteness!) of the `The` (`Unique...` would also work?).

JacquesCarette:
`Unique` has a certain charm to it, I agree.

src/Algebra/Morphism/Construct/Initial.agda
```diff
@@ -0,0 +1,73 @@
...
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Level using (Level)
+
+module Algebra.Morphism.Construct.Initial {c ℓ : Level} where
+
+open import Algebra.Bundles using (Magma)
+open import Algebra.Bundles.Raw using (RawMagma)
+open import Algebra.Construct.Initial {c} {ℓ}
```
MatthewDaggitt, line 19: It really confused me where `c` and `l` were being used. Can we move imports that rely on module parameters between the normal inputs and the variable declarations? (Might be good to add to the style guide?)

src/Algebra/Morphism/Construct/Initial.agda
```diff
@@ -0,0 +1,73 @@
...
+  homo : Homomorphic₂ zero ℤero._∙_ _∙_
+  homo ()
+
+  injective : Injective ℤero._≈_ _≈_ zero
+  injective {x = ()}
+
+------------------------------------------------------------------------
+-- Magma
+
+module _ (M : Magma m ℓm) where
```
MatthewDaggitt, line 57: Can't/shouldn't this and all the similar modules in `Terminal` be take raw versions of the bundle as parameters instead?

You could then save a lot of lines (and IMO maybe make it easier to read?) by opening `MagmaMorphisms` etc. at the top of the file without a parameter?

src/Algebra/Morphism/Construct/Terminal.agda
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+------------------------------------------------------------------------
+-- Ring
+
+module _ (R : Ring a ℓa) where
+
+  private module R = Ring R
+  open RingMorphisms R.rawRing rawRing
+  open UniqueMorphism R.+-rawMonoid
```
MatthewDaggitt, line 141: Missing empty line?

src/Algebra/Morphism/Construct/Terminal.agda
```diff
@@ -0,0 +1,146 @@
...
+        ; module RingMorphisms
+        )
+open import Data.Product.Base using (_,_)
+open import Function.Definitions using (StrictlySurjective)
+import Relation.Binary.Morphism.Definitions as Definitions
+open import Relation.Binary.Morphism.Structures
+
+private
+  variable
+    a ℓa : Level
```
MatthewDaggitt, line 38: For better or worse the current standard name is `ℓ` in the library.

jamesmckinna: Yes, but the trouble is, `c` and `ℓ` are already taken by the top-level parametrisation of the module...

src/Algebra/Morphism/Construct/Initial.agda
jamesmckinna: Hmmm... that's an interesting/tricky one. Clearly, for each of `Initial` and `Terminal`, there is the underlying module parameterised on a `Raw` bundle.

For whatever reason (?), I associate `Construct`s as being parametrised on the *bundle*, proxying for the 'name of the category'. But you're quite correct that all the other modules in that directory only parametrise on the `Raw` thing (?), so for consistency's sake, I should revise as you suggest.

jamesmckinna: Not sure I understand how to save the lines though. Will push the re-parametrisation, and then maybe you could offer me a steer on that?

MatthewDaggitt: I've rearranged it to get rid of almost 70 lines of boilerplate. Let me know if you think I've lost anything crucial?

JacquesCarette (approved):

src/Algebra/Morphism/Construct/Initial.agda
jamesmckinna: Looks good, modulo it now being harder for me to read, but that's a common problem! 

Your desire for brevity contrasts with what I regard as mine for (greater!?) clarity, but it does seem to simply be a contrast in styles. 

Happy to let yours carry the day. Not sure how well I can internalise it though! ;-)

jamesmckinna:
Happy to merge, and will try to abstain in future from my inclination towards boilerplate/bureaucracy...

PR #2300, opened 2024-02-23
Add `Effect.Foldable` and `Data.List.Effectful.Foldable` implementation

jamesmckinna:
Alternative approach to #1281 with specialised version(s) for `Data.List`. May solve #1099 ?

Currently DRAFT, for discussion, esp. wrt:
* naming
* location (currently under `Effect.*`, since ultimately it will be based off `RawFunctor`)
* DONE: properties as in #1281


Taneb:
Two worries with this approach:

* Can you define a `RawFoldable` for `Vec`?
* Can you show that FoldMap is monoid homomorphism for _commutative_ monoids from Lists up to permutations?

jamesmckinna:
@Taneb thanks for the rapid feedback!
> Two worries with this approach:

It's definitely DRAFT, but I appreciate the nudge to think about extensibility of the design. 

>     * Can you define a `RawFoldable` for `Vec`?

In the first instance (sic), surely only a *non-dependent* one (see latest commits). In general, I might expect the design to evolve to embrace:
*  `Effect.Indexed.Foldable` , based on the introduction of
*  `Algebra.Graded.*`, so that I can introduce *graded* monoids, and then generalise `foldMap` to take such a thing, and then use that `Vec; []; _++_` is graded over `ℕ; 0; _+_`...?

>     * Can you show that FoldMap is monoid homomorphism for _commutative_ monoids from Lists up to permutations?

For sure, I've given much less thought to the `Setoid` structure on the `(Raw)Monoid` argument being passed in, nor to the way properties of a `RawFoldable*` might cash out as an `IsFoldable` 'structure`. But I felt I had some liberty here, given that we (still!) haven't stabilised on how to go beyond `RawFunctor` to `Functor` with laws...

jamesmckinna:
Reply about the comment on `product` on the PR #1969 ...

jamesmckinna:
Two further comments on this evolving PR:
* I think it does serve a different/distinct purpose to that argued for by @Taneb in #1281 
* I think it is of independent interest/utility (even if following the `haskell` model a bit too closely), but conceivably all the functionality (for `List` ;-)) can be realised by suitable combinations of the monadic/applicative machinery in `Data.List.Effectful` precisely because `List`s are one version of the underlying carrier of the free `Monoid`... in which case it would make sense to abandon this PR in favour of enhancing the content there.

Regarding different potential equalities on `List`, however, at the moment, because the only structure considered is `Raw` (ie no non-trivial equations on either source or target of the `foldMap` function, in particular what it might be taken to be a homomorphism between), I don't rule out that we couldn't also tackle @Taneb 's desire to handle `(Idempotent)CommutativeMonoid`s as well. But I'm (still) inclined to think that that might better be handled by a systematic treatment of `Free` constructions cf. #1962 / #1954 and the associated discussions there.

src/Data/Vec/Effectful/Foldable.agda
```diff
@@ -0,0 +1,55 @@
...
+
+  foldMap : (A → Carrier) → Vec A n → Carrier
+  foldMap f []       = ε
+  foldMap f (x ∷ xs) = f x ∙ foldMap f xs
+
+------------------------------------------------------------------------
+-- Basic instance: using supplied defaults
+
+foldableWithDefaults : RawFoldableWithDefaults {a} (λ A → Vec A n)
+foldableWithDefaults = record { foldMap = λ M → foldMap M }
```
MatthewDaggitt, line 43: Why would we want the version with defaults?

MatthewDaggitt (commented):
I guess my general comment is that this really is the fork in the road point. Do we want this to use instance arguments ala Haskell or do we not want to use instance arguments?

src/Data/Vec/Effectful/Foldable.agda
jamesmckinna: Just for `Vec`, or in general?
Re: the latter, I suppose I should have had a `Foldable.Biased` module instead?

jamesmckinna:
As to the "fork in the road", I hadn't intended to force such an issue (not sure I'd realised it might even be such). 

Given that progress on `Algebra.Free` had stalled, which I had originally thought would be the solution to the original issue, I had thought that this was/might be a comparatively much lower-hanging proposal... if not an ideal one. 

DRAFT status means we can/should have the discussion before proceeding... interested in seeing more!

JacquesCarette:
What am I missing? I'm seeing the word 'instance' all over the comments (and in the review) but none in the code?

jespercockx:
> ... That said, it seems as though we have turned away from using actual {{...}} instance-based dictionary resolution for instantiating (sic) 'type-classes' in this style (there's a comment from @jespercockx somewhere on the main issue tracker to that effect) so perhaps this usage should be deprecated/avoided?

AFAIK the policy is to define record values that can be used as instances but do not actually mark them with the `instance` keyword in the main modules. Where it is useful, there can then also be a separate `.Instances` module that re-exports the relevant record values as proper instances.

jamesmckinna:
Thanks @jespercockx !

JacquesCarette:
> My bad, or at least, my possible misuse of the term-of-art when I should say something else, like 'dictionary'. In my mind, but not, I think (!?), unique to me, values of record type corresponding to (structures corresponding to) haskell type-classes, with such values corresponding to (dictionaries associated to) Haskell instances of such classes, get called 'instance's informally...

My take on this (with credit to Conor for making some of this clear): Haskell confuses two things, the idea of an interface, and the algorithm for giving implementations of that interface by *search*. So Haskell's typeclasses are the only mechanism for creating interfaces, which then push you into implicit resolution for implementations. That begets orphans, useless `newtype`, etc. The apparent convenience eventually engenders more pain once you reach a large enough scale.

In Agda, record types are rich enough to act as interfaces. So while it is indeed tempting to call an actual record an 'instance', as Agda also has Haskell-style typeclasses, *that* just leads to confusion. So I think Agda-speak has evolved to only call 'instance' those things that use `instance` explicitly.

JacquesCarette (approved):
Looking good.

gallais (approved):
Looks great!

jamesmckinna:
Not sure what's been going wrong with this merge?

src/Effect/Foldable.agda
```diff
@@ -0,0 +1,77 @@
...
+open import Algebra.Bundles.Raw using (RawMonoid)
+open import Algebra.Bundles using (Monoid)
+import Algebra.Construct.Flip.Op as Op
+
+open import Data.List.Base as List using (List; [_]; _++_)
+
+open import Effect.Functor as Fun using (RawFunctor)
+
+open import Function.Base using (id; flip)
+open import Function.Endomorphism.Propositional using (∘-id-monoid)
```
jamesmckinna, line 22: Something went wrong here... should be `Endo` since #2342 but my commit history on this branch had got out of sync somehow?

jamesmckinna: Hopefully fixed now!

PR #2301, opened 2024-02-26
Refactor proofs from #2023

jamesmckinna:
```quote
I would be _strongly_ tempted to refactor these proofs, either in this PR, or subsequently, in terms of properties of general binary relations:
* any reflexive relation `R` (of the correct type) satisfies `S ⇒ R ; S` and `T ⇒ T ; R`;
* any transitive relation `T` satisfies `T ; T ⇒ T`;
* `Setoid`/`IsEquivalence`, `Preorder`/`IsPreorder` satisfy reflexivity and transitivity: QED

Specialising to `Setoid` here seems ... a missed opportunity.
```
_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/pull/2023#discussion_r1501798611_
            
For the cost of only one more lemma in `Relation.Binary.Construct.Composition`, the second of the above ...

Outstanding issue: what's the Agda input sequence for `_;_`: the character "\;" (backslash-;) seems not quite right? (style-guide note?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,28 +141,28 @@ Additions to existing modules
 
 * In `Data.List.Relation.Ternary.Appending.Setoid.Properties`:
   ```agda
-  through→ : ∃[ xs ] Pointwise _≈_ as xs × Appending xs bs cs → 
+  through→ : ∃[ xs ] Pointwise _≈_ as xs × Appending xs bs cs →
              ∃[ ys ] Appending as bs ys × Pointwise _≈_ ys cs
-  through← : ∃[ ys ] Appending as bs ys × Pointwise _≈_ ys cs → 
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
> Outstanding issue: what's the Agda input sequence for _;_:

Don't know! I copied and pasted!

MatthewDaggitt (approved):

Taneb:
`;` is (according to agda-mode's "Information about the character at point" tool) a Greek question mark, which is written with `\;`, but there's a dozen characters with that and it's the 11th in the list.

I don't think we should be using that character, it's identical to the semicolon graphically (it's even normalized to the semicolon by Unicode) and bound to cause confusion. But I don't have a suggestion for what to use instead.

PR #2307, opened 2024-03-03
Refactor `Data.Integer.Divisibility.Signed`

jamesmckinna:
This is a sequel to #2294 mopping up the refactoring of `Data.Integer.Divisibility.Signed` suggested there.

NB. 
* `Data.Integer.Divisibility` is still a 'minimal' collection of properties based on the underlying `Data.Nat.Divisibility`, so this latter module has still to be imported in `Data.Integer.Divisibility.Signed`, rather than providing a 'clean' API via `Data.Integer.Divisibility`...
* Additions to `Data.Integer.Properties` also expose at least one deprecation/refactoring opportunity, between `abs-*` and `∣i*j∣≡∣i∣*∣j∣`, preferring the former over the latter, but not as yet doing the deprecation, in the interests of minimal impact (separate PR?)...
* Possibility to refactor `◃-distrib-*`?
* Removes one use of `renaming` on `import`, but `style-guide` recommendation in separate PR #2308 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,8 +134,8 @@ Additions to existing modules
 
 * In `Data.Integer.Properties`:
   ```agda
-  ◃-nonZero : {{_ : ℕ.NonZero n}} → NonZero (s ◃ n)
-  sign-*    : {{NonZero (i * j)}} → sign (i * j) ≡ sign i Sign.* sign j
+  ◃-nonZero : .{{_ : ℕ.NonZero n}} → NonZero (s ◃ n)
+  sign-*    : .{{NonZero (i * j)}} → sign (i * j) ≡ sign i Sign.* sign j
   i*j≢0     : .{{_ : NonZero i}} .{{_ : NonZero j}} → NonZero (i * j)
   i*j≢0⇒i≢0 : .{{NonZero (i * j)}} → NonZero i
   i*j≢0⇒j≢0 : .{{NonZero (i * j)}} → NonZero j
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,8 +137,6 @@ Additions to existing modules
   ◃-nonZero : .{{_ : ℕ.NonZero n}} → NonZero (s ◃ n)
   sign-*    : .{{NonZero (i * j)}} → sign (i * j) ≡ sign i Sign.* sign j
   i*j≢0     : .{{_ : NonZero i}} .{{_ : NonZero j}} → NonZero (i * j)
-  i*j≢0⇒i≢0 : .{{NonZero (i * j)}} → NonZero i
-  i*j≢0⇒j≢0 : .{{NonZero (i * j)}} → NonZero j
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,9 +127,18 @@ Additions to existing modules
   nonZeroIndex : Fin n → ℕ.NonZero n
   ```
 
-* In `Data.Integer.Divisisbility`: introduce `divides` as an explicit pattern synonym
+* In `Data.Integer.Divisibility`: introduce `divides` as an explicit pattern synonym
   ```agda
   pattern divides k eq = Data.Nat.Divisibility.divides k eq
+  ```
+
+* In `Data.Integer.Properties`:
+  ```agda
+  ◃-nonZero : {{_ : ℕ.NonZero n}} → NonZero (s ◃ n)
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):

src/Data/Integer/Properties.agda
```diff
@@ -1594,6 +1598,9 @@ private
 abs-* : ℤtoℕ.Homomorphic₂ ∣_∣ _*_ ℕ._*_
 abs-* i j = abs-◃ _ _
 
+sign-* : ∀ i j → {{NonZero (i * j)}} → sign (i * j) ≡ sign i Sign.* sign j
```
MatthewDaggitt, line 1601: Missing irrelevance annotation here?

CHANGELOG.md
```diff
@@ -127,11 +127,20 @@ Additions to existing modules
...
-* In `Data.Integer.Divisisbility`: introduce `divides` as an explicit pattern synonym
+* In `Data.Integer.Divisibility`: introduce `divides` as an explicit pattern synonym
   ```agda
   pattern divides k eq = Data.Nat.Divisibility.divides k eq
   ```
 
+* In `Data.Integer.Properties`:
+  ```agda
+  ◃-nonZero : {{_ : ℕ.NonZero n}} → NonZero (s ◃ n)
+  sign-*    : {{NonZero (i * j)}} → sign (i * j) ≡ sign i Sign.* sign j
```
MatthewDaggitt, line 138: Missing irrelevance annotation on the two lines above?

src/Data/Integer/Properties.agda
```diff
@@ -1631,6 +1638,15 @@ i*j≡0⇒i≡0∨j≡0 i p with ℕ.m*n≡0⇒m≡0∨n≡0 ∣ i ∣ (abs-cong
 ... | inj₁ ∣i∣≡0 = inj₁ (∣i∣≡0⇒i≡0 ∣i∣≡0)
 ... | inj₂ ∣j∣≡0 = inj₂ (∣i∣≡0⇒i≡0 ∣j∣≡0)
 
+i*j≢0 : ∀ i j .{{_ : NonZero i}} .{{_ : NonZero j}} → NonZero (i * j)
+i*j≢0 i j rewrite abs-* i j = ℕ.m*n≢0 ∣ i ∣ ∣ j ∣
+
+i*j≢0⇒i≢0 : ∀ i {j} → .{{NonZero (i * j)}} → NonZero i
```
MatthewDaggitt, line 1644: I am doubtful whether this and the lemma below will actually work for instance search, because `_*_` is not a constructor and therefore Agda won't be able to infer what `i` and `j` are...

Do we have any uses for these? Both on concrete and abstract variables?

I think one solution might be make `j` explicit, but testing is required.

The easy solution is just not to include them as they're not actually used currently right?

src/Data/Integer/Properties.agda
jamesmckinna: Done

CHANGELOG.md
jamesmckinna: Done

src/Data/Integer/Properties.agda
jamesmckinna: Lots of interesting commentary here.

Firstly, I introduced the three of them by analogy with the corresponding lemmas for `Nat`, to which they obviously delegate, and those haven't (so far!) been subject to the same criticism, and which have been quite heavily used (I think? 21 times after some mindless `grep`ping). Maybe need to revisit those?

Secondly, the parametrisation is identical to that in `Data.Nat.Properties`, and while I agree, none of these left-biased operations can infer their first argument (`Nat._+_`, `Nat._*_`, `List._++_`, etc.), Agda (usually) can infer the second given the first... maybe `Integer` is different? But since the definition of `_*_` itself delegates to both `Sign` and `Nat` multiplication, I had been supposing, *mutatis mutandis*...

Thirdly, the new lemma `sign-*` *does* have an irrelevant, instance) argument of the form `.{{NonZero (i * j)}}` so I suppose I had been conjecturing that whenever it gets used, there will typically be the need either to discharge it by delegation to `.{{NonZero i}}` and `.{{NonZero j}}`, or else to destruct it to get at those pieces (motivated by your criticism of #2230 I hadn't wanted to commit the user to any particular *a priori* analysis of how `.{{NonZero (i * j)}}` might arise...)

Finally, you're correct that no uses (yet!?), whereas there is the one use of `i*j≢0` in `Data.Integer.Divisibility.Signed`.

Conclusion: happy to remove, but with some nagging feeling that we might want them one day? In any case, YAGNI, for now at least. Will remove.

PR #2312, opened 2024-03-08
Tidy up functional vector permutation #2066

jamesmckinna:
This is a largely cosmetic coda to #2066 : 
* Completes the `Properties` by adding structure and bundle (NB transitivity has horrible indexed inference mess)
* Sorts out alignment, standardises `variable` names
* Cleans up `CHANGELOG`

Apologies to all for not catching this earlier in review/comments on the earlier PR.

### DIFF
--- a/src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
+++ b/src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
@@ -40,7 +40,7 @@ proj₁ (↭-sym (xs↭ys , _)) = flip xs↭ys
 proj₂ (↭-sym {x = xs} {ys} (xs↭ys , xs↭ys≡)) i = begin
   ys (flip xs↭ys ⟨$⟩ʳ i)              ≡⟨ xs↭ys≡ _ ⟨
   xs (xs↭ys ⟨$⟩ʳ (flip xs↭ys ⟨$⟩ʳ i)) ≡⟨ cong xs (inverseʳ xs↭ys) ⟩
-  xs i ∎
+  xs i                                ∎
   where open ≡-Reasoning
 
 ↭-trans : Transitive (Vector A) _↭_
--- a/src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
+++ b/src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
@@ -15,17 +15,19 @@ open import Data.Fin.Permutation using (id; flip; _⟨$⟩ʳ_; inverseʳ; _∘�
 open import Data.Vec.Functional
 open import Data.Vec.Functional.Relation.Binary.Permutation
 open import Relation.Binary.PropositionalEquality
-  using (refl; trans; _≡_; cong; module ≡-Reasoning)
-open import Relation.Binary.Indexed.Heterogeneous.Definitions
-
-open ≡-Reasoning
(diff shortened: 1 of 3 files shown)


src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
```diff
@@ -36,10 +38,32 @@ private
...
 proj₂ (↭-trans (_ , xs↭ys) (_ , ys↭zs)) _ = trans (xs↭ys _) (ys↭zs _)
+
+------------------------------------------------------------------------
+-- Structure
+
+isIndexedEquivalence : IsIndexedEquivalence (Vector A) _↭_
+isIndexedEquivalence {A = A} = record
+  { refl = ↭-refl
+  ; sym = ↭-sym
+  ; trans = λ {n₁ n₂ n₃} {xs : Vector A n₁} {ys : Vector A n₂} {zs : Vector A n₃}
```
JacquesCarette, line 57: This looks very odd - defining `i` twice? I guess it is there because Agda can't infer all the implicits?

JacquesCarette (changes requested):

src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
jamesmckinna: Indeed, Agda is maddeningly obscure if you simply try to define
```agda
; trans = ↭-trans
```
because it can't easily infer the implicit functional arguments which are the `Vector A`s `xs`, `ys`, `zs`. And they, in turn, because `Relation.Binary.Indexed.Heterogeneous.Definitions` uses full generalised transitivity, need distinct length indices...

Then there's the `i` telescopic binding, seemingly appearing twice: the first is because the `Indexed` machinery expects an implicit index `i` at which to begin a term of type generalised transitivity (I think we could make a renaming change in `Relation.Binary.Indexed.Heterogeneous.Definitions` to tweak this binding), but then the second (distinct!) binding arises because while in `Relation.Binary.Definitions` (to which the above module delegates), `Reflexivity`, `Symmetry` use `x` and `y` for their implicit arguments, a legacy we have *never* attempted to change is that `Transitivity` uses `i`, `j`, `k` for *its* implicit arguments!!! Frankly, it's a total nightmare, and if I could face the knock-on viscosity of changing `Relation.Binary.Definitions.Trans(itivity)` to use `x`, `y`, `z` then I would...
... the only saving grace is that Agda accepts (and expects) the two telescopic bindings to be given separately, even though they ostensibly have the same name. But as a ux 'bug', it takes some beating as to its 'wickedness'. 

Honestly, if you can figure out a simpler version/solution, please show me!

JacquesCarette: I can see why Agda can't do the full inference, I think. Yes, changing `Transitivity` to `x`, `y`, `z` might be a good idea.

But now I see that such a change shouldn't hold up this one.

JacquesCarette (approved):

src/Data/Vec/Functional/Relation/Binary/Permutation/Properties.agda
jamesmckinna: Thanks for your understanding!

Taneb (approved):

PR #2313, opened 2024-03-08
Tidy up `README` imports #2280

jamesmckinna:
Brings the qualified imports into line with the emerging style-guidelines in line with #2280 .

Exceptions:
* `README.Data.List.Relation.Binary.Equality` (which leaves holes open, moreover!?)
* `README.Data.List.Membership` (not clear how best to rename the qualified import names)
* `README.Text.Tabular`

### DIFF
--- a/doc/README/Data/Container/FreeMonad.agda
+++ b/doc/README/Data/Container/FreeMonad.agda
@@ -16,12 +16,12 @@ open import Data.Unit
 open import Data.Bool.Base using (Bool; true)
 open import Data.Nat
 open import Data.Sum.Base using (inj₁; inj₂)
-open import Data.Product.Base renaming (_×_ to _⟨×⟩_)
+open import Data.Product.Base as Product hiding (_×_)
 open import Data.Container using (Container; _▷_)
 open import Data.Container.Combinator
-open import Data.Container.FreeMonad as FreeMonad
+open import Data.Container.FreeMonad
 open import Data.W
-open import Relation.Binary.PropositionalEquality as P
+open import Relation.Binary.PropositionalEquality as ≡
 
 ------------------------------------------------------------------------
 -- Defining the signature of an effect and building trees describing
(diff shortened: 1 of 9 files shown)


Taneb (approved):

JacquesCarette (approved):
I'm not particularly fond of the reformatting that takes the mildly wide equational proofs and makes them more 'vertical'. I would much rather those stayed as they were.

But there's a lot of other good cleanup here, so I won't hold things up just for that. Others can feel free to disagree!

jamesmckinna:
Re: @JacquesCarette 's comment about 'horizontal' vs. 'vertical' layout of Feijen-style reasoning... 

Unfortunately, this is self-inflicted. Changing `L` to `List` as the qualified name made the lines *really* long, moreover for me at least, very hard to read. As an Algebra of Programming person, I'm used to the more generous/expensive 'vertical' style, not least because it permits attention to be focused/localised on the reasons for a rewrite in the one line, while also separating out the individual terms for similar consideration on their own line(s).

In a library module, I might not care, but in a `README`, given that it is affording documentary/tutorial description, I though the the extra space was not actually harmful, if anything an improvement. But YMMV... potato/aardappel ;-)

JacquesCarette:
Suggestion: create a local alias `L` for `List` for the scope of those proofs?

jamesmckinna:
Ach! That suggestion is tantamount to the original qualified import`as L`!?

JacquesCarette:
Local vs global...

doc/README/Function/Reasoning.agda
```diff
@@ -39,7 +39,7 @@ open import Data.String.Base as String using (String; toList; fromList)
 open import Data.String.Properties as String using (_==_)
 open import Function.Base using (_∘_)
 open import Data.Bool hiding (_≤?_)
-open import Data.Product.Base as P using (_×_; <_,_>; uncurry; proj₁)
+open import Data.Product.Base as Product using (_×_; <_,_>; uncurry; proj₁)
```
MatthewDaggitt, line 42: Suggests this `as` is not needed

doc/README/Data/Container/FreeMonad.agda
```diff
@@ -51,13 +51,13 @@ prog =
   where
   open RawMonad monad using (_>>_)
 
-runState : {S X : Set} → State S ⋆ X → (S → X ⟨×⟩ S)
+runState : {S X : Set} → State S ⋆ X → (S → X Product.× S)
```
MatthewDaggitt, line 54: Why do we actually need to qualify this? I'm not aware of any other clashing names in the library off the top my head?

MatthewDaggitt (changes requested):
Otherwise looks great. I'm agnostic as to the layout of the proofs...

doc/README/Data/Container/FreeMonad.agda
jamesmckinna: `Data.Container.Combinator._×_` clashes on import of that module. 
But it is enough to hide *that* binding instead... (although that's not possible in the *definition* of the container combinators, but that's neither here nor there in this `README` instance...)

jamesmckinna: Duly hidden.

MatthewDaggitt (approved):

PR #2314, opened 2024-03-09
make `mkRawMonad` and `mkRawApplicative` universe-polymorphic

gshen42:
A simple change. I've run `make test` and everything type checks.

### DIFF
--- a/src/Data/Product/Effectful/Left.agda
+++ b/src/Data/Product/Effectful/Left.agda
@@ -49,7 +49,7 @@ monad = record
   }
 
 -- The monad instance also requires some mucking about with universe levels.
-monadT : RawMonadT (_∘′ Productₗ)
+monadT : ∀ {ℓ} → RawMonadT {g₁ = ℓ} (_∘′ Productₗ)
 monadT M = record
   { lift = (ε ,_) <$>_
   ; rawMonad = mkRawMonad _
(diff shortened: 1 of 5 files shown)


gallais (approved):

JacquesCarette:
Unfortunately, `make test` missed something as the CI says:
```
/home/runner/work/agda-stdlib/agda-stdlib/src/Data/Product/Effectful/Examples.agda:14,1-62,37
Unsolved metas at the following locations:
  /home/runner/work/agda-stdlib/agda-stdlib/src/Data/Product/Effectful/Left.agda:52,10-19
  import Data.Product.Effectful.Left as .#Data.Product.Effectful.Left-16184305951820099816
```

jamesmckinna:
@gallais is the expert on (these particular instances of) universe polymorphism, and I seem to recall that there was a problem with `(Raw)Applicative` ie it may not be a CI problem?

JacquesCarette:
He is indeed the expert. As it turns out, I'd given a try at this a few years back and the problem then was that `Applicative` was defined in terms of the indexed version, and that would have needed some kind of level-to-level function. The better solution would have been to do a direct definition, but that was too backwards incompatible (at the time).

JacquesCarette (approved):
Looks good - let's hope the CI works this time.

Issue #2311, opened 2024-03-05
Refactor to support multiple definitions of `Data.List.Relation.Binary.Permutation`


jamesmckinna:
Recent work by 
* [Choudhury and Fiore](https://entics.episciences.org/10492/pdf), on the free commutative monoid in HoTT, and
* [Kupke, Forsberg, and Watters](https://personal.cis.strath.ac.uk/fredrik.nordvall-forsberg/papers/freshlists_aplas2023.pdf) on fresh lists

has looked at lists-modulo-permutation, with a view to establishing that (`Setoid`) structure as (a representation of) (the carrier of) the free commutative monoid, with the former alluding to difficulties in establishing such results for the `stdlib` definition of permutation, and the latter, IIUC, by observing that fresh/sorted lists, by providing a constructive representative of the equivalence-classes-modulo makes such difficulties more tractable, at the cost of a(nother) new representation (but one already in the library, and interesting in its own right). 

Each of the above, IMHO, fails to tackle the elephant in the room (incl. `stdlib` ;-), alas, by taking a prefix-closed, transitive relation which can swap only the *top* two elements of a list; it's *very* fiddly, and requires transitivity to be built in), namely that there are two 'natural' definitions of the relation, for which it is interesting (and useful/computationally relevant!) to prove the equivalence between them (something already done, but lost in the archaeology of my PhD thesis back in 1992), namely as:
* *declaratively* the least *congruence* for `[]`, `_∷_` (and hence for `_++_`) which 'makes `_++_` commutative', ie relates `as ++ bs` and `bs ++ as`
* *algorithmically* the least relation such that `[]` is related to itself, and that if `cs` is related to `as ++ bs`, then `c ∷ cs` is related to `as ++ c ∷ bs` (whence the correctness of insertion sort)

I've recently come across the following definition, due, which I learnt from Ralf Hinze (UPDATED: apparently the inspiration for him came from Choudhury and Fiore's paper above), which may easily be shown equivalent to each of these definitions (NB in the `Propositional` case, I'm still looking at the `Setoid` generalisation), as the least relation such that:
* `[]` is related to itself
* if `as` is related to `bs`, then `a ∷ as` is related to `a ∷ bs`
(so that we have left-prefix-closure for `_++_`, if not full congruence, straightaway)
[...]

PR #2317, opened 2024-03-12
Refactor `Data.List.Relation.Binary.Permutation.*`

jamesmckinna:
WORK IN PROGRESS: do not review!

Of independent interest to the various competing proposals in #2311 for the 'right' representation, as it attempts to decouple the main properties of the relation from any particular notion. In particular, the isolation of the dependencies (and improved implementation UPDATED can't seem to get the syntax declarations to work for the specialised `Propositional` version of `PermutationReasoning`!?) of the `Reasoning` module, together with a 'better' smart constructor for transitivity in the current implementation(s), as well as a lot of nitpicky streamlining. 

NEW: have removed all dependency on well-founded induction on `steps` in favour of a (structurally inductive proof of) better `↭-split` lemma which uses the fact that `Permutation R` absorbs `Pointwise R` steps provided `R` is `Transitive`.

Currently in DRAFT, but aims to, and is most of the way towards:
* redefining `Propositional` as the `setoid` instance of `Setoid`, ditto. `Properties`
* refactoring `Setoid` and `Setoid.Properties` to localise all uses of induction on the `Homogeneous` representation to properties proved in general in `Homogeneous` (thereby allowing alternative representations)
* back-porting lemmas and proofs from `Propositional` to `Setoid` or even `Homogeneous`, incl.
* I have managed to get hacky/kludge proofs to work, but there is something more systematic to investigate about the higher-dimensional structure of proofs about `Pointwise` and `Permutation` lurking offstage...!?

Outstanding issues:
* fix the `syntax` declarations for `↭-Reasoning` (esp. for `<<⟨_⟩`? see #2319 )
* should the `Properties` module(s) automatically re-export the corresponding '`Base`' module, to avoid having to explicitly import that separately?

TODO:
* `CHANGELOG` ;-)
* last missing implementations from `Propositional` (some fiddliness here, plus some deprecation opportunities once things have been suitably generalised) UPDATED only
```agda
 ↭-map-inv : List.map f xs ↭′ vs → ∃ λ ys → vs ≡ List.map f ys × xs ↭ ys
  ```
remains
* refactor `Homogeneous` into `Core`/`Base` and `Properties` (currently a total ⟪casino⟫, esp. wrt dependencies)
* refactor `Setoid.Properties` into `Core` (representation dependent; to be discussed as the best way to achieve such modularisation for (each of) the representation(s)) and `Properties` depending on that component
* exporting induction principles while keeping constructors hidden?

Etc. To be continued...

### DIFF
--- a/src/Data/List/Relation/Ternary/Interleaving/Propositional.agda
+++ b/src/Data/List/Relation/Ternary/Interleaving/Propositional.agda
@@ -34,8 +34,8 @@ pattern consʳ xs = refl ∷ʳ xs
 -- New combinators
 
 toPermutation : ∀ {l r as} → Interleaving l r as → as ↭ l ++ r
-toPermutation []         = Perm.↭-refl
-toPermutation (consˡ sp) = Perm.↭-prep _ (toPermutation sp)
+toPermutation []         = Perm.refl
+toPermutation (consˡ sp) = Perm.prep _ (toPermutation sp)
 toPermutation {l} {r ∷ rs} {a ∷ as} (consʳ sp) = begin
   a ∷ as       <⟨ toPermutation sp ⟩
   a ∷ l ++ rs  ↭⟨ shift a l rs ⟨

### CODE REVIEW

File: doc/README/Data/List/Relation/Binary/Permutation.agda [hunk 1]
```diff
@@ -48,10 +48,21 @@ open PermutationReasoning
...
 lem₂ = begin
-  1 ∷ 2 ∷ 3 ∷ []  ↭⟨ prep 1 (swap 2 3 refl) ⟩
-  1 ∷ 3 ∷ 2 ∷ []  ↭⟨ swap 1 3 refl ⟩
+  1 ∷ 2 ∷ 3 ∷ []  ↭⟨ ↭-prep 1 (↭-swap 2 3 ↭-refl) ⟩
+  1 ∷ 3 ∷ 2 ∷ []  ↭⟨ ↭-swap 1 3 ↭-refl ⟩
   3 ∷ 1 ∷ 2 ∷ []  ∎
 
+-- In practice it is further useful to extend `PermutationReasoning`
+-- with specialised combinators `<⟨_⟩` and `<<⟨_⟩` to support the
+-- distinguished use of the `↭-prep ` and `↭-prep ` steps, allowing
```
Taneb, line 57: Should one of these `↭-prep`s be a `↭-swap`?

jamesmckinna: Do you know why the `Propositional` syntax declarations don't seem to parse properly when used in `lem₂ᵣ`?

src/Relation/Binary/PropositionalEquality/Properties.agda
```diff
@@ -93,6 +93,9 @@ cong-∘ : ∀ {x y : A} {f : B → C} {g : A → B} (p : x ≡ y) →
          cong (f ∘ g) p ≡ cong f (cong g p)
 cong-∘ refl = refl
 
+sym-involutive : ∀ {x y : A} (p : x ≡ y) → sym (sym p) ≡ p
+sym-involutive refl = refl
```
jamesmckinna, line 97: Amazed this wasn't already in the library!

jamesmckinna: But have now made it local to `Data.List.Relation.Binary.Permutation.Propositional.Properties`, ditto.
```agda
trans-trans-sym : (p : x ≡ y) (q : y ≡ z) → ≡.trans (≡.trans p q) (≡.sym q) ≡ p
trans-trans-sym refl refl = refl
```

MatthewDaggitt (commented):
Before I review this, am I right in understanding there's really two changes here? Firstly moving lots of stuff from `Setoid` to `Heterogeneous` and secondly changing the implementation of `Propositional` to use `Setoid`.

If so then could we do this in two different PRs? In particular the second PR is big breaking change, and therefore there are questions about when it exactly it can be merged in.

jamesmckinna:
Thanks Matthew!
I'm just catching up with #1761 and #1354 which has made me reconsider what's here, and your comment drives the point home. 

One reason for it still being draft is it is still far from ready for review, is that I felt I needed to do both things to get a sense of what might subsequently work. 

Will attempt to refactor this into two more clearly defined pieces... so don't review yet!
(But I did have a specific question about the reasoning combinators, and any steer on that point would be helpful... ;))

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -218,138 +176,34 @@ drop-mid {A = A} {x} ws xs p = drop-mid′ p ws xs refl refl
       _ ∷ _ ∷ xs ++ _   ∎
   drop-mid′ (swap y z p) (y ∷ z ∷ ws) (z ∷ [])     refl refl = begin
       _ ∷ _ ∷ ws ++ _   <<⟨ refl ⟩
+-- *** NB the next step can't be replaced with <⟨ shift _ _ _ ⟨ for some reason? ***
       _ ∷ (_ ∷ ws ++ _) <⟨ ↭-sym (shift _ _ _) ⟩
+-- *** because of parsing problems for that use of the `Reasoning` combinators!? ***
```
jamesmckinna, line 181: This seems to be a rare bug in our `Reasoning` combinators:
* `<⟨ ↭-sym (shift _ _ _) ⟩` and
* `<⟨ shift _ _ _ ⟨`
*ought to have the same semantics here; the first succeeds, but the second fails with a parsing error. Seemingly on `master` as well, so will flag as an issue cf. #2319

jamesmckinna:
Note to self: typechecking appears not to terminate for `Data.List.Relation.Binary.Permutation.Homogeneous.Properties`!?

jamesmckinna:
Stopping here as much of the refactoring of (the properties of `Setoid` as instances of those of) `Homogeneous` is fairly settled now. So I'll now attempt to retrofit that analysis back onto the *old* versions.

jamesmckinna:
Previously @MatthewDaggitt wrote

> Before I review this, am I right in understanding there's really two changes here? Firstly moving lots of stuff from `Setoid` to `Heterogeneous` and secondly changing the implementation of `Propositional` to use `Setoid`.
> 
> If so then could we do this in two different PRs? In particular the second PR is big breaking change, and therefore there are questions about when it exactly it can be merged in.

Please now consider a possible roadmap as follows:
- [ ] (Part I) reconcile the APIs for `Propositional` and `Setoid` in their 'old' versions #2333 ; this is slightly more conservative, and doesn't even go as far as moving everything to `Homogeneous` as here
- [ ] (Part II) refactor via `Homogeneous.Properties.*` and `Propositional.Properties` and `Setoid.Properties` in terms of `Core` (based on `Homogeneous`) and 'derived' based on the respective `Core` properties
- [ ] (Part III) revisit #1761 to then reconcile the two definitions, and with minimised `Core`s via the `pattern` synonym tricks?

MatthewDaggitt:
> Stopping here

Does this mean the draft status should be removed and this PR reviewed?

jamesmckinna:
> > Stopping here
> 
> Does this mean the draft status should be removed and this PR reviewed?

Thanks @MatthewDaggitt for the prompt to return to this. I think the short answer, for the time being, is/ought to be "No":
* merge conflicts ahead of of the v2.1 release are stymying various open PRs, including the first part of this, namely #2333 , which is badged as v2.2
* the above (re-)factorisation into three parts, the first of which above, I had hoped made clear that this PR would (eventually ;-)) get broken up, part-by-part, and I haven't had the brain space to return to the second and/or third parts given that they are (potentially) `breaking`, and moreover, somewhat depend on the eventual fate of the first...

Sound OK to you?

MatthewDaggitt:
Yes, sounds reasonable.

JacquesCarette:
@jamesmckinna if you had help (i.e. someone who could follow your directions on how to push this forward), would you be able to continue work on this?

jamesmckinna:
> @jamesmckinna if you had help (i.e. someone who could follow your directions on how to push this forward), would you be able to continue work on this?

Thanks for the offer! now that #2726 has been merged, there may be a clearer path to a (more radical) approach, because both versions (`Setoid` and `Propositional`) can delegate to a version (`Algorithmic` or `Declarative`) in which the `refl` behaviour can be better handled. But it now requires someone (else?) to stand back, and get a better global picture of where we are, and a vision of where we want to go... which might end up being `breaking`, so v3.0?

PR #2321, opened 2024-03-16
Refactor `Data.List.Relation.Binary.BagAndSetEquality`

jamesmckinna:
Cherry-picked from #2317 :
* (cosmetic) rewrote the proof of `lemma` at the end to make it more readable (to me... ;-));
* (better abstraction) rewrote the proof of `↭⇒∼bag` to exploit existing lemmas in `Data.List.Relation.Binary.Permutation.Propositional.Properties`;
* (cosmetic) use of irrefutable `with` in `∼bag⇒↭`

No `CHANGELOG` required.


MatthewDaggitt (approved):

jamesmckinna:
That's great, thanks @MatthewDaggitt i
(This PR should helps clear away some cruft ahead of resuming work on #1761 ... while keeping the `Permutation` API more abstract here.)

PR #957, opened 2019-10-21
Start design doc

JacquesCarette:
Start documenting things, as per #312 .

Issue #2213, opened 2023-11-25
Documenting the design of `stdlib`


jamesmckinna:
It's come up before (esp. in discussions with @JacquesCarette ) ... but we have ended up documenting a *lot* of quite delicate/subtle design/ux decisions in the library either purely 'locally' in modules themselves, or 'worse', as part of the 'essentially ephemeral' history (*sic*: you have to be _very_ diligent to trace discussions back through the `CHANGELOG` and/or the GitHub issue/PR histories). 

And/Hence future users/developers of the library don't currently have a 'reasonable' place to consult on such things. 

So we should try to develop such a thing, but the prospect of how to lift all of that accumulated design rationale into a reference document... gives me a headache! 

Cf the philosophy underlying the main Agda issue https://github.com/agda/agda/issues/7003

PR #2322, opened 2024-03-17
Some design documentation (here: level polymorphism)

### LINKED ISSUE #312: Documentation on design decisions (re. Level polymorphism)
Is it an explicit design decision that things like
```agda
data ⊥ : Set where
```
are not level polymorphic?

I am guessing yes -- but it would be nice to have this design decision and its rationale documented in some user-facing documentation.

JacquesCarette:
This is a redo of #957 but in the setting of stdlib 2.0.  Closes #312 .


jamesmckinna:
Text looks good, but it's a bit surprising to have a `README` with no actual code in it (as opposed to the Markdown files under `doc`).

I for one would find it helpful if such a `README` were to summarise the discussion on #312 etc. eg by having sub-modules
```agda
module Empty where
  -- open monomorphic `Empty`; what's possible, what's not
module PolymorphicEmpty
  -- etc.
module Unit where
  -- etc. etc.
```

Is that doable/in scope for your PR? (more work, I know, but there's a 'unit of contribution' issue here, as well as whether the originating issue gets a fair shake at being closed by this PR: what *are* the issues at stake?). Failing that, links to follow-up reading with links to the issue tracker discussion/`readthedocs`/other archived material might help the 'level-curious'?

Also: rather than have a `Miscellany`, why not factor this out simply as `README.Design.LevelPolymorphism`?

JacquesCarette:
You're absolutely right, the bar should be higher.

I guess I only looked quickly under doc/ and only found `.agda` files. It does make much more sense to have this as markdown -- which is what it was in the dead PR!

Yes, `README.Design.LevelPolymorphism` is probably better.

MatthewDaggitt:
Or alternatively you could place this documentation in `Data.Unit`/`Data.Empty` themselves? Happy with either!

jamesmckinna:
A possibly related point: `Relation.Unary` defines `Ø` and `U` at vanilla `Set`, rather than polymorphically. 

AFAIU, this is harmless, because all of the relations/properties of `Pred`s are level-polymorphic, but I find myself scratching my head from time to time as to whether or not it is OK in general, or indeed if making the polymorphic choice would actively be a problem, because leading to unsolved metas... and I'd love a doc which lays out the issues for me!

JacquesCarette:
I've updated things to use markdown instead of comments-only `.agda`. As it turns out, `Data.Unit.Base` and `Data.Empty` already documented these issues quite well, so there was no need to add a lot of documentation there. I still think it makes sense to have something in `doc/`.

I'm going to experiment with `Relation.Unary` separately from this.

JacquesCarette:
> A possibly related point: `Relation.Unary` defines `Ø` and `U` at vanilla `Set`, rather than polymorphically.
> 
> AFAIU, this is harmless, because all of the relations/properties of `Pred`s are level-polymorphic, but I find myself scratching my head from time to time as to whether or not it is OK in general, or indeed if making the polymorphic choice would actively be a problem, because leading to unsolved metas... and I'd love a doc which lays out the issues for me!

Tried it: miserable failure. Lots of yellow in `Relation.Unary.Properties` that would be non-trivial to fix.

Taneb:
It'd definitely be worth mentioning this issue with `Relation.Unary` and advertising the existence of `Relation.Unary.Polymorphic` here

jamesmckinna:
> It'd definitely be worth mentioning this issue with `Relation.Unary` and advertising the existence of `Relation.Unary.Polymorphic` here

Thanks @Taneb for the nudge to find what I hadn't realised I'd been looking for!

But definitely a nod to "... Similar remarks apply to the liftings of ... to `Relation.Unary` predicates, with `Relation.Unary.Polymorphic` available when needed." or some such phrasing.

JacquesCarette:
Indeed - will do.

jamesmckinna (approved):
Minimal, helpful, intervention, giving useful notes in-place towards any eventual larger-scale resolution of #2213 ... Suggest we merge!

Taneb (approved):

PR #2324, opened 2024-03-21
Refactor `List.Membership.*` and `List.Relation.Unary.Any`

jamesmckinna:
By contrast with #2323 this adds to the `List.*` API, and extensively refactors proofs about `Membership` and `Any` in the light of the addition of two 'obviously missing' combinators.

Highlights of the refactoring:
* extensive debugging of `import` dependencies throughout
* revision of qualified imports in accordance with the new style guidelines #2280 
* revision of `with` usage in favour of 'irrefutable' style #2123 

The changes are mostly cosmetic, although there were some interesting pain points around eta-conversion (or lack of it!?), wrt strictness in (the structure of) values of dependent `Σ`-types.

### DIFF
--- a/src/Data/List/Membership/Propositional/Properties.agda
+++ b/src/Data/List/Membership/Propositional/Properties.agda
@@ -55,7 +55,7 @@ private
 ------------------------------------------------------------------------
 -- Publicly re-export properties from Core
 
-open import Data.List.Membership.Propositional.Properties.CoreNEW public
+open import Data.List.Membership.Propositional.Properties.Core public
 
 ------------------------------------------------------------------------
 -- Equality
--- a/src/Data/List/Membership/Propositional/Properties.agda
+++ b/src/Data/List/Membership/Propositional/Properties.agda
@@ -9,24 +9,25 @@
 module Data.List.Membership.Propositional.Properties where
 
 open import Algebra using (Op₂; Selective)
-open import Effect.Monad using (RawMonad)
-open import Data.Bool.Base using (Bool; false; true; T)
+open import Data.Bool.Base using (Bool; false; true; if_then_else_)
 open import Data.Fin.Base using (Fin)
 open import Data.List.Base as List
-open import Data.List.Relation.Unary.Any as Any using (Any; here; there)
-open import Data.List.Relation.Unary.Any.Properties
+open import Data.List.Effectful using (monad)
 open import Data.List.Membership.Propositional
 import Data.List.Membership.Setoid.Properties as Membership
 open import Data.List.Relation.Binary.Equality.Propositional
   using (_≋_; ≡⇒≋; ≋⇒≡)
-open import Data.List.Effectful using (monad)
-open import Data.Nat.Base using (ℕ; zero; suc; pred; s≤s; _≤_; _<_; _≤ᵇ_)
+open import Data.List.Relation.Unary.Any as Any using (Any; here; there)
+open import Data.List.Relation.Unary.Any.Properties
+open import Data.Nat.Base using (ℕ; suc; s≤s; _≤_; _<_; _≰_)
 open import Data.Nat.Properties
+  using (suc-injective; m≤n⇒m≤1+n; _≤?_; <⇒≢; ≰⇒>)
 open import Data.Product.Base hiding (map)
 open import Data.Product.Properties using (×-≡,≡↔≡)
 open import Data.Product.Function.NonDependent.Propositional using (_×-cong_)
 import Data.Product.Function.Dependent.Propositional as Σ
 open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂)
+open import Effect.Monad using (RawMonad)
 open import Function.Base
 open import Function.Definitions
 import Function.Related.Propositional as Related
@@ -37,14 +38,12 @@ open import Level using (Level)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Definitions as Binary hiding (Decidable)
 open import Relation.Binary.PropositionalEquality as ≡
-  using (_≡_; _≢_; refl; sym; trans; cong; subst; →-to-⟶; _≗_)
+  using (_≡_; _≢_; refl; sym; trans; cong; resp; →-to-⟶; _≗_)
 import Relation.Binary.Properties.DecTotalOrder as DTOProperties
-open import Relation.Unary using (_⟨×⟩_; Decidable)
-import Relation.Nullary.Reflects as Reflects
+open import Relation.Nullary.Decidable using (Dec; does; yes; no; ¬¬-excluded-middle)
+open import Relation.Nullary.Negation using (¬_; contradiction)
 open import Relation.Nullary.Reflects using (invert)
-open import Relation.Nullary using (¬_; Dec; does; yes; no; _because_)
-open import Relation.Nullary.Negation using (contradiction)
-open import Relation.Nullary.Decidable using (¬¬-excluded-middle)
+open import Relation.Unary using (_⟨×⟩_; Decidable)
 
 private
   open module ListMonad {ℓ} = RawMonad (monad {ℓ = ℓ})
@@ -56,7 +55,7 @@ private
 ------------------------------------------------------------------------
 -- Publicly re-export properties from Core
 
-open import Data.List.Membership.Propositional.Properties.Core public
+open import Data.List.Membership.Propositional.Properties.CoreNEW public
 
 ------------------------------------------------------------------------
 -- Equality
@@ -124,8 +123,9 @@ module _ {v : A} where
   ∈-insert xs = Membership.∈-insert (≡.setoid A) xs refl
 
   ∈-∃++ : ∀ {xs} → v ∈ xs → ∃₂ λ ys zs → xs ≡ ys ++ [ v ] ++ zs
-  ∈-∃++ v∈xs with Membership.∈-∃++ (≡.setoid A) v∈xs
-  ... | ys , zs , _ , refl , eq = ys , zs , ≋⇒≡ eq
+  ∈-∃++ v∈xs
+    with ys , zs , _ , refl , eq ← Membership.∈-∃++ (≡.setoid A) v∈xs
+    = ys , zs , ≋⇒≡ eq
 
 ------------------------------------------------------------------------
 -- concat
@@ -143,8 +143,9 @@ module _ {v : A} where
     Membership.∈-concat⁺′ (≡.setoid A) v∈vs (Any.map ≡⇒≋ vs∈xss)
 
   ∈-concat⁻′ : ∀ xss → v ∈ concat xss → ∃ λ xs → v ∈ xs × xs ∈ xss
-  ∈-concat⁻′ xss v∈c with Membership.∈-concat⁻′ (≡.setoid A) xss v∈c
-  ... | xs , v∈xs , xs∈xss = xs , v∈xs , Any.map ≋⇒≡ xs∈xss
+  ∈-concat⁻′ xss v∈c
+    with xs , v∈xs , xs∈xss ← Membership.∈-concat⁻′ (≡.setoid A) xss v∈c
+    = xs , v∈xs , Any.map ≋⇒≡ xs∈xss
 
   concat-∈↔ : ∀ {xss : List (List A)} →
               (∃ λ xs → v ∈ xs × xs ∈ xss) ↔ v ∈ concat xss
@@ -179,8 +180,9 @@ module _ (f : A → B → C) where
 
 ∈-cartesianProduct⁻ : ∀ xs ys {xy@(x , y) : A × B} →
                       xy ∈ cartesianProduct xs ys → x ∈ xs × y ∈ ys
-∈-cartesianProduct⁻ xs ys xy∈p[xs,ys] with ∈-cartesianProductWith⁻ _,_ xs ys xy∈p[xs,ys]
-... | (x , y , x∈xs , y∈ys , refl) = x∈xs , y∈ys
+∈-cartesianProduct⁻ xs ys xy∈p[xs,ys]
+  with _ , _ , x∈xs , y∈ys , refl ← ∈-cartesianProductWith⁻ _,_ xs ys xy∈p[xs,ys]
+  = x∈xs , y∈ys
 
 ------------------------------------------------------------------------
 -- applyUpTo
@@ -201,8 +203,7 @@ module _ (f : ℕ → A) where
 ∈-upTo⁺ = ∈-applyUpTo⁺ id
 
 ∈-upTo⁻ : ∀ {n i} → i ∈ upTo n → i < n
-∈-upTo⁻ p with ∈-applyUpTo⁻ id p
-... | _ , i<n , refl = i<n
+∈-upTo⁻ p with _ , i<n , refl ← ∈-applyUpTo⁻ id p = i<n
 
 ------------------------------------------------------------------------
 -- applyDownFrom
@@ -223,8 +224,7 @@ module _ (f : ℕ → A) where
 ∈-downFrom⁺ i<n = ∈-applyDownFrom⁺ id i<n
 
 ∈-downFrom⁻ : ∀ {n i} → i ∈ downFrom n → i < n
-∈-downFrom⁻ p with ∈-applyDownFrom⁻ id p
-... | _ , i<n , refl = i<n
+∈-downFrom⁻ p with _ , i<n , refl ← ∈-applyDownFrom⁻ id p = i<n
 
 ------------------------------------------------------------------------
 -- tabulate
@@ -243,10 +243,10 @@ module _ {n} {f : Fin n → A} where
 module _ {p} {P : A → Set p} (P? : Decidable P) where
 
   ∈-filter⁺ : ∀ {x xs} → x ∈ xs → P x → x ∈ filter P? xs
-  ∈-filter⁺ = Membership.∈-filter⁺ (≡.setoid A) P? (≡.subst P)
+  ∈-filter⁺ = Membership.∈-filter⁺ (≡.setoid A) P? (≡.resp P)
 
   ∈-filter⁻ : ∀ {v xs} → v ∈ filter P? xs → v ∈ xs × P v
-  ∈-filter⁻ = Membership.∈-filter⁻ (≡.setoid A) P? (≡.subst P)
+  ∈-filter⁻ = Membership.∈-filter⁻ (≡.setoid A) P? (≡.resp P)
 
 ------------------------------------------------------------------------
 -- derun and deduplicate
@@ -306,7 +306,7 @@ module _ (_≈?_ : Binary.Decidable {A = A} _≡_) where
 ------------------------------------------------------------------------
 -- length
 
-∈-length : ∀ {x : A} {xs} → x ∈ xs → 1 ≤ length xs
+∈-length : ∀ {x : A} {xs} → x ∈ xs → 0 < length xs
 ∈-length = Membership.∈-length (≡.setoid _)
 
 ------------------------------------------------------------------------
@@ -362,28 +362,27 @@ finite inj (x ∷ xs) fᵢ∈x∷xs = ¬¬-excluded-middle helper
   helper (yes (i , fᵢ≡x)) = finite f′-inj xs f′ⱼ∈xs
     where
     f′ : ℕ → _
-    f′ j with does (i ≤? j)
-    ... | true  = f (suc j)
-    ... | false = f j
+    f′ j with i ≤? j
+    ... | yes _ = f (suc j)
+    ... | no  _ = f j
 
     ∈-if-not-i : ∀ {j} → i ≢ j → f j ∈ xs
     ∈-if-not-i i≢j = not-x (i≢j ∘ f-inj ∘ trans fᵢ≡x ∘ sym)
 
-    lemma : ∀ {k j} → i ≤ j → ¬ (i ≤ k) → suc j ≢ k
+    lemma : ∀ {k j} → i ≤ j → i ≰ k → suc j ≢ k
     lemma i≤j i≰1+j refl = i≰1+j (m≤n⇒m≤1+n i≤j)
 
     f′ⱼ∈xs : ∀ j → f′ j ∈ xs
-    f′ⱼ∈xs j with i ≤ᵇ j | Reflects.invert (≤ᵇ-reflects-≤ i j)
-    ... | true  | p = ∈-if-not-i (<⇒≢ (s≤s p))
-    ... | false | p = ∈-if-not-i (<⇒≢ (≰⇒> p) ∘ sym)
+    f′ⱼ∈xs j with i ≤? j
+    ... | yes i≤j = ∈-if-not-i (<⇒≢ (s≤s i≤j))
+    ... | no  i≰j = ∈-if-not-i (<⇒≢ (≰⇒> i≰j) ∘ sym)
 
     f′-injective′ : Injective _≡_ _≡_ f′
-    f′-injective′ {j} {k} eq with i ≤ᵇ j | Reflects.invert (≤ᵇ-reflects-≤ i j)
-                                | i ≤ᵇ k | Reflects.invert (≤ᵇ-reflects-≤ i k)
-    ... | true  | p | true  | q = ≡.cong pred (f-inj eq)
-    ... | true  | p | false | q = contradiction (f-inj eq) (lemma p q)
-    ... | false | p | true  | q = contradiction (f-inj eq) (lemma q p ∘ sym)
-    ... | false | p | false | q = f-inj eq
+    f′-injective′ {j} {k} eq with i ≤? j | i ≤? k
+    ... | yes i≤j | yes i≤k = suc-injective (f-inj eq)
+    ... | yes i≤j | no  i≰k = contradiction (f-inj eq) (lemma i≤j i≰k)
+    ... | no  i≰j | yes i≤k = contradiction (f-inj eq) (lemma i≤k i≰j ∘ sym)
+    ... | no  i≰j | no  i≰k = f-inj eq
 
     f′-inj : ℕ ↣ _
     f′-inj = record
--- a/src/Data/List/Membership/Propositional/Properties/Core.agda
+++ b/src/Data/List/Membership/Propositional/Properties/Core.agda
@@ -13,8 +13,8 @@
 module Data.List.Membership.Propositional.Properties.Core where
 
 open import Data.List.Base using (List)
-open import Data.List.Relation.Unary.Any as Any using (Any; here; there)
 open import Data.List.Membership.Propositional
+open import Data.List.Relation.Unary.Any as Any using (Any; here; there)
 open import Data.Product.Base as Product using (_,_)
 open import Function.Base using (flip; id; _∘_)
 open import Function.Bundles using (_↔_; mk↔ₛ′)
--- a/src/Data/List/Membership/Propositional/Properties/Core.agda
+++ b/src/Data/List/Membership/Propositional/Properties/Core.agda
@@ -62,12 +62,12 @@ module _ {P : Pred A p} where
   ∃∈-Any : ∃∈ P xs → Any P xs
   ∃∈-Any (x , x∈xs , px) = lose {P = P} x∈xs px
 
+  ∃∈-Any∘find : (p : Any P xs) → ∃∈-Any (find p) ≡ p
+  ∃∈-Any∘find p = map∘find p refl
+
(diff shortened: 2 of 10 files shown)


src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -255,24 +263,41 @@ Any-×⁻ pq with Product.map₂ (Product.map₂ find) (find pq)
...
+
+        h : P ⋐ λ x → Any (λ y → (P x) × (Q y)) ys
+        h p = Any.map (p ,_) (lose y∈ys q)
+
+    in  begin
+      Any-×⁺ (Any-×⁻ pq)
+        ≡⟨⟩
+      Any.map h (lose x∈xs p)
+        ≡⟨ map-∘ h (λ z → resp P z p) x∈xs ⟨
+      _
```
JacquesCarette, line 297: If you're going to be this verbose, then this `_` really ought to be spelled out! (Probably the one above in the helper proof as well)

JacquesCarette (changes requested):
Looking like a great refactor - just  a couple of minor wobbles from me.

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Done, but not obvious this illuminates any better compared to the typechecker being able to infer it via `_`... but part of the verbosity was simply not being able to read, much less understand, the existing proofs...

JacquesCarette: The point is that I, as a reader, couldn't easily reconstruct what the typechecker could, so I didn't quite understand what those 2 steps were doing.

src/Data/List/Membership/Setoid.agda
```diff
@@ -48,9 +54,10 @@ mapWith∈ (x ∷ xs) f = f (here refl) ∷ mapWith∈ xs (f ∘ there)
 
 module _ {p} {P : Pred A p} where
 
-  find : ∀ {xs} → Any P xs → ∃ λ x → x ∈ xs × P x
-  find (here px)   = (_ , here refl , px)
+  find : ∀ {xs} → Any P xs → ∃∈ P xs
+  find (here px)   = _ , here refl , px
   find (there pxs) = Product.map id (Product.map there id) (find pxs)
+  -- better? let x , x∈xs , px = find pxs in x , there x∈xs , px
```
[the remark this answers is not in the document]
jamesmckinna: Done. I'd left in the comment in the hope it might get picked up. I think that with the 'better' definition, further simplifications may  be possible, but haven't had the headspace to pursue that.

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -362,28 +361,27 @@ finite inj (x ∷ xs) fᵢ∈x∷xs = ¬¬-excluded-middle helper
   helper (yes (i , fᵢ≡x)) = finite f′-inj xs f′ⱼ∈xs
     where
     f′ : ℕ → _
-    f′ j with does (i ≤? j)
-    ... | true  = f (suc j)
-    ... | false = f j
+    f′ j with i ≤? j
+    ... | yes _ = f (suc j)
+    ... | no  _ = f j
```
MatthewDaggitt, line 366: The last time I looked at the `Dec` set-up this change would mean that we now compute the proof, even when we don't need it. For example with naturals, the boolean decision procedure complexity is `O(max(log(i),log(j)))` because we use built-in arithmetic whereas the proof i s`O(max(i), max(j))` because we use unary arithmetic. 

Is this still the case?

src/Data/List/Membership/Propositional/Properties/Core.agda
```diff
@@ -12,75 +12,73 @@
...
-  rewrite find∘map x∈xs (flip (subst P) p)
-        | find-∈ x∈xs
-        = refl
+  map∘find : (p : Any P xs) → let x , x∈xs , px = find p in
+             {f : (x ≡_) ⊆ P} → f refl ≡ px →
+             Any.map f x∈xs ≡ p
+  map∘find (here  p) hyp = cong here  hyp
+  map∘find (there p) hyp = cong there (map∘find p hyp)
+
+  find∘map : ∀ {Q : Pred A q} {xs} (p : Any P xs) (f : P ⊆ Q) →
```
MatthewDaggitt, line 52: Can remove `{xs}`?

src/Data/List/Membership/Propositional/Properties/Core.agda
```diff
@@ -12,75 +12,73 @@
...
+             find (Any.map f p) ≡ Product.map id (Product.map id f) (find p)
```
MatthewDaggitt, line 53: Could use `Product.map_1` here?

src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -211,9 +210,9 @@ module _ (S : Setoid c ℓ) where
 
   ∈-∃++ : ∀ {v xs} → v ∈ xs → ∃₂ λ ys zs → ∃ λ w →
           v ≈ w × xs ≋ ys ++ [ w ] ++ zs
-  ∈-∃++ (here px)                  = [] , _ , _ , px , ≋-refl
-  ∈-∃++ (there {d} v∈xs) with ∈-∃++ v∈xs
-  ... | hs , _ , _ , v≈v′ , eq = d ∷ hs , _ , _ , v≈v′ , refl ∷ eq
+  ∈-∃++ (here px)        = [] , _ , _ , px , ≋-refl
+  ∈-∃++ (there {d} v∈xs) = let hs , _ , _ , v≈v′ , eq = ∈-∃++ v∈xs
+                           in d ∷ hs , _ , _ , v≈v′ , refl ∷ eq
```
MatthewDaggitt, line 215: This indentation style is inconsistent with the rest of the changes.

src/Data/List/Relation/Unary/All.agda
```diff
@@ -212,6 +212,9 @@ universal : Universal P → Universal (All P)
 universal u []       = []
 universal u (x ∷ xs) = u x ∷ universal u xs
 
+universal-U : Universal (All (U {A = A}))
```
MatthewDaggitt, line 215: I think `Universal (All {A = A} U)` would be neater?

src/Data/List/Relation/Unary/All.agda
```diff
@@ -116,17 +116,17 @@ unzip : All (P ∩ Q) ⊆ All P ∩ All Q
...
+  tabulateₛ : ∀∈ P xs → All P xs
+  tabulateₛ {[]}    hyp = []
+  tabulateₛ {_ ∷ _} hyp = hyp (here ≈-refl) ∷ tabulateₛ (hyp ∘ there)
 
-tabulate : (∀ {x} → x ∈ₚ xs → P x) → All P xs
+tabulate : ∀∈ₚ P xs → All P xs
 tabulate = tabulateₛ (≡.setoid _)
 
-self : ∀ {xs : List A} → All (const A) xs
+self : ∀ {xs} → All (const A) xs
```
MatthewDaggitt, line 129: Do we need `{xs}`?

src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -97,6 +97,11 @@ Any-cong {P = P} {Q = Q} {xs = xs} {ys} P↔Q xs≈ys =
 ------------------------------------------------------------------------
 -- Any.map
 
+map-ext : (f g : P ⋐ P) → (∀ {x} (p : P x) → f p ≡ g p) →
+          (p : Any P xs) → Any.map f p ≡ Any.map g p
+map-ext f g hyp (here  p) = cong here (hyp p)
+map-ext f g hyp (there p) = cong there $ map-ext f g hyp p
```
MatthewDaggitt, line 103: Missing in CHANGELOG?

src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -97,6 +97,11 @@ Any-cong {P = P} {Q = Q} {xs = xs} {ys} P↔Q xs≈ys =
 ------------------------------------------------------------------------
 -- Any.map
 
+map-ext : (f g : P ⋐ P) → (∀ {x} (p : P x) → f p ≡ g p) →
```
MatthewDaggitt, line 100: I don't quite understand why `f g` can't be from `P` to `Q`?

src/Data/List/Membership/Setoid.agda
```diff
@@ -32,14 +32,20 @@ x ∈ xs = Any (x ≈_) xs
 _∉_ : A → List A → Set _
 x ∉ xs = ¬ x ∈ xs
 
+∃∈ : ∀ {p} (P : Pred A p) → Pred (List A) _
+∃∈ P xs = ∃ λ x → x ∈ xs × P x
```
MatthewDaggitt, line 36: I'm going to be deeply unhelpful here and say that I don't really like these names without suggesting a concrete replacement. 

To me these read `Exists in P xs` and `Forall in P xs` neither of which makes sense to me... 

However, as I've said I can't come up with good alternatives. Anyone else have any good ideas?

MatthewDaggitt (changes requested):

src/Data/List/Membership/Propositional/Properties.agda
jamesmckinna: Good question!
I was focused on refactoring the 'heavy' `with`/case-analysis in the proof of `f′-injective′`, and not thinking about the computational content of `f′` itself...
... but actually that's OK, because the whole `where` block is dedicated to proving a *negative* statement, so nothing in it has/will have computational content! (And even if it did, the final `¬¬-excluded-middle` wrapper clinches it)

src/Data/List/Membership/Propositional/Properties/Core.agda
jamesmckinna: This was the original typing, and I'm conscious that systematic use of `variable` declarations can lead to unexpected re-ordering of the *implicit* argument prefix, which *may* break downstream client uses... but otherwise I would happily remove such things!

src/Data/List/Membership/Propositional/Properties/Core.agda
jamesmckinna: Well, as above, this is the legacy typing.

That said, as with [the changes to the definition of `find`](https://github.com/agda/agda-stdlib/blob/3e1c687dfa8bafe8a7efdf637a7f631304cc926a/src/Data/List/Membership/Setoid.agda#L57-L59), my temptation is actually to introduce an explicit `let`-binding here as well.

src/Data/List/Membership/Propositional/Properties/Core.agda
MatthewDaggitt: Okay fair enough!

src/Data/List/Membership/Propositional/Properties.agda
MatthewDaggitt: Okay, sounds good!

src/Data/List/Relation/Unary/All.agda
jamesmckinna: Er.... it turns out not to be possible to remove it, because we then lose the correlation between the type `List A` of `xs` and the `A` mentioned in the type, and restoring that telescopic relationship requires more syntax than we save! Unless you can find a good solution?

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Yes. I think that in fact this lemma, which never gets used, should in fact be removed... I mistakenly believed I could use to refactor `map-id` and `map-∘`...

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Indeed: see above. I don't think we need this lemma at all

src/Data/List/Membership/Setoid.agda
jamesmckinna: So...:
* the already-existing proofs `∃∈-Any` and `Any-∃∈` introduce the symbolism, so these definitions are exactly the back-formation of the associated operations, given those lemma names...
* I too wondered if they weren't so memorable/mnemonic, but see above: I didn't feel like deprecating the pre-existing names!

That said: does changing the parametrisation to 
```agda
∈ xs P = ∃ λ x → x ∈ xs × P x
```
improve matters? But even then, it can't be defined as a predicate transformer... because of the re-parametrisation!

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Aaaargh! Just realised that we also have (legacy again!):
* `Data.List.Relation.Unary.Any.Properties`
```agda
map-id : ∀ (f : P ⋐ P) → (∀ {x} (p : P x) → f p ≡ p) →
         (p : Any P xs) → Any.map f p ≡ p

map-∘ : ∀ (f : Q ⋐ R) (g : P ⋐ Q) (p : Any P xs) →
        Any.map (f ∘ g) p ≡ Any.map f (Any.map g p)
```
and
* `Data.List.Relation.Unary.All.Properties`
```agda
map-cong : ∀ {f : P ⋐ Q} {g : P ⋐ Q} (pxs : All P xs) →
           (∀ {x} → f {x} ≗ g) → All.map f pxs ≡ All.map g pxs
map-cong []         _   = refl
map-cong (px ∷ pxs) feq = cong₂ _∷_ (feq px) (map-cong pxs feq)

map-compose : ∀ {f : P ⋐ Q} {g : Q ⋐ R} (pxs : All P xs) →
              All.map g (All.map f pxs) ≡ All.map (g ∘ f) pxs
map-compose []         = refl
map-compose (px ∷ pxs) = cong (_ ∷_) (map-compose pxs)
```

So: should we
* keep the new lemma, rename it to `map-cong`, and generalise it as suggested?
* rename `map-compose` in `All.properties` to `map-∘` and deprecate the former?

Ugh: PR mission-creep!

jamesmckinna: DONE. Pushed.

src/Data/List/Membership/Setoid/Properties.agda
jamesmckinna: Hopefully commit 66a6207a14332f7e3ff15d9e15dc196159f5ace8 fixes this appropriately.

src/Data/List/Membership/Setoid.agda
JacquesCarette: I too read these as `Exists in P`. `∃∈×Pred` ? `∃∈P` ? `∃∈×P` ? ``∃-∈×P` ?

The flipped parametrisation is mildly nicer - but I agree with @MatthewDaggitt that the name needs work. Even if it means introducing a seeming inconsistency with existing naming.

src/Data/List/Relation/Binary/BagAndSetEquality.agda
```diff
@@ -388,10 +388,10 @@ drop-cons {x = x} {xs} {ys} x∷xs≈x∷ys =
     index-of (to xs≈ys (proj₂
       (from (Fin-length xs) (to (Fin-length xs) (z , p)))))   ≡⟨⟩
 
-    index-of (proj₂ (Prod.map id (to xs≈ys)
+    index-of (proj₂ (Product.map id (to xs≈ys)
```
JacquesCarette, line 391: May as well refactor to map_1 while editing this?

src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -255,24 +262,41 @@ Any-×⁻ pq with Product.map₂ (Product.map₂ find) (find pq)
...
-                               (λ y → subst Q y q)
-                               y∈ys
-           = lem₂ _ refl
+  to∘from : ∀ pq → Any-×⁺ (Any-×⁻ pq) ≡ pq
+  to∘from pq =
+    let x , x∈xs , pq′ = find pq
+
+        lem₁ : {f : (x ≡_) ⋐ λ x → Any (λ y → (P x) × (Q y)) ys} →
+               (f-refl : f refl ≡ pq′) → Any.map f _ ≡ pq
+        lem₁ {f} = map∘find pq {f = f}
```
JacquesCarette, line 271: Looking at this again: if the proof is `map∘find` why doesn't the statement seem to have a `find` in it? [My guess: the statement has an unfolding of the definition of `find` in it.]  So perhaps the lemma statement would be clearer without doing as much expansion?

src/Data/List/Relation/Unary/Any/Properties.agda
```diff
@@ -255,24 +262,41 @@ Any-×⁻ pq with Product.map₂ (Product.map₂ find) (find pq)
...
+
+        y , y∈ys , p , q = find pq′
+
+        lem₂ : {g : (y ≡_) ⋐ λ y → (P x) × (Q y)} →
+               (g-refl : g refl ≡ (p , q)) → Any.map g _ ≡ pq′
```
JacquesCarette, line 276: Same is true here. But here this lemma is only used once - so perhaps its short proof can instead be inlined at line 284 below?  And if the 'proof' is indeed `map∘find`, here and below, why doesn't the statement involve `find` ?

[Same guess; I've had to un-reduce some equational proofs LHSs to make the reason of the proof clearer in `agda-categories` - perhaps here too?]

JacquesCarette (changes requested):

src/Data/List/Membership/Setoid.agda
jamesmckinna: Will continue to ponder the alternatives. Thanks for the additional suggestions!

src/Data/List/Relation/Binary/BagAndSetEquality.agda
jamesmckinna: I don't have this line in my version: wrong branch?

jamesmckinna: Actually, I get typechecking errors if I do that!
Maybe I'm missing something?

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: This whole proof used to be incomprehensible because of the *very* tricky use of inlined-`with` to introduce lemmas whose types aren't well-formed, but whose definitions are (!?!?!?), and this very verbose version is the best I could do to reconstruct that. The use of `_` hides ...

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Well, as I first indicated, it's possible more refactoring is possible, but my brain ran out of resources...

jamesmckinna: The online g seems a good idea which moreover works... ;-) Not so sure about `lem₁` though...

src/Data/List/Relation/Binary/BagAndSetEquality.agda
Taneb: You're missing that @JacquesCarette totally _meant_ to say map_2, I think

JacquesCarette: Right...

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Thanks for continuing to nudge me towards the final version with each of `lem₁` and `lem₂` inlined, I hope it's much clearer now.

The `map∘find` arises not out of excessive expansion, but because `lose` is a disguised version of `map`on the outcome of a `find` (so *its* definition might one day usefully be inlined, but I really don't want that deprecation headache this time around...) and I think the goals are a bit confusing because of a ``map∘find∘map` composiiton which might yield to a different analysis... but again, for another time!

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: As above, inlining worked like a charm! (So I wonder if it could have already done so back in the original proofs... !?)

jamesmckinna:
So the only unresolved issue left, I think is of course: naming. 
Will continue to experiment over the coming day(s), and see if we can agree!

src/Data/List/Membership/Setoid.agda
jamesmckinna: First attempt incoming:
* keeps the names you don't like, but with `syntax` as a suffix
* introduces a more reasonable `syntax` for it

This gives us additional options for a 'real' name for the combinators, with the names I introduced available 'solely for the sake of the syntax'...?

jamesmckinna: At this stage, no attempt to propagate use of the new `syntax` to eg `Data.Membership.*.Properties`...

jamesmckinna: NB. These 'modalities' generalise `Diamond` and `Box` from `Relation.Unary.Closure.Base` so perhaps `◇∈` and `□∈` would at least be 'neutral' (if still as incomprehensible with additional `syntax`...?)

src/Data/List/Relation/Unary/Any/Properties.agda
jamesmckinna: Resolved to your satisfaction @JacquesCarette ?

src/Data/List/Membership/Setoid.agda
JacquesCarette: Certainly equally incomprehensible, but I can definitely see the appeal!

I'm worried about the over-use of `syntax`: it doesn't feel like it is a "fully designed feature", and certainly has no theoretical underpinning. That's a scary one-two.

src/Data/List/Relation/Unary/Any/Properties.agda
JacquesCarette: This is good enough. Further 'improvements' should be done by me.

src/Data/List/Membership/Setoid.agda
jamesmckinna: Well, I tend to agree with you regarding `syntax`, and that's why we have this horrible duplication fandango, so that such declarations can be hidden from `import`... but here, I think the syntax does communicate 'best' the intention of the combinator...

jamesmckinna:
Regarding `Box` and `Diamond` syntax... I realise we don't have these definitions in full generality for `R : REL A B _`, where they're usually written $\forall_{R}{Q}$ (for `Q : Pred B _`) and $\exists_{R}{P}$ (for `P : pred A _`), subject to an adjointness characterisation $$\exists_{R}{P}\subseteq{Q} \Leftrightarrow {P}\subseteq\forall_{R}{Q}$$

The symbols box and diamond come from later/subsequent uses in modal logic (IIRC, but my sense of the exact history of concepts in the field is a bit sketchy).

If we were to add *these* concepts to the library, what should we call *them*?

JacquesCarette:
Re: `∃[x∈xs] P xs = ∃ λ x → x ∈ xs × P x`. This still doesn't strongly hint to me that there's a `P` coming after that.  Though I guess if I "swap in" the version of `[_]` that I've seen Conor use a lot (not claiming he's author, just popularizer), this does make sense. But that would have to become part of our naming conventions to make wider sense?

jamesmckinna:
I'm not familiar with Conor's (use of) `[_]` (and in any case, such notation is *wildly* overloaded... I'd much rather it stood solely for 'singleton' container, and maybe it does for him, but I'm not seeing it...)

JacquesCarette:
```agda
  [_]  : {I : Set} (I -> Set) -> Set
  [ P ] = forall {i} -> P i
```

jamesmckinna:
Ah! Thanks (though as with my attempted use of the `syntax` for `Relationa.Unary.IUniversal` and `Relation.Unary.Satisfiable` I might expect some pushback from @MatthewDaggitt who I think is not a fan of 'clever' combinatory abbreviations...)

gallais:
> Regarding Box and Diamond syntax... I realise we don't have these (-monotone/functorial in each argument!) definitions in full generality for R : REL A B _, where they're usually written (for Q : Pred B _) and (for P : pred A _), subject to an adjointness characterisation 

Indeed, we only have it for homogeneous relations: http://agda.github.io/agda-stdlib/v2.0/Relation.Unary.Closure.Base.html

src/Data/List/Membership/Setoid.agda
jamesmckinna: Incoming commits pick my final proposed name. Please continue to comment, and/or change as you see fit... it's a case of search-and-replace now...

jamesmckinna:
> Re: `∃[x∈xs] P xs = ∃ λ x → x ∈ xs × P x`. This still doesn't strongly hint to me that there's a `P` coming after that. 

@JacquesCarette returning to this comment of yours after  few weeks of inactivity on this PR.

I think I don't really understand your comment: from the perspective of the `Pred`/subobject fibration, each quantifier is (a suitable adjoint to something) functorial between `Pred foo` and `Pred bar`... so why would a 'visibly obvious' quantifier such as `∃[x∈xs]` *not* suggest to you that it's going to be followed by a `P : Pred baz` and then a suitable argument to the resulting transformed `P`?

JacquesCarette:
Your first perspective (as true as it may be) doesn't seem to bear onto the problem: it's about usability. The fancier the explanation, the less likely it's going to communicate the intent to mere mortals.

But from the usability perspective... I don't know why I wasn't seeing it! Looking at it again today, it's clearly a quantifier that's still waiting for a predicate. Sigh. So: go for this notation.

JacquesCarette (approved):

src/Data/List/Membership/Setoid.agda
```diff
@@ -32,14 +32,30 @@ x ∈ xs = Any (x ≈_) xs
...
+
+syntax ∃∈-syntax (λ x → P) xs = ∃[ x ∈ xs ] P
+
+∀[x∈xs] : ∀ {p} (P : Pred A p) → Pred (List A) _
+∀[x∈xs] P xs = ∀ {x} → x ∈ xs → P x
+
+∀∈-syntax : ∀ {p} (P : Pred A p) → Pred (List A) _
+∀∈-syntax = ∀[x∈xs]
+
+syntax ∀∈-syntax (λ x → P) xs = ∀[ x ∈ xs ] P
```
MatthewDaggitt, line 49: Having looked through it again, my personal preference would be to simply remove these new definitions:
 - We've added 14 lines of code here in order to slightly shorten (not even remove) ~10 lines of code elsewhere.
 - The gains are pretty minimal `∀[x∈xs] P xs` (10 symbols) vs `∀ {x} → x ∈ xs → P x` (12 symbols)
 - I don't see the usage of these definitions increasing in the future. They're being used as a compatibility bridge to `All` and `Any`. We've crossed the bridge by proving the lemmas.
 - While the names are better now, I'm still super confused by `∃[ x ∈ xs ] P x` which looks to me like `(∃[ x ∈ xs ] P) x` but of course actually parses as `∃[ x ∈ xs ] (P x)` which is very non-intuitive to me.

However, it's not a hill I'm determined to make a stand on. If you're keen on them @jamesmckinna, then I'll bow out.

MatthewDaggitt: Actually it's quite hard to get a sense of how much they are used because of all the other changes. Given that this is the only controversial part, maybe it might be worth considering splitting out the introduction of this new syntax into a new PR?

jamesmckinna:
@MatthewDaggitt lots of food for thought in your recent round of comments. I'm travelling this week, so may not have time to weed through, but I take the criticism on board, as well as the hint to factor out the 'controversial' parts (which of course started the whole thing off in the first place, in terms of 'here's a property without the accompanying definition which might make sense of it'...) as well as my own views, which I hadn't previously indulged, about finding the whole `lose`/`find` naming scheme somewhat irritating...

Will update as soon as I get time to focus properly on it.

src/Data/List/Membership/Setoid.agda
jamesmckinna: Have removed all the new syntax, and will reconsider in a new PR in the light of all the various comments on it form the three of us.

jamesmckinna:
@MatthewDaggitt I've not opened a PR for the `syntax` extensions... yet. I'll scratch my head for a bit about the relational operator generalisations sketched above, and then try to fashion a focused PR introducing them, and specialising *their* syntax, if necessary, to the `_∈_` case...

... but in the meantime, I think this should now be ready to go!?

JacquesCarette:
Needs a re-review by @MatthewDaggitt

MatthewDaggitt (approved):

JacquesCarette (approved):

PR #2325, opened 2024-03-24
Include CHANGELOG in fix-whitespace and whitespace fixes

MatthewDaggitt:
As prompted by @andreasabel. There's no reason for the CHANGELOG not to be under whitespace fixes.

### DIFF
--- a/fix-whitespace.yaml
+++ b/fix-whitespace.yaml
@@ -8,4 +8,3 @@ included-files:
 
 excluded-files:
   - "README/Text/Tabular.agda"
-  - CHANGELOG.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,7 +170,7 @@ Additions to existing modules
   quasigroup      : Quasigroup _ _
   isLoop          : IsLoop _∙_ _\\_ _//_ ε
   loop            : Loop _ _
-  
+
   \\-leftDividesˡ  : LeftDividesˡ _∙_ _\\_
   \\-leftDividesʳ  : LeftDividesʳ _∙_ _\\_
   \\-leftDivides   : LeftDivides _∙_ _\\_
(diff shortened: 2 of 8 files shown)


fix-whitespace.yaml
```diff
@@ -7,5 +7,4 @@ included-files:
   - "*.md"
 
 excluded-files:
-  - "README/Text/Tabular.agda"
-  - CHANGELOG.md
```
jamesmckinna, line 11: AFAICT, this exclusion was added in commit by e0bd51a8c487075994a681309b62078328976adb
by @MatthewDaggitt but not sure why!?

fix-whitespace.yaml
```diff
@@ -7,5 +7,4 @@ included-files:
...
+  - "README/Text/Tabular.agda"
```
jamesmckinna, line 10: Agre that `CHANGELOG` should fall under `fix-whitespace`

jamesmckinna (approved):
Happy to restore `CHANGELOG` under `fix-whitespace` ... not sure quite why it ever got removed.

What I don't understand is how all the other whitespace violations (including some of my own!) crept in in `*.agda` modules!?

Good clean-up work @MatthewDaggitt Thanks!

fix-whitespace.yaml
JacquesCarette: Weirdly the PR that it is attached to points to commit [168f7e3](https://github.com/agda/agda-stdlib/pull/1379/commits/168f7e3d1cdbb0f7c97072f5bc58f87679dc2831) instead. It looks like this just 'snuck in' without discussion.

andreasabel:
> What I don't understand is how all the other whitespace violations (including some of my own!) crept in in `*.agda` modules!?

Probably because the whitespace check isn't run by CI.  (Or not anymore, dunno.)

Here is the whitespace watchdog now:
- #2328

andreasabel:
> There's no reason for the CHANGELOG not to be under whitespace fixes.

Technically, trailing whitespace in `.md` files stands for `<br/>`, but I think this is a misfeature that shouldn't be used.  (Use `<br/>` directly if you must.)
Another reason I can think of that folks want to use tabs, but again, one can instead use spaces.

andreasabel (approved):

PR #2326, opened 2024-03-24
Fix standard-library.agda-lib

andreasabel:
I accidentially introduce a modification in `standard-library.agda-lib` which I revert here.


jamesmckinna (approved):
Great! Thanks

PR #2329, opened 2024-03-26
[refactor] Relation.Nulllary.Decidable.Core

JacquesCarette:
- Use `Data.Unit.Polymorphic` instead of an explicit `Lift` in a couple of places
- Use `contradiction` instead of explicit `⊥-elim` in one place (other can't because of irrelevance)
- Tighten imports a bit

No need for CHANGELOG as these are all internal changes with no external effects.

### DIFF
--- a/src/Relation/Nullary/Decidable/Core.agda
+++ b/src/Relation/Nullary/Decidable/Core.agda
@@ -13,14 +13,14 @@ module Relation.Nullary.Decidable.Core where
 
 open import Level using (Level; Lift)
 open import Data.Bool.Base using (Bool; T; false; true; not; _∧_; _∨_)
-open import Data.Unit.Base using (⊤)
-open import Data.Empty using (⊥)
+open import Data.Unit.Polymorphic.Base using (⊤)
 open import Data.Empty.Irrelevant using (⊥-elim)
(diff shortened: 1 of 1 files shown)


jamesmckinna:
See also #2243 which among other things is a bit more systematic about `contradiction` and its irrelevant friends...

JacquesCarette:
Right, #2243 is indeed way more systematic. Nicely, I think that where we overlap, we made exactly the same edits. This PR is meant to be a tiny "free" improvement. The main thing it the first one (`Data.Unit.Polymorphic`) but I did more just because I was editing the file.

jamesmckinna:
Indeed: my latest commits suggest that *all* uses of `contradiction` can be made irrelevant in their first argument, i.e. every use is in fact '`weak-`' and I think moreover (not yet done) every use of `⊥-elim` can be replaced by its irrelevant `⊥-elim-irr` counterpart... this would be quite a breaking change were it to be propagated throughout the library, but I think it is the limiting endpoint of observing that `⊥-recompute` is provable...

MatthewDaggitt (approved):

MatthewDaggitt:
Added it to the merge queue. Let's get small wins in where we can.

PR #2332, opened 2024-03-27
mark variables private in Data.Container.FreeMonad

cspollard:
I assume that was the intention -- sorry if not!


jamesmckinna:
Looks reasonable to me! 
@MatthewDaggitt was this a `bug` before?

Taneb (approved):
Having publicly exported variables is very much not what we want to have. Thanks for noticing this!

gallais (approved):

Issue #2319, opened 2024-03-14
Parsing problems with the `step-swap` combinator in `Data.List.Relation.Binary.Permutation.Propositional`


jamesmckinna:
(micro-issue arising in the course of #2317 / #2311 )

The combinator, [defined](https://github.com/agda/agda-stdlib/blob/aa5d6ca30067c6b63806947d4ee0b42c6207b5b8/src/Data/List/Relation/Binary/Permutation/Propositional.agda#L95-L100) as: 
```agda
  -- Skip reasoning about the first two elements
  step-swap : ∀ x y xs {ys zs : List A} → (y ∷ x ∷ ys) IsRelatedTo zs →
              xs ↭ ys → (x ∷ y ∷ xs) IsRelatedTo zs
  step-swap x y xs rel xs↭ys = relTo (trans (swap x y xs↭ys) (begin rel))

  syntax step-swap x y xs y↭z x↭y = x ∷ y ∷ xs <<⟨ x↭y ⟩ y↭z
```
seems to work in the handful of places in the library where it is used:
* [defining `shift`](https://github.com/agda/agda-stdlib/blob/aa5d6ca30067c6b63806947d4ee0b42c6207b5b8/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda#L173)
* [defining `drop-mid′`](https://github.com/agda/agda-stdlib/blob/aa5d6ca30067c6b63806947d4ee0b42c6207b5b8/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda#L217)
* [defining `drop-mid′`](https://github.com/agda/agda-stdlib/blob/aa5d6ca30067c6b63806947d4ee0b42c6207b5b8/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda#L220)
but with some bracketing required, as well as only ever being used with `refl` as the intermediate step.

But attempting to deploy it in the associated [`README`](https://github.com/agda/agda-stdlib/blob/master/doc/README/Data/List/Relation/Binary/Permutation.agda) leads to the following unpleasant need for re-bracketing when trying to re-express example [`lem₂`](https://github.com/agda/agda-stdlib/blob/aa5d6ca30067c6b63806947d4ee0b42c6207b5b8/doc/README/Data/List/Relation/Binary/Permutation.agda#L49-L53) using the combinator, as follows:
```agda
-- In practice it is further useful to extend `PermutationReasoning`
[...]

Issue #1113, opened 2020-03-12
List permutations performance suffers from 'dumb' trans 


ajrouvoet:
The inductive definition of permutations in the stdlib uses the transitivity constructor as the implementation of 'trans':

https://github.com/agda/agda-stdlib/blob/b69dc60e3dc0a02a48a9f22354279e399823dc2e/src/Data/List/Relation/Binary/Permutation/Propositional.agda#L49

This can quickly explode to unnecessarily big terms. Imagine the following for 200k lines long:

![image](https://user-images.githubusercontent.com/475738/76530189-a9bd6600-6473-11ea-8f2d-e6efa157a78c.png)

This can easily be improved by defining a smart trans, that checks for refl appearing the left and right hand side:

```
trans : xs ↭ ys → ys ↭ zs → xs ↭ zs
trans refl ρ₂ = ρ₂
trans ρ₁ refl = ρ₁
trans ρ₁ ρ₂ = ↭-trans ρ₁ ρ₂
```

IMO this should be used by default and in the the IsEquivalence instance for Permutations. 
By switching from the 'dumb' to the 'smart' one, the term spanning 200k lines was reduced to one spanning a mere 1k. Because the smart one does not fully normalize permutations, it is computationally quite cheap.

MatthewDaggitt:
Squeezed this into the `v1.3` release

PR #2333, opened 2024-03-28
Refactor `Data.List.Relation.Binary.Permutation.*`, part I

### LINKED ISSUE #1354: Reimplement `Permutation.Propositional` in terms of `Permutation.Setoid`

jamesmckinna:
Part I of a project to fix #1354 

Highlights of the refactoring, cf. #2317 :
* Definitions don't change, so this is a 'conservative' first go, non-`breaking` for `v2.1`
* Additions to each of `Propositional` and `Setoid` and their `Properties` reconcile, more or less, the shared parts of the API
* Remove appeals to 'buggy' `PermutationReasoning` additional syntax #2319 
* General tidying up: `variable`s etc.
* Equivalence in `Propositional` of the relation and the  `Setoid (setoid _)` instance
* Refactor `Propositional.product-↭` in that spirit
* `Setoid` changes the proof of `↭-trans` in terms of new lemmas `↭-transˡ-≋` and `↭-transʳ-≋`; with them a sharpening of the analysis in `split` via a new lemma `↭-split` (purely structural; no WF-induction required)
* Move to `Homogeneous`, and deprecation in `Setoid`, of `steps`; towards removing legacy cruft now rendered obsolete by the above analysis
* (BUG?) `steps`  is deprecated in `Setoid`, but doesn't trigger the `WARNING_ON_USAGE` when the qualified-imported module `Permutation S` is `open`ed in `Properties`
* Final tidy up of `BagAndSetEquality` finishing off refactoring begun in #2321 

Not done, but could be:
* refactor `Setoid.Properties` into `Core` and derived
* ditto. for `Propositional`, and then rederived via the `Setoid (setoid _)` instance (eg even the `drop...` properties can be derived from those in `Setoid`) 

Won't do:
* move `Setoid` properties to `Homogeneous` (as in #2317) and knock-on refactoring of the `Core` properties etc.


src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -169,83 +180,81 @@ inject v ws↭ys xs↭zs = trans (++⁺ˡ _ (prep v xs↭zs)) (++⁺ʳ _ ws↭ys
...
-  drop-mid′ (swap y z p) (y ∷ z ∷ ws) (z ∷ [])     refl refl = begin
-      _ ∷ _ ∷ ws ++ _   <<⟨ refl ⟩
-      _ ∷ (_ ∷ ws ++ _) <⟨ ↭-sym (shift _ _ _) ⟩
-      _ ∷ (ws ++ _ ∷ _) <⟨ p ⟩
-      _ ∷ _             ∎
-      where open PermutationReasoning
-  drop-mid′ (swap y z p) (y ∷ z ∷ ws) (z ∷ y ∷ xs) refl refl = swap y z (drop-mid′ p _ _ refl refl)
-  drop-mid′ (trans p₁ p₂) ws  xs refl refl with ∈-∃++ (∈-resp-↭ p₁ (∈-insert ws))
-  ... | (h , t , refl) = trans (drop-mid′ p₁ ws h refl refl) (drop-mid′ p₂ h xs refl refl)
+drop-mid′ : ∀ {as bs} → as ↭ bs →
```
MatthewDaggitt, line 201: Looking at this diff, I don't understand where the definition of `drop-mid` has gone. Also is there a reason why we've now made `drop-mid′` publicly visible outside the module? You haven't listed this in the CHANGELOG so I'm guessing it's a  mistake?

MatthewDaggitt: Oh weird, just spotted the reason. The proof is now mutually dependent.... Is this refactoring really necessary? Non-mutually dependent proofs seems more desirable?

src/Data/List/Relation/Binary/Permutation/Setoid.agda
```diff
@@ -44,44 +41,66 @@ _↭_ = Homogeneous.Permutation _≈_
 ------------------------------------------------------------------------
 -- Constructor aliases
 
+-- Constructor alias
+
+↭-pointwise : _≋_ ⇒ _↭_
+↭-pointwise = refl
```
MatthewDaggitt, line 47: Again `↭-pointwise-≋`

src/Data/List/Relation/Binary/Permutation/Setoid.agda
```diff
@@ -104,12 +123,27 @@ module PermutationReasoning where
...
+
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.1
+
+steps = Homogeneous.steps {R = _≈_}
```
MatthewDaggitt, line 144: As mentioned above, not super keen on this deprecation. Can we simply re-export from `Homogeneous` for now?

src/Data/List/Relation/Binary/Permutation/Setoid.agda
```diff
@@ -44,44 +41,66 @@ _↭_ = Homogeneous.Permutation _≈_
...
+↭-transʳ-≋ (swap eq₁ eq₂ xs↭ys) (x≈w ∷ y≈z ∷ ys≋zs)
+  = swap (≈.trans eq₁ y≈z) (≈.trans eq₂ x≈w) (↭-transʳ-≋ xs↭ys ys≋zs)
+↭-transʳ-≋ (trans xs↭ws ws↭ys)  ys≋zs
+  = trans xs↭ws (↭-transʳ-≋ ws↭ys ys≋zs)
 
 ↭-trans : Transitive _↭_
-↭-trans = trans
+↭-trans (refl xs≋ys) ys↭zs = ↭-transˡ-≋ xs≋ys ys↭zs
+↭-trans xs↭ys (refl ys≋zs) = ↭-transʳ-≋ xs↭ys ys≋zs
+↭-trans xs↭ys ys↭zs        = trans xs↭ys ys↭zs
```
MatthewDaggitt, line 93: This is a big change in behaviour of `trans` and may break user's proofs? Regardless there should be documentation here explaining why we're using this non-obvious definition?

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -29,44 +28,53 @@ open import Data.List.Relation.Unary.AllPairs using (AllPairs; []; _∷_)
...
+
+private
+  variable
+    b p r : Level
+    x y z v w : A
+    xs ys zs vs ws : List A
+    P : Pred A p
+    R : Rel A r
+
+  _++[_]++_ = λ xs (z : A) ys → xs ++ [ z ] ++ ys
```
MatthewDaggitt, line 65: I have to confess I don't understand the purpose of this? As far as I can see it saves two space characters and that's it?

src/Data/Nat/Primality/Factorisation.agda
```diff
@@ -163,25 +165,26 @@ private
         where open ≡-Reasoning
 
       shuffle : ∃[ bs′ ] bs ↭ a ∷ bs′
-      shuffle with ys , zs , p ← ∈-∃++ (factorisationHasAllPrimeFactors prime[a] a∣Πbs prime[bs])
-        = ys ++ zs , ↭-trans (↭-reflexive p) (shift a ys zs)
+      shuffle
+        with ys , zs , refl ← ∈-∃++ (factorisationHasAllPrimeFactors prime[a] a∣Πbs prime[bs])
+        = ys ++ zs , shift a ys zs
```
MatthewDaggitt, line 170: Why are these changes necessary? If this PR is truly backwards compatible as you say, then can we remove all the changes in this file from this PR and make them in a separate one? Would be reassuring! (apart from the deprecated reasoning combinator!)

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -87,121 +95,188 @@ AllPairs-resp-↭ sym resp@(rʳ , rˡ) (swap eq₁ eq₂ p) ((∼₁ ∷ ∼₂)
...
-steps-respˡ (_ ∷ _ ∷ ys≋xs) (swap _ _ ys↭zs)    = cong suc (steps-respˡ ys≋xs ys↭zs)
-steps-respˡ ys≋xs           (trans ys↭ws ws↭zs) = cong (_+ steps ws↭zs) (steps-respˡ ys≋xs ys↭ws)
-
-steps-respʳ : ∀ {xs ys zs} (xs≋ys : xs ≋ ys) (zs↭xs : zs ↭ xs) →
-              steps (↭-respʳ-≋ xs≋ys zs↭xs) ≡ steps zs↭xs
-steps-respʳ _               (refl _)            = refl
-steps-respʳ (_ ∷ ys≋xs)     (prep _ ys↭zs)      = cong suc (steps-respʳ ys≋xs ys↭zs)
-steps-respʳ (_ ∷ _ ∷ ys≋xs) (swap _ _ ys↭zs)    = cong suc (steps-respʳ ys≋xs ys↭zs)
-steps-respʳ ys≋xs           (trans ys↭ws ws↭zs) = cong (steps ys↭ws +_) (steps-respʳ ys≋xs ws↭zs)
+↭-split : ∀ v as bs {xs} → xs ↭ (as ++[ v ]++ bs) →
```
MatthewDaggitt, line 120: By convention the `{xs}` argument should come first as it's mentioned first...

MatthewDaggitt (changes requested):

MatthewDaggitt:
Generally looks good though! The drastic reduction in proof length is pretty great!

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
```diff
@@ -59,3 +59,14 @@ map R⇒S (refl xs∼ys)         = refl (Pointwise.map R⇒S xs∼ys)
 map R⇒S (prep e xs∼ys)       = prep (R⇒S e) (map R⇒S xs∼ys)
 map R⇒S (swap e₁ e₂ xs∼ys)   = swap (R⇒S e₁) (R⇒S e₂) (map R⇒S xs∼ys)
 map R⇒S (trans xs∼ys ys∼zs)  = trans (map R⇒S xs∼ys) (map R⇒S ys∼zs)
+
+------------------------------------------------------------------------
+-- Steps moves here towards eventual deprecation, because it is
+-- * representation-dependent
+-- * now obsolete wrt `Setoid.Properties.{dropMiddleElement,split}`
```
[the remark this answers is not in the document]
jamesmckinna: I can roll back on that, but interestingly, the warning doesn't seem to be generated when importing `Setoid` qualified in `Setoid.Properties`... (I raise that as a potential 'bug' above...)

Mind you, you should consider that induction on `steps` is *very* representation dependent, to the point of not even being invariant under the smart constructors such as `↭-trans`, and requiring eg that `↭-sym` actually preserve the shape of derivations on the nose... which isn't likely to survive any downstream representation change?

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
jamesmckinna: Two or three 'reasons' for the refactoring:
* first I hadn't quite finished, alas, and was still doing minor tweaks; sorry for that, but now it's a helper function again...
* given the refactoring in `Setoid.Properties` of `↭-split`, I had hoped to be able to eliminate `drop-mid` in favour of `dropMiddleElement`... but then ran out of steam
* there's a very heavy reliance in the original on calling `drop-mid′` mostly with last two arguments `refl`, so it seemed worthwhile (for readability, if nothing else) to in-line those exactly in terms of `drop-mid` and hope that the termination checker could cope... and it can ;-)

src/Data/List/Relation/Binary/Permutation/Setoid.agda
jamesmckinna: Interesting!
... yes, the intentional behaviour does change, but exactly by analogy with the corresponding version in `Propositional`. But if you *really* want the intentional behaviour, you still have access to the vanilla `trans` constructor... is that (not) good enough?

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
jamesmckinna: Yeah, it's maybe taking DRY too far, but it's purely local. Elsewhere in #2317 though, I ended up having to use the qualified version of `_++_` and that was getting very heavy...

Easy to inline though!

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
jamesmckinna: Not always on top of the conventions, but yes, sure.

src/Data/Nat/Primality/Factorisation.agda
jamesmckinna: Well that's certainly true (separate PR), and maybe it's PR mission-creep, but once you start tackling `Propositional` permutation, tackling the (very modest) downstream knock-on opportunities seemed/seems worthwhile. Can revert, once again.

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
jamesmckinna: DONE, and use o0f `variable`s means it need not even be quantified...

src/Data/List/Relation/Binary/Permutation/Propositional.agda
```diff
@@ -30,21 +38,32 @@ open import Relation.Binary.Reasoning.Syntax
...
 ------------------------------------------------------------------------
 -- _↭_ is an equivalence
 
 ↭-reflexive : _≡_ ⇒ _↭_
-↭-reflexive refl = refl
+↭-reflexive refl = ↭-refl
 
-↭-refl : Reflexive _↭_
-↭-refl = refl
+↭-pointwise : _≋_ ⇒ _↭_
```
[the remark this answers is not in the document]
jamesmckinna: A-ha! Will fix... DONE

src/Data/List/Relation/Binary/Permutation/Setoid.agda
jamesmckinna: DONE, as `↭-trans′`...

jamesmckinna: Mind you, a proof I couldn't be bothered to investigate is whether the `steps` properties hold for *this* version of 'smart' `trans`... which I think in fact should be provable, given the way pointwise equality is being pushed through the derivation tree to the leaves... but as I say, that's work I *really* don't want to do.

jamesmckinna:
I think that's all of your review comments handled now! But leaving certain of the conversations unresolved until you're happy... ;-)

jamesmckinna:
Regarding (the lack of extensionality in) `steps`: 
* intensionally, given a proof `p : [] ↭ []`, `steps p` will be some value well-defined by recursion on `p`
* extensionally, it feels *very* weird (to me) that we can't define a single value for this (eg 0, or 1, as currently defined as an instance of `refl`...)

Cf. #1113 Any such proof `p` must, in fact, hereditarily be some `trans` composition of proofs of the form `refl` (hence `↭-empty-inv` is not only provable, but has the proof it has), but each of those compositions will have a *different* `steps` count.

My analogy (apologies if I am belabouring the point) would be that of a `_++_`-based representation of `List` in which we counted `[] ++ xs` as having a different length to that of `xs`... no?

More generally, 'experience' suggests to me that definitions of inductively-defined relations which build in `trans` should generally be considered ... potentially harmful (eg it's a design pattern in type theory/PL to strive as far as possible for 'transitivity elimination') for precisely such 'reason's.

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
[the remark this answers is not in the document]
jamesmckinna: I toned down the comment, but forgot to push that until now. Sorry!

CHANGELOG.md
```diff
@@ -262,6 +271,43 @@ Additions to existing modules
...
+  ↭-trans′      : Transitive _↭_
+  ```
+
+* In `Data.List.Relation.Binary.Permutation.Setoid.Properties`:
+  ```agda
+  ↭-split : xs ↭ (as ++[ v ]++ bs) →
+            ∃₂ λ ps qs → xs ≋ (ps ++[ v ]++ qs) × (ps ++ qs) ↭ (as ++ bs)
+  drop-∷  : x ∷ xs ↭ x ∷ ys → xs ↭ ys
+
+  foldr-pres-↭ : (IsCommutativeMonoid _≈_ _∙_ ε) → (foldr _∙_ ε) Preserves _↭_ ⟶ _≈_
```
jamesmckinna, line 308: It occurred to me that this deprecation, too, is premature. Will revert now.

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -454,52 +398,102 @@ module _ {ℓ} {R : Rel A ℓ} (R? : B.Decidable R) where
...
+{-# WARNING_ON_USAGE split
+"Warning: split was deprecated in v2.1.
+Please use the sharper lemma ↭-split instead."
+#-}
+
+foldr-commMonoid = foldr-pres-↭
+{-# WARNING_ON_USAGE foldr-commMonoid
+"Warning: foldr-commMonoid was deprecated in v2.1.
+Please use foldr-pres-↭ instead."
+#-}
```
jamesmckinna, line 499: Premature, esp. in view of my comment elsewhere on preferring a (future!) version of this to take the *bundle* and not the *structure* as argument. But I should save the name for such downstream refactoring! Will now revert.

src/Data/List/Relation/Binary/Permutation/Setoid.agda
jamesmckinna: Out of interest, does the introduction of `↭-trans′`  break any of *your* proofs?

JacquesCarette (changes requested):

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
```diff
@@ -59,3 +59,12 @@ map R⇒S (refl xs∼ys)         = refl (Pointwise.map R⇒S xs∼ys)
 map R⇒S (prep e xs∼ys)       = prep (R⇒S e) (map R⇒S xs∼ys)
 map R⇒S (swap e₁ e₂ xs∼ys)   = swap (R⇒S e₁) (R⇒S e₂) (map R⇒S xs∼ys)
 map R⇒S (trans xs∼ys ys∼zs)  = trans (map R⇒S xs∼ys) (map R⇒S ys∼zs)
+
+------------------------------------------------------------------------
+-- Steps moves here because it is representation-dependent
+
+steps : ∀ {R : Rel A r} {xs ys} → Permutation R xs ys → ℕ
+steps (refl _)            = 1
```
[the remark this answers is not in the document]
jamesmckinna: This is legacy code, and after @MatthewDaggitt 's (justified; correct) discouraging me from deprecating it, I don't propose to touch it as part of what is meant to be a conservative non-breaking PR...

src/Data/List/Relation/Binary/Permutation/Propositional.agda
```diff
@@ -68,6 +87,28 @@ data _↭_ : Rel (List A) a where
   { isEquivalence = ↭-isEquivalence
   }
 
+------------------------------------------------------------------------
+-- _↭_ is equivalent to `Setoid`-based permutation
```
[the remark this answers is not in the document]
jamesmckinna: I don't think that's going to work for precisely the reasons that distinguish the use of *intensional* aspects of the definitions form their *extensional ones*. In particular, I don't think it will even hold without further binding in the specific constructor definitions for the two relations. 

In any case, as this PR is intended as the first part of a longer sequence intended to conclude with a resolution of both #1354 and #2311 , I'm inclined to say this review comment as mission-creep that's out-of-scope at this stage?

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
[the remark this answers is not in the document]
JacquesCarette: Sold.

src/Data/List/Relation/Binary/Permutation/Propositional.agda
```diff
@@ -89,12 +130,12 @@ module PermutationReasoning where
   -- Skip reasoning on the first element
   step-prep : ∀ x xs {ys zs : List A} → (x ∷ ys) IsRelatedTo zs →
               xs ↭ ys → (x ∷ xs) IsRelatedTo zs
-  step-prep x xs rel xs↭ys = relTo (trans (prep x xs↭ys) (begin rel))
+  step-prep x xs rel xs↭ys = ↭-go (↭-prep x xs↭ys) rel
```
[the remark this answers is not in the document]
jamesmckinna: Again: legacy code.
All I have done here is encapsulate the permutation constructors and that of the underlying representing type `IsRelatedTo`, so that the appeal to `Reasoning.Syntax` is better encapsulated. 
NB. *no* connection here between `Homogeneous.steps` and the `step-*` combinators defined: the latter refers to steps of `Reasoning.Syntax`!

src/Data/List/Relation/Binary/Permutation/Propositional.agda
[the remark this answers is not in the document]
JacquesCarette: Ok.

src/Data/List/Relation/Binary/Permutation/Setoid.agda
```diff
@@ -104,12 +128,21 @@ module PermutationReasoning where
...
   step-swap : ∀ x y xs {ys zs : List A} → (y ∷ x ∷ ys) IsRelatedTo zs →
               xs ↭ ys → (x ∷ y ∷ xs) IsRelatedTo zs
-  step-swap x y xs rel xs↭ys = relTo (trans (swap Eq.refl Eq.refl xs↭ys) (begin rel))
+  step-swap x y xs rel xs↭ys = ↭-go (↭-swap x y xs↭ys) rel
 
   syntax step-prep x xs y↭z x↭y = x ∷ xs <⟨ x↭y ⟩ y↭z
   syntax step-swap x y xs y↭z x↭y = x ∷ y ∷ xs <<⟨ x↭y ⟩ y↭z
+
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
```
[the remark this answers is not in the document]
jamesmckinna: Good catch!

src/Data/List/Relation/Binary/Permutation/Propositional.agda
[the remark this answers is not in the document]
JacquesCarette: Ok, happy to not devolve this into mission creep.

JacquesCarette (approved):

src/Data/List/Relation/Binary/Permutation/Setoid.agda
jamesmckinna: Forgot to add the commentary you requested. Probably a bit verbose now!

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
jamesmckinna: I'll revert if you insist, but I do think an important part of `refactoring` is also to improve/streamline existing proofs... Please advise!

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
```diff
@@ -59,3 +59,12 @@ map R⇒S (refl xs∼ys)         = refl (Pointwise.map R⇒S xs∼ys)
 map R⇒S (prep e xs∼ys)       = prep (R⇒S e) (map R⇒S xs∼ys)
 map R⇒S (swap e₁ e₂ xs∼ys)   = swap (R⇒S e₁) (R⇒S e₂) (map R⇒S xs∼ys)
 map R⇒S (trans xs∼ys ys∼zs)  = trans (map R⇒S xs∼ys) (map R⇒S ys∼zs)
+
+------------------------------------------------------------------------
+-- Steps moves here because it is representation-dependent
```
MatthewDaggitt, line 64: This feels like a GitHub comment rather than a code comment. Maybe just a simple comment that says `Measures the number of constructors, can be useful for termination proofs` rather than a section heading?

MatthewDaggitt (commented):
Other than that looks great!

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
MatthewDaggitt: Yup okay fair enough. The latest refactoring makes it clear that it's not necessary to make it mutually recursive so I'm happy with it.

src/Data/List/Relation/Binary/Permutation/Propositional.agda
```diff
@@ -30,21 +38,32 @@ open import Relation.Binary.Reasoning.Syntax
...
+-- Constructor aliases
+
+↭-refl : Reflexive _↭_
+↭-refl = refl
+
+↭-prep  : ∀ x → xs ↭ ys → x ∷ xs ↭ x ∷ ys
+↭-prep = prep
+
+↭-swap  : ∀ x y → xs ↭ ys → x ∷ y ∷ xs ↭ y ∷ x ∷ ys
+↭-swap = swap
```
gallais, line 55: As always I am going to be annoying and complain that I don't like prefixed
names because users can either use the short names or import the module
qualified if they need to manually disambiguate.

The same goes for `split` being renamed to `↭-split`  (although in that case
the name `split` is already pretty poor IMO).

jamesmckinna: @gallais I don't find the comment annoying!
What I find annoying is the the abstraction failure when explicitly exporting constructors from an inductive definition which should remain (externally, to clients) abstract, IMHO (cf. `z<s` in `Data.Nat.Base`, and `0<1+n` in `Data.Nat.Properties`, not that I/we are entirely consistent in enforcing this...)
Since this is also part of a greater project of refactoring `Permutation` towards completing #1761 / #1354 I wanted to make sure that in `Properties`, references to the constructors were abstracted on those lines before proceeding further.

As regards `↭-split`, this is *not* a renaming of `split`, but a significant generalisaiton which permits a much improved proof of `dropMiddleElement`... so a cognate name change is needed to relfect the 'bettre' typing!

jamesmckinna:
Modulo @gallais 's comment, and my (attempted) rebuttal, I think that this is good to go @MatthewDaggitt unless there are review comments of yours that you are not happy have been dealt with...?

jamesmckinna:
At some point, the merge conflict resolution (probably thanks to me) went wrong, so fixing things now...

MatthewDaggitt (approved):

JacquesCarette (approved):

jamesmckinna:
Resolving latest round of merge conflicts exposed some issues with `CHANGELOG` (incl. a bug in that for #2418 ), which I've attempted to correctly rectify, but until this is merged, they'll probably go round and round in circles... sigh.

jamesmckinna:
Hmmm.... fixing the merge conflict may need a bit more thought.

Issue #2331, opened 2024-03-27
move `_≗_` from `Relation.Binary.PropositionalEquality` to `*.Core`


JacquesCarette:
It has a very straightforward definition (if you unwind the current overly complex one) and is used a fair amount, which then drags in quite a bit more than is needed. In particular, `Data.Product.Properties` is otherwise very simple, but lies higher up in the dependency graph due to this single definition.

MatthewDaggitt:
This is fine to do!

MatthewDaggitt:
And yes that definition is definitely a little overly complex!

jamesmckinna:
I agree with each of you!

PR #2334, opened 2024-03-29
Tighten imports

JacquesCarette:
I started from `Data.List.Fresh` and looked at the dependency graph (thanks for that nice script @gallais !) and tightened up  imports to make the dependency graph seem more 'natural' from a conceptual dependency point of view. While I was looking at various files, I also added some explicit `using` when it seemed to make sense.

This should be a pure clean-up with no visible consequences externally, other than making the dependency graph less convoluted. (So if anything else has slipped-in by mistake, it should be taken out!)

### DIFF
--- a/src/Algebra/Definitions/RawMagma.agda
+++ b/src/Algebra/Definitions/RawMagma.agda
@@ -10,10 +10,10 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra.Bundles using (RawMagma)
+open import Algebra.Bundles.Raw using (RawMagma)
 open import Data.Product.Base using (_×_; ∃)
 open import Level using (_⊔_)
-open import Relation.Binary.Core
+open import Relation.Binary.Core using (Rel)
 open import Relation.Nullary.Negation using (¬_)
 
 module Algebra.Definitions.RawMagma
--- a/src/Algebra/Definitions/RawMagma.agda
+++ b/src/Algebra/Definitions/RawMagma.agda
@@ -14,7 +14,7 @@ open import Algebra.Bundles.Raw using (RawMagma)
 open import Data.Product.Base using (_×_; ∃)
 open import Level using (_⊔_)
 open import Relation.Binary.Core using (Rel)
-open import Relation.Nullary.Negation using (¬_)
+open import Relation.Nullary.Negation.Core using (¬_)
 
 module Algebra.Definitions.RawMagma
   {a ℓ} (M : RawMagma a ℓ)
--- a/src/Algebra/Structures.agda
+++ b/src/Algebra/Structures.agda
@@ -21,7 +21,7 @@ module Algebra.Structures
 -- The file is divided into sections depending on the arities of the
 -- components of the algebraic structure.
 
-open import Algebra.Core
+open import Algebra.Core using (Op₁; Op₂)
 open import Algebra.Definitions _≈_
 import Algebra.Consequences.Setoid as Consequences
 open import Data.Product.Base using (_,_; proj₁; proj₂)
--- a/src/Data/List/Relation/Unary/AllPairs.agda
+++ b/src/Data/List/Relation/Unary/AllPairs.agda
@@ -18,7 +18,7 @@ open import Function.Base using (id; _∘_)
 open import Level using (_⊔_)
 open import Relation.Binary.Definitions as B
 open import Relation.Binary.Construct.Intersection renaming (_∩_ to _∩ᵇ_)
-open import Relation.Binary.PropositionalEquality
+open import Relation.Binary.PropositionalEquality.Core using (refl; cong₂)
 open import Relation.Unary as U renaming (_∩_ to _∩ᵘ_) hiding (_⇒_)
 open import Relation.Nullary.Decidable as Dec using (_×-dec_; yes; no)
 
--- a/src/Data/Product/Properties.agda
+++ b/src/Data/Product/Properties.agda
@@ -8,13 +8,14 @@
 
 module Data.Product.Properties where
 
-open import Axiom.UniquenessOfIdentityProofs
-open import Data.Product.Base
+open import Axiom.UniquenessOfIdentityProofs using (UIP; module Decidable⇒UIP)
+open import Data.Product.Base using (_,_; Σ; _×_; map; swap; ∃; ∃₂)
 open import Function.Base using (_∋_; _∘_; id)
 open import Function.Bundles using (_↔_; mk↔ₛ′)
 open import Level using (Level)
 open import Relation.Binary.Definitions using (DecidableEquality)
 open import Relation.Binary.PropositionalEquality
+  using (_≡_; _≗_; subst; cong; cong₂; cong′; refl)
 open import Relation.Nullary.Decidable as Dec using (Dec; yes; no)
 
 private
--- a/src/Function/Indexed/Relation/Binary/Equality.agda
+++ b/src/Function/Indexed/Relation/Binary/Equality.agda
@@ -8,7 +8,7 @@
 
 module Function.Indexed.Relation.Binary.Equality where
 
-open import Relation.Binary using (Setoid)
+open import Relation.Binary.Bundles using (Setoid)
 open import Relation.Binary.Indexed.Heterogeneous using (IndexedSetoid)
 
 -- A variant of setoid which uses the propositional equality setoid
--- a/src/Relation/Binary/Bundles.agda
+++ b/src/Relation/Binary/Bundles.agda
@@ -11,11 +11,10 @@
 module Relation.Binary.Bundles where
 
 open import Function.Base using (flip)
-open import Level
-open import Relation.Nullary.Negation using (¬_)
-open import Relation.Binary.Core
-open import Relation.Binary.Definitions
-open import Relation.Binary.Structures
+open import Level using (Level; suc; _⊔_)
+open import Relation.Nullary.Negation.Core using (¬_)
+open import Relation.Binary.Core using (Rel)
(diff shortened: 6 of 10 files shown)


jamesmckinna:
My instincts are strongly with you on things like this, but for the following concerns:
* given our stated policy regarding `Core` modules, are we making ourselves hostages to fortune by insisting on explicit dependency on such things? In practice, the cost of such knock-on viscosity might be small, but when I see `using` directives mention names which are already marked in open PRs for deprecation, my heart sinks a little... but maybe a (side?) effect of your work is to indicate candidate areas which indeed need such attention...
* ... thinking ... more to come

src/Algebra/Structures.agda
```diff
@@ -21,7 +21,7 @@ module Algebra.Structures
 -- The file is divided into sections depending on the arities of the
 -- components of the algebraic structure.
 
-open import Algebra.Core
+open import Algebra.Core using (Op₁; Op₂)
```
jamesmckinna, line 24: This is (semantically) redundant, as `Algebra.Core` consists of only those definitions... but was that your intention here, to draw attention to that?

I'm conscious that the library has several/many such small leaf modules, for the sake of resolving some dependency knots, so not clear about the value-add of this patch?

src/Algebra/Definitions/RawMagma.agda
```diff
@@ -10,11 +10,11 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra.Bundles using (RawMagma)
+open import Algebra.Bundles.Raw using (RawMagma)
```
jamesmckinna, line 13: Definitely in favour of such localisation where possible, given my various recent issues around `RawX` vs `IsX` dependency in the hierarchy generally...

src/Data/List/Relation/Unary/AllPairs.agda
```diff
@@ -18,7 +18,7 @@ open import Function.Base using (id; _∘_)
 open import Level using (_⊔_)
 open import Relation.Binary.Definitions as B
 open import Relation.Binary.Construct.Intersection renaming (_∩_ to _∩ᵇ_)
-open import Relation.Binary.PropositionalEquality
+open import Relation.Binary.PropositionalEquality.Core using (refl; cong₂)
```
jamesmckinna, line 21: Fine. I guess that the current decomposition of the `PropositionalEquality` modules is pretty much set in stone for the time being, so the localisation makes sense.

src/Relation/Binary/Bundles.agda
```diff
@@ -11,11 +11,10 @@
...
 open import Function.Base using (flip)
-open import Level
-open import Relation.Nullary.Negation using (¬_)
-open import Relation.Binary.Core
-open import Relation.Binary.Definitions
-open import Relation.Binary.Structures
+open import Level using (Level; suc; _⊔_)
+open import Relation.Nullary.Negation.Core using (¬_)
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Structures -- most of it
```
jamesmckinna, line 17: Should this comment be left in?

jamesmckinna (approved):
I'm basically happy with this, barring the `Algebra.Core using...` ones which take the point too far, but that's just my opinion. 

As for deprecation/renaming burden, maybe I should see this as a diagnostic tool... but I think there is a tension between making dependency explicit (which increases knock-on viscosity) and the potential 'documentation' benefit. 

This PR is small, so there's less st stake, but I think we should discuss the issue as an issue more widely. Style-guide recommendations as well?

src/Data/Product/Properties.agda
```diff
@@ -8,13 +8,14 @@
...
 module Data.Product.Properties where
 
-open import Axiom.UniquenessOfIdentityProofs
-open import Data.Product.Base
+open import Axiom.UniquenessOfIdentityProofs using (UIP; module Decidable⇒UIP)
+open import Data.Product.Base using (_,_; Σ; _×_; map; swap; ∃; ∃₂)
 open import Function.Base using (_∋_; _∘_; id)
 open import Function.Bundles using (_↔_; mk↔ₛ′)
 open import Level using (Level)
 open import Relation.Binary.Definitions using (DecidableEquality)
```
jamesmckinna, line 16: What's your take in this module on *not* importing `Relation.Binary.Core using (REL)` so that 
```agda
∃∃↔∃∃ : (R : A → B → Set ℓ) → (∃₂ λ x y → R x y) ↔ (∃₂ λ y x → R x y)
```
uses `(R : REL A B ℓ)` instead?

src/Data/Product/Properties.agda
```diff
@@ -8,13 +8,14 @@
...
 open import Relation.Binary.PropositionalEquality
+  using (_≡_; _≗_; subst; cong; cong₂; cong′; refl)
```
jamesmckinna, line 18: Ugh! I see why you might want `_≗_` in `Core`!
What's your take on `using` directives mentioning >5 names? (Style-guide again...)

src/Function/Indexed/Relation/Binary/Equality.agda
```diff
@@ -8,7 +8,7 @@
 
 module Function.Indexed.Relation.Binary.Equality where
 
-open import Relation.Binary using (Setoid)
+open import Relation.Binary.Bundles using (Setoid)
```
jamesmckinna, line 11: :+100:

JacquesCarette:
> My instincts are strongly with you on things like this, but for the following concerns:
> 
> * given our stated policy regarding `Core` modules, are we making ourselves hostages to fortune by insisting on explicit dependency on such things? 

I thought that that policy applied to code outside the library, but that code inside the library is free to use `.Core`. If that's not the case, we're going to need to create a whole lot of `.Base` modules instead.

> In practice, the cost of such knock-on viscosity might be small, but when I see `using` directives mention names which are already marked in open issues/PRs for deprecation (`Reflects.invert`, looking at you), my heart sinks a little... but maybe a (side?) effect of your work is to indicate candidate areas which indeed need such attention...

I'm certainly hoping not to entrench any such things - I'm only trying to make the concept hierarchy more sensible. I'm happy to refactor while I'm at it, if I know which direction to refactor in to.

src/Algebra/Structures.agda
JacquesCarette: I 'mechanically' put in `using` without much regard to what's in the downstream file, as a means to document what is being used in the current file, i.e. establish an explicit vocabulary / domain of discourse. That it turns out to be the whole file *right now* doesn't mean that `Algebra.Core` won't grow in the future! So these `using` future-proof things from accidental namespace pollution.

The value-add of this particular patch is tiny indeed. I still prefer my imports to be explicit, so that I don't have to keep in active memory all that's in those modules.

src/Relation/Binary/Bundles.agda
JacquesCarette: I added it on purpose. It's for readers of this file, so they know that the lack of `using` is not a mistake, and that a `using` here would be quite "large" and so the ergonomics start to be low. But it still gives information to the reader.

src/Data/Product/Properties.agda
JacquesCarette: I've got a shortly-incoming PR for #2331. It's a wart.

A `using` directive that's longer than a single line starts to be excessive. But that's my threshold, the line, not the number of names. [In my own code, I tend to go full-explicit all the time. I'm not yet fully convinced that that's the right way to do things - so I'm certainly not ready to argue for that for `stdlib`.]

src/Relation/Binary/Bundles.agda
jamesmckinna: Indeed, but I think it's non-conventional as regards existing style guidelines?
Happy to leave it in, but Y(and others')MMV...

jamesmckinna:
> > My instincts are strongly with you on things like this, but for the following concerns:
> > 
> > * given our stated policy regarding `Core` modules, are we making ourselves hostages to fortune by insisting on explicit dependency on such things?
> 
> I thought that that policy applied to code outside the library, but that code inside the library is free to use `.Core`. If that's not the case, we're going to need to create a whole lot of `.Base` modules instead.

No, you're correct. Stdlib developers are free to use `.Core`, but there's still a question about how tight we *should* make things. I guess it's a spectrum...

> > In practice, the cost of such knock-on viscosity might be small, but when I see `using` directives mention names which are already marked in open issues/PRs for deprecation (`Reflects.invert`, looking at you), my heart sinks a little... but maybe a (side?) effect of your work is to indicate candidate areas which indeed need such attention...
> 
> I'm certainly hoping not to entrench any such things - I'm only trying to make the concept hierarchy more sensible. I'm happy to refactor while I'm at it, if I know which direction to refactor in to.

Well, that would require those PRs to have advanced as far as the review stage... but there's a backlog. And in principle I like the 'documentary' function, not least because it makes for a faster-fail when loading if the names have been deprecated... (and chasing up knock-on viscosity remains a pain, so perhaps indeed the cues are helpful!)

src/Relation/Binary/Bundles.agda
JacquesCarette: Blazing new trails, that's me.

jamesmckinna: Yes, but if we invent/uncover new ways to work that we feel useful/appropriate, then we should document them!

JacquesCarette: Waiting for others like @MatthewDaggitt and @Taneb to chyme in one way or another on this.

MatthewDaggitt (approved):

MatthewDaggitt:
Opened https://github.com/agda/agda-stdlib/issues/2339 to discuss. Merging this for now.

PR #2335, opened 2024-03-29
Move pointwise equality to `.Core` module

### LINKED ISSUE #2331: move `_≗_` from `Relation.Binary.PropositionalEquality` to `*.Core`

JacquesCarette:
Closes #2331.

Also makes a few changes of imports that are related. Many of the places which look like this improvement might also help use other things from `Relation.Binary.PropositionalEquality` that should *not* be moved to `.Core`. (But it might make sense to split them off into their own lighter weight module.)

### DIFF
--- a/src/Data/List/Effectful.agda
+++ b/src/Data/List/Effectful.agda
@@ -10,14 +10,19 @@ module Data.List.Effectful where
 
 open import Data.Bool.Base using (false; true)
 open import Data.List.Base
+  using (List; map; [_]; ap; []; _∷_; _++_; concat; concatMap)
 open import Data.List.Properties
-open import Effect.Choice
-open import Effect.Empty
-open import Effect.Functor
+  using (++-identityʳ; ++-assoc; map-cong; concatMap-cong; map-concatMap;
+    concatMap-pure)
+open import Effect.Choice using (RawChoice)
+open import Effect.Empty using (RawEmpty)
+open import Effect.Functor using (RawFunctor)
 open import Effect.Applicative
+  using (RawApplicative; RawApplicativeZero; RawAlternative)
 open import Effect.Monad
-open import Function.Base
-open import Level
+  using (RawMonad; module Join; RawMonadZero; RawMonadPlus)
+open import Function.Base using (flip; _∘_; const; _$_; id; _∘′_; _$′_)
+open import Level using (Level)
 open import Relation.Binary.PropositionalEquality as ≡
   using (_≡_; _≢_; _≗_; refl)
 open ≡.≡-Reasoning
--- a/src/Data/List/Membership/Propositional/Properties.agda
+++ b/src/Data/List/Membership/Propositional/Properties.agda
@@ -37,7 +37,7 @@ open import Level using (Level)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Definitions as Binary hiding (Decidable)
 open import Relation.Binary.PropositionalEquality as ≡
-  using (_≡_; _≢_; refl; sym; trans; cong; subst; →-to-⟶; _≗_)
+  using (_≡_; _≢_; refl; sym; trans; cong; subst; _≗_)
 import Relation.Binary.Properties.DecTotalOrder as DTOProperties
 open import Relation.Unary using (_⟨×⟩_; Decidable)
 import Relation.Nullary.Reflects as Reflects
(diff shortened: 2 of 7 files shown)


src/Data/List/Effectful.agda
```diff
@@ -10,14 +10,19 @@ module Data.List.Effectful where
...
+open import Effect.Functor using (RawFunctor)
 open import Effect.Applicative
+  using (RawApplicative; RawApplicativeZero; RawAlternative)
 open import Effect.Monad
-open import Function.Base
-open import Level
+  using (RawMonad; module Join; RawMonadZero; RawMonadPlus)
+open import Function.Base using (flip; _∘_; const; _$_; id; _∘′_; _$′_)
+open import Level using (Level)
 open import Relation.Binary.PropositionalEquality as ≡
```
jamesmckinna, line 26: `.Core` here surely?

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -37,7 +37,7 @@ open import Level using (Level)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Definitions as Binary hiding (Decidable)
 open import Relation.Binary.PropositionalEquality as ≡
```
jamesmckinna, line 39: Ditto.? Or have I misunderstood something?

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -33,15 +34,15 @@ open import Data.Maybe.Relation.Unary.Any as Maybe using (just)
...
+open import Function.Bundles using (_↠_; mk↠ₛ; _⇔_; mk⇔; _↔_; mk↔ₛ′; Equivalence)
 open import Level using (Level)
 open import Relation.Binary.Core using (REL)
 open import Relation.Binary.Bundles using (Setoid)
 import Relation.Binary.Definitions as B
-open import Relation.Binary.PropositionalEquality
+open import Relation.Binary.PropositionalEquality.Core
   using (_≡_; refl; cong; cong₂; _≗_)
-open import Relation.Nullary
+open import Relation.Nullary using (¬_; does; yes; no; _because_)
```
jamesmckinna, line 45: These should be distributed over the three modules mentioned immediately below, and this import removed?
```agda
open import Relation.Nullary.Reflects using (invert)
open import Relation.Nullary.Negation using (¬_; contradiction)
open import Relation.Nullary.Decidable using (does; yes; no; _because_; ¬?; decidable-stable)```

File: src/Data/List/Effectful.agda (hunk 1 above)
Comment (JacquesCarette) [reply src/Data/List/Effectful.agda:26]: Nope: a couple of lines later there is `open ≡.≡-Reasoning` which needs the full module.

To me, that's a target for a move too (i.e. define `≡-Reasoning` in its own module, re-exported from `Relation.Binary.PropositionalEquality`, so that it can be used without dragging in so much else.)

File: src/Data/List/Membership/Propositional/Properties.agda (hunk 2 above)
Comment (JacquesCarette) [reply src/Data/List/Membership/Propositional/Properties.agda:39]: See comment above.

File: src/Data/List/Relation/Unary/All/Properties.agda (hunk 3 above)
Comment (JacquesCarette) [reply src/Data/List/Relation/Unary/All/Properties.agda:45]: Could indeed be done (if it used `.Core` for two of them) but would have very little effect on the dependency graph, as `Relation.Nullary` is fairly thin. But I guess this could be more future-proof wrt additions to `Relation.Nullary`. Sure, why not.

File: src/Data/List/Relation/Unary/All/Properties.agda [hunk 4]
```diff
@@ -11,12 +11,13 @@ module Data.List.Relation.Unary.All.Properties where
 open import Axiom.Extensionality.Propositional using (Extensionality)
 open import Data.Bool.Base using (Bool; T; true; false)
 open import Data.Bool.Properties using (T-∧)
-open import Data.Empty
+open import Data.Empty using (⊥-elim)
```
jamesmckinna, line 14: Ugh! Simultaneous import of `Data.Empty` (for `⊥-elim`) and `Relation.Nullary.Negation.Core` for `contradiction`... needs fixing at some point.

MatthewDaggitt (approved):

Issue #2330, opened 2024-03-27
move `decToMaybe` from `Data.Maybe.Base` to `Data.Maybe`


JacquesCarette:
`decToMaybe` drags in `Relation.Nullary.Reflects` and `Relation.Nullary.Decidable.Core` (which is a reasonable thing to do for its definition). But this seems like "overkill" for a `.Base` module as 'core' as for `Maybe`.

The fixes to the library are trivial - no code expects it to be that low. This, in theory, should not affect external code either, as people shouldn't be digging into `.Base` modules. But this would still be an (extremely minor) breaking change. Is that allowed for 2.1? Frankly, I don't think anyone would notice!

MatthewDaggitt:
`Base` modules are part of the public API. `Core` modules are not. So yes this would be a breaking change by our published policy, so adding this to the version 3.0 milestone.

jamesmckinna:
As discussed between me and @gallais on one of my many `Relation.Nullary` PRs, I think that `decToMaybe` belongs under `Decidable`, and *not* under `Maybe`. So probably even more so a v3.0 change, but then we could get to rename it, too...

JacquesCarette:
If we could come up with some kind of rule of thumb for where these things belong? I'm fine with this moving to `Decidable`, BTW.

Some kind of (ad hoc) ordering on concepts?

That might, for example, let us move the functions relating `Maybe` and `These` to somewhere else. It's not causing a lot of headaches, but it seems like a small wart indicative of the larger issue.

jamesmckinna:
So, for *me* at least, the ordering is induced by information loss/ornamentation/dependency:
* `Reflects` and `Decidable` are a 'richer'/'better' `Bool`, so function to `Bool` belong under `Reflects`/`Decidable` etc.
* ditto. for `Maybe`, especially if we go via `WeaklyDecidable`
* *properties*/`Pred`s are 'richer' than their underlying `Data`, so `toData` function belong under the `Pred` definition
* etc.

Is that a plausible heuristic?

jamesmckinna:
Rather than make the proposed `breaking` change, why not start more conservatively, with:
* deprecating the function in `Base` for v2.1, in favour of a definition in `Decidable`
* make the breaking change in v3.0

(Trying to channel my inner @MatthewDaggitt on this one ;-))

JacquesCarette:
[...]

PR #2336, opened 2024-03-29
Implement move-via-deprecate of `decToMaybe` as per #2330

JacquesCarette:
Make the main home of `decToMaybe` be `Relation.Nullary.Decidable.Core`, but
deprecate the one in `Data.Maybe.Base` instead of removing it entirely. Fix the library accordingly.

Note that this forces `Relation.Nullary.Decidable.Core` to use `Agda.Builtin.Maybe` to avoid a cyclic import. This can be fixed once the deprecation is done.


jamesmckinna:
Needs a `CHANGELOG`!

JacquesCarette:
Indeed it does.

jamesmckinna:
Also, is this an opportunity to follow through with @gallais 's suggestions following [his comment](https://github.com/agda/agda-stdlib/issues/2059#issuecomment-1826775133) on #2059 ? Given that once you've deprecated, you get to pick new/better names and change the API as well? or would you rather leave this to a subsequent PR? I think it would be a good idea... on whatever basis.

JacquesCarette:
`isTrue` seems sensible.

jamesmckinna:
My preference, now that it's under `Decidable`, would be #2059 's `dec⇒weaklyDec`... because all of those uses under `Solver` are about weakly deciding a property, but are written under the identification of that with `Maybe`...

... but `isBlah` is at least a conventional use. (Style guide again? Is it documented anywhere?)

I wonder if `isTrue` is a bit on the nose? (Potential conceptual clash with `True` denoting `Bool`?)

I'm really glad to be revisiting some of the choices under #2055 / #2058 / #2059 which I suspect got lost in the wash of the v2.0 release (and my mistake in tagging those as v3.0...), but it feels as though this more modest PR might be leaving downstream hostages to fortune.

Perhaps, against myself, the thing is to stick with `decToMaybe` ahead of wider consideration of the naming issues. I feel genuinely conflicted!

src/Relation/Nullary/Decidable/Core.agda
```diff
@@ -11,6 +11,9 @@
 
 module Relation.Nullary.Decidable.Core where
 
+-- this can go through `Data.Maybe.Base` once the deprecation of `decToMaybe`
+-- is fully done.
```
gallais, line 15:
```suggestion
-- decToMaybe was deprecated in v2.1
-- this can go through `Data.Maybe.Base` once the deprecation of `decToMaybe`
-- is fully done.
```
Let's reuse the magic string to make greping for things that need to be removed easier.

src/Data/Maybe/Base.agda
```diff
@@ -137,3 +131,22 @@ thisM a = maybe′ (these a) (this a)
...
+open import Relation.Nullary.Decidable.Core
+  using (_because_; Dec)
+
+decToMaybe : Dec A → Maybe A
+decToMaybe ( true because [a]) = just (invert [a])
+decToMaybe (false because  _ ) = nothing
+
+{-# WARNING_ON_USAGE decToMaybe
+"Warning: decToMaybe was deprecated in v2.1.
+Please use Relation.Nullary.Decidable's decToMaybe instead."
```
jamesmckinna, line 151: Why not simply give the full qualified name `Relation.Nullary.Decidable.decToMaybe` instead?

jamesmckinna (approved):
Having blundered in with my own commits, revising @JacquesCarette 's original PR, it's perhaps out-of-order to give an 'Approve' review... but there you have it.

JacquesCarette:
Until I hear at least one more opinion, I'm going to leave it as `decToMaybe`. I do agree its name should change!

src/Data/Maybe/Base.agda
```diff
@@ -137,3 +132,15 @@ thisM a = maybe′ (these a) (this a)
...
 thatM : Maybe A → B → These A B
 thatM = maybe′ these that
+
+------------------------------------------------------------------------
+-- Deprecated
+
+-- Version 2.1
+-- decToMaybe
+
+decToMaybe = Relation.Nullary.Decidable.Core.decToMaybe
```
MatthewDaggitt, line 142: To be honest, I think the easiest approach which would avoid all the import cycle comments would just be to entirely redefine it here. It's pretty short...

MatthewDaggitt (commented):
To be honest if we are going to change the name, I think it would be good to change it now! My main reason is that it will enhance the backwards compatbility of this change. Currently if users have both `Data.Maybe` and `Relation.Nullary` open (which is entirely possible), then they'll get a duplicate name clash.

I don't have a strong opinion on the name.

src/Data/Maybe/Base.agda
JacquesCarette: Sure, can do.

JacquesCarette:
I'm fine with renaming it, and `isYes` is indeed not the best name either. But I must say that I hate `dec⇒weaklyDec`. The `dec⇒` part is fine, it's the `weaklyDec` that I'm not happy with. What the heck is a "weak decision"? If this was long established vocabulary, sure. But I don't think that's the case. And 'weak' is ridiculously overloaded (and under-indicative).

In some ways `justYes` might be better (no, I'm not trying to pun here!) I would be fine with `dec⇒justYes`. I'm using 'just' in its 'truncate' sense here. I'm explicitly avoiding the term 'truncate' in the name, as this is a partial truncation, as that's a worse can of worms to open, IMHO.

jamesmckinna:
I'd cheerfully take responsibility for the bad name, had I introduced it, but it seems to have been written into the library a long time ago.  

I sympathise with whoever first chose "weakly"  though: Johnstone in Topos Theory (1977) called `1+_` the "decidable partial map classifier" and there seems no easy way to extract a reasonable qualified adjective out of that, given that the other ready-at-hand prefixes "semi-" and "partial-" both connote the more established recursion theoretic 'yes definitely but may not terminate if no'... and recursion theory doesn't seem to have an agreed orthodox name for this useful middle ground? 

But v3.0 as a holding pen for reconsiderations of this name, and v2.x for continuing with it for the time being?

That said, the writers of the various `Solver`s seemed to bypass the issue entirely by working with `Maybe` directly, which is a different kind of (category) mistake IMHO (one of 'data vs. property'), so `WeaklyDecidable` has not really caught on. 

I think for the sake of this PR, and its focus on simplifying dependencies, suggest sticking with the name, even if you hate it?

src/Algebra/Solver/IdempotentCommutativeMonoid.agda
```diff
@@ -26,7 +26,7 @@ import Relation.Nullary.Decidable            as Dec
 import Data.Vec.Relation.Binary.Pointwise.Inductive as Pointwise
 
 open import Relation.Binary.PropositionalEquality as ≡ using (_≡_; decSetoid)
-open import Relation.Nullary.Decidable using (Dec)
+open Dec using (Dec; decToMaybe)
```
jamesmckinna, line 29: For consistency with `Algebra.Solver.Monoid`, which uses `Dec.decToMaybe` qualified, maybe stick with the old `import` but with the qualified name `Dec`? or else expose `decToMaybe` at top-level in `...Monoid`?

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Solver.agda
```diff
@@ -30,11 +30,11 @@ open import Data.List.Properties
 open import Data.List.Relation.Binary.Sublist.Heterogeneous
   hiding (lookup)
 open import Data.List.Relation.Binary.Sublist.Heterogeneous.Properties
-open import Function
+open import Function.Base
 
 open import Relation.Binary.PropositionalEquality as ≡
   using (_≡_; _≗_; sym; cong; cong₂; subst₂)
-open import Relation.Nullary
+open import Relation.Nullary.Decidable.Core using (decToMaybe)
```
jamesmckinna, line 37: Similarly to above...

jamesmckinna:
I took the liberty of resolving the merge conflict, but without committing one way or the other to my suggestions regarding qualified names above.

src/Data/Maybe/Base.agda
JacquesCarette: Doesn't actually help: the direct definition still need `Relation.Nullary.Decidable.Core` to be imported.

jamesmckinna:
Continues to look good!
Can we get a second review, please?

jamesmckinna:
> Meeting conclusion:
> 
>     * Either change deprecation strategy to public re-export `Nullary.Core` name instead of redefining.
> 
>     * Or rename `Nullary.Core` definition

Suggest doing the first for v2.1 (hence, with no deprecation, IIUC), and deciding on the second (with consequent deprecation) as a downstream v2.2?

jamesmckinna:
Apologies! the last commit appears to have broken things, because `Relation.Binary.Consequences`, *despite* importing form the 'right' place, nevertheless triggers the deprecation warning...

JacquesCarette:
Implemented @MatthewDaggitt 's first deprecation strategy. This should be ready to go now.

JacquesCarette:
Yes, that's my thinking.

jamesmckinna:
> Yes, that's my thinking.

Happy to push this if your hands are full!

jamesmckinna:
> I'm fine with isYes and isNo.

Aaargh! Wasn't thinking...

... `isYes` is the name of the projection from `Dec` to `Bool` %!$#

jamesmckinna:
Moved name discussion back to the issue... sigh.

MatthewDaggitt:
Yes, as I said in the meeting, you can't directly attach deprecation warnings to re-exports. As @jamesmckinna suggests we'd have to remove the deprecation warning entirely and leave just a textual comment that it is deprecated. This is what we've done in the past in similar circumstances.

jamesmckinna:
Looking at the details of this PR, much of the knock-on viscosity arising from any name change to `decToMaybe` could be dealt with by fixing the various DRY violations under `Algebra.Solver.*Monoid`, and by delegation from a given `Relation.Binary.Definitions.DecidableEquality` via `Relation.Binary.Consequences.dec⇒weaklyDec`... (and we can fix *that* name downstream...)

JacquesCarette:
Yes, this works. I'm fine with `dec⇒maybe`. Thanks.

MatthewDaggitt (approved):

PR #2337, opened 2024-04-01
Remove uses of `Data.Nat.Solver` from a number of places

JacquesCarette:
Various places gratuitously used `Data.Nat.Solver` when some fairly simple direct proofs could be used instead. This made `Data.Rational` (amongst others) very 'high' on the dependency graph when this was not needed.

### DIFF
--- a/src/Data/Digit.agda
+++ b/src/Data/Digit.agda
@@ -103,7 +103,7 @@ toDigits base@(suc (suc k)) n = <′-rec Pred helper n
       x * 2+ k              ∎) ⟩
     k + x * 2+ k          ∎
     where open ≡-Reasoning
-  
+
   lem : ∀ x k r → 2 + x ≤′ r + (1 + x) * (2 + k)
   lem x k r = ≤⇒≤′ $ begin
     2 + x                             ≤⟨ m≤m+n _ _ ⟩
--- a/src/Data/Digit.agda
+++ b/src/Data/Digit.agda
@@ -9,6 +9,7 @@
 module Data.Digit where
 
 open import Data.Nat.Base
+  using (ℕ; zero; suc; _<_; _/_; _%_; sz<ss; _+_; _*_; 2+; _≤′_)
 open import Data.Nat.Properties
   using (_≤?_; _<?_; ≤⇒≤′; module ≤-Reasoning; m≤m+n; +-comm; +-assoc;
     *-distribˡ-+; *-identityʳ)
--- a/src/Data/Digit.agda
+++ b/src/Data/Digit.agda
@@ -9,20 +9,23 @@
 module Data.Digit where
 
 open import Data.Nat.Base
-open import Data.Nat.Properties using (_≤?_; _<?_; ≤⇒≤′; module ≤-Reasoning; m≤m+n)
-open import Data.Nat.Solver using (module +-*-Solver)
+open import Data.Nat.Properties
+  using (_≤?_; _<?_; ≤⇒≤′; module ≤-Reasoning; m≤m+n; +-comm; +-assoc;
+    *-distribˡ-+; *-identityʳ)
 open import Data.Fin.Base as Fin using (Fin; zero; suc; toℕ)
 open import Data.Bool.Base using (Bool; true; false)
 open import Data.Char.Base using (Char)
-open import Data.List.Base
+open import Data.List.Base using (List; replicate; [_]; _∷_; [])
 open import Data.Product.Base using (∃; _,_)
 open import Data.Vec.Base as Vec using (Vec; _∷_; [])
-open import Data.Nat.DivMod
+open import Data.Nat.DivMod using (m/n<m; _divMod_; result)
 open import Data.Nat.Induction
+  using (Acc; acc; <-wellFounded-fast; <′-Rec; <′-rec)
+open import Function.Base using (_$_)
 open import Relation.Nullary.Decidable using (True; does; toWitness)
 open import Relation.Binary.Definitions using (Decidable)
-open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl; sym; cong)
-open import Function.Base using (_$_)
+open import Relation.Binary.PropositionalEquality
+  using (_≡_; refl; sym; cong; cong₂; module ≡-Reasoning)
 
 ------------------------------------------------------------------------
 -- Digits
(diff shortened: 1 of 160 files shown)


src/Data/Integer/Properties.agda
```diff
@@ -1400,26 +1390,10 @@ private
...
-  solve 3 (λ m n o → m :+ (n :+ (con 1 :+ o)) :* (con 1 :+ m)
-                   := (con 1 :+ n) :* (con 1 :+ m) :+
-                      (m :+ o :* (con 1 :+ m)))
-         refl m n o
-*-distribʳ-+ +[1+ m ] -[1+ n ] -[1+ o ] = cong -[1+_] $
-  solve 3 (λ m n o → m :+ (con 1 :+ m :+ (n :+ o) :* (con 1 :+ m))
-                  := (con 1 :+ n) :* (con 1 :+ m) :+
-                     (m :+ o :* (con 1 :+ m)))
-         refl m n o
+*-distribʳ-+ -[1+ m ] -[1+ n ] -[1+ o ] = cong (+_) $ assoc₁ m n o
```
MatthewDaggitt, line 1393: I'm not sure whether having half the cases in one file and the other set of equally long cases in this file, is ideal? I think it would be better (everywhere!) just to expand out the cases locally?

I know GCD.Lemmas doesn't do so, but that is a very old file?

src/Data/Integer/Properties.agda
```diff
@@ -1354,14 +1344,14 @@ private
...
-*-assoc +[1+ m ] -[1+ n ] +[1+ o ] = cong -[1+_] (lemma m n o)
-*-assoc +[1+ m ] +[1+ n ] -[1+ o ] = cong -[1+_] (lemma m n o)
+*-assoc -[1+ m ] -[1+ n ] +[1+ o ] = cong (+_ ∘ suc) (inner-assoc m n o)
+*-assoc -[1+ m ] +[1+ n ] -[1+ o ] = cong (+_ ∘ suc) (inner-assoc m n o)
+*-assoc +[1+ m ] +[1+ n ] +[1+ o ] = cong (+_ ∘ suc) (inner-assoc m n o)
+*-assoc +[1+ m ] -[1+ n ] -[1+ o ] = cong (+_ ∘ suc) (inner-assoc m n o)
+*-assoc -[1+ m ] -[1+ n ] -[1+ o ] = cong -[1+_] (inner-assoc m n o)
+*-assoc -[1+ m ] +[1+ n ] +[1+ o ] = cong -[1+_] (inner-assoc m n o)
+*-assoc +[1+ m ] -[1+ n ] +[1+ o ] = cong -[1+_] (inner-assoc m n o)
+*-assoc +[1+ m ] +[1+ n ] -[1+ o ] = cong -[1+_] (inner-assoc m n o)
```
MatthewDaggitt, line 1354: Likewise here better to have a local lemma?

MatthewDaggitt (commented):
> Various places gratuitously used Data.Nat.Solver when some fairly simple direct proofs could be used instead.

Having looked at the changes, I'm not sure if some of these can be called "simple". I wonder if a parallel effort might be to reduce the dependency foot print of the solvers instead? Fundamentally the solvers should use be able to use the boolean decision procedures in the `Base` modules rather than `Dec` decision procedures in `Properties` right?

Taneb:
> > Various places gratuitously used Data.Nat.Solver when some fairly simple direct proofs could be used instead.
> 
> Having looked at the changes, I'm not sure if some of these can be called "simple". I wonder if a parallel effort might be to reduce the dependency foot print of the solvers instead? Fundamentally the solvers should use be able to use the boolean decision procedures in the `Base` modules rather than `Dec` decision procedures in `Properties` right?

You'd still need all the algebra laws, which still brings in `.Properties`...

MatthewDaggitt:
> You'd still need all the algebra laws, which still brings in .Properties...

Hmm good point!

JacquesCarette:
> Having looked at the changes, I'm not sure if some of these can be called "simple". 

Compared to a lot of proofs in `agda-categories`, they are... but point taken. Though I'd say that many of them are no worse than some of the monsters in the same files. (`Data.Integer.Properties` is seriously frightening; I'll open some issues later.)

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -1124,11 +1123,13 @@ p≤q⇒0≤q-p {p} {q} p≤q = begin
 *-inverseˡ p@(mkℚᵘ +[1+ n ] d) = *≡* $ cong +[1+_] $ begin
   (n ℕ.+ d ℕ.* suc n) ℕ.* 1 ≡⟨ ℕ.*-identityʳ _ ⟩
   (n ℕ.+ d ℕ.* suc n)       ≡⟨ cong (n ℕ.+_) (ℕ.*-suc d n) ⟩
-  (n ℕ.+ (d ℕ.+ d ℕ.* n))   ≡⟨ solve 2 (λ n d → n :+ (d :+ d :* n) := d :+ (n :+ n :* d)) refl n d ⟩
+  (n ℕ.+ (d ℕ.+ d ℕ.* n))   ≡⟨ trans (sym $ ℕ.+-assoc n d _) (trans
```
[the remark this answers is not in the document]
JacquesCarette: Will indeed get rid of them when I spot them.

src/Data/Integer/Properties.agda
JacquesCarette: I'm fine with pulling these in.

I do think if we had a big `Lemmas` file, we'd notice quite a few patterns, and could shrink things. But that can be a v3.0 kind of task.

JacquesCarette: And I now remember why I put these in a separate file: inside `Integer.Properties`, `+` and `*` are for integers, and the ones for naturals are 'long'. So these proofs end up looking much, much worse. But inside a different context, i.e. such as a separate module, this is not a problem.

So, in looking again, I'd rather not do this.

src/Data/Digit.agda
```diff
@@ -87,20 +90,32 @@ toDigits base@(suc (suc k)) n = <′-rec Pred helper n
...
+  lem′ x k = begin
+    x + x + (k + x * k)   ≡⟨ +-assoc (x + x) k _ ⟨
+    x + x + k + x * k     ≡⟨ cong (_+ x * k) (+-comm _ k) ⟩
+    k + (x + x) + x * k   ≡⟨ +-assoc k (x + x) _ ⟩
+    k + ((x + x) + x * k) ≡⟨ cong (k +_) (begin
+      (x + x) + x * k       ≡⟨ +-assoc x x _ ⟩
+      x + (x + x * k)       ≡⟨ cong (x +_) (cong (_+ x * k) (*-identityʳ x)) ⟨
+      x + (x * 1 + x * k)   ≡⟨ cong₂ _+_ (*-identityʳ x) (*-distribˡ-+ x 1 k ) ⟨
+      x * 1 + (x * suc k)   ≡⟨ *-distribˡ-+ x 1 (1 + k) ⟨
+      x * 2+ k              ∎) ⟩
```
gallais, line 103: The alignment of the LHSs got messed up here.

src/Data/Fin/Properties.agda
```diff
@@ -669,12 +668,14 @@ toℕ-combine {suc m} {n} i@0F j = begin
...
-  toℕ (n ↑ʳ combine i j)         ≡⟨ toℕ-↑ʳ n (combine i j) ⟩
-  n ℕ.+ toℕ (combine i j)        ≡⟨ cong (n ℕ.+_) (toℕ-combine i j) ⟩
-  n ℕ.+ (n ℕ.* toℕ i ℕ.+ toℕ j)  ≡⟨ solve 3 (λ n i j → n :+ (n :* i :+ j) := n :* (con 1 :+ i) :+ j) refl n (toℕ i) (toℕ j) ⟩
-  n ℕ.* toℕ (suc i) ℕ.+ toℕ j    ∎
-  where open ≡-Reasoning; open +-*-Solver
+  toℕ (combine (suc i) j)            ≡⟨⟩
+  toℕ (n ↑ʳ combine i j)             ≡⟨ toℕ-↑ʳ n (combine i j) ⟩
+  n ℕ.+ toℕ (combine i j)            ≡⟨ cong (n ℕ.+_) (toℕ-combine i j) ⟩
+  n ℕ.+ (n ℕ.* toℕ i ℕ.+ toℕ j)     ≡⟨ ℕ.+-assoc n _ (toℕ j) ⟨
+  n ℕ.+ n ℕ.* toℕ i ℕ.+ toℕ j       ≡⟨ cong (λ z → z ℕ.+ n ℕ.* toℕ i ℕ.+ toℕ j) (ℕ.*-identityʳ n) ⟨
```
gallais, line 675: Alignment issues here too (for the reasoning combinators this time).

src/Data/Digit.agda
JacquesCarette: Actually, this is intentional -- this is a nested proof, see the 'begin' on the line above.

src/Data/Fin/Properties.agda
JacquesCarette: That's odd, I don't see any alignment problems in my emacs!

src/Data/Integer/Properties.agda
MatthewDaggitt: Okay, I withdraw my objections.

MatthewDaggitt (approved):
Once the merge conflict is resolved happy to merge.

JacquesCarette:
(Normally don't merge my own, but I take the above as permission to do so rather than waiting.)

PR #2340, opened 2024-04-01
Another tweak to cong!

uncle-betty:
Hello there.

I do a lot of equational reasoning with `_/_` and `_%_`. This means `NonZero` instance arguments on both sides of my equalities. Sometimes, when `cong!` runs, the metas for the instance arguments are already resolved. But sometimes they aren't. Or one of them is, but the other isn't. It depends on how things fall into place during type-checking.

When there are such unresolved metas, `cong!` fails. For example, even an identical instance argument gets a different meta on the LHS vs. the RHS. So, anti-unification considers them a difference and places a phi. Same for an unresolved meta on one side and a resolved meta on the other.

The solution that I would like to propose is to wait for the meta variables to be resolved, before doing anti-unification. However, checking the goal for the existence of metas comes with a cost: the goal needs to be traversed, which impacts `cong!` performance.

This proposal thus only checks for metas once unification failed. The idea is to avoid burdening other people with the meta check, who might not use `_/_` and `_%_` as much as I do, and who thus might not have this problem.

In addition to the meta check, this proposal tries to make the error message issued by a failing `cong!` more meaningful to the user. `cong!` now tells them which term it tried, so that the user can troubleshoot the situation. Consider the following:
```
lemma : ∀ m n o → m + n + o ≡ n + m + o
lemma m n o = begin-equality
  m + n + o ≡⟨ cong! (+-comm m n) ⟩
  n + m + o ∎
  where open ≤-Reasoning
```
This now yields the following error:
```
cong! failed, tried:
cong (λ ϕ → ϕ + ϕ + o) (+-comm m n)
```
Previously the error was the following, which I've always found confusing:
```
m != m + n of type ℕ
```

### DIFF
--- a/doc/README/Tactic/Cong.agda
+++ b/doc/README/Tactic/Cong.agda
@@ -3,6 +3,7 @@
 module README.Tactic.Cong where
 
 open import Data.Nat
+open import Data.Nat.DivMod
 open import Data.Nat.Properties
 
 open import Relation.Binary.PropositionalEquality
@@ -160,4 +161,35 @@ module EquationalReasoningTests where
       suc (n + n)
     ≤⟨ n≤1+n _ ⟩
       suc (suc (n + n))
+    ∎
+
+module MetaTests where
+  test₁ : ∀ m n o → .⦃ _ : NonZero o ⦄ → (m + n) / o ≡ (n + m) / o
+  test₁ m n o =
+    let open ≤-Reasoning in
+    begin-equality
+      ⌞ m + n ⌟ / o
+     ≡⟨ cong! (+-comm m n) ⟩
+      (n + m) / o
+    ∎
+
+  test₂ : ∀ m n o p q r → .⦃ _ : NonZero o ⦄ → .⦃ _ : NonZero p ⦄ →
+          .⦃ _ : NonZero q ⦄ → p ≡ q ^ r → (m + n) % o % p ≡ (n + m) % o % p
+  test₂ m n o p q r eq =
+    let
+      open ≤-Reasoning
+      instance q^r≢0 = m^n≢0 q r
+    in
+    begin-equality
+      (m + n) % o % p
+     ≡⟨ %-congʳ eq ⟩
+      ⌞ m + n ⌟ % o % q ^ r
+     ≡⟨ cong! (+-comm m n) ⟩
+      ⌞ n + m ⌟ % o % q ^ r
+     ≡⟨ cong! (+-comm m n) ⟨
+      ⌞ m + n ⌟ % o % q ^ r
+     ≡⟨ cong! (+-comm m n) ⟩
+      (n + m) % o % q ^ r
+     ≡⟨ %-congʳ eq ⟨
+      (n + m) % o % p
     ∎
--- a/doc/README/Tactic/Cong.agda
+++ b/doc/README/Tactic/Cong.agda
@@ -164,6 +164,7 @@ module EquationalReasoningTests where
     ∎
 
 module MetaTests where
+
   test₁ : ∀ m n o → .⦃ _ : NonZero o ⦄ → (m + n) / o ≡ (n + m) / o
   test₁ m n o =
     let open ≤-Reasoning in
--- a/src/Tactic/Cong.agda
+++ b/src/Tactic/Cong.agda
@@ -31,13 +31,12 @@ open import Function.Base using (_$_)
 open import Data.Bool.Base            using (true; false; if_then_else_; _∧_)
 open import Data.Char.Base   as Char  using (toℕ)
 open import Data.Float.Base  as Float using (_≡ᵇ_)
-open import Data.List.Base   as List  using (List; []; _∷_; _++_)
+open import Data.List.Base   as List  using ([]; _∷_)
 open import Data.Maybe.Base  as Maybe using (Maybe; just; nothing)
 open import Data.Nat.Base    as ℕ     using (ℕ; zero; suc; _≡ᵇ_; _+_)
-open import Data.Unit.Base            using (⊤; tt)
+open import Data.Unit.Base            using (⊤)
 open import Data.Word.Base   as Word  using (toℕ)
 open import Data.Product.Base         using (_×_; map₁; _,_)
-open import Effect.Applicative        using (RawApplicative; mkRawApplicative)
 open import Function                  using (flip; case_of_)
 
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl; cong)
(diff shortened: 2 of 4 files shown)


JacquesCarette (approved):
Seems very sensible to me - but I'm also seriously not up-to-date on Agda's reflection, so my approval should be taken with a big lump of salt.

src/Tactic/Cong.agda
```diff
@@ -132,6 +146,21 @@ private
...
+
+      open Traversal applicative
+
+      actions : Actions
+      actions = record defaultActions { onMeta = λ _ x → x ∷ [] }
+
+    blockOnMetas : Term → TC ⊤
+    blockOnMetas t with traverseTerm actions (0 , []) t
+    ... | []         = pure tt
+    ... | xs@(_ ∷ _) = blockTC (blockerAll (List.map blockerMeta xs))
```
MatthewDaggitt, line 162: This feels like this (or part of it) could be extracted to a utility method in `Reflection.AST.Traversal` itself?

MatthewDaggitt (changes requested):
Otherwise looks great, and the reasoning seems sound to me! Can you add some test cases to the README file to document the behaviour and guard against regressions? Also could you add a CHANGELOG entry describing the enhancement?

src/Tactic/Cong.agda
```diff
@@ -102,6 +106,16 @@ private
...
   notEqualityError goal = typeError (strErr "Cannot rewrite a goal that is not equality: " ∷ termErr goal ∷ [])
 
+  unificationError : ∀ {A : Set} → TC Term → TC Term → TC A
+  unificationError term symTerm = do
+    term' ← term
+    symTerm' ← symTerm
+    -- Don't show the same term twice.
+    let symErr = case term' Term.≟ symTerm' of λ where
+      (yes _) → []
+      (no _) → strErr "\n" ∷ termErr symTerm' ∷ []
```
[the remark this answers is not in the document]
gallais: Yes you do.

src/Tactic/Cong.agda
```diff
@@ -236,12 +265,17 @@ macro
...
+      let makeTerm = λ lhs rhs → `cong eqGoal (antiUnify 0 lhs rhs) x≡y
       let lhs = EqualityGoal.lhs eqGoal
       let rhs = EqualityGoal.rhs eqGoal
-      -- When using ⌞_⌟ with ≡⟨ ... ⟨, (uni lhs rhs) fails and
-      -- (uni rhs lhs) succeeds.
-      catchTC (uni lhs rhs) (uni rhs lhs)
+      let term = makeTerm lhs rhs
+      let symTerm = makeTerm rhs lhs
+      let uni = _>>= flip unify hole
+      catchTC
```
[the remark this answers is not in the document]
uncle-betty: Yes! And it should.

src/Tactic/Cong.agda
uncle-betty: I like the idea! Other tactics might want to block on metas, too. But please help me understand: The `Traversal` module takes an applicative as an argument. Wouldn't a user then have to provide a "dummy" applicative in order to use `blockOnMetas`? (Which would ignore the user-provided applicative.)

uncle-betty: What I'm trying to say: Should there be a new module, `Reflection.AST.Utilities`, that would be home to `blockOnMetas`?

uncle-betty:
Thank you for your reviews!

I think I might need some guidance regarding the `blockOnMetas` (see my response to the comment), but I made the other changes.

The idea behind the `test₂` unit test, which may not be completely obvious:

* Anecdotally, I seem to have encountered issues mostly with reasoning chains that were a little longer. Hence the redundant steps.
* It tests explicitly given instances in addition to instances from instance arguments.

doc/README/Tactic/Cong.agda
```diff
@@ -161,3 +162,34 @@ module EquationalReasoningTests where
     ≤⟨ n≤1+n _ ⟩
       suc (suc (n + n))
     ∎
+
+module MetaTests where
```
MatthewDaggitt, line 166: Should be an empty line here.

src/Tactic/Cong.agda
MatthewDaggitt: Ah sorry, I missed that `Traversal` is parameterised by an applicative. Yes a new module `Reflection.TCM.Utilities` would be good. (Note I've had a change of heart from `AST` to `TCM`!)

doc/README/Tactic/Cong.agda
uncle-betty: Oh, right. Yes.

src/Tactic/Cong.agda
uncle-betty: Alright. Great. I moved it.

uncle-betty:
Thanks for taking another look. I made the changes.

jamesmckinna:
Coming to this PR quite late...
... if you find yourself doing a lot of equational reasoning with `_/_` and `_%_` then is there a higher level (as opposed to low-level `cong` reasoning) collection of `DivMod` abstractions which might streamline your work?

uncle-betty:
Excellent question. I don't know yet. I've only started toying around.

I'm curious about modeling machine integers in Agda: convert integers modulo 2^64 to `Vec Bool 64` and vice versa, prove that division by 2 of an integer corresponds to a right shift of the corresponding bit vector - this kind of thing.

Yes, there may be a much better way than what I'm currently doing. But I still think that there's benefit in making `cong!` deal properly with meta variables in goals.

jamesmckinna:
@uncle-betty you wrote: 
> I'm curious about modeling machine integers in Agda: convert integers modulo 2^64 to `Vec Bool 64` and vice versa, prove that division by 2 of an integer corresponds to a right shift of the corresponding bit vector - this kind of thing.

Makes good sense, but then I would (be tempted to) avoid `_/_` and `_%_` in favour of using the `Data.Nat.Base.⌊_/2⌋` and `Data.Nat.Base.⌈_/2⌉` functions and their properties, and perhaps consider iterates of them for `m-`ary shifts? If you're only considering base-`2^n` arithmetic, I think full 'div-mod' is overkill?

> Yes, there may be a much better way than what I'm currently doing. But I still think that there's benefit in making `cong!` deal properly with meta variables in goals.

Absolutely fair point.

uncle-betty:
Oh, nice! Thanks, @jamesmckinna, for pointing me to these functions. I appreciate the guidance.

MatthewDaggitt (approved):

MatthewDaggitt:
Apologies for the delay. Merging in now. Thanks for another great PR!

uncle-betty:
Oh, that's exciting! Great! Thank you for the merge and for your assistance.

PR #2342, opened 2024-04-03
Port two Endomorphism submodules over to new Function hierarchy

### LINKED ISSUE #2299: Update `Function.Endomorphism.*` to avoid use of deprecated `Function.Equality`
The #759 deprecation effort left some residues which still import (at least) `Function.Equality` (with a pragma to suppress the warning), notably:
- [x] `Function.Endomorphism.Setoid`
- [x] `Function.Endomorphism.Propositional`

which define the monoid etc. of endomorphisms (needed for [a development of `Foldable`](https://github.com/agda/agda-stdlib/pull/2300) cf. generalising #1281 , and also now #2350 ). 

UPDATED: ~~blocked on not-yet-merged #2240~~ each of the above also uses the long-since-deprecated (v1.5!) `Algebra.Morphism.Is*Morphism` definitions instead of updating to the ones in `Algebra.Morphism.Structures` ...

Should update to remove dependency on the deprecated module(s).

JacquesCarette:
Closes #2299 

Needed to add some missing declarations in `Algebra.Morphism.Structures` to make it all go through.


src/Algebra/Morphism/Structures.agda
```diff
@@ -68,6 +68,14 @@ module MagmaMorphisms (M₁ : RawMagma a ℓ₁) (M₂ : RawMagma b ℓ₂) wher
       ; surjective     = surjective
       }
 
+  IsSemigroupHomomorphism : (⟦_⟧ : A → B) → Set (a ⊔ ℓ₁ ⊔ ℓ₂)
+  IsSemigroupHomomorphism = IsMagmaHomomorphism
```
MatthewDaggitt, line 72: I'm slightly reluctant to go down this path, as this opens up to us adding a crazy number of synonyms for *all* the structures/bundles in `Algebra.Bundles`. This is the same reason why we don't have `RawSemigroup`...

MatthewDaggitt (changes requested):
Gah, I can't believe I missed these.

Unfortunately, I hate to sound like a broken record but these are breaking changes. I know that *I* messed up in v2.0 by not migrating these across, but that doesn't necessarily mean that we can do it v2.1.

As a fair few names are changing, It looks like some of this can be done in a backwards compatible manner by adding deprecation warnings to the old names (e.g. `^-isSemigroupMorphism`).

Can we separate this PR out into a two parts, a set of backwards compatible changes, and a set of non-backwards compatible changes and then we can evaluate from there?

src/Algebra/Morphism/Structures.agda
JacquesCarette: So the funny reason for adding this one is that the *deprecation* warning for `IsSemingroupMorphism` mentioned `IsSemigroupHomomorphism`... even though it didn't in fact exist!

Thus I decided to make the deprecation warning less of a lie.

I'm fine with not adding a crazy number of synonyms.

MatthewDaggitt: Ah my fault again. I guess we need to update the deprecation warning to `IsMagmaHomomorphism`...

JacquesCarette:
> Can we separate this PR out into a two parts, a set of backwards compatible changes, and a set of non-backwards compatible changes and then we can evaluate from there?

Sure. But I've never worried much about backwards compatibility, so I'm entirely sure which ones those are! I'd appreciate a little bit of guidance, and then I will do that.

src/Algebra/Morphism/Structures.agda
JacquesCarette: Done in PR #2344 . I'll also 'fix' this PR in that way, though I know I'll still need to split it.

jamesmckinna:
> Gah, I can't believe I missed these.

I think it's not so surprising, given that explicit pragmas had been inserted precisely to suppress the deprecation warnings... (hint: sometimes necessary, but probably harmful overall...)

In any case, the push to close #759 before v2.0 had been sufficiently effortful that there were bound to be 'loose ends'. 

This one, in any case, only emerged when I started wanting to thinking about `Endo` at the `Setoid` level... I think it often requires change in one place to 'provoke' reconsideration of older decisions, so we shouldn't feel too bad about that?

JacquesCarette:
The question remains: what should this PR actually look like? I'm having a hard time seeing the 'two parts' clearly, for some reason.

And yeah, there's a very high cost to having pragmas to suppress deprecation warnings. I would be more tempted to filter them out post-facto rather than with 'global' pragmas. [If we had more precise pragmas, that would work too.]

CHANGELOG.md
```diff
@@ -173,7 +180,7 @@ Additions to existing modules
   rawModule          : RawModule R c ℓ
   ```
 
-* In `Algebra.Properties.Group`:
+ * In `Algebra.Properties.Group`:
```
jamesmckinna, line 183: This looks like spurious whitespace?

src/Function/Endomorphism/Propositional.agda
```diff
@@ -84,17 +84,24 @@ f ^ suc n = f ∘′ (f ^ n)
...
 -- Homomorphism
 
-^-isSemigroupMorphism : ∀ f → IsSemigroupMorphism +-semigroup ∘-semigroup (f ^_)
+^-isSemigroupMorphism : ∀ f → IsMagmaHomomorphism +-rawMagma ∘-rawMagma (f ^_)
 ^-isSemigroupMorphism f = record
-  { ⟦⟧-cong = cong (f ^_)
-  ; ∙-homo  = ^-homo f
+  { isRelHomomorphism = record { cong = cong (f ^_) }
+  ; homo = ^-homo f
   }
```
jamesmckinna, line 101: It's not clear to me why this name is being retained, rather than being deprecated in favour of:
```agda
^-isMagmaHomomorphism : ∀ f → IsMagmaHomomorphism +-rawMagma ∘-rawMagma (f ^_)
```

src/Function/Endomorphism/Setoid.agda
```diff
@@ -6,90 +6,110 @@
...
+open import Function.Construct.Identity using () renaming (function to identity)
+open import Function.Construct.Composition using () renaming (function to _∘_)
+open import Function.Relation.Binary.Setoid.Equality as Eq using (_⇨_)
+open import Level using (Level; _⊔_)
+open import Relation.Binary.Core using (_Preserves_⟶_)
 
 private
-  module E = Setoid (setoid S (Trivial.indexedSetoid S))
+  module E = Setoid (S ⇨ S)
   open E hiding (refl)
```
jamesmckinna, line 33: Given the semantics are the same, IIUC, why not telescope this into:
```agda
open module E = Setoid (S ⇨ S) hiding (refl)
```

src/Function/Endomorphism/Setoid.agda
```diff
@@ -6,90 +6,110 @@
...
 -- Structures
 
 ∘-isMagma : IsMagma _≈_ _∘_
 ∘-isMagma = record
   { isEquivalence = isEquivalence
-  ; ∙-cong        = λ g f x → g (f x)
+  ; ∙-cong        = λ {_} {_} {_} {v} x≈y u≈v → S.trans u≈v (cong v x≈y)
   }
+  where
+    module S = Setoid S
```
jamesmckinna, line 68: Suggest adding this to the `private` block L43-45 and then using `S.refl` etc. subsequently throughout?

src/Function/Endomorphism/Propositional.agda
jamesmckinna: Also: can this be made an instance of the corresponding construction in `Setoid`?

CHANGELOG.md
```diff
@@ -36,6 +36,13 @@ Deprecated modules
 * `Data.List.Relation.Binary.Sublist.Propositional.Disjoint` deprecated in favour of
   `Data.List.Relation.Binary.Sublist.Propositional.Slice`.
 
+Un-deprecated modules
+---------------------
+
+* The modules `Function.Endomorphism.Propositional` and
+  `Function.Endomorphism.Setoid` previously used the old `Function`
+  hierarchy but have now been ported to using the new one.
```
jamesmckinna, line 44: Nice addition to the `CHANGELOG` rubric... and definitely preferable to inventing bogus 'alternative' names...?

jamesmckinna:
Re: @MatthewDaggitt 's "non-backwards compatible", I'm having trouble seeing what is such about @JacquesCarette 's "undeprecation" strategy. Or is something else going on that I have missed?

src/Function/Endomorphism/Propositional.agda
```diff
@@ -5,21 +5,21 @@
...
-open import Algebra.Morphism; open Definitions
+open import Algebra using (Semigroup; Magma; RawMagma; Monoid; RawMonoid)
+open import Algebra.Core
+open import Algebra.Structures using (IsMagma; IsSemigroup; IsMonoid)
+open import Algebra.Morphism
+  using (module Definitions; IsMagmaHomomorphism; IsMonoidHomomorphism)
+open Definitions using (Homomorphic₂)
 
-open import Data.Nat.Base using (ℕ; zero; suc; _+_)
+open import Data.Nat.Base using (ℕ; _+_; zero; suc; +-rawMagma; +-0-rawMonoid)
```
jamesmckinna, line 17: Why promote `_+_` in the `using` list? Surely it 'follows' `zero` and `suc` in conceptual priority, as originally?

src/Function/Endomorphism/Setoid.agda
```diff
@@ -6,90 +6,110 @@
...
-open import Function.Equality using (setoid; _⟶_; id; _∘_; cong)
-open import Data.Nat.Base using (ℕ; _+_); open ℕ
-open import Data.Nat.Properties
+open import Algebra using (Semigroup; Magma; RawMagma; Monoid; RawMonoid)
+open import Algebra.Structures using (IsMagma; IsSemigroup; IsMonoid)
+open import Algebra.Morphism
+  using (module Definitions; IsMagmaHomomorphism; IsMonoidHomomorphism)
+open Definitions using (Homomorphic₂)
+open import Data.Nat.Base using (ℕ; _+_; +-rawMagma; +-0-rawMonoid)
+open ℕ
```
jamesmckinna, line 21: This feels a bit 'occult' to me, given that you do *not* explicitly write `using (... module ℕ ...)` on the line above? And is this solely for the sake of the constructors?

JacquesCarette: Note how this occult practice was already in the original, I just made it clearer by inserting a line break!

However, I should indeed edit this to remove such shenanigans, while I'm "in" the file anyways.

src/Function/Endomorphism/Propositional.agda
JacquesCarette: I have no idea why I did that. Will revert.

CHANGELOG.md
JacquesCarette: Yup - fixed.

src/Function/Endomorphism/Setoid.agda
JacquesCarette: Good catch - done.

src/Function/Endomorphism/Setoid.agda
JacquesCarette: Yes, I should have spotted that, thanks. Done.

src/Function/Endomorphism/Propositional.agda
JacquesCarette: I've changed to the better name and deprecated the old one. And no to the instance: the types are quite different (you have to expanded them out to see it), so it's not immediate.

src/Function/Endomorphism/Setoid.agda
```diff
@@ -6,90 +6,110 @@
...
+
 ------------------------------------------------------------------------
 -- Homomorphism
 
-^-isSemigroupMorphism : ∀ f → IsSemigroupMorphism +-semigroup ∘-semigroup (f ^_)
-^-isSemigroupMorphism f = record
-  { ⟦⟧-cong = ^-cong₂ f
-  ; ∙-homo  = ^-homo f
+^-isSemigroupHomomorphism : ∀ f → IsMagmaHomomorphism +-rawMagma ∘-rawMagma (f ^_)
+^-isSemigroupHomomorphism f = record
```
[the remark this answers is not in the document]
JacquesCarette: Done.

src/Algebra/Morphism/Structures.agda
```diff
@@ -68,7 +68,6 @@ module MagmaMorphisms (M₁ : RawMagma a ℓ₁) (M₂ : RawMagma b ℓ₂) wher
       ; surjective     = surjective
       }
 
-
```
jamesmckinna, line 71: The `style-guide` indicates that there *should* be two blank lines after each internal `module`, although that rule is not observed rigidly throughout even this module, and in any case, arguably, the big section headings function more than adequately as separators... so not sure what to do here?

jamesmckinna (approved):
Otherwise all looking good, as per my earlier comments!

src/Algebra/Morphism/Structures.agda
JacquesCarette: I think that big section headings are indeed more than adequate as separators - and we should update the style guide accordingly.

JacquesCarette:
Could this get another review @MatthewDaggitt ?

src/Function/Endomorphism/Propositional.agda
```diff
@@ -84,17 +84,41 @@ f ^ suc n = f ∘′ (f ^ n)
...
+  { isMagmaHomomorphism = ^-isMagmaHomomorphism f
+  ; ε-homo = refl
   }
+
+------------------------------------------------------------------------
+-- Deprecations
+
+-- Version 2.1
+
+^-isSemigroupMorphism = ^-isMagmaHomomorphism
```
MatthewDaggitt, line 114: So for this and the name below to be backwards compatible, we actually need the old names to be attached to the old definitions here, i.e. just copy the old definition wholesale.

src/Function/Endomorphism/Setoid.agda
```diff
@@ -6,90 +6,123 @@
...
-
+  open module E = Setoid (S ⇨ S) hiding (refl)
+  module S = Setoid S
+  open Func using (cong)
 ------------------------------------------------------------------------
 -- Basic type and functions
 
 Endo : Set _
-Endo = S ⟶ S
+Endo = S ⟶ₛ S
```
MatthewDaggitt, line 38: Likewise here, we'd need to come up with an entirely new name for this (and indeed all the proofs below). That isn't really feasible, so I guess the best approach if you want to persue this is to simply leave the old module in place, and add an entirely new module under a new name with the new definitions....

MatthewDaggitt (commented):
Sorry for abandoning this half way through the conversation!

MatthewDaggitt:
Hmm so in summary, I think the only reasonable way of doing this in a backwards compatible manner would be to create entirely new modules (I would tentatively suggest `Function.Endomorphism.(Setoid/Propositional).New`?) with the new definitions. Then in `v3.0` we'll delete the current modules, and then switch the new ones in.

src/Function/Endomorphism/Setoid.agda
jamesmckinna: Hmmm... that seems... unfortunate.

jamesmckinna:
Suggested alternative: Call the new module/modules `Endo.{Setoid|Propositional}`? (and avoid the awful `.New` subhierarchy?)

JacquesCarette:
Ok, this now should be backwards compatible! If you could re-review @MatthewDaggitt and @jamesmckinna ?

src/Function/Endo/Propositional.agda
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+------------------------------------------------------------------------
+-- N-th composition
+
+infixr 8 _^_
+
+_^_ : Endo → ℕ → Endo
+f ^ zero  = id
+f ^ suc n = f ∘′ (f ^ n)
```
jamesmckinna, line 51: As with the `TCO` versions in `Algebra.Definitions.RawMonoid`, should we consider such optimisations  here?

Indeed, should we in fact have `f ^ n` defined as in `Algebra.Properties.Monoid.Mult{.TCOptimised}`/`Algebra.Definitions.RawMonoid`, for the `RawMonoid` structure on `Endo`... and then the various homormor[hism properties etc. will follow from such definitions...?

src/Function/Endo/Setoid.agda
```diff
@@ -0,0 +1,128 @@
...
+{-# WARNING_ON_USAGE ^-isSemigroupMorphism
+"Warning: ^-isSemigroupMorphism was deprecated in v2.1.
+Please use ^-isMagmaHomomorphism instead."
+#-}
+
+^-isMonoidMorphism = ^-isMonoidHomomorphism
+{-# WARNING_ON_USAGE ^-isMonoidMorphism
+"Warning: ^-isMonoidMorphism was deprecated in v2.1.
+Please use ^-isMonoidHomomorphism instead."
+#-}
```
jamesmckinna, line 128: Given that these (old) definitions now belong to entirely deprecated modules, and this module is *new*, is there any longer any need for these explicit deprecations?

src/Function/Endo/Propositional.agda
```diff
@@ -0,0 +1,124 @@
...
+{-# WARNING_ON_USAGE ^-isSemigroupMorphism
+"Warning: ^-isSemigroupMorphism was deprecated in v2.1.
+Please use ^-isMagmaHomomorphism instead."
+#-}
+
+^-isMonoidMorphism = ^-isMonoidHomomorphism
+{-# WARNING_ON_USAGE ^-isMonoidMorphism
+"Warning: ^-isMonoidMorphism was deprecated in v2.1.
+Please use ^-isMonoidHomomorphism instead."
+#-}
```
jamesmckinna, line 124: Ditto. as with `Function.Endo.Setoid` below: now that we are in a *new* module, there's no need to explicitly export deprecated names which now exist only in deprecated modules, is there?

src/Function/Endo/Setoid.agda
```diff
@@ -0,0 +1,128 @@
...
+open import Function.Construct.Composition using () renaming (function to _∘_)
+open import Function.Relation.Binary.Setoid.Equality as Eq using (_⇨_)
+open import Level using (Level; _⊔_)
+open import Relation.Binary.Core using (_Preserves_⟶_)
+
+private
+  open module E = Setoid (S ⇨ S) hiding (refl)
+  module S = Setoid S
+  open Func using (cong)
+------------------------------------------------------------------------
```
jamesmckinna, line 34: `style-guide`: Double blank line after initial `private` block, I believe...

jamesmckinna (changes requested):
This all looks good, but have advised a number of refactorings, and removal of the (now-redundant!?) deprecations... and reinstatement of the #2277 `SuccessorSet` additions to `CHANGELOG`, which seem to have got deleted in the meantime...

jamesmckinna:
Also @MatthewDaggitt : can we now remove `breaking` label?

CHANGELOG.md
```diff
@@ -235,13 +245,12 @@ Additions to existing modules
...
   ```
 
-* In `Algebra.Morphism.Structures`
+* In `Algebra.Morphism.Structures`:
   ```agda
-  module SuccessorSetMorphisms (N₁ : RawSuccessorSet a ℓ₁) (N₂ : RawSuccessorSet b ℓ₂) where
-    record IsSuccessorSetHomomorphism (⟦_⟧ : N₁.Carrier → N₂.Carrier) : Set _
-    record IsSuccessorSetMonomorphism (⟦_⟧ : N₁.Carrier → N₂.Carrier) : Set _
-    record IsSuccessorSetIsomorphism  (⟦_⟧ : N₁.Carrier → N₂.Carrier) : Set _
-
```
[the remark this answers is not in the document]
JacquesCarette: I was afraid that would happen - I tried to do this via `git revert`, got into merge-hell over `CHANGELOG` and took in the simple-looking changes to this bit (because I couldn't believe that we wanted that complicated mess in the `CHANGELOG`...)  Will fix.

src/Function/Endo/Propositional.agda
JacquesCarette: scope creep... but probably.

src/Function/Endo/Setoid.agda
JacquesCarette: yes, definitely should have cleaned that up!

CHANGELOG.md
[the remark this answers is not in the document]
jamesmckinna: Well, sorry for it being a mess...;-) wasn't sure how much detail was required... but it did get merged the first time around, so I guess it had been approved...

src/Function/Endo/Propositional.agda
jamesmckinna: Well... if you feel strongly, we can make that a downstream refactoring PR, but I think it's (eventually) what we should do?

JacquesCarette: Yes, I think we should accept that this PR is enough of an improvement as is, and then make better still - later.

jamesmckinna: Agreed. Happy to take that on downstream!

JacquesCarette:
I think this now deals with all the comments from @jamesmckinna and @MatthewDaggitt and is no longer breaking either.

jamesmckinna (approved):

jamesmckinna:
Looks good to me now! 
The HTML index run has failed... but hopefully that's a transient bug...

JacquesCarette:
It's actually not - it is failing because of warnings. A deprecated module isn't allowed to use deprecated things lest warnings get issued. I had removed the warning suppression, sigh. Will have to restore it.

[This is probably something to file upstream - deprecated modules should be allowed to use other deprecated things, there is no need for a cascade of warnings!]

jamesmckinna:
> It's actually not - it is failing because of warnings. A deprecated module isn't allowed to use deprecated things lest warnings get issued. I had removed the warning suppression, sigh. Will have to restore it.
> 
> [This is probably something to file upstream - deprecated modules should be allowed to use other deprecated things, there is no need for a cascade of warnings!]

Two or three reactions to this:
* suppressing warnings in a deprecated module seems harmless, if annoying that it should prove necessary
* that it now does so is/seems to be a knock-on consequence of #2224 although the check on deprecated modules claims to ignore warnings, so is the list broken in some way?
* maybe a clue to have (yet) another go at overhauling our CI ...?

src/Function/Endo/Propositional.agda
```diff
@@ -0,0 +1,108 @@
...
+
+∘-id-monoid : Monoid _ _
+∘-id-monoid = record { isMonoid = ∘-id-isMonoid }
+
+private
+  ∘-rawMagma : RawMagma a a
+  ∘-rawMagma = Semigroup.rawMagma ∘-semigroup
+
+  ∘-id-rawMonoid : RawMonoid a a
+  ∘-id-rawMonoid = Monoid.rawMonoid ∘-id-monoid
```
jamesmckinna, line 93: Question in my mind as to whether these should be exported at top-level?
But certainly not a deal-breaker for this PR!

src/Function/Endo/Propositional.agda
```diff
@@ -0,0 +1,108 @@
...
+
+------------------------------------------------------------------------
+-- Homomorphism
+
+^-isMagmaHomomorphism : ∀ f → IsMagmaHomomorphism +-rawMagma ∘-rawMagma (f ^_)
+^-isMagmaHomomorphism f = record
+  { isRelHomomorphism = record { cong = cong (f ^_) }
+  ; homo = ^-homo f
+  }
```
jamesmckinna, line 102: Ditto.: should `isRelHomomorphism` be lifted out here, too?

src/Function/Endo/Propositional.agda
```diff
@@ -0,0 +1,108 @@
...
+f ^ suc n = f ∘′ (f ^ n)
+
+^-homo : ∀ f → Homomorphic₂ ℕ Endo _≡_ (f ^_) _+_ _∘′_
+^-homo f zero    n = refl
+^-homo f (suc m) n = cong (f ∘′_) (^-homo f m n)
+
+------------------------------------------------------------------------
+-- Structures
+
+∘-isMagma : IsMagma _≡_ (Op₂ Endo ∋ _∘′_)
```
jamesmckinna, line 61: Do you *really* intend `_≡_` as the equality on `Endo`, rather than `_≗_`?

jamesmckinna:
Re: #2394 See [this branch](https://github.com/jamesmckinna/agda-stdlib/tree/Endomorphism-bis/src/Function/Endo) for the details: easily incorporated on top of what you have here!

src/Function/Endo/Propositional.agda
jamesmckinna: I realise that this is a legacy design decision, but given the green field reimplementation in this PR, a chance to reconsider?

JacquesCarette: Agree, this is probably a mistake.

src/Function/Endo/Propositional.agda
JacquesCarette: I guess since we're now moving to using `using` a lot, having extra names around isn't so costly, so I can make these visible beyond the bundle.

JacquesCarette:
I think the names are fine as they are (for the purposes of porting this). Made minor tweaks, but I think once the build comes back green, it can be merged.

PR #2343, opened 2024-04-04
Tighten imports some more

JacquesCarette:
A few files used very basic bits of `Data.Product` and `Function` that could instead use `.Base`. 

While in those files, fix up other things. Some files couldn't go to `.Base` after all, but other things could be fixed, so they were.

In one case, `Effect.Monad.Partiality` was using all of `Data.Nat` for a private example -- I moved that to `README.Effect.Monad.Partiality` instead.

### DIFF
--- a/doc/README/Effect/Monad/Partiality.agda
+++ b/doc/README/Effect/Monad/Partiality.agda
@@ -1,0 +1,27 @@
+-----------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Example showing the use of the partiality Monad
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe --guardedness #-}
+
+module README.Effect.Monad.Partiality where
+
+open import Codata.Musical.Notation using (♯_)
+open import Data.Bool.Base using (false; true)
+open import Data.Nat using (ℕ; _+_; _∸_; _≤?_)
+open import Effect.Monad.Partiality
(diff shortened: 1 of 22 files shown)


doc/README/Effect/Monad/Partiality.agda
```diff
@@ -0,0 +1,27 @@
+-----------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Example showing the use of the partiality Monad
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe --guardedness #-}
+
+module README.Effect.Monad.Partiality where
```
gallais, line 9: Should we move this to `Data.Nat.McCarthy` instead? We could prove the equivalence to [the tail recursive version](https://en.wikipedia.org/wiki/McCarthy_91_function#History) for instance.

gallais (approved):

doc/README/Effect/Monad/Partiality.agda
JacquesCarette: That would work too - I have no strong opinion either way.

MatthewDaggitt: I guess I'd vote against. I don't think the f91 is really a piece of functionality that is going to be useful to anyone. Indeed if they are proving properties about it, it'll almost certainly for pedagogical purposes in which case they'll want to redefine it. Therefore the README seems like a good place for it.

PR #2347, opened 2024-04-08
Tighten imports, last big versoin

JacquesCarette:
For quite a few `Data.X` imports, `Data.X.Base` was enough -- and this tends to help quite a bit with the dependencies, as `Data.X` often imports `Data.X.Properties`, which usually imports a *lot* of stuff. (Cutting that down is a 3.0 job).

### DIFF
--- a/src/Data/Fin/Substitution/Lemmas.agda
+++ b/src/Data/Fin/Substitution/Lemmas.agda
@@ -9,11 +9,12 @@
 module Data.Fin.Substitution.Lemmas where
 
 open import Data.Fin.Substitution
-open import Data.Nat hiding (_⊔_; _/_)
+open import Data.Nat.Base using (ℕ; _+_; zero; suc)
 open import Data.Fin.Base using (Fin; zero; suc; lift)
-open import Data.Vec.Base
+open import Data.Vec.Base using (lookup; []; _∷_; map)
 import Data.Vec.Properties as Vec
 open import Function.Base as Fun using (_∘_; _$_; flip)
+open import Level using (Level; _⊔_)
 open import Relation.Binary.PropositionalEquality.Core as ≡
   using (_≡_; refl; sym; cong; cong₂)
 open import Relation.Binary.PropositionalEquality.Properties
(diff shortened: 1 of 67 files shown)


src/Data/Fin/Substitution/Lemmas.agda
```diff
@@ -9,19 +9,19 @@
 module Data.Fin.Substitution.Lemmas where
 
 open import Data.Fin.Substitution
-open import Data.Nat hiding (_⊔_; _/_)
+open import Data.Nat.Base using (ℕ; _+_; zero; suc)
```
jamesmckinna, line 12: Again, `zero`, `suc` before `_+_`?

src/Data/Integer.agda
```diff
@@ -31,13 +31,13 @@ open import Data.Integer.Properties public
...
 import Data.Nat.Show as ℕ using (show)
-open import Data.Sign as Sign using (Sign)
+open import Data.Sign.Base as Sign using (Sign)
 open import Data.String.Base using (String; _++_)
 
 show : ℤ → String
 show i = showSign (sign i) ++ ℕ.show ∣ i ∣
   where
-  showSign : Sign → String
+  showSign : Sign.Sign → String
```
jamesmckinna, line 40: Why the qualified name here?

src/Data/Integer/DivMod.agda
```diff
@@ -13,7 +13,7 @@ open import Data.Integer.Properties
 open import Data.Nat.Base as ℕ using (ℕ; z≤n; s≤s; z<s; s<s)
 import Data.Nat.Properties as ℕ
 import Data.Nat.DivMod as ℕ
-import Data.Sign as S
+import Data.Sign.Base as S
```
jamesmckinna, line 16: Should be `as Sign`? #2280

src/Data/Fin/Substitution/Lemmas.agda
JacquesCarette: Fixed

src/Data/Integer.agda
JacquesCarette: No good reason (probably a result of a failed experiment that didn't get fully backed out). Fixed.

src/Data/Integer/DivMod.agda
JacquesCarette: Even better: delete altogether as it's not used!

jamesmckinna (approved):
Heroic work!

Taneb (approved):
This all looks pretty good!

gallais (approved):

Issue #2348, opened 2024-04-08
Add `Algebra.Construct.MonoidAction` and `Algebra.Construct.WreathProduct`


jamesmckinna:
The first is clear, although I wonder a bit about how best we should add `Action`s to the `Algebra` hierarchy, so going via `Construct` seems the best... for now (as opposed to a parallel hierarchy of 'things-acted-upon-by-things').

The second follows on, but is complicated by the plethora of various definitions in the literature (according to the 'thinginess' involved), and the relationship with 'semi-direct product's... so perhaps some discussion/downstream refactoring may be necessary. 

PR incoming in the next few days, I think.

JacquesCarette:
I definitely think we're going to need either `Algebra.Action` or a different hierarchy altogether for multisorted Algebra. I rather think we should strongly embrace multi-sorted, rather than trying to fit it in an ad hoc way into our singly-sorted current version.

It could be argued that `Algebra.*` should be the home for multi-sorted. It currently isn't - but that could be a reasonable evolution.

jamesmckinna:
Hmmm... I think I agree, but I'm weighing up v2.1 vs v3.0... and would like something *now*... with a multisorted refactoring downstream?

Taneb:
What do you mean by "multisorted" here?

jamesmckinna:
Well, as with `Module` (two-sorted: the `Ring` algebra acts on the `AbelianGroup`...), here there are two underlying carriers ('sorts' in the language of universal algebra, with sorts mapping to these distinct carriers in the construction of interpretations/models; but in general possible many, so 'multi-'), equality relations and operations going on: that of the thing acting (eg `Monoid`) and that of the thing acted upon (eg `Setoid`), with the action map relating the two...

jamesmckinna:
Refactoring this issue to break up the two parts. What I thought would be straightforward about Wreath products seems to require more infrastructure than I am able to find already existing at the moment... esp. wrt 'pointwise' function spaces between `Setoid`s and other kinds of bundle...

jamesmckinna:
Meanwhile, hope you're happy @JacquesCarette with my continuing with #2350 under `Algebra.Construct` ahead of any decision about where such stuff should eventually end up...?

JacquesCarette:
I certainly don't mind if a Draft PR, to shake things out, is built "in the wrong place" for parts of its contents.

jamesmckinna:
[...]

PR #2350, opened 2024-04-09
Add `Algebra.Construct.MonoidAction`

jamesmckinna:
Fixes the first part of #2348 . Currently in DRAFT for preliminary feedback.

Issues/TODOs:
- [ ] `CHANGELOG`
- [ ] uses the method of indexing a thing over a raw thing cf. #2252 ... but this seems debatable here?
- [ ] repetition between the `_⋆_` structure definable in `Raw*Action` and corresponding lifts in `*Action` and `Free`
- [ ] a `Raw*Action` might better be described as a `Setoid` homomorphism between `M.setoid × A` and `A`, only I can't seem to find a definition of the former (under `Relation.Binary.Construct.Product`?) as a product of `Setoid`s?
- [ ] the equivalence of a `*Action` and the corresponding `Monoid` homomorphism (cf. #1960 ) between `M` and `Endo A` (cf. #2342 )


jamesmckinna:
Note to self: 'tipped' lists and the adjunction preserving `List` as an initial algebra.

JacquesCarette:
What's a 'tipped' list?

jamesmckinna:
> What's a 'tipped' list?

A datatype I first learnt if from Conor (possibly invented by him): `List A`, except that at `nil` you. An store a value of some type `B` (and various games one might play as to whether that type us existentially/universally quantified, or merely supplied as an additional parameter). 

Ralf Hinze pointed out to me last week that one may easily show an isomorphism between that and `(List  A) * B` because `_ * B`, having a right adjoint, preserves colimits/initial algebras...

... and there may be some relationship with the Minamide construction (1997) in connection with tail recursion modulo constructors...

in any case, such a tipped list corresponds to a symbolic/'non-materialised' `fold`, and hence to the iterated monoid action (which materialises it!), generalising difference lists etc.

JacquesCarette:
Aha, a 2-sorted pointed unar! Nice. 
```agda
record 2PU (A B : Set) : Set where
  field
    push : (a : A) -> 2PU -> 2PU
    point : (b : B) -> 2PU
```

JacquesCarette:
Looked at the code itself right now: I expect that in `Algebra.Construct.MonoidAction`, all I will find are **constructions** of monoid actions. I did not expect to find the definition of the representation of an action to be in the same file. This belongs elsewhere.

I'm happy for the construction (at the end of the file) to stay where they are, but not so for the definition of the concepts themselves.

JacquesCarette:
`Algebra.Action.Monoid` would not be bad at all. To be more forward looking, maybe `Algebra.Action.Bundles` or something like it.

One could go crazy and do `Algebra.MultiSorted.Action.Bundles` but I think that's overkill, even if "not wrong".

jamesmckinna:
Re: your `2PU`... I was more thinking 
```agda
data TipList (A B : Set) : Set where
  [_] : A ->TipList A B
  _::_ : B -> TipList A B -> TipList A B
```
the `record` version is... the 'wrong' way round?

JacquesCarette:
Sorry, I have "free theories" on the brain. Your `TipList` is the object part of the left adjoint to the forgetful functor from `2PU` to `Set x Set`. (Urg, I hate how that sounds - I'm not a big fan of lathering on categorical lingo unless it actually enlightens.)

jamesmckinna:
> Sorry, I have "free theories" on the brain. Your `TipList` is the object part of the left adjoint to the forgetful functor from `2PU` to `Set x Set`. (Urg, I hate how that sounds - I'm not a big fan of lathering on categorical lingo unless it actually enlightens.)

Thanks for the clarification! I think I had at least glimpsed that from my reference to direction... but I hadn't joined all the dots. And so, no need to apologise, the category theory is fine!

MatthewDaggitt:
Yes, I agree with @JacquesCarette. `X.Construct` modules are for building records found in `X.Structures/Bundles`. 

I agree that splitting it into `Algebra.Action.Bundles` and `Algebra.Action.RawBundles` makes sense to me!

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
...
+  private
+
+    open module M = Monoid M using () renaming (Carrier to M)
+    open module A = Setoid A using (_≈_) renaming (Carrier to A)
+    open ≋ M.setoid using (_≋_; [] ; _∷_)
+
+    variable
+      x y z : A.Carrier
+      m n p q : M.Carrier
+      ms ns ps qs : List M.Carrier
```
MatthewDaggitt, line 114: Hmm this is the first attempt to have `variable` declarations mid-file. My opinion is that we shouldn't go here as it makes it much harder to parse a file.

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
...
+  rightAction : RightAction M setoid rawRightAction
+  rightAction = record
+    { ∙-act = λ x m n → sym (assoc x m n)
+    ; ε-act = identityʳ
+    }
+
+------------------------------------------------------------------------
+-- Distinguished *free* action: lifting a raw action to a List action
+
+module Free (M : Setoid c ℓ) (S : Setoid a r) where
```
MatthewDaggitt, line 189: This should live in `Algebra.Action.Construct.Free`?

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
...
+    ⋆-act : ∀ x ms ns → x ⋆ (ms ++ ns) ≈ x ⋆ ms ⋆ ns
+    ⋆-act _ = ⋆-act-cong M.refl A.refl
+
+    []-act : ∀ x → x ⋆ [] ≈ x
+    []-act x = []-act-cong A.refl
+
+------------------------------------------------------------------------
+-- Distinguished actions: of a module over itself
+
+module Regular (M : Monoid c ℓ) where
```
MatthewDaggitt, line 163: This should live in `Algebra.Action.Construct.Self`?

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
...
+    ⋆-act-cong refl x≈y (m ∷ ms) ns = ⋆-act-cong refl (∙-cong x≈y refl) ms ns
+
+    []-act-cong : x ≈ y → (x ᴬ⋆ᴹ []) ≈ y
+    []-act-cong = id
+
+
+------------------------------------------------------------------------
+-- Definition: indexed over an underlying raw action
+
+module _ (M : Monoid c ℓ) (A : Setoid a r) where
```
MatthewDaggitt, line 103: Should live in `Algebra.Action.Bundles`?

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
...
+    a c r ℓ : Level
+    A : Set a
+    M : Set c
+
+
+------------------------------------------------------------------------
+-- Basic definitions: fix notation for underlying 'raw' actions
+
+module _ {c a ℓ r} {M : Set c} {A : Set a}
+         (_≈ᴹ_ : Rel M ℓ) (_≈_ : Rel A r)
```
MatthewDaggitt, line 42: Should live in `Algebra.Action.Bundles.Raw`?

jamesmckinna:
Lots of helpful feedback @MatthewDaggitt thanks!

I'm undecided about `Algebra.Action.Bundles` vs `Algebra.Action.Monoid`, preferring the latter until we have more examples of the phenomenon?

As for `Raw`, I'm wondering if it's easier to start from a `Func (M.setoid ×ₛ A) A`, packaging the congruence data, *and* not introducing a new type for the sake of it...

src/Algebra/Construct/MonoidAction.agda
jamesmckinna: I'd like to push back on this, but if it's really a first use, perhaps I'm on shakier ground, but the pragmatic/ux gain seemed clear enough to me as a developer...

... but perhaps not (to you) as a reader/consumer of the module? 

Hmmm...

MatthewDaggitt:
> I'm undecided about Algebra.Action.Bundles vs Algebra.Action.Monoid, preferring the latter until we have more examples of the phenomenon?

Can you have actions on `Semigroup` or `Ring` or other structures?

src/Algebra/Construct/MonoidAction.agda
jamesmckinna: But it's possible that the suggested breakup of this into much smaller modules might in fact lift such declarations to the top-level...?

MatthewDaggitt: So the more non-local information you have to hold in your head at one time, the harder it is to understand a definition. When you have a collapse, and you have to go back and check everything to refresh it in your head, I find it much harder to build up the mental model again when I have to find and integrate non-local information in more places...

I guess it's exactly the same argument I have about having deeply nested modules within a file...

MatthewDaggitt: > But it's possible that the suggested breakup of this into much smaller modules might in fact lift such declarations to the top-level...?

In which case, fantastic!

jamesmckinna:
The classic example is "module" which is a `Ring` acting on an `AbelianGroup`... etc.

src/Algebra/Construct/MonoidAction.agda
jamesmckinna: Couldn't quite get it to work out :-(

jamesmckinna:
Temporarily defeated by the complexity of the refactoring... sigh. But nearly there!

jamesmckinna:
Currently the `IsRaw*` structures aren't really *raw*, but vanilla `Setoid` structures (with `cong`)... so further refactoring might be in order to tease those apart. But let's turn this over to review in the meantime?

src/Algebra/Construct/MonoidAction.agda
```diff
@@ -0,0 +1,227 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Monoid Actions and the free Monoid Action on a Setoid
```
jamesmckinna, line 4: This module should be removed as part of the refactoring!

src/Algebra/Action/Construct/Self.agda
```diff
@@ -0,0 +1,54 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The left- and right- regular actions: of a Monoid over itself
```
JacquesCarette, line 4: Some comment here as for 'Free'

src/Algebra/Action/Structures/Raw.agda
```diff
@@ -0,0 +1,77 @@
...
+  variable
+    x y z : A
+    m n p : M
+    ms ns ps : List M
+
+------------------------------------------------------------------------
+-- Basic definitions: fix notation
+
+record IsRawLeftAction : Set (a ⊔ r ⊔ c ⊔ ℓ) where
+  infixr 5 _ᴹ∙ᴬ_ _ᴹ⋆ᴬ_
```
JacquesCarette, line 32: I can see the point of these pre- and post- superscripts, but as you yourself rename these away almost everywhere, maybe that indicates that they're not great names after all?

src/Algebra/Action/Structures/Raw.agda
```diff
@@ -0,0 +1,77 @@
...
+
+  field
+    _ᴹ∙ᴬ_  : M → A → A
+    ∙-cong : m ≈ᴹ n → x ≈ y → (m ᴹ∙ᴬ x) ≈ (n ᴹ∙ᴬ y)
+
+-- derived form: iterated action, satisfying congruence
+
+  _ᴹ⋆ᴬ_ : List M → A → A
```
JacquesCarette, line 40: Hmm, there's also an action of non-empty lists: there is no need for a unit to exist for this to make sense.

JacquesCarette (commented):
The raft of comments should not be misleading: this seems nice. There's just a whole lot of design points that need discussed.

src/Algebra/Action/Construct/Free.agda
```diff
@@ -0,0 +1,83 @@
...
+
+open import Function.Base using (_∘_)
+
+open import Level using (Level; _⊔_)
+
+open ≋ M using (_≋_; ≋-refl; ≋-reflexive; ≋-isEquivalence; ++⁺)
+open MonoidAction
+
+------------------------------------------------------------------------
+-- First: define the Monoid structure on List M.Carrier
```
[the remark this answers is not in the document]
jamesmckinna: For *`Propositional`* equality, sure. But for `Data.List.Relation.Binary.Equality` (for `Setoid`s)? Nope!

src/Algebra/Action/Construct/Free.agda
```diff
@@ -0,0 +1,83 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The free MonoidAction on a SetoidAction
```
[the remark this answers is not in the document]
jamesmckinna: I see. Throughout I've been torn about focusing *entirely* on the `Monoid` case against the need for subsequent generalisation/refactoring later, and trying to leave enough 'hooks' in a general, but sparsely populated setup. The result seems an unhappy mixture of the two. What do you advise?

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,141 @@
...
+      ∙-act  : ∀ m n x → m M.∙ n ∙ x A.≈ m ∙ n ∙ x
+      ε-act  : ∀ x → M.ε ∙ x A.≈ x
+
+    ∙-congˡ : ∀ {m x y} → x A.≈ y → m ∙ x A.≈ m ∙ y
+    ∙-congˡ x≈y = ∙-cong M.refl x≈y
+
+    ∙-congʳ : ∀ {m n x} → m M.≈ n → m ∙ x A.≈ n ∙ x
+    ∙-congʳ m≈n = ∙-cong m≈n A.refl
+
+    ⋆-act : ∀ ms ns x → (ms ++ ns) ⋆ x A.≈ ms ⋆ ns ⋆ x
```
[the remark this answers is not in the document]
jamesmckinna: It's \* and \. and the GitHub fonts render each pretty badly compared to the emacs window... a frequent complaint on previous PRs...!?

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,141 @@
...
+  record Left (leftAction : SetoidAction.Left M.setoid A) : Set (a ⊔ r ⊔ c ⊔ ℓ)
+    where
+
+    open SetoidAction.Left leftAction public
+      using (isRawLeftAction)
+    open IsRawLeftAction isRawLeftAction public
+      renaming (_ᴹ∙ᴬ_ to _∙_; _ᴹ⋆ᴬ_ to _⋆_)
+
+    field
+      ∙-act  : ∀ m n x → m M.∙ n ∙ x A.≈ m ∙ n ∙ x
```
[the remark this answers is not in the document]
jamesmckinna: Ah, OK, I was thinking that they were 'noise', and had tuned the infix decls. to ensure I didn't need them, (the `M.` qualification is the clue we're switching type...). Sigh.

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,141 @@
...
+------------------------------------------------------------------------
+-- A Setoid action yields an iterated List action
+
+module _ {M : Setoid c ℓ} {A : Setoid a r} where
+
+  open SetoidAction
+
+  open ≋ M using (≋-setoid)
+
+  leftListAction : (leftAction : Left M A) → Left ≋-setoid A
```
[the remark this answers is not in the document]
jamesmckinna: 1. The `List` is the *carrier* of the `≋-setoid` introduced at L66
2. Initially I had had `LeftAction` etc., but now that they fall under module names `SetoidAction` and `MonoidAction` (because `Setoid` and `Monoid` are already in scope... ;-)), I couldn't stand the additional redundancy involved in writing `Action` twice... unqualified that would lead to
```agda
Algebra.Action.Bundles.MonoidAction.LeftAction
```
in which the cue word '`Action`' appears 3 times!

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,141 @@
...
+-- Basic definition: a Setoid action yields underlying raw action
+
+module SetoidAction (M : Setoid c ℓ) (A : Setoid a r) where
+
+  private
+
+    open module M = Setoid M using ()
+    open module A = Setoid A using (_≈_)
+
+  record Left : Set (a ⊔ r ⊔ c ⊔ ℓ) where
```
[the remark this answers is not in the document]
jamesmckinna: Yes, see below, and I need `SetoidAction.Left` qualified to use in `MonoidAction.Left` etc.

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,141 @@
...
+  rightListAction : (rightAction : Right M A) → Right ≋-setoid A
+  rightListAction rightAction = record
+    { act = record
+      { to = uncurry _ᴬ⋆ᴹ_ ; cong = uncurry ⋆-cong }
+    }
+    where open Right rightAction; open IsRawRightAction isRawRightAction
+
+
+------------------------------------------------------------------------
+-- Definition: indexed over an underlying raw action
```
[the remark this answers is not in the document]
jamesmckinna: Dithering about the design! I think I can't face refactoring to make the `Raw` things 'properly' raw, so I think I'll remove in favour of something simpler as `Algebra.Action.Structures`.

src/Algebra/Action/Construct/Free.agda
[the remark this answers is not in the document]
jamesmckinna: I would take the hint that it should be added to the library at some point somewhere else, but I'd be much happier to make that a downstream refactoring...

src/Algebra/Action/Structures/Raw.agda
jamesmckinna: Could be: an aspect of the evolution of the design I suppose, but at some point I had wanted a non-symmetric symbol to indicate the 'handedness' of the action, even if subsequently it is indeed a pain to write it. But using bare `_∙_` seemed to invite a horrible notational clash with that symbol coming  from `Monoid`... cf [your comment](https://github.com/agda/agda-stdlib/pull/2350/files#r1561953300) above on `Bundles`...

src/Algebra/Action/Structures/Raw.agda
jamesmckinna: A *very* good point, but I had *definitely* run out of steam for considering `Magma`/`Semigroup` actions as well, ... and their `Free` instances arising from variations on `Data.List`... but the `NonEmpty` case is one previously mentioned by @Taneb under #1281 / #2300 

But here, I think I'm strongly tempted to note this as a downstream `addition` request/suggestion, on the (oftentimes: your!) usual lines of "don't make the perfect the enemy of the good"... ;-)

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: I suppose I could eschew the renaming, and then the 'bracketing' might be clearer?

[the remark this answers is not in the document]
jamesmckinna: I ended up making different choices. Maybe not to your taste, either...!?

src/Algebra/Action/Construct/Free.agda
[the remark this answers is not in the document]
jamesmckinna: Refactored now to make `FreeMonoidAction` a sub-module. Not convinced it's better, though!

jamesmckinna:
Thanks very much to @JacquesCarette for the review comments.I've updated to partly 'streamline', and partly to leave hooks for downstream generalisation. Not convinced it's an overall improvement though!

src/Algebra/Action/Structures/Raw.agda
jamesmckinna: Plus... irritatingly, given the factorisation of `Data.List.NonEmpty` exhibiting ` List⁺` as a (proxy-) pair of an element and a `List`, then to define the `NonEmpty`/`Semigroup` actions, we need the `List`/`Monoid` ones! So... I decided that this should *definitely* be a downstream addition!

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: And reworked the comment...

jamesmckinna:
Given the trouble I've had thinking about this, no longer sure this is `low-hanging-fruit`...

src/Algebra/Action/Construct/Free.agda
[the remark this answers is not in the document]
JacquesCarette: Your 'prelude's need a TL;DR !  Ok, in your case, I'll pay more attention to that in the future.

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
JacquesCarette: A repeated comment (in the code) to that effect would be good.

src/Algebra/Action/Construct/Free.agda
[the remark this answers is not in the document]
jamesmckinna: Ahem... sorry! I'll also try to be more explicit about intentions...

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: I've left it currently as an (unrepeated) opening comment in `...Bundles`.

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: See recent commit on naming.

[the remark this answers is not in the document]
JacquesCarette: I think what confused me was that I could remember 'trivially' that `List` was the carrier of `≋-setoid`. I'm thinking that if a particular setoid has a known, particular carrier, maybe that too should be part of its name?

JacquesCarette (commented):
More comments

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: Conceivably, and from a ux perspective, I perhaps even agree with you. But are you suggesting I open the instantiated module *renaming* `≋-setoid` to include `List` somewhere in its name? That's a can of worms to want to open up!?

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: (Some of) My colleagues in WG2.1 would certainly agree with you as regards 'asymmetric symbols for asymmetric concepts', and even I do, to an extent. But *mathematically*, I find it compelling to have the pun between the action map and the monoid map, despite railing elsewhere against puns as being harmful. My 'ugly' superscripting is a way of introducing asymmetry, but then overturning it in the service of 'ordinary practice'.

As to the RHS with the 'iterated' action being given a named definition, is that (not)... overkill?

I think these naming/notation questions are tricky, and to some extent governed by convention (`List.cons` being the worst offender?), and for sure I think it's a bigger discussion than concerning this PR. Happy to put it out to (wider) discussion, but atm I'm reluctant to change... neither do I want this to be a hill on which to die!

Sigh.

src/Algebra/Action/Structures/Raw.agda
jamesmckinna: So: have I left enough room for such additions, or not? 
I think that I have, but your (pertinent!) comment makes me wonder...

CHANGELOG.md
```diff
@@ -57,6 +57,26 @@ Deprecated names
...
+  ```
+
+* The free `List` actions over a `SetoidAction`:
+  ```
+  Algebra.Action.Construct.Free
+  ```
+
+* The left- and right- regular actions (of a `Monoid`) over itself:
+  ```
+  Algebra.Action.Construct.Self
```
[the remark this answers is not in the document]
jamesmckinna: Well, @MatthewDaggitt suggested `Self`, and while I might have thought `Regular` might have been better, here I ended up thinking that a 'neutral' name, *not* informed by existing mathematical practice/conventional usage , might be preferable. It's clear that throughout this PR my instinct were pulling me in multiple directions... not always for the better!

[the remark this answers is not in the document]
JacquesCarette: I guess I didn't put 2 and 2 together because `Construct` appears in between `Action` and `Self` and I didn't get "self action"! I can get behind this name, now that it has been explained.

CHANGELOG.md
```diff
@@ -57,6 +57,26 @@ Deprecated names
...
 -----------
 
+* Bundles for left- and right- actions:
+  ```
+  Algebra.Action.Bundles
+  ```
+
+* The free `List` actions over a `SetoidAction`:
+  ```
+  Algebra.Action.Construct.Free
```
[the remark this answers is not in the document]
JacquesCarette: Unlike "self action" which does, post facto, make sense, "free action" still doesn't, as it leaves too much implicit.

Now if this was a module that gathered together a bunch of different "free actions" (of what on what?), that could make sense.

src/Algebra/Action/Structures/Raw.agda
JacquesCarette: So I think overall, it does allow enough room - but the naming doesn't.

jamesmckinna: Not sure I understand: the `NonEmpty` iterated action would have a separate name, surely?

JacquesCarette: Of course - but since `_ᴹ⋆ᴬ_` is taken, what kind of monster would be needed for that?

jamesmckinna: I'd been thinking something like `_ᴹ⁺ᴬ_`, with the '⁺' signifiying `NonEmpty`, and the `⋆` signifying `List`...

jamesmckinna: But: review comments are a good place to recommend 'better' names to a PR author ;-)

CHANGELOG.md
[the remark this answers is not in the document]
jamesmckinna: `Lift` rather than `Free`: because we 'lift' a `SetoidAction` to a `MonoidAction`?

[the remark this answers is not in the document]
JacquesCarette: It does seem a bit better - but what if there's `Lift`s to `GroupAction` too? What will that be named?

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
JacquesCarette: If the naming of `≋-setoid` is external to this PR, then that would be feature creep beyond what it reasonable to ask for. Trying to fix it locally for readability is not worth doing (sad as that may be.)

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
JacquesCarette: By 'named definition' I meant like `Commutative` and `Associative` -- maybe that wasn't clear? My feeling is that if we looked, we'd find uses of something equivalent to this action structure already in the library!

Mathematical puns are cute 'in the small' but my experience is that they don't scale well. Actions are spreading in importance (I think people are starting to spot them and name them more), so it feels like it is worth taking a little extra time to design this.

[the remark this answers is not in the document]
jamesmckinna: Yes, sorry it hadn't been clear to me.

And given the importance of the concept (I agree: it's not just me ;-)), I think the overall effect of your comments is to nudge me towards putting this back into DRAFT until I can get a clearer picture of how best to proceed forwards. A  *lot* of food for thought (the perils of being the first to contribute a 'significant' addition, one which opens a new front in the hierarchy...?)

jamesmckinna:
Recent commits:
* fix notations to be asymmetric in `Structures`
* simplify definitions of `Bundles` (so: a `Biased` version might be useful for deriving `Action`s from `Func`s of suitable type)
* rename `Free` to `FreeMonoid`, but now the module structure needs simplification

TODO (?):
* define the `Action` property in `Algebra.Definitions`... multi- vs. single-sorted distinction again...? ;-)

CHANGELOG.md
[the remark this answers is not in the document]
jamesmckinna: Now renamed `FreeMonoidAction`...

jamesmckinna:
@JacquesCarette back now ready for review, but I dare say there'll be more to discuss!

src/Algebra/Action/Structures/Raw.agda
jamesmckinna: See the latest commits for reorganisation and new (better?) notation...

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
jamesmckinna: > By 'named definition' I meant like `Commutative` and `Associative` -- maybe that wasn't clear? My feeling is that if we looked, we'd find uses of something equivalent to this action structure already in the library!

To be fair, I haven't looked in the library, but I am worried about where such a `Definition` should go... because `Algebra.Definitions` is single-sorted... and for the sake of the one definition, is it worth abstracting out? And where do you think it should go?

jamesmckinna:
I seem to be going round in circles redefining things repeatedly... DRY! Trying to fix this ...

jamesmckinna:
> @JacquesCarette back now ready for review, but I dare say there'll be more to discuss!

OK, I mean it this time...;-)

Highlights of the most recent refactoring:
* sort out the dependencies to reduce duplication/get things 'right'
* naming: (iterated) action properties now take their names from the `Monoid`, not from the underlying action map

Possible further work:
* move `ListAction` from `Bundles` to `Construct.FreeMonoid`, because they are they are the `Setoid` analogues of the free `Monoid` iterated actions... and perhaps don't need to be 'free-standing' in `Bundles`...?

src/Algebra/Action/Construct/FreeMonoid.agda
```diff
@@ -0,0 +1,82 @@
...
+open import Data.Product.Base using (_,_)
+
+open import Function.Base using (_∘_)
+
+open import Level using (Level; _⊔_)
+
+
+------------------------------------------------------------------------
+-- First: define the Monoid structure on List M.Carrier
+-- NB should be defined somewhere under `Data.List`!?
```
JacquesCarette, line 32: Data.List.Action.Construct.FreeMonoid ?

src/Algebra/Action/Construct/FreeMonoid.agda
```diff
@@ -0,0 +1,82 @@
...
+
+  leftAction : Left listAction
+  leftAction = record
+    { ∙-act = ++-act
+    ; ε-act = []-act
+    }
+
+module _ (right : SetoidAction.Right M S) where
+
+  private listAction = ListAction.rightAction right
```
JacquesCarette, line 73: Does it really make sense to create an abbreviation that's used only once? (same above)

JacquesCarette (changes requested):
One super-minor thing (and a suggestion to a question left in the code).

src/Algebra/Action/Bundles.agda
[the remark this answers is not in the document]
JacquesCarette: On the last question: `Algebra.Action.Definitions`.

src/Algebra/Action/Construct/FreeMonoid.agda
jamesmckinna: Interesting suggestion!
But NB. as with #2360 suggest that this be dealt with as a separate issue/PR. 
The code here is self-contained, so can be lifted out as a downstream refactoring?

src/Algebra/Action/Construct/FreeMonoid.agda
jamesmckinna: After my recent refactoring effort, you're absolutely right: the answer is no. 
That said, sometimes we make `private` local definitions which turn out to be useful later... and I'd been thinking of moving the code for `ListAction` here wholesale, but I guess that could happen later. Meanwhile, I'll inline.

src/Algebra/Action/Construct/FreeMonoid.agda
JacquesCarette: Agreed, that can be a separate PR.

JacquesCarette (approved):

jamesmckinna:
Thanks very much @JacquesCarette for the very detailed feedback!
I think I'd be happy to leave to a subsequent refactoring the lifting out of the `-act` properties to a separate  `Algebra.Action.Definitions` module, but I can start to see how that might go... and how that might help keep things straight when passing from one action to its iterated form...

jamesmckinna:
Coda: @JacquesCarette did your allusion to `NonEmpty` lists mean that you also might want:
```agda
module SemigroupAction (M : Semigroup c ℓ) (A : Setoid a r) where

  private

    open module M = Semigroup M using (ε; _∙_; setoid)
    open module A = Setoid A using (_≈_)

  record Left (leftAction : SetoidAction.Left setoid A) : Set (a ⊔ r ⊔ c ⊔ ℓ)
    where

    open SetoidAction.Left leftAction public

    field
      ∙-act  : ∀ m n x → m ∙ n ▷ x ≈ m ▷ n ▷ x
```
(and: `MagmaAction` even!?) with `MonoidAction` being refactored to have a `SemigroupAction` subfield...?

jamesmckinna:
I realised writing the last comment that my fixation on `Monoid` actions blinded me to the simpler structures, and with it, [Krohn-Rhodes decomposition](https://en.wikipedia.org/wiki/Krohn%E2%80%93Rhodes_theory).

Separately, by weakening things even further down to `MagmaAction`, then we would get to observe that the additional generality afforded by actions, namely that associativity is not even an optional thing for these structures, because the `∙-act` axiom essentially *forces* a version of associativity of the action map, even when/if the `_∙_` structure itself is *not* associative...

src/Algebra/Action/Bundles.agda
```diff
@@ -0,0 +1,140 @@
...
+    ++-act x ms ns = ++-act-cong A.refl ms ≋-refl
+
+    []-act : ∀ x → x ◁⋆ [] A.≈ x
+    []-act x = []-act-cong A.refl
+
+
+------------------------------------------------------------------------
+-- A Setoid action yields an iterated List action
+
+module ListAction {S : Setoid c ℓ} {A : Setoid a r} where
```
Taneb, line 89: I feel like this should be in its own module, somewhere under `Algebra.Action.Construct`

jamesmckinna: I think I'm inclined to agree. I'd cut corners trying to cram too much into/under `SetoidAction`, and I think this is a fossil residue of that which deserves its own place as a `SetoidAction` under `Construct.FreeMonoid`, and echoing earlier comments from @JacquesCarette perhaps that module needs renaming?

jamesmckinna: Have efactored to move these definitions around, and streamline/simplify the module structure in `Construct.FreeMonoid`.

jamesmckinna:
I think it now looks a lot more tight, but please feel free to re-review @JacquesCarette and @Taneb : I've tried to honour the spirit of your most recent round of comments, so many thanks for the nudge(s) to do so.

Only thing left outstanding would be to factor out the `Algebra.Action.Definitions`, but that can happen later? Happy to do it now, for completeness's sake, if you think it would improve the overall contribution.

jamesmckinna:
Converting back to `DRAFT` while I figure out:
- [x] `Algebra.Action.Definitions`
- [x] `MagmaAction`
- [ ] `SemigroupAction`
- [ ] consequences of the above for `Construct.Self` (OK for `Semigroup`, by associativity; not so for `Magma`!) and `Construct.FreeMonoid` (might want to revert to `Free` to include `Semigroup` case as well?)

JacquesCarette (approved):
What is here is great. I'll review again when more stuff gets added.

JacquesCarette:
I should reply to some specific comments: yes, definitely also including `SemigroupAction` and `MagmaAction` are things that I had in mind. `MonoidAction` as you subsequently discovered (and I discovered the hard way as well) is too special to offer sufficient insight into actions in general (the same is true of `Monoid`, and I've fallen into that trap too.)

src/Algebra/Action/Construct/FreeMonoid.agda
jamesmckinna: Suggest: `Data.List.Relation.Binary.Equality.Setoid.Properties`? which does not currently exist... but conforms to other (pre-)existing nodes in the `Data.*.Relation.Binary.*.Setoid.*` style?

jamesmckinna:
Waking this up again, prompted by the discussion on #2553 / #2558 ...

JacquesCarette (approved):

JacquesCarette:
You 'woke it up' but it is still Draft?

jamesmckinna:
> You 'woke it up' but it is still Draft?

Still DRAFT, yes, it was more a 'note to self' to mark explicit the link with #2558 and to think about more about how to streamline what is here, esp. `Magma` actions, and the way the passage from an algebra to an algebra of functions... makes things 'more associative' than the vanilla algebra.

Plus, still irritated about the (lack of) parametrisation on left/right... and want to make this 'better' if possible...?

JacquesCarette:
Is perfect the enemy of the good here? Feels like there is quite a lot of quality stuff in here that could go in to v2.3.

jamesmckinna:
Thanks @JacquesCarette , I just haven't had any spare focus to devote to this. For sure it could be v2.3-able, but given the short timescale, I'm happy to make this v2.4 or even v3.0, when I will have had more time to reconsider things?

JacquesCarette:
You tagged the 'wrong' Jacques...  I was merely asking you to think whether what's here is already (essentially) good enough to go in.

jamesmckinna:
Thanks for the nudge. I'll try to look at this over the w/e

jamesmckinna:
Some minimal restructuring (plus profiting from now actually having the `(Is)Monoid` structure/bundle on `List` with setoid equality), introducing definitions, and `MagmaAction`, but otherwise will leave anything to downstream refactoring, I think!

src/Algebra/Action/Structures.agda
```diff
@@ -0,0 +1,95 @@
...
+  variable
+    x y z : A
+    m n p : M
+    ms ns ps : List M
+
+
+------------------------------------------------------------------------
+-- Basic definitions: fix notation
+
+record IsLeftAction : Set (a ⊔ r ⊔ c ⊔ ℓ) where
```
MatthewDaggitt, line 36: Hmm is there a reason why the hierarchy in `Structures` is not the same as the hierarchy in `Bundles`? e.g. `Bundles` talks about setoids and monoids which don't feature here?

src/Algebra/Action/Definitions.agda
```diff
@@ -0,0 +1,51 @@
...
+open import Function.Base using (id)
+open import Level using (Level; _⊔_)
+
+private
+  variable
+    x y : A
+    m n : M
+
+
+
```
MatthewDaggitt, line 25: Too many empty lines?

src/Algebra/Action/Definitions.agda
```diff
@@ -0,0 +1,51 @@
...
+Actʳ = A → M → A
+
+Congruentˡ : Actˡ → Set (c ⊔ a ⊔ ℓ ⊔ r)
+Congruentˡ _▷_ = _▷_ Preserves₂ _≈ᴹ_ ⟶ _≈_ ⟶ _≈_
+
+Congruentʳ : Actʳ → Set (c ⊔ a ⊔ ℓ ⊔ r)
+Congruentʳ _◁_ = _◁_ Preserves₂ _≈_ ⟶ _≈ᴹ_ ⟶ _≈_
+
+IsActionˡ  : Actˡ → Op₂ M → Set (c ⊔ a ⊔ r)
+IsActionˡ _▷_ _∙_ = ∀ m n x → ((m ∙ n) ▷ x) ≈ (m ▷ (n ▷ x))
```
MatthewDaggitt, line 42: I don't like using the word `Is` here as it suggests that it is a `Structure`? Same for `Identityˡ`. Also by convention for the definitions we use `Left` and `Right` rather than the superscripts? Maybe `LeftAction` and `LeftIdentity`?

src/Algebra/Action/Definitions.agda
```diff
@@ -0,0 +1,51 @@
...
+--
+-- Structure of the Action of one (pre-)Setoid on another
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Relation.Binary.Core using (Rel; _Preserves₂_⟶_⟶_)
+
+module Algebra.Action.Definitions
+  {c a ℓ r} {M : Set c} {A : Set a} (_≈ᴹ_ : Rel M ℓ) (_≈_ : Rel A r)
```
MatthewDaggitt, line 12: This module parameterisation doesn't work as many of these definitions either use neither or only one of the equalities. This will leave them ill-parameterised when this module is opened unapplied.

src/Algebra/Action/Structures.agda
```diff
@@ -0,0 +1,95 @@
...
+
+-- derived form: iterated action, satisfying congruence
+
+  _▷⋆_ : List M → A → A
+  ms ▷⋆ a = foldr _▷_ a ms
+
+  _▷⁺_ : List⁺ M → A → A
+  (m ∷ ms) ▷⁺ a = m ▷ ms ▷⋆ a
+
+  ▷⋆-cong : Pointwise _≈ᴹ_ ms ns → x ≈ y → (ms ▷⋆ x) ≈ (ns ▷⋆ y)
```
MatthewDaggitt, line 51: Most of these lemmas would go in `Algebra.Action.Properties` modules by convention.

src/Algebra/Action/Construct/Self.agda
```diff
@@ -0,0 +1,40 @@
...
+module Algebra.Action.Construct.Self where
+
+open import Algebra.Action.Bundles
+open import Algebra.Action.Structures using (IsLeftAction; IsRightAction)
+open import Algebra.Bundles using (Monoid)
+
+------------------------------------------------------------------------
+-- Left- and Right- Actions of a Monoid over itself
+
+module Regular {c ℓ} (M : Monoid c ℓ) where
```
MatthewDaggitt, line 18: Why is this module called `Regular` and why is it parameterised by a `Monoid`? What other algebraic structures do we want to extend this by?

src/Algebra/Action/Construct/Self.agda
```diff
@@ -0,0 +1,40 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Left- and right- regular actions
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Algebra.Action.Construct.Self where
```
MatthewDaggitt, line 9: Should this be called `Algebra.Action.Construct.Regular`?

src/Algebra/Action/Construct/FreeMonoid.agda
```diff
@@ -0,0 +1,70 @@
...
+-- The Agda standard library
+--
+-- The free MonoidAction on a SetoidAction, arising from ListAction
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Relation.Binary.Bundles using (Setoid)
+
+module Algebra.Action.Construct.FreeMonoid
```
MatthewDaggitt, line 11: Is this module going to contain other `Free` actions? It already contains the free setoid action which suggests that it might be ill named?

MatthewDaggitt (commented):
Given that I need to make a release candidate tomorrow, and there are several questions and at least one bug in this code I'm going to bump this to v2.4

jamesmckinna:
Happy to bump while these things get sorted out!

JacquesCarette:
I'd love to get this one going into v2.4!

jamesmckinna:
> I'd love to get this one going into v2.4!

Thank-you. As with a lot of recent comments from me, I take both encouragement from that, while noting that esp. @MatthewDaggitt 's review comments suggest that there is some way to go for a 'mature' design to emerge; despite my personal attachment to what's already here, I agree that it could be improved, but I need time/space/brains to think about it...

... which I'm not getting at the moment. Moving back to DRAFT, and bumping to (at least) v3.0.

PR #2345, opened 2024-04-05
Relation.Binary.PropositionalEquality over-use,

JacquesCarette:
Many uses of `Relation.Binary.PropositionalEquality` (which is kind of 'fat' for such a low-level module) can usually be 
done instead as importing `.Core` …and/or `.Properties` instead.

PR #2354, opened 2024-04-10
Systematise use of `Relation.Binary.Definitions.DecidableEquality`

jamesmckinna:
Addresses [comment](https://github.com/agda/agda-stdlib/pull/2345#discussion_r1555854588) in #2345

No `CHANGELOG` required!


JacquesCarette (approved):
Looks great.

Too bad the better way to do it is textually longer.

PR #2355, opened 2024-04-10
Add proofs to Data.List.Properties

mildsunrise:
Hi! This is my first PR, apologies if I'm missing something.

This adds a series of proofs to `Data.List.Properties` that I consider general enough to be in the stdlib, and in most cases already exist for other list operations. This proves, in order:

 - interaction with `map`, for list operations where it makes sense and was missing
 - congruence for higher order operations, where missing (except for those acting on `Decidable`s, which I am still working on, and `unsnoc` which is challenging because of the indexed datatype)
 - for `alignWith` / `align`, `zipWith` / `zip` and `unzipWith` / `unzip`:
   - effect of swapping function arguments / inputs
   - equivalence of swapping function arguments if function is commutative
 - interaction of `take` with itself and with `drop`
 - `unzipWith` + `zipWith` being inverses (with particular cases for the non-With versions)
 - interaction with `++`, for operations where it makes sense and was missing

It also fixes a `module` that was defining a parameter that was not used in the proofs.
I've tried to respect conventions of that file where possible, please tell me if there's anything that could be improved :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,6 +310,8 @@ Additions to existing modules
   zip-unzip             : uncurry′ zip ∘ unzip ≗ id
   unzipWith-zipWith     : f ∘ uncurry′ g ≗ id → length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
   unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
+  mapMaybe-++           : mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
+  unzipWith-++          : unzipWith f (xs ++ ys) ≡ zip _++_ _++_ (unzipWith f xs) (unzipWith f ys)
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -281,6 +281,17 @@ Additions to existing modules
   reverse-upTo          : reverse (upTo n) ≡ downFrom n
   reverse-applyDownFrom : reverse (applyDownFrom f n) ≡ applyUpTo f n
   reverse-downFrom      : reverse (downFrom n) ≡ upTo n
+  mapMaybe-map          : mapMaybe f ∘ map g ≗ mapMaybe (f ∘ g)
+  map-mapMaybe          : map g ∘ mapMaybe f ≗ mapMaybe (Maybe.map g ∘ f)
+  align-map             : align (map f xs) (map g ys) ≡ map (map f g) (align xs ys)
+  zip-map               : zip (map f xs) (map g ys) ≡ map (map f g) (zip xs ys)
+  unzipWith-map         : unzipWith f ∘ map g ≗ unzipWith (f ∘ g)
+  map-unzipWith         : map (map g) (map h) ∘ unzipWith f ≗ unzipWith (map g h ∘ f)
+  unzip-map             : unzip ∘ map (map f g) ≗ map (map f) (map g) ∘ unzip
+  splitAt-map           : splitAt n ∘ map f ≗ map (map f) (map f) ∘ splitAt n
+  uncons-map            : uncons ∘ map f ≗ map (map f (map f)) ∘ uncons
+  last-map              : last ∘ map f ≗ map f ∘ last
+  tail-map              : tail ∘ map f ≗ map (map f) ∘ tail
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -305,6 +305,8 @@ Additions to existing modules
   zip-flip              : zip xs ys ≡ map swap (zip ys xs)
   unzipWith-swap        : unzipWith (swap ∘ f) ≗ swap ∘ unzipWith f
   unzip-swap            : unzip ∘ map swap ≗ swap ∘ unzip
+  take-take             : take n (take m xs) ≡ take (n ⊓ m) xs
+  take-drop             : take n (drop m xs) ≡ drop m (take (m + n) xs)
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -308,7 +308,7 @@ Additions to existing modules
   take-take             : take n (take m xs) ≡ take (n ⊓ m) xs
   take-drop             : take n (drop m xs) ≡ drop m (take (m + n) xs)
   zip-unzip             : uncurry′ zip ∘ unzip ≗ id
-  unzipWith-zipWith     : length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
+  unzipWith-zipWith     : f ∘ uncurry′ g ≗ id → length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
   unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
   ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -337,6 +337,9 @@ Additions to existing modules
   unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
   mapMaybe-++           : mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
   unzipWith-++          : unzipWith f (xs ++ ys) ≡ zip _++_ _++_ (unzipWith f xs) (unzipWith f ys)
+  catMaybes-concatMap   : catMaybes ≗ concatMap fromMaybe
+  catMaybes-++          : catMaybes (xs ++ ys) ≡ catMaybes xs ++ catMaybes ys
+  map-catMaybes         : map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -323,7 +323,7 @@ Additions to existing modules
   mapMaybe-cong         : f ≗ g → mapMaybe f ≗ mapMaybe g
   zipWith-cong          : (∀ a b → f a b ≡ g a b) → ∀ as → zipWith f as ≗ zipWith g as
   unzipWith-cong        : f ≗ g → unzipWith f ≗ unzipWith g
-  foldl-cong            : (∀ x y → f x y ≡ g x y) → d ≡ e → foldl f d ≗ foldl g e
+  foldl-cong            : (∀ x y → f x y ≡ g x y) → ∀ x → foldl f x ≗ foldl g x
   alignWith-flip        : alignWith f xs ys ≡ alignWith (f ∘ swap) ys xs
   alignWith-comm        : f ∘ swap ≗ f → alignWith f xs ys ≡ alignWith f ys xs
   align-flip            : align xs ys ≡ map swap (align ys xs)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,10 @@ Additions to existing modules
   uncons-map            : uncons ∘ map f ≗ map (map f (map f)) ∘ uncons
   last-map              : last ∘ map f ≗ map f ∘ last
   tail-map              : tail ∘ map f ≗ map (map f) ∘ tail
+  mapMaybe-cong         : f ≗ g → mapMaybe f ≗ mapMaybe g
+  zipWith-cong          : (∀ a b → f a b ≡ g a b) → ∀ as → zipWith f as ≗ zipWith g as
+  unzipWith-cong        : f ≗ g → unzipWith f ≗ unzipWith g
+  foldl-cong            : (∀ x y → f x y ≡ g x y) → d ≡ e → foldl f d ≗ foldl g e
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -307,6 +307,9 @@ Additions to existing modules
   unzip-swap            : unzip ∘ map swap ≗ swap ∘ unzip
   take-take             : take n (take m xs) ≡ take (n ⊓ m) xs
   take-drop             : take n (drop m xs) ≡ drop m (take (m + n) xs)
+  zip-unzip             : uncurry′ zip ∘ unzip ≗ id
+  unzipWith-zipWith     : length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
+  unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,6 +296,12 @@ Additions to existing modules
   zipWith-cong          : (∀ a b → f a b ≡ g a b) → ∀ as → zipWith f as ≗ zipWith g as
   unzipWith-cong        : f ≗ g → unzipWith f ≗ unzipWith g
   foldl-cong            : (∀ x y → f x y ≡ g x y) → d ≡ e → foldl f d ≗ foldl g e
+  alignWith-flip        : alignWith f xs ys ≡ alignWith (f ∘ swap) ys xs
+  alignWith-comm        : f ∘ swap ≗ f → alignWith f xs ys ≡ alignWith f ys xs
+  align-flip            : align xs ys ≡ map swap (align ys xs)
+  zip-flip              : zip xs ys ≡ map swap (zip ys xs)
+  unzipWith-swap        : unzipWith (swap ∘ f) ≗ swap ∘ unzipWith f
+  unzip-swap            : unzip ∘ map swap ≗ swap ∘ unzip
   ```
 
 * In `Data.List.Relation.Unary.All.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ Bug-fixes
 * Module `Data.List.Relation.Ternary.Appending.Setoid.Properties` no longer
   exports the `Setoid` module under the alias `S`.
 
+* Remove unbound parameter from `Data.List.Properties.length-alignWith`,
+  `alignWith-map` and `map-alignWith`.
+
 Non-backwards compatible changes
 --------------------------------
 
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -342,6 +342,21 @@ module _ {f g : These A B → C} where
   map-alignWith g (x ∷ xs) (y ∷ ys) =
     cong₂ _∷_ refl (map-alignWith g xs ys)
 
+  alignWith-flip : ∀ xs ys →
+                 alignWith f xs ys ≡ alignWith (f ∘ These.swap) ys xs
+  alignWith-flip []       []       = refl
+  alignWith-flip []       (y ∷ ys) = refl
+  alignWith-flip (x ∷ xs) []       = refl
+  alignWith-flip (x ∷ xs) (y ∷ ys) = cong (_ ∷_) (alignWith-flip xs ys)
+
+module _ {f : These A A → B} where
+
+  alignWith-comm : f ∘ These.swap ≗ f →
+                 ∀ xs ys → alignWith f xs ys ≡ alignWith f ys xs
+  alignWith-comm f-comm xs ys = trans
+    (alignWith-flip xs ys)
+    (alignWith-cong f-comm ys xs)
+
 ------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


src/Data/List/Properties.agda
```diff
@@ -1233,15 +1406,46 @@ module _ {x y : A} where
 ∷ʳ-++ : ∀ (xs : List A) (a : A) (ys : List A) → xs ∷ʳ a ++ ys ≡ xs ++ a ∷ ys
 ∷ʳ-++ xs a ys = ++-assoc xs [ a ] ys
 
+------------------------------------------------------------------------
+-- uncons
 
+module _ (f : A → B) where
```
MatthewDaggitt, line 1412: Hmm so you've made all the `f`s here and below explicit, but the existing proof `head-map` has an implicit `f`. Can `f` in general be inferable in these situations?

src/Data/List/Properties.agda
```diff
@@ -317,18 +344,48 @@ module _ {f g : These A B → C} where
...
+  alignWith-flip []       []       = refl
+  alignWith-flip []       (y ∷ ys) = refl
+  alignWith-flip (x ∷ xs) []       = refl
+  alignWith-flip (x ∷ xs) (y ∷ ys) = cong (_ ∷_) (alignWith-flip xs ys)
+
+module _ {f : These A A → B} where
+
+  alignWith-comm : f ∘ These.swap ≗ f →
+                 ∀ xs ys → alignWith f xs ys ≡ alignWith f ys xs
+  alignWith-comm f-comm xs ys = trans
```
MatthewDaggitt, line 358: In general with these `trans` proofs we usually prefer to use equational reasoning as it makes the proofs a little bit easier to read and follow?

src/Data/List/Properties.agda
```diff
@@ -317,18 +344,48 @@ module _ {f g : These A B → C} where
...
+  align-flip : align xs ys ≡ map These.swap (align ys xs)
+  align-flip = trans
+    (alignWith-flip xs ys)
+    (sym (map-alignWith These.swap ys xs))
+
 ------------------------------------------------------------------------
 -- zipWith
 
-module _ (f : A → A → B) where
+module _ {f g : A → B → C} where
```
MatthewDaggitt, line 382: Eugh, I'm torn on this one whether this is a breaking change or not. As long as the user is not passing implicits explicitly it should be fine. Anyone else has have any feelings on it? 

If we decide it is fine, then we should definitely add a CHANGELOG entry for it.

src/Data/List/Properties.agda
```diff
@@ -122,6 +122,15 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
 ------------------------------------------------------------------------
 -- mapMaybe
 
+module _ {f g : A → Maybe B} where
+
+  mapMaybe-cong : f ≗ g → mapMaybe f ≗ mapMaybe g
+  mapMaybe-cong f≗g [] = refl
+  mapMaybe-cong f≗g (x ∷ xs) rewrite f≗g x
+    with ih ← mapMaybe-cong f≗g xs | g x
```
MatthewDaggitt, line 130: Why do we need the ih in a with? Why can't we just write `cong (y ∷_) (mapMaybe-cong f≗g xs)`? Similarly for proofs below?

src/Data/List/Properties.agda
```diff
@@ -295,6 +320,8 @@ module _ {f g : These A B → C} where
   alignWith-cong f≗g (a ∷ as)   (b ∷ bs) =
     cong₂ _∷_ (f≗g (these a b)) (alignWith-cong f≗g as bs)
 
+module _ {f : These A B → C} where
```
MatthewDaggitt, line 323: Nice spot, thank you!

src/Data/List/Properties.agda
```diff
@@ -317,18 +344,48 @@ module _ {f g : These A B → C} where
...
+-- align
+
+align-map : ∀ (f : A → B) (g : C → D) →
+            ∀ xs ys → align (map f xs) (map g ys) ≡
+                      map (These.map f g) (align xs ys)
+align-map f g xs ys = trans
+  (alignWith-map f g xs ys)
+  (sym (map-alignWith (These.map f g) xs ys))
+
+module _ (xs : List A) (ys : List B) where
```
MatthewDaggitt, line 372: Let's not start sticking lists in module parameters? `∀ xs ys →` is way shorter both in the line and in terms of lines.

src/Data/List/Properties.agda
```diff
@@ -365,6 +422,37 @@ module _ (f : A → B → C) where
...
 
+  zipWith-flip : ∀ xs ys → zipWith f xs ys ≡ zipWith (flip f) ys xs
+  zipWith-flip []       []       = refl
+  zipWith-flip []       (x ∷ ys) = refl
+  zipWith-flip (x ∷ xs) []       = refl
+  zipWith-flip (x ∷ xs) (y ∷ ys) = cong (f x y ∷_) (zipWith-flip xs ys)
+
+module _ (f : A → A → B) where
+
+  zipWith-comm : (∀ x y → f x y ≡ f y x) →
```
MatthewDaggitt, line 433: Should this be generalised to `(∀ x y → f x y ≡ g y x)`?

src/Data/List/Properties.agda
```diff
@@ -365,6 +422,37 @@ module _ (f : A → B → C) where
...
+-- zip
+
+zip-map : ∀ (f : A → B) (g : C → D) →
+          ∀ xs ys → zip (map f xs) (map g ys) ≡
+                    map (Product.map f g) (zip xs ys)
+zip-map f g xs ys = trans
+  (zipWith-map _,_ f g xs ys)
+  (sym (map-zipWith _,_ (Product.map f g) xs ys))
+
+module _ (xs : List A) (ys : List B) where
```
MatthewDaggitt, line 449: Again let's not put lists in module parameters.

src/Data/List/Properties.agda
```diff
@@ -527,6 +670,12 @@ module _ {P : Pred A p} {f : A → A → A} where
 ------------------------------------------------------------------------
 -- foldl
 
+foldl-cong : ∀ {f g : B → A → B} {d e : B} →
+             (∀ x y → f x y ≡ g x y) → d ≡ e →
```
MatthewDaggitt, line 674: Having a proof of `d ≡ e` isn't very consistent with the other cong proofs above right? They omit propositional equalities, and only look at functional equalities.

MatthewDaggitt (changes requested):
Hi @mildsunrise, thanks for the great first PR! Looks very good, but I have a few minor comments which I've added below.

jamesmckinna:
Meta-comment: 
Should we fix `mapMaybe` first?
Cf. https://github.com/agda/agda/issues/5833 and links there to @omelkonian 's implementation?

MatthewDaggitt:
What would "fixing" it involve?

jamesmckinna:
> What would "fixing" it involve?

As in Orestis' prelude, write `catMaybe` *before* `mapMaybe` as the composition of `catMaybe` and `map`...

src/Data/List/Properties.agda
mildsunrise: I was mirroring the style of the existing proofs. I figured they put ih in a with to avoid repetiton in both branches, so I did the same here. I can change that if you prefer

src/Data/List/Properties.agda
mildsunrise: the diff is a little bit messy here, what happened is that [`zipWith-map` was added](https://github.com/agda/agda-stdlib/pull/2355/commits/76940a93cf9b1c36098f45988ca710a2731e1d53) (i put it at the start for consistency) and later [`zipWith-comm` was moved to the bottom](https://github.com/agda/agda-stdlib/pull/2355/commits/bb9a8a2c22266315175b2f61a36013a631e1f7e2) since it can be implemented more easily in terms of `zipWith-flip` and `zipWith-cong`, but there should be no breaking change

src/Data/List/Properties.agda
mildsunrise: that would be a breaking change (again, this was an existing proof that was moved to the bottom) but it can be done if you prefer

src/Data/List/Properties.agda
mildsunrise: yeah, that was my thought as well, but I preferred to mirror the structure of the existing `foldr-cong` proof that does accept `d ≡ e`. can be changed if you prefer

src/Data/List/Properties.agda
mildsunrise: ah, i see, i will do that. however keep in mind I cannot write `∀ xs ys →` as that would leave unbound list types. my reasoning was that the module should introduce what's necessary to bind list types A / B, but it is also reasoning that we not put lists in module parameters

src/Data/List/Properties.agda
mildsunrise: right, I tried to figure out what to do from the existing `-map` proofs but they seem to be quite inconsistent, we sometimes use implicit and sometimes explicit. in most of them we do explicit, so that's what I went with for the new `-map` proofs.

as for whether it is in general inferable, it is hard for me to say. I usually prefer the resulting proof to be completely bound by the explicit parameters, but I don't have that much experience in Agda so I'll leave that call to you :)

src/Data/List/Properties.agda
mildsunrise: very reasonable, I'll do that here and in `unzip-flip` and `align-flip`

MatthewDaggitt: Yup there's various places in this file with a similar pattern.

src/Data/List/Properties.agda
MatthewDaggitt: Ah sorry, my misunderstanding!

src/Data/List/Properties.agda
MatthewDaggitt: Ah in which case, no it shouldn't be! Sorry thought it was a new proof!

mildsunrise:
@MatthewDaggitt thanks for the thorough review! I've responded to your points and will act on them soon-ish, probably today

@jamesmckinna is it okay if I add the proposed fix to this PR, or would you rather keeping it in a separate PR, merging that one, and then rebasing this one on top?

MatthewDaggitt:
>  As in Orestis' prelude, write catMaybe before mapMaybe as the composition of catMaybe and map... which avoids the awkward with construction...

Is this is a backwards compatible change with respect to reduction behaviour? We shouldn't do it now, if it's not.

jamesmckinna:
I'll put my hobby-horse back in the stable... @MatthewDaggitt 's comment suggests we exercise caution. I'll investigate separately, and then take the knock-on hit from your additions downstream @mildsunrise so no action required on your part from me!

mildsunrise:
to my understanding it should have equivalent reduction behavior, but yes, it's good to be cautious.
okay then, I'll fix the 2 remaining issues

src/Data/List/Properties.agda
mildsunrise: done

src/Data/List/Properties.agda
mildsunrise: done

mildsunrise:
all done. let me know if you need anything else on my part, such as squashing the fixup commits into the original commits

mildsunrise:
I've added a couple more properties, `mapMaybe-++` and `unzipWith-++` since their proof was similar enough to the existing proofs. If it feels like too much, I can leave them out of this PR

src/Data/List/Properties.agda
jamesmckinna: see also [my comment](https://github.com/agda/agda-stdlib/issues/2358#issuecomment-2051480712) on #2358

CHANGELOG.md
```diff
@@ -308,8 +308,10 @@ Additions to existing modules
   take-take             : take n (take m xs) ≡ take (n ⊓ m) xs
   take-drop             : take n (drop m xs) ≡ drop m (take (m + n) xs)
   zip-unzip             : uncurry′ zip ∘ unzip ≗ id
-  unzipWith-zipWith     : length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
+  unzipWith-zipWith     : f ∘ uncurry′ g ≗ id → length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
   unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
+  mapMaybe-++           : mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
+  unzipWith-++          : unzipWith f (xs ++ ys) ≡ zip _++_ _++_ (unzipWith f xs) (unzipWith f ys)
```
jamesmckinna, line 314: Can these be analysed using `foldr-++` etc.?

src/Data/List/Properties.agda
jamesmckinna: ... and for having introduced a lemma whose name/type should be (downstream) deprecated in favour of the 'correct' version... oh dear!

CHANGELOG.md
mildsunrise: ooooh, good point, let me see

mildsunrise: oh, I see. do you have context about the "because pattern-matching definitions are generative" bit? i would like to learn more about why defining operations in Data.List.Base in terms of folds is a bad idea

JacquesCarette: What? Defining things in terms of folds should be a *good* idea!

jamesmckinna: Apologies for not being clear... but for details about generativity, you probably need to look at [the documentation](https://agda.readthedocs.io/en/latest/language/lambda-abstraction.html), which, somewhat frustratingly, doesn't make clear the relationship with [ordinary function definitions](https://agda.readthedocs.io/en/latest/language/function-definitions.html)...

That said, I have previously suggested refactoring the library to express (almost) all the recursions in `Data.List.Base` in terms of `foldr`/`foldl`, but IIRC, that was... discouraged. 

The following simple test suggests that, indeed, one should proceed with caution:
```agda
rev : List A → List A
rev = foldl (flip _∷_) []

rev′ : List A → List A
rev′ xs = go xs [] -- inline the definition using reverseAcc
  where
  go : List A → List A → List A
  go [] ys = ys
  go (x ∷ xs) ys = go xs (x ∷ ys)

test : rev ≡ rev′
test = refl
{- fails with:
foldl (λ y x → x ∷ y) [] x != Data.List.Base.go x x [] of type
List _A_1900
when checking that the expression refl has type rev ≡ rev′
-}
```

jamesmckinna: So I'm not saying it's a 'bad' idea, only that one changes the *intensional* semantics, even if not the *extensional* such. Sigh. 
For better or worse, the library privileges definitions made in direct pattern-matching style rather than by specialising higher-order operations...

jamesmckinna: See also #2267 / #2269 etc. which *do* change the intensional semantics (wrt strictness), but arguably towards 'better' definitions...

jamesmckinna: Last word: if #2355 get merged, we can fix up `mapMaybe-++` etc. downstream in #2361 assuming that gets approval...

src/Data/List/Properties.agda
jamesmckinna: See also #2123 , second checkbox, on uses of 'irrefutable' `with` (when a `let` mucks up the flow of pattern-matching; LHS `let` will be a boon here)

src/Data/List/Properties.agda
```diff
@@ -122,6 +122,15 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
...
 -- mapMaybe
 
+module _ {f g : A → Maybe B} where
+
+  mapMaybe-cong : f ≗ g → mapMaybe f ≗ mapMaybe g
+  mapMaybe-cong f≗g [] = refl
+  mapMaybe-cong f≗g (x ∷ xs) rewrite f≗g x
+    with ih ← mapMaybe-cong f≗g xs | g x
+  ... | just y  = cong (y ∷_) ih
+  ... | nothing = ih
```
jamesmckinna, line 132: Having said all of that, my understand of @JacquesCarette 's point in #2359 is that such proof patterns might be (better?) finessed using the *dependent* function `Maybe.maybe`, in which case the `ih` terms might end up being able to be `let`-bound after all...?

src/Data/List/Properties.agda
```diff
@@ -145,6 +154,29 @@ module _ (f : A → Maybe B) where
   ... | just y  = s≤s ih
   ... | nothing = m≤n⇒m≤1+n ih
 
+  mapMaybe-++ : ∀ xs ys →
+                mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
+  mapMaybe-++ []       ys = refl
+  mapMaybe-++ (x ∷ xs) ys with ih ← mapMaybe-++ xs ys | f x
+  ... | just y  = cong (y ∷_) ih
+  ... | nothing = ih
```
jamesmckinna, line 162: Ditto.

src/Data/List/Properties.agda
```diff
@@ -145,6 +154,29 @@ module _ (f : A → Maybe B) where
...
+
+module _ (f : B → Maybe C) (g : A → B) where
+
+  mapMaybe-map : mapMaybe f ∘ map g ≗ mapMaybe (f ∘ g)
+  mapMaybe-map []       = refl
+  mapMaybe-map (x ∷ xs) with ih ← mapMaybe-map xs | f (g x)
+  ... | just y  = cong (y ∷_) ih
+  ... | nothing = ih
```
jamesmckinna, line 170: ...

src/Data/List/Properties.agda
```diff
@@ -145,6 +154,29 @@ module _ (f : A → Maybe B) where
...
+
+module _ (g : B → C) (f : A → Maybe B) where
+
+  map-mapMaybe : map g ∘ mapMaybe f ≗ mapMaybe (Maybe.map g ∘ f)
+  map-mapMaybe []       = refl
+  map-mapMaybe (x ∷ xs) with ih ← map-mapMaybe xs | f x
+  ... | just y  = cong (g y ∷_) ih
+  ... | nothing = ih
```
jamesmckinna, line 178: ...

jamesmckinna (approved):
Happy with this as-is, but if the suggested use of `Maybe.maybe` permits the uses of `with` to be removed, then (probably?) so much the better! UPDATED: or not... so please ignore!

src/Data/List/Properties.agda
Taneb: The thing with `with`, as I understand it, is that proving things _about_ functions that use `with` can be inconvenient. We only very very rarely prove things about this kind of property. I think `with` is fine throughout this PR

JacquesCarette: My main 'beef' with `with` is when it is used to define functions, as that forces your hand downstream on how your proofs are forced to be done. And taking them out of proofs doesn't give as large a reward.

Having said that, I'm also interested in having proofs be less 'low level' (i.e. forcing explicit evaluation by following the definition patterns of the original function) and having more equational proofs. This would allow a smaller 'core API' that is representation dependent, and a larger set of properties which are representation *independent*.

So it's not so much 'simplification' of a proof as moving towards more intrinsic proofs.

src/Data/List/Properties.agda
mildsunrise: @MatthewDaggitt should i remove the `d ≡ e` or leave it in place to mirror `foldr-cong`? happy to do what you all prefer here

src/Data/List/Properties.agda
mildsunrise: note that the use of `with` here mirrors the style of the existing `mapMaybe` proofs. if we want to move away from it, I'd prefer to leave it to a future PR that also changes the definition in Base

mildsunrise:
(I'll be a bit busy these days, but I'd rather not block other PRs with this one, so please ping me if action is needed on my part)

src/Data/List/Properties.agda
jamesmckinna: Absolutely! See #2361 although in the end I did *not* change the proof of `mapMaybe-concatMap`... but maybe I should!?

jamesmckinna:
@MatthewDaggitt can you re-review and/or merge now, please? I think that this one is ready now ;-)

gallais:
I think we just need the `d \== e` assumption to be removed and then we're good to merge.

src/Data/List/Properties.agda
MatthewDaggitt: Please could remove it? I think in the long run we want to try and remove it from the existing proofs as well.

MatthewDaggitt:
Yup once removed, will merge :+1:

src/Data/List/Properties.agda
mildsunrise: done. `d ≡ e` was replaced by an (also explicit) `y` argument

mildsunrise:
I see that #2361 was merged before this PR. given that `catMaybes` is no longer defined in terms of `mapMaybe`, it becomes necessary to prove properties about `catMaybes` directly. And once this is done, the properties of `mapMaybe` can be simplified to use those new proofs.

#2361 already did that with `length-mapMaybe` (added `length-catMaybes`). I've went ahead and done the same with the other three mapMaybe properties.

MatthewDaggitt (approved):

PR #2357, opened 2024-04-11
[fix #2253] Deprecate `Algebra.Structures.Biased.IsRing*`

### LINKED ISSUE #2253: Redundant `zero` field in `Algebra.Structures.Biased.IsRing*`?
Issue #2195 and associated PR #2209 removed redundant `zero` field from `Algebra.Structures.IsRing`.
But what about from [`Algebra.Structures.Biased.IsRing*`](https://github.com/agda/agda-stdlib/blob/f239dc7075fcb61817dfcb5e5baf9dda520d2a5d/src/Algebra/Structures/Biased.agda#L200)?
Any other such 'leftovers'?

jamesmckinna:
Also: tweak v2.0 deprecation warnings (I think correctly...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,12 @@ Deprecated names
   _-_  ↦  _//_
   ```
 
+* In `Algebra.Structures.Biased`:
+  ```agda
+  IsRing*  ↦  Algebra.Structures.IsRing
+  isRing*  ↦  Algebra.Structures.isRing
+  ```
+
 * In `Data.Nat.Divisibility.Core`:
   ```agda
   *-pres-∣  ↦  Data.Nat.Divisibility.*-pres-∣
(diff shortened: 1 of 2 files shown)


Taneb:
This was added in #1684, as a migration path when the definition of `IsRing` was changed to not have duplicate equivalence proofs. I don't have any objection to deprecating it, but shouldn't the others added then be deprecated at the same time?

* `IsSemiringWithoutOne*`
* `IsNearSemiring*`
* `IsSemiringWithoutAnnihilatingZero*`

jamesmckinna:
I think that the substance of the issue/PR is that `zero` is definable once you have `_-_`... and us that the case for the structures you identify?

JacquesCarette (approved):

MatthewDaggitt:
@Taneb any comment on this? I'm happy to deprecate.

Taneb (approved):

Issue #2359, opened 2024-04-12
List of sub-optimal definitions in `Data.List.Base`


JacquesCarette:
While looking at PR #2355 from @mildsunrise I noticed a ridiculous number of uses of `with`. These turn out to not be localized to `Data.List.Properties` but come from uses of `with` in `Data.List.Base`. The vast majority of them can be simplified, with a really nice knock-on effect to simplify quite a few proofs.

But I'm guessing this is yet another case of breaking change, so has to wait until 3.0? I think we should outlaw uses of `with` in function definitions, if favour of defining explicit 'helper' functions.

---

- [x] mapMaybe #2361 
- [ ] unalignWith
- [x] unfold #2364 
- [x] scanr -- but already in another PR? #2395 
- [x] initLast -- ditto?
- [x] unsnoc
- [x] takeWhile #2365 
- [x] dropWhile #2365
- [x] filter #2365
- [x] partition #2365
- [x] span #2365
- [x] linesBy #2366 
- [x] wordsBy #2366
- [x] derun #2365

jamesmckinna:
This is, I think, the standing criticism from @omelkonian and others (related to the various issue he has posted on the main issue tracker about `inspect` and our attempts to deprecate it...), but only indirectly alluded to in #1937 / #2123 ... although PRs associated with those have tackled *some* of the worst offenders (esp. wrt turning `with` into `let`). 

In response to @MatthewDaggitt 's cautious nudge, @mildsunrise suggests in #2355 that the refactoring of `catMaybe`/`mapMaybe` that I suggested based on Orestis' prelude may in fact have the same behaviour as the existing definitions, so we should investigate that for all the examples you have identified... and if the computational behaviour is the same, then I think that *shouldn't* count as `breaking`...

jamesmckinna:
Could you make a task list of the 'worst'/most egregious examples? Then the analysis could be parallelised across us all...

MatthewDaggitt:
Yes if the computational behaviour is the same then it's backwards compatible.

JacquesCarette:
As long as "forcing evaluation by `with`" and "forcing evaluation through pattern-matching" are 'the same computational behaviour', then yes. They sure have very different usability behaviour!!

JacquesCarette:
Updated with an actual list.

jamesmckinna:
[...]

PR #2361, opened 2024-04-13
Improve `Data.List.Base.mapMaybe` (#2359; deprecate use of `with` #2123)

jamesmckinna:
Initial commits do side-by-side comparison of new with old.
Also: misc. tidying up. 
If we're happy with the new versions not being a breaking change, then I'll remove the old definitions/proofs.

### DIFF
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -49,6 +49,14 @@ map : (A → B) → List A → List B
 map f []       = []
 map f (x ∷ xs) = f x ∷ map f xs
 
+catMaybes′ : List (Maybe A) → List A
+catMaybes′ []             = []
+catMaybes′ (nothing ∷ xs) = catMaybes′ xs
+catMaybes′ (just x  ∷ xs) = x ∷ catMaybes′ xs
+
+mapMaybe′ : (A → Maybe B) → List A → List B
+mapMaybe′ p = catMaybes′ ∘ map p
+
 mapMaybe : (A → Maybe B) → List A → List B
 mapMaybe p []       = []
 mapMaybe p (x ∷ xs) with p x
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -62,16 +62,16 @@ private
 
 module _ {x y : A} {xs ys : List A} where
 
-  ∷-injective : x ∷ xs ≡ y List.∷ ys → x ≡ y × xs ≡ ys
+  ∷-injective : x ∷ xs ≡ y ∷ ys → x ≡ y × xs ≡ ys
   ∷-injective refl = (refl , refl)
 
-  ∷-injectiveˡ : x ∷ xs ≡ y List.∷ ys → x ≡ y
+  ∷-injectiveˡ : x ∷ xs ≡ y ∷ ys → x ≡ y
   ∷-injectiveˡ refl = refl
 
-  ∷-injectiveʳ : x ∷ xs ≡ y List.∷ ys → xs ≡ ys
+  ∷-injectiveʳ : x ∷ xs ≡ y ∷ ys → xs ≡ ys
   ∷-injectiveʳ refl = refl
 
-  ∷-dec : Dec (x ≡ y) → Dec (xs ≡ ys) → Dec (x List.∷ xs ≡ y ∷ ys)
+  ∷-dec : Dec (x ≡ y) → Dec (xs ≡ ys) → Dec (x ∷ xs ≡ y ∷ ys)
   ∷-dec x≟y xs≟ys = Decidable.map′ (uncurry (cong₂ _∷_)) ∷-injective (x≟y ×-dec xs≟ys)
 
 ≡-dec : DecidableEquality A → DecidableEquality (List A)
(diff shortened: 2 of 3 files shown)


src/Data/List/Base.agda
```diff
@@ -49,6 +49,14 @@ map : (A → B) → List A → List B
 map f []       = []
 map f (x ∷ xs) = f x ∷ map f xs
 
+catMaybes′ : List (Maybe A) → List A
+catMaybes′ []             = []
+catMaybes′ (nothing ∷ xs) = catMaybes′ xs
+catMaybes′ (just x  ∷ xs) = x ∷ catMaybes′ xs
```
JacquesCarette, line 55:
```Agda
catMaybes′ (x  ∷ xs) = maybe _∷_ id x $ catMaybes′ xs
```
?

jamesmckinna: Much better!

jamesmckinna: What about
```agda
catMaybes″ : List (Maybe A) → List A
catMaybes″ = foldr (maybe′ _∷_ id) []
```
(starting to get dizzy with the possibilities, once you start 'fixing' things...)

JacquesCarette: Personally that is my favourite because it makes reasoning about it simplest: it is very clear which lemmas will apply.

The downside is that it's not very pedagogical, but I don't think that's something we should be aiming for. [More precisely: I would be happy with a `Data.List.Traditional` module with the expanded definitions and proofs of equivalence to the less-expanded definitions used in `.Base`.

JacquesCarette:
I'd like to approve, but this PR should "make up its mind" and just replace the old definitions with the new.

jamesmckinna:
> I'd like to approve, but this PR should "make up its mind" and just replace the old definitions with the new.

Fair enough. In due course... but I think everyone (else) should be satisfied that the new definitions are appropriately equivalent to the previous one.

src/Data/List/Properties.agda
```diff
@@ -122,28 +122,50 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
...
 mapMaybe-just []       = refl
 mapMaybe-just (x ∷ xs) = cong (x ∷_) (mapMaybe-just xs)
 
+mapMaybe′-just : (xs : List A) → mapMaybe′ just xs ≡ xs
+mapMaybe′-just []       = refl
+mapMaybe′-just (x ∷ xs) = cong (x ∷_) (mapMaybe′-just xs)
+
 mapMaybe-nothing : (xs : List A) →
-                   mapMaybe {B = A} (λ _ → nothing) xs ≡ []
+                   mapMaybe (λ _ → nothing {A = A}) xs ≡ []
```
gallais, line 150: Should this be generalised?

```suggestion
                   mapMaybe (λ _ → nothing {A = B}) xs ≡ []
```

src/Data/List/Base.agda
jamesmckinna: Second point might be worth discussing separately as an issue (elsewhere people have complained to me about the lack of a(n equivalent to `haskell`) 'prelude' with which to teach (no `Level`s, 'elementary' definitions, ...)

src/Data/List/Properties.agda
jamesmckinna: And latest commit restores the binding of `B` to `mapMaybe`...

jamesmckinna:
I think the only thing remaining now would be to delete the `MapMaybeTest` module...

src/Data/List/Properties.agda
```diff
@@ -122,28 +122,57 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
...
 mapMaybe-just []       = refl
 mapMaybe-just (x ∷ xs) = cong (x ∷_) (mapMaybe-just xs)
 
 mapMaybe-nothing : (xs : List A) →
-                   mapMaybe {B = A} (λ _ → nothing) xs ≡ []
+                   mapMaybe {B = B} (λ _ → nothing) xs ≡ []
 mapMaybe-nothing []       = refl
 mapMaybe-nothing (x ∷ xs) = mapMaybe-nothing xs
 
+module MapMaybeTest where
```
JacquesCarette, line 142: Just delete this?

JacquesCarette (approved):
I'd prefer if the test module was deleted, but the rest looks good.

jamesmckinna:
@Jacques, thanks for this.
I'm happy to remove the `Test` module, but happier still that it's in the commit history, so that if anyone downstream wishes to litigate the equivalence, they can find it...

src/Data/List/Base.agda
```diff
@@ -123,11 +114,11 @@ partitionSums : List (A ⊎ B) → List A × List B
...
-merge R? []       ys       = ys
-merge R? xs       []       = xs
-merge R? (x ∷ xs) (y ∷ ys) = if does (R? x y)
-  then x ∷ merge R? xs (y ∷ ys)
-  else y ∷ merge R? (x ∷ xs) ys
+merge R? []           ys           = ys
+merge R? xs           []           = xs
+merge R? xs@(x ∷ xs₁) ys@(y ∷ ys₁) = if does (R? x y)
+  then x ∷ merge R? xs₁ ys
+  else y ∷ merge R? xs  ys₁
```
gallais, line 121:
```suggestion
merge R? x∷xs@(x ∷ xs) y∷ys@(y ∷ ys) = if does (R? x y)
  then x ∷ merge R? x∷xs ys
  else y ∷ merge R? xs   y∷ys
```
?

src/Data/List/Properties.agda
```diff
@@ -122,28 +122,34 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
 ------------------------------------------------------------------------
 -- mapMaybe
 
+length-catMaybes : ∀ xs → length (catMaybes {A = A} xs) ≤ length xs
+length-catMaybes []        = ≤-refl
+length-catMaybes (x  ∷ xs) = let ih = length-catMaybes xs
+  in maybe {B = λ x → length (catMaybes (x ∷ xs)) ≤ suc (length xs)}
```
gallais, line 128: This is maybe just me but I find this a lot less readable.
It adds to the noise by forcing us to explicitly pass the motive `B` instead of
having the *assistant* write it for us.

src/Data/List/Properties.agda
```diff
@@ -122,28 +122,34 @@ map-injective finj {x ∷ xs} {y ∷ ys} eq =
...
   ... | just y  = cong (y ∷_) ih
   ... | nothing = ih
 
   length-mapMaybe : ∀ xs → length (mapMaybe f xs) ≤ length xs
-  length-mapMaybe []       = z≤n
-  length-mapMaybe (x ∷ xs) with ih ← length-mapMaybe xs | f x
-  ... | just y  = s≤s ih
-  ... | nothing = m≤n⇒m≤1+n ih
+  length-mapMaybe xs =
+    ≤-trans (length-catMaybes (map f xs)) (≤-reflexive (length-map f xs))
```
gallais, line 152: We should use the equational reasoning kit here.

src/Data/List/Properties.agda
jamesmckinna: Yes, I agree with you! Will update.

src/Data/List/Base.agda
```diff
@@ -123,11 +114,11 @@ partitionSums : List (A ⊎ B) → List A × List B
...
-merge R? []       ys       = ys
-merge R? xs       []       = xs
-merge R? (x ∷ xs) (y ∷ ys) = if does (R? x y)
-  then x ∷ merge R? xs (y ∷ ys)
-  else y ∷ merge R? (x ∷ xs) ys
+merge R? []           ys           = ys
+merge R? xs           []           = xs
+merge R? x∷xs@(x ∷ xs) y∷ys@(y ∷ ys) = if does (R? x y)
+  then x ∷ merge R? x∷xs ys
+  else y ∷ merge R? xs   y∷ys
```
gallais, line 121: Ah! got that backwards!

```suggestion
  then x ∷ merge R? xs   y∷ys
  else y ∷ merge R? x∷xs ys
```

src/Data/List/Properties.agda
jamesmckinna: Done!

jamesmckinna:
That's enough now, I think. 
@gallais could you add to the queue please? 
(still observing the no-self-merging-rule... ;-))

JacquesCarette (approved):

gallais:
Just FTR: don't @ people in commit messages because they willl
perpetually get notifications when that same commit is pushed
to various forks.

JacquesCarette:
> Just FTR: don't @ people in commit messages because they willl perpetually get notifications when that same commit is pushed to various forks.

I was wondering what that weirdly formatted email I'd just gotten was from.

jamesmckinna:
Ah! Sorry about that!

PR #2366, opened 2024-04-15
Improve `Data.List.Base` (fix #2359; deprecate use of with #2123)

### LINKED ISSUE #2359: List of sub-optimal definitions in `Data.List.Base`

jamesmckinna:
Refactor towards `if_then_else_`:
* `wordsBy`
* `linesBy`

### DIFF
--- a/src/Data/List/Base.agda
+++ b/src/Data/List/Base.agda
@@ -428,11 +428,10 @@ linesBy {A = A} P? = go nothing where
 
   go : Maybe (List A) → List A → List (List A)
   go acc []       = maybe′ ([_] ∘′ reverse) [] acc
-  go acc (c ∷ cs) =
-    let acc′ = Maybe.fromMaybe [] acc in
-    if does (P? c)
-      then reverse acc′ ∷ go nothing cs
-      else go (just (c ∷ acc′)) cs
+  go acc (c ∷ cs) = if does (P? c)
+    then reverse acc′ ∷ go nothing cs
+    else go (just (c ∷ acc′)) cs
+    where acc′ = Maybe.fromMaybe [] acc
 
 linesByᵇ : (A → Bool) → List A → List (List A)
 linesByᵇ p = linesBy (T? ∘ p)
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

src/Data/List/Base.agda
```diff
@@ -428,9 +428,10 @@ linesBy {A = A} P? = go nothing where
 
   go : Maybe (List A) → List A → List (List A)
   go acc []       = maybe′ ([_] ∘′ reverse) [] acc
-  go acc (c ∷ cs) with does (P? c)
-  ... | true  = reverse (Maybe.fromMaybe [] acc) ∷ go nothing cs
-  ... | false = go (just (c ∷ Maybe.fromMaybe [] acc)) cs
+  go acc (c ∷ cs) = let acc′ = Maybe.fromMaybe [] acc in
```
MatthewDaggitt, line 431: Any chance we could lay this out as:
```
go acc (c ∷ cs) =
  let acc′ = Maybe.fromMaybe [] acc in
  if does (P? c)
    then reverse acc′ ∷ go nothing cs
    else go (just (c ∷ acc′)) cs
```

MatthewDaggitt (changes requested):
Otherwise looks good!

src/Data/List/Base.agda
jamesmckinna: Will do (eventually)

jamesmckinna: Done.

Taneb (approved):

src/Data/List/Base.agda
jamesmckinna: Apologies, I realise I didn't do *precisely* what you asked for, and then re-read #2369 so have updated to use a `where` clause now... and now the layout is consistent with `wordsBy` in L452-L454 below...

MatthewDaggitt (approved):

PR #2367, opened 2024-04-16
[ new ] IO buffering, and loops

gallais:


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Non-backwards compatible changes
   parametrized by _raw_ bundles, and as such take a proof of transitivity.
 * The definitions in `Algebra.Module.Morphism.Construct.Identity` are now
   parametrized by _raw_ bundles, and as such take a proof of reflexivity.
+* The module `IO.Primitive` was moved to `IO.Primitive.Core`.
```
JacquesCarette, line 29: For backwards compatibility, shouldn't there be an `IO.Primitive` that re-exports it, for now?

tests/system/io/Main.agda
```diff
@@ -0,0 +1,17 @@
...
+
+open import Data.String.Base
+open import Function.Base using (_$_)
+open import IO
+open import System.Exit using (exitSuccess)
+
+main : Main
+main = run $ do
+  hSetBuffering stdout noBuffering
+  forever $ do
```
JacquesCarette, line 13: It was really not clear to me that this would terminate, that somehow `getLine` receiving an empty line would 'escape' the forever! Perhaps comment?

JacquesCarette (changes requested):
Only a compatibility thing, otherwise this looks great.

tests/system/io/Main.agda
gallais: It's the `"exit"` keyword that makes the function exit. But I'll add a comment, sure.

JacquesCarette (approved):

src/IO.agda
```diff
@@ -27,6 +27,7 @@ private
 -- Re-exporting the basic type and functions
 
 open import IO.Base public
+open import IO.Handle public
```
MatthewDaggitt, line 30: I feel the fact that `IO` now exports a load new operations should be reflected in the CHANGELOG?

MatthewDaggitt (commented):
Otherwise looks fantastic!

src/IO.agda
gallais: Done

JacquesCarette:
(since @MatthewDaggitt 's condition has been fulfilled, hitting 'merge')

src/IO/Primitive.agda
```diff
@@ -1,38 +1,15 @@
...
-{-# COMPILE UHC pure = \_ _ x -> UHC.Agda.Builtins.primReturn x #-}
-{-# COMPILE UHC _>>=_  = \_ _ _ _ x y -> UHC.Agda.Builtins.primBind x y #-}
+open import IO.Primitive.Core public
 
--- Haskell-style alternative syntax
-return : A → IO A
-return = pure
+{-# WARNING_ON_IMPORT
+"IO.Primitive was deprecated in v2.1. Use IO.Primitive.Core instead."
+#-}
```
andreasabel, line 15: @gallais: Are there plans to use `IO.Primitive` for something else?
(Otherwise, why the deprecation?)
(Maybe the end user want to import the "top-level" modules rather than `Core` modules.)

gallais: No plans on my end. It was a cleanup because having `IO.Primitive` sit "above" other primitive-binding
modules like `IO.Primitive.Finite` seemed at odds with our current practice of the `X.Y` module being
a partial re-export of `X.Y.Z` ones.

It's named `Core` because we don't expect anyone to be importing it (or any of the `IO.Primitive` modules)
for that matter as we provide a high-level `IO` interface.

PR #2368, opened 2024-04-17
[ new ] System.Random bindings

gallais:
Following a discussion on zulip

### DIFF
--- a/src/System/Random/Primitive.agda
+++ b/src/System/Random/Primitive.agda
@@ -1,0 +1,42 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Primitive System.Random simple bindings to Haskell functions
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible #-}
+
+module System.Random.Primitive where
+
+open import Agda.Builtin.IO using (IO)
+open import Agda.Builtin.Char using (Char)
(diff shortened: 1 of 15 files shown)


JacquesCarette (approved):
Nice.

Ah, good ole `trustMe`. But what's a little `trustMe` between `IO` friends, eh?

src/System/Random/Primitive.agda
```diff
@@ -36,7 +36,7 @@ postulate
 {-# COMPILE GHC randomRIO-Int = randomRIO #-}
 {-# COMPILE GHC randomIO-Float = randomIO #-}
 {-# COMPILE GHC randomRIO-Float = randomRIO #-}
-{-# COMPILE GHC randomIO-Nat = randomIO #-}
+{-# COMPILE GHC randomIO-Nat = abs <$> randomIO #-}
```
gallais, line 39: Nasty little bug here: `Nat` is `Integer` at runtime and so we need to be more careful here!

JacquesCarette (approved):

src/System/Random/Primitive.agda
gallais: If there ever were an ad for contributing a case to the testsuite when we mess with the FFI, that's it.

MatthewDaggitt:
Looks great to me. No comments.

PR #2372, opened 2024-04-22
[ new ] Effect.Monad.Random

gallais:
Still TODO:

* [x] instances for transformers
* [x] test case


jamesmckinna:
This looks (eventually? towards `QuickCheck`-style generators for testing?) useful. Any reason it can't be tidied up and merged, if necessary with a forward pointer to (an issue/PR covering) the remaining TODOs?

JacquesCarette:
Agree with James that it would be nice to have this in sooner rather than later (i.e. by pushing the TODOs to later).

gallais:
I have some horrible nonsense to do so of course I would rather (finally) finish this up.

src/Effect/Monad/Random.agda
```diff
@@ -0,0 +1,249 @@
...
+record RawMonadRandom
+       (A : Set f)
+       (M : Set f → Set g)
+       : Set (f ⊔ g) where
+  field
+    getRandom : M A
+
+record RawMonadRandomR
+       (A : Set f) (_≤_ : Rel A f)
+       (M : Set f → Set g)
```
jamesmckinna, line 42: Is this telescope `Level` polymorphic enough? `InBounds` in the usual `Setoid`-y way, has a `Level` determined by both that of `A`, and that of `_≤_`, but here they're both squashed into `f`?

src/Effect/Monad/Random.agda
```diff
@@ -0,0 +1,249 @@
...
+-- Operations over RawMonadRandom
+
+forgetRanged : RawMonadRandomR A R M → RawMonadRandom A M
+forgetRanged m = record { getRandom = RawMonadRandomR.getRandom m }
+
+------------------------------------------------------------------------
+-- IO monad specifics
+
+module MkRandomIOInstances
+  {a} {A : Set a} (_≤_ : Rel A a)
```
jamesmckinna, line 58: Ditto. here re: `Level` polymorphism.

JacquesCarette:
@gallais are you intending to come back to this, or would you be happy if someone else pushed this to the finish line?

src/Effect/Monad/Random.agda
gallais: The problem is that having both

```agda
getRandom  : M A
getRandomR : (lo hi : A) → .(lo≤hi : lo ≤ hi) → M (InBounds _≤_ lo hi)
```

means that `M`'s level for its codomain should be the same for `A` & `InBounds _≤_ lo hi`
which constrains `_≤_`'s to be the same as `A`'s.

An alternative would be to have

```agda
getRandom  : M (Lift r A)
```

but I'm not sure that's what we want.

gallais:
I'm happy for people to take this over if I'm not responsive enough.
In this instance, I think the current design works though.

JacquesCarette (approved):

JacquesCarette:
This looks 'done' to me -- maybe @jamesmckinna and @MatthewDaggitt can take a look too?

jamesmckinna:
The patches to `CHANGELOG` look as they they will *remove* L81-102, L104, but AFAICR, those additions are (should be!?) still part of the v2.4/current `master` version? Is there some merge weirdness here? Or am I too brain-fogged by illness at the moment?

JacquesCarette:
That's probably my fault - I tried to do a merge, and I may have gotten it wrong. It looked like obsolete stuff because this PR is old. Go ahead and revert my merge.

jamesmckinna:
> ... Go ahead and revert my merge.

Actually, that's a bad idea for me at the moment. Brain fog.

jamesmckinna:
@JacquesCarette I've just tried to fix up `CHANGELOG`... will now attempt a merge conflict fix.
If successful, will merge.

PR #2373, opened 2024-04-24
Very dependent map

### LINKED ISSUE #833: Very dependent map and zipWith?
Buried in the old categories library, I found the following gems:
```agda
  map⁎ : ∀ {a b p q} {A : Set a} {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} →
        (f : (x : A) → B x) → (∀ {x} → (y : P x) → Q y (f x)) → (v : Σ A P) → Σ (B (proj₁ v)) (Q (proj₂ v))
  map⁎ f g (x , y) = (f x , g y)

  map⁎′ : ∀ {a b p q} {A : Set a} {B : A → Set b} {P : Set p} {Q : P → Set q} → (f : (x : A) → B x) → ((x : P) → Q x) → (v : A × P) → B (proj₁ v) × Q (proj₂ v)
  map⁎′ f g (x , y) = (f x , g y)

  zipWith : ∀ {a b c p q r s} {A : Set a} {B : Set b} {C : Set c} {P : A → Set p} {Q : B → Set q} {R : C → Set r} {S : (x : C) → R x → Set s} (_∙_ : A → B → C) → (_∘_ : ∀ {x y} → P x → Q y → R (x ∙ y)) → (_*_ : (x : C) → (y : R x) → S x y) → (x : Σ A P) → (y : Σ B Q) → S (proj₁ x ∙ proj₁ y) (proj₂ x ∘ proj₂ y)
  zipWith _∙_ _∘_ _*_ (a , p) (b , q) = (a ∙ b) * (p ∘ q)
```
Should these be added to the standard library?

JacquesCarette:
Closes #833 .

The names, of course, are up for debate! (And, now that I think of it, I should probably add a property or two of them in this same PR).


jamesmckinna:
Re: naming
Is it possible to reconcile the name choices for this PR, and the general principles in #278 ?

JacquesCarette:
I guess this would push towards naming the first one simply `map`? Except that `Data.Product.Base` already has such a thing, and it is much less dependent.

My feeling is that #278 deals with the case of "normally dependent" naming -- here the functions really are absurdly dependent. So they seem like they'd need their own, new column. Interestingly, the various type families in `dep-map` are left implicit, and Agda seems happy enough with that. Having said that, I tried to apply `dep-map` to non-dependent cases, and it all became a giant yellow party.

In other words, we don't need the results of @gallais 's #278 experiment to know that we can't make these very-dependent version the 'default' with the most basic name.

CHANGELOG.md
```diff
@@ -458,6 +458,21 @@ Additions to existing modules
   product-↭ : product Preserves _↭_ ⟶ _≡_
   ```
 
+* Added some very-dependent map and zipWith to `Data.Product`.
+  ```agda
+  dep-map : {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} →
+   (f : (x : A) → B x) → (∀ {x} → (y : P x) → Q y (f x)) →
+   (v : Σ A P) → Σ (B (proj₁ v)) (Q (proj₂ v))
```
jamesmckinna, line 465: I have a personal aversion to `proj₁` and `proj₂` occurring in types like this, so a possible alternative is to use an irrefutable `@`-pattern in the type, i.e.
```agda
(v@(x , y) : Σ A P) → Σ (B x) (Q y)
```
etc.

CHANGELOG.md
```diff
@@ -458,6 +458,21 @@ Additions to existing modules
...
+
+  dep-map′ : {B : A → Set b} {P : Set p} {Q : P → Set q} →
+    (f : (x : A) → B x) → ((x : P) → Q x) → (v : A × P) → B (proj₁ v) × Q (proj₂ v)
+
+  zipWith : {P : A → Set p} {Q : B → Set q} {R : C → Set r} {S : (x : C) → R x → Set s}
+    (_∙_ : A → B → C) → (_∘_ : ∀ {x y} → P x → Q y → R (x ∙ y)) →
+    (_*_ : (x : C) → (y : R x) → S x y) →
+    (x : Σ A P) → (y : Σ B Q) → S (proj₁ x ∙ proj₁ y) (proj₂ x ∘ proj₂ y)
```
jamesmckinna, line 473: Ditto.

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+     (∀ {x} → (y : P x) → h y ≡ i y) →
+     (v : Σ A P) → dep-map f h v ≡ dep-map g i v
+  dep-map-cong f≗g h≗i (x , y) = cong₂ _,_ (f≗g x) (h≗i y)
+
+------------------------------------------------------------------------
+-- dep-map′
+
+module _ {B : A → Set b} {P : Set p} {Q : P → Set q} where
+
+  dep-map′-cong : {f g : (x : A) → B x} → {h i : (x : P) → Q x} →
```
jamesmckinna, line 38: I have a slight (mathematical!) preference for `h k` as consecutive function names, rather than `h i`...

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+  variable
+    a b p q r s : Level
+    A B C : Set a
+
+------------------------------------------------------------------------
+-- dep-map
+
+module _ {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} where
+
+  dep-map-cong : {f g : (x : A) → B x} → {h i : ∀ {x} → (y : P x) → Q y (f x)} →
```
jamesmckinna, line 27: See below: `h k` for preference?

JacquesCarette:
Thanks - these are all great suggestions. (The `h i` actually come from me cutting and pasting from elsewhere in `stdlib`, but I'll accept the blame for not relabeling). Will implement as soon as I have a moment.

CHANGELOG.md
jamesmckinna: Apologies! Agda allows this in `module` telescopes (I think!?), but not in ordinary function telescopes... so you'd need 
```agda
   (v : Σ A P) → (let x , y = v) → Σ (B x) (Q y)
```
instead...
... but still worth it, I think!

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+     (∀ x → f x ≡ g x) →
```
jamesmckinna, line 28: Seeing this written out explicitly seems to suggest that `Relation.Binary.PropositionalEquality.Core._≗_` could/should be made usefully (more) dependent in its second type argument `B`?

CHANGELOG.md
jamesmckinna: Ha! Fixed in 2.6.4.*, but fails in v2.6.3... version error on my part!

jamesmckinna: or even (more) simply:
```agda
((x , y) : Σ A P) → Σ (B x) (Q y)
```

jamesmckinna: Naming : `map-Σ`? or `Σ-map`?

JacquesCarette: All good suggestions, all done.

src/Data/Product/Properties/Dependent.agda
JacquesCarette: It could... but I'm quite afraid that this would cause inference slowdowns (and maybe even some yellow). In any case, I think that should be done as a separate experiment.

src/Data/Product.agda
```diff
@@ -8,20 +8,40 @@
...
 open import Data.Product.Base public
 
+-- These are here as they are not 'basic' but instead "very dependent",
+-- i.e. the result type depends on the full input 'point' (x , y).
+map-Σ : {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} →
+   (f : (x : A) → B x) → (∀ {x} → (y : P x) → Q y (f x)) →
+   ((x , y) : Σ A P) → Σ (B x) (Q y)
+map-Σ f g (x , y) = (f x , g y)
+
+-- This is a "non-dependent" version of dep-map whereby the input is actually
```
jamesmckinna, line 31: Nitpick: need to correct the mention of `dep-map` to whatever we settle on, but very happy with you taking my suggestion of `map-Σ`...

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+  variable
+    a b p q r s : Level
+    A B C : Set a
+
+------------------------------------------------------------------------
+-- dep-map
+
+module _ {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} where
+
+  dep-map-cong : {f g : (x : A) → B x} → {h i : ∀ {x} → (y : P x) → Q y (f x)} →
```
jamesmckinna, line 27: Ditto: `map-Σ-cong` here?

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+     (∀ {x} → (y : P x) → h y ≡ i y) →
+     (v : Σ A P) → map-Σ f h v ≡ map-Σ g i v
+  dep-map-cong f≗g h≗i (x , y) = cong₂ _,_ (f≗g x) (h≗i y)
+
+------------------------------------------------------------------------
+-- dep-map′
+
+module _ {B : A → Set b} {P : Set p} {Q : P → Set q} where
+
+  dep-map′-cong : {f g : (x : A) → B x} → {h i : (x : P) → Q x} →
```
jamesmckinna, line 38: Ditto...

jamesmckinna (approved):
Very happy with this now.
Once another set of eyes has agreed to the names, suggest your rectify all uses in comments/lemmas, and then you're good to go!

src/Data/Product.agda
JacquesCarette: D'oh! Will of course fix all of these silly omissions. Hurray for code reviews.

src/Data/Product/Properties/Dependent.agda
```diff
@@ -0,0 +1,53 @@
...
+module _ {B : A → Set b} {P : A → Set p} {Q : {x : A} → P x → B x → Set q} where
+
+  map-Σ-cong : {f g : (x : A) → B x} → {h k : ∀ {x} → (y : P x) → Q y (f x)} →
+     (∀ x → f x ≡ g x) →
+     (∀ {x} → (y : P x) → h y ≡ k y) →
+     (v : Σ A P) → map-Σ f h v ≡ map-Σ g k v
+  map-Σ-cong f≗g h≗k (x , y) = cong₂ _,_ (f≗g x) (h≗k y)
+
+------------------------------------------------------------------------
+-- dep-map′
```
MatthewDaggitt, line 34: Need to update the comment headings, but otherwise, I really like the names :+1:

MatthewDaggitt (changes requested):

src/Data/Product/Properties/Dependent.agda
JacquesCarette: Fixed!

jamesmckinna:
I think we're there now!

MatthewDaggitt (approved):

PR #2374, opened 2024-04-30
Add `_>>_` for `IO.Primitive.Core`

ncfavier:
This has to be in scope for desugaring `do` blocks that don't bind intermediate results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -492,6 +492,11 @@ Additions to existing modules
   forever : IO ⊤ → IO ⊤
   ```
 
+* In `IO.Primitive.Core`:
+  ```agda
+  _>>_ : IO A → IO B → IO B
+  ```
+
 * In `Data.Word.Base`:
   ```agda
   _≤_ : Rel Word64 zero
--- a/src/IO/Primitive/Core.agda
+++ b/src/IO/Primitive/Core.agda
@@ -36,3 +36,6 @@ postulate
 -- Haskell-style alternative syntax
 return : A → IO A
 return = pure
+
+_>>_ : IO A → IO B → IO B
+a >> b = a >>= λ _ → b


JacquesCarette:
Please all add CHANGELOG entry - but otherwise looks good.

jamesmckinna:
[Comment deleted after rereading the PR name in the light of the actual commit: but it seems a bad idea in general to mention deprecated module names in PRs?]

gallais:
What is your use case? In theory you're not supposed to use `IO.Primitive` and instead rely
on `IO.Base` which does have everything you need to desugar `do` blocks.

ncfavier:
Oh, I guess a better answer would be: I want to do IO and I don't want to enable `--guardedness`.

jamesmckinna:
> Oh, I guess a better answer would be: I want to do IO and I don't want to enable `--guardedness`.

But is that sufficient grounds for us extending such functionality to all clients of `stdlib`? I'm not clear whether we should...

ncfavier:
Is there a cost?

jamesmckinna:
> Is there a cost?

So I realise that I am out of my depth here; I had imagined (at a minimum) that moving the definition of `_>>_` would at least incur a deprecation cost, but digging a bit deeper I now see that this isn't 'moving' that definition at all...

... But there remains the question of how/whether the library should support your use case... and for that, I'll throw the question back to you and @gallais ... but perhaps in this instance it's 'harmless' (although slavishly offering a 'shallow' emulation of haskell IO, and exposing that as part of the library API, seems maybe something we don't want?)...

... but then: what's the cost (to you) of working with `--guardedness`?

gallais (approved):

jamesmckinna (approved):
You convinced me!

Issue #2362, opened 2024-04-14
lib2.0 :  `ℕᵇ `(binary) can be pure unlike `ℕ`.


mechvel:
I recall it and return to pointing out that:
in lib-2.0 the part of  Data.Nat.Binary.Properties
has several proofs that cannot be considered as	reliable.
Their badness is in that the proofs
```
(Ps)   +-assoc, +-comm, *-assoc, *-comm
```
and some other,
rely on similar proofs for `ℕ `(unary representation)	by using the isomoprphism to `ℕ`.
But these proofs for `ℕ` are not implemented as the first class, are not reliable. Because  `Nat._+_` and `Nat._*_`  are implemented as built-in, not in Agda. So that these proofs for `ℕ` are not about the real algorithms for computing these operations.

And `ℕᵇ` is different. Its arithmetic is implemented totally in Agda.
The	corresponding proofs can also be implemented totally in Agda.
In this sense, `ℕᵇ`  is a better Agda model for natural	numbers than `ℕ`.
So that it is better to untie `ℕᵇ` from `ℕ`, because using proofs for `ℕ` spoils the purity of `ℕᵇ`.

gallais:
> Because Nat._+_ and Nat._*_ are implemented as built-in, not in Agda. 

That is incorrect: these definitions are implemented in Agda, and their proofs are
using the Agda definition. It's only during compilation (when the proofs are getting
erased anyway) that these definitions are substituted for their more efficient built-ins.

mechvel:
On 2024-04-14 22:29, G. Allais wrote:
>> Because Nat._+_ and Nat._*_ are implemented as built-in, not in
>> Agda.
> 
> That is incorrect: these definitions are implemented in Agda, and
> their proofs are
> using the Agda definition. It's only during compilation (when the
> proofs are getting
> erased anyway) that these definitions are substituted for their more
> efficient built-ins.
> 

Something is wrong here.
If it was in Agda, the module Data.Nat.Base  would contain, for	example,

_+_ : Op₂ ℕ   	      	  -- algorithm "Plus"
0     + n = n
suc m +	n = suc	(m + n)

Instead	it contains
```
open import Agda.Builtin.Nat public  using (_+_ ...) ...
```
This actually means
"believe that this built-in-+ returns the same result as the Plus 
program in Agda".
This is	postulation of that Plus is equivalent to this invisible 
built-in-+.

`Binary._+_`  is different, its	Agda code is in	the library written in 
Agda,
[...]

PR #2376, opened 2024-05-01
[ new ] Word8

gallais:


src/Data/Word8/Literals.agda
```diff
@@ -0,0 +1,19 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Byte Literals
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible #-}
```
JacquesCarette, line 7: Not `--safe` ?

tests/data/bytestring/expected
```diff
@@ -0,0 +1,10 @@
+"\SOH\STX\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\ETX"
+------------------------------------------------------------------------
+1 = 0b00000001
+1 = 0
```
gallais, line 4: hmmm

gallais:
I think that's now ready. Writing the test has, as always, revealed that some
useful functions were missing and so I have added them.

jamesmckinna:
Re: @JacquesCarette 's comments about `--safe`: Out of interest, what correctness guarantees/checks (if any) are provided for primitive machine arithmetic? 

Do we have anywhere a reference implementation of eg octal, hex or higher word size operations?

cf. #2362

gallais:
> what correctness guarantees/checks (if any) are provided for primitive machine arithmetic?

None. IIRC the builtin `Word64` is there primarily because of
```
primQNameToWord64s : Name → Σ Word64 (λ _ → Word64)
```
which gives us a way to define an order on `QName`s to e.g. put them as Map keys.
No interest was paid to its computational content.

src/Reflection/AST/Name.agda
```diff
@@ -10,7 +10,7 @@ module Reflection.AST.Name where
 
 open import Data.List.Base              using (List)
 import Data.Product.Properties as Prodₚ using (≡-dec)
-import Data.Word.Properties as Wₚ       using (_≟_)
+import Data.Word64.Properties as Wₚ       using (_≟_)
```
jamesmckinna, line 13: Suggest instead avoiding the `Xₚ` suffix in favour of simply using the root module/datatype name `X` cf. #2280

gallais: Note that it's not a change *I* am making.

src/Tactic/Cong.agda
```diff
@@ -35,7 +35,7 @@ open import Data.List.Base   as List  using ([]; _∷_)
 open import Data.Maybe.Base  as Maybe using (Maybe; just; nothing)
 open import Data.Nat.Base    as ℕ     using (ℕ; zero; suc; _≡ᵇ_; _+_)
 open import Data.Unit.Base            using (⊤)
-open import Data.Word.Base   as Word  using (toℕ)
+open import Data.Word64.Base   as Word64  using (toℕ)
 open import Data.Product.Base         using (_×_; map₁; _,_)
 open import Function                  using (flip; case_of_)
```
jamesmckinna, line 40: Fix knock-on effect on the alignment here?

gallais: I think we should give up on non alpha-stable alignment

jamesmckinna (approved):
I don't feel especially competent to comment on this kind of 'system'/'FFI'-level contribution, but since most of the PR consists of boilerplate renaming/deprecation, and the (new) tests pass... it seems fine to me.

JacquesCarette (approved):

PR #2377, opened 2024-05-02
fixes #2375

### LINKED ISSUE #2375: lemma for `map` for `⊆` as Subset
`lib-2.0`  needs a proof for the following lemma.
```
map⊆ :  (f : C → C') → (f Preserves _≈_ ⟶ _≈'_) → (map f) Preserves _⊆_ ⟶ _⊆'_
```
This means:
if ``(f : C → C') , (f Preserves _≈_ ⟶ _≈'_)``,
is a map between the carriers of two setoids that agrees with the two equalities, then it preserves the relation `⊆` of `Subset`.

A simlar lemma for `Sublist` is provided in ``Data.List.Relation.Binary.Sublist.Setoid.Properties``.
And the suggested lemma could probably be in
``Data.List.Relation.Binary.Subset.Setoid.Properties``.

jamesmckinna:
NB *contra* @mechvel I have left `f` *implicitly* bound, on the prototype of `Data.List.Relation.Binary.Sublist.Setoid.Properties.map⁺`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -373,6 +373,11 @@ Additions to existing modules
   map-catMaybes         : map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
   ```
 
+* In `Data.List.Relation.Binary.Subset.Setoid.Properties`
+  ```
+  map⁺ : f Preserves _≈_ ⟶ _≈′_ → as ⊆ bs → map f as ⊆′ map f bs
+  ```
+
 * In `Data.List.Relation.Unary.All.Properties`:
   ```agda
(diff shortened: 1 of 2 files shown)


Taneb:
Is there a corresponding map- that could be added at the same time?

jamesmckinna:
> Is there a corresponding map- that could be added at the same time?

Not without injectivity of`f`, I think...
... and might be considered mission creep/out of scope for this PR and its originating issue...
(There's an implied time budget here... ;-))

Taneb (approved):

JacquesCarette (changes requested):

JacquesCarette (approved):

Issue #2353, opened 2024-04-10
``_=ₛ_``  and  ``⊆ + reverse``  for lib-2.0


mechvel:
I suggest to add to standard library the following items.                       
                                                                                
To  Data.List.Relation.Binary.Subset.Setoid:                                    
```                                                                                
_=ₛ_ : Rel (List A) _            -- equality of sets                            
xs =ₛ ys =  xs ⊆ ys × ys ⊆ xs    -- represented by lists                        
                                                                                
_≠ₛ_ : Rel (List A) _                                                           
_≠ₛ_ xs =  ¬_ ∘ _=ₛ_ xs                                                         
```                                                                                                                                                             
To  Data.List.Relation.Binary.Subset.Setoid.Properties:                         
 ```                                                                               
=ₛ-refl : Reflexive _=ₛ_                                                        
=ₛ-sym : Symmetric _=ₛ_                                                         
=ₛ-trans : Transitive _=ₛ_                                                      
=ₛ-isEquivalence : IsEquivalence _=ₛ_                                           
=ₛ-reflexive = IsEquivalence.reflexive =ₛ-isEquivalence                         
                                                                                
setoidₛ : Setoid a _                                                            
                                                                                
≟ₛ : Decidable₂ _≈_ → Decidable₂ _=ₛ_                                           
                                                                                
reverse-preservesʳ-⊆ :  ∀ {xs} → reverse Preserves (xs ⊆_)                      
reverse-preservesˡ-⊆ :  ∀ {xs} → reverse Preserves (_⊆ xs)                      
                                                                                
reverse-xs-=ₛ-xs :  ∀ xs → reverse xs =ₛ xs                                     
 ```             
 This is because the above items are usable.

MatthewDaggitt:
We already have many different formalisations of this notion:

https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Relation/Binary/BagAndSetEquality.agda

https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Relation/Binary/Permutation/Setoid.agda

while I would welcome a proof that bidirectional subset inclusion implies these, I really don't think we want to be adding yet another relation.

Taneb:
If we _do_ add this new relation, I think we should name it the same as [`Relation.Unary._≐_`](http://agda.github.io/agda-stdlib/master/Relation.Unary.html#2545), which is defined in an analogous way on predicates

MatthewDaggitt:
[...]

PR #2378, opened 2024-05-02
`Data.List.Base.reverse` is self adjoint wrt `Data.List.Relation.Binary.Subset.Setoid._⊆_`

jamesmckinna:
This is a partial fix to #2353 but organised slightly differently, for consistency with existing uses of `reverse⁺` and `reverse⁻`, via a self-adjointness property of independent interest itself generalising the corresponding self-inverse property for `List.reverse`. 

Not done: extension to `Data.List.Relation.Binary.Sublist.Setoid._⊆_` (nor its `Heterogeneous` generalisation), which seems a *lot* more fiddly to achieve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -329,6 +329,12 @@ Additions to existing modules
   i*j≢0     : .{{_ : NonZero i}} .{{_ : NonZero j}} → NonZero (i * j)
   ```
 
+* In `Data.List.Membership.Setoid.Properties`:
+  ```agda
+  reverse⁺ : x ∈ xs → x ∈ reverse xs
+  reverse⁻ : x ∈ reverse xs → x ∈ xs
+  ```
+
 * In `Data.List.Properties`:
   ```agda
   length-catMaybes      : length (catMaybes xs) ≤ length xs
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):

jamesmckinna:
Possible additions? I didn't think to add these earlier, but they are in scope for this PR:
```agda
  reverse-η : ∀ {xs} → xs ⊆ reverse xs
  reverse-η = Membershipₚ.reverse⁺ S

  reverse-ε : ∀ {xs} → reverse xs ⊆ xs
  reverse-ε = Membershipₚ.reverse⁻ S
```
with the `-η`/`-ε` designations referring to the unit and counit of the self-adjunction...

JacquesCarette:
It almost feels like adding too many synonyms goes overboard? Doesn't pass the Fairbairn test. I'd be tempted to add these *as comments* in the code.

jamesmckinna:
> It almost feels like adding too many synonyms goes overboard? Doesn't pass the Fairbairn test. I'd be tempted to add these _as comments_ in the code.

Fair enough! The alternative would be to take those synonyms out of `Membershipₚ` (where they've been added as part of this PR from `Any`)  in favour of direct implementation here... but on balance your suggestion re:comment seems a good one.

Taneb (approved):
Is this also true for `Data.List.Relation.Binary.Sublist`? In which case we should add it there also, but that can be a separate PR

jamesmckinna:
> Is this also true for `Data.List.Relation.Binary.Sublist`? In which case we should add it there also, but that can be a separate PR

So that *is* a good question:
* yes, I think so
* but I found it 'not easy' to prove directly
* and so left it above as a downstream TODO...

That said, maybe it's *not* true? Either way, any attempt to prove it by going via a property of `reverseAcc` seems to get tangled up between any instances of sublist relations of the two arguments to that function...

andreasabel (approved):

Issue #2351, opened 2024-04-09
Add `Algebra.Construct.WreathProduct` and/or `Algebra.Construct.SemiDirectProduct`


jamesmckinna:
This fell originally under #2348 but I think should be factored out on its own.

Current issues:
* where is the (binary) product of `Setoid`s defined? (plus currying etc.: cartesian-closedness of `Setoid`?)
* where is the (pointwise) function space from a `Setoid` to an algebraic structure/bundle defined, and its properties established?
* which to add first: `Wreath` (my preferred target) or `SemiDirect`?
* my interest is in the wreath product of a `Monoid` with a `MonoidAction` (#2348 / #2350 ), but *many* kinds of variants exist according to how much structure is present. How/where to accommodate them all?

Taneb:
The binary product of `Setoid`s is in `Data.Product.Relation.Binary.Pointwise.NonDependent`: https://agda.github.io/agda-stdlib/master/Data.Product.Relation.Binary.Pointwise.NonDependent.html#7308. I don't know if we already have currying defined for setoids of the top of my head.

jamesmckinna:
Thanks @Taneb ! I'd just about managed to find that for myself while working on #2348 / #2350 but... not the first place I might have looked for it!

JacquesCarette:
A known, thorny issue: finding things in stdlib is very hard.

jamesmckinna:
Well, again, all good questions!
I think that as a *construction*, I should at least *also* need `Algebra.Construct.Pointwise` which does not at present exist, AFAICT. (And that is a construction... not a definition).

As for WreathProduct itself, I think that that is also a construction: given a monoid, a monoid action, then ... such and such setoid ... also admits a monoid structure...

JacquesCarette:
Good, I think this is converging to a design: indeed `WreathProduct` is a construction and so `Construct` is a great home for it. Yes, having a pointwise product construction would also be good. Whoever needs it first can build it.

jamesmckinna:
And yes: `Construct.Pointwise` (properly, should we say: `Power` and `Copower`?) should, in turn, be factored out as a distinct issue/PR!

PR #2381, opened 2024-05-05
Pointwise `Algebra`

jamesmckinna:
NB:
* missing structures/bundles (I got bored)
* `isEquivalence` might already be defined somewhere else, but defined in-place here for the time being

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,11 @@ Deprecated names
 New modules
 -----------
 
+* Pointwise algebraic structures and bundles:
+  ```
+  Algebra.Construct.Pointwise
+  ```
+
 * Raw bundles for module-like algebraic structures:
   ```
   Algebra.Module.Bundles.Raw
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,8 @@ Deprecated names
 New modules
 -----------
 
-* Pointwise algebraic structures and bundles:
+* Pointwise lifting of algebraic structures `IsX` and bundles `X` from
+  carrier set `C` to function space `A → C`:
   ```
   Algebra.Construct.Pointwise
   ```
--- a/src/Algebra/Construct/Pointwise.agda
+++ b/src/Algebra/Construct/Pointwise.agda
@@ -61,133 +61,96 @@ isEquivalence isEquivalence = record
 ------------------------------------------------------------------------
 -- Structures
 
-module _ (isMagmaC : IsMagma _≈_ _∙_) where
+isMagma : IsMagma _≈_ _∙_ → IsMagma (liftRel _≈_) (lift₂ _∙_)
+isMagma isMagma = record
+  { isEquivalence = isEquivalence M.isEquivalence
+  ; ∙-cong = λ g h x → M.∙-cong (g x) (h x)
+  }
+  where module M = IsMagma isMagma
 
-  private
-    module M = IsMagma isMagmaC
+isSemigroup : IsSemigroup _≈_ _∙_ → IsSemigroup (liftRel _≈_) (lift₂ _∙_)
+isSemigroup isSemigroup = record
+  { isMagma = isMagma M.isMagma
+  ; assoc = λ f g h x → M.assoc (f x) (g x) (h x)
+  }
+  where module M = IsSemigroup isSemigroup
 
-  isMagma : IsMagma (liftRel _≈_) (lift₂ _∙_)
-  isMagma = record
-    { isEquivalence = isEquivalence M.isEquivalence
-    ; ∙-cong = λ g h x → M.∙-cong (g x) (h x)
-    }
+isBand : IsBand _≈_ _∙_ → IsBand (liftRel _≈_) (lift₂ _∙_)
+isBand isBand = record
+  { isSemigroup = isSemigroup M.isSemigroup
+  ; idem = λ f x → M.idem (f x)
+  }
+  where module M = IsBand isBand
 
-module _ (isSemigroupC : IsSemigroup _≈_ _∙_) where
+isCommutativeSemigroup : IsCommutativeSemigroup _≈_ _∙_ →
+                         IsCommutativeSemigroup (liftRel _≈_) (lift₂ _∙_)
+isCommutativeSemigroup isCommutativeSemigroup = record
+  { isSemigroup = isSemigroup M.isSemigroup
+  ; comm = λ f g x → M.comm (f x) (g x)
+  }
+  where module M = IsCommutativeSemigroup isCommutativeSemigroup
 
-  private
-    module M = IsSemigroup isSemigroupC
+isMonoid : IsMonoid _≈_ _∙_ ε → IsMonoid (liftRel _≈_) (lift₂ _∙_) (lift₀ ε)
+isMonoid isMonoid = record
+  { isSemigroup = isSemigroup M.isSemigroup
+  ; identity = (λ f x → M.identityˡ (f x)) , λ f x → M.identityʳ (f x)
+  }
+  where module M = IsMonoid isMonoid
 
-  isSemigroup : IsSemigroup (liftRel _≈_) (lift₂ _∙_)
-  isSemigroup = record
-    { isMagma = isMagma M.isMagma
-    ; assoc = λ f g h x → M.assoc (f x) (g x) (h x)
-    }
+isCommutativeMonoid : IsCommutativeMonoid _≈_ _∙_ ε →
+                      IsCommutativeMonoid (liftRel _≈_) (lift₂ _∙_) (lift₀ ε)
+isCommutativeMonoid isCommutativeMonoid = record
+  { isMonoid = isMonoid M.isMonoid
+  ; comm = λ f g x → M.comm (f x) (g x)
+  }
+  where module M = IsCommutativeMonoid isCommutativeMonoid
 
-module _ (isBandC : IsBand _≈_ _∙_) where
+isGroup : IsGroup _≈_ _∙_ ε _⁻¹ →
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -78,6 +78,11 @@ Deprecated names
 New modules
 -----------
 
+* Pointwise algebraic structures and bundles:
+  ```
+  Algebra.Construct.Pointwise
```
JacquesCarette, line 83: The eternal naming question: I had no idea what this was until I read further down. `PointwiseLift` ? `FunctionsOver` (that's bad, I know, but something with 'Function' in the name is not unreasonable)?

src/Algebra/Construct/Pointwise.agda
```diff
@@ -0,0 +1,224 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- For each `IsX` algebraic structure `S`, lift the structure to the
+-- 'pointwise' function space `A → S`: categorically, this is the
```
JacquesCarette, line 5: (Definitely a comment, not a 'todo'): I wonder if it wouldn't be just as easy to do the n-ary version, i.e. lift to n arguments on the left. Without going through currying/uncurrying, but via the `Nary` facilities already in the library.

src/Algebra/Construct/Pointwise.agda
```diff
@@ -0,0 +1,224 @@
...
+
+module _ (isEquivalenceC : IsEquivalence _≈_) where
+
+  open IsEquivalence isEquivalenceC
+
+  isEquivalence : IsEquivalence (liftRel _≈_)
+  isEquivalence = record
+    { refl = λ {f} x → refl {f x}
+    ; sym = λ f≈g x → sym (f≈g x)
+    ; trans = λ f≈g g≈h x → trans (f≈g x) (g≈h x)
```
JacquesCarette, line 60: Have to wonder if `lift₂ trans` would work here.

src/Algebra/Construct/Pointwise.agda
```diff
@@ -0,0 +1,224 @@
...
+
+module _ (isMagmaC : IsMagma _≈_ _∙_) where
+
+  private
+    module M = IsMagma isMagmaC
+
+  isMagma : IsMagma (liftRel _≈_) (lift₂ _∙_)
+  isMagma = record
+    { isEquivalence = isEquivalence M.isEquivalence
+    ; ∙-cong = λ g h x → M.∙-cong (g x) (h x)
```
JacquesCarette, line 74: `lift₂ M.∙-cong` ?

JacquesCarette (changes requested):
It's more "consider these" than 'request changes', really.

jamesmckinna:
Great feedback, thanks!

Regarding naming, I agree that perhaps `Pointwise` is suboptimal, apart from 'ordinary' mathematical language having 'pointwise lifting' as an informal-but-always-disambiguable term-of-art...? Besides which, we had previously discussed 'pointwise' constructions under #2351 ...

I had even considered calling this `Algebra.Construct.Power{Object}`, but thought that the categorical notion might cast more heat than light... but definitely would change if you thought it fitted better?

As for proofs, one thing where I was too lazy was in formulating suitable lifting properties for the concepts in `Alfebra.Definitions`, which would make it easier to reuse in the sequel (and fill in the missing structures/bundles...). Your suggestions might also further streamline those!

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Oooh... interesting!? (Always more to have to get my head around... ;-))

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Ah!?
Is this already defined under `Function.Equality`?

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Great idea, however:
```agda
Cannot instantiate the metavariable _x_156 to solution x x₁
since it contains the variable x₁
which is not in scope of the metavariable
when checking that the expression lift₂ M.∙-cong has type
liftRel _≈_ x y →
liftRel _≈_ u v → liftRel _≈_ (lift₂ _∙_ x u) (lift₂ _∙_ y v)
```

JacquesCarette: Right - I think you'd need a dependent version of `lift₂` for that to work.

src/Algebra/Construct/Pointwise.agda
JacquesCarette: I was more thinking along the lines of the funky combinators in [Function.Base](https://agda.github.io/agda-stdlib/master/Function.Base.html) though I notice that they are currently non-dependent.

Though, as you point out, this is defined in the library already - now in `Function.Relation.Binary.Setoid.Equality`.

JacquesCarette:
`PowerObject` is worse (i.e. even fewer people will find this.)  Somehow (from #2351 ) `Copower` seems less bad.

Current favourite remains `PointwiseLift`, though it remains a 'placeholder' in my mind for something even better.

Taneb:
I think this construction is worthwhile to add to the library in some form. Further review to follow at some point.

src/Algebra/Construct/Pointwise.agda
```diff
@@ -0,0 +1,224 @@
...
+
+  lift₂ : Op₂ C → Op₂ (A → C)
+  lift₂ _∙_ g h x = (g x) ∙ (h x)
+
+  liftRel : Rel C ℓ → Rel (A → C) (a ⊔ ℓ)
+  liftRel _≈_ g h = ∀ x → (g x) ≈ (h x)
+
+
+------------------------------------------------------------------------
+-- Setoid structure: here rather than elsewhere? (could be imported?)
```
Taneb, line 50: The setoid structure should definitely be somewhere under `Relation.Binary`. I think it's related to `Relation.Binary.Indexed` in some way, although that is "more" dependent.

src/Algebra/Construct/Pointwise.agda
```diff
@@ -0,0 +1,224 @@
...
+  isEquivalence = record
+    { refl = λ {f} x → refl {f x}
+    ; sym = λ f≈g x → sym (f≈g x)
+    ; trans = λ f≈g g≈h x → trans (f≈g x) (g≈h x)
+    }
+
+------------------------------------------------------------------------
+-- Structures
+
+module _ (isMagmaC : IsMagma _≈_ _∙_) where
```
Taneb, line 66: Having anonymous modules for one definition feels overkill to me here.

Taneb:
What was the reasoning for the algebraic structures you're defining? Are you planning to define more?

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Re: `Function.Relation.Binary.Setoid.Equality`
Another case of implicit (there) vs. explicit (here) in the quantification over `x` in the definition of `_≈_` (there) and `liftRel _≈_` (here)... sigh. 
Plus, in order to use it I need to artificially impose the `Propositional` setoid structure on `A`... oh dear, and the definition of `isEquivalence` is in-lined there rather than lifted out... so perhaps some (downstream) refactoring might be in order there?
Not sure the additional complexity is worth it?

src/Algebra/Construct/Pointwise.agda
jamesmckinna: @JacquesCarette has it at `Function.Relation.Binary.Setoid.Equality`, but as above, some gymnastics required to reconcile the approaches.

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Yes, it worked out fine for the `Bundle`s, but I'd wanted the `private module ...` trick, and didn't easily see how to do that simply in a function block, but perhaps that's where `where` would be my friend. Will refactor!

jamesmckinna:
> `PowerObject` is worse (i.e. even fewer people will find this.) Somehow (from #2351 ) `Copower` seems less bad.

But `Copower` is `Σ` while `Power is `Π`, surely?

> Current favourite remains `PointwiseLift`, though it remains a 'placeholder' in my mind for something even better.

Sticking with `Pointwise` for now, but have 'improved' the `CHANGELOG` description for the time being...

CHANGELOG.md
jamesmckinna: Revised description...

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Have concluded in the interests of time/brain that this be a downstream consideration.

JacquesCarette:
About `Copower`: yes, of course. D'oh.

JacquesCarette (approved):
The only thing I'm not liking is the name - shouldn't stop progress for just that.

src/Algebra/Construct/Pointwise.agda
jamesmckinna: Or should I make the quantification implicit here? Cf. #2382

JacquesCarette: I certainly would experiment with implicit here.

jamesmckinna:
Final commit: `liftRel` now follows [`Function.Relation.Binary.Setoid.Equality`](https://agda.github.io/agda-stdlib/master/Function.Relation.Binary.Setoid.Equality.html) in being implicitly quantified, but ahead of possible merge of #2382 it doesn't make sense (to me at least) to try to import that module directly for reuse...?

JacquesCarette:
You can have one PR depend on another. There is of course a risk in that scheme.

MatthewDaggitt:
@Taneb are you willing to approve?

Taneb (approved):

PR #2382, opened 2024-05-06
Refactor `Function.Relation.Binary.Setoid.Equality` to lift out `isEquivalence`

jamesmckinna:
NB.
* no `CHANGELOG`, given that the module was only introduced in v2.1
* possible modification: should the quantification in 
```agda
infix 4 _≈_
_≈_ : (f g : Func From To) → Set (a₁ ⊔ b₂)
f ≈ g = {x : From.Carrier} → f ⟨$⟩ x To.≈ g ⟨$⟩ x
```
be implicit, as now, or explicit?


JacquesCarette (approved):

jamesmckinna:
@JacquesCarette implicit or explicit?

JacquesCarette:
Implicit. The explicit version caused a lot of noise in `agda-categories`.

Taneb (approved):
Thanks for doing this! :)

Issue #2287, opened 2024-02-10
Mechanize the Algebra hierarchy


Taneb:
Many of the modules under `Algebra` (e.g. `Algebra.Structures`, `Algebra.Bundles`, `Algebra.Morphism.Structures`...) are highly structured and repetitive. This makes it a lot of work to add something new, and _very_ easy to miss things. We should come up with a way to mechanize this and generate this portion of the library from a more concise form somehow.

## Considerations
* Concision: we want to be able to generate large amounts of the library's functionality from a small specification
* Completeness: we want to take care that anything we _want_ in these modules, we can still have in these modules.
* Ease of understanding: the documentation of this library, such as it is, is in the form of source code listings. We want users of the library to still be able to figure out what we provide
* Ease of installation: the library is generally installed by copying it verbatim somewhere onto the system. We don't have a way to run, say, a Haskell program at installation time.
* Extensibility: if we want to add a new form of algebraic structure, we should be able to do so easily. If we want to add a new expression of algebraic structures (e.g. morphism bundles, or finite algebraic structures), we should be able to do that easily also.

## Current state of affairs
We have three modules of "structures", containing records of laws for various algebraic structures, parameterized by their equivalence relation and operations. We have three modules of "bundles" containing records bundling the equivalence relation, operations, and the law structure, and three modules of "raw bundles" which omit the law structure. We have three modules of morphism structure, each of which is parameterized on the raw bundle. We have many modules of various properties of structures and morphisms, which may be harder to mechanize.

jamesmckinna:
Two comments, not (entirely; or even, at all ;-)) tongue-in-cheek:
* milestone `hypothetical-rewrite`!?
* @JacquesCarette !?

JacquesCarette:
I obviously agree emphatically with this. And, to a certain extent, this is roughly the work that Yasmine did for her Ph.D. thesis. If you couple it with the produce-documented-human-readable-code that Drasil can do, in theory I have the needed pieces in hand to do this.

In practice, these pieces are not (yet) integrated.

I would propose that we first implement this as a pre-processor. The published agda-stdlib would contain, in part, output from that pre-processor. This could be done immediately.

Then we could abstract out the features that we need to make it into a language feature.

Pinging @TOTBWF as I think he's quite interested in things like this. (Ok, what's described here is probably "baby steps" for what @TOTBWF really is interested in, but it does seem to be in a compatible direction.)

Taneb:
[...]

PR #2383, opened 2024-05-08
Add bundled homomorphisms

### LINKED ISSUE #1960: Bundles for `Algebra.*Homomorphism`, and their algebra: `Hom`-'sets' for `Algebra`

jamesmckinna:
Cherry-picked and enhanced from #1962 . Fixes #1960 .

TODO:
- [x] the `Algebra.Morphism.Construct.{Identity|Composition}` material
- [x] `RingWithoutOne` doesn't appear to export its own `Raw` substructure, so that should be added? ditto. `KleeneAlgebra`
- [x] homomorphisms for all the existing `Structures`/`Bundles`

NB
* doesn't address the `Setoid` structure on `Hom`s at all... (brain-fade; sigh)
* for bundles which export two sub-bundles, usual 'diamond' problem about re-opening those publicly for export, so someone with a good eye for such things,... please nitpick those details!
* UPDATED: `Semiring` and `Ring` homomorphisms should export a `SuccessorSetHomomorphism` structure/bundle
*only* after going back to #1363 through the medium of (the initial) `SuccessorSet` and its consequences for `Algebra.Structures` having `(Is)SuccessorSet` fields...


Taneb:
Before I look at this at all I just want to say thanks for tackling it! I've been avoiding it with the hope that we can get a solution for #2287, but that looks a long way off.

jamesmckinna:
@Taneb Indeed! This kind of boilerplate-bashing is quite fatiguing, until you run up against the 'diamond' re-export problem, at which point my powers fail me in different ways, trying to work out what should get re-exported from where... plus discovering that (still!) not everything is present which perhaps ought to be... ;-)

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,178 @@
...
+  *-monoidHomomorphism = record { isMonoidHomomorphism = *-isMonoidHomomorphism }
+
+  open MonoidHomomorphism *-monoidHomomorphism public
+
+------------------------------------------------------------------------
+-- Morphisms between RingWithoutOnes
+------------------------------------------------------------------------
+{-
+
+Problem: bundle RingWithoutOne doesn't re-export its underlying RawRingWithoutOne!
```
JacquesCarette, line 133: Seems like a bug that should be fixed within this PR as a way forward?

JacquesCarette (commented):
This all seems quite reasonable. I am doing this as a comment because it seems incomplete - but I have no specific improvement that I'd like to see in the code that is here.

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Now added to `Algebra.Bundles.RingWithoutOne`.

jamesmckinna: But still problems: see latest commits...
```agda
Cannot resolve overloaded projection Carrier because
it is not applied to a visible argument
  candidates in scope:
- Carrier : Set a
- Carrier : Set a
Algebra.Morphism.Bundles.A.Carrier has type _226
```
if declaring
```agda
field
    ⟦_⟧ : A.Carrier → B.Carrier
```
???

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,176 @@
...
+-- Morphisms between RingWithoutOnes
+------------------------------------------------------------------------
+
+record RingWithoutOneHomomorphism (A : RingWithoutOne a ℓa) (B : RingWithoutOne b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
+  private
+    module A = RingWithoutOne A
+    module B = RingWithoutOne B
+
+  field
+    ⟦_⟧ : _ → _ -- A.Carrier → B.Carrier causes an disambiguation error!?
```
jamesmckinna, line 138:
```agda
Cannot resolve overloaded projection Carrier because
it is not applied to a visible argument
  candidates in scope:
- Carrier : Set a
- Carrier : Set a
Algebra.Morphism.Bundles.A.Carrier has type _226
```
hoping this is simply a stupid error on my part, but I can't see what...

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,176 @@
...
+-- Morphisms between Rings
+------------------------------------------------------------------------
+
+record RingHomomorphism (A : Ring a ℓa) (B : Ring b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
+  private
+    module A = Ring A
+    module B = Ring B
+
+  field
+    ⟦_⟧ : A.Carrier → B.Carrier
```
jamesmckinna, line 164: Yet this works fine!?

jamesmckinna:
@JacquesCarette @Taneb I have now added the 'missing' parts of the PR (as in the revised opening preamble above), ands (hopefully) now stopping there...

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Ditto. `rawKleeneAlgebra` added to `Algebra.Bundles.KleeneAlgebra`.

src/Algebra/Morphism/Construct/Composition.agda
```diff
@@ -423,3 +424,168 @@ module _ {K₁ : RawKleeneAlgebra a ℓ₁}
     ; surjective                  = Func.surjective (_≈_ K₁) (_≈_ K₂) (_≈_ K₃) F.surjective G.surjective
     } where module F = IsKleeneAlgebraIsomorphism f-iso; module G = IsKleeneAlgebraIsomorphism g-iso
 
+-----------------------------------------------------------------------
+-- Bundled morphisms between algebras
+------------------------------------------------------------------------
+
+-- Magma
+
+module _ {M₁ : Magma a ℓ₁}
```
Taneb, line 433: I think I'd avoid these anonymous modules in the `.Construct.*` modules, they're only ever going to have a single thing in them

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,217 @@
...
+  +-monoidHomomorphism = record { isMonoidHomomorphism = +-isMonoidHomomorphism }
+
+  *-magmaHomomorphism : MagmaHomomorphism A.*-magma B.*-magma
+  *-magmaHomomorphism = record { isMagmaHomomorphism = *-isMagmaHomomorphism }
+
+------------------------------------------------------------------------
+-- Morphisms between Semirings
+------------------------------------------------------------------------
+
+record SemiringHomomorphism (A : Semiring a ℓa) (B : Semiring b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
```
Taneb, line 105: This seems to be missing a definition of `+-monoidHomomorphism` (which should be reexported downstream)

Actually, shouldn't `SemiringHomomorphism`s and `NearSemiringHomomorphism`s relate?

Taneb (changes requested):
The following don't need changes now but we should make issues for them so we don't lose track:

* Bundles for monomorphisms and isomorphisms (and epimorphisms if we add those)
* Bundles  for lattice-like and module-like morphisms

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Good catch!
The NB above noted that I'm not always good at spotting missing exports...

Taneb: That's what prompted me to look ;)

src/Algebra/Morphism/Construct/Composition.agda
jamesmckinna: DONE.

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,223 @@
...
+  semiringHomomorphism = record { isSemiringHomomorphism = isSemiringHomomorphism }
+
+  open SemiringHomomorphism semiringHomomorphism public
+    hiding (*-isMagmaHomomorphism; *-isMonoidHomomorphism)
+
+------------------------------------------------------------------------
+-- Morphisms between RingWithoutOnes
+------------------------------------------------------------------------
+
+record RingWithoutOneHomomorphism (A : RingWithoutOne a ℓa) (B : RingWithoutOne b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
```
Taneb, line 151: This should expose
* `nearSemiringHomomorphism`
* `+-magmaHomomorphism`
* `+-monoidHomomorphism`

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,223 @@
...
+  +-groupHomomorphism = record { isGroupHomomorphism = +-isGroupHomomorphism }
+
+  *-magmaHomomorphism : MagmaHomomorphism A.*-magma B.*-magma
+  *-magmaHomomorphism = record { isMagmaHomomorphism = *-isMagmaHomomorphism }
+
+------------------------------------------------------------------------
+-- Morphisms between Rings
+------------------------------------------------------------------------
+
+record RingHomomorphism (A : Ring a ℓa) (B : Ring b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
```
Taneb, line 172: This should expose
* `semiringHomomorphism`
* `ringWithoutOneHomomorphism`
* `nearSemiringHomomorphism`
* `+-magma-homomorphism`
* `+-monoid-homomorphsim`
* `*-magma-homomorphism`

Taneb (changes requested):

src/Algebra/Morphism/Bundles.agda
jamesmckinna: This turned out to be quite a bit more far-reaching than I first suspected... see incoming commits.

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Taking a deep breath after `RingWithoutOneHomomorphism`...

jamesmckinna: ... and `*-monoidHomomorphism`?

Taneb: If it's not there already (I think I convinced myself that it was earlier)

jamesmckinna: OK, I think I managed to do all of those... by filling in some more diamonds... cf. #1688 and friends #2254 etc. which might want to be revisited after these additions...?

jamesmckinna:
So I think that I now have added all the relevant (requested!) sub-{structure|bundle}s, but definitely needs another pass to check, along perhaps with a downstream issue/PR to fill in yet more gaps (including in the `Raw` sub-hierarchy, cf. #2255 ).

JacquesCarette (approved):

jamesmckinna:
Thanks @JacquesCarette ! (modulo the need for manual effort in the absence of a Drasil-based solution...)

Taneb (approved):

src/Algebra/Morphism/Bundles.agda
```diff
@@ -0,0 +1,241 @@
...
+private
+  variable
+    ℓ a ℓa b ℓb : Level
+
+
+------------------------------------------------------------------------
+-- Morphisms between Magmas
+------------------------------------------------------------------------
+
+record MagmaHomomorphism (A : Magma a ℓa) (B : Magma b ℓb) : Set (a ⊔ b ⊔ ℓa ⊔ ℓb) where
```
MatthewDaggitt, line 26: Could you add a brief comment documenting why you decided to use `Bundles` here instead of the `RawBundles` used by the morphism structures? Having thought about it for all of a minute, it's not terribly obvious to me why we should make the switch. In particular, these morphisms are often used to transfer properties from one structure, e.g. a magma, to prove that another structure is also a magma. Assuming that they're both a magma to starts with, prevents you from using these morphisms for that purpose.

MatthewDaggitt (changes requested):

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Oh-ho-ho! Good catch... will have to go back to the drawing board and scratch my head for a bit. 

My first (worst?) thought is that such a use case could be handled via the already-existing `IsMagmaHomomorphism` infrastructure... but I have *not* tested that conjecture (related to @Taneb 's on `Monomorphism`s?)... while the specific PR (#1962 ) which gave rise to (the use case giving rise to) this PR *does* already have the two `Magma` bundles at hand... and I'd never reconsidered that state of affairs...

MatthewDaggitt: So yes, it can (and is!) handled by `IsMagmaHomomorphism` and co. The question is what does making them `Bundles` instead of `RawBundles` actually gains us? I can only see downsides.

jamesmckinna: I think I'm convinced: but there's a knock-on consequence... : we can no longer define a `SetoidHomomorphism` manifest field at the bottom of the hierarchy... and probably `Relation.Morphism.Bundles` should be refactored to reflect (the eventual introduction of) `RawSetoid` as index for `SetoidHomomorphism`... ;-) cf. your #1464 / commit 981602398f183ca901a32208673e2fcf1f20caac

jamesmckinna: More knock-on:
* For `Algebra.Morphism.Construct.Identity`, we need `refl` for `IsXHomomorphism`, so do we parametrise by `xHomomorphism` by the full `X` bundle, or the `RawX` bundle, with an explicit `refl` witness...?
* Ditto. with `trans` for  `Algebra.Morphism.Construct.Composition`...

Advice welcome!

jamesmckinna: Latest commits: 
* `RawX` bundle for the *definition*
* `X` bundle for the `Construct`ions

jamesmckinna:
Currently badged as v2.2, but could merge for v2.1?

JacquesCarette (approved):

jamesmckinna:
Have now updated the `CHANGELOG` for v2.2, but there is still the unresolved discussion above about `Identity` and `Composition` apparently needing to take full bundles in order to get the correct `setoid` properties `refl` and `trans`... or should these be supplied explicitly?

jamesmckinna:
Note to self: check the `CHANGELOG` commits between v2.1 and v2.2... and `SuccessorSet` morphisms?

MatthewDaggitt (approved):

MatthewDaggitt:
I think it's fine that they take the full bundles for now.

PR #2386, opened 2024-05-14
Algebra fixity

JacquesCarette:
One more step towards #207 .   Also add to style guide.

### DIFF
--- a/doc/style-guide.md
+++ b/doc/style-guide.md
@@ -525,6 +525,54 @@ word within a compound word is capitalized except for the first word.
 * Any exceptions to these conventions should be flagged on the GitHub
   `agda-stdlib` issue tracker in the usual way.
 
+#### Fixity
+
+All functions and operators that are not purely prefix (typically
+anything that has a `_` in its name) should have an explicit fixity
+declared for it. The guidelines for these are as follows:
+
+General operations and relations:
+* binary relations of all kinds are `infix 4`
+
+* multiplication-like: `infixl 7 _*_`
+
+* addition-like  `infixl 6 _+_`
+
+* negation-like  `infix  8 ¬_`
+
+* and-like  `infixr 7 _∧_`
+
+* or-like  `infixr 6 _∨_`
+
+* post-fix inverse  `infix  8 _⁻¹`
+
+* bind `infixl 1 _>>=_`
+
+* list concat-like `infixr 5 _∷_`
+
+* ternary reasoning `infix 1 _⊢_≈_`
+
+* composition `infixr 9 _∘_`
+
+* application `infixr -1 _$_ _$!_`
+
+* combinatorics `infixl 6.5 _P_ _P′_ _C_ _C′_`
+
+* pair `infixr 4 _,_`
+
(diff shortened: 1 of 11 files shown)


Taneb (approved):

doc/style-guide.md
```diff
@@ -525,6 +525,62 @@ word within a compound word is capitalized except for the first word.
...
+
+* and-like  `infixr 7 _∧_`
+
+* or-like  `infixr 6 _∨_`
+
+* post-fix inverse  `infix  8 _⁻¹`
+
+* bind `infixl 1 _>>=_`
+
+* list concat-like `infixr 5 _∷_`
```
jamesmckinna, line 556: NB `Data.List.Base._++_` follows this scheme, yet it is also a `Monoid`-like addition operator (L544 above suggests it should therefore be `infixl 6 _++_`)... how should we resolve such inconsistencies?

JacquesCarette: So I think that 'addition-like' means much closer to 'arithmetic' than something more general, i.e. that addition-like things could happen inside concat-like things without too much need for parens.

Or it could be that this is all a hack that happens to work by serendipitous accident rather than by conscious design!

doc/style-guide.md
```diff
@@ -525,6 +525,62 @@ word within a compound word is capitalized except for the first word.
...
+
+* unary prefix relations `infix 4 ε∣_`
+
+* unary postfix relations `infixr 8 _∣0`
+
+* multiplication-like: `infixl 7 _*_`
+
+* addition-like  `infixl 6 _+_`
+
+* negation-like  `infix  8 ¬_`
```
jamesmckinna, line 546: Aaaargh!
* `Relation.Nullary.Negation.Core` declares `infix 3 ¬_`, and is (definitely!?) a special case
* `Data.Integer.Base.-_` etc. otherwise follow the pattern
```suggestion
* unary prefix negation-like  `infix  8 -_` (NB. *logical* negation `infix 3 ¬_` is a special case)
```

jamesmckinna: And ... what about an additional/separate entry for *binary* `_-_` (as `infixl 6 `_-_`) as in `Algebra.Structures.IsGroup` and `Data.Integer.Base`?

jamesmckinna (changes requested):

doc/style-guide.md
JacquesCarette: Good points. I'm using slight variants of the ideas, but basically 'yes' to both.

jamesmckinna (approved):

src/Algebra/Bundles.agda
```diff
@@ -917,6 +917,7 @@ record NonAssociativeRing c ℓ : Set (suc (c ⊔ ℓ)) where
     using () renaming (magma to *-magma; identity to *-identity)
 
 record Nearring c ℓ : Set (suc (c ⊔ ℓ)) where
+  infix  8 -_
```
jamesmckinna, line 920: Good catch!

Issue #2369, opened 2024-04-17
To let or not to let


JacquesCarette:
Right now, `let` in Agda generates substitutions (as confirmed by Andreas [here](https://github.com/agda/agda/issues/5497) amongst other places). The [attempted fix](https://github.com/agda/agda/issues/3094) for this has stalled even though it is an [extremely old](https://github.com/agda/agda/issues/514) issue.

The advice is to favour `where` over `let`.

The question is: do we program stdlib to current Agda or some optimistic future Agda that will have a better implementation of `let`?

Given just how long this has been a wart, I'm of the opinion that being optimistic would be ill-placed (even though I'm actually normally fairly optimistic!)

gallais:
FTR, these cannot be turned into `where` definitions because these get lifted to the
toplevel and that is not done in a modality-respecting way. In other words: turning
these into `where` definitions leads to errors about irrelevant values being used in
a relevant context.

https://github.com/gallais/agda-stdlib/blob/system-random/src/System/Random.agda#L119-L133

JacquesCarette:
Good point: indeed there are times when we have no choice, because the implementation scheme for `where` is also sub-optimal.

jamesmckinna:
Time to revisit after 2.7.0's introduction of `using`?

JacquesCarette:
That would be great - but we'd need to know how `using` is implemented to see if it's actually a win. Perhaps @jespercockx can enlighten us?

jespercockx:
I was not the one who implemented it, but I believe `using` is desugared to an auxiliary function, so I'm afraid it's not really better performance-wise than a `where` declaration.

I started working on implementing proper `let`-bindings for Agda's internal syntax during the last Agda meeting (see https://github.com/agda/agda/pull/7253) but it's a lot more complicated that I had anticipated so don't expect anything from it soon.

PR #2389, opened 2024-05-17
More list properties about `catMaybes/mapMaybe`

omelkonian:
- Follow-up to PR #2361
- Ported from my [formal-prelude](https://github.com/omelkonian/formal-prelude/tree/c10fe94876a7378088f2e4cf68d9b18858dcc256)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -343,47 +343,55 @@ Additions to existing modules
 
 * In `Data.List.Properties`:
   ```agda
-  length-catMaybes      : length (catMaybes xs) ≤ length xs
-  applyUpTo-∷ʳ          : applyUpTo f n ∷ʳ f n ≡ applyUpTo f (suc n)
-  applyDownFrom-∷ʳ      : applyDownFrom (f ∘ suc) n ∷ʳ f 0 ≡ applyDownFrom f (suc n)
-  upTo-∷ʳ               : upTo n ∷ʳ n ≡ upTo (suc n)
-  downFrom-∷ʳ           : applyDownFrom suc n ∷ʳ 0 ≡ downFrom (suc n)
-  reverse-selfInverse   : SelfInverse {A = List A} _≡_ reverse
-  reverse-applyUpTo     : reverse (applyUpTo f n) ≡ applyDownFrom f n
-  reverse-upTo          : reverse (upTo n) ≡ downFrom n
-  reverse-applyDownFrom : reverse (applyDownFrom f n) ≡ applyUpTo f n
-  reverse-downFrom      : reverse (downFrom n) ≡ upTo n
-  mapMaybe-map          : mapMaybe f ∘ map g ≗ mapMaybe (f ∘ g)
-  map-mapMaybe          : map g ∘ mapMaybe f ≗ mapMaybe (Maybe.map g ∘ f)
-  align-map             : align (map f xs) (map g ys) ≡ map (map f g) (align xs ys)
-  zip-map               : zip (map f xs) (map g ys) ≡ map (map f g) (zip xs ys)
-  unzipWith-map         : unzipWith f ∘ map g ≗ unzipWith (f ∘ g)
-  map-unzipWith         : map (map g) (map h) ∘ unzipWith f ≗ unzipWith (map g h ∘ f)
-  unzip-map             : unzip ∘ map (map f g) ≗ map (map f) (map g) ∘ unzip
-  splitAt-map           : splitAt n ∘ map f ≗ map (map f) (map f) ∘ splitAt n
-  uncons-map            : uncons ∘ map f ≗ map (map f (map f)) ∘ uncons
-  last-map              : last ∘ map f ≗ map f ∘ last
-  tail-map              : tail ∘ map f ≗ map (map f) ∘ tail
-  mapMaybe-cong         : f ≗ g → mapMaybe f ≗ mapMaybe g
-  zipWith-cong          : (∀ a b → f a b ≡ g a b) → ∀ as → zipWith f as ≗ zipWith g as
-  unzipWith-cong        : f ≗ g → unzipWith f ≗ unzipWith g
-  foldl-cong            : (∀ x y → f x y ≡ g x y) → ∀ x → foldl f x ≗ foldl g x
-  alignWith-flip        : alignWith f xs ys ≡ alignWith (f ∘ swap) ys xs
-  alignWith-comm        : f ∘ swap ≗ f → alignWith f xs ys ≡ alignWith f ys xs
-  align-flip            : align xs ys ≡ map swap (align ys xs)
-  zip-flip              : zip xs ys ≡ map swap (zip ys xs)
-  unzipWith-swap        : unzipWith (swap ∘ f) ≗ swap ∘ unzipWith f
-  unzip-swap            : unzip ∘ map swap ≗ swap ∘ unzip
-  take-take             : take n (take m xs) ≡ take (n ⊓ m) xs
-  take-drop             : take n (drop m xs) ≡ drop m (take (m + n) xs)
-  zip-unzip             : uncurry′ zip ∘ unzip ≗ id
-  unzipWith-zipWith     : f ∘ uncurry′ g ≗ id → length xs ≡ length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
-  unzip-zip             : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
-  mapMaybe-++           : mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
-  unzipWith-++          : unzipWith f (xs ++ ys) ≡ zip _++_ _++_ (unzipWith f xs) (unzipWith f ys)
-  catMaybes-concatMap   : catMaybes ≗ concatMap fromMaybe
-  catMaybes-++          : catMaybes (xs ++ ys) ≡ catMaybes xs ++ catMaybes ys
-  map-catMaybes         : map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
+  length-catMaybes       : length (catMaybes xs) ≤ length xs
+  applyUpTo-∷ʳ           : applyUpTo f n ∷ʳ f n ≡ applyUpTo f (suc n)
+  applyDownFrom-∷ʳ       : applyDownFrom (f ∘ suc) n ∷ʳ f 0 ≡ applyDownFrom f (suc n)
+  upTo-∷ʳ                : upTo n ∷ʳ n ≡ upTo (suc n)
+  downFrom-∷ʳ            : applyDownFrom suc n ∷ʳ 0 ≡ downFrom (suc n)
+  reverse-selfInverse    : SelfInverse {A = List A} _≡_ reverse
+  reverse-applyUpTo      : reverse (applyUpTo f n) ≡ applyDownFrom f n
+  reverse-upTo           : reverse (upTo n) ≡ downFrom n
+  reverse-applyDownFrom  : reverse (applyDownFrom f n) ≡ applyUpTo f n
+  reverse-downFrom       : reverse (downFrom n) ≡ upTo n
+  mapMaybe-map           : mapMaybe f ∘ map g ≗ mapMaybe (f ∘ g)
+  map-mapMaybe           : map g ∘ mapMaybe f ≗ mapMaybe (Maybe.map g ∘ f)
+  align-map              : align (map f xs) (map g ys) ≡ map (map f g) (align xs ys)
+  zip-map                : zip (map f xs) (map g ys) ≡ map (map f g) (zip xs ys)
+  unzipWith-map          : unzipWith f ∘ map g ≗ unzipWith (f ∘ g)
+  map-unzipWith          : map (map g) (map h) ∘ unzipWith f ≗ unzipWith (map g h ∘ f)
+  unzip-map              : unzip ∘ map (map f g) ≗ map (map f) (map g) ∘ unzip
+  splitAt-map            : splitAt n ∘ map f ≗ map (map f) (map f) ∘ splitAt n
+  uncons-map             : uncons ∘ map f ≗ map (map f (map f)) ∘ uncons
+  last-map               : last ∘ map f ≗ map f ∘ last
+  tail-map               : tail ∘ map f ≗ map (map f) ∘ tail
+  mapMaybe-cong          : f ≗ g → mapMaybe f ≗ mapMaybe g
+  zipWith-cong           : (∀ a b → f a b ≡ g a b) → ∀ as → zipWith f as ≗ zipWith g as
+  unzipWith-cong         : f ≗ g → unzipWith f ≗ unzipWith g
+  foldl-cong             : (∀ x y → f x y ≡ g x y) → ∀ x → foldl f x ≗ foldl g x
+  alignWith-flip         : alignWith f xs ys ≡ alignWith (f ∘ swap) ys xs
+  alignWith-comm         : f ∘ swap ≗ f → alignWith f xs ys ≡ alignWith f ys xs
+  align-flip             : align xs ys ≡ map swap (align ys xs)
+  zip-flip               : zip xs ys ≡ map swap (zip ys xs)
+  unzipWith-swap         : unzipWith (swap ∘ f) ≗ swap ∘ unzipWith f
+  unzip-swap             : unzip ∘ map swap ≗ swap ∘ unzip
+  take-take              : take n (take m xs) ≡ take (n ⊓ m) xs
+  take-drop              : take n (drop m xs) ≡ drop m (take (m + n) xs)
+  zip-unzip              : uncurry′ zip ∘ unzip ≗ id
+  unzipWith-zipWith      : f ∘ uncurry′ g ≗ id →
+                           length xs ≡ length ys →
+                           unzipWith f (zipWith g xs ys) ≡ (xs , ys)
+  unzip-zip              : length xs ≡ length ys → unzip (zip xs ys) ≡ (xs , ys)
+  mapMaybe-++            : mapMaybe f (xs ++ ys) ≡ mapMaybe f xs ++ mapMaybe f ys
+  unzipWith-++           : unzipWith f (xs ++ ys) ≡
+                           zip _++_ _++_ (unzipWith f xs) (unzipWith f ys)
+  catMaybes-concatMap    : catMaybes ≗ concatMap fromMaybe
+  catMaybes-++           : catMaybes (xs ++ ys) ≡ catMaybes xs ++ catMaybes ys
+  map-catMaybes          : map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
+  Any-catMaybes⁺         : Any (M.Any P) xs → Any P (catMaybes xs)
+  mapMaybeIsInj₁∘mapInj₁ : mapMaybe isInj₁ (map inj₁ xs) ≡ xs
+  mapMaybeIsInj₁∘mapInj₂ : mapMaybe isInj₁ (map inj₂ xs) ≡ []
+  mapMaybeIsInj₂∘mapInj₂ : mapMaybe isInj₂ (map inj₂ xs) ≡ xs
+  mapMaybeIsInj₂∘mapInj₁ : mapMaybe isInj₂ (map inj₁ xs) ≡ []
   ```
 
 * In `Data.List.Relation.Binary.Sublist.Setoid.Properties`:
@@ -496,7 +504,15 @@ Additions to existing modules
 
 * Added new proofs to `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
-  product-↭ : product Preserves _↭_ ⟶ _≡_
+  product-↭   : product Preserves _↭_ ⟶ _≡_
+  catMaybes-↭ : xs ↭ ys → catMaybes xs ↭ catMaybes ys
+  mapMaybe-↭  : xs ↭ ys → mapMaybe f xs ↭ mapMaybe f ys
+  ```
+
+* Added new proofs to `Data.List.Relation.Binary.Permutation.Setoid.Properties.Maybe`:
+  ```agda
+  catMaybes-↭ : xs ↭ ys → catMaybes xs ↭ catMaybes ys
+  mapMaybe-↭  : xs ↭ ys → mapMaybe f xs ↭ mapMaybe f ys
   ```
 
 * Added new functions in `Data.String.Base`:
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -718,29 +718,25 @@ map-concatMap f g xs = begin
 -- catMaybes
 
 catMaybes-concatMap : catMaybes {A = A} ≗ concatMap fromMaybe
-catMaybes-concatMap [] = refl
-catMaybes-concatMap (mx ∷ xs) with ih ← catMaybes-concatMap xs | mx
-... | just x  = cong (x ∷_) ih
-... | nothing = ih
+catMaybes-concatMap []             = refl
+catMaybes-concatMap (just x  ∷ xs) = cong (x ∷_) $ catMaybes-concatMap xs
+catMaybes-concatMap (nothing ∷ xs) = catMaybes-concatMap xs
 
 length-catMaybes : ∀ xs → length (catMaybes {A = A} xs) ≤ length xs
-length-catMaybes [] = ≤-refl
-length-catMaybes (mx ∷ xs) with ih ← length-catMaybes xs | mx
-... | just _  = s≤s ih
-... | nothing = m≤n⇒m≤1+n ih
+length-catMaybes []             = ≤-refl
+length-catMaybes (just _  ∷ xs) = s≤s $ length-catMaybes xs
+length-catMaybes (nothing ∷ xs) = m≤n⇒m≤1+n $ length-catMaybes xs
 
 catMaybes-++ : (xs ys : List (Maybe A)) →
                catMaybes (xs ++ ys) ≡ catMaybes xs ++ catMaybes ys
-catMaybes-++ []        _  = refl
-catMaybes-++ (mx ∷ xs) ys with ih ← catMaybes-++ xs ys | mx
-... | just x  = cong (x ∷_) ih
-... | nothing = ih
+catMaybes-++ []             _  = refl
+catMaybes-++ (just x  ∷ xs) ys = cong (x ∷_) $ catMaybes-++ xs ys
+catMaybes-++ (nothing ∷ xs) ys = catMaybes-++ xs ys
 
 map-catMaybes : (f : A → B) → map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
-map-catMaybes _ [] = refl
-map-catMaybes f (mx ∷ xs) with ih ← map-catMaybes f xs | mx
-... | just x  = cong (f x ∷_) ih
-... | nothing = ih
+map-catMaybes _ []             = refl
+map-catMaybes f (just x  ∷ xs) = cong (f x ∷_) $ map-catMaybes f xs
+map-catMaybes f (nothing ∷ xs) = map-catMaybes f xs
 
 Any-catMaybes⁺ : ∀ {P : Pred A ℓ} {xs : List (Maybe A)} →
                  Any (MAny P) xs → Any P (catMaybes xs)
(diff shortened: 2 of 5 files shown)


src/Data/List/Properties.agda
```diff
@@ -83,71 +82,6 @@ private
 ≡-dec _≟_ []       (y ∷ ys) = no λ()
 ≡-dec _≟_ (x ∷ xs) (y ∷ ys) = ∷-dec (x ≟ y) (≡-dec _≟_ xs ys)
 
-------------------------------------------------------------------------
```
JacquesCarette, line 86: Any strong reason to re-order the file this way?

src/Data/List/Properties.agda
```diff
@@ -740,6 +714,39 @@ map-concatMap f g xs = begin
...
+-- catMaybes
+
+catMaybes-concatMap : catMaybes {A = A} ≗ concatMap fromMaybe
+catMaybes-concatMap []             = refl
+catMaybes-concatMap (just x  ∷ xs) = cong (x ∷_) (catMaybes-concatMap xs)
+catMaybes-concatMap (nothing ∷ xs) = catMaybes-concatMap xs
+
+length-catMaybes : ∀ xs → length (catMaybes {A = A} xs) ≤ length xs
+length-catMaybes []             = ≤-refl
+length-catMaybes (just x  ∷ xs) = s≤s (length-catMaybes xs)
```
JacquesCarette, line 727: Personally I think it's clearer to not do 2-level pattern matching but to delegate the job to `maybe`, i.e.
```agda
length-catMaybes (x  ∷ xs) = let l = length-catMaybes xs) in maybe (\ _ → s≤s l) (m≤n⇒m≤1+n l) x
```
which more cleanly separates the list-based reasoning from the maybe-based reasoning.

(In my own code, I'd got further and use higher-order instead of `let` but I've been convinced that this is a step too far for stdlib)

src/Data/List/Properties.agda
```diff
@@ -792,6 +799,26 @@ module _ (g : B → C) (f : A → Maybe B) where
     mapMaybe (Maybe.map g) (map f xs)  ≡⟨ mapMaybe-map _ f xs ⟩
     mapMaybe (Maybe.map g ∘ f) xs      ∎
 
+mapMaybeIsInj₁∘mapInj₁ : (xs : List A) → mapMaybe (isInj₁ {B = B}) (map inj₁ xs) ≡ xs
+mapMaybeIsInj₁∘mapInj₁ = λ where
```
JacquesCarette, line 803: why not pattern-match ? This doesn't match the rest of the style.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -373,3 +374,28 @@ product-↭ (swap {xs} {ys} x y r) = begin
...
   where open ≡-Reasoning
+
+------------------------------------------------------------------------
+-- catMaybes
+
+catMaybes-↭ : xs ↭ ys → catMaybes xs ↭ catMaybes ys
+catMaybes-↭ refl = refl
+catMaybes-↭ (trans xs↭ ↭ys) = trans (catMaybes-↭ xs↭) (catMaybes-↭ ↭ys)
+catMaybes-↭ {xs = .(mx ∷ _)} {.(mx ∷ _)} (prep mx xs↭)
+  with mx
```
JacquesCarette, line 385: `maybe` would be better than `with` here.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -373,3 +374,28 @@ product-↭ (swap {xs} {ys} x y r) = begin
...
+... | nothing = catMaybes-↭ xs↭
+... | just x  = prep x $ catMaybes-↭ xs↭
+catMaybes-↭ {xs = .(mx ∷ my ∷ _)} {.(my ∷ mx ∷ _)} (swap mx my xs↭)
+  with mx | my
```
JacquesCarette, line 389: whereas this indicates a missing combinator in `Data.Maybe.Base`, a more general version of `zipWith`.

JacquesCarette (changes requested):

src/Data/List/Properties.agda
jamesmckinna: Also: `where` instead of `let` cf. your own #2369 ?

src/Data/List/Properties.agda
```diff
@@ -48,19 +49,17 @@ open import Relation.Unary using (Pred; Decidable; ∁)
...
-  variable
-    a b c d e p : Level
-    A : Set a
-    B : Set b
-    C : Set c
-    D : Set d
-    E : Set e
-    x y z w : A
-    xs ys zs ws : List A
+private variable
```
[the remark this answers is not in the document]
omelkonian: I don't see this anywhere in `docs/style-guide.md` btw. Maybe add an example like the following to make it clear:
```agda
private 
  abstract 
    f : ...
    f = ...
```

src/Data/List/Properties.agda
omelkonian: I prefer less ink on the page, especially if it the added information is completely useless/redundant (in this case repeating the name `mapMaybeIsInj₁∘mapInj₁`).

It gets even worse when you have to repeat arguments as well; maybe the case of `Any-catMaybes⁺` convinces you...

src/Data/List/Properties.agda
omelkonian: >  Personally I think it's clearer to not do 2-level pattern matching but to delegate the job to maybe

Your suggested change doesn't typecheck though: using eliminators in dependent situations often requires that you explicitly provide the 'motive' (in this case generalising over `mx`). Indeed that's what people used to do before proper dependent pattern matching came along:
```agda
length-catMaybes : ∀ xs → length (catMaybes {A = A} xs) ≤ length xs
length-catMaybes []        = ≤-refl
length-catMaybes (mx ∷ xs) = let IH = length-catMaybes xs in
  maybe {B = λ mx → length (catMaybes (mx ∷ xs)) ≤ length (mx ∷ xs)}
        (λ _ → s≤s IH)
        (m≤n⇒m≤1+n IH)
        mx
```
It gets even worse in more complicated scenarios like `catMaybes-↭` where you need to introduce more variables, etc.

A quick `grep` through the codebase reveals no other explicit motives for `maybe`: are you sure you want this change (and most importantly, maintaining this code)?

Btw, `length-catMaybes` was already there :rofl:

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
omelkonian: See comment above.

omelkonian: For reference, this is what you would need to write:
```agda
catMaybes-↭ {xs = .mx ∷ xs} {.mx ∷ ys} (prep mx xs↭)
  = maybe {B = λ mx → catMaybes (mx ∷ xs) ↭ catMaybes (mx ∷ ys)}
          (λ x → prep x $ catMaybes-↭ xs↭)
          (catMaybes-↭ xs↭)
          mx
```
Notice the necessity to introduce variables for the tails of the two lists.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
omelkonian: I don't get what you're suggesting here: what does `zipWith` have to do with this proof? 

Given your previous comments, you probably want an eliminator for two maybes simultaneously?
```agda
maybe² : ∀ {A : Set a} {B : Maybe A → Maybe A → Set b}
  → (∀ x y → B (just x) (just y))
  → (∀ x → B (just x) nothing)
  → (∀ y → B nothing (just y))
  → B nothing nothing
  → (x y : Maybe A)
  → B x y
maybe² jj jn nj nn = λ where
  (just x) (just y) → jj x y
  (just x) nothing  → jn x
  nothing  (just y) → nj y
  nothing  nothing  → nn
```
Ofc, the motive will again be necessary:
```agda
catMaybes-↭ {xs = .mx ∷ .my ∷ xs} {.my ∷ .mx ∷ ys} (swap mx my xs↭)  =
    let IH = catMaybes-↭ xs↭ in
    maybe² {B = λ mx my → catMaybes (mx ∷ my ∷ xs) ↭ catMaybes (my ∷ mx ∷ ys)}
           (λ x y → swap x y IH)
           (λ x → prep x IH)
           (λ y → prep y IH)
           IH
           mx my
```

omelkonian:
@JacquesCarette To sum up, let's agree that there are both pros and cons to using old-school eliminators in contrast to immediately using (dependent) pattern matching:
- *PRO:* reusability! e.g. if we have some lemmas about `maybe` we could reuse them for deriving other proofs
- *CONS:*
   + type inference (e.g. explicitly giving the motives in the examples above)
   + someone wanting to prove something about these definitions can immediately see what they have to do case analysis on, while with `maybe` that have to perform one additional mental step
   
Given that there is only one lemma about `maybe` :disappointed: (`maybe-map`) and that the inductive hypotheses in the proofs above can easily be shared using `with IH ← ...`, I have not followed your suggestion to use eliminators and stuck with matching on the maybes and so forth.

Feel free to push another commit on top of my branch to convert to eliminator-style proofs though...

JacquesCarette:
@omelkonian I'll comment on various of the details a bit later, but long story short: you make a very convincing case. I was overly optimistic.

src/Data/List/Properties.agda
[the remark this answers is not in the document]
jamesmckinna: or, indeed, with a 
```agda
private
  variable
    x y z : A
```
block...

src/Data/List/Properties.agda
```diff
@@ -740,6 +714,50 @@ map-concatMap f g xs = begin
...
+    with IH ← map-catMaybes xs
+    with mx
+  ... | just x  = cong (f x ∷_) IH
+  ... | nothing = IH
+
+Any-catMaybes⁺ : ∀ {P : Pred A ℓ} {xs : List (Maybe A)}
+  → Any (MAny P) xs → Any P (catMaybes xs)
+Any-catMaybes⁺ {xs = nothing ∷ xs} (there x∈)       = Any-catMaybes⁺ x∈
+Any-catMaybes⁺ {xs = just x  ∷ xs} (here (just px)) = here px
+Any-catMaybes⁺ {xs = just x  ∷ xs} (there x∈)       = there $ Any-catMaybes⁺ x∈
```
jamesmckinna, line 759: Why here rather than added after L150 in the original?

jamesmckinna: And for the sake of juxtaposition, I'd swap L758 and L759, as well as suppressing `x` and `xs`:
```suggestion
Any-catMaybes⁺ : ∀ {P : Pred A ℓ} {xs : List (Maybe A)}
  → Any (MAny P) xs → Any P (catMaybes xs)
Any-catMaybes⁺ {xs = nothing ∷ _} (there x∈)       = Any-catMaybes⁺ x∈
Any-catMaybes⁺ {xs = just _  ∷ _} (there x∈)       = there $ Any-catMaybes⁺ x∈
Any-catMaybes⁺ {xs = just _  ∷ _} (here (just px)) = here px
```

CHANGELOG.md
```diff
@@ -496,7 +503,9 @@ Additions to existing modules
 
 * Added new proofs to `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
-  product-↭ : product Preserves _↭_ ⟶ _≡_
+  product-↭   : product Preserves _↭_ ⟶ _≡_
+  catMaybes-↭ : xs ↭ ys → catMaybes xs ↭ catMaybes ys
+  mapMaybe-↭  : xs ↭ ys → mapMaybe f xs ↭ mapMaybe f ys
```
jamesmckinna, line 508: Are these not (also) provable for the `Setoid` version?
If so, please add! (for me, that relation is conceptually prior to the `Propositional` one cf. #1761

src/Data/List/Properties.agda
```diff
@@ -740,6 +714,50 @@ map-concatMap f g xs = begin
...
+... | nothing = m≤n⇒m≤1+n IH
+
+catMaybes-++ : (xs ys : List (Maybe A)) →
+               catMaybes (xs ++ ys) ≡ catMaybes xs ++ catMaybes ys
+catMaybes-++ []        _  = refl
+catMaybes-++ (mx ∷ xs) ys
+  with IH ← catMaybes-++ xs ys
+  with mx
+... | just x  = cong (x ∷_) IH
+... | nothing = IH
```
jamesmckinna, line 743: Notwithstanding your discussion with @JacquesCarette , I fail to see how these proofs improve on the original three-liners of L126-L149. What am I not grasping here? Esp. given your noted/notable aversion to `with`-based programming...

src/Data/List/Properties.agda
```diff
@@ -740,6 +714,50 @@ map-concatMap f g xs = begin
...
+
+module _ (f : A → B) where
+
+  map-catMaybes : map f ∘ catMaybes ≗ catMaybes ∘ map (Maybe.map f)
+  map-catMaybes [] = refl
+  map-catMaybes (mx ∷ xs)
+    with IH ← map-catMaybes xs
+    with mx
+  ... | just x  = cong (f x ∷_) IH
+  ... | nothing = IH
```
jamesmckinna, line 753: Ditto.

src/Data/List/Properties.agda
jamesmckinna, line 759: And here, the proof format follow the original style!?

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -33,11 +34,11 @@ open import Relation.Binary.PropositionalEquality.Core as ≡
 open import Relation.Binary.PropositionalEquality.Properties using (module ≡-Reasoning)
 open import Relation.Nullary
 
-private
-  variable
-    a b p : Level
-    A : Set a
-    B : Set b
+private variable
```
jamesmckinna, line 37: May not be in `style-guide`, but the additional blankline+indent  is customary everywhere else in `stdlib`...

jamesmckinna (changes requested):
Great additions, but need to fix up layout and some of the `with`-based proofs in favou rof the (simpler!) originals, I think.

src/Data/List/Properties.agda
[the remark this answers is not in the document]
JacquesCarette: #2390 at least documents that lacuna

src/Data/List/Properties.agda
JacquesCarette: I was definitely overly optimistic regarding `maybe` being able to infer the motive. As it can't, then this 3-line proof is indeed the better one.

src/Data/List/Properties.agda
JacquesCarette: While I greatly sympathize with this 'less ink on the page', I have had to adjust my own style when contributing to `stdlib` where I think a certain amount of uniformity of style is better.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
JacquesCarette: Right, that's worse.

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
JacquesCarette: I meant exactly this `maybe²` which can legitimately also be called `zipWith` for 2 `Maybe`s.

Having said that, clearly it wouldn't improve things (here). There might be a more-dependent version that would be able to deal with this, but that's out-of-scope for this PR.

JacquesCarette:
As I agree with @jamesmckinna 's comments, I won't approve this PR (yet), even though I'm not happy with the parts that I had commented on myself.

src/Data/List/Properties.agda
omelkonian: I used irrefutable `with` for reusability: the induction hypothesis is now only written once and shared across the branches of the case analysis. (the new feature for [left-hand side lets via the `using` keyword](https://github.com/agda/agda/pull/7078) would be even better, but this will have to wait until the next release)

Re. my `with`-scepticism, it mostly refers to basic definitions like `mapMaybe` and not so frequently when you prove properties about them. Especially when a use of `with` is unnecessary, that is you do not really need *dependent* pattern matching; here you actual need the dependency. All at the same time considering the reduction behaviour which might now favour the use of `with`, as mentioned in [@gallais' comment](https://github.com/agda/agda-stdlib/issues/1937#issuecomment-1483389966) in the original discussion about this.

 **However**, just to contradict my own argument above, when one gets into the business of proving meta-properties about these properties (e.g. you view proofs of membership [constructively as indices](https://omelkonian.github.io/formal-prelude/Prelude.Lists.Membership.html) --- ironically I also needed that for [permutations](https://omelkonian.github.io/formal-prelude/Prelude.Lists.PermutationsMeta.html) --- also c.f. notes about "meta-properties" in my [thesis](https://omelkonian.github.io/data/phd-thesis.pdf)), there is no distinction between "base" definitions and "properties". Having said that, the standard library does not include any of these to my knowledge, so we can just ignore these ultra-rare cases for the sake of practicality.

Therefore, I still think the current form sits nicely between the eliminator-style approach requested by Jacques and the original fully-direct non-reusing pattern matching.

src/Data/List/Properties.agda
omelkonian: > Why here rather than added after L150 in the original?

Grouped the properties of `catMaybe` with the ones for `mapMaybe`, as it seemed more intuitive.
Also, lemma `catMaybes-concatMap` mentions `concatMap` which comes just before in the file.

omelkonian: Yep, here there is no shared expression to reuse, therefore I classify this case under 'unnecessary `with`s'.

JacquesCarette: I definitely prefer the 3-liner over this!

jamesmckinna: Well, perhaps there is no 'right' order for module contents, so the rule-of-thumb ends up being: try, where possible, to follow in `Properties` the order in `Base`... but even that seems hard to stick to.

CHANGELOG.md
omelkonian: Indeed they are, but require another module/file that is more general than `Data.List.Relation.Binary.Permutation.Setoid.Properties`, i.e. without a fixed module parameter for setoid equality on `A`. 

I added this as `Data.List.Relation.Binary.Permutation.Setoid.Properties.Maybe`.

src/Data/List/Properties.agda
jamesmckinna: I'm sympathetic to the argument based on sharing, but where it falls down for me is in the simple-minded LOC here... and some kind of trade-off in cognitive load that follows from that... sharing a recursive call both affords, but also *enforces*, a particular understanding of the call structure which is quite distinct from the one-clause-one-call-structure style. You might argue it's 'richer', but you also pay a higher price for it!

My instinct for `stdlib` goes towards a more 'middlebrow' approach... with a correspondingly lower bar to entry for newcomers?

jamesmckinna: As for 'meta' properties, I find increasingly the need for proofs of 'higher-dimensional' lifting lemmas (eg in [`Data.List.Relation.Binary.BagAndSetEquality`](https://github.com/agda/agda-stdlib/blob/c5255d0c28ca39b3233f4ade7c21a0a96b7a5b69/src/Data/List/Relation/Binary/BagAndSetEquality.agda#L579-L592)...

CHANGELOG.md
jamesmckinna: Ugh! yes... a wrinkle that recurs through the library. 
But can't such proofs follow eg the model of `Data.List.Relation.Binary.Permutation.Setoid.Properties.map`?

omelkonian: Unfortunately not: after flatting the maybes, you need to instantiate `catMaybes-↭` with type `B` instead of the (global) type `A`.

src/Data/List/Properties.agda
omelkonian: There are actually already occurrences of the `with ih ← ...` pattern already in the codebase + one could say that this is more intuitive to beginners than simultaneous deep pattern matching maybe?

CHANGELOG.md
jamesmckinna: Grrr. OK.

src/Data/List/Relation/Binary/Permutation/Setoid/Properties/Maybe.agda
```diff
@@ -0,0 +1,76 @@
...
+module _ (sᴬ : Setoid a ℓ) (sᴮ : Setoid b ℓ′) where
+  open Setoid sᴬ using () renaming (_≈_ to _≈ᴬ_)
+  open Permutation sᴬ using () renaming (_↭_ to _↭ᴬ_)
+
+  open Setoid sᴮ using () renaming (_≈_ to _≈ᴮ_)
+  open Permutation sᴮ using () renaming (_↭_ to _↭ᴮ_)
+
+  private sᴹᴮ = setoidᴹ sᴮ
+  open Setoid sᴹᴮ using () renaming (_≈_ to _≈ᴹᴮ_)
+  open Permutation sᴹᴮ using () renaming (_↭_ to _↭ᴹᴮ_)
```
jamesmckinna, line 71: This `open` not required!

src/Data/List/Properties.agda
jamesmckinna: Indeed! ... I'm responsible for them... ;-)

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -373,3 +375,26 @@ product-↭ (swap {xs} {ys} x y r) = begin
...
+-- catMaybes
+
+catMaybes-↭ : xs ↭ ys → catMaybes xs ↭ catMaybes ys
+catMaybes-↭ refl = refl
+catMaybes-↭ (trans xs↭ ↭ys) = trans (catMaybes-↭ xs↭) (catMaybes-↭ ↭ys)
+catMaybes-↭ {xs = .mx ∷ xs} {.mx ∷ ys} (prep mx xs↭)
+  with ih ← catMaybes-↭ xs↭ | mx
+... | nothing = ih
+... | just x  = prep x ih
+catMaybes-↭ {xs = .mx ∷ .my ∷ xs} {.my ∷ .mx ∷ ys} (swap mx my xs↭)
```
jamesmckinna, line 389: In light of the `Setoid.Properties.Maybe` versions below, is there any longer any need to expose the forced implicit bindings for `xs` and `ys`? I'd reckon probably not...?

jamesmckinna:
Thanks for the continued updates, but I won't re-review until you signal it's time to do so though!

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -144,7 +144,7 @@ steps-respʳ ys≋xs           (trans ys↭ws ws↭zs) = cong (steps ys↭ws +_)
 ------------------------------------------------------------------------
 -- map
 
-module _ (T : Setoid b ℓ) where
+module _ {ℓ} (T : Setoid b ℓ) where
```
omelkonian, line 147: Note that this is a side-fix: previously the function's co-domain (equivalence) was unnecessarily restricted to live at the same level as the domain (equivalence).

omelkonian:
@jamesmckinna ready for review

CHANGELOG.md
```diff
@@ -343,47 +343,54 @@ Additions to existing modules
...
+  alignWith-comm         : f ∘ swap ≗ f → alignWith f xs ys ≡ alignWith f ys xs
+  align-flip             : align xs ys ≡ map swap (align ys xs)
+  zip-flip               : zip xs ys ≡ map swap (zip ys xs)
+  unzipWith-swap         : unzipWith (swap ∘ f) ≗ swap ∘ unzipWith f
+  unzip-swap             : unzip ∘ map swap ≗ swap ∘ unzip
+  take-take              : take n (take m xs) ≡ take (n ⊓ m) xs
+  take-drop              : take n (drop m xs) ≡ drop m (take (m + n) xs)
+  zip-unzip              : uncurry′ zip ∘ unzip ≗ id
+  unzipWith-zipWith      : f ∘ uncurry′ g ≗ id → length xs ≡
+                           length ys → unzipWith f (zipWith g xs ys) ≡ (xs , ys)
```
andreasabel, line 381: Linebreak in awkward place here (break at "->" instead).

src/Data/List/Properties.agda
```diff
@@ -792,6 +798,22 @@ module _ (g : B → C) (f : A → Maybe B) where
     mapMaybe (Maybe.map g) (map f xs)  ≡⟨ mapMaybe-map _ f xs ⟩
     mapMaybe (Maybe.map g ∘ f) xs      ∎
 
+mapMaybeIsInj₁∘mapInj₁ : (xs : List A) → mapMaybe (isInj₁ {B = B}) (map inj₁ xs) ≡ xs
+mapMaybeIsInj₁∘mapInj₁ []       = refl
+mapMaybeIsInj₁∘mapInj₁ (x ∷ xs) = cong (x ∷_) (mapMaybeIsInj₁∘mapInj₁ xs)
```
andreasabel, line 803: Do we have more general versions of this?  Looks like a property of a retract (inj1) aka embedding-projection pair.
So if `proj . emb = just`, then `mapMaybe proj . map emb = id`.

andreasabel (commented):
Good stuff.

CHANGELOG.md
omelkonian: Not my change, but fixed that since you spotted it :)

src/Data/List/Properties.agda
omelkonian: You're absolutely right, now generalised.

andreasabel (approved):
Thanks for the update!

omelkonian:
@JacquesCarette @jamesmckinna Ready to merge?

JacquesCarette (approved):
Yes from me.

CHANGELOG.md
jamesmckinna: See also #2397 this may require downstream refactoring...

jamesmckinna (commented):
I don't think the new `with`-based code (and the accompanying code motion) is an improvement over the existing pattern-matching based definitions, but you have two approvals from others, so I'm not going to stand in the way.

omelkonian:
> I don't think the new `with`-based code (and the accompanying code motion) is an improvement over the existing pattern-matching based definitions, but you have two approvals from others, so I'm not going to stand in the way.

Isn't the reuse of the inductive hypothesis (`ih`) clearly an improvement?

jamesmckinna:
> > I don't think the new `with`-based code (and the accompanying code motion) is an improvement over the existing pattern-matching based definitions, but you have two approvals from others, so I'm not going to stand in the way.
> 
> Isn't the reuse of the inductive hypothesis (`ih`) clearly an improvement?

In this case, I (personally) actually think no, it's not, at least not enough to justify the rewriting of existing code. Funnily enough, I think that both you and @JacquesCarette have convinced me, in different ways, and from different starting points, to view such code style with... scepticism for the time being. And one reason I dismissed my review was to avoid going back over this ground...so: I stand aside, as previously.

andreasabel:
Mmh, I tend to agree with James @jamesmckinna that `with` should be avoided where possible.  The reason is that it internally generates a helper function which Agda might sometimes leak in printing.  Also, it bears some hazards in proving things about the parent function as the name of the auxiliary function is not available to the user.
So, I think the general thrust should be to get rid of `with` where possible rather than introducing more `with`s.

omelkonian:
> While I very much agree with @andreasabel 's general stance, especially for any definition that computes, I've gotten more relaxed about "pure proofs".

See my comment above on "meta-properties", which nullifies the distinction between programs and "pure proofs".

> Really what's we're sorely missing (and this is, I believe, what @omelkonian was using it for) is a proper pattern-matching `let` that doesn't unfold eagerly. This is a real wart in the current Agda language (along with the same issue wrt module parameters).

Isn't the upcoming `using` syntax that I mentioned above already what you wish for (although I'm not sure about its unfolding behaviour, but the `opaque` feature should cover that).

JacquesCarette:
Yes, meta-properties, being proofs-about-proofs, do blur things. I've hated doing such proofs so much (in MLTT) that I block them out; when done 'properly', a la category theory where one adds higher-level cells, then these become sane again. `subst` hell is really the worst hell of them all.

I don't understand (yet!) enough of the details of `using` to know if it really going to be 'the' feature needed. I suspect that as long as there isn't some kind of `let`-like construct in the internal AST, it won't be, it'll merely be pleasant syntactic sugar.

andreasabel:
@omelkonian: Thanks, Orestis, for your patience and willingness to revert changes!

jamesmckinna:
> @omelkonian: Thanks, Orestis, for your patience and willingness to revert changes! (`with` is gone again)

Seconded!

PR #2392, opened 2024-05-20
`style-guide` rule for initial `private` block in modules

### LINKED ISSUE #2390: Document `variable` block indentation style
As uncovered by PR #2389 it seems that the preferred style
```agda
private
  variable
    a : Level
```
isn't actually documented.

jamesmckinna:
Fixes #2390.

NB. The aim of the PR is to document what we (mostly) currently do, rather than fix in stone what we want to be able to do... the text is always changeable, and adherence to the guidelines a matter of manual review/checking!

### DIFF
--- a/doc/style-guide.md
+++ b/doc/style-guide.md
@@ -229,6 +229,34 @@ automate most of this.
     }
   ```
 
+#### Layout of initial `private` block
+
+* Since the introduction of generalizable `variable`s (see below),
+  this block provides a very useful way to 'fix'/standardise notation
+  for the rest of the module, as well as introducing local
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
I agree with this style. But we should probably get various people's buy-in, i.e. this should have a higher threshold than usual.

jamesmckinna:
> I agree with this style. But we should probably get various people's buy-in, i.e. this should have a higher threshold than usual.

Absolutely!

andreasabel (approved):

gallais (approved):

Taneb (approved):

Issue #2391, opened 2024-05-20
[DRY] what's the best way to `public`ly re-export properties/structure?


jamesmckinna:
Cf. #2255 

Thanks to the sophistication of the re-exporting strategies embodied in `Relation.Binary.Bundles`, we could streamline concrete instances of such strategies, such as in [`Data.List.Binary.Relation.Equality.DecSetoid`](https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Relation/Binary/Equality/DecSetoid.agda). 

Compare the following (23 LOC; 4 `import`s; *1* explicit definition; and 2 `public` re-exports, the first of which also introduces the correct fixity, the second shared with the existing version):
```agda
open import Relation.Binary.Bundles using (DecSetoid)

module Data.List.Relation.Binary.Equality.DecSetoid
  {a ℓ} (DS : DecSetoid a ℓ) where

import Data.List.Relation.Binary.Equality.Setoid as SetoidEquality
import Data.List.Relation.Binary.Pointwise as Pointwise
open import Level using (_⊔_)

------------------------------------------------------------------------
-- Additional properties

≋-decSetoid : DecSetoid a (a ⊔ ℓ)
≋-decSetoid = Pointwise.decSetoid DS

open DecSetoid ≋-decSetoid public
  using (setoid)
  renaming (isDecEquivalence to ≋-isDecEquivalence; _≟_ to _≋?_)

------------------------------------------------------------------------
-- Make all definitions from setoid equality available

open SetoidEquality setoid public
```
with the existing version (31 LOC; 7 `import`s; 3 explicit definitions, incl. the fixity declaration; and the 1 `public` re-export shared with the above).

So we trade economy of expression/export for explicitness of definition... although in each case, largely by delegation, so it's not immediately clear that there is much gain in understanding, except the documentation afforded by the types...

Which is 'better', and which style should the library adopt from v2.* onwards?

jamesmckinna:
Similarly, existing style is to build up, definition-by-definition, towards eg `Setoid` instances, as in eg. [Data.List.Binary.Relation.Equality.Setoid](https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Relation/Binary/Equality/Setoid.agda), rather than define the `Setoid` first, and then `open` it `public`ly and suitably `renaming` its components...

jamesmckinna:
[...]

PR #2393, opened 2024-05-21
Add the `Setoid`-based `Monoid` on `(List, [], _++_)`

### LINKED ISSUE #2360: Add the `Setoid`al structure of a (free) `Monoid` on `List`
(Extracted from #2350)
Module `Data.List.Relation.Binary.Equality` defines the `Setoid` lifting of equality from a `Setoid` on `A` to one on `List A`.

But nowhere that I can find is the `++-[]-monoid` bundle/structure etc. defined for *this* `Setoid`... so we should add it.

See ~~[this blob](https://github.com/agda/agda-stdlib/blob/63ef07ed4e67d03942e9c092afc27d5a4b7fcbeb/src/Algebra/Action/Construct/Free.agda#L37-L58)~~ #2393  for one possible implementation.

Issues:
* where should this construction live?
* what else should be added along these lines?
* what other 'obvious' `Setoid` liftings are missing?

NB [this comment seems never to have been acted upon](https://github.com/agda/agda-stdlib/blob/69de98beb2da9e647597afd5f8726b8a892f226c/src/Data/List/Properties.agda#L7-L8) cf. #568

jamesmckinna:
Refactored from #2350 : fixes #2360

Outstanding `library-design` issue (from #2360 ): where should this live?

Downstream issues: what else belongs here (or wherever this ends up)?

### DIFF
--- a/src/Data/List/Relation/Binary/Equality/Setoid/Properties.agda
+++ b/src/Data/List/Relation/Binary/Equality/Setoid/Properties.agda
@@ -12,8 +12,8 @@ module Data.List.Relation.Binary.Equality.Setoid.Properties
   {c ℓ} (S : Setoid c ℓ)
   where
 
-open import Algebra.Bundles using (Monoid)
-open import Algebra.Structures using (IsMonoid)
+open import Algebra.Bundles using (Magma; Semigroup; Monoid)
+import Algebra.Structures as Structures
 open import Data.List.Base using (List; []; _++_)
 import Data.List.Properties as List
 import Data.List.Relation.Binary.Equality.Setoid as ≋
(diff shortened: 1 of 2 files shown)


andreasabel (approved):

gallais (approved):

src/Data/List/Relation/Binary/Equality/Setoid/Properties.agda
```diff
@@ -0,0 +1,46 @@
...
+
+------------------------------------------------------------------------
+-- The []-++-Monoid
+
+-- Structure
+
+isMonoid : IsMonoid _≋_ _++_ []
+isMonoid = record
+  { isSemigroup = record
+    { isMagma = record
```
Taneb, line 34: Could we have these `isMagma` and `isSemigroup` defined at the top level? In case someone needs them it'll be more convenient than digging into this. And also, I think it looks neater

Taneb (changes requested):

src/Data/List/Relation/Binary/Equality/Setoid/Properties.agda
jamesmckinna: Makes sense, and while I was working on this PR, the ideas behind #2391 were coming into focus for me: so to make this sharper, do I 
* `open` the `Monoid monoid` structure to `public`ly export (a/the relevant subcollection of) its fields, 
* or build up to `isMonoid` by progressive instantiation?
Current library style is the first option, so I guess I'll follow that, but already I can imagine a time downstream where I/we opt for the second.

JacquesCarette (approved):

PR #2395, opened 2024-05-23
Refactor `Data.List.Base.scan*` and their properties

### LINKED ISSUE #2267: A 'better' (less strict) `Data.List.Base.tails`?

jamesmckinna:
Refactored/uniformised from #2258 and #2269 , which now close unmerged instead...
Fixes #2267 and relevant item of #2359 .
Many thanks to all who commented on/reviewed the earlier PRs!


src/Data/List/NonEmpty/Base.agda
```diff
@@ -143,6 +143,24 @@ concatMap f = concat ∘′ map f
...
+  go (x ∷ xs) = List.[ x ] ∷ List.map (x ∷_) (go xs)
+
+-- Tails
+
+tails : List A → List⁺ (List A)
+tails xs = xs ∷ go xs
+  where
+  go : List A → List (List A)
+  go []       = []
+  go (x ∷ xs) = xs ∷ go xs
```
andreasabel, line 162: Ah, so the implementation of inits/tails isn't shared between list and non-empty list.  
Is this because of module dependency issues?

src/Data/List/Scans.agda
```diff
@@ -0,0 +1,106 @@
...
+    where
+    go : A → List B → List A
+    go _ []       = []
+    go e (x ∷ xs) = let fex = f e x in fex ∷ go fex xs
+
+  scanl : A → List B → List A
+  scanl e = toList ∘ scanl⁺ e
+
+------------------------------------------------------------------------
+-- Properties
```
Taneb, line 57: I can't recall if there was discussion previously about having the properties in a separate file... I think it would greatly help the dependency graph if you only need the definitions to separate them

jamesmckinna: Not separating them was indeed pure laziness on my part, and doing so instead would improve the dependency graph...
... so I'll switch to Draft until that's completed.

src/Data/List/NonEmpty/Base.agda
jamesmckinna: Indeed, `inits` gets used under `Membership.Propositional` in such a way as to (seem to) greatly complicate the dependency graph if I were to try to factor out a common root definition under `NonEmpty`... but I I will investigate further while undertaking the refactoring requested below.

jamesmckinna: Suggest instead: `inits-tail` and `tails-tail` for the auxiliaries?

src/Data/List/Scans.agda
jamesmckinna: Done (I think)

src/Data/List/NonEmpty/Base.agda
jamesmckinna: Ah! Much better... and yes, AFAIK, we would/should/do allow `tail∘inits` and `tail∘tails` (and isn't that better, for uniformity's sake?)

andreasabel (approved):
Nice!  Now the recursion in inits and tails is _iteration_ (with a name one can form lemmas about), the purest form, best for proofs!

src/Data/List/Scans/Properties.agda
```diff
@@ -0,0 +1,80 @@
...
+  private
+    h = List.foldr f e
+
+  scanr⁺-defn : scanr⁺ f e ≗ List⁺.map h ∘ List⁺.tails
+  scanr⁺-defn []       = refl
+  scanr⁺-defn (x ∷ xs) = let eq = scanr⁺-defn xs
+    in cong₂ (λ z → f x z ∷_) (cong List⁺.head eq) (cong toList eq)
+
+  scanr-defn : scanr f e ≗ List.map h ∘ List.tails
+  scanr-defn xs = begin
```
Taneb, line 46: I wonder if this proof (resp. `scanl-defn` below) can be made prettier by using `scanr⁺-defn` (resp `scanl⁺-defn`)

Taneb (approved):

src/Data/List/Scans/Properties.agda
```diff
@@ -0,0 +1,80 @@
...
+    scanr f e xs
+      ≡⟨ cong toList (scanr⁺-defn xs) ⟩
```
jamesmckinna, line 48: Isn't here where `scanr⁺-defn`gets used? Not sure how to make it prettier, though?

src/Data/List/Scans/Properties.agda
```diff
@@ -0,0 +1,80 @@
...
+  scanl⁺-defn : ∀ e → scanl⁺ f e ≗ List⁺.map (h e) ∘ List⁺.inits
+  scanl⁺-defn e []       = refl
+  scanl⁺-defn e (x ∷ xs) = let eq = scanl⁺-defn (f e x) xs in
+    cong (e ∷_) $ cong (f e x ∷_) $ trans (cong List⁺.tail eq) (List.map-∘ _)
+
+  scanl-defn : ∀ e → scanl f e ≗ List.map (h e) ∘ List.inits
+  scanl-defn e []       = refl
+  scanl-defn e (x ∷ xs) = cong (e ∷_) $ begin
+    scanl f (f e x) xs
+      ≡⟨ scanl-defn (f e x) xs ⟩
```
jamesmckinna, line 73: I agree that here there's a direct appeal to IH, so perhaps it can be untangled in favour of `scanl⁺-defn`?

src/Data/List/Scans/Properties.agda
jamesmckinna: Yes, latest commits fold in the new defns of `inits` etc. thanks to @andreasabel 's suggestions, so indeed it is as you conjectured!

CHANGELOG.md
```diff
@@ -65,6 +65,30 @@ Deprecated names
   isRing*  ↦  Algebra.Structures.isRing
   ```
 
+* In `Data.List.Base`:
+  ```agda
+  scanr  ↦  Data.List.Scans.scanr
```
jamesmckinna, line 70: Need to fix the destination and the deprecation warnings!

PR #2399, opened 2024-05-31
Lists: decidability of Subset+Disjoint relations

omelkonian:


src/Data/List/Relation/Binary/Disjoint/DecSetoid.agda
```diff
@@ -0,0 +1,40 @@
...
+open import Relation.Nullary using (¬?; _×-dec_)
+
+module Data.List.Relation.Binary.Disjoint.DecSetoid {c ℓ} (S : DecSetoid c ℓ) where
+
+open DecSetoid S
+open import Data.List.Relation.Binary.Equality.DecSetoid S
+open import Data.List.Relation.Binary.Disjoint.Setoid setoid public
+open import Data.List.Membership.DecSetoid S
+
+¬∉⇒∈ : ∀ {x xs} → ¬ (x ∉ xs) → x ∈ xs
```
jamesmckinna, line 28: This lemma is `decidable-stable` for membership, being `Decidable` via `Any`... suggest refactoring along those lines?

src/Data/List/Membership/DecSetoid.agda
```diff
@@ -30,3 +31,6 @@ x ∈? xs = any? (x ≟_) xs
 
 _∉?_ : Decidable _∉_
 x ∉? xs = ¬? (x ∈? xs)
+
+∈-stable : ∀ {xs ys} → Stable (xs ∈ ys)
```
jamesmckinna, line 35: Quantification over `x` and `xs`, surely?
```suggestion
∈-stable : ∀ {x xs} → Stable (x ∈ xs)
```

src/Data/List/Membership/DecSetoid.agda
```diff
@@ -9,10 +9,11 @@
 open import Relation.Binary.Definitions using (Decidable)
 open import Relation.Binary.Bundles using (DecSetoid)
 open import Relation.Nullary.Decidable using (¬?)
+open import Relation.Nullary using (Stable; decidable-stable)
+open import Data.List.Relation.Unary.Any using (any?)
```
jamesmckinna, line 13: Of these, L10 is the only one required for the opening module header L15 to typecheck, so `style-guide` [requires](https://github.com/agda/agda-stdlib/blob/59d7b2ec356d062245a0fdcd8141b3fca404023d/doc/style-guide.md?plain=1#L105-L121) moving all the other `import`s *below* that line...

src/Data/List/Membership/DecSetoid.agda
jamesmckinna: Also, my previous observation about `decidable-stable` was more in the spirit of making the previous versions of the lemma, and in fact *also* this one itself, redundant... (Fairbairn threshhold again)... since `_∈?_` is the one doing the actual heavy lifting?

jamesmckinna:
Some minor gripes and nitpicks (mostly `style-guide` related) ahead of a proper review, but looks great otherwise...
... apart from the need for`CHANGELOG` entries

src/Data/List/Membership/DecSetoid.agda
omelkonian: I would prefer the inverse policy:
> Only put imports below the parametrized module that actually get instantiated with the module params.

But since you've already decided, I'll go with the flow.

The alphabetical requirement must be a joke, right?
> * The list of imports should be declared in alphabetical order.

JacquesCarette (approved):

jamesmckinna (approved):
Looks good to me!

PR #2402, opened 2024-06-05
Add `IsIdempotentMonoid` and `IsCommutativeBand` to `Algebra.Structures`

### LINKED ISSUE #2138: Add `IdempotentMonoid` and `CommutativeBand`
These feel fairly obvious omissions to the algebra hierarchy.

A commutative band is the same thing as a semilattice but I think it's worth duplicating

Example of a commutative band that is not an idempotent commutative monoid: minimum on naturals

Example of an idempotent monoid that is not an idempotent commtuative monoid: first non-nothing element on `Maybe`

jamesmckinna:
UPDATED Fixes #2138 (and a bit more!)

* Introduces the new `Algebra.Structures` and `Algebra.Bundles`
* Redefines `Algebra.Lattice.Structures.IsSemilattice` and `Algebra.Lattice.Structures.IsBoundedSemilattice` as aliases for them, and fixes the knock-on dependency issues arising for `Algebra.Lattice.Bundles`,  `Algebra.Lattice.Construct.*`
* Adds more manifest fields for substructures arising from this in eg `IsIdempotentSemiring`

NB: (possible downstream PRs?)
* no `Biased` smart constructors (neither under `Algebra.Structures`, nor `Algebra.Lattice.Structures`)
* no `Consequences` (ditto.) to reflect the new arrangements
* no concrete example implementations in `Data.*.Properties` as per #2138 
* no deprecations of the (nevertheless now-redundant?) `Lattice` aliases.

### DIFF
--- a/src/Algebra/Structures.agda
+++ b/src/Algebra/Structures.agda
@@ -249,7 +249,7 @@ record IsIdempotentCommutativeMonoid (∙ : Op₂ A)
 
   isCommutativeBand : IsCommutativeBand ∙
   isCommutativeBand = record { isBand = isBand ; comm = comm }
-    
+
 ------------------------------------------------------------------------
 -- Structures with 1 binary operation, 1 unary operation & 1 element
 ------------------------------------------------------------------------
--- a/src/Algebra/Structures.agda
+++ b/src/Algebra/Structures.agda
@@ -614,6 +614,20 @@ record IsIdempotentSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
 
   open IsSemiring isSemiring public
 
+  +-isIdempotentCommutativeMonoid : IsIdempotentCommutativeMonoid + 0#
+  +-isIdempotentCommutativeMonoid = record
+    { isCommutativeMonoid = +-isCommutativeMonoid
+    ; idem = +-idem
+    }
+
+  open IsIdempotentCommutativeMonoid +-isIdempotentCommutativeMonoid public
+    using ()
+    renaming ( isCommutativeBand to +-isCommutativeBand
+             ; isBand to +-isBand
+             ; isIdempotentMonoid to +-isIdempotentMonoid
+             )
+
+
 record IsKleeneAlgebra (+ * : Op₂ A) (⋆ : Op₁ A) (0# 1# : A) : Set (a ⊔ ℓ) where
   field
     isIdempotentSemiring  : IsIdempotentSemiring + * 0# 1#
--- a/src/Algebra/Structures.agda
+++ b/src/Algebra/Structures.agda
@@ -162,7 +162,7 @@ record IsCommutativeBand (∙ : Op₂ A) : Set (a ⊔ ℓ) where
   open IsBand isBand public
 
   isCommutativeSemigroup : IsCommutativeSemigroup ∙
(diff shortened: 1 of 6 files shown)


Taneb:
I think we should change `Algebra.Lattice` to refer to these structures somehow

jamesmckinna:
> I think we should change `Algebra.Lattice` to refer to these structures somehow

Downstream PR?

MatthewDaggitt (approved):

jamesmckinna:
Have redefined `IsSemilattice` and `IsBoundedSemilattice` as aliases for the new structures, but have *not* deprecated the old names (lots of knock-on viscosity elsewhere in the library).
Consider deprecation of the 'old' names as a downstream issue/PR?

JacquesCarette (approved):

Taneb (changes requested):
I'm reviewing this in a train station, so I could be wrong, but I think you're missing that a commutative idempotent monoid is a commutative band?

And also IdempotentSemiring is missing... A lot of these, maybe now without be a good opportunity to add the links

src/Algebra/Structures.agda
```diff
@@ -216,9 +241,14 @@ record IsIdempotentCommutativeMonoid (∙ : Op₂ A)
...
-  isBand : IsBand ∙
-  isBand = record { isSemigroup = isSemigroup ; idem = idem }
+  isIdempotentMonoid : IsIdempotentMonoid ∙ ε
+  isIdempotentMonoid = record { isMonoid = isMonoid ; idem = idem }
+
+  open IsIdempotentMonoid isIdempotentMonoid public
+    using (isBand)
 
+  isCommutativeBand : IsCommutativeBand ∙
+  isCommutativeBand = record { isBand = isBand ; comm = comm }
```
jamesmckinna, line 251: @Taneb So I think that this is what you were looking for?

jamesmckinna:
> I'm reviewing this in a train station, so I could be wrong, but I think you're missing that a commutative idempotent monoid is a commutative band?

See https://github.com/agda/agda-stdlib/blob/31193c68ea97c341062de656c56218937fca1b12/src/Algebra/Structures.agda#L250-L251

> And also IdempotentSemiring is missing... A lot of these, maybe now without be a good opportunity to add the links

Actually, I'm going to push back here and say these are (perhaps easy) downstream PRs, and I'm trying to avoid to much PR-creep these days in the interests of a tighter merge cycle... ?

Separately, I worry a bit about `IdempotentSemiring` as a name: do we intend (probably?) the `_+_` to be idempotent... in which case what do you call a `Semiring` in which `_*_` is idempotent?... and what's the relationship with `KleeneAlgebra`?

jamesmckinna:
> I'm reviewing this in a train station, so I could be wrong, but I think you're missing that a commutative idempotent monoid is a commutative band?

See comment above: L250-L251.

> And also IdempotentSemiring is missing... A lot of these, maybe now without be a good opportunity to add the links

Latest commit augments `IsIdempotentSemiring` with the relevant substructures (I hope! please check...).

After that, I'd be tempted to say that additional (sub)structures should be downstream PRs, unless you find any *glaring* omissions!?

Taneb (approved):

src/Algebra/Structures.agda
jamesmckinna: Oops... `Bundles` as per original issue...

JacquesCarette (approved):

PR #2407, opened 2024-06-11
Refactor `Algebra.Solver.*Monoid`

### LINKED ISSUE #2403: [DRY] Refactor `Algebra.Solver.*Monoid` (or deprecate entirely?)
At present, there are 3 such modules, two of which (`Commutative` and `IdempotentCommutative`) are not yet superseded by counterparts under `Tactic.*Solver` (only `Monoid` has been reimplemented).

There is a *huge* amount of duplicated code between the three, which could/should be refactored into a common core, basically consisting of:
* a `FreeMonoid` construction
* the existing generic solver based on `Relation.Binary.Reflection`, exporting `solve`
* a 'Tactic' module, exporting `prove`, parametrised on a suitable API `Normal` for normal forms

The three individual modules would then reduce entirely to their construction of a suitable implementation of the `Normal` API, together with appropriate `public` re-export of `solve` and `prove` from the (instantiated) generic constructions.

NB. `Tactic.MonoidSolver` *also* contains an implementation (a different one! the 'functorial' version of variables, rather than the `Fin`-based one) of the free monoid, together with its semantics via the Cayley construction, so downstream there would/should be ample opportunity to refactor that as well... at such time as we have a common implementation of `FreeMonoid` (cf. #1962 ) on which we can agree.

UPDATED: indeed, further downstream refactoring would be possible (see [discussion here](https://github.com/agda/agda-stdlib/pull/2407#issuecomment-2258285549) or else #2457 ). Once you have  a suitable API of `Normal` forms (parameterised by a `RawMonoid M`):
* the `normalise` function can be expressed generically, given a `rawMonoid` structure on `Normal` forms, together with an interpretation of `var`
* the `correct` function can be expressed generically, given a `MonoidHomomorphism` from `Normal` to `M`, together with the 'obvious' correctness/coherence condition on (the interpretation of) `var`, *on the assumption that `M` admits `IsMonoid`* (ie `M` is actually a `Monoid`)
* at which point, all the additional existing machinery involving `prove` etc. can be folded in on top of the above two definable operations.

jamesmckinna:
Fixes #2403 

NB: 
* for backwards compatibility, each module exports the `Expr` API publicly (but not the new API for `Normal` forms)
* UPDATED: module `Monoid.Solver` adopts the API for `*CommutativeMonoid`, but maybe the `Monoid.prove` type should standardise on this as well?
* misc. tweaks to existing other `Solver` modules


JacquesCarette (approved):

jamesmckinna:
Converted to DRAFT for further refactoring

jamesmckinna:
I'd converted back to draft with a view to more drsatic refactoring, but then lost my nerve over the knock-on consequences for deprecation/backwards-compatibility, so I'll leave those for another time... ;-)

And in the meanwhile: merge conflict hell!

jamesmckinna:
Will leave the `CHANGELOG` until after v2.1

src/Algebra/Solver/CommutativeMonoid/Normal.agda
```diff
@@ -0,0 +1,135 @@
...
+-- A singleton bag.
+
+sg : (i : Fin n) → Normal n
+sg zero    = 1 ∷ empty
+sg (suc i) = 0 ∷ sg i
+
+-- The composition of normal forms.
+infixr 10 _•_
+
+_•_  : (v w : Normal n) → Normal n
```
JacquesCarette, line 79: `zipWith` ?

src/Algebra/Solver/IdempotentCommutativeMonoid/Normal.agda
```diff
@@ -0,0 +1,160 @@
...
+
+sg : (i : Fin n) → Normal n
+sg zero    = true ∷ empty
+sg (suc i) = false ∷ sg i
+
+-- The composition of normal forms.
+infixr 10 _•_
+
+_•_ : (v w : Normal n) → Normal n
+[]      • []      = []
```
JacquesCarette, line 81: `zipWith` ?

JacquesCarette (changes requested):
A couple of minor improvements suggested.

On a bigger scale: this seems like a lot of duplication. Feels like one could abstract over the differences between `CommutativeMonoid/Normal` and `IdempotentCommutativeMonoid/Normal` instead of this copy-pasta.  But that's an easy comment for me to make!!! I should actually try it.

jamesmckinna:
> A couple of minor improvements suggested.
> 
> On a bigger scale: this seems like a lot of duplication. Feels like one could abstract over the differences between `CommutativeMonoid/Normal` and `IdempotentCommutativeMonoid/Normal` instead of this copy-pasta. But that's an easy comment for me to make!!! I should actually try it.

For sure, the duplication was *already there in the library*. This PR merely refactors it to make it evident, so in that sense this PR should be regarded as 'provisional', even if mergeable. 

The downstream refactoring alluded to in the preamble would spruce up the homomorphism correctness proof machinery, but at the cost, again, of having to re-export stuff, now from a different, but again shared, module (I've experimented with this already, but fresh eyes might be welcome!), else have a heavier-weight move/rename-plus-deprecate refactoring, which I just couldn't face right now.

src/Algebra/Solver/IdempotentCommutativeMonoid/Normal.agda
jamesmckinna: !!! nice observation !!!

jamesmckinna: (I'd not attempted to actually analyse existing code, only move it around...)

jamesmckinna:
Latest commit removes dependency on `Data.Nat.GeneralisedArithmetic.fold` in favour of `Algebra.Definitions.RawMonoid._×_` cf. #2446

jamesmckinna:
@JacquesCarette you wrote:
> On a bigger scale: this seems like a lot of duplication. Feels like one could abstract over the differences between `CommutativeMonoid/Normal` and `IdempotentCommutativeMonoid/Normal` instead of this copy-pasta. But that's an easy comment for me to make!!! I should actually try it.

See `Algebra.Solver.*Monoid.*` refactoring on branch [jamesmckinna:new-solver-refactor](https://github.com/jamesmckinna/agda-stdlib/tree/new-solver-refactor) for a first second go in this direction...

JacquesCarette (approved):

JacquesCarette:
This looks good to me.

I tried to find the second refactor on the branch you pointed to, but I wasn't quite sure what to look at (i.e. which commits are for this?). Maybe make a tentative PR for that as well, to make it easier to review?

jamesmckinna:
Given the additional subtleties involved in pursuing your further ideas regarding refactoring in #2457 , does it make sense to aim to merge this in the meantime?

JacquesCarette:
> [...] does it make sense to aim to merge this in the meantime?

As long as the additional refactor doesn't introduce compatibility issues, then yes, it makes sense to merge this in first.

jamesmckinna:
> > [...] does it make sense to aim to merge this in the meantime?
> 
> As long as the additional refactor doesn't introduce compatibility issues, then yes, it makes sense to merge this in first.

Well, that's a good question! I guess that the NormalAPI is nowhere exposed as such in the (existing) 'top level' `Algebra.Solver.*Monoid` modules, so as long as we feel free to change that API under the abstract `Tactic` interface, it should be OK, right? 

Or not!?

jamesmckinna:
And... as for "compatibility issues", I think there won't be any arising from #2457 ... given the (relative) care taken to preserve (even if to deprecate) the v1.*/v2.1 API ...

src/Algebra/Solver/Monoid/Tactic.agda
```diff
@@ -0,0 +1,52 @@
...
+--
+-- A solver for equations over monoids
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (Monoid)
+import Algebra.Solver.Monoid.Expression as Expression
+
+module Algebra.Solver.Monoid.Tactic {a c ℓ} (M : Monoid c ℓ)
```
MatthewDaggitt, line 12: Very minor, but I would prefer we didn't use the term `Tactic` outside of the top-level tactic folder as this does not export a tactic macro per se. I'm open to other suggestions, but we could call this `Implementation` or proof?

src/Algebra/Solver/Monoid/Tactic.agda
```diff
@@ -0,0 +1,52 @@
...
+  (open Monoid M using (rawMonoid; setoid; _≈_))
+  (open Expression rawMonoid using (Expr; Env; var; ⟦_⟧; NormalAPI))
+  (N : NormalAPI {a}) where
```
MatthewDaggitt, line 15: This rather suggests to me that the level parameter should be explicit?

src/Algebra/Solver/Monoid/Tactic.agda
```diff
@@ -0,0 +1,52 @@
...
+
+private
+  variable
+    n : ℕ
+
+
+------------------------------------------------------------------------
+-- Tactic
+
+-- We can also give a sound, but not necessarily complete, procedure
```
MatthewDaggitt, line 37: What does the `also` refer to here?

src/Algebra/Solver/Monoid/Tactic.agda
```diff
@@ -0,0 +1,52 @@
...
+
+open NormalAPI N
+open module R = Reflection setoid var ⟦_⟧ (⟦_⟧⇓ ∘ normalise) correct public
+  using (solve; _⊜_)
+
+private
+  variable
+    n : ℕ
+
+
```
MatthewDaggitt, line 33: One too many empty lines?

src/Algebra/Solver/Monoid/Tactic.agda
```diff
@@ -0,0 +1,52 @@
...
+open import Data.Maybe.Base as Maybe
+  using (Maybe; From-just; from-just)
+open import Data.Nat.Base using (ℕ)
+open import Function.Base using (_∘_; _$_)
+open import Relation.Binary.Consequences using (dec⇒weaklyDec)
+open import Relation.Binary.PropositionalEquality.Core using (_≡_; cong)
+import Relation.Binary.Reflection as Reflection
+
+open NormalAPI N
+open module R = Reflection setoid var ⟦_⟧ (⟦_⟧⇓ ∘ normalise) correct public
```
MatthewDaggitt, line 26: This module `R` is publicly exported? Should it be/is there a reason why we're naming it?

CHANGELOG.md
```diff
@@ -45,9 +62,26 @@ New modules
   Algebra.Properties.IdempotentCommutativeMonoid
   ```
 
+* Refactoring of the `Algebra.Solver.*Monoid` implementations, via
+  a single `Tactic` module API based on the existing `Expr`, and
+  a common `Normal`-form API:
```
MatthewDaggitt, line 67: I would add that you're still supposed to access the solvers through the existing files.

src/Algebra/Solver/CommutativeMonoid/Normal.agda
```diff
@@ -0,0 +1,140 @@
...
+-- Constructions on normal forms
+
+-- The empty bag.
+
+empty : Normal n
+empty = replicate _ 0
+
+-- A singleton bag.
+
+sg : (i : Fin n) → Normal n
```
MatthewDaggitt, line 73: Could we take this opportunity to deprecate this cryptic name and replace it with `singleton`?

src/Algebra/Solver/IdempotentCommutativeMonoid.agda
```diff
@@ -10,208 +10,35 @@
...
-sg : (i : Fin n) → Normal n
-sg zero    = true ∷ empty
-sg (suc i) = false ∷ sg i
-
--- The composition of normal forms.
-
-_•_ : (v w : Normal n) → Normal n
-[]      • []      = []
-(l ∷ v) • (m ∷ w) = (l ∨ m) ∷ v • w
+open module N = Normal M public
```
MatthewDaggitt, line 24: Again is there a reason we are deliberately exporting a module named `N`? (See other files as well)

src/Algebra/Solver/IdempotentCommutativeMonoid/Normal.agda
```diff
@@ -0,0 +1,157 @@
...
+sg-correct : (x : Fin n) (ρ : Env n) → ⟦ sg x ⟧⇓ ρ ≈ lookup ρ x
+sg-correct zero (x ∷ ρ) = begin
+    x ∙ ⟦ empty ⟧⇓ ρ   ≈⟨ ∙-congˡ (empty-correct ρ) ⟩
+    x ∙ ε              ≈⟨ identityʳ _ ⟩
+    x                  ∎
+sg-correct (suc x) (m ∷ ρ) = sg-correct x ρ
+
+-- Normal form composition corresponds to the composition of the monoid.
+
+distr : ∀ a b c → a ∙ (b ∙ c) ≈ (a ∙ b) ∙ (a ∙ c)
```
MatthewDaggitt, line 103: This feels like it should be a lemma in `Properties.IdempotentCommutativeMonoid`?

src/Algebra/Solver/IdempotentCommutativeMonoid/Normal.agda
```diff
@@ -0,0 +1,157 @@
...
+-- Constructions on normal forms
+
+-- The empty set.
+
+empty : Normal n
+empty = replicate _ false
+
+-- A singleton set.
+
+sg : (i : Fin n) → Normal n
```
MatthewDaggitt, line 73: Again `singleton`?

src/Algebra/Solver/Monoid.agda
```diff
@@ -6,133 +6,48 @@
...
 ------------------------------------------------------------------------
--- "Tactic.
-
-open module R = Relation.Binary.Reflection
-                  setoid var ⟦_⟧ (⟦_⟧⇓ ∘ normalise) normalise-correct
-  public using (solve; _⊜_)
+-- Tactic
 
--- We can decide if two normal forms are /syntactically/ equal.
+open module T = Tactic M (record { N }) public
```
MatthewDaggitt, line 30: Again named modules being exported?

MatthewDaggitt (changes requested):
This looks really good! Added a few small comments, many of which whose causes aren't directly introduced by you.

MatthewDaggitt:
After those are addressed, we should definitely merge!

src/Algebra/Solver/Monoid/Tactic.agda
jamesmckinna: `Solver`?

src/Algebra/Solver/Monoid/Tactic.agda
jamesmckinna: Comment copied over from original. Have rejigged the module contents to make the exports more explicit, and hopefully the "also" now makes more sense?

src/Algebra/Solver/Monoid/Tactic.agda
jamesmckinna: OK, I thought the `style-guide` recommended two after the initial `private` block`, but maybe that's been superseded now...

src/Algebra/Solver/IdempotentCommutativeMonoid/Normal.agda
jamesmckinna: It is now, but the PR that introduced that was submitted *after* this PR, but *merged* before... ;-)

src/Algebra/Solver/IdempotentCommutativeMonoid.agda
jamesmckinna: Fixed (I hope!)

src/Algebra/Solver/Monoid.agda
jamesmckinna: Ditto.

src/Algebra/Solver/Monoid/Tactic.agda
jamesmckinna: Sadly, this is legacy from the [original module API](https://github.com/agda/agda-stdlib/blob/355ac2548fb60adae218e32f5466160d15ecb1ef/src/Algebra/Solver/Monoid.agda#L109-L111)

src/Algebra/Solver/IdempotentCommutativeMonoid.agda
jamesmckinna: Fixed

jamesmckinna:
@MatthewDaggitt thanks for the very detailed feedback. Summarising:
* I've fixed the newly-introduced module exports to hopefully now not introduce new names, but the original export of `module R` in `Algebra.Solver.Monoid` is legacy, so leaving it for now.
* I've done the suggested deprecations

@JacquesCarette Against the discussion on #2457 , which I think can follow-on downstream, I still think this is worth merging now, as a refactoring-with-deprecations. The rest can follow later?

src/Algebra/Solver/Monoid/Tactic.agda
MatthewDaggitt: Oh sorry, I haven't seen that entry in the style guide before! That's a new one to me, apologies.

jamesmckinna:
Thanks @MatthewDaggitt for resolving the merge conflict... I'd have got there once I'd woken up... ;-)

jamesmckinna:
Apologies for inaccuracy in my summary of changes: it seems as though I *did* in fact make the `breaking` change of making the module `R` private in `Algebra.Solver.Monoid.Solver`, so it's *not* now exported publicly from  `Algebra.Solver.Monoid`.

I've suggested on the downstream issue #2472 that we take no action on this, but advice/counter-suggestions welcome!

PR #2409, opened 2024-06-15
Update LICENSE

lexvanderstoep:
Update the year to reflect the present.

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,4 +1,4 @@
-iCopyright (c) 2007-2024 Nils Anders Danielsson, Ulf Norell, Shin-Cheng
+Copyright (c) 2007-2024 Nils Anders Danielsson, Ulf Norell, Shin-Cheng
 Mu, Bradley Hardy, Samuel Bronson, Dan Doel, Patrik Jansson,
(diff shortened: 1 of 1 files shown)


lexvanderstoep:
Bit of a nit, but noticed the license was off.

LICENCE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2007-2023 Nils Anders Danielsson, Ulf Norell, Shin-Cheng
+iCopyright (c) 2007-2024 Nils Anders Danielsson, Ulf Norell, Shin-Cheng
```
gallais, line 1: What's with the `i`?

```suggestion
Copyright (c) 2007-2024 Nils Anders Danielsson, Ulf Norell, Shin-Cheng
```

jamesmckinna (changes requested):
Thanks for the nitpicking! 
Approve subject to following @gallais 's suggestion!

LICENCE
lexvanderstoep: I think that's a remnant of me using Emacs for the first time after using Vim for a while... Fixed.

jamesmckinna (approved):

PR #2410, opened 2024-06-16
Fix #2396 by removing redundant zero in IsNonAssociativeRing

### LINKED ISSUE #2396: [DRY] More redundant `zero` fields in `Algebra.Structures`
Cf. #2195 and #2253 

* `IsNonAssociativeRing` admits enough properties to make the `zero` field redundant by exactly the same argument as for `IsRingWithoutOne` (so the redundancy is a `bug`,  and removing  it would be `breaking`)
* correspondingly, the argument for `IsRingWithoutOne` makes no use of `*-assoc`...

Possible remedy: introduce a further refinement in which a new `IsNonAssociativeRingWithoutOne` would the 'right' home for such an argument, with each of the above two structures inheriting from that...?

lexvanderstoep:
The zero field in the IsNonAssociativeRing was redundant, and could be replaced with a proof based on the other properties.


jamesmckinna:
Tests show that replacing the (redundant) explicit field with a definition/manifest field is `breaking` behaviour... so we need to tread a bit more carefully!

Issues:
* knock-on consequences: need to update uses/definitions to accommodate the change
* `breaking` changes are generally being punted to v3.0, rather than any v2.x minor version
* needs `CHANGELOG`

lexvanderstoep:
Thanks for reviewing this so quickly! I was about to mark the PR as a draft to fix the usage of `IsNonAssociativeRing`, but you beat me to it.

About the change being `breaking` and only included in v3.0, makes sense. What is the way forward? Do we leave the MR until we're ready to release the next major version of the library?

jamesmckinna:
I'd definitely go with DRAFT for the time being...
I could well imagine some of the other maintainers weighing in on how best to take this forward... and when. But we're a bit more focused on clearing the v2.1 milestone, so this one will definitely be on the back burner for a bit, I think.

lexvanderstoep:
Sounds like a plan!

JacquesCarette (approved):
Approving these changes "on their own" but agree with @jamesmckinna about the breaking aspects.

jamesmckinna:
Re: the discussion on #2396 I'd be tempted to take this off `DRAFT` status now, as it's 'ready to go' (or at least 'ready to review' ;-))... the only problem being that between the next release and v3.0 (hopefully sooner rather than later, but we'll see in the new year), there'll be an almighty `CHANGELOG` merge conflict, but should be easy to resolve in place then.

lexvanderstoep:
@jamesmckinna Updated the PR to be 'ready to go'. I'll tackle the almighty `CHANGELOG` merge conflict when the time comes...

MatthewDaggitt:
Merged this is in and think it is ready to go!

MatthewDaggitt (approved):

gallais (approved):

MatthewDaggitt:
Finally merging, long overdue. Thanks very much @lexvanderstoep!

gallais (approved):

lexvanderstoep:
Thanks for merging! @MatthewDaggitt

PR #2413, opened 2024-06-19
fixes #2411

### LINKED ISSUE #2411: Naming of new functions `tail∘inits` and `tail∘tails`
Just going through tidying up the CHANGELOG and came across PR https://github.com/agda/agda-stdlib/pull/2395 which introduced `tail∘inits` and `tail∘tails`.

This is the first time ever we've used unicode names for functions, and has potential to disrupt naming schemes in the `Properties` files.

Are we sure we want to do this? Could we come up with more conventional names for these functions? For example: 
 - `tailInits`/`tailTails`

jamesmckinna:
NB. I haven't tried doing this with local named module defining an auxiliary function cf. #2411 because the scoping of such modules means that I end up having to write:
```agda
inits : List A → List (List A)
inits {A = A} = ([] ∷_) ∘ tail
  module Inits where
    tail : List A → List (List A)
    tail []       = []
    tail (x ∷ xs) = [ x ] ∷ map (x ∷_) (tail xs)

tails {A = A} = λ (xs : List A) → xs ∷ tail xs
  module Tails where
    tail : List A → List (List A)
    tail []       = []
    tail (_ ∷ xs) = xs ∷ tail xs
```
which isn't obviously more readable/intelligible because of the artificiality of introducing an explicit `λ ...` on the RHS?

UPDATED: have gone with (a cleaned-up version of) this style now!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -498,8 +498,8 @@ Additions to existing modules
 
 * In `Data.List.Base` redefine `inits` and `tails` in terms of:
   ```agda
-  initsTail : List A → List (List A)
-  tailsTail : List A → List (List A)
+  Inits.tail : List A → List (List A)
+  Tails.tail : List A → List (List A)
   ```
 
 * In `Data.List.Membership.Propositional.Properties.Core`:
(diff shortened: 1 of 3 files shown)


JacquesCarette (approved):

MatthewDaggitt:
Thanks, merging in!

Taneb (approved):

MatthewDaggitt (approved):

PR #2414, opened 2024-06-21
Add a property relating `maybe` to composition

WhatisRT:
This is part of upstreaming properties that seem generally useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,11 @@ Additions to existing modules
   ++⁺ˡ : Reflexive R → ∀ zs → (_++ zs) Preserves (Pointwise R) ⟶ (Pointwise R)
   ```
 
+* In `Data.Maybe.Properties`:
+  ```agda
+  maybe′-∘ : ∀ {b} (f : B → C) (g : A → B) → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
+  ```
+
 * New lemmas in `Data.Nat.Properties`:
   ```agda
   m≤n⇒m≤n*o : ∀ o .{{_ : NonZero o}} → m ≤ n → m ≤ n * o
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@ Additions to existing modules
 
 * In `Data.Maybe.Properties`:
   ```agda
-  maybe′-∘ : ∀ {b} (f : B → C) (g : A → B) → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
+  maybe′-∘ : ∀ f g → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
   ```
 
 * New lemmas in `Data.Nat.Properties`:
--- a/src/Data/Maybe/Properties.agda
+++ b/src/Data/Maybe/Properties.agda
@@ -97,6 +97,10 @@ maybe′-map : ∀ j (n : C) (f : A → B) ma →
              maybe′ j n (map f ma) ≡ maybe′ (j ∘′ f) n ma
 maybe′-map = maybe-map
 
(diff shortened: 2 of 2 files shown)


JacquesCarette (changes requested):
You'll need to add to CHANGELOG too. But it is a nice little lemma that does seem to be missing, AFAIK.

src/Data/Maybe/Properties.agda
```diff
@@ -97,6 +97,10 @@ maybe′-map : ∀ j (n : C) (f : A → B) ma →
              maybe′ j n (map f ma) ≡ maybe′ (j ∘′ f) n ma
 maybe′-map = maybe-map
 
+maybe-∘ : ∀ {c x} (f : B → C) (g : A → B) → f (maybe g c x) ≡ maybe (f ∘ g) (f c) x
+maybe-∘ {x = just _}  _ _ = refl
```
jamesmckinna, line 101: Pattern-matching on an implicit argument like this doesn't quite smell right... and in any case, this lemma can probably (better) be given a direct one-line definition, again in terms of `maybe`...?

jamesmckinna: Also, on the model of `maybe′-map`, should the lemma be stated (and named) explicitly involving `maybe′` rather than `maybe`?

jamesmckinna:
And as for `CHANGELOG`... suggest this wait until this has been reset for the v2.2 milestone after the planned v2.1 release...

src/Data/Maybe/Properties.agda
WhatisRT: Here's a proof that uses `maybe` instead of matching. I think it's clearly worse, and I don't think it can be reduced further:
```
maybe-∘ {x = x} f _ = maybe {B = λ y → f (maybe _ _ y) ≡ maybe _ _ y} (λ _ → refl) refl x
```
I don't mind converting `x` to an explicit argument, but there's plenty of good reasons to match on a hidden one, so I don't see that as a reason to do so.

WhatisRT: Hmm, I'm not sure it's worth weakening the resulting statement just so that the proof is a bit simpler. And arguably it's not even actually simpler (up to inlining), it's just a difference in the surface language code.

jamesmckinna: Not sure I understand how this is 'weaker' as a statement? In any downstream client use where `f` and/or `g` might depend on a given `x`, then appropriate arguments to the lemma can still be supplied, or not? Their *types* don't depend on `x`, so each of the arguments is 'independent' in the telescope?

WhatisRT: Ah, yes, I was confused. We're not actually using the more general type of `maybe` here.

WhatisRT:
Done

CHANGELOG.md
```diff
@@ -71,6 +71,11 @@ Additions to existing modules
   ++⁺ˡ : Reflexive R → ∀ zs → (_++ zs) Preserves (Pointwise R) ⟶ (Pointwise R)
   ```
 
+* In `Data.Maybe.Properties`:
+  ```agda
+  maybe′-∘ : ∀ {b} (f : B → C) (g : A → B) → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
```
jamesmckinna, line 76: Sorry to miss this earlier: we generally suppress leading quantifier prefixes where they can (easily ;-)) be inferred directly, or by reference to the actual code, so
```suggestion
  maybe′-∘ : ∀ f g → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
```
should suffice!

jamesmckinna:
Looks great! otherwise sorry for the last nitpick re `CHANGELOG`... :-(

JacquesCarette (approved):

Issue #2415, opened 2024-06-21
comments on lib-2.1 candidate 1 


mechvel:
(1) Visibility of  Data.Product.zipWith

Checking ForTests (/home/mechvel/inAgda/doconA/3.2/source/ForTests.agda).
Multiple definitions of  zipWith.
Previous definition at
/home/mechvel/agda/stLib/2.1-rc1/src/Data/List/Base.agda:83,1-8
when scope checking the declaration
  open import Data.Product public hiding (map; zip)

Is this backwards compatible?

(2) What is the matter with
    Data.List.Properties.filter-accept, filter-reject ?

Unlike with lib-2.0, it appears "Unsolved", and it requires implicit
arguments to be set
- this is under the same type checker
  Agda 2.6.4.3  Built with flags (cabal -f)
  - optimise-heavily,
  the type check keys are  "--auto-inline --guardedness"

And it occurs non-backwards compatible.
The effect often occurs that Agda silently hangs forever, then one has to guess and search for ``filter-accept`` in the currently processed module.
My program has probably thousand of places where the implicit arguments ``filter-accept/reject `` need to be inserted.
I wonder of why the effect is produced while the type checker version is the same and the function definition is the same.

Can you explain the effect?

(3)  ``reverse`` and `⊆`.

lib-2.1-rc1 has in Data.List.Relation.Binary.Subset.Setoid.Properties
```
reverse-selfAdjoint : as ⊆ reverse bs → reverse as ⊆ bs
reverse⁺            : as ⊆ bs → reverse as ⊆ reverse bs
reverse⁻            : reverse as ⊆ reverse bs → as ⊆ bs
```
But there are also needed
```
reverse-preservesʳ-⊆ :  ∀ {xs} → rev Preserves (xs ⊆_)
reverse-preservesʳ-⊆ {xs} {ys} xs⊆ys =  xs⊆rev-ys
  where
  xs⊆rev-ys : xs ⊆ (rev ys)
  xs⊆rev-ys = reverse∈⁺ S ∘ xs⊆ys

reverse-preservesˡ-⊆ :  ∀ {xs} → rev Preserves (_⊆ xs)
reverse-preservesˡ-⊆ {xs} {ys} ys⊆xs =  rev-ys⊆xs
  where
  rev-ys⊆xs : (rev ys) ⊆ xs
  rev-ys⊆xs {z} z∈rev-ys = ys⊆xs z∈ys
    where
[...]

PR #2418, opened 2024-06-24
Add proofs on truth value

mildsunrise:
Adds a handful of "congruence" proofs on the truth value of a decidable, and also `≐?`:

- `Relation.Nullary.Decidable`:
  - `does-≡`: decidables over the same type have the same truth value
  - `does-⇔`: generalization of `does-≡` to mutually implied types

- `Relation.Unary.Properties`:
  - `≐?`: generalization of `Decidable.map` to predicates (if two predicates are equivalent, one is decidable if the other is)
  - `does-≡`: generalization of `does-≡` to predicates
  - `does-≐`: generalization of `does-⇔` to predicates

Using these, together with the definitions in `Decidable.Core`, one can prove the effect of type operations on truth value. For example, proving that `Dec (¬ A)` and `Dec A` have negated truth values can be done as `does-≡ ¬a? (¬? a?)`. Same with `×` / `∧` and `⊎` / `∨`.


src/Relation/Nullary/Decidable.agda
```diff
@@ -80,3 +81,10 @@ dec-yes-irr a? irr a with a′ , eq ← dec-yes a? a rewrite irr a a′ = eq
 
 ⌊⌋-map′ : ∀ t f (a? : Dec A) → ⌊ map′ {B = B} t f a? ⌋ ≡ ⌊ a? ⌋
 ⌊⌋-map′ t f a? = trans (isYes≗does (map′ t f a?)) (sym (isYes≗does a?))
+
+does-⇔ : A ⇔ B → (a? : Dec A) → (b? : Dec B) → does a? ≡ does b?
+does-⇔ A⇔B a? (yes b) = dec-true a? (Equivalence.from A⇔B b)
+does-⇔ A⇔B a? (no ¬b) = dec-false a? (¬b ∘ Equivalence.to A⇔B)
+
+does-≡ : (a? b? : Dec A) → does a? ≡ does b?
+does-≡ = does-⇔ (mk⇔ id id)
```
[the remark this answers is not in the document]
mildsunrise: thanks, done!

MatthewDaggitt (approved):

gallais (approved):

mildsunrise:
(I've simplified the proofs, but the statements remain unchanged)

src/Relation/Unary/Properties.agda
```diff
@@ -18,9 +17,8 @@ open import Relation.Binary.Definitions
   hiding (Decidable; Universal; Irrelevant; Empty)
 open import Relation.Binary.PropositionalEquality.Core using (refl; _≗_)
 open import Relation.Unary
-open import Relation.Nullary.Decidable using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does; does-⇔)
+open import Relation.Nullary.Decidable using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does) renaming (does-≡ to does-≡′)
```
jamesmckinna, line 20: Would (better?) suggest making this a qualified import to avoid the `renaming` directive cf. #2308

src/Relation/Nullary/Decidable.agda
```diff
@@ -80,3 +80,10 @@ dec-yes-irr a? irr a with a′ , eq ← dec-yes a? a rewrite irr a a′ = eq
 
 ⌊⌋-map′ : ∀ t f (a? : Dec A) → ⌊ map′ {B = B} t f a? ⌋ ≡ ⌊ a? ⌋
 ⌊⌋-map′ t f a? = trans (isYes≗does (map′ t f a?)) (sym (isYes≗does a?))
+
+does-≡ : (a? b? : Dec A) → does a? ≡ does b?
+does-≡ a? (yes a) = dec-true a? a
+does-≡ a? (no ¬a) = dec-false a? ¬a
```
jamesmckinna, line 86: I would always be tempted, where possible, to make such definitions lazier wrt their `Dec A` arguments, as follows:
```suggestion
does-≡ a? (true because [a]) = dec-true a? (invert [a])
does-≡ a? (false because [¬a]) = dec-false a? (invert [¬a])
```
because, in general, `yes` and `no` are unnecessarily strict as constructors.

mildsunrise: hmm, I see. your concern is that `yes` / `no` makes the function strict on the Reflects, right? I can see the advantage of your proposed code, but the downside is making the proof more inscrutable to the casual reader... and bearing in mind this is computing an identity, not a "useful" value as in `Data.List.Base`, I'm not sure this extra lazyness matters here?

(I wish Agda had lazy patterns like Haskell... :pensive:)

src/Relation/Unary/Properties.agda
mildsunrise: ah, thanks a lot, totally missed that

src/Relation/Nullary/Decidable.agda
jamesmckinna: It's a fair point... so I'm happy to give way on this one. But long term I would like to refactor `Dec` to avoid `yes` and `no` entirely in favour of such lazier counterparts... something I began with #2055 and subsequent PRs such as #2149 

But the recent issue #2415 and its retrospective consequences for #2359 / #2365 suggest that a bit more thinking is (still) required. 

Thanks for at least considering the suggestion. And as for the 'casual' reader of library source, there are certainly interesting trade-offs to be made between such an audience, and that of developers/maintainers... as I have learnt over the last 3 years ;-)

src/Relation/Unary/Properties.agda
```diff
@@ -17,7 +17,8 @@ open import Relation.Binary.Definitions
   hiding (Decidable; Universal; Irrelevant; Empty)
 open import Relation.Binary.PropositionalEquality.Core using (refl; _≗_)
 open import Relation.Unary
-open import Relation.Nullary.Decidable using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does) renaming (does-≡ to does-≡′)
+import Relation.Nullary.Decidable as Dec
+open import Relation.Nullary.Decidable using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does)
```
jamesmckinna, line 21: You can telescope this with the preceding line simply as:
```suggestion
open import Relation.Nullary.Decidable as Dec using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does)
```
and kill two birds with one stone... :-)

mildsunrise: thanks again :)

src/Relation/Nullary/Decidable.agda
mildsunrise: yeah, there's certainly a tradeoff... I feel like I could be convinced either way tbh

src/Relation/Unary/Properties.agda
```diff
@@ -15,9 +15,10 @@ open import Level using (Level)
 open import Relation.Binary.Core as Binary
 open import Relation.Binary.Definitions
   hiding (Decidable; Universal; Irrelevant; Empty)
-open import Relation.Binary.PropositionalEquality.Core using (refl)
+open import Relation.Binary.PropositionalEquality.Core using (refl; _≗_)
 open import Relation.Unary
-open import Relation.Nullary.Decidable using (yes; no; _⊎-dec_; _×-dec_; ¬?)
+import Relation.Nullary.Decidable as Dec
+open import Relation.Nullary.Decidable as Dec using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does)
```
jamesmckinna, line 21: Apologies for returning to this after a few days' away.

L20 is now redundant, being superseded/'overridden' by L21, so there are several possible tweaks to eliminate such redundancy:
* if we want to preserve the distinction between L20 and L21, by virtue of the fact that L21 imports from `Decidable.Core` (reexported publicly by `Decidable`), then change L21 to the narrower import
* if we want to preserve L20 in its own right, then we could 'optimise' L21 simply to `open Dec using...`, since that local/qualified module name is after L20 already in scope
* but, to my mind, simplest and best of all, would simply be to delete L20, leaving just
```suggestion
open import Relation.Nullary.Decidable as Dec using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does)
```

src/Relation/Unary/Properties.agda
```diff
@@ -200,6 +201,10 @@ U-Universal = λ _ → _
 ------------------------------------------------------------------------
 -- Decidability properties
 
+≐? : {P : Pred A ℓ₁} {Q : Pred A ℓ₂} →
+     P ≐ Q → Decidable P → Decidable Q
+≐? (P⊆Q , Q⊆P) P? x = map′ P⊆Q Q⊆P (P? x)
```
jamesmckinna, line 206: Two comments about this lemma:
* name: not quite sure about the name, but I can't think of the better at the moment
* definition: Can this be more 'simply' (compactly) expressed in terms of `Dec.map`? Delegation back to the conceptually prior, and differently 'simpler' `map′` might feel like a win, but as we are already in a (module-)scope in which `Dec.map` is defined and available, it seems 'better' to use it? (the heavy use of scare quotes is intentional in drawing attention to my ambivalence about the range of choices here)

The second point amounts really to saying: this is the 'obvious' lifting to `Unary` predicates of `Dec.map` on `Nullary` propositions... so "for the casual reader" it might be better to make that evident...?

jamesmckinna: yeah... not sure what name I'd choose, but fortunately, blocking on the v2.1 release means that we collectively have plenty of time to think about this ;-)

JacquesCarette: I get strong `map` vibes from this. The current names makes me think that I'll be deciding an equality between functions.

JacquesCarette (changes requested):
Other than the one name choice, this is great.

src/Relation/Unary/Properties.agda
mildsunrise: let's go with `map` then

MatthewDaggitt:
@JacquesCarette  are you willing to approve now?

JacquesCarette (approved):

CHANGELOG.md
```diff
@@ -53,3 +53,16 @@ Additions to existing modules
...
   _≡?_ : DecidableEquality (Vec A n)
   ```
+
+* In `Relation.Nullary.Decidable`:
+  ```agda
+  does-⇔  : A ⇔ B → (a? : Dec A) → (b? : Dec B) → does a? ≡ does b?
+  does-≡  : (a? b? : Dec A) → does a? ≡ does b?
+  ```
+
+* In `Relation.Nullary.Properties`:
```
jamesmckinna, line 63: We missed this (copy/paste error?) during review: should be `Relation.Unary.Properties`!

MatthewDaggitt: Nice spot! Any chance of a quick PR?

jamesmckinna: Well, I have already fixed it on (a couple of ;-)) open branches as part of merge conflict resolution, so the alternative would be to merge those... but if you want a cleaner fine-grained git history, happy to do so.

PR #2419, opened 2024-06-24
Add some type relations and isomorphisms

WhatisRT:
Upstreaming some functions that I've used to prove things about sets (not the Agda kind).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,11 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Data.Product.Function.Dependent.Propositional`:
+  ```agda
+  cong′ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
+  ```
+
 * In `Data.List.Properties`:
   ```agda
   product≢0 : All NonZero ns → NonZero (product ns)
@@ -92,6 +97,15 @@ Additions to existing modules
   ```agda
   _≡?_ : DecidableEquality (Vec A n)
   ```
+
+* In `Function.Related.TypeIsomorphisms`:
+  ```agda
+  Σ-distribˡ-⊎ : {P : A → Set a} {Q : A → Set b} → (∃ λ a → P a ⊎ Q a) ↔ (∃ P ⊎ ∃ Q)
+  Σ-distribʳ-⊎ : {P : Set a} {Q : Set b} {R : P ⊎ Q → Set c} → (Σ (P ⊎ Q) R) ↔ (Σ P (R ∘ inj₁) ⊎ Σ Q (R ∘ inj₂))
+  ×-distribˡ-⊎′ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
+  ×-distribʳ-⊎′ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
+  ∃-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
+ ```
 
 * In `Relation.Nullary.Decidable`:
   ```agda
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
 
-In `Function.Related.TypeIsomorphisms`, the unprimed versions are more level polymorphic; and the primed versions retain `Level` homogeneous types for the `Semiring` axioms to hold. 
+In `Function.Related.TypeIsomorphisms`, the unprimed versions are more level polymorphic; and the primed versions retain `Level` homogeneous types for the `Semiring` axioms to hold.
 
 Minor improvements
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -314,7 +314,7 @@ Additions to existing modules
 
 * In `Data.Product.Function.Dependent.Propositional`:
   ```agda
-  cong′ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
+  congˡ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
   ```
 
 * New lemma in `Data.Vec.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
 
+In `Function.Related.TypeIsomorphisms`, the unprimed versions are more level polymorphic; and the primed versions retain `Level` homogeneous types for the `Semiring` axioms to hold. 
+
 Minor improvements
 ------------------
 
@@ -329,10 +331,10 @@ Additions to existing modules
 
 * In `Function.Related.TypeIsomorphisms`:
   ```agda
-  Σ-distribˡ-⊎ : {P : A → Set a} {Q : A → Set b} → (∃ λ a → P a ⊎ Q a) ↔ (∃ P ⊎ ∃ Q)
-  Σ-distribʳ-⊎ : {P : Set a} {Q : Set b} {R : P ⊎ Q → Set c} → (Σ (P ⊎ Q) R) ↔ (Σ P (R ∘ inj₁) ⊎ Σ Q (R ∘ inj₂))
-  ×-distribˡ-⊎′ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
-  ×-distribʳ-⊎′ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
+  Σ-distribˡ-⊎ : (∃ λ a → P a ⊎ Q a) ↔ (∃ P ⊎ ∃ Q)
+  Σ-distribʳ-⊎ : (Σ (A ⊎ B) P) ↔ (Σ A (P ∘ inj₁) ⊎ Σ B (P ∘ inj₂))
+  ×-distribˡ-⊎ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
+  ×-distribʳ-⊎ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
   ∃-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
  ```
 
--- a/src/Data/Product/Function/Dependent/Propositional.agda
+++ b/src/Data/Product/Function/Dependent/Propositional.agda
@@ -17,7 +17,7 @@ open import Function.Related.Propositional
          implication; reverseImplication; equivalence; injection;
          reverseInjection; leftInverse; surjection; bijection)
 open import Function.Base using (_$_; _∘_; _∘′_)
-open import Function.Properties.Inverse using (↔⇒↠; ↔⇒⟶; ↔⇒⟵; ↔-sym; ↔⇒↩)
+open import Function.Properties.Inverse using (↔⇒↠; ↔⇒⟶; ↔⇒⟵; ↔-sym; ↔⇒↩; refl)
 open import Function.Properties.RightInverse using (↩⇒↪; ↪⇒↩)
 open import Function.Properties.Inverse.HalfAdjointEquivalence
   using (↔⇒≃; _≃_; ≃⇒↔)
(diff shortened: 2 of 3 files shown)


src/Data/Product/Function/Dependent/Propositional.agda
```diff
@@ -316,3 +316,8 @@ cong {B = B} {k = reverseInjection}   I↔J A↢B = Σ-↣ (↔-sym I↔J) (swap
 cong {B = B} {k = leftInverse}        I↔J A↩B = ↩⇒↪ (Σ-↩ (↔⇒↩ (↔-sym I↔J)) (↪⇒↩ (swap-coercions {k = leftInverse} B I↔J A↩B)))
 cong {k = surjection}                 I↔J A↠B = Σ-↠ (↔⇒↠ I↔J) A↠B
 cong {k = bijection}                  I↔J A↔B = Σ-↔ I↔J A↔B
+
+cong′ : ∀ {k} → {A : Set a} {B₁ : A → Set b} {B₂ : A → Set c} →
```
MatthewDaggitt, line 320: Can we use `I` for the index type instead of `A`, and `A` and `B` instead of `B₁`/`B₂`. Then we can use the variables and not have to add them to the type directly.

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -332,3 +352,11 @@ True↔ ( true because  [p]) irr =
   mk↔ₛ′ (λ _ → invert [p]) (λ _ → _) (irr _) (λ _ → refl)
 True↔ (false because ofⁿ ¬p) _ =
   mk↔ₛ′ (λ()) (invert (ofⁿ ¬p)) (λ x → flip contradiction ¬p x) (λ ())
+
+------------------------------------------------------------------------
+-- Relating a predicate to an existentially quantified one with the
+-- restriction that the quantified variable is equal to the given one
+
+∃-≡ : ∀ {A : Set a} (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
```
MatthewDaggitt, line 360: Again `A` is already a variable so doesn't need to be declared.

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -106,24 +106,44 @@ private
...
   (uncurry [ (λ _ _ → refl) , (λ _ _ → refl) ])
 
+------------------------------------------------------------------------
+-- Properties of × and ⊎
+
+-- × distributes over ⊎
+-- primed variants are more level polymorphic
+
+×-distribˡ-⊎′ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
+×-distribˡ-⊎′ = Σ-distribˡ-⊎
```
MatthewDaggitt, line 136: Following the convention of `Function.Base` then the primes should be on the less polymorphic lemmas rather than the more polymorphic lemmas?

MatthewDaggitt (changes requested):

MatthewDaggitt:
Ah, and a CHANGELOG entry would be great!

src/Function/Related/TypeIsomorphisms.agda
WhatisRT: Yes, I wasn't sure about backwards compatibility. If you think it's fine I'll swap the two. Do you think `×-distrib-⊎` should also be primed for consistency, even though it probably doesn't make sense to add a non-primed version?

WhatisRT:
I'll add the CHANGELOG once 2.1 is out.

src/Function/Related/TypeIsomorphisms.agda
JacquesCarette: I think `×-distrib-⊎` should stay as is because it is already non-dependent because it uses `_×_` rather than `∃`. The additional advantage is that this doesn't have backward compatibility problems!

jamesmckinna:
@WhatisRT any progress on the suggested changes? This should be easy to merge once you've done that...

WhatisRT:
I've added the changes to the CHANGELOG. I don't really understand what's the final verdict on the naming, should I change some of the names or not?

src/Data/Product/Function/Dependent/Propositional.agda
```diff
@@ -316,3 +316,6 @@ cong {B = B} {k = reverseInjection}   I↔J A↢B = Σ-↣ (↔-sym I↔J) (swap
 cong {B = B} {k = leftInverse}        I↔J A↩B = ↩⇒↪ (Σ-↩ (↔⇒↩ (↔-sym I↔J)) (↪⇒↩ (swap-coercions {k = leftInverse} B I↔J A↩B)))
 cong {k = surjection}                 I↔J A↠B = Σ-↠ (↔⇒↠ I↔J) A↠B
 cong {k = bijection}                  I↔J A↔B = Σ-↔ I↔J A↔B
+
+cong′ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
+cong′ = cong (refl _)
```
jamesmckinna, line 321: By analogy with the derived forms for `∙-cong` for a binary operator under `Algebra.Structures`, I think this should better be called `congˡ`?

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -106,24 +106,44 @@ private
...
+  (uncurry [ curry inj₁ , curry inj₂ ])
+  [ Product.dmap inj₁ id , Product.dmap inj₂ id ]
   [ (λ _ → refl) , (λ _ → refl) ]
   (uncurry [ (λ _ _ → refl) , (λ _ _ → refl) ])
 
+------------------------------------------------------------------------
+-- Properties of × and ⊎
+
+-- × distributes over ⊎
+-- primed variants are more level polymorphic
```
jamesmckinna, line 133: I'd simply remove this line (L133) as both incorrect, and out-of-scope, for this addition!

src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -332,3 +352,11 @@ True↔ ( true because  [p]) irr =
   mk↔ₛ′ (λ _ → invert [p]) (λ _ → _) (irr _) (λ _ → refl)
 True↔ (false because ofⁿ ¬p) _ =
   mk↔ₛ′ (λ()) (invert (ofⁿ ¬p)) (λ x → flip contradiction ¬p x) (λ ())
+
+------------------------------------------------------------------------
+-- Relating a predicate to an existentially quantified one with the
+-- restriction that the quantified variable is equal to the given one
+
+∃-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
+∃-≡ P {x} = mk↔ₛ′ (λ Px → x , refl , Px) (λ where (_ , (refl , Py)) → Py)
```
jamesmckinna, line 361: The nested parentheses `(_ , (refl , Py))` in the pattern are superfluous/redundant here?
```suggestion
∃-≡ P {x} = mk↔ₛ′ (λ Px → x , refl , Px) (λ where (_ , refl , Py) → Py)
```

JacquesCarette:
Yes, some of the names do need changing.

CHANGELOG.md
```diff
@@ -93,6 +98,15 @@ Additions to existing modules
   _≡?_ : DecidableEquality (Vec A n)
   ```
 
+* In `Function.Related.TypeIsomorphisms`:
+  ```agda
+  Σ-distribˡ-⊎ : {P : A → Set a} {Q : A → Set b} → (∃ λ a → P a ⊎ Q a) ↔ (∃ P ⊎ ∃ Q)
+  Σ-distribʳ-⊎ : {P : Set a} {Q : Set b} {R : P ⊎ Q → Set c} → (Σ (P ⊎ Q) R) ↔ (Σ P (R ∘ inj₁) ⊎ Σ Q (R ∘ inj₂))
+  ×-distribˡ-⊎′ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
+  ×-distribʳ-⊎′ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
+  ∃-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
```
jamesmckinna, line 107: Is there also a place for the *right* Kan extension version ('Yoneda') of this lemma (characterising the *left* Kan extension of `P` along `_≡_`, cf. 'coYoneda'), viz:
```agda
∀-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∀ {y} → y ≡ x → P y)
```

jamesmckinna: Oh... the R-to-L inverse map seems to need `Extensionality` and/or (?) `with-K`? But L-to-R works fine... so can the statement be suitably weakened (I'm not up-to-speed with this bit of the library and all the `*Inverse*` definitions...

jamesmckinna: Postpone to a downstream PR!

jamesmckinna:
I think that it would be good to try to merge this sooner rather than later, and AFAICT, it's down to agreeing names for the new additions... so a re-review seems helpful!?

JacquesCarette:
Looks like the master merge broke the build?

jamesmckinna:
> Looks like the master merge broke the build?

Nope... it's the rebuild of `alex` and `happy` which fails even before the attempted `master` merge is tested...

MatthewDaggitt:
Reading the comments, it looks like we have got consensus on the names?

jamesmckinna:
> Reading the comments, it looks like we have got consensus on the names?

Yes, I'd been waiting on @WhatisRT to update this PR, but if you're keen to merge now (and for sure, this PR prompted me to think harder about some other type isos which might usefully be added downsream, esp. as regards refactoring `Data.List.Relation.Binary.BagAndSetEquality` (again!)), I'd be happy to push updates from my end...

src/Data/Product/Function/Dependent/Propositional.agda
jamesmckinna, line 321:
```suggestion
congˡ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
congˡ = cong (refl _)
```

src/Function/Related/TypeIsomorphisms.agda
jamesmckinna, line 133:
```suggestion
```

src/Function/Related/TypeIsomorphisms.agda
jamesmckinna, line 136:
```suggestion
×-distribˡ-⊎ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
×-distribˡ-⊎ = Σ-distribˡ-⊎
```

CHANGELOG.md
```diff
@@ -312,6 +312,11 @@ Additions to existing modules
   m≤pred[n]⇒suc[m]≤n : .{{NonZero n}} → m ≤ pred n → suc m ≤ n
   ```
 
+* In `Data.Product.Function.Dependent.Propositional`:
+  ```agda
+  cong′ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
```
jamesmckinna, line 317:
```suggestion
  congˡ : ∀ {k} → (∀ {x} → A x ∼[ k ] B x) → Σ I A ∼[ k ] Σ I B
```

CHANGELOG.md
```diff
@@ -322,6 +327,15 @@ Additions to existing modules
   _≡?_ : DecidableEquality (Vec A n)
   ```
 
+* In `Function.Related.TypeIsomorphisms`:
+  ```agda
+  Σ-distribˡ-⊎ : {P : A → Set a} {Q : A → Set b} → (∃ λ a → P a ⊎ Q a) ↔ (∃ P ⊎ ∃ Q)
+  Σ-distribʳ-⊎ : {P : Set a} {Q : Set b} {R : P ⊎ Q → Set c} → (Σ (P ⊎ Q) R) ↔ (Σ P (R ∘ inj₁) ⊎ Σ Q (R ∘ inj₂))
+  ×-distribˡ-⊎′ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
+  ×-distribʳ-⊎′ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
```
jamesmckinna, line 335:
```suggestion
  ×-distribˡ-⊎ : (A × (B ⊎ C)) ↔ (A × B ⊎ A × C)
  ×-distribʳ-⊎ : ((A ⊎ B) × C) ↔ (A × C ⊎ B × C)
```

src/Function/Related/TypeIsomorphisms.agda
jamesmckinna: Apologies for the recent flurry of suggestion here. I realise that I wasn't quite understanding how this PR adds more polymorphic versions, and reimplements the prior functionality in these terms, but I'd be tempted to suggest that we 'just' take the more polymorphic versions... would this be a `breaking` change? Or potentially give rise to unsolved metas?

WhatisRT:
Sorry, I think I missed some of the naming suggestions. But it seems you all have way stronger opinions about that than I do so I'd rather let you do the final adjustments.

jamesmckinna:
I think if @JacquesCarette or @MatthewDaggitt can resolve the questions about level polymorphism and its impact on names (I managed to confuse myself too much over this to do so :-() then following @WhatisRT 's most recent comment, then you should make those changes yourselves and merge!

jamesmckinna:
Looking again at all this: the less/insufficiently polymorphic versions of the distributive laws are *also* present in `Data.Product.Algebra`, so it's tempting to argue for making the (possibly) `breaking` change of making `Function.Related.TypeIsomorphisms` carry only the more polymorphic version, and re-implement the less polymorphic versions in `Data.Product.Algebra` using that functionality? (Or vice versa?)
Also (but separate from this PR): in each module, `assoc` for each of sum and (non-dependent) product are less insufficiently polymorphic... suggesting that the library design might benefit from larger scale rethinking of these questions... or have I misunderstood something fundamental to what's going on here?

jamesmckinna:
How should we proceed with this? Merge as is, and refactor later?

JacquesCarette:
Requested changes from @MatthewDaggitt have been done, as verified by @gallais @jamesmckinna and I. Merging.

gallais (approved):

JacquesCarette (approved):

PR #2422, opened 2024-07-01
fixes #2408

### LINKED ISSUE #2408: Add `Algebra.Properties.IdempotentCommutativeMonoid`
Cf. #2138 / #2402 as well as #2403 / #2407 together with
* `Algebra.Properties.IsBand`
* etc.
Eg `Algebra.Solver.IdempotentCommutativeMonoid.distr` could/should be lifted out ... what else though?

jamesmckinna:
NB:
* UPDATED: `CHANGELOG` added (but merge conflicts until v2.1 is released)
* no `public` re-export of properties of `(Commutative)Monoid` inheriting from `(Commutative)Semigroup` (and perhaps this is a downstream issue to rationalise the hierarchy of `Algebra.Properties.*` to better reflect such inheritance hierarchies...?)
* knock-on consequences for #2407 /downstream (whether or not this is merged first)...

Additionally (but out of scope here):
* possible other [DRY] redundancies/refactoring opportunities exposed in `Algebra.Properties.CommutativeSemigroup` and elsewhere (eg. the `semimedial` properties are all instances of `interchange`; the pattern `sym (assoc x y z)` is already defined by `Algebra.Properties.Semigroup.x∙yz≈xy∙z` etc.)


JacquesCarette (approved):

src/Algebra/Properties/IdempotentCommutativeMonoid.agda
```diff
@@ -0,0 +1,38 @@
...
+  using (_DistributesOverˡ_; _DistributesOverʳ_; _DistributesOver_)
+open import Algebra.Properties.CommutativeSemigroup commutativeSemigroup
+  using (interchange)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+
+------------------------------------------------------------------------
+-- Distributivity
+
+distrˡ : _∙_ DistributesOverˡ _∙_
```
MatthewDaggitt, line 28: I'm wondering whether we should make it explicit in the name that the operation is distributing over itself, i.e. `∙-distrˡ-∙`? It's somewhat unusual to have an operator related to itself, so it might reduce user surprise and increase readability?

jamesmckinna: Pro: the argument you offer
Contra: `IdempotentCommutativeMonoid` only has 1 binary operation, so (albeit implicitly) there is only one 'distributivity' available/expressible?

As for relating operators to themselves, `interchange : Interchangable _∙_ _∙_` under `CommutativeSemigroup` provides a counter-example to your suggested naming, unless you think that that, too, should be `∙-interchange-∙`?

The originating issue/PR #2403 / #2407 previously had `distr` for what was clearly a *left* distributivity, so all I had done was rectify the orientation of that name, and then flesh out the right- version and their conjunction...

But it's not (especially) a deal-breaker for me either way which name/naming style we go for, but consistency seems more important in the long run. I suppose that *abstract* structures eg. `IsSemiring` don't mention their operators in their property names, but concrete instances for eg `Nat` do... and this is an 'abstract' property... should it in fact not be here at all, but be a manifest field of the `IsIdempotentCommutativeMonoid` structure instead?

jamesmckinna: Another (possible) contra: #1544 ?

jamesmckinna: Anyway: pushed now (but leaving `interchange` alone, for the time being... ;-))

MatthewDaggitt: > As for relating operators to themselves, interchange : Interchangable _∙_ _∙_ under CommutativeSemigroup provides a counter-example to your suggested naming, unless you think that that, too, should be ∙-interchange-∙?

Yes, I admit it's inconsistent at the moment. I'm not aware we have a naming convention for interchange yet, I would probably argue for something like `∙-∙-interchange`

>  I suppose that abstract structures eg. IsSemiring don't mention their operators in their property names, but concrete instances for eg Nat do... and this is an 'abstract' property... 

They do when there's potential ambiguity. 

> should it in fact not be here at all, but be a manifest field of the IsIdempotentCommutativeMonoid structure instead?

I don't think we can add every derived property as a manifest field otherwise the records are going to bloat like crazy.

MatthewDaggitt (approved):

PR #2424, opened 2024-07-01
Vec.Properties: introduce `≈-cong′`

mildsunrise:
## Additions

This PR introduces the `≈-cong′` lemma:

``` agda
≈-cong′ : ∀ {f-len : ℕ → ℕ} (f : ∀ {n} → Vec A n → Vec B (f-len n))
          {xs : Vec A m} {ys : Vec A n} .{eq} → xs ≈[ eq ] ys →
          f xs ≈[ cong f-len eq ] f ys
≈-cong′ f {xs = []}     {ys = []}     refl = cast-is-id refl (f [])
≈-cong′ f {xs = x ∷ xs} {ys = y ∷ ys} refl = ≈-cong′ (f ∘ (x ∷_)) refl
```

This lemma does for `≈[_]` what `cong` does for `≡`: given `f` and `xs ≈[ _ ] ys` it proves `f xs ≈[ _ ] f ys`.
More specifically, it proves that any `Vec A n → Vec B m` function that is polymorphic in `n` must preserve `≈[_]`.

## Current status

Compare this with the existing `≈-cong` lemma (introduced by @shhyou with #2067 along with the reasoning system):

``` agda
≈-cong : ∀ .{l≡o : l ≡ o} .{m≡n : m ≡ n} {xs : Vec A m} {ys zs} (f : Vec A o → Vec A n) →
         xs ≈[ m≡n ] f (cast l≡o ys) → ys ≈[ l≡o ] zs → xs ≈[ m≡n ] f zs
≈-cong f xs≈fys ys≈zs = trans xs≈fys (cong f ys≈zs)
```

This lemma works for a non-polymorphic `f`, but it requires the user to supply proof that `f` preserves `≈[_]`.
We already have a handful of these proofs written for some functions:

``` agda
map-cast : (f : A → B) .(eq : m ≡ n) (xs : Vec A m) →
           map f (cast eq xs) ≡ cast eq (map f xs)
map-cast {n = zero}  f eq []       = refl
map-cast {n = suc _} f eq (x ∷ xs)
  = cong (f x ∷_) (map-cast f (suc-injective eq) xs)

cast-++ˡ : ∀ .(eq : m ≡ o) (xs : Vec A m) {ys : Vec A n} →
           cast (cong (_+ n) eq) (xs ++ ys) ≡ cast eq xs ++ ys
cast-++ˡ {o = zero}  eq []       {ys} = cast-is-id refl (cast eq [] ++ ys)
cast-++ˡ {o = suc o} eq (x ∷ xs) {ys} = cong (x ∷_) (cast-++ˡ (cong pred eq) xs)

cast-++ʳ : ∀ .(eq : n ≡ o) (xs : Vec A m) {ys : Vec A n} →
           cast (cong (m +_) eq) (xs ++ ys) ≡ xs ++ cast eq ys
cast-++ʳ {m = zero}  eq []       {ys} = refl
cast-++ʳ {m = suc m} eq (x ∷ xs) {ys} = cong (x ∷_) (cast-++ʳ eq xs)

cast-∷ʳ : ∀ .(eq : suc n ≡ suc m) x (xs : Vec A n) →
          cast eq (xs ∷ʳ x) ≡ (cast (cong pred eq) xs) ∷ʳ x
cast-∷ʳ {m = zero}  eq x []       = refl
cast-∷ʳ {m = suc m} eq x (y ∷ xs) = cong (y ∷_) (cast-∷ʳ (cong pred eq) x xs)
cast-∷ʳ _ x _ = ≈-cong′ (_∷ʳ x) refl

cast-reverse : ∀ .(eq : m ≡ n) → cast eq ∘ reverse {A = A} {n = m} ≗ reverse ∘ cast eq
cast-reverse {n = zero}  eq []       = refl
cast-reverse {n = suc n} eq (x ∷ xs) = begin
  reverse (x ∷ xs)           ≂⟨ reverse-∷ x xs ⟩
  reverse xs ∷ʳ x            ≈⟨ ≈-cong (_∷ʳ x) (cast-∷ʳ eq x (reverse xs))
                                       (cast-reverse (cong pred eq) xs) ⟩
  reverse (cast _ xs) ∷ʳ x   ≂⟨ reverse-∷ x (cast (cong pred eq) xs) ⟨
  reverse (x ∷ cast _ xs)    ≈⟨⟩
  reverse (cast eq (x ∷ xs)) ∎
  where open CastReasoning
```

## Simplifications

`≈-cong′` is used in a similar way, except that it doesn't require manually proving preservation of `cast` for `f`.
So, most uses of `≈-cong` (including all current ones) can be migrated to `≈-cong′` just by removing the 2nd argument:

``` diff
-≈-cong (_∷ʳ x) (cast-∷ʳ (cong suc (List.length-reverse xs)) _ _) (fromList-reverse xs)
+≈-cong′ (_∷ʳ x) (fromList-reverse xs)
```

An extreme example is `example4-cong²`, where we show chaining `≈-cong`. This can also be done with `≈-cong′`, but there we can further express it as a single `≈-cong′` of the composition, as we tend to do with `cong`:

``` diff
-≈-cong reverse (cast-reverse (cong (_+ n) (+-comm 1 m)) ((xs ∷ʳ a) ++ ys))
-        (≈-cong (_++ ys) (cast-++ˡ (+-comm 1 m) (xs ∷ʳ a))
-                (unfold-∷ʳ _ a xs))
+≈-cong′ (reverse ∘ (_++ ys)) (unfold-∷ʳ _ a xs)
```

And the 'cast preservation' proofs above, should you still need them, are just special cases of `≈-cong′`:

``` agda
map-cast f _ _ = sym (≈-cong′ (map f) refl)
cast-++ˡ _ _ {ys} = ≈-cong′ (_++ ys) refl
cast-++ʳ _ xs = ≈-cong′ (xs ++_) refl
cast-∷ʳ _ x _ = ≈-cong′ (_∷ʳ x) refl
cast-reverse _ _ = ≈-cong′ reverse refl
```

### DIFF
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -386,10 +386,10 @@ cast-sym eq {xs = x ∷ xs} {ys = y ∷ ys} xxs[eq]≡yys =
   in cong₂ _∷_ (sym x≡y) (cast-sym (suc-injective eq) xs[eq]≡ys)
 
 ≈-cong′ : ∀ {f-len : ℕ → ℕ} (f : ∀ {n} → Vec A n → Vec B (f-len n))
-          {xs : Vec A m} {ys : Vec A n} .{eq} → xs ≈[ eq ] ys →
+          {m n} {xs : Vec A m} {ys : Vec A n} .{eq} → xs ≈[ eq ] ys →
           f xs ≈[ cong f-len eq ] f ys
-≈-cong′ f {xs = []}     {ys = []}     refl = cast-is-id refl (f [])
-≈-cong′ f {xs = x ∷ xs} {ys = y ∷ ys} refl = ≈-cong′ (f ∘ (x ∷_)) refl
+≈-cong′ f {m = zero}  {n = zero}  {xs = []}     {ys = []}     refl = cast-is-id refl (f [])
+≈-cong′ f {m = suc m} {n = suc n} {xs = x ∷ xs} {ys = y ∷ ys} refl = ≈-cong′ (f ∘ (x ∷_)) refl
 
 ------------------------------------------------------------------------
 -- map
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -1035,8 +1035,7 @@ reverse-++-eqFree : ∀ (xs : Vec A m) (ys : Vec A n) → let eq = +-comm m n in
 reverse-++-eqFree {m = zero}  {n = n} []       ys = ≈-sym (++-identityʳ-eqFree (reverse ys))
 reverse-++-eqFree {m = suc m} {n = n} (x ∷ xs) ys = begin
   reverse (x ∷ xs ++ ys)              ≂⟨ reverse-∷ x (xs ++ ys) ⟩
-  reverse (xs ++ ys) ∷ʳ x             ≈⟨ ≈-cong (_∷ʳ x) (cast-∷ʳ (cong suc (+-comm m n)) x (reverse (xs ++ ys)))
-                                                (reverse-++-eqFree xs ys) ⟩
+  reverse (xs ++ ys) ∷ʳ x             ≈⟨ ≈-cong′ (_∷ʳ x) (reverse-++-eqFree xs ys) ⟩
   (reverse ys ++ reverse xs) ∷ʳ x     ≈⟨ ++-∷ʳ-eqFree x (reverse ys) (reverse xs) ⟩
   reverse ys ++ (reverse xs ∷ʳ x)     ≂⟨ cong (reverse ys ++_) (reverse-∷ x xs) ⟨
   reverse ys ++ (reverse (x ∷ xs))    ∎
@@ -1087,8 +1086,7 @@ map-ʳ++ {ys = ys} f xs = begin
                 cast eq ((xs ++ ys) ʳ++ zs) ≡ ys ʳ++ (xs ʳ++ zs)
 ++-ʳ++-eqFree {m = m} {n} {o} xs {ys} {zs} = begin
   ((xs ++ ys) ʳ++ zs)              ≂⟨ unfold-ʳ++ (xs ++ ys) zs ⟩
-  reverse (xs ++ ys) ++ zs         ≈⟨ ≈-cong (_++ zs) (cast-++ˡ (+-comm m n) (reverse (xs ++ ys)))
-                                             (reverse-++-eqFree xs ys) ⟩
+  reverse (xs ++ ys) ++ zs         ≈⟨ ≈-cong′ (_++ zs) (reverse-++-eqFree xs ys) ⟩
   (reverse ys ++ reverse xs) ++ zs ≈⟨ ++-assoc-eqFree (reverse ys) (reverse xs) zs ⟩
   reverse ys ++ (reverse xs ++ zs) ≂⟨ cong (reverse ys ++_) (unfold-ʳ++ xs zs) ⟨
   reverse ys ++ (xs ʳ++ zs)        ≂⟨ unfold-ʳ++ ys (xs ʳ++ zs) ⟨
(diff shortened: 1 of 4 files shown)


mildsunrise:
(draft because we still need to figure out a good name for this lemma, and also update the documentation at `doc/README/Data/Vec/Relation/Binary/Equality/Cast.agda`)

src/Data/Vec/Properties.agda
```diff
@@ -381,6 +381,12 @@ cast-sym eq {xs = x ∷ xs} {ys = y ∷ ys} xxs[eq]≡yys =
   let x≡y , xs[eq]≡ys = ∷-injective xxs[eq]≡yys
   in cong₂ _∷_ (sym x≡y) (cast-sym (suc-injective eq) xs[eq]≡ys)
 
+≈-cong′ : ∀ {f-len : ℕ → ℕ} (f : ∀ {n} → Vec A n → Vec B (f-len n))
```
shhyou, line 384: Is it `Vec A (f-len n)`?

shhyou: I see what I have missed. It's correct to have `B` in the type. However, I do feel like it is better to place `≈-cong′` in `Data.Vec.Relation.Binary.Equality.Cast.CastReasoning`.

shhyou:
This improvement is awesome! In the long term, it'd be great if it's possible to prioritize this new lemma over the existing `≈-cong`, perhaps taking up its name and giving the old one a different name.

src/Data/Vec/Properties.agda
mildsunrise: yeah, that was my first impulse as well, but it's unfortunately not possible to place it there, precisely because of that `B`. to do it, we should first convert the `A` module parameter into a variable. shall I do it?

shhyou: Yes, I think it's worth it to turn the module parameter `A` into `variable`s in order to group `≈-cong′` and other combinators together.

I think using `variable`s makes `≈-sym` and `≈-trans` uglier since `{A}` will have to be explicitly supplied, but I don't have a better solution.

MatthewDaggitt: Unfortunately, if we turn the module parameter into a variable in `Data.Vec.Relation.Binary.Equality.Cast.CastReasoning` we will have to make this is `v3.0` PR as that is breaking change. I suggest therefore we stick with the current location?

MatthewDaggitt: Ah except... that the module parameter is implicit. In which case I think it's fine to move it there :+1:

MatthewDaggitt (approved):

JacquesCarette (approved):

src/Data/Vec/Properties.agda
```diff
@@ -381,6 +381,12 @@ cast-sym eq {xs = x ∷ xs} {ys = y ∷ ys} xxs[eq]≡yys =
...
+          {xs : Vec A m} {ys : Vec A n} .{eq} → xs ≈[ eq ] ys →
+          f xs ≈[ cong f-len eq ] f ys
+≈-cong′ f {xs = []}     {ys = []}     refl = cast-is-id refl (f [])
+≈-cong′ f {xs = x ∷ xs} {ys = y ∷ ys} refl = ≈-cong′ (f ∘ (x ∷_)) refl
```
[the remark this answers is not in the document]
mildsunrise: interesting... i tend not to care about the computational character of lemmas that produce identities, but I can definitely apply that change. will do both soon-ish :)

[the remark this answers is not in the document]
mildsunrise: done

src/Data/Vec/Properties.agda
mildsunrise: done. I think the signatures were left unchanged (there was no unwanted repositioning of implicit args), please review

src/Data/Vec/Relation/Binary/Equality/Cast.agda
```diff
@@ -10,8 +10,10 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-module Data.Vec.Relation.Binary.Equality.Cast {a} {A : Set a} where
+module Data.Vec.Relation.Binary.Equality.Cast where
 
+open import Level using (Level)
+open import Function using (_∘_)
```
shhyou, line 16: Maybe use `Function.Base`?

src/Data/Vec/Properties.agda
```diff
@@ -369,7 +369,7 @@ lookup∘update′ {i = i} {j} i≢j xs y = lookup∘updateAt′ i j i≢j xs
 -- cast
 
 open VecCast public
-  using (cast-is-id; cast-trans)
+  using (cast-is-id; cast-trans; ≈-cong′)
```
shhyou, line 372: I think `≈-cong′` should not be put under `public`.

src/Data/Vec/Properties.agda
shhyou: ha! I didn't realize that the type could be added to `Sym` and `Trans`.

jamesmckinna:
This seems like a great contribution... but I'm not across what's needed to get this over the line for v2.2 
@mildsunrise how would you like to proceed? 
Is there something (further) that the other (approving) maintainers could contribute here?

MatthewDaggitt:
Remaining to do?

1. Both of @shhyou's comment should be implemented I think
2. We should then implement the old `≈-cong` in terms of `≈-cong′` for sanity
3. We should deprecate `≈-cong`?

mildsunrise:
Hi! Sorry for the absence, I'll try to unblock this as much as possible.

> Both of @shhyou's comment should be implemented I think

Done!

> We should then implement the old `≈-cong` in terms of `≈-cong′` for sanity

I don't think we can, because the old `≈-cong` is more general than the new one: the new one only works for functions that are polymorphic in the input length, and where the output length depends only on the input length.

> We should deprecate `≈-cong`?

I am indifferent. I guess it could be argued that the new `≈-cong′` is the one that actually shows congruence, and the old one is merely an adapter from a sort of congruence proof given for `cast` to a congruence proof for `≈[_]`.

MatthewDaggitt:
> I don't think we can, because the old ≈-cong is more general than the new one: the new one only works for functions that are polymorphic in the input length, and where the output length depends only on the input length.

Ah okay, that makes sense!

>  I am indifferent. I guess it could be argued that the new ≈-cong′ is the one that actually shows congruence, and the old one is merely an adapter from a sort of congruence proof given for cast to a congruence proof for ≈[_].

Yup. I was advocating for deprecation based on the misconception that the new definition was more general. 

Okay great, I think that tidies everything up. I'll merge this in and squeeze it into v2.2

PR #2426, opened 2024-07-02
Add left- and right- `Pointwise` congruence for `_++_` on `List`

### LINKED ISSUE #1131: Add left and right versions of `++⁺` for pointwise relations
Data.List.Relation.Binary.Pointwise  has 
```
module _ {_∼_ : REL A B ℓ} where                                                                                                                                
  ++⁺ : ∀ {ws xs ys zs} → Pointwise _∼_ ws xs → Pointwise _∼_ ys zs →             
                                        Pointwise _∼_ (ws ++ ys) (xs ++ zs)  
```
According to the library style, needs it to be   
``++-cong₂ :  _++_ Preserves₂ (Pointwise _~_) ⟶ (Pointwise _~_) ⟶ (Pointwise _~_)`` 
?
There remains a question of the difference between ``REL A B _`` and ``Rel A _, Rel B _``.
So that I wonder.
Anyway, it has, probably, sense to add lemmas derived from ``++⁺``, named ``++-cong<foo>``,
and having ``Preserves/Preserves₂`` in the signature.
For example, in my program, I write for lists over a setoid:
```
  ++cong1=ₗ :   ∀ {ys} → (_++ ys) Preserves _=ₗ_ ⟶ _=ₗ_
  ++cong1=ₗ {ys} ws=ₗxs =   ++⁺ {_∼_ = _≈_} ws=ₗxs =ₗrefl 
```

jamesmckinna:
Fixes #1131 

NB: no `CHANGELOG` (yet), but adds:
* to `Data.List.Relation.Binary.Pointwise`
```agda
  ++⁺ʳ : Reflexive R → ∀ xs → Pointwise R ys zs → Pointwise R (xs ++ ys) (xs ++ zs)
  ++⁺ˡ : Reflexive R → Pointwise R ws xs → ∀ zs → Pointwise R (ws ++ zs) (xs ++ zs)
```
* to `Data.List.Relation.Binary.Equality.Setoid`
```agda
  ++⁺ʳ : ∀ xs {ys zs} → ys ≋ zs → xs ++ ys ≋ xs ++ zs
  ++⁺ˡ : ∀ {ws xs} → ws ≋ xs → ∀ zs → ws ++ zs ≋ xs ++ zs
```

Issue:
* the [originating issue](https://github.com/agda/agda-stdlib/issues/1131) is 4 years old... (and should we have a policy on pruning such things?)
* arguably, such lemmas (at least: the second pair) should be *automatically* derivable as `∙-congʳ` and `∙-congˡ` for the `(Is)Magma` structure/bundle defined on `List` in #2393 ...
* so it's possible (desirable?) that we mark this PR as `status:duplicate` and not merge after all (except perhaps the first pair of lemmas?)

UPDATED: added `CHANGELOG`, and revised the types of lemmas in response to @JacquesCarette 's comments below.


MatthewDaggitt (approved):

jamesmckinna:
Thanks @MatthewDaggitt : any opinions on the discussion points at the top?

src/Data/List/Relation/Binary/Pointwise.agda
```diff
@@ -166,6 +166,15 @@ tabulate⁻ {n = suc n} (x∼y ∷ xs∼ys) (fsuc i) = tabulate⁻ xs∼ys i
 ++-cancelʳ {xs = xs}     (y ∷ ys) []       eq   =
   contradiction (≡.trans (≡.sym (length-++ (y ∷ ys))) (Pointwise-length eq)) (m≢1+n+m (length xs) ∘ ≡.sym)
 
+module _ (rfl : Reflexive R) where
+
+  ++⁺ʳ : ∀ xs → Pointwise R ys zs → Pointwise R (xs ++ ys) (xs ++ zs)
```
JacquesCarette, line 171: Aren't you inconsistent between these two wrt to where you put the quantified variable?

Also, I would think that the `xs` (here) and `zs` (below) could both be implicit? They're not needed in the proofs at all, which is a strong sign that they are not strongly relevant.

JacquesCarette (changes requested):
While it would be nice to derive these through automation, we're kind of far from that. Let's have a full-featured library now, and worry about that later. So I'm all for merging this in.

src/Data/List/Relation/Binary/Pointwise.agda
jamesmckinna: I'll leave this TODO for now, but I suspect that something similar is at play in the problems with #2415 ...? (Forward proof permits weaker quantification, while backward proof demands stronger thanks to the semantics of matching/unification?)

MatthewDaggitt:
> the https://github.com/agda/agda-stdlib/issues/1131 is 4 years old... (and should we have a policy on pruning such things?)

I'm happy for old issues to still float around if they haven't been solved.

> so it's possible (desirable?) that we mark this PR as status:duplicate and not merge after all (except perhaps the first pair of lemmas?)

I think better that they exist first. We can then always refactor the implementation later.

src/Data/List/Relation/Binary/Pointwise.agda
jamesmckinna: Have now added revised types for the new lemmas, and note the following:
* `++⁺ʳ` now has a type corresponding to the explicitly quantified version of `Algebra.Definitions.LeftCongruent`
* `++⁺ˡ` now has a type corresponding to the explicitly quantified version of `Algebra.Definitions.RightCongruent`

Which nomenclature/notation should be taken as canon?

jamesmckinna:
Barring the resolution of the `Left`/`Right` question above, I think this is good to go now.

MatthewDaggitt:
> Which nomenclature/notation should be taken as canon? (Potential issue: I think that the use(s) of left/right superscript annotations in the Data.List.Relation* hierarchy are opposite to those of the Algebra.* congruence properties... sigh)

Sigh as well. I think it's better to be consistent within `Data.List.Relation`...

jamesmckinna:
> > Which nomenclature/notation should be taken as canon? (Potential issue: I think that the use(s) of left/right superscript annotations in the Data.List.Relation* hierarchy are opposite to those of the Algebra.* congruence properties... sigh)
> 
> Sigh as well. I think it's better to be consistent within `Data.List.Relation`...

Ok, I'll leave the names as is, but `v3.0` or `hypothetical-rewrite` to correct this inconsistency... one day?

jamesmckinna:
@JacquesCarette are you now happy with the revisions to this?

JacquesCarette (approved):

PR #2429, opened 2024-07-04
[v2.1-rc1] fixes #2400: use explicit quantification instead

### LINKED ISSUE #2400: What's the 'right' notion of equality between functions?
We have [the pointwise `PropositionalEquality` definition `_≗_`](https://github.com/agda/agda-stdlib/blob/c719430e5f27a2df47a18df8781cb5824ba5b601/src/Relation/Binary/PropositionalEquality/Core.agda#L35-L42) (moved in #2335 )
```agda
_≗_ {A = A} {B = B} f g = ∀ x → f x ≡ g x
```
and [the `Function` definition `_≈_`](https://github.com/agda/agda-stdlib/blob/c719430e5f27a2df47a18df8781cb5824ba5b601/src/Function/Relation/Binary/Setoid/Equality.agda#L25-L27) (since #2240 and its antecedent issues back to #1467 )
```agda
_≈_ : (f g : Func From To) → Set (a₁ ⊔ b₂)
f ≈ g = {x : From.Carrier} → f ⟨$⟩ x To.≈ g ⟨$⟩ x
```
and... we are (once again!) inconsistent as to implicit/explicit quantification.

@JacquesCarette has argued for the implicit version; @Taneb 's recent [topic thread on Zulip](https://agda.zulipchat.com/#narrow/stream/263194-categories/topic/Should.20Sets.20use.20.E2.89.97.3F) suggests that this is not always the right choice.

Maybe there's room for both, but it doesn't feel quite right to me... or else, the distinction should be documented.

jamesmckinna:
Breaks #2240 but as that was added in v2.1 suggest we merge this to supersede the definition there.
~~TODO~~ DONE: Knock-on effect on #2299 

NB no `CHANGELOG` required?

### DIFF
--- a/src/Function/Endo/Setoid.agda
+++ b/src/Function/Endo/Setoid.agda
@@ -58,7 +58,7 @@ private
 ∘-isMagma : IsMagma _≈_ _∘_
 ∘-isMagma = record
   { isEquivalence = isEquivalence
-  ; ∙-cong        = λ {_} {_} {_} {v} x≈y u≈v → S.trans u≈v (cong v x≈y)
+  ; ∙-cong        = λ {_} {_} {_} {k} f≈g h≈k x → S.trans (h≈k _) (cong k (f≈g x))
   }
 
 ∘-magma : Magma (c ⊔ e) (c ⊔ e)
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

JacquesCarette:
Breaks the build... (`Function.Relation.Binary.Setoid.Equality`) but otherwise, I agree.

MatthewDaggitt (approved):
Yup looks great!

PR #2430, opened 2024-07-05
Vec.Properties: drop eq parameter when it is a property

mildsunrise:
This PR applies the changes discussed in #2421 for v2.2.
Since these aren't "usual" deprecations where a lemma changes name, I've put them in a separate section at the end of the file.

Note that this conflicts with #2424, so one of those two should be merged so I have time to resolve the conflicts in the other one.

PR #2431, opened 2024-07-05
Vec.Properties: drop eq parameter when it is a property (v3.0 changes)

### LINKED ISSUE #2421: Proofs in `Data.Vec.Properties` take general properties as inputs
Looking at `Data.Vec.Properties`, I see many properties accept an irrelevant `eq` parameter that is actually a well-known property on naturals and has nothing relevant to the particular input. For example:

``` agda
++-∷ʳ : ∀ .(eq : suc (m + n) ≡ m + suc n) z (xs : Vec A m) (ys : Vec A n) →
        cast eq ((xs ++ ys) ∷ʳ z) ≡ xs ++ (ys ∷ʳ z)
++-∷ʳ {m = zero}  eq z []       []       = refl
++-∷ʳ {m = zero}  eq z []       (y ∷ ys) = cong (y ∷_) (++-∷ʳ refl z [] ys)
++-∷ʳ {m = suc m} eq z (x ∷ xs) ys       = cong (x ∷_) (++-∷ʳ (cong pred eq) z xs ys)
```

Instead of requesting a proof for `suc (m + n) ≡ m + suc n`, we could fill it in ourselves with `+-suc` which is already in scope:

``` agda
++-∷ʳ : ∀ z (xs : Vec A m) (ys : Vec A n) → let eq = sym (+-suc m n) in
        cast eq ((xs ++ ys) ∷ʳ z) ≡ xs ++ (ys ∷ʳ z)
++-∷ʳ {m = zero}  z []       []       = refl
++-∷ʳ {m = zero}  z []       (y ∷ ys) = cong (y ∷_) (++-∷ʳ′ z [] ys)
++-∷ʳ {m = suc m} z (x ∷ xs) ys       = cong (x ∷_) (++-∷ʳ′ z xs ys)
```

This not only makes `++-∷ʳ` easier to use, it also simplifies the code for the proof itself (no extra `eq` argument, no need for `refl` and `cong pred eq`). And because `eq` is irrelevant in `cast`, we don't lose generality -- the original statement is just `const ++-∷ʳ`:

``` agda
++-∷ʳ′ : ∀ .(eq : suc (m + n) ≡ m + suc n) z (xs : Vec A m) (ys : Vec A n) →
         cast eq ((xs ++ ys) ∷ʳ z) ≡ xs ++ (ys ∷ʳ z)
++-∷ʳ′ _ = ++-∷ʳ
```

So I wanted to ask, what's the motivation for writing properties the current way? Was it because of limitations of Agda back when these proofs were written (wrt irrelevancy and so on)? Am I myself missing Later in the file I can see properties that use the second style:

``` agda
fromList-reverse : (xs : List A) → cast (List.length-reverse xs) (fromList (List.reverse xs)) ≡ reverse (fromList xs)
```

And the follow up question is: would it be a good idea to change the properties to the second style in the next compat-breaking version?

mildsunrise:
Follow up to #2430 with the changes for v3. See #2421 for context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,19 @@ Deprecated names
   toDec    ↦  Relation.Nullary.Decidable.Core.fromSum
   ```
 
+* In `Data.Vec.Properties`:
+  ```agda
+  ++-assoc-eqFree      ↦  ++-assoc
+  ++-identityʳ-eqFree  ↦  ++-identityʳ
+  unfold-∷ʳ-eqFree     ↦  unfold-∷ʳ
+  ++-∷ʳ-eqFree         ↦  ++-∷ʳ
+  ∷ʳ-++-eqFree         ↦  ∷ʳ-++
+  reverse-++-eqFree    ↦  reverse-++
+  ∷-ʳ++-eqFree         ↦  ∷-ʳ++
+  ++-ʳ++-eqFree        ↦  ++-ʳ++
+  ʳ++-ʳ++-eqFree       ↦  ʳ++-ʳ++
+  ```
+
 * In `IO.Base`:
   ```agda
   untilRight  ↦  untilInj₂
--- a/doc/README/Data/Vec/Relation/Binary/Equality/Cast.agda
+++ b/doc/README/Data/Vec/Relation/Binary/Equality/Cast.agda
@@ -92,7 +92,7 @@ example1a-fromList-∷ʳ x xs eq = begin
   cast eq₂ (cast eq₁ (fromList (xs List.++ List.[ x ])))
     ≡⟨ cong (cast eq₂) (fromList-++ xs) ⟩
   cast eq₂ (fromList xs ++ [ x ])
-    ≡⟨ ≈-sym (unfold-∷ʳ′ x (fromList xs)) ⟩
+    ≡⟨ ≈-sym (unfold-∷ʳ-eqFree x (fromList xs)) ⟩
   fromList xs ∷ʳ x
     ∎
   where
@@ -114,7 +114,7 @@ example1b-fromList-∷ʳ x xs eq = begin
   fromList (xs List.++ List.[ x ])
     ≈⟨ fromList-++ xs ⟩
   fromList xs ++ [ x ]
-    ≈⟨ unfold-∷ʳ′ x (fromList xs) ⟨
+    ≈⟨ unfold-∷ʳ-eqFree x (fromList xs) ⟨
   fromList xs ∷ʳ x
     ∎
   where open CastReasoning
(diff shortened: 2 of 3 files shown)


Taneb:
@mildsunrise now that the next release will be 3.0, do you want to revisit this?

jamesmckinna:
So I couldn't help noticing, while trying to figure out the merge conflicts, that @shhyou 's original proofs made extensive use of the `... ≈[ ... ] ...` syntax, whereas the proofs in this PR reinstate the explicit unfolding of that syntax into `cast ... ... ≡ ...` form.

Is that a Good Idea, or a Bad One?

Otherwise put: is it a `style-guide`/`library-design` precept that we *don't* use `syntax` explicitly in library modules?

JacquesCarette:
I like the earlier syntax. [In part it is because it is a degenerate case of 'display map' where on thing holds over another, in this case, one equation over another.]

shhyou:
No, I think this PR is not related to the difference between `... ≈[ ... ] ...` and `cast ... ... ≡ ...`. The current PR is to finish the postponed breaking changes in #2430, namely replace these lemmas on the left by their `-eqFree` variants on the right:
```
  ++-assoc _      ↦  ++-assoc-eqFree
  ++-identityʳ _  ↦  ++-identityʳ-eqFree
  unfold-∷ʳ _     ↦  unfold-∷ʳ-eqFree
  ++-∷ʳ _         ↦  ++-∷ʳ-eqFree
  ∷ʳ-++ _         ↦  ∷ʳ-++-eqFree
  reverse-++ _    ↦  reverse-++-eqFree
  ∷-ʳ++ _         ↦  ∷-ʳ++-eqFree
  ++-ʳ++ _        ↦  ++-ʳ++-eqFree
  ʳ++-ʳ++ _       ↦  ʳ++-ʳ++-eqFree
```

Previously, the `ʳ++-ʳ++` lemmas use `... ≈[ ... ] ...` syntax because they directly reason over `cast` equationally. What looks like "unfolding" at first sight is due to #2733's new proof strategy: from `toList-injective` and various `toList-X` lemmas, @jmougeot reuses list lemmas in `Data.List.Properties` to prove properties about vectors. Therefore `cast` is gone and plain equational reasoning suffices. I believe these differences do not fit @jamesmckinna's question.

I think the renaming from `*-eqFree` to `*` is still needed, but the current proof strategy is good.

jamesmckinna:
Thanks @shhyou for the more refined analysis than my first superficial re-reading of this PR. And, to be clear, I think this is an important followup change! I had just been trying to get my head around it before trying to resolve the merge conflicts and then try to merge in to `stdlib` sooner rather than later...

PR #2438, opened 2024-07-11
remove unnecessary condition in the `Relation.Binary.Reasoning.Base.Apartness`

damhiya:
`Transitive _#_` condition in the module `Relation.Binary.Reasoning.Base.Apartness` is unnecessary. Indeed, apartness relation can be transitive only if it is empty, and I think the module name is not appropriate since this module just works for any symmetric relation.


JacquesCarette (changes requested):
Agreed - but needs a CHANGELOG entry!

damhiya:
Is there any opinion about changing the module name?

MatthewDaggitt:
Sorry for the long reply time. 

> Is there any opinion about changing the module name?

As always, it's easy to say the current name isn't the best. Do you have a better suggestion we can compare it with? I personally don't think the current name is terrible.

@JacquesCarette would you be willing to approve now?

MatthewDaggitt (approved):

damhiya:
I recommend `Relation.Binary.Reasoning.Base.Symmetric`.

JacquesCarette (approved):
I'm fine with this PR doing just the removal of the condition. Name changes can be in a different PR.

JacquesCarette:
Looking in more detail: what's in the file seems quite consistent with its current name - it is base reasoning about an apartness relation!

MatthewDaggitt:
I'm also tempted to agree with @JacquesCarette. It's also not as if it's a very widely used module outside of the library.

PR #2439, opened 2024-07-17
Setup for v2.2 development

MatthewDaggitt:
After this is merged, we can start merging in v2.2 PRs.


jamesmckinna (approved):
Besides @gallais 's suggestion, I think that this looks good to go.

jamesmckinna:
Should we also update the `release-guide.md` to document this (new) workflow/branch discipline... and set in stone blessed names for these 'release' branches... cf. #2445 ?

JacquesCarette:
Makes sense to me!

PR #2440, opened 2024-07-19
Implement ≤-total in terms of _≤?_

### LINKED ISSUE #2436: Data.Nat.Properties.≤-total should be implemented with _<ᵇ_
It's very slow for `≤-total` to decide which constructor to use, especially compared to `≤?`. This is an issue using `Data.Nat` with code generic over a `TotalOrder` - it's far slower than it could be.

Taneb:
Fixes #2436

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -185,11 +185,6 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 ≤-trans z≤n       _         = z≤n
 ≤-trans (s≤s m≤n) (s≤s n≤o) = s≤s (≤-trans m≤n n≤o)
 
-≤-total : Total _≤_
-≤-total zero    _       = inj₁ z≤n
-≤-total _       zero    = inj₂ z≤n
-≤-total (suc m) (suc n) = Sum.map s≤s s≤s (≤-total m n)
-
 ≤-irrelevant : Irrelevant _≤_
 ≤-irrelevant z≤n        z≤n        = refl
 ≤-irrelevant (s≤s m≤n₁) (s≤s m≤n₂) = cong s≤s (≤-irrelevant m≤n₁ m≤n₂)
@@ -207,6 +202,11 @@ m ≤? n = map′ (≤ᵇ⇒≤ m n) ≤⇒≤ᵇ (T? (m ≤ᵇ n))
 
 _≥?_ : Decidable _≥_
 _≥?_ = flip _≤?_
+
+≤-total : Total _≤_
+≤-total m n with m ≤? n
+... | yes m≤n = inj₁ m≤n
+... | no  m≰n = inj₂ (≰⇒≥ m≰n)
 
 ------------------------------------------------------------------------
 -- Structures
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -162,6 +162,11 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 -- Properties of _≤_
 ------------------------------------------------------------------------
 
+≰⇒≥ : _≰_ ⇒ _≥_
+≰⇒≥ {m} {zero} m≰n = z≤n
+≰⇒≥ {zero} {suc n} m≰n = contradiction z≤n m≰n
+≰⇒≥ {suc m} {suc n} m≰n = s≤s (≰⇒≥ (m≰n ∘ s≤s))
+
 ------------------------------------------------------------------------
 -- Relational properties of _≤_
 
@@ -330,9 +335,6 @@ n≤1⇒n≡0∨n≡1 (s≤s z≤n) = inj₂ refl
 ≰⇒> {zero}          z≰n = contradiction z≤n z≰n
 ≰⇒> {suc m} {zero}  _   = z<s
 ≰⇒> {suc m} {suc n} m≰n = s<s (≰⇒> (m≰n ∘ s≤s))
-
-≰⇒≥ : _≰_ ⇒ _≥_
-≰⇒≥ = <⇒≤ ∘ ≰⇒>
 
 ≮⇒≥ : _≮_ ⇒ _≥_
 ≮⇒≥ {_}     {zero}  _       = z≤n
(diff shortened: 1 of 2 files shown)


gallais:
You'll need a CHANGELOG entry

Taneb:
That's a good point, I'll add a changelog after #2439 gets merged

JacquesCarette:
I'll wait to do a proper review once the CHANGELOG is in place.

jamesmckinna (approved):
Modulo:
* `CHANGELOG`
* `breaking`

This looks great!

jamesmckinna:
How does this PR contribute to (positively or negatively...) the issues about 'concrete' orderings raised in issue #1179 ? 
It seems as though both are trying to tackle a pervasive problem; this PR, locally, with the bigger issue at a (more) global level.

Taneb:
@jamesmckinna they're related, but I don't think either of them solve or get in the way of the other

jamesmckinna:
I think when you resolve the merge conflict in `CHANGELOG`, you'll see a couple of entries from v2.1 which have crept in/been left in by mistake, and which should now be removed.

src/Data/Nat/Properties.agda
```diff
@@ -203,6 +203,11 @@ m ≤? n = map′ (≤ᵇ⇒≤ m n) ≤⇒≤ᵇ (T? (m ≤ᵇ n))
 _≥?_ : Decidable _≥_
 _≥?_ = flip _≤?_
 
+≤-total : Total _≤_
+≤-total m n with m ≤? n
+... | yes m≤n = inj₁ m≤n
+... | no  m≰n = inj₂ (≰⇒≥ m≰n)
```
JacquesCarette, line 209: You've improved the first case of this proof but not the second, as this still dispatches to a proof by recursion. Now laziness probably kicks in -- and yet, if you're worried about efficiency, shouldn't all the cases by made equally efficient?

JacquesCarette (changes requested):

src/Data/Nat/Properties.agda
Taneb: I think that anything that actually needs the value of the `m ≤ n` witness is going to need to do a proof by recursion to construct the value, including existing `_≤?_`. What matters here is the _decision_, whether we have `inj₁` or `inj₂`

jamesmckinna: Moreover, proofs of `_≤_` are irrelevant, so no client code should ever in poractice be exploiting the exact nature of the proof being computed. That said, there may be a way to concretise observation that via suitable use of `recompute`?

Taneb: I think adding an unnecessary `recompute` here would only confuse things...

JacquesCarette: I think "proofs of `_≤_` are irrelevant" is being abused for what it's not: yes, all such proofs are equal, BUT they still contain information (which can be extracted), namely the difference (as a Nat) between the numbers.

Irrelevant does not mean information-free.

jamesmckinna: I think I'd be groping towards something (esp. wit the mention of `recompute`) along the lines of: the implication `≰⇒≥` is based on a *negated* premise, which we know can have no computational content (cf. `Relation.Nullary.Recomputable`)... so while the conclusion `m ≥ n` indeed does have computational content (the difference), this can't arise *per se* from the proof of `m ≰ n`... so: either it *must* be by reference to the actual `m` and `n` concerned, or else that it, too, can be obtained by `recompute` from... something irrelevant. But I admit, it would be better if I had been able to demonstrate this, rather than speculate philosophically on the point.

MatthewDaggitt:
The question is do we view this as a backwards compatible breaking change? If it was just a change in behaviour then I would say yes and bump it to v3.0. If however, we argue that this is in fact a performance bug then the maybe we can justify adding it to v2.2?

src/Data/Nat/Properties.agda
JacquesCarette: Re: efficiency. I think I can see that.

JacquesCarette (approved):

Taneb:
@MatthewDaggitt while I would like it in sooner, of course, the position that has the better argument is that we don't make any performance guarantees at all (our readme says more or less this right near the beginning). Based on that, it probably ought wait for 3.0

jamesmckinna:
> @MatthewDaggitt while I would like it in sooner, of course, the position that has the better argument is that we don't make any performance guarantees at all (our readme says more or less this right near the beginning). Based on that, it probably ought wait for 3.0

Well, if you were to make that sacrifice, that would ease the path to a `v2.2` release following #2473 ...

src/Data/Nat/Properties.agda
gallais, line 209: Perhaps avoid using `yes` and `no`, cf. https://github.com/agda/agda-stdlib/pull/2365/files#r1566627871

gallais (approved):

src/Data/Nat/Properties.agda
jamesmckinna: Thanks for the shoutout, I've been wondering whether I've ridden that hobby-horse a bit too hard, but I do like the new, lazier versions.

jamesmckinna:
> Any objections to this being merged now?

Against our most recent discussion on the topic, we still do have a (placeholder?) non-`breaking` v2.3 milestone open, whereas `breaking` issues/PRs are typically being considered as v3.0? So maybe, ready though it may be, we can hold off merging for a bit? or is this mission critical to another project?

JacquesCarette:
Do we really wish to regard relying on reduction behaviour of *proofs* as 'breaking'? I completely agree that, given how Agda (MLTT, really) works, lots of proofs inevitably rely on the reduction behaviour of the terms they speak about. But lifting that up a level seems a bit too far? Do we have many theorems about properties of properties?

I suspect that this change would, in fact, not affect anyone. ['suspect' indicates lower confidence that 'bet' as I'm not a big gambler.]

jamesmckinna:
@JacquesCarette let me clarify my position:
* my understanding was that we had agreed that this was `breaking`, but I am happy if we row back on such a judgment
* my intention in  the above was to draw attention to PRs which are clearly non-`breaking`/v2.3, in contrast to those which might better/best be judged v3.0...

If everyone is happy, I'd merge this for v2.3, but would advise, for belt-and-braces, that there be a "non-backwards-compatible" entry in `CHANGELOG` to record what's happening?

jamesmckinna:
Or else, that we abandon the distinction v2.3/v3.0, and fast-forward to admit all the breaking changes? (We may have agreed that at the last maintainers' meeting, but for whatever reason, I'm dragging my heels...)

PR #2444, opened 2024-07-29
Run CI tests on release branches as well

MatthewDaggitt:
So I can test for the release of v2.1.1 for compatibility with Agda 2.7.0

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - experimental
+      - 'v*-branch'
   pull_request:
     branches:
       - master
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - master
       - experimental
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

.github/workflows/ci-ubuntu.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - experimental
+      - 'v*-branch'
```
jamesmckinna, line 7: Nitpick more than anything: is there any reason to bless `-branch` as a suffix in the name of version branches, newly introduced by #2439 ? (I didn't catch it there)

It's not as if we have `master-branch` as a name... or is the intention that these names stand out because they'll be (eventually) ephemeral?

Would `v*.*` suffice instead?

MatthewDaggitt: The problem with `v*.*` is that then we have branches and tags with the same name, and Git complains in certain circumstances because of the ambiguity. 

I'd certainly be happy with `release-v*`, I'll change it :+1:

MatthewDaggitt:
Renamed it in the CI and will rename the branch now.

PR #2447, opened 2024-07-29
Added AVL.Map suggestion, explanation of MkValue args.

mars0i:
I had trouble understanding `MkValue` in the example `Tree` definition.  The proposed addition after the `Tree` definition briefly summarize what I learned (for future readers with ignorance like mine), without going into what I thought might be unnecessary detail.

I also added a suggestion and note about `Data.Tree.AVL.Map` in the "Further reading" section at the end.  I feel that `AVL.Map` satisfies the simplest use cases for the `Data.Tree.AVL`, so it's worth mentioning here.

(Sorry about the two commits with the same message.  I think I did something wrong when I tried to rebase them into one commit.  If it bothers anyone, I'll figure out how to fix it.)

### DIFF
--- a/doc/README/Data/Tree/AVL.agda
+++ b/doc/README/Data/Tree/AVL.agda
@@ -28,20 +28,20 @@ open import Relation.Binary.PropositionalEquality
 open Data.Tree.AVL <-strictTotalOrder renaming (Tree to Tree′)
 Tree = Tree′ (MkValue (Vec String) (subst (Vec String)))
 
--- The first argument to MkValue should be a function from a key
--- (a ℕ in this case) to a value type (a String vector type).
--- [Since (Vec String) is missing Vec's second parameter, a ℕ,
--- it's the required kind of function.]
+-- The first argument to `MkValue` should be a function from a key
+-- (a `ℕ` in this case) to a value type (a `String` vector type).
+-- Since `(Vec String)` is missing `Vec`'s second parameter, a `ℕ`,
+-- it's equivalent to `λ n → Vec String n`.
 
--- The second argument to MkValue is a function that can accept
+-- The second argument to `MkValue` is a function that can accept
 -- a proof that two keys are equal, and then substitute a unique key
---  for an equivalent key that is passed in.  Applying subst to the
--- key-to-value-type function passed as MkValue's first argument is
+-- for an equivalent key that is passed in.  Applying `subst` to the
+-- key-to-value-type function passed as `MkValue`'s first argument is
 -- a normal way to create such a function.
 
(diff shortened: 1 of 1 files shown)


doc/README/Data/Tree/AVL.agda
```diff
@@ -28,6 +28,21 @@ open import Relation.Binary.PropositionalEquality
 open Data.Tree.AVL <-strictTotalOrder renaming (Tree to Tree′)
 Tree = Tree′ (MkValue (Vec String) (subst (Vec String)))
 
+-- The first argument to MkValue should be a function from a key
+-- (a ℕ in this case) to a value type (a String vector type).
+-- [Since (Vec String) is missing Vec's second parameter, a ℕ,
```
MatthewDaggitt, line 33: Code is usually wrapped in backticks (e.g. `Vec String`) rather than brackets by convention. It would be good if you could wrap other code in them as well, e.g. `MkValue`

doc/README/Data/Tree/AVL.agda
```diff
@@ -28,6 +28,21 @@ open import Relation.Binary.PropositionalEquality
...
+-- it's the required kind of function.]
```
MatthewDaggitt, line 34: I would say, "it's equivalent to the function `\lambda n -> Vec String n`" (with appropriate unicode) rather than "it's the required kind of function". I'd also take the square brackets away entirely.

doc/README/Data/Tree/AVL.agda
```diff
@@ -28,6 +28,21 @@ open import Relation.Binary.PropositionalEquality
...
+
+-- The second argument to MkValue is a function that can accept
+-- a proof that two keys are equal, and then substitute a unique key
+--  for an equivalent key that is passed in.  Applying subst to the
```
MatthewDaggitt, line 38: Rogue space at the start of the line, and after "in."

MatthewDaggitt (changes requested):
Thanks, this looks great! Really appreciate the PR. I've got a few small formatting comments, but pretty minor.

mars0i:
Great--thanks.  Will do as soon as I get a moment.

MatthewDaggitt (approved):

MatthewDaggitt:
Great, thanks! Merging in.

jamesmckinna:
> Great, thanks! Merging in.

??? @MatthewDaggitt All I did was enable the workflow... but we're still waiting on a `CHANGELOG`, IIUC?

jamesmckinna:
(Sorry for last comment... deleted because getting mixed up with several other PRs :facepalm: )

PR #2449, opened 2024-07-31
Make `_<″_`/`_≤″_` inversions more lazy

jamesmckinna:
Extracted from the discussion on [this Zulip thread](https://agda.zulipchat.com/#narrow/stream/259644-newcomers/topic/What.20is.20going.20on.20with.20this.20proof.20normalization.20performance.3F)

NB:
* no `CHANGELOG`: only a potential performance improvement
* should become redundant if we manage to deprecate these relations entirely


Taneb (approved):
This seems sensible

JacquesCarette (approved):

PR #2450, opened 2024-07-31
[ add ] `Bool` action on a `RawMonoid` plus properties

jamesmckinna:
Prompted by @JacquesCarette 's desire to further systematise the refactoring in #2407 , this adds two operations which appear *ad hoc* in the normal form semantics for `IdempotentCommutativeMonoid`s, but are yet more instances of a `(Raw)MonoidAction` cf. #2348 / #2350 by analogy with the corresponding actions in the normal form semantics for `CommutativeMonoid`s. 

NB 
* Nat-action arises because (Nat, _+_, 0) is the free commutative monoid on one generator...
* similarly, (Bool, /\, true) is, I think, the free idempotent commutative monoid on one generator?

Could be left as DRAFT, pending further work cf. #2351 , but worth inviting review now, I think.

NB. possible issues:
* names of things: current notation is *terrible*, unfortunately, so suggestions welcome!
* left-/right- symmetric versions, ie. `x + n × y` as well as `n × x + y` etc.
* possible refactoring of the existing definitions as instances of these generalised 'affine' ones?
* refactoring not yet applied to `Algebra.Solver.*Monoid` etc.
* ... further systematisation of existing `Algebra.Definitions.RawMonoid._×_` operation etc. possible as well?
* Fairbairn threshhold?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,20 @@ Additions to existing modules
   ⊕-∧-booleanRing   : BooleanRing _ _
   ```
 
+* In `Algebra.Module.Properties.LeftModule`:
+  ```agda
+  -1#*ₗm≈-ᴹm : ∀ m → - 1# *ₗ m ≈ᴹ -ᴹ m
+  -‿distrib-*ₗ : ∀ r m → - r *ₗ m ≈ᴹ -ᴹ (r *ₗ m)
+  -ᴹ‿distrib-*ₗ : ∀ r m → r *ₗ (-ᴹ m) ≈ᴹ -ᴹ (r *ₗ m)
+  ```
+
+* In `Algebra.Module.Properties.RightModule`:
+  ```agda
+  -1#*ₗm≈-ᴹm : m*ᵣ-1#≈-ᴹm : ∀ m → m *ᵣ (- 1#) ≈ᴹ -ᴹ m
+  -‿distrib-*ᵣ : ∀ m r → m *ᵣ (- r) ≈ᴹ -ᴹ (m *ᵣ r)
+  -ᴹ‿distrib-*ᵣ : ∀ m r → (-ᴹ m) *ᵣ r ≈ᴹ -ᴹ (m *ᵣ r)
+  ```
+
 * In `Algebra.Properties.Monoid.Mult` properties of the Boolean action on a RawMonoid:
   ```agda
   ?>₀-homo-true  : true ?>₀ x ≈ x
@@ -331,6 +345,17 @@ Additions to existing modules
   ```agda
   ⟨_⟩⊢_ : (A → B) → Pred A ℓ → Pred B _
   [_]⊢_ : (A → B) → Pred A ℓ → Pred B _
+  ```
+
+* In `Relation.Unary.Properties`
+  ```agda
+  _map-⊢_   : P ⊆ Q → f ⊢ P ⊆ f ⊢ Q
+  map-⟨_⟩⊢_ : P ⊆ Q → ⟨ f ⟩⊢ P ⊆ ⟨ f ⟩⊢ Q
+  map-[_]⊢_ : P ⊆ Q → [ f ]⊢ P ⊆ [ f ]⊢ Q
+  ⟨_⟩⊢⁻_    : ⟨ f ⟩⊢ P ⊆ Q → P ⊆ f ⊢ Q
+  ⟨_⟩⊢⁺_    : P ⊆ f ⊢ Q → ⟨ f ⟩⊢ P ⊆ Q
+  [_]⊢⁻_    : Q ⊆ [ f ]⊢ P → f ⊢ Q ⊆ P
+  [_]⊢⁺_    : f ⊢ Q ⊆ P → Q ⊆ [ f ]⊢ P
   ```
 
 * In `System.Random`:
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -76,8 +76,7 @@ jobs:
           if [[ '${{ github.ref }}' == 'refs/heads/experimental' \
              || '${{ github.base_ref }}' == 'experimental' ]]; then
             # Pick Agda version for experimental
-            echo "AGDA_COMMIT=3d04bacca842729f9c0869b9287256321b5f450f" >> "${GITHUB_ENV}";
-              # Andreas, 2025-10-07: 3d04bacca842729f9c0869b9287256321b5f450f is tags/v2.8.0
+            echo "AGDA_COMMIT=tags/v2.8.0" >> "${GITHUB_ENV}";
             echo "AGDA_HTML_DIR=html/experimental" >> "${GITHUB_ENV}"
           else
             # Pick Agda version for master
@@ -170,8 +169,7 @@ jobs:
 
       - name: Golden testing
         run: |
-          ${{ env.CABAL_V1_INSTALL }} clock
-          make testsuite INTERACTIVE='' AGDA_EXEC='~/.cabal/bin/agda'
+          make testsuite INTERACTIVE='' AGDA_EXEC='agda' GHC_EXEC='ghc'
 
 
 ########################################################################
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ GenerateEverything
 Haskell
 html
 log
+logs/
 MAlonzo
 output
 runtests
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,3 +1,4 @@
+GHC_EXEC ?= ghc
 AGDA_EXEC ?= agda
 AGDA_OPTIONS=-Werror
 AGDA_RTS_OPTIONS=+RTS -M4.0G -H3.5G -A128M -RTS
@@ -14,7 +15,7 @@ test: doc/Everything.agda check-whitespace
 	cd doc && $(AGDA) README.agda
 
 testsuite:
-	$(MAKE) -C tests test AGDA="$(AGDA)" AGDA_EXEC="$(AGDA_EXEC)" only=$(only)
+	$(MAKE) -C tests test GHC_EXEC="$(GHC_EXEC)" AGDA="$(AGDA)" only=$(only)
 
 fix-whitespace:
 	$(CABAL_EXEC) exec -- fix-whitespace
--- a/README.md
+++ b/README.md
@@ -32,6 +32,17 @@ in glorious clickable HTML.
 
 ## Installation instructions
 
+### Automated installation (currently experimental)
+
+On Unix-style systems, it should be as simple as running the following command in
+a shell:
+
+```shell
+sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh)"
+```
+
+### Manual installation
+
 See the [installation instructions](https://github.com/agda/agda-stdlib/blob/master/doc/installation-guide.md) for the latest version of the standard library.
 
 #### Old versions of Agda
--- a/doc/installation-guide.md
+++ b/doc/installation-guide.md
@@ -3,6 +3,21 @@ Installation instructions
 
 Note: the full story on installing Agda libraries can be found at [readthedocs](http://agda.readthedocs.io/en/latest/tools/package-system.html).
 
+
+## Automated installation on Unix-style systems
+
+Running the following command in a terminal will download the
+[installation script](../stdlib-install.sh) and run it. It will
+ask you some questions about your setup and automatically install
+a version of the standard library compatible with your installed
+version of Agda.
+
+```shell
+sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh)"
+```
+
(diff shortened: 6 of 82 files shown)


jamesmckinna:
Ooops. Moving to DRAFT while I rethink this.

gallais:
Conor's disembodied voice tells me we should use a non-symmetric infix symbol for this non-symmetric action.

jamesmckinna:
> Conor's disembodied voice tells me we should use a non-symmetric infix symbol for this non-symmetric action.

I'm genuinely torn over this. I understand, appreciate, and even largely agree with, the doctrine (prevalent in WG2.1) that non-symmetric/commutative operations should be given non-symmetric notation. Nevertheless, we don't insist that that a `Group` operation be given a non-symmetric symbol... even if we *do* distinguish it as `_+_` (or something else 'obviously connoting commutativity') in the `AbelianGroup` case.

Nevertheless, I think we do ourselves (or else: type theory as a principled metalanguage for the prosecution of such doctrine) a disservice by taking the 'old-fashioned' CFG view of notation not coming *a priori* with a type attached.

The case here, as more generally in the [draft PR](https://github.com/agda/agda-stdlib/pull/2350) on `Algebra.Action`, is that the point of an action is that it is *extending* the domain of applicability of a given operation (however symbolised) to the structure on which it acts... and the generalised associativity property of a `MagmaAction` is *precisely* where the extended notation bites, visually/mnemonically at least.

You might object that the `Bool` action defined here, or `Algebra.Definitions.RawMonoid. _×_` aren't in and of themselves 'commutative'... but that obscures the fact that their `flip`ped counterparts are in fact extensionally equal to these definitions.

So perhaps the way out of the bottle is to tag the symbol with `ˡ` and `ʳ` annotations, and use infix symbol positions to record where the tail-recursive accumulator argument should go in the extended notation, yielding four compound symbols in all?

I guess I could live with that... modulo leaving `_×_` unchanged as 'legacy' (yuk)?

src/Algebra/Properties/Monoid/Mult.agda
```diff
@@ -78,3 +79,16 @@ open import Algebra.Definitions.RawMonoid rawMonoid public
   n × x + m × n × x     ≈⟨ +-congˡ (×-assocˡ x m n) ⟩
   n × x + (m ℕ.* n) × x ≈⟨ ×-homo-+ x n (m ℕ.* n) ⟨
   (suc m ℕ.* n) × x     ∎
+
+-- _∧_ is homomorphic with respect to _Bool∧_.
```
MatthewDaggitt, line 83: The comment should be `Bool._∧_`

src/Algebra/Properties/Monoid/Mult.agda
```diff
@@ -78,3 +79,16 @@ open import Algebra.Definitions.RawMonoid rawMonoid public
...
+
+∧-homo-true : ∀ x → true ∧ x ≈ x
+∧-homo-true x = refl
+
+∧-assocˡ : ∀ b b′ x → b ∧ (b′ ∧ x) ≈ (b Bool.∧ b′) ∧ x
+∧-assocˡ false b x = refl
+∧-assocˡ true  b x = refl
+
+∧∙-≗∧+ : ∀ b x y → b ∧′ x ∙ y ≈ (b ∧ x) + y
```
MatthewDaggitt, line 92: This is a very incomprehensible name. Maybe `b∧x∙y≈b∧x+y` would be better?

MatthewDaggitt (commented):
Otherwise looks good.

src/Algebra/Definitions/RawMonoid.agda
```diff
@@ -65,3 +66,18 @@ suc n ×′ x = n ×′ x + x
...
 sum = Vector.foldr _+_ 0#
+
+------------------------------------------------------------------------
+-- 'Conjunction' with a Boolean: action of the Boolean (true,∧)-rawMonoid
+------------------------------------------------------------------------
+
+infixr 8 _∧_
+
+_∧_ : Bool → Carrier → Carrier
+b ∧ x = if b then x else 0#
```
JacquesCarette, line 77: Strictly speaking, this could be defined for any `Pointed` structure with an apartness relation, as an action on a `Pointed` structure.

How is the above 'conjuction'?

src/Algebra/Definitions/RawMonoid.agda
```diff
@@ -65,3 +66,18 @@ suc n ×′ x = n ×′ x + x
...
+
+-- tail-recursive optimisation
```
JacquesCarette, line 79: optimisation of *what*? It doesn't seem like an optimisation of the above? What am I missing?

src/Algebra/Properties/Monoid/Mult.agda
```diff
@@ -78,3 +79,16 @@ open import Algebra.Definitions.RawMonoid rawMonoid public
   n × x + m × n × x     ≈⟨ +-congˡ (×-assocˡ x m n) ⟩
   n × x + (m ℕ.* n) × x ≈⟨ ×-homo-+ x n (m ℕ.* n) ⟨
   (suc m ℕ.* n) × x     ∎
+
+-- _∧_ is homomorphic with respect to _Bool∧_.
+
+∧-homo-true : ∀ x → true ∧ x ≈ x
+∧-homo-true x = refl
```
JacquesCarette, line 86: Here and below: I quite like the use of `_` for arguments that you're never going to look at in the rhs.

JacquesCarette (changes requested):

src/Algebra/Definitions/RawMonoid.agda
jamesmckinna: OK, I think that's fair... I was being pretty *ad hoc* in wanting to add this by analogy with the `Nat` action already defined there, as then these two instances of a phenomenon (eventually: wreath product) are the relevant ones for the `Algebra.Solver.*CommutativeMonoid` refactorings #2407 #2457 ... and I'm still unhappy about the general picture for `Pointed` things in the library, so don't (yet) want to reopen that can of worms...

As for it being 'conjunction', well it clearly generalises it: for `Carrier` being that of the `Monoid` structure on `Bool` given by `_∨_` with unit `false`, 
```agda
b ∧ x = if b then x else false
```
is a definition of 'conjunction'... yes!?

src/Algebra/Definitions/RawMonoid.agda
jamesmckinna: As with the TCOptimised forms of (monoid) sum and multiplication, the 'optimisation' lies in avoiding having to explicitly track terms of the form `x + 0#` when these can be computed away instead to `x`... But I admit I haven't *quite* got my story straight yet... ;-)

jamesmckinna: As witnessed by `b∧x∙y≈b∧x+y`...

src/Algebra/Definitions/RawMonoid.agda
JacquesCarette: Yes, that is the operational definition of 'conjunction' on `Bool`. 

I guess that I was thinking along the lines of "if b is the least element then x else some-distinguished-point" as the general meaning. It feels more action-y. 'conjunction' feels like it's a coincidence when `Carrier` is `Bool`.

src/Algebra/Definitions/RawMonoid.agda
JacquesCarette: It still feels like a stretch. I do like the optimization of an action not doing a (statically known) null operation, a lot. I'm still worried that this particular case is a coincidence rather than fundamental!

jamesmckinna: Well I think it extends to the `Nat` cases as well (but obscured in the current presentation, which doesn't introduce the tail-recursive forms...)

JacquesCarette:
What prevents this from losing its DRAFT status?

jamesmckinna:
> What prevents this from losing its DRAFT status?

Off the top of my head:
* no bandwidth to think about the problem... incl revisiting
* does it solve the problem I wanted it to?
* does it help with #2457 as I had hoped?

All of the above are slightly beyond my capacity to think right now, so `DRAFT` seemed to safest way to mark "I'll come back to this (at some point)"

jamesmckinna:
Notes to self (and any other interested parties here):
* #2892 provides more primitive infrastructure for definitions such as those given here?
* wreath product constructions cf. #2351
* ... 'actions induced by homomorphisms' more generally

It may well be that this all needs reorganisation down the road, at which point it may be clearer (than it is to me now) what would be necessary/useful to isolate 'sensible' common abstractions.

src/Algebra/Properties/Monoid/Mult.agda
```diff
@@ -60,7 +61,7 @@ open import Algebra.Definitions.RawMonoid rawMonoid public
 ×-homo-+ : ∀ x m n → (m ℕ.+ n) × x ≈ m × x + n × x
 ×-homo-+ x 0       n = sym (+-identityˡ (n × x))
 ×-homo-+ x (suc m) n = begin
-  x + (m ℕ.+ n) × x    ≈⟨ +-cong refl (×-homo-+ x m n) ⟩
+  x + (m ℕ.+ n) × x    ≈⟨ +-congˡ (×-homo-+ x m n) ⟩
```
JacquesCarette: cong followed by sym-assoc is what some of those combinators for Semigroup does in one go.

CHANGELOG.md
```diff
@@ -48,6 +48,19 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Definitions.RawMonoid` action of a Boolean on a RawMonoid:
+  ```agda
+  _∧_    : Bool → Carrier → Carrier
```
JacquesCarette: I'm with @gallais here (and Conor too): I find this symmetric symbol quite misleading.

src/Algebra/Definitions/RawMonoid.agda
JacquesCarette: Revisiting: maybe better thought as a 'guarded' value?

CHANGELOG.md
jamesmckinna: Yes, I wonder if these things could be (appropriately) re-notated in terms of #2894 , for all its faults?

src/Algebra/Definitions/RawMonoid.agda
jamesmckinna: Yes, I guess that this is a classic instance of (idealised) haskell's `MonadZero.guard` (if I've got this right; hackage seems to suggest that's not quite the case...)

gallais: `b ∧ x` is essentially the pattern `x <$ guard b` which AFAIK does not have a name
but is very useful.

CHANGELOG.md
gallais: Could be something like `?>` by analogy to `$>`.

jamesmckinna: Apologies for brain fade... but where is `$>` defined? (in Agda or in haskell...)

jamesmckinna: But I otherwise quite like `?>` (although `?` is usually a postfix cue that we are using `Decidable` things... not sure about prefix uses?)

gallais: `$>` is `Functor m => m a -> b -> m b`. `<` and `>` always point to the value that is kept in the result.

src/Algebra/Properties/Monoid/Mult.agda
jamesmckinna: But such refactorings are out of scope for this PR, surely?

Ah... caught on my own tidying mentality; this was part of the original PR, so should either be pursued further, or else reverted and done elsewhere as part of using `Semigroup` properties. Sigh.

src/Algebra/Properties/Monoid/Mult.agda
```diff
@@ -59,10 +61,7 @@ open import Algebra.Definitions.RawMonoid rawMonoid public
 
 ×-homo-+ : ∀ x m n → (m ℕ.+ n) × x ≈ m × x + n × x
 ×-homo-+ x 0       n = sym (+-identityˡ (n × x))
-×-homo-+ x (suc m) n = begin
-  x + (m ℕ.+ n) × x    ≈⟨ +-cong refl (×-homo-+ x m n) ⟩
-  x + (m × x + n × x)  ≈⟨ sym (+-assoc x (m × x) (n × x)) ⟩
-  x + m × x + n × x    ∎
+×-homo-+ x (suc m) n = sym (uv≈w⇒xu∙v≈xw (sym (×-homo-+ x m n)) x)
```
jamesmckinna, line 64: Refactored following @JacquesCarette 's suggestion.

jamesmckinna:
Converting back from `DRAFT`:
* fixed proofs (incl. @JacquesCarette 's suggestions)
* fixed notation to be (more like) @gallais 's suggestion (subscript-0 for the 'common' case; I don't really like the use of primes in the names for the accumulator generalisations in `Algebra.Definitions.RawMonoid`...)
* fixed merge conflicts to resolve `CHANGELOG` problems etc.

I never quite got as far as finding a common generalisation between the `Nat`- and `Bool`-actions, I think that will need to wait until #2350 eventually (!?) lands, so for now, I think this should just stand as-is, modulo nitpicking about names, which I'd be happy to engage with!

JacquesCarette:
Still needs a second approver...

PR #2456, opened 2024-08-13
Proof that `pred` is right adjoint to `suc` on `ℕ._≤_`

jamesmckinna:
I'm reluctant to add yet more to `Data.Nat.Properties`, but these lemmas seemed missing, and offered a sweet-spot in an application I've been working on, and as a by-product the left-to-right direction permits a possibly useful refactoring of `≤-pred`.

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -272,8 +272,11 @@ _≥?_ = flip _≤?_
 s≤s-injective : {p q : m ≤ n} → s≤s p ≡ s≤s q → p ≡ q
 s≤s-injective refl = refl
 
+sm≤n⇒m≤pn : suc m ≤ n → m ≤ pred n
+sm≤n⇒m≤pn {n = suc _} = s≤s⁻¹
+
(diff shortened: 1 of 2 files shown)


src/Data/Nat/Properties.agda
```diff
@@ -272,8 +272,14 @@ _≥?_ = flip _≤?_
 s≤s-injective : {p q : m ≤ n} → s≤s p ≡ s≤s q → p ≡ q
 s≤s-injective refl = refl
 
+sm≤n⇒m≤pn : suc m ≤ n → m ≤ pred n
```
MatthewDaggitt, line 275: We usually use the full `suc` and `pred` in lemma names, i.e. `suc[m]≤n⇒m≤pred[n]`

MatthewDaggitt (changes requested):
Other than that looks good!

gallais (approved):

JacquesCarette (approved):

MatthewDaggitt (approved):

PR #2460, opened 2024-08-15
Add a couple lemmas about product in Data.List.Properties

dolio:
These are a couple of lemmas I found useful when trying to formalize Euclid's proof of the infinitude of primes.
- the product of non-zero naturals is non-zero
-  any element of a list of non-zero naturals is less-or-equal to the product of the list

Hopefully I didn't just miss these somewhere else, but the divisibility result above this was about the only theorem I could find about `product`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Data.List.Properties`:
+  ```agda
+  nonZero-product : All NonZero ns → NonZero (product ns)
+  ∈⇒≤product : n ∈ ns → All NonZero ns → n ≤ product ns
+  ```
+
 * In `Data.List.Relation.Unary.All`:
   ```agda
   search : Decidable P → ∀ xs → All (∁ P) xs ⊎ Any P xs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,8 +50,8 @@ Additions to existing modules
 
 * In `Data.List.Properties`:
   ```agda
-  nonZero-product : All NonZero ns → NonZero (product ns)
-  ∈⇒≤product : n ∈ ns → All NonZero ns → n ≤ product ns
+  product≢0 : All NonZero ns → NonZero (product ns)
+  ∈⇒≤product : All NonZero ns → n ∈ ns → n ≤ product ns
   ```
 
 * In `Data.List.Relation.Unary.All`:
@@ -73,6 +73,8 @@ Additions to existing modules
 
 * New lemmas in `Data.Nat.Properties`: adjunction between `suc` and `pred`
   ```agda
+  m≤n⇒m≤n*o : ∀ o .{{_ : NonZero o}} → m ≤ n → m ≤ n * o
+  m≤n⇒m≤o*n : ∀ o .{{_ : NonZero o}} → m ≤ n → m ≤ o * n
   suc[m]≤n⇒m≤pred[n] : suc m ≤ n → m ≤ pred n
   m≤pred[n]⇒suc[m]≤n : .{{NonZero n}} → m ≤ pred n → suc m ≤ n
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,10 +71,14 @@ Additions to existing modules
   ++⁺ˡ : Reflexive R → ∀ zs → (_++ zs) Preserves (Pointwise R) ⟶ (Pointwise R)
   ```
 
-* New lemmas in `Data.Nat.Properties`: adjunction between `suc` and `pred`
+* New lemmas in `Data.Nat.Properties`:
   ```agda
   m≤n⇒m≤n*o : ∀ o .{{_ : NonZero o}} → m ≤ n → m ≤ n * o
   m≤n⇒m≤o*n : ∀ o .{{_ : NonZero o}} → m ≤ n → m ≤ o * n
+  ```
+
+  adjunction between `suc` and `pred`
+  ```agda
   suc[m]≤n⇒m≤pred[n] : suc m ≤ n → m ≤ pred n
   m≤pred[n]⇒suc[m]≤n : .{{NonZero n}} → m ≤ pred n → suc m ≤ n
   ```
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -842,25 +842,20 @@ sum-++ (x ∷ xs) ys = begin
 ∈⇒∣product {n} {n ∷ ns} (here  refl) = divides (product ns) (*-comm n (product ns))
 ∈⇒∣product {n} {m ∷ ns} (there n∈ns) = ∣n⇒∣m*n m (∈⇒∣product n∈ns)
 
-nonZero-product : ∀ {ns} → All NonZero ns → NonZero (product ns)
-nonZero-product [] = _
-nonZero-product {n ∷ ns} (nzn ∷ nzns) = m*n≢0 n (product ns)
+product≢0 : ∀ {ns} → All NonZero ns → NonZero (product ns)
+product≢0 [] = _
+product≢0 {n ∷ ns} (nzn ∷ nzns) = m*n≢0 n (product ns)
   where instance
     _ = nzn
-    _ = nonZero-product nzns
+    _ = product≢0 nzns
 
-∈⇒≤product : ∀ {n ns} → n ∈ ns → All NonZero ns → n ≤ product ns
-∈⇒≤product {n} {m ∷ ms} (here n≡m) (_ ∷ nzms) rewrite n≡m =
+∈⇒≤product : ∀ {n ns} → All NonZero ns → n ∈ ns → n ≤ product ns
+∈⇒≤product {n} {m ∷ ms} (_ ∷ nzms) (here refl) =
   m≤m*n m (product ms)
-  where instance _ = nonZero-product nzms
-∈⇒≤product {n} {m ∷ ms} (there n∈ns) (nz ∷ nzns) = begin-ordered
-  n               ≤⟨ ∈⇒≤product n∈ns nzns ⟩
-  product ms      ≤⟨ m≤n*m (product ms) m ⟩
-  m * product ms  ∎-ordered
-  where
-  instance _ = nz
-  open ≤-Reasoning using (step-≤)
-    renaming ( begin_ to begin-ordered_ ; _∎ to _∎-ordered)
+  where instance _ = product≢0 nzms
+∈⇒≤product {n} {m ∷ ms} (nz ∷ nzns) (there n∈ns) =
+  m≤n⇒m≤o*n m (∈⇒≤product nzns n∈ns)
+  where instance _ = nz
 
 
 ------------------------------------------------------------------------
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -842,6 +842,19 @@ sum-++ (x ∷ xs) ys = begin
 ∈⇒∣product {n} {n ∷ ns} (here  refl) = divides (product ns) (*-comm n (product ns))
 ∈⇒∣product {n} {m ∷ ns} (there n∈ns) = ∣n⇒∣m*n m (∈⇒∣product n∈ns)
 
+nonZero-product : ∀ {ns} → All NonZero ns → NonZero (product ns)
+nonZero-product [] = _
+nonZero-product (nzn ∷ nzns) = m*n≢0 _ _ ⦃ nzn ⦄ ⦃ nonZero-product nzns ⦄
+
+∈⇒≤product : ∀ {n ns} → n ∈ ns → All NonZero ns → n ≤ product ns
+∈⇒≤product {n} {n ∷ ms} (here  refl) (_ ∷ nzms) =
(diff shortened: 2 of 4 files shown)


src/Data/List/Properties.agda
```diff
@@ -842,6 +842,19 @@ sum-++ (x ∷ xs) ys = begin
 ∈⇒∣product {n} {n ∷ ns} (here  refl) = divides (product ns) (*-comm n (product ns))
 ∈⇒∣product {n} {m ∷ ns} (there n∈ns) = ∣n⇒∣m*n m (∈⇒∣product n∈ns)
 
+nonZero-product : ∀ {ns} → All NonZero ns → NonZero (product ns)
+nonZero-product [] = _
+nonZero-product (nzn ∷ nzns) = m*n≢0 _ _ ⦃ nzn ⦄ ⦃ nonZero-product nzns ⦄
+
+∈⇒≤product : ∀ {n ns} → n ∈ ns → All NonZero ns → n ≤ product ns
+∈⇒≤product {n} {n ∷ ms} (here  refl) (_ ∷ nzms) =
+  ≤-trans (≤-reflexive (sym (*-identityʳ n)))
```
JacquesCarette, line 851: I think using `≤-Reasoning` would make this proof (and the one below) much clearer.

JacquesCarette (changes requested):

src/Data/List/Properties.agda
```diff
@@ -842,6 +842,19 @@ sum-++ (x ∷ xs) ys = begin
 ∈⇒∣product {n} {n ∷ ns} (here  refl) = divides (product ns) (*-comm n (product ns))
 ∈⇒∣product {n} {m ∷ ns} (there n∈ns) = ∣n⇒∣m*n m (∈⇒∣product n∈ns)
 
+nonZero-product : ∀ {ns} → All NonZero ns → NonZero (product ns)
```
jamesmckinna, line 845: This proof was described [here](https://github.com/agda/agda-stdlib/pull/1969#discussion_r1435032512) in the course of reviewing @Taneb 's #1969 ...
... and I think is a bit cleaner wrt the use of instance inference?

src/Data/List/Properties.agda
jamesmckinna: It might also permit a proof in which you don't force evaluation of the proof in the `here` pattern to be `refl`? I haven't checked this, though!

jamesmckinna:
And: `CHANGELOG` entries please!

src/Data/List/Properties.agda
```diff
@@ -842,6 +842,19 @@ sum-++ (x ∷ xs) ys = begin
...
+nonZero-product [] = _
+nonZero-product (nzn ∷ nzns) = m*n≢0 _ _ ⦃ nzn ⦄ ⦃ nonZero-product nzns ⦄
+
+∈⇒≤product : ∀ {n ns} → n ∈ ns → All NonZero ns → n ≤ product ns
+∈⇒≤product {n} {n ∷ ms} (here  refl) (_ ∷ nzms) =
+  ≤-trans (≤-reflexive (sym (*-identityʳ n)))
+          (*-monoʳ-≤ n (>-nonZero⁻¹ (product ms)))
+  where instance _ = nonZero-product nzms
+∈⇒≤product {n} {m ∷ ms} (there n∈ns) (nz ∷ nzns) =
+  ≤-trans (≤-reflexive (sym (+-identityʳ n)))
```
jamesmckinna, line 855: [DRY] repetition here of a sub-term/proof you usein the previous clause... lift out as a lemma, or find the relevant thing if it exists, in `Data.Nat.Properties`?

src/Data/List/Properties.agda
dolio: I can do this, but I don't know if it's cleaner. The syntax for the cleanest version conflicts with the `≡-Reasoning` that's opened earlier in the module, so I have to redefine the equation part of the syntax to use it. Does that still sound worth it?

jamesmckinna: Alternatively, use `m≤m*n` in the`here` branch, and `m≤n*m` in the `there` branch...?

As for the global opening of a reasoning module, and clashing with a local use of another, it's a fair point... we would now advise against it in the style guide, but this module carries way too much baggage (and for `Data.Nat.Properties` even more so). but I think the above suggestions finesse all this...?

src/Data/List/Properties.agda
jamesmckinna: See above

src/Data/List/Properties.agda
dolio: Okay, yeah, those do the job well.

CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Data.List.Properties`:
+  ```agda
+  nonZero-product : All NonZero ns → NonZero (product ns)
```
jamesmckinna, line 53: I think that `product≢0 ` is a better name, for consistency with other `*≢0 ` lemmas in `Data.Nat.Properties`?

src/Data/List/Properties.agda
jamesmckinna: And naming: see above?

src/Data/List/Properties.agda
```diff
@@ -842,6 +842,27 @@ sum-++ (x ∷ xs) ys = begin
...
+  m≤m*n m (product ms)
+  where instance _ = nonZero-product nzms
+∈⇒≤product {n} {m ∷ ms} (there n∈ns) (nz ∷ nzns) = begin-ordered
+  n               ≤⟨ ∈⇒≤product n∈ns nzns ⟩
+  product ms      ≤⟨ m≤n*m (product ms) m ⟩
+  m * product ms  ∎-ordered
+  where
+  instance _ = nz
+  open ≤-Reasoning using (step-≤)
+    renaming ( begin_ to begin-ordered_ ; _∎ to _∎-ordered)
```
jamesmckinna, line 863: Oh dear, yes, that `renaming` is a bit heavy/unpleasant!
Would
```agda
renaming ( begin_ to begin-≤_ ; _∎ to _≤-∎)
```
be any better?

jamesmckinna (approved):
This seems fine as it is, but my nitpicks below might streamline things a bit, aesthetically at least?

src/Data/List/Properties.agda
jamesmckinna: Maybe we should simply refactor `Data.Nat.Properties` to have lemmas of the form
```agda
m≤n⇒m≤o*n : {{NonZero o}} → m ≤ n → m ≤ o * n
m≤n⇒m≤n*o : {{NonZero o}} → m ≤ n → m ≤ n * o
```
and thereby localise all the `Nat`-based `≤-Reasoning` where it belongs... rather than in `Data.List.Properties`?

jamesmckinna:
NB. It looks as though you have forked the main repo and then done this PR directly on *your* `master` branch`... *much* better to do PRs independently on separate branches off your `master`... as described in section 5 of `HACKING`...

CHANGELOG.md
jamesmckinna: And that also means that [`Data.Nat.Primality`](https://github.com/agda/agda-stdlib/blob/0da5ba7629ca1a911958f155fbecf2cc85646c4a/src/Data/Nat/Primality.agda#L325-L330) can be trivially refactored to make use of this lemma...

src/Data/List/Properties.agda
JacquesCarette: I agree that this is too heavyweight. Probably back to the `trans` version is going to be cleanest.

jamesmckinna: Or: as indicated above: localise the reasoning with `ℕ._≤_` and `ℕ._*_` in `Data.Nat.Properties`, and then appeal to it when proving properties of `Data.List.Base.product` in `Data.List.Properties`... noting that the lemmas I propose above are 'hidden' appeals to transitivity on top of the existing `m≤m*n` etc...

src/Data/List/Properties.agda
dolio: Okay, I'll try that out.

dolio: Okay, I've updated to take this approach and I think address the other comments.

I also noticed that there were already `<` versions of these m-n-o lemmas that are special cases, so I replaced their implementation with just the `≤` one.

src/Data/Nat/Properties.agda
```diff
@@ -1008,13 +1014,10 @@ m<m*n m@(suc m-1) n@(suc (suc n-2)) (s≤s (s≤s _)) = begin-strict
   m * n       ∎
 
 m<n⇒m<n*o : ∀ o .{{_ : NonZero o}} → m < n → m < n * o
-m<n⇒m<n*o {n = n} o m<n = <-≤-trans m<n (m≤m*n n o)
+m<n⇒m<n*o = m≤n⇒m≤n*o
```
jamesmckinna, line 1017: Nice!

src/Data/Nat/Properties.agda
```diff
@@ -1008,13 +1014,10 @@ m<m*n m@(suc m-1) n@(suc (suc n-2)) (s≤s (s≤s _)) = begin-strict
...
 
 m<n⇒m<o*n : ∀ {m n} o .{{_ : NonZero o}} → m < n → m < o * n
-m<n⇒m<o*n {m} {n} o m<n = begin-strict
-  m     <⟨ m<n⇒m<n*o o m<n ⟩
-  n * o ≡⟨ *-comm n o ⟩
-  o * n ∎
+m<n⇒m<o*n = m≤n⇒m≤o*n
```
jamesmckinna, line 1020: Even better!

src/Data/List/Properties.agda
```diff
@@ -842,25 +842,20 @@ sum-++ (x ∷ xs) ys = begin
...
-  where instance _ = nonZero-product nzms
-∈⇒≤product {n} {m ∷ ms} (there n∈ns) (nz ∷ nzns) = begin-ordered
-  n               ≤⟨ ∈⇒≤product n∈ns nzns ⟩
-  product ms      ≤⟨ m≤n*m (product ms) m ⟩
-  m * product ms  ∎-ordered
-  where
-  instance _ = nz
-  open ≤-Reasoning using (step-≤)
-    renaming ( begin_ to begin-ordered_ ; _∎ to _∎-ordered)
+  where instance _ = product≢0 nzms
```
jamesmckinna, line 855: Implicit argument `n` is never used, and you introduce an artificial change-of-name with `m`... much better to simply use `{ns = n ∷ ns}` as the first binding...
```agda
∈⇒≤product {ns = n ∷ ns} (_ ∷ nzns) (here refl) =
  m≤m*n n (product ns)
  where instance _ = product≢0 nzns
```

src/Data/List/Properties.agda
```diff
@@ -842,25 +842,20 @@ sum-++ (x ∷ xs) ys = begin
...
+∈⇒≤product {n} {m ∷ ms} (nz ∷ nzns) (there n∈ns) =
+  m≤n⇒m≤o*n m (∈⇒≤product nzns n∈ns)
+  where instance _ = nz
```
jamesmckinna, line 858: Similarly:
```suggestion
∈⇒≤product {ns = n ∷ _} (nz ∷ nzns) (there n∈ns) =
  m≤n⇒m≤o*n n (∈⇒≤product nzns n∈ns)
  where instance _ = nz
```

jamesmckinna:
Thanks for a very nice first PR! (NB better to fix up your `git` branches before the next one... ;-))
A couple of final refinements, and we should be good to go! 
@JacquesCarette are you happy with the refactoring?

JacquesCarette (approved):

src/Data/List/Properties.agda
```diff
@@ -842,6 +842,22 @@ sum-++ (x ∷ xs) ys = begin
 ∈⇒∣product {n} {n ∷ ns} (here  refl) = divides (product ns) (*-comm n (product ns))
 ∈⇒∣product {n} {m ∷ ns} (there n∈ns) = ∣n⇒∣m*n m (∈⇒∣product n∈ns)
 
+product≢0 : ∀ {ns} → All NonZero ns → NonZero (product ns)
+product≢0 [] = _
+product≢0 {n ∷ ns} (nzn ∷ nzns) = m*n≢0 n (product ns)
+  where instance
+    _ = nzn
+    _ = product≢0 nzns
```
MatthewDaggitt, line 850: Sorry to unmerge, but would writing this line as:
```agda
product≢0 {n ∷ ns} (n≢0 ∷ ns≢0) = m*n≢0 {{n≢0}} {{product≢0 ns≢0}} 0 n (product ns)
```
be much cleaner than the four lines we currently have? Same with the proof below.

MatthewDaggitt: Also marginally more efficient as Agda doesn't have to do instance search...

jamesmckinna: @MatthewDaggitt I think on reflection that you're right, and my apologies to @dolio for my earlier suggestion(s) which led to the changes towards this more verbose style. 

I'm not necessarily convinced that inlining explicit instances is more readable, but for sure it spares the checker the cost of the (otherwise quite narrow, because local/immediate) search.

PR #2465, opened 2024-08-29
Add two lemmas to `Data.List.Membership.Setoid.Properties` 

### LINKED ISSUE #2463: two proofs suggested for `Data.List.Membership.Setoid.Properties`
I suggest the following two proofs for standard library
(probably for Data.List.Membership.Setoid.Properties).
These are is over ``S : Setoid _ _``.
```
any∈sym :  ∀ {xs ys} → Any (_∈ ys) xs → Any (_∈ xs) ys
any∈sym {xs} {ys} any-xs∈ys =
  let (x , x∈xs , x∈ys) = find {P = (_∈ ys)} any-xs∈ys
  in
  lose {P = (_∈ xs)} (∈-resp-≈ S) x∈ys x∈xs

all∉sym :  ∀ {xs ys} → All (_∉ ys) xs → All (_∉ xs) ys
all∉sym {xs} {ys} all-xs∉ys =
                  ¬Any⇒All¬ ys (¬any-xs∈ys ∘ any∈sym)
  where
  ¬any-xs∈ys =  All¬⇒¬Any all-xs∉ys
```
I do not set dashes, for example, write `any∈sym` rather than `any-∈-sym`.
Because
(1) this makes the proof texts shorter,
(2) the symbols like ``∈, ∉, *, _≈_, ...`` denote operators and are distinguished themselves inside a word.
But if the team insists, let them set dashes.

jamesmckinna:
Fixes #2463 

Breaks dependency cycle by introducing new `Data.List.Relation.Unary.All.Properties.Core` module.

### DIFF
--- a/src/Data/List/Membership/Setoid/Properties.agda
+++ b/src/Data/List/Membership/Setoid/Properties.agda
@@ -456,9 +456,9 @@ module _ (S : Setoid c ℓ) where
   open Setoid S using (sym)
   open Membership S
 
-  any∈sym :  ∀ {xs ys} → Any (_∈ ys) xs → Any (_∈ xs) ys
-  any∈sym = Any.swap ∘ Any.map (Any.map sym)
+  Any-∈-swap :  ∀ {xs ys} → Any (_∈ ys) xs → Any (_∈ xs) ys
+  Any-∈-swap = Any.swap ∘ Any.map (Any.map sym)
 
-  all∉sym :  ∀ {xs ys} → All (_∉ ys) xs → All (_∉ xs) ys
-  all∉sym p = All.¬Any⇒All¬ _ ((All.All¬⇒¬Any p) ∘ any∈sym)
+  All-∉-swap :  ∀ {xs ys} → All (_∉ ys) xs → All (_∉ xs) ys
+  All-∉-swap p = All.¬Any⇒All¬ _ ((All.All¬⇒¬Any p) ∘ any∈sym)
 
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):

src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -446,3 +447,18 @@ module _ (S : Setoid c ℓ) where
...
+
+------------------------------------------------------------------------
+-- Any/All symmetry wrt _∈_/_∉_
+
+module _ (S : Setoid c ℓ) where
+
+  open Setoid S using (sym)
+  open Membership S
+
+  any∈sym :  ∀ {xs ys} → Any (_∈ ys) xs → Any (_∈ xs) ys
```
MatthewDaggitt, line 459: There's got to better names somewhere. They are very similar in form to the existing lemma `All-swap` that was moved below. So how about `Any-swap-∈` and `All-swap-∉`? At least they would be consistent...

MatthewDaggitt (changes requested):

src/Data/List/Membership/Setoid/Properties.agda
jamesmckinna: Thanks for the suggestions. I'd taken the name directly from @mechvel 's original issue, without re-thinking them, so the suggestions are welcome, and these seem to fit much better with our house style!

MatthewDaggitt (approved):

PR #2468, opened 2024-08-30
Typo in `Data.List.Relation.Unary.All`

### LINKED ISSUE #2467: somethig like a typo in  lib-2.1.1-rc2
Data/List/Relation/Unary/All.agda 
starts with 
```
All.agda:module _(S : Setoid a ℓ) ...
``` 
Probably it is better to set a space after `_`

jamesmckinna:
Fixes #2467

No `CHANGELOG`.

Applies to `master` as well as to `v2.1.1`.

### DIFF
--- a/src/Data/List/Relation/Unary/All.agda
+++ b/src/Data/List/Relation/Unary/All.agda
@@ -115,7 +115,7 @@ zip = zipWith id
 unzip : All (P ∩ Q) ⊆ All P ∩ All Q
 unzip = unzipWith id
 
-module _(S : Setoid a ℓ) {P : Pred (Setoid.Carrier S) p} where
+module _ (S : Setoid a ℓ) {P : Pred (Setoid.Carrier S) p} where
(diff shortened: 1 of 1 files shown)


Taneb (approved):

jamesmckinna:
I just noticed that how/whenever this was merged previously, it seems to have been undone on `master`: (link copied just now)
https://github.com/agda/agda-stdlib/blob/577008ed887eb9d33be6eb3b216ecfdd852667ff/src/Data/List/Relation/Unary/All.agda#L197
Not quite sure how/why this happens/happened!?

JacquesCarette:
`git blame` says the current master for that line is from f0ef8a647a which is pre-fix. So it looks like it was indeed undone. Except that `git log` says that the latest merge into this file is indeed cc69ea07b !

JacquesCarette:
Mystery solved: this problem occurs twice in this file, you fixed the one at line 118 previously but didn't catch the one at line 197 !!!

PR #2478, opened 2024-09-09
fixes #2466: add equivalence proofs

### LINKED ISSUE #2466: Equivalence of characterisations of `Data.List.Membership.Setoid._∉_`
Issue #2463 got me thinking... why do we not have the equivalence anywhere, for given `x` and `xs`, of:
* `x ∉ xs = ¬ x ∈ xs` defined in `Data.List.Membership.Setoid`
* `All (x ≉_) xs` defined via `Data.List.Relation.Unary.All`

Where would such a statement live? Currently `Data.List.Membership.Setoid.Properties` defines only a `private` (moreover: unidirectional) version, as:
```agda
  private
    ∉×∈⇒≉ : ∀ {x y xs} → All (y ≉_) xs → x ∈ xs → x ≉ y
    ∉×∈⇒≉ = All.lookupWith λ y≉z x≈z x≈y → y≉z (trans (sym x≈y) x≈z)
```
But for reflexive `_≈_` this would follow immediately from the equivalence...?

jamesmckinna:
NB.:
* enabled by the refactoring in #2463 in terms of `Data.List.Relation.Unary.All.Properties.Core`
* easy refactoring of the `private` lemma
* Fairbairn threshold? The additions are (specialised) versions of imported lemmas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,8 +120,10 @@ Additions to existing modules
 
 * In `Data.List.Membership.Setoid.Properties`:
   ```agda
-  Any-∈-swap :  Any (_∈ ys) xs → Any (_∈ xs) ys
-  All-∉-swap :  All (_∉ ys) xs → All (_∉ xs) ys
+  Any-∈-swap : Any (_∈ ys) xs → Any (_∈ xs) ys
+  All-∉-swap : All (_∉ ys) xs → All (_∉ xs) ys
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

JacquesCarette:
On Fairbairn: my feeling is that the lemmas here are somehow "more findable" than the more general version, and so likely to be more useful. So I'm happy with this 'duplication'.

PR #2479, opened 2024-09-10
Lists: porting several lemmas from other libraries

omelkonian:
Porting useful list properties that we needed in the [Cardano ledger project](https://github.com/IntersectMBO/formal-ledger-specifications), several of which are themselves ported from my [personal prelude](https://github.com/omelkonian/formal-prelude/).

I am totally fine with some of those not being merged due to the 'Fairbairn threshold' or just because it is too niche of a utility to be included in the standard library; I'll let the reviewer(s) decide.

NB: each commit is somewhat independent, so reviews can proceed in a per-commit fashion (unless a change is requested globally, e.g. cosmetic changes w.r.t. to the style guide).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,18 @@ Additions to existing modules
   concatMap⁻ : Any P (concatMap f xs) → Any (Any P ∘ f) xs
   ```
 
+* In `Data.List.Relation.Unary.Unique.Setoid.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique S (x ∷ y ∷ xs) → Unique S (x ∷ xs)
+  Unique∷⇒head∉tail : Unique S (x ∷ xs) → x ∉ xs
+  ```
+
+* In `Data.List.Relation.Unary.Unique.Propositional.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique (x ∷ y ∷ xs) → Unique (x ∷ xs)
+  Unique∷⇒head∉tail : Unique (x ∷ xs) → x ∉ xs
+  ```
+
 * In `Data.List.Relation.Binary.Equality.Setoid`:
   ```agda
   ++⁺ʳ : ∀ xs → ys ≋ zs → xs ++ ys ≋ xs ++ zs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,30 +125,32 @@ Additions to existing modules
 
 * In `Data.List.Membership.Setoid.Properties`:
   ```agda
-  Any-∈-swap    :  Any (_∈ ys) xs → Any (_∈ xs) ys
-  All-∉-swap    :  All (_∉ ys) xs → All (_∉ xs) ys
-  ∈-map∘filter⁻ : y ∈₂ map f (filter P? xs) → ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
-  ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
-                  ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
-                  y ∈₂ map f (filter P? xs)
-  ∈-concatMap⁺  : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
-  ∈-concatMap⁻  : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
-  ∉[]           : x ∉ []
+  Any-∈-swap     : Any (_∈ ys) xs → Any (_∈ xs) ys
+  All-∉-swap     : All (_∉ ys) xs → All (_∉ xs) ys
+  ∈-map∘filter⁻  : y ∈₂ map f (filter P? xs) → ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
+  ∈-map∘filter⁺  : f Preserves _≈₁_ ⟶ _≈₂_ →
+                   ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
+                   y ∈₂ map f (filter P? xs)
+  ∈-concatMap⁺   : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
+  ∈-concatMap⁻   : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
+  ∉[]            : x ∉ []
+  deduplicate-∈⇔ : _≈_ Respectsʳ (flip R) → z ∈ xs ⇔ z ∈ deduplicate R? xs
   ```
 
 * In `Data.List.Membership.Propositional.Properties`:
   ```agda
-  ∈-AllPairs₂   : AllPairs R xs → x ∈ xs → y ∈ xs → x ≡ y ⊎ R x y ⊎ R y x
-  ∈-map∘filter⁻ : y ∈ map f (filter P? xs) → (∃[ x ] x ∈ xs × y ≡ f x × P x)
-  ∈-map∘filter⁺ : (∃[ x ] x ∈ xs × y ≡ f x × P x) → y ∈ map f (filter P? xs)
-  ∈-concatMap⁺  : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
-  ∈-concatMap⁻  : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
-  ∈-++          : v ∈ xs ++ ys ⇔ (v ∈ xs ⊎ v ∈ ys)
-  []∉map∷       : [] ∉ map (x ∷_) xss
-  map∷-decomp∈  : (x ∷ xs) ∈ map (y ∷_) xss → x ≡ y × xs ∈ xss
-  map∷-decomp   : xs ∈ map (y ∷_) xss → ∃[ ys ] ys ∈ xss × y ∷ ys ≡ xs
-  ∈-map∷⁻       : xs ∈ map (x ∷_) xss → x ∈ xs
-  ∉[]           : x ∉ []
+  ∈-AllPairs₂    : AllPairs R xs → x ∈ xs → y ∈ xs → x ≡ y ⊎ R x y ⊎ R y x
+  ∈-map∘filter⁻  : y ∈ map f (filter P? xs) → (∃[ x ] x ∈ xs × y ≡ f x × P x)
+  ∈-map∘filter⁺  : (∃[ x ] x ∈ xs × y ≡ f x × P x) → y ∈ map f (filter P? xs)
+  ∈-concatMap⁺   : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
+  ∈-concatMap⁻   : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
+  ++-∈⇔          : v ∈ xs ++ ys ⇔ (v ∈ xs ⊎ v ∈ ys)
+  []∉map∷        : [] ∉ map (x ∷_) xss
+  map∷-decomp∈   : (x ∷ xs) ∈ map (y ∷_) xss → x ≡ y × xs ∈ xss
+  map∷-decomp    : xs ∈ map (y ∷_) xss → ∃[ ys ] ys ∈ xss × y ∷ ys ≡ xs
+  ∈-map∷⁻        : xs ∈ map (x ∷_) xss → x ∈ xs
+  ∉[]            : x ∉ []
+  deduplicate-∈⇔ : z ∈ xs ⇔ z ∈ deduplicate _≈?_ xs
   ```
 
 * In `Data.List.Membership.Propositional.Properties.WithK`:
@@ -228,13 +230,11 @@ Additions to existing modules
 * In `Data.List.Relation.Binary.Disjoint.Setoid.Properties`:
   ```agda
   deduplicate⁺ : Disjoint S xs ys → Disjoint S (deduplicate _≟_ xs) (deduplicate _≟_ ys)
-  ∈-dedup      : x ∈ xs ⇔ x ∈ deduplicate _≟_ xs
   ```
 
 * In `Data.List.Relation.Binary.Disjoint.Propositional.Properties`:
   ```agda
   deduplicate⁺ : Disjoint xs ys → Disjoint (deduplicate _≟_ xs) (deduplicate _≟_ ys)
-  ∈-dedup      : x ∈ xs ⇔ x ∈ deduplicate _≟_ xs
   ```
 
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties.WithK`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,13 +125,19 @@ Additions to existing modules
 
 * In `Data.List.Membership.Setoid.Properties`:
   ```agda
-  Any-∈-swap :  Any (_∈ ys) xs → Any (_∈ xs) ys
-  All-∉-swap :  All (_∉ ys) xs → All (_∉ xs) ys
+  Any-∈-swap    :  Any (_∈ ys) xs → Any (_∈ xs) ys
+  All-∉-swap    :  All (_∉ ys) xs → All (_∉ xs) ys
+  ∈-map∘filter⁻ : y ∈₂ map f (filter P? xs) → ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
+  ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
+                  ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
+                  y ∈₂ map f (filter P? xs)
   ```
 
 * In `Data.List.Membership.Propositional.Properties`:
   ```agda
-  ∈-AllPairs₂ : AllPairs R xs → x ∈ xs → y ∈ xs → x ≡ y ⊎ R x y ⊎ R y x
+  ∈-AllPairs₂   : AllPairs R xs → x ∈ xs → y ∈ xs → x ≡ y ⊎ R x y ⊎ R y x
+  ∈-map∘filter⁻ : y ∈ map f (filter P? xs) → (∃[ x ] x ∈ xs × y ≡ f x × P x)
+  ∈-map∘filter⁺ : (∃[ x ] x ∈ xs × y ≡ f x × P x) → y ∈ map f (filter P? xs)
   ```
 
 * In `Data.List.Membership.Propositional.Properties.WithK`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,11 @@ Additions to existing modules
   All-∉-swap :  All (_∉ ys) xs → All (_∉ xs) ys
   ```
 
+* In `Data.List.Membership.Propositional.Properties.WithK`:
+  ```agda
+  unique∧set⇒bag : Unique xs → Unique ys → xs ∼[ set ] ys → xs ∼[ bag ] ys
+  ```
+
 * In `Data.List.Properties`:
   ```agda
   product≢0 : All NonZero ns → NonZero (product ns)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,16 @@ Additions to existing modules
   ++⁺ˡ : Reflexive R → ∀ zs → (_++ zs) Preserves (Pointwise R) ⟶ (Pointwise R)
   ```
 
+* In `Data.List.Relation.Binary.Sublist.Heterogeneous.Properties`:
+  ```agda
+  Sublist-[]-universal : Universal (Sublist R [])
+  ```
+
+* In `Data.List.Relation.Binary.Sublist.Setoid.Properties`:
+  ```agda
+  []⊆-universal : Universal ([] ⊆_)
+  ```
+
 * In `Data.Maybe.Properties`:
   ```agda
   maybe′-∘ : ∀ f g → f ∘ (maybe′ g b) ≗ maybe′ (f ∘ g) (f b)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,8 @@ Additions to existing modules
   ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
                   ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
                   y ∈₂ map f (filter P? xs)
+  ∈-concatMap⁺  : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
+  ∈-concatMap⁻  : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
   ```
 
 * In `Data.List.Membership.Propositional.Properties`:
@@ -138,6 +140,8 @@ Additions to existing modules
   ∈-AllPairs₂   : AllPairs R xs → x ∈ xs → y ∈ xs → x ≡ y ⊎ R x y ⊎ R y x
   ∈-map∘filter⁻ : y ∈ map f (filter P? xs) → (∃[ x ] x ∈ xs × y ≡ f x × P x)
   ∈-map∘filter⁺ : (∃[ x ] x ∈ xs × y ≡ f x × P x) → y ∈ map f (filter P? xs)
+  ∈-concatMap⁺  : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
+  ∈-concatMap⁻  : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
   ```
 
 * In `Data.List.Membership.Propositional.Properties.WithK`:
@@ -147,13 +151,20 @@ Additions to existing modules
 
 * In `Data.List.Properties`:
   ```agda
-  product≢0 : All NonZero ns → NonZero (product ns)
-  ∈⇒≤product : All NonZero ns → n ∈ ns → n ≤ product ns
+  product≢0    : All NonZero ns → NonZero (product ns)
+  ∈⇒≤product   : All NonZero ns → n ∈ ns → n ≤ product ns
+  concatMap-++ : concatMap f (xs ++ ys) ≡ concatMap f xs ++ concatMap f ys
   ```
 
 * In `Data.List.Relation.Unary.All`:
   ```agda
   search : Decidable P → ∀ xs → All (∁ P) xs ⊎ Any P xs
+  ```
+
+* In `Data.List.Relation.Unary.Any.Properties`:
+  ```agda
+  concatMap⁺ : Any (Any P ∘ f) xs → Any P (concatMap f xs)
+  concatMap⁻ : Any P (concatMap f xs) → Any (Any P ∘ f) xs
   ```
 
 * In `Data.List.Relation.Binary.Equality.Setoid`:
@@ -170,7 +181,7 @@ Additions to existing modules
 
 * In `Data.List.Relation.Binary.Subset.Propositional.Properties`:
   ```agda
-  concatMap⁺ : ∀ (f : A → List B) → xs ⊆ ys → concatMap f xs ⊆ concatMap f ys
+  concatMap⁺ : xs ⊆ ys → concatMap f xs ⊆ concatMap f ys
   ```
 
 * In `Data.List.Relation.Binary.Sublist.Heterogeneous.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@ Additions to existing modules
                   y ∈₂ map f (filter P? xs)
   ∈-concatMap⁺  : Any ((y ∈_) ∘ f) xs → y ∈ concatMap f xs
   ∈-concatMap⁻  : y ∈ concatMap f xs → Any ((y ∈_) ∘ f) xs
+  ∉[]           : x ∉ []
   ```
 
 * In `Data.List.Membership.Propositional.Properties`:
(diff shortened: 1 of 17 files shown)


src/Data/List/Relation/Unary/Unique/Setoid/Properties.agda
```diff
@@ -156,3 +158,21 @@ module _ (S : Setoid a ℓ) {P : Pred _ p} (P? : Decidable P) where
...
+
+------------------------------------------------------------------------
+-- ∷
+
+module _ (S : Setoid a ℓ) where
+
+  open Setoid S renaming (Carrier to A)
+  open Membership S using (_∈_; _∉_)
+
+  private variable x y : A; xs : List A
```
jamesmckinna, line 170: As with the other (excellent!) introductions of local `private variable` declarations, suggest spacing this out as:
```suggestion
  private
    variable
      x y : A
      xs : List A
```

src/Data/List/Relation/Binary/Permutation/Propositional/Properties/WithK.agda
```diff
@@ -0,0 +1,62 @@
...
+
+open import Data.Sum as Sum using (_⊎_)
+open import Data.Sum.Function.Propositional using (_⊎-cong_)
+
+open import Relation.Binary.Definitions using (DecidableEquality)
+
+------------------------------------------------------------------------
+-- deduplicate
+
+module _ {a}{A : Set a} (_≟_ : DecidableEquality A) where
```
jamesmckinna, line 36: Missing whitespace:
```suggestion
module _ {a} {A : Set a} (_≟_ : DecidableEquality A) where
```

src/Data/List/Relation/Unary/Unique/Setoid/Properties.agda
```diff
@@ -9,12 +9,14 @@
...
+import Data.List.Membership.Setoid as Membership
 open import Data.List.Membership.Setoid.Properties
 open import Data.List.Relation.Binary.Disjoint.Setoid
 open import Data.List.Relation.Binary.Disjoint.Setoid.Properties
+open import Data.List.Relation.Unary.Any using (here; there)
 open import Data.List.Relation.Unary.All as All using (All; []; _∷_)
 open import Data.List.Relation.Unary.All.Properties using (All¬⇒¬Any)
-open import Data.List.Relation.Unary.AllPairs as AllPairs using (AllPairs)
+open import Data.List.Relation.Unary.AllPairs as AllPairs using (AllPairs; _∷_)
 open import Data.List.Relation.Unary.Unique.Setoid
```
jamesmckinna, line 20: Given that your adding properties of `Unique` in this PR, it's probably worth *also* refactoring this imported module to simplify it (and its `Propositional` friend), because `Distinct` defined `private`ly there is (now) superseded by `Setoid._≉_`...

src/Data/List/Relation/Binary/Subset/Setoid/Properties.agda
```diff
@@ -41,6 +42,18 @@ private
...
+------------------------------------------------------------------------
+
+module _ (S : Setoid a ℓ) where
+
+  open Setoid S using (refl)
+  open Subset S
+
+  ⊈[] : ∀ {x xs} → x ∷ xs ⊈ []
+  ⊈[] p = Membershipₚ.∉[] S $ p (here refl)
+
```
jamesmckinna, line 56: Suggest also adding the corollary to this?
```suggestion
  ⊆[]⇒≡[] : (_⊆ []) ⋐ (_≡ [])
  ⊆[]⇒≡[] {x = []}    _ = ≡.refl
  ⊆[]⇒≡[] {x = _ ∷ _} p with () ← ⊈[] p
```
for the cost of an additional initial
```agda
open import Relation.Binary.PropositionalEquality.Core as ≡ using (_≡_)
```

src/Data/List/Relation/Binary/Subset/Propositional/Properties.agda
```diff
@@ -52,8 +52,16 @@ private
...
+    x y : A
     ws xs ys zs : List A
 
+------------------------------------------------------------------------
+-- Basics
+------------------------------------------------------------------------
+
+⊈[] : x ∷ xs ⊈ []
+⊈[] = Subset.⊈[] (setoid _)
+
```
jamesmckinna, line 64: And add its corollary, as below for `Setoid`?

src/Data/List/Relation/Binary/Disjoint/Setoid/Properties.agda
```diff
@@ -8,49 +8,65 @@
...
+  sym : Symmetric (Disjoint S)
+  sym xs#ys (v∈ys , v∈xs) = xs#ys (v∈xs , v∈ys)
 
-  open Setoid S
+  ------------------------------------------------------------------------
+  -- Relationship with other predicates
+  ------------------------------------------------------------------------
 
   Disjoint⇒AllAll : ∀ {xs ys} → Disjoint S xs ys →
                     All (λ x → All (λ y → ¬ x ≈ y) ys) xs
```
jamesmckinna, line 43: `Setoid` now defines `_≉_` so could (more) simply write
```suggestion
                    All (λ x → All (x ≉_) ys) xs
```
or even in the spirit of #2478 use `All (_∉ ys) xs` instead?

src/Data/List/Membership/Setoid/Properties.agda
```diff
@@ -363,6 +388,33 @@ module _ (S : Setoid c ℓ) {P : Pred (Carrier S) p}
...
 
+------------------------------------------------------------------------
+-- map∘filter
+
+module _
+  (S₁ : Setoid c₁ ℓ₁) (S₂ : Setoid c₂ ℓ₂)
+  {P : Pred (S₁ .Carrier) p} (P? : Decidable P) (resp : P Respects (S₁ .Setoid._≈_))
+  {f xs y} where
+
+  open Setoid     S₁ renaming (_≈_ to _≈₁_)
```
jamesmckinna, line 399: Ugh! Is it perhaps cleaner to `open Setoid S₁` first?
```suggestion
  (open Setoid S₁ renaming (Carrier to A; _≈_ to _≈₁_))
  {P : Pred A p} (P? : Decidable P) (resp : P Respects _≈₁_)
  {f xs y} where

```

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -96,14 +100,14 @@ map-mapWith∈ = Membership.map-mapWith∈ (≡.setoid _)
...
   ∈-map⁺ = Membership.∈-map⁺ (≡.setoid A) (≡.setoid B) (cong f)
 
-  ∈-map⁻ : ∀ {y xs} → y ∈ map f xs → ∃ λ x → x ∈ xs × y ≡ f x
+  ∈-map⁻ : y ∈ map f xs → ∃ λ x → x ∈ xs × y ≡ f x
   ∈-map⁻ = Membership.∈-map⁻ (≡.setoid A) (≡.setoid B)
 
-  map-∈↔ : ∀ {y xs} → (∃ λ x → x ∈ xs × y ≡ f x) ↔ y ∈ map f xs
-  map-∈↔ {y} {xs} =
+  map-∈↔ : (∃ λ x → x ∈ xs × y ≡ f x) ↔ y ∈ map f xs
+  map-∈↔ {xs}{y} =
```
jamesmckinna, line 110: I like the refactoring to use `private` variables, but here it does mean that the implicit argument changes, which is technically a `breaking` change, but I think of a kind that we have elsewhere accepted without worrying too much.

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -24,23 +25,23 @@ open import Data.List.Relation.Unary.Any.Properties
...
-open import Function.Bundles using (_↔_; _↣_; Injection)
+open import Function.Bundles using (_↔_; _↣_; Injection; _⇔_; mk⇔)
 open import Function.Related.TypeIsomorphisms using (×-comm; ∃∃↔∃∃)
 open import Function.Construct.Identity using (↔-id)
 open import Level using (Level)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Definitions as Binary hiding (Decidable)
 open import Relation.Binary.PropositionalEquality.Core as ≡
-  using (_≡_; _≢_; refl; sym; trans; cong; cong₂; resp; _≗_)
+  using (_≡_; _≢_; refl; sym; trans; cong; cong₂; resp; _≗_; subst)
```
jamesmckinna, line 44: The import (and use!) of `subst` is avoidable here: see below

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -401,3 +433,39 @@ there-injective-≢∈ : ∀ {xs} {x y z : A} {x∈xs : x ∈ xs} {y∈xs : y 
...
+map∷-decomp∈ {xss = _ ∷ _} = λ where
+  (here refl) → refl , here refl
+  (there p)   → map₂ there $ map∷-decomp∈ p
+
+map∷-decomp : xs ∈ map (y ∷_) xss → ∃[ ys ] ys ∈ xss × y ∷ ys ≡ xs
+map∷-decomp               {xss = _  ∷ _} (here refl) = -, here refl , refl
+map∷-decomp {xs = []}     {xss = _  ∷ _} (there xs∈) = ⊥-elim ([]∉map∷ xs∈)
+map∷-decomp {xs = x ∷ xs} {xss = _  ∷ _} (there xs∈) =
+  let eq , p = map∷-decomp∈ xs∈
+  in -, there p , subst (λ ◆ → ◆ ∷ _ ≡ _) eq refl
```
jamesmckinna, line 466: `subst` can avoided in favour of `cong`:
```suggestion
  in -, there p , cong (_∷ xs) (sym eq)
```

src/Data/List/Base.agda
```diff
@@ -229,6 +229,12 @@ lookup : ∀ (xs : List A) → Fin (length xs) → A
 lookup (x ∷ xs) zero    = x
 lookup (x ∷ xs) (suc i) = lookup xs i
 
+lookupMaybe : List A → ℕ → Maybe A
+lookupMaybe = λ where
+  []       _       → nothing
+  (x ∷ _)  zero    → just x
+  (_ ∷ xs) (suc n) → lookupMaybe xs n
+
```
jamesmckinna, line 237: So... how useful/necessary is this operation?
It's noticeable that you don't add any lemmas about it to `Data.List.Properties`?

CHANGELOG.md
```diff
@@ -118,23 +123,74 @@ Additions to existing modules
   Env = Vec Carrier
  ```
 
+* In `Data.List.Base`:
+  ```agda
+  lookupMaybe : List A → ℕ → Maybe A
```
jamesmckinna, line 128: See below.

CHANGELOG.md
```diff
@@ -118,23 +123,74 @@ Additions to existing modules
...
 
+* In `Data.List.Relation.Unary.Any.Properties`:
+  ```agda
+  concatMap⁺ : Any (Any P ∘ f) xs → Any P (concatMap f xs)
+  concatMap⁻ : Any P (concatMap f xs) → Any (Any P ∘ f) xs
+  ```
+
+* In `Data.List.Relation.Unary.Unique.Setoid.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique S (x ∷ y ∷ xs) → Unique S (x ∷ xs)
```
jamesmckinna, line 184: Suggest not adding this name? See comment on the module.

CHANGELOG.md
```diff
@@ -118,23 +123,74 @@ Additions to existing modules
...
+  Unique∷⇒head∉tail : Unique S (x ∷ xs) → x ∉ xs
+  ```
+
+* In `Data.List.Relation.Unary.Unique.Propositional.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique (x ∷ y ∷ xs) → Unique (x ∷ xs)
```
jamesmckinna, line 190: Ditto. re: dropping this?

CHANGELOG.md
```diff
@@ -118,23 +123,74 @@ Additions to existing modules
...
+* In `Data.List.Relation.Unary.Any.Properties`:
+  ```agda
+  concatMap⁺ : Any (Any P ∘ f) xs → Any P (concatMap f xs)
+  concatMap⁻ : Any P (concatMap f xs) → Any (Any P ∘ f) xs
+  ```
+
+* In `Data.List.Relation.Unary.Unique.Setoid.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique S (x ∷ y ∷ xs) → Unique S (x ∷ xs)
+  Unique∷⇒head∉tail : Unique S (x ∷ xs) → x ∉ xs
```
jamesmckinna, line 185: This PR introduces a lot of new names, some of which, like this one, are pretty hard to read to begin with... so maybe an alternative is preferable?
```suggestion
  Unique[x∷xs]⇒x∉xs : Unique S (x ∷ xs) → x ∉ xs
```
Etc.

CHANGELOG.md
```diff
@@ -118,23 +123,74 @@ Additions to existing modules
...
+  ```
+
+* In `Data.List.Relation.Unary.Unique.Propositional.Properties`:
+  ```agda
+  Unique-dropSnd    : Unique (x ∷ y ∷ xs) → Unique (x ∷ xs)
+  Unique∷⇒head∉tail : Unique (x ∷ xs) → x ∉ xs
```
jamesmckinna, line 191: Ditto. re: changing name?

jamesmckinna (changes requested):
Many thanks for a huge contribution!

Lots of nitpicky feedback, but I think quite a few changes (removing redundancy etc.) would make this a lot tighter.

src/Data/List/Relation/Binary/Disjoint/Setoid/Properties.agda
jamesmckinna: Sorry! Out of scope for this PR!

src/Data/List/Membership/Setoid/Properties.agda
omelkonian: Unfortunately, introducing `` in the module telescope (instead of a statement in the body of the module) leads to this cryptic error?!
```agda
module _
  (S₁ : Setoid c₁ ℓ₁) (S₂ : Setoid c₂ ℓ₂)
  (open Setoid S₁ renaming (Carrier to A; _≈_ to _≈₁_))
  {P : Pred _ p} (P? : Decidable P) (resp : P Respects _)
  {f xs y} where

  open Setoid     S₂ renaming (_≈_ to _≈₂_; sym to sym₂)
...
  ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
```
```
Carrier S₁ !=< Data.List.Membership.Setoid.Properties.GeneralizeTel
when checking that the expression _≈₁_ has type
Rel (Carrier S₁) _ℓ₁_1616
```

The best I could come up with is using an extra inner module:
```agda
module _ (S₁ : Setoid c₁ ℓ₁) (S₂ : Setoid c₂ ℓ₂) where
  open Setoid S₁ renaming (Carrier to A; _≈_ to _≈₁_)
  open Setoid S₂ renaming (Carrier to B; _≈_ to _≈₂_; sym to sym₂)
  module _ {P : Pred A p} (P? : Decidable P) (resp : P Respects _≈₁_) {f xs y} where
    open Membership S₁ renaming (_∈_ to _∈₁_)
    open Membership S₂ renaming (_∈_ to _∈₂_)

    ∈-map∘filter⁻ : y ∈₂ map f (filter P? xs) →
                    ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
    ∈-map∘filter⁻ h =
      let x , x∈ , y≈ = ∈-map⁻ S₁ S₂ h
          y∈ , Py     = ∈-filter⁻ S₁ P? resp x∈
      in x , y∈ , y≈ , Py

    ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
                    ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
                    y ∈₂ map f (filter P? xs)
    ∈-map∘filter⁺ pres (x , x∈ , y≈ , Px)
      = ∈-resp-≈ S₂ (sym₂ y≈) 
      $ ∈-map⁺ S₁ S₂ pres (∈-filter⁺ S₁ P? resp x∈ Px)
```

Or, alternatively, letting Agda infer ` _≈₁_`:
```agda
module _
  (S₁ : Setoid c₁ ℓ₁) (S₂ : Setoid c₂ ℓ₂)
  (open Setoid S₁ renaming (Carrier to A))
  {P : Pred A p} (P? : Decidable P) (resp : P Respects _)
  {f xs y} where

  open Setoid     S₁ renaming (_≈_ to _≈₁_)
  open Setoid     S₂ renaming (_≈_ to _≈₂_; sym to sym₂)
  open Membership S₁ renaming (_∈_ to _∈₁_)
  open Membership S₂ renaming (_∈_ to _∈₂_)

  ∈-map∘filter⁻ : y ∈₂ map f (filter P? xs) →
                  ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
  ∈-map∘filter⁻ h =
    let x , x∈ , y≈ = ∈-map⁻ S₁ S₂ h
        y∈ , Py     = ∈-filter⁻ S₁ P? resp x∈
    in x , y∈ , y≈ , Py

  ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
                  ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
                  y ∈₂ map f (filter P? xs)
  ∈-map∘filter⁺ pres (x , x∈ , y≈ , Px)
    = ∈-resp-≈ S₂ (sym₂ y≈)
    $ ∈-map⁺ S₁ S₂ pres (∈-filter⁺ S₁ P? resp x∈ Px)
```
...and to avoid the telescope import altogether:
```agda
module _
  (S₁ : Setoid c₁ ℓ₁) (S₂ : Setoid c₂ ℓ₂)
  {P : Pred _ p} (P? : Decidable P) (resp : P Respects _)
  {f xs y} where
```

Shall I go with the last one?

src/Data/List/Membership/Propositional/Properties.agda
omelkonian: Indeed it seems extremely tedious to worry about backwards compatibility w.r.t. (hidden) variable argument order.

I would urge you to only worry about backw.compat. only modulo arg-order, especially given that Agda does not provide any flexibility for that (e.g. topological/dependency order instead of the current strict linear order).

CHANGELOG.md
omelkonian: I've renamed to your suggestion for now, although I have to point out that I find this naming approach redundant: why would you want to repeat the type as-is? This is mere duplication, while names should provide "shortcuts" that are more intuitive/abstract/natural than the type that accompanies them (e.g. ⁺/⁻ for intro/elim).

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -401,3 +433,39 @@ there-injective-≢∈ : ∀ {xs} {x y z : A} {x∈xs : x ∈ xs} {y∈xs : y 
...
+[]∉map∷ {xss = _ ∷ _} (there p) = []∉map∷ p
+
+map∷-decomp∈ : (List A ∋ x ∷ xs) ∈ map (y ∷_) xss → x ≡ y × xs ∈ xss
+map∷-decomp∈ {xss = _ ∷ _} = λ where
+  (here refl) → refl , here refl
+  (there p)   → map₂ there $ map∷-decomp∈ p
+
+map∷-decomp : xs ∈ map (y ∷_) xss → ∃[ ys ] ys ∈ xss × y ∷ ys ≡ xs
+map∷-decomp               {xss = _  ∷ _} (here refl) = -, here refl , refl
+map∷-decomp {xs = []}     {xss = _  ∷ _} (there xs∈) = ⊥-elim ([]∉map∷ xs∈)
```
omelkonian, line 463: Also changed this to `contradiction`.

CHANGELOG.md
jamesmckinna: I take your point, just as I'm grateful you took my suggestion. I confess I don't think my position is necessarily 'correct' or (cognitively) 'optimal', but sometimes I think there is a place for redundancy... 

... hence the call-out to a second reviewer.

src/Data/List/Membership/Setoid/Properties.agda
jamesmckinna: Ugh! That's really unfortunate... (caused by a clash between `variable` generalisation at the top of the module, and locally via the `open`ing of `Setoid`?)

... I'd be happy with the last approach, but again, I'd go with the view of the second reviewer, so leaving this as 'unresolved' for the time being...

jamesmckinna (approved):

CHANGELOG.md
```diff
@@ -147,6 +196,53 @@ Additions to existing modules
   ++⁺ˡ : Reflexive R → ∀ zs → (_++ zs) Preserves (Pointwise R) ⟶ (Pointwise R)
   ```
 
+* In `Data.List.Relation.Binary.Subset.Setoid.Properties`:
+  ```agda
+  ⊈[]    : x ∷ xs ⊈ []
```
jamesmckinna, line 201: Should this in fact be `∷⊈[]` (since `[] ⊆ []` is provable!)

CHANGELOG.md
```diff
@@ -147,6 +196,53 @@ Additions to existing modules
...
+  ⊆∷⇒∈∨⊆ : xs ⊆ y ∷ ys → y ∈ xs ⊎ xs ⊆ ys
+  ⊆∷∧∉⇒⊆ : xs ⊆ y ∷ ys → y ∉ xs → xs ⊆ ys
+  ```
+
+* In `Data.List.Relation.Binary.Subset.Propositional.Properties`:
+  ```agda
+  ⊈[]    : x ∷ xs ⊈ []
```
jamesmckinna, line 208: Ditto. to L201 above...

src/Data/List/Relation/Binary/Subset/Setoid/Properties.agda
```diff
@@ -51,8 +52,12 @@ module _ (S : Setoid a ℓ) where
...
   open Subset S
 
-  ⊈[] : ∀ {x xs} → x ∷ xs ⊈ []
-  ⊈[] p = Membershipₚ.∉[] S $ p (here refl)
+  ∷⊈[] : ∀ {x xs} → x ∷ xs ⊈ []
+  ∷⊈[] p = Membershipₚ.∉[] S $ p (here refl)
+
+  ⊆[]⇒≡[] : (_⊆ []) ⋐ (_≡ [])
+  ⊆[]⇒≡[] {x = []}    _ = ≡.refl
+  ⊆[]⇒≡[] {x = _ ∷ _} p with () ← ⊈[] p
```
jamesmckinna, line 60:
```suggestion
⊆[]⇒≡[] {x = _ ∷ _} p with () ← ∷⊈[] p
```

src/Data/List/Base.agda
omelkonian: Actually, this is often useful when writing meta-programs, and there is no real need for lemmas since we do not typically prove our meta-programs correct :)

Taneb (approved):

CHANGELOG.md
```diff
@@ -120,21 +125,65 @@ Additions to existing modules
...
 * In `Data.List.Membership.Setoid.Properties`:
   ```agda
-  Any-∈-swap :  Any (_∈ ys) xs → Any (_∈ xs) ys
-  All-∉-swap :  All (_∉ ys) xs → All (_∉ xs) ys
+  Any-∈-swap    :  Any (_∈ ys) xs → Any (_∈ xs) ys
+  All-∉-swap    :  All (_∉ ys) xs → All (_∉ xs) ys
+  ∈-map∘filter⁻ : y ∈₂ map f (filter P? xs) → ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x
+  ∈-map∘filter⁺ : f Preserves _≈₁_ ⟶ _≈₂_ →
+                  ∃[ x ] x ∈₁ xs × y ≈₂ f x × P x →
+                  y ∈₂ map f (filter P? xs)
```
gallais, line 133: These feel super ad-hoc. Why `map` and `filter` in particular?
Why not e.g. `mapMaybe`? Or `unalignWith`?

CHANGELOG.md
```diff
@@ -147,6 +196,53 @@ Additions to existing modules
...
+
+* In `Data.List.Relation.Binary.Sublist.Setoid.Properties`:
+  ```agda
+  []⊆-universal : Universal ([] ⊆_)
+  ```
+
+* In `Data.List.Relation.Binary.Disjoint.Setoid.Properties`:
+  ```agda
+  deduplicate⁺ : Disjoint S xs ys → Disjoint S (deduplicate _≟_ xs) (deduplicate _≟_ ys)
+  ∈-dedup      : x ∈ xs ⇔ x ∈ deduplicate _≟_ xs
```
gallais, line 231: Shouldn't this be `dedup⁺` in the membership module instead?

src/Data/List/Membership/Propositional/Properties.agda
```diff
@@ -123,10 +126,13 @@ module _ {v : A} where
   ∈-++⁻ : ∀ xs {ys} → v ∈ xs ++ ys → (v ∈ xs) ⊎ (v ∈ ys)
   ∈-++⁻ = Membership.∈-++⁻ (≡.setoid A)
 
+  ∈-++ : v ∈ xs ++ ys ⇔ (v ∈ xs ⊎ v ∈ ys)
```
gallais, line 129: This should probably mention `⇔` in the name (cf. `concat-∈↔`  below)

omelkonian: Wouldn't it be better to have a consistent naming convention for bundling introductions & eliminations?
Since intros are named `<property_name>⁺` (modulo left/right qualifiers) and eliminators `<property_name>⁻`, my suggestion was to just drop the polarities (i.e. `<property_name>`) to communicate that this includes both intro & elim. 

Another naming convention that I find intuitive is to explicitly include both polarities: `<property_name>⁺⁻`, e.g. `∈-++⁺⁻`.

The things I don't like about the current naming of such bundles are the following:
- The exact arrow being used will change in each case (whether it's an equivalence  `⇔`, an inverse `↔`, a bi-inverse, etc...); isn't it preferable to have a universal convention that hides such differences (much like we do for intros/elims which might take different forms in each individual case but are always named with `⁺` and `⁻`.
- names like `concat-∈↔` are not immediately linked to the intros/elims that it contains  (i.e. the `<property_name>` part is not retained)

So, if I follow the current practices, I would have to name this `++-∈⇔` :(
I can definitely do that, but I find it ugly and unintuitive.

CHANGELOG.md
omelkonian: Yep, I agree, and I was actually reluctant to include this in this PR for the same reason.

The only argument I find here is that these are operations that are so incredibly common, that if one user goes through the trouble of defining the composition to re-use it across their codebase/application, then probably it is worth including in the stdlib for other users. (It is, however, difficult to provide evidence for what is common and what is not.)

Of course, it is infeasible to be *complete* in any way, i.e. cover all possible combinations of list operations, but maybe a "the-more-the-merrier" strategy is better in this case, and we should encourage users to submit such combinations as they arise?

src/Data/List/Relation/Binary/Permutation/Propositional/Properties/WithK.agda
```diff
@@ -36,27 +34,21 @@ open import Relation.Binary.Definitions using (DecidableEquality)
...
-    ∼⟨ ∈-dedup≡ ⊎-cong ∈-dedup≡ ⟩
-      (x ∈ dedup≡ xs ⊎ x ∈ dedup≡ ys)
-    ∼⟨ SK-sym ∈-++ ⟩
-      x ∈ dedup≡ xs ++ dedup≡ ys
-    ∎ where open EquationalReasoning
+    λ {x} → begin
+    x ∈ dedup≡ (xs ++ ys)           ∼⟨ SK-sym ∈-dedup≡ ⟩
+    x ∈ xs ++ ys                    ∼⟨ ++-∈⇔ ⟩
+    (x ∈ xs ⊎ x ∈ ys)               ∼⟨ ∈-dedup≡ ⊎-cong ∈-dedup≡ ⟩
+    (x ∈ dedup≡ xs ⊎ x ∈ dedup≡ ys) ∼⟨ SK-sym ++-∈⇔ ⟩
```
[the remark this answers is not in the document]
omelkonian: Seems there's no such syntax?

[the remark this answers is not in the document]
jamesmckinna: Module `Function.Related.Propositional.Equational Reasoning` opens (L316) module `Relation.Binary.Reasoning.Syntax.↔-syntax` (L349) which defines this syntax (L354)...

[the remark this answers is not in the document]
jamesmckinna: ... Issue #1138 fixed by #2160 !?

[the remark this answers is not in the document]
jamesmckinna: And... you could avoid the explicit named `using (SK-sym; ` import at the top...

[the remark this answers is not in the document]
jamesmckinna: Apologies, I hadn't looked closely enough at this code, nor tested it: we have 'symmetric' equality combinators for `↔-Reasoning`(`_[ bag ]_`), but *not* for `_[ set ]_`... which we should. I'll open an issue/PR.

JacquesCarette (approved):

jamesmckinna:
Thanks @JacquesCarette ! merging now...

jamesmckinna:
3rd time lucky?

PR #2481, opened 2024-09-11
Add properties/structures/bundles related to decidability for `Relation.Binary.Construct.Interior.Symmetric`

jamesmckinna:
First steps towards #2353 , but of independent interest/utility, I hope.

NB. Issue:
* we don't seem to have `IsDecPreorder`/`DecPreorder` defined in `Relation.Binary.Structures`/`Relation.Binary.Bundles`? now raised as #2482 so move discussion there.

### DIFF
--- a/src/Relation/Binary/Construct/Interior/Symmetric.agda
+++ b/src/Relation/Binary/Construct/Interior/Symmetric.agda
@@ -8,15 +8,18 @@
 
 module Relation.Binary.Construct.Interior.Symmetric where
 
-open import Function.Base using (flip)
-open import Level
+open import Data.Bool.Base using (_∧_)
+open import Function.Base using (flip; _∘_)
+open import Level using (Level)
 open import Relation.Binary
+open import Relation.Nullary.Decidable.Core
+open import Relation.Nullary.Reflects
 
 private
   variable
-    a b c ℓ r s t : Level
+    a ℓ : Level
     A : Set a
-    R S T : Rel A r
+    R S T : Rel A ℓ
 
 ------------------------------------------------------------------------
 -- Definition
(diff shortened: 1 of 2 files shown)


src/Relation/Binary/Construct/Interior/Symmetric.agda
```diff
@@ -56,28 +59,59 @@ transitive tr = trans tr tr
 asymmetric⇒empty : Asymmetric R → Empty (SymInterior R)
 asymmetric⇒empty asym (r , r′) = asym r r′
 
+-- Decidability
+decidable : Decidable R → Decidable (SymInterior R)
+does  (decidable R? x y) = does (R? x y) ∧ does (R? y x)
+proof (decidable R? x y) = proof (R? x y) reflects proof (R? y x)
+  where
+  _reflects_ : ∀ {bxy byx} → Reflects (R x y) bxy → Reflects (R y x) byx →
```
Taneb, line 67: Is it worth pulling this to the top level?

Taneb (approved):

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: That's a nice question!
My first thought was that this was a fairly slavish re-emulation of the corresponding (top-level) principle `_×-reflects_` defined in `Relation.Nullary.Reflects`, so didn't merit elevation to the top-level; not least because it really only has this one use, namely to establish decidability... (in that sense, almost *none* of `Reflects` should really be exposed outside `Relation.Nullary.Decidable.*`?).
Not sure how best to reply except that: if we think we might use it later/downstream, we could always refactor to lift it back out...? But going the other way is ... more work ;-)
Not the most 'dignified' answer!

JacquesCarette: I'm with @Taneb here, this does seem to be of independent interest... sorry!

JacquesCarette (approved):
I likewise won't hold this one up for the one refactor... but it should get done.

src/Relation/Binary/Construct/Interior/Symmetric.agda
jamesmckinna: @JacquesCarette comments:
> I'm with @Taneb here, this does seem to be of independent interest... sorry!

I think that if you the two of you can come up with a good name for this lemma, I'll do the refactoring...
... else let's merge this now against such a name choice emerging later...?

MatthewDaggitt (approved):

PR #2349, opened 2024-04-09
[ new ] ⁻¹-anti-homo-(// | \\)

gallais:

Issue #1118, opened 2020-03-16
strictly positive number types


riz0id:
proposing non-zero naturals and positive rationals and completing proofs for their properties with the algebraic morphism modules if necessary.

MatthewDaggitt:
Hmm do you have a definition in mind? In practice due to the issue of converting backwards and forwards between them, it's often easier simply to enforce a non-zero requirement where needed. See `Data.Nat.DivMod` for an example of how to use implicit arguments to avoid having to pass the proof every time and see #921 for a proposed improvement.

jamesmckinna:
In the light of #2644 do we need to reconsider this? or close?

riz0id:
Seems like people are going for Cauchy sequences.. and that is kind of where I found myself trying to do all of this some time back. I don't really have time to spare to contribute at the moment. Probably fine to close this.

jamesmckinna:
> Seems like people are going for Cauchy sequences.. and that is kind of where I found myself trying to do all of this some time back. I don't really have time to spare to contribute at the moment. Probably fine to close this.

Well, #2487 is moving towards an 'O Connor style version based on regular functions on ℚ⁺ rather than Cauchy sequences, with ℚ⁺ handled conventionally via the `Positive` predicate, in the spirit of @MatthewDaggitt 's suggestion above. It'll be interesting to see, once that PR lands, how badly it may be affected by #2644 ...

But yes, happy to close this (for the time being, if necessary; can always re-open if we eg move to ℚ⁺ in terms of `Data.Refinement`... or some such).

PR #2487, opened 2024-09-27
Real numbers, based on Cauchy sequences

Taneb:
***Very much work-in-progress***

From previous attempts, multiplication of real numbers will be the stalling point. But the HoTT book suggests this can be implemented in terms of addition, subtraction, and multiplication by a rational, which are all achievable, and squaring, which I don't know about.

### DIFF
--- a/src/Data/Rational/Properties.agda
+++ b/src/Data/Rational/Properties.agda
@@ -49,6 +49,7 @@ open import Data.Sum.Base as Sum using (inj₁; inj₂; [_,_]′; _⊎_)
 import Data.Sign.Base as Sign
 open import Function.Base using (_∘_; _∘′_; _∘₂_; _$_; flip)
 open import Function.Definitions using (Injective)
+open import Function.Metric.Rational as Metric hiding (Symmetric)
 open import Level using (0ℓ)
 open import Relation.Binary
 open import Relation.Binary.Morphism.Structures
@@ -1719,6 +1720,91 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
 ∣∣p∣∣≡∣p∣ : ∀ p → ∣ ∣ p ∣ ∣ ≡ ∣ p ∣
 ∣∣p∣∣≡∣p∣ p = 0≤p⇒∣p∣≡p (0≤∣p∣ p)
 
+------------------------------------------------------------------------
+-- Metric space
+------------------------------------------------------------------------
+
+private
+  d : ℚ → ℚ → ℚ
+  d p q = ∣ p - q ∣
+
+d-cong : Congruent _≡_ d
+d-cong = cong₂ _
+
+d-nonNegative : ∀ {p q} → 0ℚ ≤ d p q
+d-nonNegative {p} {q} = nonNegative⁻¹ _ {{∣-∣-nonNeg (p - q)}}
+
+d-definite : Definite _≡_ d
+d-definite {p} refl = cong ∣_∣ (+-inverseʳ p)
+
+d-indiscernable : Indiscernable _≡_ d
+d-indiscernable {p} {q} ∣p-q∣≡0 = begin
+  p               ≡˘⟨ +-identityʳ p ⟩
+  p - 0ℚ          ≡˘⟨ cong (_-_ p) (∣p∣≡0⇒p≡0 (p - q) ∣p-q∣≡0) ⟩
+  p - (p - q)     ≡⟨ cong (_+_ p) (neg-distrib-+ p (- q)) ⟩
+  p + (- p - - q) ≡˘⟨ +-assoc p (- p) (- - q) ⟩
+  (p - p) - - q   ≡⟨ cong₂ _+_ (+-inverseʳ p) (⁻¹-involutive q) ⟩
+  0ℚ + q          ≡⟨ +-identityˡ q ⟩
+  q               ∎
+  where
+    open ≡-Reasoning
+    open GroupProperties +-0-group
+
+d-sym : Metric.Symmetric d
+d-sym p q = begin
+  ∣ p - q ∣       ≡˘⟨ cong (λ # → ∣ # - q ∣) (⁻¹-involutive p) ⟩
+  ∣ - - p - q ∣   ≡˘⟨ cong ∣_∣ (neg-distrib-+ (- p) q) ⟩
+  ∣ - (- p + q) ∣ ≡⟨ cong (λ # → ∣ - # ∣) (+-comm (- p) q) ⟩
+  ∣ - (q - p) ∣   ≡⟨ ∣-p∣≡∣p∣ (q - p) ⟩
+  ∣ q - p ∣       ∎
+  where
+    open ≡-Reasoning
+    open GroupProperties +-0-group
+
+d-triangle : TriangleInequality d
+d-triangle p q r = begin
+  ∣ p - r ∣             ≡˘⟨ cong (λ # → ∣ # - r ∣) (+-identityʳ p) ⟩
+  ∣ p + 0ℚ - r ∣        ≡˘⟨ cong (λ # → ∣ p + # - r ∣) (+-inverseˡ q) ⟩
+  ∣ p + (- q + q) - r ∣ ≡˘⟨ cong (λ # → ∣ # - r ∣) (+-assoc p (- q) q) ⟩
+  ∣ ((p - q) + q) - r ∣ ≡⟨ cong ∣_∣ (+-assoc (p - q) q (- r)) ⟩
+  ∣ (p - q) + (q - r) ∣ ≤⟨ ∣p+q∣≤∣p∣+∣q∣ (p - q) (q - r) ⟩
+  ∣ p - q ∣ + ∣ q - r ∣ ∎
+  where open ≤-Reasoning
+
+d-isProtoMetric : IsProtoMetric _≡_ d
+d-isProtoMetric = record
+  { isPartialOrder = ≤-isPartialOrder
+  ; ≈-isEquivalence = isEquivalence
+  ; cong = cong₂ _
+  ; nonNegative = λ {p q} → d-nonNegative {p} {q}
+  }
+
+d-isPreMetric : IsPreMetric _≡_ d
+d-isPreMetric = record
+  { isProtoMetric = d-isProtoMetric
+  ; ≈⇒0 = d-definite
+  }
+
+d-isQuasiSemiMetric : IsQuasiSemiMetric _≡_ d
+d-isQuasiSemiMetric = record
+  { isPreMetric = d-isPreMetric
+  ; 0⇒≈ = d-indiscernable
+  }
+
+d-isSemiMetric : IsSemiMetric _≡_ d
+d-isSemiMetric = record
+  { isQuasiSemiMetric = d-isQuasiSemiMetric
+  ; sym = d-sym
+  }
+
+d-isMetric : IsMetric _≡_ d
+d-isMetric = record
+  { isSemiMetric = d-isSemiMetric
+  ; triangle = d-triangle
+  }
+
+d-metric : Metric _ _
+d-metric = record { isMetric = d-isMetric }
 
 ------------------------------------------------------------------------
 -- DEPRECATED NAMES
--- a/src/Data/Real/Base.agda
+++ b/src/Data/Real/Base.agda
@@ -1,0 +1,73 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Real numbers
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --without-K --guardedness #-}
+
+module Data.Real.Base where
+
+open import Codata.Guarded.Stream
+open import Codata.Guarded.Stream.Properties
+open import Data.Integer.Base using (+<+)
+open import Data.Nat.Base as ℕ using (z≤n; s≤s)
+import Data.Nat.Properties as ℕ
+open import Data.Product.Base hiding (map)
+open import Data.Rational.Base as ℚ hiding (_+_)
+open import Data.Rational.Properties
+open import Data.Rational.Solver
+open import Data.Rational.Unnormalised using (*<*)
+open import Function.Base using (_$_)
+open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong; cong₂)
+
+open import Function.Metric.Rational.CauchySequence d-metric public renaming (CauchySequence to ℝ)
+
+fromℚ : ℚ → ℝ
+fromℚ = embed
+
+_+_ : ℝ → ℝ → ℝ
+sequence (x + y) = zipWith ℚ._+_ (sequence x) (sequence y)
+isCauchy (x + y) ε = proj₁ [x] ℕ.+ proj₁ [y] , λ {m} {n} m≥N n≥N → begin-strict
+  ∣ lookup (zipWith ℚ._+_ (sequence x) (sequence y)) m - lookup (zipWith ℚ._+_ (sequence x) (sequence y)) n ∣
+    ≡⟨ cong₂ (λ a b → ∣ a - b ∣)
+      (lookup-zipWith m ℚ._+_ (sequence x) (sequence y))
+      (lookup-zipWith n ℚ._+_ (sequence x) (sequence y))
+    ⟩
+  ∣ (lookup (sequence x) m ℚ.+ lookup (sequence y) m) - (lookup (sequence x) n ℚ.+ lookup (sequence y) n) ∣
+    ≡⟨ cong ∣_∣ (lemma (lookup (sequence x) m) (lookup (sequence y) m) (lookup (sequence x) n) (lookup (sequence y) n)) ⟩
+  ∣ (lookup (sequence x) m - lookup (sequence x) n) ℚ.+ (lookup (sequence y) m - lookup (sequence y) n) ∣
+    ≤⟨ ∣p+q∣≤∣p∣+∣q∣
+      (lookup (sequence x) m - lookup (sequence x) n)
+      (lookup (sequence y) m - lookup (sequence y) n)
+    ⟩
+  ∣ lookup (sequence x) m - lookup (sequence x) n ∣ ℚ.+ ∣ lookup (sequence y) m - lookup (sequence y) n ∣
+    <⟨ +-mono-<
+      (proj₂ [x]
+        (ℕ.≤-trans (ℕ.m≤m+n (proj₁ [x]) (proj₁ [y])) m≥N)
+        (ℕ.≤-trans (ℕ.m≤m+n (proj₁ [x]) (proj₁ [y])) n≥N)
+      )
+      (proj₂ [y]
+        (ℕ.≤-trans (ℕ.m≤n+m (proj₁ [y]) (proj₁ [x])) m≥N)
+        (ℕ.≤-trans (ℕ.m≤n+m (proj₁ [y]) (proj₁ [x])) n≥N)
+      )
+    ⟩
+  ½ * ε ℚ.+ ½ * ε
+    ≡˘⟨ *-distribʳ-+ ε ½ ½ ⟩
+  1ℚ * ε
+    ≡⟨ *-identityˡ ε ⟩
+  ε ∎
+  where
+    open ≤-Reasoning
+    instance _ : Positive (½ * ε)
+    _ = positive {½ * ε} $ begin-strict
+      0ℚ     ≡˘⟨ *-zeroˡ ε ⟩
+      0ℚ * ε <⟨ *-monoˡ-<-pos ε {0ℚ} {½} (*<* (+<+ (s≤s z≤n))) ⟩
+      ½ * ε  ∎
+    [x] = isCauchy x (½ * ε)
+    [y] = isCauchy y (½ * ε)
+
+    lemma : ∀ a b c d → (a ℚ.+ b) - (c ℚ.+ d) ≡ (a - c) ℚ.+ (b - d)
+    lemma = solve 4 (λ a b c d → ((a :+ b) :- (c :+ d)) , ((a :- c) :+ (b :- d))) refl
+      where open +-*-Solver
+
--- a/src/Function/Metric/Rational/CauchySequence.agda
+++ b/src/Function/Metric/Rational/CauchySequence.agda
@@ -1,0 +1,83 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Cauchy sequences on a metric over the rationals
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible --guardedness #-}
+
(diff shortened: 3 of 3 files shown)


src/Data/Rational/Properties.agda
```diff
@@ -1719,6 +1720,91 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
...
+d-nonNegative {p} {q} = nonNegative⁻¹ _ {{∣-∣-nonNeg (p - q)}}
+
+d-definite : Definite _≡_ d
+d-definite {p} refl = cong ∣_∣ (+-inverseʳ p)
+
+d-indiscernable : Indiscernable _≡_ d
+d-indiscernable {p} {q} ∣p-q∣≡0 = begin
+  p               ≡˘⟨ +-identityʳ p ⟩
+  p - 0ℚ          ≡˘⟨ cong (_-_ p) (∣p∣≡0⇒p≡0 (p - q) ∣p-q∣≡0) ⟩
+  p - (p - q)     ≡⟨ cong (_+_ p) (neg-distrib-+ p (- q)) ⟩
```
MatthewDaggitt, line 1744: It feels like `p - (p - q) ≡ q` should be a lemma in `GroupProperties`?

src/Data/Rational/Properties.agda
```diff
@@ -1719,6 +1720,91 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
...
+  where
+    open ≡-Reasoning
+    open GroupProperties +-0-group
+
+d-sym : Metric.Symmetric d
+d-sym p q = begin
+  ∣ p - q ∣       ≡˘⟨ cong (λ # → ∣ # - q ∣) (⁻¹-involutive p) ⟩
+  ∣ - - p - q ∣   ≡˘⟨ cong ∣_∣ (neg-distrib-+ (- p) q) ⟩
+  ∣ - (- p + q) ∣ ≡⟨ cong (λ # → ∣ - # ∣) (+-comm (- p) q) ⟩
+  ∣ - (q - p) ∣   ≡⟨ ∣-p∣≡∣p∣ (q - p) ⟩
```
MatthewDaggitt, line 1758: Same here. It feels like `- (p - q) ≡ q - p` should be a lemma in `GroupProperties`?

src/Data/Rational/Properties.agda
```diff
@@ -1719,6 +1720,91 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
...
+
+d-nonNegative : ∀ {p q} → 0ℚ ≤ d p q
+d-nonNegative {p} {q} = nonNegative⁻¹ _ {{∣-∣-nonNeg (p - q)}}
+
+d-definite : Definite _≡_ d
+d-definite {p} refl = cong ∣_∣ (+-inverseʳ p)
+
+d-indiscernable : Indiscernable _≡_ d
+d-indiscernable {p} {q} ∣p-q∣≡0 = begin
+  p               ≡˘⟨ +-identityʳ p ⟩
```
MatthewDaggitt, line 1742: Obviously it would be great if we could use the new symmetric reasoning syntax.

src/Data/Rational/Properties.agda
```diff
@@ -1719,6 +1720,91 @@ toℚᵘ-homo-∣-∣ (mkℚ -[1+ _ ] _ _) = *≡* refl
 ∣∣p∣∣≡∣p∣ : ∀ p → ∣ ∣ p ∣ ∣ ≡ ∣ p ∣
 ∣∣p∣∣≡∣p∣ p = 0≤p⇒∣p∣≡p (0≤∣p∣ p)
 
+------------------------------------------------------------------------
+-- Metric space
+------------------------------------------------------------------------
+
+private
+  d : ℚ → ℚ → ℚ
+  d p q = ∣ p - q ∣
```
MatthewDaggitt, line 1729: This is really very frustrating. Anyone have any nice ideas how we can make this a function in `Data.Rational.Base`? I guess we could use some weird unicode bars?

src/Function/Metric/Rational/CauchySequence.agda
```diff
@@ -0,0 +1,83 @@
...
+open import Data.Integer.Base using (+<+)
+open import Data.Product.Base
+open import Data.Rational.Base
+open import Data.Rational.Properties
+open import Function.Base
+open import Level
+open import Relation.Binary
+open import Relation.Binary.PropositionalEquality using (cong₂)
+
+record CauchySequence : Set a where
```
MatthewDaggitt, line 28: So I think this and the equality definition needs to be defined over a `DistanceFunction` rather than a full `Metric` with laws. Otherwise we're forced to define completeness over the whole metric space bundle, and we try and avoid defining predicates over bundles like that in the library.

src/Function/Metric/Rational/CauchySequence.agda
```diff
@@ -0,0 +1,83 @@
...
+  field
+    sequence : Stream Carrier
+    isCauchy : ∀ ε → {{Positive ε}} → Σ[ N ∈ ℕ ] ∀ {m n} → m ℕ.≥ N → n ℕ.≥ N → d (lookup sequence m) (lookup sequence n) < ε
```
MatthewDaggitt, line 31: Should the positive proof be irrelevant? If not, it would be good to document why not?

src/Function/Metric/Rational/CauchySequence.agda
```diff
@@ -0,0 +1,83 @@
...
+  ½ * ε + ½ * ε
+    ≡˘⟨ *-distribʳ-+ ε ½ ½ ⟩
+  1ℚ * ε
+    ≡⟨ *-identityˡ ε ⟩
+  ε
+    ∎
+  where
+    open ≤-Reasoning
+    instance _ : Positive (½ * ε)
+    _ = positive {½ * ε} $ begin-strict
```
MatthewDaggitt, line 70: Can we use an existing proof that two positive numbers multiplied together is another positive number?

src/Function/Metric/Rational/CauchySequence.agda
MatthewDaggitt: The added advantage of this, is that we could definite the notion of Cauchy sequence generically, rather than just for `Rational`.

MatthewDaggitt: This record would also benefit from a constructor I think? Which would allow us to avoid writing `sequence x` and `sequence y` everywhere and instead write `x` and `y`?

MatthewDaggitt (commented):
This looks great!

I've left a review on the two files outside of `Data.Real`. I think it's probably easier to try and merge them in first and then tackle the `Real` files?

src/Data/Rational/Properties.agda
Taneb: It actually requires commutativity! (this is hidden in `neg-distrib-+`). As far as I can tell it doesn't exist yet.

src/Data/Rational/Properties.agda
Taneb: Yes! It's `⁻¹-anti-homo-//`

src/Data/Rational/Properties.agda
Taneb: I continue to strongly dislike the new symmetric reasoning syntax. emacs treats `⟨` and `⟩` as brackets, which means if I miss a `)` (which I often do if I write a whole sequence of reasoning in one go) I can't use emacs' bracket-matching to find where it should go

src/Function/Metric/Rational/CauchySequence.agda
Taneb: Yes, it should.

src/Function/Metric/Rational/CauchySequence.agda
Taneb: I don't think we currently have this proof.

Taneb: I've added it, and some related lemmata

src/Function/Metric/Rational/CauchySequence.agda
Taneb: Hmm. I want to make `isCauchy` irrelevant. Would both doing that and introducing a constructor run into Irrelevant Projections stuff?

MatthewDaggitt: Well we use a record constructor with irrelevant fields in Data.Rational.Base and I'm not aware we've run into any problems there...

src/Data/Rational/Properties.agda
MatthewDaggitt: The following proof looks commutativity-free to me?
```
p - (p - q) = p + - (p + - q) = p + (-p + q) = (p + -p) + q = 0# + q = q
```

src/Data/Rational/Properties.agda
MatthewDaggitt: We have officially deprecated the old syntax: https://github.com/agda/agda-stdlib/blob/45a46f7691c328364856eeda8b042bf6f389d8a8/src/Relation/Binary/Reasoning/Syntax.agda#L270

It's just that Agda doesn't transfer deprecation warnings properly for parametrised modules.

src/Data/Rational/Properties.agda
Taneb: The step `- (p + - q) = - p + q` requires commutativity

MatthewDaggitt: Oh whoops. Yes, you don't get distributivity for free.

src/Data/Real/Base.agda
```diff
@@ -0,0 +1,224 @@
...
+  ½ * ε ℚ.+ ½ * ε
+    ≡⟨ *-distribʳ-+ ε ½ ½ ⟨
+  1ℚ * ε
+    ≡⟨ *-identityˡ ε ⟩
+  ε ∎
+  where
+    open ≤-Reasoning
+    instance _ : Positive (½ * ε)
+    _ = pos*pos⇒pos ½ ε
+    [x] = isCauchy x (½ * ε)
```
JacquesCarette, line 68: you name `isCauchy` but only use its first component -- maybe name that instead, to save all the projections above?

src/Data/Real/Base.agda
```diff
@@ -0,0 +1,224 @@
...
+    open ≤-Reasoning
+
+    B : ℕ.ℕ
+    B = proj₁ (isCauchy x 1ℚ)
+
+    b : ℚ
+    b = 1ℚ ℚ.+ ∣ lookup (sequence x) B ∣
+
+    instance _ : Positive b
+    _ = positive $ begin-strict
```
JacquesCarette, line 185: shouldn't you prove pos+pos=>pos and pos1, then this is an easy consequence?

src/Data/Real/Base.agda
```diff
@@ -0,0 +1,224 @@
...
+      ∣ lookup (sequence x) B ∣        ≡⟨ +-identityˡ ∣ lookup (sequence x) B ∣ ⟨
+      0ℚ ℚ.+ ∣ lookup (sequence x) B ∣ <⟨ +-monoˡ-< ∣ lookup (sequence x) B ∣ {0ℚ} {1ℚ} (*<* (+<+ (s≤s z≤n))) ⟩
+      1ℚ ℚ.+ ∣ lookup (sequence x) B ∣ ≡⟨⟩
+      b                                ∎
+
+    instance _ : NonZero b
+    _ = pos⇒nonZero b
+
+    instance _ : Positive (b ℚ.+ b)
+    _ = positive $ begin-strict
```
JacquesCarette, line 196: so this should also be pos+pos=>pos ?

src/Data/Real/Properties.agda
```diff
@@ -0,0 +1,80 @@
...
+  where
+    open ℚ.≤-Reasoning
+    instance _ : ℚ.Positive (ℚ.½ ℚ.* ε)
+    _ = ℚ.pos*pos⇒pos ℚ.½ ε
+
+    lemma : ∀ a b c d → (a ℚ.+ b) ℚ.- (c ℚ.+ d) ≡ (a ℚ.- c) ℚ.+ (b ℚ.- d)
+    lemma = solve 4 (λ a b c d → ((a :+ b) :- (c :+ d)) , ((a :- c) :+ (b :- d))) refl
+      where open +-*-Solver
+
++-assoc : Associative _+_
```
JacquesCarette, line 54: all the next properties seem to be lifting properties, i.e. if some binary operation _+_ has property P then its lift to streams will have the same property.

JacquesCarette (commented):
This is looking really great. A variety of comments as I was leafing through the code.

src/Data/Real/Base.agda
Taneb: I do use the second component, on lines 50 and 54. But I use the first component a lot more

src/Data/Real/Base.agda
Taneb: pos1 is already inferrable, bot pos+pos=>pos would be a good addition

jamesmckinna:
I've written this on *every* channel concerned with constructive formalisations of the reals, but for the sake of clarity, I'll repeat myself here: please please please do NOT base the representation on Cauchy sequences, because the rates of convergence (be it by `1/n` convergence, or by `1/2^n`...) are *so* bad, and so much pointless arithmetic is involved, but instead use the development in [Russell O' Connor's PhD thesis](http://r6.ca/Thesis/) in terms of the monad for metric space completion, based on *functions from `ε ∈ ℚ⁺` to `ℚ`* representing nested decreasing `ε`-balls around a point... (IIRC; but it's all *very* clear in Russell's thesis, Coq or no Coq...)

The opportunity when passing from 'classical'/'orthodox' constructions in mathematics to ones 'better adapted' to the activity for formalisation, is one of the outstanding research challenges for our field, and a much-neglected one. We can do better!

src/Data/Rational/Properties.agda
jamesmckinna: See my global comment: we should have `ℚ⁺` as a first class thing in the library before we proceed!

Taneb:
@jamesmckinna I'm reading Russel O'Connor's construction now. I haven't fully understood it yet, but I will note: I'm not using a fixed rate of convergence, so I don't think I have the rate of convergence issues you mention. For example, real number expressions involving only rational numbers so far converge _immediately_.

jamesmckinna:
To press on my point: I think it's much more *natural* (Occam's Razor) to consider (regular) *functions* rather than *streams* for encoding the basic data underlying the 'Cauchy' property... and, given that `stdlib` does not commit to using HoTT-like quotients based on HI(I)Ts, in favour of a (legacy/historical, but still, ontologically less committed/more neutral) choice based on setoids, I don't see any especial need to mutually define distance-related concepts with the metric completion functor. But other opinions are available!

src/Data/Rational/Properties.agda
jamesmckinna: See `Algebra.Properties.AbelianGroup.⁻¹-anti-homo‿-` added in #2349 ?

src/Data/Rational/Properties.agda
jamesmckinna: Re: notation.
It *is* a norm, so does using the following offer any assistance?
```suggestion
-- distance function
∥_─_∥ : ℚ → ℚ → ℚ
∥ p ─ q ∥ = ∣ p - q ∣
```
(reusing the 'long minus' from `Data.List.Relation.Unary.Any`, whose Unicode codepoint/Agda input-encoding I confess not ever to be able to remember, so this might do more harm than good...). The "weird Unicode bars" here are in fact simply given by `\ | |`... so they aren't too bad, but the long minus is annoying...

src/Function/Metric/Rational/CauchySequence.agda
jamesmckinna: As for irrelevant projections, the discussion on https://github.com/agda/agda/issues/7063 , upstream from #2199 (and others `label:irrelevance`/`label:irrelevant-projections`)  suggest... (probably) not?

Taneb:
This has stalled while I've been going on a sidequest to implement O'Connor-style reals. It's a _lot_ more up-front work, but it does seem like it's going to win some advantages:

* It avoids going via a stream of rationals, skipping the potentially long indexing step and avoiding needing `--guardedness`
* A lot of the work implementing operators is done "up front", making the actual operations a lot easier

Both in this branch and my branch implementing the O'Connor reals (not currently online), I'm using `Data.Rational.Solver` a lot. This would be a little nicer perhaps if #1879 had been done, and if we were more willing to use these solvers within stdlib itself. It would also be very nice to have an inequality solver for rationals, but I have no idea how one could go about implementing that. Alternatively, I could put a bit of effort into factoring out common lemmas (mostly relating to `½ * x + ½ * x ≡ x`) which might be enough to make this a lot less tedious.

Another thing that would be useful would be a type of strictly positive rationals, for example:

```agda
record ℚ⁺ : Set where
  constructor 𝕢⁺
  field
    unℚ⁺ : ℚ
    instance .{{unℚ⁺-pos}} : Positive unℚ⁺
```

(see also #1118). This looks _like_ `Data.Refinement`, but it's not because of the instance-iness. 

O'Connor defines metric spaces in terms of balls rather than distance functions, which allows him to define "real" metrics prior to having real numbers. Most of the challenge has been defining completion as a (strong!) monad on the category of these metric spaces and universally continuous functions.

jamesmckinna:
@Taneb writes:
> Another thing that would be useful would be a type of strictly positive rationals, for example:
> 
> ```agda
> record ℚ⁺ : Set where
>   constructor 𝕢⁺
>   field
>     unℚ⁺ : ℚ
>     instance .{{unℚ⁺-pos}} : Positive unℚ⁺
> ```
> 
> (see also #1118). This looks _like_ `Data.Refinement`, but it's not because of the instance-iness.

I'm confused... I though that the `{{...}}` bracketing would already mark the `unℚ⁺-pos` field as an instance, and available for proof search? What's the difference between your definition and eg.

```agda
record ℚ⁺ : Set where
  constructor 𝕢⁺
  field
    unℚ⁺ : ℚ
    .{{unℚ⁺-pos}} : Positive unℚ⁺
```
As for constructor/field names, my (ongoing; following PhIl Wadler) preference might be for 'minimum-ink' solutions, such as a postfix constructor `_⁺` and, whatever the *actual* field name chosen for the underlying value (though `value` springs to mind, echoing `Data.Refinement`... though I'd be inclined to keep it `private`?), a postfix synonym `_⁻` to strip off the `_⁺`... but perhaps that might cause too much eye/mind strain? And `pos` would seem canonical for the associated irrelevant instance name?

Taneb:
Thanks @jamesmckinna , I've taken those suggestions into account. I'd been over-cautious with the `instance` declaration, and assumed I was right when Agda didn't complain about the redundant keyword

MatthewDaggitt:
Removing this from v2.2

Taneb:
I'd appreciate the help! I've been alternatingly ill, busy, and distracted for a little while. I'll message you on Zulip about this later.

MatthewDaggitt:
@Taneb I think there's already an implementation of O'Connor style reals out there here: 

https://github.com/bobatkey/agda-metric-reals

We should check with @bobatkey, but I think he'd be happy to have them form the basis of something added here?

Taneb:
@MatthewDaggitt while that looks interesting, it doesn't seem to contain a full definition of real numbers, and isn't using O'Connor style reals (so far).

MatthewDaggitt:
Ah sorry, my bad! I have to confess I'm not too across the different formalisations and when I read the citations at the bottom, I assumed it was.

jamesmckinna:
The `merge` conflict now reported is strange (and irritating! `git` can't do better?), so I wonder if, in the interests of making incremental progress, that the `Data.Rational` commits be split out as a smaller, simpler PR... or even two? 

* one for the `nonPos`/`nonNeg` instance arithmetic
* another for the distance metric function, esp. modulo @JacquesCarette 's comment about 
> No idea on how to name this properly in `Data.Rational.Base`. But I was surprised that it is ℚ-valued - do we not have non-negative rationals as a type? That way this would be nonNegative by construction instead of requiring a proof.

Each of those would seem a useful contribution in and of themselves as 'infrastructure'...?

Taneb:
I already pulled the `nonPos`/`nonNeg` etc instance arithmetic out into PR #2496, it has since been merged, but I never updated this branch. Getting the distance metric function in would be great though.

bobatkey:
> @MatthewDaggitt while that looks interesting, it doesn't seem to contain a full definition of real numbers, and isn't using O'Connor style reals (so far).

Not everything is finished (esp multiplication) but my development does include a definition of the reals using O'Connor's definition (and the upper-semi-continuous reals):
1. There is a definition of [Upper Semi-continuous reals](https://github.com/bobatkey/agda-metric-reals/blob/main/src/Data/Real/UpperClosed.agda#L59), which turns O'Connor's definitions using balls into a more standard looking pseudometric space definition.
2. The [completion monad](https://github.com/bobatkey/agda-metric-reals/blob/main/src/MetricSpace/Completion.agda) is defined on pseudometric spaces, using regular functions.
3. The [Cauchy Reals](https://github.com/bobatkey/agda-metric-reals/blob/main/src/MetricSpace/CauchyReals.agda) are the completion of the rationals as a metric space.
4. Addition and negation are defined, with the expected laws. Bounded multiplication is done, but unrestricted multiplication is unfinished because I never worked out how to incorporate the scaling properly. My memory is misty, but I think O'Connor does everything using uniformly continuous functions, and I'm using non-expansive functions, which made things a bit harder.

I do mean to go back and finish this, but the first thing I want to do is tidy up the `Data.Real.UpperClosed` to do proper Dedekind reals (with variations) before the reals as a metric space completion.

jamesmckinna:
> I'd appreciate the help! I've been alternatingly ill, busy, and distracted for a little while. I'll message you on Zulip about this later.

Sorry not have looped back to this before now. Happy to contribute, but it would be good to start form a clean PR, rebased against #2496 before looking at the diffs with @bobatkey 's incomplete development?

bsaul:
This progress on the Reals is exciting to see. I've been noodling on Reals via the so-called [Eudoxus construction](https://bradleysaul.us/bcs-0040.xml) here: https://git.sr.ht/~bradleysaul/eudoxus. I'm not nearly as far along in terms of theorems about the Reals as this PR, but I thought I'd share for comparison's sake.

JacquesCarette:
I see there's been quite a bit of progress since the last discussion happened. Perhaps @Taneb can give us an 'in words' status update?

jamesmckinna:
> Thanks @jamesmckinna , I've taken those suggestions into account. I'd been over-cautious with the `instance` declaration, and assumed I was right when Agda didn't complain about the redundant keyword

Looping back to this earlier thread: see also https://github.com/agda/agda/pull/8367#issuecomment-3864646908 and https://github.com/agda/agda-stdlib/pull/2935#issuecomment-3864299246 ... these are/will become distinct ways of specifying an `instance` field !? So my understanding above appears, now, to have been faulty.

Issue #2274, opened 2024-02-03
[DRY] why is `_≉_` defined both for `Algebra.Bundles.Raw.RawX` bundles, and via `Setoid` instances, for `Algebra.Bundles.X`?


jamesmckinna:
This is another instance of the issues arising when adding 'derived forms' to a(n algebraic) structure/bundle, namely where should they be added:
* to the `RawX` bundle? this is currently the case for `_≉_`, but AFAICT, never used, except to hereditarily be re-exported to richer structures;
* to the `IsX` structure? *not* currently the case for  `_≉_` via an `IsEquivalence` instance, but is for `IsGroup._-_` cf. #2251 ;
* to the `X` bundle? this is *also* the case for `_≉_` via the `Setoid` instance (#2095 , #2162 ), which is in fact a 'derived form' defined in the `IsX` structure... *and* by the opening of the underlying `RawX` bundle.

The first and third cases seem to me to represent a clear DRY violation, and probably arise from the accumulation of  'legacy' design decisions. But what, if any, should be our refactoring policy towards a  revised design?

jamesmckinna:
In the absence of further discussion yet, note there are two possible scenarios:
* where the derived operation indeed is derivable from the `Raw` *signature* (these are the 'easy' examples being considered under #2251 / #2268 );
* where the operation is only definable *under the assumptions that the `IsX` axiomatisation holds* (for some reason, the first example that I can think of is the 'snake lemma' for Abelian categories... but presumably 'easier' naturally occurring examples arise for theories simpler than the essentially-algebraic fragment? Maybe not!?)

I'm not sure what the official (universal algebra) nomenclature is for each kind, but on the model of the theory of formal systems, I supposed the first should be called 'derivable', while the second 'admissible'?

In any case, this issue concerns derivable operations being defined 'too late' to be exported to any subsequently definable extension of the algebra. And that does seem to be a DRY problem... over and above the 'where should such things live?' question...

JacquesCarette:
Finally catching up on really old stuff - my mailbox is quite full of saved email about items that I meant to comment on.  Still have 30 or so from 2024 alone, never mind a whole bunch from 2023!

I do agree that the library is over-eager to define 'helper' conservative extensions that turn out to be seldom used. (I don't know what the official nomenclature is either, but 'derivable' and 'admissible' make a lot of sense.)

[...]

PR #2491, opened 2024-10-02
Add `Raw` bundles to `Relation.Binary.*` hierarchy

jamesmckinna:
This addresses the `Relation.Binary.Bundles.Raw` [part of #2274](https://github.com/agda/agda-stdlib/issues/2274#issuecomment-2112489369) ahead of any subsequent refactoring of the `Algebra` hierarchy in earnest, to exploit the standardisation of vocabulary/fixity envisaged there. 

Adds:
* `RawX` bundles systematically to mirror those in `Relation.Binary.Bundles.X`
* `rawX` manifest fields to each of those in `Relation.Binary.Bundles.X`

NB. Possible issues:
* By contrast with the main `Relation.Bundles` definitions, this introduces `rawSetoid` fields at the top level to each `RawX` bundle, in order to also introduce the negated equality symbol `_≉_` uniformly.
* The `DecX` hierarchy is not mirrored with a parallel `RawDecX` hierarchy... which might otherwise violate the proposed alternative dependency structure outlined in #2252 .
* In the spirit of #2274 the current design for `ApartnessRelation` introduces its negated relation symbol in the `Structures` hierarchy (cf. `Algebra.Structures.IsGroup`...), moreover *without* any fixity declaration; so there are/might be opportunities to rectify this design via `public` re-export...?

### DIFF
--- a/src/Relation/Binary/Bundles/Raw.agda
+++ b/src/Relation/Binary/Bundles/Raw.agda
@@ -32,15 +32,15 @@ record RawSetoid a ℓ : Set (suc (a ⊔ ℓ)) where
 
 
 ------------------------------------------------------------------------
--- RawPreorder
+-- RawRelation: basis for Relation.Binary.Bundles.*Order
 ------------------------------------------------------------------------
 
-record RawPreorder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
-  infix 4 _≈_ _≲_
+record RawRelation c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  infix 4 _≈_ _∼_
   field
     Carrier    : Set c
     _≈_        : Rel Carrier ℓ₁  -- The underlying equality.
-    _≲_        : Rel Carrier ℓ₂  -- The relation.
+    _∼_        : Rel Carrier ℓ₂  -- The underlying relation.
 
   rawSetoid : RawSetoid c ℓ₁
   rawSetoid = record { _≈_ = _≈_ }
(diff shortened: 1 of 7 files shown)


Taneb:
This looks pretty good, but I do think that the changes to `Relation.Binary.Bundles` to export these should happen in this PR

jamesmckinna:
The latest round of debugging commits suggest that the `Eq` submodules of `Reation.Binary.Bundles.X` seem to exist solely to play the 'helper' role envisaged by @JacquesCarette in exporting an extended signature for `Setoid`s with the negated equality symbol... so a 'natural' further refactoring would be to remove all this `Eq` indirection *completely*? (At the cost of reifying my difference of opinion with Jacques...). I'm not clear about the history of these things, but why *do* we have the `Eq` submodules factored out in this way?

jamesmckinna:
> This looks pretty good, but I do think that the changes to `Relation.Binary.Bundles` to export these should happen in this PR

Now added, and manifest at the top updated to reflect those additions.

JacquesCarette:
`git blame` on `Relation.Binary.Bundles` tells me the first appearance of such a `module Eq` dates from September 2019 by @MatthewDaggitt back when the file was `Relation.Binary.Packages` in commit 7e00c92194.

JacquesCarette (commented):
I'm somewhat undecided if this is the right thing to do. The code seems fine. It is definitely more consistent with what has been done elsewhere, so as a default, this is moving in a coherent direction. I'm just not quite ready to 'approve'.

jamesmckinna:
@JacquesCarette writes: 
> I'm somewhat undecided if this is the right thing to do. The code seems fine. It is definitely more consistent with what has been done elsewhere, so as a default, this is moving in a coherent direction. I'm just not quite ready to 'approve'.

Beyond your comments on the originating issue #2274 can you say (more) about why you are/might be hesitating?

JacquesCarette:
Teaching with PLFA right now, have updated to agda-2.7 and stdlib-2.2, and some things have gotten worse. I've got some issues to bring up when I have the time. One example: having `Irrelevant` show up in some of the goals when working on some stuff to do with Negation was unexpected and unwelcome. There's more.

jamesmckinna:
> Teaching with PLFA right now, have updated to agda-2.7 and stdlib-2.2, and some things have gotten worse. I've got some issues to bring up when I have the time. One example: having `Irrelevant` show up in some of the goals when working on some stuff to do with Negation was unexpected and unwelcome. There's more.

Are these issues to do with this PR, or general comments deserving a separate issue for v2.1/v2.2, or even on the PLFA tracker? hard to see how `Relation.Nullary.Irrelevant` might show up, except via `Relation.Nullary.Negation.Core`, other than... that should have happened before v2.0...?

JacquesCarette:
> Are these issues to do with this PR, or general comments deserving a separate issue for v2.1/v2.2,

The example is of the latter kind. But the reason I'm hesitating is related to me finding teaching with post-v2.0 stdlib a bit awkward compared to 1.7.

jamesmckinna:
> > Are these issues to do with this PR, or general comments deserving a separate issue for v2.1/v2.2,
> 
> The example is of the latter kind. But the reason I'm hesitating is related to me finding teaching with post-v2.0 stdlib a bit awkward compared to 1.7.

Still: suggest you move this thread off this PR?

wenkokke:
> > Are these issues to do with this PR, or general comments deserving a separate issue for v2.1/v2.2,
> 
> 
> 
> The example is of the latter kind. But the reason I'm hesitating is related to me finding teaching with post-v2.0 stdlib a bit awkward compared to 1.7.

Could you submit this as a separate issue, so that we can discuss these issues in detail?

JacquesCarette:
@wenkokke wrote:
> Could you submit this as a separate issue, so that we can discuss these issues in detail?

planning to! I've been swamped, so this is still in my ToDo queue.

MatthewDaggitt:
> I'm not clear about the history of these things, but why do we have the Eq submodules factored out in this way? UPDATED: oh, I see that it's one way to ensure that refl and trans as field names don't clash between the underlying isEquivalence and the corresponding IsPreorder properties. I wonder if that was the good/better/best design choice, in favour of having fields names ≲-refl, ≲-trans etc.? Qualified names... hmmm.

Yes, in retrospect it wasn't the best idea, but at the time, it was the only way to add them that preserved backwards compatibility.

src/Relation/Binary/Bundles/Raw.agda
```diff
@@ -0,0 +1,118 @@
...
+  infix 4 _≉_
+  _≉_ : Rel Carrier _
+  x ≉ y = ¬ (x ≈ y)
+
+
+------------------------------------------------------------------------
+-- RawPreorder
+------------------------------------------------------------------------
+
+record RawPreorder c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
```
MatthewDaggitt, line 38: Hmm so following the pattern of `Algebra.Bundles.Raw`, then bundle equivalence is determined not by the names of the relation, but by their signature. For example both `CommutativeMonoid` and `Monoid` both have a single `RawMonoid` base.

Therefore to be consistent I think `RawPreorder`, `RawPartialOrder`, `RawStrictPartialOrder` and `RawApartness` relation should all be a single record... With what name I don't know...

If the purpose of this is to introduce syntax, then I think we need to use some other mechanism other than `Raw` bundles...

MatthewDaggitt (changes requested):
The other refactorings look good though!

src/Relation/Binary/Bundles/Raw.agda
jamesmckinna: Two things:
* good point about shared signatures (and `\~` would be the 'canonical' choice if symbol here?), and then let the corresponding `Raw`sub bundles in `Bundles` handle the renaming to orthodox/rectified syntax?
* as with discussions with @JacquesCarette , for me the issue is one if being able, for a given raw signature, to enrich it with 'obvious' derived operations which allow smoother treatment of subsequent properties etc. The 'syntax' aspect here is really about standardising notation for these otherwise useful/important derived notions...?

jamesmckinna:
So, while I would still be happy/happier to see this merged (despite @JacquesCarette 's hesitation about doing so), it's somewhat ironic that we seem, via this route, to have reinstated the 'old' vocabulary (and then erased it via `renaming`... ;-)), cf. #2099 and its precursor issues/PRs. But I think that each time I/we go round this merry-go-round, things do actually improve?

gallais (approved):
This nicely refactors concepts that were free-floating until now.

JacquesCarette (approved):
I relent. This is an improvement, on the balance of things, esp. for v2.*.

JacquesCarette:
@jamesmckinna @gallais and I have verified that the changes asked for have been done. So we're over-riding the blocked merge from @MatthewDaggitt .

PR #2492, opened 2024-10-02
Add properties of `Data.Refinement` and refactor module structure

jamesmckinna:
NB.
* could be folded into `Data.Refinement` directly?
* Fairbairn threshold? is this injectivity proof not *immediate*?

### DIFF
--- a/src/Data/Refinement/Properties.agda
+++ b/src/Data/Refinement/Properties.agda
@@ -1,0 +1,24 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of refinement types
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Refinement.Properties where
+
+open import Level
+open import Data.Refinement
+open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
+open import Relation.Nullary.Recomputable
+
+private
+  variable
+    a p : Level
+    A : Set a
+
+module _ {P : A → Set p} where
+
+  value-injective : {v w : Refinement A P} → value v ≡ value w → v ≡ w
+  value-injective refl = refl
--- a/src/Data/Refinement/Properties.agda
+++ b/src/Data/Refinement/Properties.agda
@@ -11,7 +11,6 @@ module Data.Refinement.Properties where
 open import Level
 open import Data.Refinement
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
-open import Relation.Nullary.Recomputable
 
 private
   variable
--- a/src/Data/Refinement/Properties.agda
+++ b/src/Data/Refinement/Properties.agda
@@ -9,15 +9,17 @@
 module Data.Refinement.Properties where
 
 open import Level
-open import Data.Refinement
-open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
+open import Data.Refinement.Base
+open import Relation.Binary.Definitions using (DecidableEquality)
+open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl; cong)
+import Relation.Nullary.Decidable.Core as Dec
 
 private
   variable
     a p : Level
     A : Set a
     P : A → Set p
-    v w : Refinement A P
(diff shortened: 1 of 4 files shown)


src/Data/Refinement/Properties.agda
```diff
@@ -0,0 +1,23 @@
...
+open import Level
+open import Data.Refinement
+open import Relation.Binary.PropositionalEquality.Core using (_≡_; refl)
+
+private
+  variable
+    a p : Level
+    A : Set a
+
+module _ {P : A → Set p} where
```
MatthewDaggitt, line 20: As `P` is used in the type of `Refinement`, it's fine to add it as a variable. In fact you could do this in `Data.Refinement` as well if you were feeling keen.

MatthewDaggitt (approved):
Yup, I this is definitely useful! Also I think `Properties` is definitely the right place for it, even though it seems like overkill at the moment. I'm sure more stuff will be added to it!

gallais (approved):

jamesmckinna:
> Yup, I this is definitely useful! Also I think `Properties` is definitely the right place for it, even though it seems like overkill at the moment. I'm sure more stuff will be added to it!

Two things: 
* I do have some additional properties (concerning iterated refinement) which occurred to me during the writing of this (downstream PR?)
* Should we followup with a refactoring into an orthodox `.Base`/`.Properties` division? (also could be handled downstream)

src/Data/Refinement/Properties.agda
jamesmckinna: Will do!

gallais:
Yes please: if it grows enough, let's split it up according to the hierarchy!

jamesmckinna:
OK, I've added a proof of (inheritance of) decidable equality, and refactored as `Base`/`Properties`. I'll stop there.

jamesmckinna:
Does @fredrikNordvallForsberg 's resolution of #2495 mean that we could simplify the implementation of `Data.Refinement` to simply make the `proof` field take the `.` modality?
```agda
record Refinement {a p} (A : Set a) (P : A → Set p) : Set (a ⊔ p) where
  constructor _,_
  field value : A
        .proof : P value
infixr 4 _,_
open Refinement public

module _ {P : A → Set p} {Q : B → Set q} where

  map : (f : A → B) → ∀[ P ⇒ f ⊢ Q ] →
        [ a ∈ A ∣ P a ] → [ b ∈ B ∣ Q b ]
  map f prf (a , p) = f a , prf p
```
etc.

Not for this PR, but a potential simplifying refactoring/deprecation downstream?

gallais:
You don't get irrelevant projections so IME it's nicer to package up things affected by the irrelevant
modality so that you can project them out and pass them around to combinators.

JacquesCarette:
The old category library used to use irrelevance *a lot* and then broke in ways we couldn't fix at some point in Agda's evolution. So I'd be quite concerned that we'd run up against bugs "just too late" if we start using it too much in the library.

Irrelevant projections (and the lack thereof) are a big deal. Conor and I more recently tried to use irrelevance quite aggressively to force Agda to ensure we stay honest (we're doing quotients "by hand"), and gave up. Now we stay honest by sheer force of will.

jamesmckinna:
Ready to merge now?

src/Data/Refinement/Properties.agda
```diff
@@ -0,0 +1,35 @@
...
+-- Basic properties
+
+value-injective : value v ≡ value w → v ≡ w
+value-injective refl = refl
+
+module _ (eq? : DecidableEquality A) where
+
+  infix 4 _≟_
+  _≟_ : DecidableEquality [ x ∈ A ∣ P x ]
+  v ≟ w = Dec.map′ value-injective (cong value) (eq? (value v) (value w))
```
jamesmckinna, line 35: This is clearly a pattern, which currently is reified only under `Relation.Decidable.Nullary.Decidable.via-injection`, but it seems excessive to create an `Injection` bundle solely for the sake of applying the lemma, so suggest some downstream refactoring to make this a property of `Injective` (bare) functions?

JacquesCarette (approved):

PR #2494, opened 2024-10-04
Fix type for ≡-syntax in heterogeneous equality

ruifengx:
The original definition forces LHS and RHS of the result type to be the same. I do not find any reason for this restriction, and it prevents writing the following code (which I believe is intention of `≡-syntax` operators):
```agda
example = begin
  some expression        ≡⟨⟩
  some other expression  ≡⟨⟩
  ...
```
where `some expression` and `some other expression` are definitionally equal.

I have verified locally that the modified code type-checks with Agda 2.7.0.1.

### DIFF
--- a/src/Relation/Binary/HeterogeneousEquality.agda
+++ b/src/Relation/Binary/HeterogeneousEquality.agda
@@ -240,16 +240,16 @@ module ≅-Reasoning where
   start : ∀ {x : A} {y : B} → x IsRelatedTo y → x ≅ y
   start (relTo x≅y) = x≅y
 
-  ≡-go : ∀ {A : Set a} → Trans {A = A} {C = A} _≡_ _IsRelatedTo_ _IsRelatedTo_
+  ≡-go : ∀ {A : Set ℓ} {B : Set ℓ} → Trans {A = A} {C = B} _≡_ _IsRelatedTo_ _IsRelatedTo_
   ≡-go x≡y (relTo y≅z) = relTo (trans (reflexive x≡y) y≅z)
 
   -- Combinators with one heterogeneous relation
   module _ {A : Set ℓ} {B : Set ℓ} where
     open begin-syntax (_IsRelatedTo_ {A = A} {B}) start public
+    open ≡-syntax (_IsRelatedTo_ {A = A} {B}) ≡-go public
 
   -- Combinators with homogeneous relations
   module _ {A : Set ℓ} where
-    open ≡-syntax (_IsRelatedTo_ {A = A}) ≡-go public
     open end-syntax (_IsRelatedTo_ {A = A}) (relTo refl) public
 
   -- Can't create syntax in the standard `Syntax` module for


jamesmckinna:
When preparing a branch towards a PR, *please* do not push commits to your local `master` branch, as this makes it very difficult to do a `git fetch` from your branch in order to be able to review your code (name clash with local `master` branch`)... instead, as described in the `HACKING` guide, please make a new branch off your own `master`, as eg `git checkout -b heterogeneous-equality-levels`.

src/Relation/Binary/HeterogeneousEquality.agda
```diff
@@ -240,16 +240,16 @@ module ≅-Reasoning where
   start : ∀ {x : A} {y : B} → x IsRelatedTo y → x ≅ y
   start (relTo x≅y) = x≅y
 
-  ≡-go : ∀ {A : Set a} → Trans {A = A} {C = A} _≡_ _IsRelatedTo_ _IsRelatedTo_
+  ≡-go : ∀ {A : Set ℓ} {B : Set ℓ} → Trans {A = A} {C = B} _≡_ _IsRelatedTo_ _IsRelatedTo_
```
jamesmckinna, line 243: This, I think, will still not have as much level polymorphism as you intend with this PR: level `ℓ` will be generalised, as per the `variable` declarations governing the module, but `A` and `B`, being explicitly mentioned in the type here, will each be bound *at the same level*.
```suggestion
  ≡-go : ∀ {A : Set a} {B : Set b} → Trans {A = A} {C = B} _≡_ _IsRelatedTo_ _IsRelatedTo_
```

src/Relation/Binary/HeterogeneousEquality.agda
```diff
@@ -240,16 +240,16 @@ module ≅-Reasoning where
...
   ≡-go x≡y (relTo y≅z) = relTo (trans (reflexive x≡y) y≅z)
 
   -- Combinators with one heterogeneous relation
   module _ {A : Set ℓ} {B : Set ℓ} where
```
jamesmckinna, line 247: Hence the levels here, as a knock-on change, should also be changed...
```suggestion
  module _ {A : Set a} {B : Set b} where
```

jamesmckinna (changes requested):
Thanks for a first PR!
Some suggested changes will get you closer to what I understand to be the intention, namely that heterogeneous equality (and its equational `Reasoning` combinators/syntax, should be made more level-polymorphic.

You will also need a `CHANGELOG` entry recording the change: suggest under the heading `Minor improvements`?

ruifengx:
> When preparing a branch towards a PR, _please_ do not push commits to your local `master` branch

Sorry, I should have checked the guide carefully, and I will remember this next time. GitHub does not allow me to change the source branch after the PR is created, but I can reopen a new one if that is more convenient for you.

src/Relation/Binary/HeterogeneousEquality.agda
ruifengx: Right, I forgot that HEq also supports different levels. I just modified `_IsRelatedTo_`, `≡-go`, and the anonymous module accordingly.

src/Relation/Binary/HeterogeneousEquality.agda
```diff
@@ -234,22 +234,22 @@ module ≅-Reasoning where
 
   infix 4 _IsRelatedTo_
 
-  data _IsRelatedTo_ {A : Set ℓ} {B : Set ℓ} (x : A) (y : B) : Set ℓ where
+  data _IsRelatedTo_ {A : Set a} {B : Set b} (x : A) (y : B) : Set a where
```
jamesmckinna, line 237: And... the `Level` of the definition now also needs to be changed: if you write `Set _`, then Agda will correctly infer the (minimum necessary) level as `Set (a ⊔ b)` but better to be explicit here, I think.
```suggestion
  data _IsRelatedTo_ {A : Set a} {B : Set b} (x : A) (y : B) : Set (a ⊔ b) where
```

ruifengx: I have verified that just `a` works (it type-checks locally with Agda 2.7.0.1), because the level `b` will actually be equal to `a` in the single constructor `relTo`. Due to the same reason, the equality itself also have a similar level constraint:
```agda
data _≅_ {A : Set a} (x : A) : {B : Set b} → B → Set a where
   refl : x ≅ x
```

jamesmckinna: Good catch! My mistake.

jamesmckinna:
> > When preparing a branch towards a PR, _please_ do not push commits to your local `master` branch
> 
> Sorry, I should have checked the guide carefully, and I will remember this next time. GitHub does not allow me to change the source branch after the PR is created, but I can reopen a new one if that is more convenient for you.

Given the size of the changes you propose, this was easy enough to duplicate locally to experiment with, but in general (and from next time... ;-) indeed please have a fresh branch for each PR.

JacquesCarette (changes requested):
Needs a CHANGELOG - but otherwise looks good now.

ruifengx:
Thanks! I added an entry to the CHANGELOG under minor improvements.

I am a bit hesitant on whether this counts as a bug fix or a minor improvement. In the version before this change, the operators from `≡-syntax` do not have the correct type to fit into equational reasoning (using any of them forces the resulting equality to have the same type for LHS and RHS, which defeats the purpose of HEq), so arguably this can be considered as a bug.

JacquesCarette (approved):
I'm not entirely sure either, but moving things around before release in CHANGELOG is easy enough, so I'm approving now.

jamesmckinna (approved):

PR #2496, opened 2024-10-14
Lemmas for Positive, Negative, etc. and _+_ and _*_ for rationals

Taneb:
This is pulled out from my branch for real numbers

A discovered annoyance (bug?) is that `*-monoˡ-≤-nonNeg` corresponds to `*-monoʳ-<-pos`, _not_ `*-monoˡ-<-pos` as one may expect. But I don't want to fix that in this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -316,6 +316,18 @@ Additions to existing modules
   m≤pred[n]⇒suc[m]≤n : .{{NonZero n}} → m ≤ pred n → suc m ≤ n
   ```
 
+* New lemmas in `Data.Rational.Properties`:
+  ```agda
+  nonNeg+nonNeg⇒nonNeg : ∀ p .{{_ : NonNegative p}} q .{{_ : NonNegative q}} → NonNegative (p + q)
+  nonPos+nonPos⇒nonPos : ∀ p .{{_ : NonPositive p}} q .{{_ : NonPositive q}} → NonPositive (p + q)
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

jamesmckinna:
> This is pulled out from my branch for real numbers

Looks good!

> A discovered annoyance (bug?) is that `*-monoˡ-≤-nonNeg` corresponds to `*-monoʳ-<-pos`, _not_ `*-monoˡ-<-pos` as one may expect. But I don't want to fix that in this PR

Yes, I think this is an(other) example of left/right being mixed up in the library, and at some point, we should rationalise (and correctly implement) our conventions. Cf. https://github.com/agda/agda-stdlib/issues/1436#issuecomment-1243841859 and #1579 etc.

jamesmckinna (approved):
Two comments:
* it would be great to find a way to make even the arguments `p`, `q` themselves inferrable, but I suspect this isn't possible without quotation/indirection, because of the usual 'arithmetic operations can't infer their arguments' objection (but I still have the fantasy that one day we might overcome that)
* all the lemmas are in `Data.Rational.Properties`: is there any relationship with/need for, any (additional) ones in `Data.Rationa.Unnormalised.Properties`? (I've never been quite clear about the division of labour between the two...)

Taneb:
> all the lemmas are in `Data.Rational.Properties`: is there any relationship with/need for, any (additional) ones in `Data.Rational.Unnormalised.Properties`? (I've never been quite clear about the division of labour between the two...)

I'll admit I didn't do that simply out of laziness. But looking at it now, I don't think in this case doing it that way would be helpful.

PR #2499, opened 2024-11-13
Mess around with IsSemiringWithoutOne reexports

Taneb:
It was missing several and had some duplicated, which was causing problems when trying to use those names

I don't know how to changelog this

### DIFF
--- a/src/Algebra/Structures.agda
+++ b/src/Algebra/Structures.agda
@@ -415,15 +415,22 @@ record IsSemiringWithoutOne (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
     zero                  : Zero 0# *
 
   open IsCommutativeMonoid +-isCommutativeMonoid public
-    using (setoid)
+    using (setoid; isEquivalence)
     renaming
-    ( comm                   to +-comm
+    ( ∙-cong                 to +-cong
+    ; ∙-congˡ                to +-congˡ
+    ; ∙-congʳ                to +-congʳ
+    ; assoc                  to +-assoc
+    ; identity               to +-identity
+    ; identityˡ              to +-identityˡ
+    ; identityʳ              to +-identityʳ
+    ; comm                   to +-comm
     ; isMonoid               to +-isMonoid
     ; isCommutativeMagma     to +-isCommutativeMagma
     ; isCommutativeSemigroup to +-isCommutativeSemigroup
     )
 
-  open Setoid setoid public
+  open IsEquivalence isEquivalence public
 
   *-isMagma : IsMagma *
   *-isMagma = record
@@ -443,6 +450,12 @@ record IsSemiringWithoutOne (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
     ( ∙-congˡ to *-congˡ
     ; ∙-congʳ to *-congʳ
     )
+
+  distribˡ : * DistributesOverˡ +
+  distribˡ = proj₁ distrib
+
+  distribʳ : * DistributesOverʳ +
+  distribʳ = proj₂ distrib
 
   zeroˡ : LeftZero 0# *
   zeroˡ = proj₁ zero
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
 
+* The names exposed by the `IsSemiringWithoutOne` record have been altered to
+  better correspond to other algebraic structures. In particular:
+  * `Carrier` is no longer exposed.
+  * Several laws have been re-exposed from `IsCommutativeMonoid +` renaming
+    them to name the operation `+`.
+  * `distribˡ` and `distribʳ` are defined in the record.
+
 Minor improvements
 ------------------
 


src/Algebra/Structures.agda
```diff
@@ -415,15 +415,22 @@ record IsSemiringWithoutOne (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
...
+    ; identityˡ              to +-identityˡ
+    ; identityʳ              to +-identityʳ
+    ; comm                   to +-comm
     ; isMonoid               to +-isMonoid
     ; isCommutativeMagma     to +-isCommutativeMagma
     ; isCommutativeSemigroup to +-isCommutativeSemigroup
     )
 
-  open Setoid setoid public
+  open IsEquivalence isEquivalence public
```
JacquesCarette, line 433: Doesn't this removal of `open Setoid setoid public` change the export, i.e. `Carrier` is no longer visible?

JacquesCarette (changes requested):
Also need a Changelog entry.

src/Algebra/Structures.agda
Taneb: Yes, this is intentional - no other structure exposes `Carrier`, and doing so was causing issues when trying to use the corresponding bundle.

MatthewDaggitt:
> I don't know how to changelog this

I would document this as a bug fix. Noting the removal of the `Carrier` export and then listing the new exports.

MatthewDaggitt (approved):

jamesmckinna:
As the author of the prior, now offending, commit, I [drew attention](https://github.com/agda/agda-stdlib/issues/1917#issuecomment-1719329631) in the discussion comments on the relevant PR to the possibility of having introduced such a bug...

... having said that, the `setoid` export was for a reason, and I think that the revisions here revert the relevant consequence (namely to export negated equality...), so I think I'd prefer it if instead 
* either `Carrier` is part if an explicit `hiding` directive when exporting `setoid` (open world assumption)
* or, that the exports are specifically restricted via a `using` directive

There's a meta-issue here:
* the offending commit passed an approving review by the author of the present PR
* the originating history of the offending commit does not seem to have been referred back to for context/rationale/explanation

This is not to throw shade either on me (fir a buggy commit), or @Taneb for the fix here, but to point out that, wittingly or unwittingly, revert prior commits may have further subsequent negative downstream consequences. 

Happy to be corrected if I've is construed the situation here!

src/Algebra/Structures.agda
jamesmckinna: See longer comment below

src/Algebra/Structures.agda
```diff
@@ -415,15 +415,22 @@ record IsSemiringWithoutOne (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
     zero                  : Zero 0# *
 
   open IsCommutativeMonoid +-isCommutativeMonoid public
-    using (setoid)
+    using (setoid; isEquivalence)
```
jamesmckinna, line 418: I'd double down on @Taneb 's suggestions, and remove the `setoid` export entirely. As noted in my longer meta-comment, I think it's a design error for a `Structure` to export a sub-`Bundle`... one to chalk up to my relative inexperience/lack of understanding when raising and then tackling #1917 which caused all this in the first place. hey-ho!

src/Algebra/Structures.agda
jamesmckinna: Indeed, no (other) `Structure` exports `setoid`, nor I think, should.

jamesmckinna (approved):
Suggest removing the export of `setoid` completely... but not (yet) sure what the knock-on consequences of that might be...

jamesmckinna:
> Suggest removing the export of `setoid` completely... but not (yet) sure what the knock-on consequences of that might be...

Ugh. Worse that I thought: the `unique`ness properties of inverse, defined as manifest fields, in `Algebra.Structures.IsGroup`, rely on `Algebra.Consequences.Setoid`, which depends, as the name suggests, on having a `Bundle`d `Setoid`, rather than the underlying `IsEquivalence` for access to the properties of equality. So following my suggestion (to have `IsMagma` *not* export a `setoid` field) would actually have a larger scale knock-on effect.

Nevertheless, I think deleting the `setoid` export from `IsSemiringWithoutOne` should be harmless?

src/Algebra/Structures.agda
jamesmckinna: Re: the export of `setoid` from `IsMagma`: this is used by `IsGroup` in this module, so can't be removed at this stage.

MatthewDaggitt:
Gah these algebra imports, re-exports are always nightmarish. I'm not across the exact details, but my understanding aligns with both @jamesmckinna and @Taneb in that we try to avoid structures exporting bundles.

src/Algebra/Structures.agda
```diff
@@ -444,6 +451,12 @@ record IsSemiringWithoutOne (+ * : Op₂ A) (0# : A) : Set (a ⊔ ℓ) where
     ; ∙-congʳ to *-congʳ
     )
 
+  distribˡ : * DistributesOverˡ +
+  distribˡ = proj₁ distrib
+
+  distribʳ : * DistributesOverʳ +
+  distribʳ = proj₂ distrib
+
```
jamesmckinna, line 459: These additions make sense, but if you are going to make them, then perhaps you should re-export them downstream from eg `IsRingWithoutOne` via opening the `isSemiringWithoutOne` manifest field... etc. which might simplify those exports elsewhere, too?

jamesmckinna (commented):
My suggestions could be considered downstream, but I think are worth doing here as immediate consequences of the new exports you propose?

jamesmckinna:
@Taneb do you want me to add a `CHANGELOG` entry and merge this?

jamesmckinna:
Lots of other merges happening right now... so `CHANGELOG` conflict sadly inevitable!

jamesmckinna:
@JacquesCarette I think you can now approve the changes and merge?

JacquesCarette (approved):

PR #2501, opened 2024-11-18
Add filter-≐

Taneb:


JacquesCarette (commented):
Overall I think this is a good addition, but I do have a question about the implementation, which is why I'm neither approving nor requesting a change.

JacquesCarette (approved):

PR #2503, opened 2024-11-23
Add `≤‴-irrelevant` to `Data.Nat.Properties`

tsung-ju:


jamesmckinna:
Thanks for the proposed contribution.

Two quick comments in place of a proper review:
* besides 'completeness', what's the rationale for the addition? (this relation is rarely used in the library, and never, AFAIK, in a context where its being `Irrelevant` is important)
* the proofs do look very gnarly (esp. the use of `subst`), so I wonder if they can be simplified.

If we do agree to add this, you will need a `CHANGELOG` entry as well.

src/Data/Nat/Properties.agda
```diff
@@ -2237,6 +2237,24 @@ _>‴?_ = flip _<‴?_
...
 ≤‴⇒≤ = ≤″⇒≤ ∘ ≤‴⇒≤″
 
+≤‴-irrelevant : Irrelevant {A = ℕ} _≤‴_
+≤‴-irrelevant ≤‴-refl q = lemma q refl
+  where
+  lemma : ∀ {m n} → (p : m ≤‴ n) → (e : m ≡ n) → subst (m ≤‴_) e ≤‴-refl ≡ p
+  lemma ≤‴-refl       e rewrite ≡-irrelevant e refl = refl
+  lemma (≤‴-step m<m) refl = contradiction (≤‴⇒≤ m<m) (<-irrefl refl)
+≤‴-irrelevant (≤‴-step x<x) ≤‴-refl = contradiction (≤‴⇒≤ x<x) (<-irrefl refl)
+≤‴-irrelevant (≤‴-step p)   (≤‴-step q) rewrite ≤‴-irrelevant p q = refl
```
jamesmckinna, line 2247: Suggest replacing the use of `rewrite` with one of `cong` (less strict; more robust):
```suggestion
≤‴-irrelevant (≤‴-step p)   (≤‴-step q) = cong ≤‴-step $ ≤‴-irrelevant p q
```

src/Data/Nat/Properties.agda
```diff
@@ -2237,6 +2237,24 @@ _>‴?_ = flip _<‴?_
 ≤‴⇒≤ : _≤‴_ ⇒ _≤_
 ≤‴⇒≤ = ≤″⇒≤ ∘ ≤‴⇒≤″
 
+≤‴-irrelevant : Irrelevant {A = ℕ} _≤‴_
+≤‴-irrelevant ≤‴-refl q = lemma q refl
+  where
+  lemma : ∀ {m n} → (p : m ≤‴ n) → (e : m ≡ n) → subst (m ≤‴_) e ≤‴-refl ≡ p
+  lemma ≤‴-refl       e rewrite ≡-irrelevant e refl = refl
+  lemma (≤‴-step m<m) refl = contradiction (≤‴⇒≤ m<m) (<-irrefl refl)
+≤‴-irrelevant (≤‴-step x<x) ≤‴-refl = contradiction (≤‴⇒≤ x<x) (<-irrefl refl)
```
jamesmckinna, line 2246: These two clauses suggest a refactoring to abstract out an appeal to `Irreflexive _≡_ _≤‴_` as follows:
```suggestion
  lemma (≤‴-step m<m) e with () <‴-irrefl e m<m
≤‴-irrelevant (≤‴-step m<m) ≤‴-refl  with () <‴-irrefl refl m<m
```
with a prior definition (before L2240) as follows:
```agda
<‴-irrefl : Irreflexive _≡_ _<‴_
<‴-irrefl eq = <-irrefl eq ∘ ≤‴⇒≤
```

jamesmckinna (changes requested):
Suggest adding the proof of `Irreflexive _≡_ _≤‴_` in order to make use of it.

Needs `CHANGELOG`.

tsung-ju:
Hi, thanks for the feedback!

My use case for `m ≤‴ n` is to represent numbers with a lower bound `m`, using `≤‴-refl` and `≤‴-step` to adjust the lower bound. I needed irrelevance to prove decidable equality for a data structure containing these proofs. The proof was tricky under without-K, so I thought it might be a helpful addition. That said, I’m not opposed to deprecating or revising `_≤‴_`, as it can be difficult to work with in its current form.

I’ve addressed the easier suggestions and will update the CHANGELOG soon. I’ll also explore possible simplifications using `cast`.

jamesmckinna:
> I’ve addressed the easier suggestions and will update the CHANGELOG soon. I’ll also explore possible simplifications using `cast`.

This last part is *entirely* optional, except to draw your attention to the `subst` (fully general)/`cast` (specific on a per-type/-per-predicate/per-relation basis) distinction as a basis for 'subst-like' reasoning where matching on `refl` is actually doing too much, or the wrong kind of, evaluation (in particular, work that isn't erasable at run-time in a backend).

jamesmckinna (approved):

MatthewDaggitt (approved):

MatthewDaggitt:
Thanks for the contribution! Merging in now.

PR #2505, opened 2024-11-24
Add `partition-is-foldr` to `Data.List.Properties`

carlostome:


src/Data/List/Properties.agda
```diff
@@ -1294,6 +1294,14 @@ module _ {P : Pred A p} (P? : Decidable P) where
   ...  | true  = Product.map s≤s m≤n⇒m≤1+n ih
   ...  | false = Product.map m≤n⇒m≤1+n s≤s ih
 
+  partition-is-foldr : partition P? ≗ foldr (λ x → if does (P? x) then Product.map₁ (x ∷_)
+                                                                  else Product.map₂ (x ∷_))
+                                            ([] , [])
```
jamesmckinna, line 1299: I'd be tempted to insert some linebreaks:
```suggestion
  partition-is-foldr : partition P? ≗ foldr
    (λ x → if does (P? x) then Product.map₁ (x ∷_) else Product.map₂ (x ∷_))
    ([] , [])
```
here, but not necessarily in the `CHANGELOG`.

jamesmckinna (approved):

gallais:
My only worry is that given that `foldr` can encode any function, should we have such a
lemma no matter how complex the `cons` step is? Relatedly, do we want to simplify this
one by defining e.g.

```
inj : Bool → A → A ⊎ A
_⊎∷_ : A ⊎ B → List A × List B → List A × List B

partition P? = foldr (\ x -> inj (does P? x) x ⊎∷_) ([], [])
```

jamesmckinna:
@gallais 

I think there is indeed a question as why the definitions are given in `Data.List.Base` are given in 'plain' form, and *not* as instances of `fold`, but given that they are, it makes/made sense to me to accompany each of them with a lemma of this kind. But if we are going to widen the discussion to consider all of those definitions in the library, then... maybe a separate `library-design` issue worth opening?

As for the (straw-man?) refactoring you propose, I think it falls to Occam's Razor?

JacquesCarette:
I wouldn't call this `partition-is-foldr` but `partition-as-foldr` , or `partition-via-foldr`. And yes, `foldr` being the `List` eliminator, we need to be careful to not have an explosion of these. Having said that, I'm fine with `partition` as it is an important function.

jamesmckinna:
> I wouldn't call this `partition-is-foldr` but `partition-as-foldr` , or `partition-via-foldr`. And yes, `foldr` being the `List` eliminator, we need to be careful to not have an explosion of these. Having said that, I'm fine with `partition` as it is an important function.

But we already have
* `id-is-foldr`
* `++-is-foldr`
* `map-is-foldr`
* ...

So it does explode into a deprecation-fest if we follow your lead? v3.0?

JacquesCarette:
Sigh - yes, v3.0. Inconsistency is worse than misleading names.

JacquesCarette (approved):

PR #2507, opened 2024-11-30
add `¬First⇒All` lemma to `Data.List.Relation.Unary.First.Properties`

wasabi315:
This PR adds a `¬Any⇒All¬` counterpart for `First`:
```agda
¬First⇒All : ∁ Q ⊆ P → ∀ xs → ∁ (First P Q) xs → All P xs

-- cf.
-- In Data.List.Relation.Unary.All.Properties.Core
¬Any⇒All¬ : ∀ xs → ¬ Any P xs → All (¬_ ∘ P) xs
```

Issues
- Should `xs` be an explicit argument or implicit?
  - Explicit for consistency with `¬Any⇒All¬` for now
- Is`∁ (First P Q) xs` preferable over `¬ First P Q xs`?

### DIFF
--- a/src/Data/List/Relation/Unary/First/Properties.agda
+++ b/src/Data/List/Relation/Unary/First/Properties.agda
@@ -59,6 +59,12 @@ module _ {a p q} {A : Set a} {P : Pred A p} {Q : Pred A q} where
   First⇒¬All q⇒¬p [ qx ]     (px ∷ pxs) = q⇒¬p qx px
   First⇒¬All q⇒¬p (_ ∷ pqxs) (_ ∷ pxs)  = First⇒¬All q⇒¬p pqxs pxs
 
+  ¬First⇒All : ∁ Q ⊆ P → ∀ xs → ∁ (First P Q) xs → All P xs
+  ¬First⇒All ¬q⇒p []       _      = []
+  ¬First⇒All ¬q⇒p (x ∷ xs) ¬pqxxs =
+    let px = ¬q⇒p (¬pqxxs ∘ [_]) in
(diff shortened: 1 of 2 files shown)


src/Data/List/Relation/Unary/First/Properties.agda
```diff
@@ -59,6 +59,12 @@ module _ {a p q} {A : Set a} {P : Pred A p} {Q : Pred A q} where
   First⇒¬All q⇒¬p [ qx ]     (px ∷ pxs) = q⇒¬p qx px
   First⇒¬All q⇒¬p (_ ∷ pqxs) (_ ∷ pxs)  = First⇒¬All q⇒¬p pqxs pxs
 
+  ¬First⇒All : ∁ Q ⊆ P → ∀ xs → ∁ (First P Q) xs → All P xs
```
ncfavier, line 62: Why not `∁ (First P Q) ⊆ All P`?

wasabi315: My idea was to make `¬First⇒All`'s signature close to `¬Any⇒All¬`'s where the argument `xs` is explicit. Using `P Relation.Unary.⊆ Q`, it will be implicit on the other hand.
But, I just found `Relation.Unary._⊆′_`, the explicit version of `_⊆_`. I should have used it instead.

As mentioned in the first comment, I am not sure whether to make the argument `xs` explicit or implicit.
In `¬Any⇒All¬` it is explicit, but why is it?

wasabi315: I am not sticking to having it explicit, so I will update the signature to `¬First⇒All : ∁ Q ⊆ P → ∁ (First P Q) ⊆ All P` then.

MatthewDaggitt (approved):

PR #2510, opened 2024-12-06
[ add ] lemma relating `Propositional` and `Setoid` versions of `Sublist`

jamesmckinna:
This implements the lemma requested by @mechvel on [zulip](https://agda.zulipchat.com/#narrow/channel/264623-stdlib/topic/.60.60_.E2.8A.86.E2.82.80_.20.E2.87.92.20_.E2.8A.86.E2.82.97_.60.60.20for.20sublist.20by.20propositional.20vs.20by.20Setoid)

[DELETED]

UPDATED:
* lemma in `Propositional.Properties`
* reduces to a triviality involving `Setoid.map`
* Fairbairn threshold?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,6 +192,11 @@ Additions to existing modules
                                             ([] , [])
   ```
 
+* In `Data.List.Relation.Binary.Sublist.Propositional.Properties`:
+  ```agda
+  ⊆⇒⊆~ : IsEquivalence ≈  → ∀ {as bs} → as ⊆ bs → Setoid._⊆_ setoid≈ as bs
+  ```
+
 * In `Data.List.Relation.Unary.Any.Properties`:
   ```agda
   concatMap⁺ : Any (Any P ∘ f) xs → Any P (concatMap f xs)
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
I really do feel that this property should live in `Propositional.Properties`... I've made a PR (https://github.com/agda/agda-stdlib/pull/2514) that should make this possible. Let's see if that passes first?

jamesmckinna:
> I really do feel that this property should live in `Propositional.Properties`

As with the original Zulip discussion, I'm agnostic about where it best should live, but in fact to my surprise, it was super easy to add the 'right' version of the proof to `Setoid`, *independent of* the considerations of #2514 ... but happy to be steered by you downstream... (sic!)

MatthewDaggitt:
> But perhaps it makes for a slightly 'cleaner' dependency graph? (Propositional.Properties depends on Setoid, but not Setoid.Properties depends on Propositional?)

Yes, I think that this is the key. We should try and have the `Propositional` files only depend the `Setoid` files, and not vice-versa.

JacquesCarette (approved):

jamesmckinna:
I'll revise in favour of the dependency you favour.

PR #2512, opened 2024-12-08
Update installation instructions

### LINKED ISSUE #2500: Update installation instructions wrt install dir?
The installation directions suggest that the agda library information directory lives at $HOME/.agda/libraries. However, the documentation at `https://agda.readthedocs.io/en/latest/tools/package-system.html` suggests that this is out of date (though still supported), and suggests identifying the location using `agda --print-agda-app-dir` ... which indeed, on my macOS system, returns $HOME/.config/agda rather than $HOME/.agda. So perhaps the installation directions here are out of date? Or perhaps there should just be a link to the `readthedocs.io` site?

Apologies if I'm wading into a religious debate that I'm not aware of :).

jamesmckinna:
Fixes #2500 .

### DIFF
--- a/doc/installation-guide.md
+++ b/doc/installation-guide.md
@@ -12,7 +12,7 @@ Use version v2.1.1 of the standard library with Agda 2.7.0. You can find the cor
    done manually by visiting the Github repository for the library, or via the
    command line as follows:
    ```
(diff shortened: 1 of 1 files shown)


jamesmckinna:
@jbclements can you take a look at this please, as the author of the OP?

MatthewDaggitt (approved):

jbclements:
Many thanks for the pull request; your change is better than mine would have been!

PR #2515, opened 2024-12-09
[ refactor ] Exchange components of `Relation.Binary.Definitions._Respects₂_`

### LINKED ISSUE #2471: `Relation.Binary.Definitions._Respects₂_` seems to exchange 'left' and 'right' in its left/right projections?

jamesmckinna:
Fixes #2471

Lots of (trivial) knock-on viscosity: PR takes the opportunity to rename/refactor proofs in (hopefully!) helpful ways.

Is this v2.x or v3.0, because `breaking`?


Taneb (approved):

MatthewDaggitt:
This is definitely v3.0 I'm afraid. Large breaking change...

JacquesCarette (approved):
v3.0 it might be, but here's my approval anyways.

jamesmckinna:
Will resolve the merge conflict only once `CHANGELOG` moves from v2.2 to v2.3... but even that is likely to be premature ;-)

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Bug-fixes
 Non-backwards compatible changes
 --------------------------------
 
+* [issue #2471](https://github.com/agda/agda-stdlib/issues/2471) In `Relation.Binary.Definitions`, the left/right order of the components of `_Respects₂_` have been swapped.
```
MatthewDaggitt, line 15: Can we provide a motivation here? Could we even label it a bug fix? This is going to be really painful for a lot of my code, and it would be good to get the community on board as much as possible by explaining the reasons for it!

MatthewDaggitt (changes requested):

CHANGELOG.md
jamesmckinna: @MatthewDaggitt thanks for the pushback on something which I had thought was previously agreed to, in principle.
By contrast with #2532 , where there is a genuine *semantic* issue about 'left' and 'right, here it is simply (to me, at least) the cognitive dissonance arising from
* the *left* (`proj₁`) component of a pair actually corresponding to a 'right'-respecting property `_Respectsʳ_`
* and *vice versa*

Now that I reflect a bit on the pushback, I *suppose* I could see it as an instance of #2532 where we have an underlying definition (of '-respecting'), whose handedness (left/right) is at odds with both *name* and *semantics*. If the definitions of being 'left'-respecting and 'right'-respecting were swapped, and the corresponding superscripts `ˡ`/`ʳ` exchanged also, then (perhaps? probably?) this PR would no longer be necessary... but the corresponding changes would themselves be as `breaking` as the ones at hand here!

... that said, I had identified this before #2532 as a 'low-hanging' issue which drew attention to a `bug` in things-as-they-are.

If you prefer, we could tag this `status:blocked-by-issue` ahead of a more profound discussion of #2532 but as I indicate above, the reconciliation envisaged here would *still* be necessary, but *mutatis mutandis*. Or at least I think so as of this morning!

jamesmckinna: TL;DR I could try to summarise the above (and earlier discussion of it) as a longer `CHANGELOG` entry, but initially the issue seemed so 'obvious' to me that I didn't quite know what else to say...

MatthewDaggitt: > @MatthewDaggitt thanks for the pushback on something which I had thought was previously agreed to, in principle.

Apologies that it came across as pushback, that wasn't the intention. I'm certainly on board and think that this is a good change! However, I just want to explain to the users in the CHANGELOG why we're doing it given that it will break code.

It can be as saying something as simple as:

> In `Relation.Binary.Definitions`, the left/right components of `_Respects₂_` have been swapped as previously the position of the `_Respectsˡ_` (respects left) was placed on the right hand side of the pair and `_Respects_` (respects right) was placed on the left hand side of the pair. By switching them the names are now consistent with their location.

jamesmckinna: Ah! Sorry to be/have been so defensive...
... and your offer of words is admirably crisp compared to my own ramblings! Will update accordingly...

jamesmckinna:
I'll only return to the merge conflicts after v2.3 gets released (or else we fast-forward directly to v3.0)

JacquesCarette (approved):
re-approving, as there were changes.

jamesmckinna:
@MatthewDaggitt it's just up to you to approve now

MatthewDaggitt (approved):

PR #2518, opened 2024-12-12
[ refactor ] Change definition of `Data.Nat.Base._≤‴_`

### LINKED ISSUE #2504: Refactor/deprecate `Data.Nat.Base._≤‴_` and its properties
@tsung-ju 's recent #2503  drew my attention to two things:
* the relation is, in fact, never used in `stdlib`, so is a good candidate for deprecation; cf. #1726 
* failing that (maybe someone out there does, in fact, use it? UPDATED: @tsung-ju in fact!), the constructor `≤‴-refl : ∀{m} → m ≤‴ m`, with its repeated occurrence of the `m`, in fact gives rise to unification problems for the typechecker under the ambient options `without-K` or `cubical-compatible`, requiring in the open PR a circumlocution via a lemma which cuts in an explicit appeal to equality...

If we opt to deprecate the relation, then all well and good as to the first point... but the second suggests a wider `library-design` issue/pattern that we might wish to more systematically deploy: avoid repeated arguments in `data` constructors... unless `with-K` is in operation?

Eg. replacing the current definition with
```agda
data _≤‴_ (m n : ℕ) : Set

_<‴_ : (m n : ℕ) → Set
m <‴ n = suc m ≤‴ n

data _≤‴_ m n where
  ≤‴-reflexive : m ≡ n → m ≤‴ n
  ≤‴-step : m <‴ n → m ≤‴ n

pattern ≤‴-refl = ≤‴-reflexive refl
```
(and hence a possible additional refactoring in terms of `Relation.Binary.Construct.Closure.Reflexive` etc.)

UPDATED: the above definition moreover admits (much?) more direct proofs of `≤‴⇒≤` and `≤⇒≤‴` (aah... but with bad complexity, cf. #2440 #2442 )
```agda
≤‴⇒≤ : m ≤‴ n → m ≤ n
≤‴⇒≤ (≤‴-reflexive eq) = ≤-reflexive eq
≤‴⇒≤ (≤‴-step m<n) = <⇒≤ (≤‴⇒≤ m<n)

s≤‴s : m ≤‴ n → suc m ≤‴ suc n
s≤‴s (≤‴-reflexive m≡n) = ≤‴-refl $ cong suc m≡n
s≤‴s (≤‴-step m<n) = ≤‴-step (s≤‴s m<n)

0≤‴n : ∀ n → 0 ≤‴ n
0≤‴n zero    = ≤‴-refl
0≤‴n (suc n) = ≤‴-step (s≤‴s (0≤‴n n))

≤⇒≤‴ : _≤_ ⇒ _≤‴_
≤⇒≤‴ z≤n       = 0≤‴n _
≤⇒≤‴ (s≤s m≤n) = s≤‴s $ ≤⇒≤‴ m≤n
```
But with cleaner proofs of
```agda
≤‴⇒≤″ : m ≤‴ n → m ≤″ n
≤‴⇒≤″  ≤‴-refl      = _ , +-identityʳ _
≤‴⇒≤″ (≤‴-step m≤n) = _ , trans (+-suc _ _) (≤″-proof (≤‴⇒≤″ m≤n))

m≤‴m+k : m + k ≡ n → m ≤‴ n
m≤‴m+k {k = zero}  = ≤‴-reflexive ∘ trans (sym (+-identityʳ _))
m≤‴m+k {k = suc _} = ≤‴-step ∘ m≤‴m+k ∘ trans (sym (+-suc _ _))
```
and from #2503 
```agda
≤‴-irrelevant : Irrelevant _≤‴_
≤‴-irrelevant (≤‴-reflexive eq₁) (≤‴-reflexive eq₂) = cong ≤‴-reflexive (≡-irrelevant eq₁ eq₂)
≤‴-irrelevant (≤‴-reflexive eq₁) (≤‴-step q)        with () ← <‴-irrefl eq₁ q
≤‴-irrelevant (≤‴-step p)        (≤‴-reflexive eq₂) with () ← <‴-irrefl eq₂ p
≤‴-irrelevant (≤‴-step p)        (≤‴-step q)        = cong ≤‴-step (≤‴-irrelevant p q)
```

jamesmckinna:
Fixes #2504 , following the discussion there and on #2503 , which this also refactors.

Tagged as `breaking`, although the only known use of this relation in the wild is from the OP of #2503 , and the `CHANGELOG` documents the only possible (?) source of `breaking` behaviour, namely supplying implicit arguments to the new pattern synonym for the former non-linear constructor `≤‴-refl`. Hence v2.2, not v3.0... but I'm happy to row back on that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Non-backwards compatible changes
     them to name the operation `+`.
   * `distribˡ` and `distribʳ` are defined in the record.
 
+* [issue #2504]() In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the chnage in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
+
 Minor improvements
 ------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ Non-backwards compatible changes
     them to name the operation `+`.
   * `distribˡ` and `distribʳ` are defined in the record.
 
-* [issue #2504]() In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the chnage in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
+* [issue #2504](https://github.com/agda/agda-stdlib/issues/2504) In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the chnage in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
 
 Minor improvements
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ Non-backwards compatible changes
     them to name the operation `+`.
   * `distribˡ` and `distribʳ` are defined in the record.
 
-* [issue #2504](https://github.com/agda/agda-stdlib/issues/2504) In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the chnage in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
+* [issue #2504](https://github.com/agda/agda-stdlib/issues/2504) In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the change in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
 
 Minor improvements
 ------------------
--- a/src/Data/Nat/Base.agda
+++ b/src/Data/Nat/Base.agda
@@ -335,7 +335,7 @@ suc n ! = suc n * n !
 -- _≤′_: this definition is more suitable for well-founded induction
 -- (see Data.Nat.Induction)
 
-infix 4 _<′_ _≤′_ _≥′_ _>′_
+infix 4 _≤′_ _<′_ _≥′_ _>′_
 
 data _≤′_ (m : ℕ) : ℕ → Set where
   ≤′-refl :                         m ≤′ m
--- a/src/Data/Nat/Base.agda
+++ b/src/Data/Nat/Base.agda
@@ -335,7 +335,7 @@ suc n ! = suc n * n !
 -- _≤′_: this definition is more suitable for well-founded induction
 -- (see Data.Nat.Induction)
 
-infix 4 _≤′_ _<′_ _≥′_ _>′_
+infix 4 _<′_ _≤′_ _≥′_ _>′_
 
 data _≤′_ (m : ℕ) : ℕ → Set where
   ≤′-refl :                         m ≤′ m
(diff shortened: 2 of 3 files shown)


jamesmckinna:
@tsung-ju Could you please check that your development is compatible with these changes?

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ Non-backwards compatible changes
     them to name the operation `+`.
   * `distribˡ` and `distribʳ` are defined in the record.
 
+* [issue #2504](https://github.com/agda/agda-stdlib/issues/2504) In `Data.Nat.Base` the definition of `_≤‴_` has been modified to make the witness to equality explicit in a new `≤‴-reflexive` constructor; a pattern synonym ≤‴-refl` has been added for backwards compatibility but NB. the chnage in parametrisation means that this pattern is *not* well-formed if the old implicit arguments `m`,`n` are supplied explicitly.
```
jamesmckinna, line 33: I'm aware of the typo, but don't wish to waste cycles pushing `CHANGELOG`-local-only commits until this has been reviewed.

src/Data/Nat/Base.agda
```diff
@@ -335,7 +335,7 @@ suc n ! = suc n * n !
 -- _≤′_: this definition is more suitable for well-founded induction
 -- (see Data.Nat.Induction)
 
-infix 4 _≤′_ _<′_ _≥′_ _>′_
+infix 4 _<′_ _≤′_ _≥′_ _>′_
```
jamesmckinna, line 338: This is spurious... don't know how it crept in! Will fix later.
Mind you, it draws attention to the fact that `_≤′_` *also* violates the 'no non-linear constructor' proposal in #2519 ...

tsung-ju:
Hi, I tried the new version, and it indeed only breaks in places where I supplied implicit argument to `≤‴-refl`. After changing those to `≤‴-reflexive {m=m} refl` and adding the necessary imports, everything worked as before.

Is it possible to add an implicit argument to `≤‴-refl`? I tried something like
```agda
pattern ≤‴-refl {m} = ≤‴-reflexive {m=m} refl
```
but that doesn’t seem to work.

jamesmckinna:
@tsung-ju thanks very much for doing a smoke test on these changes!

I, too, had problems defining a `pattern` synonym which takes an additional implicit argument; I imagined that this might be because `m` and `n` are declared as *parameters* to the relation, rather than *indices*... so, in some sense, in any context in which `≤‴-reflexive` (or `≤‴-refl`) would be a legal pattern, the values of those parameters *must already be known*. I may raise on issue on the main Agda tracker about this, but only when/if this PR gets approval to merge. I certainly found it puzzling...

In the meantime, I conjecture that *no* use of the implicit argument(s) should actually be needed, but I'd be interested to see examples where that is actually *necessary*... so I am surprised that you did seem to need them, and found the solution to be to explicitly supply them via an inlined version. If you are willing to post them, suggest that you do so on issue #2504 itself (where the design and its behaviour is discussed), rather than this PR (where those are implemented). 

In the meantime, thanks again, and for the original PR #2503 which prompted these investigations!

src/Data/Nat/Properties.agda
```diff
@@ -2240,14 +2240,11 @@ _>‴?_ = flip _<‴?_
...
-  lemma : ∀ {m n} → (e : m ≡ n) → (q : m ≤‴ n) → subst (m ≤‴_) e ≤‴-refl ≡ q
-  lemma {m} e    ≤‴-refl       = cong (λ e → subst (m ≤‴_) e ≤‴-refl) $ ≡-irrelevant e refl
-  lemma     refl (≤‴-step m<m) with () ← <‴-irrefl refl m<m
-≤‴-irrelevant (≤‴-step m<m) ≤‴-refl     with () ← <‴-irrefl refl m<m
-≤‴-irrelevant (≤‴-step p)   (≤‴-step q) = cong ≤‴-step $ ≤‴-irrelevant p q
+≤‴-irrelevant : Irrelevant _≤‴_
+≤‴-irrelevant (≤‴-reflexive eq₁) (≤‴-reflexive eq₂) = cong ≤‴-reflexive (≡-irrelevant eq₁ eq₂)
+≤‴-irrelevant (≤‴-reflexive eq₁) (≤‴-step q)        with () ← <‴-irrefl eq₁ q
+≤‴-irrelevant (≤‴-step p)        (≤‴-reflexive eq₂) with () ← <‴-irrefl eq₂ p
+≤‴-irrelevant (≤‴-step p)        (≤‴-step q)        = cong ≤‴-step (≤‴-irrelevant p q)
```
MatthewDaggitt, line 2247: Nice refactor!

MatthewDaggitt (approved):
Okay, looks reasonable.

CHANGELOG.md
jamesmckinna: Pushed.

src/Data/Nat/Base.agda
jamesmckinna: Pushed.

JacquesCarette (approved):
Seems reasonable to me as well - but not pushing 'merge', as 'seems reasonable' feels like too low a bar.

jamesmckinna:
@JacquesCarette this PR and #2523 are individual instances of refactoring along the lines of #2519 , which I took directly from your comments on #2504 .

Very happy to pause developments in this direction, but please, comment on #2519 if you have objections to it!?

JacquesCarette:
I think the changes here are fine, I'm probably uncomfortable because `Data.Nat.Base._≤‴_` shouldn't exist at all (in that module, at least), so that's clouding my vision of the local value of this change. So I'm asking for another opinion.

jamesmckinna:
Agree that long-term (post-v3.0?) we shouldn't pollute `Data.Nat.Base` with all these relations, but for v2.x...

PR #2521, opened 2024-12-15
[ add ] Lemmas about `takeWhile` and `dropWhile` to `Data.List.Unary.All.Properties`

### LINKED ISSUE #2520: 3 new proofs suggested for ``takeWhile, dropWhile``.
To  Data.List.Relation.Unary.All.Properties :                                                                                                                   
```                                                                             
all⇒takeWhile=id :                                                              
  (P? : Decidable P) → (∀ {xs} → All P xs → takeWhile P? xs ≡ xs)               
all⇒takeWhile=id _ {[]}     _            =  PE.refl                             
  all⇒takeWhile=id P? {x ∷ xs} (Px ∷ P-xs)  with P? x                           
  ... | yes _  =  cong (x ∷_) (all⇒takeWhile=id P? P-xs)                        
  ... | no ¬Px =  contradiction Px ¬Px                                          
```                                                                             
The name and implementation can be changed, as the team likes.                  
```                                                                             
all⇒dropWhile≡[] :                                                              
  (P? : Decidable P) → (∀ {xs} → All P xs → dropWhile P? xs ≡ [])               
all⇒dropWhile≡[] _ {[]} _ =  PE.refl                                            
all⇒dropWhile≡[] P? {x ∷ xs} (Px ∷ P-xs)  with P? x                             
  ... | yes _  =  all⇒dropWhile≡[] P? P-xs                                      
  ... | no ¬Px =  contradiction Px ¬Px                                          
```                                                                             
To  Data.List.Relation.Unary.Any.Properties :                                   
```                                                                             
any¬⇒dropWhile≢[] :                                                             
  (P? : Decidable P) → (∀ {xs} → Any (¬_ ∘ P) xs → dropWhile P? xs ≢ []         
```

jamesmckinna:
Fixes #2520 with implementations using `Relation.Nullary.Decidable.dec-true` to simplify the case analysis.

NB only the first two lemmas suggested in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -332,6 +332,8 @@ Additions to existing modules
 * In `Data.List.Relation.Unary.All.Properties`:
   ```agda
   all⊆concat : (xss : List (List A)) → All (Sublist._⊆ concat xss) xss
+  all⇒dropWhile≡[] : (P? : Decidable P) → All P xs → dropWhile P? xs ≡ []
+  all⇒takeWhile≗id : (P? : Decidable P) → All P xs → takeWhile P? xs ≡ xs
   ```
 
 * In `Data.List.Relation.Unary.Any.Properties`:
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -47,7 +47,7 @@ open import Relation.Binary.PropositionalEquality.Core
 open import Relation.Nullary.Reflects using (invert)
 open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 open import Relation.Nullary.Decidable
-  using (Dec; does; yes; no; _because_; ¬?; decidable-stable)
+  using (Dec; does; yes; no; _because_; ¬?; decidable-stable; dec-true)
 open import Relation.Unary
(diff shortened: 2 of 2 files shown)


MatthewDaggitt (approved):

JacquesCarette (approved):
I may yet regret 'Approving' more `rewrite`s, but oh well.

jamesmckinna:
I wonder, how can we systematise either the grounds for your 'complaint's against `rewrite`, or else grounds under which we agree it's OK to use?

Or else, systematise the use of alternatives (eg `cast`?) to `subst`, and insist on it in the `style-guide`? 

It had occurred to me that there would/should be a definable `cast` for `All`... but my own idiomatic use of Agda does use `rewrite` (often in conjunction with `with`), so this issue may well recur in subsequent PRs...

JacquesCarette:
Grounds for complaint: I find `rewrite` quite fragile, and debugging a broken proof that uses `rewrite` well nigh impossible. Was reminded of that recently with me tweaking things in PLFA (not its fault). `subst` is equally bad, in entirely different ways.

Certainly we should outlaw `rewrite` outside of `Properties` (or morally `Properties`) modules, for the same reason that using `with` to do a non-degenerate case-split is a real pain to work with. In `Properties`, as you have it here, it does make some proofs quite a bit more succinct. And, usually, I like succinct (within limits, of course).

PR #2513, opened 2024-12-08
Add `all⊆concat`

jamesmckinna:
Fixes #2509

NB. Tweaks the `import`s in order to have clean separation of `Sublist._⊆_` and its properties from `Subset._⊆_`.

Issue #2509, opened 2024-12-04
suggested  "`concat` members are sublists"


mechvel:
I suggest for standard library a proof for that each member of ``(concat xss)``
is a sublist in the result:
```
import      Data.List.Relation.Binary.Sublist.Setoid.Properties
            as SublistProps
open import Data.List.Relation.Binary.Sublist.Propositional using ()
            renaming (_⊆_ to _⊆ₗ_; ⊆-refl to ⊆ₗrefl)
S = PE.setoid A  -- the propositional setoid

all⊆ₗconcat : (xss : List (List A)) → All (_⊆ₗ (concat xss)) xss
all⊆ₗconcat [] = []
all⊆ₗconcat (xs ∷ xss) =
  xs⊆ₗunion ∷ (AllM.map ⊆ₗunion'⇒⊆ₗunion all-xss⊆ₗunion')
  where
  union' = concat xss
  union = xs ++ union'
  xs⊆ₗunion = SublistProp.++⁺ʳ S {xs}{xs} union' ⊆ₗrefl
  all-xss⊆ₗunion' = all⊆ₗconcat xss
  ⊆ₗunion'⇒⊆ₗunion = \{ys} → SublistProps.++⁺ˡ S {ys}{union'} xs
```
The team can set a nicer implementation if any.

jamesmckinna:
I think you want something a bit simpler:
```agda
import Data.List.Relation.Binary.Sublist.Propositional as Sublist
import Data.List.Relation.Binary.Sublist.Propositional.Properties as Sublist
all⊆concat : (xss : List (List A)) → All (Sublist._⊆ (concat xss)) xss
all⊆concat [] = []
all⊆concat (xs ∷ xss) =
  Sublist.++⁺ʳ (concat xss) Sublist.⊆-refl ∷ All.map (Sublist.++⁺ˡ xs) (all⊆concat xss)
```

It's possible that further simplification is possible?

mechvel:
James wrote 
> It's possible that further simplification is possible?

You write in one line by composing of  what I wrote in seven lines.
To my mind, the reader would understand these seven lines more easily.
As to me, this difference is not important.
Let the team chose the variant as it likes.

jamesmckinna:
Sergei, thanks for the feedback. 
[...]

PR #2522, opened 2024-12-15
Deprecate `Data.List.Relation.Unary.All.Properties.takeWhile⁻`

jamesmckinna:
This is a coda to #2520 / #2521 : the type of `takeWhile⁻` is insufficiently general, and generalising it makes the proof trivially delegate to that of `all-takeWhile`; another instance of 'non-linear bindings considered harmful' ...

NB. UPDATED: following @MatthewDaggitt 's suggestions below, can now ignore first three items here
* not sure how to record this in `CHANGELOG`: under `Minor improvements`? 
* possible deprecation of `takeWhile⁻` now possible? or else of `all-takeWhile`? 
* not `breaking`, but not clear whether the generalisation may lead to `unsolved metas` errors downstream?
* any other similar refactorings possible?

### DIFF
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -43,7 +43,7 @@ open import Relation.Binary.Core using (REL)
 open import Relation.Binary.Bundles using (Setoid)
 import Relation.Binary.Definitions as B
 open import Relation.Binary.PropositionalEquality.Core
-  using (_≡_; refl; sym; cong; cong₂; _≗_)
+  using (_≡_; refl; sym; trans; cong; cong₂; _≗_)
 open import Relation.Nullary.Reflects using (invert)
 open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 open import Relation.Nullary.Decidable
@@ -444,15 +444,21 @@ drop⁺ (suc n) (px ∷ pxs) = drop⁺ n pxs
 
 dropWhile⁺ : (Q? : Decidable Q) → All P xs → All P (dropWhile Q? xs)
 dropWhile⁺               Q? []         = []
-dropWhile⁺ {xs = x ∷ xs} Q? (px ∷ pxs) with does (Q? x)
+dropWhile⁺ {xs = x ∷ xs} Q? px∷pxs@(_ ∷ pxs) with does (Q? x)
 ... | true  = dropWhile⁺ Q? pxs
-... | false = px ∷ pxs
+... | false = px∷pxs
 
-dropWhile⁻ : (P? : Decidable P) → dropWhile P? xs ≡ [] → All P xs
-dropWhile⁻ {xs = []}     P? eq = []
-dropWhile⁻ {xs = x ∷ xs} P? eq with P? x
-... | yes px = px ∷ (dropWhile⁻ P? eq)
-... | no ¬px = case eq of λ ()
+module _ (P? : Decidable P) where
+
+  dropWhile++⁻ : ∀ xs {ys} → dropWhile P? (xs ++ ys) ≡ ys → All P xs
+  dropWhile++⁻ []       eq = []
+  dropWhile++⁻ (x ∷ xs) eq with P? x
+  ... | yes px = px ∷ dropWhile++⁻ xs eq
+  ... | no  _  = case List.++-cancelʳ _ _ [] eq of λ ()
+
+  dropWhile⁻ : dropWhile P? xs ≡ [] → All P xs
+  dropWhile⁻ {xs = xs}
+    with prf ← dropWhile++⁻ xs {ys = []} rewrite List.++-identityʳ xs = prf
 
 all-head-dropWhile : (P? : Decidable P) →
                      ∀ xs → Maybe.All (∁ P) (head (dropWhile P? xs))
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -49,7 +49,7 @@ open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 open import Relation.Nullary.Decidable
   using (Dec; does; yes; no; _because_; ¬?; decidable-stable)
 open import Relation.Unary
-  using (Decidable; Pred; Universal; ∁; _∩_; _⟨×⟩_) renaming (_⊆_ to _⋐_)
+  using (Decidable; Pred; ∁; _∩_; _⟨×⟩_) renaming (_⊆_ to _⋐_)
 open import Relation.Unary.Properties using (∁?)
 
 private
@@ -466,17 +466,20 @@ take⁺ zero    pxs        = []
 take⁺ (suc n) []         = []
 take⁺ (suc n) (px ∷ pxs) = px ∷ take⁺ n pxs
 
+takeWhileP⇒Q⇒R⁺ : (∀ {x} → P x → Q x → R x) → (Q? : Decidable Q) →
+                  All P xs → All R (takeWhile Q? xs)
+takeWhileP⇒Q⇒R⁺ {P = P} {Q = Q} {R = R} P⇒Q⇒R Q? = go where
+  go : All P xs → All R (takeWhile Q? xs)
+  go [] = []
+  go {xs = x ∷ xs} (px ∷ pxs) with Q? x
+  ... | yes qx = P⇒Q⇒R px qx ∷ go pxs
+  ... | no _ = []
+
 takeWhile⁺ : (Q? : Decidable Q) → All P xs → All P (takeWhile Q? xs)
-takeWhile⁺               Q? []         = []
-takeWhile⁺ {xs = x ∷ xs} Q? (px ∷ pxs) with does (Q? x)
-... | true  = px ∷ takeWhile⁺ Q? pxs
-... | false = []
+takeWhile⁺ = takeWhileP⇒Q⇒R⁺ λ p _ → p
 
 all-takeWhile : (P? : Decidable P) → ∀ xs → All P (takeWhile P? xs)
-all-takeWhile P? []       = []
-all-takeWhile P? (x ∷ xs) with P? x
-... | yes px = px ∷ all-takeWhile P? xs
-... | no  _ = []
+all-takeWhile P? = takeWhileP⇒Q⇒R⁺ (λ _ q → q) P? ∘ All.universal-U
 
 ------------------------------------------------------------------------
 -- applyUpTo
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
To be honest I would recommend just deprecating `takeWhile⁻` entirely, rather than make this change at all. 

1. It's a trivial generalisation of `all-takeWhile`
2. It's really not the correct name for that lemma under the style guide, as it doesn't take in an argument of the form `All P (takeWhile ...)`, 

> not sure how to record this in CHANGELOG: under Minor improvements?

If we were keeping it I would explicitly say what the generalisation is under "Minor improvements". I think there's probably no need to mention the delegation of the proof, as that's not something that is going to affect users?

> possible deprecation of takeWhile⁻ now possible? or else of all-takeWhile?

See above.

> not breaking, but not clear whether the generalisation may lead to unsolved metas errors downstream?

I agree this could lead to unsolved metas. But irrelevant if we just deprecate it :smile:

jamesmckinna:
UPDATED: now opt for deprecation

Of these only the last point now stands further consideration: 
* type of `dropWhile⁻` is similarly not consistent with `style-guide`
* types of `dropWhile⁺` and `dropWhile⁺` are insufficiently general: without any relationship between `P` and `Q`, these are map properties of `All P` wrt the sublist/prefix relation
* if generalised appropriately, then at least the existing `all-takeWhile` could even be deprecated as well.

Accordingly will convert to DRAFT while I figure this out.

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -466,23 +466,20 @@ take⁺ zero    pxs        = []
...
+  ... | yes qx = P⇒Q⇒R px qx ∷ go pxs
+  ... | no _ = []
 
-takeWhile⁻ : (P? : Decidable P) → takeWhile P? xs ≡ xs → All P xs
-takeWhile⁻ {xs = []}     P? eq = []
-takeWhile⁻ {xs = x ∷ xs} P? eq with P? x
-... | yes px = px ∷ takeWhile⁻ P? (List.∷-injectiveʳ eq)
-... | no ¬px = case eq of λ ()
+takeWhile⁺ : (Q? : Decidable Q) → All P xs → All P (takeWhile Q? xs)
+takeWhile⁺ = takeWhileP⇒Q⇒R⁺ λ p _ → p
```
jamesmckinna, line 479: Now a possible deprecation opportunity?

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -466,23 +466,20 @@ take⁺ zero    pxs        = []
...
 
 all-takeWhile : (P? : Decidable P) → ∀ xs → All P (takeWhile P? xs)
-all-takeWhile P? []       = []
-all-takeWhile P? (x ∷ xs) with P? x
-... | yes px = px ∷ all-takeWhile P? xs
-... | no ¬px = []
+all-takeWhile P? = takeWhileP⇒Q⇒R⁺ (λ _ q → q) P? ∘ All.universal-U
```
jamesmckinna, line 482: Ditto.?

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -444,15 +444,21 @@ drop⁺ (suc n) (px ∷ pxs) = drop⁺ n pxs
 
 dropWhile⁺ : (Q? : Decidable Q) → All P xs → All P (dropWhile Q? xs)
```
jamesmckinna, line 445: I can't quite see what a judicious generalisation of this should be from which this would follow by analogy with `takeWhile⁺` below!?

JacquesCarette: Should this be a corollary of something about `All P xs` implies `All P ys` whenever ys is a subset of xs?

jamesmckinna: Indeed. 

As with #2509 / #2513 I'm worried that this needs much more careful consideration of the complicated interlocking dependencies between `All`/`Any` and `Sublist`/`Subset`, and in this case, I'd been after something closer to the `takeWhileP⇒Q⇒R⁺` introduced by this PR.

The 'problem' is that `take`/`drop` are really concerned with `Prefix`/`Suffix`... so perhaps a larger scale refactoring might be necessary later?

jamesmckinna:
If there was a button "Convert PR to an issue", I'd hit it, under a title like "overhaul theory of `takeWhile` and `dropWhile` for `All`", because it feels as though this is really a patch on something which is actually broken (wrt our `style-guide` precepts, *and* to the 'right' level of abstraction/generalisation), and needs a more serious fix?

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -466,23 +472,20 @@ take⁺ zero    pxs        = []
 take⁺ (suc n) []         = []
 take⁺ (suc n) (px ∷ pxs) = px ∷ take⁺ n pxs
 
-takeWhile⁺ : (Q? : Decidable Q) → All P xs → All P (takeWhile Q? xs)
-takeWhile⁺               Q? []         = []
-takeWhile⁺ {xs = x ∷ xs} Q? (px ∷ pxs) with does (Q? x)
-... | true  = px ∷ takeWhile⁺ Q? pxs
-... | false = []
+takeWhileP⇒Q⇒R⁺ : (∀ {x} → P x → Q x → R x) → (Q? : Decidable Q) →
+                  All P xs → All R (takeWhile Q? xs)
```
gallais, line 476: Isn't this a a combination of `All Q (takeWhile Q? xs)`, the old `takeWhile⁺`  and `zipWith`?
It feels like we're dropping a separation of concerns in favour of a swiss knife approach.

src/Data/List/Relation/Unary/All/Properties.agda
JacquesCarette: Well, `take` / `drop` don't really respect `Prefix` / `Suffix`, do they? `takeWhile` and `dropWhile` do. And so for them, it does seem to make sense to think more deeply about what the whole structure should be.

src/Data/List/Relation/Unary/All/Properties.agda
JacquesCarette: It does rather feel like a composition of things, doesn't it?

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: Hmm. Both `take`/`drop` (provided the argument `n` is within bounds...;-)) and `takeWhile`/`dropWhile` enjoy a characteristic property that their append yields the original list?
But agree: more thought required, or else actually abandoning this PR!

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: Thanks very much for these observations: I hadn't realised how much of a Swiss-Army-type person I can be! Very helpful to have it pointed out. I think this is further support for an abandon.

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: > Should this be a corollary of something about `All P xs` implies `All P ys` whenever ys is a subset of xs?

Indeed, and this appears *twice* in the library:
* `Data.List.Relation.Unary.All.Properties.anti-mono : xs Subset.⊆ ys → All P ys → All P xs` (and `Sublist` is a stronger property than `Subset`)
* `Data.List.Relation.Binary.Sublist.Propositional.Properties.All-resp-⊆ : {P : Pred A ℓ} → (All P) Respects _⊇_` and this should be refactored to a proof in `...Sublist.Setoid.Properties`, under assumptions that `P` respects the underlying `Setoid` equality?

MatthewDaggitt (approved):

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: Followups on #2525 ...

jamesmckinna:
After a bit more detective work...
`Data.List.Relation.Binary.Sublist.Setoid.Properties.takeWhile-⊆ : ∀ xs → takeWhile P? xs ⊆ xs`
so @JacquesCarette 's comment above has extra force... should we move *all* the properties of `*While` wrt `All` out of `All.Properties` altogether?

MatthewDaggitt:
> should we move all the properties of *While wrt All out of All.Properties altogether? (not withstanding... ahem, the recently merged https://github.com/agda/agda-stdlib/pull/2521 ...;-))

Sorry, I'm afraid I don't quite understand why we would do this and where would we put them?

jamesmckinna:
@MatthewDaggitt I guess I'd been thinking, after @JacquesCarette 's comment:
* these properties are consequences of a general `Sublist`/`Subset` property of `All` (discussion elsewhere about this), so perhaps should better be rephrased in terms of such;
* the general drift of other recent discussion (with you) has been to avoid dragging additional dependency on `Sublist`/`Subset` into the `import`s of `All.Properties`
* ...

But it doesn't affect this PR, so happy to punt this discussion downstream. HTH.

MatthewDaggitt:
> As to where we'd put them, probably in Sublist.Properties where we already have takeWhile-⊆ and dropWhile-⊆ as consequences of the corresponding Heterogeneous versions...?

Personally it would be very weird for me for `All` properties to be found under `Sublist`... but yes, happy to move it to a downstream discussion

JacquesCarette (approved):
The changes in this PR make sense to me.

The discussion, less so. I'm having a hard time extracting exactly what is being proposed from the thread.

jamesmckinna:
@JacquesCarette writes:
> The discussion, less so. I'm having a hard time extracting exactly what is being proposed from the thread.

At present: nothing beyond what is in the current commits, which I don't plan to add to.

In future, the 'issue(s)' identified above may broadly be covered by [ DRY ]:
* should  properties of `*While` wrt `All` and `Any` be proved directly, as now (and hence be located under `Properties` of `Data.List.Relation.Unary.{All|Any}`
* or else as consequences of a general functoriality (co- for `Any`, contra- for `All`) wrt `Subset`/`Sublist`

Discussion:
* the first alternative is (basically) the *status quo ante*, so nothing as such is proposed (and indeed, for this PR, nothing further is towards these downstream issues)
* the second flows from your comment about functoriality (see #2525 for separate discussion, lifted out from here), but potentially carries the additional cost of a more complex dependency graph, or else things living in not-quite-the-right place, or refactoring to make the functoriality properties `Core`, so that their `import` can be uncoupled from the various other 'right places' where things do/might/should live.

Followup discussion to #2525 ?

JacquesCarette:
Right - so the discussion should be moved to an issue, as this is ready to be merged, which I will do now.

Issue #2519, opened 2024-12-12
`data` constructors should *not* be non-linear unless defined within the scope of `with-K`


jamesmckinna:
See the discussion on #2504 for a specific instance. 

This raises the issue to library level, given the ambient `stdlib` assumption of `cubical-compatible`/`without-K`.

`data` constructors which exploit non-linearity are making *implicit* appeal to reflexivity of (propositional) equality, and that should instead be required as an *explicit* witness. Specialising the constructor to supply `refl` as a default witness should be the preferred approach to offering a non-linear constructor. Cf. `Data.Nat.Divisibility.Core` #2013

MatthewDaggitt:
I agree this is a good principle going forward. However, how many definitions fall foul of this currently? Depending on the numbers I would be weary of how big a breaking change this would be to change all the existing definitions?

jamesmckinna:
I haven't done a survey of the relevant definitions, besides `Data.Nat.Base._≤′_`, which seems to fall even more easily to the corresponding refactoring as #2504 .

JacquesCarette:
I agree with the design principle.

My gut feel is that this isn't likely to occur very often. But this is rather hard to search for! So piecemeal refactoring is indeed a rather reasonable way to go.

Taneb:
Can you clarify what exactly you mean by "non-linear" here? Is it that the same variable should not appear in two different indices of a data constructor?

jamesmckinna:
> Can you clarify what exactly you mean by "non-linear" here? Is it that the same variable should not appear in two different indices of a data constructor?

Yes, exactly. I'll change the title of the issue.

In the originating PR #2504 it's actually even stronger than that, because it meant that those indices could even become *parameters* to the `data` definition.

jamesmckinna:
Looping back to #1344 ... where @MatthewDaggitt had the same idea, but with different motivations, in [this comment](https://github.com/agda/agda-stdlib/issues/1344#issuecomment-768686234).

Issue #2442, opened 2024-07-26
Linear-time implementation of `Data.Nat.Properties.≤′⇒≤`?


jamesmckinna:
In a [recent Zulip thread](https://agda.zulipchat.com/#narrow/stream/264623-stdlib/topic/Data.2ENat.2EProperties.2E.3C.E2.80.B2.E2.87.92.3C.20unnecessarily.20quadratic.3F) the question was raised as to whether a linear-time implementation of the coercion from `_≤′_` to `_≤_` was possible. After some experiments (and not going via `recompute`), I wondered about the following:
``agda
m≤′n⇒m≤o+n′ : m ≤′ n → ∀ {o p} → .(o + n ≡ p) → m ≤ p
m≤′n⇒m≤o+n′ ≤′-refl {o = o} eq = cast-≤ʳ (trans (+-comm _ o) eq) (m≤m+n _ o)
m≤′n⇒m≤o+n′ (≤′-step m≤′n) {o = o} eq = m≤′n⇒m≤o+n′ m≤′n {o = suc o} (trans (sym (+-suc o _)) eq)

m≤′n⇒m≤o+n : m ≤′ n → ∀ o → m ≤ o + n
m≤′n⇒m≤o+n m≤′n o = m≤′n⇒m≤o+n′ m≤′n refl

≤′⇒≤ : _≤′_ ⇒ _≤_
≤′⇒≤ m≤′n = m≤′n⇒m≤o+n m≤′n 0
```
with (independently useful... for replacing uses of `subst` when dealing with `_≤_`...?) linear-time coercions given by:
```agda
cast-≤ˡ : .(m ≡ o) → m ≤ n → o ≤ n
cast-≤ˡ {o = zero}  _  z≤n       = z≤n
cast-≤ˡ {o = suc _} eq (s≤s m≤n) = s≤s (cast-≤ˡ (cong pred eq) m≤n)

cast-≤ʳ : .(n ≡ o) → m ≤ n → m ≤ o
cast-≤ʳ             _  z≤n       = z≤n
cast-≤ʳ {o = suc _} eq (s≤s m≤n) = s≤s (cast-≤ʳ (cong pred eq) m≤n)
```

More generally, should we (now start to) reconsider `Data.Nat.Properties` from the perspective of making its *proofs* (as well as the definitions in `Data.Nat.Base`) more computationally efficient?

NB further optimisation may be possible in the base case
```agda
[...]

PR #2523, opened 2024-12-16
[ refactor ] Change definition of `Data.Nat.Base._≤′_`

jamesmckinna:
This is exactly analogous to #2504 ,  in line with #2519 .
Potentially `breaking`, but all tests pass on `stdlib` without changes.
Possible downstream improvement for tackling #2442 ?

### DIFF
--- a/src/Data/Nat/Base.agda
+++ b/src/Data/Nat/Base.agda
@@ -338,8 +338,10 @@ suc n ! = suc n * n !
 infix 4 _≤′_ _<′_ _≥′_ _>′_
 
 data _≤′_ (m : ℕ) : ℕ → Set where
-  ≤′-refl :                         m ≤′ m
-  ≤′-step : ∀ {n} (m≤′n : m ≤′ n) → m ≤′ suc n
+  ≤′-reflexive : ∀ {n} → m ≡ n → m ≤′ n
+  ≤′-step : ∀ {n} → m ≤′ n → m ≤′ suc n
+
+pattern ≤′-refl = ≤′-reflexive refl
 
 _<′_ : Rel ℕ 0ℓ
 m <′ n = suc m ≤′ n
--- a/src/Data/Nat/Base.agda
+++ b/src/Data/Nat/Base.agda
@@ -341,7 +341,7 @@ data _≤′_ (m : ℕ) : ℕ → Set where
   ≤′-reflexive : ∀ {n} → m ≡ n → m ≤′ n
   ≤′-step : ∀ {n} → m ≤′ n → m ≤′ suc n
 
-pattern ≤′-refl = ≤′-reflexive refl
+pattern ≤′-refl {m} = ≤′-reflexive {n = m} refl
 
 _<′_ : Rel ℕ 0ℓ
 m <′ n = suc m ≤′ n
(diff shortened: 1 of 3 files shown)


JacquesCarette (approved):
This seems very reasonable to me.

MatthewDaggitt (approved):

jamesmckinna:
Re: `breaking`/`v3.0` I guess you're right @MatthewDaggitt although as with #2504 I'm not clear whether this relation has any external use; internally it exists as a mediator for `Data.Nat.Induction`, moreover one which has been previously refactored extensively so that the single proof involved `<⇒<′` already goes via a `pattern` synonym `<′-base`... which in fact is (equivalent to) the one introduced by this PR...

... but I should consider uncoupling the last commit to `CHANGELOG`, as this was made on the assumption that the two PRs would be merged under v2.2...

MatthewDaggitt:
I havered on this. I mean the previous change was also similarly breaking and got merged. I think the rule we seem to be converging on is that if we break implicit arguments then it is fine?

jamesmckinna:
> I havered on this. 

Fair enough! This relation *is* actually used in `stdlib` (but, as usual, we can't speak for external client uses), so I took care to investigate *how*, and hence whether the consequences of this breaking change would cash out adversely. But I concluded: not.

> I mean the previous change was also similarly breaking and got merged. I think the rule we seem to be converging on is that if we break implicit arguments then it is fine?

That's as may be, but AFAIAC, this is purely in the spirit of #2519 : ie. make the `reflexivity` constructor take a proof of equality explicitly. The implicit/explicit distinction is moot as far as this PR is concerned, I think?

src/Data/Nat/Base.agda
```diff
@@ -338,8 +338,10 @@ suc n ! = suc n * n !
 infix 4 _≤′_ _<′_ _≥′_ _>′_
 
 data _≤′_ (m : ℕ) : ℕ → Set where
-  ≤′-refl :                         m ≤′ m
-  ≤′-step : ∀ {n} (m≤′n : m ≤′ n) → m ≤′ suc n
+  ≤′-reflexive : ∀ {n} → m ≡ n → m ≤′ n
+  ≤′-step : ∀ {n} → m ≤′ n → m ≤′ suc n
+
+pattern ≤′-refl = ≤′-reflexive refl
```
jamesmckinna, line 344:
```suggestion
pattern ≤′-refl {m} = ≤′-reflexive {n = m} refl
```

would, I (now) think, be a *non*-`breaking` change.

jamesmckinna:
The argument for making a `breaking` change in #2504 was that there were *no* uses of it in the library, and only one external user (whose PR prompted this whole cycle of refactoring), who could absorb the change without difficulty. 

But hopefully the non-breaking suggestion that I make above (which I conjecture is even equivalent to the previous version, because of the way unification works; it might only fail if `{n = m}` were ever to be supplied explicitly) mops up any 'accidnetal' breakges...?

JacquesCarette:
These suggestions make sense to me.

src/Data/Nat/Base.agda
MatthewDaggitt: If pattern synonyms support that, then yes, let's do it!

MatthewDaggitt:
Yup, upon reflection happy to merge this either way, but even better if we can make it non-breaking.

PR #2524, opened 2024-12-17
[ refactor ] Move `Data.List.Relation.Unary.All.Properties.all⊆concat` to `Data.List.Relation.Binary.Sublist.Setoid.Properties`

### LINKED ISSUE #2517: [ refactor ] Possible refactoring of #2509 / #2513 ?
This issue may be regarded (and hence instantly dismissed) as a `status:duplicate` of #2509... but posted for 'documentation'/alternative implementation purposes.

I'm happy with the merged #2513 , fixing #2509 , and so too is the author of the OP. Thanks to @MatthewDaggitt for the speedy turnround time!

That said, a brain-worm has been worrying away at me as to whether a pattern-matching-free solution was available (*slightly* better abstraction wrt presentation/implementation of the `Sublist` relation...?), and it seems there is: relative to the existing solution in `Data.List.Relation.Unary.All.Properties`, we can refactor as follows
```
all⊆concat : (xss : List (List A)) → All (Sublist._⊆ (concat xss)) xss
all⊆concat {A = A} xss = All.tabulate xs∈xss⇒xs⊆concat[xss] where
  xs∈xss⇒xs⊆concat[xss] : xs ∈ xss → xs Sublist.⊆ concat xss
  xs∈xss⇒xs⊆concat[xss] {xs = xs} xs∈xss
    with prf ← Sublist.concat⁺ (Sublist.map Sublist.⊆-reflexive (Sublist.from∈ xs∈xss))
    rewrite List.++-identityʳ xs
    = prf
```
(and further simplification may be even be possible *in the definition of `prf` itself*? The `⊆-reflexive` step is to pass from a rather simple relation via `map` to the more complex iterated `Sublist (Sublist R)` ... so this seems to have the flavour of the triangular identities for `Sublist` as a suitable 'monad' on relations... but I can't quite see the details yet)

This suggests that we might, instead, refactor by lifting out the above `where`-bound lemma into a self-standing one in `Data.List.Relation.Binary.Sublist.Propositional.Properties` (maybe even in `Setoid`? but then the proposed refactoring seems to stumble on the parameterisation on the `Setoid`, and the lemma shifts from `A` to `List A`... :-(, so perhaps the answer is to pull back all the way to `Heterogeneous.Properties`?) of the form
```agda
xs∈xss⇒xs⊆concat[xss] : xs ∈ xss → xs ⊆ concat xss
xs∈xss⇒xs⊆concat[xss] {xs = xs} xs∈xss
  with prf ← concat⁺ (Sublist.map ⊆-reflexive (from∈ xs∈xss))
  rewrite List.++-identityʳ xs
  = prf
```

leaving the final implementation (under `Data.List.Relation.Unary.All.Properties`? or should this move as well?) simply as:
```agda
all⊆concat : (xss : List (List A)) → All (Sublist._⊆ (concat xss)) xss
all⊆concat {A = A} xss = All.tabulate Sublist.xs∈xss⇒xs⊆concat[xss]
```

jamesmckinna:
Fixes #2517 following discussion there. 
Re-exports to `Data.List.Relation.Binary.Sublist.Propositional.Properties` automatically, so this strictly generalises #2513 as a solution to #2509 .

### DIFF
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -46,7 +46,7 @@ open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 open import Relation.Nullary.Decidable
   using (Dec; does; yes; no; _because_; ¬?; decidable-stable; dec-true)
 open import Relation.Unary
-  using (Decidable; Pred; Universal; ∁; _∩_; _⟨×⟩_) renaming (_⊆_ to _⋐_)
+  using (Decidable; Pred; ∁; _⟨×⟩_) renaming (_⊆_ to _⋐_)
 open import Relation.Unary.Properties using (∁?)
 
 private
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -384,12 +384,7 @@ concat⁺ (pxs ∷ pxss) = ++⁺ pxs (concat⁺ pxss)
 concat⁻ : ∀ {xss} → All P (concat xss) → All (All P) xss
 concat⁻ {xss = []}       []  = []
 concat⁻ {xss = xs ∷ xss} pxs = ++⁻ˡ xs pxs ∷ concat⁻ (++⁻ʳ xs pxs)
-{-
-all⊆concat : (xss : List (List A)) → All (Sublist._⊆ concat xss) xss
-all⊆concat [] = []
-all⊆concat (xs ∷ xss) =
-  Sublist.++⁺ʳ (concat xss) Sublist.⊆-refl ∷ All.map (Sublist.++⁺ˡ xs) (all⊆concat xss)
--}
+
 ------------------------------------------------------------------------
 -- snoc
 
--- a/src/Data/List/Relation/Unary/All/Properties.agda
+++ b/src/Data/List/Relation/Unary/All/Properties.agda
@@ -20,9 +20,6 @@ import Data.List.Membership.Setoid as SetoidMembership
 import Data.List.Properties as List
 import Data.List.Relation.Binary.Equality.Setoid as ≋
 open import Data.List.Relation.Binary.Pointwise.Base using (Pointwise; []; _∷_)
-import Data.List.Relation.Binary.Sublist.Propositional as Sublist
-import Data.List.Relation.Binary.Sublist.Propositional.Properties
-  as Sublist
 import Data.List.Relation.Binary.Subset.Propositional as Subset
 open import Data.List.Relation.Unary.All as All using
   ( All; []; _∷_; lookup; updateAt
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):
Makes me wonder if we have equational proof combinators for ⊆ reasoning.

jamesmckinna:
> Makes me wonder if we have equational proof combinators for ⊆ reasoning.

We do, AFAIK, but would they help here? 

Oh, I suppose you could eliminate my go-to pattern of `with ... rewrite ...`. Hmm, I could try that... but I'm still thinking about your comments elsewhere about the fragility of `rewrite`, and would like to see some concrete examples of it working, then breaking. Plenty of examples of it not working, but they seem legitimately to arise from "I'm not sure there is a case for `refl` when attempting to unify..." diagnostics under `cubical-compatible`/`without-K`, which I guess we have to bite the bullet on...

JacquesCarette:
Will try to collect examples of fragility. My memory is that it mostly arises in two cases: 1) Agda upgrades, 2) (what ought to be) minor refactorings. `rewrite` is extremely syntatic, and so prone to fail whenever any kind of minor disturbance happens. Of course, some equational proofs also fail in that way! The biggest difference is often `rewrite` is used to fix up a type, often enough an implicit one, and so why it's failing can be extremely hard to figure out.

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -388,11 +385,6 @@ concat⁻ : ∀ {xss} → All P (concat xss) → All (All P) xss
 concat⁻ {xss = []}       []  = []
 concat⁻ {xss = xs ∷ xss} pxs = ++⁻ˡ xs pxs ∷ concat⁻ (++⁻ʳ xs pxs)
 
-all⊆concat : (xss : List (List A)) → All (Sublist._⊆ concat xss) xss
-all⊆concat [] = []
-all⊆concat (xs ∷ xss) =
-  Sublist.++⁺ʳ (concat xss) Sublist.⊆-refl ∷ All.map (Sublist.++⁺ˡ xs) (all⊆concat xss)
```
MatthewDaggitt, line 394: Is this a breaking change?

MatthewDaggitt:
I'm with @JacquesCarette on this one. It is also my experience that rewrite proofs tend to be very fragile, and refactoring proofs that use extensive rewriting is extremely painful. In particular, it's very non-obvious which and how many terms in the goal are being rewritten. 

To be honest, I would be happy with a blanket policy of no rewrite proofs in the library, but I realise that might be a bit extreme.

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: It's not clear how this could be, when this lemma was only introduced last week!
I thikn that's why I was keen to do this refactoring as soon as possible, to rectify what I think  is the mistake we made by merging #2513 as a solution to #2509 without thinking more about where this lemma should live...

jamesmckinna:
I'll fix up the `rewrite` step...

jamesmckinna:
> Makes me wonder if we have equational proof combinators for ⊆ reasoning.

It turns out I was wrong. There *is* a module `⊆-Reasoning` for *`Subset`*, but not, it seems, for `Sublist`. Sigh.

src/Data/List/Relation/Unary/All/Properties.agda
MatthewDaggitt: Ah sorry! I didn't realise it was introduced last week! Ignore me.

MatthewDaggitt (approved):

jamesmckinna:
Given 2 approvals, are you each happy for me *also* to fold in tightening the `import`s as identified in #2533 ?

JacquesCarette (approved):

JacquesCarette:
I'm always pro tighter imports.

JacquesCarette (approved):

PR #2528, opened 2024-12-19
Add two lemmas on _∸_ and _≤_ in `Data.Nat.Properties`

Necried:
Added two lemmas on _∸_ and _≤_ . I'm unsure if I got the naming conventions of the lemmas correct, suggestions are welcome!

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1589,6 +1589,19 @@ m≤n⇒n∸m≤n : m ≤ n → n ∸ m ≤ n
 m≤n⇒n∸m≤n z≤n       = ≤-refl
 m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
 
+m≤n⇒1+[n∸m]≡[1+n]∸m : m ≤ n → suc (n ∸ m) ≡ suc n ∸ m
+m≤n⇒1+[n∸m]≡[1+n]∸m {zero} {zero} m≤n = refl
+m≤n⇒1+[n∸m]≡[1+n]∸m {zero} {suc n} m≤n = refl
+m≤n⇒1+[n∸m]≡[1+n]∸m {suc m} {suc n} (s≤s m≤n) = m≤n⇒1+[n∸m]≡[1+n]∸m m≤n
+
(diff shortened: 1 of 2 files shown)


src/Data/Nat/Properties.agda
```diff
@@ -1589,6 +1589,19 @@ m≤n⇒n∸m≤n : m ≤ n → n ∸ m ≤ n
 m≤n⇒n∸m≤n z≤n       = ≤-refl
 m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
 
+m≤n⇒1+[n∸m]≡[1+n]∸m : m ≤ n → suc (n ∸ m) ≡ suc n ∸ m
+m≤n⇒1+[n∸m]≡[1+n]∸m {zero} {zero} m≤n         = refl
+m≤n⇒1+[n∸m]≡[1+n]∸m {zero} {suc n} m≤n        = refl
+m≤n⇒1+[n∸m]≡[1+n]∸m {suc m} {suc n} (s≤s m≤n) = m≤n⇒1+[n∸m]≡[1+n]∸m m≤n
+
+m≤n⇒n∸[n∸m]≡m : m ≤ n → n ∸ (n ∸ m) ≡ m
```
jamesmckinna, line 1597: How is this different from L1661
```agda
m∸[m∸n]≡n : ∀ {m n} → n ≤ m → m ∸ (m ∸ n) ≡ n
```

src/Data/Nat/Properties.agda
```diff
@@ -1589,6 +1589,19 @@ m≤n⇒n∸m≤n : m ≤ n → n ∸ m ≤ n
 m≤n⇒n∸m≤n z≤n       = ≤-refl
 m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
 
+m≤n⇒1+[n∸m]≡[1+n]∸m : m ≤ n → suc (n ∸ m) ≡ suc n ∸ m
```
jamesmckinna, line 1592: Is this not equivalent to `+-∸-assoc 1`?

jamesmckinna:
I believe that this PR duplicates proofs already present in `Data.Nat.Properties`... so perhaps another failure of discoverability.

Necried:
You're right! I wasn't looking at the correct places when using this module.

PR #2530, opened 2024-12-20
[ add ] `concat [ xs ] ≡  xs`

jamesmckinna:
I'm persuaded by the [discussion on Zulip](https://agda.zulipchat.com/#narrow/channel/264623-stdlib/topic/.60concat-single.60.20.3F) (and my own thoughtlessness in responding there) that this is a simple worthwhile addition, with its potential downstream use in #2524 (and elsewhere). Thanks to @ncfavier for clarifying the triangle laws for the `List` monad!

Possible issue(s):
* naming (as ever); alternatives `concat∘[_]≗id`? (or some prefix version?)
* use of mixfix notation; but here it seems to chime well with `[_]` as an operator?
* rename existing `concat-[-]` to eg. `concat∘map[_]≗id`? (or some more informative prefix version?)

### DIFF
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -681,6 +681,9 @@ concat-[-] : concat {A = A} ∘ map [_] ≗ id
 concat-[-] [] = refl
 concat-[-] (x ∷ xs) = cong (x ∷_) (concat-[-] xs)
 
+concat-[_] : concat {A = A} ∘ [_] ≗ id
+concat-[ xs ] = ++-identityʳ xs
+
 ------------------------------------------------------------------------
 -- concatMap
 
--- a/src/Data/List/Properties.agda
+++ b/src/Data/List/Properties.agda
@@ -677,9 +677,9 @@ concat-concat (xss ∷ xsss) = begin
   concat xss ++ concat (concat xsss) ≡⟨ concat-++ xss (concat xsss) ⟩
   concat (concat (xss ∷ xsss))       ∎
 
-concat-[-] : concat {A = A} ∘ map [_] ≗ id
-concat-[-] [] = refl
-concat-[-] (x ∷ xs) = cong (x ∷_) (concat-[-] xs)
+concat-map-[_] : concat {A = A} ∘ map [_] ≗ id
+concat-map-[ [] ]     = refl
+concat-map-[ x ∷ xs ] = cong (x ∷_) (concat-map-[ xs ])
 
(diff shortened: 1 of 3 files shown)


JacquesCarette (approved):

src/Data/List/Properties.agda
```diff
@@ -681,6 +681,9 @@ concat-[-] : concat {A = A} ∘ map [_] ≗ id
 concat-[-] [] = refl
 concat-[-] (x ∷ xs) = cong (x ∷_) (concat-[-] xs)
 
+concat-[_] : concat {A = A} ∘ [_] ≗ id
```
MatthewDaggitt, line 684: I would really like to use the same naming convention as `concat-[-]` above, which really should be called `concat-map-[-]`. However, I don't see a way to do so in the short-term. My suggestions would therefore be keep the name `concat-[_]` and deprecate the name `concat-[-]` to either

1. `concat-map-[_]` to be consistent with the new name
2. `concat-map-[-]` with the aim of eventually deprecating the new name `concat-[_]` to concat-[-]`

Taneb: If we want to start that process in this pull request I prefer the first option

jamesmckinna: One way to move in favour of 2 would be to label this v3.0, and simply make the `breaking` change...
... but it occurs to me that neither option actually precludes an eventual deprecation in favour of the `[-]` idiom?

I'll go for 1.

JacquesCarette (approved):

src/Data/List/Properties.agda
```diff
@@ -677,9 +677,12 @@ concat-concat (xss ∷ xsss) = begin
   concat xss ++ concat (concat xsss) ≡⟨ concat-++ xss (concat xsss) ⟩
   concat (concat (xss ∷ xsss))       ∎
 
-concat-[-] : concat {A = A} ∘ map [_] ≗ id
-concat-[-] [] = refl
-concat-[-] (x ∷ xs) = cong (x ∷_) (concat-[-] xs)
+concat-map-[_] : concat {A = A} ∘ map [_] ≗ id
```
Taneb, line 680: This deprecation/rename should be in the changelog

Taneb (approved):

src/Data/List/Properties.agda
jamesmckinna: Another :facepalm:

JacquesCarette (approved):

Issue #2532, opened 2024-12-22
[ refactor ] Recomcile left/right congruence statements and proofs


jamesmckinna:
> Have now added revised types for the new lemmas, and note the following:
* `++⁺ʳ` now has a type corresponding to the *explicitly* quantified version of `Algebra.Definitions.LeftCongruent`
* `++⁺ˡ` now has a type corresponding to the *explicitly* quantified version of `Algebra.Definitions.RightCongruent`
(and explicit because "`_++_` can't infer its arguments"... at least in the first case)

Which nomenclature/notation should be taken as canon? (Potential issue: I think that the use(s) of left/right superscript annotations in the `Data.List.Relation*` hierarchy are opposite to those of the `Algebra.*` congruence properties... sigh)

Barring the resolution of this question, I think this is (should be! ;-)) good to go now.

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2426#issuecomment-2254107549_

cf. #1436

JacquesCarette:
Could you phrase your "decisions to be made" crisply? I feel like I can't given any opinion right now, as I'm not quite sure what I have to give an opinion on!  Maybe split things into 1) crisp problem statement, and 2) discussion of the problem ?

jamesmckinna:
Ouch! Will attempt to do so...

jamesmckinna:
Here goes (but given I never seem to be able to meet your standards of crispness/non-verbosity, I'm not confident): there are 2 mismatches at issue, *as I see things*:
* between the abstract (`Algebra`) and concrete (`List`/`Vec`) versions of left/right congruence
* between implicit quantification (in `Algebra.Definitions.*Congruent`, in particular) and explicit quantification (`style-guide` recommendations regarding when arguments in `Algebra.Definitions.*` should be given explicitly in general)

As to discussion of the above, I've probably written a lot/too much already, but I'll return to this provided the above meets the required crispness threshhold.

JacquesCarette:
Those are quite crisp, thank you.
- I do think we should have a simple interpretation of what left/right means, and be able to clearly articulate it. And that this is absolutely worth fixing. 
- I also agree that guidelines for what should be explicit / implicit are worthwhile. Here we will have to settle for guidelines as I think there will always be exceptions. So the guidelines should probably include 'rules' for how to justify exceptions!

Those should likely be two issues, as they are not related to each other. Given the title of this issue, let's keep this one for left/right.

[...]

PR #2535, opened 2024-12-28
[ refactor ] exchange left and right in `Data.List.Relation.Binary.Pointwise` etc. congruence rules for `_++_`

jamesmckinna:
This is a do-over of #2426 , which I think should be regarded as having introduced a bug, which this PR fixes, as well as the original issue #1131 . 

This also addresses the left/right issue discussed extensively on the original PR and in #2532 .

Fortunately the original PR was v2.2, so this is *not* breaking, indeed should be a 'silent' refactoring.

Outstanding issue: refactor further to make use of `LeftCongruent` and `RightCongruent` from `Algebra.Definitions`, which changes explicit quantification as now to implicit quantification... I'm inclined to do this, so feedback welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,8 +356,8 @@ Additions to existing modules
 
 * In `Data.List.Relation.Binary.Equality.Setoid`:
   ```agda
-  ++⁺ʳ : ∀ xs → ys ≋ zs → xs ++ ys ≋ xs ++ zs
-  ++⁺ˡ : ∀ zs → ws ≋ xs → ws ++ zs ≋ xs ++ zs
+  ++⁺ˡ : ∀ xs → ys ≋ zs → xs ++ ys ≋ xs ++ zs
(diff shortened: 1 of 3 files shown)


JacquesCarette:
I don't want to go either way on this until we decide what left and right mean.

jamesmckinna:
> I don't want to go either way on this until we decide what left and right mean.

As discussed also on #2532 there is an internally/locally coherent notion of left/right already present in the affected modules, which the previous PR violated (so: in that sense, a 'bug'), and this restores that internal/local coherence.

The 'global' questions of what coherent/consistent design does the library adopt will clearly break some modules in favour of maintaining the choices in others, but that seems a different question to the one at hand of restoring the 'local' choices here.

JacquesCarette:
I'm willing to allow it for #2532 . Local coherence is good.

JacquesCarette (approved):
Ok, restoring local coherence is sufficient reason to go ahead now.

PR #2536, opened 2024-12-31
[ refactor ] add `rawSetoid` manifest fields to `Algebra.Bundles.Raw`

jamesmckinna:
This tackles the second part of #2274 as a followup to #2491 by replacing the in-place definition of negated equality in `Algebra.Bundles.Raw.*` with a delegation to its definition in the underlying `RawSetoid` bundle via an additional `rawSetoid` manifest field.

Outstanding issues:
* should the new field itself be publicly exported throughout, rather than simply the negated equality, as previously?
* does this addition give rise to similar problems as those recently fixed by #2499 ?

### DIFF
--- a/src/Algebra/Bundles/Raw.agda
+++ b/src/Algebra/Bundles/Raw.agda
@@ -10,6 +10,7 @@ module Algebra.Bundles.Raw where
 
 open import Algebra.Core
 open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Bundles.Raw using (RawSetoid)
 open import Level using (suc; _⊔_)
 open import Relation.Nullary.Negation.Core using (¬_)
 
@@ -27,6 +28,11 @@ record RawSuccessorSet c ℓ : Set (suc (c ⊔ ℓ)) where
     suc#    : Op₁ Carrier
     zero#   : Carrier
 
+  rawSetoid : RawSetoid c ℓ
+  rawSetoid = record { _≈_ = _≈_ }
+
+  open RawSetoid rawSetoid public using (_≉_)
+
 ------------------------------------------------------------------------
 -- Raw bundles with 1 binary operation
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

JacquesCarette:
"new field" ? I make a difference between a 'field' in a record and a definitional extension - they differ quite a bit. For example, you can only assign to fields (as seen when doing copatterns) and not extensions.

You can test what's in a module by putting the name of a module in a hole and then doing C-c C-o. This will list all the names that are visible in that module. That should let you figure out item 2.

jamesmckinna:
Apologies. What you call a "definitional extension" I habitually call a manifest field, and hence, sloppily, simply a field. 

Mind you, it would be great if they were overridable in the same way as `field`s, as that would make emulation of haskell-style defaulting sooo much easier...

src/Algebra/Bundles/Raw.agda
```diff
@@ -27,6 +28,11 @@ record RawSuccessorSet c ℓ : Set (suc (c ⊔ ℓ)) where
     suc#    : Op₁ Carrier
     zero#   : Carrier
 
+  rawSetoid : RawSetoid c ℓ
```
Taneb, line 31: I'm surprised that this is here rather than (say) `Relation.Binary.Bundles.Raw`. Sorry if I'm retreading on already-dealt-with topics here, I haven't been following the topic

jamesmckinna: So #2194 *defines* `Relation.Binary.Bundles.Raw` (following a suggestion of yours ;-)), and then this PR is designed to leverage that in order to avoid the duplication of the definition of `_≉_` previously under `Algebra.Bundles`... HTH

JacquesCarette:
I'm fine with "manifest field" (though I'll keep using "definitional extension"). I guess it's shortening that to "field" that bugs me!

Yes, I too would like some kind of 'default' mechanism. Though I think that, in Agda, one ought to be forced to prove that the alternate definition is equivalent to the default one. Which I guess means that providing a default on something with no notion of equivalence would be disallowed. It certainly would be interesting to play around with that!

src/Algebra/Bundles/Raw.agda
Taneb: Ach! I misread this as _defining_ the notion of `RawSetoid`. Serves me right for reviewing PRs overtired and slightly hung over, I guess.

Taneb (approved):

src/Algebra/Bundles/Raw.agda
jamesmckinna: Happy New Year! ;-)

PR #2540, opened 2025-01-03
Minor tidying up of CHANGELOG and prep for v2.2-rc1

MatthewDaggitt:


CHANGELOG.md
```diff
@@ -128,36 +128,28 @@ New modules
...
-  ```agda
-  sum-↭ : sum Preserves _↭_ ⟶ _≡_
-  ```
 
-* Added `Data.List.Relation.Binary.Permutation.Propositional.Properties.WithK`
-
-* Refactored out from `Data.List.Relation.Unary.All.Properties` in order to break a dependency cycle with `Data.List.Membership.Setoid.Properties`:
-  ```agda
-  Data.List.Relation.Unary.All.Properties.Core
+* Properties of list permutations that require the `--with-k` flag:
```
jamesmckinna, line 150: `--with-K`

jamesmckinna (approved):
Nice work!
Thanks very much indeed @MatthewDaggitt

Issue #2542, opened 2025-01-03
on lib-2.2-rc1


mechvel:
I have tested it under                                                          
Agda 2.7.0.1 built with flags (cabal -f  optimise-heavily)                      
ghc-9.0.2.  Debian Linux 12                                                     
on my large library of computer algebra.                                        
                                                                                
It works all right.                                                             
Please, find my remarks below.                                                  
                                                                                                                                                                
* Algebra.Definitions.RawMagma  defines                                         
```                                                                             
x ∤∤ y = ¬ x ∣∣ y                                                               
```                                                                             
I doubt, may be it is better to denote this `¬∣∣`.                              
Because seeing ``x ∤∤ y`` the reader could think of   "neither `x ∣ y` nor `y ∣ x`".                                                   
                                                                                
-----------------------------------------------------                           
Data.List.Base:                                                                 
```                                                                             
deduplicate : ∀ {R : Rel A p} → B.Decidable R → List A → List A                 
```                                                                             
The names ``take, drop, span, takeWhile, filter ...``    come from the Haskell library.                
And `deduplicate` corresponds to `nub` of Haskell library.                      
Right?                                                                          
So, I wonder of whether `deduplicate` needs to be renamed to `nub`              
(also I do not know the mnemonic of the name `nub`).                            
                                                                                
It is also needed a proof for the main properties of   `deduplicate`, the ones that explain why this function is needed:
(a) ``ys = (deduplicate xs)`` and `xs` need to be equal as sets  (to be subsets of each other),    
(b) `ys` must not have repetitions.                                             
                                                                                            
For example, I have in my library                                               
```                                                                             
  data AnyRelates (_~_ : Rel A β) :  Pred (List A) (α ⊔ β)                      
       where                                                                    
       here  :  ∀ {x xs} → Any (x ~_) xs → AnyRelates _~_ (x ∷ xs)              
       there :  ∀ {x xs} → AnyRelates _~_ xs → AnyRelates _~_ (x ∷ xs)          
                                                                                                               
AnyRepeats : Pred (List C) (α ⊔ α≈)     -- over a Setoid                          
AnyRepeats = AnyRelates _≈_                                                     
                                                                                
Over DecSetoid:                                                                 
nub :  (xs : List C) → Σ (List C) (\ys → (xs =ₛ ys) × ¬ AnyRepeats ys)          
```
  
---------------------------------------------                                   
Why is it introduced  ``∃[ x ] (P x)``                                          
[...]

PR #2545, opened 2025-01-12
[ refactor ] proofs under `Data.List.Membership.*.Properties`

jamesmckinna:
This fixes the `low-hanging-fruit` item under issue #2542 concerning the switch to a postfix definition of: 
`_∉[] : (x : A) → x ∉ []` in each of 
*  `Data.List.Membership.Setoid.Properties`
*  `Data.List.Membership.Propositional.Properties.Core`

Additionally:
* cleans up `import`s: `resp` rather than its equivalent `subst`;
* refactors to remove all uses of `rewrite` in the second module above, in the spirit of recent discussions. 

UPDATED:
* most recent (knock-on) commit exposes a slight wrinkle in the parameterisation: use here of the lemma, instantiated to `Setoid S` arises there as `(S Membershipₚ.∉[]) _` which looks... not quite right wrt the original problem (and removing parentheses causes a parse error!) but tweaked now to look a bit more reasonable?


MatthewDaggitt (approved):

MatthewDaggitt:
Thanks @jamesmckinna !

MatthewDaggitt:
> most recent (knock-on) commit exposes a slight wrinkle in the parameterisation: use here of the lemma, instantiated to Setoid S arises there as (S Membershipₚ.∉[]) _ which looks... not quite right wrt the original problem (and removing parentheses causes a parse error!) but tweaked now to look a bit more reasonable?

This has just sunk in. Yes, this is why we don't use misfix proofs as a rule, because complex module parameterisations often screw them over. I'd vote for staying with the current policy of not using misfix for proofs therefore (and codifying it in the style guide?).... Sorry, this is my fault, I knew this many years ago, but no longer.

jamesmckinna:
Thanks for the careful consideration @MatthewDaggitt !

So... I'll revert the controversial commits, but cherry-pick the equational refactorings as useful in their own right?

jamesmckinna:
Ugh... that didn't seem to work, sigh, because I hadn't been fine-grained enough about this. Will raise a fresh PR instead,

jamesmckinna:
Nope. Closing in frustration!

Issue #2553, opened 2025-01-16
[ refactor ] Remove `sum` and `product` (and their properties) from `Data.Nat.*`


jamesmckinna:
This is a long-standing legacy issue that causes problems esp. regarding:
* the dependency graph: `Data.Nat.Base` and `Properties` need to import `Data.List.*` otherwise unnecessarily; also, how much of `Algebra` should be imported by 'concrete' `Data` modules? `Data.List.Properties` ends up being where properties of `sum` and `product` get proved... 
* abstraction level: these are operations definable on `(Raw)Monoid`s, and should be treated as such
* extensibility: every time a 'new' property needs to be proven, the above questions need to be re-examined (esp. wrt dependency, eg do we want to import `Data.List.Relation.Binary.Permutation.*` into `Data.Nat.Properties`, etc.?)
* etc.

Just as, in the end, it was (judged to be) 'better' to lift out `scan`s from `Data.List.Base`, `sum` and `product` should be lifted out from `Data.Nat.Base` (and correspondingly their properties from `Data.Nat.Properties`) into (at least!) a separate module.

Lots of design choices here:
* a general theory under `Algebra.Definitions.RawMonoid`/`Algebra.Properties.Monoid`? Deprecation problems wrt the existing definitions and properties of `sum` etc. so potentially lots of `breaking` changes, otherwise a mess of deprecations?
* theory of interaction(s) between them (esp. wrt distributivity) under `Algebra.Properties.(Commutative)(Semi)Ring` etc.
* instantiation for `Nat` (and the other numeric types `Integer` and `Rational`?): what should the module be called?
* ...

Possible implementation pathway:
* v2.3, a 'minimal, deprecation-only refactoring
* v3.0, a breaking version to get things 'right'

mechvel:
> the dependency graph: Data.List.Base and Properties need to import Data.Nat.Divisibility otherwise unnecessarily; also, 
> how much of Algebra should be imported by 'concrete' Data modules? Data.List.Properties ends up being where properties > of sum and product get proved...

It is not natural for Data.List  to use the function of product of a list of natural numbers, and for  Data.List.Properties  to 
prove the properties of this product. 
But  it is possible to introduce, say Data.Nat.List and put there  the items for List Nat.

JacquesCarette:
I emphatically agree with removing `sum` and `product` from `Data.List.Base`. That was a grave mistake that is relatively harmless in Haskell-like languages and deadly in Agda-like languages.

[...]

Issue #2446, opened 2024-07-29
Refactor ` Data.Nat.GeneralisedArithmetic`


jamesmckinna:
Now that we have `Algebra.Structures.IsSuccessorSet`/`Algebra.Bundles.SuccessorSet`, should we refactor this module (and others!?) to reflect the action/behaviour/homomorphism properties of the *initial* `SuccessorSet`, viz. `Nat`...? Cf. `Algebra.Definitions.RawMonoid.(Sum|Mult|Exp)` etc. ...

JacquesCarette:
Seems very reasonable to me. Although I've got to wonder how much of this is essentially going to be a halfway house to the categorical properties of a NNO?

jamesmckinna:
> Seems very reasonable to me. Although I've got to wonder how much of this is essentially going to be a halfway house to the categorical properties of a NNO?

Indeed. `Free` is lurking off-stage, as with so many constructions in the library...?

PR #2558, opened 2025-01-16
[ refactor ] Add `Data.Nat.ListAction`

jamesmckinna:
Implements the 'provisional' v2.3 fix for #2553 .

Issues (summarising discussion below):
* naming+location of new module: under `Data.Nat` or `Data.List` (or elsewhere?) UPDATED to `Data.Nat.ListAction`
* scope of PR: minimal for v2.3, or maximal/breaking for v3.0?

NB.:
* fixes a spurious deprecation message from v2.0 in `Data.List.Properties` concerning `sum-++-commute`, but now the message is... anachronistic. Wasn't sure how to deal with this properly!


Taneb:
I have to say, I don't like the module name... there are other monoids on Nat (e.g. (max, 0)) whose folds we'd also want to have in the same module if we have them.

jamesmckinna:
Thanks @Taneb !

Bear in mind that I regard this PR as provisional, indeed minimal wrt *not* refactoring to lift out the actual `Monoid` abstractions as such, before a 'proper' (`breaking`) refactoring which will deprecate/remove this module, and all traces of the original definitions. In particular, I would want this PR only to emulate existing functionality, rather than try to generalise it to arbitrary monoid structures arising on `Nat`. Perhaps that's a failure of imagination/nerve?

Eg, I could even imagine doing that generalisation (in whatever form) downstream, and then simply deprecating this module with its 'clunky' name. perhaps I even thought that would be inevitable, so either didn't care, or actively chose, to make the name 'clunky'... on a 'build one to throw away' principle...

That said, would (any of) the following be better candidate names?
* `Data.Nat.MonoidAction`/`Data.Nat.FreeMonoidAction` (a bit too general, since we're only considering the action of `Nat` on *itself*...?); it's a pity I haven't/hadn't managed to finish #2350 before this, but I've been trying to not let the 'perfect design' get in the way of progress... ;-), but eventually, as with #2446 , there ought to be a grand refactoring of all this stuff... 
* `Data.Nat.Monoid`

Other suggestions welcome! @JacquesCarette any thoughts on this?

jamesmckinna:
> As I said elsewhere, fully agree with moving these out of where they currently are.

Thanks. I had the feeling that the consensus was already 'out there' (it seems to have been a recurring issue *en passant* for years without anyone tackling it seriously...)

> Hmm, why `Data.Nat.*` though? I was thinking of these as "special cases of `List`-based computation" and so under `Data.List`. `Data.List.NatActions`? [Obviously not wedded to that awkward name, but pushing things along.]

Interesting... but a rather different perspective to mine (although I could agree with `Data.Nat.Actions, as a version of one of the choice above...) precisely because:
* as @Taneb points out, each of these (and more...) arises from a `Monoid` on `Nat`, and hence a `MonoidAction` of `Nat` on itself, ...
* and then these operations are the canonical *lifting* of that action to that of the `Free Monoid` on the carrier of `Nat`, by iteration (over the underlying list representation; but others are possible!)

So from my point of view, `List` is *an implementation detail*, when the 'real' action (sic) lies with (actions on) `Nat. That's why I chose `Data.Nat` as the root...

> If it were in `Data.Nat` then I'd see it being way more general, i.e. over any foldable container.

Well, indeed, but that would be a generalisation beyond the scope of this v2.3-badged PR, which I've deliberately tried to keep as minimal as possible, while opening up the design/discussion space for subsequent downstream refactorings. Besides which, I suspect (strongly) that any `Foldable` functor `F` can have its `fold`/`foldMap` definition for `Monoid M` factored through first the (carrier of the) free monoid on M, viz. `List M` followed by the iterated `List`-action `List M → M`... (this must be in the automata literature somewhere?)

Taneb:
Do we also want to move `and` and `or` out (to something under `Data.Bool` if `sum` and `product` are under `Data.Nat`)? What about `any` and `all`?

jamesmckinna:
> Do we also want to move `and` and `or` out (to something under `Data.Bool` if `sum` and `product` are under `Data.Nat`)? What about `any` and `all`?

Well... actually no,  not right now, but it might very well make sense to do so in downstream PRs? The focus of this PR, and the originating linked issue was for the 'problematic' `Nat` iterated actions. But happy to see the debate widened to other base types... but suggest such discussion move to #2553 ? or a fresh issue?

jamesmckinna:
Hmmm... @Taneb 's comments (in one direction), and @JacquesCarette 's (a bit in another) suggest perhaps this should be refactored in terms of `Data.List.Effectful.Foldable`, but nevertheless should live under `Data.Nat` for `sum`/`product`, and under `Data.Bool` for `and`/`or, and even `all`/`any`? As e.g `

And indeed that `Effect.Foldable` (or its `List` implementation) should also take responsibility for the `CommutativeMonoid`/`Permutation` interaction... as a downstream refactoring?

JacquesCarette:
As I said elsewhere, agree with @Taneb that those should be moved as well. Also agree that while an issue could cover moving both sets of things, doing it in smaller PRs is more to my taste.

My feeling is indeed that this stuff belongs under `Data.Nat` and `Data.Bool`. Willing to be convinced otherwise on that aspect.

And yes, the generalizations of this (to more general containers + algebras) absolutely should be done as a separate, downstream refactor.

jamesmckinna:
So:
* I'll rename to `Data.Nat.ListAction`, as some sort of median-of-three arbitration of the discussion above?
* I'll raise a separate analogous PR for `Bool`?

Comments, as ever, welcome!

JacquesCarette:
`Data.Nat.ListAction` seems a quite reasonable compromise name to me.

CHANGELOG.md
```diff
@@ -21,8 +21,30 @@ Deprecated modules
...
 
+* In `Data.List.Base`:
+  ```agda
+  sum       ↦  Data.Nat.ListAction.sum
+  product   ↦  Data.Nat.ListAction.product
+  ```
+
+* In `Data.List.Properties`:
+  ```agda
+  sum-++       ↦  Data.Nat.ListAction.sum-++
```
JacquesCarette, line 32: Hmm, I think these ought to be in `Data.Nat.ListAction.Properties`. I still will want to have `sum` and `product` available with a decently small footprint (dependency-wise) and bundling properties with definitions defeats that.

src/Data/Tree/Binary/Zipper.agda
```diff
@@ -10,9 +10,10 @@ module Data.Tree.Binary.Zipper where
 
 open import Level using (Level; _⊔_)
 open import Data.Tree.Binary as BT using (Tree; node; leaf)
-open import Data.List.Base as List using (List; []; _∷_; sum; _++_; [_])
+open import Data.List.Base as List using (List; []; _∷_; _++_; [_])
 open import Data.Maybe.Base using (Maybe; nothing; just)
 open import Data.Nat.Base using (ℕ; suc; _+_)
+open import Data.Nat.ListAction using (sum)
```
JacquesCarette, line 16: This is one example use where bundling properties will have made the dependency graph worse.

JacquesCarette (changes requested):
Other than property placement, the rest looks good!

CHANGELOG.md
jamesmckinna: And a corresponding `Base` module for the definitions?

jamesmckinna:
> Other than property placement, the rest looks good!

@JacquesCarette Have made the requested changes. 
Seeking yours, and second approval (@Taneb ?), after the successful merge of #2561

JacquesCarette (approved):

Taneb (approved):

MatthewDaggitt (approved):

PR #2561, opened 2025-01-22
[ refactor ] Add `Data.Bool.ListAction`

jamesmckinna:
This is the `Bool` analogue of #2558; see discussion there and  on #2553 .

Issue:
* intended as a v2.3 compatibility stub module, ahead of a more thorough refactoring in terms of `Data.List.Effectful.Foldable.foldMap`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ Deprecated names
 New modules
 -----------
 
-* `Data.List.Base.{and|or|any|all}` and their properties have been lifted out into `Data.Bool.ListAction`.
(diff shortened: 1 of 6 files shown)


jamesmckinna:
Aaargh! Once again, the deprecations warnings are only biting when the HTML indexing phase takes place, so these errors aren't turning up locally when I run `make test`.

jamesmckinna:
> Surprised that none of those functions have properties in the library!

Well, I think that once each of `all` and `any` are shown equivalent (via `T`) to their corresponding `Relation.Unary` predicate forms `All` and `Any`, perhaps they don't need to?

gallais (approved):

Issue #2559, opened 2025-01-16
`IntegralSemiring`  proposal


mechvel:
atthew Daggitt  wrote on lib-2.2-rc1
on January 4, 2025

> product≢0 : All NonZero ns → NonZero (product ns) -
> while I agree that it would be good to develop the theory in terms of an
> appropriate algebraic structure, unfortunately the library is not
> currently set up to do so. We would need someone to be motivated to make
> those contributions, but regardless those changes would also not make it
> into v2.2.

The  --attached archive--  contains the .agda files that set up lib-2.2 to do so,
which makes it  lib-2.2-next.
It is set in the standard library style.
It consists of the following files.

Algebra-Definitions.agda
  to be merged to  Algebra/Definitions.agda  of lib-2.2.
It defines the property NoZeroDivisor
``zero _∙_ = {x y : A} → (x ∙ y) ≈ zero → x ≈ zero ⊎ y ≈ zero``

Algebra-Properties-Monoid.agda
  to be set as  Algebra/Properties/Monoid.agda.
It defines
```
Π₁ : C → List C → C    -- product of a nonempty list of elements,
Π : List C → C         -- product of a list of elements.
It proves
Π[x] : ∀ x → Π (x ∷ []) ≈ x
```

Algebra-Structures.agda
  to be merged to  Algebra/Structures.agda.
It defines
```
record IsIntegralSemiring (+ * : Op₂ A) (0# 1# : A) : Set (a ⊔ ℓ) where
  field
    isSemiring    : IsSemiring + * 0# 1#
    0≉1           : ¬ 0# ≈ 1#
    noZeroDivisor : NoZeroDivisor 0# *

  open IsSemiring isSemiring public
```

Algebra-Bundles.agda
  to merge to  Algebra/Bundles.agda.
It defines
```
record IntegralSemiring c ℓ : Set (suc (c ⊔ ℓ)) where
  ...
  isIntegralSemiring : IsIntegralSemiring _≈_ _+_ _*_ 0# 1#
```

Algebra-Properties-Semiring-Integral.agda
  to set as  Algebra/Properties/Semiring/Integral.agda.
It proves
```
nonzero*nonzero :  ∀ {x y} → x ≉ 0# → y ≉ 0# → (x * y) ≉ 0#
nonzero*nonzero {x} {y} x≉0 y≉0 xy≈0 with noZeroDivisor xy≈0
... | inj₁ x≈0 = x≉0 x≈0
[...]

PR #2563, opened 2025-01-24
[ add ] `Algebra.Definitions.Integral` and its consequences

### LINKED ISSUE #2554: [ add ] Definition, consequences, properties, structures, and bundles for `Algebra.*`s with the 'no non-zero divisors' property
Lifted from #2542 , but has arisen before in various scattered proposals to extend the `Algebra` hierarchy to include:
* `IntegralDomain`
* `EuclideanDomain`
* `UniqueFactorisationDomain` 
* `PrincipalIdealDomain`
* ...

which is (probably) too big a project to identify as a single issue/PR?

jamesmckinna:
Fixes #2554 . With thanks to @mechvel for discussion and his contributions on #2559 , from which this obviously diverges slightly.

TODO:
- [x] `Algebra.Definitions.Integral` etc. and one `Algebra.Consequences.Base`
- [ ] `(Is)Semiring` etc. up to `(Is)IntegralCommutativeRing`
- [ ] `(Is)IntegralDomain` as the orthodox special case of `(Is)IntegralCommutativeRing` satisfying `1#≉0# : 1# ≉ 0#`, and hence `NoZeroDivisors 0# _•_` outright

### DIFF
--- a/src/Algebra/Consequences/Base.agda
+++ b/src/Algebra/Consequences/Base.agda
@@ -15,6 +15,7 @@ open import Algebra.Definitions
 open import Data.Sum.Base
 open import Relation.Binary.Core
 open import Relation.Binary.Definitions using (Reflexive)
+open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 
 module _ {ℓ} {_•_ : Op₂ A} (_≈_ : Rel A ℓ) where
 
@@ -27,6 +28,14 @@ module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) where
                                      SelfInverse _≈_ f →
                                      Involutive _≈_ f
   reflexive∧selfInverse⇒involutive refl inv _ = inv refl
+
+module _ {ℓ} {_•_ : Op₂ A} {0# 1# : A} (_≈_ : Rel A ℓ) where
+
+  integral⇒noZeroDivisors : Integral _≈_ 1# 0# _•_ → ¬ (1# ≈ 0#) →
+                            NoZeroDivisors _≈_ 0# _•_
+  integral⇒noZeroDivisors (inj₁ 1#≈0#)          = contradiction 1#≈0#
+  integral⇒noZeroDivisors (inj₂ noZeroDivisors) = λ _ → noZeroDivisors
+
 
 ------------------------------------------------------------------------
 -- DEPRECATED NAMES
--- a/src/Algebra/Definitions.agda
+++ b/src/Algebra/Definitions.agda
@@ -66,6 +66,12 @@ RightZero z _∙_ = ∀ x → (x ∙ z) ≈ z
 Zero : A → Op₂ A → Set _
 Zero z ∙ = (LeftZero z ∙) × (RightZero z ∙)
 
+NoZeroDivisors : A → Op₂ A → Set _
+NoZeroDivisors z _∙_ = ∀ {x y} → (x ∙ y) ≈ z → x ≈ z ⊎ y ≈ z
+
+Integral : A → A → Op₂ A → Set _
+Integral 1# 0# _∙_ = 1# ≈ 0# ⊎ NoZeroDivisors 0# _∙_
+
 LeftInverse : A → Op₁ A → Op₂ A → Set _
 LeftInverse e _⁻¹ _∙_ = ∀ x → ((x ⁻¹) ∙ x) ≈ e
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,12 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Consequences.Base`:
+  ```agda
+  integral⇒noZeroDivisors : Integral _≈_ 1# 0# _•_ → ¬ (1# ≈ 0#) →
+                            NoZeroDivisors _≈_ 0# _•_
+  ```
+
 * In `Algebra.Construct.Pointwise`:
   ```agda
   isNearSemiring                  : IsNearSemiring _≈_ _+_ _*_ 0# →
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,3 +85,10 @@ Additions to existing modules
   quasiring                       : Quasiring c ℓ → Quasiring (a ⊔ c) (a ⊔ ℓ)
   commutativeRing                 : CommutativeRing c ℓ → CommutativeRing (a ⊔ c) (a ⊔ ℓ)
   ```
+
+* In `Algebra.Definitions`:
+  ```agda
+  NoZeroDivisors : A → Op₂ A → Set _
+  Integral       : A → A → Op₂ A → Set _
+  ```
+  (see [discussion on issue #2554](https://github.com/agda/agda-stdlib/issues/2554))


mechvel:
James, 
thank you for correcting the proposal.
Is it possible to get the archive of the source of the pull request   3f6bdbb  you have done?
If it is possible, I would copy it, install it as the library version,  with calling it for myself say  lib-2.3-rc1
(earlier I had difficulties with pull requests, it is easier for me to deal with a release candidate).
Then I could continue it by proposing the items that you announce as  TODO.
 ?

Taneb:
@mechvel you can download an archive at that commit here: https://github.com/agda/agda-stdlib/archive/3f6bdbb86b3c45f8eb1af8e9d042dac26893123f.zip

I found this link by clicking the commit, pressing the "Browse files" button, and then looking in the green "Code" button drop down

jamesmckinna:
@mechvel Sergei, I chose to begin pushing to this PR before it was finished (so, properly, I should have marked it as `DRAFT`!), but as you perhaps can see, I am still actively making commits. Suggest that you hold off from downloading just at the moment?

mechvel:
James, please inform me when it appears  a reasonable moment for me to download the library version (I hope I will be able to download the archive of the source).
There remain a couple of  items in my proposal  onIntegralSemiring.zip which are not currently in CHANGELOG.md.

mechvel:
Taneb wrote
> @mechvel, you can download an archive at that commit here: https://github.com/agda/agda-stdlib/archive
> /3f6bdbb86b3c45f8eb1af8e9d042dac26893123f.zip
> 
> I found this link by clicking the commit, pressing the "Browse files" button, and then looking in the green "Code" button > drop down

Thank you. I shall try this, where James signals that the commit is ready.

mechvel:
@jamesmckinna asks 
> deprecate (Is)CancellativeCommutativeSemiring in favour of Integral versions?

For `Ring`, I believe, these two are equivalent. So that `CancellativeRing`  is not needed.

But how do you derive  ``Integral ==> LeftCancellative`` for `Semiring`?
For `Ring' , I derive this this way:
Let x*y == x*z and x \= 0 ,  Then  x*y  -  x*z  == 0  ==  x*(y-z).   Then it follows from `Integral` that  y-z == 0.
But in `Semiring`, the above subtraction does not exist. And I do not know how to prove this, so that  for `Semiring` I provide both `Integral` and `Cancellative`.
What people think of this?

src/Algebra/Consequences/Base.agda
```diff
@@ -28,6 +33,30 @@ module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) where
...
+
+module _ {ℓ} {_∙_ : Op₂ A} {0# : A} (_≈_ : Rel A ℓ) where
+
+  private
+    _≉_ : Rel A ℓ
+    x ≉ y = ¬ (x ≈ y)
+
+  noZeroDivisors⇒x≉0∧y≉0⇒xẏ≉0 : NoZeroDivisors _≈_ 0# _∙_ →
+                                x ≉ 0# → y ≉ 0# → (x ∙ y) ≉ 0#
+  noZeroDivisors⇒x≉0∧y≉0⇒xẏ≉0 noZeroDivisors x≉0 y≉0 xy≈0 with noZeroDivisors xy≈0
```
JacquesCarette, line 55: don't have time to fully try, but could this be
```agda
either x≉0 y≉0 (noZeroDivisors xy≈0)
```

JacquesCarette (commented):
other than the one item, this is looking nice to me.

jamesmckinna:
@JacquesCarette : thanks for the early review
@mechvel : this is ready to look at now, but please be aware that the general 'NonZero' properties will *not* get covered here; there's too much going on with #2558 #2561 and then rationalising the designs for v3.0 to consider such additions now, I think.

Also: your comments about `Cancellative` are sensible: it is more that I believed we now have a redundant' entry in the namespace, but I need to think more about your comments about the `Semiring` structures/bundles.

src/Algebra/Consequences/Base.agda
jamesmckinna: Ah!? Will investigate tomorrow...

jamesmckinna:
> James wrote
> 
>     * fold definition of Trivial better into Algebra.Definitions so that Algebra.Definitions.Integral can make use of it?
> 
>     * breaking rephrase Algebra.Properties.Semiring.Primality in terms of Trivial?
> 
> 
> Is `Trivial` ever defined in lib-2.2 Algebra?

No, this is new to this v2.3-badged PR.

mechvel:
I wrote  about  trying to derive Integral ==> LeftCancellative for Semiring.
Each S : Ring  has the commutative additive Monoid M.  M can be extended to A : AbeleanGroup (probably called enveloping group for M).   So there is a hope that IntegralSemiring can be extended to IntegralRing, and  by this,
the equivalence IntegralSemigring <--> LeftCancellativeSemiring can be proved.
It is not so difficult to verify. But setting this whole proof  to Standard library will need to add many items.
So that (Commutative)CancellativeSemiring  cannot be replaced, so far, with IntegralSemiring
(may be there is possible a simpler proof , I do not know).

mechvel:
The definition of AlmostLeft(Right)Cancellative  
``∀ x y z → ¬ x ≈ e → (x • y) ≈ (x • z) → y ≈ z``
is given (by myself) several years ago again with ignoring the case of undecidable equality.
If the team cares for this feature, we have to change this to disjunction.
?

Taneb:
I'm not convinced these should directly be in `Algbera.{Structures,Bundles}`. As I've said elsewhere, there's two different classically equivalent constructively distinct definitions we want here (especially in the future when we've got real numbers). I'd like to suggest `Algebra.Discrete.{Structures,Bundles}` as a counterpart to the existing `Algebra.Apartness`

src/Algebra/Definitions.agda
```diff
@@ -66,6 +66,12 @@ RightZero z _∙_ = ∀ x → (x ∙ z) ≈ z
 Zero : A → Op₂ A → Set _
 Zero z ∙ = (LeftZero z ∙) × (RightZero z ∙)
 
+NoZeroDivisors : A → Op₂ A → Set _
+NoZeroDivisors z _∙_ = ∀ {x y} → (x ∙ y) ≈ z → x ≈ z ⊎ y ≈ z
+
+Integral : A → A → Op₂ A → Set _
+Integral 1# 0# _∙_ = 1# ≈ 0# ⊎ NoZeroDivisors 0# _∙_
```
Taneb, line 73: Assuming that `0#` and `1#` have their normal properties wrt multiplication, `1# ≈ 0#` implies all other equalities:
x = 1*x = 0*x = 0 = 0*y = 1*y = y

mechvel:
I think Taneb is right.
```
lemma1 :  (1# ≈ 0# ⊎ NoZeroDivisors 0# _∙_) → NoZeroDivisors 0# _∙_
lemma1 (inj₁ 1≈0)    = noZeroDivisors  -- because everything equals zero due to `1≈0`.
lemma1 (inj₂ noZDiv) = noZDiv

lemma2 :  NoZeroDivisors 0# _∙_ → (1# ≈ 0# ⊎ NoZeroDivisors 0# _∙_)
lemma2 =  inj₂	
```
Therefore   ``1# ≈ 0# ⊎ NoZeroDivisors 0# _∙_``
is equivalent to  `` NoZeroDivisors 0# _∙_``.
Right?
Further, the initial (my suggested and classical one too) definition was

(D)
record IsIntegralSemiring ...
  where
  field
    1≉0 : 1 ≉ 0#                                                  -- (I)
    noZeroDivisors : NoZeroDivisors 0# _∙_   -- (II)

We cannot remove (I), because the classical definition requires `IntegralSemiring` to be non-zero.

Neither we need to remove (I) with changing (II) to
``1# ≈ 0# ⊎ (NoZeroDivisors 0# _∙_)``.
Because of the equivalence given by  `lemma1` and `lemma2`.
I am sorry for that I missed this when agreed with changing the definition (D).

It comes out that we have to
a) return to (D),
b) remove the definition `Integral`,
c) remove from the library these wise comments about "positive" definitions with relation to the case of undecidable equality -- for thease of	`Integral'
   (for	other cases it is all right, but for this particular case of the property 1# ≈ 0# and zero ring, it is not needed).

Do I mistake?
If I do not mistake, then I am sorry for spending our effort.

mechvel:
But I have lost the first archive that I uploaded to here. Can people find it on GitHub?

jamesmckinna:
Ai! What can I have been thinking/doing!?
Apologies to all, I'd better have a long rethink.

src/Algebra/Consequences/Base.agda
JacquesCarette: The eliminator for disjoint union.

JacquesCarette:
This is not `breaking` so could it be v2.4 rather than v3.0?

jamesmckinna:
> This is not `breaking` so could it be v2.4 rather than v3.0?

I guess so, but six months later, I would need to page a lot back in to do so, and there are more pressing v2.3/v2.4 issues/PRs currently occupying me ;-)

Issue #2565, opened 2025-01-28
[ refactor ] `Function.Definitions` and `Function.Consequences`?


jamesmckinna:
Against #2514 , and continuing to think about #1436 etc., cf. #2318 

I think that it might be more ergonomic to refactor `Function.Definitions` to separate out:
* those `Definitions` parametrised on two equality relations, which would now become *top-level* module parameters
* the `Strictly*` definitions, parametrised on a *single* equality relation, which I would move to a distinct new `Function.Definitions.Strict` module, again parametrised on the relation (and indeed, the relative importance of these definitions relative to the others has been discussed at length; it feels as though they are 'special', and not for 'everyday' use, except for the `Propositional` instantiation with `_≡_` in `Data.Nat.Properties` etc.)
* knock-on consequences: `Function.Consequences` and `Algebra.Definitions`, where in particular, one can immediately instantiate the two relations with the underlying equality of the `Algebra` (this would be the ergonomic payoff).

Draft PR in preparation.

MatthewDaggitt:
This is exactly the structure we used to have but it didn't work because of the way that Agda handles transitive top-level parameters. Therefore we switched to the current design in a breaking change. Let me try and dig up the issue.

MatthewDaggitt:
See
- https://github.com/agda/agda-stdlib/issues/1970
- https://github.com/agda/agda-stdlib/pull/1156

There's also an upstream issue on the Agda repo describing the underlying problem with top-level module parameters where if I remember correctly there was some debate about whether the current behaviour that caused this is a bug or a feature. However, I can't seem to find it....

jamesmckinna:
Wow!?

MatthewDaggitt:
So if I understand it correctly your proposal is go back to the old design, but fix the problem by not having the auxiliary modules re-exported by the top level `Definitions` module?

jamesmckinna:
I'll post the PR... shortly

jamesmckinna:
So, I suspect that this is now moot, given the larger-scale developments in #2568 / #2569 but I think may still be worthwhile downstream as part of tackling #2570 ... so marking as `blocked-by-issue` for the time being.

Retrospective update: I no longer think blocked by #2570 as such, nor indeed by now-closed #2569 , but almost surely by the much more consequential #2583 ... so leaving the label attached ;-)

jamesmckinna:
[...]

PR #2566, opened 2025-01-28
[ add ] `Relation.Binary._Reflects_⟶_` as a companion to `_Preserves_⟶_`

jamesmckinna:
A preparatory PR ahead of tackling #1436 in earnest.
NB:
* illustrative refactoring of `Function.Definitions.Injective`
* surprisingly, attempting to do the same for `Function.Definitions.Congruent` in terms of `Preserves` results in many, many, unsolved metas... sigh. See #2565 for details

### DIFF
--- a/src/Function/Definitions.agda
+++ b/src/Function/Definitions.agda
@@ -10,9 +10,10 @@
 
 module Function.Definitions where
 
-open import Data.Product.Base using (∃; _×_)
+open import Data.Product.Base using (∃; _×_; proj₁)
+open import Function.Base using (_∘_)
 open import Level using (Level)
-open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Core using (Rel; _Reflects_⟶_)
 
 private
   variable
@@ -31,7 +32,7 @@ module _
   Congruent f = ∀ {x y} → x ≈₁ y → f x ≈₂ f y
 
   Injective : (A → B) → Set _
-  Injective f = ∀ {x y} → f x ≈₂ f y → x ≈₁ y
+  Injective = _Reflects _≈₂_ ⟶ _≈₁_
 
   Surjective : (A → B) → Set _
   Surjective f = ∀ y → ∃ λ x → ∀ {z} → z ≈₁ x → f z ≈₂ y
--- a/src/Function/Definitions.agda
+++ b/src/Function/Definitions.agda
@@ -10,7 +10,7 @@
 
 module Function.Definitions where
 
-open import Data.Product.Base using (∃; _×_; proj₁)
+open import Data.Product.Base using (∃; _×_)
 open import Function.Base using (_∘_)
 open import Level using (Level)
 open import Relation.Binary.Core using (Rel; _Reflects_⟶_)
--- a/src/Function/Definitions.agda
+++ b/src/Function/Definitions.agda
@@ -11,7 +11,6 @@
 module Function.Definitions where
 
 open import Data.Product.Base using (∃; _×_)
-open import Function.Base using (_∘_)
 open import Level using (Level)
 open import Relation.Binary.Core using (Rel; _Reflects_⟶_)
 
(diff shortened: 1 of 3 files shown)


gallais:
What's the intuition for the `Reflects` name? Does it have any relationship to the ssreflect
style `Reflects` that we already have? http://agda.github.io/agda-stdlib/v2.1.1/Relation.Nullary.Reflects.html#1010

jamesmckinna:
> What's the intuition for the `Reflects` name? Does it have any relationship to the ssreflect style `Reflects` that we already have? http://agda.github.io/agda-stdlib/v2.1.1/Relation.Nullary.Reflects.html#1010

Nothing to do with `Reflects` as in ssreflect, nor `Relation.Nullary.Reflects`!
But absolutely conventional usage in first-order logic/relational algebra (eg. discussion in McKinna&Pollack 1999 on preserving and reflecting α-conversion in λ-calculus representations...)

src/Function/Definitions.agda
```diff
@@ -31,7 +31,7 @@ module _
   Congruent f = ∀ {x y} → x ≈₁ y → f x ≈₂ f y
 
   Injective : (A → B) → Set _
-  Injective f = ∀ {x y} → f x ≈₂ f y → x ≈₁ y
+  Injective = _Reflects _≈₂_ ⟶ _≈₁_
```
MatthewDaggitt, line 34: I'm kind of against this change personally. I read `Injective f = ∀ {x y} → f x ≈₂ f y → x ≈₁ y` and I understand exactly what it says.

I read `Injective = _Reflects _≈₂_ ⟶ _≈₁_` and I have to do the following:
1. Expand the missing function to  `Injective f = f Reflects _≈₂_ ⟶ _≈₁_` (non-trivial for beginners)
2. Expand Reflects to `(_≈₁_ on f) ⇒ _≈₂_` which is equally baffling.
3. Expand `on` ...
4. Expand `_⇒_`...
5. Only now do I get to the actual definition.

All of these definitions are in different modules and Agda currently provides no easy method of expanding them. In my mind the very small benefit of this change is outweighed by the quite large decrease in usability!

jamesmckinna: @MatthewDaggitt very sharp critique from the ux perspective (I think perhaps we should have `ux` as a new label category for tagging issues/PRs), for which all thanks.

As the preamble to this PR indicates, I wanted this *concept*/definition in `stdlib` so that (eventually) all the `Cancellative` properties could be rephrased in terms of it. Currently, under #1436 / #2573 we only treat the cancellative properties for *equality*... if you will, perhaps as another, more directly graspable concept, we should get rid of `Cancellative` and friends altogether in favour of saying that eg `m +_` is `Injective` on `Nat` with `_≡_` (that would be the other way to go).

It's perhaps the case that my ideological/reformatory zeal on [DRY] issues leads to (over-)generalisations like those considered here (or most recently, in #2581 ), and that reflects what is (probably!) the greater abstraction power of higher-order languages such as Agda's type theory relative to the abstraction power of users (esp. wrt definitional expansion across modules, as you indicate so forcefully).

Against your suggestion, my own preference would be to record as a comment, the 'conventional' usage as being a definitional expansion of the abstract one. Something like
```agda
  Injective : (A → B) → Set _
  Injective = _Reflects _≈₂_ ⟶ _≈₁_
-- i.e.  Injective f = ∀ {x y} → f x ≈₂ f y → x ≈₁ y
```
Of course, that would be a break from existing practice in the library, as well as opening the door to bit-rot if/when definitions change, but here I don't think we are at quite such a risk... precisely because the definition is so 'stable'?

jamesmckinna: All of that said, happy to revert this particular instance of the definition, but then what about `Cancellative` in all its variety?

MatthewDaggitt: I think adding non-type checked comments is very susceptible to bit rot as you say. I would be more open to a private definition that performed an equality check.

However, I'm still not entirely sure what advantages the generalisation brings in this case. It's not like we have either a) a set of properties/theory about these kinds of definitions in general or b) lots of definitions in this form? `Injective`, `Cancellative`, anything else?

Ditto, I'm not sure that I would be keen on replacing the definitions of `Cancellative` for exactly the same reasons...

jamesmckinna: I guess I would still try (but ) to justify the addition
* on formal grounds, as the 'opposite' of `Respects`
* on abstraction grounds (I clearly prefer greater degrees of abstraction, yes ,for its own sake, so I'll try to keep that in mind in future)

but this particular stone seems to be getting harder to push uphill. I'll stop!

PR #2567, opened 2025-01-28
[ refactor ] `Function.Definitions`, adding `Function.Definitions.Strict`  module

### LINKED ISSUE #2565: [ refactor ] `Function.Definitions` and `Function.Consequences`?

jamesmckinna:
See #2565 
NB:
* *lots* of (trivial `import`) knock-on consequences
* little exploration of the possible ergonomic benefits yet
* `breaking` because basically impossible to deprecate wrt the old parameterisation...

UPDATE: I've cherry picked the non-breaking changes for #2569 , so I won't fight for this one...


jamesmckinna:
The *reason* for this PR was to factor out (sub-)modules with different dependencies/parametrisation.
The *issue* perhaps is moot? But eg #2888 and related issues/PRs suggest that *some* refactoring might offer a better UX/ergonomics?

JacquesCarette:
I think the changes to `Function.Consequences.Propositional` are nice and should not be lost -- and don't seem to depend on the movement?

I must admit, I'm not very fond of the name `Functions.Definitions.Strict` as it feels like we're abusing 'strict' to mean too many things. Either that or there should be non-trivial comments in the new file justifying the terminology.

jamesmckinna:
> I think the changes to `Function.Consequences.Propositional` are nice and should not be lost -- and don't seem to depend on the movement?

Thanks! Indeed, they seem independent, so could simply be lifted out as a small PR in their own right, but the value seems to come as much from the shift in parametrisation? That does arise from the code motion/reorganisation...

> I must admit, I'm not very fond of the name `Functions.Definitions.Strict` as it feels like we're abusing 'strict' to mean too many things. Either that or there should be non-trivial comments in the new file justifying the terminology.

Agreed, 'Strict' is frustratingly underdetermined because so overloaded... so a better name would be welcome!

JacquesCarette:
`git blame` shows that @alexarice and @MatthewDaggitt were the most involved in those changes. Perhaps they should chyme in as to names?

alexarice (approved):

alexarice:
On the naming, I feel I originally introduced these with the opposite naming, i.e. the ones that are now called strict are the non-strict ones, and the ones that are now the "normal" ones are strict ones, so maybe this terminology is confusing. I'm afraid I don't have a better suggestion though

jamesmckinna:
I confess I don't have the requisite muscle-memory as to terminology here... are these `Split` notions, in which case `Function.Definitions.Split` could be ... better? UPDATED: no, `Split` notions also require `cong` for the relevant gadgets.

`OnTheNose`? That's terrible, but in practice we no longer really use any of the `Strict` machinery, do we? Only in these modules:
* `Algebra.Morphism.Construct.Terminal` uses `StrictlySurjective`
* `Data.Fin.Permutation` uses `StrictlyInverseˡ` and `StrictlyInverseʳ`
* `Data.Product.Function.Dependent.Setoid` uses `StrictlySurjective`, `StrictlyInverseˡ`, and `StrictlyInverseʳ`; `Propositional` uses only `StrictlySurjective`
* `Data.Sum.Function.Setoid` uses only `StrictlySurjective`
* `Function.Consequences{.*}`
* `Function.Structures`
* `Function.Structures.Biased`

Arguably, we can eliminate `StrictlySurjective` altogether, via
```agda
strictlySurjective⇒strictlyInverseˡ : (surj : StrictlySurjective ≈₂ f) →
                                      StrictlyInverseˡ ≈₂ f (proj₁ ∘ surj)
strictlySurjective⇒strictlyInverseˡ surj = proj₂ ∘ surj

strictlyInverseˡ⇒strictlySurjective : StrictlyInverseˡ ≈₂ f f⁻¹ →
                                      StrictlySurjective ≈₂ f
strictlyInverseˡ⇒strictlySurjective {f⁻¹ = f⁻¹} sinv y = f⁻¹ y , sinv _
```
but... legacy!?

jamesmckinna:
OK, I went back to #1156 so I get why @alexarice might not want `Strict` as the module name... and I agree, but absent better suggestions, suggest refactoring so that we have
* `Function.Definitions.Strictly` (*sic*)
* in which `StrictlySurjective` etc. simply get defined as `Surjective`
* and in `Consequences` etc., we import the new module qualified as `Strictly`, so that `StrictlySurjective` would in fact now be `Strictly.Surjective` etc. (so: deprecations are in order!)

Then, downstream, we can make one choice about what we should write instead of `Strictly` and rename accordingly.

UPDATED what about `NonCompositional`!?

JacquesCarette:
I think we need to think harder. We need to understand what mathematical / computing concept it is we're capturing in the `strictly*` versions (especially in contrast with the 'other' versions of surjectivity). I wonder if we should ask on the category theory Zulip?

jamesmckinna:
> I think we need to think harder. We need to understand what mathematical / computing concept it is we're capturing in the `strictly*` versions (especially in contrast with the 'other' versions of surjectivity). I wonder if we should ask on the category theory Zulip?

@JacquesCarette I was a bit surprised to read this. I think (sic) that the hard thinking was done in response to #1156 and associated PRs by Alex and others:
* the current so-called `Strictly*` definitions enjoy poor compositionality in the absence of `Congruent` properties (so of course, fine in the `Propositional` case)
* the others in `Function.Definitions` *do*, moreover strictly/on-the-nose wrt the proofs of such things, so they are *definitely* better to work with
* much more minor, but still the source of my gripe/peeve: the associated parametrisation (or lack of it) of the module actually worsens the ergonomics of using these definitions, because the main set are parametrised on two relations, and the `Strictly*` only on one, so we don't get the same top-level behaviour as we do with `Algebra.Definitions`/`Algebra.Structures` etc.

The aim of this PR, as @MatthewDaggitt observed earlier in its history, is to break up `Function.Definitions`, in order to allow each set of definitions to have the requisite top-level parameterisation, plus:
* sorting out `Function.Consequences` wrt implicit/explicit, because at present the UX there is pretty horrible as well

This was intended to be 'pure plumbing', rather than anything more consequential in terms of semantics. By my most recent (uncommitted) experiments now incline to a slightly more consequential conclusion: that the `Strictly*` definitions not only should be put in a module on their own, but that we should even deprecate it, or else not export it directly as part of `Function`, and strive to avoid adding new lemmas proving anything of the form `f (g x) ≈ ...` *directly*.

It's possible that I have been over-thinking things/going down rabbit-holes, but in this case, I think that you might be!

jamesmckinna:
Oh, and in any case, I would rather merge the non-DRAFT #2583 and #2599 first... these *do* have semantic/ergonomic consequences! In the case of #2583 , *precisely* as a result of the shift to the #1156 -new definition of `Surjective`... it just took along time for those consequences to shake out!

UPDATED: 2026-07-13 now merged, so let's see what's what...

JacquesCarette:
Sorry if I wasn't clear: all I mean is that we ought to find a better name than 'Strict' and 'strictly' *or* convince ourselves (with a reasoning that we can write down as documentation ) why it is a good name.

The idea of this refactor is otherwise very good.

jamesmckinna:
> Sorry if I wasn't clear: all I mean is that we ought to find a better name than 'Strict' and 'strictly' _or_ convince ourselves (with a reasoning that we can write down as documentation ) why it is a good name.

Further suggestions (as preferable to `OnTheNose`;-) ...):
* `Pointwise`
* `Local`
* or else, if we take my hint about above about deprecating `StrictlySurjective`, we could 'just' consider the `StrictlyInverse*` definitions, and call it `StrictInverses`, or `PointwiseInverses` or something... (starting to get a bit unwieldy...)

Or even:
* deprecate the module altogether and bite the bullet of refactoring the above 5 non-`Function` modules, and in line the relevant lemmas in-place in the `Function` hierarchy?

jamesmckinna:
Refactored to isolate appeals to `Function.Definitions.Strictly`: consequences external to `Function.*` hierarchy being
* `Data.Fin.Permutation`: looks fiddly to refactor, but obviously 'possible'
* `Data.List.Relation.Unary.Enumerates.Setoid.Properties`: one use, easily refactored
* `Data.Product.Function.Dependent.Setoid`: appeal to `Strictly` definitions could be inlined for subsequent more aggressive refactoring, given that they are only used as staging posts on the way to suitable `*Inverse` properties
* `Data.Sum.Function.Setoid`: ditto., but simpler: `⊎-surjective` gets used, and `⊎-strictlySurjective` is there for 'completeness`
* `Algebra.Morphism.Construct.Terminal` could easily be replaced with an appeal to `Surjective`, and some (small?) downstream cost

Accordingly, putting this back on the review stack, with idea for a downstream refactoring to be able to deprecate `Strictly` altogether, after enough plumbing/scaffolding has been put in place to make this work?

UPDATED: a by product of the refactoring is to have made `Function.Consequences` needlessly complicated wrt anonymous `module` structure capturing the already quite complicated implicit/explicit dependencies, which as previously observed, don't quite work 'right'. It may be easier simply to revert this particular module back to its original, but I somehow hoped/imagined that this process would clarify what we do/don't need from `Consequences`, as well as how 'best' to parametrise it. But I don't seem to have quite got there yet!

DONE: `CHANGELOG`!

JacquesCarette:
> deprecating `StrictlySurjective`

I was going to say that I use these in my code -- but that's not true, it is `strictlyInverse` (left and right) that I do use a fair bit. They 'fit' the proof obligations more nicely in my context.

jamesmckinna:
> > deprecating `StrictlySurjective`
> 
> I was going to say that I use these in my code -- but that's not true, it is `strictlyInverse` (left and right) that I do use a fair bit. They 'fit' the proof obligations more nicely in my context.

OK, well deprecations had better wait for v4.0 then ;-)

In the meantime, I think I'd better stop fiddling with this PR. 

Question: what should I call this wretched subsidiary module!? `Pointwise` seems the best of what I've come up with so far...

PR #2569, opened 2025-01-28
[ refactor ] fixes #2568; proves full symmetry for `Bijection`

### LINKED ISSUE #2568: [ DRY ] Refactor `Function.*` to rationalise the existence of a section to a given `Surjective f`
Lots of repetition in the hierarchy of various `private` definitions of a (not-necessarily `Congruent`)  `section : B → A` to a given `Surjective f` for `f : A → B`, which should be rationalised into an appropriate eg. manifest field of the various `Structures` and `Bundles`...

Issues:
* is a property, derived from `Function.Definitions.Surjective`?
* is it a manifest field of `Function.Structures.IsSurjection` cf. `IsGroup`
* is it a manifest field of `Function.Bundles.Surjection`

plus
* naming: `section` (neutral?), `to⁻` (as now), or `from` emulating usage already present in other `Structures` with already a well-defined section, moreover `Congruent`?
* ...

jamesmckinna:
A comprehensive overhaul building on the proposed refactoring in #2568 . Finally finished, I think!

UPDATED: now all 'done', and a lot tighter/smoother/cleaner
* refactor `Function.Consequences`? (check implicit/explicit bindings of lemmas there, esp. in their use later)
* BUT: *full* symmetry for `Bijection` now proved, old weaker form `sym-≡` etc. now deprecated (this was a `bug`!)
* AND can prove `Congruent ≈₂ ≈₁ section` outright under similar equational assumptions!
* HENCE: in `Function.Construct.Symmetry`
  - indeed, can remove hypothesis of `Congruent ≈₁ ≈₂ f` from `{injective|bijective}`
  - hence, symmetry of `IsBijective`, and of `Bijection`, *without* detour via `Inverse`

TODO:
- [x] `CHANGELOG`
- [x] refactor `Function.Construct.Symmetry` `Function.Properties.Bijection` to prove full symmetry of `IsBijective` etc.

NB.
* the deprecations considered in #2568 are optional, it's more a case of agreeing the 'right' names; existing commits make the change so as to highlight where the work has happened
* the other changes above *do* necessitate deprecations, of the former, weaker (!), proofs
* these are badged as v2.3 deprecations in favour of `*WithoutCongruence` constructions, but a v3.0 cleanup PR would remove the old proofs entirely and rename to remove that suffix
* not entirely sure I've quite got the deprecation messages correct, given how gnarly the refactoring has been.... it's typically not a straightforward like-for-like deprecation, because the types have changed
* likewise, there may still be one or two outstanding DRY opportunities for further refactoring, but downstream!


jamesmckinna:
Ooops! Need to fix the version number in sprechstimme warnings. Will fix tomorrow.

src/Function/Consequences.agda
```diff
@@ -112,3 +113,44 @@ strictlyInverseʳ⇒inverseʳ : Transitive ≈₁ →
...
+  module _ (refl : Reflexive ≈₁) where
+
+    strictlySurjective : StrictlySurjective ≈₂ f
+    strictlySurjective = surjective⇒strictlySurjective ≈₂ refl surj
+
+    strictlyInverseˡ : StrictlyInverseˡ ≈₂ f section
+    strictlyInverseˡ _ = inverseˡ refl
+
+    strictlyInverseʳ : StrictlyInverseʳ ≈₂ section f
+    strictlyInverseʳ _ = inverseˡ refl
```
jamesmckinna, line 137: Is this the right definition to export?
Or do we want, later, to have `strictlyInverseʳ : StrictlyInverseʳ ≈₂ f section` as a consequence of `inverseʳ : Transitive ≈₂ → Inverseʳ ≈₁ ≈₂ f section`?

src/Function/Structures.agda
```diff
@@ -123,7 +139,16 @@ record IsRightInverse (to : A → B) (from : B → A) : Set (a ⊔ b ⊔ ℓ₁
     renaming (cong to to-cong)
 
   strictlyInverseʳ : StrictlyInverseʳ _≈₁_ to from
-  strictlyInverseʳ x = inverseʳ Eq₂.refl
+  strictlyInverseʳ = inverseʳ⇒strictlyInverseʳ _≈₁_ _≈₂_ Eq₂.refl inverseʳ
+
+  injective : Injective _≈₁_ _≈₂_ to
+  injective = inverseʳ⇒injective {f⁻¹ = from} _≈₂_ to Eq₂.refl Eq₁.sym Eq₁.trans inverseʳ
```
jamesmckinna, line 145: Yuk! Can this not be refactored/improved in some way?
It certainly seems to hint that the parameterisation in `Function.Consequences` may be suboptimal?

jamesmckinna: Most recent commit at least managed to remove `{f⁻¹ = from}`...

CHANGELOG.md
```diff
@@ -85,3 +116,59 @@ Additions to existing modules
...
+  ```agda
+  section          : B → A
+  inverseˡ         : Inverseˡ _≈₁_ _≈₂_ to section
+  strictlyInverseˡ : StrictlyInverseˡ _≈₂_ to section
+  ```
+
+* In `Function.Consequences`: the theory of the left inverse of a surjective function
+  ```agda
+  module Section (surj :  Surjective ≈₁ ≈₂ f)
+  ```
```
jamesmckinna, line 152: Should this module be lifted out from `Consequences` to the top-level of `Function.*`? Or somewhere else?

src/Function/Structures.agda
jamesmckinna: Probably the best solution is to re-instantiate `module Section` in `Function.Consequences.Setoid` so that all the equality axioms are plumbed-in in place, and then import *that* in `Function.Structures` and `Function.Bundles`... d'oh!

src/Function/Consequences.agda
```diff
@@ -112,3 +113,44 @@ strictlyInverseʳ⇒inverseʳ : Transitive ≈₁ →
...
 strictlyInverseʳ⇒inverseʳ trans cong sinv {x} y≈f⁻¹x =
   trans (cong y≈f⁻¹x) (sinv x)
+
+------------------------------------------------------------------------
+-- Theory of the section of a Surjective function
+
+module Section (≈₂ : Rel B ℓ₂) (surj :  Surjective {A = A} ≈₁ ≈₂ f) where
+
+  section : B → A
+  section = proj₁ ∘ surj
```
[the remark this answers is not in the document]
JacquesCarette: Not only is 'section' occult, it is also used internally by Agda to give a name to (pieces of) sections that are otherwise anonymous. So error messages in actual usage could be extremely confusing. I'm definitely in the `from` camp.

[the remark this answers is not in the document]
jamesmckinna: Ugh! which is prior, haskell or math? Oh dear, haskell it seems ;-)
That said, I don't mind 'conventional' usage...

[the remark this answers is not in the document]
jamesmckinna: Double ugh! the declaration of `f⁻¹` as a `variable` in `Function.Consequences` means it can't be used neutrally as a name in `Function.Consequences`, whereas we can use `from` as a `renaming` of `section` in `Function.Structures` and `Function.Bundles`, which is probably, in the end, what matters most?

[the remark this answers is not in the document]
jamesmckinna: WIP on relevant commit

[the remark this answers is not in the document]
jamesmckinna: All that said, the comment in `Function.Bundles` on `SplitSurjection` nevertheless refers to this map as a 'section'... but only in the comments ;-)

src/Function/Construct/Symmetry.agda
```diff
@@ -31,70 +33,40 @@ private
...
-  inverseʳ inv = inv
+  inverseʳ = id
 
   inverseˡ : Inverseʳ ≈₁ ≈₂ f f⁻¹ → Inverseˡ ≈₂ ≈₁ f⁻¹ f
-  inverseˡ inv = inv
+  inverseˡ = id
 
   inverseᵇ : Inverseᵇ ≈₁ ≈₂ f f⁻¹ → Inverseᵇ ≈₂ ≈₁ f⁻¹ f
-  inverseᵇ (invˡ , invʳ) = (invʳ , invˡ)
+  inverseᵇ = swap
```
jamesmckinna, line 46: It's tempting to want to deprecate/remove this whole section, else move to `Function.Consequences`.
In any case, the use of the lemmas in this module has been inlined away, against such a move.

CHANGELOG.md
jamesmckinna: Given that the only property used in the `Section` module is `surjective⇒strictlySurjective`, it is tempting to inline that, against lifting out the whole module, and deprecating `Function.Consequences.*` entirely, as those properties are nowhere else used in `stdlib`? ... downstream PR... ?

src/Function/Bundles.agda
```diff
@@ -20,16 +20,16 @@
 module Function.Bundles where
 
 open import Function.Base using (_∘_)
+open import Function.Consequences.Propositional
+  using (strictlySurjective⇒surjective; strictlyInverseˡ⇒inverseˡ; strictlyInverseʳ⇒inverseʳ)
```
jamesmckinna, line 24: It is tempting to want to inline these, as in the `Propositional` case, they take on a particularly simple form which is obscured by the three levels of indirection in these definitions...

CHANGELOG.md
jamesmckinna: Hmmm... not 'nowhere':
* `Data.Product.Function.Dependent.Setoid`
* `Data.Nat.Binary.Properties`

src/Function/Consequences.agda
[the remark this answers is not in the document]
JacquesCarette: "declaration of `f⁻¹` as a `variable`": I was going to say that these are 'purely local', until I remembered that they can end up in the implicits list, and thus on the API. Indeed, for consistency, I like the rename to `from`.

src/Function/Bundles.agda
JacquesCarette: inline these where?

src/Function/Consequences.agda
```diff
@@ -23,22 +24,22 @@ private
     a b ℓ₁ ℓ₂ : Level
     A B : Set a
     ≈₁ ≈₂ : Rel A ℓ₁
-    f f⁻¹ : A → B
+    to f f⁻¹ : A → B
```
JacquesCarette, line 27: I understand that the way variables work, this declaration of `f` and `f⁻¹` will actually have the 'right' things in practice, I find the declaration here confusion, i.e. I would prefer an additional line with `f⁻¹ : B → A` for clarity.

And I thought you'd rename `f` to `to` and `f⁻¹` to `from` everywhere?

JacquesCarette (changes requested):
Spectacular work. Just one question and one nitpick.

src/Function/Consequences.agda
jamesmckinna: Re: inlining. Two things about this:
* I'd thought these lemmas weren't used quite as much as they are (I was really focused on their use in `Function.*`), so the case for inlining is less strong now
* that said, the `Function.Consequences.Propositional` version probably *should* be inlined/redefined in that module, because in that setting, the way `cong` and `trans` are defined for `_≡_` means that they take a *much* simpler form than that inheriting from `Setoid`...

Re: `f` and `f⁻¹`: I'd though to have `to` and `from` for the *new* uses (and not simply those in `Structures`/`Bundles`), but I'm not really a fan of 'gratuitous' alpha-conversion elsewhere, if only for the sake of keeping the footprint of the PR smaller. But I could be persuaded... ;-)

src/Function/Bundles.agda
jamesmckinna: Firstly, in `Function.Consequences.Propositional` (see latest commit), but one could even imagine inlining these definitions (given how simple they are) at any call-site...?

jamesmckinna: Last words on this:
* I think it is better, given that the types of the corresponding lemmas in `Function.Consequences.Propositional` *differ* from those in `Function.Consequences{.Setoid}` (by making `f` explicit), we can, moreover should, feel free to reimplement them, as is now done; 
* the explicit quantification seems necessary in the handful of call-sites where these lemmas are actually used (and with enough of them, that I no longer propose inlining them there)...
* ... BUT is wrong for `strictlyInverseʳ⇒inverseʳ` where it is `f⁻¹` which should be supplied explicitly, and not `f`! Fixing this should form part of #2570

src/Function/Consequences.agda
```diff
@@ -81,6 +82,23 @@ strictlySurjective⇒surjective : Transitive ≈₂ →
 strictlySurjective⇒surjective trans cong surj x =
   Product.map₂ (λ fy≈x z≈y → trans (cong z≈y) fy≈x) (surj x)
 
+module IsSurjective {f : A → B} (surjective :  Surjective ≈₁ ≈₂ f) where
```
MatthewDaggitt: Named modules like this are usually an indication that they deserve their own top-level module? Do we really need a named module?

src/Function/Construct/Symmetry.agda
```diff
@@ -35,21 +37,26 @@ module _ {≈₁ : Rel A ℓ₁} {≈₂ : Rel B ℓ₂} {f : A → B}
...
-    f⁻¹      = proj₁ ∘ surj
-    f∘f⁻¹≡id = proj₂ ∘ surj
+  open IsSurjective {≈₂ = ≈₂} surj
+    using (section; section-strictInverseˡ)
+
+  module _ (refl : Reflexive ≈₁) where
+
+    private
+
+      f∘section≡id = section-strictInverseˡ refl
```
MatthewDaggitt: This seems to be a large increase in both lines and complexity of code for comparatively little benefit down below?

src/Function/Properties/Bijection.agda
```diff
@@ -30,16 +31,33 @@ private
...
 refl = Identity.bijection _
 
--- Can't prove full symmetry as we have no proof that the witness
--- produced by the surjection proof preserves equality
-sym-≡ : Bijection S (setoid B) → Bijection (setoid B) S
-sym-≡ = Symmetry.bijection-≡
+-- Now we *can* prove full symmetry as we now have a proof that
+-- the witness produced by the surjection proof preserves equality
+sym : Bijection S T → Bijection T S
+sym = Inverse⇒Bijection ∘ Symmetry.inverse ∘ Bijection⇒Inverse
```
MatthewDaggitt: This change is great though!

jamesmckinna: This refers to an out-of-date commit now?

src/Function/Construct/Symmetry.agda
jamesmckinna: This refers to an out-of-date commit now?

src/Function/Properties/Bijection.agda
jamesmckinna: I had first managed to prove symmetry by going via `Inverse`, as above, but the current state of the PR now proves `sym` *directly*, thanks to `injective` `Surjection`s automatically satisfying `Congruent from`...

src/Function/Consequences.agda
jamesmckinna: So... I've had time to sleep on this, and I'm coming round to your way of thinking, or at least, some way towards it:
* initially, I followed the path of 
  - start with properties of an arbitrary `Surjective f` relative to arbitrary 'equalities', in `Function.Consequences`
  - specialise to those equalities being derived from `Setoid`s, in `Function.Consequences.Setoid`
  - `import` those in `Function.Structures`
  - and in `Function.Bundles`
  - deploy in `Function.Construct.Symmetry`
  (Of course, there's a lot of post-hoc rationalisation involved here... but the layering of abstractions did seem at least congruent with other existing development paths in `stdlib`, so I'm not going to litigate this idealised design process further)

To your question about separate top-level modules, I suppose we could now say: only the last three points of the list above matter, and the only real deployment of module `Section` is in `Function.Structures.{IsSurjection|IsBijection}`, notwithstanding the deprecations in `Function.Construct.Symmetry`.

So a possible refactoring ahead of merging might be to lift all of `Function.Consequences.Setoid.Section` directly into `Function.Structures.{IsSurjection|IsBijection}` (and hence no real need for `Function.Consequences.Section` either). 

Is that more what you had in mind?

jamesmckinna:
Summarising the discussion above:
* if we want the new definitions to be inlined in `Function.Structures` and `Function.Bundles`, then `Function.Construct.Symmetry` really can't have the extensive suite of deprecations without excessive duplication... so maybe that's sufficiently `breaking` to make this v3.0, unless we argue that we are fixing a 'bug' with non-backwards-compatible changes...
* if we want a fully backwards-compatible v2.3 approach, then the current architecture developed so far is the way to go.

Let me know how you see this going forward!

jamesmckinna:
I'll only return to the merge conflicts once we choose between this v2.3 version, and my (now-) preferred `breaking` (but *much* cleaner) #2583

jamesmckinna:
After much consideration, I'm going to withdraw this explicitly conservative/v2.3 version in favour of the more radical, yet more streamlined, v3.0 version #2583

Issue #2502, opened 2024-11-22
Refactor `Algebra.Consequences.*`?


jamesmckinna:
These modules are structured around:
* `Base`, depending only on the underlying `Carrier` set `A`, whose properties make explicit a dependency on a(n equality) relation `_≈_ : Rel A ℓ`;
* `Setoid`, depending on a `Setoid`, giving rise to a relation as in `Base`, but with the `IsEquivalence`/`Reasoning` properties in scope;
* `Propositional`, essential being `Setoid` instantiated to `Relation.Binary.PropositionalEquality.Properties.setoid A`

and are used in various places to augment the signature of both specific `Algebra.Structures` (esp. `IsGroup`, `Is*Ring*`), and also various `Algebra.Construct.*`.

@Taneb 's recent #2499 draws attention to the fact that this organisation privileges `Bundle`d `Setoid`s over their underlying `Relation.Binary.Structures.IsEquivalence` fields... and thus violates the (implicit) dependency restriction of `Structures` only on other `Structures`, and not on `Bundles`...

So: as a `hypothetical-rewrite` or `v3.0` refactoring, we should (perhaps!?) consider refactoring to ... sort this out. Discussion welcome!

jamesmckinna:
On further investigation:
* the admissible properties `Algebra.Structure.IsGroup.uniqueˡ-⁻¹` and `Algebra.Structure.IsGroup.uniqueʳ-⁻¹`, which depend on `Algebra.Consequences.Setoid` (and hence on a public export of `setoid`), can de deprecated in favour of `Algebra.Properties.Group.inverseˡ-unique` and `Algebra.Properties.Group.inverseʳ-unique` respectively; (so: perhaps/probably should be?)
* either way, those properties are nowhere used except to be (renamed and) re-exported by the `IsAbelianGroup` substructures in `Algebra.Module.Structures`, so presumably those re-exports can be handled by refactoring/reorganising in like fashion via `Algebra.Module.Properties.*`?
* moreover, the properties are never actually used, *except* (!!!) in `deprecated` definitions under `Algebra.Morphism`, specifically those which exhibit that it suffices to define a `IsGroupHomomorphism` via a single `IsMonoidHomomorphism` field, because the inverse homomorphism property is admissible... which surely should either be refactored as a `Biased` smart constructor, or else, now that homomorphisms are defined relative to the underlying `Raw` bundles... set aside entirely (cf. the discussion of admissible properties in #2383 ...)

[...]

PR #2571, opened 2025-02-03
[ refactor ] deprecate `Algebra.Structures.IsGroup.{uniqueˡ-⁻¹|uniqueʳ-⁻¹}` with knock-ons for `Algebra.Module.*`

jamesmckinna:
This is a v2.3-compatible non-`breaking` fix for the `Group` uniqueness of inverses part of #2502 , avoiding the larger-scale v3.0 structural changes proposed there.

NB. the wrinkle is that an `Bundles.AbelianGroup` has to be created on the fly in order to exploit its `Properties`, but that can (perhaps!?) be eliminated as part of a downstream refactoring. UPDATED have now managed to eliminate this thanks to the prompt from @MatthewDaggitt below!


JacquesCarette (approved):

src/Algebra/Module/Structures.agda
```diff
@@ -210,10 +211,12 @@ module _ (ring : Ring r ℓr)
       ( isGroup    to +ᴹ-isGroup
       ; inverseˡ   to -ᴹ‿inverseˡ
       ; inverseʳ   to -ᴹ‿inverseʳ
-      ; uniqueˡ-⁻¹ to uniqueˡ‿-ᴹ
-      ; uniqueʳ-⁻¹ to uniqueʳ‿-ᴹ
       )
 
+    open AbelianGroupProperties record { isAbelianGroup = +ᴹ-isAbelianGroup } public
+      using () renaming (inverseˡ-unique to uniqueˡ‿-ᴹ; inverseʳ-unique to uniqueʳ‿-ᴹ)
```
MatthewDaggitt, line 217: If `uniqueˡ-⁻¹` didn't belong in `Algebra.Structures`, then morally these shouldn't belong in `Algebra.Module.Structures`? If the `Properties` file already exists then they could be transferred there? Otherwise we could leave a comment?

MatthewDaggitt (approved):

src/Algebra/Module/Structures.agda
jamesmckinna: Thanks! I think this represents a wobble/failure of nerve on my part, while *not* trying to solve the problem of violating the meta-level principle '`Structures` shouldn't depend on `Bundles`'.

But in a certain sense that is *already* a bigger problem: that `Algebra.Properties.X` shouldn't be parameterised on a `Bundle` `x : X, but rather on the underlying `Structure` `isX : IsX` which supplies the 'axiomatisation' of `X`s...

jamesmckinna: But on that basis, I'd be happy to put this back in `DRAFT`, or even to close this PR, against a wider rethink of these issues...

MatthewDaggitt: No, no I think that this is an improvement over what we currently have.

jamesmckinna: OK, I'll see what/how best to tackle the above, and then deal with the outstanding merge conflict.

jamesmckinna: So... revisiting this now, I'd happily move the definitions entirely to `Algebra.Module.Properties.*module`, but for the fact that it would mean leaving the names as they are currently in-place, and duplicating them in `Algebra.Module.Properties` (so introducing a possible bug on `import`), because the `renaming` actually chooses the 'correct' names.

For `Group`, the deprecated names aren't identical to the 'correct' names in `Algebra.Properties.Group`, so the issue doesn't arise, and I can't really face introduce 'fake' names for `Module` simply for the sake of specialising the deprecation in-place in order to point to the 'correct' version elsewhere. 

Or is that the 'right' thing to do?

jamesmckinna:
Latest commit https://github.com/agda/agda-stdlib/commit/e12f94278055356a1aa747176410a4b1cf4bc54b updates the PR to propagate 'corresponding' changes to Algebra.Module.Structures and Algebra.Module.Properties.*, with suitable additional modules in the latter hierarchy, and 'corresponding' name choices for `Module` as for `Group`...
... accordingly, re-requesting a review from @MatthewDaggitt

src/Algebra/Module/Properties.agda
```diff
@@ -18,7 +18,8 @@ module Algebra.Module.Properties
 
 open Module mod
 open import Algebra.Module.Properties.Semimodule semimodule public
-open import Algebra.Properties.Group using (⁻¹-involutive)
-
--ᴹ-involutive : Involutive _≈ᴹ_ -ᴹ_
--ᴹ-involutive = ⁻¹-involutive +ᴹ-group
+open import Algebra.Module.Properties.LeftModule leftModule public
+  using (identityˡ-uniqueᴹ; identityʳ-uniqueᴹ)
```
MatthewDaggitt, line 22: This API change isn't listed in the CHANGELOG?

MatthewDaggitt (approved):
Apart from my single comment, looks good!

src/Algebra/Module/Properties.agda
jamesmckinna: Good catch! Thanks.

src/Algebra/Structures.agda
```diff
@@ -323,10 +323,18 @@ record IsGroup (_∙_ : Op₂ A) (ε : A) (_⁻¹ : Op₁ A) : Set (a ⊔ ℓ) w
...
+  Please use Algebra.Properties.Group.inverseˡ-unique instead. "
+  #-}
 
   uniqueʳ-⁻¹ : ∀ x y → (x ∙ y) ≈ ε → y ≈ (x ⁻¹)
   uniqueʳ-⁻¹ = Consequences.assoc∧id∧invˡ⇒invʳ-unique
                 setoid ∙-cong assoc identity inverseˡ
+  {-# WARNING_ON_USAGE uniqueʳ-⁻¹
+  "Warning: uniqueʳ-⁻¹ was deprecated in v2.3.
+  Please use Algebra.Properties.Group.inverseʳ-unique instead. "
+  #-}
```
jamesmckinna, line 337: Aaargh. Switching to DRAFT while I figure whether the types of the new versions are even correct!

jamesmckinna: The above *are* fine, but not so for the new additions under `Algebra.Module.*`. Now FIXED (I hope)!

jamesmckinna:
Phew, I *think* that's all fixed now.

PR #2572, opened 2025-02-06
[ refactor ] Add equality as a parameter to `Algebra.Consequences.Base`

jamesmckinna:
See [this comment](https://github.com/agda/agda-stdlib/issues/2502#issuecomment-2613113410) on issue #2502 .

By analogy with `Algebra.Definitions` and `Algebra.Structures`... moreover `non-breaking`.
Indeed, `Base` is only imported by `Algebra.Consequences.Setoid` (and by the now-deprecated module `Algebra.FunctionProperties.Consequences.Core`), so there should be *no* knock-on consequences (sic) for client modules.

PR #2573, opened 2025-02-06
[ refactor ] Revise definitions, consequences, and use, of `Algebra.Definitions.(Almost)*Cancellative`

### LINKED ISSUE #1436: Standardise implicit arguments of cancellation properties.

jamesmckinna:
Fixes #1436 , ...
... *at least as far as 'AlmostCancellative' properties of operations wrt **equality** relation; the various order-theoretic generalisations are **not** covered*
... so the issue probably ought to be kept open even if this gets merged? cf. #2580 

NB.:
* see issue for discussion of the design strategy (could be less `breaking`? `naming` issues... etc.)
* `Algebra.Consequences.Base` refactoring #2572 should probably be merged first; fold in the `instance`-based analysis for `Decidable` (and hence `Recomputable`) predicates into `Except_-*Cancellative_`?
* only equality-based properties tackled so far... `Algebra.Definitions` only imported at the equality relation, but might better be left parameterised, so that the various `Cancellative` properties can be made generic...
* for `Integer` and `Rational`, the `Cancellative` properties wrt multiplication don't easily cash out via the 'generic' design strategy of going via `AlmostCancellative`, so I haven't (yet) include such statements, nor refactored to make these the primary notions.


src/Algebra/Consequences/Base.agda
```diff
@@ -13,14 +13,48 @@ module Algebra.Consequences.Base
...
+                   Except_-LeftCancellative_ _≈_ P _•_
+  almost⇒exceptˡ cancelˡ {x} with cancelˡ x
+  ... | inj₁ px     = contradiction px
+  ... | inj₂ cancel = const cancel
+
+  almost⇒exceptʳ : _-AlmostRightCancellative_ _≈_ P _•_ →
+                   Except_-RightCancellative_ _≈_ P _•_
+  almost⇒exceptʳ cancelʳ {x} with cancelʳ x
+  ... | inj₁ px     = contradiction px
+  ... | inj₂ cancel = const cancel
```
jamesmckinna, line 41: Moderately sure that @JacquesCarette would prefer a definition in terms of `Data.Sum.Base.[_,_]` ;-)
Next iteration...

src/Algebra/Consequences/Base.agda
```diff
@@ -13,14 +13,48 @@ module Algebra.Consequences.Base
...
+                   _-AlmostLeftCancellative_ _≈_ P _•_
+  except⇒almostˡ cancelˡ x with dec x
+  ... | yes px = inj₁ px
+  ... | no ¬px = inj₂ (cancelˡ ¬px)
+
+  except⇒almostʳ : Except_-RightCancellative_ _≈_ P _•_ →
+                   _-AlmostRightCancellative_ _≈_ P _•_
+  except⇒almostʳ cancelʳ x with dec x
+  ... | yes px = inj₁ px
+  ... | no ¬px = inj₂ (cancelʳ ¬px)
```
jamesmckinna, line 56: Indeed, I couldn't be bothered to make these lazier as per my previous strictures against `yes` and `no`... sigh ;-)

src/Algebra/Consequences/Setoid.agda
```diff
@@ -157,17 +157,19 @@ module _ {_∙_ : Op₂ A} (comm : Commutative _∙_) {e : A} where
 
   comm∧almostCancelˡ⇒almostCancelʳ : AlmostLeftCancellative e _∙_ →
                                      AlmostRightCancellative e _∙_
-  comm∧almostCancelˡ⇒almostCancelʳ cancelˡ-nonZero x y z x≉e yx≈zx =
-    cancelˡ-nonZero x y z x≉e $ begin
+  comm∧almostCancelˡ⇒almostCancelʳ almostCancelˡ x with almostCancelˡ x
+  ... | inj₁ x≈e     = inj₁ x≈e
+  ... | inj₂ cancelˡ = inj₂ λ y z yx≈zx → cancelˡ y z $ begin
```
jamesmckinna, line 162: Ditto. as above in terms of `[_,_]`?

src/Algebra/Consequences/Setoid.agda
```diff
@@ -157,17 +157,19 @@ module _ {_∙_ : Op₂ A} (comm : Commutative _∙_) {e : A} where
...
       z ∙ x ≈⟨ comm z x ⟩
       x ∙ z ∎
 
   comm∧almostCancelʳ⇒almostCancelˡ : AlmostRightCancellative e _∙_ →
                                      AlmostLeftCancellative e _∙_
-  comm∧almostCancelʳ⇒almostCancelˡ cancelʳ-nonZero x y z x≉e xy≈xz =
-    cancelʳ-nonZero x y z x≉e $ begin
+  comm∧almostCancelʳ⇒almostCancelˡ almostCancelʳ x with almostCancelʳ x
+  ... | inj₁ x≈e     = inj₁ x≈e
+  ... | inj₂ cancelʳ = inj₂ λ y z xy≈xz → cancelʳ y z $ begin
```
jamesmckinna, line 172: Ditto.

src/Algebra/Definitions.agda
```diff
@@ -140,20 +141,36 @@ SelfInverse f = ∀ {x y} → f x ≈ y → f y ≈ x
...
-RightCancellative _•_ = ∀ x y z → (y • x) ≈ (z • x) → y ≈ z
+RightCancellative _•_ = ∀ x → RightCancellativeAt x _•_
 
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+_-AlmostLeftCancellative_ Except_-LeftCancellative_ : ∀ {p} (P : Pred A p) →
+                                                      Op₂ A → Set _
+P -AlmostLeftCancellative _•_ = ∀ x → P x ⊎ LeftCancellativeAt x _•_
+Except P -LeftCancellative _•_ = ∀ {x} → ¬ (P x) → LeftCancellativeAt x _•_
```
jamesmckinna, line 162: Refactor in terms of `{{¬ (P x)}}`? 
Might be better to have `{{N x}}` with some sort of relationship between `N x` and `¬ (P x)`?

src/Algebra/Definitions.agda
```diff
@@ -140,20 +141,36 @@ SelfInverse f = ∀ {x y} → f x ≈ y → f y ≈ x
...
+
 AlmostLeftCancellative : A → Op₂ A → Set _
-AlmostLeftCancellative e _•_ = ∀ x y z → ¬ x ≈ e → (x • y) ≈ (x • z) → y ≈ z
+AlmostLeftCancellative e = (_≈ e) -AlmostLeftCancellative_
+
+_-AlmostRightCancellative_ Except_-RightCancellative_ : ∀ {p} (P : Pred A p) →
+                                                         Op₂ A → Set _
+P -AlmostRightCancellative _•_ = ∀ x → P x ⊎ RightCancellativeAt x _•_
+Except P -RightCancellative _•_ = ∀ {x} → ¬ (P x) → RightCancellativeAt x _•_
```
jamesmckinna, line 170: Ditto re. `{{N x}}`

jamesmckinna:
Latest commit rejigs the statements in terms of `instance` arguments for `Except_-*Cancellative_`, but it's not obvious to me that the ergonomics actually improve?

src/Algebra/Properties/Semiring/Primality.agda
```diff
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Some theory for CancellativeCommutativeSemiring.
+-- Properties of Coprimality and Irreducibility for Semiring.
```
jamesmckinna, line 4: This should lift out to a separate PR

JacquesCarette (commented):
Seems all quite reasonable.

jamesmckinna:
I think I can't face doing more with this PR at this stage: the new definitions are in-place, with one specimen instantiation for `Nat`, but it isn't clear to me how best to proceed with:
- [ ] the rest of `Data.Nat.Properties`
- [ ] `Data.Integer.Properties`
- [ ] `Data.Rational.Properties`

without a lot more refactoring, which I suggest take place downstream?

JacquesCarette (approved):
Seems like a nice PR as-is, more can be done subsequently.

jamesmckinna:
> Seems like a nice PR as-is, more can be done subsequently.

Thanks @JacquesCarette I suppose my concern was that if it turned out those downstream PRs were still stymied by 'bad' design decisions here, then ... I'm not sure what. Certainly it seemed a *big* job to try to refactor existing proofs to try to conform to the new APIs...

JacquesCarette:
To allay your fears, might be worth taking `Data.Rational.Properties` (in a new branch based off of this one) and porting it, but without aiming to make that port a PR, just a "is this a good idea" test.

jamesmckinna:
> To allay your fears, might be worth taking `Data.Rational.Properties` (in a new branch based off of this one) and porting it, but without aiming to make that port a PR, just a "is this a good idea" test.

So to allay the fears about whether it's too much work to make the change, I may as well (accept having to) do the additional work to make the change? Sigh...

JacquesCarette:
> So to allay the fears about whether it's too much work to make the change, I may as well (accept having to) do the additional work to make the change? Sigh...

Well, it might turn out to be easy! Or you might be right, and be able to estimate just how much work it is, and let us know.

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -44,6 +43,7 @@ import Relation.Binary.Properties.Poset as PosetProperties
 import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
 open import Relation.Binary.Reasoning.Syntax
 
+open import Algebra.Consequences.Propositional {A = ℚᵘ}
```
MatthewDaggitt, line 46: This seems like an unneeded import?

src/Data/Rational/Properties.agda
```diff
@@ -64,6 +63,7 @@ open import Relation.Nullary.Decidable.Core as Dec
   using (yes; no; recompute; map′; _×-dec_)
 open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 
+open import Algebra.Consequences.Propositional {A = ℚ}
```
MatthewDaggitt, line 66: Why do we now need to import it with the type specifically applied? Is this a breaking change?

MatthewDaggitt (changes requested):
There are lots of good changes in here, but it's difficult to parse all the different definitions of cancellative being added. I'll have another look if you can lay them out on separate lines?

doc/README.agda
```diff
@@ -3,7 +3,7 @@
 module README where
 
 ------------------------------------------------------------------------
--- The Agda standard library, version 2.3-dev
+-- The Agda standard library, version 3.0-dev
```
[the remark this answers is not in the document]
jamesmckinna: committed, not yet pushed.

CHANGELOG.md
```diff
@@ -147,3 +158,8 @@ Additions to existing modules
   ```agda
   ```
+
+* In `Data.Nat.Properties`:
+  ```agda
+  *-almostCancelʳ-≡ : AlmostRightCancellative 0 _*_
+  ```
```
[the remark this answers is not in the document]
jamesmckinna: Could very well be... I'll try to come back to this this week.

src/Algebra/Definitions.agda
```diff
@@ -140,20 +140,40 @@ SelfInverse f = ∀ {x y} → f x ≈ y → f y ≈ x
...
 
 RightCancellative : Op₂ A → Set _
-RightCancellative _•_ = ∀ x y z → (y • x) ≈ (z • x) → y ≈ z
+RightCancellative _•_ = ∀ x → RightCancellativeAt x _•_
 
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+_-AlmostLeftCancellative_ Provided_-LeftCancellative_
+  Except_-LeftCancellative_ : ∀ {p} (P : Pred A p) → Op₂ A → Set _
```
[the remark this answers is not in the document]
jamesmckinna: OK re: line separation

As for names/naming, indeed I'm not thrilled by the names, the first item on my 'issue/NB' list in the preamble.
The hyphens were some attempt to render the English-language hyphenation, but clearly not the best choice.

Suggestions for these auxiliary notions very welcome! Also the prioritisation of the `Almost` definition, with the `Except`/`Provided` ones being subordinate with the (irrelevant) `instance` arguments. In some ways, by making their names unwieldy, I was attempting to 'downgrade' them. But again: suggestions welcome!

src/Data/Rational/Unnormalised/Properties.agda
jamesmckinna: It's needed at L766:
```agda
```

src/Data/Rational/Properties.agda
jamesmckinna: And in fact, *this* one is the unneeded `import`!

CHANGELOG.md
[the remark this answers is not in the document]
jamesmckinna: So, I'm not sure what's missing, as `Data.Nat.Properties` only has this one specimen lemma added.
For `Data.Rational{.*}.Properties`, only the types of certain lemmas as part of the API change... but maybe *that* is missing?

src/Algebra/Definitions.agda
```diff
@@ -140,20 +140,42 @@ SelfInverse f = ∀ {x y} → f x ≈ y → f y ≈ x
...
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+_AlmostLeftCancellative′_    : ∀ {p} (P : Pred A p) → Op₂ A → Set _
+Provided_LeftCancellative_  : ∀ {p} (P : Pred A p) → Op₂ A → Set _
+Except_LeftCancellative_    : ∀ {p} (P : Pred A p) → Op₂ A → Set _
+
+P AlmostLeftCancellative′ _•_   = ∀ x → P x ⊎ LeftCancellativeAt x _•_
+Provided P LeftCancellative _•_ = ∀ x y z → .{{P x}} → (x • y) ≈ (x • z) → y ≈ z
+Except P LeftCancellative _•_   = Provided (∁ P) LeftCancellative _•_
```
jamesmckinna, line 164: I've put al lthe names/types on separate lines, and removed the hyphens, at the cost of needing to standardise apart `_AlmostLeftCancellative′_` and `AlmostLeftCancellative`. Further suggestions re names welcome!

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -792,8 +792,8 @@ neg⇒nonZero (mkℚᵘ (-[1+ _ ]) _) = _
 +-inverse : Inverse _≃_ 0ℚᵘ -_ _+_
 +-inverse = +-inverseˡ , +-inverseʳ
 
-+-cancelˡ : ∀ {r p q} → r + p ≃ r + q → p ≃ q
-+-cancelˡ {r} {p} {q} r+p≃r+q = begin-equality
++-cancelˡ : LeftCancellative _≃_ _+_
++-cancelˡ r p q r+p≃r+q = begin-equality
```
jamesmckinna, line 796: How should this type change be reflected in the `CHANGELOG`, given that this is now the 'original' intention?

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -805,12 +805,8 @@ neg⇒nonZero (mkℚᵘ (-[1+ _ ]) _) = _
...
   q            ∎ where open ≤-Reasoning
 
-+-cancelʳ : ∀ {r p q} → p + r ≃ q + r → p ≃ q
-+-cancelʳ {r} {p} {q} p+r≃q+r = +-cancelˡ {r} $ begin-equality
-  r + p ≃⟨ +-comm r p ⟩
-  p + r ≃⟨ p+r≃q+r ⟩
-  q + r ≃⟨ +-comm q r ⟩
-  r + q ∎ where open ≤-Reasoning
++-cancelʳ : RightCancellative _≃_ _+_
++-cancelʳ = Consequences.comm∧cancelˡ⇒cancelʳ ≃-setoid +-comm +-cancelˡ
```
jamesmckinna, line 809: Ditto.

jamesmckinna:
Besides the still-outstanding merge conflict, some improvement (I hope!) in the `CHANGELOG` documentation of the API change(s).

jamesmckinna:
OK... I might need to go back to the drawing board, a bit, now that we have, for `Data.Nat.Properties` (L927-L931):
```agda
*-cancelʳ-≡ : ∀ m n o .{{_ : NonZero o}} → m * o ≡ n * o → m ≡ n
*-cancelʳ-≡ m n o = almost⇒exceptʳ _ *-almostCancelʳ-≡ o m n {{≢-nonZero⁻¹ _}}

*-cancelˡ-≡ : ∀ m n o .{{_ : NonZero o}} → o * m ≡ o * n → m ≡ n
*-cancelˡ-≡ m n o rewrite *-comm o m | *-comm o n = *-cancelʳ-≡ m n o
```
as the *status quo ante*, and with the new definitions, instead we would have (NB: argument order!)
```agda
*-cancelʳ-≡ : Provided NonZero RightCancellative _*_
*-cancelʳ-≡ m n o = almost⇒exceptʳ _ *-almostCancelʳ-≡ m n o {{≢-nonZero⁻¹ _}}

*-cancelˡ-≡ : Provided NonZero LeftCancellative _*_
*-cancelˡ-≡ m n o rewrite *-comm m o | *-comm m n = *-cancelʳ-≡ m n o
```

So...: what *is*/should be the correct argument order in all the new definitions, or do we simply carry on breaking the API?

jamesmckinna:
Drawback 2.

Suppose we want to rewrite eg. `Data.Nat.Properties` L967-973
```agda
*-cancelʳ-≤ : ∀ m n o .{{_ : NonZero o}} → m * o ≤ n * o → m ≤ n
*-cancelʳ-≤ zero    _       _         _  = z≤n
*-cancelʳ-≤ (suc m) (suc n) o@(suc _) le =
  s≤s (*-cancelʳ-≤ m n o (+-cancelˡ-≤ _ _ _ le))

*-cancelˡ-≤ : ∀ o .{{_ : NonZero o}} → o * m ≤ o * n → m ≤ n
*-cancelˡ-≤ {m} {n} o rewrite *-comm o m | *-comm o n = *-cancelʳ-≤ m n o
```
to use the new definitions, then, modulo Drawback 1., we would (be able to) have:
```agda
*-cancelʳ-≡ : Provided NonZero RightCancellative _*_
*-cancelˡ-≡ : Provided NonZero LeftCancellative _*_
```
modulo fiddling the imports so that we replace `_≡_` with `_≤_` in the definitions... so far so good.

BUT: for `Data.Rational.Properties` L1360, we can't rewrite:
```agda
*-cancelʳ-≤-neg : ∀ r .{{_ : Negative r}} → p * r ≤ q * r → p ≥ q
```
in terms of `Provided Negative RightCancellative _*_`... because the relation has flipped. So... probably, everything should move to `Relation.Binary.Definitions` to allow independent parametrisation on the two relations, and then import *those* definitions in `Algebra.Definitions` at the *same* relation. 

My brain hurts. Doable, but *exhausting*... :-(

jamesmckinna:
(Potential) Drawback 3.
Are all/any of the other potential uses of the `Provided`/`Except` family of properties expressible  as 'algebraic' instances of `Almost`, for suitable choices of the underlying 'equality' relation in terms of some `e`?

src/Data/Rational/Unnormalised/Properties.agda
JacquesCarette: I think what's in the `CHANGELOG` now reflects this change quite well.

JacquesCarette (approved):

JacquesCarette:
From the code you show, the argument order seems fine? Is there something I'm missing?

It also seems to me that it is `*-cancelʳ-≤-neg` that is weird, not what you've written. It inlines a 'relative complement' operation in its type, which isn't something generic combinators really should be expected to deal with, not unless we over-generalize.

jamesmckinna:
Re: argument order
It's more a question: do we try to keep the old order of arguments (`o m n` in the above example), or commit to the new order induced (`m n o`) by the particular 'design' of the new predicates. Otherwise put: should the new predicates be redesigned to retain the old argument order (fiddly, but doable, I think, and maybe not even sensible,  but I've lost the capacity to resolve such decisions)

Re: `*-cancelʳ-≤-neg`
I think I don't quite understand your comment? It's 'obviously' a cancellation property, but the 'flip'/generalisation in the conclusion ordering arises because of the nature of the proviso `Negative r`... so I don't quite see how the 'relative complement' plays a role?

JacquesCarette:
I'm still not quite getting it. `o m n` is a bizarre order, `m n o` is better. But that's just order of the variables. You probably mean something deeper that is clear in your head, but probably need to unwind some layers of definitions to see?

Re: `*-cancelʳ-≤-neg`. The fact that `≤` becomes `≥` in the conclusion is highly non-trivial. It's just seems trivial to use because of how long ago we've integrated the action of negatives on ordering relations into our knowledge base. The point is that this isn't 'structural' in any reasonable way, so there is no actual pattern to be leveraged here (well, there is, but it's too general to be of much use in practice, IMHO).

src/Algebra/Definitions.agda
```diff
@@ -140,20 +140,42 @@ SelfInverse f = ∀ {x y} → f x ≈ y → f y ≈ x
...
 RightCancellative : Op₂ A → Set _
-RightCancellative _•_ = ∀ x y z → (y • x) ≈ (z • x) → y ≈ z
+RightCancellative _•_ = ∀ x → RightCancellativeAt x _•_
 
 Cancellative : Op₂ A → Set _
 Cancellative _•_ = (LeftCancellative _•_) × (RightCancellative _•_)
 
+_AlmostLeftCancellative′_   : ∀ {p} (P : Pred A p) → Op₂ A → Set _
+Provided_LeftCancellative_  : ∀ {p} (P : Pred A p) → Op₂ A → Set _
+Except_LeftCancellative_    : ∀ {p} (P : Pred A p) → Op₂ A → Set _
```
MatthewDaggitt, line 160: I realise that this highlights the similarity. However we don't separate type signatures from definitions like this elsewhere in the library unless it is necessary for mutual recursion. In this PR we should probably follow the standard style?

MatthewDaggitt (approved):

jamesmckinna:
Merge conflict fixed, and so what's left
* as I understand @JacquesCarette comments above, we shouldn't try to (over-)generalise to multiple relations as part of this PR
* not yet fixed the argument order, but I'm tempted simply to not try this time around!

If you think this is mergeable, I'll let go for the time being.

JacquesCarette (approved):

JacquesCarette:
I think this PR is a good unit of progress as-is.

jamesmckinna:
Not quite sure how to go back to this after more than a year... and all the merge conflicts are definitely a headache!

jamesmckinna:
Merge conflicts are one thing, but I think I'd (now!? v3.0 etc.)  like to be more radical in making this `breaking`, by simply changing the API to make `Almost*` offer the positive/constructive/disjunctive version *only*. Comments welcome!

jamesmckinna:
Beyond considerations of what this PR might/should actually contain, I've brought it up to date, and with two approvals from @JacquesCarette and @MatthewDaggitt suggest merging now/soon and then seeing how the dust settles?

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
```diff
@@ -12,29 +12,36 @@ module Algebra.Properties.CancellativeCommutativeSemiring
...
-... | inj₁ x≈0 = x≉0 x≈0
-... | inj₂ y≈0 = y≉0 y≈0
+module _ (_≈?_ : Decidable _≈_) where
+
+  xy≈0⇒x≈0∨y≈0 : x * y ≈ 0# → x ≈ 0# ⊎ y ≈ 0#
+  xy≈0⇒x≈0∨y≈0 {x} {y} xy≈0 =
+    map₂ (λ cancel → cancel _ _ (trans xy≈0 (sym (zeroʳ x)))) (*-cancelˡ-nonZero x)
+
+  x≉0∧y≉0⇒xy≉0 : x ≉ 0# → y ≉ 0# → x * y ≉ 0#
+  x≉0∧y≉0⇒xy≉0 x≉0 y≉0 xy≈0 = [ x≉0 , y≉0 ]′ (xy≈0⇒x≈0∨y≈0 xy≈0)
```
jamesmckinna: So... we've somehow now managed to eliminate the `Decidable` module hypothesis, so it's removable, but at the cost of a `breaking` API change for this module, ... but we're in v3.0, so why not!?

UPDATED and indeed, this property appears never to be used in `stdlib`!?

JacquesCarette: Go ahead and break.

JacquesCarette:
I agree with my past self: this PR, as of today, is a worthwhile series of improvements.

The more radical version might be good too, but feels like a new step, best performed on top of this one.

src/Algebra/Properties/CancellativeCommutativeSemiring.agda
jamesmckinna: Asked, answered, and now DONE.

JacquesCarette (approved):

PR #2576, opened 2025-02-07
Rename overloaded sym in apartness

bsaul:
Currently in `IsHeytingCommutativeRing`, `sym` is overloaded as it is exported both from `IsCommutativeRing` and `IsApartnessRelation`. This PR renames the apartness sym to `#-sym`.

NOTE: `#-sym` is also removed from `Algebra.Apartness.Properties.HeytingCommutativeRing` as it is now redundant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,15 @@ Highlights
 Bug-fixes
 ---------
 
+* In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
+  to `#-sym` in order to avoid overloaded projection.
+  `irrefl` and `cotrans` are similarly renamed.
+
 Non-backwards compatible changes
 --------------------------------
 
 Minor improvements
 ------------------
-
-* In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
-  to `#-sym` in order to avoid overloaded projection.
-
 
 Deprecated modules
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Non-backwards compatible changes
 Minor improvements
 ------------------
 
+* In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
+  to `#-sym` in order to avoid overloaded projection.
+
+
 Deprecated modules
 ------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Bug-fixes
 
 * In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
   to `#-sym` in order to avoid overloaded projection.
-  `irrefl` and `cotrans` are similarly renamed.
+  `irrefl` and `cotrans` are similarly renamed for the sake of consistency.
 
 Non-backwards compatible changes
 --------------------------------
--- a/src/Algebra/Apartness/Structures.agda
+++ b/src/Algebra/Apartness/Structures.agda
@@ -34,7 +34,11 @@ record IsHeytingCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
 
   open IsCommutativeRing isCommutativeRing public
   open IsApartnessRelation isApartnessRelation public
-    renaming (sym to #-sym)
+    renaming 
+      ( irrefl  to #-irrefl
+      ; sym     to #-sym
+      ; cotrans to #-cotrans 
+      )
 
   field
     #⇒invertible : ∀ {x y} → x # y → Invertible 1# _*_ (x - y)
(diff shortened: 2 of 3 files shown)


jamesmckinna:
Hmmm... interesting. Thanks for the PR!

Ahead of a proper review:
* just removing the definition from `Algebra.Apartness.Properties.HeytingCommutativeRing` would nominally be a `breaking` change, so for v2.3 we should add a deprecation warning indicating the new source for the property; otherwise, for v3.0 we can simply make the break;
* the PR itself draws attention to the fact that `Symmetric _#_` is in fact a redundant part of the definition of `IsHeytingCommutativeRing`, so it *might* be worth having an `Algebra.Apartness.Structures.Biased` smart constructor which puts the pieces together, but this could be developed downstream;
* the whole definition of HCR now seems a bit odd (compared to HCF?): why should a `Ring` admit inverses of elements apart from `0`? As opposed to simply being a `CommutativeRing` with a `Tight` apartness relation? There might be a wider issue to raise about refactoring the whole caboodle?

cspollard:
Hi @jamesmckinna . I'm not at all opposed to having things re-arranged, especially if it can become more accessible/ergonomic and, of course, more correct.

I remember a discussion on why we should _not_ call this a `LocalRing` [eventually it became `HeytingCommutativeRing`], but I cannot find the discussion at the moment. I believe `LocalRing` had too many potential meanings and/or wasn't very precise. These structures are split into what we call `HeytingCommutativeRing` and `HeytingField` based on

I did manage to find some discussions on this in zulip from a while ago:

https://agda.zulipchat.com/#narrow/channel/264623-stdlib/topic/Fields.2C.20OrderedFields.2C.20etc.20in.20the.20standard.20library.3F

but not a direct decision on the naming convention chosen.

jamesmckinna:
Thanks @cspollard for the back pointer to the extensive prior discussion on Zulip

src/Algebra/Apartness/Structures.agda
```diff
@@ -34,6 +34,7 @@ record IsHeytingCommutativeRing : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
 
   open IsCommutativeRing isCommutativeRing public
   open IsApartnessRelation isApartnessRelation public
+    renaming (sym to #-sym)
```
MatthewDaggitt, line 37: I think we should take this opportunity to rename all three fields with a `#` to be consistent?

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ Non-backwards compatible changes
 Minor improvements
 ------------------
 
+* In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
+  to `#-sym` in order to avoid overloaded projection.
```
MatthewDaggitt, line 19: Can we move this to bug fixes section?

MatthewDaggitt (changes requested):
Otherwise, this looks good to me! @jamesmckinna raises some good points for downstream, but I don't think should stop this immediate bug fix :smile:

src/Algebra/Apartness/Structures.agda
jamesmckinna: Do you mean in-place in this `import`, or actually making the `breaking` change back in the `record` definition in `Relation.Binary.Structures.IsApartnessRelation`?

MatthewDaggitt: Good question. In this import!

jamesmckinna:
@bsaul are you happy undertaking the requested changes?

bsaul:
> @bsaul are you happy undertaking the requested changes?

Yes. Will do (probably later today).

CHANGELOG.md
bsaul: Done in d5d902bbe3cd2d80f55b59f0f9ec9ea219c776c6.

jamesmckinna (approved):
I've made suggestion by way of emphasis, but it's not a deal-breaker.
Otherwise this looks great!
(And I've stashed a copy of the old prof of `#-sym` so that it can be incorporated into a hypothetical future `Algebra.Apartness.Structures.Biased` etc.)

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Highlights
 Bug-fixes
 ---------
 
+* In `Algebra.Apartness.Structures`, renamed `sym` from `IsApartnessRelation`
+  to `#-sym` in order to avoid overloaded projection.
+  `irrefl` and `cotrans` are similarly renamed.
```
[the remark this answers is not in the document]
bsaul: The suggestion is fine by me.

JacquesCarette (approved):

jamesmckinna:
@MatthewDaggitt are you happy with @bsaul 's changes per you review? if so, let's merge!

MatthewDaggitt (approved):

PR #2579, opened 2025-02-10
[ refactor ] `Algebra.Properties.Semiring.Divisibility` and friends

jamesmckinna:
Cosmetic (non-`breaking`): : simplifies `import`s and some proofs. Lifted out from #2573 

NB. Could go further (but then would be `breaking`):
* define `NonTrivial` to mean `1# ≉ 0#` (rather than `0# ≉ 1#` as used currently), and refactor to use `NonTrivial` as hypothesis (Blast! clashes with `Data.Nat.Base.NonTrivial` with a different semantics...)
* redefine mutual divisibility using `Relation.Binary.Construct.Interior.Symmetric`, and exploit the generic properties of such relations
* ...

### DIFF
--- a/src/Algebra/Properties/Semigroup/Divisibility.agda
+++ b/src/Algebra/Properties/Semigroup/Divisibility.agda
@@ -6,12 +6,13 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra using (Semigroup)
-open import Data.Product.Base using (_,_)
-open import Relation.Binary.Definitions using (Transitive)
+open import Algebra.Bundles using (Semigroup)
 
 module Algebra.Properties.Semigroup.Divisibility
   {a ℓ} (S : Semigroup a ℓ) where
+
+open import Data.Product.Base using (_,_)
+open import Relation.Binary.Definitions using (Transitive)
 
 open Semigroup S
 
@@ -25,7 +26,7 @@ open import Algebra.Properties.Magma.Divisibility magma public
 
 ∣-trans : Transitive _∣_
 ∣-trans (p , px≈y) (q , qy≈z) =
-  q ∙ p , trans (assoc q p _) (trans (∙-congˡ px≈y) qy≈z)
+  (q ∙ p) , trans (assoc q p _) (trans (∙-congˡ px≈y) qy≈z)
 
 ------------------------------------------------------------------------
 -- Properties of _∥_
--- a/src/Algebra/Properties/Semigroup/Divisibility.agda
+++ b/src/Algebra/Properties/Semigroup/Divisibility.agda
@@ -26,7 +26,7 @@ open import Algebra.Properties.Magma.Divisibility magma public
 
 ∣-trans : Transitive _∣_
 ∣-trans (p , px≈y) (q , qy≈z) =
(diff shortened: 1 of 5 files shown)


JacquesCarette (approved):

src/Algebra/Properties/Semigroup/Divisibility.agda
```diff
@@ -25,7 +26,7 @@ open import Algebra.Properties.Magma.Divisibility magma public
 
 ∣-trans : Transitive _∣_
 ∣-trans (p , px≈y) (q , qy≈z) =
-  q ∙ p , trans (assoc q p _) (trans (∙-congˡ px≈y) qy≈z)
+  (q ∙ p) , trans (assoc q p _) (trans (∙-congˡ px≈y) qy≈z)
```
Taneb, line 29: Why did you introduce brackets here? Did something change the fixities? Or it it just for clarity?

jamesmckinna: Good catch! While thinking about how to simplify the `import`s (because divisibility is now given a custom `record` definition, rather than delegating to `Data.Product.Base.Σ`), I thought I could remove `import`s of `Data.Product.Base._,_`... doing so then causes a parse error (corrected by adding the above parentheses!), but actually can't quite be fully realised, because *mutual* divisibility again relies on `_×_`... whence my comment about redefining *that* relation in terms of `Relation.Binary.Construct.Interior.Symmetric`...

So it really draws attention to the fact that there are *several* uses of `_,_` as a `constructor`, but somehow different ones give rise to different parses, so is itself a distinct (and tricky!) issue to want to iron out maybe?

jamesmckinna: Can revert the parentheses back (actually: should!), but still think it's worth thinking about the points above.

src/Algebra/Properties/Semiring/Primality.agda
```diff
@@ -1,22 +1,31 @@
...
   {a ℓ} (R : Semiring a ℓ)
   where
 
-open Semiring R renaming (Carrier to A)
+open import Data.Sum.Base using (reduce)
+open import Function.Base using (flip)
+open import Relation.Binary.Definitions using (Symmetric)
+
+open Semiring R
+  using (_≉_; sym; trans; 0#; 1#; zeroˡ; rawSemiring)
```
MatthewDaggitt, line 20: Is this refinement of `using` worth the extra two lines? We're in `Properties.Semiring` so if we have name clashes with the remaining fields in`Semiring` I feel that its probably a bug and it would be better to detect it.

jamesmckinna: Yes, I agree. Will revert!

src/Algebra/Properties/Semigroup/Divisibility.agda
jamesmckinna: I think it may be a fixity issue: overloaded constructors are indeed fine, but I haven't checked what each individual one is declared with...

jamesmckinna:
Suggest: merging this as-is, but lifting out the `breaking` changes alluded to above as a separate issue #2582 for downstream consideration?

JacquesCarette (approved):

MatthewDaggitt (approved):

PR #2580, opened 2025-02-10
[ refactor ] Restate, and  use, the definitions of `Monotonic` etc. operations

### LINKED ISSUE #1579: Inconsistent naming between `Data.(Nat/Integer).Properties`

jamesmckinna:
Aim: towards tackling #1579 

TODO:
- [x] introduce the relevant definitions and consequences
- [x] deploy in `Data.Nat.*` UPDATED: unsolved metas suggest some more revision is necessary :-(
- [x] deploy in `Data.Integer.*`
- [x] deploy in `Data.Rational.*`
- [ ] make the `breaking` changes (v3.0: separate PR once the rest is done?)


JacquesCarette:
Thanks - these already are easier to understand.

MatthewDaggitt (approved):

jamesmckinna:
How do I remove #1579 from the `Development` sidebar entry?

MatthewDaggitt:
> How do I remove https://github.com/agda/agda-stdlib/issues/1579 from the Development sidebar entry?

Not sure how you did it, but it's gone for me :tada:

src/Relation/Binary/Consequences.agda
```diff
@@ -87,24 +87,38 @@ module _ {_≈_ : Rel A ℓ₁} {_≤_ : Rel A ℓ₂} where
...
   mono⇒cong : Symmetric ≈₁ → ≈₁ ⇒ ≤₁ → Antisymmetric ≈₂ ≤₂ →
-              ∀ {f} → f Preserves ≤₁ ⟶ ≤₂ → f Preserves ≈₁ ⟶ ≈₂
+              ∀ {f} → Monotonic₁ ≤₁ ≤₂ f → Monotonic₁ ≈₁ ≈₂ f
   mono⇒cong sym reflexive antisym mono x≈y = antisym
     (mono (reflexive x≈y))
     (mono (reflexive (sym x≈y)))
 
   antimono⇒cong : Symmetric ≈₁ → ≈₁ ⇒ ≤₁ → Antisymmetric ≈₂ ≤₂ →
-                  ∀ {f} → f Preserves ≤₁ ⟶ (flip ≤₂) → f Preserves ≈₁ ⟶ ≈₂
+                  ∀ {f} → f Preserves ≤₁ ⟶ (flip ≤₂) → Monotonic₁ ≈₁ ≈₂ f
```
jamesmckinna, line 96: This unfortunately is stated so as *not* to be able to exploit the refactoring redefinition of `Antitonic₁` in `Relation.Binary.Definitions` below (wrong relation gets `flip`ped!). Downstream refactoring?

src/Relation/Binary/Definitions.agda
```diff
@@ -155,19 +155,25 @@ Monotonic₁ : Rel A ℓ₁ → Rel B ℓ₂ → (A → B) → Set _
 Monotonic₁ _≤_ _⊑_ f = f Preserves _≤_ ⟶ _⊑_
 
 Antitonic₁ : Rel A ℓ₁ → Rel B ℓ₂ → (A → B) → Set _
-Antitonic₁ _≤_ _⊑_ f = f Preserves (flip _≤_) ⟶ _⊑_
+Antitonic₁ _≤_ = Monotonic₁ (flip _≤_)
+
+LeftMonotonic : Rel B ℓ₁ → Rel C ℓ₂ → (A → B → C) → Set _
+LeftMonotonic _≤_ _⊑_ _∙_ = ∀ {x} → Monotonic₁ _≤_ _⊑_ (x ∙_)
```
jamesmckinna, line 161: Implicit or explicit quantification over `x` here?

src/Relation/Binary/Definitions.agda
```diff
@@ -155,19 +155,25 @@ Monotonic₁ : Rel A ℓ₁ → Rel B ℓ₂ → (A → B) → Set _
...
+
+RightMonotonic : Rel A ℓ₁ → Rel C ℓ₂ → (A → B → C) → Set _
+RightMonotonic _≤_ _⊑_ _∙_ = ∀ {y} → Monotonic₁ _≤_ _⊑_ (_∙ y)
```
jamesmckinna, line 164: Ditto. `y`

jamesmckinna:
Now that the abstract definitions are mostly in place, can look at the instantiations in `Data.Nat` etc. Still questions in mind, as ever, about implicit/explicit quantification in `Left`/`RightMonotonic`, plus dependency on `Relation.Binary` adding one additional layer of indirection (cf. #2573 #2566 etc.) so a nudge from reviewers would be welcome before proceeding...

CHANGELOG.md
```diff
@@ -80,6 +80,16 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Consequences.Base`:
+  ```agda
+  module Congruence (cong : Congruent₂ _≈_ _∙_) (refl : Reflexive _≈_)
```
JacquesCarette, line 85: This is not particularly informative -- what is in that module? Same below.

src/Relation/Binary/Definitions.agda
JacquesCarette: The only real way to find out is to use it a dozen times for different things and see what comes up.

JacquesCarette (changes requested):
Coming along nicely.

src/Relation/Binary/Definitions.agda
jamesmckinna: Well, yes. I was really referring to the fact that the typical 'concrete' use that these definitions are attempting to abstract over is eg. from `Data.Nat.Properties`
```agda
+-mono-≤ : _+_ Preserves₂ _≤_ ⟶ _≤_ ⟶ _≤_
+-monoˡ-≤ : ∀ n → (_+ n) Preserves _≤_ ⟶ _≤_
+-monoʳ-≤ : ∀ n → (n +_) Preserves _≤_ ⟶ _≤_
```

The originating issue, and #1436 similarly, emphasise the need for explicit quantification for partially applied binary operators in concrete algebras, because the unifier can't infer arguments when `_∙_` is given concrete form. But the lemmas 

So there are two issues here:
* are the new definitions (and the lemmas which establish them) sufficient for the applications? (more investigation required...)
* do/should we make the `breaking` changes to the `Properties` modules if they turn out to be?

Obviously, I'm merely thinking out loud at the moment, as the PR develops, but I wanted to forward point to the issue, for comment.

CHANGELOG.md
jamesmckinna: I think I've never understood how much detail, and of what character, should be included in `CHANGELOG` when adding a new submodule:
* (show) the module contents themselves  ('types document themselves')
* (tell) informal description of what that contents is, but without detail...

Meta-question: what belongs in a `CHANGELOG` (entry) anyway? 
(I often used to end up being too verbose... and now I seem to have overcompensated in the other direction)

src/Relation/Binary/Definitions.agda
jamesmckinna: latest commits make them explicit...

jamesmckinna:
Remark: all of the `Data.X.Properties` module duplicate the same abstract proof that (anti-)monotonicity implies distributivity over (lub)glb... so we should downstream lift *that* argument out once and for all? etc. DRY

jamesmckinna:
So this should now be ready, modulo:
* reconsideration of `Left`/`Right` ... so no `breaking` changes for this PR cf. #2532 
* refactoring towards `_-AlmostMonotonic`, on the model of #2573

src/Data/Rational/Unnormalised/Properties.agda
```diff
@@ -37,7 +37,7 @@ open import Relation.Binary.Bundles
 open import Relation.Binary.Structures
   using (IsEquivalence; IsDecEquivalence; IsApartnessRelation; IsTotalPreorder; IsPreorder; IsPartialOrder; IsTotalOrder; IsDecTotalOrder; IsStrictPartialOrder; IsStrictTotalOrder; IsDenseLinearOrder)
 open import Relation.Binary.Definitions
-  using (Reflexive; Symmetric; Transitive; Cotransitive; Tight; Decidable; Antisymmetric; Asymmetric; Dense; Total; Trans; Trichotomous; Irreflexive; Irrelevant; _Respectsˡ_; _Respectsʳ_; _Respects₂_; tri≈; tri<; tri>)
+  using (Reflexive; Symmetric; Transitive; Cotransitive; Tight; Decidable; Antisymmetric; Asymmetric; Dense; Total; Trans; Trichotomous; Irreflexive; Irrelevant; _Respectsˡ_; _Respectsʳ_; _Respects₂_; tri≈; tri<; tri>; Monotonic₁; Monotonic₂; LeftMonotonic; RightMonotonic)
```
MatthewDaggitt, line 40: At this point this `using` just needs to go?

src/Data/Rational/Properties.agda
```diff
@@ -1331,26 +1331,26 @@ module _ where
 *-cancelˡ-≤-pos : ∀ r .{{_ : Positive r}} → r * p ≤ r * q → p ≤ q
 *-cancelˡ-≤-pos {p} {q} r rewrite *-comm r p | *-comm r q = *-cancelʳ-≤-pos r
 
-*-monoʳ-≤-nonNeg : ∀ r .{{_ : NonNegative r}} → (_* r) Preserves _≤_ ⟶ _≤_
+*-monoʳ-≤-nonNeg : ∀ r .{{_ : NonNegative r}} → Monotonic₁ _≤_ _≤_ (_* r)
```
MatthewDaggitt, line 1334: Feels like the 8 lemmas below here should use `Left`/`RightMonotonic`? Maybe same in `Data.Integer.Properties`?

src/Data/Nat/Properties.agda
```diff
@@ -968,25 +972,25 @@ n≢0∧m>1⇒m*n>1 m n rewrite *-comm m n = m≢0∧n>1⇒m*n>1 n m
...
 
-*-monoˡ-≤ : ∀ n → (_* n) Preserves _≤_ ⟶ _≤_
-*-monoˡ-≤ n m≤o = *-mono-≤ m≤o (≤-refl {n})
+*-monoˡ-≤ : RightMonotonic _≤_ _≤_ _*_
+*-monoˡ-≤ = mono₂⇒monoʳ {≤₁ = _≤_} {≤₂ = _≤_} {≤₃ = _≤_} ≤-refl *-mono-≤
 
-*-monoʳ-≤ : ∀ n → (n *_) Preserves _≤_ ⟶ _≤_
-*-monoʳ-≤ n m≤o = *-mono-≤ (≤-refl {n}) m≤o
+*-monoʳ-≤ : LeftMonotonic _≤_ _≤_ _*_
+*-monoʳ-≤ = mono₂⇒monoˡ {≤₁ = _≤_} {≤₂ = _≤_} {≤₃ = _≤_} ≤-refl *-mono-≤
```
MatthewDaggitt, line 983: This providing of implicits is annoying. Suggests that some of them should be explicit in the proof?

src/Data/Nat/Properties.agda
```diff
@@ -702,31 +706,31 @@ m+n≤o⇒n≤o : ∀ m {n o} → m + n ≤ o → n ≤ o
...
 
-+-monoˡ-≤ : ∀ n → (_+ n) Preserves _≤_ ⟶ _≤_
-+-monoˡ-≤ n m≤o = +-mono-≤ m≤o (≤-refl {n})
++-monoˡ-≤ : RightMonotonic _≤_ _≤_ _+_
++-monoˡ-≤ = mono₂⇒monoʳ {≤₃ = _≤_} ≤-refl +-mono-≤
 
-+-monoʳ-≤ : ∀ n → (n +_) Preserves _≤_ ⟶ _≤_
-+-monoʳ-≤ n m≤o = +-mono-≤ (≤-refl {n}) m≤o
++-monoʳ-≤ : LeftMonotonic _≤_ _≤_ _+_
++-monoʳ-≤ = mono₂⇒monoˡ {≤₃ = _≤_} ≤-refl +-mono-≤
```
MatthewDaggitt, line 717: Same here.

src/Algebra/Consequences/Base.agda
```diff
@@ -13,14 +13,24 @@ module Algebra.Consequences.Base
...
+  sel⇒idem : Selective _≈_ _∙_ → Idempotent _≈_ _∙_
   sel⇒idem sel x = reduce (sel x x)
 
+  module Congruence (cong : Congruent₂ _≈_ _∙_) (refl : Reflexive _≈_) where
+
+    ∙-congˡ : LeftCongruent _≈_ _∙_
+    ∙-congˡ = mono₂⇒monoˡ {≤₁ = _≈_} {≤₂ = _≈_} {≤₃ = _≈_} refl cong _
+
+    ∙-congʳ : RightCongruent _≈_ _∙_
+    ∙-congʳ = mono₂⇒monoʳ {≤₁ = _≈_} {≤₂ = _≈_} {≤₃ = _≈_} refl cong _
```
MatthewDaggitt, line 32: Why were the old proofs in `Algebra.Structures` so simple and these are now so complicated?

src/Algebra/Consequences/Base.agda
```diff
@@ -13,14 +13,24 @@ module Algebra.Consequences.Base
...
 open import Relation.Binary.Definitions using (Reflexive)
 
-module _ {ℓ} {_•_ : Op₂ A} (_≈_ : Rel A ℓ) where
+module _ {ℓ} {_∙_ : Op₂ A} (_≈_ : Rel A ℓ) where
 
-  sel⇒idem : Selective _≈_ _•_ → Idempotent _≈_ _•_
+  sel⇒idem : Selective _≈_ _∙_ → Idempotent _≈_ _∙_
   sel⇒idem sel x = reduce (sel x x)
 
+  module Congruence (cong : Congruent₂ _≈_ _∙_) (refl : Reflexive _≈_) where
```
MatthewDaggitt, line 26: Can we de-nest this module for readability and move it above the proofs about selectivity? We could also probably make `{ℓ} {_∙_ : Op₂ A}` variables? I assume you've checked that it requires that `_≈_` be explicit as well?

MatthewDaggitt (changes requested):
Otherwise looks really good! Massive improvement.

CHANGELOG.md
```diff
@@ -98,6 +98,20 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Consequences.Base`:
+  ```agda
+  module Congruence (cong : Congruent₂ _≈_ _∙_) (refl : Reflexive _≈_) where
```
MatthewDaggitt, line 103: This module parameterisation is currently missing equality! Evidence for the idea that my suggestion of denesting the modules would improve code readabilty!

src/Data/Rational/Unnormalised/Properties.agda
jamesmckinna: Absolutely! DONE

src/Data/Rational/Properties.agda
jamesmckinna: This is why I'm currently stalling on the reviews of #2573 , because this PR will then itself need the analogues of `_-AlmostLeftMonotonic_` etc., and I didn't quite have the energy to add them... yet.

src/Data/Nat/Properties.agda
jamesmckinna: Yes, I've been scratching my head over how it might be avoided... without a conclusion yet.

src/Algebra/Consequences/Base.agda
jamesmckinna: Sigh... because of my DRY-inspired insistence on having `Congruent` delegate to `Monotonic`... and I might need to row back on that design decision. Oh dear...

src/Algebra/Consequences/Base.agda
jamesmckinna: Re: `variable`s... see #2592 ;-) which this PR could be marked as blocking on...
Re: should `_≈_` be implicit or explicit? Good question... (see also monotonicty/congruence lemmas above ;-))

CHANGELOG.md
jamesmckinna: OK... I was clearly too influenced by my own thinking towards #2572 . Of course, for `Base`, the implicit/explicit distinction is moot, as most deployments in client modules will delegate to `Algebra.Consequences.{Setoid|Propositional}`, where the actual equality will be hidden...

jamesmckinna:
I'll resolve the merge conflict later, meanwhile tricky questions above needing more thought... ;-)

src/Data/Rational/Properties.agda
MatthewDaggitt: Oh right of course. Ignore this comment. Best to get this merged in if we can.

src/Algebra/Structures.agda
```diff
@@ -54,13 +54,14 @@ record IsMagma (∙ : Op₂ A) : Set (a ⊔ ℓ) where
...
   setoid = record { isEquivalence = isEquivalence }
-
+{-
   ∙-congˡ : LeftCongruent ∙
   ∙-congˡ y≈z = ∙-cong refl y≈z
 
   ∙-congʳ : RightCongruent ∙
   ∙-congʳ y≈z = ∙-cong y≈z refl
-
+-}
```
MatthewDaggitt: Delete?

src/Data/Nat/Properties.agda
JacquesCarette: So if you make all 3 underscores, which ones are yellow? I notice that the previous proof provided the `n` explicitly, might that work here as well?

src/Data/Rational/Unnormalised/Properties.agda
JacquesCarette, line 40: long line?

JacquesCarette (changes requested):

src/Data/Nat/Properties.agda
jamesmckinna: It seems that only `{≤₃ = _≤_}` is required here...

jamesmckinna: Aaargh: there's a difference depending on whether we consider left or right, presumably because of the left-biased implementations. Will need to investigate further. Sigh.
```agda
*-monoˡ-≤ : RightMonotonic _≤_ _≤_ _*_
*-monoˡ-≤ = mono₂⇒monoʳ {≤₃ = _≤_} ≤-refl *-mono-≤

*-monoʳ-≤ : LeftMonotonic _≤_ _≤_ _*_
*-monoʳ-≤ = mono₂⇒monoˡ {≤₁ = _≤_} {≤₂ = _≤_} {≤₃ = _≤_} ≤-refl *-mono-≤
```

src/Data/Nat/Properties.agda
jamesmckinna: But here just 1 suffices:
```agda
+-monoˡ-≤ : RightMonotonic _≤_ _≤_ _+_
+-monoˡ-≤ = mono₂⇒monoʳ {≤₃ = _≤_} ≤-refl +-mono-≤

+-monoʳ-≤ : LeftMonotonic _≤_ _≤_ _+_
+-monoʳ-≤ = mono₂⇒monoˡ {≤₃ = _≤_} ≤-refl +-mono-≤
```

src/Data/Rational/Unnormalised/Properties.agda
jamesmckinna: Fixed by earlier commits, so merge conflict resolution should fix this.

src/Algebra/Consequences/Base.agda
```diff
@@ -10,22 +10,35 @@
...
+  ∙-congʳ : RightCongruent _∙_
+  ∙-congʳ {x} = mono₂⇒monoʳ {≤₁ = _≈_} {≤₃ = _≈_} (refl {x = x}) cong x
+
+module _ {ℓ} {_∙_ : Op₂ A} (_≈_ : Rel A ℓ) (open Definitions _≈_) where
+
+  sel⇒idem : Selective _∙_ → Idempotent _∙_
   sel⇒idem sel x = reduce (sel x x)
 
-module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) where
+module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) (open Definitions _≈_) where
```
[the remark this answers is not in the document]
jamesmckinna: Because in each case, my conviction is:
* it's easier to read the hypotheses and statements, without the noise of the additional `_≈_` arguments
* that this is further weight towards my proposed refactoring in #2502 / #2572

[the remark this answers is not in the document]
JacquesCarette: I think you misunderstood my point. I want that `open` there, just inside the module instead of in the telescope!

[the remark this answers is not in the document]
jamesmckinna: OK, but I had been maintaining consistency with the other two, where the `open` declaration *is* required in the module telescope, in order to be able to state the hypotheses without reference to equality...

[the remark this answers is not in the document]
JacquesCarette: Ah, that's the part that I had not seen (because I was looking at just the diffs). I'm still not keen, but consistency wins me over for this one case.

JacquesCarette (approved):

jamesmckinna:
Have fixed the merge conflict, and... not sure what else to do! Maybe it can still be improved wrt implicits?

JacquesCarette (approved):

JacquesCarette:
Just needs a second reviewer.

src/Algebra/Structures.agda
jamesmckinna: Thanks! You beat me to it.

Issue #2570, opened 2025-01-29
[ DRY ] Streamline `Function.*` hierarchy


jamesmckinna:
Now that #2568 / #2569 establish that (for `Setoid`s) `Bijection` and `Inverse` are equivalent concepts, we should perhaps do a wholesale refactoring of the `Function` hierarchy to (choose and) deprecate whichever definitions (and their `Propositional` friends`) we now deem to be redundant?

JacquesCarette:
Can you give us a preview of what you think this would entail?

jamesmckinna:
Well, the obvious 'minimal' fix/streamlining I envisaged after #2568 would be to:
* pick *one* of `Bijection` or `Inverse` (and their `Propositional` counterparts), and deprecate the other in its favour, with `Biased` implementations replacing them?
* all the knock-ons which might flow from such

But, the more I worked on that PR, the more threads seemed to get pulled... so it's a bit hard to see exactly how things might cash out?

jamesmckinna:
> Can you give us a preview of what you think this would entail?

@JacquesCarette The *other* way to answer your question might be: I have no idea, my brain hurts, so I'd like someone *else* to think about the possible knock-on consequences of #2569 ... ;-) (with shoutouts to @alexarice for the redefinition of `Surjective` and to @MatthewDaggitt for all his heroic work in landing #759 , without which, ...)

jamesmckinna:
See also: my recent annotations to #2569 identifying possible inlining/deprecation opportunities.

jamesmckinna:
#2569 is now (long since!) closed in favour of #2583 so closing this issue in favour of followups on that PR, and any subsequent downstream issues/PRs?

JacquesCarette:
So my thinking would be to *not* do this, i.e. have only one of `Bijection` and `Inverse`. Yes, they are equivalent -- so what? It is often very useful to have multiple interfaces to the same concepts (see the 'Realms' paper). Another way to see this is as a dual to views! Those are definitely good.

From a UX perspective, I'm definitely pro redundancy. The added maintenance is quite small. [And I'm normally a huge proponent of DRY!]

jamesmckinna:
But see also the redundancy in #3069 ...

PR #2583, opened 2025-02-13
[ refactor ] Symmetry of `Bijection` as a consequence of properties of a given `Surjective` function

### LINKED ISSUE #2568: [ DRY ] Refactor `Function.*` to rationalise the existence of a section to a given `Surjective f`

jamesmckinna:
Alternative version of #2569 which dispenses with much of the deprecation cruft, in favour of badging this as a `bug` fix. 
No separate `module Section` added, in favour of folding everything into the definitions of `Function.Structures.IsBijection` and `Function.Bundles.Bijection`.

NB. there may now be some further redundancy that could be pruned out: manifest fields, or `Properties`?


JacquesCarette (approved):

jamesmckinna:
@JacquesCarette which of this and #2569 do you prefer? We should (probably) only merge one of them...

jamesmckinna:
I'll only return to the merge conflicts once we choose between this v3.0 version, and the clunkier v2.3 #2569

JacquesCarette:
> which of this and https://github.com/agda/agda-stdlib/pull/2569 do you prefer?

There's a lot in common between them, so I'm having a hard time spotting exactly where the differences are?

jamesmckinna:
> > which of this and #2569 do you prefer?
> 
> There's a lot in common between them, so I'm having a hard time spotting exactly where the differences are?

The main difference is that
* the v2.3 version carries around a lot of extra cruft, in order to be backwards-compatible
* the v3.0 is much more streamlined, as I was able to jettison all that cruft

This was the one that made me seriously think we should simply leapfrog to a full, breaking, v3.0... and I am still considering withdrawing the v2.3 one as too compromised to be worth the effort? Hence the request for comment.

jamesmckinna:
I've downgraded/closed the earlier #2569 in favour of this one. Will fix the merge conflicts, and tidy up the comments to sort out the discussion of congruence.

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Bug-fixes
   the record constructors `_,_` incorrectly had no declared fixity. They have been given
   the fixity `infixr 4 _,_`, consistent with that of `Data.Product.Base`.
 
+* A major overhaul of the `Function` hierarchy sees the systematic development
+  and use of the theory of the left inverse `from` to a given `Surjective` function
+  `to`, as a consequence of which we can achieve full symmetry of `Bijection`, in
+  `Function.Properties.Bijection`/`Function.Construct.Symmetry`, rather than the
+  restricted versions considered to date. NB. this is non-backwards compatible
+  because the types of various properties are now sharper, and some previous lemmas
+  are no longer present, due to the complexity their deprecation would entail.
```
MatthewDaggitt, line 26: Can we be more explicit here? Which property types are now sharper? Which lemmas are not present?

JacquesCarette (approved):

jamesmckinna:
To my amazement, no further inherited merge conflicts since 2025! 
UPDATED: Ha! More fool me for merging #2996 round we go again...

UPDATED: moved `CHANGELOG` text from `bug fixes` to `non-backwards compatible changes`, it seems a better fit.

jamesmckinna:
I'll wait to resolve the merge conflicts until some of the other PRs in v3.0 milestone have been merged.
And it's just as well, because all the deprecation messages need to be updated from v2.3 to v3.0!!! DONE.

jamesmckinna:
Note to self to (re-)consider the `*WithoutCongruence` lemmas mentioned in #2570 ...

CHANGELOG.md
```diff
@@ -68,6 +68,23 @@ Bug-fixes
...
+  Specifically:
+  - `Function.Construct.Symmetry.isBijection` no longer requires the hypothesis
+    `Congruent ≈₂ ≈₁ f⁻¹` for `f⁻¹ = B.from`.
+  - `Function.Construct.Symmetry.isBijection-≡` is now redundant, as an instance
```
MatthewDaggitt, line 81: Might we be able to deprecate instead of delete all the removed lemmas? Just because we _can_ break backwards compatibility for 3.0 doesn't mean that we shouldn't do our best to maintain it where possible?

MatthewDaggitt (approved):
Otherwise looks good to me!

CHANGELOG.md
jamesmckinna: Hmm. There used to be a v2.4-compatible version #2569  which did as much of that as possible, and back then it seemed 'easier' simply to prioritise the 'right' version. 

Some deprecations I think would be painful, if I recall correctly, because of the dependency graph. Again I'll take a look. 

Once upon a time merging that version for v2.4 and then coming back to this one might have been a possible path... but that was then. Sigh.

jamesmckinna: It'll be Monday before I can look at this I expect.

jamesmckinna: Looking at this again, I have now reinstated those deprecations from #2569 which do *not* touch the new `Function.Consequences.Section` lemmas introduced in that PR as part of earlier refactoring. 

Instead, at least if I understand my own thought processes 18+ months years later, those `Consequences` have now all been folded in-place into the corresponding `Function.Bundles`, `Function.Structures`, and `Function.Properties`, and it's not obvious to me that anything is to be gained from lifting them out as free-standing `Consequences` in their own right...

jamesmckinna:
Finally (!?):
* have reinstated some of the earlier deprecations from #2569 
* have *not* reinstated the lemmas from that PR's introduction of a submodule `Function.Consequences.Section`, in favour of the alternative approach in *this* PR, where all such `Properties` are defined 'in-place'
* I now observe that, esp. internally wrt appeals to symmetry, that some refactoring could be undertaken downstream in the light of #2599 ... but not this time around!

Consequently, with 2 approving reviews, I'm going to add this to the queue...

CHANGELOG.md
MatthewDaggitt: Thanks, okay that makes sense to me!

jamesmckinna:
> NB. there may now be some further redundancy that could be pruned out: manifest fields, or `Properties`?

The knock-on might be that we refactor `Function.Bundles` and `Function.Properties` to reflect these new additions. See #3069 for a possible *post hoc* example... but a more extensive pruning might be in order.

PR #2586, opened 2025-02-17
Add `suc-≢-injective` to `Fin`

javierdiaz72:
This PR adds the `suc-≢-injective` property to `Data.Fin.Properties`.

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -92,6 +92,10 @@ nonZeroIndex {n = suc _} _ = _
 suc-injective : Fin.suc i ≡ suc j → i ≡ j
 suc-injective refl = refl
 
+suc-≢-injective : Fin.suc i ≢ suc j → i ≢ j
+suc-≢-injective {i = zero} {j = zero} p = contradiction refl p
+suc-≢-injective {_}        {_}        p = λ q → contradiction (cong suc q) p
+
 infix 4 _≟_
 
 _≟_ : DecidableEquality (Fin n)


Taneb:
Thanks for the contribution! However, I'm (softly) against including this. A better definition is `suc-≢-injective = _∘ cong suc` (as it doesn't need to recurse), and this isn't really passing the Fairbairn threshold.

jamesmckinna:
Thanks @javierdiaz72 for the offer of the contribution,
thanks @Taneb for the review/feedback, with which I entirely agree 
(although I might think that your one-line definition is perhaps a bit too terse for beginners/users rather than `stdlib` developers... ;-))

To flesh out the criticism:
* besides the proof offered above, the lemma already has a *generic* proof as [Function.Consequences.contraInjective](https://github.com/agda/agda-stdlib/blob/94906c00c9b0afab2283149fa33875e31b226a61/src/Function/Consequences.agda#L28-L33)...
  ```agda
-- Injective


  contraInjective : ∀ (≈₂ : Rel B ℓ₂) → Injective ≈₁ ≈₂ f →
                  ∀ {x y} → ¬ (≈₁ x y) → ¬ (≈₂ (f x) (f y))
  contraInjective _ inj p = contraposition inj p
  ```
* ... which draws attention to the fact that there is really no need to consider a specific `Fin`-based pattern-matching analysis, as you supply (see what happens when you delete the `i = zero`, `j = zero` clause from your definition)
* and, finally, Nathan's observation that we (informally) require a minimum-sized delta for any contributed PR to be worth considering for a merge, and on the basis of the above, it would be hard to justify.

Many thanks again though for a first attempt!

Issue #2219, opened 2023-11-30
Define `LocalRing`


jamesmckinna:
Some delicacy required over the precise axiomatisation, in order to avoid too much decidability (and with the Maximal Ideal 'Theorem'/Axiom also lurking offstage), so perhaps using the geometric version: Every element is either zero or else has a multiplicative inverse?

Nagata, (1962, Wiley), "Local Rings"
[nlab page](https://ncatlab.org/nlab/show/local+ring)

mechvel:
The most general construct here is a Fraction Domain over a  (commutative?) Ring R by a Multiplicative system S in R, where the denominators for fractions over R  are taken only from S. 
And a subset S is required to be closed under `*` (I do not recall, maybe some more requirement is set for S).
A local ring is a special case of  FractionDomain over S,  where  ``S = R \ I```,   where `I` is a prime ideal.
So, the  most reasonable way (for an Applied Library !) is to 
1) define what is a multiplicative system, probably as the membership function agreed with  ≈  and *,
2) define the ideal notion, again, via  the membership function agreed with  ≈  and operations.
3) define the constructs of (FractionDomai S),   (LacalRing primeIdeal-I).

But I am sure that these constructs are not for Standard library for Agda.
Such an approach leads to that Standard library will include the algebra library having million pages of the source code.
Local rings, and many other subtle and advance constructs are only for Applied libraries.

As to Standard library, it is sufficient  FractionField  over a GCDDomain.
I have such in in my Applied library.
But i do not believe that I am able to make it standard.
Because it relies on the general notion of GCD,  GCDDomain, and I failed to make standard even these ones.

jamesmckinna:
As to a `stdlib` with millions of lines of code, this is exactly the state of affairs for the Lean `mathlib` and the Rocq `MathComponents` libraries as I understand it. I think we should be bold enough to imagine a (near?) future in which Agda enjoys similar expressiveness.

Resource constraints on checking such a thing are another matter, as, ahem, are constraints on *developer* resources... ;-)

jamesmckinna:
> Growing `stdlib` a lot would be nice.

;-) one PR at a time...

> We should pay close attention to who are the main Agda users who use `stdlib` (and ignore cubical Agda users, they are a different breed): lots and lots use it to do the meta-theory of programming languages and the like. So our biggest impact would be to support those activities as much as possible.

[...]

PR #2588, opened 2025-02-18
[ refactor ] (Re)define `(Is)TightApartness` and `(Is)HeytingCommutativeRing`/`(Is)HeytingField`

### LINKED ISSUE #2587: [ refactor ] Definitions of `Relation.Binary.Apartness` and `Algebra.*.*HeytingCommutativeRing`
```quote
* the PR itself draws attention to the fact that `Symmetric _#_` is in fact a redundant part of the definition of `IsHeytingCommutativeRing`, so it *might* be worth having an `Algebra.Apartness.Structures.Biased` smart constructor which puts the pieces together, but this could be developed downstream;
* the whole definition of HCR now seems a bit odd (compared to HCF?): why should a `Ring` admit inverses of elements apart from `0`? As opposed to simply being a `CommutativeRing` with a `Tight` apartness relation?
```

_Excepted from original post by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2576#issuecomment-2644745606_

jamesmckinna:
Fixes #2587 

Adds:
* `Algebra.Apartness.Properties.HeytingField`, superseding `Algebra.Apartness.Properties.HeytingCommutativeRing`

NB:
* latter module should now be deprecated (?) in favour of moving its one remaining property to `Algebra.Properties.Ring` (or even `Algebra.Properties.AbelianGroup`...) DONE properties moved; module NOT deprecated
* `Data.Rational.Unnormalised.Properties` should be refactored to make use of `Relation.Binary.Properties.DecSetoid` for its `(Is)*ApartnessRelation` definitions DONE

Issue:
* is this v2.3 (as a `bug` fix of the various APIs involved?) or v3.0 (as a large `breaking` change?)
* re: @MatthewDaggitt 's comment below about the mathematics, perhaps if someone could check what is said in eg. Troelstra and van Dalen, that might shape how this goes forward, but my own view is: authors seem to disagree about `Apartness` vs. `TightApartness`, etc. but that the fundamental distinction should be between `*CommutativeRing` and `*Field` (and yes: perhaps `HeytingLocalRing` would have been better, but see also #2219 ) in that the latter should have inverses, while the former need not. The existing APIs make the distinction turn on whether the apartness is tight or not, which seems 'wrong'... even to the point of being a `bug`?

RFC from @cspollard and @bsaul ...


src/Relation/Binary/Properties/DecSetoid.agda
```diff
@@ -36,8 +37,19 @@ open SetoidProperties setoid
   ; cotrans = ≉-cotrans
   }
 
+≉-tight : Tight _≈_ _≉_
+≉-tight x y = decidable-stable (x ≟ y)
```
MatthewDaggitt, line 41: Missing addition from Changelog.

src/Relation/Binary/Definitions.agda
```diff
@@ -147,7 +147,7 @@ Cotransitive : Rel A ℓ → Set _
 Cotransitive _#_ = ∀ {x y} → x # y → ∀ z → (x # z) ⊎ (z # y)
 
 Tight : Rel A ℓ₁ → Rel A ℓ₂ → Set _
-Tight _≈_ _#_ = ∀ x y → (¬ x # y → x ≈ y) × (x ≈ y → ¬ x # y)
+Tight _≈_ _#_ = ∀ x y → ¬ x # y → x ≈ y
```
MatthewDaggitt, line 150: This change (with the explicit change in definition) should be documented in the Changelog?

src/Algebra/Apartness/Properties/HeytingField.agda
```diff
@@ -0,0 +1,108 @@
...
+  using (-0#≈0#; -‿distribˡ-*; -‿distribʳ-*; -‿anti-homo-+; -‿involutive)
+open import Relation.Binary.Definitions using (Symmetric)
+import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
+open import Algebra.Properties.CommutativeMonoid *-commutativeMonoid
+
+private variable
+  x y z : Carrier
+
+
+open import Algebra.Apartness.Properties.HeytingCommutativeRing heytingCommutativeRing public
```
MatthewDaggitt, line 31: Hmm I was going to say this should go above the variable field by convention, but then I spotted the public at the end. I really feel the `public` key word should have gone in the front of the statement if I could design Agda again...

src/Algebra/Apartness/Bundles.agda
```diff
@@ -36,8 +36,8 @@ record HeytingCommutativeRing c ℓ₁ ℓ₂ : Set (suc (c ⊔ ℓ₁ ⊔ ℓ
   commutativeRing : CommutativeRing c ℓ₁
   commutativeRing = record { isCommutativeRing = isCommutativeRing }
 
-  apartnessRelation : ApartnessRelation c ℓ₁ ℓ₂
-  apartnessRelation = record { isApartnessRelation = isApartnessRelation }
```
MatthewDaggitt, line 40: Is `apartnessRelation` still exported from somewhere? I can't spot it. If not, we should reopen `tightApartnessRelation`?

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Non-backwards compatible changes
   significantly faster. However, its reduction behaviour on open terms may have
   changed.
 
+* The definitions of `Algebra.Structures.IsHeyting*` and
```
MatthewDaggitt, line 20: I know this is a draft, but in the final version it would be good to explain broadly what the refactorings actually are here?

MatthewDaggitt (changes requested):
Otherwise, code-wise it looks good. Can't comment on the mathematics :smile:

src/Relation/Binary/Properties/DecSetoid.agda
jamesmckinna: yeah, I wasn't sure how to record this... it's a *redefinition*, rather than an addition, because the definition of `Tight` has been changed in `Relation.Binary.Definitions` to drop the redundant second conjunct, which is equivalent to `Irreflexive`, as noted on the issue.

What's the best way to do this?

src/Algebra/Apartness/Properties/HeytingField.agda
jamesmckinna: That said, I've now managed to delete this re-export completely, which is one solution to your problem ;-)

src/Algebra/Apartness/Bundles.agda
jamesmckinna: Ah... I slightly screwed up the latest round of commits wrt the merge conflict arising from #2576 so will do another round of checking...

jamesmckinna:
I think this is now ready for review!
many thanks to @MatthewDaggitt for the pre-review, hopefully the latest rpound of (many!) commits have put things on a more sound footing. As in updated preamble above:
* is this v2.3 or v3.0? (maybe the latter is safer?)
* I've left `Algebra.Apartness.properties.HeytingCommutativeRing` as an empty stub, with two deprecations, rather than deprecate the module altogether, against possible future additions to this?

JacquesCarette (approved):
Spent a lot of time reviewing this... and no comment to speak of. I did spot several proofs that I'd want to change, but I should add some reasoning combinators first that would show how these embody repeated patterns.

src/Algebra/Apartness/Properties/HeytingField.agda
```diff
@@ -0,0 +1,85 @@
...
+invertibleʳ⇒# = invertible⇒# ∘ invertibleʳ⇒invertible
+
+1#0 : 1# # 0#
+1#0 = invertibleˡ⇒# (1# , 1*[x-0]≈x)
+  where
+  1*[x-0]≈x : 1# * (x - 0#) ≈ x
+  1*[x-0]≈x {x} = trans (*-identityˡ (x - 0#)) (x-0#≈x x)
+
+x#0y#0→xy#0 : x # 0# → y # 0# → x * y # 0#
+x#0y#0→xy#0 {x} {y} x#0 y#0 = helper (#⇒invertible x#0) (#⇒invertible y#0)
```
jamesmckinna, line 48: @JacquesCarette 's wanting to refactor the proofs. 
It occurs to me that some (all?) of these arguments could be refactored by going via `Algebra.Properties.CancellativeCommutativeSemiring`, after establishing the relevant structure/bundle?

JacquesCarette: Could be. I was thinking of analogs to the `elim` and `cancel` reasoning combinators of `agda-categories`. While at it, I would have ported push/pull (but renamed them) too. Many of these already work on `Monoid`.

jamesmckinna:
> Otherwise, code-wise it looks good. Can't comment on the mathematics 😄

Hopefully the discussion on #2587 clarifies this?

jamesmckinna:
I'll return to the merge conflicts after v2.3, given that this is a breaking change.

src/Algebra/Apartness/Properties/HeytingField.agda
JacquesCarette: Hoping to do that via proxy (if in the next month) or perhaps myself after that.

JacquesCarette (approved):

JacquesCarette (approved):

jamesmckinna:
@JacquesCarette wrote:
> Spent a lot of time reviewing this... and no comment to speak of. I did spot several proofs that I'd want to change, but I should add some reasoning combinators first that would show how these embody repeated patterns.

Much of the refactoring here (essentially code motion, little else, in service of the new, hopefully more streamlined abstractions) leaves original legacy proofs untouched. We may now have made progress elsewhere on enriching the reasoning infrastructure (have we?), so there's an opportunity here to weigh in on what you would want to change here, or else to open a fresh issue/PR documenting such. Let me know what you'd like us to do!

JacquesCarette:
@jamesmckinna needs yet another merge. @MatthewDaggitt this is then going to be waiting on you.

jamesmckinna:
Thanks @JacquesCarette will look at this this evening. Battling insomnia and some slightly more urgent stuff. DONE.

jamesmckinna:
So... I did finally consult Troelstra and van Dalen (Vol.2, which is a bit harder to get hold of, and is also 2x length of Vol. 1...), and their chapter on 'Algebra' has this kind of thing largely based on a KNAW article by Heyting from 1941 which I have yet to read...

... nevertheless they draw attention to the need for *additional* axioms governing the relationship between `_#_` and the `_+_` and `_*_` operators. So I'm going to push this to DRAFT for the time being, and then come back once I've added the extras back in. DONE.

Rationale: we're already making a breaking change as to the definitions of `Tight` and `(Is)TightApartness`, so we might as well get the mathematics right. Thanks to @MatthewDaggitt for drawing attention to this!

jamesmckinna:
Candidate `Algebra.Apartness.Properties.HeytingCommutativeRing` (so that that module deserves not to be deprecated!):
* prove that `_-_` is `StronglyExtensional _#_`.
* `Biased` smart constructors so that `TightApartness` and `StronglyExtensional` imply `Congruent` (and thus the underlying `IsCommutativeRing` axioms can be weakened.
* `0# # 1#`, or else require that as part of the axiomatisation (!)

All of the above, except the last perhaps, are downstream considerations, so stopping here.

UPDATED: spoke too soon. Need to fix axiomatisation, and prove the extra pieces for `Data.Rational`. Ooops!

UPDATED: extra `StronglyExtensional` properties now proved. Moving off `DRAFT` for review.

src/Algebra/Apartness/Properties/HeytingCommutativeRing.agda
```diff
@@ -9,76 +9,31 @@
...
-
-#-congˡ : y ≈ z → x # y → x # z
-#-congˡ y≈z x#y = #-sym (#-congʳ y≈z (#-sym x#y))
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.3
```
jamesmckinna: Aaargh
```suggestion
-- Version 3.0
```

src/Algebra/Apartness/Properties/HeytingCommutativeRing.agda
```diff
@@ -9,76 +9,31 @@
...
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 3.0
+
+x-0≈x = x-0#≈x
+{-# WARNING_ON_USAGE x-0≈x
+"Warning: x-0≈x was deprecated in v2.3.
```
jamesmckinna:
```suggestion
"Warning: x-0≈x was deprecated in v3.0.
```

src/Algebra/Apartness/Properties/HeytingCommutativeRing.agda
```diff
@@ -9,76 +9,31 @@
...
+Please use Algebra.Properties.Ring.x-0#≈x instead."
+#-}
+
+open HeytingCommutativeRing heytingCommutativeRing public using (#-sym)
+{-# WARNING_ON_USAGE #-sym
+"Warning: #-sym was deprecated in v2.3.
```
jamesmckinna:
```suggestion
"Warning: #-sym was deprecated in v3.0.
```

jamesmckinna:
Last call for @MatthewDaggitt to (re-)review!? I think the substance of your criticism revolved round adequately accounting for (the intent behind) this refactoring in the `CHANGELOG`. Plus the (hopefully correct!) re-export of `(Is)ApartnessRelation` and its accounting in `CHANGELOG`.

I'd be happy to clear this off the docket before another round of merge conflicts torpedo it...

src/Algebra/Apartness/Properties/HeytingCommutativeRing.agda
```diff
@@ -9,76 +9,31 @@
...
-open import Algebra.Properties.Ring ring
-  using (-0#≈0#; -‿distribˡ-*; -‿distribʳ-*; -‿anti-homo-+; -‿involutive)
-open import Relation.Binary.Definitions using (Symmetric)
-import Relation.Binary.Reasoning.Setoid as ≈-Reasoning
-open import Algebra.Properties.CommutativeMonoid *-commutativeMonoid
 
-private variable
-  x y z : Carrier
-
-invertibleˡ⇒# : LeftInvertible _≈_ 1# _*_ (x - y) → x # y
```
MatthewDaggitt, line 30: We seemed to have removed a lot of definitions from this file without either re-exporting them from elsewhere, deprecating them or documenting it in the CHANGELOG? Maybe I'm missing something though...

MatthewDaggitt (approved):
Other than that, looks good.

src/Algebra/Apartness/Properties/HeytingCommutativeRing.agda
jamesmckinna: OK, thanks for this. 

I think I had (at least!) two/three/four problems:
* needing to move code from `HeytingCommutativeRing` to `HeytingField` (specifically, the invertibility properties you highlight) but without renaming/deprecation, because not wanting to invent 'bogus' new names for the sake of standardising things apart (or: dependency cycles!? they can't be re-exported from `HField` to `HCRing`...); this is a 'bug' in the original definitions, so, or so I thought, needing 'fixing' by putting things in the 'right' place...
* changing certain other definitions (specifically `Tight` as a property of an `ApartnessRelation`) to better account for the original *intention* behind the `(Is)Heyting*` definitions
* adding the 'official' Heyting-like properties of the algebra operations being suitably 'congruential' wrt apartness, by positive phrasing the analogues of the contrapositives of `Congruent` (cf. `Cancellable` etc.)
* how best to account for these things in the `CHANGELOG`.

I clearly need to do better as to the last point, so I'll have another go at fleshing out what's what here...

CHANGELOG.md
jamesmckinna: I've now added more detail here since your most recent round of review, asking, as I understood it, for the `CHANGELOG` to better account for what's going on here.

PR #2589, opened 2025-02-18
Add filter-↭

Taneb:
Filtering two lists that are permutations of each other leaves you with two (possibly smaller) lists which are permutations of each other

### DIFF
--- a/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
+++ b/src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
@@ -26,7 +26,7 @@ open import Data.Product.Base using (_,_; _×_; ∃; ∃₂)
 open import Data.Maybe.Base using (Maybe; just; nothing)
 open import Function.Base using (_∘_; _⟨_⟩_; _$_)
 open import Level using (Level)
-open import Relation.Unary using (Pred)
+open import Relation.Unary as Pred using (Pred)
 open import Relation.Binary.Core using (Rel; _Preserves_⟶_; _Preserves₂_⟶_⟶_)
 open import Relation.Binary.Definitions using (_Respects_; Decidable)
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):
Wow, `with` *and* `rewrite` and I still approve! What am I becoming...  [I can clearly see that not using these tools would make the resulting code much larger without any real gain.]

src/Data/List/Relation/Binary/Permutation/Propositional/Properties.agda
```diff
@@ -372,6 +372,21 @@ module _ {ℓ} {R : Rel A ℓ} (R? : Decidable R) where
...
+------------------------------------------------------------------------
+-- filter
+
+filter-↭ : ∀ {p} {P : Pred A p} (P? : Pred.Decidable P) → xs ↭ ys → filter P? xs ↭ filter P? ys
+filter-↭ P? refl = refl
+filter-↭ P? (prep x xs↭ys) with P? x
+... | yes _ = prep x (filter-↭ P? xs↭ys)
+... | no _  = filter-↭ P? xs↭ys
+filter-↭ P? (swap x y xs↭ys) with P? x in eqˣ | P? y in eqʸ
+... | yes _ | yes _ rewrite eqˣ rewrite eqʸ = swap x y (filter-↭ P? xs↭ys)
```
Taneb, line 384: I'm annoyed at having to match on `P? x` and `P? y` _twice_ each in each case to get the `filter` to normalize properly. Is there a way to avoid this?

gallais: You could prove an unfolding lemma:
```
P x -> filter P? (x :: xs)  == x :: filter P? xs
```

MatthewDaggitt (approved):

PR #2577, opened 2025-02-07
[ re #523 ] Refactorings using wlog

gallais:
* [x] `*-distribˡ-∣-∣ : _*_ DistributesOverˡ ∣_-_∣` 
* [x] `∣m⊝n∣≤m⊔n : ∀ m n → ∣ m ⊖ n ∣ ℕ.≤ m ℕ.⊔ n`

PR #2592, opened 2025-02-19
[ refactor ] use `variable`s in `Algebra.Consequences.Base`

jamesmckinna:
This lifts out the refactoring aspects of #2572 and should be merged first of the two.

It also rectifies two bugs
* use of deprecated names in `Data.Rational.Unnormalised.Properties` (2026-06-12: I no longer know what this refers to; so I guess it must have been fixed in the intervening time ;-))
* changing to `variable`s means that the names of implicit parameters now matter; but `Base` *still* used the incorrect symbol for the binary operation, which has now been corrected.

NB One (and only one!) knock-on change needed when supplying an implicit `Op₂` parameter!
UPDATED: in the `Propositional` case, instantiating proofs from `Setoid` with instances of `cong` should be accompanied by a change in parametrisation from implicit to explicit, because (`PropositionalEquality.Core.`)`cong` can't (necessarily) infer its argument, whereas the typechecker can infer `_∙_` when deploying general `Setoid` proofs taking an explicit `cong : Congruent₂ _∙_` argument ...


src/Algebra/Consequences/Base.agda
```diff
@@ -7,27 +7,40 @@
...
 
-  sel⇒idem : Selective _≈_ _•_ → Idempotent _≈_ _•_
+  sel⇒idem : Selective _≈_ _∙_ → Idempotent _≈_ _∙_
   sel⇒idem sel x = reduce (sel x x)
 
-module _ {ℓ} {f : Op₁ A} (_≈_ : Rel A ℓ) where
+------------------------------------------------------------------------
+-- SelfInverse
+
+module _  (_≈_ : Rel A ℓ) where
```
Taneb, line 39: Does there need to be two separate modules here? Can/should the `module _ (_\~~_ : Rel A \ell) where` headers be merged?

jamesmckinna: Hmmm... yes, but I had been trying to restore `Algebra.Consequences.base` to as-near-as-possible its original form, which had the two distinct internal modules.

Waiting on @MatthewDaggitt 's review comments on the two PRs split out of the orginal... but could indeed make your suggested change!

JacquesCarette (approved):

jamesmckinna:
Hmmm... recent merge conflicts were a bit strange, hopefully fixed now. 
Also fixed: bug left behind after splitting the original PR into two

src/Algebra/Consequences/Base.agda
jamesmckinna: Resolving in favour of addressing the equality refactoring in #2572

src/Data/Nat/Properties.agda
```diff
@@ -1848,7 +1848,7 @@ m∸n≤∣m-n∣ m n with ≤-total m n
   n ∸ m     ∎
 
 *-distribˡ-∣-∣ : _*_ DistributesOverˡ ∣_-_∣
-*-distribˡ-∣-∣ a = wlog ≤-total (comm⇒sym[distribˡ] {_◦_ = _*_} ∣-∣-comm a)
+*-distribˡ-∣-∣ a = wlog ≤-total (comm⇒sym[distribˡ] ∣-∣-comm {_◦_ = _*_} a)
```
MatthewDaggitt, line 1851: This makes this look like not a cosmetic change but a breaking change?

jamesmckinna: Sigh.
The proof above was itself only introduced last week in #2577 ... and I guess that it is inevitable that the move to using `variable` has the potential *always* to reorder the order of *implicit* arguments in proofs. In the past we haven't (apparently?) regarded this as `breaking` behaviour, possible because the 'real' underlying issue may be that the ambient parameterisation on the binary operation ought to be explicit. Will investigate whether there's a better fix to the above.

jamesmckinna: You're probably right it should be v3.0, and then we can fix the parametrisation problems exposed above as well.

jamesmckinna: ... and I'd press on the argument in favour of #2572 (backwards compatible!) as well as part of such v3.0 changes...

jamesmckinna:
I'll return to the merge conflict(s) after v2.3, given that this is a breaking change.

jamesmckinna:
Ugh. Fixing the merge conflicts turned out to be more fiddly than I'd hoped. Some brain-, as well as bit-, rot I fear...

JacquesCarette (approved):

jamesmckinna:
@Taneb @MatthewDaggitt would you be happy with merging this now?

jamesmckinna:
I'd like a second approval, so that this can merge and thus unblock #2572 .

Taneb (approved):

Issue #2594, opened 2025-02-19
A Question on Apartness definitions -- particularly `irrefl`


bsaul:
I have an observation/question about the apartness/inequality definitions (sorry if this is not a particularly well formed yet). As I read about apartness in constructive math texts, it's generally defined on top of a notion of *inequality* (though perhaps *inequivalence* is a better term). For example, in [_A Course in Constructive Algebra_](https://link.springer.com/book/10.1007/978-1-4419-8640-5) by Mines, Richman, and Ruitenberg on p.8 (1988 ed): "We regard the relation of *inequality* as conventional and not necessarily the denial of equality. ... A consistent [which is what they call irreflexive], symmetric, cotranstive inequality is called an *apartness*." On the surface, this all well and good and the definition of apartness seems to match what's in the stdlib (at least of v2.2 without looking back in history). 

But after reading [_Apartness and Uniformity_](https://link.springer.com/book/10.1007/978-3-642-22415-7) by Bridges and Vîţă, I began to wonder if the stdlib has the relation between inequivalence and equivalence correct. Here's how Bridges and Vîţă (p. 8) define inequality:

![Image](https://github.com/user-attachments/assets/32d0f022-9ee6-4e04-9f97-590fae0eb866)

Note the condition that $\neq$ (which translates to `_#_` in stdlib notation) implies the negation of equality (i.e. the "denial inequality" in the sense of Hines et al). Currently we have have an an `ApartnessRelation` is `Irreflexive _≈_ _#_`, which translates to `∀ {x y} → x ≈ y → ¬ (x # y)`, but this is the opposite of the Bridges definition, as I understand. Is the stdlib definition incorrect? or what am I missing here?

This is tangentially related to #2588 and Heyting algebra stuff more generally as that's were apartness starts to show up in the `Algebra` modules.

Taneb:
I don't have access to that particular book, but this is the difference between an apartness relation and a _tight_ apartness relation; see, for example, [Wikipedia](https://en.wikipedia.org/wiki/Apartness_relation#Definition)) and this library's definition of [`Tight`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Definitions.html#3826).

jamesmckinna:
[...]

PR #2595, opened 2025-02-19
Adds unary disjoint relation

bsaul:
This PR adds the relation between two predicates expressing disjointness.

I am not at all wedded to the `_#_` notation.

See [the discussion on the Agda Zulip](https://agda.zulipchat.com/#narrow/channel/264623-stdlib/topic/A.20couple.20of.20additions.20to.20.60Relation.2EUnary.60.20to.20consider) for more background.


src/Relation/Unary.agda
```diff
@@ -207,7 +207,7 @@ infixr 8 _⇒_
 infixr 7 _∩_
 infixr 6 _∪_
 infixr 6 _∖_
-infix 4 _≬_
+infix 4 _≬_ _#_ _#′_
```
jamesmckinna, line 210: What about `\perp`/`⊥` instead? (*not* `\bot` BTW, but indistinguishable in many fonts, alas...)

jamesmckinna:
(in lieu of my (now deleted) comment, I took the liberty of adding the link myself to your OP... hope that's OK)

bsaul:
> (I took the liberty of adding the link myself to your OP... hope that's OK)

Of course, thanks!

src/Relation/Unary.agda
bsaul: That's Ok by me. That said, I already find some of the symbols used in `Relation.Unary` to be difficult to distinguish.

JacquesCarette: paging in @WolframKahl who might also have a suggestion.

JacquesCarette: Because `\bot` and `\perp` are so hard to distinguish, I'm not pro. I think using the same symbol as `apartness` makes sense.

bsaul: I just pushed a version using `\perp`, so folks can look at it.

jamesmckinna: Actually, at least in `github`, I find I can distinguish it from `\bot`. 
FWIW, I think I'd rather have a distinct symbol than need to have `hiding` directives/qualified `import`s for the sake of reusing `_#_`...

jamesmckinna:
cf. #2594 ... Can you also add some properties of these relations? Or other extensions, eg.:
* prove `Symmetric _⊥_`, `Symmetric _≬_`
* lift `_⊥_` to `Relation.Binary.Definitions` (and even: `Relation.Nullary`?), and redefine `Irreflexive` using it?
* redefine relational composition using `_≬_`? (there may be dependency issues in trying to do this; partly because of the `Nullary`/`Unary`/`Binary` distinction, and partly because neither of `Unary` nor `Nullary` split off distinct `Definitions` modules...)
* ...

All of these strictly go outside the scope of what you have so far, but might also give life to the new definitions. 

You also alluded to how they are part of your basic toolkit when developing in Agda, so what's a specimen client application/deployment context which might also be generally useful as an addition to the library?

bsaul:
> Can you also add some properties of these relations?

Yes, will do. Those are good suggestions.

bsaul:
> You also alluded to how they are part of your basic toolkit when developing in Agda, so what's a specimen client application/deployment context which might also be generally useful as an addition to the library?

I'm working on formalizing parts probability/statistics/causal inference wherein disjoint "events" or sets are sometimes a precondition in a theorem, e.g. Pearl's definition of [d-separation](https://git.sr.ht/~bradleysaul/liatris/tree/main/item/src/Graph/Graph/DAG/Construct/dSeparation.lagda.md).

src/Relation/Unary.agda
bsaul: Another option is to be more verbose as in `_IsDisjointFrom_`.

jamesmckinna:
> > You also alluded to how they are part of your basic toolkit when developing in Agda, so what's a specimen client application/deployment context which might also be generally useful as an addition to the library?
> 
> I'm working on formalizing parts probability/statistics/causal inference wherein disjoint "events" or sets are sometimes a precondition in a theorem, e.g. Pearl's definition of [d-separation](https://git.sr.ht/~bradleysaul/liatris/tree/main/item/src/Graph/Graph/DAG/Construct/dSeparation.lagda.md).

Nice! But not sure quite how in scope (yet) that might be for `stdlib`. Still the two concepts introduced are familiar to me from constructive approaches to (the foundations of) probability, cf. eg. Jan van Plato's book from some years ago now...

bsaul:
> Nice! But not sure quite how in scope (yet) that might be for `stdlib`.

I wouldn't expect most of what I'm noodling on to be in scope for `stdlib` -- I'm just trying to upstream things that might be.

bsaul:
> cf. eg. Jan van Plato's book from some years ago now...

Thanks for the reference -- looks quite interesting.

bsaul:
> * lift `_⊥_` to `Relation.Binary.Definitions`? and redefine Irreflexive using it?

Trying to import `Relation.Unary` into `Relation.Binary.Defintions` introduces a cyclic dependency, so I suspect there would be  a bit of refactoring to make this happen.

src/Relation/Unary/Properties.agda
```diff
@@ -18,7 +18,8 @@ open import Relation.Binary.Definitions
 open import Relation.Binary.PropositionalEquality.Core using (refl; _≗_)
 open import Relation.Unary
 open import Relation.Nullary.Decidable as Dec using (yes; no; _⊎-dec_; _×-dec_; ¬?; map′; does)
-open import Function.Base using (id; _$_; _∘_)
+open import Relation.Nullary.Negation using (¬_)
```
jamesmckinna, line 21:
```suggestion
open import Relation.Nullary.Negation.Core using (¬_)
```

bsaul: Done.

CHANGELOG.md
```diff
@@ -143,3 +143,16 @@ Additions to existing modules
   ```agda
   ```
+
+* In `Relation.Unary`:
+  ```agda
+  _⊥_ _⊥′_ : Pred A ℓ₁ → Pred A ℓ₂ → Set _
+  ```
+
+* In `Relation.Unary.Definitions`:
```
jamesmckinna, line 152:
```suggestion
* In `Relation.Unary.Properties`:
```

jamesmckinna (changes requested):
Thanks very much for the PR, it looks in good shape for merging once the `CHANGELOG` is fixed.

I'm happy with `\perp` as the symbol, but I'll leave it to the other maintainers to rule on this.

src/Relation/Unary.agda
jamesmckinna: Perhaps, 'ideally', use of a ['fat \perp'](https://tex.stackexchange.com/questions/79434/double-perpendicular-symbol-for-independence) might be best, but I don't think we can shoehorn that into the `agda-input-mode`...?

JacquesCarette: that would be nice (fat perp)

WolframKahl:
On Tue, Feb 25, 2025 at 05:20:24PM -0800, Jacques Carette wrote:
> that would be nice (fat perp)

Fat perp is apparently not on offer by Unicode, but see:

https://www.fileformat.info/info/unicode/char/search.htm?q=perp&han=Y

https://www.fileformat.info/info/unicode/char/search.htm?q=tack&han=Y

jamesmckinna:
In favour of `\perp`:
* easy to type
* mildly distinct visually from `\bot`
* definitely distinct visually/lexically from `_#_`

In favour of `_#_`
* semantically it is an apartness
* ... what else?

Regarding @WolframKahl 's links to Unicode, I wonder if perhaps we should make every module preamble comment include the new (infix/mixfix) symbols introduced with their character encodings (`emacs` and Unicode?), as 'self-documentation', rather than trying to build a big lookup table in `style-guide`?

JacquesCarette:
> I wonder if perhaps we should make every module preamble comment include the new (infix/mixfix) symbols introduced with their character encodings (emacs and Unicode?), as 'self-documentation', rather than trying to build a big lookup table in style-guide?

That seems like a good idea.

bsaul:
> > I wonder if perhaps we should make every module preamble comment include the new (infix/mixfix) symbols introduced with their character encodings (emacs and Unicode?), as 'self-documentation', rather than trying to build a big lookup table in style-guide?
> 
> That seems like a good idea.

Agreed.

WolframKahl:
On Wed, Feb 26, 2025 at 03:58:22AM -0800, jamesmckinna wrote:
> In favour of `\perp`:
> * [...]
> * mildly distinct visually from `\bot`

With my current fonts, I can see that when they are side-by-side,
but I wouldn't be able to distinguish individual occurrences.
I'd therefore argue against using both symbols in Agda.

jamesmckinna:
> On Wed, Feb 26, 2025 at 03:58:22AM -0800, jamesmckinna wrote: In favour of `\perp`: * [...] * mildly distinct visually from `\bot`
> With my current fonts, I can see that when they are side-by-side, but I wouldn't be able to distinguish individual occurrences. I'd therefore argue against using both symbols in Agda.

Hmm... that's a shame, I think. I'd maybe try to influence things further by adding
* `\perp` as a definition *is* lifting/proxying for the underlying appeal to `bot`, so 'it makes sense' (?) to use cognate symbols

But I'd better stop beating this particular donkey now.

MatthewDaggitt:
I have no strong opinions on the choice of symbol. We use quite a lot of close cognates in the library already, but I can understand the urge not to add more.

JacquesCarette (approved):
I will relent on the name.

jamesmckinna:
Merge conflicts resolved; additions improve level polymorphism, but also (hopefully) avoid potential unsolved metas when applying the `*-sym` proofs, but given my involvement, a second review/fresh pair of eyes would be helpful before merging.

src/Relation/Unary/Properties.agda
```diff
@@ -198,6 +199,27 @@ U-Universal = λ _ → _
...
+≬-symmetric = Product.map₂ swap
+
+⊥-symmetric : Sym {A = Pred A ℓ₁} {B = Pred A ℓ₂} _⊥_ _⊥_
+⊥-symmetric = _∘ swap
+
+≬-sym : Symmetric {A = Pred A ℓ₁} _≬_
+≬-sym = ≬-symmetric
+
+⊥-sym : Symmetric {A = Pred A ℓ₁} _⊥_
+⊥-sym = ⊥-symmetric
```
gallais, line 215: Why do we want two names for the same proofs?

jamesmckinna: Maybe I've misunderstood, but the `-symmetric` forms are more level polymorphic than the `-sym` ones?

jamesmckinna:
Very happy/happier still with fat perp over perp... however we end up supporting it?

jamesmckinna:
> I've opened https://github.com/agda/agda/pull/7791/files to add the relevant unicode symbols

Oh I see, `\bbot` is 'short' like `\bot`, rather than 'tall' like `\perp`. Hmmm...

jamesmckinna:
Once we can and *do* agree on the symbol, suggest updating the declaration of the symbol in the spirit of #2690 ...

jamesmckinna:
Can we (@JacquesCarette @gallais) agree this for v2.3?

PR #2599, opened 2025-02-19
[ refactor ] Redefine `Relation.Binary.Definitions.Adjoint`, plus knock-on `Function.Consequences`

### LINKED ISSUE #2581: [ refactor ] Reconcile `Relation.Binary.Definitions.Adjoint` and `Function.Definitions.Inverse*`
Meditating on #2580 (and some other recent PRs, including those which touch `Function.*`), it's hard not to observe the following:
* `Relation.Binary.Definitions.Adjoint` universally quantifies the product of two implications, rather than separating them out as distinct (universally quantified!) implicaitons, and then taking their product, cf. `Function.Definitions.Inverse` in terms of `Inverseˡ` and `Inverseʳ`
* That the latter properties (intended to be interpreted relative to  `Setoid`s which supply the 'equality' relations...) could, modulo `flip`-symmetry, be re-expressed as instances of the corresponding components of the above refactoring of `Relation.Binary.Definitions.Adjoint`

Accordingly: propose:
* separate out (`naming`?) `HalfLeftAdjoint` and `HalfRightAdjoint` as
  ```agda
  HalfLeftAdjoint HalfRightAdjoint : Rel A ℓ₁ → Rel B ℓ₂ → (A → B) → (B → A) → Set _
  HalfLeftAdjoint _≤_ _⊑_ f g = ∀ {x y} → (f x ⊑ y → x ≤ g y)
  HalfRightAdjoint _≤_ _⊑_ f g = ∀ {x y} → (x ≤ g y → f x ⊑ y)
  ...
  Adjoint _≤_ _⊑_ f g = (HalfLeftAdjoint _≤_ _⊑_ f g) × (HalfRightAdjoint _≤_ _⊑_ f g)
  ```
* redefine
  ```agda
  Inverseʳ : (A → B) → (B → A) → Set _
  Inverseʳ f g = HalfLeftAdjoint ≈₁ ≈₂ f g
  Inverseˡ : (A → B) → (B → A) → Set _
  Inverseˡ f g = HalfRightAdjoint ≈₁ ≈₂ f g
  ```

NB left<->right mismatch, as ever... cf. #2532 

Pro: DRY, refactoring `Function.Properties.Inverse.HalfAdjointEquivalence` (?), ... `Inverse` as a special case of `Adjoint`, hence... (?), ... etc.
Con: *very* `breaking`, and perhaps requires good upside-down-and-backwards spectacles through which to look at everything...

jamesmckinna:
Fixes #2581 .


jamesmckinna:
I'll return to the merge conflicts after v2.3, given that this is a breaking change.

src/Function/Consequences/Setoid.agda
```diff
@@ -16,13 +16,15 @@ module Function.Consequences.Setoid
...
+open import Relation.Binary.Definitions
+  using (HalfLeftAdjoint; HalfRightAdjoint; Adjoint)
 open import Relation.Nullary.Negation.Core using (¬_)
 
 import Function.Consequences as C
 
 private
-  open module S = Setoid S using () renaming (Carrier to A; _≈_ to ≈₁)
-  open module T = Setoid T using () renaming (Carrier to B; _≈_ to ≈₂)
+  open module S = Setoid S using () renaming (Carrier to A; _≈_ to ≈₁; sym to sym₁)
```
JacquesCarette, line 26: Not for this PR, but S,T to A,B and 1,2, is horrible. But it is consistent with what's already here, so I'll just review the added functionality and ignore the old style.

JacquesCarette (approved):

src/Function/Consequences/Setoid.agda
jamesmckinna: Nice observation: you're right that it would be better to use qualified names in preference to the renaming nonsense but that is 'old'/'legacy'-style, so we should fix that up (eventually, but could do now, esp. given yet-another merge conflict!)

jamesmckinna:
Given how many of these `breaking` changes arise from the analysis developed in specific issues, I've evolved my approach to the `CHANGELOG` by explicitly linking to those issues under each item. Feedback welcome on whether this is desirable or not!

At present, two existing such items are badly formatted, so at some point, we should fix those up, but I won't try to as part of this PR (though at some level, to me at least, it would 'make sense' to do so, to be consistent with the entry added here... ;-))

src/Function/Consequences/Setoid.agda
JacquesCarette: I'd much rather see such a fix in a separate PR. We don't want to mix the discussions, nor slow down a good fix because of arguing over another.

JacquesCarette (approved):
It is easy to remove these links later if we wish, *way harder* to add them post facto. So I'm all for having them in!

jamesmckinna:
Just fixed a typo, and with it the chance to look over everything once again. I'd like to merge this, but at that point, we basically will have duplicate families `Half*Adjoint` and `Inverse*`, so downstream we should try to reconcile everything under a single sensible naming scheme...

JacquesCarette (approved):

JacquesCarette:
I don't understand your comment about duplication? Yes, these are equivalent notions, but with rather different UX, so having them both is definitely a plus.

jamesmckinna:
Thanks @JacquesCarette re:duplication. Will not fiddle any further.

jamesmckinna:
@MatthewDaggitt still pending a review from you, but I can rescind that if you're swamped at the moment.

MatthewDaggitt (approved):

PR #2600, opened 2025-02-20
end construct import cleaning

jmougeot:


JacquesCarette (changes requested):
The changes themselves are good, but this also fails the whitespace check.

JacquesCarette (approved):

Taneb (approved):

PR #2604, opened 2025-02-20
[ fix ] in non-commutative settings distinguish _∣ˡ_ and _∣ʳ_

gallais:


JacquesCarette (approved):
Nice.

src/Data/List/Relation/Binary/Suffix/Homogeneous/Properties.agda
```diff
@@ -45,3 +45,20 @@ isDecPartialOrder dpo = record
...
   } where module DPO = IsDecPartialOrder dpo
+
+
+module _ {A : Set a} where
+
+  open import Data.List.Base using (List; []; _∷_)
+  open import Data.List.Properties using (++-monoid)
+  open import Algebra.Properties.Monoid.Divisibility (++-monoid A)
+  open import Relation.Binary.PropositionalEquality using (_≡_; refl; cong)
+  open import Relation.Binary using (_⇒_)
```
MatthewDaggitt, line 56: These imports should live at the top with `Algebra.Properties.Monoid.Divisibility` renamed as `MonoidDivisibility` and opened here?

src/Data/List/Relation/Binary/Suffix/Homogeneous/Properties.agda
```diff
@@ -45,3 +45,20 @@ isDecPartialOrder dpo = record
...
+
+  Suffix-as-∣ʳ : Suffix _≡_ ⇒ _∣ʳ_
```
MatthewDaggitt, line 58: Should these lemmas live in a new file `Suffix.Propositional` and `Suffix.Propositional.Properties`?

src/Data/List/Relation/Binary/Prefix/Homogeneous/Properties.agda
```diff
@@ -45,3 +46,20 @@ isDecPartialOrder dpo = record
...
+
+
+
+module _ {A : Set a} where
+
+  open import Data.List.Base using (List; []; _∷_)
+  open import Data.List.Properties using (++-monoid)
+  open import Algebra.Properties.Monoid.Divisibility (++-monoid A)
+  open import Relation.Binary.PropositionalEquality using (_≡_; refl)
+  open import Relation.Binary using (_⇒_)
```
MatthewDaggitt, line 58: Identical comments here.

MatthewDaggitt (changes requested):
Otherwise looks good!

src/Data/List/Relation/Binary/Suffix/Homogeneous/Properties.agda
gallais: Fair; moved everything to `Suffix.Propositional.Properties`.
I did not define `Suffix.Propositional` because, just like the non-existent `Suffix.Homogeneous`,
it would be a single definition of an alias for the type constructor, wouldn't it?

jamesmckinna:
Very nice!

JacquesCarette (approved):

MatthewDaggitt (approved):

src/Data/List/Relation/Binary/Suffix/Homogeneous/Properties.agda
MatthewDaggitt: Yup that seems fair!

gallais:
#2631 kicked up a conflict with this PR and the principled fix is ironically to stick with what I
have a thus reverse the refactoring?!

jamesmckinna:
> #2631 kicked up a conflict with this PR and the principled fix is ironically to stick with what I have and thus reverse the refactoring?!

I think I do not understand why this should be the 'principled' fix. One way to piggy-back on that PR is to propagate the `trans refl` optimisation to each of your new proofs, or have I really not understood what is going on here?

Separately (downstream, or expanding this PR?) we should perhaps consider/have already considered adding `Algebra.Properties.CommutativeMonoid.Divisibility`, and making *that* module (and `CommutativeMagma`/`CommutativeSemigroup` 'above' it) be the repository of properties of `_∣_`?

PR #2607, opened 2025-02-20
Import cleaning Module 1

jmougeot:


JacquesCarette (approved):

src/Algebra/Module/Definitions/Bi/Simultaneous.agda
```diff
@@ -6,13 +6,13 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Relation.Binary
+open import Relation.Binary using (Rel)
```
jamesmckinna, line 9: Oooh... I didn't manage to look at this before it was merged: should really be 
```suggestion
open import Relation.Binary.Core using (Rel)
```

jamesmckinna: Maybe the best thing is to raise yet another `import cleaning PR`, or else fold this in to an existing open one?

JacquesCarette: @jmougeot can you please explicitly comment on what you will / have done about this?

PR #2609, opened 2025-02-20
Import cleaning Module 3

jmougeot:


src/Algebra/Module/Structures/Biased.agda
```diff
@@ -26,7 +29,7 @@ private
     m ℓm r ℓr s ℓs : Level
     M : Set m
 
-module _ (commutativeSemiring : CommutativeSemiring r ℓr) where
+module _ (commutativeSemiring :   CommutativeSemiring r ℓr) where
```
JacquesCarette, line 32: extra space?

JacquesCarette (changes requested):

src/Algebra/Module/Structures/Biased.agda
jmougeot: fix-whitespace told me everything was good. Doesn’t it apply in this case?

JacquesCarette: No - it is just about trailing spaces and spaces on blank lines. Adding extra space after a `:` can be a proper stylistic choice in some cases -- but not here.

JacquesCarette (approved):

src/Algebra/Module/Bundles.agda
```diff
@@ -26,15 +26,18 @@
...
+  using (IsLeftSemimodule; IsRightSemimodule; IsBisemimodule;
+  IsSemimodule; IsLeftModule; IsRightModule; IsModule; IsBimodule)
 open import Algebra.Module.Definitions
-open import Algebra.Properties.Group
-open import Function.Base
-open import Level
+open import Function.Base using (flip)
+open import Level using (Level; _⊔_; suc)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Nullary    using (¬_)
```
jamesmckinna, line 42: Another uncaught 'clean': this `import` is removable entirely.

PR #2613, opened 2025-02-21
[Import] `Algebra.Properties`

jmougeot:


src/Algebra/Properties/BooleanAlgebra.agda
```diff
@@ -26,8 +26,7 @@ open BooleanAlgebra B
 
 import Algebra.Properties.DistributiveLattice as DistribLatticeProperties
 
-open import Algebra.Structures _≈_
-open import Relation.Binary
+open import Relation.Binary using (Rel)
```
JacquesCarette, line 29: I think `Rel` is defined in a much smaller module than `Relation.Binary`. It would be worth narrowing this down.

src/Algebra/Properties/CommutativeSemigroup.agda
```diff
@@ -13,7 +13,6 @@ module Algebra.Properties.CommutativeSemigroup
   where
 
 open CommutativeSemigroup CS
-
```
JacquesCarette, line 16: Actually, should keep this blank line: it is an `open` of something specific, which is different than a bunch of `import` of useful functionality.

src/Algebra/Properties/DistributiveLattice.agda
```diff
@@ -9,11 +9,10 @@
 -- Disabled to prevent warnings from deprecated names
 {-# OPTIONS --warn=noUserWarning #-}
 
-open import Algebra.Lattice.Bundles
-open import Algebra.Lattice.Structures.Biased
-open import Relation.Binary
+open import Algebra.Lattice.Bundles using (DistributiveLattice; Lattice)
+open import Algebra.Lattice.Structures.Biased using (isDistributiveLatticeʳʲᵐ)
+open import Relation.Binary using (Rel)
```
JacquesCarette, line 14: and of course my previous comments holds everywhere

JacquesCarette (changes requested):

JacquesCarette (approved):

src/Algebra/Properties/AbelianGroup.agda
```diff
@@ -6,7 +6,7 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra
+open import Algebra using (AbelianGroup)
```
jamesmckinna, line 9:
```suggestion
open import Algebra.Bundles using (AbelianGroup)
```
as elsewhere

src/Algebra/Properties/BooleanAlgebra.agda
```diff
@@ -26,8 +26,7 @@ open BooleanAlgebra B
 
 import Algebra.Properties.DistributiveLattice as DistribLatticeProperties
 
-open import Algebra.Structures _≈_
```
jamesmckinna, line 29: Should probably be placed lower down in the ordering? Or is it needed for subsequent `import`s?

jamesmckinna (changes requested):
Overall improvement to structure of `import`s, but consistency over importing `Algebra.Bundles` should also be kept in mind.

src/Algebra/Properties/BooleanAlgebra.agda
jamesmckinna: My mistake: this is another deprecated module, so there really shouldn't be any commits to it, unless/until it gives rise to errors during CI?

src/Algebra/Properties/BooleanAlgebra.agda
jamesmckinna: As below, though: these are commits to a DEPRECATED module, and as such, I think should be avoided unless absolutely necessary?

src/Algebra/Properties/BooleanAlgebra.agda
JacquesCarette: Right, as @jamesmckinna says, don't change any deprecated module, it is not worth it. These will eventually be deleted.

JacquesCarette (changes requested):

jamesmckinna (changes requested):
Beside reverting the changes to `Algebra.Properties.BooleanAlgebra`
(and reapplying the 'cleaning' to `Algebra.Lattice.Properties.BooleanAlgebra` ;-))
this all looks straightforward

jamesmckinna (changes requested):
Let me double down on my earlier review comments:
please revert ALL changes to 
- [x] `Algebra.Properties.BooleanAlgebra` (done I think)
- [x] `Algebra.Properties.DistributiveLattice`
- [ ] `Algebra.Properties.Lattice`

because DEPRECATED modules should NOT be touched during refactoring, unless they break CI, and AFAIU, this is *not* the case here.

JacquesCarette (approved):

src/Algebra/Properties/RingWithoutOne.agda
```diff
@@ -6,7 +6,7 @@
 
 {-# OPTIONS --cubical-compatible --safe #-}
 
-open import Algebra
+open import Algebra using (RingWithoutOne)
```
jamesmckinna, line 9:
```suggestion
open import Algebra.Bundles using (RingWithoutOne)
```

jamesmckinna (approved):
Have added one more (missing) `Bundles`

PR #2615, opened 2025-02-21
[Import] `Axiom.*`

jmougeot:


JacquesCarette (approved):

src/Axiom/UniquenessOfIdentityProofs.agda
```diff
@@ -9,10 +9,13 @@
...
-open import Relation.Binary.Core
-open import Relation.Binary.Definitions
-open import Relation.Binary.PropositionalEquality.Core
+open import Relation.Nullary.Decidable.Core using
+  (recompute; recompute-constant)
+open import Relation.Binary.Core using (Rel; _⇒_)
+open import Relation.Binary.Definitions using
+  (Sym; Irrelevant; DecidableEquality)
+open import Relation.Binary.PropositionalEquality.Core using
+  (_≡_; refl; trans; sym; cong)
```
jamesmckinna, line 18: What is the rational for having `using` on the line *above* the list of names?
At least if you instead write
```suggestion
open import Relation.Nullary.Decidable.Core
  using (recompute; recompute-constant)
open import Relation.Binary.Core using (Rel; _⇒_)
open import Relation.Binary.Definitions
  using (Sym; Irrelevant; DecidableEquality)
open import Relation.Binary.PropositionalEquality.Core
  using (_≡_; refl; trans; sym; cong)
```
then it's *not* a bug to comment out the `using` directives. This seems more robust.

src/Axiom/Extensionality/Heterogeneous.agda
```diff
@@ -9,10 +9,12 @@
 module Axiom.Extensionality.Heterogeneous where
 
 import Axiom.Extensionality.Propositional as P
+  using (Extensionality; lower-extensionality)
```
jamesmckinna, line 12: Inconsistency with your style of `using` on one line, and the list of used names on another below. I think what you have on L12 is 'better', see below.

jamesmckinna (changes requested):
Consistency is above all the most important style quality (even if "the hobgoblin of small minds"), but I think your changes/additions to `using` directives are going in the 'wrong' direction...

src/Axiom/UniquenessOfIdentityProofs.agda
JacquesCarette: That is an excellent suggestion! I think we should make that part of `style-guide`.

PR #2617, opened 2025-02-21
[Import] `Codata.Musical.*`

jmougeot:


JacquesCarette (approved):

src/Codata/Musical/Stream.agda
```diff
@@ -9,7 +9,7 @@
 module Codata.Musical.Stream where
 
 open import Level using (Level)
-open import Codata.Musical.Notation
+open import Codata.Musical.Notation using (♭; ∞; ♯_)
```
jamesmckinna, line 12: I think that this revision (and all the others under `Codata.*` is unnecessary, even undesirable, given that this module is simply a `public` re-exporting stub module for `Agda.Builtin.Coinduction`, so you're potentially making the `import` strategy/dependency structure of the library less robust/more viscous (wrt knock-on viscosity) than before?

jamesmckinna (changes requested):
I would seriously reconsider the wisdom of qualifying the `import` of `Codata.Musical.Notation` with a `using` directive throughout.

JacquesCarette (approved):

jamesmckinna (approved):
Thanks!

PR #2619, opened 2025-02-21
[Import] `Codata.Sized.*`

jmougeot:


JacquesCarette (approved):
Approving the code, but the CI has not finished yet.

src/Codata/Sized/Delay/Effectful.agda
```diff
@@ -8,13 +8,15 @@
 
 module Codata.Sized.Delay.Effectful where
 
-open import Codata.Sized.Delay
+open import Codata.Sized.Delay using
+  (Delay; now; never; later; map; bind; zipWith; alignWith)
```
jamesmckinna, line 12:
```suggestion
open import Codata.Sized.Delay
  using (Delay; now; never; later; map; bind; zipWith; alignWith)
```
as elsewhere in this suite of PRs.

jamesmckinna (changes requested):
I think that `using` directives should be on the same line as the names that are used.

jamesmckinna (approved):
Thanks!

PR #2622, opened 2025-02-24
[imports] `Data.Container.*` .. `Data.Fin.Subset.*`

jmougeot:


JacquesCarette (approved):

jamesmckinna:
As with all of your other PRs doing "import cleaning", I've reluctantly made peace with the very uninformative PR titles on GitHub, but this one appears to demonstrate what's going on: Without further editing, GitHib will take the *first* commit message as title for a PR... and when that commit is a *merge* commit, then we have really lost any sense of what is going on (plus illustrating that the PRs aren't being based against up-to-date `master`...)

For the rest of the 'cleaning' ones that you are posting, it serves no purpose to change the titles, but this one... I think I need a more informative title please!

jamesmckinna:
> Thanks for your feedback. How would you like me to handle the PR titles in the future? Should I change them to be more descriptive?

Well... you already have 9 that have been merged, and another 19 still open, so it seems a it late to change the naming policy now, ... but for the fact that even across those 28 PRs, there isn't precise consistency. 

My own view now is that there is not *much* to be done... about *this* collection of PRs. But in future, may I suggest that, as with any other academic writing, the 'title' should at least give the reader chance to understand approximately what the PR is about, without having to read the detail of the chnaned files...

Similarly, the initial 'preamble' comment on a PR ought ot be the place to flag both more detail, as well as any outstanding issues which might be controversial and/or out-of-scope for the PR.

src/Data/Fin/Subset/Properties.agda
```diff
@@ -9,8 +9,10 @@
 module Data.Fin.Subset.Properties where
 
 import Algebra.Definitions as AlgebraicDefinitions
-import Algebra.Structures as AlgebraicStructures
-import Algebra.Lattice.Structures as AlgebraicLatticeStructures
+import Algebra.Structures as AlgebraicStructures using (IsMagma; IsSemigroup; IsMonoid; IsBand;
+  IsCommutativeMonoid; IsIdempotentCommutativeMonoid)
+import Algebra.Lattice.Structures as AlgebraicLatticeStructures using
+  (IsSemilattice; IsLattice; IsDistributiveLattice; IsBooleanAlgebra)
```
jamesmckinna, line 15: Usual layout considerations regarding `using` directives

jamesmckinna (changes requested):
Looks great otherwise!

jamesmckinna (approved):
Thanks!

src/Data/Container/Morphism.agda
```diff
@@ -8,8 +8,8 @@
 
 module Data.Container.Morphism where
 
-open import Data.Container.Core
-import Function.Base as F
+open import Data.Container.Core using (Container; ⟦_⟧; _⇒_; shape; position)
+import Function.Base as F using (id; _∘′_)
```
jamesmckinna, line 12: I realise that this is something else that I do not understand: why are we naming definitions that nevertheless need to be mentioned within qualified scope, as eg `F.id`? What advantage is there to explicitly listing them? (maybe some redundant recoding/documentary commentary advantage?)

JacquesCarette: That's why *I* will sometimes do it. (There are also advantages when making copies of modules, but that's not the case here).  I had not asked for this particular use of explicit `using`, I think that's @jmougeot being 'thorough' while not noticing that these mean something slightly different than the usual cases.

PR #2624, opened 2025-02-25
[imports] `Data.List.Effectful.*` .. `Relation.Nullary`

jmougeot:
Refactor import statements for clarity and consistency, and introduce a new module for irrelevant types in the relation library.


src/Data/List/Fresh/Relation/Unary/All.agda
```diff
@@ -53,7 +54,7 @@ module _ {R : Rel A r} where
 
 module _ {R : Rel A r} {P : Pred A p} {Q : Pred A q} where
 
-  map : ∀ {xs : List# A R} → ∀[ P ⇒ Q ] → All P xs → All Q xs
+  map : ∀ {xs : List# A R} →  ∀[ P ⇒ Q ] → All P xs → All Q xs
```
JacquesCarette, line 57: stray space?

src/Relation/Nullary.agda
```diff
@@ -19,20 +18,14 @@ private
...
 open import Relation.Nullary.Negation.Core public
 open import Relation.Nullary.Reflects public hiding (recompute; recompute-constant)
 open import Relation.Nullary.Decidable.Core public
-
-------------------------------------------------------------------------
--- Irrelevant types
-
-Irrelevant : Set p → Set p
-Irrelevant P = ∀ (p₁ p₂ : P) → p₁ ≡ p₂
+open import Relation.Nullary.Irrelevant public
```
JacquesCarette, line 28: Why did you delete the definition of `Irrelevant`? Surely that's needed?

src/Relation/Nullary/Irrelevant.agda
```diff
@@ -0,0 +1,19 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Irrelevant types
```
JacquesCarette, line 4: Oh, I see. While in theory this is a good change, in practice this is an API change, which will break some people's code who assume that `Irrelevant`  is defined in `Relation.Nullary`.

You shouldn't do this in a "cleaning" PR. It should be done separately. And then we can discuss how to do it so that no outside code breaks.

JacquesCarette (changes requested):

src/Relation/Nullary/Irrelevant.agda
jmougeot: Reed helped me for this one

jamesmckinna: Yes. PLEASE DO NOT DO THIS.
As a general rule, we try to stick to the rule
"one PR for one purpose only", backed up by:
* starting with an issue, where the overall problem and the design choices underlying possible solutions can be discussed;
* identifying well defined parts that can be tackled as a single PR

This often has the (very!) desirable side effect of minimising the reviewing load of any given PR, especially if the 'big' design decisions have already been discussed.

By all means raise an issue about moving `Relation.Nullary.Irrelevant`, (and even develop an accompanying PR) but it's hard to see that this would have high priority at the moment. 

And as @JacquesCarette points out, it would be a breaking change to the library API, so at the very least, would require a lot more infrastructure (eg deprecation warnings and redefinitions, over and above import refactoring) before we would proceed...

jamesmckinna: And, as with previous comments, there should be some indication in the title and preamble to any OR that this is part of the intended/expected consequence of merging. But that's not the case here!

jamesmckinna: And FTR, I'm not against doing this, *in principle*, but not in this PR.

src/Data/List/Relation/Binary/Equality/Propositional.agda
```diff
@@ -15,9 +15,10 @@ open import Relation.Binary.Core using (_⇒_)
 module Data.List.Relation.Binary.Equality.Propositional {a} {A : Set a} where
 
 open import Data.List.Base
-import Data.List.Relation.Binary.Equality.Setoid as SetoidEquality
+import Data.List.Relation.Binary.Equality.Setoid as SetoidEquality using 
+    ([]; _≋_; ≋-refl; _∷_)
```
jamesmckinna, line 19: ditto.

src/Data/List/Relation/Binary/Infix/Heterogeneous.agda
```diff
@@ -8,13 +8,13 @@
...
-open import Level
+open import Level using (Level; _⊔_)
 open import Relation.Binary.Core using (REL; _⇒_)
 open import Data.List.Base as List using (List; []; _∷_; _++_)
 open import Data.List.Relation.Binary.Pointwise
   using (Pointwise)
-open import Data.List.Relation.Binary.Prefix.Heterogeneous
-  as Prefix using (Prefix; []; _∷_; _++ᵖ_)
+open import Data.List.Relation.Binary.Prefix.Heterogeneous as Prefix
+  using (Prefix; []; _∷_; _++ᵖ_)
```
jamesmckinna, line 17: Good change, nice!

src/Data/List/Relation/Binary/Infix/Homogeneous/Properties.agda
```diff
@@ -8,16 +8,16 @@
...
-open import Level
+open import Data.List.Relation.Binary.Pointwise as Pointwise using (Pointwise)
+open import Data.List.Relation.Binary.Infix.Heterogeneous using (Infix)
+open import Data.List.Relation.Binary.Infix.Heterogeneous.Properties
+  using (fromPointwise; infix?; trans; antisym)
 open import Function.Base using (_∘′_)
+open import Level using (Level; _⊔_)
 open import Relation.Binary.Core using (REL)
-open import Relation.Binary.Structures
+open import Relation.Binary.Structures
```
jamesmckinna, line 18: This will break the whitespace check on commit.
Use `fix-whitespace src/Data/List/Relation/Binary/Infix/Homogeneous/Properties.agda` to fix this.

src/Data/List/Relation/Binary/Lex/NonStrict.agda
```diff
@@ -12,20 +12,24 @@
...
-open import Level
-open import Relation.Nullary
+open import Level using (Level; _⊔_)
+open import Relation.Nullary.Negation using (¬_)
+open import Relation.Nullary using (yes; no)
 open import Relation.Binary.Core using (Rel; _⇒_)
 open import Relation.Binary.Bundles
-  using (Poset; StrictPartialOrder; DecTotalOrder; StrictTotalOrder; Preorder)
+  using (Poset; StrictPartialOrder; DecTotalOrder; StrictTotalOrder;
+  Preorder)
```
jamesmckinna, line 26: Note sure this change is worth it.

src/Data/List/Relation/Binary/Lex/NonStrict.agda
```diff
@@ -12,20 +12,24 @@
...
 open import Relation.Binary.Structures
-  using (IsEquivalence; IsPartialOrder; IsStrictPartialOrder; IsTotalOrder; IsStrictTotalOrder; IsPreorder; IsDecTotalOrder)
+  using (IsEquivalence; IsPartialOrder; IsStrictPartialOrder; IsTotalOrder;
+  IsStrictTotalOrder; IsPreorder; IsDecTotalOrder)
```
jamesmckinna, line 29: semicolons/linebreaks, as elsewhere

src/Data/List/Relation/Binary/Lex/NonStrict.agda
```diff
@@ -12,20 +12,24 @@
...
 open import Relation.Binary.Definitions
-  using (Irreflexive; _Respects₂_; Antisymmetric; Asymmetric; Symmetric; Transitive; Decidable; Total; Trichotomous)
+  using (Irreflexive; _Respects₂_; Antisymmetric; Asymmetric; Symmetric;
+  Transitive; Decidable; Total; Trichotomous)
```
jamesmckinna, line 32: ditto

src/Relation/Nullary/Irrelevant.agda
TOTBWF: As part of the culprit here, I fail to see how this is a breaking change. `Relation.Nullary` re-exports the definition, so it should be completely transparent to downstream users. The only possible breaking change would be if someone wrote their own `Relation.Nullary.Irrelevant`, but the same argument could be made for adding _any_ module.

TOTBWF: As an aside, `WeaklyDecidable` should also probably go; importing a module for a type alias seems like a bad idea.

MatthewDaggitt: So this is a backwards compatible change and doesn't need a deprecating warning because as @TOTBWF says users are still accessing the _same_ definition from the same module. The time when we need a deprecation warning is when we have to duplicate the definition when we move it in order to avoid dependency cycles.

But yes, this should be done as a separate PR :+1:

src/Data/List/Relation/Binary/Lex/NonStrict.agda
jmougeot: I worked on the wrong branch by mistake and didn't want these changes to appear in this PR or be published yet. So I redid everything cleanly in the next PR, removing the changes here.  Sorry about that.

src/Data/List/Fresh/Relation/Unary/Any/Properties.agda
```diff
@@ -58,7 +58,7 @@ module _ {R : Rel A r} {P : Pred A p} {Q : Pred A q} (P? : Decidable P) where
   ¬All⇒Any : {xs : List# A R} → ¬ (All P xs) → Any (∁ P) xs
   ¬All⇒Any {xs = []}      ¬ps = ⊥-elim (¬ps [])
   ¬All⇒Any {xs = x ∷# xs} ¬ps with P? x
-  ... |  true because  [p] = there (¬All⇒Any (¬ps ∘′ (invert [p] ∷_)))
+  ... | true because  [p] = there (¬All⇒Any (¬ps ∘′ (invert [p] ∷_)))
```
jamesmckinna, line 61: `true` and `false` are of different lengths, so *some* alignment is required if we want (the usually desirable?) alignment of the ` = definiens` part of each `with` clause. What's the right thing to do? Not the change you've made as it stands, i think.

src/Relation/Nary.agda
```diff
@@ -15,18 +15,20 @@ module Relation.Nary where
...
+open import Data.Empty using (⊥; ⊥-elim)
 open import Data.Nat.Base using (zero; suc)
 open import Data.Product.Base as Product using (_×_; _,_)
 open import Data.Product.Nary.NonDependent
 open import Data.Sum.Base using (_⊎_)
 open import Function.Base using (_$_; _∘′_)
 open import Function.Nary.NonDependent
 open import Relation.Nullary.Negation using (¬_)
-open import Relation.Nullary.Decidable as Dec using (Dec; yes; no; _because_; _×-dec_)
+open import Relation.Nullary.Decidable as Dec
```
jamesmckinna, line 28:
```suggestion
open import Relation.Nullary.Decidable.Core as Dec
```

jamesmckinna (changes requested):
yet more nitpicking, sorry.

JacquesCarette (approved):

jamesmckinna (approved):

src/Relation/Nullary/Irrelevant.agda
jamesmckinna: And such PR is now #2676 ?

PR #2625, opened 2025-02-27
[Import] `Data.List.*`

jmougeot:


jamesmckinna:
Something's broken concerning (missing import?) `Pointwise` equality on lists.

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
```diff
@@ -8,13 +8,14 @@
 
 module Data.List.Relation.Binary.Disjoint.Propositional.Properties where
 
+open import Agda.Builtin.List using (List; []; _∷_)
```
JacquesCarette, line 11: We shouldn't **add** imports to `Agda.Builtin.List` (or any other `Agda.Builtin`. These should only be imported in `Data.X.Base`, which then re-exports them. So `using (List; []; _∷_)` should be moved to `Data.List.Base` below.

src/Data/List/NonEmpty/Effectful.agda
```diff
@@ -8,14 +8,14 @@
 
 module Data.List.NonEmpty.Effectful where
 
-open import Agda.Builtin.List
+open import Agda.Builtin.List using (List; _∷_; [])
```
JacquesCarette, line 11: Going directly into `Data.Builtin.List` is not a good idea here, it should use `Data.List.Core` instead.

JacquesCarette (changes requested):

src/Data/List/NonEmpty/Effectful.agda
jamesmckinna: @JacquesCarette you meant `Agda.Builtin.List` and `Data.List.Base`, I think?

Not quite sure why this module seemed to be originally based off `Agda.Builtin.List`, there are usually only two (acceptable!?) reasons for that:
* breaking a dependency cycle (occasionally) 
* a MWE mindest of using only 'minimal commitments' to dependency (almost never now)

In any case, @jmougeot : please do not touch *any* imports from `Agda.Builtin.*` (nor from modules marked altogether as `DEPRECATED`...)

src/Data/List/Membership/DecPropositional.agda
```diff
@@ -13,8 +13,6 @@ module Data.List.Membership.DecPropositional
 
 open import Relation.Binary.PropositionalEquality.Properties using (decSetoid)
 
-open import Relation.Binary.PropositionalEquality.Properties using (decSetoid)
-
```
jamesmckinna, line 17: Wow! Clearly no-one's looked at this in a while... good catch.

src/Data/List/NonEmpty/Instances.agda
```diff
@@ -8,9 +8,10 @@
 
 module Data.List.NonEmpty.Instances where
 
-open import Data.List.NonEmpty.Effectful
-import Data.List.NonEmpty.Effectful.Transformer as Trans
-
+open import Data.List.NonEmpty.Effectful using
+  (functor; applicative; monad; comonad)
+import Data.List.NonEmpty.Effectful.Transformer as Trans using
+  (functor; applicative; monad; monadT)
```
jamesmckinna, line 14: Sigh... usual comment about `using`

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
jamesmckinna: Yes, this looks like legacy code.
If the module imports *both* `Agda.Builtin.List` *and* `Data.List.Base` then something is wrong. 
if the `Builtin` import can be deleted safely, please do so!

src/Data/List/Relation/Binary/Infix/Heterogeneous/Properties.agda
```diff
@@ -8,28 +8,29 @@
...
 open import Relation.Binary.PropositionalEquality.Core using (_≢_; refl; cong)
-open import Data.List.Relation.Binary.Pointwise.Base as Pointwise using (Pointwise)
-open import Data.List.Relation.Binary.Infix.Heterogeneous
-open import Data.List.Relation.Binary.Prefix.Heterogeneous
-  as Prefix using (Prefix; []; _∷_)
-import Data.List.Relation.Binary.Prefix.Heterogeneous.Properties as Prefix
-open import Data.List.Relation.Binary.Suffix.Heterogeneous
-  as Suffix using (Suffix; here; there)
+open import Data.List.Relation.Binary.Pointwise.Base as Pointwise
+  using (Pointwise)
```
jamesmckinna, line 32: Curious as to why this import is here, rather than above the block of `Relation.*` imports?

jamesmckinna (changes requested):
Mostly 'the usual' comments about layout style, `using` etc.
Otherwise looks fine.

src/Data/List/NonEmpty/Effectful.agda
JacquesCarette: Yes, typo in both cases!!! Serves me right for doing this fairly late after a long day.

I'm pretty sure this use of `Agda.Builtin.List` is a mistake - but you're right that it likely should be cleaned up in a separate PR.

jamesmckinna, line 11:
```suggestion
open import Data.List.Base using (List; []; _∷_)
```

Appears to be the correct fix.

jamesmckinna (approved):
Removed the appeal to `Agda.Builtin`, otherwise all looks fine.

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
jamesmckinna: I think this was marked as resolved prematurely

jamesmckinna:
```suggestion
```

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
```diff
@@ -8,13 +8,14 @@
...
 open import Level using (Level)
 open import Function.Base using (_∘_)
 open import Function.Bundles using (_⇔_)
-open import Data.List.Base
-open import Data.List.Relation.Binary.Disjoint.Propositional
-import Data.List.Relation.Unary.Any as Any
-open import Data.List.Relation.Unary.All as All
+open import Data.List.Base using (concat; deduplicate)
```
jamesmckinna:
```suggestion
open import Data.List.Base using (List; []; _∷_; concat; deduplicate)
```

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
jamesmckinna: My mistake, my browser had been showing me cached version of these pages... sigh.

jamesmckinna: Please ignore!

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
jamesmckinna: This suggestion is now out-of-sync.

src/Data/List/Relation/Binary/Disjoint/Propositional/Properties.agda
```diff
@@ -8,18 +8,19 @@
...
 
-open import Level using (Level)
-open import Function.Base using (_∘_)
-open import Function.Bundles using (_⇔_)
-open import Data.List.Base
-open import Data.List.Relation.Binary.Disjoint.Propositional
-import Data.List.Relation.Unary.Any as Any
-open import Data.List.Relation.Unary.All as All
+
+open import Data.List.Base using (concat; deduplicate; List; []; _∷_)
```
jamesmckinna, line 12:
```suggestion
open import Data.List.Base using (List; []; _∷_; concat; deduplicate)
```
By convention, types and constructors should always precede functions defined on them!

src/Data/List/NonEmpty/Effectful.agda
jamesmckinna: Actually, I took the executive decision to resolve this as part of your 'cleanup': precisely *because* the overall strategy of narrowing imports via `using` directives actually resolves a non-trivial dependency/ambiguity issue which AFAICT was responsible for the original appeal to `Agda.Builtin`. A separate PR would have been excessive, I think.

JacquesCarette (approved):

JacquesCarette (approved):

PR #2627, opened 2025-02-27
[ new ] module Data.Sign.Show

gallais:


Taneb (approved):
Seems a pretty reasonable addition

src/Data/Sign/Show.agda
```diff
@@ -0,0 +1,16 @@
...
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Sign.Show where
+
+open import Data.Sign.Base using (Sign; +; -)
+open import Data.String.Base using (String)
+
+show : Sign → String
+show + = "+"
+show - = "-"
```
jamesmckinna, line 16: Now that we are being *much* more explicit about `import`s, esp. wrt `using` directives, is this improved, or worsened by writing instead
```suggestion
open import Data.Sign.Base using (module Sign)
open import Data.String.Base using (String)

open Sign

show : Sign → String
show + = "+"
show - = "-"
```
now that `data` declarations give rise to corresponding `module` ones?

jamesmckinna (approved):
Approved, despite being troubled by my own question...

src/Data/Sign/Show.agda
gallais: It'd be a complete departure from the style we have adopted so far so I think
it should be discussed in its own issue.

Issue #2629, opened 2025-02-27
Optimising `Setoid`s/reasoning by 'rewriting' otherwise higher-dimensional equalities


jamesmckinna:
For `Relation.Binary.PropositionalEquality`, we have (at least) the following definitional equalities
* `sym refl = refl`
* `trans refl = id` (but `trans p refl` is only *propositionally* equal to `p`...)
* `resp P refl = id`
* `cong f refl = refl`
* etc.

whereas for `Setoid`s, we can form the same LHS combinations (or their *mutatis mutandis* variants, modulo suitable assumptions about respectfulness etc.), but for which we do not even necessarily have the above equalities even as provable equalities between proofs ... never mind any higher-dimensional coherent iterations of such ideas a la HoTT.

But there are various places where it might indeed be useful/more efficient (eg in proofs of divisibility in `Algebra`) to *optimise* such combinations, *as if* those equations *did* hold, moreover definitionally, *and* without regard to the definitional biases in `trans` etc. Example: in [`Algebra.Properties.Magma.Divisibility`](https://github.com/agda/agda-stdlib/blob/547a06a63e45d6c68bf834c34abeed7c24a489c4/src/Algebra/Properties/Magma/Divisibility.agda#L41-L42) we see
```agda
xy≈z⇒y∣z : ∀ x y {z} → x ∙ y ≈ z → y ∣ z
xy≈z⇒y∣z x y xy≈z = ∣-respʳ-≈ xy≈z (x∣yx y x)
```
Now, inlining the definitions yields `xy≈z⇒y∣z x y xy≈z = x , trans refl xy≈z` which we may then, by *fiat*, rewrite as `xy≈z⇒y∣z x _ xy≈z = x , xy≈z` as a RHS with the correct type, moreover one which has better reduction behaviour now that we have removed the blocking non-redex `trans refl` etc.
(Similar examples are available for all the various combinations of left/right respects, left/right transitivity etc.)

Proposal: to go through the library in search of such 'locally optimisable' RHS of definitions in terms of `Setoid` combinators, and suitably 'optimise' them. Cf. @JacquesCarette 's #2288

JacquesCarette:
While I'm generally in favour, I'd also like to know: what function is responsible for introducing the `trans refl`? It feels like it's the "guilty party" here and should be fixed. Without knowing that, I'd have a hard time searching through the library for such potential optimisations!

jamesmckinna:
[...]

PR #2631, opened 2025-02-28
[ refactor ] `Algebra.Properties.Magma.Divisibility`

jamesmckinna:
Fixes:
* the inciting incident from #2629 
* an uncaught use of a deprecated name

### DIFF
--- a/src/Algebra/Properties/CommutativeMagma/Divisibility.agda
+++ b/src/Algebra/Properties/CommutativeMagma/Divisibility.agda
@@ -28,7 +28,7 @@ x∣xy : ∀ x y → x ∣ x ∙ y
 x∣xy x y = y , comm y x
 
 xy≈z⇒x∣z : ∀ x y {z} → x ∙ y ≈ z → x ∣ z
-xy≈z⇒x∣z x y xy≈z = {!∣-respʳ-≈ xy≈z (x∣xy x y)!}
+xy≈z⇒x∣z x y xy≈z = ∣-respʳ-≈ xy≈z (x∣xy x y)
 
 x|xy∧y|xy : ∀ x y → (x ∣ x ∙ y) × (y ∣ x ∙ y)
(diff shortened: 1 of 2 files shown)


src/Algebra/Properties/CommutativeMagma/Divisibility.agda
```diff
@@ -28,7 +28,7 @@ x∣xy : ∀ x y → x ∣ x ∙ y
 x∣xy x y = y , comm y x
 
 xy≈z⇒x∣z : ∀ x y {z} → x ∙ y ≈ z → x ∣ z
-xy≈z⇒x∣z x y xy≈z = ∣-respʳ xy≈z (x∣xy x y)
+xy≈z⇒x∣z x y xy≈z = ∣-respʳ-≈ xy≈z (x∣xy x y)
```
JacquesCarette, line 31: I don't understand why this doesn't have the same redundancy problem as the one in `Magma`?

jamesmckinna: Good question! (Probably an additional appeal to `comm`?)

jamesmckinna: Indeed, the normal form of the RHS here is simply `y , trans (comm y x) xy≈z`, so there isn't the same opportunity for `trans refl` to 'reduce'.

jamesmckinna: In respect of the originating issue #2629 , though, ... examples such as these potentially give rise to subsequent 'redexes' of the form `trans (comm y x) (trans (comm x y) ...)` which should also be 'reduced' away.

JacquesCarette: Agreed: but is there anything left to do here? I don't think so. So I'll mark this one as resolved.

JacquesCarette (approved):

Taneb (approved):

PR #2632, opened 2025-02-28
[ refactor ] Define and use `Relation.Binary.Definitions.Refl`

### LINKED ISSUE #2628: [ DRY ] Add `Relation.Binary.Definitions.Refl`
Given the legacy design choice of 
```agda
-- Reflexivity - defined without an underlying equality. It could
-- alternatively be defined as `_≈_ ⇒ _∼_` for some equality `_≈_`.

-- Confusingly the convention in the library is to use the name "refl"
-- for proofs of Reflexive and `reflexive` for proofs of type `_≈_ ⇒ _∼_`,
-- e.g. in the definition of `IsEquivalence` later in this file. This
-- convention is a legacy from the early days of the library.

Reflexive : Rel A ℓ → Set _
Reflexive _∼_ = ∀ {x} → x ∼ x
```

I wonder if we shouldn't simply accept things as they are, but then go on to introduce
```agda
Refl : Rel A ℓ₁ → Rel A ℓ₂ →  → Set _
Refl _≈_ _∼_ = _≈_ ⇒ _∼_
```
(with an `{-# INLINE #-}` directive?), and thus avoid the endless `import` of `Relation.Binary.Core using (_⇒_)` whenever we want to state and prove `reflexive` properties of relations?

This would at least be moderately consistent with our current ambient policy of having 'generalised' definitions involving `REL` (here: `_⇒_`) then delegating to specialised ones involving `Rel`.

It would also reserve the name against any possible (unimaginable now?) downstream refactoring to swap the two definitions over, so that `refl` would become a proof of `Refl`, and `reflexive` a proof of `Reflexive` etc.

Cosmetic; possibly useless; but to this tormented mind, somehow 'more streamlined'...

jamesmckinna:
Fixes #2628

Probably sheer folly without also making the swap of `Refl` and `Reflexive`, but this could be the enabling groundwork for that.

DRAFT for comment/discussion; will only fix merge conflicts if it's worth pursuing at all.

### DIFF
--- a/src/Algebra/Properties/Magma/Divisibility.agda
+++ b/src/Algebra/Properties/Magma/Divisibility.agda
@@ -39,7 +39,7 @@ x∣yx : ∀ x y → x ∣ y ∙ x
 x∣yx x y = y , refl
 
 xy≈z⇒y∣z : ∀ x y {z} → x ∙ y ≈ z → y ∣ z
-xy≈z⇒y∣z x y xy≈z = ∣-respʳ-≈ xy≈z (x∣yx y x)
+xy≈z⇒y∣z x y xy≈z = {!∣-respʳ-≈ xy≈z (x∣yx y x)!}
 
 ------------------------------------------------------------------------
 -- Properties of non-divisibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,3 +149,8 @@ Additions to existing modules
   ```agda
   filter-↭ : ∀ (P? : Pred.Decidable P) → xs ↭ ys → filter P? xs ↭ filter P? ys
   ```
+
+* In `Relation.Binary.Definitions`:
+  ```agda
+  Refl : Rel A ℓ → Rel A ℓ → Set _
+  ```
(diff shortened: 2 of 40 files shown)


src/Algebra/Properties/Magma/Divisibility.agda
```diff
@@ -39,7 +39,7 @@ x∣yx : ∀ x y → x ∣ y ∙ x
 x∣yx x y = y , refl
 
 xy≈z⇒y∣z : ∀ x y {z} → x ∙ y ≈ z → y ∣ z
-xy≈z⇒y∣z x y xy≈z = ∣-respʳ-≈ xy≈z (x∣yx y x)
+xy≈z⇒y∣z x y xy≈z = {!∣-respʳ-≈ xy≈z (x∣yx y x)!}
```
JacquesCarette, line 42: oops: left in a hole (in something that seems irrelevant?)

JacquesCarette (changes requested):

JacquesCarette:
I'm not feeling this one at all. It feels like a solid step in exactly the wrong direction. Doubling-down on historically bad names, and replacing a nice infix by an more obscure prefix, combine to, indeed, 'sheer folly'.

src/Algebra/Properties/Magma/Divisibility.agda
jamesmckinna: Oops, wrong branch I think

PR #2636, opened 2025-03-03
docs: proof-reading the tutorial

ulidtko:
Hi, Agda team!

Here's a small patch in the tutorial, `doc/README*`, with minor editorials: typos and the like.

I know it's always easier for (somebody else's) fresh set of eyeballs to do proof-reads. Thus, here goes.

Hoping for easy merge — there's no code change, just the docstrings.

### DIFF
--- a/doc/README.agda
+++ b/doc/README.agda
@@ -113,7 +113,7 @@ import README.IO
 -- • Tactic
 --     Tactics for automatic proof generation
 
--- ∙ Text
+-- • Text
 --     Format-based printing, Pretty-printing, and regular expressions
 
 
--- a/doc/README/Design/Hierarchies.agda
+++ b/doc/README/Design/Hierarchies.agda
@@ -34,7 +34,7 @@ private
 --   ∙ Relation.Binary
 --   ∙ Relation.Binary.Indexed
 --
--- A given hierarchy `X` is always split into 4 seperate folders:
+-- A given hierarchy `X` is always split into 4 separate folders:
 --   ∙ X.Core
 --   ∙ X.Definitions
 --   ∙ X.Structures
@@ -66,7 +66,7 @@ private
 
 -- The Core module contains the basic units of the hierarchy.
 
--- For example for binary relations these are homoegeneous and
+-- For example, in binary relations these are homogeneous and
 -- heterogeneous binary relations:
 
 REL : Set a → Set b → (ℓ : Level) → Set (a ⊔ b ⊔ suc ℓ)
@@ -90,8 +90,7 @@ Op₂ A = A → A → A
 -- The Definitions module defines the various properties that the
 -- basic units of the hierarchy may have.
 
--- For example in Relation.Binary this includes reflexivity,
--- transitivity etc.
+-- Examples in Relation.Binary include reflexivity, transitivity, etc.
 
 Reflexive : Rel A ℓ → Set _
 Reflexive _∼_ = ∀ {x} → x ∼ x
@@ -105,7 +104,7 @@ Transitive _∼_ = ∀ {x y z} → x ∼ y → y ∼ z → x ∼ z
 Total : Rel A ℓ → Set _
 Total _∼_ = ∀ x y → x ∼ y ⊎ y ∼ x
 
--- For example in Algebra these are associativity, commutativity.
+-- Examples in Algebra include associativity, commutativity.
 -- Note that all definitions for Algebra are based on some notion of
 -- underlying equality.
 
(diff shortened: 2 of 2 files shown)


ulidtko:
BTW @andreasabel, thanks for reminding me!

This has been collecting dust for... turns out, exactly a year. I was initially meaning to go over all readme modules; alas, that hadn't happened. Uneasy times.

doc/README.agda
```diff
@@ -113,7 +113,7 @@ import README.IO
 -- • Tactic
 --     Tactics for automatic proof generation
 
--- ∙ Text
+-- • Text
```
JacquesCarette, line 116: what is the change here? A different unicode dot?

ulidtko: Yes; this one was an oddball, the diff makes it consistent with other heading-level dots:
> -- • Level
-- • Reflection
-- • Relation
-- • Size
-- • Strict
-- ∙ Text        --- was
-- • Tactic   --- fixed

JacquesCarette (approved):

doc/README/Design/Hierarchies.agda
```diff
@@ -236,17 +234,13 @@ record Semigroup : Set (suc (a ⊔ ℓ)) where
   magma : Magma a ℓ
   magma = record { isMagma = isMagma }
 
--- Note that the Semigroup record does not include a Magma field.
```
Taneb, line 239: Why was this paragraph removed?

ulidtko: @Taneb well, two reasons.

One: it seemed a bit misplaced... Just below, there's detailed exposition of `Bundle re-exporting`.

Two: it looks possibly outdated; talks about `record Semigroup` having a `"repackaging function" semigroup which converts a Magma to a Semigroup`  — meanwhile, what's actually there in the definition, is the opposite: a function `magma` that projects the implied Magma out of the Semigroup.

Perhaps I misunderstood; please confirm. I can recover the paragraph easily, it just didn't read to me as correct neither helpful.

doc/README/Design/Hierarchies.agda
```diff
@@ -66,7 +66,7 @@ private
 
 -- The Core module contains the basic units of the hierarchy.
 
--- For example for binary relations these are homoegeneous and
+-- For example, in binary relations these are homogeneous and
```
jamesmckinna, line 69: Comma seems worthwhile, but not sure about the increase in legibility/intelligibility arising from the change in/of preposition, maybe going all out and writing
```suggestion
-- For example, in the case of binary relations these are homogeneous and
```
(possibly with knock-on consequences for the line-breaking/layout)

jamesmckinna (approved):
Thanks for offering this. I think it would be good to nail down the discussion with @Taneb before merging, though.

doc/README/Design/Hierarchies.agda
jamesmckinna: Actually, I think there's preferable solution to this problem, namely to rewrite the whole paragraph so as to explain (hopefully: better!) what's actually going on ... but for the unfortunate lack of distinction between a 'primitive'/'definitional' `field` and what elsewhere in PL would be called a 'manifest field', *viz.* the definitions made either explicitly (as with the `magma` example at hand), or by being brought into scope by an `open` declaration (as with `isMagma` on which it depends)

jamesmckinna: Suggest, therefore instead, as a drop-in replacement for the deleted paragraph (GitHub won't apparently let me add this as a `suggestion`, alas):
```agda
-- Note that the Semigroup record does not include a (primitive; definitional) Magma field, by contrast with the `IsSemigroup` structure which *does* include an `isMagma` field as primitive.
-- Instead, the Semigroup record includes an additional declaration (a 'manifest field' of the `record`) defining a `Magma` bundle, in terms of that exported `isMagma` field.
-- In this way, 'inheritance' is *automatic* for the `IsX` sub*structures* of a given bundle, while supporting the *optional* export of inherited sub*bundles*.
```

Now, I fully expect colleagues to object to my rephrasing... ;-) which is why I haven't tried to push such a commit to this PR without further discussion!

jamesmckinna (changes requested):
Suggest rewriting (rather than deleting) the paragraph in favour of (some version of) the suggestion above.

doc/README/Design/Hierarchies.agda
jamesmckinna: @Taneb @ulidtko if you're happy with the suggested changes, please say so, or else offer alternatives.
propose that we try to merge sooner rather than later?

jamesmckinna:
Six weeks later, let me ask explicitly again for comments on my proposed rephrasing...
@Taneb @MatthewDaggitt @gallais

MatthewDaggitt:
Sorry for the long reply time, yes that rephrasing sounds fine to me @jamesmckinna, thanks for suggesting!

jamesmckinna:
Thanks @MatthewDaggitt ... paragraph (tweaked) now committed.

MatthewDaggitt (approved):

MatthewDaggitt:
Thanks for the tweaks @ulidtko and @jamesmckinna, merging in now.

PR #2643, opened 2025-03-05
[Import] `Reflection`

jmougeot:


src/Reflection/AST/Universe.agda
```diff
@@ -8,12 +8,12 @@
 
 module Reflection.AST.Universe where
 
-open import Data.List.Base             using (List)
-open import Data.String.Base           using (String)
-open import Data.Product.Base          using (_×_)
-open import Reflection.AST.Argument    using (Arg)
+open import Data.List.Base using (List)
+open import Data.String using (String)
```
JacquesCarette, line 12: why remove the `.Base` here?

JacquesCarette (changes requested):

JacquesCarette:
Strictly speaking, I personally like lining up the `using`. But it is not done consistently through the library (or even inside single files), so I'm okay with removing it. Others might not agree.

jamesmckinna:
> Strictly speaking, I personally like lining up the `using`. But it is not done consistently through the library (or even inside single files), so I'm okay with removing it. Others might not agree.

I think the most robust way to 'preserve' the alignment is to lay off `using` directives on new lines?
Others have observed that the alignment (on the same line) is not, in general, going to be/stay robust...

JacquesCarette (approved):

PR #2647, opened 2025-03-06
Add Dec for unit and empty types

carlostome:
Adds missing `Dec` values for unit type (nullary product type) and empty type (nullary sum type).


gallais (commented):
Looks good; please add a CHANGELOG entry!

Taneb (approved):

PR #2652, opened 2025-03-11
[Refactor] Change ⊥-elim to contradiction in `src/Codata/Guarded/Stream/Relation/Unary/Any`

jmougeot:


JacquesCarette (approved):
First PR in using `contradiction` over `⊥-elim` when it makes sense (such as here) for clearer code.

Taneb (approved):
This is an obvious win here

PR #2658, opened 2025-03-12
[Refractor] contradiction over ⊥-elim in commutes-with-∪ def

jmougeot:


jamesmckinna:
Duplicated by #2663 ?

JacquesCarette (approved):
Does look like there's duplication, so I'll approve this one and then I think the other will end up conflicting if it really is the same.

PR #2660, opened 2025-03-12
[Refractor] contradiction over ⊥-elim in as def

jmougeot:


jamesmckinna (changes requested):
More narrowing possible

JacquesCarette (approved):

jamesmckinna (approved):
Looks good!

PR #2661, opened 2025-03-12
[Refractor] contradiction over ⊥-elim in ¬-not def

jmougeot:


src/Data/Bool/Properties.agda
```diff
@@ -657,10 +657,10 @@ not-¬ {true}  refl ()
 not-¬ {false} refl ()
 
 ¬-not : ∀ {x y} → x ≢ y → x ≡ not y
-¬-not {true}  {true}  x≢y = ⊥-elim (x≢y refl)
+¬-not {true}  {true}  x≢y = contradiction refl x≢y
```
JacquesCarette, line 660: I wonder if this means that Agda would already know that `x≢y` is impossible, i.e. if you replace it by `()` on the left, it might typecheck as is. Same below.

JacquesCarette (changes requested):

src/Data/Bool/Properties.agda
jamesmckinna: Same as my remark on #2664 ... nope, Agda can't do it :-(

Taneb (approved):

JacquesCarette (approved):

PR #2664, opened 2025-03-12
[Refractor] contradiction over ⊥-elim `src/Data/Vec/Functional/Properties.agda`

jmougeot:


src/Data/Vec/Functional/Properties.agda
```diff
@@ -70,7 +70,7 @@ updateAt-updates (suc i) xs = updateAt-updates i (tail xs)
 
 updateAt-minimal : ∀ (i j : Fin n) {f : A → A} (xs : Vector A n) →
                    i ≢ j → updateAt xs j f i ≡ xs i
-updateAt-minimal zero    zero    xs 0≢0 = ⊥-elim (0≢0 refl)
+updateAt-minimal zero    zero    xs 0≢0 = contradiction refl 0≢0
```
JacquesCarette, line 73: as with a previous one: likely this can be made even simpler as I would think Agda would know that `0≢0` is impossible.

JacquesCarette (changes requested):

src/Data/Vec/Functional/Properties.agda
jamesmckinna: Sadly, I think because we are working without K, Agda offers instead:
```agda
zero ≢ zero should be empty, but that's not obvious to me
when checking the clause left hand side
updateAt-minimal zero zero xs ()
```

jamesmckinna: But given that this PR (and others?) rests on the `contradiction refl 0≢0` pattern for (an obviously false) `0≢0`, it's tempting to want to abstract over this, rather than endlessly repeat the pattern...
... in `Relation.Binary.PropositionalEquality.Properties` for example?

jamesmckinna (approved):
But see comment about abstracting out the `x≢x → Whatever` pattern.

src/Data/Vec/Functional/Properties.agda
JacquesCarette: As this PR is a straight improvement, I'll approve, but I agree that a second pass that rests on abstracting out `contradiction refl` with something defined in that file is a good idea. Something for @jmougeot to tackle.

JacquesCarette (approved):

PR #2665, opened 2025-03-12
[Refractor] contradiction over ⊥-elim in  antisymmetric  def

jmougeot:


src/Data/List/Relation/Binary/Lex.agda
```diff
@@ -8,23 +8,24 @@
 
 module Data.List.Relation.Binary.Lex where
 
-open import Data.Empty using (⊥; ⊥-elim)
 open import Data.Unit.Base using (⊤; tt)
```
jamesmckinna, line 11: This line can also be removed (so I'm wondering how/why it escaped the last round of `import` simplification?)

jmougeot: The reason is that there is no warning for ⊥ and ⊥-elim indicating that they are not used in the file while it's the case for the other importations

JacquesCarette (approved):

src/Data/List/Relation/Binary/Lex.agda
```diff
@@ -116,3 +117,4 @@ module _ {a ℓ₁ ℓ₂} {A : Set a} {P : Set}
     decidable (x ∷ xs) []       = no λ()
     decidable (x ∷ xs) (y ∷ ys) =
       Dec.map ∷<∷-⇔ (dec-≺ x y ⊎-dec (dec-≈ x y ×-dec decidable xs ys))
+
```
jamesmckinna, line 120:
```suggestion
```
as elsewhere, please dont add blank lines/change existing whitespace

jamesmckinna (changes requested):

jamesmckinna (approved):

PR #2667, opened 2025-03-12
[Refractor] contradiction over ⊥-elim in

jmougeot:


Taneb (approved):

JacquesCarette (changes requested):
remove import of `⊥-elim `? Rest looks good.

jmougeot:
I can't remove it in this file since it's still use in the file : 
 ∈-remove x∈xs y∈xs x≉y = fromInj₂ (⊥-elim ∘′ x≉y) (remove-inv x∈xs y∈xs)

JacquesCarette (approved):

JacquesCarette:
you might be able to replace that with `flip contradiction` but I'm not sure if that would make things clearer. So I think it best to leave it as is, at least for this pass.

PR #2672, opened 2025-03-13
[Refractor] contradiction over ⊥-elim in  `Data/List/Relation/Binary/Heterogeneous/Propreties`

jmougeot:


JacquesCarette (approved):

jamesmckinna (approved):
Similar to my remarks about the `contradiction refl  ...` pattern, I think it might be a better intervention (downstream) to abstract over proofs of the form `contradiction x<x ...` when `_<_ is an `Irreflexive` relation...

PR #2674, opened 2025-03-13
[Refractor] contradiction over ⊥-elim in `Data/List/Relation/Binary/Prefix/Heterogenous/Propreties`

jmougeot:


JacquesCarette (approved):

src/Data/List/Relation/Binary/Prefix/Heterogeneous/Properties.agda
```diff
@@ -8,27 +8,29 @@
...
+open import Function.Base using (flip; _∘_; _$_)
 open import Relation.Binary.Core using (Rel; REL; _⇒_)
 open import Relation.Binary.Definitions
   using (Trans; Antisym; Irrelevant; Decidable)
-open import Relation.Binary.PropositionalEquality.Core using (_≡_; _≢_; refl; cong₂)
+open import Relation.Binary.PropositionalEquality.Core
+  using (_≡_; _≢_; refl; cong₂)
+open import Relation.Nullary.Decidable.Core as Dec
+  using (_×-dec_; yes; no; _because_)
+open import Relation.Nullary.Negation.Core using (contradiction; ¬_)
```
jamesmckinna, line 32:
```suggestion
open import Relation.Nullary.Negation.Core using (¬_; contradiction)
```
This import ought to be appear in the same form every time it occurs, and that form moreover should have `using` order follow dependency order?

jamesmckinna (changes requested):
This suite of PRs should aim for consistency at the global level in terms of a stereotype for the additional imports... (?)

jamesmckinna (approved):
Thanks!

PR #2677, opened 2025-03-15
Match on _IsRelatedTo_ later for the PartialSetoid Reasoning

Ailrun:
## Main Motivation of This PR
The current version does not work in the following case, for example:
```
begin A ≈⟨ some-partial-equiv ⟩
      B ≡⟨ some-complex-eq ⟩
      C ∎
```
when `some-complex-eq` is not definitionally equal to [`refl`](https://agda.github.io/agda-stdlib/master/Agda.Builtin.Equality.html#207). Because of that, [`step`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Syntax.html#11989) of the noncomputing syntax will not be evaluated, and in turn, [`∼-go`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Base.Partial.html#1185) does not produce any [`multiStep`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Base.Partial.html#1131) constructor. This blocks the decision procedure [`IsMultiStep?`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Base.Partial.html#1702).

One can make this work using `subst`, but more complex reasoning becomes very nasty with that.

This PR makes the above example work.

## Related Issue
One related, remaining issue is that the reversed case still does not work:
```
begin A' ≡⟨ some-complex-eq' ⟩
      B' ≈⟨ some-partial-equiv' ⟩
      C' ∎
```
As `some-complex-eq` is not definitionally equal to [`refl`](https://agda.github.io/agda-stdlib/master/Agda.Builtin.Equality.html#207), [`step`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Syntax.html#11989) of the noncomputing syntax will not be evaluated. Thus, like before, [`IsMultiStep?`](https://agda.github.io/agda-stdlib/master/Relation.Binary.Reasoning.Base.Partial.html#1702) cannot decide whether this reasoning contains multiple steps or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,18 @@ Deprecated names
   *ₗ-assoc+comm⇒*ₗ-*ᵣ-assoc   ↦  *ₗ-assoc∧comm⇒*ₗ-*ᵣ-assoc
   ```
 
+* In `Algebra.Modules.Structures.IsLeftModule`:
+  ```agda
+  uniqueˡ‿⁻ᴹ   ↦  Algebra.Module.Properties.LeftModule.inverseˡ-uniqueᴹ
+  uniqueʳ‿⁻ᴹ   ↦  Algebra.Module.Properties.LeftModule.inverseʳ-uniqueᴹ
+  ```
+
+* In `Algebra.Modules.Structures.IsRightModule`:
+  ```agda
+  uniqueˡ‿⁻ᴹ   ↦  Algebra.Module.Properties.RightModule.inverseˡ-uniqueᴹ
+  uniqueʳ‿⁻ᴹ   ↦  Algebra.Module.Properties.RightModule.inverseʳ-uniqueᴹ
+  ```
+
 * In `Algebra.Properties.Magma.Divisibility`:
   ```agda
   ∣∣-sym       ↦  ∥-sym
@@ -92,6 +104,12 @@ Deprecated names
   ∣-trans    ↦  ∣ʳ-trans
   ```
 
+* In `Algebra.Structures.Group`:
+  ```agda
+  uniqueˡ-⁻¹   ↦  Algebra.Properties.Group.inverseˡ-unique
+  uniqueʳ-⁻¹   ↦  Algebra.Properties.Group.inverseʳ-unique
+  ```
+
 * In `Data.List.Base`:
   ```agda
   and       ↦  Data.Bool.ListAction.and
@@ -118,6 +136,8 @@ Deprecated names
 
 New modules
 -----------
+
+* `Algebra.Module.Properties.{Bimodule|LeftModule|RightModule}`.
 
 * `Data.List.Base.{and|or|any|all}` have been lifted out into `Data.Bool.ListAction`.
 
@@ -159,6 +179,12 @@ Additions to existing modules
   kleeneAlgebra                   : KleeneAlgebra c ℓ → KleeneAlgebra (a ⊔ c) (a ⊔ ℓ)
   quasiring                       : Quasiring c ℓ → Quasiring (a ⊔ c) (a ⊔ ℓ)
   commutativeRing                 : CommutativeRing c ℓ → CommutativeRing (a ⊔ c) (a ⊔ ℓ)
+  ```
+
+* In `Algebra.Modules.Properties`:
+  ```agda
+  inverseˡ-uniqueᴹ : x +ᴹ y ≈ 0ᴹ → x ≈ -ᴹ y
+  inverseʳ-uniqueᴹ : x +ᴹ y ≈ 0ᴹ → y ≈ -ᴹ x
   ```
 
 * In `Algebra.Properties.Magma.Divisibility`:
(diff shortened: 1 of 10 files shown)


jamesmckinna:
I would personally find an example helpful here to illustrate the discussion. 

Specifically, to understand in your hypothetical first situation if, and then why, the following reasoning would not work instead:
```agda
begin A ≈⟨ some-partial-equiv ⟩
      B ≈⟨ some-complex-eq ⟩
      C ∎
```

It may be that some (other) plumbing may be required, such as observing that reflexivity holds in PERs at any `B` such that `p : A ≈ B` for some `p`, `A`, and whether or not that might also work in the second situation?

Certainly, the library has no examples of reasoning in PERs that I'm aware of, so it probably hasn't been stress tested enough for the situations you sketch.

Ailrun:
@jamesmckinna `some-complex-eq` is a proof of propositional equality, not an equivalence I am using, so it is impossible to use `≈⟨⟩` as far as I know. Maybe I should have clarified that I have a PER different from the propositional equality.

jamesmckinna:
( I updated my posting to try to make clearer my intentions)

Ailrun:
@jamesmckinna The problem there is, `_≈_` is a PER, so `B ≡ C` does not give `B ≈ C`. That entire thing works only because there is
`A ≈ B` already where we can apply a substitution on (namely, substitute `C` for `B` in `A ≈ B` using `B ≡ C`).

Ailrun:
@jamesmckinna One can prove something like
```
begin A ≈⟨ some-partial-equiv ⟩
      B ≈⟨ right-reflexivity some-partial-equiv some-complex-eq ⟩
      C ∎
```
which derive `B ≈ C` out of `A ≈ B` and `B ≡ C`, but then the problem is when we have multiple steps of equalities:
```
begin A ≈⟨ some-partial-equiv ⟩
      B ≡⟨ eq1 ⟩
      C ≡⟨ eq2 ⟩
      D ≡⟨ eq3 ⟩
      E ≡⟨ eq4 ⟩
      F ≡⟨ eq5 ⟩
      G ∎
```
Now, to change `eq5` into `F ≈ G`, we need to derive `X ≈ F` for some `X` (or the symmetric one), but if no other information is given, the only obvious such proof is applying `subst` on `some-partial-equiv` to change `B` with `F`, which is basically the whole proof.

jamesmckinna:
@Ailrun thanks very much for the clarification, and sorry that my efforts last night weren't quite enough to make myself clear. 

What I had in mind was the following plumbing, for PER `_≈_`:
```agda
   trans-reflˡ : ∀ {x y z} → x ≡ y → y ≈ z → x ≈ z
  trans-reflˡ ≡.refl p = p

  trans-reflʳ : ∀ {x y z} → x ≈ y → y ≡ z → x ≈ z -- by analogy with `Relation.Binary.PropositionalEquality.trans-reflʳ`
  trans-reflʳ p ≡.refl = p

  p-reflˡ : ∀ {x y} → x ≈ y → x ≈ x -- 'partial reflexivity on the left'
  p-reflˡ p = trans p (sym p)

  p-reflʳ : ∀ {x y} → x ≈ y → y ≈ y
  p-reflʳ p = trans (sym p) p

  p-reflexiveˡ : ∀ {x y z} → x ≈ y → x ≡ z → x ≈ z
  p-reflexiveˡ p ≡.refl = p-reflˡ p

  p-reflexiveʳ : ∀ {x y z} → x ≈ y → y ≡ z → y ≈ z
  p-reflexiveʳ p ≡.refl = p-reflʳ p
```
and then to replace your
```agda
begin A ≈⟨ p ⟩
      B ≡⟨ q ⟩
      C ∎
```
by
```agda
begin A ≈⟨ p ⟩
      B ≈⟨ p-reflexiveʳ p q ⟩
      C ∎
```
and the second example similarly (?). 

But your question does indeed draw attention to problems with the current setup! And that my suggestions are certainly not *per se* part of the current `Reasoning` setup. The chains of reasoning argument is also quite compelling, even if `B ≡ F` would be provable by a sub-chain of `_≡_`-reasoning. 

I've encountered something similar in the `Pointwise` theory of `List`, but 'solved' it by appeals to such reflexivity wrappers.

Does the solution in the current PR completely solve the problem, however? I'd be happier if we had did have such...

Ailrun:
To be honest, I don't think this PR's solution is really a complete solution. I intended to give a solution with a minimal impact in the sense that it does not "reduce" the reduction behavior. If we want a more complete solution, my guess is changing `_IsRelated_` to track equalities as well (I believe preorder case has such tracking). That change would solve both of the "problematic" cases I mentioned in the original post.

MatthewDaggitt:
> To be honest, I don't think this PR's solution is really a complete solution. I intended to give a solution with a minimal impact in the sense that it does not "reduce" the reduction behavior. If we want a more complete solution, my guess is changing _IsRelated_ to track equalities as well (I believe strict order case has such tracking). That change would solve both of the "problematic" cases I mentioned in the original post.

I agree with your analysis. Changing `_IsRelated_` to track equalities seems like the more principled solution. Do you think you would have the time to have a crack at it @Ailrun?

Ailrun:
@MatthewDaggitt Unfortunately not within a month. After a month I might have some more time.

Ailrun:
Actually, I have some time to fix that now. See #2689.

PR #2678, opened 2025-03-17
[ add ] `Relation.Binary.Properties.PartialSetoid`

jamesmckinna:
This adds the properties of PERs discussed in #2677 .

Issues:
* naming: `p-` prefix to signify 'partial'
* where to add: 
   - `Properties` (as here)
   - `Consequences` (contra: dependency on `PropositionalEquality`)
   - or as manifest fields of `Relation.Binary.Structures.IsPartialEquivalence` (contra: every `IsEquivalence` gets bloated by the addition)

Downstream:
* potential refactoring of `Relation.Binary.Properties.Setoid` to inherit from this?

### DIFF
--- a/src/Relation/Binary/Properties/PartialSetoid.agda
+++ b/src/Relation/Binary/Properties/PartialSetoid.agda
@@ -30,18 +30,18 @@ trans-reflˡ ≡.refl p = p
 trans-reflʳ : RightTrans _≈_ _≡_
 trans-reflʳ p ≡.refl = p
 
-p-reflˡ : x ≈ y → x ≈ x
-p-reflˡ p = trans p (sym p)
+partial-reflˡ : x ≈ y → x ≈ x
+partial-reflˡ p = trans p (sym p)
 
-p-reflʳ : x ≈ y → y ≈ y
-p-reflʳ p = trans (sym p) p
+partial-reflʳ : x ≈ y → y ≈ y
+partial-reflʳ p = trans (sym p) p
 
-p-refl : x ≈ y → x ≈ x × y ≈ y
-p-refl p = p-reflˡ p , p-reflʳ p
+partial-refl : x ≈ y → x ≈ x × y ≈ y
+partial-refl p = partial-reflˡ p , partial-reflʳ p
 
-p-reflexiveˡ : x ≈ y → x ≡ z → x ≈ z
-p-reflexiveˡ p ≡.refl = p-reflˡ p
+partial-reflexiveˡ : x ≈ y → x ≡ z → x ≈ z
+partial-reflexiveˡ p ≡.refl = partial-reflˡ p
 
-p-reflexiveʳ : x ≈ y → y ≡ z → y ≈ z
-p-reflexiveʳ p ≡.refl = p-reflʳ p
+partial-reflexiveʳ : x ≈ y → y ≡ z → y ≈ z
+partial-reflexiveʳ p ≡.refl = partial-reflʳ p
 
--- a/src/Relation/Binary/Properties/PartialSetoid.agda
+++ b/src/Relation/Binary/Properties/PartialSetoid.agda
@@ -12,6 +12,7 @@ module Relation.Binary.Properties.PartialSetoid
   {a ℓ} (S : PartialSetoid a ℓ) where
 
 open import Data.Product.Base using (_,_; _×_)
+open import Relation.Binary.Definitions using (LeftTrans; RightTrans)
 open import Relation.Binary.PropositionalEquality.Core as ≡ using (_≡_)
 
 open PartialSetoid S
@@ -23,10 +24,10 @@ private
 ------------------------------------------------------------------------
 -- Proofs for partial equivalence relations
 
-trans-reflˡ : x ≡ y → y ≈ z → x ≈ z
+trans-reflˡ : LeftTrans _≡_ _≈_
 trans-reflˡ ≡.refl p = p
 
-trans-reflʳ : x ≈ y → y ≡ z → x ≈ z
+trans-reflʳ : RightTrans _≈_ _≡_
 trans-reflʳ p ≡.refl = p
 
 p-reflˡ : x ≈ y → x ≈ x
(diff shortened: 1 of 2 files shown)


src/Relation/Binary/Properties/PartialSetoid.agda
```diff
@@ -0,0 +1,46 @@
...
+------------------------------------------------------------------------
+-- Proofs for partial equivalence relations
+
+trans-reflˡ : x ≡ y → y ≈ z → x ≈ z
+trans-reflˡ ≡.refl p = p
+
+trans-reflʳ : x ≈ y → y ≡ z → x ≈ z
+trans-reflʳ p ≡.refl = p
+
+p-reflˡ : x ≈ y → x ≈ x
```
MatthewDaggitt, line 32: What does `p` stand for here?

src/Relation/Binary/Properties/PartialSetoid.agda
```diff
@@ -0,0 +1,46 @@
...
+open PartialSetoid S
+
+private
+  variable
+    x y z : Carrier
+
+------------------------------------------------------------------------
+-- Proofs for partial equivalence relations
+
+trans-reflˡ : x ≡ y → y ≈ z → x ≈ z
```
MatthewDaggitt, line 26: We could use the generalised `Trans` here?

src/Relation/Binary/Properties/PartialSetoid.agda
jamesmckinna: see above! under "issues: naming" ;-)

src/Relation/Binary/Properties/PartialSetoid.agda
jamesmckinna: yes!

JacquesCarette (approved):

JacquesCarette:
I'm actually fine with the names as they are.

src/Relation/Binary/Properties/PartialSetoid.agda
MatthewDaggitt: My vote would be for expanding these out to `partial`, as I don't think they are guessable from the `p` alone?

src/Relation/Binary/Properties/PartialSetoid.agda
```diff
@@ -0,0 +1,47 @@
...
+p-reflˡ : x ≈ y → x ≈ x
+p-reflˡ p = trans p (sym p)
+
+p-reflʳ : x ≈ y → y ≈ y
+p-reflʳ p = trans (sym p) p
+
+p-refl : x ≈ y → x ≈ x × y ≈ y
+p-refl p = p-reflˡ p , p-reflʳ p
+
+p-reflexiveˡ : x ≈ y → x ≡ z → x ≈ z
```
MatthewDaggitt, line 42: This is `Trans _≈_ _≡_ _≈_` (and similarly below), suggesting that `reflexive` may not be quite the right name? Similarly is this not just a variant of `subst`?

src/Relation/Binary/Properties/PartialSetoid.agda
```diff
@@ -0,0 +1,47 @@
...
+
+private
+  variable
+    x y z : Carrier
+
+------------------------------------------------------------------------
+-- Proofs for partial equivalence relations
+
+trans-reflˡ : LeftTrans _≡_ _≈_
+trans-reflˡ ≡.refl p = p
```
MatthewDaggitt, line 28: Do we actually need to match on the proof here? Is this not just a variant of `subst`?

MatthewDaggitt:
`Properties` seems the right place to me

src/Relation/Binary/Properties/PartialSetoid.agda
jamesmckinna: Happy to do so!

src/Relation/Binary/Properties/PartialSetoid.agda
jamesmckinna: Hmmm... I see what you mean... Will ponder this for a bit!

src/Relation/Binary/Properties/PartialSetoid.agda
JacquesCarette: I think `partial-` is better than `p-`. I do like to import a variety of things qualified, so I would personally drop the `partila-` completely, but I'm definitely not going to insist on that.

src/Relation/Binary/Properties/PartialSetoid.agda
JacquesCarette: This is not `Trans _≈_ _≡_ _≈_` because the second `x` would be a `y` in that case.

This is an extremely weird property which basically says: if x is related to something (anything) and we have evidence of that, and x and z are identical, then we can get evidence that x and z are related.

I'm at a loss for what to name it.

jamesmckinna: I don't know about 'weird'... these properties are characteristic of being a PER rather than a full `IsEquivalence`!?

jamesmckinna: As to names: I could call it (something like) `conditional-reflexive`, but then we would lose that the `≡.refl` instantiation of such a lemma should be `partial-refl`... so those lemmas would also need to be renamed. And then... names too long?

JacquesCarette: Weird likely because I generically find PERs weird.

However, your comment hits the real issue on the head: _only on its domain of definition_. We really ought to define an 'is defined' construction (a la `Trans` and `Symmetric`, etc) and use it here. Then I would have understood that you were asking for some indirect evidence of being defined. Then we'll be in a decent position to think about names.

At least now I understand this combinator's meaning (when x is defined, then equality that involve it is reflexive), and that is no longer weird at all.

jamesmckinna: Not sure I see clearly what the `IsDefined` construction would look like, and whether it would be 'free-standing', or tied to being in a PER? 

Cf. Scott's "Identity and Existence in Intuitionistic Logic" (LNM 753) which explores variations on the theme of `E t` vs. `t = t` as expressions of '`t` is defined'/'`t` exists', but does seem tied to `_=_` being an 'equality' (Leibniz-like) relation...?

PR #2688, opened 2025-04-01
[Add] Consequences of associativity for `Semigroup`s

jmougeot:
Introduce a new module for equational reasoning in semigroups, providing utilities for associativity reasoning and operations.


src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Equational reasoning for semigroups
+-- (Utilities for associativity reasoning, pulling and pushing operations)
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra using (Semigroup)
```
JacquesCarette, line 10: tighter import?

JacquesCarette (changes requested):

JacquesCarette:
For others: the names are taken from the combinators in `agda-categories`. Those were inherited from older versions of that library. I do *not* like them! So please do suggest better ones.

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@ New modules
 
 * `Data.Sign.Show` to show a sign
 
+* `Algebra.Reasoning.SemiGroup` adding reasoning combinator for semigroup
```
jamesmckinna, line 126:
```suggestion
* `Algebra.Reasoning.SemiGroup` adding reasoning combinators for semigroups
```

src/Algebra/Reasoning/Semigroup.agda
jamesmckinna, line 10:
```suggestion
open import Algebra.Bundles using (Semigroup)
```

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+module Pulls (ab≡c : a ∙ b ≈ c) where
+    pullʳ : ∀ {x} → (x ∙ a) ∙ b ≈ x ∙ c
+    pullʳ {x = x} = begin
+        (x ∙ a) ∙ b ≈⟨ assoc x a b ⟩
+        x ∙ (a ∙ b) ≈⟨ ∙-cong refl ab≡c ⟩
+        x ∙ c       ∎
+
+    pullˡ : ∀ {x} → a ∙ (b ∙ x) ≈ c ∙ x
+    pullˡ {x = x} = begin
+        a ∙ (b ∙ x) ≈⟨ sym (assoc a b x) ⟩
```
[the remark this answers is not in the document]
jamesmckinna: Here:
```suggestion
        a ∙ (b ∙ x) ≈⟨ assoc a b x ⟨
```

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+  α = ((a ∙ b) ∙ c) ∙ d
+  β = (a ∙ (b ∙ c)) ∙ d
+  γ = (a ∙ b) ∙ (c ∙ d)
+  δ = a ∙ ((b ∙ c) ∙ d)
+  ε = a ∙ (b ∙ (c ∙ d))
+
+  Only reassociations that need two assoc steps are defined here.
+  -}
+  assoc²αδ : ((a ∙ b) ∙ c) ∙ d ≈ a ∙ ((b ∙ c) ∙ d)
+  assoc²αδ = trans (∙-cong (assoc a b c) refl) (assoc a (b ∙ c) d)
```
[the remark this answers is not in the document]
jamesmckinna: As in my comments elsewhere, I think `w x y z` ought to be the preferred variable names, rather than `a b c d`, which (tend to be) reserved for `Level` variables... not sure what `style-guide` recommends here?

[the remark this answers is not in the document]
jamesmckinna: As to @JacquesCarette 's
> For the purposes of the standard library, I'm thinking that all of these should be equational proofs?

Actually, with one eye on #2629 , I think the opposite, as most of these are 'simple' instances of `trans`, esp. when written with `∙-congˡ` and `∙-congʳ`...

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+    pull-first : ∀ {x y} → a ∙ ((b ∙ x) ∙ y) ≈ c ∙ (x ∙ y)
+    pull-first {x = x} {y = y} = begin
+      a ∙ ((b ∙ x) ∙ y) ≈⟨ ∙-cong refl (assoc b x y) ⟩
+      a ∙ (b ∙ (x ∙ y)) ≈⟨ pullˡ ⟩
+      c ∙ (x ∙ y)       ∎
+
+    pull-center : ∀ {x y} → x ∙ (a ∙ (b ∙ y)) ≈ x ∙ (c ∙ y)
+    pull-center {x = x} {y = y} = ∙-cong refl (pullˡ)
+
+    -- could be called pull₃ʳ
```
jamesmckinna, line 98: But MUST not, IMNSVHO, because how on earth will a reader/user remember such names, without an additional mode d'emploi by their side?

jamesmckinna (changes requested):
Lots of comments, I'm requesting changes based on a single one, but can be summarised as:
*please* give lemmas names that can be remembered without needing a lookup table or other instruction manual to explain what they are/how they behave!

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+  assoc²εα : a ∙ (b ∙ (c ∙ d)) ≈ ((a ∙ b) ∙ c) ∙ d
+  assoc²εα = sym (trans (assoc (a ∙ b) c d) (assoc a b (c ∙ d)))
+
+  assoc²εβ : a ∙ (b ∙ (c ∙ d)) ≈ (a ∙ (b ∙ c)) ∙ d
+  assoc²εβ = sym (trans (assoc a (b ∙ c) d) (∙-cong refl (assoc b c d)))
+
+open Assoc4 public
+
+module Pulls (ab≡c : a ∙ b ≈ c) where
+    pullʳ : ∀ {x} → (x ∙ a) ∙ b ≈ x ∙ c
```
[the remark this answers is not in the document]
jamesmckinna: Also, the use of `⇒` in the lemma names would avoid the need for the submodules here to be named; they could (more) simply be anonymous `module _ (hyps) where`...

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+
+-- operate on the left part, then the right part
+center⁻¹ : a ∙ b ≈ c → x ∙ y ≈ z →  a ∙ ((b ∙ x) ∙ y) ≈ c ∙ z
+center⁻¹ {a = a} {b = b} {c = c} {x = x} {y = y} {z = z} eq eq′ = begin
+  a ∙ ((b ∙ x) ∙ y) ≈⟨ ∙-cong refl (pullʳ eq′) ⟩
+  a ∙ (b ∙ z)       ≈⟨ pullˡ eq ⟩
+  c ∙ z             ∎
+
+push-center : a ∙ b ≈ c → x ∙ (c ∙ y) ≈ x ∙ (a ∙ (b ∙ y))
+push-center eq = sym (pull-center eq)
```
jamesmckinna, line 133: So... each of these is parametrised on an assumption of the form `a ∙ b ≈ c`, so why is this not abstracted out as a `module` parameter?

jamesmckinna:
```suggestion
-- operate in the 'center' instead (like pull/push)

module _ (a∙b≈c : a ∙ b ≈ c) where

  center : (d ∙ a) ∙ (b ∙ e) ≈ d ∙ (c ∙ e)
  center = pullʳ (pullˡ a∙b≈c)

... etc.
```

jamesmckinna: But see elsewhere about 'better' names!

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra using (Semigroup)
+
+module Algebra.Reasoning.Semigroup {o ℓ} (S : Semigroup o ℓ) where
+
+open Semigroup S
+    using (Carrier; _∙_; _≈_; setoid; trans ; refl; sym; assoc; ∙-cong)
```
jamesmckinna, line 15: Indentation is by *2* spaces, not *4* as you have here... see `style-guide`.
```suggestion
  using (Carrier; _∙_; _≈_; setoid; trans ; refl; sym; assoc; ∙-cong; ∙-congˡ; ∙-congʳ)
```

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+      x ∙ (y ∙ c)       ∎
+
+open Pulls public
+
+module Pushes (ab≡c : a ∙ b ≈ c) where
+    pushʳ : x ∙ c ≈ (x ∙ a) ∙ b
+    pushʳ {x = x} = begin
+        x ∙ c       ≈⟨ sym (∙-cong refl ab≡c) ⟩
+        x ∙ (a ∙ b) ≈⟨ sym (assoc x a b) ⟩
+        (x ∙ a) ∙ b ∎
```
jamesmckinna, line 112: Successive appeals to `sym` suggest that it should be lifted out (permutation of `sym` through proofs), but in fact doing so quickly reveals the following:
```suggestion
 pushʳ = sym (pullʳ ab≡c)
```

src/Algebra/Reasoning/Semigroup.agda
```diff
@@ -0,0 +1,157 @@
...
+    pull-last : ∀ {x y} → (x ∙ (y ∙ a)) ∙ b ≈ x ∙ (y ∙ c)
+    pull-last {x = x} {y = y} = begin
+      (x ∙ (y ∙ a)) ∙ b ≈⟨ assoc x (y ∙ a) b ⟩
+      x ∙ ((y ∙ a) ∙ b) ≈⟨ ∙-cong refl (pullʳ {x = y}) ⟩
+      x ∙ (y ∙ c)       ∎
+
+open Pulls public
+
+module Pushes (ab≡c : a ∙ b ≈ c) where
+    pushʳ : x ∙ c ≈ (x ∙ a) ∙ b
```
jamesmckinna, line 108: Again: too much indentation

jamesmckinna:
OK... thanks for the recent commits, but I'm going to carry on nitpicking:
* the `Assoc4` proofs miss *lots* of symmetry, with 5 'independent' lemmas each with their symmetric counterpart, and since `Reasoning.Setoid` *does* have the syntax for applying the symmetric equivalent of an equation, we maybe don't even need the 5 equivalents...
  - `u∙[[v∙w]∙x]≈[[u∙v]∙w]∙x = sym [[u∙v]∙w]∙x≈u∙[[v∙w]∙x]`
  - `u∙[v∙[w∙x]]≈[[u∙v]∙w]∙x = sym [[u∙v]∙w]∙x≈u∙[v∙[w∙x]]`
  - `[u∙v]∙[w∙x]≈[u∙[v∙w]]∙x = sym [u∙[v∙w]]∙x≈[u∙v]∙[w∙x]`
  - `u∙[v∙[w∙x]]≈[u∙[v∙w]]∙x = sym [u∙[v∙w]]∙x≈u∙[v∙[w∙x]]`
  - `u∙[v∙w]]∙x]≈[u∙v]∙[w∙x] = sym [u∙v]∙[w∙x]≈u∙[[v∙w]]∙x]`
  where this last one, `[u∙v]∙[w∙x]≈u∙[[v∙w]]∙x]` is defined as a step-by-step proof, rather than the others, just defined in terms of `trans`, so could instead consider
  - `[u∙v]∙[w∙x]≈u∙[[v∙w]]∙x] = trans (assoc u v (w ∙ x)) (sym (∙-congˡ (assoc v w x)))`
* similarly, while the `Pushes` are symmetric equivalents of something in `Pulls`, not all the `Pulls` have their counterparts... but as above, maybe (probably) we don't even need the `Pushes` at all...
* the square brackets do improve things, I think, but you've overdone them, as there's no need to have each *innermost* group neither bracketed, nor mention the operation explicitly, cf. `x∙yz≈xy∙z`, so I should apologise for not expressing my suggestion clearly enough (or else falling into the same trap myself!), so eg. `[[u∙v]∙w]∙x≈u∙[[v∙w]∙x]` could be simplified to `[uv∙w]∙x≈u∙[vw∙x]` and that would be, I think, much more readable (with less ink!). Or even `[[uv]w]x≈u[[vw]x]` but that exchanges one character for two, so is perhaps suboptimal. Etc.

jamesmckinna:
Also:
> Introduce a new module for equational reasoning in semigroups, providing utilities for associativity reasoning and operations.

As with previous comments, we don't need a new module (even if it's moved location), these could all go in `Algebra.Properties.Semigroup`, and *should*, I think (another instance of Occam's Eraser: don't multiply `module`s unnecessarily)

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,139 @@
...
+open import Algebra.Bundles using (Semigroup)
+
+module Algebra.Properties.Semigroup.Reasoning {o ℓ} (S : Semigroup o ℓ) where
+
+open import Algebra.Structures using (IsMagma)
+
+open Semigroup S
+    using (Carrier; _∙_; _≈_; setoid; trans; refl; sym; assoc; ∙-cong; isMagma)
+open import Relation.Binary.Reasoning.Setoid setoid
+open IsMagma isMagma using (∙-congˡ; ∙-congʳ)
```
jamesmckinna, line 19: This, I think, a real mistake.
The `Algebra` hierarchy is designed so that each `Structure`/`Bundle` re-exports all of its substructure(s)/sub-bundles, so explicitly to 'undo' that design by exposing `isMagma` and then exporting from that... is recapitulating a design pattern *which has already been implemented*, so it's
* a DRY failure (and indeed, if the design pattern hasn't been implemented properly, then such a bug would be exposed by trying simply to get everything you need from a single `open Semigroup` declaration, so there's a robustness aspect as well)
* an abstraction failure (we shouldn't need to know how `Semigroup` gives rise to ... whatever it is that supports `∙-congˡ`, `∙-congʳ`)
* an "Occam's Eraser" violation: don't multiply modules (either by creation, or import, or re-export) unnecessarily

Strongly suggest
```suggestion
    using (Carrier; _∙_; _≈_; setoid; trans; refl; sym; assoc; ∙-cong; ∙-congˡ; ∙-congʳ)
open import Relation.Binary.Reasoning.Setoid setoid
```
or even, as I've suggested before, drop the `using` directive entirely.

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,139 @@
...
+
+  extendʳ : u ∙ (v ∙ y) ≈ w ∙ (x ∙ y)
+  extendʳ {y = y} = begin
+    u ∙ (v ∙ y) ≈⟨ pullˡ s ⟩
+    (w ∙ x) ∙ y ≈⟨ assoc w x y ⟩
+    w ∙ (x ∙ y) ∎
+
+  extend² : ∀ y z → (y ∙ u) ∙ (v ∙ z) ≈ (y ∙ w) ∙ (x ∙ z)
+  extend² y z = begin
+    (y ∙ u) ∙ (v ∙ z) ≈⟨ pullʳ (extendʳ {y = z}) ⟩
```
jamesmckinna, line 134: This step is perhaps where the use of *implicit* `y` in `extendʳ` is a mistake: the need to make it explicit is perhaps/probably a code-smell, part of the general problem that in any *concrete* instantiation of `Semigroup`, the typechecker will not (necessarily) be able to infer arguments to any sub-term `y ∙ u`, so the defensive programming style woul dbe to make the argument explicit, and if necessary, use `_` at call-sites. Cf. type of `assoc` itself...

src/Algebra/Reasoning/Semigroup.agda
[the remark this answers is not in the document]
JacquesCarette: While `x∙y≈z⇒[w∙x]∙y≈w∙z` is a logical name, it is also quite long. As these combinators that re-associate are book-keeping things, I'd like to find something shorter / less noisy.

I do hate push/pull. I'm partial to `on-right`.

src/Algebra/Properties/Semigroup/Reasoning.agda
JacquesCarette: strongly agree with the suggestion

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,139 @@
...
+    using (Carrier; _∙_; _≈_; setoid; trans; refl; sym; assoc; ∙-cong; isMagma)
+open import Relation.Binary.Reasoning.Setoid setoid
+open IsMagma isMagma using (∙-congˡ; ∙-congʳ)
+
+private
+ variable
+    u v w x y z : Carrier
+
+module Assoc4  {u v w x : Carrier} where
+  [[u∙v]∙w]∙x≈u∙[[v∙w]∙x] : ((u ∙ v) ∙ w) ∙ x ≈ u ∙ ((v ∙ w) ∙ x)
```
JacquesCarette, line 26: While I don't like the 'explicit' names for push/pull, I'm quite fine with these names for the re-associators that work on 4 things!

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,139 @@
...
+  [[u∙v]∙w]∙x≈u∙[[v∙w]∙x] = trans (∙-congʳ (assoc u v w)) (assoc u (v ∙ w) x)
```
JacquesCarette, line 27: The strong reason to have push/pull defined before these is that they can be used here. You should indeed move those up above `Assoc4` and refactor the proofs here to use push/pull.

src/Algebra/Properties/Semigroup/Reasoning.agda
JacquesCarette: This is probably true; I guess we can get away with it in the categorical case because the types will end up uniquely identify things?

JacquesCarette (changes requested):

src/Algebra/Reasoning/Semigroup.agda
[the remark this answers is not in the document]
jamesmckinna: Well, with my suggested notational optimisations, this would/could become...`xy≈z⇒wx∙y≈wz`, which is about as short as I can make it?!

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,110 @@
...
+
+  u[v∙wx]≈[uv∙w]x : u ∙ (v ∙ (w ∙ x)) ≈ ((u ∙ v) ∙ w) ∙ x
+  u[v∙wx]≈[uv∙w]x = sym (trans (assoc (u ∙ v) w x) (assoc u v (w ∙ x)))
+
+  u[v∙wx]≈[u∙vw]x : u ∙ (v ∙ (w ∙ x)) ≈ (u ∙ (v ∙ w)) ∙ x
+  u[v∙wx]≈[u∙vw]x  = sym (trans (assoc u (v ∙ w) x) (∙-congˡ (assoc v w x)))
+
+open Assoc4 public
+
+module Pulls (uv≡w : u ∙ v ≈ w) where
```
jamesmckinna, line 59: Suggest you don't use `uv≡w` as the name of a proof of `u ∙ v ≈ w`... but use `uv≈w` instead?
(`≡` should only ever be reserved for `PropositionalEquality`)

src/Algebra/Reasoning/Semigroup.agda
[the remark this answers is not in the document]
jamesmckinna: As for `on-right` I'm *really* not sure what that is supposed to signify, given that the lemma is not `cong` like, nor is the action obviously happening on the right? (Never mind our ongoing debates about left/right distinctions giving rise to confusion...)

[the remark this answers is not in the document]
JacquesCarette: `on-right` as in `apply this equality on the right`, after a re-focus. It's very much `cong`-like to me, it just has a re-association done first, to put "the right" in focus.

JacquesCarette:
The 'new' names in `Assoc4` are better. The rest are hideous and make things quite unreadable. It was a nice experiment... which I would call a clear failure. I most definitely prefer push/pull over these.

jamesmckinna:
> The 'new' names in `Assoc4` are better. The rest are hideous and make things quite unreadable. It was a nice experiment... which I would call a clear failure. I most definitely prefer push/pull over these.

Again, agree to disagree. No 'clear' failure on my side, but happy to try to find a good outcome to the naming discussion.

jamesmckinna (changes requested):
Placeholder review: this will have to wait until at least the weekend before I can return to it.
But I think in the meantime there still seem to be spacing issues, the existing question of where these things should live, as well as those regarding whether explicitly-named proofs of 'symmetric' versions are really necessary.

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@ New modules
 
 * `Data.Sign.Show` to show a sign
 
+* `AlgebraPropreties.Semigroup.Reasoning` adding reasoning combinators for semigroups
```
MatthewDaggitt, line 126: Lots of typos in the name?

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- uv≈wuational reasoning for semigroups
```
MatthewDaggitt, line 4: Typo?

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@ New modules
...
+
```
MatthewDaggitt, line 127: I'm also not sure whether we should refer to these as `reasoning combinators`. The `Reasoning` terminology has quite a specific use in the library for syntax that chains together nicely. I'm not sure these lemmas qualify?

MatthewDaggitt (commented):
I agree with a lot of what @jamesmckinna is saying. 

- This doesn't need it's own module, i.e. `Algebra.Properties.Semigroup` is fine.
- I much prefer the current set of names.

src/Algebra/Reasoning/Semigroup.agda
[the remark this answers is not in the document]
jamesmckinna: I'd tried to reply, but my comment got lost: if you're wedded to an ASCII/prefix-Lisp name for these things, then I'd much prefer `refocus-on-right`, or even `refocus-right` (the `-on` isn't then doing much work), or even `refocusʳ`... in the style of other lemmas in the library.

CHANGELOG.md
jamesmckinna: And indeed, if we do as I have suggested several times, and *not* make this a new module, but under `additions to existing modules`, then it should be under `Algebra.Properties.Semigroup`...

CHANGELOG.md
jamesmckinna: We won't need `CHANGELOG` text if it moves to `Additions to...`, but the opening comment block should be rephrased...

src/Algebra/Properties/Semigroup/Reasoning.agda
jamesmckinna: Yeah, looks like an unwitting `global-search-and-replace` snafu.

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+open Semigroup S
+  using (Carrier; _∙_; _≈_; setoid; trans; refl; sym; assoc; ∙-cong; isMagma
+  ; ∙-congˡ; ∙-congʳ)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+private
+ variable
+    u v w x y z : Carrier
+
+module Pulls (uv≈w : u ∙ v ≈ w) where
```
jamesmckinna, line 23: Latest reiteration: why bother to have a named module here? When this moves to `Algebra.Properties.Semigroup`, we could maybe have a comment, but given that you immediately reopen the module publicly, I see no advantage to naming it; any subsequent client import of the top-level module can use `using` to restrict the import interface... adding an additional layer of indirection via `module Pulls` will simply complicate that for... no obvious advantage?

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+    uv≈w⇒xu∙v≈xw : ∀ {x} → (x ∙ u) ∙ v ≈ x ∙ w
+    uv≈w⇒xu∙v≈xw {x = x} =  trans (assoc x u v) (∙-congˡ uv≈w)
```
jamesmckinna, line 25: I think this is going to need `x` to be explicit:
```suggestion
    uv≈w⇒xu∙v≈xw : ∀ x → (x ∙ u) ∙ v ≈ x ∙ w
    uv≈w⇒xu∙v≈xw x =  trans (assoc x u v) (∙-congˡ uv≈w)
```

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+    uv≈w⇒u∙vx≈wx : ∀ {x} → u ∙ (v ∙ x) ≈ w ∙ x
+    uv≈w⇒u∙vx≈wx {x = x} = trans (sym (assoc u v x)) (∙-congʳ uv≈w)
```
jamesmckinna, line 28: Same
```suggestion
    uv≈w⇒u∙vx≈wx : ∀ x → u ∙ (v ∙ x) ≈ w ∙ x
    uv≈w⇒u∙vx≈wx x = trans (sym (assoc u v x)) (∙-congʳ uv≈w)
```

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+    uv≈w⇒u[vx∙y]≈w∙xy : ∀ {x y} → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ (x ∙ y)
+    uv≈w⇒u[vx∙y]≈w∙xy {x = x} {y = y} = trans (∙-congˡ (assoc v x y)) uv≈w⇒u∙vx≈wx
```
jamesmckinna, line 31: Ditto.
```suggestion
    uv≈w⇒u[vx∙y]≈w∙xy : ∀ x y → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ (x ∙ y)
    uv≈w⇒u[vx∙y]≈w∙xy x y = trans (∙-congˡ (assoc v x y)) uv≈w⇒u∙vx≈wx
```

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+    uv≈w⇒x[uv∙y]≈x∙wy : ∀ {x y} → x ∙ (u ∙ (v ∙ y)) ≈ x ∙ (w ∙ y)
```
jamesmckinna, line 33: Ditto

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+    uv≈w⇒x[uv∙y]≈x∙wy = ∙-congˡ uv≈w⇒u∙vx≈wx
+
+    uv≈w⇒[x∙yu]v≈x∙yw : ∀ {x y} → (x ∙ (y ∙ u)) ∙ v ≈ x ∙ (y ∙ w)
```
jamesmckinna, line 36: Ditto.

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+    uv≈w⇒[x∙yu]v≈x∙yw  {x = x} {y = y} = trans (assoc x (y ∙ u) v) (∙-congˡ (uv≈w⇒xu∙v≈xw {x = y}))
+
+    uv≈w⇒[xu∙v]y≈x∙wy : ∀ {x y} → ((x ∙ u) ∙ v) ∙ y ≈ x ∙ (w ∙ y)
```
jamesmckinna, line 39: Ditto.

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+    uv≈w⇒[xu∙v]y≈x∙wy = trans (∙-congʳ uv≈w⇒xu∙v≈xw ) (assoc _ _ _)
+
+    uv≈w⇒[xy∙u]v≈x∙yw : ∀ {x y} → ((x ∙ y) ∙ u) ∙ v ≈ x ∙ (y ∙ w)
```
jamesmckinna, line 42: Ditto.

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+    uv≈w⇒w∙xy≈u[vx∙y] : ∀ {x y} → w ∙ (x ∙ y) ≈ u ∙ ((v ∙ x) ∙ y)
+    uv≈w⇒w∙xy≈u[vx∙y] {x = x} {y = y} = sym (uv≈w⇒u[vx∙y]≈w∙xy uv≈w)
+
+    uv≈w⇒x∙wy≈x[u∙vy] : ∀ {x y} → x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
+    uv≈w⇒x∙wy≈x[u∙vy] {x = x} {y = y} = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w)
+
+    uv≈w⇒x∙yw≈[x∙yu]v : ∀ {x y} → x ∙ (y ∙ w) ≈ (x ∙ (y ∙ u)) ∙ v
+    uv≈w⇒x∙yw≈[x∙yu]v {x = x} {y = y} = sym (uv≈w⇒[x∙yu]v≈x∙yw uv≈w)
+
+open Pushes public
```
jamesmckinna, line 63: I'd be happy to have all these symmetric versions (cf. `x∙yz≈xy∙z`!) provided, once again, `module Pushes` is made anonymous, but I still think it's a margin call whether such duplication is *really* necessary?

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+module Center (uv≈w : u ∙ v ≈ w) where
+  uv≈w⇒xu∙vy≈x∙wy : (x ∙ u) ∙ (v ∙ y) ≈ x ∙ (w ∙ y)
+  uv≈w⇒xu∙vy≈x∙wy = uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx uv≈w)
+
+  uv≈w⇒xy≈z⇒u[vx∙y]≈wz : ∀ z → x ∙ y ≈ z → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ z
+  uv≈w⇒xy≈z⇒u[vx∙y]≈wz z xy≈z = trans (∙-congˡ (uv≈w⇒xu∙v≈xw xy≈z)) (uv≈w⇒u∙vx≈wx uv≈w)
```
jamesmckinna, line 70: Ironically, in *this* instance, because `z` occurs 'free' on the RHS of the equation, it (probbaly) won't need to be quantified, but almost surely once again `x` and `y` *will*:
```suggestion
  uv≈w⇒xy≈z⇒u[vx∙y]≈wz : ∀ x y → x ∙ y ≈ z → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ z
  uv≈w⇒xy≈z⇒u[vx∙y]≈wz x y xy≈z = trans (∙-congˡ (uv≈w⇒xu∙v≈xw xy≈z)) (uv≈w⇒u∙vx≈wx uv≈w)
```

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+  uv≈w⇒x∙wy≈x∙[u∙vy] : x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
```
jamesmckinna, line 72: Make `x` and `y` explicit again?

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+  uv≈w⇒x∙wy≈x∙[u∙vy] = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w)
+
+open Center public
+
+module Assoc4  {u v w x : Carrier} where
```
jamesmckinna, line 77:
```suggestion
module _ where
```
The implicit parameters in the telescope are covered by the `variable` declaration, so aren't needed. Similarly, as throughout, I don't think we need the module to be named?

However, as with the above observations, and the fact that `assoc` itself makes its arguments explicit, even if they are seldom required in practice, probably they should be explicit here:
```suggestion
module _ (u v w x : Carrier) where
```

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+
+  u[v∙wx]≈[uv∙w]x : u ∙ (v ∙ (w ∙ x)) ≈ ((u ∙ v) ∙ w) ∙ x
+  u[v∙wx]≈[uv∙w]x = sym [uv∙w]x≈u[v∙wx]
+
+  u[v∙wx]≈[u∙vw]x : u ∙ (v ∙ (w ∙ x)) ≈ (u ∙ (v ∙ w)) ∙ x
+  u[v∙wx]≈[u∙vw]x  = sym [u∙vw]x≈u[v∙wx]
+
+open Assoc4 public
+
+module Extends {u v w x : Carrier} (s : u ∙ v ≈ w ∙ x) where
```
jamesmckinna, line 110:
```suggestion
module _ (s : u ∙ v ≈ w ∙ x) where
```
Here, the implicits are already covered by the `variable` declaration. 
`s` isn't such a great name for the assumption, so maybe `uv≈wx` instead?

src/Algebra/Properties/Semigroup/Reasoning.agda
```diff
@@ -0,0 +1,121 @@
...
+  uv≈wx⇒yu∙v≈yw∙x : (y ∙ u) ∙ v ≈ (y ∙ w) ∙ x
+  uv≈wx⇒yu∙v≈yw∙x  {y = y} = trans (uv≈w⇒xu∙v≈xw s) (sym (assoc y w x))
```
jamesmckinna, line 112:
```suggestion
  uv≈wx⇒yu∙v≈yw∙x : ∀ y → (y ∙ u) ∙ v ≈ (y ∙ w) ∙ x
  uv≈wx⇒yu∙v≈yw∙x y = trans (uv≈w⇒xu∙v≈xw s) (sym (assoc y w x))
```

jamesmckinna (changes requested):
Lots of comments and suggestion, but mostly a do-over of things I have already commented on before, and were not (yet) changed. TL;DR:
* move all the lemmas to `Algebra.Properties.Semigroup`
* anonymous modules
* fix implicit/explicit quantification

jamesmckinna:
Following @MatthewDaggitt 's comment about 'reasoning combinators', I've taken the liberty of renaming the PR to hopefully better reflect the intention of the contribution(s) here.

jmougeot:
I moved everything into Semigroups and made the variables explicit. I tried to address all the comments, but I probably missed some. Sorry for not taking these changes into account earlier

CHANGELOG.md
```diff
@@ -123,7 +123,7 @@ New modules
 
 * `Data.Sign.Show` to show a sign
 
-Additions to existing modules
+* `Algebra.Properties.Semigroup` adding consequences for associtvity for semigroups
```
jamesmckinna, line 126: Typo
```suggestion
* `Algebra.Properties.Semigroup` adding consequences for associativity for semigroups
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+    uv≈w⇒x[uv∙y]≈x∙wy x y = ∙-congˡ (uv≈w⇒u∙vx≈wx y)
+
+    uv≈w⇒[x∙yu]v≈x∙yw : ∀ x y → (x ∙ (y ∙ u)) ∙ v ≈ x ∙ (y ∙ w)
+    uv≈w⇒[x∙yu]v≈x∙yw  x y = trans (assoc x (y ∙ u) v) (∙-congˡ (uv≈w⇒xu∙v≈xw y))
+
+    uv≈w⇒[xu∙v]y≈x∙wy : ∀ x y → ((x ∙ u) ∙ v) ∙ y ≈ x ∙ (w ∙ y)
+    uv≈w⇒[xu∙v]y≈x∙wy x y = trans (∙-congʳ (uv≈w⇒xu∙v≈xw x)) (assoc _ _ _)
+
+    uv≈w⇒[xy∙u]v≈x∙yw : ∀ x y → ((x ∙ y) ∙ u) ∙ v ≈ x ∙ (y ∙ w)
+    uv≈w⇒[xy∙u]v≈x∙yw x y = trans (∙-congʳ (assoc x y u)) (uv≈w⇒[x∙yu]v≈x∙yw x  y )
```
jamesmckinna, line 58: whitespace
```suggestion
    uv≈w⇒[xy∙u]v≈x∙yw x y = trans (∙-congʳ (assoc x y u)) (uv≈w⇒[x∙yu]v≈x∙yw x y)
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+    uv≈w⇒w∙xy≈u[vx∙y] x y = sym (uv≈w⇒u[vx∙y]≈w∙xy uv≈w x y)
+
+    uv≈w⇒x∙wy≈x[u∙vy] : ∀ x y → x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
+    uv≈w⇒x∙wy≈x[u∙vy] x y = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w x y)
+
+    uv≈w⇒x∙yw≈[x∙yu]v : ∀ x y  → x ∙ (y ∙ w) ≈ (x ∙ (y ∙ u)) ∙ v
+    uv≈w⇒x∙yw≈[x∙yu]v x y  = sym (uv≈w⇒[x∙yu]v≈x∙yw uv≈w x y)
+
+    uv≈w⇒xu∙vy≈x∙wy : (x ∙ u) ∙ (v ∙ y) ≈ x ∙ (w ∙ y)
+    uv≈w⇒xu∙vy≈x∙wy = uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx uv≈w _) _
```
jamesmckinna, line 77: I think that this needs, for the usual reasons about the unifier/typechecker
```suggestion
    uv≈w⇒xu∙vy≈x∙wy : ∀ x y → (x ∙ u) ∙ (v ∙ y) ≈ x ∙ (w ∙ y)
    uv≈w⇒xu∙vy≈x∙wy x y = uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx uv≈w x) y
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+
+    uv≈w⇒xy≈z⇒u[vx∙y]≈wz : ∀ z → x ∙ y ≈ z → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ z
+    uv≈w⇒xy≈z⇒u[vx∙y]≈wz z xy≈z = trans (∙-congˡ (uv≈w⇒xu∙v≈xw xy≈z v)) (uv≈w⇒u∙vx≈wx uv≈w z)
+
+    uv≈w⇒x∙wy≈x∙[u∙vy] : x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
+    uv≈w⇒x∙wy≈x∙[u∙vy] = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w _ _)
```
jamesmckinna, line 83: And here
```suggestion
    uv≈w⇒x∙wy≈x∙[u∙vy] : ∀ x y → x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
    uv≈w⇒x∙wy≈x∙[u∙vy] x y = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w x y)
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+    uv≈w⇒x∙wy≈x[u∙vy] x y = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w x y)
+
+    uv≈w⇒x∙yw≈[x∙yu]v : ∀ x y  → x ∙ (y ∙ w) ≈ (x ∙ (y ∙ u)) ∙ v
+    uv≈w⇒x∙yw≈[x∙yu]v x y  = sym (uv≈w⇒[x∙yu]v≈x∙yw uv≈w x y)
+
+    uv≈w⇒xu∙vy≈x∙wy : (x ∙ u) ∙ (v ∙ y) ≈ x ∙ (w ∙ y)
+    uv≈w⇒xu∙vy≈x∙wy = uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx uv≈w _) _
+
+    uv≈w⇒xy≈z⇒u[vx∙y]≈wz : ∀ z → x ∙ y ≈ z → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ z
+    uv≈w⇒xy≈z⇒u[vx∙y]≈wz z xy≈z = trans (∙-congˡ (uv≈w⇒xu∙v≈xw xy≈z v)) (uv≈w⇒u∙vx≈wx uv≈w z)
```
jamesmckinna, line 80: And here:
```suggestion
    uv≈w⇒xy≈z⇒u[vx∙y]≈wz : ∀ x y → x ∙ y ≈ z → u ∙ ((v ∙ x) ∙ y) ≈ w ∙ z
    uv≈w⇒xy≈z⇒u[vx∙y]≈wz x y xy≈z = trans (∙-congˡ (uv≈w⇒xu∙v≈xw xy≈z v)) (uv≈w⇒u∙vx≈wx uv≈w _)
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+
+    uv≈w⇒x∙wy≈x∙[u∙vy] : x ∙ (w ∙ y) ≈ x ∙ (u ∙ (v ∙ y))
+    uv≈w⇒x∙wy≈x∙[u∙vy] = sym (uv≈w⇒x[uv∙y]≈x∙wy uv≈w _ _)
+
+module _ {u v w x : Carrier} where
```
jamesmckinna, line 85: These will need to be *explicit*... for the usual reasons, but there's no need to make `Carrier` explicit:
```suggestion
module _ u v w x where
```

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+    u[vw∙x]≈uv∙wx : u ∙ ((v ∙ w) ∙ x) ≈ (u ∙ v) ∙ (w ∙ x)
+    u[vw∙x]≈uv∙wx = sym uv∙wx≈u[vw∙x]
+
+    u[v∙wx]≈[uv∙w]x : u ∙ (v ∙ (w ∙ x)) ≈ ((u ∙ v) ∙ w) ∙ x
+    u[v∙wx]≈[uv∙w]x = sym [uv∙w]x≈u[v∙wx]
+
+    u[v∙wx]≈[u∙vw]x : u ∙ (v ∙ (w ∙ x)) ≈ (u ∙ (v ∙ w)) ∙ x
+    u[v∙wx]≈[u∙vw]x = sym [u∙vw]x≈u[v∙wx]
+
+module _ {u v w x : Carrier} (uv≈wx : u ∙ v ≈ w ∙ x) where
```
jamesmckinna, line 116: Finally, here:
```suggestion
module _ (uv≈wx : u ∙ v ≈ w ∙ x) where
```
because the `variable` declarations take care of the initial implicit prefix...

jamesmckinna (changes requested):
Last round of changes, I hope, because I've looked at this to many times now... law of diminishing returns...

NB if you adopt my suggestions, there will no doubt be knock-on consequences (wrt parametrisation) which you should fix before committing... for the sake of any other reviewers!

CHANGELOG.md
jamesmckinna: And, in case it wasn't clear before: you *do* need to list, explicitly, every single lemma as an addition to that module (follow the model of the entries below), so this line should also move downwards to a separate block between `Algebra.Properties.Monoid.Divisibility` and `Algebra.Properties.Semigroup.Divisibility`...
... AND you've somehow deleted the line
`Additions to existing modules` at L126... which you need to restore, sigh.

CHANGELOG.md
```diff
@@ -123,7 +123,7 @@ New modules
 
 * `Data.Sign.Show` to show a sign
 
-Additions to existing modules
+* `Algebra.Properties.Semigroup` adding consequences for associativity for semigroups
```
jamesmckinna, line 126: (I made this comment on #2692 but it belongs here):

No, you need to revert this change.
The lemmas added to `Algebra.Properties.Semigroup` should each be added in an Agda code block under this line moved to be a sub-heading underneath the original section heading...

src/Algebra/Properties/Semigroup.agda
jamesmckinna: I'm not sure why this was marked as `resolved`?

src/Algebra/Properties/Semigroup.agda
jamesmckinna: Nor this, because it hasn't been fixed yet.

CHANGELOG.md
```diff
@@ -126,6 +126,8 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Properties.Semigroup` adding consequences for associativity for semigroups
+
```
jamesmckinna, line 130: Need to insert here all the new names, once we have (eventually) agree them...
```suggestion
```agda
  uv≈w⇒xu∙v≈xw : ∀ x → (x ∙ u) ∙ v ≈ x ∙ w
```

etc.

src/Algebra/Reasoning/Semigroup.agda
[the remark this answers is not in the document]
JacquesCarette: I'm going to make some proposal on naming as a full comments instead of inline.

JacquesCarette:
As I've indicated, I'm not a big fan of push/pull because they are not very mnemomic; the superscripts don't help either. Indeed, if one looks at
```agda
module Pulls (x∙y≈z : x ∙ y ≈ z) where
  x∙y≈z⇒[w∙x]∙y≈w∙z : (w ∙ x) ∙ y ≈ w ∙ z
```
what it really does is
1. refocus on the right
2. apply an equality there.

So that's what I'd like the naming to reflect. Now, I quite like the 'synonym' `_⟩∙⟨_` for `∙-cong` (hoping to introduce that synonym too.) So then if we take `⭆` to be refocus (on the right), then the above could be `⭆∙⟨_`. We would similarly have a refocus on the left then act on that side as `⭅⟨_∙`.  I would be fine with putting in a marker for where "nothing happens". So we could have:
- `⭆-I∙⟨_` : refocus right then apply equation there
- `⭅-_⟩∙I` : refocus left then apply equation there
- `I∙_⟩-⭅` : apply equation on right then refocus left
- `_⟩∙I-⭆` : apply equation on left then refocus right

Note that these are meant to have the 3 variables be *implicit*, on purpose: if they cannot be figured out by Agda, then it's probably better to use the 'long form' of the reasoning (i.e. explicit calls to `assoc`).

jamesmckinna:
> As I've indicated, I'm not a big fan of push/pull because they are not very mnemomic; the superscripts don't help either. Indeed, if one looks at
> 
> ```agda
> module Pulls (x∙y≈z : x ∙ y ≈ z) where
>   x∙y≈z⇒[w∙x]∙y≈w∙z : (w ∙ x) ∙ y ≈ w ∙ z
> ```
> 
> what it really does is
> 
>     1. refocus on the right
> 
>     2. apply an equality there.
> 
> 
> So that's what I'd like the naming to reflect. Now, I quite like the 'synonym' `_⟩∙⟨_` for `∙-cong` (hoping to introduce that synonym too.) So then if we take `⭆` to be refocus (on the right), then the above could be `⭆∙⟨_`. We would similarly have a refocus on the left then act on that side as `⭅⟨_∙`. I would be fine with putting in a marker for where "nothing happens". So we could have:
> 
>     * `⭆-I∙⟨_` : refocus right then apply equation there
> 
>     * `⭅-_⟩∙I` : refocus left then apply equation there
> 
>     * `I∙_⟩-⭅` : apply equation on right then refocus left
> 
>     * `_⟩∙I-⭆` : apply equation on left then refocus right
> 
> 
> Note that these are meant to have the 3 variables be _implicit_, on purpose: if they cannot be figured out by Agda, then it's probably better to use the 'long form' of the reasoning (i.e. explicit calls to `assoc`).

OK, I guess it is/was good to see how far apart the various contributors are on how to think about these naming issues, and their intended semantic underpinnings. But I have to say that I find the gap between the symbolic names you propose ((NB in my Firefox font for GitHub, your symbol `⭅` doesn't even display!) and those intended meanings so large that I simply cannot imagine ever using these principles except via uncomprehending cut-and-paste from the module, and with the above explanation by my side as user manual.

In the face of this, I am going to suggest
* either that we roll back to a text-based name, with your clear desire to inject 'refocus' as an idiom as a basis for the name (and I even suggested names based around that above)...
* or that we redraw the 'implicational' forms I suggested to make the `x∙y≈z⇒`prefix implicit as an infix argument position, with eg.
  - `x∙y≈z⇒[w∙x]∙y≈w∙z : (w ∙ x) ∙ y ≈ w ∙ z` becoming `[wx]y≈[_]wz` or `[wx]y≈⟨_⟩wz`, but again, we have up until now avoided making operators for equational proofs into infixes/mixfixes... except in the 'equation over equation' presentation of equality-modulo-`cast` in `Data.Vec.Relation.Binary.Equality.Cast`
* or simply that I withdraw from even expressing a view about this (much less merge the PR), other than to say that adopting such a change as you propose would mark such a meta-level breaking change from existing practice in `stdlib` that... I am completely lost for words! 

Barn-burning.

MatthewDaggitt:
Conclusion: accept but note that there should be better solution

src/Algebra/Properties/Semigroup.agda
```diff
@@ -28,3 +34,91 @@ alternative = alternativeˡ , alternativeʳ
...
+  uv≈w⇒[x∙yu]v≈x∙yw x y = trans (assoc x (y ∙ u) v) (∙-congˡ (uv≈w⇒xu∙v≈xw y))
+
+  uv≈w⇒[xu∙v]y≈x∙wy : ∀ x y → ((x ∙ u) ∙ v) ∙ y ≈ x ∙ (w ∙ y)
+  uv≈w⇒[xu∙v]y≈x∙wy x y = trans (∙-congʳ (uv≈w⇒xu∙v≈xw x)) (assoc x w y)
+
+  uv≈w⇒[xy∙u]v≈x∙yw : ∀ x y → ((x ∙ y) ∙ u) ∙ v ≈ x ∙ (y ∙ w)
+  uv≈w⇒[xy∙u]v≈x∙yw x y = trans (∙-congʳ (assoc x y u)) (uv≈w⇒[x∙yu]v≈x∙yw x  y )
+
+module _ (uv≈w : u ∙ v ≈ w) where
+  uv≈w⇒xw≈xu∙v : ∀ x → x ∙ w ≈ (x ∙ u) ∙ v
```
JacquesCarette: It was suggested that all the ones which start with `sym` in their proof should be removed, since that's easier to obtain via the reasoning combinators themselves.

MatthewDaggitt (approved):

MatthewDaggitt:
@JacquesCarette ?

JacquesCarette (approved):

jamesmckinna:
Sorry for not having rescinded the old review! Thanks for merging!

Issue #2654, opened 2025-03-12
Left/Right inconsistency


JacquesCarette:
Issue #2532 from @jamesmckinna draws our attention to some of the left/right inconsistencies. There are... more. I attach two lists (courtesy of @jmougeot ), one which details definitions that use subscript r and l (rather than the majority superscript), and also other items that use either `Left` or `Right` in their names, or 'r'/'l' superscripts (or even both!!).  And these are *just* definition sites, not all lines. There are quite a lot of these.  So I am rather at a loss on how to check 'consistency' when the scale is this large. But at least now we have the raw data in hand.

- Subscripts: [lower.txt](https://github.com/user-attachments/files/19200974/lower.txt)
- Left/Right: [matches.txt](https://github.com/user-attachments/files/19200973/matches.txt)

jamesmckinna:
Heroic work, but indeed: words fail me as to how we might reconcile all this!

PR #2692, opened 2025-04-10
[Add] Consequences of identity for `monoids`

jmougeot:
Introduce reasoning combinators for monoids. 

This PR is intended to focus on the `Monoid.Reasoning` filr as the discussion regarding the `Semigroup.reasonig` file is in the PR : #2688

I went with the names from agda-categories since they seemed to make sense.


jamesmckinna:
Probably the 'best' thing to do is to mark this as blocked (on #2688 ), otherwise you'll end up with merge conflicts when that PR gets merge, and indeed to start afresh with this PR being on a branch based on *that* branch. That keeps both the locality of each PR smaller, but also the (sequential) dependency between them.

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+-- (Utilities for identity and cancellation reasoning, extending semigroup reasoning)
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (Monoid)
+
+module Algebra.Properties.Monoid.Reasoning {o ℓ} (M : Monoid o ℓ) where
+
+open import Algebra.Structures using (IsMagma)
```
jamesmckinna, line 14: As with #2688 there's no need to make this explicit, it's even an abstraction failure/violation to do so

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (Monoid)
+
+module Algebra.Properties.Monoid.Reasoning {o ℓ} (M : Monoid o ℓ) where
+
+open Monoid M
+    using (Carrier; _∙_; _≈_; setoid; isMagma; semigroup; ε; sym; identityˡ
```
jamesmckinna, line 15: Default indentation, here, as throughout the module is *2* spaces, not 4 as you have here.
Given that this is the `(n+1)`th time of making this observation, it suggests that you need to fix the defaults in your editor perhaps?

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+    ; identityʳ ; ∙-cong; refl; assoc; ∙-congˡ; ∙-congʳ; trans)
+open import Relation.Binary.Reasoning.Setoid setoid
+open import Algebra.Properties.Semigroup.Reasoning semigroup public
+
```
jamesmckinna, line 19: Suggest declaring a `private` block of `variable`s here, again as usual, following #2688 ...

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+module Identity {a : Carrier } where
```
jamesmckinna, line 20: ... So that there is no need here for either the implicit module scope, the module name, or even the declaration of amodule at all!

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+    id-unique : (∀ b → b ∙ a ≈ b) → a ≈ ε
+    id-unique b∙a≈b = trans (sym (identityˡ a)) (b∙a≈b ε)
+
+    id-comm : a ∙ ε ≈ ε ∙ a
+    id-comm = trans (identityʳ a) (sym (identityˡ a))
+
+    id-comm-sym : ε ∙ a ≈ a ∙ ε
+    id-comm-sym = sym id-comm
```
jamesmckinna, line 28: I now think that these explicit symmetric forms are excessive/superfluous.

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+    ; identityʳ ; ∙-cong; refl; assoc; ∙-congˡ; ∙-congʳ; trans)
+open import Relation.Binary.Reasoning.Setoid setoid
+open import Algebra.Properties.Semigroup.Reasoning semigroup public
+
+module Identity {a : Carrier } where
+    id-unique : (∀ b → b ∙ a ≈ b) → a ≈ ε
+    id-unique b∙a≈b = trans (sym (identityˡ a)) (b∙a≈b ε)
+
+    id-comm : a ∙ ε ≈ ε ∙ a
+    id-comm = trans (identityʳ a) (sym (identityˡ a))
```
jamesmckinna, line 25: Against the above comment, this *may* need `a` to be explicit... try some tests to see in eg. `Data.Nat.Properties`?

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+
+    id-comm-sym : ε ∙ a ≈ a ∙ ε
+    id-comm-sym = sym id-comm
+
+open Identity public
+
+module IntroElim {a b : Carrier} (a≈ε : a ≈ ε) where
```
jamesmckinna, line 32: My instincts tell me:
* `a` can be implicit, and covered by a `variable` declaration
* `b` should be explicit

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+    elimʳ = trans (∙-congˡ a≈ε) (identityʳ b)
+
+    elimˡ : a ∙ b ≈ b
+    elimˡ = trans (∙-congʳ a≈ε) (identityˡ b)
+
+    introʳ : a ≈ ε → b ≈ b ∙ a
+    introʳ a≈ε = sym elimʳ
+
+    introˡ : a ≈ ε → b ≈ a ∙ b
+    introˡ a≈ε = sym elimˡ
```
jamesmckinna, line 43: Argument `a≈ε : a ≈ ε` is already covered by the `module` telescope. Delete here.
```suggestion
    introʳ : b ≈ b ∙ a
    introʳ = sym elimʳ

    introˡ : b ≈ a ∙ b
    introˡ = sym elimˡ
```

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+
+    introcenter : ∀ c → b ∙ c ≈ b ∙ (a ∙ c)
+    introcenter c = trans (∙-congˡ (sym (identityˡ c))) (∙-congˡ (∙-congʳ (sym a≈ε)))
+
+open IntroElim public
+
+module Cancellers {a b c : Carrier} (inv : a ∙ c ≈ ε) where
```
jamesmckinna, line 50: Implicits `a`, `c` via `variable`s again... but `b` should be explicit, I think.

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,68 @@
...
+  insertˡ = sym cancelˡ
+
+  insertʳ : b ≈ (b ∙ a) ∙ c
+  insertʳ = sym cancelʳ
+
+  cancelInner : ∀ {g} → (b ∙ a) ∙ (c ∙ g) ≈ b ∙ g
+  cancelInner {g = g} = trans (sym (assoc (b ∙ a) c g)) (∙-congʳ cancelʳ)
+
+  insertInner : ∀ {g} → b ∙ g ≈ (b ∙ a) ∙ (c ∙ g)
+  insertInner = sym cancelInner
```
jamesmckinna, line 68: Two questions:
* Why `g`, and not `d`?
* Why implicit, and not explicit, quantification?

src/Algebra/Definitions.agda
```diff
@@ -26,6 +25,7 @@ module Algebra.Definitions
 open import Algebra.Core using (Op₁; Op₂)
 open import Data.Product.Base using (_×_; ∃-syntax)
 open import Data.Sum.Base using (_⊎_)
+open import Relation.Nullary.Negation.Core using (¬_)
```
jamesmckinna, line 28: Nice, although I thought this might have been done in one of your earlier rounds of `import` cleaning up?

src/Algebra/Properties/Monoid/Reasoning.agda
jamesmckinna: Secondly, for the `*Inner` lemmas, can they not be re-expressed in terms of the refocusing lemmas for `Semigroup`? It seems as though these are `insert`/`cancel` 'under a `b`', using them?

src/Algebra/Properties/Monoid/Reasoning.agda
JacquesCarette: `id-comm-sym` (and `id-comm`) are used surprisingly often in `agda-categories`. But I think that might be an artifact of how classical category theory is "formalized" in textbooks that when *actually* formalized adds a ridiculous number of `id` that need to be shuffled around.

It also happens when doing naive `foldr` or naive partial evaluation: you get spurious `id`s, and sometimes you need to shuffle these around. So while I'm not claiming `id-comm-sym` would pay its weight here, there's a good reason why it is being proposed.

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+  ; identityʳ ; ∙-cong; refl; assoc; ∙-congˡ; ∙-congʳ; trans)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+open import Algebra.Properties.Semigroup semigroup public
+
+private
+    variable
+        a b c d : Carrier
+
+module _ where
```
JacquesCarette, line 25: the `module _ where` is now not needed.

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+    id-unique : ∀ a → (∀ b → b ∙ a ≈ b) → a ≈ ε
+    id-unique a b∙a≈b = trans (sym (identityˡ a)) (b∙a≈b ε)
+
+    id-comm : ∀ a → a ∙ ε ≈ ε ∙ a
+    id-comm a = trans (identityʳ a) (sym (identityˡ a))
+
+    id-comm-sym : ∀ a → ε ∙ a ≈ a ∙ ε
+    id-comm-sym a = sym (id-comm a)
+
+module _ {a b : Carrier} (a≈ε : a ≈ ε) where
```
JacquesCarette, line 35: the `variable` block above means you don't need `{a b : Carrier}` anymore

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+    elimʳ : ∀ b → b ∙ a ≈ b
+    elimʳ b = trans (∙-congˡ a≈ε) (identityʳ b)
+
+    elimˡ : ∀ b → a ∙ b ≈ b
+    elimˡ b = trans (∙-congʳ a≈ε) (identityˡ b)
+
+    introʳ : ∀ b → b ≈ b ∙ a
+    introʳ b = sym (elimʳ b)
```
JacquesCarette, line 43: eta-reduce, here and next

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+
+    introˡ : ∀ b → b ≈ a ∙ b
+    introˡ b = sym (elimˡ b)
+
+    introcenter : ∀ c → b ∙ c ≈ b ∙ (a ∙ c)
+    introcenter c = trans (∙-congˡ (sym (identityˡ c))) (∙-congˡ (∙-congʳ (sym a≈ε)))
```
JacquesCarette, line 49: I wonder if we should use a superscript `m` (for 'middle') or `c` (for, indeed, 'center') instead of spelling it out?

JacquesCarette (changes requested):

src/Algebra/Properties/Monoid/Reasoning.agda
jamesmckinna: Interesting, thanks for the insight. All of this seems only further evidence towards developing `Solver`-style/normal form equational reasoning as in #2629 ...

src/Algebra/Properties/Monoid/Reasoning.agda
jamesmckinna: Meta: a good general point that, besides 'left' and 'right' (modulo collective mutual misunderstanding/inconsistency about what such conventions might even mean #2654 ), we don't have good heuristics about super-/sub-scripts and their intended semantics. I *think* I'd be happy with `ᶜ` or `ᵐ`... but nervous about establishing a precedent without further thought/discussion?

CHANGELOG.md
```diff
@@ -123,7 +123,7 @@ New modules
 
 * `Data.Sign.Show` to show a sign
 
-Additions to existing modules
+* `Algebra.Properties.Semigroup` adding consequences for associatvity for semigroups
```
jamesmckinna, line 126: No, you need to revert this change.
The lemmas added to `Algebra.Properties.Semigroup` should each be added in an Agda code block under this line moved to be a sub-heading *underneath* this section heading...

src/Algebra/Properties/Monoid/Reasoning.agda
```diff
@@ -0,0 +1,70 @@
...
+--
+-- Equational reasoning for monoids
+-- (Utilities for identity and cancellation reasoning, extending semigroup reasoning)
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (Monoid)
+
+module Algebra.Properties.Monoid.Reasoning {o ℓ} (M : Monoid o ℓ) where
```
jamesmckinna, line 12: As usual: these lemmas belong in `Algebra.Properties.Monoid`, not in a separate module as you have it.

By all means do initial development in that style before raising a PR, or else mark it as `DRAFT` (and hence not ready-for-review) but once you do raise the PR/ask for review, we do need it to be compatible both with the existing `stdlib` hierarchy/structure, but also with the `style-guide`, or else we will expend a lot of extra reviewing energy on things which should have been fixed ahead of time.

CHANGELOG.md
```diff
@@ -113,6 +113,8 @@ Deprecated names
 New modules
 -----------
 
+* `Algebra.Properties.Monoid` adding consequences for identity for monoids
+
```
jamesmckinna, line 117: Not here
```suggestion
```

CHANGELOG.md
```diff
@@ -126,6 +128,8 @@ New modules
 Additions to existing modules
 -----------------------------
 
+* In `Algebra.Properties.Semigroup` adding consequences for associativity for semigroups
+
```
jamesmckinna, line 132: But here instead:
```suggestion
* `Algebra.Properties.Monoid` adding consequences for identity for monoids

* In `Algebra.Properties.Semigroup` adding consequences for associativity for semigroups

```
plus *all* the actual named lemmas listed.

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,70 @@
...
+    introˡ : ∀ b → b ≈ a ∙ b
+    introˡ = sym ∘ elimˡ
+
+    introcenter : ∀ c → b ∙ c ≈ b ∙ (a ∙ c)
+    introcenter c = trans (∙-congˡ (sym (identityˡ c))) (∙-congˡ (∙-congʳ (sym a≈ε)))
+
+module _ (inv : a ∙ c ≈ ε) where
+
+  cancelʳ : ∀ b → (b ∙ a) ∙ c ≈ b
+  cancelʳ b = trans (assoc b a c) (trans (∙-congˡ inv) (identityʳ b))
```
JacquesCarette, line 54: I'm quite sure this can be done in 2 steps via one of the Semigroup combinators. Same for on the left.

JacquesCarette (changes requested):

MatthewDaggitt:
Given that this is not properly rebased, and the naming scheme is different from the semigroup PR and that I need to make a release candidate tomorrow, I'm going to remove this from the v2.3 milestone tomorrow @JacquesCarette. If you would like this to urgently get in then please bring it up to date.

JacquesCarette:
I will try.

JacquesCarette (approved):
I think this is in excellent shape now.

JacquesCarette:
@jmougeot pushed through the work (thanks!)

Yes, the naming scheme here is quite different. However, I claim that the names in this PR *make sense*. The ones for original semigroup PR didn't (i.e. the ones based on push and pull which are quite arbitrary). When you read a proof that uses the names here, you get a 'semantic sense' of what the proof does. They've also been used a lot in `agda-categories` and adopted in other libraries as well. And we've agreed that the current names in semigroup are an unsatisfying compromise, i..e the names are at best 'not wrong'.

I'm fine if this doesn't make it in to v2.3 -- but it shouldn't be because of the names!

jamesmckinna:
I'm (qualified) happy with this, but would like us to revisit the names before v3.0, when we can break/deprecate to improve matters. My concern would be that we choose bad names, use them, extensively even, and then have more knock-on viscosity once we find the 'right' ones... but that is, here, I think outweighed by having them available for experiment, so their utility (and memorability!) can be experimented with... 
... but as I'm on cell signal, I'll defer the heavy lifting and final approval to the release manager... ;-)

src/Algebra/Properties/Monoid/Reasoning.agda
jamesmckinna: On balance, and with `cancelInner` below, I now think I'd prefer the superscript version, but don't mind which, maybe a slight preference for 'c' (possible Anglophone/Francophone bias?)

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,70 @@
...
+  elimˡ = trans (∙-congʳ a≈ε) ∘ identityˡ
+
+  introʳ : ∀ b → b ≈ b ∙ a
+  introʳ = sym ∘ elimʳ
+
+  introˡ : ∀ b → b ≈ a ∙ b
+  introˡ = sym ∘ elimˡ
+
+  introcenter : ∀ c → b ∙ c ≈ b ∙ (a ∙ c)
+  introcenter c = trans (∙-congˡ (sym (identityˡ c))) (∙-congˡ (∙-congʳ (sym a≈ε)))
```
jamesmckinna: On reflection, I think i do prefer:
```suggestion
  introᶜ : ∀ c → b ∙ c ≈ b ∙ (a ∙ c)
  introᶜ c = trans (∙-congˡ (sym (identityˡ c))) (∙-congˡ (∙-congʳ (sym a≈ε)))
```

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,70 @@
...
+  insertˡ = sym ∘ cancelˡ
+
+  insertʳ : ∀ b → b ≈ (b ∙ a) ∙ c
+  insertʳ = sym ∘ cancelʳ
+
+  cancelInner : ∀ b d → (b ∙ a) ∙ (c ∙ d) ≈ b ∙ d
+  cancelInner b d = trans (uv≈w⇒xu∙vy≈x∙wy inv b d) (∙-congˡ (identityˡ d))
+
+  insertInner : ∀ b d → b ∙ d ≈ (b ∙ a) ∙ (c ∙ d)
+  insertInner = λ b d → sym (cancelInner b d)
```
jamesmckinna: Likewise:
```suggestion
  cancelᶜ : ∀ b d → (b ∙ a) ∙ (c ∙ d) ≈ b ∙ d
  cancelᶜ b d = trans (uv≈w⇒xu∙vy≈x∙wy inv b d) (∙-congˡ (identityˡ d))

  insertᶜ : ∀ b d → b ∙ d ≈ (b ∙ a) ∙ (c ∙ d)
  insertᶜ = λ b d → sym (cancelInner b d)
```

src/Algebra/Properties/Monoid/Reasoning.agda
JacquesCarette: I'm fine with that renaming.

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,70 @@
...
+  insertˡ = sym ∘ cancelˡ
+
+  insertʳ : ∀ b → b ≈ (b ∙ a) ∙ c
+  insertʳ = sym ∘ cancelʳ
+
+  cancelᶜ : ∀ b d → (b ∙ a) ∙ (c ∙ d) ≈ b ∙ d
+  cancelᶜ b d = trans (uv≈w⇒xu∙vy≈x∙wy inv b d) (∙-congˡ (identityˡ d))
+
+  insertᶜ : ∀ b d → b ∙ d ≈ (b ∙ a) ∙ (c ∙ d)
+  insertᶜ = λ b d → sym (cancelInner b d)
```
JacquesCarette:
```suggestion
  insertᶜ = λ b d → sym (cancelᶜ b d)
```

jamesmckinna:
Although I've previously approved these, I'd like us to reconsider the Fairbairn penalty for having left/right `intro`/`elim` pairs with the one being the `sym` of the other , and likewise with `insert`/`cancel`...
... just as with #2688 not all uses of `sym` were eliminated (and another one, where the `sym` is disguised by flipping the arguments)

Suggestion: a follow-up PR to prune out the redundancies? (Ideally: before the v2.3 `CHANGELOG` is set in stone...)
But here: we need to commit to which we prefer, `intro` over `elim` (or vice versa), `insert` over `cancel`?
@JacquesCarette or would you argue the Fairbairn cost is worth paying?

JacquesCarette:
In this particular case, I will definitely argue that that Fairbairn cost is worth paying, again largely because of readability. When you seem something being eliminated, it's nice to see the proof say that. When the proof says 'intro' but it reads as the opposite, it takes a while to see that for this one line, the reasoning is backwards. 

I think that `intro` / `elim` and `insert` / `cancel` are the only times when I feel that way! So I'm not about to advocate for the multiplication of named symmetric combinators.

src/Algebra/Properties/Monoid.agda
```diff
@@ -0,0 +1,70 @@
...
+  ; identityʳ ; ∙-cong; refl; assoc; ∙-congˡ; ∙-congʳ; trans)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+open import Algebra.Properties.Semigroup semigroup public
+
+private
+    variable
+        a b c d : Carrier
+
+id-unique : ∀ a → (∀ b → b ∙ a ≈ b) → a ≈ ε
```
Taneb: Can we name these with `ε` rather than `id`?

Taneb:
Do we want to (can we please) reexport these for rings with suitable renamings?

src/Algebra/Properties/Monoid.agda
JacquesCarette: Makes sense.

JacquesCarette:
> Do we want to (can we please) reexport these for rings with suitable renamings?

I'd be happy to!

JacquesCarette:
@Taneb I've done the renaming. I'll do the `Ring` part in a different PR, so as to not block this one.

jamesmckinna:
> I'm (qualified) happy with this, but would like us to revisit the names before v3.0, when we can break/deprecate to improve matters. My concern would be that we choose bad names, use them, extensively even, and then have more knock-on viscosity once we find the 'right' ones... but that is, here, I think outweighed by having them available for experiment, so their utility (and memorability!) can be experimented with... ... but as I'm on cell signal, I'll defer the heavy lifting and final approval to the release manager... ;-)

@Taneb 's problems on #2851 suggests that we may indeed have been too hasty in accepting the `cancel` names... sigh.

PR #2693, opened 2025-04-10
[ add ] obvious lemma about self-contradiction to `Relation.Nullary.Negation.Core`

jamesmckinna:
Self-explanatory, but I'm open to:
* alternative suggestions for the name of this lemma
* suggestions that we don't even need this ;-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,3 +244,8 @@ Additions to existing modules
   ⊤-dec : Dec {a} ⊤
   ⊥-dec : Dec {a} ⊥
   ```
+
+* In `Relation.Nullary.Negation.Core`:
+  ```agda
+  self-contradiction : (A → ¬ A) → ¬ A
+  ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -247,5 +247,5 @@ Additions to existing modules
 
 * In `Relation.Nullary.Negation.Core`:
   ```agda
-  self-contradiction : (A → ¬ A) → ¬ A
+  contra-diagonal : (A → ¬ A) → ¬ A
   ```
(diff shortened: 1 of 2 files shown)


Taneb:
I'd be happier with this if this had a shorter (but equally clear) name...

jamesmckinna:
> I'd be happier with this if this had a shorter (but equally clear) name...

An alternative, not necessarily better, in the spirit of #2688 might be 
```agda
```

JacquesCarette:
`¬fix¬` ? `contra-diag` ?

jamesmckinna:
Thanks @JacquesCarette ... not sure about either of the suggestions, though I have a slight preference for `contra-diag`, but honestly, I think the `contra` is superfluous (though semantically meaningful, both the type and the fact that this is in the `Negation` module...) and `diag` feels to me as though it leans too much towards short-for-the-sake-of-it... given that this module contains many 'long-form' names, seemingly without problems... which is why I'd suggested `diagonalisation`, with `diagonal` as *my* concession towards abbreviation... ;-)

jamesmckinna:
> I've been scouring Wikipedia pages about logic trying to find a good name... it's _almost_ negation introduction, but not quite - that's `(A -> B) -> (A -> ¬ B) -> ¬ B`, if you pass `id` as the first argument you get this. 

Hmm... I might even take that as an argument *not* to add the new lemma, in favour of `¬-intro`, although we tend not to have such 'Hilbert-style' principles in the library, because the `A ->_` is essentially handled by parametrisation/scope, and then delegation to `contradiction`... NB did you mean `... -> ¬ A` in the above? 

> Of the suggestions so far, I like `contra-diagonal` the best

@JacquesCarette would you be OK with this?

JacquesCarette:
Fine with `contra-diagonal` (though it's barely shorter than `self-contradiction`...)

JacquesCarette (approved):

PR #2694, opened 2025-04-11
[ fix ] attach fixity to the new declaration of `Function.Base.case_returning_of_`

jamesmckinna:
Leftover from #2157


gallais:
We probably need to keep both for backwards compatibility reasons?

jamesmckinna:
> We probably need to keep both for backwards compatibility reasons?

Ah, maybe? The original deprecation is from v2.0, so I somehow imagined that 'backwards compatibility' now need not extend to the old name. But I guess yours is the more conservative/compatible solution... will update.

Taneb (approved):

PR #2703, opened 2025-04-16
[add] properties of zipWith to Pointwise (Data.Vec)

carlostome:
Add some properties of zipWith w.r.t. Pointwise for `Data.Vec`

### DIFF
--- a/src/Data/Vec/Relation/Binary/Pointwise/Inductive.agda
+++ b/src/Data/Vec/Relation/Binary/Pointwise/Inductive.agda
@@ -8,6 +8,8 @@
 
 module Data.Vec.Relation.Binary.Pointwise.Inductive where
 
+open import Algebra.Definitions
+  using (Associative; Commutative; LeftIdentity; RightIdentity; Congruent₂)
 open import Data.Fin.Base using (Fin; zero; suc)
 open import Data.Nat.Base using (ℕ; zero; suc)
 open import Data.Product.Base using (_×_; _,_; uncurry; <_,_>)
@@ -33,6 +35,7 @@ private
     B : Set b
     C : Set c
     D : Set d
+    n : ℕ
 
 ------------------------------------------------------------------------
 -- Definition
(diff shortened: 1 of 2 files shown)


gallais (approved):

src/Data/Vec/Relation/Binary/Pointwise/Inductive.agda
```diff
@@ -232,6 +235,42 @@ module _ {_∼_ : Rel A ℓ} (refl : Reflexive _∼_) where
   cong-[ zero ]≔  p (_   ∷ eqn) = refl ∷ eqn
   cong-[ suc i ]≔ p (x∼y ∷ eqn) = x∼y  ∷ cong-[ i ]≔ p eqn
 
+------------------------------------------------------------------------
+-- zipWith
+
+module _ {_∼_ : Rel A ℓ} where
+  module _ {f : A → A → A} where
```
JacquesCarette, line 242: If you're going to go with nested anonymous modules (rather than private variables), then I would also lift `assoc : Associative _∼_ f` to the module signature, so that it doesn't have to be repeated in the code below. Same with all the other modules.

But I'm going to leave this as a suggestion rather than a must-do. I don't think we have fully settled on a style for this yet.

JacquesCarette (approved):

src/Data/Vec/Relation/Binary/Pointwise/Inductive.agda
carlostome: For consistency, I followed the style in `Data.Vec.Properties` (https://github.com/agda/agda-stdlib/blob/8c5dee34ebde4734f8da5369c3f735576a6311fc/src/Data/Vec/Properties.agda#L572).

JacquesCarette: Right - consistency wins! Ignore my suggestion for this PR.

PR #2709, opened 2025-04-24
[ refactor ] generalise inclusion between `Propositional` permutation, and arbitrary `Setoid`-based version on same carrier

### LINKED ISSUE #2708: This implication should work for any setoid-based permutation
This implication should work for any setoid-based permutation, not just the equality setoid:

https://github.com/agda/agda-stdlib/blob/3880dec52b0b97e600c302f26315184142d4fd2f/src/Data/List/Relation/Binary/Permutation/Propositional.agda#L98

jamesmckinna:
One approach which fixes #2708. 

NB:
* v2.3 compatible
* bit clunky, esp. wrt naming
* `CHANGELOG` entry doesn't explicitly link `isEquivalence` argument with target relation
* fixes typo in subsequent entry
* v3.0 version might form part of ongoing (but stalled) refactoring project #1354 and associated PRs, with better choices of names in particular

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,7 +243,12 @@ Additions to existing modules
   map-downFrom : ∀ (f : ℕ → A) n → map f (downFrom n) ≡ applyDownFrom f n
   ```
 
-* In `Data.List.Relation.Binary.Permutation.PropositionalProperties`:
+* In `Data.List.Relation.Binary.Permutation.Propositional`:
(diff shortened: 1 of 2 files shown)


jamesmckinna:
@onestruggler @MatthewDaggitt if you each happy with this as us, suggest that we merge, and do the better job above for v3.0

MatthewDaggitt (approved):

Issue #2712, opened 2025-05-05
Incorrect deprecation target in `Relation.Binary.OrderMorphism` warning


jmougeot:
In [src/Relation/Binary/OrderMorphism.agda](https://github.com/agda/agda-stdlib/blob/master/src/Relation/Binary/OrderMorphism.agda)  the deprecation warning currently states:

`{-# WARNING_ON_IMPORT
"Relation.Binary.OrderMorphism was deprecated in v1.5.
Use Relation.Binary.Reasoning.Morphism instead."
#-}`

However, the module `Relation.Binary.Reasoning.Morphism` does not exist. It seems the intended replacement is:
[src/Relation/Binary/Morphism/Bundles.agda](https://github.com/agda/agda-stdlib/blob/master/src/Relation/Binary/Morphism/Bundles.agda)

JacquesCarette:
Indeed - @jmougeot could you put together a PR to fix it?

jamesmckinna:
As I commented on the PR, though: the module where this occurs is deprecated, moreover since v1.5, so will in any case be removed soon, and anyone using `stdlib` should *not* ordinarily be expecting to use this module. Are we creating (unnecessary) work for ourselves here (it feels as though I am, even typing this...)?

jamesmckinna:
I'll stop banging my particular drum!

jamesmckinna:
Closed by #2713

PR #2713, opened 2025-05-06
Fix bug in Relation.Binary.Morphism.Bundles usage

jmougeot:
Update the deprecation warning to correctly reference Relation.Binary.Morphism.Bundles instead of Relation.Binary.Reasoning.Morphism.

### DIFF
--- a/src/Relation/Binary/OrderMorphism.agda
+++ b/src/Relation/Binary/OrderMorphism.agda
@@ -11,7 +11,7 @@ module Relation.Binary.OrderMorphism where
 
 {-# WARNING_ON_IMPORT
 "Relation.Binary.OrderMorphism was deprecated in v1.5.
-Use Relation.Binary.Reasoning.Morphism instead."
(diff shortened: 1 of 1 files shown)


jamesmckinna:
As a matter of practice, we don't accept PRs on deprecated modules unless this actually gives rise to a bug...
```agda
-- The Agda standard library
--
-- This module is DEPRECATED. Please use `Relation.Binary.Morphism`
-- instead.

module Relation.Binary.OrderMorphism where

{-# WARNING_ON_IMPORT
"Relation.Binary.OrderMorphism was deprecated in v1.5.
Use Relation.Binary.Reasoning.Morphism instead."
#-}
```
What exactly is the bug here?

JacquesCarette:
The bug is documented in issue #2712 which @jmougeot should have linked.

JacquesCarette (approved):

jamesmckinna (approved):

PR #2715, opened 2025-05-09
Adds Algebra.Morphism.Construct.DirectProduct

carlostome:
This PR adds proofs that projections from the product type are structure preserving maps (morphisms) from the direct product construction.

This not complete whatsoever, but at least it gives an initial template that can be extended over time.

### DIFF
--- a/src/Algebra/Morphism/Construct/DirectProduct.agda
+++ b/src/Algebra/Morphism/Construct/DirectProduct.agda
@@ -1,0 +1,66 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The projection morphisms for alegraic structures arising from the
+-- direct product construction
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Algebra.Morphism.Construct.DirectProduct where
+
+open import Algebra.Bundles
+open import Algebra.Morphism.Structures
+  using ( module MagmaMorphisms
+        ; module MonoidMorphisms
+        )
+open import Data.Product
+open import Level using (Level)
+open import Relation.Binary.Definitions using (Reflexive)
+open import Algebra.Construct.DirectProduct
+
+private
+  variable
+    c ℓ : Level
+
+------------------------------------------------------------------------
+-- Magmas
+
+module _ (M N : RawMagma c ℓ) (open RawMagma M) (refl : Reflexive _≈_) where
+  open MagmaMorphisms (rawMagma M N) M
+
+  isMagmaHomomorphism-proj₁ : IsMagmaHomomorphism proj₁
+  isMagmaHomomorphism-proj₁ = record
+    { isRelHomomorphism = record { cong = λ {x} {y} z → z .proj₁ }
+    ; homo              = λ _ _ → refl
+    }
+
+module _ (M N : RawMagma c ℓ) (open RawMagma N) (refl : Reflexive _≈_) where
+  open MagmaMorphisms (rawMagma M N) N
+
+  isMagmaHomomorphism-proj₂ : IsMagmaHomomorphism proj₂
+  isMagmaHomomorphism-proj₂ = record
+    { isRelHomomorphism = record { cong = λ {x} {y} z → z .proj₂ }
+    ; homo              = λ _ _ → refl
+    }
+
+------------------------------------------------------------------------
+-- Monoids
+
+module _ (M N : RawMonoid c ℓ) (open RawMonoid M) (refl : Reflexive _≈_) where
+  open MonoidMorphisms (rawMonoid M N) M
+
+  isMonoidHomomorphism-proj₁ : IsMonoidHomomorphism proj₁
+  isMonoidHomomorphism-proj₁ = record
+    { isMagmaHomomorphism = isMagmaHomomorphism-proj₁ _ _ refl
+    ; ε-homo = refl
+    }
+
+module _ (M N : RawMonoid c ℓ) (open RawMonoid N) (refl : Reflexive _≈_) where
+  open MonoidMorphisms (rawMonoid M N) N
+
+  isMonoidHomomorphism-proj₂ : IsMonoidHomomorphism proj₂
+  isMonoidHomomorphism-proj₂ = record
+    { isMagmaHomomorphism = isMagmaHomomorphism-proj₂ _ _ refl
+    ; ε-homo = refl
+    }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,6 +155,8 @@ New modules
 
 * `Algebra.Module.Properties.{Bimodule|LeftModule|RightModule}`.
 
+* `Algebra.Morphism.Construct.DirectProduct`.
+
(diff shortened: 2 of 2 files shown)


JacquesCarette (approved):
looks good to me

jamesmckinna:
Hmm: some things to maybe (re) consider 
* names?
* tighter `import`s
* module organisation, including the inlining of the `IsRelHomomorphism` definitions, seems sub-optimal
* the interesting morphism, namely into the product object, isn't defined, not a deal breaker but defining it might expose further some of the issues above?

src/Algebra/Morphism/Construct/DirectProduct.agda
```diff
@@ -0,0 +1,66 @@
...
+  variable
+    c ℓ : Level
+
+------------------------------------------------------------------------
+-- Magmas
+
+module _ (M N : RawMagma c ℓ) (open RawMagma M) (refl : Reflexive _≈_) where
+  open MagmaMorphisms (rawMagma M N) M
+
+  isMagmaHomomorphism-proj₁ : IsMagmaHomomorphism proj₁
```
MatthewDaggitt, line 32: By convention in the library these should be named `proj₁-isMagmaHomomorphism`

MatthewDaggitt (changes requested):
And I also agree with @jamesmckinna's suggestion that the actual morphism should be listed explicitly.

src/Algebra/Morphism/Construct/DirectProduct.agda
```diff
@@ -0,0 +1,66 @@
...
+  isMagmaHomomorphism-proj₁ = record
+    { isRelHomomorphism = record { cong = λ {x} {y} z → z .proj₁ }
```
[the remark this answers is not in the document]
JacquesCarette: Why bind names that are not used? Could we even eta reduce to just `proj₁` ?

carlostome:
I've tried to address the comments above.

However, I can't understand how the suggestion in:

> * slice the module structure you have orthogonally: so that eg. a `module Proj₁ (refl : Reflexive M._≈_) where` defines the various versions of `proj₁` for each of the algebraic categories... etc. (WIP)

would work. By this I mean, that the argument `(refl : Reflexive M._≈_)` already depends on a chosen algebraic structure 
`M`, and thus it does not make sense to define the "various versions" of functions under such module.

I already rebased this branch on top of (a squashed version of) `jamesmckinna:rel-product` (see #2720) to have access to "relational homomorphisms".

jamesmckinna:
Thanks @carlostome yes I think you're right that my suggestion about restructuring the module organisation was wrong.
That said, I'd be tempted to replay the 'export' trick from `Relation.Binary.Morphism.Construct.Product` to make your definitions in `Algebra.Morphism.Construct.DirectProduct` a bit less unwieldy, both to define, and to deploy?

src/Algebra/Morphism/Construct/DirectProduct.agda
```diff
@@ -59,13 +58,34 @@ module Magma (M : RawMagma a ℓ₁) (N : RawMagma b ℓ₂) where
...
 
+-- Package for export
+module Magma-Export {M : RawMagma a ℓ₁} {N : RawMagma b ℓ₂} where
+  open Magma
+
+  private
+    module M = RawMagma M
+    module N = RawMagma N
+
+  module _ {refl : Reflexive M._≈_} where
```
JacquesCarette, line 76: why use an anonymous module here, and no explicit signature? This feels like it is making things obscure.

src/Algebra/Morphism/Construct/DirectProduct.agda
```diff
@@ -107,3 +127,24 @@ module Monoid (M : RawMonoid a ℓ₁) (N : RawMonoid b ℓ₂) where
...
+    proj₁ = Proj₁.isMonoidHomomorphism M M refl
+
+  module _ {refl : Reflexive N._≈_} where
+    proj₂ = Proj₂.isMonoidHomomorphism M N refl
+
+  module _ {P : RawMonoid c ℓ₃} where
+
+    private
+      module P = RawMonoid P
```
JacquesCarette: It doesn't look like you are using this below, you are using the record, not the module, no?

carlostome: This was an overlook

src/Algebra/Morphism/Construct/DirectProduct.agda
carlostome: I followed the approach in #2720 (for consistency). Meaning, that implicit parameters are wrapped in anonymous modules

JacquesCarette: I wish I had noticed that in that PR! But I can't fault you for being consistent.

JacquesCarette (approved):

src/Algebra/Morphism/Construct/DirectProduct.agda
jamesmckinna: And #2720 is yet to be merged, so feel free @JacquesCarette to comment there on my design (which you had previously approved?)

jamesmckinna:
Now that #2720 has been merged, suggest we try to tie this one up for v2.3 also?

src/Algebra/Morphism/Construct/DirectProduct.agda
JacquesCarette: I'm actually fine with anonymous modules. It's the proliferation of tiny named modules that I'm less keen on. I notice that it uses that too, and I had missed that.

src/Algebra/Morphism/Construct/DirectProduct.agda
```diff
@@ -0,0 +1,142 @@
...
+                          IsMagmaHomomorphism P M f →
+                          IsMagmaHomomorphism P N h →
+                          IsMagmaHomomorphism P (rawMagma M N) (Product.< f , h >)
+    isMagmaHomomorphism F H = record
+      { isRelHomomorphism = RP.< F.isRelHomomorphism , H.isRelHomomorphism >
+      ; homo              = λ x y → F.homo x y , H.homo x y
+      }
+      where
+        module F = IsMagmaHomomorphism F
+        module H = IsMagmaHomomorphism H
```
jamesmckinna: It's purely cosmetic, but I don't personally care for the choice of `f`, `h` as root names for the two legs of the product diagram: suggest instead
* __either__ `f` and `g` (and `module G = Is*Homomorphism G` etc.)
* __or__ `h` and `k`

jamesmckinna (changes requested):
Apart form the final nitpick, looks good to go. 
Many thanks for the contribution, and your patience during the review process!

src/Algebra/Morphism/Construct/DirectProduct.agda
jamesmckinna: You renamed `H` to `G`, but not (yet) `h` to `g`... but otherwise, this now looks good to go.

JacquesCarette:
So we need an approve from @MatthewDaggitt .

MatthewDaggitt (approved):

PR #2719, opened 2025-05-26
Superset relation on Data.Fin.Subsets

Taneb:


src/Data/Fin/Subset.agda
```diff
@@ -77,6 +77,18 @@ p ⊂ q = p ⊆ q × ∃ (λ x → x ∈ q × x ∉ p)
 _⊄_ : Subset n → Subset n → Set
 p ⊄ q = ¬ (p ⊂ q)
 
+_⊇_ : Subset n → Subset n → Set
+p ⊇ q = ∀ {x} → x ∈ q → x ∈ p
```
MatthewDaggitt, line 81: I think it would be safer just to define all of these as e.g. `flip _⊆_`? That's how we usually define these operators and semantically what they should be.

MatthewDaggitt (changes requested):
Otherwise looks good to me!

MatthewDaggitt (approved):

PR #2720, opened 2025-05-26
[ add ] product structure on `RawSetoid`

jamesmckinna:
Lifts out the relevant constructions on `RawSetoid` from #2715 which is now marked blocking on this PR.

NB.
* constructions defined in named modules with explicit parametrisation, each as the relevant `IsRelHomomorphism`
* *then* defined for export in terms of the 'usual' expected names, with *implicit* parametrisation
* should the underlying `Pointwise` product of `RawSetoid`s be lifted out (trivially) to `Relation.Binary.Construct.Product`? (see discussion with/comments from @JacquesCarette below...)

### DIFF
--- a/src/Relation/Binary/Morphism/Construct/Product.agda
+++ b/src/Relation/Binary/Morphism/Construct/Product.agda
@@ -2,7 +2,7 @@
 -- The Agda standard library
 --
 -- The projection morphisms for relational structures arising from the
--- product construction
+-- non-dependent product construction
 ------------------------------------------------------------------------
 
 {-# OPTIONS --safe --cubical-compatible #-}
--- a/src/Relation/Binary/Morphism/Construct/Product.agda
+++ b/src/Relation/Binary/Morphism/Construct/Product.agda
@@ -66,9 +66,12 @@ module _ (M : RawSetoid a ℓ₁) (N : RawSetoid b ℓ₂) where
 ------------------------------------------------------------------------
 -- package up for export
 
-module _ {M : RawSetoid a ℓ₁} {N : RawSetoid b ℓ₂} {P : RawSetoid c ℓ} where
+module _ {M : RawSetoid a ℓ₁} {N : RawSetoid b ℓ₂} where
 
   proj₁ = Proj₁.isRelHomomorphism M N
   proj₂ = Proj₂.isRelHomomorphism M N
+
+module _ {M : RawSetoid a ℓ₁} {N : RawSetoid b ℓ₂} {P : RawSetoid c ℓ} where
+
   <_,_> = Pair.isRelHomomorphism M N P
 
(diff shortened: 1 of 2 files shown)


jamesmckinna:
Don't really understand why the tests are failing, but seems to be a GitHub problem, rather than at my end?

src/Relation/Binary/Morphism/Construct/Product.agda
```diff
@@ -0,0 +1,77 @@
...
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Relation.Binary.Morphism.Construct.Product where
+
+import Data.Product.Base as Product using (<_,_>; proj₁; proj₂)
+open import Data.Product.Relation.Binary.Pointwise.NonDependent as Pointwise
+  using (Pointwise)
+open import Level using (Level)
+--open import Relation.Binary.Construct using ()
```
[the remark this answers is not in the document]
jamesmckinna: Good catch! I'd been havering about lifting out the construction of the product `RawSetoid` to (a hypothetical) `Relation.Binary.Construct.Product` for the sake of structural consistency wrt the hierarchies, but didn't in the end feel it was worth the additional effort? 

Please advise if you think it is!

Otherwise, will remove this, and the subsequent comment on L26...

src/Relation/Binary/Morphism/Construct/Product.agda
```diff
@@ -0,0 +1,77 @@
...
+open import Relation.Binary.Bundles.Raw using (RawSetoid)
+open import Relation.Binary.Morphism.Structures using (IsRelHomomorphism)
+
+private
+  variable
+    a b c ℓ₁ ℓ₂ ℓ : Level
+
+
+------------------------------------------------------------------------
+-- product construction on RawSetoid belongs in Relation.Binary.Construct.Product?
```
jamesmckinna, line 26: Is this comment worth following up on, so that the construction of `P` is moved to such a destination, and then re-imported here?

jamesmckinna: As above: possible downstream refactoring!

jamesmckinna:
Last last thought: should this be parametrised on `RawSetoid`, or simply on a (pair of) bare relation(s) `_≈_` (as in the `Function.Construct.*` and existing `Relation.Binary.Construct.*` hierarchies)?

JacquesCarette:
> Last last thought: should this be parametrised on `RawSetoid`, or simply on a (pair of) bare relation(s) `_≈_` (as in the `Function.Construct.*` and existing `Relation.Binary.Construct.*` hierarchies)?

Ah, the perennial (un)bundling question.

The biggest advantage of the unbundled version is that having a shared carrier of two setoids becomes easy. Having to bundle obscures things in that case and downstream uses might then struggle to recover that information.

That is, however, a pure hypothetical. However, as you mention, `Relation.Binary.Construct.*` seems to make the opposite decision, and being consistent with that does seem like a good idea? I would want these new functions to interact smoothly with the parts of the library that are 'related' (pun unintended) as the main design principle.

Issue #2717, opened 2025-05-12
Clean up basic order theory definitions


jmougeot:
Was working with @TOTBWF to port over some domain theory from the 1lab, and noticed that the definitions of joins/meets in preorders are a bit all over the place both in terms of naming and semantics.

- What is typically called a top element is referred to as a [Maximum](https://github.com/agda/agda-stdlib/blob/9bf80c4c209715be2a212509f29f098cb62f4509/src/Relation/Binary/Definitions.agda#L129); likewise for bottom elements and [Minimum](https://github.com/agda/agda-stdlib/blob/9bf80c4c209715be2a212509f29f098cb62f4509/src/Relation/Binary/Definitions.agda#L139).
- Similarly, the definitions of [Suprema](https://github.com/agda/agda-stdlib/blob/9bf80c4c209715be2a212509f29f098cb62f4509/src/Relation/Binary/Lattice/Definitions.agda#L28) and [Infinima](https://github.com/agda/agda-stdlib/blob/9bf80c4c209715be2a212509f29f098cb62f4509/src/Relation/Binary/Lattice/Definitions.agda#L32) are defined in terms of binary operators that assign pairs of elements to a binary join/meet: this makes it impossible to say that a particular element is an join/meet of two elements. Also, these are defined under Relation.Binary.Lattice.Definitions.
- There are no definitions for joins/meets of families (that we are aware of!)

I suggest that we perform the following rationalisation:

1. Deprecate Maximum and Minimum in favor of IsTop and IsBottom
2. Deprecate Supremum and Infininum in favor of IsBinaryJoin and IsBinaryMeet in Relation.Binary.Definitions, which both state that a *particular* element is a binary meet/join of a pair of elements.
3. Add IsJoin and IsMeet to Relation.Binary.Definitions, which express that a *particular* element is a meet/join of a family.
4. Provided bundled versions of all the above.

I'm open to alternative names and locations. Note that this would be a breaking change, so we would have to take the necessary steps towards deprecation.

jamesmckinna:
Nice issue to raise... the challenge/paradox of making naming choices in a 'standard' library: 
* what *may* be "typically" called one thing (eg 'top') in one area of mathematics (eg. domain theory) *may* be called something else (eg. 'maximum') in another area (even a cognate one, eg. general theory of ordered structures)
[...]

PR #2721, opened 2025-05-26
 [Add] Initial files for Domain theory 

jmougeot:
This draft pull request introduces three new files under src/Relation/Binary/Domain, providing foundational modules for domain theory in the Agda standard library. The new modules include definitions and structures such as DCPO, Lub, and ScottContinuous, aiming to extend the library's support for domain-theoretic formalizations. This is an initial draft and feedback is welcome on the design and implementation.

The definition of Lub provided here is likely better placed elsewhere in the library and may require a different name.  For further discussion, see issue [#2717](https://github.com/agda/agda-stdlib/issues/2717).


jamesmckinna:
Any particular reason to introduce this material as a separate sub-hierarchy under `Relation.Binary.Domain`?

src/Relation/Binary/Domain/Definitions.agda
```diff
@@ -0,0 +1,28 @@
...
+  variable
+    c o ℓ e o' ℓ' e' ℓ₁ ℓ₂ : Level
+    Ix A B : Set o
+    P : Poset c ℓ e
+
+------------------------------------------------------------------------
+-- Directed families
+------------------------------------------------------------------------
+
+IsSemidirectedFamily : (P : Poset c ℓ₁ ℓ₂) → ∀ {Ix : Set c} → (s : Ix → Poset.Carrier P) → Set _
```
MatthewDaggitt: This doesn't match the format of definitions elsewhere in the library. 

- In `Module.Definitions` we still just parameterise them by the raw relations/predicates/sets. We don't parameterise the definitions by the bundles themselves.
- The name `Is` is reserved for "Structures", I would just call this `SemiDirected`.
- The name for standard index sets is just `I`, but I think in this case it's not actually being used an index in the type so the standard name `A` or `B` would be better?
- What does the name `s` stand for?

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+open import Data.Nat.Properties using (≤-trans)
+open import Function using (_∘_)
+open import Level using (Level; _⊔_; suc)
+open import Relation.Binary.Bundles using (Poset)
+open import Relation.Binary.Domain.Definitions
+open import Relation.Binary.Morphism.Structures using (IsOrderHomomorphism)
+
+private variable
+  o ℓ e o' ℓ' e' ℓ₂ : Level
+  Ix A B : Set o
```
MatthewDaggitt: A lot of these variables don't seem to be needed (and indeed are not standard names in the library) while other names used throughout the file (e.g. `c`, `ℓ₁`) could be variables.

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+
+module _ {c ℓ₁ ℓ₂ : Level} (P : Poset c ℓ₁ ℓ₂) where
+  open Poset P
+
+  record IsDirectedFamily {Ix : Set c} (s : Ix → Carrier) : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
```
MatthewDaggitt: Why is this record called `IsDirectedFamily` but the property references `SemiDirected`?

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+    no-eta-equality
+    field
+      elt : Ix
```
MatthewDaggitt: What does `elt` stand for?

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+      SemiDirected : IsSemidirectedFamily P s
+
+  record IsLub {Ix : Set c} (s : Ix → Carrier) (lub : Carrier) : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      is-upperbound : ∀ i → s i ≤ lub
```
MatthewDaggitt: Feels like this should be a definition in `Definitions`

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+      is-least : ∀ y → (∀ i → s i ≤ y) → lub ≤ y
```
MatthewDaggitt: This line could then reuse it.

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+      ⋁ : ∀ {Ix : Set c}
+        → (s : Ix → Carrier)
+        → IsDirectedFamily s
+        → Carrier
+      ⋁-isLub : ∀ {Ix : Set c}
+        → (s : Ix → Carrier)
+        → (dir : IsDirectedFamily s)
+        → IsLub s (⋁ s dir)
+
+    module _ {Ix : Set c} {s : Ix → Carrier} {dir : IsDirectedFamily s} where
```
MatthewDaggitt: What's the point of this module's renaming? Could we document it?

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+      open IsLub (⋁-isLub s dir)
+        renaming (is-upperbound to ⋁-≤; is-least to ⋁-least)
+        public
+
+module _ {c ℓ₁ ℓ₂ : Level} {P : Poset c ℓ₁ ℓ₂} {Q : Poset c ℓ₁ ℓ₂} where
```
MatthewDaggitt: Do the posets really have to live in the same universes?

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+
+  private
+    module P = Poset P
+    module Q = Poset Q
+
+  record IsScottContinuous (f : P.Carrier → Q.Carrier) : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+    field
+      PreserveLub : ∀ {Ix : Set c} {s : Ix → P.Carrier}
```
MatthewDaggitt: Names should lower case and grammatical.

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+
+  record IsLub {Ix : Set c} (s : Ix → Carrier) (lub : Carrier) : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      is-upperbound : ∀ i → s i ≤ lub
+      is-least : ∀ y → (∀ i → s i ≤ y) → lub ≤ y
+
+  record IsDCPO : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+    field
+      ⋁ : ∀ {Ix : Set c}
+        → (s : Ix → Carrier)
```
MatthewDaggitt: Please use the standard formatting for laying out types over multiple lines (see the library style guide)

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,66 @@
...
+    field
+      elt : Ix
+      SemiDirected : IsSemidirectedFamily P s
+
+  record IsLub {Ix : Set c} (s : Ix → Carrier) (lub : Carrier) : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
+    field
+      is-upperbound : ∀ i → s i ≤ lub
+      is-least : ∀ y → (∀ i → s i ≤ y) → lub ≤ y
+
+  record IsDCPO : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
```
MatthewDaggitt: Please use the full expanded acronym in the name.

src/Relation/Binary/Domain/Bundles.agda
```diff
@@ -0,0 +1,53 @@
...
+    f             : Poset.Carrier P → Poset.Carrier Q
+    Scottfunction : IsScottContinuous {P = P} {Q = Q} f
+
+------------------------------------------------------------------------
+-- Lubs
+------------------------------------------------------------------------
+
+record Lub {c ℓ₁ ℓ₂ : Level} {P : Poset c ℓ₁ ℓ₂} {Ix : Set c} (s : Ix → Poset.Carrier P) : Set (c ⊔ ℓ₁ ⊔ ℓ₂) where
+  private
+    module P = Poset P
```
MatthewDaggitt: A better way of doing this would be `open Poset P`, then you don't need the qualifiers. 

Also every bundle should have a corresponding structure...

src/Relation/Binary/Domain/Bundles.agda
```diff
@@ -0,0 +1,53 @@
...
+
+------------------------------------------------------------------------
+-- Scott-continuous functions
+------------------------------------------------------------------------
+
+record ScottContinuous {c ℓ₁ ℓ₂ : Level} {P : Poset c ℓ₁ ℓ₂} {Q : Poset c ℓ₁ ℓ₂} :
+  Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
+  field
+    f             : Poset.Carrier P → Poset.Carrier Q
+    Scottfunction : IsScottContinuous {P = P} {Q = Q} f
```
MatthewDaggitt: Why do you need to provide `P` and `Q` explicitly? If we need to, that's a bug in the implicitness of the arguments that we should fix.

src/Relation/Binary/Domain/Bundles.agda
```diff
@@ -0,0 +1,53 @@
...
+private
+  variable
+    o ℓ e o' ℓ' e' ℓ₂ : Level
+    Ix A B : Set o
+
+------------------------------------------------------------------------
+-- DCPOs
+------------------------------------------------------------------------
+
+record DCPO (c ℓ₁ ℓ₂ : Level) : Set (suc (c ⊔ ℓ₁ ⊔ ℓ₂)) where
```
MatthewDaggitt: Please use the full expanded name.

MatthewDaggitt (changes requested):

src/Relation/Binary/Domain/Bundles.agda
jmougeot: When I remove the explicit arguments {P = P} and {Q = Q}, I get the following error:

`All Goals, Errors
_ℓ₂_65 : Level
_P._≈__69 : Relation.Binary.Core.Rel (Poset.Carrier P) _ℓ₁_64
_P._≤__70 : Relation.Binary.Core.Rel (Poset.Carrier P) _ℓ₂_65
_P.isPartialOrder_71 : Relation.Binary.Structures.IsPartialOrder _P._≈__69 _P._≤__70
_Q._≈__73 : Relation.Binary.Core.Rel (Poset.Carrier Q) _ℓ₁_64
_Q._≤__74 : Relation.Binary.Core.Rel (Poset.Carrier Q) _ℓ₂_65
_Q.isPartialOrder_75 : Relation.Binary.Structures.IsPartialOrder _Q._≈__73 _Q._≤__74`

I’m not sure whether this is due to an actual bug in the implicit resolution or if it’s caused by some ambiguity introduced by the IsScottContinuous record, which might be making the posets harder to infer.

src/Relation/Binary/Domain/Definitions.agda
jmougeot: I’ve updated the definition to avoid using the Poset bundle and instead parameterised it by the raw relation, as I understood from your note about Module.Definitions. I also renamed s to f for clarity and dropped the Is prefix in favour of semidirected, following the naming conventions for properties rather than structures.

That said, I’m not entirely sure if this is what you had in mind — especially regarding the commented-out IsSemidirectedFamily.

src/Padrightpropertiesdraft.agda
```diff
@@ -0,0 +1,105 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Padrightpropertiesdraft where
+
+open import Data.Fin.Base using (Fin; zero; suc; inject≤)
+open import Data.Nat.Base using (ℕ; zero; suc; _+_; _≤_; z≤n; s≤s; _∸_)
+open import Data.Nat.Properties using (≤-refl; ≤-trans; m≤m+n; m+n∸m≡n)
+open import Data.Vec.Base
```
JacquesCarette: `using` ?

src/Padrightpropertiesdraft.agda
```diff
@@ -0,0 +1,105 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of padRight for Vec
```
JacquesCarette: Oops, this should have been in this PR!

src/Padrightpropertiesdraft.agda
```diff
@@ -0,0 +1,105 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Padrightpropertiesdraft where
```
JacquesCarette: Probably should all be added to `Data.Vec.Properties`

src/Padrightpropertiesdraft.agda
```diff
@@ -0,0 +1,105 @@
...
+padRight-map : ∀ (f : A → B) (m≤n : m ≤ n) (a : A) (xs : Vec A m) →
+               map f (padRight m≤n a xs) ≡ padRight m≤n (f a) (map f xs)
+padRight-map f z≤n a [] = map-replicate f a _
+padRight-map f (s≤s m≤n) a (x ∷ xs) = cong (f x ∷_) (padRight-map f m≤n a xs)
+
+------------------------------------------------------------------------
+-- Interaction with lookup
+
+padRight-lookup : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) (i : Fin m) →
+                  lookup (padRight m≤n a xs) (inject≤ i m≤n) ≡ lookup xs i
```
JacquesCarette: Feels like this should be derived via lemmas instead of directly by induction.

`lookup ys (inject ≤ i m≤n)` is `lookup (take i ys) i` and `take i (padRight m≤n a xs)` is `xs`.

src/Padrightpropertiesdraft.agda
```diff
@@ -0,0 +1,105 @@
...
+padRight-lookup (s≤s m≤n) a (x ∷ xs) zero = refl
+padRight-lookup (s≤s m≤n) a (x ∷ xs) (suc i) = padRight-lookup m≤n a xs i
+
+------------------------------------------------------------------------
+-- Interaction with zipWith
+
+-- When both vectors have the same original length
+padRight-zipWith : ∀ (f : A → B → C) (m≤n : m ≤ n) (a : A) (b : B)
```
JacquesCarette: This also feels like it could be proved by a sequence of lemmas. [I will stop review this now, and just review the Domain stuff.]

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,79 @@
...
+open import Level using (Level; _⊔_; suc)
+open import Relation.Binary.Bundles using (Poset)
+open import Relation.Binary.Domain.Definitions
+
+private variable
+  a b c ℓ ℓ₁ ℓ₂ : Level
+  A B : Set a
+
+
+module _ {c ℓ₁ ℓ₂ : Level} (P : Poset c ℓ₁ ℓ₂) where
```
JacquesCarette: Do these definitions really need a `Poset` or just a relation? I don't immediately see the use of any of the other fields of `Poset` below.

src/Relation/Binary/Domain/Structures.agda
jmougeot: I added the renaming to simplify domain properties proofs for instance in the following proof : 

-  With the renaming you can prove immediately:
```
myUpperBound
  : ∀ {B : Set c} (f : B → Carrier) (dir : IsDirectedFamily f) (i : B)
  → f i ≤ ⋁ f dir
myUpperBound f dir i = ⋁-≤ f dir i
```

-  Without the renaming you’d have to write something like:
```
myUpperBound
  : ∀ {B : Set c} (f : B → Carrier) (dir : IsDirectedFamily f) (i : B)
  → f i ≤ ⋁ f dir
myUpperBound′ f dir i = IsLub.isUpperBound (IsDirectedCompletePartialOrder.⋁-isLub f dir) i
```
But let me know if changes are needed

src/Relation/Binary/Domain/Structures.agda
jmougeot: elt stands for element, should I change it ?

src/Relation/Binary/Domain/Structures.agda
TOTBWF: The definition doesn't need a Poset, but this is essentially always going to be used with a Poset/Preorder. Using a bare relation could cause rather poor inference later down the line, so I'd hesitate a bit on this generalization.

We could also use a Preorder here, but that doesn't really buy us much. Doing posetal completions is really, really cheap with setoids, and the preorder versions of lemmas would be essentially the same thing as the plugging in a posetal completion to a poset-based version.

TOTBWF: Speaking of which: is posetal completion anywhere in the stdlib?

src/Relation/Binary/Domain/Definitions.agda
```diff
@@ -0,0 +1,38 @@
...
+private
+  variable
+    a b ℓ : Level
+    A B : Set a
+
+------------------------------------------------------------------------
+-- Directed families
+------------------------------------------------------------------------
+
+-- IsSemidirectedFamily : (P : Poset c ℓ₁ ℓ₂) → ∀ {Ix : Set c} → (s : Ix → Poset.Carrier P) → Set _
```
JacquesCarette: Please delete this (obsolete) comment.

src/Relation/Binary/Domain/Definitions.agda
```diff
@@ -0,0 +1,38 @@
...
+--
+-- Definitions for domain theory
+------------------------------------------------------------------------
+
+
+
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Relation.Binary.Domain.Definitions where
```
JacquesCarette, line 12: Given that neither of the definitions here depend on `Poset` (now), maybe they should go where all the other binary relations are defined?

JacquesCarette (changes requested):

JacquesCarette:
Could you please resolve the CHANGELOG conflicts?

JacquesCarette:
Also: you don't seem to have addressed the latest 2 items that I asked about?

src/Relation/Binary/Domain/Definitions.agda
JacquesCarette: What I mean is that all 3 definitions below should be moved to `Relation.Binary.Definitions`, with the naming made consistent with the names already in that file.

src/Relation/Binary/Domain/Structures.agda
```diff
@@ -0,0 +1,76 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Structures for domain theory
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Relation.Binary.Domain.Structures where
```
JacquesCarette, line 9: To 'fit' the rest of `stdlib`, even though this is all "most useful" for a full `Poset`, these structures should be in terms of a `RawRelation`, as that is all that is used in the definitions.

JacquesCarette:
Depending on what others think (and my review being taken into account), I do think this could make it to 2.3.

src/Relation/Binary/Domain/Structures.agda
jamesmckinna: Looping back to @TOTBWF 's question re poset completion: no. Please consider adding to `stdlib`!
As for 'not buying much' between `Preorder` and `Poset`, we *tend* to try to state (and prove!) things wrt maximum-sensible generality, and antisymmetry feels like 'noise' much of the time in developing the basics... but YMMV.

src/Relation/Binary/Domain/Structures.agda
jamesmckinna: See my latest round of comments on #2809 ... further input from you would be welcome!

jamesmckinna:
Suggest that we close in favour of #2809 ?

src/Relation/Binary/Domain/Structures.agda
jamesmckinna: Apologies, I think I was wrong on poset completion: I think that https://github.com/agda/agda-stdlib/blob/54f5c38799f6c0125f7dc914e073670b549bcbe6/src/Relation/Binary/Construct/Interior/Symmetric.agda#L84-L107
gives you what you want, but not necessarily in a recognisable form... I had to think about it a bit harder while looping back to #2070 / #2071

JacquesCarette:
Since the attributions are correct in that further PR, it's probably best to close this one.

PR #2723, opened 2025-05-30
Insertion sort and its properties. A bug in MergeSort.agda is fixed.

onestruggler:
Besides the sorted and permutation peroperties, we also proved a congruence between permutation relation and pointwise relation on lists. And then we prove permuation relation is decidable.


src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+... | yes x≤z | no y≤z = ⊥-elim (y≤z (≤-respˡ-≈ eq x≤z))
+... | no x≤z | no y≤z = Eq.refl ∷ insert-congˡ xs eq
+
+insert-cong : ∀ {x y xs ys} → x ≈ y → xs ≋ ys →
+              insert x xs ≋ insert y ys
+insert-cong {x} {y} {xs} {ys} eq1 eq2 =
+  transₚₜ (insert-congˡ xs eq1) (insert-congʳ y eq2)
+
+sort-cong : ∀ {xs ys} → xs ≋ ys → sort xs ≋ sort ys
+sort-cong {xs} {ys} [] = []
```
JacquesCarette, line 143: as you don't need these named implicits in the rhs, please elide

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+
+insert-swap : ∀ x y xs → insert x (insert y xs) ≋ insert y (insert x xs)
+insert-swap x y [] with x ≤? y | y ≤? x
+... | yes x≤y | yes y≤x = eq ∷ (Eq.sym eq) ∷ []
+  where eq = antisym x≤y y≤x
+... | no x≤y | yes y≤x = reflₚₜ
+... | yes x≤y | no y≤x = reflₚₜ
+... | no x≤y | no y≤x = (Eq.sym eq) ∷ (eq ∷ [])
+  where eq = ≥-antisym (≰⇒≥ x≤y) (≰⇒≥ y≤x)
+insert-swap x y (z ∷ xs) with x ≤? z in xz | y ≤? z in yz |
```
JacquesCarette, line 154: I think that using a 'long line' here would make things easier to read (even though we normally don't let things get too wide). But that's more of an opinion than a "please change".

JacquesCarette (changes requested):
This is generally looking really nice. I've only spotted some small improvements that ought to be made, and also left a suggestion.

src/Data/List/Sort/InsertionSort.agda
onestruggler: done, and similar implicits are also done.

jamesmckinna:
So the `bug` that is fixed is the generalisation from `Propositional` permutation to `Setoid` permutation? or something else? I wasn't sure just looking at the diffs.

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+  module Eq ; _≈_ ; ≤-respʳ-≈ ; ≤-respˡ-≈ ; antisym)
+  
+open import Data.List.Relation.Binary.Equality.Setoid Eq.setoid
+  using (_≋_)
+open import Data.List.Relation.Binary.Permutation.Setoid Eq.setoid
+open import Data.List.Relation.Unary.Sorted.TotalOrder totalOrder
+  using (Sorted)
+open import Data.List.Sort.Base totalOrder
+open import Relation.Binary.Reasoning.Setoid (setoid Eq.setoid)
+  as SetoidReasoning using ()
```
jamesmckinna, line 46: Why not simply
```suggestion
import Relation.Binary.Reasoning.Setoid (setoid Eq.setoid)
  as SetoidReasoning
```
Although the proliferation of `setoid`s makes it hard to see that this is declaring a reasoning module for `_≋_`... so maybe could be improved by renaming to
```suggestion
import Relation.Binary.Reasoning.Setoid (setoid Eq.setoid)
  as ≋-Reasoning
```

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+
+open Setoid (setoid Eq.setoid)
+  renaming (refl to reflₚₜ ; trans to transₚₜ) using ()
```
jamesmckinna, line 49: This seems *very* strange: surely the (eventually) defined `reflₚₜ` is an alias for `Data.List.Relation.Binary.Equality.Setoid.≋-refl` etc.? So could be fixed up in the `import` on L39-L40?

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+insert-↗ x [] = [-]
+insert-↗ x ([-] {y}) with (x ≤? y)
+... | yes x≤y = x≤y ∷ [-]
+... | no x≤y = ≰⇒≥ x≤y ∷ [-]
+insert-↗ x (_∷_ {y} {z} {ys} y≤z z≤ys) with (x ≤? y)
+... | yes x≤y = x≤y ∷ y≤z ∷ z≤ys
+... | no x≤y with insert-↗ x z≤ys
+... | sd with (x ≤? z)
+... | yes r = ≰⇒≥ x≤y ∷ sd
+... | no r = y≤z ∷ sd
```
jamesmckinna, line 102: This can be streamlined to
```suggestion
... | no x≤y with sd ← insert-↗ x z≤ys | x ≤? z
...   | yes r = ≰⇒≥ x≤y ∷ sd
...   | no r = y≤z ∷ sd
```
(and where, again, the `r`s are redundant, so could be replaced by `_`)

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+insert-swap x y (z ∷ xs) | no x≤z | yes y≤z | no x≤y | no y≤x
+  rewrite xy | yx | yz | xz = reflₚₜ
+insert-swap x y (z ∷ xs) | no x≤z | no y≤z | yes x≤y | yes y≤x
+  rewrite xy | yx | yz | xz = Eq.refl ∷ (insert-swap x y xs)
+insert-swap x y (z ∷ xs) | no x≤z | no y≤z | yes x≤y | no y≤x
+  rewrite xy | yx | yz | xz = Eq.refl ∷ (insert-swap x y xs)
+insert-swap x y (z ∷ xs) | no x≤z | no y≤z | no x≤y | yes y≤x
+  rewrite xy | yx | yz | xz = Eq.refl ∷ (insert-swap x y xs)
+insert-swap x y (z ∷ xs) | no x≤z | no y≤z | no x≤y | no y≤x
+  rewrite xy | yx | yz | xz = Eq.refl ∷ (insert-swap x y xs)
```
jamesmckinna, line 189: Yikes!?
Ordinarily, we try to avoid use of `rewrite` in `stdlib`... but is it avoidable here?
Given the common structure, can this be refactored into `wlog`-like lemmas, to appeal to the symmetry?

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+insert-congˡ {x} {y} (z ∷ xs) eq with x ≤? z | y ≤? z
+... | yes x≤z | yes y≤z = eq ∷ reflₚₜ
+... | no x≤z | yes y≤z = ⊥-elim (x≤z (≤-respˡ-≈ (Eq.sym eq) y≤z))
+... | yes x≤z | no y≤z = ⊥-elim (y≤z (≤-respˡ-≈ eq x≤z))
+... | no x≤z | no y≤z = Eq.refl ∷ insert-congˡ xs eq
+
+insert-cong : ∀ {x y xs ys} → x ≈ y → xs ≋ ys →
+              insert x xs ≋ insert y ys
+insert-cong {x} {y} {xs} {ys} eq1 eq2 =
+  transₚₜ (insert-congˡ xs eq1) (insert-congʳ y eq2)
```
jamesmckinna, line 140: Any reason to use `transₚₜ` rather than `SetoidReasoning` here?

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+  insert x′ (insert y′ ys) ≈⟨ insert-swap x′ y′ ys ⟩
+  insert y′ (insert x′ ys) ∎
+  where open SetoidReasoning
+
+-- Ideally, we want:
+
+--   property1 : ∀ {xs ys} → xs ↭ ys →
+--               Sorted xs → Sorted ys → xs ≋ ys
+
+-- But the induction over xs ↭ ys is hard to do. So instead we have a
```
jamesmckinna, line 205: In what way hard, especially in view of `insert-swap` above? Would the abstract proof via `Sorted` be any worse?

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,230 @@
...
+
+------------------------------------------------------------------------
+-- Decidability property
+
+infix 4 _↭?_
+
+_↭?_ : Decidable _↭_
+xs ↭? ys with decidable Eq._≟_ (sort xs) (sort ys)
+... | yes eq = yes (begin
+  xs      ↭⟨ ↭-sym (sort-↭ xs) ⟩
```
jamesmckinna, line 225:
```suggestion
  xs      ↭⟨ sort-↭ xs ⟨
```
would be cleaner/clearer?

src/Data/List/Sort/InsertionSort.agda
jamesmckinna: Ah... transitivity. A more serious bug in the definition of `Permutation`: transitivity is eliminable!

jamesmckinna (approved):
Lots of suggestions, but a very nice piece of work.

onestruggler:
> So the `bug` that is fixed is the generalisation from `Propositional` permutation to `Setoid` permutation? or something else? I wasn't sure just looking at the diffs.

Luckily, the diffs are the only modification to fix the bug mentioned. The bug is: mergesort used a DecTotalOrder to sort, whose underlying equality is not propositional. Before modification, the permutation property only works for a DecTotalOrder with propositional equality, but now any setoid equivalence.

src/Data/List/Sort/InsertionSort.agda
onestruggler: Nice, I didn't know symmetry can be used this way, and I thought it is a typo there. Now I know. Thanks.

src/Data/List/Sort/InsertionSort.agda
onestruggler: Thanks. Using symmetry is a very good suggestion.

src/Data/List/Sort/InsertionSort.agda
onestruggler: no. just because in this way, the code is shorter. Should I change it?

src/Data/List/Sort/InsertionSort.agda
onestruggler: no longer a problem now.

src/Data/List/Sort/InsertionSort.agda
onestruggler: Now it is just a single line of "trans" (without a linebreak), which is succinct, and looks better than SetoidReasoning.

src/Data/List/Sort/InsertionSort.agda
```diff
@@ -0,0 +1,194 @@
...
+
+insertionSort : SortingAlgorithm
+insertionSort = record
+  { sort   = sort
+  ; sort-↭ = sort-↭
+  ; sort-↗ = sort-↗
+  }
+
+------------------------------------------------------------------------
+-- Congruence properties
```
JacquesCarette, line 102: Hmm, now I notice that usually such things would be in `Data.List.Sort.InsertionSort.Properties` rather than directly in the main file.

JacquesCarette (changes requested):
Sorry, one more thing. This is looking really good now.

src/Data/List/Sort/InsertionSort.agda
jamesmckinna: See #2311

jamesmckinna:
`fix-whitespace` needed!

src/Data/List/Sort/InsertionSort/Properties.agda
```diff
@@ -0,0 +1,182 @@
...
+  insert y′ (insert x′ ys) ∎
+  where open ≋-Reasoning
+
+-- Ideally, we want:
+
+--   property1 : ∀ {xs ys} → xs ↭ ys → Sorted xs → Sorted ys → xs ≋ ys
+
+-- But the induction over xs ↭ ys is hard to do for the "transitive"
+-- constructor. So instead we have a similar property that depends on
+-- the particular sorting algorithm used.
```
MatthewDaggitt, line 161: As you say in the comment, this isn't really the form of the two lemmas we want to prove. Yes, the transitive constructor makes the proof hard, but its not impossible to do. 

I'd really prefer to have the correct lemma in the library, rather than the "wrong" one that we have to go to the faff of removing later... 

If the full lemma is too hard to prove, then I suggest we simply remove these two lemmas.

MatthewDaggitt (changes requested):

MatthewDaggitt:
Also needs a CHANGELOG entry.

src/Data/List/Sort/InsertionSort/Properties.agda
onestruggler: I have tried to prove the full lemma using the current permutation definition, but failed. Issue #2311 started one year ago was trying to fix the definition, but not ready yet.

Ignore the comment for a moment, the property sort-cong-↭ is just a congruence for sort, which itself as a property for sort should be there, but it is implied by the full lemma. Can we change the proof of sort-cong-↭ later and keep this property now? I also have subsequent developement depending on the decidability of permutation. If not, I have no problem removing them.

jamesmckinna: FWIW, #2311 stalled because of the various discussions around making `breaking` changes for v3.0. But as an addition with yet another definition, and a proof of equivalence with curent versions, sure I can add one. 

Incoming... #2726

MatthewDaggitt: @onestruggler I've proved the general lemma 
```agda
↗↭↗⇒≋ : ∀ {xs ys} → Sorted O xs → Sorted O ys → xs ↭ ys → xs ≋ ys
```
in the branch `sorting-perm-equality2`. A PR will be incoming in the next few days, so can you remove these lemmas from your PR. You can then add any extra results you need based on the new lemma in a new PR?

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Minor improvements
 * Moved the concept `Irrelevant` of irrelevance (h-proposition) from `Relation.Nullary`
   to its own dedicated module `Relation.Nullary.Irrelevant`.
 
+* Permutaions in `Data.List.Sort.MergenSort` and `Data.List.Sort.Base` are now setoid based rather than propositional equality based only.
```
MatthewDaggitt: Lots of typos.
- Permutaions
- MergenSort

MatthewDaggitt (changes requested):
Ah also I've just realised a really crucial point! The permutations were propositional based because that is actually the stronger result than them being setoid based permutations! 

Please can you revert the changes back so that they are propositional permutations! (and of course remove that statement from the CHANGELOG!)

MatthewDaggitt:
It might be good to add a comment as well explaining why the proofs are propositional based. Note that it is very easy to convert from propositional permutation to setoid-based permutation (but not the other way around!)

onestruggler:
> Ah also I've just realised a really crucial point! The permutations were propositional based because that is actually the stronger result than them being setoid based permutations!
> 
> Please can you revert the changes back so that they are propositional permutations! (and of course remove that statement from the CHANGELOG!)

I must misundertand something. I think setoid-based sorting is more general (stronger) than propositional based. If you want to sort propositionally you just use a DecTotalOrder whose underlying equivalence is propositional. Then you use the following to get to the "real propsitional":

https://github.com/agda/agda-stdlib/blob/ac6571b1f1831bfad1f3939a052291bbd129b0eb/src/Data/List/Relation/Binary/Permutation/Propositional.agda#L118

But if you only have propositional sorting, you cannot sort setoidly. 

Every propositional equality is a setoid. You can always turn propostional equaltiy into a setoid equivalence by refl but not visa vera, that is setoid is more general.

src/Data/List/Sort/InsertionSort/Properties.agda
onestruggler: The "property1 : ∀ {xs ys} → xs ~ ys → Sorted xs → Sorted ys → xs ≋ ys" is still hard to prove using the new definition in #2726 . Induction over ⋎ cannot go through, becasue I need "b ∷ cs" to be sorted in order to call IH on it. It is unreasonable to ask "b ∷ cs" to be sorted.

onestruggler: If you squint at the following line in the new def of permuation in #2726 

_⋎_  : as ∼ b ∷ cs → a ∷ cs ∼ bs → a ∷ as ∼ b ∷ bs

this line is essentially transitivity (plus swap): as ~ cs -> cs ~ bs -> as ~ bs.

The reason we cannot get rid of transitivity is that transitivity is composition of permutations. We cannot list all N! permutations in our inductive data type, we can only choose some generators, e.g. right now all swaps (transpositions) (we know transpositions generates all permuations) , so we have to compose generators to get all permutations. So transitivity is non-elimable. So induction proof can never avoid transivity, hence hard to go through.

This suggests the "full lemma" as called by Matt is hard to prove using whatever defs of permutations. Hence the not-so-full lemma depending on a particular sorting algorithm seems the only way to do it.

onestruggler: > @onestruggler I've proved the general lemma
> 
> ```agda
> ↗↭↗⇒≋ : ∀ {xs ys} → Sorted O xs → Sorted O ys → xs ↭ ys → xs ≋ ys
> ```
> 
> in the branch `sorting-perm-equality2`. A PR will be incoming in the next few days, so can you remove these lemmas from your PR. You can then add any extra results you need based on the new lemma in a new PR?

Yes, I am going to make a new PR later. Your proof is impressive! Thanks.

MatthewDaggitt:
> I must misundertand something. I think setoid-based sorting is more general (stronger) than propositional based.

The crucial thing is that sorting should preserve the elements in the list right? Even if you are using a setoid based `DecTotalOrder`, you still want to enforce that the sorting algorithm should not be allowed to actually mutate the elements (even if the mutation is equal under the setoid-based equality.)

For example, imagine sorting unnormalised rationals quotiented over by the setoid in the obvious way (e.g. 1/2 = 2/4 in the setoid). If you say that sorting only needs to return a setoid-based permutation then legally `sort [1/3,1/2]` could return `[1000001/2000002, 123/369]`. If you saw this, then you'd think the sorting algorithm had gone very wrong right?

Crucially, it's actually impossible for the sorting algorithm to do that as the way we have created the interface, it is impossible to mutate the elements of the list at all as the interface is entirely generic as to the carrier set being passed in.

src/Data/List/Sort/InsertionSort/Properties.agda
jamesmckinna: @onestruggler thanks for the comments on #2726 suggest we move the discussion there? But yes, the merge constructor indeed incorporates *some* transitivity, but in a more constrained way, moreover list `cs` is *shorter* than `as` or `bs`  (which drives the proof of full transitivity)

jamesmckinna:
Very interesting discussions... who knew insertion sort still had things to teach us!?

onestruggler:
> > I must misundertand something. I think setoid-based sorting is more general (stronger) than propositional based.
> 
> The crucial thing is that sorting should preserve the elements in the list right? Even if you are using a setoid based `DecTotalOrder`, you still want to enforce that the sorting algorithm should not be allowed to actually mutate the elements (even if the mutation is equal under the setoid-based equality.)

Yes, it should preserve elements. But if I used a setoid based `DecTotalOrder` with the setoid being propostional equality, it also never changes elements.

> 
> For example, imagine sorting unnormalised rationals quotiented over by the setoid in the obvious way (e.g. 1/2 = 2/4 in the setoid). If you say that sorting only needs to return a setoid-based permutation then legally `sort [1/3,1/2]` could return `[1000001/2000002, 123/369]`. If you saw this, then you'd think the sorting algorithm had gone very wrong right?

In a setoid, 1/3 is 123/369, the return looks good to me. Also, at least for insertion sort, it does not change the elements whatsoever: it compares and puts everything back.

The old interface cannot sort unnormalised rationals (because the underlying equivalence is non-propositional equivalence), but can sort normalised rationals (because the underlying equivalence is propositional equality), while the new sort can sort both. When the new one sorts normalised rationals, it behaves the same as the old one (guarantted by ↭ₛ⇒↭ : _↭ₛ_ ⇒ _↭_ ). That is the new one is stronger.

I need to hear more voices on this: @JacquesCarette @jamesmckinna @larrytheliquid

src/Data/List/Sort/InsertionSort/Properties.agda
onestruggler: Hi @jamesmckinna ,  sure, let's move to #2726 after this post. Matt has a proof that avoids induction over transitivity, so different defs for permutation is no longer urgent for proving the "full lemma". But simple defs is definitely preferable.

jamesmckinna:
I'm now in meetings for the rest of the day, but lots of food for thought regarding `Propositional` vs. `Setoid` representations wrt stability of sorting... and having not really thought about these things since my PhD thesis (1988--92 ;-)) I need to scratch my head a bit before saying anything stupid here...

MatthewDaggitt:
> Yes, it should preserve elements. But if I used a setoid based DecTotalOrder with the setoid being propostional equality, it also never changes elements.

This is beside the point.

> In a setoid, 1/3 is 123/369, the return looks good to me.

Really, a sorting algorithm that magically changes 1/3 to 123/369 looks good to you?

>  Also, at least for insertion sort, it does not change the elements whatsoever: it compares and puts everything back.

Yes, exactly! That's my point! It's impossible to write a generic sorting algorithm that changes the elements. Hence this should be reflected in the type signature.

> The old interface cannot sort unnormalised rationals (because the underlying equivalence is non-propositional equivalence), but can sort normalised rationals (because the underlying equivalence is propositional equality), while the new sort can sort both. When the new one sorts normalised rationals, it behaves the same as the old one (guarantted by ↭ₛ⇒↭ : ↭ₛ ⇒ ↭ ). That is the new one is stronger.

This is not true. With the current interface you can pass in the total order over unnormalised rationals which uses the setoid-based equality and then derive both the setoid and the non-setoid permutation result:
```agda
open import Data.Rational.Unnormalised
open import Data.Rational.Unnormalised.Properties
open import Data.List
open import Data.List.Sort ≤-decTotalOrder
import Data.List.Relation.Binary.Permutation.Homogeneous as Homo
import Data.List.Relation.Binary.Permutation.Setoid ≃-setoid as Setoid
import Data.List.Relation.Binary.Permutation.Propositional as Prop

l : List ℚᵘ
l = 1ℚᵘ ∷ 0ℚᵘ ∷ []

sl : List ℚᵘ
sl = sort l

sorted-prop-↭ : sl Prop.↭ l
sorted-prop-↭ = sort-↭ l

sorted-setoid-↭ : sl Setoid.↭ l
sorted-setoid-↭ = Homo.map ≃-reflexive (Prop.↭⇒↭ₛ (sorted-prop-↭))
```
So challenge for you. Can you derive `sorted-prop-↭` after your alterations while keeping the imports identical (i.e. while passing in `≤-decTotalOrder` as the argument to `List.Sort`)?

onestruggler:
> Yes, exactly! That's my point! It's impossible to write a generic sorting algorithm that changes the elements. Hence this should be reflected in the type signature.

I finally see your point. Thanks for the clarification. I indeed misunderstood it. I apologize for being stupid. 

In some sense, for sorting, the setoid equivalence in a DecTotalOrder is never used, in other words, in some sense, in the eyes of sorting, any DecTotalOrder is always propostional-based. (the sentence before is just a vague idea, please do not take it literally). So restricting to propostional permutation loses nothing and is indeed stronger than setoid permutation (as the challenge shows, you can never recover propostional equality from setoid equivalence).

This should be documented in case of someone like me misunderstand it again.

JacquesCarette:
Glad it was sorted (ahem) out. Yes, this is super subtle.

Happy to help document -- but where? I guess it doesn't matter as much as to where we put pointers to such documentation, so that you might have seen it while doing this work.

gallais:
While trying to cleanup `insert-swap-≤` (without much success), I proved a few auxiliary things
that may be useful:

```agda

insert-spec : ∀ x xs → insert x xs ≋ takeWhile (¬? ∘ (x ≤?_)) xs ++ x ∷ dropWhile (¬? ∘ (x ≤?_)) xs
insert-spec x [] = Pointwise.refl Eq.refl
insert-spec x (y ∷ xs) with x ≤? y
... | yes x≤y = Pointwise.refl Eq.refl
... | no  x≰y = Eq.refl ∷ insert-spec x xs

insert-here : ∀ {x y ys} → x ≤ y → insert x (y ∷ ys) ≋ x ∷ y ∷ ys
insert-here {x} {y} x≤y with x ≤? y
... | yes _  = Pointwise.refl Eq.refl
... | no x≰y = contradiction x≤y x≰y

insert-there-≰ : ∀ {x y ys} → x ≰ y → insert x (y ∷ ys) ≋ y ∷ insert x ys
insert-there-≰ {x} {y} {ys} x≰y with x ≤? y
... | yes x≤y = contradiction x≤y x≰y
... | no _ = Pointwise.refl Eq.refl
```

MatthewDaggitt:
@onestruggler I think the easiest way to get a win at this point is just to submit a new PR with only `InsertionSort.Base` and `InsertionSort.Properties` in?

We can then refactor `MergeSort` into separate modules in a separate PR if we want to.

onestruggler:
> @onestruggler I think the easiest way to get a win at this point is just to submit a new PR with only `InsertionSort.Base` and `InsertionSort.Properties` in?
> 
> We can then refactor `MergeSort` into separate modules in a separate PR if we want to.

agreed. I will do it soon.

PR #2724, opened 2025-06-02
Some preparatory proofs for proving sorting+permutation is equality

MatthewDaggitt:
I've been sniped by https://github.com/agda/agda-stdlib/pull/2723. This is a preparatory PR for the full proof that sorting two lists that are permutations of each other result in equal lists.


MatthewDaggitt:
Messed this up closing.

PR #2490, opened 2024-10-01
[DRY] Refactor `Data.List.Relation.Binary.Equality.Setoid` exports

jamesmckinna:
Addresses the two examples of the general issue #2391 , but leaving that open for further downstream instances to be tackled.

NB.:
* no `CHANGELOG` required, as a 'pure' refactoring

Issue #2489, opened 2024-10-01
[DRY] Refactor algebraic properties of types


jamesmckinna:
There's (a lot of) duplication between `Function.Related.TypeIsomorphisms` and `Data.Product.Algebra` (and possibly elsewhere: eg the duplication of `True↔` between the first of these, and `Relation.Nullary.Decidable`...).

This might be bad enough to want to deprecate whole modules, but there is a subtle problem exposed by #2419 , namely that not all the lemmas are as level-polymorphic as they might be, esp. those in `Data.Product.Algebra`, where the use of `Algebra.Definitions` imposes a homogeneous level restriction.

So a refined refactoring might be to make`Function.Related.TypeIsomorphisms` home for the heterogeneous level-polymorphic definitions (the direction in #2419 ), while making `Data.Product.Algebra` home for the homogeneous/algebraic versions?

This feels as though it might be a useful clean-up, but potentially `breaking` wrt possibly unresolved `Level` metas and/or name clashing which no obvious (or at least: sensible) deprecation strategy can comfortably accommodate...

JacquesCarette:
`Function.Related.TypeIsomorphism` is both quite old and very useful. The level polymorphism is definitely something worth having.

I wonder if the (deeper) issue is that type "algebra" isn't quite 1-sorted algebra. So that our current `Data.*.Algebra` is correct wrt the usual notion of 'algebra', but we're encountering a genuine generalization here.

jamesmckinna:
> `Function.Related.TypeIsomorphism` is both quite old and very useful. The level polymorphism is definitely something worth having.

I agree (re: level polymorphism... but where these things live is moot, AFAIAC)

jamesmckinna:
> I wonder if the (deeper) issue is that type "algebra" isn't quite 1-sorted algebra. So that our current `Data.*.Algebra` is correct wrt the usual notion of 'algebra', but we're encountering a genuine generalization here.

I'm not sure I quite understand this: some things (esp. `Level`-polymorphic such) are 'large' notions (moreover: typically 'type constructors' in some form or another), whereas relative to a given carrier, the same things are 'small', and more like 'properties'/'structure'...? I was trying to cash that distinction out by saying we should... split them between different modules!?

JacquesCarette:
If Agda had `--cumulativity` as default (which it doesn't, and I'm fine with that), one could see the stuff in `Function.Related.TypeIsomorphism` as 'algebra' for `Set\omega`.

[...]

PR #2725, opened 2025-06-02
 Some preparatory proofs for proving sorting+permutation is equality #2724 

MatthewDaggitt:
I've been sniped by https://github.com/agda/agda-stdlib/pull/2723. This is a preparatory PR for the full proof that sorting two lists that are permutations of each other result in equal lists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,13 +296,15 @@ Additions to existing modules
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
   xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
+  ¬[]↭x∷xs : ¬ ([] ↭ x ∷ xs)
+  ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
   toFin-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (toFin xs↭ys) i)
   ```
 
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
   filter-↭ : ∀ (P? : Pred.Decidable P) → xs ↭ ys → filter P? xs ↭ filter P? ys
-	```
+        ```
 
 * In `Data.List.Relation.Binary.Pointwise.Properties`:
   ```agda
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,7 +285,7 @@ Additions to existing modules
 
 * In `Data.List.Relation.Binary.Permutation.Homogeneous`:
   ```agda
-  toFin : Permutation R xs ys → Fin.Permutation (length xs) (length ys)
+  onIndices : Permutation R xs ys → Fin.Permutation (length xs) (length ys)
   ```
 
 * In `Data.List.Relation.Binary.Permutation.Propositional`:
@@ -296,7 +296,6 @@ Additions to existing modules
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
   xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
-  ¬[]↭x∷xs : ¬ ([] ↭ x ∷ xs)
   ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
   toFin-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (toFin xs↭ys) i)
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -297,7 +297,7 @@ Additions to existing modules
   ```agda
   xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
   ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
-  toFin-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (toFin xs↭ys) i)
+  onIndices-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (onIndices xs↭ys) i)
   ```
 
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,7 +293,7 @@ Additions to existing modules
   ↭⇒↭ₛ′ : IsEquivalence _≈_ → _↭_ ⇒ _↭ₛ′_
   ```
 
-* In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
+* In `Data.List.Relation.Binary.Permutation.Setoid.Properties`:
   ```agda
   xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
   ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -235,6 +235,23 @@ Additions to existing modules
   ∙-cong-∣ : x ∣ y → a ∣ b → x ∙ a ∣ y ∙ b
   ```
 
+* In `Data.Fin.Base`:
+  ```agda
+  _≰_ : ∀ {n} → Rel (Fin n) 0ℓ
+  _≮_ : ∀ {n} → Rel (Fin n) 0ℓ
+  ```
+
+* In `Data.Fin.Permutation`:
+  ```agda
+  cast-id : .(m ≡ n) → Permutation m n
+  swap : Permutation m n → Permutation (suc (suc m)) (suc (suc n))
+  ```
+
+* In `Data.Fin.Properties`:
+  ```agda
+  cast-involutive : .(eq₁ : m ≡ n) .(eq₂ : n ≡ m) → ∀ k → cast eq₁ (cast eq₂ k) ≡ k
+  ```
+
 * In `Data.Fin.Subset`:
   ```agda
   _⊇_ : Subset n → Subset n → Set
@@ -266,6 +283,11 @@ Additions to existing modules
   map-downFrom : ∀ (f : ℕ → A) n → map f (downFrom n) ≡ applyDownFrom f n
   ```
 
+* In `Data.List.Relation.Binary.Permutation.Homogeneous`:
+  ```agda
+  toFin : Permutation R xs ys → Fin.Permutation (length xs) (length ys)
+  ```
+
 * In `Data.List.Relation.Binary.Permutation.Propositional`:
   ```agda
   ↭⇒↭ₛ′ : IsEquivalence _≈_ → _↭_ ⇒ _↭ₛ′_
@@ -273,7 +295,18 @@ Additions to existing modules
 
 * In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
   ```agda
+  xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
+  toFin-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (toFin xs↭ys) i)
+  ```
+
+* In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
+  ```agda
   filter-↭ : ∀ (P? : Pred.Decidable P) → xs ↭ ys → filter P? xs ↭ filter P? ys
+	```
+
+* In `Data.List.Relation.Binary.Pointwise.Properties`:
+  ```agda
+  lookup-cast : Pointwise R xs ys → .(∣xs∣≡∣ys∣ : length xs ≡ length ys) → ∀ i → R (lookup xs i) (lookup ys (cast ∣xs∣≡∣ys∣ i))
   ```
 
 * In `Data.Product.Function.Dependent.Propositional`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,8 +237,8 @@ Additions to existing modules
 
 * In `Data.Fin.Base`:
   ```agda
-  _≰_ : ∀ {n} → Rel (Fin n) 0ℓ
-  _≮_ : ∀ {n} → Rel (Fin n) 0ℓ
+  _≰_ : Rel (Fin n) 0ℓ
+  _≮_ : Rel (Fin n) 0ℓ
   ```
 
 * In `Data.Fin.Permutation`:
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -19,7 +19,7 @@ open import Level using (0ℓ)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; _≢_; refl; cong)
 open import Relation.Binary.Indexed.Heterogeneous.Core using (IRel)
-open import Relation.Nullary.Negation.Core using (contradiction)
+open import Relation.Nullary.Negation.Core using (¬_; contradiction)
 
 private
   variable
@@ -271,7 +271,7 @@ pinch {suc n} (suc i) (suc j) = suc (pinch i j)
 ------------------------------------------------------------------------
 -- Order relations
 
-infix 4 _≤_ _≥_ _<_ _>_
+infix 4 _≤_ _≥_ _<_ _>_ _≰_ _≮_
 
 _≤_ : IRel Fin 0ℓ
 i ≤ j = toℕ i ℕ.≤ toℕ j
@@ -285,6 +285,11 @@ i < j = toℕ i ℕ.< toℕ j
 _>_ : IRel Fin 0ℓ
 i > j = toℕ i ℕ.> toℕ j
 
+_≰_ : ∀ {n} → Rel (Fin n) 0ℓ
+i ≰ j = ¬ (i ≤ j)
+
+_≮_ : ∀ {n} → Rel (Fin n) 0ℓ
+i ≮ j = ¬ (i < j)
 
 ------------------------------------------------------------------------
 -- An ordering view.
--- a/src/Data/Fin/Permutation.agda
+++ b/src/Data/Fin/Permutation.agda
@@ -261,27 +261,7 @@ insert {m} {n} i j π = permutation to from inverseˡ′ inverseʳ′
 -- Note: should be refactored as a special-case when we add the
 -- concatenation of two permutations
 swap : Permutation m n → Permutation (suc (suc m)) (suc (suc n))
-swap {m} {n} π = permutation to from inverseˡ′ inverseʳ′
-  where
-  to : Fin (suc (suc m)) → Fin (suc (suc n))
-  to 0F      = 1F
-  to 1F      = 0F
-  to (suc (suc i)) = suc (suc (π ⟨$⟩ʳ i))
-
-  from : Fin (suc (suc n)) → Fin (suc (suc m))
-  from 0F            = 1F
-  from 1F            = 0F
-  from (suc (suc i)) = suc (suc (π ⟨$⟩ˡ i))
-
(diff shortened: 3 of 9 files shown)


src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
```diff
@@ -33,37 +37,39 @@ data Permutation {A : Set a} (R : Rel A r) : Rel (List A) (a ⊔ r) where
...
 -- Measures the number of constructors, can be useful for termination proofs
-
-steps : ∀ {R : Rel A r} {xs ys} → Permutation R xs ys → ℕ
+steps : ∀ {xs ys} → Permutation R xs ys → ℕ
 steps (refl _)            = 1
 steps (prep _ xs↭ys)      = suc (steps xs↭ys)
 steps (swap _ _ xs↭ys)    = suc (steps xs↭ys)
 steps (trans xs↭ys ys↭zs) = steps xs↭ys + steps ys↭zs
+
+toFin : ∀ {xs ys} → Permutation R xs ys → Fin.Permutation (length xs) (length ys)
```
Taneb: I don't like this name, it's not converting to a Fin but to a permutation on Fins. Maybe `onIndices` as an alternative?

src/Data/Fin/Permutation.agda
```diff
@@ -221,6 +248,35 @@ insert {m} {n} i j π = permutation to from inverseˡ′ inverseʳ′
     punchIn j (punchOut j≢k)                                               ≡⟨ punchIn-punchOut j≢k ⟩
     k                                                                      ∎
 
+------------------------------------------------------------------------
+-- Swapping
+
+-- Takes a permutation m → n and creates a permutation
+-- suc (suc m) → suc (suc n) by mapping 0 to 1 and 1 to 0 and
+-- then applying the input permutation to everything else
+swap : Permutation m n → Permutation (suc (suc m)) (suc (suc n))
```
JacquesCarette, line 263: This is a special case of \oplus of two permutations.

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
JacquesCarette: I agree that `toFin` is a bad name. `onIndices` is quite reasonable. (It's also a 'design smell', but that's another matter.)

src/Data/List/Relation/Binary/Permutation/Setoid.agda
```diff
@@ -31,7 +31,7 @@ open ≋ S using (_≋_; _∷_; ≋-refl; ≋-sym; ≋-trans)
 -- Definition, based on `Homogeneous`
 
 open Homogeneous public
-  using (refl; prep; swap; trans)
+  using (refl; prep; swap; trans; toFin)
```
JacquesCarette: why re-export `toFin` here? That doesn't seem justified.

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -100,6 +102,21 @@ AllPairs-resp-↭ sym resp (trans p₁ p₂)    pxs             =
...
+xs↭ys⇒|xs|≡|ys| (refl eq)            = Pointwise.Pointwise-length eq
+xs↭ys⇒|xs|≡|ys| (prep eq xs↭ys)      = ≡.cong suc (xs↭ys⇒|xs|≡|ys| xs↭ys)
+xs↭ys⇒|xs|≡|ys| (swap eq₁ eq₂ xs↭ys) = ≡.cong (λ x → suc (suc x)) (xs↭ys⇒|xs|≡|ys| xs↭ys)
+xs↭ys⇒|xs|≡|ys| (trans xs↭ys xs↭ys₁) = ≡.trans (xs↭ys⇒|xs|≡|ys| xs↭ys) (xs↭ys⇒|xs|≡|ys| xs↭ys₁)
+
+¬[]↭x∷xs : ∀ {x xs} → ¬ ([] ↭ x ∷ xs)
+¬[]↭x∷xs = contraposition xs↭ys⇒|xs|≡|ys| λ()
+
+¬x∷xs↭[] : ∀ {x xs} → ¬ (x ∷ xs ↭ [])
+¬x∷xs↭[] = contraposition xs↭ys⇒|xs|≡|ys| λ()
```
JacquesCarette, line 115: Faibairn threshold? Feels like we only need one direction, given that permutations are symmetric.

src/Data/List/Relation/Binary/Pointwise/Properties.agda
```diff
@@ -75,3 +77,15 @@ irrelevant : Irrelevant R → Irrelevant (Pointwise R)
...
+Pointwise-length : ∀ {xs ys} → Pointwise R xs ys → length xs ≡ length ys
+Pointwise-length []            = ≡.refl
+Pointwise-length (x∼y ∷ xs∼ys) = ≡.cong ℕ.suc (Pointwise-length xs∼ys)
+
+lookup-cast : ∀ {xs ys} →
+              Pointwise R xs ys →
+              .(∣xs∣≡∣ys∣ : length xs ≡ length ys) →
+              ∀ i →
+              R (lookup xs i) (lookup ys (cast ∣xs∣≡∣ys∣ i))
+lookup-cast (Rxy ∷ Rxsys) eq zero = Rxy
```
JacquesCarette: why name `eq` on the lhs if it is irrelevant (and not used in the rhs)?

JacquesCarette (changes requested):

src/Data/List/Relation/Binary/Permutation/Homogeneous.agda
MatthewDaggitt: Changed to `onIndices`

src/Data/List/Relation/Binary/Permutation/Setoid.agda
MatthewDaggitt: I'd argue that everything that can't be specialised should be publicly re-exported from the `Homogeneous` module. That means `steps` should be re-exported as well?

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
MatthewDaggitt: That's fair, removed a direction.

src/Data/List/Relation/Binary/Pointwise/Properties.agda
MatthewDaggitt: Removed the name.

src/Data/Fin/Permutation.agda
MatthewDaggitt: I agree (`lift0` should be as well!), but we don't have such an operation in the library. I have added a note that we should refactor it when we add that operator.

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -429,6 +443,15 @@ module _{_∙_ : Op₂ A} {ε : A}
     where open ≈-Reasoning CM.setoid
   foldr-commMonoid (trans xs↭ys ys↭zs) = CM.trans (foldr-commMonoid xs↭ys) (foldr-commMonoid ys↭zs)
 
+toFin-lookup : ∀ (xs↭ys : xs ↭ ys) →
```
Taneb: This should be renamed to match `onIndices`

CHANGELOG.md
```diff
@@ -266,14 +283,31 @@ Additions to existing modules
...
+  ```agda
+  onIndices : Permutation R xs ys → Fin.Permutation (length xs) (length ys)
+  ```
+
 * In `Data.List.Relation.Binary.Permutation.Propositional`:
   ```agda
   ↭⇒↭ₛ′ : IsEquivalence _≈_ → _↭_ ⇒ _↭ₛ′_
   ```
 
+* In `Data.List.Relation.Binary.Permutation.Propositional.Properties`:
```
Taneb: These are in `....Setoid.Properties`, right?

src/Data/List/Relation/Binary/Permutation/Setoid.agda
jamesmckinna: As you (already: discussion on #2333 if not) know: I think that, in particular, `steps` should **not** be re-exported from `Homogeneous`... and indeed that if I ever manage to finish the refactoring under #2317 that various other legacy aspects of these modules will have been removed/deprecated...

MatthewDaggitt: Yup fair.

src/Data/Fin/Permutation.agda
JacquesCarette: I have it buried in my Rig Categories repo. I need to port the basic operations on permutations over. I've got the full semiring defined there.

JacquesCarette (approved):

src/Data/Fin/Permutation.agda
jamesmckinna: But also: `Function.Related.TypeIsomorphisms` etc. as per #2489 ?

jamesmckinna:
Having looked at this, and its followup, in particular lemma [`↗↭↗⇒≤`](https://github.com/agda/agda-stdlib/commit/69b2578a091b53e9f8edcd05e972a5ba386c0b8e#diff-786f60f1efe1d992f4fcc500b1ddfdb17bf4c5bd88574074fb45117ce9b60453), do I understand correctly that the gist of argument is (Cantor-Schroeder-Bernstein redux):
* `Sorted O xs` implies that `lookup : Fin (length xs) → O.Carrier` is an order *mono*morphism, so has a monotonic inverse on its range
* for `Sorted O xs`, `Sorted O ys`,  `xs ↭ ys` via `onIndices` thus induces an order *iso*morphism between `Fin n` and itself, which is then necessarily the *identity*
* thus the `O`-monotone map from `xs` to `ys` induced via conjugation with `lookup` and its inverse, is also the identity (via faithfulness of the functor induced by `onIndices`?)
* hence `xs ≋ ys`...

If so, then I think this PR and its followup may be simplifiable via the already existing `Data.Fin.Properties.injective⇒≤` (and/or `pigeonhole`/`cantor-schroeder-bernstein`)?

UPDATED: on closer examination, maybe this is what you have!

src/Data/Fin/Base.agda
```diff
@@ -285,6 +285,11 @@ i < j = toℕ i ℕ.< toℕ j
 _>_ : IRel Fin 0ℓ
 i > j = toℕ i ℕ.> toℕ j
 
+_≰_ : ∀ {n} → Rel (Fin n) 0ℓ
+i ≰ j = ¬ (i ≤ j)
+
+_≮_ : ∀ {n} → Rel (Fin n) 0ℓ
+i ≮ j = ¬ (i < j)
```
jamesmckinna, line 292: Should we define these here, or have them created automatically by (defining and then) opening the corresponding bundle(s) in `Data.Fin.Properties`? cf. #2391 / #2490

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -100,6 +102,18 @@ AllPairs-resp-↭ sym resp (trans p₁ p₂)    pxs             =
 Unique-resp-↭ : Unique Respects _↭_
 Unique-resp-↭ = AllPairs-resp-↭ (_∘ ≈-sym) ≉-resp₂
 
+------------------------------------------------------------------------
+-- Length
+
+xs↭ys⇒|xs|≡|ys| : ∀ {xs ys} → xs ↭ ys → length xs ≡ length ys
+xs↭ys⇒|xs|≡|ys| (refl eq)            = Pointwise.Pointwise-length eq
+xs↭ys⇒|xs|≡|ys| (prep eq xs↭ys)      = ≡.cong suc (xs↭ys⇒|xs|≡|ys| xs↭ys)
+xs↭ys⇒|xs|≡|ys| (swap eq₁ eq₂ xs↭ys) = ≡.cong (λ x → suc (suc x)) (xs↭ys⇒|xs|≡|ys| xs↭ys)
```
jamesmckinna: How about
```suggestion
xs↭ys⇒|xs|≡|ys| (swap eq₁ eq₂ xs↭ys) = ≡.cong 2+_ (xs↭ys⇒|xs|≡|ys| xs↭ys)
```
(possibly at the cost of an additional import from `Data.Nat.Base`)?

jamesmckinna (commented):
Some nitpicks. 
Happy to approve, but I too have been sniped by this... would love to be able to simplify if at all possible!

src/Data/Fin/Permutation.agda
jamesmckinna: Plus, of course: we only need this operation because `Permutation.Setoid` stipulates `swap` as a constructor... ;-)

CHANGELOG.md
```diff
@@ -235,6 +235,23 @@ Additions to existing modules
   ∙-cong-∣ : x ∣ y → a ∣ b → x ∙ a ∣ y ∙ b
   ```
 
+* In `Data.Fin.Base`:
+  ```agda
+  _≰_ : ∀ {n} → Rel (Fin n) 0ℓ
+  _≮_ : ∀ {n} → Rel (Fin n) 0ℓ
```
jamesmckinna: Can we, by convention, avoid writing the `∀ {n} → ` prefix here?

src/Data/List/Relation/Binary/Pointwise.agda
```diff
@@ -130,13 +130,6 @@ AllPairs-resp-Pointwise resp@(respₗ , respᵣ) (x∼y ∷ xs) (px ∷ pxs) =
 
 ------------------------------------------------------------------------
 -- Relationship to functions over lists
-------------------------------------------------------------------------
--- length
-
-Pointwise-length : Pointwise R xs ys → length xs ≡ length ys
-Pointwise-length []            = ≡.refl
-Pointwise-length (x∼y ∷ xs∼ys) = ≡.cong ℕ.suc (Pointwise-length xs∼ys)
-
```
jamesmckinna, line 139: No deprecation warning for this?

src/Data/Fin/Permutation.agda
JacquesCarette: Yes, absolutely, there is some great stuff in `Function.Related.TypeIsomorphism`. And a rather apt 'of course'.

CHANGELOG.md
```diff
@@ -266,14 +283,31 @@ Additions to existing modules
...
 * In `Data.List.Relation.Binary.Permutation.Propositional`:
   ```agda
   ↭⇒↭ₛ′ : IsEquivalence _≈_ → _↭_ ⇒ _↭ₛ′_
   ```
 
+* In `Data.List.Relation.Binary.Permutation.Setoid.Properties`:
+  ```agda
+  xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
+  ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
+  toFin-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (toFin xs↭ys) i)
```
jamesmckinna: I think this should be:
```suggestion
  onIndices-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (onIndices xs↭ys) i)
```
and again, do we need the `∀ i → ` prefix in `CHANGELOG`?

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -17,6 +17,7 @@ module Data.List.Relation.Binary.Permutation.Setoid.Properties
 
 open import Algebra
 open import Data.Bool.Base using (true; false)
+open import Data.Fin using (zero; suc)
```
jamesmckinna:
```suggestion
open import Data.Fin.Base using (zero; suc)
```

src/Data/Fin/Permutation.agda
```diff
@@ -221,6 +248,35 @@ insert {m} {n} i j π = permutation to from inverseˡ′ inverseʳ′
...
+swap {m} {n} π = permutation to from inverseˡ′ inverseʳ′
```
jamesmckinna: Is this definition easier to describe/define indirectly, via the composition of
* `eval ((0F , 1F) ∷ (1F , 0F))` using `Data.Fin.Permutation.Transposition.List`
* `lift₀ {suc m} {suc n} (lift₀ {m} {n} π)`

or indeed (more) directly, as `transpose 0F 1F ∘ₚ lift₀ {suc m} {suc n} (lift₀ {m} {n} π)`?

CHANGELOG.md
Taneb: (explicit arguments should definitely be in the changelog - but does `i` need to be explicit?)

jamesmckinna: @Taneb  I'm surprised to learn that (is it documented?): given that `CHANGELOG` isn't machine-checked, I use it to get (a sense of) the high-level gist of what's been added, and then look at the module for the exact details of the quantification/parametrisation etc. 

As for whether `i` needs to be explicit in this lemma, I expect so, if only by analogy with (all the) lemmas about `lookup` in `Data.List.Properties`...?

src/Data/Fin/Base.agda
MatthewDaggitt: Unsure. If we do so then we should do everywhere in all numeric data modules, consistently. Feel free to open an issue?

CHANGELOG.md
MatthewDaggitt: Thanks, fixed the changelog.

src/Data/Fin/Permutation.agda
MatthewDaggitt: Yup, much nicer! Good spot!

MatthewDaggitt:
> Having looked at this, and its followup, in particular lemma [↗↭↗⇒≤](https://github.com/agda/agda-stdlib/commit/69b2578a091b53e9f8edcd05e972a5ba386c0b8e#diff-786f60f1efe1d992f4fcc500b1ddfdb17bf4c5bd88574074fb45117ce9b60453), do I understand correctly that the gist of argument is (Cantor-Schroeder-Bernstein redux):

Yes, I think so. If it can be simplified then I'm all for that :+1:

CHANGELOG.md
```diff
@@ -266,14 +283,31 @@ Additions to existing modules
...
 * In `Data.List.Relation.Binary.Permutation.Propositional`:
   ```agda
   ↭⇒↭ₛ′ : IsEquivalence _≈_ → _↭_ ⇒ _↭ₛ′_
   ```
 
+* In `Data.List.Relation.Binary.Permutation.Setoid.Properties`:
+  ```agda
+  xs↭ys⇒|xs|≡|ys| : xs ↭ ys → length xs ≡ length ys
+  ¬x∷xs↭[] : ¬ (x ∷ xs ↭ [])
+  onIndices-lookup : ∀ i → lookup xs i ≈ lookup ys (Inverse.to (onIndices xs↭ys) i)
```
jamesmckinna, line 300: Now fixed!

src/Data/List/Relation/Binary/Permutation/Setoid/Properties.agda
```diff
@@ -29,19 +30,20 @@ import Data.List.Relation.Unary.Unique.Setoid as Unique
 import Data.List.Membership.Setoid as Membership
 open import Data.List.Membership.Setoid.Properties using (∈-∃++; ∈-insert)
 import Data.List.Properties as List
-open import Data.Nat.Base using (ℕ; suc; _<_; z<s; _+_)
+open import Data.Nat.Base using (ℕ; suc; 2+; _<_; z<s; _+_)
```
jamesmckinna, line 33: I'm irritated that we didn't make this pattern synonym `2+_`... leading to the mistake in my suggestion here. But should be good now!

jamesmckinna:
I'll try to review properly this week, but for now, I've just fixed two things that slipped through the net...

JacquesCarette (approved):

Taneb (approved):

PR #2726, opened 2025-06-02
[ add ] Choudhury and Fiore's alternative definition of `Permutation` for `Setoid`s

jamesmckinna:
Alternative definition, as described in #2311, and towards open problems arising from #2723 and #2725 

NB.:
* UPDATED: I've previously misattributed this to Ralf Hinze, but credit properly belongs to Choudhury & Fiore, apparently
* module name? `ChoudhuryFiore`, like `Hinze` before, it, doesn't sit well with library style? But will change for the time being, against the possibility of eventually superseding `Permutation.Setoid` with this definition?)
* a bit rough around the edges; maybe should be `DRAFT` for the time being... or else could be fixed up in v3.0 once we agree which representations to bless for that release? UPDATED: have now done so

TODO:
* `Base`/`Properties` refactoring
* `CHANGELOG`

Regarding the `Sorted` lemma of #2723
```agda
property1 : xs ↭ ys → Sorted xs → Sorted ys → xs ≋ ys
```
suggest that this be handled in a downstream PR:
* subrelation `_∼ₛ_` in which list `cs` is `Sorted`, is equivalent to `_∼_` (given a sorting algorithm,  or at least the existence of a sorted representative of the `_∼_` equivalence class)
* `_∼ₛ_`, with some gymnastics, satisfies `property1`, but the paper proof I have needs tidying up


jamesmckinna:
Ugh. File conversion from DOS (don't ask). GitHub doesn't like it: what do I need to do to convert the file? (Emacs displays it as `DOS`, but doesn't show the `CRLF` control characters, so it looks like 'ordinary' Agda source text...

jamesmckinna:
@MatthewDaggitt 's approach to `property1` via `Fin n` as a canonical representing type for `Sorted` lists seems much better, so good not to pursue this further here!

JacquesCarette:
This is also equivalent to the more general version using `Inverse`, no?  If you're not going to pursue further, close?

jamesmckinna:
> This is also equivalent to the more general version using `Inverse`, no? If you're not going to pursue further, close?

No to closing, the comment was about pursuing a proof of `property1`. The rest I *definitely* want to pursue, because this is (IMHO: *much*) 'better' than what we already have in terms of `prep` and `swap` etc.

What is the "more general version using `Inverse`"?

JacquesCarette:
A `Permutation` is a special case (for finite types) of a type isomorphism, i.e. the case for `Fin n`. The category (Nat, Permutation) is the skeleton of (FinSet, Isomorphism). (In fact both are rig categories, and they are rig isomorphic).

jamesmckinna:
@JacquesCarette wrote:
> A `Permutation` is a special case (for finite types) of a type isomorphism, i.e. the case for `Fin n`. The category (Nat, Permutation) is the skeleton of (FinSet, Isomorphism). (In fact both are rig categories, and they are rig isomorphic).

Fantastic! Thanks so much for this (and apologies for not going back to #2311 where you do reference your own developments, and I never properly followed those up :-():
* `Rig` not `Semiring` (still sad the library blesses the latter nomenclature over the former)
* The machinery for all this is in `stdlib`, but not the actual theorem (never mind the 'skeleton of...' property: but did you mean rig-isomorphic, and not rig-equivalent? can a skeleton of a category `ℂ` ever be *isomorphic* to `ℂ` unless it's already skeletal? maybe I need to go back to CT school)

Nevertheless, all that said (and doubtless more that could be said), at least one reason I favour this ~~Hinze~~ definition over others, including anything based on coordinatisation via `Data.Fin.Permutation` would be, among other qualities:
* coordinate-free: no tabulation of the lists required (so the key `onIndices`/`lookup` machinery in @MatthewDaggitt 's lemma `↗↭↗⇒≤` might actually fall to a simpler analysis? cf. https://github.com/agda/agda-stdlib/pull/2725#issuecomment-2943100636)
* algorithmically efficient (compared to the `insert` version in #2311 ), while still being 'declarative'
* *cognitively* efficient: no 'permutations are generated by swaps' clumsiness (beyond [@onestruggler 's observation](https://github.com/agda/agda-stdlib/pull/2723#discussion_r2124672189); this is *implicit* in the definition)
* easy to prove equivalent to the full declarative version (which I should also add, either here, or as yet another definition), and hence the path to `Free CommutativeMonoid`
* ...

Now, this comment (and perhaps others above) properly belong to #2311, as 'design' rather than 'implementation', so I'll stop here.

gallais:
@jamesmckinna In that last commit I think you forgot to add the moved file so it just looks like
you nuked everything.

As for the naming debate, I would say stay away from naming it after people unless it's
already an established convention (e.g. Boolean/Heyting/Kripke). For instance we called this
module [`Data.List.Relation.Unary.Sufficient`](https://github.com/agda/agda-stdlib/blob/master/src/Data/List/Relation/Unary/Sufficient.agda#L12) rather than `Data.List.Relation.Unary.McKinna` ;)

jamesmckinna:
:facepalm:  as to deleting the addition; now restored. Thanks @gallais !
Plus: having the courage of my convictions from #2311 and calling this `Algorithmic`... will consider here, or downstream, adding `Permutation.Declarative` as well, and refactoring their properties to witness the equivalence (and probably delete the equivalence with existing `_↭_`, to avoid downstream naming/deprecation issues...)

JacquesCarette:
> but did you mean rig-isomorphic ?

I meant *naturally isomorphic as rig categories* (though I did omit 'in the presence of the axiom of choice').

JacquesCarette:
Having said that: there are strong reasons for having *many* different representations of permutations in `stdlib`. They have their (powerful, disparate) pros and cons. Where it is not always clear that different representations of `_<_` are worth it (though the cognoscenti do know), for permutations, it is a very different ball game. The equivalence of the different representations is usually seriously non-trivial too.

I'm also with @gallais on naming. Especially for permutations. While you may have learned X or Y permutation from a specific paper by specific people, usually those representations are decades if not centuries old. We need to hunt down the 'proper' names.

src/Data/List/Relation/Binary/Permutation/Algorithmic.agda
```diff
@@ -0,0 +1,150 @@
...
+    n : ℕ
+
+
+-------------------------------------------------------------------------------
+-- Definition
+
+infix  4  _∼_
+infix  5 _⋎_ _⋎[_]_
+
+data _∼_ : List A → List A → Set (s ⊔ ℓ) where
```
MatthewDaggitt: Is there a reason to choose a different symbol for the relation than we have in the existing permutation module? Note this symbol is re-used for preorders in the library and this relation is definitely not a preorder. I really think we should just use the existing symbol.

MatthewDaggitt:
@jamesmckinna is currently away and it's not like anyone is relying on this to make it into v2.3 so removing from the milestone.

src/Data/List/Relation/Binary/Permutation/Algorithmic.agda
jamesmckinna: Done!

jamesmckinna:
Happy with the milestone bump, but otherwise modules should be ready to go into `master` after the v2.3 release.

Issue #2037, opened 2023-07-31
Add surjective homomorphisms


Taneb:
In `Algebra.Morphism.Structures` and the similar modules for lattices and modules, we define `IsXHomomorphism`, `IsXMonomorphism`, and `IsXIsomorphism`. We do not currently define `IsXEpimorphism`, for morphisms that are surjective but not necessarily injective. This feels an obvious omission, and would be useful for defining, for example, projective objects (cf. [wp:Projective module](https://en.wikipedia.org/wiki/Projective_module))

jamesmckinna:
... And for proving that the unique morphism from any `Algebra` to its corresponding terminal object is `StrictlySurjective` #2296

PR #2729, opened 2025-06-07
Modular arithmetic in terms of ideals

Taneb:
Opening this PR to share my WIP. I've got a messy proof of the Chinese remainder theorem for arbitrary rings, but in porting it from my standalone library to this I've somehow made some parameters not infer properly

### DIFF
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -1,0 +1,140 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Quotient groups
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group)
+open import Algebra.NormalSubgroup using (NormalSubgroup)
+
+module Algebra.Construct.Quotient.Group  {c ℓ c′ ℓ′} (G : Group c ℓ) (N : NormalSubgroup G c′ ℓ′) where
+
+open Group G
+
+import Algebra.Definitions as AlgDefs
+open import Algebra.Morphism.Structures
+open import Algebra.Properties.Group G
+open import Algebra.Structures using (IsGroup)
+open import Data.Product.Base
+open import Level using (_⊔_)
+open import Relation.Binary.Core using (_⇒_)
+open import Relation.Binary.Definitions using (Reflexive; Symmetric; Transitive)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+open NormalSubgroup N
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → x // y ≈ ι g → x ≋ y
+
+≋-refl : Reflexive _≋_
+≋-refl {x} = N.ε by begin
+  x // x ≈⟨ inverseʳ x ⟩
+  ε      ≈⟨ ι.ε-homo ⟨
+  ι N.ε  ∎
+
+≋-sym : Symmetric _≋_
+≋-sym {x} {y} (g by x//y≈ιg) = g N.⁻¹ by begin
+  y // x      ≈⟨ ⁻¹-anti-homo-// x y ⟨
+  (x // y) ⁻¹ ≈⟨ ⁻¹-cong x//y≈ιg ⟩
+  ι g ⁻¹      ≈⟨ ι.⁻¹-homo g ⟨
+  ι (g N.⁻¹)  ∎
+
+
+≋-trans : Transitive _≋_
+≋-trans {x} {y} {z} (g by x//y≈ιg) (h by y//z≈ιh) = g N.∙ h by begin
+  x // z              ≈⟨ ∙-congʳ (identityʳ x) ⟨
+  x ∙ ε // z          ≈⟨ ∙-congʳ (∙-congˡ (inverseˡ y)) ⟨
+  x ∙ (y \\ y) // z   ≈⟨ ∙-congʳ (assoc x (y ⁻¹) y) ⟨
+  (x // y) ∙ y // z   ≈⟨ assoc (x // y) y (z ⁻¹) ⟩
+  (x // y) ∙ (y // z) ≈⟨ ∙-cong x//y≈ιg y//z≈ιh ⟩
+  ι g ∙ ι h           ≈⟨ ι.∙-homo g h ⟨
+  ι (g N.∙ h)         ∎
+
+≋-isEquivalence : IsEquivalence _≋_
+≋-isEquivalence = record
+  { refl = ≋-refl
+  ; sym = ≋-sym
+  ; trans = ≋-trans
+  }
+
+≈⇒≋ : _≈_ ⇒ _≋_
+≈⇒≋ {x} {y} x≈y = N.ε by begin
+  x // y ≈⟨ x≈y⇒x∙y⁻¹≈ε x≈y ⟩
+  ε      ≈⟨ ι.ε-homo ⟨
+  ι N.ε  ∎
+
+open AlgDefs _≋_
+
+≋-∙-cong : Congruent₂ _∙_
+≋-∙-cong {x} {y} {u} {v} (g by x//y≈ιg) (h by u//v≈ιh) = g N.∙ normal h y .proj₁ by begin
+  x ∙ u // y ∙ v              ≈⟨ ∙-congˡ (⁻¹-anti-homo-∙ y v) ⟩
+  x ∙ u ∙ (v ⁻¹ ∙ y ⁻¹)       ≈⟨ assoc (x ∙ u) (v ⁻¹) (y ⁻¹) ⟨
+  (x ∙ u // v) // y           ≈⟨ ∙-congʳ (assoc x u (v ⁻¹)) ⟩
+  x ∙ (u // v) // y           ≈⟨ ∙-congʳ (∙-congˡ u//v≈ιh) ⟩
+  x ∙ ι h // y                ≈⟨ ∙-congʳ (∙-congˡ (identityˡ (ι h))) ⟨
+  x ∙ (ε ∙ ι h) // y          ≈⟨ ∙-congʳ (∙-congˡ (∙-congʳ (inverseˡ y))) ⟨
+  x ∙ ((y \\ y) ∙ ι h) // y   ≈⟨ ∙-congʳ (∙-congˡ (assoc (y ⁻¹) y (ι h))) ⟩
+  x ∙ (y \\ y ∙ ι h) // y     ≈⟨ ∙-congʳ (assoc x (y ⁻¹) (y ∙ ι h)) ⟨
+  (x // y) ∙ (y ∙ ι h) // y   ≈⟨ assoc (x // y) (y ∙ ι h) (y ⁻¹) ⟩
+  (x // y) ∙ (y ∙ ι h // y)   ≈⟨ ∙-cong x//y≈ιg (proj₂ (normal h y)) ⟩
+  ι g ∙ ι (normal h y .proj₁) ≈⟨ ι.∙-homo g (normal h y .proj₁) ⟨
+  ι (g N.∙ normal h y .proj₁) ∎
+
+≋-⁻¹-cong : Congruent₁ _⁻¹
+≋-⁻¹-cong {x} {y} (g by x//y≈ιg) = normal (g N.⁻¹) (y ⁻¹) .proj₁ by begin
+  x ⁻¹ ∙ y ⁻¹ ⁻¹                      ≈⟨ ∙-congʳ (identityˡ (x ⁻¹)) ⟨
+  (ε ∙ x ⁻¹) ∙ y ⁻¹ ⁻¹                ≈⟨ ∙-congʳ (∙-congʳ (inverseʳ (y ⁻¹))) ⟨
+  ((y ⁻¹ ∙ y ⁻¹ ⁻¹) ∙ x ⁻¹) ∙ y ⁻¹ ⁻¹ ≈⟨ ∙-congʳ (assoc (y ⁻¹) ((y ⁻¹) ⁻¹) (x ⁻¹)) ⟩
+  y ⁻¹ ∙ (y ⁻¹ ⁻¹ ∙ x ⁻¹) ∙ y ⁻¹ ⁻¹   ≈⟨ ∙-congʳ (∙-congˡ (⁻¹-anti-homo-∙ x (y ⁻¹))) ⟨
+  y ⁻¹ ∙ (x ∙ y ⁻¹) ⁻¹ ∙ y ⁻¹ ⁻¹      ≈⟨ ∙-congʳ (∙-congˡ (⁻¹-cong x//y≈ιg)) ⟩
+  y ⁻¹ ∙ ι g ⁻¹ ∙ y ⁻¹ ⁻¹             ≈⟨ ∙-congʳ (∙-congˡ (ι.⁻¹-homo g)) ⟨
+  y ⁻¹ ∙ ι (g N.⁻¹) ∙ y ⁻¹ ⁻¹         ≈⟨ proj₂ (normal (g N.⁻¹) (y ⁻¹)) ⟩
+  ι (normal (g N.⁻¹) (y ⁻¹) .proj₁)   ∎
+
+quotientIsGroup : IsGroup _≋_ _∙_ ε _⁻¹
+quotientIsGroup = record
+  { isMonoid = record
+    { isSemigroup = record
+      { isMagma = record
+        { isEquivalence = ≋-isEquivalence
+        ; ∙-cong = ≋-∙-cong
+        }
+      ; assoc = λ x y z → ≈⇒≋ (assoc x y z)
+      }
+    ; identity = record
+      { fst = λ x → ≈⇒≋ (identityˡ x)
+      ; snd = λ x → ≈⇒≋ (identityʳ x)
+      }
+    }
+  ; inverse = record
+    { fst = λ x → ≈⇒≋ (inverseˡ x)
+    ; snd = λ x → ≈⇒≋ (inverseʳ x)
+    }
+  ; ⁻¹-cong = ≋-⁻¹-cong
+  }
+
+quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
+quotientGroup = record { isGroup = quotientIsGroup }
+
+η : Group.Carrier G → Group.Carrier quotientGroup
+η x = x -- because we do all the work in the relation
+
+η-isHomomorphism : IsGroupHomomorphism rawGroup (Group.rawGroup quotientGroup) η
+η-isHomomorphism = record
+  { isMonoidHomomorphism = record
+    { isMagmaHomomorphism = record
+      { isRelHomomorphism = record
+        { cong = ≈⇒≋
+        }
+      ; homo = λ _ _ → ≋-refl
+      }
+    ; ε-homo = ≋-refl
+    }
+  ; ⁻¹-homo = λ _ → ≋-refl
+  }
+
--- a/src/Algebra/Construct/Quotient/Ring.agda
+++ b/src/Algebra/Construct/Quotient/Ring.agda
@@ -1,0 +1,57 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Quotient rings
+------------------------------------------------------------------------
+
+open import Algebra.Bundles using (Ring)
+open import Algebra.Ideal using (Ideal)
+
+module Algebra.Construct.Quotient.Ring {c ℓ c′ ℓ′} (R : Ring c ℓ) (I : Ideal R c′ ℓ′) where
+
+open Ring R
+open Ideal I
+
+open import Algebra.Construct.Quotient.Group +-group normalSubgroup
+  public
+  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup)
+  renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
+
+open import Algebra.Definitions _≋_
+open import Algebra.Properties.Ring R
+open import Algebra.Structures
+open import Level
+open import Relation.Binary.Reasoning.Setoid setoid
+
+≋-*-cong : Congruent₂ _*_
+≋-*-cong {x} {y} {u} {v} (j by x-y≈ιj) (k by u-v≈ιk) = j I.*ᵣ u I.+ᴹ y I.*ₗ k by begin
+    x * u - y * v                       ≈⟨ +-congʳ (+-identityʳ (x * u)) ⟨
+    x * u + 0# - y * v                  ≈⟨ +-congʳ (+-congˡ (-‿inverseˡ (y * u))) ⟨
+    x * u + (- (y * u) + y * u) - y * v ≈⟨ +-congʳ (+-assoc (x * u) (- (y * u)) (y * u)) ⟨
+    ((x * u - y * u) + y * u) - y * v   ≈⟨ +-assoc (x * u - y * u) (y * u) (- (y * v)) ⟩
+    (x * u - y * u) + (y * u - y * v)   ≈⟨ +-cong ([y-z]x≈yx-zx u x y) (x[y-z]≈xy-xz y u v) ⟨
+    (x - y) * u + y * (u - v)           ≈⟨ +-cong (*-congʳ x-y≈ιj) (*-congˡ u-v≈ιk) ⟩
+    ι j * u + y * ι k                   ≈⟨ +-cong (ι.*ᵣ-homo u j) (ι.*ₗ-homo y k) ⟨
+    ι (j I.*ᵣ u) + ι (y I.*ₗ k)         ≈⟨ ι.+ᴹ-homo (j I.*ᵣ u) (y I.*ₗ k) ⟨
+    ι (j I.*ᵣ u I.+ᴹ y I.*ₗ k)          ∎
+
+quotientIsRing : IsRing _≋_ _+_ _*_ (-_) 0# 1#
+quotientIsRing = record
+  { +-isAbelianGroup = record
+    { isGroup = quotientIsGroup
+    ; comm = λ x y → ≈⇒≋ (+-comm x y)
+    }
+  ; *-cong = ≋-*-cong
+  ; *-assoc = λ x y z → ≈⇒≋ (*-assoc x y z)
+  ; *-identity = record
+    { fst = λ x → ≈⇒≋ (*-identityˡ x)
+    ; snd = λ x → ≈⇒≋ (*-identityʳ x)
+    }
+  ; distrib = record
+    { fst = λ x y z → ≈⇒≋ (distribˡ x y z)
+    ; snd = λ x y z → ≈⇒≋ (distribʳ x y z)
+    }
+  }
+
+quotientRing : Ring c (c ⊔ ℓ ⊔ c′)
+quotientRing = record { isRing = quotientIsRing }
--- a/src/Algebra/Construct/Quotient/Ring.agda
+++ b/src/Algebra/Construct/Quotient/Ring.agda
@@ -4,6 +4,8 @@
 -- Quotient rings
 ------------------------------------------------------------------------
 
+{-# OPTIONS --safe --cubical-compatible #-}
+
 open import Algebra.Bundles using (Ring)
 open import Algebra.Ideal using (Ideal)
 
--- a/src/Algebra/NormalSubgroup.agda
+++ b/src/Algebra/NormalSubgroup.agda
@@ -1,0 +1,34 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definition of normal subgroups
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group; RawGroup)
+
+module Algebra.NormalSubgroup {c ℓ} (G : Group c ℓ)  where
+
+open Group G
+
+open import Algebra.Structures using (IsGroup)
+open import Algebra.Morphism.Structures
+import Algebra.Morphism.GroupMonomorphism as GM
+open import Data.Product.Base
+open import Level using (suc; _⊔_)
+
+record NormalSubgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    -- N is a subgroup of G
+    N : RawGroup c′ ℓ′
+    ι : RawGroup.Carrier N → Carrier
+    ι-monomorphism : IsGroupMonomorphism N rawGroup ι
+    -- every element of N commutes in G
+    normal : ∀ n g → ∃[ n′ ] g ∙ ι n ∙ g ⁻¹ ≈ ι n′
+
+  module N = RawGroup N
+  module ι = IsGroupMonomorphism ι-monomorphism
+
+  N-isGroup : IsGroup N._≈_ N._∙_ N.ε N._⁻¹
+  N-isGroup = GM.isGroup ι-monomorphism isGroup
--- a/src/Algebra/Ideal.agda
+++ b/src/Algebra/Ideal.agda
@@ -1,0 +1,47 @@
+------------------------------------------------------------------------
+-- Ideals of a ring
+--
+-- The Agda standard library
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles
+
+module Algebra.Ideal {c ℓ} (R : Ring c ℓ) where
+
+open Ring R
+
+open import Algebra.Module.Bundles.Raw
+import Algebra.Module.Construct.TensorUnit as TU
+open import Algebra.Module.Morphism.Structures
+open import Algebra.NormalSubgroup (+-group)
+open import Data.Product.Base
+open import Level
+open import Relation.Binary.Reasoning.Setoid setoid
+
+record Ideal c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    I : RawModule Carrier c′ ℓ′
+    ι : RawModule.Carrierᴹ I → Carrier
+    ι-monomorphism : IsModuleMonomorphism I (TU.rawModule {R = rawRing}) ι
+
+  module I = RawModule I
+  module ι = IsModuleMonomorphism ι-monomorphism
+
+  normalSubgroup : NormalSubgroup c′ ℓ′
+  normalSubgroup = record
+    { N = I.+ᴹ-rawGroup
+    ; ι = ι
+    ; ι-monomorphism = ι.+ᴹ-isGroupMonomorphism
+    ; normal = λ n g → record
+      { fst = n
+      ; snd = begin
+        g + ι n - g     ≈⟨ +-assoc g (ι n) (- g) ⟩
+        g + (ι n - g)   ≈⟨ +-congˡ (+-comm (ι n) (- g)) ⟩
+        g + (- g + ι n) ≈⟨ +-assoc g (- g) (ι n) ⟨
+        g - g + ι n     ≈⟨ +-congʳ (-‿inverseʳ g) ⟩
+        0# + ι n        ≈⟨ +-identityˡ (ι n) ⟩
+        ι n             ∎
+      }
+    }
--- a/src/Algebra/Ideal/Construct/Intersection.agda
+++ b/src/Algebra/Ideal/Construct/Intersection.agda
@@ -1,0 +1,64 @@
+------------------------------------------------------------------------
+-- Intersection of ideals
+--
+-- The Agda standard library
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles
+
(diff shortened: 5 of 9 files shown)


JacquesCarette:
Would you want some help to get this further along?

Taneb:
Yes, actually. I've been working on a module for the special case of ideals of the ring of integers, and I've been struggling to prove that (for a non-zero modulus) it's finite, which I think it important for the "yes this is modular arithmetic as you know it" feel. I'll post a WIP commit shortly

JacquesCarette:
Ok, once my students make further progress on the ones they are currently working on, I'll get them to look at this.

jamesmckinna:
And, ... having murdered my darlings #2257 #2292 I should turn my attention to helping with this!

src/Data/Integer/ModularArithmetic.agda
```diff
@@ -0,0 +1,96 @@
...
+      -- something about x % m being smaller than m and at least 0?
+      -- so k ≡ x / m - y / m
+      -- and so x % m - y % m
+      -- ∴ x % m ≡ y % m
+
+_≋?_ : Decidable _≋_
+-- need to check whether m is 0
+_≋?_ with ℕ.nonZero? ∣ m ∣
+... | yes p = _≋?′_ {{p}}
+... | no ¬p = {!!}
```
jamesmckinna, line 84: My first thought on reading this code is that `.{{_ : NonZero m}}` should in fact also be part of the `module` parametrisation, and thus
* the anonymous `module _ .{{_ : NonZero m}} where` can be removed
* `_≋?_` just *is* `_≋?′_`

jamesmckinna: My second thought is that, for general `Ring`s:
* arithmetic `mod 0` does make sense; quotienting by the zero ideal `R/(0)` is isomorphic to `R`?
* arithmetic `mod 1` also makes sense; `R/(1)` is the trivial `Ring`

In either case, for sure the equality on the quotient would be decidable, but perhaps requires that we analyse these three cases separately:
* `m = 0`
* `m = 1`
* `NonTrivial m` (and that might involve defining that notion for `ℤ` on the model of `NonZero`?).

So the question is really: do you/we want to consider these special cases for `ℤ/mℤ`, or would it be 'better' to qualify the module with `.{{_ : NonTrivial m}}` and infer `{{NonZero m}}` where needed?

jamesmckinna:
Some errors thrown up by checking with `agda-v2.8.0` (beyond the failing tests above):
* `Data.Fin.Properties` on this branch still has the now-erroneous `--warn=noUserWarning` which should be fixed after a rebase/merge with the current `master`?
* `Data.Integer.Properties` now triggers a warning about a null `rewrite` step on L1417 (which I hadn't seen caught anywhere else? but the offending line doesn't appear any more on `master`, so...)

src/Data/Integer/ModularArithmetic.agda
Taneb: I'm not sure why we need to treat the m = 1 case specially. It's not a particularly interesting case, but it should work the same as any other nonzero modulus

jamesmckinna: Fair enough, but the path to 
```agda
_≋?_ with ℕ.nonZero? ∣ m ∣
... | yes p = _≋?′_ {{p}}
... | no ¬p = {!!}
```
is then via `_≟_ : DecidableEquality ℤ` having first shown `R/(0)` iso to `R` for any `R : Ring _ _`?
(And that `¬p` implies `m ≡ 0`...)

Taneb: Yes, that was my plan

Taneb:
@jamesmckinna I've merged in master, are those two errors fixed now?

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,34 @@
...
+open import Level using (suc; _⊔_)
+
+record NormalSubgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    -- N is a subgroup of G
+    N : RawGroup c′ ℓ′
+    ι : RawGroup.Carrier N → Carrier
+    ι-monomorphism : IsGroupMonomorphism N rawGroup ι
+    -- every element of N commutes in G
+    normal : ∀ n g → ∃[ n′ ] g ∙ ι n ∙ g ⁻¹ ≈ ι n′
```
jamesmckinna: So... I was a bit surprised that this was an 'easier' formulation than
```suggestion
    normal : ∀ n g → ∃[ n′ ] g ∙ ι n ≈ ι n′ ∙ g
```
which I wondered as to whether 
* it would be easier/smoother to show gives rise to an equivalence relation on the quotient?
* it would generalise (better) to `Loop`, `Quasigroup` or even `Magma`/`Semigroup`?
* for *commutative* operation, every subgroup is automatically `normal`...

Cf. comments elsewhere from @JacquesCarette about defining 'ideal' via 'sink'...

src/Algebra/Ideal/Construct/Principal.agda
```diff
@@ -0,0 +1,57 @@
...
+            ; ε-homo = zeroˡ a
+            }
+          ; ⁻¹-homo = λ x → sym (-‿distribˡ-* x a)
+          }
+        ; *ₗ-homo = λ r x → *-assoc r x a
+        ; *ᵣ-homo = λ r x → begin
+          x * r * a   ≈⟨ *-assoc x r a ⟩
+          x * (r * a) ≈⟨ *-congˡ (*-comm r a) ⟩
+          x * (a * r) ≈⟨ *-assoc x a r ⟨
+          x * a * r   ∎
```
jamesmckinna, line 52: This kind of argument occurs in `Algebra.Properties.CommutativeSemigroup`, and might usefully be re-used here?

src/Algebra/Ideal/Construct/Principal.agda
```diff
@@ -0,0 +1,57 @@
...
+        }
+      }
+    ; injective = λ p → p
```
jamesmckinna, line 55:
```suggestion
    ; injective = id
```
???

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,140 @@
...
+open import Relation.Binary.Definitions using (Reflexive; Symmetric; Transitive)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+open NormalSubgroup N
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → x // y ≈ ι g → x ≋ y
```
jamesmckinna: Similarly to the type of `NormalSubgroup.normal`, is it 'easier' to write
```suggestion
  _by_ : ∀ g → x ≈ ι g ∙ y → x ≋ y
```
???

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,140 @@
...
+  ι (g N.⁻¹)  ∎
+
+
+≋-trans : Transitive _≋_
+≋-trans {x} {y} {z} (g by x//y≈ιg) (h by y//z≈ιh) = g N.∙ h by begin
+  x // z              ≈⟨ ∙-congʳ (identityʳ x) ⟨
+  x ∙ ε // z          ≈⟨ ∙-congʳ (∙-congˡ (inverseˡ y)) ⟨
+  x ∙ (y \\ y) // z   ≈⟨ ∙-congʳ (assoc x (y ⁻¹) y) ⟨
+  (x // y) ∙ y // z   ≈⟨ assoc (x // y) y (z ⁻¹) ⟩
+  (x // y) ∙ (y // z) ≈⟨ ∙-cong x//y≈ιg y//z≈ιh ⟩
```
jamesmckinna: Already covered (in part) by `Algebra.Properties.Monoid`...? And if not, could/should be added?
Or: see above!

src/Data/Integer/ModularArithmetic.agda
```diff
@@ -0,0 +1,96 @@
...
+ℤ/mℤ  = quotientRing
+
+-- todo:
+-- * chinese remainder theorem specialized to integers
+-- * finiteness: for non-zero
+
+module _ .{{_ : NonZero m}} where
+
+  from-% : ∀ {x y} → x % m ≡ y % m → x ≋ y
+  from-% {x} {y} x%m≡y%m = x / m - y / m by begin
```
jamesmckinna, line 40: How much of this argument recapitulates concretely reasoning steps already used abstractly in CRT?

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: Yielding:
```agda
≋-refl : Reflexive _≋_
≋-refl {x} = N.ε by begin
  x         ≈⟨ identityˡ _ ⟨
  ε ∙ x     ≈⟨ ∙-congʳ (ι.ε-homo) ⟨
  ι N.ε ∙ x ∎

≋-sym : Symmetric _≋_
≋-sym {x} {y} (g by x≈ιg∙y) = g N.⁻¹ by begin
  y              ≈⟨ y≈x\\z _ _ _ (sym x≈ιg∙y) ⟩
  ι g ⁻¹ ∙ x     ≈⟨ ∙-congʳ (ι.⁻¹-homo g) ⟨
  ι (g N.⁻¹) ∙ x ∎

≋-trans : Transitive _≋_
≋-trans {x} {y} {z} (g by x≈ιg∙y) (h by y≈ιh∙z) = g N.∙ h by begin
  x               ≈⟨ x≈ιg∙y ⟩
  ι g ∙ y         ≈⟨ ∙-congˡ y≈ιh∙z ⟩
  ι g ∙ (ι h ∙ z) ≈⟨ assoc _ _ _ ⟨
  ι g ∙ ι h ∙ z   ≈⟨ ∙-congʳ (ι.∙-homo g h) ⟨
  ι (g N.∙ h) ∙ z ∎
```
and thus being admissible in any ~~`Quasigroup`~~ `Monoid` (an associative `Loop` is a group)? (Well, `refl` and `trans` at least...)

Indeed, these are properties (modulo `ι`) of the abstract `Divisibility` relations on `Magma` and their properties... as structure is successively enriched to `Semigroup` (for `trans`) and `Monoid` (for `refl`)! So we should add `Group` divisibility to inherit those, with `sym` becoming provable...?

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,140 @@
...
+  { refl = ≋-refl
+  ; sym = ≋-sym
+  ; trans = ≋-trans
+  }
+
+≈⇒≋ : _≈_ ⇒ _≋_
+≈⇒≋ {x} {y} x≈y = N.ε by begin
+  x // y ≈⟨ x≈y⇒x∙y⁻¹≈ε x≈y ⟩
+  ε      ≈⟨ ι.ε-homo ⟨
+  ι N.ε  ∎
```
jamesmckinna: Similarly
```agda
≈⇒≋ {x} {y} x≈y = N.ε by begin
  x         ≈⟨ x≈y ⟩
  y         ≈⟨ identityˡ _ ⟨
  ε ∙ y     ≈⟨ ∙-congʳ (ι.ε-homo) ⟨
  ι N.ε ∙ y ∎
```

jamesmckinna:
Have suggested some possible refactorings to make the constructions/lemmas more reusable, and to be able to reuse `Divisibility` for abstract algebra... or not, exactly... :-(

jamesmckinna:
So for the 'ALT' version of `NormalSubgroup`, I ended up with:
```agda
open import Algebra.Bundles using (Group; RawGroup)

module Algebra.NormalSubgroupALT {c ℓ} (G : Group c ℓ)  where

open import Algebra.Structures using (IsGroup)
open import Algebra.Morphism.Structures
import Algebra.Morphism.GroupMonomorphism as GM
open import Data.Product.Base
open import Level using (suc; _⊔_)

private
  module G = Group G
  open G using (_≈_; _∙_)


record NormalSubgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
-- firstly: N is a subgroup of G
  field
    N : RawGroup c′ ℓ′

  module N = RawGroup N

  field
    ι : N.Carrier → G.Carrier
    ι-monomorphism : IsGroupMonomorphism N G.rawGroup ι

  module ι = IsGroupMonomorphism ι-monomorphism

  isGroup : IsGroup N._≈_ N._∙_ N.ε N._⁻¹
  isGroup = GM.isGroup ι-monomorphism G.isGroup

  group : Group _ _
  group = record { isGroup = isGroup }

-- secondly: every element of N commutes in G
  field
    normal : ∀ n g → ∃[ n′ ] g ∙ ι n ≈ ι n′ ∙ g
```

jamesmckinna:
... and for `Algebra.Construct.Quotient.Group` I obtained
```agda
open import Algebra.Bundles using (Group; RawGroup)
open import Algebra.NormalSubgroupALT using (NormalSubgroup)

module Algebra.Construct.Quotient.GroupALT
  {c ℓ} (G : Group c ℓ) {c′ ℓ′} (N : NormalSubgroup G c′ ℓ′) where

open import Algebra.Morphism.Structures
open import Algebra.Structures using (IsGroup)
open import Data.Product.Base
open import Level using (_⊔_)
open import Relation.Binary.Core using (_⇒_)
open import Relation.Binary.Definitions using (Reflexive; Symmetric; Transitive)
open import Relation.Binary.Structures using (IsEquivalence)

import Algebra.Definitions as AlgDefs
import Relation.Binary.Reasoning.Setoid as ≈-Reasoning

open import Algebra.Properties.Group G

private
  module G = Group G
  open G using (_≈_; _∙_; ε;  _⁻¹)
  open import Algebra.Properties.Monoid G.monoid
  module N = NormalSubgroup N
  open N using (ι; normal; module N)
  open ≈-Reasoning G.setoid

infix 0 _by_

data _≋_ (x y : G.Carrier) : Set (c ⊔ ℓ ⊔ c′) where
  _by_ : ∀ n → x ≈ ι n ∙ y → x ≋ y

quotientRawGroup : RawGroup _ _
quotientRawGroup = record { _≈_ = _≋_ ; _∙_ = _∙_ ; ε = ε ; _⁻¹ = _⁻¹ }

≈⇒≋ : _≈_ ⇒ _≋_
≈⇒≋ {x} {y} x≈y = N.ε by begin
  x         ≈⟨ x≈y ⟩
  y         ≈⟨ G.identityˡ _ ⟨
  ε ∙ y     ≈⟨ G.∙-congʳ (ι.ε-homo) ⟨
  ι N.ε ∙ y ∎

≋-refl : Reflexive _≋_
≋-refl {x} = ≈⇒≋ G.refl

≋-sym : Symmetric _≋_
≋-sym {x} {y} (g by x≈ιg∙y) = g N.⁻¹ by begin
  y              ≈⟨ y≈x\\z _ _ _ (G.sym x≈ιg∙y) ⟩
  ι g ⁻¹ ∙ x     ≈⟨ G.∙-congʳ (ι.⁻¹-homo g) ⟨
  ι (g N.⁻¹) ∙ x ∎

≋-trans : Transitive _≋_
≋-trans {x} {y} {z} (g by x≈ιg∙y) (h by y≈ιh∙z) = g N.∙ h by begin
  x               ≈⟨ x≈ιg∙y ⟩
  ι g ∙ y         ≈⟨ G.∙-congˡ y≈ιh∙z ⟩
  ι g ∙ (ι h ∙ z) ≈⟨ G.assoc _ _ _ ⟨
  ι g ∙ ι h ∙ z   ≈⟨ G.∙-congʳ (ι.∙-homo g h) ⟨
  ι (g N.∙ h) ∙ z ∎

≋-isEquivalence : IsEquivalence _≋_
≋-isEquivalence = record
  { refl = ≋-refl
  ; sym = ≋-sym
  ; trans = ≋-trans
  }

open AlgDefs _≋_

≋-∙-cong : Congruent₂ _∙_
≋-∙-cong {x} {y} {u} {v} (g by x≈ιg∙y) (h by u≈ιh∙v) =
  let k , y∙ιh≈ιk∙y = normal h y in g N.∙ k by begin
  x ∙ u                 ≈⟨ G.∙-cong x≈ιg∙y u≈ιh∙v ⟩
  (ι g ∙ y) ∙ (ι h ∙ v) ≈⟨ uv≈w⇒xu∙vy≈x∙wy y∙ιh≈ιk∙y _ _ ⟩
  ι g ∙ ((ι k ∙ y) ∙ v) ≈⟨ G.assoc _ _ _ ⟨
  ι g ∙ (ι k ∙ y) ∙ v   ≈⟨ G.∙-congʳ (G.assoc _ _ _) ⟨
  ι g ∙ ι k ∙ y ∙ v     ≈⟨ G.assoc _ _ _ ⟩
  (ι g ∙ ι k) ∙ (y ∙ v) ≈⟨ G.∙-congʳ (ι.∙-homo g k) ⟨
  ι (g N.∙ k) ∙ (y ∙ v) ∎

≋-⁻¹-cong : Congruent₁ _⁻¹
≋-⁻¹-cong {x} {y} (g by x≈ιg∙y) =
  let h , y⁻¹∙ιg⁻¹≈ιh∙y⁻¹ = normal (g N.⁻¹) (y ⁻¹)
  in h by begin
  x ⁻¹              ≈⟨ G.⁻¹-cong x≈ιg∙y ⟩
  (ι g ∙ y) ⁻¹      ≈⟨ ⁻¹-anti-homo-∙ _ _ ⟩
  y ⁻¹ ∙ ι g ⁻¹     ≈⟨ G.∙-congˡ (ι.⁻¹-homo _) ⟨
  y ⁻¹ ∙ ι (g N.⁻¹) ≈⟨ y⁻¹∙ιg⁻¹≈ιh∙y⁻¹ ⟩
  ι h ∙ y ⁻¹        ∎

quotientIsGroup : IsGroup _≋_ _∙_ ε _⁻¹
quotientIsGroup = record
  { isMonoid = record
    { isSemigroup = record
      { isMagma = record
        { isEquivalence = ≋-isEquivalence
        ; ∙-cong = ≋-∙-cong
        }
      ; assoc = λ x y z → ≈⇒≋ (G.assoc x y z)
      }
    ; identity = record
      { fst = λ x → ≈⇒≋ (G.identityˡ x)
      ; snd = λ x → ≈⇒≋ (G.identityʳ x)
      }
    }
  ; inverse = record
    { fst = λ x → ≈⇒≋ (G.inverseˡ x)
    ; snd = λ x → ≈⇒≋ (G.inverseʳ x)
    }
  ; ⁻¹-cong = ≋-⁻¹-cong
  }

quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
quotientGroup = record { isGroup = quotientIsGroup }

module _/_ = Group quotientGroup

η : G.Carrier → _/_.Carrier
η x = x -- because we do all the work in the relation

η-isHomomorphism : IsGroupHomomorphism G.rawGroup quotientRawGroup η
η-isHomomorphism = record
  { isMonoidHomomorphism = record
    { isMagmaHomomorphism = record
      { isRelHomomorphism = record
        { cong = ≈⇒≋
        }
      ; homo = λ _ _ → ≋-refl
      }
    ; ε-homo = ≋-refl
    }
  ; ⁻¹-homo = λ _ → ≋-refl
  }
```
In each case, feel free to adapt as you see fit. (I'm almost tempted to inline `η = id` in the definition of `η-isGroupHomomorphism`, for example, and dispense with a redundant definition, which might in any case be want to be called `π` or `proj`?

Also: `IsGroupEpimorphism`? cf. your #2037 ...

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,34 @@
...
+
+  module N = RawGroup N
+  module ι = IsGroupMonomorphism ι-monomorphism
+
+  N-isGroup : IsGroup N._≈_ N._∙_ N.ε N._⁻¹
+  N-isGroup = GM.isGroup ι-monomorphism isGroup
```
jamesmckinna, line 34: Should there also be a `Group` defined?
```suggestion
  N-isGroup : IsGroup N._≈_ N._∙_ N.ε N._⁻¹
  N-isGroup = GM.isGroup ι-monomorphism isGroup
  
  N-group : Group _ _
  N-group = record { isGroup = N-isgroup }
```
plus: what should be exported `public`ly from this module?

jamesmckinna:
And for `Algebra.Construct.Quotient.Ring`, the relevant proof then changes to
```agda
≋-*-cong : Congruent₂ _*_
≋-*-cong {x} {y} {u} {v} (j by x≈ιj+y) (k by u≈ιk+v) = j I.*ᵣ u I.+ᴹ y I.*ₗ k by begin
    x * u                                ≈⟨ *-congʳ x≈ιj+y ⟩
    (ι j + y) * u                        ≈⟨ distribʳ _ _ _ ⟩
    ι j * u + y * u                      ≈⟨ +-congˡ (*-congˡ u≈ιk+v) ⟩
    ι j * u + y * (ι k + v)              ≈⟨ +-congˡ (distribˡ _ _ _) ⟩
    ι j * u + (y * ι k + y * v)          ≈⟨ +-assoc _ _ _ ⟨
    (ι j * u + y * ι k) + y * v          ≈⟨ +-congʳ (+-cong (ι.*ᵣ-homo u j) (ι.*ₗ-homo y k)) ⟨
    ι (j I.*ᵣ u) + ι (y I.*ₗ k)  + y * v ≈⟨ +-congʳ (ι.+ᴹ-homo (j I.*ᵣ u) (y I.*ₗ k)) ⟨
    ι (j I.*ᵣ u I.+ᴹ y I.*ₗ k) + y * v   ∎
```

jamesmckinna:
Suggested refactoring for `Ideal`: 
* use `comm` to show that *any* subgroup of of an `AbelianGroup` is normal
* use `+-comm` of the (module structure of a) `Ring` to obtain the `normalSubgroup` field of the `Ideal`

src/Data/Integer/ModularArithmetic.agda
jamesmckinna: Or better (?): on my revised account of equality in the quotient:
```agda
module ℤ/mℤ = Ring quotientRing

-- todo:
-- * chinese remainder theorem specialized to integers
-- * finiteness: for non-zero

module _ .{{_ : NonZero m}} where

  module _ z where
    private
      z%m = + (z % m)
      z/m = z / m
    
    z≋z%m : z ≋ z%m
    z≋z%m = z/m by begin
      z             ≡⟨ a≡a%n+[a/n]*n z m ⟩
      z%m + z/m * m ≡⟨ +-comm z%m (z/m * m) ⟩
      z/m * m + z%m ∎
      where open ≡-Reasoning

  from-% : ∀ {x y} → x % m ≡ y % m → x ≋ y
  from-% {x} {y} x%m≡y%m = begin
    x ≈⟨ z≋z%m x ⟩
    + (x % m) ≈⟨ ℤ/mℤ.reflexive (≡.cong +_ x%m≡y%m) ⟩
    + (y % m) ≈⟨ z≋z%m y ⟨
    y     ∎
    where open ≈-Reasoning ℤ/mℤ.setoid
```
This kind of argument can doubtless be generalised: `z ≋ π z` in the quotient, for any such, while `π z ≡ z % m` for the particular case `ℤ/mℤ`? UPDATED: er, no, sadly, actual reasoning is required wrt the boundedness, it seems.

src/Data/Integer/ModularArithmetic.agda
```diff
@@ -0,0 +1,96 @@
...
+    (+ (x % m) + ((x / m) * m - + (y % m))) - (y / m) * m   ≡⟨ ≡.cong (λ z → (+ (x % m) + z) - (y / m) * m) (+-comm ((x / m) * m) (- + (y % m))) ⟩
+    (+ (x % m) + (- + (y % m) + (x / m) * m)) - (y / m) * m ≡⟨ ≡.cong (λ z → z - (y / m) * m) (+-assoc (+ (x % m)) (- + (y % m)) ((x / m) * m)) ⟨
+    ((+ (x % m) - + (y % m)) + (x / m) * m) - (y / m) * m   ≡⟨ +-assoc (+ (x % m) - + (y % m)) ((x / m) * m) (- ((y / m) * m)) ⟩
+    (+ (x % m) - + (y % m)) + ((x / m) * m - (y / m) * m)   ≡⟨ ≡.cong₂ (λ a b → (+ a - + (y % m)) + ((x / m) * m + b)) x%m≡y%m (neg-distribˡ-* (y / m) m) ⟩
+    (+ (y % m) - + (y % m)) + ((x / m) * m + - (y / m) * m) ≡⟨ ≡.cong₂ _+_ (+-inverseʳ (+ (y % m))) (≡.sym (*-distribʳ-+ m (x / m) (- (y / m)))) ⟩
+    0ℤ + (x / m - y / m) * m                                ≡⟨ +-identityˡ ((x / m - y / m) * m) ⟩
+    (x / m - y / m) * m                                     ∎
+    where open ≡-Reasoning
+
+  to-% : ∀ {x y} → x ≋ y → x % m ≡ y % m
```
jamesmckinna, line 45: On whatever account of `_≋_` and `π` as the quotient map from `R` to `R / I`, this is the composition of `π-cong` (which ought generally to be provable, plus `π z ≡ z % m`? or is there more to this?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,69 @@
...
+quotientRawRing = record
+  { Carrier = Carrier
+  ; _≈_ = _≋_
+  ; _+_ = _+_
+  ; _*_ = _*_
+  ; -_ = -_
+  ; 0# = 0#
+  ; 1# = 1#
+  }
+
```
jamesmckinna, line 47: Should insert here that the quotient map on the underlying additive subgroup of the module in fact extends to a ring homomorphism from `R` to `R / I`...

... which given that the underlying map is `id` is pretty easy by hand.

src/Algebra/Construct/Quotient/Group.agda
Taneb: It's a shame that the iota means I can't reuse the divisibility machinery...

jamesmckinna: Yes, and I'm not sure we're quite ready to embrace the least common generalisation of the two... but one day perhaps!?

Taneb:
I think (while travelling) this should be generalized to ideals over modules, rather than just rings. This shouldn't need much change to the code.

Taneb:
> I think (while travelling) this should be generalized to ideals over modules, rather than just rings. This shouldn't need much change to the code.

I had the chance to give this a go, and ran into an obstacle. For `Kernel` to specialize properly to rings, we need to have defined "polymorphic" (on the underlying rings) module morphisms. This isn't too difficult, but it's tedious and requires a little thinking

src/Data/Integer/ModularArithmetic.agda
jamesmckinna: Ah... there is: a lemma more or less of the form
```agda
lemma : ∀ {p q} → (+ p) ≋ (+ q) → p ℕ.< ∣ m ∣ → q ℕ.< ∣ m ∣ → p ≡ q
```
is needed, and for that, some of the machinery in `Data.Nat.Bounded.Properties` developed in #2257 such as `≡-mod⇒%≡%` looks to be what you are after?

UPDATED: wading through the old PR, it seems the operative core properties needed from `Data.Nat.DivMod` are:
* `m<n⇒m%n≡m`
* `[m+kn]%n≡m%n`

but some wrappers around those for `Integer`s might be necessary.

jamesmckinna:
Not sure you need the generalisations at this stage?
But for `to-%` it does unfortunately seem as though some actual `DivMod`/`Bounded` arithmetic is needed ;-)

src/Data/Integer/ModularArithmetic.agda
jamesmckinna, line 84: ... with the modified definition of `_≋_`:
```suggestion
_≋?_ : Decidable _≋_
_≋?_ with ∣ m ∣ ℕ.≟ 0
... | yes |m|≡0 = λ x y → map′ ≡⇒≋ ≋⇒≡ (x ≟ y)
  where
  open ≡-Reasoning
  ≡⇒≋ : _≡_ ⇒ _≋_
  ≡⇒≋ {x = x} {y = y} x≡y = +0 by begin
    x           ≡⟨ x≡y ⟩
    y           ≡⟨ +-identityˡ y ⟨
    +0 + y      ≡⟨⟩
    +0 * +0 + y ∎
  ≋⇒≡ : _≋_ ⇒ _≡_
  ≋⇒≡ {x = x} {y = y} (k by x≡k0+y) = begin
    x           ≡⟨ x≡k0+y ⟩
    k * m + y   ≡⟨ ≡.cong (λ m → k * m + y) (∣i∣≡0⇒i≡0 |m|≡0) ⟩
    k * +0 + y  ≡⟨ ≡.cong (_+ y) (*-comm k +0) ⟩
    +0 * k + y  ≡⟨ +-identityˡ y ⟩
    y           ∎
... | no  m≢0 = _≋?′_ {{ℕ.≢-nonZero m≢0}}
```

src/Data/Integer/ModularArithmetic.agda
```diff
@@ -0,0 +1,96 @@
...
+    where
+      open ≡-Reasoning
+      lemma : x % m ⊖ y % m ≡ (k - (x / m) + (y / m)) * m
+      lemma = begin
+        x % m ⊖ y % m               ≡⟨ m-n≡m⊖n (x % m) (y % m) ⟨
+        + (x % m) - + (y % m)       ≡⟨ {!!} ⟩
+        (k - (x / m) + (y / m)) * m ∎
+
+      bound : ∣ x % m ⊖ y % m ∣ ℕ.< ∣ m ∣
+      bound = {!!}
```
jamesmckinna, line 56:
```suggestion  to-% : ∀ {x y} → x ≋ y → x % m ≡ y % m
  to-% {x} {y} x≋y
    using x%m ← x % m
    using y%m ← y % m
    with k by x%m≡km+y%m ← (let open ≈-Reasoning ℤ/mℤ.setoid in begin
      + x%m ≈⟨ z≋z%m x ⟨
      x ≈⟨ x≋y ⟩
      y ≈⟨ z≋z%m y ⟩
      + y%m ∎)
    = begin
      x%m                   ≡⟨ ℕ.m<n⇒m%n≡m (n%d<d x m) ⟨
      x%m ℕ.% ∣ m ∣         ≡⟨⟩
      (+ x%m) % m           ≡⟨ ≡.cong (_% m) x%m≡km+y%m ⟩
      (k * m + (+ y%m)) % m ≡⟨ [km+n]%m≡n%m k (+ y%m) ⟩
      (+ y%m) % m           ≡⟨⟩
      y%m ℕ.% ∣ m ∣         ≡⟨ ℕ.m<n⇒m%n≡m (n%d<d y m) ⟩
      y%m ∎
    where
    open ≡-Reasoning
    [km+n]%m≡n%m : ∀ k n → (k * m + n) % m ≡ n % m
    [km+n]%m≡n%m k n = {!!}
```
where this last lemma is the 'obvious' extension of `Data.Nat.DivMod.[m+kn]%n≡m%n` to `Integer`... but is still ... irritating to have to prove directly from first principles.

jamesmckinna:
Yet another alternative definition of
```agda
data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
```
namely, to symmetrise wrt multiplication by elements of the normal subgroup, ie
```agda
record _≋_ (x y : G.Carrier) : Set (c ⊔ ℓ ⊔ c′) where
  field
    {l} {r} : _
    [ιl]x≈[ιr]y : ι l ∙ x G.≈ ι r ∙ y
```
on the basis that this mirrors the revised ('symmetrised') definition of `normal`...?
(never mind the relationship with `Rational`, and the 'INT construction' more generally)

Transitivity is a bit harder to prove, but I wonder if this might overall enjoy a smoother development?

Or else refactor as a lemma/view expressing the decomposition?

src/Algebra/NormalSubgroup.agda
Taneb: I wrote the definition I did because I went to the Wikipedia page for "normal subgroup" and translated the definition in the opening paragraph into Agda. 
Your definition does seem to be equivalent, and somehow nicer. I'm a little worried that it's somehow "detached" from traditional definitions, but I don't think it's far removed enough for that to be a blocker

src/Algebra/Ideal/Construct/Principal.agda
Taneb: Yes indeed - the lines you highlighted are exactly `xy∙z≈x∙zy` from that module, specialized for multiplication. See my comment here: https://github.com/agda/agda-stdlib/pull/2765#issuecomment-3065984036 but this doesn't need to wait for that to be dealt with

src/Algebra/Construct/Quotient/Group.agda
Taneb: Code duplication with divisibility shouldn't be a blocker - I think your suggestion is a good one and I'll modify the code accordingly.

src/Algebra/NormalSubgroup.agda
jamesmckinna: Elsewhere I think that @JacquesCarette has commented on not (necessarily) letting 'traditional' definitions be our guide, with his citation of 'Post-Modern Algebra' as a rational reconstruction which perhaps avoids the... drawbacks... of 'tradition'.

src/Algebra/Ideal/Construct/Principal.agda
jamesmckinna: (And yes: improving `Algebra.Properties.Ring` and its ancestors should also get done! cf. #2804 )

src/Data/Integer/ModularArithmetic.agda
jamesmckinna: ... where the argument for the case `m≡0` could/should be generalised for arbitrary `Ring`, with the machinery `(∣i∣≡0⇒i≡0 |m|≡0)` being all that's then needed to aplly it for the `Integer` case...

src/Data/Integer/ModularArithmetic.agda
jamesmckinna: I'd hoped I could simplify this further, but ran out of energy! Hopefully your holiday has re-energised you!

jamesmckinna:
> I think (while travelling) this should be generalized to ideals over modules, rather than just rings. This shouldn't need much change to the code.

This was one of Nagata's original motivation (see 'Local Rings' for details) for introducing the 'ring of dual numbers' construction now in `Algebra.Module.Construct.Idealisation`:
* the lattice of ideals of the ring R ⋉ M is in 1-1 correspondence with that of sub modules of an R-module M
* so the generalisation makes sense, but maybe is worthwhile as a downstream PR? What you have seem pretty close now?

src/Algebra/NormalSubgroup.agda
JacquesCarette: And there's a body of precedent being built where even in Lean, which is unabashedly classical, has often chosen non-traditional definitions because they turned out to *still* be too hard to work with in practice. 'Tradition' is for paper-math; fully formal math needs to carve its own path because it aims higher.

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: Also... it occurs to me somewhat late in the day, for all these definitions of (equivalence relations defining the) quotient structure... which is preferable (if, indeed, such a question has a clear answer/rationale for it):
* one-constructor `data` definition?
* or, `record` type definition?

Taneb: If there is a rationale for that class of decisions I'd love to learn it

jamesmckinna: Hmmm... let's see if I can at least rationalise it *post hoc*, although given the number of people who make/have made such suggestions to me at AIM, perhaps there are *real* reasons:
pro
* 'eta for free' so no (real) overhead to pattern matching on the record (debatable?); but maybe this is also a 'contra'?
* projections/copatterns are 'better' (are they?) if you don't need access to all the substructure of a given expression (is this correct?)
* informal subtyping via the signature of the associated `module`

contra
* the main one I see is having, in this example, to give `n` a name *as a `field`*, whereas in the `data` version, it's 'just an argument in a telescope' defining a function, so 'less' namespace pollution (but that can be controlled by how much you expose `field` names by `open`ing

Er... that's all I've got? Anyone else care to offer a suggestion?

JacquesCarette: There are also differences in how goals are displayed, Agda's ability to infer things, and how much evaluation happens, at least between direct / one-field-record / one-item-data.

jamesmckinna: Well, there is also the problem/issue: `record`s can't be used if you have a GADT-/inductive-family-style definition, or rather they can, at the cost of a suitable additional `field` defining the 'equation' (usually in terms of `_≡_`) which holds between the (arbitrary) parameters to the `record`, and the (possibly) green slime occurring in the conclusion of the constructor.

What's interesting about these definitions is that the equations defining the concepts in the `data` versions *are* the specialisations of such 'green slime', but to the underlying `Setoid` equality, rather than `_≡_`... so would be some more work to do to ensure that any inductive family-style definition enjoyed suitable closure/'respects' properties wrt that equality, on the model of `subst` ... so it might end up being more painful in the end? UPDATED: or indeed, impossible. How do we show that the `_by_` decomposition is *unique*? If we can't, then an inductive family version isn't going to fly, I think :-(

We're not especially good in `stdlib` at doing 'Red team'/'Blue team' adversarial design in order to determine which choice is better, but I wonder sometimes (if we had a larger developer pool!) if it might be worthwhile? And indeed, if my reviewing style is evolving in such a direction... ;-)

JacquesCarette: Agreed: I'd say we're still figuring out how to do modeling in dependent type theory (i.e. the whole community, not just `stdlib` developers).

jamesmckinna:
I've `resolved` some of the above discussions, but others, I think, remain useful to leave open while you tidy up this draft.
But I'm very happy with the conclusion of our f2f discussion this morning, in the form of refactoring this into smaller PRs which build the various pieces of infrastructure before going all out for the 'main' theorem downstream...

PR #2730, opened 2025-06-08
Some lemmata for lists and non-empty lists

pmbittner:
These are some lemmata we needed in our project [Vatras](https://github.com/VariantSync/Vatras) for reasoning on non-empty lists. It seemed these lemmata could be generally useful to other people as well.

-- _PR is part of Zurihac 2025_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -288,6 +288,7 @@ Additions to existing modules
   ∷⁺→∷ : (x List⁺.∷ xs) ≡ (y List⁺.∷ ys) →
          (x List.∷ xs) ≡ (y List.∷ ys)
   length-⁺++⁺ : (xs ys : List⁺ A) → length (xs ⁺++⁺ ys) ≡ length xs + length ys
+  length-⁺++⁺-comm : ∀ (xs ys : List⁺ A) → length (xs ⁺++⁺ ys) ≡ length (ys ⁺++⁺ xs)
   length-⁺++⁺-≤ˡ : (xs ys : List⁺ A) → length xs ≤ length (xs ⁺++⁺ ys)
   length-⁺++⁺-≤ʳ : (xs ys : List⁺ A) → length ys ≤ length (xs ⁺++⁺ ys)
   map-⁺++⁺ : ∀ (f : A → B) xs ys → map f (xs ⁺++⁺ ys) ≡ map f xs ⁺++⁺ map f ys


Taneb:
These look like good additions! Can you update `Changelog.md` with these?

pmbittner:
Ok, I updated the changelog as well. :)

pmbittner:
I also migrated some additional lemmas from `Data.List.Properties`.

CHANGELOG.md
```diff
@@ -276,6 +276,27 @@ Additions to existing modules
   filter-↭ : ∀ (P? : Pred.Decidable P) → xs ↭ ys → filter P? xs ↭ filter P? ys
   ```
 
+* In `Data.List.NonEmpty.Properties`:
+  ```agda
+  ∷→∷⁺ : ∀ {x y : A} {xs ys : List A} →
```
JacquesCarette: We don't tend to put the implicits in the CHANGELOG, (We don't tend to use qualified names either but, in this case, it's probably warranted).

CHANGELOG.md
```diff
@@ -276,6 +276,27 @@ Additions to existing modules
...
+         (x List.∷ xs) ≡ (y List.∷ ys) →
+         (x List⁺.∷ xs) ≡ (y List⁺.∷ ys)
+  ∷⁺→∷ : ∀ {x y : A} {xs ys : List A} →
+         (x List⁺.∷ xs) ≡ (y List⁺.∷ ys) →
+         (x List.∷ xs) ≡ (y List.∷ ys)
+  length-⁺++⁺ : (xs ys : List⁺ A) →
```
JacquesCarette: The next 3 should all be on a single line (each).

JacquesCarette (changes requested):
A few nits, but otherwise this is looking nice.

[Not a big fan of `rewrite`, but it's not a showstopper.]

pmbittner:
Thank you for the review. :) I integrated the requested changes. I also got rid of the `rewrite` by introducing an analogous lemma to `Data.List.Properties` and then reusing that.

JacquesCarette (approved):

CHANGELOG.md
```diff
@@ -260,6 +260,7 @@ Additions to existing modules
 
 * In `Data.List.Properties`:
   ```agda
+  length-++-≤ : ∀ (xs : List A) → length xs ≤ length (xs ++ ys)
```
gallais: Feels like this should have a `ˡ` suffix and we should have the `ʳ` variant too.

CHANGELOG.md
```diff
@@ -276,6 +277,22 @@ Additions to existing modules
...
   ```
 
+* In `Data.List.NonEmpty.Properties`:
+  ```agda
+  ∷→∷⁺ : (x List.∷ xs) ≡ (y List.∷ ys) →
+         (x List⁺.∷ xs) ≡ (y List⁺.∷ ys)
+  ∷⁺→∷ : (x List⁺.∷ xs) ≡ (y List⁺.∷ ys) →
+         (x List.∷ xs) ≡ (y List.∷ ys)
+  length-⁺++⁺ : (xs ys : List⁺ A) → length (xs ⁺++⁺ ys) ≡ length xs + length ys
+  length-⁺++⁺-≤ : (xs ys : List⁺ A) → length xs ≤ length (xs ⁺++⁺ ys)
```
gallais: Same remark about suffixes.

gallais (commented):
Happy with this; apart from two small naming issues (and related completeness question).

src/Data/List/NonEmpty/Properties.agda
```diff
@@ -69,6 +72,47 @@ toList->>= f (x ∷ xs) = begin
...
+
+-- ... and vice versa
+∷⁺→∷ : ∀ {x y : A} {xs ys : List A} →
+      (x List⁺.∷ xs) ≡ (y List⁺.∷ ys) →
+      (x List.∷ xs) ≡ (y List.∷ ys)
+∷⁺→∷ refl = refl
+
+------------------------------------------------------------------------
+-- _⁺++⁺_
+length-⁺++⁺ : (xs ys : List⁺ A) →
```
MatthewDaggitt, line 90: Very minor but can we have an empty line between this lemm and the section heading?

MatthewDaggitt (approved):

pmbittner:
Hi everyone,

I implemented the requested changes. Implementing the `length-++-≤ʳ` theorem required some additional lemmata, which unfortunately required another theorem (`++-identityʳ`) which was located below in that file. So I had to reorder some theorems, including module definitions. Please check whether my changes remain idiomatic to the standard library.

(Maybe there is a simpler proof for `length-++-≤ʳ` but I did not find it yet.)

src/Data/List/Properties.agda
```diff
@@ -190,6 +189,47 @@ module _ {A : Set a} where
...
+length-++-sucˡ : ∀ (x : A) (xs ys : List A) →
+                 length (x ∷ xs ++ ys) ≡ suc (length (xs ++ ys))
+length-++-sucˡ _ _ _ = refl
+
+length-++-sucʳ : ∀ (xs : List A) (y : A) (ys : List A) →
+                 length (xs ++ y ∷ ys) ≡ suc (length (xs ++ ys))
+length-++-sucʳ []       _ _  = refl
+length-++-sucʳ (_ ∷ xs) y ys = cong suc (length-++-sucʳ xs y ys)
+
+length-++-comm : ∀ (xs ys : List A) →
```
JacquesCarette, line 201: I wonder if this would be easier to prove as a corollary of length being preserved by permutations?

pmbittner: If I understand correctly, this means to

1. Establish a theorem that says that `length` is preserved under permutation.
2. Prove that `xs ++ ys` is a permutation of `ys ++ xs`.
3. Prove `length-++-comm` as a corollary.

This should be possible and the theorems 1 and 2 should be useful as well. I am not sure this would be a simpler as a sole means to prove `length-++-comm` though.

Given that there are multiple definitions of permutations, which one should be used here? `Data.List.Relation.Binary.Permutation.Setoid` or `Data.List.Relation.Binary.Permutation.Propositional`?

JacquesCarette: I know step 1 already exists, I've used it recently. I'm almost sure step 2 does as well. 

I'd use `Propositional` as it is more precise in what it asserts.

Taneb: I'd caution against doing this for dependency tree reasons...

pmbittner: So how should we proceed with this PR?

JacquesCarette: (Note that I approved the changes, my suggestion above was for a potential future, not this. Now it's up to someone else to also approve the latest changes. Maybe @MatthewDaggitt can re-review?)

MatthewDaggitt:
Thanks @pmbittner for the great contribution and apologies for the delay. Merging them in now.

PR #2733, opened 2025-06-09
Refactor `Data.Vec.Properties` : Add toList-injective and new lemmas 

jmougeot:
- Replace `Data.List.Base as List` and `Data.List.Properties as List` with shorter `L` aliases throughout the module for better readable proof 

- Add new utility lemma `toList-injective` that proves vector equality from list equality using heterogeneous equality

- Add new equation-free lemmas that leverage `toList-injective` for cleaner proofs:
  - `fromList-reverse`: proves `fromList (L.reverse xs) ≈[ L.length-reverse xs ] reverse (fromList xs)`
  - `++-ʳ++-eqFree`: proves associativity-like property for reverse-append without explicit equation parameter
  - `ʳ++-ʳ++-eqFree`: proves composition property for reverse-append operations

- Update all `List.` references to `L.` in function types and implementations

- Import `Tactic.Cong` module to support the `cong!` tactic used in several proofs

This change improves code readability and adds a fundamental lemma for  reasoning about vector equality via list conversion, which is used in several proofs throughout the module.

### DIFF
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -1361,20 +1361,30 @@ fromList-reverse xs =
     toList (reverse (fromList xs))      ∎
     where open ≡-Reasoning
 
+
 ------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 2.3
+
+∷-ʳ++-eqFree : ∀ a (xs : Vec A m) {ys : Vec A n} → let eq = sym (+-suc m n) in
+               cast eq ((a ∷ xs) ʳ++ ys) ≡ xs ʳ++ (a ∷ ys)
+∷-ʳ++-eqFree a xs {ys} = ʳ++-ʳ++-eqFree (a ∷ []) {ys = xs} {zs = ys}
+{-# WARNING_ON_USAGE ∷-ʳ++-eqFree
+"Warning: ∷-ʳ++-eqFree was deprecated in v2.3.
+Please use ʳ++-ʳ++-eqFree instead, which does not take eq."
+#-}
+
+-- Version 2.2
+
 -- TRANSITION TO EQ-FREE LEMMA
-------------------------------------------------------------------------
+--
 -- Please use the new proofs, which do not require an `eq` parameter.
 -- In v3, `name` will be changed to be the same lemma as `name-eqFree`,
 -- and `name-eqFree` will be deprecated.
-
-++-assoc : ∀ .(eq : (m + n) + o ≡ m + (n + o)) (xs : Vec A m) (ys : Vec A n) (zs : Vec A o) →
-           cast eq ((xs ++ ys) ++ zs) ≡ xs ++ (ys ++ zs)
-++-assoc _ = ++-assoc-eqFree
-{-# WARNING_ON_USAGE ++-assoc
-"Warning: ++-assoc was deprecated in v2.2.
-Please use ++-assoc-eqFree instead, which does not take eq."
-#-}
 
 ++-identityʳ : ∀ .(eq : n + zero ≡ n) (xs : Vec A n) → cast eq (xs ++ []) ≡ xs
 ++-identityʳ _ = ++-identityʳ-eqFree
@@ -1423,15 +1433,6 @@ Please use reverse-++-eqFree instead, which does not take eq."
 Please use ∷-ʳ++-eqFree instead, which does not take eq."
 #-}
 
-∷-ʳ++-eqFree : ∀ a (xs : Vec A m) {ys : Vec A n} → let eq = sym (+-suc m n) in
-               cast eq ((a ∷ xs) ʳ++ ys) ≡ xs ʳ++ (a ∷ ys)
-∷-ʳ++-eqFree a xs {ys} = ʳ++-ʳ++-eqFree (a ∷ []) {ys = xs} {zs = ys}
-
-{-# WARNING_ON_USAGE ∷-ʳ++-eqFree
-"Warning: ∷-ʳ++-eqFree was deprecated in v2.2.
-Please use ʳ++-ʳ++-eqFree instead, which does not take eq."
-#-}
-
 ++-ʳ++ : ∀ .(eq : m + n + o ≡ n + (m + o)) (xs : Vec A m) {ys : Vec A n} {zs : Vec A o} →
          cast eq ((xs ++ ys) ʳ++ zs) ≡ ys ʳ++ (xs ʳ++ zs)
 ++-ʳ++ _ = ++-ʳ++-eqFree
@@ -1448,11 +1449,13 @@ Please use ++-ʳ++-eqFree instead, which does not take eq."
 Please use ʳ++-ʳ++-eqFree instead, which does not take eq."
 #-}
 
-------------------------------------------------------------------------
--- DEPRECATED NAMES
-------------------------------------------------------------------------
--- Please use the new names as continuing support for the old names is
--- not guaranteed.
+++-assoc : ∀ .(eq : (m + n) + o ≡ m + (n + o)) (xs : Vec A m) (ys : Vec A n) (zs : Vec A o) →
+           cast eq ((xs ++ ys) ++ zs) ≡ xs ++ (ys ++ zs)
+++-assoc _ = ++-assoc-eqFree
+{-# WARNING_ON_USAGE ++-assoc
+"Warning: ++-assoc was deprecated in v2.2.
+Please use ++-assoc-eqFree instead, which does not take eq."
+#-}
 
 -- Version 2.0
 
--- a/src/Data/Vec/Properties.agda
+++ b/src/Data/Vec/Properties.agda
@@ -54,7 +54,7 @@ private
     ws xs ys zs : Vec A n
 
 ------------------------------------------------------------------------
--- Properties to List 
+-- Properties to List
 
 toList-injective
   : ∀ {m n}
@@ -63,7 +63,7 @@ toList-injective
   → toList xs ≡ toList ys
   → xs ≈[ m=n ] ys
 toList-injective m=n [] [] xs=ys = refl
-toList-injective m=n (x ∷ xs) (y ∷ ys) xs=ys = 
+toList-injective m=n (x ∷ xs) (y ∷ ys) xs=ys =
   cong₂ _∷_ (L.∷-injectiveˡ xs=ys) (toList-injective (cong pred m=n) xs ys (L.∷-injectiveʳ xs=ys))
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


JacquesCarette:
> Replace Data.List.Base as List and Data.List.Properties as List with shorter L aliases throughout the module for better readable proof

While I happen to like that, this goes specifically against the stdlib [naming convention](https://github.com/agda/agda-stdlib/blob/master/doc/style-guide.md) which states that it should be `List`.

> import Tactic.Cong module to support the cong! tactic used in several proofs

It has been an (unwritten!) convention to not use tactics in fairly low-level parts of the library, as we don't want those parts of the library to depend on the `Reflection` machinery. The dependency graph is nasty enough as it is, we don't really want to make it worse.

src/Data/Vec/Properties.agda
```diff
@@ -1025,17 +1039,32 @@ map-reverse f (x ∷ xs) = begin
...
-  reverse (xs ++ ys) ∷ʳ x             ≈⟨ ≈-cong′ (_∷ʳ x) (reverse-++-eqFree xs ys) ⟩
-  (reverse ys ++ reverse xs) ∷ʳ x     ≈⟨ ++-∷ʳ-eqFree x (reverse ys) (reverse xs) ⟩
-  reverse ys ++ (reverse xs ∷ʳ x)     ≂⟨ cong (reverse ys ++_) (reverse-∷ x xs) ⟨
-  reverse ys ++ (reverse (x ∷ xs))    ∎
-  where open CastReasoning
+reverse-++-eqFree : ∀ (xs : Vec A m) (ys : Vec A n)
+                  → reverse (xs ++ ys) ≈[ +-comm m n ] reverse ys ++ reverse xs
+reverse-++-eqFree {m = m} {n = n} xs ys =
+  toList-injective (+-comm m n) (reverse (xs ++ ys)) (reverse ys ++ reverse xs) $
+  begin
```
MatthewDaggitt, line 1055: Is this really a cleaner proof? To my eyes it looks longer and more complicated?

src/Data/Vec/Properties.agda
```diff
@@ -1318,17 +1353,14 @@ fromList-++ : ∀ (xs : List A) {ys : List A} →
...
-  fromList (List.reverse xs) ++ [ x ]           ≈⟨ unfold-∷ʳ-eqFree x (fromList (List.reverse xs)) ⟨
-  fromList (List.reverse xs) ∷ʳ x               ≈⟨ ≈-cong′ (_∷ʳ x) (fromList-reverse xs) ⟩
-  reverse (fromList xs) ∷ʳ x                    ≂⟨ reverse-∷ x (fromList xs) ⟨
-  reverse (x ∷ fromList xs)                     ≈⟨⟩
-  reverse (fromList (x List.∷ xs))              ∎
-  where open CastReasoning
+fromList-reverse : (xs : List A) → (fromList (List.reverse xs)) ≈[ List.length-reverse xs ] reverse (fromList xs)
+fromList-reverse xs = toList-injective (List.length-reverse xs) (fromList (List.reverse xs)) (reverse (fromList xs)) $
+  begin
+    toList (fromList (List.reverse xs)) ≡⟨ toList∘fromList (List.reverse xs) ⟩
```
MatthewDaggitt, line 1358: Can we align the reasoning combinators?

src/Data/Vec/Properties.agda
```diff
@@ -1304,8 +1339,8 @@ cast-fromList {xs = List.[]}     {ys = List.[]}     eq = refl
 cast-fromList {xs = x List.∷ xs} {ys = y List.∷ ys} eq =
   let x≡y , xs≡ys = List.∷-injective eq in begin
   x ∷ cast (cong (pred ∘ List.length) eq) (fromList xs) ≡⟨ cong (_ ∷_) (cast-fromList xs≡ys) ⟩
-  x ∷ fromList ys                                       ≡⟨ cong (_∷ _) x≡y ⟩
-  y ∷ fromList ys                                       ∎
+  x ∷ fromList ys                                    ≡⟨ cong (_∷ _) x≡y ⟩
```
MatthewDaggitt: The only change here seems to be to misalign the combinators?

src/Data/Vec/Properties.agda
```diff
@@ -52,6 +53,19 @@ private
     m n o : ℕ
     ws xs ys zs : Vec A n
 
+------------------------------------------------------------------------
+-- Properties to List
```
MatthewDaggitt: This should be `Properties of toList`

src/Data/Vec/Properties.agda
```diff
@@ -52,6 +53,19 @@ private
...
+
+toList-injective
+  : ∀ {m n}
+  → .(m=n : m ≡ n)
```
MatthewDaggitt: Can we follow the style guide about the placement of arrows on new lines?

Also can the name of this proof use `≡` instead of `=`?

src/Data/Vec/Properties.agda
```diff
@@ -52,6 +53,19 @@ private
     m n o : ℕ
     ws xs ys zs : Vec A n
 
+------------------------------------------------------------------------
+-- Properties to List
+
+toList-injective
+  : ∀ {m n}
```
MatthewDaggitt: This are variables and therefore not needed.

src/Data/Vec/Properties.agda
```diff
@@ -1025,17 +1039,32 @@ map-reverse f (x ∷ xs) = begin
...
 
 -- append and reverse
+toList-∷ʳ : ∀ x (xs : Vec A n) → toList (xs ∷ʳ x) ≡ toList xs List.++ List.[ x ]
+toList-∷ʳ x []       = refl
+toList-∷ʳ x (y ∷ xs) = cong (y List.∷_) (toList-∷ʳ x xs)
+
+toList-reverse : ∀ (xs : Vec A n) → toList (reverse xs) ≡ List.reverse (toList xs)
+toList-reverse [] = refl
+toList-reverse (x ∷ xs) = begin
+  toList (reverse (x ∷ xs))          ≡⟨ cong toList (reverse-∷ x xs) ⟩
```
MatthewDaggitt: Can we align the reasoning combinators? (you may need to install a monospaced font to make this easier)

src/Data/Vec/Properties.agda
```diff
@@ -1061,53 +1090,55 @@ foldr-ʳ++ B f {e} xs = foldl-fusion (foldr B f e) refl (λ _ _ → refl) xs
...
-∷-ʳ++-eqFree a xs {ys} = begin
-  (a ∷ xs) ʳ++ ys         ≂⟨ unfold-ʳ++ (a ∷ xs) ys ⟩
-  reverse (a ∷ xs) ++ ys  ≂⟨ cong (_++ ys) (reverse-∷ a xs) ⟩
-  (reverse xs ∷ʳ a) ++ ys ≈⟨ ∷ʳ-++-eqFree a (reverse xs) ⟩
-  reverse xs ++ (a ∷ ys)  ≂⟨ unfold-ʳ++ xs (a ∷ ys) ⟨
-  xs ʳ++ (a ∷ ys)         ∎
-  where open CastReasoning
+toList-ʳ++ : ∀ (xs : Vec A m) {ys : Vec A n} → toList (xs ʳ++ ys) ≡ (toList xs) List.ʳ++ toList ys
+toList-ʳ++ xs {ys} = begin
+  toList (xs ʳ++ ys)                    ≡⟨ cong toList (unfold-ʳ++ xs ys) ⟩
```
MatthewDaggitt: Can we align the reasoning combinators? (for all new proofs in this file)

src/Data/Vec/Properties.agda
```diff
@@ -1061,53 +1090,55 @@ foldr-ʳ++ B f {e} xs = foldl-fusion (foldr B f e) refl (λ _ _ → refl) xs
...
+  map f (xs ʳ++ ys)              ≡⟨ cong (map f) (unfold-ʳ++ xs ys) ⟩
+  map f (reverse xs ++ ys)       ≡⟨ map-++ f (reverse xs) ys ⟩
+  map f (reverse xs) ++ map f ys ≡⟨ cong (_++ map f ys) (map-reverse f xs) ⟩
   reverse (map f xs) ++ map f ys ≡⟨ unfold-ʳ++ (map f xs) (map f ys) ⟨
   map f xs ʳ++ map f ys          ∎
   where open ≡-Reasoning
 
-∷-ʳ++-eqFree : ∀ a (xs : Vec A m) {ys : Vec A n} → let eq = sym (+-suc m n) in
-               cast eq ((a ∷ xs) ʳ++ ys) ≡ xs ʳ++ (a ∷ ys)
-∷-ʳ++-eqFree a xs {ys} = begin
```
MatthewDaggitt, line 1073: This lemma seems to have been deleted without replacement?

MatthewDaggitt (changes requested):
Generally the lemmas look good, but I agree with @JacquesCarette point about the use of tactics.

src/Data/Vec/Properties.agda
jmougeot: From my perspective, the advantages of this version are:
	•	It uses equational reasoning instead of induction
	•	It relies on existing properties from the List module
	•	It avoids the use of brackets from cast reasoning, which I personally find harder to follow
But if it's not what we want to take,  please let me know!

src/Data/Vec/Properties.agda
jmougeot: This lemma was previously used in the proof of` ʳ++-ʳ++-eqFree`, but since we now use the corresponding property directly from the List module, this lemma has become redundant `ʳ++-ʳ++-eqFree` is strictly stronger: 

```
∷-ʳ++-eqFree a xs {ys}  = ʳ++-ʳ++-eqFree (a ∷ []) {ys = xs} {zs = ys}

```
Should we keep it anyway ?

MatthewDaggitt: Yes, in general we can't delete code from the library without deprecating it as other users may depend on it downstream! Please could you put it back?

src/Data/Vec/Properties.agda
MatthewDaggitt: Okay, those are good reasons.

MatthewDaggitt:
Otherwise once the old lemma is added back, then it looks good to go!

src/Data/Vec/Properties.agda
jmougeot: I added it back as you asked, but shouldn’t I put it with the deprecated functions?

MatthewDaggitt: If the old version is strictly stronger, then yes we can deprecate it. Please remember to add an entry in the CHANGELOG!

src/Data/Vec/Properties.agda
```diff
@@ -38,6 +38,7 @@ open import Relation.Unary using (Pred; Decidable)
 open import Relation.Nullary.Decidable.Core
   using (Dec; does; yes; _×-dec_; map′)
 open import Relation.Nullary.Negation.Core using (contradiction)
+open import Tactic.Cong using (cong!)
```
JacquesCarette: Could you remove this please.

JacquesCarette (changes requested):
Other than the missing `CHANGELOG` bit that @MatthewDaggitt asked for, and the `cong!`, this looks good to me too.

MatthewDaggitt (approved):

JacquesCarette (approved):

PR #2736, opened 2025-06-19
Some extra support for homogeneous n-ary products

MatthewDaggitt:
Lately my students and I have been writing a lot of code with N-ary products where all elements either belong to types with the same universe level or belong to the same types. This adds some extra infrastructure to make this all a little bit nicer.

If `⨆` (which acts as a fold) took a level for the `n = zero` case instead of defaulting to `Level.zero` then we could prove that `HomoProduct n A : Set a` instead of `HomoProduct n A : Set (lconst n a)`. However, a) that is a breaking change and b) there's no nice way that I can see to make that cast so that the other n-ary functions over n-ary products can see through the cast. Therefore leaving as is in order to make incremental progress

### DIFF
--- a/src/Data/Product/Nary/NonDependent.agda
+++ b/src/Data/Product/Nary/NonDependent.agda
@@ -256,7 +256,7 @@ updateₙ′ n k = updateₙ n k
 -- Please use the new names as continuing support for the old names is
 -- not guaranteed.
 
--- Version 1.7
+-- Version 2.3
 
 Allₙ = Pointwiseₙ
 {-# WARNING_ON_USAGE Allₙ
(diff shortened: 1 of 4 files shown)


JacquesCarette (approved):

src/Data/Product/Nary/NonDependent.agda
```diff
@@ -240,8 +245,20 @@ updateₙ : ∀ n {ls lᵘ} {as : Sets n ls} k {aᵘ : _ → Set lᵘ} (f : ∀
...
            Product n as → Product n (Updateₙ as k aᵘ)
 updateₙ′ n k = updateₙ n k
+
+------------------------------------------------------------------------
+-- DEPRECATED NAMES
+------------------------------------------------------------------------
+-- Please use the new names as continuing support for the old names is
+-- not guaranteed.
+
+-- Version 1.7
```
MatthewDaggitt: Typo

jamesmckinna:
```suggestion
-- Version 2.3
```
???

PR #2738, opened 2025-06-20
[ add ] `Relation.Nullary.Decidable.dec-yes-recompute`

jamesmckinna:
This PR sharpens the type of `Relation.Nullary.Decidable.dec-yes` by supplying an explicit existential witness:
* `dec-yes-recompute` weakens the hypothesis to allow *irrelevant* argument `.(a : A)`
* by witnessing the existential with `recompute a? a`
* `dec-yes` then follows immediately.

It introduces two new lemmas:
* `Relation.Nullary.Recomputable.recompute-irrelevant-id`: for propositionally irrelevant `A`, `recompute` is the identity
* `Relation.Nullary.Decidable.Core.recompute-irrelevant-id`, encapsulating the above

UPDATED:
* refactors `Relation.Nullary.Recomputable` to lift definitions and basic properties out into `Relation.Nullary.Recomputable.Core`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,11 @@ Deprecated names
   left-inverse ↦ rightInverse
   ```
 
+* In `Relation.Nullary.Decidable`:
+  ```agda
+  dec-yes  ↦ dec-yes-recompute
+  ```
+
 New modules
 -----------
 
@@ -310,6 +315,11 @@ Additions to existing modules
   <₋-accessible-⊥₋ : Acc _<₋_ ⊥₋
   <₋-accessible[_] : Acc _<_ x → Acc _<₋_ [ x ]
   <₋-wellFounded   : WellFounded _<_ → WellFounded _<₋_
+  ```
+
+* In `Relation.Nullary.Decidable`:
+  ```agda
+  dec-yes-recompute : (a? : Dec A) → .(a : A) → a? ≡ yes (recompute a? a)
   ```
 
 * In `Relation.Nullary.Decidable.Core`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,8 +152,9 @@ Deprecated names
 
 * In `Relation.Nullary.Decidable`:
   ```agda
-  dec-yes  ↦ dec-yes-recompute
+  dec-yes  ↦ dec-yes-recompute : (a? : Dec A) → .(a : A) → a? ≡ yes (recompute a? a)
   ```
+  with a more precise type.
 
 New modules
 -----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,12 +150,6 @@ Deprecated names
   left-inverse ↦ rightInverse
   ```
 
-* In `Relation.Nullary.Decidable`:
-  ```agda
-  dec-yes  ↦ dec-yes-recompute : (a? : Dec A) → .(a : A) → a? ≡ yes (recompute a? a)
-  ```
-  with a more precise type.
-
 New modules
 -----------
 
@@ -361,12 +355,6 @@ Additions to existing modules
 * In `Relation.Nullary.Negation.Core`:
   ```agda
   contra-diagonal : (A → ¬ A) → ¬ A
-  ```
-
-* In `Relation.Nullary.Recomputable`:
-  ```agda
-  recompute-irrelevant-id : (recompute : Recomputable A) → Irrelevant A →
-                            (a : A) → recompute a ≡ a
   ```
 
 * In `Relation.Nullary.Reflects`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,7 +352,7 @@ Additions to existing modules
   ```agda
   ⊤-dec : Dec {a} ⊤
   ⊥-dec : Dec {a} ⊥
-  recompute-irr≗id : (a? : Decidable A) → Nullary.Irrelevant A →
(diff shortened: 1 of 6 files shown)


jamesmckinna:
> As this is likely a rather obscure bit of functionality with few users, I don't mind the name churn.

Well, I'm not thrilled by the name churn, but I wasn't quite prepared to just grab the old name and give it the new type... UPDATED but in a `breaking` v3.0 change, I might be tempted to rethink all the various names of these highly niche lemmas?

As it is, and seeing the *single* downstream use of the new construct, I think it can even be avoided altogether (or rather: it is still useful, but localised to its use as a step on a better path to `dec-yes-irr`, which is the enabling technology in `Relation.Binary.PropositionalEquality` for leveraging UIP for `DecidableEquality`...)... see #2740 

UPDATED: the use of the new lemma in `Data.Fin.Permutation` is superseded by the changes in #2740 but I think that these two PRs are orthogonal, modulo rebasing, so I've removed the `blocked-by-issue` label. Think that this would be good to merge now, unless the right thing to do is revert the knock-on changes to `Data.Fin.Permutation`?

src/Relation/Nullary/Recomputable.agda
```diff
@@ -34,10 +35,18 @@ Recomputable : (A : Set a) → Set a
...
+-- Fundamental properties:
+-- 'promotion' is a constant function.
+-- it is moreover the identity, if `A` is propositionally irrelevant.
+
+module _ (promote : Recomputable A) where
+
+  recompute-constant : (p q : A) → promote p ≡ promote q
+  recompute-constant _ _ = refl
+
+  recompute-irr≗id : Nullary.Irrelevant A → (a : A) → promote a ≡ a
```
MatthewDaggitt: Is there a reason why we need to qualify this via `Nullary.`? If not we should probably remove it?

I'm also not super happy with the name. Could be `recompute-irrelevant`? Or `recompute-id`? Or `recompute-irrelevant-id`?

src/Relation/Nullary/Recomputable.agda
```diff
@@ -34,10 +35,18 @@ Recomputable : (A : Set a) → Set a
 Recomputable A = .A → A
 
 ------------------------------------------------------------------------
--- Fundamental property: 'promotion' is a constant function
+-- Fundamental properties:
+-- 'promotion' is a constant function.
+-- it is moreover the identity, if `A` is propositionally irrelevant.
+
+module _ (promote : Recomputable A) where
```
MatthewDaggitt: As all the names refer to `recompute`, shouldn't the function be called that here? Otherwise it's a little confusing to the reader.

CHANGELOG.md
```diff
@@ -150,6 +150,11 @@ Deprecated names
   left-inverse ↦ rightInverse
   ```
 
+* In `Relation.Nullary.Decidable`:
+  ```agda
+  dec-yes  ↦ dec-yes-recompute
```
MatthewDaggitt: We probably want a note here as this isn't a straight deprecation as the new lemma is not equivalent.

CHANGELOG.md
```diff
@@ -312,17 +317,30 @@ Additions to existing modules
...
   ```
 
 * In `Relation.Nullary.Negation.Core`:
   ```agda
   contra-diagonal : (A → ¬ A) → ¬ A
   ```
 
+* In `Relation.Nullary.Recomputable`:
+  ```agda
+  recompute-irr≗id : (promote : Recomputable A) → Nullary.Irrelevant A →
```
MatthewDaggitt: Same comments? `promote` -> `recompute`? No `Nullary.`?

MatthewDaggitt (changes requested):
Otherwise looks good!

CHANGELOG.md
jamesmckinna: The use of `Nullary. Irrelevant` is necessary here, otherwise there is a clash with `Data.Irrelevant.Irrelevant`. @JacquesCarette had previously made a suggestion to refactor `Relation.Nullary.Recomputable` into `Base` and `Properties`, which I think would solve the name clash, but I hadn't at this stage wanted to fold such a thing in to this PR. But perhaps your comment suggest that it should be!?

src/Relation/Nullary/Recomputable.agda
jamesmckinna: As above: alas yes, regarding the qualified name.

As for `naming`, I like `recompute-irrelevant-id` best of the suggestions. Thanks!

CHANGELOG.md
jamesmckinna: Now refactored to split the module dependency.

jamesmckinna:
I suppose that the last step would be to make the `breaking` change, and simply give `dec-yes` the new type of `dec-yes-recompute`... but then we'd be bumping to v3.0 rather than v2.3. Still, worth considering downstream? NB. There are now *no* uses of `dec-yes` elsewhere in `stdlib`.

MatthewDaggitt:
How about leaving the existing definition as is?

jamesmckinna:
Latest commits should fix things as agreed 2025-06-25.

JacquesCarette (approved):
Confirming that I approve of the changes since my last approval.

jamesmckinna:
I removed the coupling with `Data.Fin.Permutation`, so that this is now disjoint/orthogonal to #2740 . 
Just waiting on @MatthewDaggitt 's approval in order to merge.

MatthewDaggitt (approved):

PR #1732, opened 2022-03-04
[ fix #645 ] Irrelevant Data.Empty

gallais:
Shall I also rename `Data.Erased` to `Data.Irrelevant`? I remember people complaining that
it's not the correct name for it. `Data.Erased` should instead be:

```agda
record Erased (A : Set a) where
  field @0 erased : A
```

PR #2740, opened 2025-06-21
[ refactor ] `Data.Fin.Properties` of decidable equality, plus knock-ons

jamesmckinna:
This PR supersedes the knock-on changes for `Permutation` in #2738 and adds some new lemmas about `transpose` which simplify the dependency graph. The contribution can be seen as a belated corollary to the changes introduced in #645 / #1732 .

Downstream, we might want to revisit the dependency graph now between 
* `Relation.Nullary.Decidable` 
* `Axiom.UniquenessOfIdentityProofs`
* `Relation.Binary.PropositionalEquality` 

in putting these pieces together?

Outstanding (naming) issue: I followed the existing `Data.Nat.Properties.≟-diag` convention in `Data.Fin.Properties`, and then backported the `refl` instantiation of this to `Data.Nat.Properties.≟-diag-refl`... but it occurs to me that it's hard to imagine instances of `≟-diag` which would use a general equality, and even if they were to, the use of `diag` *ought* (?) to be a codeword for a repeated argument/reflexive equality? I just couldn't face the deprecation headache though... so a v3.0 downstream `breaking` change might be in order?


jamesmckinna:
In the absence of a review yet, I'll lift out the bug fixes as a separate PR, as per @JacquesCarette 's comment on #2743

src/Data/Fin/Permutation.agda
```diff
@@ -166,19 +168,27 @@ remove {m} {n} i π = permutation to from inverseˡ′ inverseʳ′
 
   inverseʳ′ : StrictlyInverseʳ _≡_ to from
   inverseʳ′ j = begin
-    from (to j)                                                      ≡⟨⟩
-    punchOut {i = i} {πˡ (punchIn (πʳ i) (punchOut to-punchOut))} _  ≡⟨ punchOut-cong′ i (cong πˡ (punchIn-punchOut _)) ⟩
-    punchOut {i = i} {πˡ (πʳ (punchIn i j))}                      _  ≡⟨ punchOut-cong i (inverseˡ π) ⟩
-    punchOut {i = i} {punchIn i j}                                _  ≡⟨ punchOut-punchIn i ⟩
-    j                                                                ∎
+    from (to j)
```
JacquesCarette: I know the original is likely too wide, but it is oh so much more readable! Could we refrain from reformatting in those cases?

src/Data/Fin/Permutation.agda
```diff
@@ -166,19 +168,27 @@ remove {m} {n} i π = permutation to from inverseˡ′ inverseʳ′
...
+    j ∎
 
   inverseˡ′ : StrictlyInverseˡ _≡_ to from
   inverseˡ′ j = begin
-    to (from j)                                                       ≡⟨⟩
-    punchOut {i = πʳ i} {πʳ (punchIn i (punchOut from-punchOut))}  _  ≡⟨ punchOut-cong′ (πʳ i) (cong πʳ (punchIn-punchOut _)) ⟩
-    punchOut {i = πʳ i} {πʳ (πˡ (punchIn (πʳ i) j))}               _  ≡⟨ punchOut-cong (πʳ i) (inverseʳ π) ⟩
-    punchOut {i = πʳ i} {punchIn (πʳ i) j}                         _  ≡⟨ punchOut-punchIn (πʳ i) ⟩
-    j                                                                 ∎
+    to (from j)
```
JacquesCarette: same here

src/Data/Fin/Permutation/Components.agda
```diff
@@ -42,17 +35,31 @@ transpose i j k with does (k ≟ i)
 --  Properties
 ------------------------------------------------------------------------
 
+transpose-iij : ∀ {n} (i j : Fin n) → transpose i i j ≡ j
+transpose-iij i j with j ≟ i in j≟i
+... | true  because [j≡i] = sym (invert [j≡i])
```
JacquesCarette: since you're actually using the proof here, does laziness (i.e. avoiding `yes`) matter?

src/Data/Fin/Properties.agda
```diff
@@ -100,6 +103,18 @@ zero  ≟ suc y = no λ()
 suc x ≟ zero  = no λ()
 suc x ≟ suc y = map′ (cong suc) suc-injective (x ≟ y)
 
+≡-irrelevant : Irrelevant {A = Fin n} _≡_
+≡-irrelevant = Decidable⇒UIP.≡-irrelevant _≟_
+
+≟-diag : (eq : i ≡ j) → (i ≟ j) ≡ yes eq
```
JacquesCarette: I almost commented on the `CHANGELOG` that surely these lemmas are instances of things that are true about combinations of Decidable and Irrelevant, but decided to wait until I saw the proofs... and indeed.

So, do these pay for themselves? I do agree that at the use-site, it sure looks nicer. I'm inclined to say yes.

src/Data/Nat/Properties.agda
```diff
@@ -110,6 +110,12 @@ m ≟ n = map′ (≡ᵇ⇒≡ m n) (≡⇒≡ᵇ m n) (T? (m ≡ᵇ n))
 ≟-diag : (eq : m ≡ n) → (m ≟ n) ≡ yes eq
 ≟-diag = ≡-≟-identity _≟_
 
+≟-diag-refl : ∀ n → (n ≟ n) ≡ yes refl
+≟-diag-refl _ = ≟-diag refl
```
JacquesCarette: This one, on the other hand, seems to fail the Fairbairn test.

JacquesCarette (changes requested):

src/Data/Fin/Permutation.agda
jamesmckinna: Ok, but I can't help but disagree, not least given that in Feijen's original presentation, the justifications were indeed supposed to be interleaved line-by-line with the expressions being calculated. The long lines break any kind of readability. 

But happy to revert against a downstream discussion issue/PR.

src/Data/Fin/Permutation/Components.agda
jamesmckinna: Perhaps not!

src/Data/Nat/Properties.agda
jamesmckinna: Well, the reflexive case us what I use for Fin, and I dislike having the APIs for `Nat` and `Fin` diverging... what is really prefer would be to only have `diag` refer to the `refl` version but... legacy names!

src/Data/Fin/Properties.agda
jamesmckinna: Tricky judgment call, isn't it!? I'm perhaps more willing to pay a tax for specialising a lemma, for sake of avoiding a more convoluted import/instantiation chain in place, but  I guess people's threshold for this is... variable.

src/Data/Nat/Properties.agda
jamesmckinna: But in the interests of merging, I'll reconcile the API to `Nat` style.

src/Data/Fin/Permutation.agda
```diff
@@ -334,17 +335,12 @@ insert-remove : ∀ i (π : Permutation (suc m) (suc n)) → insert i (π ⟨$
 insert-remove {m = m} {n = n} i π j with i ≟ j
 ... | yes i≡j = cong (π ⟨$⟩ʳ_) i≡j
 ... | no  i≢j = begin
-  punchIn (π ⟨$⟩ʳ i) (punchOut (punchInᵢ≢i i (punchOut i≢j) ∘ sym ∘ Injection.injective (↔⇒↣ π))) ≡⟨ punchIn-punchOut _ ⟩
-  π ⟨$⟩ʳ punchIn i (punchOut i≢j) ≡⟨ cong (π ⟨$⟩ʳ_) (punchIn-punchOut i≢j) ⟩
-  π ⟨$⟩ʳ j ∎
+  punchIn (π ⟨$⟩ʳ i) (punchOut (punchInᵢ≢i i (punchOut i≢j) ∘ sym ∘ Injection.injective (↔⇒↣ π)))                             ≡⟨ punchIn-punchOut _ ⟩
```
MatthewDaggitt: This is funny alignment.

src/Data/Fin/Properties.agda
```diff
@@ -100,6 +103,18 @@ zero  ≟ suc y = no λ()
...
+≟-diag = ≡-≟-identity _≟_
```
MatthewDaggitt: I would vote for calling this `≟-diag-≡` and `≟-off-diag` as `≟-diag-≢` for consistency? Actually do we even need the `diag` in the names? Doesn't really do much...

src/Data/Fin/Permutation/Components.agda
```diff
@@ -42,17 +34,31 @@ transpose i j k with does (k ≟ i)
 --  Properties
 ------------------------------------------------------------------------
 
+transpose-iij : ∀ {n} (i j : Fin n) → transpose i i j ≡ j
```
MatthewDaggitt: Normal naming conventions would probably be `transpose[i,i,j]≡j`? Same for the three below?

MatthewDaggitt (changes requested):
Otherwise looks good!

MatthewDaggitt:
Again, going to bump to v2.4

jamesmckinna:
These ones I think I can fix just now, but sure, no deal-breaker in the bump.

src/Data/Fin/Properties.agda
jamesmckinna: That works for me... I'd been following the precedent in `Data.Nat.Properties`, so I'll leave fixing *that* as a downstream rename/deprecate PR.

src/Data/Fin/Permutation.agda
jamesmckinna: Fixed by restoring to the original.

JacquesCarette (approved):

MatthewDaggitt (approved):

PR #2742, opened 2025-06-21
Add `map⁻` to `Unique`

javierdiaz72:
This PR adds `map⁻` to `Data.List.Relation.Unary.Unique`.

### DIFF
--- a/src/Data/List/Relation/Unary/AllPairs/Properties.agda
+++ b/src/Data/List/Relation/Unary/AllPairs/Properties.agda
@@ -47,6 +47,11 @@ module _ {R : Rel A ℓ} {f : B → A} where
   map⁺ []           = []
   map⁺ (x∉xs ∷ xs!) = All.map⁺ x∉xs ∷ map⁺ xs!
 
+  map⁻ : ∀ {xs} → AllPairs R (map f xs) →
+         AllPairs (λ x y → R (f x) (f y)) xs
+  map⁻ {[]}     _              = []
+  map⁻ {_ ∷ _} (fx∉fxs ∷ fxs!) = All.map⁻ fx∉fxs ∷ map⁻ fxs!
+
 ------------------------------------------------------------------------
 -- ++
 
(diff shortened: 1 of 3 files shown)


src/Data/List/Relation/Unary/AllPairs/Properties.agda
```diff
@@ -47,6 +47,11 @@ module _ {R : Rel A ℓ} {f : B → A} where
   map⁺ []           = []
   map⁺ (x∉xs ∷ xs!) = All.map⁺ x∉xs ∷ map⁺ xs!
 
+  map⁻ : ∀ {xs} → AllPairs R (map f xs) →
+         AllPairs (λ x y → R (f x) (f y)) xs
+  map⁻ {[]}     _              = []
+  map⁻ {_ ∷ _} (fx∉fxs ∷ fxs!) = All.map⁻ fx∉fxs ∷ map⁻ fxs!
```
jamesmckinna: The only thing I might suggest as a cosmetic improvement (but: is it?) would be to add `_on_` to the import list from `Function.Base` above, and have this as:
```suggestion
  map⁻ : ∀ {xs} → AllPairs R (map f xs) →
         AllPairs (R on f) xs
  map⁻ {[]}     _              = []
  map⁻ {_ ∷ _} (fx∉fxs ∷ fxs!) = All.map⁻ fx∉fxs ∷ map⁻ fxs!
```
with a corresponding back-port to the type of `map⁺` above.

jamesmckinna:
And I forgot to say: we'll need `CHANGELOG` entries under the heading `Additions to existing modules`... let me know if you have trouble with that, but the existing format should be self-explanatory!

javierdiaz72:
> And I forgot to say: we'll need `CHANGELOG` entries under the heading `Additions to existing modules`... let me know if you have trouble with that, but the existing format should be self-explanatory!

Resolved in ac6c1a45259c310a488e31331c884d44787e9d20.

jamesmckinna:
Pushed some fix-ups, then merging. Thanks @javierdiaz72 !

gallais:
Fixed a needless assumption (functions are always congruent with respect to `_≡_`)

jamesmckinna:
Nice catch @gallais ! Removed the associated import as well.

PR #2744, opened 2025-06-23
[ refactor ] (more) decidable `Data.Fin.Properties`

jamesmckinna:
This PR refactors some proofs about `Decidable` predicates on `Fin n`, to clarify some, and to add others, including some additional `syntax`, around a independently interesting minimisation principle from which certain others follow.

A downstream refactoring might reconsider these things in terms of specialisation of lemmas in `Induction.InfiniteDescent`.

Outstanding issue: could be more systematic about use of `variable`s to clean the whole module up?


src/Data/Fin/Properties.agda
```diff
@@ -949,26 +949,51 @@ module _ {p} {P : Pred (Fin (suc n)) p} where
...
+
+Universal< : ∀ {p} → Pred (Fin n) p → Pred (Fin n) p
+Universal< P i = (j : Fin′ i) → P (inject j)
+
+syntax Universal< (λ j → P) i = ∀[ j < i ] P
+
+MinimalCounterexample : ∀ {p} (P : Pred (Fin n) p) → Set p
+MinimalCounterexample P = ∃[ i ] ¬ P i × ∀[ j < i ] P j
+
+syntax MinimalCounterexample P = μ⟨¬ P ⟩
```
MatthewDaggitt: I don't think we ever try and publicly define new types and syntax in a `Properties` file. To be honest, I really think at this point we should be lifting all these non-trivial decidability results out to a new module `Data.Fin.Decidability`?

src/Data/Fin/Properties.agda
```diff
@@ -949,26 +949,51 @@ module _ {p} {P : Pred (Fin (suc n)) p} where
...
+  module DecFinSubset where
+  cons : (Q 0F → P 0F) → (Q ∘ suc ⊆ P ∘ suc) → Q ⊆ P
+  cons q₀⊆p₀ f = ∀-cons {P = λ x → Q x → P x} q₀⊆p₀ (λ- f) $-
+  ih : Dec (Q ∘ suc ⊆ P ∘ suc)
+  ih = decFinSubset (Q? ∘ suc) P?
+  Q⊆P⇒Q∘suc⊆P∘suc : Q ⊆ P → Q ∘ suc ⊆ P ∘ suc
+  Q⊆P⇒Q∘suc⊆P∘suc f {x} = f {suc x}
+  dec[Q⊆P] : Dec (Q ⊆ P)
+  dec[Q⊆P] with Q? zero
+  ... | false because [¬Q0] = let ¬q₀ = invert [¬Q0] in
```
MatthewDaggitt: Can we not just use `yes` and `no` here?

src/Data/Fin/Properties.agda
```diff
@@ -949,26 +949,51 @@ module _ {p} {P : Pred (Fin (suc n)) p} where
...
-... | false because [¬Q0] | cons =
-  map′ (λ f {x} → cons (⊥-elim ∘ invert [¬Q0]) (λ x → f {x}) x)
-       (λ f {x} → f {suc x})
-       (decFinSubset (Q? ∘ suc) P?)
-... | true  because  [Q0] | cons =
-  map′ (uncurry λ P0 rec {x} → cons (λ _ → P0) (λ x → rec {x}) x)
-       < _$ invert [Q0] , (λ f {x} → f {suc x}) >
-       (P? (invert [Q0]) ×-dec decFinSubset (Q? ∘ suc) P?)
+decFinSubset {suc n} {P = P} {Q} Q? P? = dec[Q⊆P]
+  module DecFinSubset where
```
MatthewDaggitt: Some empty lines between the definitions would definitely help readability here!

src/Data/Fin/Properties.agda
```diff
@@ -949,26 +949,51 @@ module _ {p} {P : Pred (Fin (suc n)) p} where
...
+Universal< P i = (j : Fin′ i) → P (inject j)
+
+syntax Universal< (λ j → P) i = ∀[ j < i ] P
+
+MinimalCounterexample : ∀ {p} (P : Pred (Fin n) p) → Set p
+MinimalCounterexample P = ∃[ i ] ¬ P i × ∀[ j < i ] P j
+
+syntax MinimalCounterexample P = μ⟨¬ P ⟩
+
+min? : ∀ {p} {P : Pred (Fin n) p} → Decidable P → Π[ P ] ⊎ μ⟨¬ P ⟩
```
MatthewDaggitt: We should be able to come up with a better name for this. I guess find minimum? 

This return type is also going to be somewhat baffling. Is writing `Π[ P ] ⊎ μ⟨¬ P ⟩` really more informative than `Universal P ⊎ MinimalCounterexample P`?

src/Data/Fin/Properties.agda
```diff
@@ -949,26 +949,51 @@ module _ {p} {P : Pred (Fin (suc n)) p} where
...
+all? {n = zero}  P? = yes λ()
+all? {n = suc _} P? = Dec.map ∀-cons-⇔ (P? zero ×-dec all? (P? ∘ suc))
+
+Universal< : ∀ {p} → Pred (Fin n) p → Pred (Fin n) p
+Universal< P i = (j : Fin′ i) → P (inject j)
+
+syntax Universal< (λ j → P) i = ∀[ j < i ] P
+
+MinimalCounterexample : ∀ {p} (P : Pred (Fin n) p) → Set p
+MinimalCounterexample P = ∃[ i ] ¬ P i × ∀[ j < i ] P j
```
MatthewDaggitt: This should probably be `ExistsMinimalCounterexample`?

jamesmckinna:
Thanks @MatthewDaggitt for the feedback:
* regarding lifting out these properties to a new module, that does seem like a good idea
* regarding naming/(use of) `syntax`, indeed there are lots of stylistic questions about the current version.

I'll move to DRAFT for the time being, and remove from v2.3, while I reconsider things.

src/Data/Fin/Properties.agda
jamesmckinna: `Data.Fin.Decidable`? `Data.Fin.Decidable.Properties`?
`Data.Fin.Relation.Unary.Decidable`?

src/Data/Fin/Properties.agda
jamesmckinna: Yes, but it's my obsessive avoidance of `yes` and `no` in favour of their lazier versions... as discussed at length in previous PRs...

src/Data/Fin/Properties.agda
jamesmckinna: Regarding naming, the analogous proof for `List` is `Data.List.Relation.Unary.All.search`, but doesn't have a minimality property (we don't prove that it returns the *first* counterexample)

Regarding the type, yes, I'm aware that not everyone likes the `Unary` quantifiers `Π[_]` etc., whereas I find them convenient shorthand. Will reconsider...

src/Data/Fin/Properties.agda
jamesmckinna: At the cost of... a horribly long name? (more ammunition towards a `syntax` abbreviation?)

src/Data/Fin/Properties.agda
jamesmckinna: Regarding `min?` itself as a name:
* postfix `?` to denote that this is a decision procedure (but maybe we should reserve such usage for functions returning `Dec _` results?)
* `min`/`μ` by association with Kleene's minimisation operator.

src/Data/Fin/Properties.agda
jamesmckinna: I've opted for the last one...

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,166 @@
...
+    i j : Fin n
+    P : Pred (Fin n) p
+    Q : Pred (Fin n) q
+
+
+------------------------------------------------------------------------
+-- Quantification
+------------------------------------------------------------------------
+
+module _ {P : Pred (Fin (suc n)) p} where
```
JacquesCarette: Why are these here? None of these are about a decidable predicate. I think these should be either in the main file or, at worst, a `.Core` version, but not here.

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,166 @@
...
+-- When P is a decidable predicate over a finite set the following
+-- lemma can be proved.
+
+¬∀⟶∃¬ : ∀ n (P : Pred (Fin n) p) → Decidable P →
+          ¬ (∀ i → P i) → (∃ λ i → ¬ P i)
+¬∀⟶∃¬ n P P? ¬P = map id proj₁ (¬∀⟶∃¬-smallest n P P? ¬P)
+
+
+------------------------------------------------------------------------
+-- Effectful
```
JacquesCarette: Likewise, there does not seem to be a good reason to move these here from the main file. This almost feels like a cut-and-paste error?

src/Data/Fin/Properties.agda
JacquesCarette: `findMinimum` seems like a good name. `searchForMin` is another viable option.  postfix `?` to me always indicates `Dec`.

jamesmckinna:
Hmmm... @JacquesCarette the code motion was in response to earlier review comments from @MatthewDaggitt , but maybe I misunderstood his intention. Nevertheless, the `Quantification` section is necessary for the subsequent `decFinSubset` and `all?`/`any?` definitions... so it seemed cognate to keep them together (esp. given the `public` re-export in `Data.Fin.Properties`...)

JacquesCarette:
I had seen some of the conversation. This is why maybe a larger break up (`Data.Fin.Properties.Quantification` too?) might make more sense. Same with the `Effectful` stuff, it most definitely does not belong in `.Properties.Decidable`. The main code motion does make a lot of sense, it just has knock-on consequences.

src/Data/Fin/Properties.agda
jamesmckinna: `searchMin`?

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: Well, they are more about Unary predicates on Fin than anything else, so moving them untied a tricky dependency ambiguity (having `Unary` and `Binary` versions of `Decidable` imported together... annoying).

And as above, the public re-export means that the API for `Data.Fin.Properties` does not change...

src/Data/Fin/Properties.agda
JacquesCarette: The github email made it unclear whether the `?` was in the name or not...  `searchMin` is fine with me.

src/Data/Fin/Relation/Unary/Decidable.agda
JacquesCarette: `Data.Fin.Relation.Unary.Effectful` ?

jamesmckinna: Now in `Data.Fin.Relation.Unary.Base`, in the interests of not proliferating too many too-small modules...

src/Data/Fin/Properties.agda
jamesmckinna: Now in terms of `yes`, `no`.

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: Now in a `.Base`, imported by `.Decidable`, and re-exported in `Properties`.

jamesmckinna:
Last addition: dual version of `searchMin` so that client use doesn't have to figure out the DNE...

jamesmckinna:
@JacquesCarette If you are happy with the revisions, then we could add this to the v2.3 docket?

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of Decidable unary predicates on Fin
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Fin.Relation.Unary.Decidable where
```
MatthewDaggitt, line 9: Why do we need `Decidable` in the name? Isn't everything over `Fin` going to be decidable?

src/Data/Fin/Relation/Unary/Base.agda
```diff
@@ -0,0 +1,78 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Properties of Decidable unary predicates on Fin
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Data.Fin.Relation.Unary.Base where
```
MatthewDaggitt, line 9: Why do we need this `Base` module?

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
...
+  -- The boolean component of the result is exactly the
+  -- obvious fold of boolean tests (`foldr _∧_ true`).
+  note : ∀ {P : Pred (Fin 3) p} (P? : Decidable P) →
+         ∃ λ z → Dec.does (all? P?) ≡ z
+  note P? = Dec.does (P? 0F) ∧ Dec.does (P? 1F) ∧ Dec.does (P? 2F) ∧ true
+          , refl
+
+------------------------------------------------------------------------
+-- search
+-- a Decidable predicate is either always true, or has a smallest counterexample
```
MatthewDaggitt, line 59: Capitalisation and formatting of this header could be better?

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
...
+
+
+Universal< : Pred (Fin n) p → Pred (Fin n) p
+Universal< P i = (j : Fin′ i) → P (inject j)
+
+syntax Universal< (λ j → P) i = ∀[ j < i ] P
+
+ExistsMinimalCounterexample : (P : Pred (Fin n) p) → Set p
+ExistsMinimalCounterexample P = ∃[ i ] ¬ P i × ∀[ j < i ] P j
```
MatthewDaggitt, line 68: We could just call this `MinimalCounterexample`? Making this a record with named fields would make it easier to read as well?

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
...
+
+syntax ExistsMinimalCounterexample P = μ⟨¬ P ⟩
```
MatthewDaggitt, line 70: I'm still kind of against this for readability reasons, but why does this have to be a syntax declaration rather than an ordinary infix declaration?

src/Data/Fin/Properties.agda
```diff
@@ -123,6 +96,36 @@ suc x ≟ suc y = map′ (cong suc) suc-injective (x ≟ y)
...
+------------------------------------------------------------------------
+-- Reexport properties of Unary predicates on Fin
+------------------------------------------------------------------------
+
+open import Data.Fin.Relation.Unary.Base public
+open import Data.Fin.Relation.Unary.Decidable public
+
+------------------------------------------------------------------------
+-- Fin
+------------------------------------------------------------------------
```
MatthewDaggitt, line 108: Is there a reason this and `Bundles` have been moved? To be they logically make sense to live at the top of the file as they related to the type itself, rather than functions or relations over it.

src/Data/Fin/Properties.agda
```diff
@@ -123,6 +96,36 @@ suc x ≟ suc y = map′ (cong suc) suc-injective (x ≟ y)
   { isDecEquivalence = ≡-isDecEquivalence {n}
   }
 
+------------------------------------------------------------------------
+-- Reexport properties of Unary predicates on Fin
+------------------------------------------------------------------------
+
+open import Data.Fin.Relation.Unary.Base public
+open import Data.Fin.Relation.Unary.Decidable public
```
MatthewDaggitt, line 104: I'm not convinced about this direction of dependencies.

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
...
+  note P? = Dec.does (P? 0F) ∧ Dec.does (P? 1F) ∧ Dec.does (P? 2F) ∧ true
+          , refl
+
+------------------------------------------------------------------------
+-- search
+-- a Decidable predicate is either always true, or has a smallest counterexample
+
+
+Universal< : Pred (Fin n) p → Pred (Fin n) p
+Universal< P i = (j : Fin′ i) → P (inject j)
```
MatthewDaggitt, line 63: Why do we use `Fin'` and `inject` here rather than an explicit `j < i` proof? Generally `Fin'` is quite difficult to work with.

src/Data/Fin/Relation/Unary/Decidable.agda
```diff
@@ -0,0 +1,122 @@
...
+
+syntax Universal< (λ j → P) i = ∀[ j < i ] P
+
+ExistsMinimalCounterexample : (P : Pred (Fin n) p) → Set p
+ExistsMinimalCounterexample P = ∃[ i ] ¬ P i × ∀[ j < i ] P j
+
+syntax ExistsMinimalCounterexample P = μ⟨¬ P ⟩
+
+searchMin : Decidable P → (∀ i → P i) ⊎ ∃[ i ] ¬ P i × ∀[ j < i ] P j
```
MatthewDaggitt, line 72: How about calling this `searchForCounterexample` and the one below `searchForWitness`?

MatthewDaggitt:
Again has merge conflicts, no one waiting on it particularly and @jamesmckinna is away so bumping to v2.4

jamesmckinna:
Happy to bump! There was enough feedback/pushback on some of the choices I'd made that this one deserves a bit more time/energy/thought.

jamesmckinna:
I'm going to close this one for now, not least because, as is usual for me, the PR tries to do too many things at once:
* refactor existing proofs via an intermediate result with 'better' constructive content (disjunction *vs.* implication from negated antecedent); 
* introduce new definitions to package such reasoning: bounded quantifiers on `Fin` (should be introduced separately), with `searchForWitness` and `searchForCounterexample` as derived forms; 
* `syntax` declarations for the above, by analogy with existing forms under `Relation.Unary` (but this seems to be more controversial)
* code-motion to try to sort out the dependency cycles which attempting all of the above seemed to entail.

It's a mess that I don't quite feel like untangling, except as separate PRs for each of the above, and certainly not now.

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: Re: naming
* I tried to factor out properties of `Decidable` `Unary` predicates... so the name seemed to 'fit'

Re: are all `Unary` predicates on decidable?
* I don't think so... by pulling back any predicate `P` on `ℕ` such that `P 0` iff some undecidable problem, and then consider `P ∘ toℕ`? or am I hopelessly misguided?

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: Using a `record` not a bad idea...

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: This is pure legacy! The refactoring here was designed to make an existing (however poorly expressed) statement derivable from something stronger...

src/Data/Fin/Relation/Unary/Decidable.agda
JacquesCarette: I think you're missing the part that everything in Agda terminates. I think that to do this kind of encoding, you'd need to encode a non-enumerable subset of ℕ into `P`. You can indeed use `Fin 2` instead of `Bool` to encode undecidability. But it the predicate computes and only has a finite number of cases to check, that is decidable.

JacquesCarette:
(Oops, didn't notice it was subsequently closed before commenting.)

src/Data/Fin/Relation/Unary/Decidable.agda
jamesmckinna: Hmmmph... back to school for me!

PR #2747, opened 2025-06-25
Lemmata for `if_then_else_`

pmbittner:
We used `if_then_else_` a lot in one of [our projects](https://github.com/VariantSync/Vatras) and it turned out that sometimes these basic lemmata on `if_then_else_` are quite useful. While it is typically straightforward to resolve `if_then_else_` expressions by pattern matching on the boolean condition instead, sometimes it is convenient to use these lemmata when an `if_then_else_` occurs somewhere deep in an equational-reasoning chain.

### DIFF
--- a/src/Data/Bool/Properties.agda
+++ b/src/Data/Bool/Properties.agda
@@ -751,12 +751,14 @@ if-eta false = refl
 if-eta true  = refl
 
 if-idem-then : ∀ b {x y : A} →
-               (if b then (if b then x else y) else y) ≡ (if b then x else y)
+               (if b then (if b then x else y) else y)
+             ≡ (if b then x                    else y)
 if-idem-then false = refl
 if-idem-then true  = refl
 
 if-idem-else : ∀ b {x y : A} →
-               (if b then x else (if b then x else y)) ≡ (if b then x else y)
+               (if b then x else (if b then x else y))
+             ≡ (if b then x else y)
 if-idem-else false = refl
 if-idem-else true  = refl
 
@@ -790,7 +792,8 @@ if-∨ false = refl
 if-∨ true  = refl
 
 if-xor : ∀ b {c} {x y : A} →
-         (if b xor c then x else y) ≡ (if b then (if c then y else x) else (if c then x else y))
+         (if b xor c then x else y)
+       ≡ (if b then (if c then y else x) else (if c then x else y))
 if-xor false = refl
 if-xor true {false} = refl
 if-xor true {true } = refl
@@ -808,19 +811,23 @@ if-xor true {true } = refl
 -- unfortunately fails to resolve the omitted argument.)
 
 if-cong : ∀ {b c} {x y : A} → b ≡ c →
-          (if b then x else y) ≡ (if c then x else y)
+          (if b then x else y)
+        ≡ (if c then x else y)
 if-cong refl = refl
 
 if-cong-then : ∀ b {x y z : A} → x ≡ z →
-               (if b then x else y) ≡ (if b then z else y)
+               (if b then x else y)
+             ≡ (if b then z else y)
 if-cong-then _ refl = refl
 
 if-cong-else : ∀ b {x y z : A} → y ≡ z →
-               (if b then x else y) ≡ (if b then x else z)
+               (if b then x else y)
+             ≡ (if b then x else z)
 if-cong-else _ refl = refl
 
 if-cong₂ : ∀ b {x y z w : A} → x ≡ z → y ≡ w →
-           (if b then x else y) ≡ (if b then z else w)
+           (if b then x else y)
+         ≡ (if b then z else w)
 if-cong₂ _ refl refl = refl
 
 ------------------------------------------------------------------------


src/Data/Bool/Properties.agda
```diff
@@ -745,6 +745,50 @@ if-float : ∀ (f : A → B) b {x y} →
 if-float _ true  = refl
 if-float _ false = refl
 
+if-idemp : ∀ b {x : A} →
```
Taneb: Normally we abbreviate idempotent to `idem`

gallais: Is it more of an `if-eta`?

src/Data/Bool/Properties.agda
```diff
@@ -745,6 +745,50 @@ if-float : ∀ (f : A → B) b {x y} →
...
+           (if b then x else x) ≡ x
+if-idemp false = refl
+if-idemp true  = refl
+
+if-swap : ∀ b c {x y : A} →
```
gallais: Is this `if-swapʳ`? Which suggests there's an opportunity for a left version too.

Taneb: I'd prefer it was called `if-swap-else`, there's three sides to every if (and similar for the `if-cong`s below)

src/Data/Bool/Properties.agda
pmbittner: I think `if-eta` is a good name because it removes the unnecessary if-abstraction. Idempotence might rather be something like the following, which I am going to add:
```agda
if-idem : ∀ b {x y : A} → (if b then (if b then x else y) else y) ≡ (if b then x else y)
if-idem false = refl
if-idem true  = refl
```

src/Data/Bool/Properties.agda
pmbittner: Ok, I will rename the theorem to `if-swap-else`, and I will also add the `if-swap-then` case.

Speaking of the `cong`-theorems, I noticed that a `cong`-theorem for the boolean condition is missing, so I am going to add that as well.

src/Data/Bool/Properties.agda
pmbittner: Ah and this is again rather an `if-idem-then` and there is also an `if-idem-else`.

src/Data/Bool/Properties.agda
```diff
@@ -745,6 +745,72 @@ if-float : ∀ (f : A → B) b {x y} →
...
+if-∨ false = refl
+if-∨ true  = refl
+
+if-xor : ∀ b {c} {x y : A} →
+         (if b xor c then x else y) ≡ (if b then (if c then y else x) else (if c then x else y))
+if-xor false = refl
+if-xor true {false} = refl
+if-xor true {true } = refl
+
+if-cong : ∀ {b c} {x y : A} → b ≡ c →
```
Taneb, line 813: Thinking about it, isn't this just `cong (if_then x else y)`? And similar for the other cong lemmas. Is it worth including?

pmbittner: Yes, it is indeed exactly that.

The benefit of the if-cong lemmas is that we can infer some of the branches automatically with implicit arguments, so we can omit them. That was the reason why I created these cong-lemmas in our project because sometimes we had large expressions in these branches and it was tedious to spell them out. Example:

```agda
open import Data.Bool using (if_then_else_)
open import Data.Bool.Properties using (if-cong-else)
open import Data.Nat using (ℕ; _+_)
open import Data.Nat.Properties using (+-comm)
open import Relation.Binary.PropositionalEquality as Eq using (_≡_; refl; cong)

looooooong-expression = 1

with-if-cong : ∀ b (x y : ℕ)
  → (if b then looooooong-expression else x + y)
  ≡ (if b then looooooong-expression else y + x)
with-if-cong b x y = if-cong-else b (+-comm x y)

with-cong : ∀ b (x y : ℕ)
  → (if b then looooooong-expression else x + y)
  ≡ (if b then looooooong-expression else y + x)
with-cong b x y = cong (if b then looooooong-expression else_) (+-comm x y)
```

JacquesCarette (approved):
Nice.

A lot of these generalize to partitions and piecewise functions. Which I should implement.

src/Data/Bool/Properties.agda
pmbittner: This does not help with all elements within an `if_then_else_` though because we still have to spell out the condition (`b` in the example above).

MatthewDaggitt: Can you not just write `cong (if b then _ else_) (+-comm x y)` in that second proof? Does it actually not infer it? I assumed that anything that could be inferred by implicit arguments should also be inferable via underscores?

MatthewDaggitt:
I'm dubious about the utility of the `cong` proofs but happy to approve if either i) they are removed or ii) they turn out to be actually needed. See my comment above.

src/Data/Bool/Properties.agda
pmbittner: Good catch. I expected your idea to work but surprisingly it does not type check. (I admit that I do not know when it is safe to use `_` and I was not able to find documentation on it in the Agda documentation in a quick search.)

image

MatthewDaggitt: Wow, that's very weird. Okay, happy to accept it then. Can you write a couple of lines above the `cong` lemma section explaining why we're adding them then! I'll then accept and merge for v2.3.

pmbittner: I added the comment ([3eb32c3](https://github.com/agda/agda-stdlib/pull/2747/commits/3eb32c383344b4bdf6bee1a567e35075f539db2d)). I also noticed that some lemmas are longer than the style guide's line limit, so I added line breaks to some lemmas ([0d55cb9](https://github.com/agda/agda-stdlib/pull/2747/commits/0d55cb9c7c9fb79378082f295c7a487a0d92d1ed)). These line breaks also make it easier to see which terms in an `if_then_else_` are affected by a substitution.

PR #2754, opened 2025-06-28
Correct option --warn to --warning for agda/agda#7966

andreasabel:
Correct option --warn to --warning for agda/agda#7966

The official name of the option is `--warning` and the prefix `--warn` is only accepted because we have no options like `--warner` or `--warner-brothers-owns-us`, but we might have so in the future.

### DIFF
--- a/CHANGELOG/v1.3.md
+++ b/CHANGELOG/v1.3.md
@@ -226,7 +226,7 @@ Deprecated modules
 
 * A warning is now raised whenever you import a deprecated module. This should
   aid the transition to the new modules. These warnings can be disabled locally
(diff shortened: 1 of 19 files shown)


jamesmckinna:
Good catch!

andreasabel:
I suppose this could be cherry-picked onto `master` as well.  The patch currently only sits on `experimental`.
Shall I pick it to `master`?

jamesmckinna:
> I suppose this could be cherry-picked onto `master` as well. The patch currently only sits on `experimental`. Shall I pick it to `master`?

No need I think. We're planning a v2.3 release to synchronise with Agda v2.8.0, and part of the `stdlib` release cycle involves merging in `experimental` (I htink @MatthewDaggitt is busy with that at the moment), so no need on your part to act now...

PR #2757, opened 2025-07-01
[ add ] `∸-suc` lemma for natural numbers

pmbittner:
~~These lemmata were~~This lemma was useful in inductive proofs involving `∸` ~~and `≤`~~ and it seems ~~these two lemmata are~~ this lemma is not yet part of the standard library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,3 +26,9 @@ New modules
 
 Additions to existing modules
 -----------------------------
+
+* In `Data.Nat.Properties`:
+  ```agda
+  ∸-suc : m ≤ n → suc n ∸ m ≡ suc (n ∸ m)
+  ```
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@ Non-backwards compatible changes
 Minor improvements
 ------------------
 
+* Refactored usages of `+-∸-assoc 1` to `∸-suc` in:
+  ```agda
+  README.Data.Fin.Relation.Unary.Top
+  Algebra.Properties.Semiring.Binomial
+  Data.Fin.Subset.Properties
+  Data.Nat.Binary.Subtraction
+  Data.Nat.Combinatorics
+  ```
+
 Deprecated modules
 ------------------
 
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1637,7 +1637,7 @@ m<n+o⇒m∸n<o (suc m) (suc n)             lt = m<n+o⇒m∸n<o m n  (s<s⁻¹ 
 m+n≤o⇒m≤o∸n : ∀ m {n o} → m + n ≤ o → m ≤ o ∸ n
 m+n≤o⇒m≤o∸n zero    le       = z≤n
 m+n≤o⇒m≤o∸n (suc m) (s≤s le)
-  rewrite ∸-suc(m+n≤o⇒n≤o m le) = s≤s (m+n≤o⇒m≤o∸n m le)
+  rewrite ∸-suc (m+n≤o⇒n≤o m le) = s≤s (m+n≤o⇒m≤o∸n m le)
 
 m≤o∸n⇒m+n≤o : ∀ m {n o} (n≤o : n ≤ o) → m ≤ o ∸ n → m + n ≤ o
 m≤o∸n⇒m+n≤o m         z≤n       le rewrite +-identityʳ m = le
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1524,6 +1524,10 @@ pred[m∸n]≡m∸[1+n] (suc m) (suc n) = pred[m∸n]≡m∸[1+n] m n
 ------------------------------------------------------------------------
 -- Properties of _∸_ and _≤_/_<_
 
+∸-suc : m ≤ n → suc n ∸ m ≡ suc (n ∸ m)
+∸-suc z≤n       = refl
+∸-suc (s≤s m≤n) = ∸-suc m≤n
+
 m∸n≤m : ∀ m n → m ∸ n ≤ m
 m∸n≤m n       zero    = ≤-refl
 m∸n≤m zero    (suc n) = ≤-refl
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1618,12 +1618,11 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
 ∸-+-assoc (suc m) (suc n) o = ∸-+-assoc m n o
 
 +-∸-assoc : ∀ m {n o} → o ≤ n → (m + n) ∸ o ≡ m + (n ∸ o)
-+-∸-assoc m (z≤n {n = n})             = begin-equality m + n ∎
-+-∸-assoc m (s≤s {m = o} {n = n} o≤n) = begin-equality
-  (m + suc n) ∸ suc o  ≡⟨ cong (_∸ suc o) (+-suc m n) ⟩
-  suc (m + n) ∸ suc o  ≡⟨⟩
-  (m + n) ∸ o          ≡⟨ +-∸-assoc m o≤n ⟩
-  m + (n ∸ o)          ∎
++-∸-assoc zero    {n = n} {o = o} _   = begin-equality n ∸ o ∎
++-∸-assoc (suc m) {n = n} {o = o} o≤n = begin-equality
+  suc (m + n) ∸ o   ≡⟨ ∸-suc (m≤n⇒m≤o+n m o≤n) ⟩
+  suc ((m + n) ∸ o) ≡⟨ cong suc (+-∸-assoc m o≤n) ⟩
+  suc (m + (n ∸ o)) ∎
 
 m≤n+o⇒m∸n≤o : ∀ m n {o} → m ≤ n + o → m ∸ n ≤ o
 m≤n+o⇒m∸n≤o      m  zero    le = le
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -251,6 +251,12 @@ Additions to existing modules
   ∙-cong-∣ : x ∣ y → a ∣ b → x ∙ a ∣ y ∙ b
   ```
 
+* In `Data.Nat.Properties`:
+  ```agda
+  ∸-suc : n ≤ m → suc m ∸ n ≡ suc (m ∸ n)
+  ≤-suc : n ≤ m → n ≤ suc m
```
jamesmckinna: This is a duplicate of `m≤n⇒m≤1+n` from L293 below, so should be removed from the PR.

CHANGELOG.md
```diff
@@ -251,6 +251,12 @@ Additions to existing modules
   ∙-cong-∣ : x ∣ y → a ∣ b → x ∙ a ∣ y ∙ b
   ```
 
+* In `Data.Nat.Properties`:
+  ```agda
+  ∸-suc : n ≤ m → suc m ∸ n ≡ suc (m ∸ n)
```
jamesmckinna: This might better be called `suc-∸` (we tend to favour L-R reading order for lemmas about composition of functions).

Also, following the `style-guide` precepts about alphabetical mention of specimen names, esp. those bound in `variable` declarations, the (name and) type of this lemma should probably be
```suggestion
  suc-∸ : m ≤ n → suc n ∸ m ≡ suc (n ∸ m)
```

CHANGELOG.md
pmbittner: Oops, sorry. I removed the respective commits.

CHANGELOG.md
pmbittner: I chose `∸-suc` as a symmetric name to `+-suc` and `*-suc` which look quite similar. I have no problem with a renaming though.

Sorry for getting the variable names wrong, fixed now.

jamesmckinna: Let's see what our colleagues say about names!

JacquesCarette: Tricky: I prefer James' suggestion, but it's also true that this is consistent with pre-existing names! So I don't know.

jamesmckinna: Yes,... maybe I was being too doctrinaire, moreover for a doctrine which is more often honoured in the breach... sigh.

pmbittner:
I rebased onto master and now an internal CI error occurred (i.e., something in the pipeline, not the agda standard library failed).

pmbittner:
It seems that `∸-suc` is already part of the standard library in terms of `+-∸-assoc 1` as well. Sometimes it is not easy to spot whether a theorem you need is already there.... :sweat_smile:

jamesmckinna:
> It seems that `∸-suc` is already part of the standard library in terms of `+-∸-assoc 1` as well. Sometimes it is not easy to spot whether a theorem you need is already there.... 😅

Ah! Yes! Good catch.but indeed the combination you describe us a common case, so I'm wondering if it is worth the Fairbairn threshold penalty to include your version as precisely thus instantiating the general case? Not sure, but suggest we perhaps in any case push to c2.4 to allow us all more time to discuss?

jamesmckinna:
Suggestion: 
* we add your lemma (under whatever name ;-))
* we refactor the subsequent proof of `+-∸-assoc` as follows:
  ```agda
  +-∸-assoc : ∀ m {n o} → o ≤ n → (m + n) ∸ o ≡ m + (n ∸ o)
  +-∸-assoc zero    {n = _} {o = _} _  = refl
  +-∸-assoc (suc m) {n = n} {o = o} le = begin-equality
    suc (m + n) ∸ o   ≡⟨ ∸-suc (m≤n⇒m≤o+n m le) ⟩
    suc ((m + n) ∸ o) ≡⟨ cong suc (+-∸-assoc m le) ⟩
    suc (m + (n ∸ o)) ∎
  ```
  which I think is a modest improvement, and the instantiation at `m = 1` which occurs twice elsewhere in `Data.Nat.Properties` can be refactored to use your new lemma.

pmbittner:
Sounds good to me, given that there are instantiations of `+-∸-assoc 1`. :slightly_smiling_face: I pushed the refactorings of the  `+-∸-assoc 1` usages. Do you want to do the refactoring of `+-∸-assoc` or should I push the change?

jamesmckinna:
> Sounds good to me, given that there are instantiations of `+-∸-assoc 1`. 🙂 I pushed the refactorings of the `+-∸-assoc 1` usages. Do you want to do the refactoring of `+-∸-assoc` or should I push the change?

Happy for you to push the change!

jamesmckinna:
Aaargh! as I've learnt to my cost in the past, if you `@` someone in a commit message, they get a shit-ton of auto-generated (and redundant/spurious) email from GitHiub when eventually merged (both into `master` and then into any subsequent `merge` anywhere else, and then the final release...). but thanks for the shoutout...

src/Data/Nat/Properties.agda
```diff
@@ -1630,7 +1633,7 @@ m<n+o⇒m∸n<o (suc m) (suc n)             lt = m<n+o⇒m∸n<o m n  (s<s⁻¹
 m+n≤o⇒m≤o∸n : ∀ m {n o} → m + n ≤ o → m ≤ o ∸ n
 m+n≤o⇒m≤o∸n zero    le       = z≤n
 m+n≤o⇒m≤o∸n (suc m) (s≤s le)
-  rewrite +-∸-assoc 1 (m+n≤o⇒n≤o m le) = s≤s (m+n≤o⇒m≤o∸n m le)
+  rewrite ∸-suc(m+n≤o⇒n≤o m le) = s≤s (m+n≤o⇒m≤o∸n m le)
```
jamesmckinna:
```suggestion
  rewrite ∸-suc (m+n≤o⇒n≤o m le) = s≤s (m+n≤o⇒m≤o∸n m le)
```

src/Data/Nat/Properties.agda
```diff
@@ -1610,12 +1614,11 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
 ∸-+-assoc (suc m) (suc n) o = ∸-+-assoc m n o
 
 +-∸-assoc : ∀ m {n o} → o ≤ n → (m + n) ∸ o ≡ m + (n ∸ o)
-+-∸-assoc m (z≤n {n = n})             = begin-equality m + n ∎
-+-∸-assoc m (s≤s {m = o} {n = n} o≤n) = begin-equality
-  (m + suc n) ∸ suc o  ≡⟨ cong (_∸ suc o) (+-suc m n) ⟩
-  suc (m + n) ∸ suc o  ≡⟨⟩
-  (m + n) ∸ o          ≡⟨ +-∸-assoc m o≤n ⟩
-  m + (n ∸ o)          ∎
++-∸-assoc zero    {n = _} {o = _} _   = refl
```
jamesmckinna: And in the spirit of the original proof:
```suggestion
+-∸-assoc zero    {n = n} {o = o} _   = begin-equality n ∸ o ∎
```
This is pure cosmetics, however, and I'm not really sure that `begin-equality x ∎` is quite the best way to write `refl` when you want to explicitly mention `x`?

src/Data/Nat/Properties.agda
```diff
@@ -1610,12 +1614,11 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
...
++-∸-assoc (suc m) {n = n} {o = o} o≤n = begin-equality
+  suc (m + n) ∸ o   ≡⟨ ∸-suc (m≤n⇒m≤o+n m o≤n) ⟩
+  suc ((m + n) ∸ o) ≡⟨ cong suc (+-∸-assoc m o≤n) ⟩
```
jamesmckinna, line 1624: Can we drop any parentheses here? I had been being hyper-cautious when writing the new proof...

pmbittner:
Oops, sorry. I can apply your suggested changes and remove the "@" from the commit message as well.

src/Data/Nat/Properties.agda
pmbittner: We can simplifiy `(m + n) ∸ o` to `m + n ∸ o` but I am sceptic whether this is more readable, especially in a proof about associativity?

src/Data/Nat/Properties.agda
pmbittner: Keeping the old style is probably a good idea. I also do not know what is the best style. Alternatively, we could do `refl {x = n ∸ o}`.

jamesmckinna (approved):
All good now. In fact: better than before, I think!

jamesmckinna:
Additional uses of the lemma which could usefully be refactored as part of this:
* `README.Data.Fin.Relation.Unary.Top` L79
* `Algebra.Properties.Semiring.Binomial` L222
* `Data.Fin.Subset.Properties` L370
* `Data.Nat.Binary.Subtraction` L91
* `Data.Nat.Combinatorics` L125

pmbittner:
This is a good idea! Thanks for collecting these references. I will refactor them.

pmbittner:
I refactored the references now and rebased onto main.

There seems to be a type error in `README.Data.Fin.Relation.Unary.Top` within the definition of `>-weakInduction` (unrelated to this PR). I opened #2791 to fix that.

jamesmckinna:
> I refactored the references now and rebased onto main.
> 
> There seems to be a type error in `README.Data.Fin.Relation.Unary.Top` within the definition of `>-weakInduction` (unrelated to this PR). I opened #2791 to fix that.

That's #2789 opened by me over the w/e:-) I é closed mine in favour of your better one, and queued it for merging now.

JacquesCarette (approved):

PR #2758, opened 2025-07-03
Cosmetics: delete 2 rewrite steps that do not fire

andreasabel:
Detected by a new warning for superfluous rewrites: agda/agda#7973

Note: Agda `master` will soon warn about these, but 2.8.0 won't (unless maybe its release is further delayed).

Anyway, this PR can be merged immediately since it removes rewrite steps that are superfluous independent on the Agda version.


Taneb (approved):
It looks like this compiles OK, so let's get this merged

JacquesCarette:
Really quite intriguing that there were indeed 'useless' steps in proofs in stdlib. Definite improvement to have these be gone (and have warnings that help us keep it that way).

PR #2759, opened 2025-07-07
[ refactor ] cosmetic improvement to `Algebra.Construct.Initial`

jamesmckinna:
PR #2546 fixed a v2.8.0 compatibility issue regarding inference of implicit arguments in instantiation of `record` fields.

This PR tidies up those changes:
* to be less noisy
* to add two more definitions which don't incur the problem.

Badged as v2.3, because it 'fixes' an earlier v2.3 issue/pr but could easily be bumped to v2.4.

The original refactoring hoped to define a single module `ℤero` from which all subsequent definitions could be given by record subtyping simply as `... = record { ℤero }`, but that dream has now evaporated. So it might be worth simply abandoning the explicit definition altogether, although the book-keeping overhead outweighs the benefit. Could make this `breaking` for v3.0, though!?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,12 @@ Additions to existing modules
     ∙-congʳ : RightCongruent _≈_ _∙_
   ```
 
+* In `Algebra.Construct.Initial`:
+  ```agda
+  assoc : Associative _≈_ _∙_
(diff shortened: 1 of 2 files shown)


JacquesCarette (approved):

JacquesCarette:
I agree, this PR should be v2.3.

Issue #2704, opened 2025-04-17
Do we need a definition for `(Is)BooleanRing`?


onestruggler:
Here, boolean algebra gives a commutative ring, but it shoud give more than that, .i.e.,  a boolean ring. On the other hand, there is no more other boolean ring except for the one comes from a boolean algebra.  I am not sure if we need such a defintion.  On the other other hand, at least, a usage will be clarifing the correspondence between boolean algebras and boolean rings.

https://github.com/agda/agda-stdlib/blob/6db08b3b2af9f220c016d2a79ace400ae60006a8/src/Algebra/Lattice/Properties/BooleanAlgebra.agda#L527

jamesmckinna:
There probably is/may be some archeology to go through here (NB I have not myself done so!) about the `DefaultXor` implementation, as well as `Biased` versions of `Algebra.Lattice.{Structures|Bundles}`?

JacquesCarette:
While it "makes sense" to add `IsBooleanRing` to `Algebra.Lattice.Structures.Biased`, it also pretty much insures no one's going to find it...  `IsBooleanRing` in my mind is part of the 'usual' `Algebra` hierarchy. It's a *theorem* that it is isomorphic to something that naturally finds a home in `Algebra.Lattice`.

jamesmckinna:
Hmmm, aren't all the `Biased` constructions expressions of 'theorems'... as in, 'smart constructors' are 'proofs' of derivable/admissible properties, under PAT. Cf. `Group`s as `Loop`s/`Quasigroup`s?

As for discoverability (but, as ever, `grep`, though a very blunt instrument, would find it for you?) I'd be happy with it under `Algebra.Structures`, but once the theorem is proved (moreover: where?) is there subsequently any further role for it? (Apologies for my ignorance on this one, for all I know, the ring aspect is more important than the order-theoretic one?)

jamesmckinna:
Hmmm... I've been experimenting towards a PR for this... but I'm not even clear what the API should be:
* the 'obvious' would/should be
  ```agda
  record IsBooleanRing
         (_+_ _*_ : Op₂ A) (-_ : Op₁ A) (0# 1# : A) : Set (a ⊔ ℓ) where
  field
    isRing : IsRing _+_ _*_ -_ 0# 1#
    *-idem : Idempotent _*_
  ```
  from which we can then *prove* (eg see https://en.wikipedia.org/wiki/Boolean_ring#Properties_of_Boolean_rings)
[...]

PR #2761, opened 2025-07-08
[ add ] rule of signs for `RingWithoutOne`

jamesmckinna:
Surprised to find this was not anywhere in the library! 
Warm-up for tackling #2704 , so could instead be folded into a PR for that, but this is of independent interest/importance?

Took the liberty of using the PR also to do some tidying up of the module... and to fix the markup in `CHANGELOG`

... prompting me to revisit #1875 : I'd still push for deprecating `x+x≈x⇒x≈0` in favour of `+-identityˡ-unique`, otherwise if we think this lemma important enough, then perhaps it should be re-exported from where it's first definable, namely all the way up (down?) the hierarchy in `Algebra.Loop.Properties`...


jamesmckinna:
A general point about `Algebra.Properties.X` vs. `Algebra.Consequences.Setoid`: should this lemma be in `Properties.RingWithoutOne`, or proved first in `Consequences`, for a longish list of hypotheses?

src/Algebra/Properties/RingWithoutOne.agda
```diff
@@ -36,6 +36,12 @@ open AbelianGroupProperties +-abelianGroup public
   ; ⁻¹-∙-comm        to -‿+-comm
   )
 
+x+x≈x⇒x≈0 : ∀ x → x + x ≈ x → x ≈ 0#
```
Taneb, line 39: Can this be (under a different name) in `Algebra.Properties.Group`?

jamesmckinna: See preamble: this is legacy code, and was argued for in #1875 which introduced it.

My considered judgment  is now that that was an mistake, and in any case, that if we are to keep it, that it properly belongs under `Algebra.Properties.Loop` I think, with downstream re-export to `(Abelian)Group` under suitbale names and then finally here with this name. But for now, this is merely a cosmetic code motion from the original, so that `(Abelian)Group` properties of `RingWithoutOne` are kept separate from those involving distributivity and negation proper.

CHANGELOG.md
```diff
@@ -275,7 +279,7 @@ Additions to existing modules
   uv≈wx⇒yu∙v≈yw∙x       : ∀ y → (y ∙ u) ∙ v ≈ (y ∙ w) ∙ x
   uv≈wx⇒u∙vy≈w∙xy       : ∀ y → u ∙ (v ∙ y) ≈ w ∙ (x ∙ y)
   uv≈wx⇒yu∙vz≈yw∙xz     : ∀ y z → (y ∙ u) ∙ (v ∙ z) ≈ (y ∙ w) ∙ (x ∙ z)
-```
+  ```
```
JacquesCarette: Did you really mean to add these spaces here? I doubt it.

JacquesCarette (changes requested):

JacquesCarette:
> A general point about Algebra.Properties.X vs. Algebra.Consequences.Setoid: should this lemma be in Properties.RingWithoutOne, or proved first in Consequences, for a longish list of hypotheses?

I'm thinking `Properties.RingWithoutOne`. The only reason I would put it in `Consequences` is if the proof uses a 'weird' set of properties that amounts to an algebra without a name that is unlikely to see much use if it were created.

CHANGELOG.md
jamesmckinna: Yes, actually. The code/`agda` markdown should be offset as we're in a `*` bullet-pointed list, at least AFAIU?

JacquesCarette: I think that markdown rendering automatically indents inside a list.

jamesmckinna: Oh, OK (I've maybe learnt something), but still: consistency with every other entry?

JacquesCarette: Consistency (very) frequently wins such arguments...

MatthewDaggitt:
In the interest of getting a release out, I'm bumping this to v2.4

MatthewDaggitt (approved):

jamesmckinna:
@JacquesCarette are you happy to remove your caveat on this, and merge?

JacquesCarette (approved):

PR #2769, opened 2025-07-14
Add padRight properties to Data.Vec.Properties

e-mniang:
This PR adds several equational properties for padRight to Data.Vec.Properties, extending the standard vector operations toolkit with useful lemmas for reasoning about padding on the right.

**Added properties:**

-     `padRight-lookup`: accessing a padded vector retrieves the original values at valid indices

-     `padRight-map`: padRight commutes with `map`

-     `padRight-zipWith`: padRight commutes with `zipWith`

-    ` padRight-zipWith₁`: generalizes `padRight-zipWith` for different vector lengths

-     `padRight-take`: recovers the original prefix with `take`

-     `padRight-drop`: the suffix added by padding equals `replicate`

-    ` padRight-updateAt`: updates on padded vectors correspond to updates on the original vectors

Each property is proven by structural induction and uses standard lemmas such as `zipWith-replicate, map-replicate, and inject≤.`

---

Authored by : @jmougeot


src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,61 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+
+padRight-zipWith₁ : ∀ {p} (f : A → B → C) (m≤n : m ≤ n) (p≤m : p ≤ m)
+                    (a : A) (b : B) (xs : Vec A m) (ys : Vec B p) →
+                    zipWith f (padRight m≤n a xs) (padRight (≤-trans p≤m m≤n) b ys) ≡
+                    padRight m≤n (f a b) (zipWith f xs (padRight p≤m b ys))
+padRight-zipWith₁ {p} f m≤n p≤m a b xs ys = trans (cong (zipWith f (padRight m≤n a xs)) (padRight-trans p≤m m≤n b ys))
+                                            (padRight-zipWith f m≤n a b xs (padRight p≤m b ys))
+
+padRight-take : ∀ {p} (m≤n : m ≤ n) (a : A) (xs : Vec A m) (n≡m+p : n ≡ m + p) →
+                take m (subst (Vec A) n≡m+p (padRight m≤n a xs)) ≡ xs
```
jamesmckinna: When dealing with `Fin` or `Vec`, `subst` is (almost) always to be deprecated in favour of `cast`
```suggestion
padRight-take : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
                take m (cast n≡m+o (padRight m≤n a xs)) ≡ xs
```
where also:
* variable `o` is already declared, so should be used in preference to `p`, which has been used for a `Level` variable...
* the equality `n≡m+o` is declared *irrelevant*, because `cast` can't inspect it

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,61 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-take {m = zero} z≤n a [] refl = refl
+padRight-take {m = suc m} {p = p} (s≤s m≤n) a (x ∷ xs) refl = cong (x ∷_) (padRight-take {p = p} m≤n a xs refl)
+
+padRight-drop : ∀ {p} (m≤n : m ≤ n) (a : A) (xs : Vec A m) (n≡m+p : n ≡ m + p) →
+                drop m (subst (Vec A) n≡m+p (padRight m≤n a xs)) ≡ replicate p a
```
jamesmckinna: Ditto. re: `subst` and `cast`:
```suggestion
padRight-drop : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
                drop m (cast n≡m+o (padRight m≤n a xs)) ≡ replicate o a
```
where, again, `o` instead of `p`, and the equation marked as irrelevant.

jamesmckinna:
See also: #2770 . Even if the types of the `Base` operations are not (yet) changed, their properties should, where possible, be written to reflect that they only make *irrelevant* use of their `m ≤ n` arguments... UPDATED hmm, but it's not quite as simple as I'd hoped (and suggested as changes) *unless* we also make the changes proposed in #2770 which allow `padRight` to reduce more eagerly under weaker assumptions about the proofs of `m ≤ n`...

That being the case, I'll shout out to @JacquesCarette as to whether we go for this PR as is, and refactor downstream, or mark this blocking on #2770 ?

UPDATED: see also #2787

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,61 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-drop {m = zero} z≤n a [] refl = refl
+padRight-drop {m = suc m} {p = p} (s≤s m≤n) a (x ∷ xs) refl = padRight-drop {p = p} m≤n a xs refl
```
jamesmckinna: Because `cast` doesn't (*can't*) make use of its argument, these proofs should not do pattern-matching on `refl` either as proofs of `n≡m+o : n ≡ m + o`... but at the cost of needing one more lemma:
```suggestion
padRight-drop {m = zero}              _   a []       eq = cast-replicate eq a
padRight-drop {m = suc _} {n = suc _} m≤n a (x ∷ xs) eq = padRight-drop (s≤s⁻¹ m≤n) a xs (suc-injective eq)
```
where the following should appear before `padRight-drop` in the section on `replicate`...
```agda
cast-replicate : ∀ .(m≡n : m ≡ n) (x : A) → cast m≡n (replicate m x) ≡ replicate n x
cast-replicate {m = zero}  {n = zero}  _  _  = refl
cast-replicate {m = suc _} {n = suc _} eq x = cong (x ∷_) (cast-replicate (suc-injective eq) x)
```

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,61 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+                    (a : A) (b : B) (xs : Vec A m) (ys : Vec B p) →
+                    zipWith f (padRight m≤n a xs) (padRight (≤-trans p≤m m≤n) b ys) ≡
+                    padRight m≤n (f a b) (zipWith f xs (padRight p≤m b ys))
+padRight-zipWith₁ {p} f m≤n p≤m a b xs ys = trans (cong (zipWith f (padRight m≤n a xs)) (padRight-trans p≤m m≤n b ys))
+                                            (padRight-zipWith f m≤n a b xs (padRight p≤m b ys))
+
+padRight-take : ∀ {p} (m≤n : m ≤ n) (a : A) (xs : Vec A m) (n≡m+p : n ≡ m + p) →
+                take m (subst (Vec A) n≡m+p (padRight m≤n a xs)) ≡ xs
+padRight-take {m = zero} z≤n a [] refl = refl
+padRight-take {m = suc m} {p = p} (s≤s m≤n) a (x ∷ xs) refl = cong (x ∷_) (padRight-take {p = p} m≤n a xs refl)
```
jamesmckinna: Leading to the less strict pattern-matching:
```suggestion
padRight-take {m = zero}              _   a []       _  = refl
padRight-take {m = suc _} {n = suc _} m≤n a (x ∷ xs) eq = cong (x ∷_) (padRight-take (s≤s⁻¹ m≤n) a xs (suc-injective eq))
```

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,61 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-drop : ∀ {p} (m≤n : m ≤ n) (a : A) (xs : Vec A m) (n≡m+p : n ≡ m + p) →
+                drop m (subst (Vec A) n≡m+p (padRight m≤n a xs)) ≡ replicate p a
+padRight-drop {m = zero} z≤n a [] refl = refl
+padRight-drop {m = suc m} {p = p} (s≤s m≤n) a (x ∷ xs) refl = padRight-drop {p = p} m≤n a xs refl
+
+padRight-updateAt : ∀ (m≤n : m ≤ n) (xs : Vec A m) (f : A → A) (i : Fin m) (x : A) →
+                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
+                    padRight m≤n x (updateAt xs i f)
+padRight-updateAt (s≤s m≤n) (y ∷ xs) f zero x = refl
+padRight-updateAt (s≤s m≤n) (y ∷ xs) f (suc i) x = cong (y ∷_) (padRight-updateAt m≤n xs f i x)
```
jamesmckinna: And finally...
```suggestion
padRight-updateAt : ∀ .(m≤n : m ≤ n) (xs : Vec A m) (f : A → A) (i : Fin m) (x : A) →
                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
                    padRight m≤n x (updateAt xs i f)
padRight-updateAt {n = suc _} m≤n (y ∷ xs) f zero    x = refl
padRight-updateAt {n = suc _} m≤n (y ∷ xs) f (suc i) x = cong (y ∷_) (padRight-updateAt (s≤s⁻¹ m≤n) xs f i x)
```

jamesmckinna:
I should have minded my manners better, and thanked you for the PR before piling in with my nitpicks! These are all, in whatever form they eventually take, a useful contribution... and since `padRight` goes hand-in-hand with `truncate`, we should perhaps welcome a follow-up PR with the corresponding lemmas for `truncate` too?

e-mniang (commented):
Hi!
Thanks for your comments! I've made the necessary changes based on your feedback. I just had a small issue with the last point — the irrelevant part of the type signature.
Looking forward to your review of this version!

Great idea for the lemmas for `truncate`, will definitely look into that!

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,67 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+
+
+cast-replicate : ∀ .(m≡n : m ≡ n) (x : A) → cast m≡n (replicate m x) ≡ replicate n x
+cast-replicate {m = zero}  {n = zero}  _  _  = refl
+cast-replicate {m = suc _} {n = suc _} eq x = cong (x ∷_) (cast-replicate (suc-injective eq) x)
+
+padRight-drop : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
+                drop m (cast n≡m+o (padRight m≤n a xs)) ≡ replicate o a
+padRight-drop {m = zero}  z≤n a [] eq = cast-replicate eq a
+padRight-drop {m = suc _} {n = suc _} (s≤s m≤n) a (x ∷ xs) eq = padRight-drop m≤n a xs ((suc-injective eq))
```
jamesmckinna:
```suggestion
padRight-drop {m = suc _} {n = suc _} (s≤s m≤n) a (x ∷ xs) eq = padRight-drop m≤n a xs (suc-injective eq)
```
double parentheses?

src/Data/Vec/Properties.agda
```diff
@@ -1184,6 +1168,67 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+
+
+padRight-updateAt : ∀ (m≤n : m ≤ n) (xs : Vec A m) (f : A → A) (i : Fin m) (x : A) →
+                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
+                    padRight m≤n x (updateAt xs i f)
+padRight-updateAt {n = suc n} (s≤s m≤n) (y ∷ xs) f zero x = refl
+padRight-updateAt {n = suc n} (s≤s m≤n) (y ∷ xs) f (suc i) x = cong (y ∷_) ((padRight-updateAt m≤n xs f i x))
```
jamesmckinna:
```suggestion
padRight-updateAt {n = suc n} (s≤s m≤n) (y ∷ xs) f (suc i) x = cong (y ∷_) (padRight-updateAt m≤n xs f i x)
```
ditto.

src/Data/Vec/Properties.agda
jamesmckinna: Looks good now, but I'd bump the `cast-replicate` proof further up the module to the section on `replicate` from L1128 onwards...

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-refl : (a : A) (xs : Vec A n) → padRight ≤-refl a xs ≡ xs
+padRight-refl a []       = refl
+padRight-refl a (x ∷ xs) = cong (x ∷_) (padRight-refl a xs)
+
+padRight-replicate : (m≤n : m ≤ n) (a : A) → replicate n a ≡ padRight m≤n a (replicate m a)
+padRight-replicate z≤n       a = refl
+padRight-replicate (s≤s m≤n) a = cong (a ∷_) (padRight-replicate m≤n a)
+
+padRight-trans : ∀ {p} (m≤n : m ≤ n) (n≤p : n ≤ p) (a : A) (xs : Vec A m) →
+            padRight (≤-trans m≤n n≤p) a xs ≡ padRight n≤p a (padRight m≤n a xs)
```
jamesmckinna: As previously, please don't use `p` here, when `o` would conventionally be better.
```suggestion
padRight-trans : ∀ (m≤n : m ≤ n) (n≤o : n ≤ o) (a : A) (xs : Vec A m) →
            padRight (≤-trans m≤n n≤o) a xs ≡ padRight n≤o a (padRight m≤n a xs)
```

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-trans z≤n       n≤p       a []       = padRight-replicate n≤p a
+padRight-trans (s≤s m≤n) (s≤s n≤p) a (x ∷ xs) = cong (x ∷_) (padRight-trans m≤n n≤p a xs)
```
jamesmckinna: LIkewise
```suggestion
padRight-trans z≤n       n≤o       a []       = padRight-replicate n≤o a
padRight-trans (s≤s m≤n) (s≤s n≤o) a (x ∷ xs) = cong (x ∷_) (padRight-trans m≤n n≤o a xs)
```

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+                   (xs : Vec A m) (ys : Vec B m) →
+                   zipWith f (padRight m≤n a xs) (padRight m≤n b ys) ≡ padRight m≤n (f a b) (zipWith f xs ys)
+padRight-zipWith f z≤n a b [] [] = zipWith-replicate f a b
+padRight-zipWith f (s≤s m≤n) a b (x ∷ xs) (y ∷ ys) = cong (f x y ∷_) (padRight-zipWith f m≤n a b xs ys)
+
+padRight-zipWith₁ : ∀ {p} (f : A → B → C) (m≤n : m ≤ n) (p≤m : p ≤ m)
+                    (a : A) (b : B) (xs : Vec A m) (ys : Vec B p) →
+                    zipWith f (padRight m≤n a xs) (padRight (≤-trans p≤m m≤n) b ys) ≡
+                    padRight m≤n (f a b) (zipWith f xs (padRight p≤m b ys))
+padRight-zipWith₁ {p} f m≤n p≤m a b xs ys = trans (cong (zipWith f (padRight m≤n a xs)) (padRight-trans p≤m m≤n b ys))
```
jamesmckinna: Whatever else above regarding `m n o/p` ordering of the variables...
```suggestion
padRight-zipWith₁ f m≤n p≤m a b xs ys = trans (cong (zipWith f (padRight m≤n a xs)) (padRight-trans p≤m m≤n b ys))
```
Implicit `p` isn't even used here, so it's much better to omit it to make your code less brittle.

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-zipWith : ∀ (f : A → B → C) (m≤n : m ≤ n) (a : A) (b : B)
+                   (xs : Vec A m) (ys : Vec B m) →
+                   zipWith f (padRight m≤n a xs) (padRight m≤n b ys) ≡ padRight m≤n (f a b) (zipWith f xs ys)
+padRight-zipWith f z≤n a b [] [] = zipWith-replicate f a b
+padRight-zipWith f (s≤s m≤n) a b (x ∷ xs) (y ∷ ys) = cong (f x y ∷_) (padRight-zipWith f m≤n a b xs ys)
+
+padRight-zipWith₁ : ∀ {p} (f : A → B → C) (m≤n : m ≤ n) (p≤m : p ≤ m)
+                    (a : A) (b : B) (xs : Vec A m) (ys : Vec B p) →
+                    zipWith f (padRight m≤n a xs) (padRight (≤-trans p≤m m≤n) b ys) ≡
+                    padRight m≤n (f a b) (zipWith f xs (padRight p≤m b ys))
```
jamesmckinna: Ugh!
* `p` shouldn't be used here as an implicit; it's previously bound at module-level scope as a `variable` of type `Level`, while `o` is bound at type `Nat`, so use that instead
* what's the correct order of assumptions here? that `m ≤ n` *precede* `p ≤ m`, or that you instead order them so that `m ≤ n ≤ o` and then fix up the relevant assumption *uses* to reflect that?

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-zipWith₁ {p} f m≤n p≤m a b xs ys = trans (cong (zipWith f (padRight m≤n a xs)) (padRight-trans p≤m m≤n b ys))
+                                            (padRight-zipWith f m≤n a b xs (padRight p≤m b ys))
+
+padRight-take : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
+                take m (cast n≡m+o (padRight m≤n a xs)) ≡ xs
+padRight-take m≤n a [] p = refl
+padRight-take (s≤s m≤n) a (x ∷ xs) p = cong (x ∷_) (padRight-take m≤n a xs (suc-injective p))
```
jamesmckinna: Please don't use `p` here as a pattenr variable for an equation, especially given that it is already implicitly bound above at type Level... etc. 
Instead use `eq` or the assumption name you have already chosen in the lemma statement (reduce cognitive load on the reader as the implied design heuristic here)

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-drop : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
+                drop m (cast n≡m+o (padRight m≤n a xs)) ≡ replicate o a
+padRight-drop {m = zero}  z≤n a [] eq = cast-replicate eq a
+padRight-drop {m = suc _} {n = suc _} (s≤s m≤n) a (x ∷ xs) eq = padRight-drop m≤n a xs (suc-injective eq)
+
+padRight-updateAt : ∀ (m≤n : m ≤ n) (xs : Vec A m) (f : A → A) (i : Fin m) (x : A) →
+                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
+                    padRight m≤n x (updateAt xs i f)
+padRight-updateAt {n = suc n} (s≤s m≤n) (y ∷ xs) f zero x = refl
+padRight-updateAt {n = suc n} (s≤s m≤n) (y ∷ xs) f (suc i) x = cong (y ∷_) (padRight-updateAt m≤n xs f i x)
```
jamesmckinna:
```suggestion
padRight-updateAt {n = suc _} (s≤s m≤n) (y ∷ xs) f zero x = refl
padRight-updateAt {n = suc _} (s≤s m≤n) (y ∷ xs) f (suc i) x = cong (y ∷_) (padRight-updateAt m≤n xs f i x)
```
The actual `n` isn't used subsequently on the LHS, nor on the RHS, so omitting it in favour of a wildcard `_` makes the code less brittle.

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-take (s≤s m≤n) a (x ∷ xs) p = cong (x ∷_) (padRight-take m≤n a xs (suc-injective p))
+
+padRight-drop : ∀ (m≤n : m ≤ n) (a : A) (xs : Vec A m) .(n≡m+o : n ≡ m + o) →
+                drop m (cast n≡m+o (padRight m≤n a xs)) ≡ replicate o a
+padRight-drop {m = zero}  z≤n a [] eq = cast-replicate eq a
+padRight-drop {m = suc _} {n = suc _} (s≤s m≤n) a (x ∷ xs) eq = padRight-drop m≤n a xs (suc-injective eq)
+
+padRight-updateAt : ∀ (m≤n : m ≤ n) (xs : Vec A m) (f : A → A) (i : Fin m) (x : A) →
+                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
+                    padRight m≤n x (updateAt xs i f)
```
jamesmckinna: What's the right argument order here? A heuristic I learnt from @JacquesCarette would be: args that vary most should be rightmost, those that are invariant/vary less should be leftmost, together with the heuristic that says that argument order for lemmas should try to follow the argument order for the 'main' function being characterised, so here at least something like:
```suggestion
padRight-updateAt : ∀ (m≤n : m ≤ n) (x : A) (xs : Vec A m) (f : A → A) (i : Fin m) →
                    updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡
                    padRight m≤n x (updateAt xs i f)
```
with the position of `f` being moot, but perhaps should be rightmost to reflect *its* arg position in the function calls?

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
 toList-replicate zero    x = refl
 toList-replicate (suc n) x = cong (_ List.∷_) (toList-replicate n x)
 
+cast-replicate : ∀ .(m≡n : m ≡ n) (x : A) → cast m≡n (replicate m x) ≡ replicate n x
+cast-replicate {m = zero}  {n = zero}  _  _  = refl
+cast-replicate {m = suc _} {n = suc _} eq x = cong (x ∷_) (cast-replicate (suc-injective eq) x)
```
jamesmckinna: Alignment?
```suggestion
cast-replicate {m = zero}  {n = zero}  _  _ = refl
cast-replicate {m = suc _} {n = suc _} eq x = cong (x ∷_) (cast-replicate (suc-injective eq) x)
```

jamesmckinna (changes requested):
More nitpicks, I'm afraid, but otherwise looks good to go...
... but you'll need to write `CHANGELOG` entries as well. Suggest you do this once the reset after the v2.3 release is complete, though...

... UPDATED which has now taken place. So, if you look at `CHANGELOG.md`, you'll see under `Additions to existing modules`... please add the lemma names with their types, and you can see existing versions under eg `CHANGELOG/v2.2.md` to see formatting conventions etc.

CHANGELOG.md
```diff
@@ -444,6 +444,24 @@ Additions to existing modules
   fromList-reverse : (xs : List A) → (fromList (List.reverse xs)) ≈[ List.length-reverse xs ] reverse (fromList xs)
 
   fromList∘toList : ∀  (xs : Vec A n) → fromList (toList xs) ≈[ length-toList xs ] xs
+
+  padRight-lookup : ∀ {m n} → (m≤n : m ≤ n) (a : A) (xs : Vec A m) (i : Fin m) → lookup (padRight m≤n a xs) (inject≤ i m≤n) ≡ lookup xs i
```
jamesmckinna:
```suggestion
  padRight-lookup : (m≤n : m ≤ n) (a : A) (xs : Vec A m) (i : Fin m) → lookup (padRight m≤n a xs) (inject≤ i m≤n) ≡ lookup xs i
```
We tend to suppress ('obviously') inferrable initial quantifier prefixes from the types of `CHANGELOG` entries.
And in the rest of the additions here...

src/Data/Vec/Properties.agda
```diff
@@ -1184,13 +1168,71 @@ toList-replicate : ∀ (n : ℕ) (x : A) →
...
+padRight-updateAt {n = suc _} (s≤s m≤n) x (y ∷ xs) f zero = refl
+padRight-updateAt {n = suc _} (s≤s m≤n) x (y ∷ xs) f (suc i) = cong (y ∷_) (padRight-updateAt m≤n x xs f i)
+
 ------------------------------------------------------------------------
 -- iterate
 
 iterate-id : ∀ (x : A) n → iterate id x n ≡ replicate n x
 iterate-id x zero    = refl
 iterate-id x (suc n) = cong (_ ∷_) (iterate-id (id x) n)
-
```
jamesmckinna, line 1193: I think we should keep this blank line? How did this get deleted?

CHANGELOG.md
e-mniang: Hi! Thanks for the correction. I looked at v2.2 and the ones already on that version, and thought it could be done."

src/Data/Vec/Properties.agda
e-mniang: must’ve been a mistype on my part, sorry about that!

jamesmckinna:
As I already wrote on #2795 ...
Once the backports https://github.com/agda/agda-stdlib/pull/2799 from v2.3 have been merged into `master`, it will make sense to fix up the `CHANGELOG` conflicts, merge `master` into this branch, and then look at the suggested fixes.

UPDATED: I resolved the merge conflicts (which were easy), so I'll re-review, but this looks good now.

MatthewDaggitt (approved):

PR #2774, opened 2025-07-18
Fix breaking introduction of variable in `Semiring.Primality`

### LINKED ISSUE #2772: Maybe avoid  introducing almost undefined  arguments?
The master branch of July 15, 2025 has
```
module Algebra.Properties.Semiring.Primality  {a ℓ} (R : Semiring a ℓ) where
...
private  variable    x p : A
...
Irreducible⇒≉0 : 0# ≉ 1# → Irreducible p → p ≉ 0#
...
```

The old version of   ``Irreducible⇒≉0 : 0# ≉ 1# → (∀ {p } → Irreducible p → p ≉ 0#)     -- (I)``
looks easier to read and understand, to my mind.
Because it looks like a mathematical statement.
(and the change is backwards incompatible. Is it in CHANGELOG ? ).

I had  under lib-2.2  ``a≉0   = Irreducible⇒≉0 0≉1 {a} irred-a`` 
Now Agda-2.8.0+master  report "implicit instead of explicit".
I remove curly braces. And it reports "unequal terms".
Then I remove `a`. Now it is type-checked.
Then I look into Standard library ...

It is frightening to see a variable name which has come somehow from nowhere.
I recall there are many such places in Standard library.
?

MatthewDaggitt:
Fixes accidental breaking change in https://github.com/agda/agda-stdlib/issues/2772.

### DIFF
--- a/src/Algebra/Properties/Semiring/Primality.agda
+++ b/src/Algebra/Properties/Semiring/Primality.agda
@@ -22,8 +22,7 @@ open import Algebra.Properties.Semiring.Divisibility R
 
 private
   variable
-    x p : A
-
+    x : A
 
 ------------------------------------------------------------------------
 -- Re-export primality definitions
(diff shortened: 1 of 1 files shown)


jamesmckinna (approved):
For the sake of v2.3, I'm happy to revert a prior change.

But the general principle of avoiding implicits in favour of `variable`s is still a good one, and will, from time to time, introduce `breaking` changes because of re-ordering of telescopic bindings.

User code which relies on explicit instantiation of implicit arguments is inherently brittle, unless this points to the need to make the argument explicit, which should be flagged as a separate issue?

I hope that for v3.0 we can make all of this more systematic?

MatthewDaggitt:
I agree with all your points, particularly about how using implicits widely is very fragile! But I do think that we should be careful in introducing `variable`s in minor version bumps and try to prioritise making breaking ones in major versions bumps nonetheless.

Issue #2771, opened 2025-07-15
[ refactor ] `Algebra.Structures.IsQuasiring`?


jamesmckinna:
Why in this definition is the multiplicative structure (L670-L672)
https://github.com/agda/agda-stdlib/blob/5c641176b3008e0a21419fcbe338f9bf3a0df6fd/src/Algebra/Structures.agda#L667-L674
not telescoped back into 
```agda
*-isMonoid : IsMonoid * 1#
```
Instead of the more involved unbundling of such a field into its component fields?

Maybe this has come up before, but I can't quite now see why this choice was made...

Ah... the problem of re-`open`ing the substructures to then have to figure out which underlying `setoid` is in play... fair enough, but it looks a bit weird sometimes!

JacquesCarette:
So maybe this warrants a comment in the file, so that future maintainers don't end up tripping on the same issue?

jamesmckinna:
> So maybe this warrants a comment in the file, so that future maintainers don't end up tripping on the same issue?

And I should learn this lesson and transfer it over to #2763 and I should track down the original issue which drew attention to the problem... #1617 I think (with shades of #898 too)

jamesmckinna:
Badged `status: duplicate` as subsumed by #2781 (and how *that* got fixed in #1684 )

PR #2781, opened 2025-07-22
[ add ] commentary explaining the `IsXRing` design problem (issues #1617 and #2771)

jamesmckinna:
> So maybe this warrants a comment in the file, so that future maintainers don't end up tripping on the same issue? 

 _Originally posted by @JacquesCarette in [#2771](https://github.com/agda/agda-stdlib/issues/2771#issuecomment-3099081788)_

This PR precisely attempts to offer such commentary in `Algebra.Structures`, but it's not clear that this is the right place to do it, nor that the 'first' instance of the phenomenon, `IsNearSemiring`, is where users will find such explanation useful.

Meta-issue: when we have a substantive `library-design` issue such as #1617, and its resolution in #1684 , how/where should we document it *beyond* the comment trail on GitHub?

### DIFF
--- a/doc/README/Design/Hierarchies.agda
+++ b/doc/README/Design/Hierarchies.agda
@@ -9,6 +9,7 @@
 module README.Design.Hierarchies where
 
 open import Data.Sum.Base using (_⊎_)
+open import Data.Product.Base using (_×_)
 open import Level using (Level; _⊔_; suc)
 open import Relation.Binary.Core using (_Preserves₂_⟶_⟶_)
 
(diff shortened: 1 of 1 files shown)


JacquesCarette:
That Meta-issue why I'm keen to start on the stdlib design paper! We do have some design documentation in `doc/README/Design/*`.

The 'real' solution to this problem would have us pull out the `Setoid` over which the structures are defined as a parameter instead of the 'raw' approach we're taking now. More generally, I think support for what others are calling "displayed algebra" will become necessary. No idea if @AndrasKovacs is working on that now, but he's certainly one of the leading candidates!

jamesmckinna:
I did look in the `README.Design.*` tree, and couldn't really find what I needed. But perhaps the text I've written here should be moved there? Ironically (sic), I find I increasingly don't bother to read the `README`s, and as they aren't re-generated by `make test`, although they are re-checked at least (?), there is the risk that their contents drifts from our current internal mental model of what is/should be going on...?

JacquesCarette:
I'd say *copied* there rather than moved. There are times when redundancy is good.

jamesmckinna:
I'll convert to `DRAFT` while I do the copying over to `README`?
UPDATED: turns out a lot more material was needed in order to be able to typecheck the `Hierarchies` module!
Now, I'll ping again and then we can get this off the docket provided you're both still happy.

MatthewDaggitt (approved):

jamesmckinna:
@JacquesCarette @gallais are you happy to merge now, with the additional commentary in `README.Design.Hierarchies`?

JacquesCarette (approved):

Issue #2746, opened 2025-06-25
[ refactor ] application, and perhaps even proof, of `Data.Fin.Properties.pigeonhole`?


jamesmckinna:
Currently, we have https://github.com/agda/agda-stdlib/blob/9fca9dcdf6ac75f219100be4489961ad1f1390c4/src/Data/Fin/Properties.agda#L1006-L1023

Regarding the two lemmas which follow `pigeonhole`, we have:
* `injective⇒≤` proved by a separate (and complicatedly `contraPositive`) argument from that of `pigeonhole`, despite (essentially) recapiltulating the same/similar call-pattern
* from which `<⇒notInjective` then follows by (another contrapositive heavy!) application of `ℕ.≤⇒≯`

By contrast, we could reorder the lemmas so that the second follows as an immediate corollary:
* ```agda
  <⇒notInjective : ∀ {f : Fin m → Fin n} → n ℕ.< m → ¬ (Injective _≡_ _≡_ f)
  <⇒notInjective {f = f} n<m inj = let i , j , i<j , fᵢ≡fⱼ = pigeonhole n<m f in <-irrefl (inj fᵢ≡fⱼ) i<j
  ```
* followed by the 'opposite' contrapositive argument for the first
   ```agda
   injective⇒≤ : ∀ {f : Fin m → Fin n} → Injective _≡_ _≡_ f → m ℕ.≤ n
  injective⇒≤ = ℕ.≮⇒≥ ∘ flip <⇒notInjective
  ```

Regarding the proof of `pigeonhole` itself (the present version is merely a cosmetic refactoring of @MatthewDaggitt 's original in https://github.com/agda/agda-stdlib/pull/387 ), there is (still, to my mind) something strange about the argument: namely that the argument in the second `no` branch makes a recursive call on `pigeonhole m<n`, but now with a function *which has no relation to the original `f`*... (except perhaps indirectly via `f₀≢fₖ`). How is this possible? Can this be simplified further?

JacquesCarette:
I agree that refactoring the two lemmas that follow is a good idea.

Note that the recursive call is, in a very hidden way, about `f` ! It 'punches out' a part of `f`'s codomain. You have to delta-expand quite a lot to see this. 

[...]

PR #2782, opened 2025-07-25
[ refactor ] `Data.Fin.Properties` following #2746

jamesmckinna:
UPDATED: 
* no `CHANGELOG` required! Only existing proofs have been refactored. 
* issue below resolved moderately successfully via @andreasabel 's trick with `flip contradiction` cf. #2784 

NB. I tried to simplify the proof of `injective⇒existsPivot` to use `<⇒notInjective` instead, but it causes a weird explicit/implicit type inference error when trying to obtain a contradiction from `f∘inject!-injective` and `<⇒notInjective`...:
```agda
injective⇒existsPivot {f = f} f-injective i
  with any? (λ j → j ≤? i ×-dec i ≤? f j)
... | yes result = result
... | no ¬result = contradiction f∘inject!-injective (<⇒notInjective {f = f∘inject! } (ℕ.n<1+n (toℕ i)))
  where
  ...
```
yields
```agda
/home/james/RESEARCH/Agda-stdlib-dev-v2.3/src/Data/Fin/Properties.agda:1077.55-104: error: [UnequalHiding]
f∘inject! _x_4739 ≡ f∘inject! _y_4740 → _x_4739 ≡ _y_4740 !=
{x y : Fin (toℕ (suc i))} → f∘inject! x ≡ f∘inject! y → x ≡ y
because one is an implicit function type and the other is an
explicit function type
<⇒notInjective {f = f∘inject!} (ℕ.n<1+n (toℕ i)) has type
¬ (f∘inject! _x_4739 ≡ f∘inject! _y_4740 → _x_4739 ≡ _y_4740)
```
(this kind of thing *really* aggravates me: we have two sub-proofs of `a : A` and `¬a : ¬ A` here, with `A = Injective _≡_ _≡_ f∘inject!`, but (re-)typechecking `contradiction a ¬a` causes a type error... :-( See https://github.com/agda/agda/issues/8026

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -46,7 +46,8 @@ open import Relation.Binary.PropositionalEquality.Properties as ≡
   using (module ≡-Reasoning)
 open import Relation.Nullary.Decidable as Dec
   using (Dec; _because_; yes; no; _×-dec_; _⊎-dec_; map′)
-open import Relation.Nullary.Negation.Core using (¬_; contradiction)
+open import Relation.Nullary.Negation.Core
+  using (¬_; contradiction; contradiction′)
 open import Relation.Nullary.Reflects using (Reflects; invert)
 open import Relation.Unary as U
   using (U; Pred; Decidable; _⊆_; Satisfiable; Universal)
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

JacquesCarette:
I'm fairly sure the reason for the error is that Agda eagerly inserts leading implicits but does not insert any past an explicit parameter.  So 'indexed' cases such as this would possibly need its own version of 'indexed contradiction`.

jamesmckinna:
> I'm fairly sure the reason for the error is that Agda eagerly inserts leading implicits but does not insert any past an explicit parameter. So 'indexed' cases such as this would possibly need its own version of 'indexed contradiction`.

Well... maybe, but that's *bad*? (from a user/UX perspective? it *might* be tolerable from a developer perspective, but I don't think it *should* be...)

JacquesCarette:
The follow-ups on the Agda issue shed quite a lot of light on this (and confirm my rough guess). Andreas' latest message agrees that the UX is bad, but that there are non-trivial technical reasons for it. And a nifty work-around!

jamesmckinna:
~~Blocking on #2784 for the time being.~~ Removed the block, so that the use of `flip contradiction` can instead be refactored in the course of *that* PR...

jamesmckinna:
Ha! I'll fix the merge conflict once #2784 has merged. DONE.

MatthewDaggitt (approved):

PR #2748, opened 2025-06-26
Proved sorted permutations are equal

MatthewDaggitt:
The main result. Would like to try and sneak this into v2.3 for @onestruggler.

Have also added an explanation of the choice of propositional permutation in `SortingAlgorithm` and a lemma that proves the setoid version.

Issue #2346, opened 2024-04-08
Deprecate existing definition of `⊥-elim` in favour of its irrelevant version `⊥-elim-irr`


jamesmckinna:
I further conjecture on #2329 that all uses of `⊥-elim` can be replaced by their irrelevant version `⊥-elim-irr`... so question for the maintainers, are we prepared to embrace such a (breaking?) change to the type of `⊥-elim`?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/issues/2243#issuecomment-2021111222_
            
This v3.0 issue concerns the `breaking` change not undertaken as part of #2243 . 

With it, among other things, might go the unification (also `breaking`) of the types of `contradiction` and `contradiction-irr` in `Relation.Nullary.Negation.Core` and a more systematic treatment of 'computational irrelevance of negated propositions'... considered in #2199.

PR #2783, opened 2025-07-25
[ refactor ] make `n≢i : n ≢ toℕ i` argument to `lower₁` irrelevant

jamesmckinna:
PR #2748 made me rethink whether `Data.Fin.Base.lower₁` is fit-for-purpose (as an inverse to `inject₁`), or could instead be deprecated in favour of the new `Data.Fin.Base.lower`...

This PR doesn't entirely solve that issue, but does prove the two definitions extensionally equal on their domains, as a consequence, perhaps more importantly, of weakening the type of `lower₁` so that its precondition is made irrelevant.

Two (possibly more downstream) knock-on consequences:
* ~~lemma `lower₁-¬0≢0 : ∀ {ℓ} {A : Set ℓ} → .(0 ≢ 0) → A` encapsulates a repeated pattern of (`¬-recompute`) reasoning, which ideally would be made `private`, but is needed in *both* `Base` and `Properties`... so, reluctantly, has been added;~~ avoided in favour of #2785 on which this PR is now `blocked`.
* many of the proofs about `lower₁` could/should be simplified by delegation to those for  `lower`...?
* #2790 `Data.Fin.Base.punchOut` and its properties could similarly be weakened by making its `i≢j : i ≢ j` argument irrelevant! What else might be susceptible to this kind of refactoring?

NB. As observed/observable in `README.Data.Fin.Relation.Unary.Top`, we can actually avoid having *any* uses of `lower₁` in the library, so `deprecation` seems possible/desirable #2786 

UPDATED: no longer `blocked` on #2785 .

### DIFF
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -19,8 +19,7 @@ open import Level using (0ℓ)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; _≢_; refl; cong)
 open import Relation.Binary.Indexed.Heterogeneous.Core using (IRel)
-open import Relation.Nullary.Negation.Core using (¬_; contradiction)
-open import Relation.Nullary.Recomputable using (¬-recompute)
+open import Relation.Nullary.Negation.Core using (¬_; contradiction-irr)
 
 private
   variable
@@ -117,11 +116,8 @@ inject≤ {n = suc _} (suc i) m≤n = suc (inject≤ i (ℕ.s≤s⁻¹ m≤n))
 
 -- lower₁ "i" _ = "i".
 
-lower₁-¬0≢0 : ∀ {ℓ} {A : Set ℓ} → .(0 ≢ 0) → A
-lower₁-¬0≢0 0≢0 = contradiction refl (¬-recompute 0≢0)
-
 lower₁ : ∀ (i : Fin (suc n)) → .(n ≢ toℕ i) → Fin n
-lower₁ {zero}  zero    ne = lower₁-¬0≢0 ne
+lower₁ {zero}  zero    ne = contradiction-irr refl ne
 lower₁ {suc n} zero    _  = zero
 lower₁ {suc n} (suc i) ne = suc (lower₁ i (ne ∘ cong suc))
 
@@ -257,7 +253,7 @@ opposite {suc n} (suc i) = inject₁ (opposite i)
 -- McBride's "First-order unification by structural recursion".
 
 punchOut : ∀ {i j : Fin (suc n)} → i ≢ j → Fin n
-punchOut {_}     {zero}   {zero}  i≢j = contradiction refl i≢j
+punchOut {_}     {zero}   {zero}  i≢j = contradiction-irr refl i≢j
 punchOut {_}     {zero}   {suc j} _   = j
 punchOut {suc _} {suc i}  {zero}  _   = zero
 punchOut {suc _} {suc i}  {suc j} i≢j = suc (punchOut (i≢j ∘ cong suc))
--- a/src/Data/Fin/Base.agda
+++ b/src/Data/Fin/Base.agda
@@ -20,6 +20,7 @@ open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.PropositionalEquality.Core using (_≡_; _≢_; refl; cong)
 open import Relation.Binary.Indexed.Heterogeneous.Core using (IRel)
 open import Relation.Nullary.Negation.Core using (¬_; contradiction)
+open import Relation.Nullary.Recomputable using (¬-recompute)
 
 private
   variable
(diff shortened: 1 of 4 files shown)


src/Data/Fin/Base.agda
```diff
@@ -116,8 +117,11 @@ inject≤ {n = suc _} (suc i) m≤n = suc (inject≤ i (ℕ.s≤s⁻¹ m≤n))
 
 -- lower₁ "i" _ = "i".
 
-lower₁ : ∀ (i : Fin (suc n)) → n ≢ toℕ i → Fin n
-lower₁ {zero}  zero    ne = contradiction refl ne
+lower₁-¬0≢0 : ∀ {ℓ} {A : Set ℓ} → .(0 ≢ 0) → A
```
JacquesCarette: This seems wildly contorted to me: under any false assumption (not just 0 ≢ 0) you can derive anything you want. Surely this already exists?

src/Data/Fin/Base.agda
```diff
@@ -116,8 +117,11 @@ inject≤ {n = suc _} (suc i) m≤n = suc (inject≤ i (ℕ.s≤s⁻¹ m≤n))
...
+lower₁-¬0≢0 0≢0 = contradiction refl (¬-recompute 0≢0)
+
+lower₁ : ∀ (i : Fin (suc n)) → .(n ≢ toℕ i) → Fin n
+lower₁ {zero}  zero    ne = lower₁-¬0≢0 ne
```
JacquesCarette: The previous proof for this case was much clearer. Maybe what we need is a `contradiction` that is happy to have its second argument be irrelevant?

src/Data/Fin/Properties.agda
```diff
@@ -506,30 +506,30 @@ inject!-< {suc n} {suc i} (suc k) = s≤s (inject!-< k)
 -- lower₁
 ------------------------------------------------------------------------
 
-toℕ-lower₁ : ∀ i (p : n ≢ toℕ i) → toℕ (lower₁ i p) ≡ toℕ i
-toℕ-lower₁ {ℕ.zero}  zero    p = contradiction refl p
-toℕ-lower₁ {ℕ.suc m} zero    p = refl
-toℕ-lower₁ {ℕ.suc m} (suc i) p = cong ℕ.suc (toℕ-lower₁ i (p ∘ cong ℕ.suc))
+toℕ-lower₁ : ∀ i .(p : n ≢ toℕ i) → toℕ (lower₁ i p) ≡ toℕ i
+toℕ-lower₁ {ℕ.zero}  zero    0≢0 = lower₁-¬0≢0 0≢0
```
JacquesCarette: Comments above would carry-over to here, i.e. to use `irrelevant-contradiction` (or whatever it gets called).

src/Data/Fin/Properties.agda
```diff
@@ -561,6 +561,17 @@ lower-injective {n = suc n} zero    zero    eq = refl
 lower-injective {n = suc n} (suc i) (suc j) eq =
   cong suc (lower-injective i j (suc-injective eq))
 
+lower₁≗lower : ∀ (i : Fin (suc n)) .(n≢i : n ≢ toℕ i) →
```
JacquesCarette: These lemmas feel odd: basically saying `lower₁` is the same as `lower` with 2nd argument an utter mess. Can't that argument be replaced with something much more generic, i.e. because here too we shouldn't care exactly what proof we get of that fact?

JacquesCarette (changes requested):

src/Data/Fin/Base.agda
jamesmckinna: The point was that the specialised version of `contradiction refl <x≢x>` clutters the code of `lower₁` *and* its properties, so I wanted a canned version for each of those irritatting, but unavoidable, impossible branches. But maybe this solution isn't the one you care for... but I don't find `contradiction refl ne` any  more perspicuous!

src/Data/Fin/Base.agda
jamesmckinna: I did originally have such a thing when refactoring `Relation.Nullary.Negation.Core`, but IIRC, the fact that that meant that there were *3* distinct types for `contradiction` was ... not acceptable? Cf. #2346 

This was one source for the 'Harrop' analysis of `Relation.Nullary.Recomputable`, but exposing that for the sake of the common case of negated propositions also seems like too-heavy a hammer. 

Suggestions welcome!

src/Data/Fin/Properties.agda
jamesmckinna: Maybe, but we want something that will typecheck with `toℕ i < n` in a context where all we know is `n≢i : n ≢ toℕ i`... the fact that the actual form is irrelevant is part of the reason to want dotted hypotheses, but we still have to be able to instantiate them!

Is this an argument in favour of irrelevant *instance*s instead? A more drastic change in the type signature, though...

src/Data/Fin/Base.agda
jamesmckinna: But, on reflection, I think you're right. Will open a separate PR to make this change to the type of `contradiction-irr`... #2785

src/Data/Fin/Base.agda
jamesmckinna: Now avoided by reverting the addition in favour of using `contradiction-irr` for the updated type of that lemma.

src/Data/Fin/Properties.agda
jamesmckinna: In a way, the two lemmas are really the justification for subsequent deprecation of `lower₁`, but I held off doing so for now. Happy to be nudged towards that path, though!?

JacquesCarette: If these lemmas are, in a sense, meta-theorems that justify something else, but not really important / useful on their own, then I'd erase them from the library and put them only in the github-level comments for the PR that deprecates `lower₁`

JacquesCarette (commented):
Modulo my comment on whether we even need those extra 2 lemmas, I'm now happy with this.

src/Data/Fin/Properties.agda
jamesmckinna: I'll lift these out to the comments in #2786 , but in the meantime, cf. `Data.Fin.Properties.{punchOut-cong|punchOut-cong′}` which is engaged in similar shenanigans, esp. eg in their deployment in `Data.Fin.Permutation`, which I might hope could be drastically simplified by making `punchOut` similarly take an irrelevant argument.

jamesmckinna:
Revisiting this, I'm happy to badge this as `breaking`/v3.0, given that the intensional behaviour of the function is being changed.

JacquesCarette (approved):

jamesmckinna:
Given all the drama about irrelevance at the moment, I'm going to mark this as `blocked by-issue`...

jamesmckinna:
And now, in the spirit of clearing things off the docket, I'm going to close for the time being, in favour of further discussion on #2941 (and perhaps generalising that issue somewhat?).

Issue #2784, opened 2025-07-26
[ breaking ] Flip arguments to `Relation.Nullary.negation.Core.contradiction`


jamesmckinna:
> Workaround: You would not get this problem if `contradiction` had the opposite argument ordering:
> ```
> contradiction : ¬ X → X → Whatever
> contradiction ¬a a with () ← ¬a a
> 
> test : Whatever
> test = contradiction ¬∀P ∀P
> ```
> Here, the equation `¬ X = ¬ (∀ {a} → P a)` is produced which solves correctly to `X = ∀ {a} → P a`.
> 
>  

 _Originally posted by @andreasabel in [#8026](https://github.com/agda/agda/issues/8026#issuecomment-3117932494)_

My recent woes in #2782 and my followup issue https://github.com/agda/agda/issues/8026 on the main agda tracker lead me to the somewhat uncomfortable conclusion that it would be better to break the current API for `contradiction` by flipping its arguments, or else, perhaps more modestly, but more of a redundancy/Fairbairn threshold code smell, that we give `flip contradiction` a first-class name for such situations?

JacquesCarette:
I think giving `flip contradiction` a first class name is a better idea. Here it is Agda itself which forces our hand. Higher concerns (redundancy, Fairbairn threshold) get derailed as 'not applicable' since `contradiction` and `flip contradiction` *are not equivalent*!!

jamesmckinna:
Two followups:
* can we somehow (sensibly) reconcile the addition/modification with #2785 
* what name for this new lemma!? Suggestions welcome! I'd tried `contradictionᵒ` before... but pushback on that, including from you!

gallais:
> what name for this new lemma!?

EMP, short for Exploding Modus Ponens? 🤭

jamesmckinna:
> > what name for this new lemma!?
> 
> EMP, short for Exploding Modus Ponens? 🤭

Oh! I want to be angry and humourless about this, but it has a certain something...
... that said, I'm still tempted to have a `flip`ped `contradiction-irr` (with a new name, such as @JacquesCarette 's `irrelevant-contradiction`), and use that to drive everything? On second thoughts, coupling the two things is a terrible idea, but having three distinct combinators also seems bad!

MatthewDaggitt:
[...]

PR #2785, opened 2025-07-26
[ refactor ] weaken type of `Relation.Nullary.Negation.Core.contradiction-irr`

jamesmckinna:
For justification, see `CHANGELOG`: this is really a consequence of #2199 and subsequent refactorings of `Relation.Nullary.Recomputable{.Properties}`. Cf. #2346 which might be worth re-opening?

Very little is affected, except that `contradiction` now needs to be eta-expanded because there is no subtyping between `A → B` and `.A → B`.

For candidate use, see the discussion in #2783 .

### DIFF
--- a/src/Relation/Nullary/Negation/Core.agda
+++ b/src/Relation/Nullary/Negation/Core.agda
@@ -47,11 +47,11 @@ _¬-⊎_ = [_,_]
 ------------------------------------------------------------------------
 -- Uses of negation
 
-contradiction-irr : .A → ¬ A → Whatever
+contradiction-irr : .A → .(¬ A) → Whatever
 contradiction-irr a ¬a = ⊥-elim-irr (¬a a)
 
 contradiction : A → ¬ A → Whatever
-contradiction a = contradiction-irr a
+contradiction a ¬a = contradiction-irr a ¬a
 
 contradiction₂ : A ⊎ B → ¬ A → ¬ B → Whatever
 contradiction₂ (inj₁ a) ¬a ¬b = contradiction a ¬a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Non-backwards compatible changes
 Minor improvements
 ------------------
 
+* The type of `Relation.Nullary.Negation.Core.contradiction-irr` has been further
+  weakened so that the negated hypothesis `¬ A` is marked as irrelevant. This is
+  safe to do, in view of `Relation.Nullary.Recomputable.Properties.¬-recompute`.
+
 Deprecated modules
 ------------------
 


JacquesCarette (approved):
Minor but important change. Well spotted.

jamesmckinna:
> Minor but important change. Well spotted.

Thanks! But mostly I'm annoyed with myself that this kind of downstream consequence of my introduction and refactoring of `Recomputable` could/should have been done more or less straight after #2199 , but I was too tired/distracted to follow through, or remember to loop back to these things until now. Better late than never though!

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -47,11 +47,11 @@ _¬-⊎_ = [_,_]
 ------------------------------------------------------------------------
 -- Uses of negation
 
-contradiction-irr : .A → ¬ A → Whatever
+contradiction-irr : .A → .(¬ A) → Whatever
 contradiction-irr a ¬a = ⊥-elim-irr (¬a a)
 
 contradiction : A → ¬ A → Whatever
-contradiction a = contradiction-irr a
+contradiction a ¬a = contradiction-irr a ¬a
```
MatthewDaggitt, line 54: Was this changed forced? And if so, is this therefore a breaking change?

jamesmckinna: Two good questions, to which my evasive, equivocating answers would be: "yes, but...; no, but..."
* re: the change itself: yes, the eta-expansion was forced because of the (lack of!) subtyping issue between `A → B` and `.A → B`, BUT the original definition could have been written in eta-expanded form, with equivalent semantics, so it's really an artefact of the (my!) lazy choice not to have done so originally...;
* re: is it `breaking`: I would argue no, in terms of the behaviour of the exported definition of `contradiction` itself, as there is no difference to the type, or uses, or call-sites of `contradiction`; as to `contradiction-irr`, yes, indeed so as to enable #2784 #2786 #2790 downstream, BUT I would argue that it is a `bug` that we didn't follow through on the consequences of #2199 

As in the `CHANGELOG`, I regard this as a `minor improvement `: non-breaking, but arguably better move the commentary to `non-backwards compatible` (at least for the downstream consequences, because the unfolding behaviour of the proposed revised function definitions changes subtly)?

While the suggestion been punted before, I might still argue that the type of `contradiction` itself should be changed to `.A → .(¬ A) → Whatever`, but I thought that would be a step too far here... but perhaps one way out is simply to agree here and now that that is what we will do for v3.0 (and get rid of `contradiction-irr` completely!), and I'll move the milestone on the downstream PRs, but I don't want to have to relitigate these things for a third time then... ;-)

MatthewDaggitt (approved):

PR #2793, opened 2025-07-30
[ refactor ] `Data.Fin.Properties.decFinSubset`

jamesmckinna:
Cherry-picked from #2744 . *Almost* purely cosmetic:
* introduces `variable`s. NB. cf. #2774  for problems with *non-prenex* use of such things
* simplifies `all?` to avoid use of `decFinSubset` (idiotic complexity blow-up)
* tidies up proof of latter
* no `CHANGELOG`

NB.
* local lemma `DecFinSubset.cons` makes successful use of `λ-` and `$-`, but I don't seem able to do so for `Q⊆P⇒Q⊆ₛP` thanks to unsolved metas!?
* yet another `flip contradiction`! #2784

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -1039,20 +1039,20 @@ private
 -- If a decidable predicate P over a finite set is sometimes false,
 -- then we can find the smallest value for which this is the case.
 
-¬∀⟶∃¬-smallest : ∀ n {p} (P : Pred (Fin n) p) → Decidable P →
+¬∀⇒∃¬-smallest : ∀ n {p} (P : Pred (Fin n) p) → Decidable P →
                  ¬ (∀ i → P i) → ∃ λ i → ¬ P i × ((j : Fin′ i) → P (inject j))
-¬∀⟶∃¬-smallest zero    P P? ¬∀P = contradiction (λ()) ¬∀P
-¬∀⟶∃¬-smallest (suc n) P P? ¬∀P with P? zero
+¬∀⇒∃¬-smallest zero    P P? ¬∀P = contradiction (λ()) ¬∀P
+¬∀⇒∃¬-smallest (suc n) P P? ¬∀P with P? zero
 ... | false because [¬P₀] = (zero , invert [¬P₀] , λ ())
 ... | true  because  [P₀] = map suc (map id (∀-cons (invert [P₀])))
-  (¬∀⟶∃¬-smallest n (P ∘ suc) (P? ∘ suc) (¬∀P ∘ (∀-cons (invert [P₀]))))
+  (¬∀⇒∃¬-smallest n (P ∘ suc) (P? ∘ suc) (¬∀P ∘ (∀-cons (invert [P₀]))))
 
 -- When P is a decidable predicate over a finite set the following
 -- lemma can be proved.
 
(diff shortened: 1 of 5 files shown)


src/Data/Fin/Properties.agda
```diff
@@ -54,10 +54,13 @@ open import Relation.Unary.Properties using (U?)
 
 private
   variable
-    a : Level
+    a p q : Level
     A : Set a
     m n o : ℕ
     i j : Fin n
+    P : Pred (Fin n) p
+    Q : Pred (Fin n) q
```
MatthewDaggitt: These predicates aren't allowed to be made variables by the rules in the style guide?

MatthewDaggitt (changes requested):
Otherwise I approve

src/Data/Fin/Properties.agda
jamesmckinna: Strict, but fair... sigh.

I'm *strongly* tempted to raise a v3.0 issue to extend the `variable` section to include not just quantification over the `data` being defined, but also over `Pred`s and `Rel`s defined over it.
Else, we should refactor `Properties` modules to a finer granularity, where 'properties of predicates on `X`' permit quantification in this style...

JacquesCarette: It's certainly worth considering that design, i.e. allow quantification over `Pred` and `Rel`.

jamesmckinna:
@MatthewDaggitt can you re-review ahead of a possible merge?

src/Data/Fin/Properties.agda
```diff
@@ -1066,6 +1054,30 @@ private
           ¬ (∀ i → P i) → (∃ λ i → ¬ P i)
 ¬∀⟶∃¬ n P P? ¬P = map id proj₁ (¬∀⟶∃¬-smallest n P P? ¬P)
 
+-- Kleisli lifting of Dec over Unary subset relation
```
jamesmckinna: Hmm... this isn't actually Kleisli! My mistake, fix incoming...

jamesmckinna:
Hmm... some weird sync error. Have fast-forwarded with `master` in the hope of fixing things...

src/Data/Fin/Properties.agda
```diff
@@ -1067,6 +1054,30 @@ private
...
+  ih : Dec Q⊆ₛP
+  ih = decFinSubset (Q? ∘ suc) P?
+
+  Q⊆P⇒Q⊆ₛP : Q ⊆ P → Q⊆ₛP
+  Q⊆P⇒Q⊆ₛP q⊆p {x} = q⊆p {suc x}
+
+  dec[Q⊆P] : Dec (Q ⊆ P)
+  dec[Q⊆P] with Q? zero
+  ... | no ¬q₀ = map′ (cons (contradiction′ ¬q₀)) Q⊆P⇒Q⊆ₛP ih
+  ... | yes q₀ = map′ (uncurry (cons ∘ const)) < _$ q₀ , Q⊆P⇒Q⊆ₛP > (P? q₀ ×-dec ih)
```
jamesmckinna, line 1079: Note to self: this will need fixing if and when #2843 gets merged!

jamesmckinna:
Rebasing against `master` before adding to the queue to be on the safe side.

Issue #2539, opened 2025-01-01
The design of 'free objects'


JacquesCarette:
This issue is a place to discuss the design of what we'd like. @jamesmckinna gave a first PR (#1962 ) that has been long stalled (because of, well, me!)

So let's try to get it going again. My biggest worry is the ridiculous amount of boilerplate. This is not @jamesmckinna 's fault, it is inherent to attempting to define 'free object' completely *by hand* by *unrolling* all the definitions.  Not only are the definitions unrolled, a whole lot of useful 'kit' is also provided (normally a wonderful thing!) also by unrolling all the definitions. Wikipedia's [Free object](https://en.wikipedia.org/wiki/Free_object) article is quite readable to get a sense of what all that is.  Its [list of free objects](https://en.wikipedia.org/wiki/Free_object#List_of_free_objects) shows us how much this is going to explore. Plus that list is far from complete.

Maybe we should instead bite the bullet and do this [properly categorically](https://ncatlab.org/nlab/show/free+object).

Another approach would be to follow something largely along @jamesmckinna 's design but, rather than doing it all by hand, doing it via meta-programming. We couldn't do it using Agda's own meta-programming features (I don't think) but we could certainly do it in Haskell -- maybe even by using Agda itself as a library. (Maybe @TOTBWF might be interested in that.)

JacquesCarette:
I should also link in #2287 "Mechanize the Algebra hierarchy". I think that the design space for both are quite similar and should likely have a similar solution.

jamesmckinna:
Thanks @JacquesCarette for opening this issue after all my nagging about it.

I guess my original 'by hand' version(s) were a ham-fisted attempt towards provoking such discussion, and as a way of exploring the space of how to present `FreeX` as a *syntactic* construction based on the term algebra for the signature of `X`. And... syntax breeds boilerplate? (I don't think it's *merely* a consequence of my own more verbose, 'didactic' style of writing Agda?)

The abstract 'universal property' version of `FreeX` seems much closer to a 'finally tagless' *semantic* (Church-style) characterisation, which clearly has advantages in (any) functional context... but for the 'usual' problems of Church-style wrt induction/pattern-matching as a 'programmer's API' for the construction, as well as universe level polymorphism in the predicative setting.

(More later, but I thought it worth putting those two markers down: I think each approach carries 'boilerplate' with it, but we seem to have different tolerance levels regarding the two, as well perhaps as ideological/religious positions as to what is tolerable, and why...?)

[...]

PR #2796, opened 2025-07-30
[ add ] partial elements in `Effect.Monad.Partial` module

e-mniang:
This PR  introduces the `↯` type for partial elements along with its associated functions for binding, mapping, and applying. 

Partial computations — such as non-terminating functions or computations undefined on some input — are represented by a domain of definition and an associated function. The type ↯ A ℓ encapsulates such partial values over a type A by:

```
record ↯ {ℓ} (A : Set ℓ) (ℓ' : Level) : Set (ℓ ⊔ suc ℓ') where
  field
    Dom : Set ℓ'
    elt : Dom → A
```

**Basic constructors:**
```
    never : nowhere-defined partial element

    always : total value lifted into a partial one
```

 **Functions:**
```
bind↯ : monadic bind

map↯ : functorial map

ap↯ : applicative operator
```
-----------------------------------------------
Source: https://1lab.dev/Data.Partial.Base.html#1132 and @TOTBWF


MatthewDaggitt:
Hi @e-mniang, thanks for the contribution! I've requested reviews from people who may be more capable of assessing this than me. However, for starters please could you make sure that your proposed new file has the same formatting as all the other files in the standard library?

e-mniang (commented):
Hi! Thanks for your comment.
I made some changes and I hope the file is now in the correct format. :)

jamesmckinna:
@MatthewDaggitt requested my review, but ahead of any such, my first thought would be: why is the `Level` of the domain *universally quantified* in the definition, and not *existentially*?:
```agda
record ↯ {a} (A : Set a) : Set ω where
  field
    ℓ : Level
    Dom : Set ℓ
    elt : Dom → A
```

I realise that `Setω` is not everyone's cup of tea, but it strikes me that such a definition is closer in spirit to saying "there is *some* domain of definition" which I (perhaps mistakenly) presume to be part of the data for 'partiality'.

An alternative *might* be to define 'partial function from `A` to `B`', in which we could constrain the level of the domain to be bounded by that of `A` (actually: could we do that here, simply by taking `a` as the level of `Dom`?) and then the type of 'partial elements of `A`' would be given by 'partial function from `Data.Unit.Polymorphic.⊤` to `A`'?

I'll try to look at the 1lab to see what's going on there (oh, some quite delicate things under the hood about a small type of propositions... and the level that such a thing should live at. Hmmm). And perhaps also go back to Eugenio Moggi's PhD thesis, and Scott's "Identity and Existence in Intuitionistic Logic" (Springer LNM 753) for a refresher course...

Also: `Codata.Sized.Delay` and friends?

gallais:
> Also: `Codata.Sized.Delay` and friends?

We indeed have e.g. https://agda.github.io/agda-stdlib/v2.2/Effect.Monad.Partiality.html

jamesmckinna:
> > Also: `Codata.Sized.Delay` and friends?
> 
> We indeed have e.g. https://agda.github.io/agda-stdlib/v2.2/Effect.Monad.Partiality.html

So... that leaves some tricky design choices as to how to proceed with this PR, if we think it is worth doing so:
* for example, can we show that `Effect.Monad.Partiality` gives rise to an instance of the structure you define above? Eg. consider the subset of `A ⊥` with domain defined by `_⇓` (via `Data.Refinement` or a vanilla `Σ`-type...), with inclusion given by the obvious thing...
* can you go in the opposite direction (that seems harder?)
* what, if any, are the consequences of (giving up on the details of) the `1lab` type `Ω` of small propositions etc. in your definition?
* what properties of 'partial' elements/functions in `1lab` are missing from the analysis/properties of the `_⊥` monad?

So... overall my conclusion (for the time being, at least) is that this PR probably shouldn't proceed, unless there are compelling answers to the above questions?

TOTBWF:
Delay monads like the one in `Effect.Monad.Partiality` require some degree of countable choice to be equivalent to the partiality monad here: I think "Quotienting the delay monad by weak bisimilarity" by Chapman et al. has the details. The problem is that delay is too intentional: two things that eventually arrive at the same value might take different numbers of steps.  IIRC you still get a monad on the category of setoids if you pick the right equivalence relation, but it's quite delicate, and I don't recall if it actually classifies partial maps.

As for impredicativity, this isn't essential aside from the usual caveats about predicative order theory: see Tom de Jong's thesis for more.

Finally, we could prove the monad laws, though this would require some setoid-fu that quotients away the choice of witnesses of definedness.

TOTBWF:
Also, I forgot to mention that modulo proof-relevance, this is the free pointed DCPO, whereas delay-like monads are aiming to be the free omega-CPO.

JacquesCarette:
That level of 'documentation' would be worth putting in the PR, including adding some to the current Partiality monad.

jamesmckinna:
@TOTBWF wrote:
> Also, I forgot to mention that modulo proof-relevance, this is the free pointed DCPO, whereas delay-like monads are aiming to be the free omega-CPO.

Thanks Reed for this. Does this (ie. 'free pointed DCPO') mean that only *decidable* domains of partial definitions are being considered (because we can distinguish the point?), or is that a misunderstanding based on `stdlib` (erroneous?) use of `Maybe` for capturing pointedness?

Also: *a propos* free-ness (and #1962 #2539 etc. for @JacquesCarette 's shopping list of examples with which to explore candidate designs), are we going to add the relevant universal properties, suitably phrased, at some point downstream... ?

jamesmckinna:
After the recent slew of PRs based on the 'subset-as-family' idiom, I'm happy to row back on my suggestions regarding `Level` parameterisation/(im)predicativity, and take a fresh(er) look at the current draft.

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Effect.Monad.Partial where
+
+open import Level using (Level; suc; zero;_⊔_)
+open import Data.Product using (_×_; Σ; Σ-syntax; _,_)
+open import Data.Empty using (⊥-elim; ⊥)
+open import Data.Unit using (⊤)
```
jamesmckinna: I'm going to guess/recommend that these should import the *polymorphic* versions of `Unit` and `Empty`
```suggestion
open import Data.Empty.Polymorphic using (⊥-elim; ⊥)
open import Data.Unit.Polymorphic using (⊤)
```

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+
+private  
+  variable
+    ℓ ℓ' : Level
+    A B : Set ℓ
```
jamesmckinna: (Current) `stdlib` conventions would have this as:
```suggestion
    a ℓ ℓ' : Level
    A B : Set a
```
based on the `A : Set a` idiom seen in most/many other modules...

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+
+------------------------------------------------------------------------
+-- The partial monad
+
+record ↯ {ℓ} (A : Set ℓ) (ℓ' : Level) : Set (ℓ ⊔ suc ℓ') where
```
jamesmckinna: Accordingly,
```suggestion
record ↯ (A : Set a) (ℓ : Level) : Set (a ⊔ suc ℓ) where
```
where we now profit from having globally declared `a` as inferrable of type `Level`.

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+  field
+    Dom : Set ℓ'
```
jamesmckinna: And hence
```suggestion
    Dom : Set ℓ
```

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+    elt : Dom → A
```
jamesmckinna: This seems *much* more debatable!
`Dom`as a (capitalised) name seems fine for 'domain of the family map defining a subset of `A`', but while I see the *possible* merit of `elt` (wanting to connote 'the family map represents elementhood of the subset'), i think it's just not quite ... enough to carry that weight.

Two alternatives might be
* `ι` ($\iota$), as emblematic of a (monomorphic) injection
* `dom` as the name of a function taking elements of `Dom`

Other suggestions/discussion welcome!

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+
+open ↯
+
+never : ↯ A zero
+never .Dom = ⊥
+never .elt = ⊥-elim
```
jamesmckinna: I'm always a bit nervous about leaving implicit `Level` inference as a hostage to fortune, But I think deciding that `never` and `always` be tied to `Level` *zero* is over-determining the design.
```suggestion
never : ↯ A ℓ
never {ℓ = ℓ} .Dom = ⊥ {ℓ = ℓ}
never .elt = ⊥-elim
```

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+
+always : A → ↯ A zero
+always a .Dom = ⊤
+always a .elt _ = a
```
jamesmckinna: Ditto.
```suggestion
always : A → ↯ A ℓ
always {ℓ = ℓ} a .Dom = ⊤ {ℓ = ℓ}
always a .elt _ = a
```

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+↯-bind a↯ f .Dom = Σ[ a↓ ∈ a↯ .Dom ] f (a↯ .elt a↓) .Dom
+↯-bind a↯ f .elt (a↓ , fa↓) = f (a↯ .elt a↓) .elt fa↓
+
+↯-map : (A → B) → ↯ A ℓ → ↯ B ℓ
+↯-map f a↯ .Dom = a↯ .Dom
+↯-map f a↯ .elt d = f (a↯ .elt d)
+
+↯-ap : ↯ (A → B) ℓ → ↯ A ℓ' → ↯ B (ℓ ⊔ ℓ')
+↯-ap a→b↯ a↯ .Dom = a→b↯ .Dom × a↯ .Dom
+↯-ap a→b↯ a↯ .elt (f↓ , a↓) = a→b↯ .elt f↓ (a↯ .elt a↓)
```
jamesmckinna: Cool.
What's the unicode name for the down-lightning-bolt symbol, given that this is going to be in frequent use, if deployed?

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,51 @@
...
+
+
```
jamesmckinna, line 84: Too much whitespace! We seem inconsistent as to whether we take *no* blank lines, or else simply a single one, as module terminator, but two... too much!

jamesmckinna (dismissed):
Very nice, and I look forward to seeing this land, after appropriate revisions.

src/Effect/Monad/Partial.agda
jamesmckinna: Hmmm, thus may need an explicit named telescope to bind `{ℓ = ℓ}` on the LHS. Apologies for posting without testing first!

src/Effect/Monad/Partial.agda
JacquesCarette: I'm fine with both those suggestions. @TOTBWF ?

src/Effect/Monad/Partial.agda
JacquesCarette: I definitely agree that we should be more `Level`-polymorphic whenever we can.

jamesmckinna:
Re: the construction as a partial map classifier.

At the moment, this only provides the 'raw' type constructor, without any additional `Setoid`-fu to make it a construction fitting into the existing `Setoid`-based hierarchy of 'sets and functions' as the (concrete) basis for (abstract) algebra.

Fine as far as that goes, as we don't even have machinery (yet) for partial functions, besides my speculative #2816 , so I'm interested in where this PR might go downstream in terms of fleshing out that, or any alternative, vision for such things.

src/Effect/Monad/Partial.agda
e-mniang: Here is the unicode : U+21AF

jamesmckinna: Is there an Emacs `agda-input-mode` version?

e-mniang: Yes (sorry for the delay), it's \dz.

JacquesCarette:
So it looks like this one might be ready after all? @jamesmckinna what is outstanding?

jamesmckinna:
> So it looks like this one might be ready after all? @jamesmckinna what is outstanding?

I'll need to come back to this and think! Plus get more sleep than I'm currently managing to achieve.

UPDATED: needs `CHANGELOG` at a minimum. Other stuff I'm still thinking about, but I would double down on makin g `never` and `always` `Level`-polymorphic... the previous suggestions weren't applied! UPDATED: now done.

@e-mniang are you still interested in working on this, and/or able to do so? If not, I think that between @JacquesCarette and myself we can try to merge this in soon. In any case, converting back from `DRAFT` with this in mind...

src/Effect/Monad/Partial.agda
jamesmckinna: I took the initiative with `dom`...

jamesmckinna:
@JacquesCarette I think this now is a minimal commit-able bunch of code, but lacking @TOTBWF 's gloss/commentary. Will add later, unless  Reed, do you want to?

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,63 @@
...
+    A B : Set a
+
+
+------------------------------------------------------------------------
+-- Object part: type definition
+
+record ↯ (A : Set a) (ℓ : Level) : Set (a ⊔ suc ℓ) where
+  field
+    Dom : Set ℓ
+    dom : Dom → A
```
jamesmckinna, line 38: Question: can we improve the ergonomics of this definition by making `ℓ` implicit rather than explicit?
```suggestion
record ↯ {ℓ} (A : Set a) : Set (a ⊔ suc ℓ) where
  field
    Dom : Set ℓ
    dom : Dom → A
```

jamesmckinna:
Question for @JacquesCarette and @TOTBWF : 
* the difference between this contribution for `stdlib` and that for 1lab is that `Dom` isn't required to be a `Prop`/subsingleton/subobject of 1, so should that in fact be part of the definition?
* wrt this, `never` corresponds to the subsingleton which is *actually* empty; `always` to the definitely inhabited singleton, as the 'two' extremal examples...
* ...  but what about 
  ```agda
  weird : ↯ A _
  weird {A = A} = record { Dom = A ; dom = id }
  ```

On its face, this typechecks correctly (at level `a` if `A : Set a`), but what on earth does it correspond to, considered as a 'partial' inhabitant of `A`. If `A` is already a subsingleton (I'd be tempted to qualify the `Dom` field declaration with an irrelevance modality, but as discussed elsewhere, this may not be the moment to open that can of worms... cf. https://github.com/agda/agda/issues/8557), then `weird`, although *weird*, is more-or-less 'obviously equivalent' to `always`. But without any such stipulation... what exactly do we have here?

Proposal: we park this PR unmerged, until such time as we resolve the irrelevance/`Prop` issue, or else ... well I'm not quite sure!?

Please chime in on this, otherwise I'm not sure that this should go through even as it now is. 

(Reading back over my earliest comment, I did ask about what might be the consequences of dropping the `Prop`/subsingleton restriction... sorry it's taken so long to return to my first thought!)

src/Effect/Monad/Partial.agda
jamesmckinna: Marginal, given my concerns below, so resolving for now.

TOTBWF:
If you remove the subsingleton condition then you get a multi-map classifier instead of a partial map classifier.

```agda
A → ↯ B 
≅ A → Σ (Dom : Set) (Dom → B) -- by defn
≅ A → (B → Set) -- Fibre-family equivalence[1]
```

After some thought, I think this is actually what you'd want in the stdlib. If you are working with setoids, then the right notion of subsingleton is `∀ x y → x ≈ y`. This has some consequences for the theory of partial functions: the right notion of partiality is that the domain of definition is a setoidal subsingleton. This means things like the codiscrete setoid will have an underlying multimap, *not* an underlying partial map.

As for `weird`, this is actually a very handy trick. One way of viewing 
`↯ A` is that it's a value of `A` guarded behind some assumptions that we will discharge later. From this perspective, `weird` lets you add an assumption that you will need to discharge later. This is *very* useful when programming in the partiality/multimap monad; it's basically a proof-relevant, witness providing form of `guard :: Bool -> m ()` or `assert`.

I would definitely *not* make Dom a Prop or irrelevant though! Both will be rather ill-behaved when you want to use `weird` (or it's subsingleton version where the domain of definition is `Σ (x : A) (∀ y → x = y)`) to cook up things like termination witnesses. Moreover, neither Prop nor irrelevance supports unique choice[^2] which will cause problems when you try to take directed joins.

[^1] This step may or may not need univalence depending on the notion of isomorphism/equivalence of types you use.

[^2] Modulo the bug where you get global choice lol

JacquesCarette:
I believe the "global choice" bug has been fixed by Jesper in current development Agda, and ought to be in 2.9.

jamesmckinna:
On the basis of @TOTBWF 's comments, I'd be happy to merge now.

CHANGELOG.md
```diff
@@ -270,6 +270,8 @@ New modules
   Data.Tree.Rose.Show
   ```
 
+* `Effect.Monad.Partial` based on domain `Dom` and injection `dom`.
```
JacquesCarette: There should be more information here (from comments and from PR).

src/Effect/Monad/Partial.agda
```diff
@@ -0,0 +1,63 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- The partial monad cf. https://1lab.dev/Data.Partial.Base.html
+--
+-- Modulo proof-relevance, this defines the free pointed DCPO,
+-- whereas delay-like monads, as in `Effect.Monad.Partiality`,
+-- are aiming to define the free ωCPO.
+-- NB. in each case, there are additional 'up to' considerations
+-- wrt 'appropriate' setoid equality/quotient/bisimilarity.
```
JacquesCarette, line 10: Probably some of the PR information should also be added here).

JacquesCarette (changes requested):

gallais (changes requested):
I'm happy with this except for:

* `↯-` prefix in names (we should always avoid prefixes as you can import the module qualified if you need to disambiguate clashing names)
* Should `always` be called `pure` and moved to the previous section?

Observing these structures, there is of course opportunity for more combinators e.g. `never` could be the neutral element for an alternative `_<|>_`

src/Effect/Monad/Partial.agda
jamesmckinna: What more? Please feel free to add as you see fit!

jamesmckinna:
@gallais I fixed the prefixing of names
@JacquesCarette I realised two things:
* I'm happy to merge this into the library
* But, my understand is still a bit dim about many of the issues discussed by @TOTBWF above, so I punted the documentation issues you raise to a link in the `CHANGELOG` to this discussion.

Not sure there is much more I can manage right now!

PR #2801, opened 2025-08-02
[ add ] clean version of `Data.Fin.Properties.searchMinimalCounterexample`

jamesmckinna:
Revised version of the relevant code from #2744 incorporating: UPDATED
* ~~`MinimalCounterexample`/`Least⟨¬_⟩`~~ `Least⟨_⟩` now defined as a `record`, not as `syntax`
* ~~infix versions now as (duplicate) definitions of the~~ removed verbose CamelCase versions
* use of `inject` deprecated in favour of `_<_`; add `inject-<` as lemma to smooth the path

Outstanding issues, if this PR gets merged:
* UPDATED lemmas `¬∀⟶∃¬-smallest` (definitely?) and `¬∀⟶∃¬` (maybe not?) should be deprecated, as superseded by `search-least⟨¬_⟩`?

### DIFF
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -29,7 +29,7 @@ open import Data.Product.Properties using (,-injective)
 open import Data.Product.Algebra using (×-cong)
 open import Data.Sum.Base as Sum using (_⊎_; inj₁; inj₂; [_,_]; [_,_]′)
 open import Data.Sum.Properties using ([,]-map; [,]-∘)
-open import Function.Base using (_∘_; id; _$_; flip)
+open import Function.Base using (_∘_; id; _$_; flip; const; λ-; _$-)
 open import Function.Bundles using (Injection; _↣_; _⇔_; _↔_; mk⇔; mk↔ₛ′)
 open import Function.Definitions using (Injective; Surjective)
 open import Function.Consequences.Propositional using (contraInjective)
@@ -62,6 +62,7 @@ private
     A : Set a
     m n o : ℕ
     i j : Fin n
+
 
 ------------------------------------------------------------------------
 -- Fin
@@ -1000,7 +1001,7 @@ pinch-injective {i = suc i} {suc j} {suc k} 1+i≢j 1+i≢k eq =
 -- Quantification
 ------------------------------------------------------------------------
 
-module _ {p} {P : Pred (Fin (suc n)) p} where
+module _ {P : Pred (Fin (suc n)) p} where
 
   ∀-cons : P zero → Π[ P ∘ suc ] → Π[ P ]
   ∀-cons z s zero    = z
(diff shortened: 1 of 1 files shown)


src/Data/Fin/Properties.agda
```diff
@@ -1048,16 +1053,57 @@ private
...
+μ⟨_⟩ : Pred (Fin n) p → Set p
+μ⟨_⟩ = MinimalExample
+
+μ⟨¬_⟩ : Pred (Fin n) p → Set p
+μ⟨¬_⟩ = MinimalCounterexample
+
+¬¬μ⇒μ : ∀ {P : Pred (Fin n) p} → Decidable P → μ⟨¬ ∁ P ⟩ → μ⟨ P ⟩
+¬¬μ⇒μ P? (μ i ¬¬pᵢ ∀[j<i]) = μ i (decidable-stable (P? i) (¬-recompute ¬¬pᵢ)) ∀[j<i]
+
+searchMinimalCounterexample : ∀ {P : Pred (Fin n) p} → Decidable P → (∀ i → P i) ⊎ μ⟨¬ P ⟩
```
gallais: We have a notation for this in case you are not aware:

```suggestion
searchMinimalCounterexample : ∀ {P : Pred (Fin n) p} → Decidable P → Π[ P ] ⊎ μ⟨¬ P ⟩
```

src/Data/Fin/Properties.agda
```diff
@@ -1048,16 +1053,57 @@ private
...
+searchMinimalCounterexample {zero}  {P = _} P? = inj₁ λ()
+searchMinimalCounterexample {suc _} {P = P} P? with P? zero
+... | no ¬p₀ = inj₂ (μ zero ¬p₀ λ())
+... | yes p₀ = Sum.map (∀-cons p₀) μ⁺ (searchMinimalCounterexample (P? ∘ suc))
+  where
+  μ⁺ : μ⟨¬ P ∘ suc ⟩ → μ⟨¬ P ⟩
```
gallais: Should this be a toplevel lemma?

```
∀[ Decidable ⇒ suc ⊢ μ⟨¬_⟩ ⇒ μ⟨¬_⟩ ]
```

src/Data/Fin/Properties.agda
jamesmckinna: So to this, I am going to hint that this suggestion (and the `syntax` underlying it) doesn't find universal (*sic*) favour with the other maintainers (see the feedback on the original #2744 from which this is derived)... and my choice of `μ⟨¬_⟩` is deliberately as a synonym/alias, and not as `syntax`, as a consequence. 

Shorter lines won me over to the notation!

Ob: `Π[ P ]` why was this notation chosen in preference to `∀[ P ]`, given that `∃⟨ P ⟩` was chosen in preference to `Σ⟨ P ⟩`?

src/Data/Fin/Properties.agda
jamesmckinna: Hmmm... *I* perhaps find that formulation compelling, if, as with all these things, the notational economy comes with an associated decompression penalty...
... while the simplicity of the proof makes me wonder whether it's 'worth' putting at top-level? The advantage of local lemmas is that they can always be promoted downstream, but adding new top-level lemmas requires (slightly) more justification?

Still, it is a direct analogue of `∀-cons`, so maybe that's reason enough?

Oh... hang on, where does `p₀ : P zero` fit in your re-formulation?

src/Data/Fin/Properties.agda
gallais: `∀[ P ]` also exists for quantifying over an implicit index

src/Data/Fin/Properties.agda
gallais: Oops, my bad. I did not look at the code too closely and so did not spot the
locally bound  `P zero`.

src/Data/Fin/Properties.agda
jamesmckinna: Of course! D'oh...

JacquesCarette (commented):
I like all of this, except for the name ` μ`. It's a cute pun, but caters to a rather small crowd.

How about calling the constructor `least` instead?  If you want a pun, then `μ⟨_⟩` can be called `the-least`. I don't think `μ⟨¬_⟩` passes the Fairbairn threshold.

jamesmckinna:
> I like all of this, except for the name `μ`. It's a cute pun, but caters to a rather small crowd.

It was *absolutely* not intended as a pun:
* for the 'small crowd' of recursion theorists/computer scientists who know or care about recursion theory, I regard `μ` as *absolutely conventional usage* denoting the Godel-Kleene minimisation operator, here specialised to the (much) easier case of `Fin`;
* `Relation.Unary` defines a number of 'Greek' short form syntaxes/definitions `∃⟨_⟩`, `∀[_]` for things which otherwise would be more verbose, and I'm conscious that there are at least two schools of thought about their appropriateness for `stdlib`, viz. favouring ASCII/CamelCase or else 'Greek' forms, and this PR follows that model: long-form names, with short-form 'abbreviations'

> How about calling the constructor `least` instead? If you want a pun, then `μ⟨_⟩` can be called `the-least`. I don't think `μ⟨¬_⟩` passes the Fairbairn threshold.

Well...
* the use of `μ` as a *constructor* for the two records introduced here is, of course,  both optional (no constructor required), and replaceable, so I'd be happy to change it for *something*, but 'least' seems itself to be working a bit *too* hard cognitively/semantically, and above all, I wanted something 'minimum ink' (following Wadler, whose taste in these matters usually seems reliable...)
* the Fairbairn objection to `μ⟨¬_⟩` I actively disagree with: because the types `MinimalExample` and `MinimalCounterexample` are subtly different, to try to (more) accurately reflect the computation irrelevance of fields/arguments. So if they are to have short-form abbreviation, then they each deserve one!
* but again, maybe (short) ASCII alternative would be better, but then I'd be tempted simply to resort *only* to the existing `camelCase : CamelCase` idiomatics, but at the (quite high!) cost of adding some *very* long names... :-(

jamesmckinna:
@JacquesCarette You were right, and I was wrong!
Latest commits hopefully simplify the code enough to answer (most of) your outstanding criticisms.
I couldn't tackle the Fairbairn one, though: `irrelevant-projections` warning/errors stymied me at an awkward moment.
If you're able to simplify further, I'd be happy to see the details!

JacquesCarette:
> @JacquesCarette You were right

I somehow never tire of hearing that 😏 .  I'm quite happy with the current results.

src/Data/Fin/Properties.agda
```diff
@@ -1049,16 +1054,51 @@ private
...
+-- A decidable predicate P over a finite set is either always true,
+-- or else we can find the smallest value for which P is false.
+
+module _ (P : Pred (Fin n) p) where
+
+  record Least⟨¬_⟩ : Set p where
+    constructor least
+    field
+      witness : Fin n
+      .contra : ¬ P witness
```
gallais: Is this necessary; all negated proposition are already irrelevant.
Declaring a field irrelevant however means we cannot use the projection as irrelevant projections are unsafe.

jamesmckinna: Yes, you're probably right (notwithstanding https://github.com/agda/agda/issues/7063 ... which left me puzzled as to Agda's intended vs. actual behaviour).

Separately, there's the distinction between definitional vs. propositional irrelevance, but happy to simplify what's here. That might enable @JacquesCarette 'a earlier comment about not needing both definitions...?

jamesmckinna: Indeed: latest commits simplify a lot!

PR #2798, opened 2025-08-01
[ add ] `style-guide` recommendation preferring `contradiction` over `⊥-elim`

jamesmckinna:
Fix #2653 .
Placeholder text for now; suggestions for improvement welcome (esp. wrt rationale/links to associated issues...)... and received.

PR #2797, opened 2025-07-31
[ add ] flipped version of `Relation.Nullary.Negation.Core.contradiction`

jamesmckinna:
Fixes #2784 .

Issue #2653, opened 2025-03-11
Prefer `contradiction` over `⊥-elim` when it makes sense?


JacquesCarette:
See PR #2652 . I think @jamesmckinna had already started down this road, this is continuing that. But I want to make sure that this is the design we want (and so should codify in `style-guide`).

jamesmckinna:
As @Taneb notes on the PR, this is/can be a clear win.

My original motivation, after extensive investigation/refactoring of negation while developing `Relation.Nullary.Recomputable` and its consequences, was to try to avoid explicit appeal to `⊥-elim` altogether, in favour of `contradiction` etc. as an encapsulated abstraction. 

This would/could subsequently be weakened to permit further exploitation of the *irrelevance* of arguments of negated type. See 

So I'm definitely in favour!

JacquesCarette:
So I'm going to let @jmougeot know that he can go ahead and submit PRs to continue this - but single-file PRs for this, as I suspect some of them will not be 'clear wins'.

jamesmckinna:
One other thing I have suggested in the past is also to replace explicit use of ` = ⊥-elim p` with `with () ← p`... in the 'other' cases. 

I still regard that, in and of itself, as a 'clear win', but the lack of response to #2123 suggests/suggested that I was, perhaps, alone in thinking that... and in my jeremiad against overly explicit exposure of the underlying representation of `⊥` (part of a wider issue about how much we are willing to sacrifice 'abstractness' in the library API in favour of exposing such representations...)

jamesmckinna:
> So I'm going to let [@jmougeot](https://github.com/jmougeot) know that he can go ahead and submit PRs to continue this - but single-file PRs for this, as I suspect some of them will not be 'clear wins'.

Indeed, regarding adding this to the `style-guide`, it might be worth canvassing beyond my own (perhaps) idiosyncratic opinion...

gallais:
I thought it was already in the style guide as @MatthewDaggitt has been pushing
for `contradiction` over `⊥-elimpretty consistently in the past.

jamesmckinna:
> I thought it was already in the style guide as [@MatthewDaggitt](https://github.com/MatthewDaggitt) has been pushing for `contradiction` over `⊥-elim` pretty consistently in the past.

Sadly, not in `style-guide` (yet)... but good to know that you and @MatthewDaggitt agree!

jamesmckinna:
[...]

PR #2802, opened 2025-08-04
[ refactor ] make `contradiction` and friends entirely definitionally proof-irrelevant

jamesmckinna:
The inevitable `breaking` consequence of #2785 , designed as a 'cleaning of the stables'. 
Cf. #2346 whose intention it follows.
Currently duplicates functionality from #2803 ; will tidy up once that is merged.

Downstream consequence:
* some uses of `contradiction` internally need eta-expansion (and this may be regarded as a UX failure...?)
* consequently, so too may some client uses of this definition elsewhere (but no others in `stdlib` that I can find)

TODO:
- [ ] revert all uses of `contradiction-irr` introduced as consequences of #2785 
- [ ] ...

Consider also:
- [ ] refactoring `Relation.Nullary.Reflects` to make its various arguments definitionally proof-irrelevant
- [ ] deprecating/removing `Relation.Nullary.Negation.Core._¬-⊎_` in favour of `contradiction₂`, or else change its type as well; its one use in `stdlib` is in`Reflects`, so could be refactored in any case via `¬-recompute` and inlining...
- [ ] ...

### DIFF
--- a/src/Data/Vec/Relation/Unary/Unique/Setoid/Properties.agda
+++ b/src/Data/Vec/Relation/Unary/Unique/Setoid/Properties.agda
@@ -39,7 +39,7 @@ module _ (S : Setoid a ℓ₁) (R : Setoid b ℓ₂) where
 
   map⁺ : ∀ {f} → (∀ {x y} → f x ≈₂ f y → x ≈₁ y) →
          ∀ {n xs} → Unique S {n} xs → Unique R {n} (map f xs)
-  map⁺ inj xs! = AllPairs.map⁺ (AllPairs.map (λ x≈y → contraposition inj x≈y) xs!)
+  map⁺ inj xs! = AllPairs.map⁺ (AllPairs.map (contraposition inj) xs!)
 
 ------------------------------------------------------------------------
 -- take & drop


src/Algebra/Module/Properties/Semimodule.agda
```diff
@@ -16,25 +16,28 @@ module Algebra.Module.Properties.Semimodule
...
 open CommutativeSemiring semiring
 open Semimodule          semimod
 open import Relation.Binary.Reasoning.Setoid ≈ᴹ-setoid
-open import Relation.Nullary.Negation using (contraposition)
+
+------------------------------------------------------------------------
 
 x≈0⇒x*y≈0 : ∀ {x y} → x ≈ 0# → x *ₗ y ≈ᴹ 0ᴹ
-x≈0⇒x*y≈0 {x} {y} x≈0 = begin
+x≈0⇒x*y≈0 {x = x} {y = y} x≈0 = begin
```
JacquesCarette: Why is this change needed?

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -409,17 +409,17 @@ module Antisymmetry
     length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
     length zs        ≤⟨ ℕ.n≤1+n (length zs) ⟩
     length (z ∷ zs)  ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
+    length ys₁       ∎) (ℕ.<-irrefl ≡.refl)
```
JacquesCarette: perhaps don't roll in such pure, non-breaking style changes in to something `breaking` and make a separate (if tiny) PR?

src/Data/List/Relation/Unary/First/Properties.agda
```diff
@@ -54,7 +54,7 @@ module _ {a p q} {A : Set a} {P : Pred A p} {Q : Pred A q} where
 module _ {a p q} {A : Set a} {P : Pred A p} {Q : Pred A q} where
 
   All⇒¬First : P ⊆ ∁ Q → All P ⊆ ∁ (First P Q)
-  All⇒¬First p⇒¬q (px ∷ pxs) [ qx ]   = contradiction qx (p⇒¬q px)
+  All⇒¬First p⇒¬q (px ∷ pxs) [ qx ]   = p⇒¬q px qx
```
JacquesCarette, line 57: Doesn't this go in the opposite direction of preferring `contradiction`?

src/Relation/Nullary/Negation.agda
```diff
@@ -72,7 +72,7 @@ open import Relation.Nullary.Negation.Core public
 -- ⊥).
 
 call/cc : ((A → Whatever) → DoubleNegation A) → DoubleNegation A
-call/cc hyp ¬a = hyp (flip contradiction ¬a) ¬a
+call/cc hyp ¬a = hyp (λ a → contradiction a ¬a) ¬a
```
JacquesCarette: Can't you use `contradiction'` to avoid eta-expanding?

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -67,17 +77,11 @@ contra-diagonal self a = self a a
 
 -- Everything is stable in the double-negation monad.
 stable : ¬ ¬ Stable A
-stable ¬[¬¬a→a] = ¬[¬¬a→a] (contradiction (¬[¬¬a→a] ∘ const))
+stable ¬[¬¬a→a] = ¬[¬¬a→a] λ ¬¬a → contradiction (¬[¬¬a→a] ∘ const) ¬¬a
```
JacquesCarette, line 81: same here and below: `contradiction'`?

src/Relation/Nullary/Negation/Core.agda
```diff
@@ -26,38 +26,48 @@ infix 3 ¬_
 ¬_ : Set a → Set a
 ¬ A = A → ⊥
 
+-- Note the following use of flip:
+private
+  note : (A → ¬ B) → B → ¬ A
```
JacquesCarette: If you're going to bother to move this to the top of the file, maybe also document where the 'note' is useful in reading below?

src/Algebra/Module/Morphism/ModuleHomomorphism.agda
```diff
@@ -51,7 +51,7 @@ x≈0⇒fx≈0 {x} x≈0 = begin⟨ B.≈ᴹ-setoid ⟩
   B.0ᴹ   ∎
 
 fx≉0⇒x≉0 : ∀ {x} → f x B.≉ᴹ B.0ᴹ → x A.≉ᴹ A.0ᴹ
-fx≉0⇒x≉0 = contraposition x≈0⇒fx≈0
+fx≉0⇒x≉0 x≈0 = contraposition x≈0⇒fx≈0 x≈0
```
JacquesCarette: Can you spell out why this expansion is needed (in the PR text, not there)? I'm puzzled.

jamesmckinna:
Thanks @JacquesCarette for the speedy feedback (on a `DRAFT` PR for v3.0!). 

For most of your queries, eta-expansion is needed because there's no (contravariant) subtyping between `.A → B` and `A → B`.

As for the ones which seemingly contradict #2653 , see #2803 : don't use `contradiction` where simple application will do instead! (Ie use of negation sometimes happens in the minimal logic fragment, without needing *ex falso* at all)

src/Algebra/Module/Properties/Semimodule.agda
jamesmckinna: Bug! Caused by trying to introduce `variable`s, but then they muddle the quantifier prefix order...

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
jamesmckinna: Not a style change; it's yet another 'feature' of irrelevant function spaces: `_$_` is no longer well-typed!

src/Data/List/Relation/Unary/First/Properties.agda
jamesmckinna: See #2803 for a related instance.

src/Relation/Nullary/Negation/Core.agda
jamesmckinna: Well... for starters, `breaking`, so code motion ought to be permissible?

Secondly, more importantly, *some* of the properties in this module (eg `_¬-⊎_` and `¬¬-η`) are consequences of the *minimal logic* interpretation of `¬_`, and they are conceptually prior to the ones which make genuine use of *ex falso* `⊥-elim(-irr)`... so it made sense to me that they occur first in the module. Cf. #2803 

As to the best way to document this separation of concerns, suggestions welcome!

src/Algebra/Module/Morphism/ModuleHomomorphism.agda
jamesmckinna: The eta-expansion is needed because there's no (contravariant) subtyping between .A → B and A → B.

src/Relation/Nullary/Negation/Core.agda
jamesmckinna: The eta-expansion is needed because there's no (contravariant) subtyping between .A → B and A → B.

src/Relation/Nullary/Negation.agda
jamesmckinna: No... The eta-expansion is needed because there's no (contravariant) subtyping between .A → B and A → B.

src/Relation/Nullary/Negation/Core.agda
jamesmckinna: Maybe one way to resolve this is to interpolate a v2.4 non-`breaking` PR which reorganises `Negation.Core` to distinguish
* definition
* consequences/properties *per* minimal logic:  `note`, `¬¬-η`, `contra-diagonal`...
* consequences/properties *per* use of *ex falso*

rather than try to do this as part of this PR?

jamesmckinna: And/or add some text to #2798 to emphasise the distinction?

jamesmckinna: See: #2805

jamesmckinna:
It's still `DRAFT`, and in any case v3.0, but it'll need to wait for any dust to settle from #2797 #2803 and #2805 etc. before proceeding!

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
JacquesCarette: So do we need a new variant for `_$_`  that would be?

src/Relation/Nullary/Negation/Core.agda
JacquesCarette: I'm quite happy with #2805 . I think that will indeed take care of my issue here as well.

jamesmckinna:
Reconsidering whether this is even worthwhile: logically it makes sense, but ergonomically, it's such a pain.

JacquesCarette:
Reading this code, I get the same impression: ergonomically, agda's proof irrelevance leaves much to be desired. Is there any conclusion from here that can be up-streamed?

jamesmckinna:
> Reading this code, I get the same impression: ergonomically, agda's proof irrelevance leaves much to be desired. Is there any conclusion from here that can be up-streamed?

Perhaps, but in the interests of cleaning up my own cluttered brain, I'd probably rather punt this. I conjecture/suspect that much of the current effort being expended on modalities in Agda more generally might hint that the `.` modality may one day end up deprecated?

In any case, I think that I myself may have misconstrued how best to deploy these ideas (re: `¬-recompute` and the Harrop fragment more generally): if we have an *irrelevant* negated argument in scope, and from which we wish to derive a contradiction, then it can be promoted via `¬-recompute` and then appeal to `contradiction`.

Whereas, if we have an irrelevant *positive* assumption `a : A`, and seek a contradiction that and `¬ A`, then... we already have `contradiction-irr` for that. We *could* insist on replacing each of `contradiction` and `contradiction′` with these irrelevant versions, but then the UX problems bite again...  whereas the current status quo gives us eta-contractible version of those functions, nevertheless with each being given by delegation to the irrelevant version...

... conclusion: closing for now.

PR #2805, opened 2025-08-05
[ refactor ] reorganise `Relation.Nullary.Negation.Core` to make clean separation of properties

jamesmckinna:
This is somehow an 'intermediate' PR following on from #2803 and before the more radical #2802 , aiming to systematise the organisation of the module into definitions and properties which depend on negation being a connective in minimal logic, and those (fewer, but more powerful) which genuinely rely on *ex falso* in the form of `Data.Empty.⊥-elim`.

NB. More in the same spirit may be possible downstream, eg.
* monadic structure on `DoubleNegation` requires only the minimal properties (modulo definitional proof-irrelevance... which now could also be tackled by #2802 )
* more nuanced consideration of the style-guide recommendations in #2798 
* etc.

### DIFF
--- a/src/Relation/Nullary/Negation/Core.agda
+++ b/src/Relation/Nullary/Negation/Core.agda
@@ -20,37 +20,56 @@ private
     Whatever : Set w
 
 ------------------------------------------------------------------------
--- Negation.
+-- Definition.
 
 infix 3 ¬_
 ¬_ : Set a → Set a
 ¬ A = A → ⊥
 
 ------------------------------------------------------------------------
--- Stability.
+-- Properties, I: as a definition in *minimal logic*
+
+-- Note the following use of flip:
+private
+  note : (A → ¬ B) → B → ¬ A
+  note = flip
+
+-- Contraposition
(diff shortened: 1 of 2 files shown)


src/Relation/Nullary/Negation/Core.agda
```diff
@@ -20,37 +20,56 @@ private
...
 
 infix 3 ¬_
 ¬_ : Set a → Set a
 ¬ A = A → ⊥
 
 ------------------------------------------------------------------------
--- Stability.
+-- Properties, I: as a definition in *minimal logic*
+
+-- Note the following use of flip:
```
JacquesCarette: I know this was in the file before, but there are *no* uses of `flip`, at this type or any other, (left?) in this file, so this note does not seem to be meaningful.

jamesmckinna: I hadn't anticipated this, but I'd be only to happy to remove `note` now! 
The 'intuition' it carries seems now much better represented in the reorganisation, IMHO.

Issue #2525, opened 2024-12-18
[ refactor ] Functoriality of `Data.List.Relation.Binary.Sublist|Subset` wrt `All` and `Any`


jamesmckinna:
> Should this be a corollary of something about `All P xs` implies `All P ys` whenever ys is a subset of xs?

Indeed, and this appears *twice* in the library:
* `Data.List.Relation.Unary.All.Properties.anti-mono : xs Subset.⊆ ys → All P ys → All P xs` (and `Sublist` is a stronger property than `Subset`)
* `Data.List.Relation.Binary.Sublist.Propositional.Properties.All-resp-⊆ : {P : Pred A ℓ} → (All P) Respects _⊇_` and this should be refactored to a proof in `...Sublist.Setoid.Properties`, under assumptions that `P` respects the underlying `Setoid` equality?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/pull/2522#discussion_r1889111306_

MatthewDaggitt:
Eek that's not ideal that it appears twice :sweat: 

My instincts say that it should live under `Sublist`...

jamesmckinna:
> My instincts say that it should live under Sublist...

As would mine, except for the fact that the `Sublist` versions morally should delegate to the `Subset` versions via `Sublist⇒Subset`? But then we have the problem that this lemma is under `...Subset.Setoid.Properties`... sigh!?

MatthewDaggitt:
> As would mine, except for the fact that the Sublist versions morally should delegate to the Subset versions via Sublist⇒Subset?

Err, that's not a principle that any of the existing code has been designed around. It might be possible to do but it would involve a wholesale refactoring of the `Sublist` modules I suspect.

JacquesCarette:
(There's quite a bit of discussion in #2422 that is relevant.)

This is a much more general design issue: should we prove things directly or as special cases of more general results? That's probably the real discussion to be having. I'm fairly sure that has come up before! There are pros and cons, and my opinion on that has changed (evolved?) multiple times over the years.

jamesmckinna:
#2522 ?

JacquesCarette:
Oops, right - fixed.

jamesmckinna:
@JacquesCarette 's deeper question

Are we able to distil, from the long history `stdlib` commits, any exemplary case studies to illustrate the pros and cons of each choice (wrt the general design question, rather than that of the OP)?

[...]

PR #2808, opened 2025-08-13
[ refactor ] `Data.List.Relation.Binary.Sublist.Propositional.Properties`

jamesmckinna:
See #2525 . This tackles the refactoring part, but does not resolve the duplication problem.

UPDATED: yuk... some tricky intensional equality snafus caused by the refactoring. FIXED now, but at the cost of a `breaking` (intensional to extensional) change...

Not only that, but attempting to deprecate `anti-mono`/`all-anti-mono` causes a dependency cycle between
* `Data.List.Relation.Unary.All.Properties`
* `Data.List.Relation.Binary.Subset.Propositional.Properties`

I'm tempted to suggest that we make the `breaking` change simply of moving the lemmas to the latter module, and marking this as `non-backwards-compatible` for the sake of a simpler dependency graph?

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
@@ -137,6 +137,12 @@ All-resp-⊆-trans {τ = []      } ([]        ) []       = refl
 ------------------------------------------------------------------------
 -- Functor laws for Any-resp-⊆ / lookup
 
+lookup≗Any-resp-⊆ : ∀ {P : Pred A ℓ} (xs⊆ys : xs ⊆ ys) →
+                    lookup xs⊆ys ≗ Any-resp-⊆ {P = P} xs⊆ys
+lookup≗Any-resp-⊆ (y ∷ʳ xs⊆ys)          = cong there ∘ lookup≗Any-resp-⊆ xs⊆ys
+lookup≗Any-resp-⊆ (x ∷ xs⊆ys) (here px) = refl
+lookup≗Any-resp-⊆ (x ∷ xs⊆ys) (there p) = cong there (lookup≗Any-resp-⊆ xs⊆ys p)
+
 -- First functor law: identity.
 
 Any-resp-⊆-refl : ∀ {P : Pred A ℓ} →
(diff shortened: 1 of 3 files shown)


src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
```diff
@@ -167,7 +180,7 @@ lookup-⊆-trans = Any-resp-⊆-trans
 lookup-injective : ∀ {P : Pred A ℓ} {τ : xs ⊆ ys} {i j : Any P xs} →
                    lookup τ i ≡ lookup τ j → i ≡ j
 lookup-injective {τ = _  ∷ʳ _}                     = lookup-injective ∘′ there-injective
-lookup-injective {τ = x≡y ∷ _} {here  _} {here  _} = cong here ∘′ subst-injective x≡y ∘′ here-injective
+lookup-injective {τ = refl ∷ _} {here  _} {here  _} = cong here ∘′ here-injective
```
JacquesCarette, line 183: This change now makes the comment right after it not make sense?

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
```diff
@@ -106,6 +109,23 @@ module _ (≈-assoc : ∀ {w x y z} (p : w ≈ x) (q : x ≈ y) (r : y ≈ z) 
...
 
 
+------------------------------------------------------------------------
+-- Relationships to other predicates
+
+module _ {P : Pred A p} (resp : P Respects _≈_) where
+
+-- All P is a contravariant functor from _⊆_ to Set.
+
+  All-resp-⊆ : (All P) Respects _⊇_
```
JacquesCarette, line 119: The name says the opposite of the property?!?!?

jamesmckinna: The name is legacy, back-ported from `Propositional`. But I agree, it's maybe idiotic to have retained it.

jamesmckinna:
> I think it would make sense to separate the pure additions from the modifications into 2 PRs.

Ah... yes, you're probably right. But as I comment on the original issue, it's perhaps not even obvious what the right way forward is here.

src/Data/List/Relation/Binary/Sublist/Setoid/Properties.agda
JacquesCarette: As it's new, there is still time to fix it, right?

jamesmckinna: OK, but what about its `Propositional` specialisation? Add a new name, and deprecate the old one? That might be better, I suppose!?

jamesmckinna: Oh... lots of knock-on nonsense. Sigh.

JacquesCarette: That would be the logical thing to do (new name, deprecate old).

src/Data/List/Relation/Binary/Sublist/Propositional/Properties.agda
jamesmckinna: Oh, yes. Yuk. Back to the drawing board...

PR #2810, opened 2025-08-15
Module morphisms polymorphic in the underlying ring structure

Taneb:
This was a side-experiment in #2729 that I separated out so it could get feedback without hindering that PR. It is neither blocking nor blocked by that PR.

- [x] Adds a subtraction operator for modules
- [x] Adds a notion of "polymorphic" module morphisms, which can change the underlying ring
- [x] Redefines "monomorphic" module morphisms in terms of these
- [ ] Properties of polymorphic monomorphisms


jamesmckinna:
Maybe I'm missing something, but I'm not clear about the (eventual) pragmatics/uses of this design, compared with factorising any X-module morphism between `M` and `N` over two distinct X-rings of coefficients `R` and `S`, related by a X-ring homomorphism `h`
* a morphism relative to `S` between `h*N` and `N`
* a morphism between `M` and `h*N` relative to `R`
* .. for a suitable definition of ... ah, now I perhaps see.

But still, documenting the rationale for the design as part of the PR documentation, if not as commentary in the files themselves, would be useful, I think. As to unsolved metas/need for explicit arguments, that seems to be a recurring problem with Agda's (weakened) mechanisms for inferring such things in the setting of the more complicate flags/modality/erasure/polarity analyses now being instrumented?

UPDATED:
* suggest that the addition of a binary subtraction operator (inherited from the underlying `+-abelianGroup` of a module?) be lifted out as a separate PR, if it's useful...
* revisiting my earlier comment, maybe it *is* simpler to reconsider this PR in terms of a `Construct` on `XModule`s of 'change-of-base along a homomorphism of coefficients' `h* M`, and then your new composite 'polymorphic' notion of homomorphism as being given definitionally as a 'monomorphic' homomorphism between `M` and `h * N`? And certainly, that wouldn't require the new explicit parametrisation?

JacquesCarette:
I agree that more documentation would have been helpful.

From what I can see from the code:
1. I really want this capability,
2. but not as a modification of the current code,
3. rather as a different kind of module morphism.

jamesmckinna:
@JacquesCarette does the change-of-base construction give you what you want, as suggested above?

JacquesCarette:
> does the change-of-base construction give you what you want?

Quite likely. But the devil's in the details for such things, thus my hedging.

src/Algebra/Module/Construct/DirectProduct.agda
```diff
@@ -184,7 +184,7 @@ leftModule : {R : Ring r ℓr} →
              LeftModule R m′ ℓm′ →
              LeftModule R (m ⊔ m′) (ℓm ⊔ ℓm′)
 leftModule M N = record
-  { -ᴹ_ = map M.-ᴹ_ N.-ᴹ_
+  { -ᴹ_ = map (M.-ᴹ_) (N.-ᴹ_)
```
jamesmckinna, line 187: Do the parentheses help here? I'm not convinced...

src/Algebra/Module/Morphism/Construct/Composition.agda
```diff
@@ -39,23 +39,23 @@ module _
   isLeftSemimoduleHomomorphism f-homo g-homo = record
     { +ᴹ-isMonoidHomomorphism = isMonoidHomomorphism ≈ᴹ₃-trans F.+ᴹ-isMonoidHomomorphism G.+ᴹ-isMonoidHomomorphism
     ; *ₗ-homo                 = λ r x → ≈ᴹ₃-trans (G.⟦⟧-cong (F.*ₗ-homo r x)) (G.*ₗ-homo r (f x))
-    } where module F = IsLeftSemimoduleHomomorphism f-homo; module G = IsLeftSemimoduleHomomorphism g-homo
+    } where module F = IsLeftSemimoduleHomomorphism M₁ M₂ f-homo; module G = IsLeftSemimoduleHomomorphism M₂ M₃ g-homo
```
jamesmckinna, line 42: So I agree with you that it's annoying to have to be explicit with these arguments... I suppose that's one way of nudging the typechecker to know which scalars are involved at a given call-/use- site in this new setup, but perhaps simply supplying the `R`/`S` arguments would be the other way to do it?

src/Algebra/Module/Construct/DirectProduct.agda
Taneb: It's been a while, but I think without them Agda doesn't know if I mean that or a use of the `_-ᴹ_` operator

src/Algebra/Module/Morphism/Definitions.agda
```diff
@@ -9,7 +9,9 @@
 open import Relation.Binary.Core using (Rel)
 
 module Algebra.Module.Morphism.Definitions
-  {r} (R : Set r) -- The underlying ring
+  {r} (R : Set r) -- The underlying ring of the domain
+  {s} (S : Set s) -- The underlying ring of the codomain
+  ([_] : R → S)   -- The homomorphism between the underlying rings
```
jamesmckinna, line 14: As I've indicated before, i think this design is trying to do too many things at once which could be usefully factored into smaller pieces, such as #2913 ?

Taneb: Yes, I've come to agree with you on this.

Taneb:
Closing in favour of the approach described in #2913

src/Algebra/Module/Construct/DirectProduct.agda
jamesmckinna: Ah... I see, but still worth thinking about the earlier comment about having a small PR adding that subtraction operator  in its own right?

src/Algebra/Module/Morphism/Definitions.agda
jamesmckinna: OK, let's see what comes out in the wash!

PR #2811, opened 2025-08-20
Predomains

jamesmckinna:
Intended for comparison only with #2721 / #2809 by way of offering an 'alternative' (simpler?) design...

NB I'd be very tempted to refactor eg `UpperBound` and `MinimalUpperBoundAbove` in terms of intersection-closed predicates, ie to develop a theory of objects  '`≤`-minimal such that `P`' for predicates `P` under suitable closure conditions, but that's a tale for another day, as well as the theory of 'inclusions' between DC-'subsets' of a given preorder, so that the various lubs/mubs could be compared according to their indexing sets `I`... etc.

UPDATED: definition of Mub *might* require the additional precondition that `x` is a *lower* bound for the image of `f`, to rule out nonsensical definitions... oops!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,11 @@ Additions to existing modules
                     updateAt (padRight m≤n x xs) (inject≤ i m≤n) f ≡ padRight m≤n x (updateAt xs i f)
   ```
 
+* In `Relation.Binary.Definitions`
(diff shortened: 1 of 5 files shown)


jamesmckinna:
Closing this now in the interests of not further muddying the water.

Issue #2814, opened 2025-08-21
[ add ] Upwards- and downwards- closed subsets of a given ordered set


jamesmckinna:
DRAFT

* Subsets are a pain
* Sigma types are a pain
* Insisting on setoids  (and hence: Sigma types, else `Data.Refinement` types, but then the ergonomics of irrelevance gets differently painful?) rather than partial setoids to capture 'sub'-ness is also something of a pain...

Issue: it would be 'useful' infrastructure to add the objects of the issue title, but the design choices are... various, and variously debatable.

Added: label `subsets` because this may become a pervasive problem going forward?

JacquesCarette:
I'd say
- 'on the nose' subsets are a pain in dependent type theory, because they need to be encoded via a predicate
- subsets of X encoded as (Y, injection from Y to X) are actually not so bad to work with.

These are, of course, equivalent. But the ergonomics differ a lot. I guess this is yet more Fam-Pow stuff.

jamesmckinna:
Thanks @JacquesCarette . I guess I shouldn't have opened a 'draft' issue when my thoughts had not yet crystallised.
Trying to define `{y : A | x ≲ y}` initially stumbled on the realisation that any naive definition of this would only be a *partial* setoid if `A` were taken as the `Carrier`, and hence not able to be an instance of `IsPreorder` etc., because of our axiomatisation in terms of `IsEquivalence` as the basis for `Relation.Binary.Structures`...

... so one solution (a canonical choice of `Fam`-style injection: `proj₁`) would be first to define '`Setoid` from `PartialSetoid`' as a `Relation.Binary.Construct` (or else as additional lemmas in `Relation.Binary.Properties.PartialSetoid`... which I now got as far as developing), and then define upward closed subset in terms of the obvious partial equivalence structure? Cf. #2816

JacquesCarette:
My instinct here is that design smells strongly of classical-math / set-theoretical thinking, and may not lead to improvements in MLTT.

jamesmckinna:
Re: orthodox set-theoretical thinking.
I guess so, given that our `Function`/`Algebra` hierarchies are themselves based on `Setoid`s...
[...]

Issue #1899, opened 2023-01-03
Substructures and quotients in the `Algebra.*` hierarchy


jamesmckinna:
The recent discussion in issue #1888 concerning the correct definition of `Module` drew my attention to the (almost?) complete (?) lack of any treatment of algebraic substructures, and the corresponding notions of 'things-which-give-rise-to-quotients': 
- [ ] submonoid of a monoid;
- [ ] (normal) subgroup of a group;
- [ ] (left- and right-) ideal of a (non-commutative) ring;
- [ ] others? (eg do we care about submagmas? subsemirings etc.)

together with the associated 'free' things, viz.
- [ ] the submonoid generated by a subset; 
- [ ] the subgroup generated by ...; 
- [ ] the ideal generated by...; 
- [ ] the terminal ('zero'?) (sub)object as the free thing on the empty type of generators...; EDITED apologies for my ignorance of `Algebra.Construct.Zero` which does define these (but none of the associated homomorphisms); see also PR #1902
- [ ] etc. 

So this is (the beginnings of) a shopping list for the above, and some proposals for how to represent them. 

A left- (resp. right-) ideal of a `Ring R` with `Carrier` given by `A` should be given by: 
- a left- (resp. right-) `R`-module, with carrier type `I` for representing the subset in question; 
- an injective map `h : I -> A` which *is a left- (resp. right-) `R`-module homomorphism* 

TODO: 
- work out the details! (eg: injective map or monomorphism? are they same for `Setoid`s? etc. plus: level issues?)
- related matters: quotients, plus short/long exact sequences to characterise things? 
- what else?

JacquesCarette:
On the 'what else', Yasmine Sharoda's [PhD Thesis](https://github.com/ysharoda/PhD-Thesis) has a nice list of things that are well-enough understood that our original plan was to generate them all. Eventually [our code](https://github.com/ysharoda/Deriving-Definitions) generated fewer of these (in Agda and in Lean).  She mined a lot of universal algebra texts to find the most common constructions we should be looking at.

There's a lot of work to be still to be done to actually come up with a good shopping list, and even more work to find good definitions to actually use. The naive definitions in undergraduate textbooks tend to be flawed, and indeed one needs to look at category theory to find stable patterns (as already listed: terminal objects, initial objects, free objects -- but there are lots of very useful variations).

[...]

PR #2816, opened 2025-08-24
[ add ] `Setoid` from `PartialSetoid`

jamesmckinna:
This tackles one version of how to develop 'subsetoid's via PERs. Cf. #2814 

Not necessarily the 'right' way to go, but still functionality which may be useful in the future

UPDATED: to live under `Relation.Binary.Construct`

### DIFF
--- a/src/Relation/Binary/Construct/Kernel.agda
+++ b/src/Relation/Binary/Construct/Kernel.agda
@@ -1,0 +1,65 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Equaliser of a kernel pair in Setoid
+------------------------------------------------------------------------
+
+open import Relation.Binary.Bundles using (PartialSetoid; Setoid)
+
+module Relation.Binary.Construct.Kernel
+  {a i ℓ} {I : Set i}
+  (S : PartialSetoid a ℓ) (let module S = PartialSetoid S)
+  (f g : I → S.Carrier)
+  where
+
+open import Function.Base using (id; _∘_; _on_)
+open import Level using (Level; _⊔_)
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Definitions using (Defined)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Relation.Binary.Morphism.Structures
+  using (IsRelHomomorphism; IsRelMonomorphism)
+
+import Relation.Binary.Properties.PartialSetoid S as Properties
+
+
+------------------------------------------------------------------------
+-- Definitions
+
+record Carrier : Set (a ⊔ i ⊔ ℓ) where
+
+  field
+    h    : I
+    refl : (f h) S.≈ (g h)
+
+  ι : S.Carrier
+  ι = f h
+
+
+open Carrier public using (ι)
+
+_≈_ : Rel Carrier _
+_≈_ = S._≈_ on ι
+
+-- Structure
+
+isEquivalence : IsEquivalence _≈_
+isEquivalence = record
+  { refl = λ {x = x} → Properties.partial-reflˡ (Carrier.refl x)
+  ; sym = S.sym
+  ; trans = S.trans
+  }
+
+-- Bundle
+
+setoid : Setoid _ _
+setoid = record { isEquivalence = isEquivalence }
+
+-- Monomorphism
+
+isRelHomomorphism : IsRelHomomorphism _≈_ S._≈_ ι
+isRelHomomorphism = record { cong = id }
+
+isRelMonomorphism : IsRelMonomorphism _≈_ S._≈_ ι
+isRelMonomorphism = record { isHomomorphism = isRelHomomorphism ; injective = id }
+
--- a/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
+++ b/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
@@ -12,7 +12,6 @@ module Relation.Binary.Construct.SetoidFromPartialSetoid
 open import Data.Product.Base using (_,_; Σ; proj₁; proj₂)
 open import Function.Base using (id; _on_)
 open import Relation.Binary.Core using (Rel)
-open import Relation.Binary.Definitions using (Reflexive)
 open import Relation.Binary.Structures
   using (IsPartialEquivalence; IsEquivalence)
 open import Relation.Binary.Morphism.Structures
@@ -33,16 +32,11 @@ Carrier = Σ S.Carrier λ x → x S.≈ x
 _≈_ : Rel Carrier _
 _≈_ = S._≈_ on proj₁
 
--- Properties
-
-refl : Reflexive _≈_
-refl {x = x} = proj₂ x
-
 -- Structure
 
 isEquivalence : IsEquivalence _≈_
 isEquivalence = record
-  { refl = λ {x} → refl {x = x}
+  { refl = λ {x = x} → proj₂ x
   ; sym = S.sym
   ; trans = S.trans
   }
--- a/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
+++ b/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
@@ -9,8 +9,8 @@ open import Relation.Binary.Bundles using (PartialSetoid; Setoid)
 module Relation.Binary.Construct.SetoidFromPartialSetoid
   {a ℓ} (S : PartialSetoid a ℓ) where
 
-open import Data.Product.Base using (_,_; Σ; proj₁)
-open import Function.Base using (id)
+open import Data.Product.Base using (_,_; Σ; proj₁; proj₂)
+open import Function.Base using (id; _on_)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Definitions using (Reflexive)
 open import Relation.Binary.Structures
@@ -31,12 +31,12 @@ Carrier : Set _
 Carrier = Σ S.Carrier λ x → x S.≈ x
 
 _≈_ : Rel Carrier _
-x≈x@(x , _) ≈ y≈y@(y , _) = x S.≈ y
+_≈_ = S._≈_ on proj₁
 
 -- Properties
 
 refl : Reflexive _≈_
-refl {x = _ , x≈x} = x≈x
+refl {x = x} = proj₂ x
 
 -- Structure
 
--- a/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
+++ b/src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
@@ -9,34 +9,37 @@ open import Relation.Binary.Bundles using (PartialSetoid; Setoid)
 module Relation.Binary.Construct.SetoidFromPartialSetoid
   {a ℓ} (S : PartialSetoid a ℓ) where
 
-open import Data.Product.Base using (_,_; Σ; proj₁; proj₂)
 open import Function.Base using (id; _on_)
+open import Level using (_⊔_)
 open import Relation.Binary.Core using (Rel)
 open import Relation.Binary.Structures
-  using (IsPartialEquivalence; IsEquivalence)
+  using (IsEquivalence)
 open import Relation.Binary.Morphism.Structures
   using (IsRelHomomorphism; IsRelMonomorphism)
 
 private
   module S = PartialSetoid S
 
-  variable
-    x y z : S.Carrier
 
 ------------------------------------------------------------------------
 -- Definitions
 
-Carrier : Set _
-Carrier = Σ S.Carrier λ x → x S.≈ x
+record Carrier : Set (a ⊔ ℓ) where
+
+  field
+    ι    : S.Carrier
+    refl : ι S.≈ ι
+
+open Carrier public using (ι)
 
 _≈_ : Rel Carrier _
-_≈_ = S._≈_ on proj₁
+_≈_ = S._≈_ on ι
 
 -- Structure
 
 isEquivalence : IsEquivalence _≈_
 isEquivalence = record
-  { refl = λ {x = x} → proj₂ x
+  { refl = λ {x = x} → Carrier.refl x
   ; sym = S.sym
   ; trans = S.trans
   }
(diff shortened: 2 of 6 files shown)


src/Relation/Binary/Properties/PartialSetoid.agda
```diff
@@ -45,3 +53,37 @@ partial-reflexiveˡ p ≡.refl = partial-reflˡ p
 partial-reflexiveʳ : x ≈ y → y ≡ z → y ≈ z
 partial-reflexiveʳ p ≡.refl = partial-reflʳ p
 
+------------------------------------------------------------------------
+-- Setoids from partial setoids
```
JacquesCarette: My instinct would be to put all of these in a (named) module rather than adding more things to the exported namespace of this module. That would also obviate the need for the awkward names with subscripts.

Also: definitely a construction. In fact, why not build the `Setoid` bundle directly?

JacquesCarette (changes requested):

src/Relation/Binary/Properties/PartialSetoid.agda
jamesmckinna: Are you saying that I 'just' build a value of type `Setoid`, and not explicitly define all the other fields at top-level? 
I'm not quite sure what you mean otherwise, but it would be good to be clear...

Our custom to date seems to be to such structures/bundle explicitly piece-by-piece, but my suggestion in #2391 was that, at least for *additional* structure, that we could apply a `Construct`, and then publicly re-open it to expose its subfields. But that's not  the same as making a `Construct` definition from scratch, as here?

jamesmckinna: In any case, will move to `Relation.Binary.Construct.SetoidFromPartialSetoid`, unless a better name is available/recommended?

jamesmckinna: Or did you mean to build the relevant `Function.Bundles` object(s) directly from `isRelHomomorphism`? That's not possible, precisely because the `Function` hierarchy is based on `Setoid`s as source and target, ... and we only have the former here!

JacquesCarette: > Are you saying that I 'just' build a value of type Setoid, and not explicitly define all the other fields at top-level?

Yes. Because the pieces are so trivial.

Having all the pieces exposed in `SetoidFromPartialSetoid` does make sense. It's just pollution here.

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
```diff
@@ -0,0 +1,61 @@
...
+    x y z : S.Carrier
+
+------------------------------------------------------------------------
+-- Definitions
+
+Carrier : Set _
+Carrier = Σ S.Carrier λ x → x S.≈ x
+
+_≈_ : Rel Carrier _
+x≈x@(x , _) ≈ y≈y@(y , _) = x S.≈ y
```
JacquesCarette:
```
x ≈ y = proj₁ x S.≈ proj₁ y
```
?  Or even 
```
_≈_ = S._≈_ on proj₁
```
I certainly don't like binding names that are not used on the rhs as you do here.

JacquesCarette (changes requested):
One quibble, but otherwise I like this now.

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
```diff
@@ -0,0 +1,61 @@
...
+--
+-- Conversion of a PartialSetoid into a Setoid
+------------------------------------------------------------------------
+
+open import Relation.Binary.Bundles using (PartialSetoid; Setoid)
+
+module Relation.Binary.Construct.SetoidFromPartialSetoid
+  {a ℓ} (S : PartialSetoid a ℓ) where
+
+open import Data.Product.Base using (_,_; Σ; proj₁; proj₂)
```
jamesmckinna: I don't know if it would be worse, or better, to have something like
```suggestion
open import Data.Product.Base using (_,_; Σ)
  renaming (proj₁ to ι; proj₂ to x≈x)
```
so that subsequent definitions can exploit such 'abstract' vocabulary/notation, rather than expose the concrete projection names?

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
```diff
@@ -0,0 +1,61 @@
...
+Carrier : Set _
+Carrier = Σ S.Carrier λ x → x S.≈ x
+
+_≈_ : Rel Carrier _
+_≈_ = S._≈_ on proj₁
+
+-- Properties
+
+refl : Reflexive _≈_
+refl {x = x} = proj₂ x
```
jamesmckinna: Given Agda's inability to profit from making this an explicit definition in the `IsEquivalence.refl` field below, maybe this should be removed, and the definition inlined there?

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
jamesmckinna: Or perhaps better still, remove/inline `refl` as below, and make 
```agda
ι : Carrier → S.Carrier
ι = proj₁
```
so that it may subsequently be exported/exportable as such?

JacquesCarette: If you're going to go that route (second suggestion), why not bite the bullet and make the thing a record? Post-facto adding aliases for projections is what `agda-unimath` does... and it eventually leads to unreadable goals.

I like the first suggestion even less.

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
JacquesCarette: Probably.

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
jamesmckinna: Well, in a way, the `record`-based version *is* (arguably) cleaner:
```agda
record SubSetoid : Set (a ⊔ ℓ) where

  field
    ι    : S.Carrier
    refl : ι S.≈ ι

open SubSetoid public using (ι)

_≈_ : Rel SubSetoid _
_≈_ = S._≈_ on ι

-- Structure

isEquivalence : IsEquivalence _≈_
isEquivalence = record
  { refl = λ {x = x} → SubSetoid.refl x
  ; sym = S.sym
  ; trans = S.trans
  }

-- Bundle

setoid : Setoid _ _
setoid = record { isEquivalence = isEquivalence }

-- Monomorphism

isRelHomomorphism : IsRelHomomorphism _≈_ S._≈_ ι
isRelHomomorphism = record { cong = id }

isRelMonomorphism : IsRelMonomorphism _≈_ S._≈_ ι
isRelMonomorphism = record { isHomomorphism = isRelHomomorphism ; injective = id }
```

JacquesCarette: I certainly would argue so!

jamesmckinna: In the interests of clarity: are you saying that you would prefer that it be changed in favour of the `record` version? 

More than happy to do so if so!

JacquesCarette: Yes

jamesmckinna: Done. I've retained `Carrier` as the name of the underlying type, but I suppose something else, such as `SubSetoid`, would work as well. Field name `refl` is kept hidden, so that its 'rectification' (wrt implicit quantification) as `Setoid.refl` is the way to access it.

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
```diff
@@ -0,0 +1,58 @@
...
+  using (IsRelHomomorphism; IsRelMonomorphism)
+
+private
+  module S = PartialSetoid S
+
+
+------------------------------------------------------------------------
+-- Definitions
+
+record Carrier : Set (a ⊔ ℓ) where
```
JacquesCarette, line 27: I must say that I expected the carrier to be called `Carrier` (and not `ι`) with the record called something like 'Reflective' or some such. 

Thinking some more, I think having a definition for `Defined x` (expanding to `x S.≈ x`) might make the overall definition more perspicuous.

JacquesCarette (changes requested):

src/Relation/Binary/Construct/SetoidFromPartialSetoid.agda
jamesmckinna: So... 

* `Carrier` (capitalised) is the carrier *type*; did you envisage `carrier` (lower-case) for the name of its inhabitant?
I was motivated by my proposal under #1899 to have a standardised name which then becomes the underlying *function* of the homo-/mono-morphism injecting `Carrier` into `S.Carrier`... but part of my thinking was also that the record itself doesn't merit an 'interesting' name... because it should only be exposed through its `Setoid` API, when it would moreover, again receive the name `Carrier`... (and be renamed, presumably, at any client use-site, if needed)
* adding `Defined` sounds a good suggestion for `Relation.Binary.Definitions`... DRY says we should then redefine `Reflexive` in terms of it?

~~On naming, I think I was struck looking back over #2071 that the carrier type of the symmetric interior is given an 'interesting'/'informative' name, when in fact that's another case (I now think), where the `record` type is a concrete implementation detail which should only really be accessed/accessible via the API for the `Construct` being defined... esp. given that the module import path *already* cues the construction as that of the symmetric interior...~~

UPDATED: I've gone back over that last paragraph, and it doesn't any longer make sense!? Sorry for the noise!

Let me retry : 
#2071 defines a construction *on binary relations*, not on *types*, whereas `Carrier` constructs a type towards building a `Setoid`, so it's not obvious that it belongs under `Construct`? 

The operating construct on binary relations involved here is simply `_on_`, but I'd hesitate before putting the derived constructions as `Properties` of `Relation.Binary.Construct.On`...?

jamesmckinna: D'oh!?

I've ended up defining
`subset` = `equaliser of kernel pair`
for a/the special case of the 'inclusion'...

So it is probably better to refactor this as
`kernel pair of arbitrary function` (`Fam` again!?), and then specialise that at `id`?

Or: take a breath and sit on my hands for a bit...

JacquesCarette: Good realization - I certainly had not spotted that. Perhaps sitting is indeed wise.

jamesmckinna:
UPDATED: 
* converted to `DRAFT`
* added `Relation.Binary.Construct.Kernel` and `Relation.Binary.Construct.SubSetoid` for consideration

We can discuss where this might go, if and when, ... later.

src/Relation/Binary/Construct/Kernel.agda
```diff
@@ -0,0 +1,65 @@
...
+open import Relation.Binary.Morphism.Structures
+  using (IsRelHomomorphism; IsRelMonomorphism)
+
+import Relation.Binary.Properties.PartialSetoid S as Properties
+
+
+------------------------------------------------------------------------
+-- Definitions
+
+record Carrier : Set (a ⊔ i ⊔ ℓ) where
```
JacquesCarette, line 29: I'm still not happy with this record being called `Carrier` and its 'carrier' field being called `h`. To me, `Carrier` will always mean the underlying type of something `Setoid`-based. So re-using that name confounds expectations.

It feels like `Carrier` is instead some kind of `Index`-type with a distinguished point `h`.

JacquesCarette (changes requested):

src/Relation/Binary/Construct/Kernel.agda
jamesmckinna: > I'm still not happy with this record being called `Carrier` and its 'carrier' field being called `h`. To me, `Carrier` will always mean the underlying type of something `Setoid`-based. So re-using that name confounds expectations.

@JacquesCarette I'm reluctant to antagonise the collective wisdom of McMaster any more than I already seem to be doing over order-theoretic concepts, so I'll try to restate my position FTR, and then happily let others decide what becomes of this:
* one of the end products of this PR is precisely to establish `Carrier` (the `record` type) as the `Carrier` (field value) of "something `Setoid` based", namely the bundle defining the 'subset on which `f` and `g` equalise'
* it's clear you don't like the pun; fair enough, but I would *never* expect this type to be explicitly exported, except through the `Setoid` interface (cf. #2391 ); I guess I could make it `private`, to improve matters? Alternatively, it could indeed be renamed, but as I say, I regard it as an abstract data type, existing solely to be a choice of domain for the `isRelMonomorphism` definition... viz. as a way of speaking about 'the subset on which `f` and `g` equalise'
* it's conventional, given `f`, and `g` defining some diagram (here: a parallel pair with domain `I`, as per the 'sub-thing as given by a map on indices' construction, to have `h`, or perhaps better here, `k` (for `Kernel`, except that old-timey category theory books tend to use `h` for the universal equalising map, and `k` for the corresponding *co*equalising map in a coequaliser diagram) define the (appropriate) unique mediating morphism (I won't debate uniqueness or otherwise here, given higher-minded, and higher-dimensional, sensitivities about 'uniqueness up to *what*?'); given the way that `field` name become projection functions, that, again precisely, establishes `h`, or whatever we end up calling it, as a *function* of type `Carrier → I`, and again, conventionally, in such a way as to fit the diagram of parallel arrows `f`, `g`...
* I've proposed/queried elsewhere the use of a standardised idiom for describing the eventual map `ι : Carrier → S.Carrier`, as, again precisely, corresponding to 'the injection into `S` of the subset on which `f` and `g` equalise', but again, the *actual* name seems really neither here nor there, but for the fact that, again from the `public` exports, `h` is never again mentioned (so *its* name is similarly immaterial/abstract)

I'd be quite happy to say 'puns considered harmful', and rename everything, but for the fact that these objects/constructions giving rise to the (intended for `public` export) definitions
* `_≈_`
* `isEquivalence`
* `setoid`
* `isRelMonomorphism`

*do* all need to be named. Perhaps it was failure of imagination on my part not to be able to come up with better ones, but in the end, I thought, it turns out mistakenly, that the ones I chose, for the reasons above, made the best 'fit'.

HTH!

> It feels like `Carrier` is instead some kind of `Index`-type with a distinguished point `h`.

See above: `h` defines, as a projection, a *function* from `Carrier` to `I`.

JacquesCarette: Making it `private` would go part of the way, for sure. Some puns I like, but somehow not this one.

The argument "establishing this as the `Carrier` of ..." doesn't go very far, as it could be extended to so many more things. 

Have you considered `Equalizer` as the name? `EqualAt`, if we want a less fancy name?

jamesmckinna:
@JacquesCarette can we try to progress this, or should I close for the time being?

JacquesCarette:
We can try to make progress.

PR #2555, opened 2025-01-16
Add a few algebraic structures missing from the Algebra.Construct.Pointwise

bsaul:
This PR adds a few structures/bundles currently missing from `Algebra.Construct.Pointwise`, namely:

* `isNearSemiring`
* `isSemiringWithoutOne`
* `isCommutativeSemiringWithoutOne`
* `isCommutativeSemiring`
* `isIdempotentSemiring`
* `isKleeneAlgebra`
* `isQuasiring`
* `isCommutativeRing`

Issue #1907, opened 2023-01-11
`DirectProduct`s etc. in the `Algebra.*` hierarchy


jamesmckinna:
`Algebra.Construct.DirectProduct` and `Algebra.Module.Construct.DirectProduct` define some objects which are the *binary* products of the appropriate algebras, which are moreover in some case also biproducts (products and coproducts together). 

This is fine as far as it goes, but does not extend to `I`-ary products and sums of (at least) modules (sums have only finitely-many non-zero components, whereas products can have arbitrary `I`-ary functions in the carrier)

Analogously to #1906 / #1902 , suggest a refactor to put things 'in the right place': 
* define `Product` to introduce all the product algebras; 
* define `Sum` (or `Coproduct`) to introduce the coproducts;
* re-export those which *are* biproduct objects in`DirectProduct`; 
* consider deprecating that module in favour of `Biproduct`

It seems easiest perhaps to begin by localising things first in `Algebra.Module.Construct.*` as a prototype for the the larger task in `Algebra.Construct.*`

JacquesCarette:
Agda will have the best engineered library of them all! (I mean that).

jamesmckinna:
I'll take that as encouragement!

JacquesCarette:
That's how it was meant. What I really want to do next is to emulate (the good parts of) Kevin Buzzard and put as many students on implementing things as possible. The more ready designs we have, the faster that will go.

jamesmckinna:
Suggest that this wait until after the v2.0 release. ;-)

jamesmckinna:
And its now over a year since I last commented... or (anyone) did anything about this... must remedy that!

jamesmckinna:
> And its now over a year since I last commented... or (anyone) did anything about this... must remedy that!

*bis repetita placent*

jamesmckinna:
Preliminary investigation of #2174 in #3081 stalls on the mismatch between the `Data.Product` definitions underlying the current implementation of `DirectProduct` and the general picture sketched above. 

So while these issues are not coupled as such, working on this one might unblock/remove the need for extensive re-engineering as a knock on consequence of redefining `Pointwise` as a record....

PR #2817, opened 2025-08-28
[ add ] Pointwise lifting of algebra to `Data.Vec.Functional` (Functional vector module #1945 redux)

e-mniang:
This PR updates #1945 by bringing the functional vector and module structures/bundles in line with current stdlib conventions and removing nested submodules.

Next, I’ll add the necessary entry to the CHANGELOG.

Reviews and comments are welcome.


src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,142 @@
...
+-- Binary operation lifting
+lift₂ᴹ : {A : Set a} → Op₂ A → Op₂ (Vector A n)
+lift₂ᴹ _∙_ = zipWith _∙_
+
+-- Unary operation lifting
+lift₁ᴹ : {A : Set a} → Op₁ A → Op₁ (Vector A n)
+lift₁ᴹ f = map f
+
+-- Vector RawMagma construction
+VecRawMagma : RawMagma a ℓ → (n : ℕ) → RawMagma a ℓ
```
JacquesCarette: I would use (say) `Data.Sum.Algebra` as a guide for naming conventions of these.

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,142 @@
...
+
+-- Vector RawRing construction
+VecRawRing : RawRing a ℓ → (n : ℕ) → RawRing a ℓ
+VecRawRing R n =
+  record
+    { RawSemiring (VecRawSemiring R.rawSemiring n)
+    ; -_ = lift₁ᴹ R.-_
+    }
+  where module R = RawRing R
+
```
JacquesCarette, line 81: remove extra blanks

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,142 @@
...
+
+
+-- Left scalar action obtained from the ring/semiring multiplication
+_*ₗ_ : {S : Set a} → Op₂ S → Opₗ S (Vector S n)
+_*ₗ_ _*_ r xs = map (λ x → _*_ r x) xs
```
JacquesCarette:
```suggestion
_*ₗ_ _*_ r xs = map (r *_) xs
```

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,142 @@
...
+
+-- Right scalar action (same underlying Op₂, but as a right action)
+_*ᵣ_ : {S : Set a} → Op₂ S → Opᵣ S (Vector S n)
+_*ᵣ_ _*_ xs r = map (λ x → _*_ x r) xs
```
JacquesCarette:
```suggestion
_*ᵣ_ _*_ xs r = map (_* r) xs
```

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,734 @@
...
+    module ≈ᴹ = ADefinitions RNS._≈_
+
+  vec1# : Vector NS.Carrier n
+  vec1# = λ _ → SR.1#
+
+  isSemiringVec : IsSemiring RNS._≈_ RNS._+_ RNS._*_ RNS.0# vec1#
+  isSemiringVec = record {
+    isSemiringWithoutAnnihilatingZero = record
+      { +-isCommutativeMonoid = isCommutativeMonoid n +-isCommutativeMonoid
+      ; *-cong    = λ x≈y u≈v i → *-cong (x≈y i) (u≈v i)
```
JacquesCarette: there are many combinators in `Function.Base` that could be used here.

JacquesCarette (changes requested):
A few minor style things, but this looks quite good.

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,735 @@
...
+
+  private
+    *ₗ-*ᵣ-coincident : SimultaneousBiDefs.Coincident Carrier RNS._≈_ LSM._*ₗ_ RSM._*ᵣ_
+    *ₗ-*ᵣ-coincident r xs i = *-comm r (xs i)
+
+  isModule : IsModule commutativeRing RNS._≈_ RNS._+_ RNS.0# VG._⁻¹ LSM._*ₗ_ RSM._*ᵣ_
+  isModule = record
+    { isBimodule = isBimodule
+    ; *ₗ-*ᵣ-coincident = λ r xs i → *-comm r (xs i)
+    }
```
jamesmckinna: This is a striking DRY violation, unless the typechecker is (now) unable to check
```agda
isModule = record
    { isBimodule = isBimodule
    ; *ₗ-*ᵣ-coincident = *ₗ-*ᵣ-coincident
  }
```

It might be! In which case, remove the whole `private` block?

jamesmckinna:
No review as such, but an observation about a global problem with both `Base` and `Properties`, echoing @JacquesCarette 's 'style' comment: in many/most/all(?) of the definitions, there is a need to standardise apart the name of a derived construction from that of its 'base' counterpart, and this PR adopts three distinct strategies to achieve this
* superscript `̌`
* `vec-`/`Vec` prefixing
* `base` as a generic name for the 'base' notion (eg `IsCommutativeMonoid`), and the 'usual' `isCamelCase : IsCamelCase` idiom for naming the derived notion

Some of this seems to come about because of name clashes, either directly, or as a result of local `open`ing of a base `record`.
It would be great if, even at the cost of some slightly more involved qualified imports/private module declarations, that such things could be avoided, moreover in such a way as to make the above '(in)elegant variation' go away...

Suggest also that, for all these update/'redux' PRs:
* the title of the PR be a copy/tidying up of the original, but with the additional pointer to the original as you have it, eg here "[ add ] `Pointwise` lifting of algebra to `Data.Vec.Functional` (#1945 redux)", rather than the more gnomic "Update of...", which requires the reader to follow one level of indirection in order to know what's going on...

src/Data/Vec/Functional/Algebra/Base.agda
jamesmckinna: Also, worth adding some global `variable` declarations to avoid having to mention `{S : Set a}` prefixes in types at all?

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+open import Data.Vec.Functional
+open import Algebra.Core
+open import Algebra.Bundles
+open import Algebra.Module
+open import Relation.Binary using (Rel)
+open import Data.Vec.Functional.Relation.Binary.Pointwise using (Pointwise)
+
+private variable
+  a ℓ : Level
+  A : Set ℓ
```
jamesmckinna: Conventionally:
```suggestion
  A : Set a
```

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+-- Vector-lifted relations & operators
+
+pointwiseᵛ : {A : Set a} ( _≈_ : Rel A ℓ ) → Rel (Vector A n) ℓ
+pointwiseᵛ _≈_ = Pointwise _≈_
+
+zipWithᵛ : {A : Set a} → Op₂ A → Op₂ (Vector A n)
+zipWithᵛ _∙_ = zipWith _∙_
+
+mapᵛ : {A : Set a} → Op₁ A → Op₁ (Vector A n)
+mapᵛ f = map f
```
jamesmckinna: So that all these initial `{A : Set a}` prefixes can be avoided!

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,735 @@
...
+open import Data.Nat using (ℕ)
+open import Data.Fin using (Fin; zero; suc)
+open import Function.Base using (_∘′_; _ˢ_; _∘_)
+open import Data.Vec.Functional
+open import Algebra.Core
+open import Algebra.Bundles
+open import Algebra.Module
+open import Algebra.Module.Structures
+open import Relation.Binary
+import Algebra.Definitions as ADefinitions
```
jamesmckinna: No other `Definitions` modules are in (qualified) scope, so why not simply
```suggestion
import Algebra.Definitions as Definitions
```

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+private variable
+  a ℓ : Level
+  A : Set a
+  n : ℕ
+
+------------------------------------------------------------------------
+-- Vector-lifted relations & operators
+
+pointwiseᵛ : ( _≈_ : Rel A ℓ ) → Rel (Vector A n) ℓ
+pointwiseᵛ _≈_ = Pointwise _≈_
```
jamesmckinna: Is this (redundant) repetition worth the gain (eg in uniformity of naming?)
I'd be tempted to do two things here:
* make the `import` on L20 qualified, `as Pointwise`
* inline all uses of `pointwiseᵛ` in favour simply of `Pointwise` as it stands...

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+
+zipWithᵛ : Op₂ A → Op₂ (Vector A n)
+zipWithᵛ _∙_ = zipWith _∙_
+
+mapᵛ : Op₁ A → Op₁ (Vector A n)
+mapᵛ f = map f
```
jamesmckinna: Ditto. (and all the more so, now that 3 imported definitions are being subject to the same treatment...)

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+
+------------------------------------------------------------------------
+-- Vector-lifted raw structures
+
+rawMagmaᵛ : RawMagma a ℓ → (n : ℕ) → RawMagma a ℓ
+rawMagmaᵛ M n =
```
jamesmckinna: Now that M is being referenced only *locally* via the `where` clause, again... I'd drop the superscript.

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+    { RawSemiring (rawSemiringᵛ R.rawSemiring n)
+    ; -_ = mapᵛ R.-_
+    }
+  where module R = RawRing R
+
+------------------------------------------------------------------------
+-- Vector actions (left/right scalar multiplication)
+
+_*ₗᵛ_ : {S : Set a} → Op₂ S → Opₗ S (Vector S n)
+_*ₗᵛ_ _*_ r xs = map (r *_) xs
```
jamesmckinna: Two suggestions:
* eta-contract (no need to mention `xs` on either LHS or RHS)
* now you've used `map`, rather than `mapᵛ`, its alias...

... so I'd go all the way now:
```suggestion
_*ₗ_ : {S : Set a} → Op₂ S → Opₗ S (Vector S n)
_*ₗ_ _*_ r = map (r *_)
```

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,139 @@
...
+
+_*ᵣᵛ_ : {S : Set a} → Op₂ S → Opᵣ S (Vector S n)
+_*ᵣᵛ_ _*_ xs r = map (_* r) xs
```
jamesmckinna: Ditto.

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,725 @@
...
+------------------------------------------------------------------------
+-- Raw structure properties
+
+module RawMagmaProperties (rawMagma : RawMagma a ℓ) (n : ℕ) where
+  private
+    magmaᵛ = rawMagmaᵛ rawMagma n
+    module VM = RawMagma magmaᵛ
+    module M = RawMagma rawMagma
+
+  open IsEquivalence
```
jamesmckinna: Is this needed?

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,725 @@
...
+
+  isMagmaᵛ : IsMagma M._≈_ M._∙_ → IsMagma VM._≈_ VM._∙_
+  isMagmaᵛ base = record
+    { isEquivalence = flip Pointwise.isEquivalence _ B.isEquivalence
+    ; ∙-cong        = λ x≈y u≈v i → B.∙-cong (x≈y i) (u≈v i)
+    }
+    where module B = IsMagma base
```
jamesmckinna: Now you've chosen `base` for the 'base' argument, you won't need the superscript...?

As for `module B = IsMagma base` why change the names `module CM = ...`, `module SG = ...` of the subsequent locally-bound modules... when `B` would suffice in those instances as well?

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,725 @@
...
+    }
+    where module CS = IsCommutativeSemigroup base
+
+------------------------------------------------------------------------
+
+module RawMonoidProperties (rawMonoid : RawMonoid a ℓ) (n : ℕ) where
+  private
+    monoidᵛ = rawMonoidᵛ rawMonoid n
+    module VM = RawMonoid monoidᵛ
+    module RM = RawMonoid rawMonoid
```
jamesmckinna: Again, I see no use of `monoidᵛ` besides the definition of `module VM`, so consider inlining it:
```suggestion
    module VM = RawMonoid (rawMonoid M n)
    module M = RawMonoid M
```
etc. No need for `RM` here?

src/Data/Vec/Functional/Algebra/Properties.agda
```diff
@@ -0,0 +1,725 @@
...
+    }
+
+  isModule : IsModule commutativeRing RNS._≈_ RNS._+_ RNS.0# VG._⁻¹ LSM._*ₗ_ RSM._*ᵣ_
+  isModule = record
+    { isBimodule = isBimodule
+    ; *ₗ-*ᵣ-coincident = λ r xs i → *-comm r (xs i)
+    }
+
+  module' : Module _ _ _
+  module' = record { isModule = isModule }
```
jamesmckinna: For the underlying name of the bundle, `stdlib` uses in eg. `Algebra.Module.Construct.TensorUnit` the name `⟨module⟩`, and I would follow suit here, for the sake of consistency.

jamesmckinna:
Thanks for the updates @e-mniang ! My recent round of nitpicks really home in on wanting to avoid the superscripts altogether:
* for the function/relation names on `Vector`s, the existing infrastructure suffices?
* for your additions of structure/bundle inheritance, with careful choice of parameter names, there's no need either!

Hope this 
* is clear
* helps!

JacquesCarette:
FYI: @e-mniang says she'll finish this once she's established a proper rhythm with this term's courses, back in France.

jamesmckinna:
> FYI: @e-mniang says she'll finish this once she's established a proper rhythm with this term's courses, back in France.

Safe journeys back! 

Given the current merge hiatus thanks to GitHub, it seems fine to also have a hiatus in fretting too much about when updates get pushed...

... while commentary remains the stream that can never be shut off!

e-mniang:
Hi! Thanks for the comments. I’ve made changes to address each point; I hope it’s in good shape now.

jamesmckinna:
`fix-whitespace src/Data/Vec/Functional/Algebra/Base.agda` (DONE)

jamesmckinna:
@JacquesCarette I've fixed the whitespace and `CHANGELOG` issues; can you re-review ahead of merging?

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Some Vector-related module Definitions
+-- Some VecF.Vector-related module Definitions
```
JacquesCarette, line 4: While importing `Data.Vec.Functional` as `VecF` makes sense, `VecF` in the top-comment does not. I would expand this out to "Functional Vector-related"

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -12,128 +12,116 @@ open import Level using (Level; suc; _⊔_)
...
+import Data.Vec.Functional.Relation.Binary.Pointwise as Pointwise
 
 private variable
   a ℓ : Level
   A : Set a
   n : ℕ
 
 ------------------------------------------------------------------------
--- Vector-lifted relations & operators
+-- VecF.Vector-lifted raw structures
```
JacquesCarette, line 28: Same comment. I would be tempted to go with just "Lifted raw structures" or if one wants more, "Raw structures lifted to Functional Vectors".

JacquesCarette (changes requested):
Not so happy with the changes to the comments, happy with the rest.

jamesmckinna:
While merging is blocked, can we (re)consider where all this material belongs?
* currently, under `Data.Vec.Functional.Algebra.*`
* but should it, in fact, belong better under `Algebra.Construct.FunctionalVector.*`

The latter emphasises that we have constructions on algebras, rather than being 'accidental' algebraic consequences for functional vectors... but I confess I find the question hard to resolve, and its taken this long even for it to crystallise...

src/Data/Vec/Functional/Algebra/Base.agda
```diff
@@ -0,0 +1,127 @@
...
+
+module Data.Vec.Functional.Algebra.Base where
+
+open import Level using (Level; suc; _⊔_)
+open import Function using (_$_)
+open import Data.Nat using (ℕ)
+open import Data.Fin using (Fin)
+import Data.Vec.Functional as VecF
+open import Algebra.Core
+open import Algebra.Bundles
```
JacquesCarette, line 17: `using` , here and below?

JacquesCarette (changes requested):

JacquesCarette:
I like the current location. Because, to me `Algebra.Construct` is where *constructions* belong. What you do here is "the algebra (and properties) of functional vectors", so `Data.Vec.Functional.Algebra` seems like a great home (to me).

src/Data/Vec/Functional/Algebra/Base.agda
jamesmckinna: I presume that this is (something like) a global-search-and-replace that didn't get weeded out?

CHANGELOG.md
```diff
@@ -61,6 +61,8 @@ New modules
 
 * `Data.List.Relation.Binary.Permutation.Declarative{.Properties}` for the least congruence on `List` making `_++_` commutative, and its equivalence with the `Setoid` definition.
 
+* `Data.Vec.Functional.Algebra(.{Base|Properties})` - structures and bundles about functional vectors and modules.
```
jamesmckinna:
```suggestion
* `Data.Vec.Functional.Algebra.{Base|Properties}` - structures and bundles about functional vectors and modules.
```
I think that the extra parentheses are... too much!

jamesmckinna:
I've been revisiting this, with a view to tidying up to be able to merge it in, but I'm now having second thoughts:
* for `Algebra.Structures.IsX`/`Algebra.Bundles.X`, I'm happy with the theorems for bundles as witnessed here: if the `Carrier` of an `X`-algebra admits an `IsX` structure, then `Vector n Carrier` also admits an `IsX` structure (though this PR is a very telescoped down version of the more elaborate #1945 which preceded it)
* But for `Algebra.Module.Bundles.Y`, I think this PR subtly misses the point: rather than have that `M` admits an `R{,S}-IsYmodule` structure (and hence bundles... etc.) implies that `Vector n M.Carrier` also carries an `R{,S}-IsYmodule` structure (and bundles...), this PR restricts to the special case where `M` is given by *`R` itself*, which can be seen as a instantiation of a more general lemma to the special case of the `TensorUnit` construction: where `R` acts on itself to give rise to a suitable `IsYmodule` structure... UPDATED and this seems to have been a defect inherited from #1945 

So: suggest that we refactor this so as to
* either only admit the part of the PR concerning the `X`-algebra constructions
* or else generalise the `Y`-module development to make the parametrisation appropriately general (see below)

What's the best way to proceed?

In any case, I don't think I can give an `Approve` review yet... :-(

jamesmckinna:
To give an indication of (perhaps) what the module definitions might look like:
```agda
module _ (M : RawLeftSemimodule R a ℓ) where

  private
    module M = RawLeftSemimodule M

  rawLeftSemimodule : (n : ℕ) → RawLeftSemimodule R a ℓ
  rawLeftSemimodule n =
    record
      { Carrierᴹ = Vector M.Carrierᴹ n
      ; _≈ᴹ_     = Pointwise M._≈ᴹ_
      ; _+ᴹ_     = zipWith M._+ᴹ_
      ; _*ₗ_     = λ r → Vector.map (r M.*ₗ_)
      ; 0ᴹ       = λ _ → M.0ᴹ
      }
 ```
where `R` is globally bound simply as a `private variable R : Set r`... (as in the definition of `RawLeftSemimodule`) rather than as a `Raw{Near}Semiring`... which would otherwise be required by the `TensorUnit` construction.

UPDATED: there may even be a question here (@Taneb can you comment?) as to the 'correct' parametrisations in `Algebra.Module.Construct.TensorUnit`... the current version perhaps demands too much of the underlying scalar algebra when defining the `{Left|Right}SemiY` and `PreY` constructions?

MORE UPDATES: aaargh. The above construction is strictly less general that the `zipWith` version, but so too are the existing ones... back to the drawing board! Sigh.

**Comment by jamesmckinna** [comment]
More thoughts, now on `Properties`...
* I'm not sure the design of these incremental modules follows the hierarchical design principles for the main `Algebra` hierarchy...
* ... specifically, `+-isMonoid` recapitulates a construction already under `RawMonoidProperties`, by delegation to `SemigroupProperties`; I think the 'usual' hierarchy would inherit from (all of) `RawMonoidProperties`, and then add back in the derived `Semigroup` properties not specifically covered by such re-export?
* Ditto. `+-isGroup` etc. for `GroupProperties` etc.

**Comment by jamesmckinna** [comment]
Revised (and now hopefully improved!) versions of the `YModule` constructions:
```agda
rawLeftSemimodule-diag : (M : RawLeftSemimodule R a ℓ) (n : ℕ) →
                         RawLeftSemimodule R a ℓ
rawLeftSemimodule-diag M n = record
  { Carrierᴹ = Vector M.Carrierᴹ n
  ; _≈ᴹ_     = Pointwise M._≈ᴹ_
  ; _+ᴹ_     = zipWith M._+ᴹ_
  ; _*ₗ_     = λ r m → Vector.map (r M.*ₗ_) m
  ; 0ᴹ       = λ _ → M.0ᴹ
  }
  where module M = RawLeftSemimodule M

rawLeftSemimodule : (M : RawLeftSemimodule R a ℓ) (n : ℕ) →
                    RawLeftSemimodule (Vector R n) a ℓ
rawLeftSemimodule M n = record
  { Carrierᴹ = Vector M.Carrierᴹ n
  ; _≈ᴹ_     = Pointwise M._≈ᴹ_
  ; _+ᴹ_     = zipWith M._+ᴹ_
  ; _*ₗ_     = zipWith M._*ₗ_
  ; 0ᴹ       = λ _ → M.0ᴹ
  }
  where module M = RawLeftSemimodule M
```

The idea being that the `*-diag` versions keep the underlying (semi)ring of scalars fixed, while the `*` versions also vectorise the scalars.

The `-diag` suffix because these forms amount to composing the vectorised version with the pullback/change-of-base functor #2913 induced by the diagonal map from `R` to `Vector R n`, ie `replicate`. So they are derivable, and hence not needed to be included, except here to demonstrate the range of expressivity available beyond what is already in this PR.

jamesmckinna:
Getting on a high(er) horse for a moment, I'm going to double down on the suggestion to refactor this PR so as to achieve:
* a separation of the `Algebra.Base` and `Algebra.Properties` parts from those for `Module`s, 
* which I think should live separately under `Algebra.Module.Base` and `Algebra.Module.Properties`, for hygiene's sake, and consistency with the main `Algebra` hierarchy
* that we consider the 'vectorised' module constructions as above under `Algebra.Module.Pointwise.Base` etc., and the 'diagonal' ones under yet another layer `Algebra.Module.Diagonal.Base` etc. (suggestions welcome!)

Against @JacquesCarette 's earlier arguments, I might then also want to revisit whether these should live under `Algebra.Construct.Vector` or `Algebra.Construct.Product` or even `Algebra.Construct.Pointwise` (sic), because they amount to special cases (for index type `I = Fin n`) of the `I`-ary product construction #1907  on algebras? Ditto. `Module`s etc.

Shoutouts to @Taneb @MatthewDaggitt @gallais for your opinions on this!?

Under "the perfect should not be the enemy of the good", I'd be happy for the `I = Fin n` versions to live where they currently are, but with a downstream refactoring to exploit such instantiation of a more general `Algebra.Construct.Product`...

UPDATE: (with apologies for my failing memory) if I'm not mistaken, `Algebra.Construct.Pointwise`, already introduced in #2381 and updated in #2555 , gives (most of) the necessary machinery for `Algebra`s at least. So we could radically trim this PR down to focus solely on the `Ymodule` cases?

src/Data/Vec/Functional/Algebra/Properties.agda
jamesmckinna: Agree: I think we should (now) perhaps view the contents of this PR as a special case of `I`-ary products #1907 and hence be more systematic about regarding them as `I`-ary functions...

jamesmckinna:
@JacquesCarette @Taneb @e-mniang How should we proceed with this PR?

Finally dredging up `Algebra.Construct.Pointwise` from memory suggests (to me, at least) that we can significantly refactor and simplify this PR: `Data.Vec.Functional.Algebra` can be given as the special case `{C = Fin n}` ? There would definitely be some plumbing to do to organise such instantiation, and to extend the `Pointwise` constructions to `Algebra.Module.Construct.Pointwise`...

jamesmckinna:
@JacquesCarette any progress on this towards landing it for v2.4?

JacquesCarette:
None. And my spare time between now and mid-April is tiny.

jamesmckinna:
:disappointed: but them's the breaks. Will see what, if anything, I can try and do.

But perhaps we simply punt this to v3.0 for the time being?

MatthewDaggitt (approved):
This looks good to me!

jamesmckinna:
@JacquesCarette can we allocate some time together to sort this out, or else agree to close, unless @e-mniang wants to take this forward?

JacquesCarette:
Let's first see if we can sort this out @jamesmckinna . I can definitely allocate time on next Wednesday (accounting for time zones); might have some windows before that, but can't quite allocate the time at the moment.

jamesmckinna:
> Let's first see if we can sort this out @jamesmckinna . I can definitely allocate time on next Wednesday (accounting for time zones); might have some windows before that, but can't quite allocate the time at the moment.

Next Wednesday, but later than our usual 2pm BST, so maybe also better for you, might very well work for me.

JacquesCarette:
I'm available until 10PM BST, pick your time!

jamesmckinna:
> I'm available until 10PM BST, pick your time!

Let's do 18:00 BST.

jamesmckinna:
Further to our discussion yesterday, I'd lifted out #2915 as a separate issue tackling what I think is the main work remaining here. If your eventual (reworking of this) PR lands, we should think about closing that issue as part of the tidy up...

JacquesCarette:
Closing in favour of #3077 .

PR #2821, opened 2025-08-29
[draft] Polynomial

samparkky:
This draft pull request introduces a new way to represent a polynomial over commutative ring  -- as a vector over the commutative ring -- and a proof that polynomials themselves make up a commutative ring.

### DIFF
--- a/src/Algebra/Polynomial/Base1.agda
+++ b/src/Algebra/Polynomial/Base1.agda
@@ -1,17 +1,17 @@
-open import Algebra.Bundles using (Semiring)
+open import Algebra.Bundles using (Ring)
 open import Data.Nat.Base as ℕ using (ℕ; _⊔_; suc; pred)
 open import Data.Product.Base using (_,_)
 open import Data.Vec.Base as Vec using ([]; _∷_)
 import Level as L
 
-module Algebra.Polynomial.Base
-  {ℓ₁ ℓ₂} (SR : Semiring ℓ₁ ℓ₂)
+module Algebra.Polynomial.Base1
+  {ℓ₁ ℓ₂} (R : Ring ℓ₁ ℓ₂)
   where
 
-open import Algebra.Polynomial.Poly.Base SR as Poly
+open import Algebra.Polynomial.Poly.Base1 R as Poly
   using (Poly; zeros)
 
-open Semiring SR
+open Ring R
   using (0#; 1#)
   renaming (Carrier to A)
 
@@ -33,7 +33,7 @@ record Polynomial : Set ℓ₁ where
 
 private
   variable
-    P Q R : Polynomial
+    P Q W : Polynomial
 
 -- Equivalence relation for representations of polynomials
 infix 4 _≈_
@@ -46,7 +46,7 @@ refl = Poly.refl
 sym : P ≈ Q → Q ≈ P
 sym = Poly.sym
 
-trans : P ≈ Q → Q ≈ R → P ≈ R
+trans : P ≈ Q → Q ≈ W → P ≈ W
 trans = Poly.trans
 
 --------------------------------------------------
@@ -64,6 +64,10 @@ one = (1 , (1# ∷ []))
 -- Multiply the polynomial by a factor of x
 shift : Polynomial → Polynomial
 shift (m , p) = (suc m , Poly.shift p)
+
+-- Negation
+-_ : Polynomial → Polynomial
+- (m , p) = (m , Poly.- p)
 
 -- Scaling
 _·_ : A → Polynomial → Polynomial
--- a/src/Algebra/Polynomial/Base2.agda
+++ b/src/Algebra/Polynomial/Base2.agda
@@ -1,17 +1,17 @@
-open import Algebra.Bundles using (Semiring)
+open import Algebra.Bundles using (CommutativeRing)
 open import Data.Nat.Base as ℕ using (ℕ; _⊔_; suc; pred)
 open import Data.Product.Base using (_,_)
 open import Data.Vec.Base as Vec using ([]; _∷_)
 import Level as L
 
-module Algebra.Polynomial.Base
-  {ℓ₁ ℓ₂} (SR : Semiring ℓ₁ ℓ₂)
+module Algebra.Polynomial.Base2
+  {ℓ₁ ℓ₂} (CR : CommutativeRing ℓ₁ ℓ₂)
   where
 
-open import Algebra.Polynomial.Poly.Base SR as Poly
+open import Algebra.Polynomial.Poly.Base2 CR as Poly
   using (Poly; zeros)
 
-open Semiring SR
+open CommutativeRing CR
   using (0#; 1#)
   renaming (Carrier to A)
 
@@ -64,6 +64,10 @@ one = (1 , (1# ∷ []))
 -- Multiply the polynomial by a factor of x
 shift : Polynomial → Polynomial
 shift (m , p) = (suc m , Poly.shift p)
+
+-- Negation
+-_ : Polynomial → Polynomial
+- (m , p) = (m , Poly.- p)
 
 -- Scaling
 _·_ : A → Polynomial → Polynomial
--- a/src/Algebra/Polynomial/Morphism/Morphism.agda
+++ b/src/Algebra/Polynomial/Morphism/Morphism.agda
@@ -1,0 +1,510 @@
+-- Polynomial Morphisms
+open import Algebra.Bundles
+  using (CommutativeRing ; CommutativeSemiring; Ring;
+         SemiringWithoutAnnihilatingZero; Semiring)
+
+module ALgebra.Polynomial.Morphism.Morphism
+  {ℓ₁ ℓ₂}
+  (R : CommutativeRing ℓ₁ ℓ₂)
+  where
+
+open import Algebra.Bundles.Raw using (RawRing)
+open import Algebra.Morphism.Bundles
+  using (RingHomomorphism ; MagmaHomomorphism ; MonoidHomomorphism)
+open import Algebra.Morphism.Structures
+  using (IsRingHomomorphism ; IsMagmaHomomorphism ; IsSemiringHomomorphism;
+         IsNearSemiringHomomorphism ; IsMonoidHomomorphism)
+open import Algebra.Polynomial.Poly.Base2 R
+  using (Poly; []≈ ; ≈[]; cons ; cast-lem ; consˡ ; consʳ)
+  renaming ( sym to P-sym ; _·_ to _·P_
+           ; _+_ to _+P_ ; _*_ to _*P_; _≈_ to _≈P_; -_ to -P_)
+open import Algebra.Polynomial.Poly.Properties2 R
+  using (cast-p≈p; IsZero⇒≈0; ≈0⇒IsZero; a·p≈a∷[]*p ; *-distrib-shift)
+  renaming  (+-identityˡ to +-P-identityˡ ; +-identityʳ to +-P-identityʳ
+            ; zeroˡ to P-zeroˡ; zeroʳ to P-zeroʳ ; *-comm to *-P-comm )
+open import Algebra.Polynomial.Base2 R
+open import Algebra.Polynomial.Properties2 R as P
+  using (+-identityˡ; +-identityʳ; +-*-commutativeRing; *-zeroˡ ; *-zeroʳ)
+
+import Algebra.Definitions.RawSemiring as Def
+import Algebra.Definitions as D 
+open import Algebra.Properties.Group
+open import Algebra.Properties.Ring
+open import Data.Nat.Base as ℕ using (ℕ; suc; pred; _⊔_)
+open import Data.Vec.Base as Vec using (Vec; []; _∷_)
+open import Data.Vec.Relation.Unary.All
+open import Data.Product.Base using (proj₁ ; proj₂)
+import Level as L
+import Relation.Binary.Morphism.Structures as S
+  using (IsRelHomomorphism)
+open import Relation.Binary.PropositionalEquality.Core using (cong)
+
+module R = CommutativeRing R
+open R
+  hiding (distrib)
+  renaming
+  ( Carrier to A
+  ; _≈_ to _≈A_
+  ; _+_ to _+A_
+  ; _*_ to _*A_
+  ; -_ to -A_
+  ; 0# to A0#
+  ; 1# to A1#)
+  
+private
+  variable
+    m n k l : ℕ
+    a b c d : A
+    p : Poly m
+    q : Poly n
+    P : Polynomial
+    Q : Polynomial
+
+module PolynomialRingMorphism
+  {ℓ₃ ℓ₄}
+  (B : CommutativeRing ℓ₃ ℓ₄)
+  (β : CommutativeRing.Carrier B) 
+
+  (f : RingHomomorphism {ℓ₁} {ℓ₂}
+       (CommutativeRing.rawRing R) (CommutativeRing.rawRing B) )
+  where
+
+  module B = CommutativeRing B 
+  open B
+    using (semiringWithoutOne ; isSemiringWithoutOne ; distrib)
+    renaming
+    (_≈_ to _≈B_
+    ; _+_ to _+B_
+    ; _*_ to _*B_
+    ; -_ to -B_
+    ; setoid to B-Set)
+    using (0#; 1#)
+
+  open import Relation.Binary.Reasoning.Setoid B-Set
+    
+  open RingHomomorphism f using (⟦_⟧)
+  open D {A = B.Carrier} (_≈B_)
+    using (_MiddleFourExchange_; _DistributesOverˡ_)
+
+  eval-p : Poly m → CommutativeRing.Carrier B
+  eval-p [] = 0#
+  eval-p (a ∷ p) =  ⟦ a ⟧ +B (β *B eval-p p )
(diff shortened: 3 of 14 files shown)


jamesmckinna:
Zeroth thoughts:
* thanks for the contribution!
* It's complicated, though, given that we already have at least one such representation (in the `RingSolver` architecture, based on an optimised Horner normal form, IIRC), and committing the namespace to have `Algebra.Polynomial` be the home for this presentation of the dense representation, while natural, seems ... premature without further discussion. I might tentatively suggest `Algebra.Polynomial.Dense.*` as a better root path for these additions?
* if it is indeed 'draft', then it would be helpful to avoid any premature detailed review by marking it as such.

First thoughts:
* there is no need to base this on `CommutativeRing`: the only place that I can see where `*-comm` is used is in the `shift` lemmas for multiplication, where one of the arguments is `0#`, and `0#` commutes with *everything* multiplicatively (as does `1#`...); the use of `Consequences` to streamline proofs in the presence of commutativity is a short-cut, but not actually *necessary*, IIUC
* so: suggest at least that the PR be refactored to base on `Ring`, with additional properties when multiplication is `Commutative`

Second thought:
* no need for subtraction/unary negation either! So please consider doing this for `Semiring` instead? Indeed, the `stdlib` style might be to try to identify the minimal axiomatisation for which `Polynomial` makes sense, and build up from there. This PR makes a lot of commitments (esp. to naming) which might prove costly to refactor if we wanted to generalise later?

Third thought:
* (but this could be added downstream) any version of `Polynomial` ought *at least* to have (some version of) its characteristic universal property, or even simply `eval` given an element of the underlying `Carrier`?

Fourth thoughts:
* ergonomics: which version of `Vec` works best? the 'structural' account (`Vec` as you have it), or the 'functional' account (as in `Data.Vec.Functional`); or even a *bounded* version for normalising away leading zeroes? It's possible that each of these would require `cast` in order to state, and prove, basic properties, but should be investigated further...
* some of your `cast` lemmas, as well as some of the actual `suc`/`pred` arithmetic, may already be present in the library, and it might be sensible to refactor to lift any missing others out as `Data.Vec.Properties` and `Data.Nat.Properties`?

src/Algebra/Properties/Group.agda
```diff
@@ -127,6 +127,19 @@ x≈y⇒x∙y⁻¹≈ε {x} {y} x≈y = begin
   y ∙ y ⁻¹ ≈⟨ inverseʳ y ⟩
   ε        ∎
 
+x≈ε⇒x⁻¹≈ε : ∀ {x} → x ≈ ε → x ⁻¹ ≈ ε
+x≈ε⇒x⁻¹≈ε {x} x≈ε = begin
+  x ⁻¹   ≈⟨ ⁻¹-cong x≈ε ⟩
+  ε ⁻¹   ≈⟨ ε⁻¹≈ε ⟩
+  ε ∎
```
jamesmckinna, line 134: I wasn't sure if
* this lemma met the Fairbairn threshhold (moreover, it's used only once in this PR)
* this proof was the most 'efficient', compared to
  ```agda
  x≈ε⇒x⁻¹≈ε = identityˡ-unique _ _ ∘ trans (inverseˡ _) ∘ sym
  ```
which uses only a `Loop` property beyond the basic `IsGroup` axiomatisation.

src/Algebra/Properties/Group.agda
```diff
@@ -127,6 +127,19 @@ x≈y⇒x∙y⁻¹≈ε {x} {y} x≈y = begin
...
+
+x⁻¹≈ε⇒x≈ε : ∀ {x} → x ⁻¹ ≈ ε → x ≈ ε
+x⁻¹≈ε⇒x≈ε {x} x⁻¹≈ε = begin
+  x       ≈⟨ ⁻¹-involutive x ⟨
+  x ⁻¹ ⁻¹ ≈⟨ ⁻¹-cong x⁻¹≈ε ⟩
+  ε ⁻¹    ≈⟨  ε⁻¹≈ε ⟩
+  ε ∎
```
jamesmckinna, line 141: Ditto, only more so:
```agda
x⁻¹≈ε⇒x≈ε = identityʳ-unique _ _ ∘ trans (inverseˡ _) ∘ sym
```

jamesmckinna: Alternatively, you could bootstrap
```suggestion
x⁻¹≈ε⇒x≈ε {x} x⁻¹≈ε = begin
  x       ≈⟨ ⁻¹-involutive x ⟨
  x ⁻¹ ⁻¹ ≈⟨ x≈ε⇒x⁻¹≈ε x⁻¹≈ε ⟩
  ε ∎
```
which is 'just'
```agda
x⁻¹≈ε⇒x≈ε = trans (sym (¹-involutive _)) ∘ x≈ε⇒x⁻¹≈ε
```

Taneb:
I've been working on a slightly different implementation of this. Let me find what I've done and compare - notably, I used lists rather than vectors, and didn't have a notion of degree

Taneb:
@jamesmckinna 
> there is no need to base this on CommutativeRing: the only place that I can see where *-comm is used is in the shift lemmas for multiplication, where one of the arguments is 0#, and 0# commutes with everything multiplicatively (as does 1#...); the use of Consequences to streamline proofs in the presence of commutativity is a short-cut, but not actually necessary, IIUC

I think you need commutativity of multiplication to show that polynomial evaluation preserves multiplication

src/Algebra/Polynomial/Poly/Base.agda
```diff
@@ -0,0 +1,149 @@
...
+
+------------------------------------------------------------------
+-- The representation of a polynomial is not unique
+-- because we allow coefficients to be zero
+-- We define an equivalence relation which takes this into account 
+
+data _≈_ : (Poly m) → (Poly n) → Set (ℓ₁ L.⊔ ℓ₂) where
+  []≈  : (q : Poly n) → IsZero q → [] ≈ q
+  ≈[]  : (p : Poly m) → IsZero p → p ≈ []
+  cons : (a ≈A b) → (p ≈ q) → (a ∷ p) ≈ (b ∷ q)
```
Taneb, line 70: This particular definition has two separate proofs that `[]` is equal to `[]`... not necessarily the end of the world, but it bothers me

jamesmckinna:
@Taneb 

> > there is no need to base this on CommutativeRing: the only place that I can see where *-comm is used is in the shift lemmas for multiplication, where one of the arguments is 0#, and 0# commutes with everything multiplicatively (as does 1#...); the use of Consequences to streamline proofs in the presence of commutativity is a short-cut, but not actually necessary, IIUC
> 
> I think you need commutativity of multiplication to show that polynomial evaluation preserves multiplication

Hmmm. https://math.stackexchange.com/questions/806119/do-polynomials-make-sense-over-non-commutative-rings 

Otherwise could we even speak about polynomials over rings such as square matrices (for the Cayley-Hamilton theorem, for example, even if that's usually only considered for matrices over commutative rings, such matrices don't enjoy commutative multiplication... while for any polynomial and its values, it *is* the case that *powers* of an element *do* commute...)?

I think the real issue is making sure that multiplication is defined correctly... see also https://en.wikipedia.org/wiki/Monoid_ring

JacquesCarette:
Non-commutative polynomial rings certainly do make sense and are useful. See for examples the [skew polynomials](https://en.wikipedia.org/wiki/Ore_extension) that arise when modeling various kinds of operators.

Axiom does allow polynomials over coefficients rings like square matrices.

Taneb:
For reference, here's my implementation of polynomials: https://gist.github.com/Taneb/3257f47436ec6b11de403d839d7c413b

samparkky:
Thank you for the detailed feedback and the comments and I apologize for the delayed response. Although it will take some time, I will try implement the changes using the semiring instead of the commutative ring and eventually prove the universal property of polynomial rings. Since this is only a draft which I should have made it clear, any further comments or suggestions are welcome.

src/Algebra/Polynomial/Base1.agda
```diff
@@ -0,0 +1,82 @@
+open import Algebra.Bundles using (Ring)
+open import Data.Nat.Base as ℕ using (ℕ; _⊔_; suc; pred)
+open import Data.Product.Base using (_,_)
+open import Data.Vec.Base as Vec using ([]; _∷_)
+import Level as L
+
+module Algebra.Polynomial.Base1
```
JacquesCarette, line 7: `Base1` is quite a rotten name. I'm sure we can help find a better name, once we know the intent?

src/Algebra/Polynomial/Base1.agda
```diff
@@ -0,0 +1,82 @@
...
+    m n k : ℕ
+    p : Poly m
+    q : Poly n
+    r : Poly k
+
+---------------------------------------------------
+-- Types
+
+record Polynomial : Set ℓ₁ where
+  constructor _,_
```
JacquesCarette, line 29: I strongly object to overloading `_,_` for this constructor!

src/Algebra/Polynomial/Base1.agda
```diff
@@ -0,0 +1,82 @@
...
+  field
+    degree : ℕ
```
JacquesCarette, line 31: There should be a comment here that the "degree" is off-by-one from the usual notion of the degree of a polynomial. It is more like a length than a degree.

src/Algebra/Polynomial/Base1.agda
```diff
@@ -0,0 +1,82 @@
...
+zero : Polynomial
+zero = (0 , [])
+
+one : Polynomial
+one = (1 , (1# ∷ []))
+
+---------------------------------------------------
+-- Operations
+
+-- Multiply the polynomial by a factor of x
```
JacquesCarette, line 64: I'd say "by x" rather than "by a factor of x", as I wondering "what factor?" 

And why not implement `shift :: ℕ → Polynomial → Polynomial instread?

src/Algebra/Polynomial/Base2.agda
```diff
@@ -0,0 +1,82 @@
+open import Algebra.Bundles using (CommutativeRing)
+open import Data.Nat.Base as ℕ using (ℕ; _⊔_; suc; pred)
+open import Data.Product.Base using (_,_)
+open import Data.Vec.Base as Vec using ([]; _∷_)
+import Level as L
+
+module Algebra.Polynomial.Base2
```
JacquesCarette, line 7: Same comment about the name

src/Algebra/Polynomial/Base2.agda
```diff
@@ -0,0 +1,82 @@
...
+    m n k : ℕ
+    p : Poly m
+    q : Poly n
+    r : Poly k
+
+---------------------------------------------------
+-- Types
+
+record Polynomial : Set ℓ₁ where
+  constructor _,_
```
JacquesCarette, line 29: and the name of this constructor (won't repeat other comments that also apply)

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+open import Data.Nat.Base as ℕ using (ℕ; suc; pred; _⊔_)
+open import Data.Vec.Base as Vec using (Vec; []; _∷_)
+open import Data.Vec.Relation.Unary.All
+open import Data.Product.Base using (proj₁ ; proj₂)
+import Level as L
+import Relation.Binary.Morphism.Structures as S
+  using (IsRelHomomorphism)
+open import Relation.Binary.PropositionalEquality.Core using (cong)
+
+module R = CommutativeRing R
```
JacquesCarette, line 42: I think that `module A = CommutativeRing R` and then using `A.foo` below would be much nicer than renaming.

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+    a b c d : A
+    p : Poly m
+    q : Poly n
+    P : Polynomial
+    Q : Polynomial
+
+module PolynomialRingMorphism
+  {ℓ₃ ℓ₄}
+  (B : CommutativeRing ℓ₃ ℓ₄)
+  (β : CommutativeRing.Carrier B)
```
JacquesCarette, line 66: Why are Homomorphisms always scaled? This seems like an odd choice.

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+  eval (ℕ.zero , []) = 0#
+  eval (suc n , (a ∷ p)) = ⟦ a ⟧ +B (β *B eval (n , p) )
+
+ 
+
+  eval-equiv : (p : Poly m) → eval-p p ≈B (eval (m , p))
+  eval-equiv {ℕ.zero} [] = B.refl
+  eval-equiv {suc m} (a ∷ p) = begin
+    ⟦ a ⟧ +B β *B eval-p p
+      ≈⟨ B.+-congˡ (B.*-congˡ (eval-equiv p) ) ⟩
```
JacquesCarette, line 104: There's no point in using equational reasoning when the proof is exactly 1 inference.

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+  open RingHomomorphism f using (⟦_⟧)
+  open D {A = B.Carrier} (_≈B_)
+    using (_MiddleFourExchange_; _DistributesOverˡ_)
+
+  eval-p : Poly m → CommutativeRing.Carrier B
+  eval-p [] = 0#
+  eval-p (a ∷ p) =  ⟦ a ⟧ +B (β *B eval-p p )
+
+
+  eval : Polynomial → CommutativeRing.Carrier B
```
JacquesCarette, line 94: Bad name: it took me a while to figure out that this "evaluate at 1 scaled by β" or equivalently "evaluate at β". The evaluation point should be a (local) parameter, not a module parameter.

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+    eval ((m , p) + (n , q))
+      ≈⟨ eval-equiv (p +P q) ⟨
+    eval-p (p +P q)
+      ≈⟨ eval-p-+-homomorphic p q ⟩
+    eval-p p +B eval-p q
+      ≈⟨ B.+-cong (eval-equiv p) (eval-equiv q) ⟩
+    eval (m , p) +B eval (n , q)
+      ∎
+
+  eval-p-1#-homomorphic : eval-p (A1# ∷ []) ≈B 1#
```
JacquesCarette, line 205: Inlining the definition of a constant polynomial makes this definition harder to read. Plus this is true for any member of the ground ring.

src/Algebra/Polynomial/Morphism/Morphism.agda
```diff
@@ -0,0 +1,510 @@
...
+      ≈⟨ B.+-congʳ (RingHomomorphism.-‿homo f a) ⟩
+    -B ⟦ a ⟧ +B -B (β *B eval-p p)
+      ≈⟨ B.+-comm (-B (β *B eval-p p)) (-B ⟦ a ⟧) ⟨
+    -B (β *B eval-p p) +B -B ⟦ a ⟧
+      ≈⟨ ⁻¹-anti-homo-∙ B.+-group ⟦ a ⟧ (β *B eval-p p) ⟨
+    -B eval-p (a ∷ p)
+      ∎
+        where
+          step1 = begin
+            β *B -B eval-p p
```
JacquesCarette, line 332: This seems like a special case of commutativity of scalars.

JacquesCarette (changes requested):
Partial review of these additions.

src/Algebra/Polynomial/Base2.agda
samparkky: I sincerely appreciate all your comments and apologize for the delayed response. I’ve been postponing the implementation of the suggested edits but hope to resume work soon. I made several poor choices regarding naming and implementation and was careless to post this version prematurely. I’ll do my best to reflect all your feedback as soon as possible.

gallais:
Given that we have very few (scaling & shifting) degree-respecting operations,
that our notions of addition or equality are heterogeneous, do we really need
to have an index-enforcing representation? Especially if it forces us to have a
bunch of `cast` all over the place?

How about defining `degree` as a function (maybe even one that disregards
irrelevant `0#`s for semiring where that's decidable) and proving its properties
separately?

JacquesCarette:
@samparkky any chance you will finish this?

samparkky:
Yes, sorry for the delay. I only recently took over the project after graduating, but I plan to implement all the requested changes by the end of the summer.

src/Algebra/Polynomial/Poly/Base1.agda
```diff
@@ -0,0 +1,148 @@
...
+
+module R = Ring R
+open R
+ renaming
+  ( Carrier to A
+  ; _≈_ to _≈A_
+  ; _+_ to _+A_
+  ; _*_ to _*A_
+  ; -_ to -A_)
+ using (0#; 1#)
```
jamesmckinna, line 24: if, instead, you write:

```suggestion
private
  open module R = Ring R
      using (0#; 1#)
      renaming (Carrier to A)
```
then the remaining parts of the signature are available as projections `R._+_` etc. from module `R`, and it's perhaps worth saving the legibility/maintenance cost of 'custom' names `_+A_`.

PR #2823, opened 2025-08-31
[ add ] `PartialOrder` structure/bundle on `_→_` (fixes issue #2820)

### LINKED ISSUE #2820: [ add ] `Setoid`/`IsEquivalence` bundle/structure of propositions modulo bi-implication
This issue is derived from the OP of #2070 .

Given the 'bare' preorder of sets (propositions) under implication `_→_`, we can now add the structure/bundle using the constructions in `Relation.Binary.Construct.Interior.Symmetric` added in #2071 , so we should... somewhere.

jamesmckinna:
Specimen solution to #2820 , but I'm not entirely sure about
* what the eventual `public` exports should be
* what the relationship with `Equivalence`/`_↔_` should be, and where it should be established
* names/locations of the contribution

TODO:
* `CHANGELOG` additional exports, depending on what we eventually decide about such


JacquesCarette (commented):
This part of the PR makes sense to me.

PR #2824, opened 2025-09-03
Add `cartesianProductWith⁻` and `cartesianProduct⁻` to `All`

javierdiaz72:
This PR adds `cartesianProductWith⁻` and `cartesianProduct⁻` to `Data.List.Relation.Unary.All.Properties`.


CHANGELOG.md
```diff
@@ -103,3 +103,16 @@ Additions to existing modules
   ```agda
   ¬¬-η : A → ¬ ¬ A
   ```
+
+* In `Data.List.Relation.Unary.All.Properties`:
+  ```agda
+  cartesianProductWith⁻ : ∀ f →
+                          f Preserves₂ _≈₁_ ⟶ _≈₂_ ⟶ _≡_ →
```
JacquesCarette, line 110: There is no point right-indenting so much, go ahead and only indent the type by 2-4 spaces, so as to use fewer lines

src/Data/List/Relation/Unary/All/Properties.agda
```diff
@@ -419,14 +421,31 @@ module _ (S₁ : Setoid a ℓ₁) (S₂ : Setoid b ℓ₂) where
...
                       All P (cartesianProduct xs ys)
   cartesianProduct⁺ = cartesianProductWith⁺ _,_
 
+  cartesianProductWith⁻ : ∀ f →
+                          f Preserves₂ _≈₁_ ⟶ _≈₂_ ⟶ _≡_ →
+                          ∀ xs ys →
+                          All P (cartesianProductWith f xs ys) →
+                          (∀ {x y} → x ∈₁ xs → y ∈₂ ys → P (f x y))
+  cartesianProductWith⁻ {P = P} f pres (x′ ∷ xs) ys Ps {y = y} x∈x′∷xs y∈ys
+    with x∈x′∷xs   | ++⁻ (map (f x′) ys) Ps
```
JacquesCarette, line 438: I think it would be nicer the use plain pattern-matching for the ` x∈x′∷xs` argument and then dispatch on `with`.

Also: is there a way to use equational reasoning instead of `rewrite`? This proof is far from limpid.

JacquesCarette (changes requested):
A few nitpicks.

jamesmckinna:
The `Preserves` assumption on `f` seems suspiciously strong, with its use of propositional equality. In other words the lemmas added seem to be insufficiently generalised. Can you improve matters by making a suitable additional `Respects` assumption about `P`? 

UPDATED: I think that the 'real' `cartesianProductWith⁻` should take into account a `Setoid`/`IsEquivalence` structure on the codomain of `f`... and that `f` should be suitably `Congruent₂` wrt its input arguments, and that result structure... in such a way that when specialised to the `Pointwise` structure on the product, in the special case of `cartesianProduct`, such congruence property becomes automatic. Additionally, `P` should suitably respect such structure. 

I think that this is the 'usual' problem with `_∧_`/ `_×_` as a connective: we *can* present it (Prawitz-style) just in terms of `_,_`/`proj₁`/`proj₂`, but it is 'better' to consider it Gentzen-/Schroeder-Heister- style in terms of `_,_` and a left rule which is more like pattern-matching `let`, with result type/conclusion in 'general position' as arbitrary proposition `C`. The extension to considering `Setoid` lifting of these concepts seems to demand that we take the more general option?

CHANGELOG.md
jamesmckinna: Also: no need for the explicit quantifiers for `f` and `xs`/`ys` in this type?
(But there seems to be no explicitly documented consensus on how much detail of such kind we expect in a `CHANGELOG` entry...?)

Indeed, are you sure that `xs`/`ys` even need to be explicitly quantified at all here?

src/Data/List/Relation/Unary/All/Properties.agda
jamesmckinna: And rather than use `with`, as the result of that computation has a single-constructor type, it can be irrefutably matched with a pattern-matching `let` instead, or even the new `using` syntax?

jamesmckinna: But, if the issue is one of *laziness*, namely that different projections are used in each branch, then better to use `proj₁`/`proj₂` instead?

jamesmckinna: Also: in each case branch, pointless eta-expansion wrt the `y∈ys` argument...

jamesmckinna:
@javierdiaz72 besides fixing the merge conflicts that have arisen since the v2.4 release, and current developments on `master` towards v3.0, how would you like to proceed with this? There hasn't been any response from you to the review feedback from myself and @JacquesCarette since already 9 months, so we're thinking about closing this PR unmerged. 

Please let us know what you would like to do with this contribution!

jamesmckinna:
I think we'll close this tomorrow if we still haven't had a response.

jamesmckinna:
Closing unmerged.

Issue #2303, opened 2024-02-27
Find better new symbol for `;` 


jamesmckinna:
Responding to my query on #2301 , @Taneb wrote:

            `;` is (according to agda-mode's "Information about the character at point" tool) a Greek question mark, which is written with `\;`, but there's a dozen characters with that and it's the 11th in the list.

I don't think we should be using that character, it's identical to the semicolon graphically (it's even normalized to the semicolon by Unicode) and bound to cause confusion. But I don't have a suggestion for what to use instead. EDIT: `⨾` (first in the list of `\;` but also \z;`) is "Z NOTATION RELATION COMPOSITION" per the Unicode standard. So it's at least semantically appropriate

_Originally posted by @Taneb in https://github.com/agda/agda-stdlib/issues/2301#issuecomment-1965933649_

silentstormm:
`⨾ U+2A3E : Z NOTATION RELATIONAL COMPOSITION` shows up as the first suggestion for `\;`, doesn't seem to be used anywhere else _and_ is intended for this exact purpose, is there any reason it's not used here?

Taneb:
Yes, that's the symbol I suggested in the linked PR

JacquesCarette:
Personally I much prefer U+02A1F , i.e.
```
02A1F ⨟ # \zcmp mathop oz = \semi (oz), = \fatsemi (stmaryrd), Z NOTATION SCHEMA COMPOSITION
```
which I believe is also available in newer agdas.

MatthewDaggitt:
@JacquesCarette  what do you see as the difference between your symbol and the one suggested by @silentstormm above?

JacquesCarette:
U+02A1F is larger than  U+2A3E which I've found to act as a nice visual delimiter. It also has the advantage of having been in a standard latex package for longer.

jamesmckinna:
What are the 'natural' symbols for the right adjoints to these candidate 'semicolon' symbols for composition? (Maybe de Moor and Bird came up with 'reasonable' names for them in *Algebra of Programming*?)

(otherwise, giving in to the temptation to repeat myself: why *not* `\otimes`, which has the advantage of having been in standard LaTeX packages for *a lot* longer... I'm not sure that Z notation is necessarily a good guide to choices we might commit to in `stdlib`?)

PR #2825, opened 2025-09-04
Add formalization of substructural logics in src

mlebar-CU:
I've added a formalization of several substructural logics in the src folder, based on chapter 2 of Greg Restall's "An Introduction to Substructural Logics". This is work I did as part of a master's research project. My hope is that the work I've done here can be a useful resource for anyone wanting to study logics in Agda. Please let me know if there any changes I can make to help this fit the Agda standard library.


CHANGELOG/mlebar-CU.md
```diff
@@ -0,0 +1 @@
+* Added a folder "Logic" in src. This provides a formalization of several substructural logics based on chapter 2 of Greg Restall's "An Introduction to Substructural Logics". My hope is that this can provide a useful framework for anyone wanting to study a variety of logics in Agda. This was a master's research project done under Professor Stuart Kurtz, and I'm really happy with my work here - I would love to contribute to the standard library, and I am happy to make any appropriate modification to help with that.
```
JacquesCarette: We try not to have long lines - can you please cut these down to ~80 characters wide?

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Connectives.And-Logic where
+
+open import Logic.Logic
+
+record And-Logic
+  {Lang : Set} {Struct : Set} (S : Lang → Struct)
```
JacquesCarette, line 14: It would probably make sense to bundle all these things up in a record?

Certainly it would make sense to explain (in the CHANGELOG? A README?) what the overall design is.

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct )
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _∧_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
```
JacquesCarette, line 20: Shouldn't this field be a parameter? You're building And-logic on top of implicational logic, right?

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+    and-introduction :
+      ∀ {X : Struct} {A B : Lang} →
+      X ⊢ A →
+      X ⊢ B →
+      -------------------
+      X ⊢ (A ∧ B)
+    and-elimination-1 :
```
JacquesCarette, line 27: I think `left` and `right` would make more sense than -1 and -2. I know you're following Restall, but he chose bad names, we don't have to follow suit!

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+      -------------------
+      X ⊢ A
+    and-elimination-2 :
+      ∀ {X : Struct} {A B : Lang} →
+      X ⊢ (A ∧ B) →
+      -------------------
+      X ⊢ B
+
+if-and-distrib :
+  ∀ (Lang : Set) (Struct : Set)
```
JacquesCarette, line 39: This repetition of arguments screams for having `module _ ...` encapsulating that.

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+
+record And-Logic
+  {Lang : Set} {Struct : Set} (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct )
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _∧_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+    and-introduction :
```
JacquesCarette, line 21: As I see these used 'in practice' below, I think `∧-I` (and `∧-E`) would be much nicer. You could even consider making `∧-I` infix.

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
...
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _∧_ : Lang → Lang → Lang) →
+  And-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _∧_ → ∀ {A B C : Lang}  →
+  -----------------------------------------------------------------
+  (S ((A ⇒ B) ∧ (A ⇒ C))) ⊢ (A ⇒ (B ∧ C))
+
+-- proof from page 34
+if-and-distrib Lang Struct S _⊢_ C⟨_⟩ _⨾_ _⇒_ _∧_ x =
+  Logic.if-introduction y
```
JacquesCarette, line 51: I think you should 'open' more things, so as to make this proof human-readable.

src/Logic/Connectives/Backif-Logic.agda
```diff
@@ -0,0 +1,62 @@
+----------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Logic with backward conditional connective +
+-- related proof
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Connectives.Backif-Logic where
```
JacquesCarette, line 10: Of course, many of the comments on `And-logic` apply here as well (and below) and will not be repeated. Only new comments will be added.

src/Logic/Connectives/Backif-Logic.agda
```diff
@@ -0,0 +1,62 @@
...
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct )
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _⇐_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+
+    backif-introduction :
+      ∀ {X : Struct} {A B : Lang} →
+      ((S A) ⨾ X) ⊢ B →
```
JacquesCarette, line 26: You should give appropriate fixities to `_⨾_` and `_⊢_` so that no parens are needed here at all.

src/Logic/Connectives/Backif-Logic.agda
```diff
@@ -0,0 +1,62 @@
...
+  ∀ (Lang : Set) (Struct : Set) (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _⇐₁_ _⇐₂_ : Lang → Lang → Lang) →
+  Backif-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _⇐₁_ →
+  Backif-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _⇐₂_ →
+  ∀ {A B : Lang} →
+  ---------------------------------------------------------
+  (S (A ⇐₁ B)) ⊢ (A ⇐₂ B) × (S (A ⇐₂ B)) ⊢ (A ⇐₁ B)
```
JacquesCarette, line 49: It would make sense to define "logical equivalence" in `Logic.logic` and then use it here, instead of expanding it out.

src/Logic/Connectives/Backif-Logic.agda
```diff
@@ -0,0 +1,62 @@
...
+-- Uniqueness of back-conditionals
+backif-unique :
+  ∀ (Lang : Set) (Struct : Set) (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _⇐₁_ _⇐₂_ : Lang → Lang → Lang) →
+  Backif-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _⇐₁_ →
+  Backif-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _⇐₂_ →
+  ∀ {A B : Lang} →
```
JacquesCarette, line 47: Should probably make `A` and `B` variables.

src/Logic/Connectives/Fusion-Logic.agda
```diff
@@ -0,0 +1,62 @@
+----------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Logic with 'fusion' connective + related proof
```
JacquesCarette, line 4: Should have more comments about what 'fusion' is, since it is not so common.

src/Logic/Connectives/Fusion-Logic.agda
```diff
@@ -0,0 +1,62 @@
...
+
+open import Logic.Logic
+open import Data.Product using (_×_; proj₁; proj₂) renaming (_,_ to ⟨_,_⟩)
+
+record Fusion-Logic
+  {Lang : Set} {Struct : Set} (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _∘_ : Lang → Lang → Lang) : Set where
```
JacquesCarette, line 19: `_∘_` seems like a fairly unfortunate choice of symbols?

src/Logic/Connectives/Or-Logic.agda
```diff
@@ -0,0 +1,89 @@
...
+record Or-Logic
+  {Lang : Set} {Struct : Set} (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct )
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ _∨_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+
+    product-context :
```
JacquesCarette, line 24: Why does this logic have 3 additional context assumptions that others don't? Also, shouldn't these be parameters rather than fields?

src/Logic/Connectives/Trivial-Logic.agda
```diff
@@ -0,0 +1,60 @@
...
+  (⊤ ⊥ : Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+
+    ⊤-introduction :
+      ∀ {X : Struct} →
+      ---------
+      X ⊢ ⊤
+
+    ⊥-elimination :
```
JacquesCarette, line 33: As the comment at the top only mentions 'trivial truth', I was surprised to see false also show up.

jamesmckinna:
Zeroth thought: this looks really nice!

First thought: we might need to scratch out heads a while about opening up/committing the `Logic.*` namespace to this?

Second thought: could this be refactored to prune out a lot of the DRY repetition of eg the initial parameter telescopes of (most) of the definitions? 
* Either, by reifying them as a `record`, eg `RawLogic` on the model of `Algebra.Bundles.Raw`...?
* Or, by making them top-level parameters of each (sub-)module?

Third thought: everything is tied to '`Logic` over `Set`, ie at `Level zero`, rather than making it `Level`-polymorphic. Presumably (?) it would be straightforward to generalise everything, albeit a lot of tedious editing to do so...

Fourth thought: On the model of the various `README` modules which document/show exemplary usage, are there good such 'example' files which you could add to this development (esp. wrt the slightly 'occult' use of language such as `Struct` to embed propositions as ... wff-like things; readers unfamiliar with Restall's book, or general approach, might find this hard going at first?)

Fifth, nitpicky, thought: the fat semicolon is... annoying cf. #2303

JacquesCarette:
I've got a detailed review incoming. Agree with thoughts 0-4.

MatthewDaggitt:
At the risk of being a dissenting opinion, is this something that might be better as a standalone library? My initial reaction is that this feels very similar to https://github.com/agda/agda-stdlib/blob/master/src/Data/Fin/Substitution.agda and its child modules which I think we agreed at some point probably shouldn't live in the standard library?

src/Logic/Connectives/Truth-Logic.agda
```diff
@@ -0,0 +1,55 @@
...
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Connectives.Truth-Logic where
+
+open import Logic.Punctuation.Leftid-Logic
+open import Logic.Logic
+open import Data.Product using (_×_; proj₁; proj₂) renaming (_,_ to ⟨_,_⟩)
+
+
+record Truth-Logic {Lang : Set} {Struct : Set} (S : Lang → Struct) (_⊢_ : Struct → Lang → Set) (C⟨_⟩ : Struct → Struct ) (_⨾_ : Struct → Struct → Struct) (∅ : Struct) (_⇒_ : Lang → Lang → Lang) (t : Lang) : Set where
```
JacquesCarette, line 16: Inconsistent formatting from other files

src/Logic/Connectives/Truth-Logic.agda
```diff
@@ -0,0 +1,55 @@
...
+-- The Agda standard library
+--
+-- Logic with 'truth' connective + related proof
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Connectives.Truth-Logic where
+
+open import Logic.Punctuation.Leftid-Logic
```
JacquesCarette, line 11: 'Punctuation' ?

src/Logic/Connectives/Truth-Logic.agda
```diff
@@ -0,0 +1,55 @@
...
+open import Data.Product using (_×_; proj₁; proj₂) renaming (_,_ to ⟨_,_⟩)
+
+
+record Truth-Logic {Lang : Set} {Struct : Set} (S : Lang → Struct) (_⊢_ : Struct → Lang → Set) (C⟨_⟩ : Struct → Struct ) (_⨾_ : Struct → Struct → Struct) (∅ : Struct) (_⇒_ : Lang → Lang → Lang) (t : Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+
+    is-leftid-Logic : Leftid-Logic S _⊢_ C⟨_⟩ _⨾_ ∅ _⇒_
+
+    t-introduction : ∅ ⊢ t
```
JacquesCarette, line 22: Why `t` here where \top was used in Trivial-logic?

src/Logic/Logic.agda
```diff
@@ -0,0 +1,94 @@
...
+      (S A) ⊢ A
+
+    context :
+      ∀ {X X' : Struct} {A : Lang} →
+      (C⟨ X ⟩ ⊢ A → C⟨ X' ⟩ ⊢ A) →
+      X ⊢ A →
+      ---------
+      X' ⊢ A
+
+    semicolon--context-l :
```
JacquesCarette: really sub-optimal name for a rule, same below.

src/Logic/Logic.agda
```diff
@@ -0,0 +1,94 @@
...
+      ----------------------------------------
+      (C⟨ Y ⨾ X ⟩ ⊢ A → C⟨ Y ⨾ X' ⟩ ⊢ A)
+
+    semicolon--context-r :
+      ∀ {X X' Y : Struct} {A : Lang} →
+      (C⟨ X ⟩ ⊢ A → C⟨ X' ⟩ ⊢ A) →
+      ----------------------------------------
+      (C⟨ X ⨾ Y ⟩ ⊢ A → C⟨ X' ⨾ Y ⟩ ⊢ A)
+
+    if-introduction :
```
JacquesCarette: wouldn't implication-introduction or `⇒-introduction` (or something shorter) would be more intuitive?

src/Logic/Logic.agda
```diff
@@ -0,0 +1,94 @@
...
+
+    if-elimination :
+      ∀ {X Y : Struct} {A B : Lang} →
+      X ⊢ (A ⇒ B) →
+      Y ⊢ A →
+      -------------
+      (X ⨾ Y) ⊢ B
+
+-- proof from page 23
+chain-ifs :
```
JacquesCarette: As there is no `if` in the assumption, this doesn't quite seem like 'chaining'? I understand that after `if-introduction`, it does -- but still.

src/Logic/Logics/BCK-Logic.agda
```diff
@@ -0,0 +1,28 @@
+----------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definition for Positive Affine Logic (BCK in Restall)
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Logics.BCK-Logic where
```
JacquesCarette, line 9: `Logic.Logics` already clearly says that we're introducing logics... so maybe `BCK` is quite sufficient?

src/Logic/Logics/J-Logic.agda
```diff
@@ -0,0 +1,28 @@
...
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Logics.J-Logic where
+
+open import Logic.Logic
+open import Logic.Logics.BCK-Logic
+open import Logic.Structural-Rules.W-Logic
+
+record J-Logic
```
JacquesCarette, line 15: If you're going to call it `J`, you need to explain why!

src/Logic/Logics/L-Logic.agda
```diff
@@ -0,0 +1,32 @@
...
+
+module Logic.Logics.L-Logic where
+
+open import Logic.Logic
+open import Logic.Structural-Rules.B-Logic
+open import Logic.Structural-Rules.B^c-Logic
+open import Logic.Connectives.Backif-Logic
+open import Logic.Connectives.Fusion-Logic
+
+record L-Logic
```
JacquesCarette, line 17: `Lambek` instead of `L`, unless there is precedent?

src/Logic/Logics/Or-Comma-Logic.agda
```diff
@@ -0,0 +1,42 @@
...
+  (_⨾_ _/_ : Struct → Struct → Struct)
+  (_⇒_ _∨_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_  _⇒_
+
+    is-or-logic : Or-Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ _∨_
+
+    is-comma-logic : Comma-Logic S _⊢_ C⟨_⟩ _⨾_ _/_ _⇒_
+
+    comma-product-context-l :
```
JacquesCarette, line 31: surely a better name can be found?

src/Logic/Structural-Rules/B'-logic.agda
```diff
@@ -0,0 +1,26 @@
...
+-- proof
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Structural-Rules.B'-Logic where
+
+open import Logic.Logic
+
+record B'-Logic
```
JacquesCarette, line 14: Is this a standard name for this logic? reference?

src/Logic/Structural-Rules/B^c-logic.agda
```diff
@@ -0,0 +1,26 @@
...
+record Bᶜ-Logic
+  {Lang : Set} {Struct : Set} (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ : Lang → Lang → Lang) : Set where
+  field
+    is-logic : Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_
+
+    converse-assoc :
```
JacquesCarette, line 22: `sym-assoc` ?

JacquesCarette (changes requested):

JacquesCarette:
> At the risk of being a dissenting opinion, is this something that might be better as a standalone library? 

I certainly agree that the bar for introducing logics into `stdlib` ought to be high. But my work on MathScheme strongly hints that 'logics' are both quite algebraic and 'fit' quite well in a setting like `stdlib` if done well and in a certain style.

jamesmckinna:
Re: @MatthewDaggitt on `Data.Fin.Substitution`
The full discussion is on the PR which refactored that to move the untyped lambda calculus example to `README`, but I didn't think our intention had been to remove it entirely...

So, to the 'dissenting' view, I agree to the extent that I wouldn't be happy to give top-level `Logic.` over to this, but as `Logic.Substructural`... sure, why not?

I'd be interested to see @JacquesCarette develop his argument from MathScheme. 'Algebra of logic' seems a very respectable kind of addition... not least given my own interests towards developing (universes for) proof-theoretic constructions such as those of Harrop and successors, towards internalising certain kinds of proof procedures around classical logic/`Decidable` propositions.

mlebar-CU:
Thanks everyone for the detailed feedback. Jacques, I especially appreciate you taking the time to go through my code in such detail. I'll take some time to modify with these comments in mind and resubmit.

MatthewDaggitt:
> I certainly agree that the bar for introducing logics into stdlib ought to be high. But my work on MathScheme strongly hints that 'logics' are both quite algebraic and 'fit' quite well in a setting like stdlib if done well and in a certain style.

> 'Algebra of logic' seems a very respectable kind of addition...

Okay, that's fine! Just wanted to raise it as a possibility, if you're both happy with the addition, let's go for it.

mlebar-CU:
Hey all, just wanted to say this is still on my to do list, but the quarter has gotten away from me a bit! I'm hoping to get back to it soon.

JacquesCarette:
Think you'll get back to this @mlebar-CU ?

mlebar-CU:
> Think you'll get back to this @mlebar-CU ?

Yes! I start a PhD program in CS in ~1 month, I am planning to work on this then. Haven't had as much time in the transition period as I was hoping. Sorry for the delay!

src/Logic/Logic.agda
```diff
@@ -0,0 +1,94 @@
...
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒₁_ _⇒₂_ : Lang → Lang → Lang) →
+  Logic S _⊢_ C⟨_⟩ _⨾_ _⇒₁_ →
+  Logic S _⊢_ C⟨_⟩ _⨾_ _⇒₂_ →
+  ∀ {A B : Lang} →
+  ------------------------------------------------
+  S (A ⇒₁ B) ⊢ (A ⇒₂ B) × S (A ⇒₂ B) ⊢ (A ⇒₁ B)
+
+conditional-unique Lang Struct S _⊢_ C⟨_⟩ _⨾_ _⇒₁_ _⇒₂_ x y =
+  ⟨ Logic.if-introduction y
```
jamesmckinna, line 87: Surely this should be
```suggestion
  ⟨ Logic.⇒-introduction y
```
etc. throughout this proof?
or am I missing something?

src/Logic/Logic.agda
```diff
@@ -0,0 +1,94 @@
...
+  ∀ (Lang : Set) (Struct : Set) (S : Lang → Struct)
+  (_⊢_ : Struct → Lang → Set)
+  (C⟨_⟩ : Struct → Struct)
+  (_⨾_ : Struct → Struct → Struct)
+  (_⇒_ : Lang → Lang → Lang) →
+  Logic S _⊢_ C⟨_⟩ _⨾_ _⇒_ →
+  ∀ {A B C : Lang} →
+  (S A) ⊢ B →
+  ------------------------
+  (S (B ⇒ C)) ⊢ (A ⇒ C)
```
jamesmckinna, line 68: One way to avoid this kind of recapitulation of a quantifier prefix would be to make this definition a manifest field of the record, and then the salient structure of the construction
```agda
⇒-introduction x (⇒-elimination x (hypothesis x) y)
```
might actually be(come) visible... The rest is just... 'noise'!?

src/Logic/Logic.agda
mlebar-CU: Yes, I just missed that in my revisions! Will fix

src/Logic/Logic.agda
jamesmckinna: To illustrate: I refactored your `Logic.Logic` (a name that surely cannot stand... ;-)) as follows
```agda
-- The Agda standard library
--
-- Logic + some basic proofs about conditional

{-# OPTIONS --without-K --safe #-}

module Logic.Substructural
  {l s ℓ} {Lang : Set l} {Struct : Set s} (S : Lang → Struct)
  (_⊢_ : Struct → Lang → Set ℓ)
  (C⟨_⟩ : Struct → Struct)
  (_∙_ : Struct → Struct → Struct)
  where

open import Data.Product using (_×_; _,_; proj₁; proj₂)

private
  variable
    A B C : Lang
    X Y Z : Struct
    _⇒₁_ _⇒₂_ : Lang → Lang → Lang

record Logic (_⇒_ : Lang → Lang → Lang) : Set (l ⊔ s ⊔ ℓ) where
  field
    hypothesis :
      -------------------
      S A ⊢ A

    context :
      (C⟨ Y ⟩ ⊢ A → C⟨ X ⟩ ⊢ A) →
      Y ⊢ A →
      ---------
      X ⊢ A

    ∙-context-l :
      (C⟨ Y ⟩ ⊢ A → C⟨ X ⟩ ⊢ A) →
      ----------------------------------------
      C⟨ Z ∙ Y ⟩ ⊢ A → C⟨ Z ∙ X ⟩ ⊢ A

    ∙-context-r :
      (C⟨ Y ⟩ ⊢ A → C⟨ X ⟩ ⊢ A) →
      ----------------------------------------
      C⟨ Y ∙ Z ⟩ ⊢ A → C⟨ X ∙ Z ⟩ ⊢ A

    if-introduction :
      (X ∙ S A) ⊢ B →
      -------------
      X ⊢ (A ⇒ B)

    if-elimination :
      X ⊢ (A ⇒ B) →
      Y ⊢ A →
      -------------
      (X ∙ Y) ⊢ B

  chain-ifs : -- proof from page 23
    S A ⊢ B →
    ------------------------
    S (B ⇒ C) ⊢ (A ⇒ C)

  chain-ifs = if-introduction ∘ if-elimination hypothesis

-- Lemma 2.22 (Uniqueness of conditionals)

conditional-unique-lemma : 
  Logic _⇒₁_ →
  Logic _⇒₂_ →
  ------------------------------------------------
  S (A ⇒₁ B) ⊢ (A ⇒₂ B)

conditional-unique-lemma L₁ L₂ =
  if-introduction (if-elimination hypothesis hypothesis)
  where
  open Logic L₁ using (if-elimination; hypothesis)
  open Logic L₂ using (if-introduction)

conditional-unique :
  Logic _⇒₁_ →
  Logic _⇒₂_ →
  ------------------------------------------------
  S (A ⇒₁ B) ⊢ (A ⇒₂ B) × S (A ⇒₂ B) ⊢ (A ⇒₁ B)

conditional-unique L₁ L₂ =
  conditional-unique-lemma L₁ L₂ , conditional-unique-lemma L₂ L₁
```

And observe two things as a consequence:
* at a surface level, Agda's support for scoping mechanisms (`variable` declarations, `module` parameters) makes it possible to write the whole thing in a much more idiomatic way (YMMV?)
* at a conceptual level, there's an evident mismatch/asymmetry when eg writing the proof of `conditional-unique-lemma`, we have to choose which `Logic` from which to appeal to `hypothesis`, even though in theory they are shared in common between ` L₁` and  `L₂`...

Can/should you refactor these definitions so that the basic implicational consequence rules `⇒-introduction`, `⇒-elimination` are separated out from the common substructural framework given by `hypothesis` etc.?

What about `cut`?

jamesmckinna (changes requested):
In the absence of a 'proper' review beyond my specimen refactoring of `Logic.Logic`, I nevertheless am going to double down on my initial thoughts 1-5 as each wanting serious attention before we could consider adding this to `stdlib`. Initial thought 0 should be reconsidered in the light of any refactoring to make the code more readable, eg as suggested above.

CHANGELOG/mlebar-CU.md
```diff
@@ -0,0 +1,28 @@
+* Added a folder "Logic" in src. This provides a formalization of several 
+substructural logics based on chapter 2 of Greg Restall's "An Introduction to 
+Substructural Logics". This provides a framework for anyone wanting to study
+substructural logics.
```
jamesmckinna, line 4: This paragraph would make sense as an entry in the main `CHANGELOG` under
```agda
New modules
-----------
```

We don't add 'hierarchical' `CHANGELOG`s as perhaps you envisage(d) here. The rest of the text (together with ... more examples/documentation) should properly belong under `doc/README/Logic/...` where the `README` structure reflects the hierarchical `module` structure under `src`. Or else even simply as comments for the 'top-level' `Logic.Logic` (But I suspect `README/*` would be a better place for this.

Similarly, there's no need for your `fix-whitespace`... which is already part of our CI. See `HACKING.md`.

src/Logic/Connectives/And-Logic.agda
```diff
@@ -0,0 +1,63 @@
+----------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Logic with 'and' connective + related proof
+----------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Logic.Connectives.And-Logic where
```
jamesmckinna, line 9: As a specimen review comment:
```suggestion
module Logic.Connectives.And where
```

the redundant naming scheme you have under `Logic.Connectives` and `Logic.Logics` with the repetition of `-Logic` as a suffix is truly exhausting to read, for this reader at least.

src/Logic/Logics/BCK-Logic.agda
jamesmckinna: 💯

mlebar-CU:
> In the absence of a 'proper' review beyond my specimen refactoring of `Logic.Logic`, I nevertheless am going to double down on my initial thoughts 1-5 as each wanting serious attention before we could consider adding this to `stdlib`. Initial thought 0 should be reconsidered in the light of any refactoring to make the code more readable, eg as suggested above.

This makes sense. It's probably going to take me at least the better part of the semester to work through the refactoring, but I'm happy to work through all of this. I hope it doesn't ping you every time I push, but if it does let me know and I can find a different workflow.

Very much appreciate the time both of you put into reviewing this code and writing out comments, and I apologize it's taken me so long to get back to it - the time between the end of my master's and starting my PhD was quite busy.

jamesmckinna:
@mlebar-CU thanks for your thoughtful reply. I'd say that you shouldn't feel you need to be in a rush on this one, not least because this PR is opening whole top-level chunk of the namespace. In the meantime, I'll do my best to try to offer more (hopefully constructive!) criticism and suggestions.

PR #2829, opened 2025-09-11
Filters - alternative to the base of Domain theory

gabriellisboaconegero:
This work aims to be an alternative to the base of domain theory (#2809). There were a lot of discussions about the decisions, and talking with @TOTBWF, a good alternative would be using filters. So, we introduced a generalised way.

## Added
- Up and Down directness.
- Up and Down closure.
- IsFilter
- IsIdeal (Dual of Filter)
- Two facts about filters in relation to lattices.

## Structure
I know the current implementation does not follow the stdlib patterns, but some of the facts and definitions I was not sure about were to be put.

### DIFF
--- a/src/Relation/Binary/Filter.agda
+++ b/src/Relation/Binary/Filter.agda
@@ -1,0 +1,89 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Structures for order-theoretic filters
+------------------------------------------------------------------------
+-- As per our discussion, we should add the definition of a filter to
+-- agda-stdlib.
+
+-- ## Background
+
+-- First, some background. Let (P, ≤) be a preorder (or more generally, I
+-- guess just a relation `R : P → P → Set`). A *filter* of size κ in P consists of
+-- a predicate `F : P → Set κ` that satisfies the following conditions:
+
+-- - `F` is upwards closed: if x ≤ y and F(x), then F(y)
+-- - `F` is downwards directed: there exists some x with F(x), and for
+-- - every x, y : P with F(x) and F(y), there exists some z : P with
+--   F(z) × z ≤ x × z ≤ y
+
+-- Some useful facts:
+
+-- - If P is a meet semilattice, then conditions 1 and 2 jointly imply that
+--   F(⊤), and conditions 1 and 3 mean that F(x) × F(y) → F(x ∧ y), where ∧
+--   is the meet.
+-- - Alternatively, if P is a meet semilattice, then a κ-small filter is
+--   equivalent to a meet semilattice homomorphism P → Type κ, where the
+--   unit type is the top element, and meets are given by products of
+--   types.
+
+-- ## Programming
+
+-- There does need to be some thought as to organization. We should go with
+-- the definition I listed first as our primary definition, as it is the
+-- most general. The useful facts would be nice to prove as theorems.
+
+-- Moreover, I think we should place the definition either in it's own file
+-- `Relation.Binary.Filter`, or inside of a folder like
+-- `Relation.Binary.Diagram.Filter`; not sure here. Theorems would be
+-- somewhere in Relation.Binary.Lattice.Filter or something.
+
+-- We might want to factor out the "downward directed" and "upwards closed" bits into their own
+-- definitions? Could also place all of this in `Relation.Binary.Subset` or
+-- something, as they are all definitions regarding subsets of binary
+-- relations? Note that there are also "downwards closed" and "upwards
+-- directed" sets, so be careful with names!
+
+-- PS: for bonus points, you could also add *ideals* in a preorder: these
+-- are the duals of filters; IE: downwards-closed, upwards directed.
+
+-- Hope that makes sense,
+-- Reed :)
+{-# OPTIONS --cubical-compatible --safe #-}
+
+module Relation.Binary.Filter where
+
+open import Relation.Binary.Core using (Rel)
+open import Data.Product.Base using (∃-syntax; _×_; _,_)
+open import Level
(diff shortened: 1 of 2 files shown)


src/Relation/Binary/Filter.agda
```diff
@@ -0,0 +1,89 @@
...
+open import Relation.Binary.Structures using (IsPreorder)
+open import Function.Base using (flip)
+
+private
+  variable
+    a κ ℓ₁ ℓ₂ : Level
+    A : Set a
+
+UpwardClosure : (A → Set ℓ₁) → Rel A ℓ₂ → Set _
+UpwardClosure {A = A} F _≤_ = ∀ (x y : A) → (x ≤ y) × (F x) → F y
```
jamesmckinna: [DRY]
This is redundant, as an instance of `Relation.Binary.Definitions._Respects_`... noting that the use of of `_×_` again involves redundant currying/uncurrying.

src/Relation/Binary/Filter.agda
```diff
@@ -0,0 +1,89 @@
...
+-- ## Background
+
+-- First, some background. Let (P, ≤) be a preorder (or more generally, I
+-- guess just a relation `R : P → P → Set`). A *filter* of size κ in P consists of
+-- a predicate `F : P → Set κ` that satisfies the following conditions:
+
+-- - `F` is upwards closed: if x ≤ y and F(x), then F(y)
+-- - `F` is downwards directed: there exists some x with F(x), and for
+-- - every x, y : P with F(x) and F(y), there exists some z : P with
+--   F(z) × z ≤ x × z ≤ y
```
jamesmckinna, line 18: I suppose that the thing which makes me uneasy about such a definition is that it fuses two aspects of the definition, in the interest of avoiding some of the difficulties that Reed alludes to in #2814 , namely:
* the 'subset' implicitly defined by the predicate `F`
* the abstract property of a relation, here '`_≤_` restricted to the subset defined by `F`', as being directed

I think as far as `stdlib` is concerned, it would be (more) helpful to isolate the second one, as for example in #2813 (where it would make sense simply to cherry-pick the definition, and perhaps some of its properties, and leave the `Pointed` extension construction to a separate PR), and as to the first, perhaps introduce a/the construction, analogous to `Function.Base._on_`, corresponding to the first one?

Not sure whether we should continue the discussion here, or else on #2814 though.

TOTBWF:
@gabriellisboaconegero and I have had some offline discussions about using a homomorphism-based definition that seems to be the One True Way. The idea is that we can define a setoidal analog of the order on propositions by defining a poset whose carrier is `Set κ`, where equivalence is bi-implication and `P ≤ Q` as `P → Q`, as in the following code.
```agda
module Cool where

open import Level

open import Data.Product

open import Function.Base

open import Relation.Binary.Bundles

open IsPreorder
open IsPartialOrder

record _↔_ {ℓ ℓ'} (P : Set ℓ) (Q : Set ℓ') : Set (ℓ ⊔ ℓ') where
  field
    to : P → Q
    from : Q → P

open _↔_

private variable
  ℓ : Level
  P Q R : Set ℓ

id↔ : P ↔ P
id↔ .to = id
id↔ .from = id

_∙↔_ : P ↔ Q → Q ↔ R → P ↔ R
(f ∙↔ g) .to = g .to ∘ f .to
(f ∙↔ g) .from = f .from ∘ g .from

_↔⁻¹ : P ↔ Q → Q ↔ P
(f ↔⁻¹) .to = f .from
(f ↔⁻¹) .from = f .to

BiImplicationIsEquivalence : ∀ {ℓ} → IsEquivalence (_↔_ {ℓ})
BiImplicationIsEquivalence .IsEquivalence.refl = id↔
BiImplicationIsEquivalence .IsEquivalence.sym = _↔⁻¹
BiImplicationIsEquivalence .IsEquivalence.trans = _∙↔_

Omega : ∀ (κ : Level) → Poset (suc κ) κ κ
Omega κ .Poset.Carrier = Set κ
Omega κ .Poset._≈_ P Q = P ↔ Q
Omega κ .Poset._≤_ P Q = P → Q
Omega κ .Poset.isPartialOrder .isPreorder .isEquivalence = BiImplicationIsEquivalence
Omega κ .Poset.isPartialOrder .isPreorder .reflexive p↔q = p↔q .to
Omega κ .Poset.isPartialOrder .isPreorder .trans p→q q→r = q→r ∘ p→q
Omega κ .Poset.isPartialOrder .antisym p→q q→r .to = p→q
Omega κ .Poset.isPartialOrder .antisym p→q q→r .from = q→r
```

This removes the need for upwards and downwards closure, as a homomorphism `P → Omega κ` is exactly an upwards closed subset (and likewise for `Opposite P → Omega κ` and downwards closure).

Next, we can define the order-theoretic analog of the category of elements. This transforms a monotone map `F : P → Omega κ` into an order `Restrict F` whose carrier is `Σ (p : P.Carrier) (F p)`, where the order and equivalence are given purely by the `p` component. In other words, this lets us restrict an order to an upwards closed subset.

Finally, we can define a filter as a monotone map `F : P → Omega κ` such that `Restrict F` is downwards directed. This seems like a nice point in the design space, as we get to re-use a lot of pre-existing notions. Moreover, this is a direct decategorification of the definition of a [flat](flat functor), which is always a good sign.

EDIT: It was pointed out on zulip that this equivalence relation already exists as https://agda.github.io/agda-stdlib/master/Function.Properties.Equivalence.html#1935, so it should be really easy to assemble this from kit that we already have.

JacquesCarette:
Given that the design, such as it is, has evolved to no longer resemble this PR at all, best to just close this one completely. Category of elements and filters can be defined again later, when needed.

PR #2830, opened 2025-09-22
Add `doc/` to the target in Makefile and make the `cabal` command customizable

shhyou:
The commits include two changes:
1. Add `doc/` prefix to the `Everything.agda` targets in GNUmakefile
2. Make `cabal` a makefile variable, thus allowing the user to substitute a different command (or designate a specific binary).

I believe (1) is a leftover of #2184. This change stops two consecutive `make listings` calls from re-generating the documentation.

(2) allows the user to use custom commands, e.g. using stack:
```
$ ln -s stack-9.10.2.yaml stack.yaml
$ stack build
$ make listings CABAL_RUN_COMMAND='stack exec' AGDA_OPTIONS=
```

However, `fix-whitespace`, `check-whitespace` and `test` seem to be cabal-specific, so I only make the cabal path customizable.

CI: shhyou/agda-stdlib#1

### DIFF
--- a/stack-9.10.3.yaml
+++ b/stack-9.10.3.yaml
@@ -1,5 +1,5 @@
-snapshot: lts-24.11
-compiler: ghc-9.10.2
+snapshot: lts-24.17
+compiler: ghc-9.10.3
 compiler-check: match-exact
 packages:
 - .


shhyou:
I believe GenerateEverything already puts Everything.agda in doc since #2184. This PR does not change its `--out-dir`.

MatthewDaggitt:
This seems a reasonable set of changes to me!

shhyou:
@gallais For the change from `Everything.agda:` to `doc/Everything.agda:` in Makefile, is that a correct change? Because I see that `GenerateEverything` already puts `Everything.agda` in `doc/`.

GNUmakefile
```diff
@@ -2,34 +2,36 @@ AGDA_EXEC ?= agda
 AGDA_OPTIONS=-Werror
 AGDA_RTS_OPTIONS=+RTS -M4.0G -H3.5G -A128M -RTS
 AGDA=$(AGDA_EXEC) $(AGDA_OPTIONS) $(AGDA_RTS_OPTIONS)
+CABAL_EXEC ?= cabal
```
andreasabel, line 5: Did you mean to call this `CABAL` (because you are referring to it as such below).

shhyou: Oops thanks for catching the bug. I'll fix them later.

andreasabel: Ok then I flag this as "changes requested".

andreasabel (changes requested):
Fix the CABAL env var

GNUmakefile
shhyou: @andreasabel finished!

shhyou:
I think this is ready.

andreasabel (approved):
A sorry, I seemed to have blocked this with my review.
Changes accepted!

PR #2832, opened 2025-09-26
De Morgan's laws for Pred

### LINKED ISSUE #2831: ``¬∃⇒∀¬``
Needs this proof to be in the library?
```
¬∃⇒∀¬ :  ... {P : Pred A p} → ¬ ∃ (\x → P x) → (∀ x → ¬ P x)
¬∃⇒∀¬ ¬∃P x Px =   ¬∃P (x , Px)
```
Also I think that the `reverse' cannot be proved:
```
¬∀⇒∃¬ :  ... ¬ (∀ x → P x)  → ∃ (\x → ¬ P x)
```
Can it?

Taneb:
Closes #2831


src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
 ∁U-Empty : Empty {A = A} (∁ U)
 ∁U-Empty = λ x x∈∁U → x∈∁U _
 
+------------------------------------------------------------------------
+-- De Morgan's laws
+
+¬∃⟨P⟩⇒Π[∁P] : ∀ {P : Pred A ℓ} → ¬ ∃⟨ P ⟩ → Π[ ∁ P ]
+¬∃⟨P⟩⇒Π[∁P] ¬sat x Px = ¬sat (x , Px)
```
JacquesCarette: I wonder if this can be simplified to just `uncurry'` ?

JacquesCarette: (I'm not saying to should be implemented that way...)

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+¬∃⟨P⟩⇒∀[∁P] : ∀ {P : Pred A ℓ} → ¬ ∃⟨ P ⟩ → ∀[ ∁ P ]
+¬∃⟨P⟩⇒∀[∁P] ¬sat Px = ¬sat (-, Px)
+
+∃⟨∁P⟩⇒¬Π[P] : ∀ {P : Pred A ℓ} → ∃⟨ ∁ P ⟩ → ¬ Π[ P ]
+∃⟨∁P⟩⇒¬Π[P] (x , ¬Px) ΠP = ¬Px (ΠP x)
+
+∃⟨∁P⟩⇒¬∀[P] : ∀ {P : Pred A ℓ} → ∃⟨ ∁ P ⟩ → ¬ ∀[ P ]
+∃⟨∁P⟩⇒¬∀[P] (x , ¬Px) ∀P = ¬Px ∀P
```
JacquesCarette: Replace `x` by `_`

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+Π[∁P]⇒¬∃[P] : ∀ {P : Pred A ℓ} → Π[ ∁ P ] → ¬ ∃⟨ P ⟩
+Π[∁P]⇒¬∃[P] Π∁P (x , Px) = Π∁P x Px
```
JacquesCarette, line 71: (again, would `curry'` typecheck?_

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+∀[∁P]⇒¬∃[P] : ∀ {P : Pred A ℓ} → ∀[ ∁ P ] → ¬ ∃⟨ P ⟩
+∀[∁P]⇒¬∃[P] ∀∁P (x , Px) = ∀∁P Px
```
JacquesCarette: elide `x` here too

JacquesCarette (changes requested):
Two comments, two requests for change.

Massively surprised this wasn't here already!

src/Relation/Unary/Properties.agda
Taneb: I get the following error when I try, which I don't really understand:
```
/home/taneb/Documents/misc/agda/stdlib/master/src/Relation/Unary/Properties.agda:59.15-22: error: [UnequalTerms]
Data.Irrelevant.Irrelevant Data.Empty.Empty !=<
(y : _B_94 x) → _C_95 (x , y)
when checking that the expression uncurry has type
¬ Satisfiable P → Universal (∁ P)
```

JacquesCarette (approved):

src/Relation/Unary/Properties.agda
jamesmckinna: It might need more eta expansion?

The `...Irrelevant.Irrelevant` is coming from the empty type (I conjecture), so the type checker seems to having trouble unfolding things enough to unify properly... weird though!?

jamesmckinna:
So... I think that all of these things are already covered by [`Relation.Nullary.Negation`](https://github.com/agda/agda-stdlib/blob/54f5c38799f6c0125f7dc914e073670b549bcbe6/src/Relation/Nullary/Negation.agda#L32-L48) (including even @JacquesCarette 's observations about `curry` etc.)? Or not!?

JacquesCarette:
I think you are correct @jamesmckinna . But there must also be something very slightly different about them? Perhaps some subtlety about irrelevance?

jamesmckinna:
The differences, AFAICT, are that the lemmas in `Relation.Nullary.Negation` only concern the `Universal` (explicitly quantified) cases, whereas this PR also covers the `IUniversal` (implicitly quantified) ones as well. But, FTR, using the suggested import:
```agda
open import Relation.Nullary.Negation using (¬_; ∃⟶¬∀¬; ∀⟶¬∃¬; ¬∃⟶∀¬; ∀¬⟶¬∃; ∃¬⟶¬∀)
module _ {P : Pred A ℓ} where

  ¬∃⟨P⟩⇒Π[∁P] : ¬ ∃⟨ P ⟩ → Π[ ∁ P ]
  ¬∃⟨P⟩⇒Π[∁P] = ¬∃⟶∀¬

  ¬∃⟨P⟩⇒∀[∁P] : ¬ ∃⟨ P ⟩ → ∀[ ∁ P ]
  ¬∃⟨P⟩⇒∀[∁P] ¬sat = ¬∃⟶∀¬ ¬sat _

  ∃⟨∁P⟩⇒¬Π[P] : ∃⟨ ∁ P ⟩ → ¬ Π[ P ]
  ∃⟨∁P⟩⇒¬Π[P] = ∃¬⟶¬∀

  ∃⟨∁P⟩⇒¬∀[P] : ∃⟨ ∁ P ⟩ → ¬ ∀[ P ]
  ∃⟨∁P⟩⇒¬∀[P] ∃CP ∀P = ∃¬⟶¬∀ ∃CP λ _ → ∀P

  Π[∁P]⇒¬∃[P] : Π[ ∁ P ] → ¬ ∃⟨ P ⟩
  Π[∁P]⇒¬∃[P] = ∀¬⟶¬∃

  ∀[∁P]⇒¬∃[P] : ∀[ ∁ P ] → ¬ ∃⟨ P ⟩
  ∀[∁P]⇒¬∃[P] ∀∁P = ∀¬⟶¬∃ λ _ → ∀∁P
```

So... I think this is worth badging as `status:duplicate` or even `status:invalid`?

src/Relation/Unary/Properties.agda
jamesmckinna: No, it looks as though I was wrong; it's really the 'usual' implicit vs. explicit quantification nonsense, it would seem, given the (correct) solutions I offer below. The error messages have got a lot worse in diagnosing/being able to self-diagnose this, however... :-(

jamesmckinna:
Do we want to merge this, or close, and follow up with a rename+deprecate as discussed in #2831 ?

I'm (largely) agnostic/uncommitted on this issue (apart from the redundancy involved), but proofs which differ at all from those versions offered above would be... redundancy too far, IMNSVHO ;-)

src/Relation/Unary/Properties.agda
jamesmckinna: See below

JacquesCarette:
I think you're right about all of this being redundant. I think the main remaining point is around findability and naming:
1. Neither @Taneb nor Sergei were able to find these
2. `Relation.Nullary` is a hideous prefix. Are there such things as non-trivial nullary relations? This feels like a horrible hack from ages ago that 3.0 should get rid of.

jamesmckinna:
I broadly agree:
* for v2.4, I think we should get the names/namingconvention right see #2838 
* for v3.0, we can indeed revisit these larger-scale modular hierarchies

But it's a testament to Stockholm Syndrome that I *have* come to internalise the `Nullary`/`Unary`/`Binary` taxonomy ;-)

jamesmckinna:
@Taneb suggest merging your commits here, or my suggestions above, to #2838 so that we can knock this off, at least for v2.4?

Taneb:
Why don't we merge both PRs separately? fwiw I'm (very slightly) against your suggestions: they add a new import to the dependency graph to turn a bunch of one line definitions to a bunch of slightly more abbreviated one line definitions

jamesmckinna:
> Why don't we merge both PRs separately? fwiw I'm (very slightly) against your suggestions: they add a new import to the dependency graph to turn a bunch of one line definitions to a bunch of slightly more abbreviated one line definitions

Fair enough, but, *pace* the discoverability problem (not trying to discount it), I'm not sure the duplication in `Unary` makes (better?) sense... but for the sake of v2.4 (and the opportunity for a `breaking` redesign in v3.0), happy to merge this too.

jamesmckinna (changes requested):
Happy to merge this, but I'd push a bit harder on @JacquesCarette 's suggestion to use `curry`/`uncurry` where possible... (eg by inlining the definitions given above?)

JacquesCarette (approved):

src/Relation/Unary/Properties.agda
```diff
@@ -8,7 +8,7 @@
 
 module Relation.Unary.Properties where
 
-open import Data.Product.Base as Product using (_×_; _,_; swap; proj₁; zip′)
+open import Data.Product.Base as Product using (_×_; _,_; -,_; swap; proj₁; zip′; curry)
```
jamesmckinna, line 11:
```suggestion
open import Data.Product.Base as Product using (_×_; _,_; -,_; swap; proj₁; zip′; curry; uncurry)
```
As before, I'd rather use the generic operations where possible.

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
 ∁U-Empty = λ x x∈∁U → x∈∁U _
 
+------------------------------------------------------------------------
+-- De Morgan's laws
+
+¬∃⟨P⟩⇒Π[∁P] : ∀ {P : Pred A ℓ} → ¬ ∃⟨ P ⟩ → Π[ ∁ P ]
+¬∃⟨P⟩⇒Π[∁P] ¬sat = curry ¬sat
+
+¬∃⟨P⟩⇒∀[∁P] : ∀ {P : Pred A ℓ} → ¬ ∃⟨ P ⟩ → ∀[ ∁ P ]
+¬∃⟨P⟩⇒∀[∁P] ¬sat = curry ¬sat _
```
jamesmckinna, line 62: Maybe coupling is considered harmful, but given the amount of duplication/repetition/redundancy here, I would prefer to emphasise it by delegation:
```suggestion
¬∃⟨P⟩⇒∀[∁P] ¬sat = ¬∃⟨P⟩⇒Π[∁P] ¬sat _
```

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+∃⟨∁P⟩⇒¬Π[P] : ∀ {P : Pred A ℓ} → ∃⟨ ∁ P ⟩ → ¬ Π[ P ]
+∃⟨∁P⟩⇒¬Π[P] (x , ¬Px) ΠP = ¬Px (ΠP x)
```
jamesmckinna, line 65:
```suggestion
∃⟨∁P⟩⇒¬Π[P] (_ , ¬Px) ΠP = ¬Px (ΠP _)
```
using the underscores here:
* is more robust (debatable?)
* formally mirrors the subsequent lemma

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+∃⟨∁P⟩⇒¬∀[P] : ∀ {P : Pred A ℓ} → ∃⟨ ∁ P ⟩ → ¬ ∀[ P ]
+∃⟨∁P⟩⇒¬∀[P] (_ , ¬Px) ∀P = ¬Px ∀P
+
+Π[∁P]⇒¬∃[P] : ∀ {P : Pred A ℓ} → Π[ ∁ P ] → ¬ ∃⟨ P ⟩
+Π[∁P]⇒¬∃[P] Π∁P (x , Px) = Π∁P x Px
```
jamesmckinna, line 71:
```suggestion
Π[∁P]⇒¬∃[P] = uncurry
```
suffices!

src/Relation/Unary/Properties.agda
```diff
@@ -52,6 +52,27 @@ U-Universal = λ _ → _
...
+
+∀[∁P]⇒¬∃[P] : ∀ {P : Pred A ℓ} → ∀[ ∁ P ] → ¬ ∃⟨ P ⟩
+∀[∁P]⇒¬∃[P] ∀∁P (_ , Px) = ∀∁P Px
```
jamesmckinna, line 74:
```suggestion
∀[∁P]⇒¬∃[P] = Π[∁P]⇒¬∃[P] (λ _ → ∀∁P)
```
again, emphasise the delegation to the explicit version.

jamesmckinna (approved):
I won't stand in the way of what's here already, but I do think the suggestions I make improve things, but without adding the extra node to the dependency graph, while emphasising the generic/delegation aspects of all these proofs.

jamesmckinna:
There are (lots of!?) potential `library-design` issues raised by this PR:
* navigability/discoverability, as already noted
* redundant recoding as a side effect of the above
* finessing implicit vs. explicit function spaces (I find `Function.Base.{_$-|λ-}` too fiddly in practice to use, but maybe they could be deployed here?)
* the `Relation.*ary` hierarchy, and 'opportunities' to refactor it
* ...

so perhaps worth lifting out as a new issue?

jamesmckinna:
Do you want the refactoring suggestions, or shall we simply merge as is @Taneb ?

jamesmckinna:
Happy to (re-)consider my suggestions re: delegation/refactoring downstream, so merging now.

PR #2833, opened 2025-09-26
More distributivity properties for nats

Taneb:
I based the types for the list action distributivity properties on `Algebra.Properties.Semiring.Sum`, which uses `Vector` rather than `List` (and of course is for an arbtirary `Semiring`)

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1068,7 +1068,7 @@ m<n⇒m<o*n = m≤n⇒m≤o*n
   (m * (m ^ n)) * (m ^ o) ∎
 
 ^-distribʳ-* : ∀ m n o → (n * o) ^ m ≡ n ^ m * o ^ m
-^-distribʳ-* zero n o = refl
+^-distribʳ-* zero n o = sym (*-identityˡ 1)
 ^-distribʳ-* (suc m) n o = begin-equality
   (n * o) * (n * o) ^ m     ≡⟨ cong (n * o *_) (^-distribʳ-* m n o) ⟩
   (n * o) * (n ^ m * o ^ m) ≡⟨ [m*n]*[o*p]≡[m*o]*[n*p] n o (n ^ m) (o ^ m) ⟩


src/Data/Nat/ListAction/Properties.agda
```diff
@@ -51,6 +53,14 @@ sum-++ (m ∷ ms) ns = begin
   (m + sum ms) + sum ns  ∎
   where open ≡-Reasoning
 
+*-distribˡ-sum : ∀ m ns → m * sum ns ≡ sum (map (m *_) ns)
+*-distribˡ-sum m [] = *-zeroʳ m
+*-distribˡ-sum m (n ∷ ns) = trans (*-distribˡ-+ m n (sum ns)) (cong (m * n +_) (*-distribˡ-sum m ns))
+
+*-distribʳ-sum : ∀ m ns → sum ns * m ≡ sum (map (_* m) ns)
+*-distribʳ-sum m [] = *-zeroˡ m
```
Taneb, line 61: I use an explicit invocation of `*-zeroˡ` rather than `refl` to keep the proofs symmetric

jamesmckinna:
I'm happy with the local details of this PR, but I'm starting (since already a while now) to be very dissatisfied with each of the underlying design decisions, notwithstanding my own refactorings which, like this one, have perpetuated suboptimal choices...
* `Vector` rather than `Vec` (but this ought to be inessential, in view of the following)
* `sum` as an instance of a `Foldable` operation/operation on a `Foldable` functor.

I'm grateful to Oleg Kiselyov for drawing my attention to various invited talks by Guy Steele over the 2010s about this in the context of parallel programming in general, and FORTRAN/Fortress in particular, but also to a (much) longer history (1970s?  1980s?) in the Dijkstra et al. lineage of so-called "Eindhoven Quantifier Notation" (discussed eg [here](https://link.springer.com/chapter/10.1007/11783596_7)), as well as Bertot et al. ["Canonical Big Operators" in Coq/SSReflect](https://link.springer.com/chapter/10.1007/978-3-540-71067-7_11) (as it then was) in the 2000s.

We can, and should, do better! 

(And yes, before @JacquesCarette feels obliged to remind me about about the good/perfect friend/enemy spectrum of trade-offs, I should lift this out as a separate `library-design` issue)

Shoutouts also to Johannes Waldmannn for his [bracing 2018 essay on lists in haskell](https://arxiv.org/abs/1808.08329) ...

src/Data/Nat/Properties.agda
```diff
@@ -1067,6 +1067,13 @@ m<n⇒m<o*n = m≤n⇒m≤o*n
   m * ((m ^ n) * (m ^ o)) ≡⟨ sym (*-assoc m _ _) ⟩
   (m * (m ^ n)) * (m ^ o) ∎
 
+^-distribʳ-* : ∀ m n o → (n * o) ^ m ≡ n ^ m * o ^ m
+^-distribʳ-* zero n o = refl
```
jamesmckinna: But no explicit appeals here to `*-zero`?

Taneb: This is the `(n * o) ^ 0 ≡ n ^ 0 * o ^ 0`, which simplifies to `1 ≡ 1 * 1` and then `1 ≡ 1`

jamesmckinna: Of course! (sorry, brain fog).
But still, for the sake of (the `Semiring`) abstraction, is it worth writing `sym (*-identityˡ _)` instead?

Taneb:
> `Vector` rather than `Vec` (but this ought to be inessential, in view of the following)

@jamesmckinna 

Is that relevant here? This PR is specifically about `List`, even though I chose the types based on an earlier implementation on `Vector`.

I accept that your point about a hypothetical `Foldable` stands, however.

jamesmckinna:
> > `Vector` rather than `Vec` (but this ought to be inessential, in view of the following)
> 
> @jamesmckinna
> 
> Is that relevant here? This PR is specifically about `List`, even though I chose the types based on an earlier implementation on `Vector`.
> 
> I accept that your point about a hypothetical `Foldable` stands, however.

Sorry @Taneb I think I should look a bit more closely (and your review request means that I will, but perhaps not straightaway ;-) I'm getting over a very unpleasant bug just atm) at your code. The `List` version morally seems just as bad as any other choice of representation, but I suppose we do eventually have to pick one!?

Taneb:
> Against that, I'm surprised that
> 
>     * `*-distribˡ-sum` isn't _easily_ derivable by commutativity of `_*_` from `*-distribʳ-sum`

I hadn't previously even tried this. It _can_ be done as the follows:

```agda
*-distribʳ-sum′ m ns = begin
  sum ns * m          ≡⟨ *-comm (sum ns) m ⟩
  m * sum ns          ≡⟨ *-distribˡ-sum m ns ⟩
  sum (map (m *_) ns) ≡⟨ cong sum (map-cong (*-comm m) ns) ⟩
  sum (map (_* m) ns) ∎
```
But this is longer than defining it directly, and there aren't any other uses for pulling out a lemma like `Algebra.Consequences.Propositional.comm∧distrʳ⇒distrˡ`

>     * that while `*-distribʳ-sum` and `^-distribʳ-product` do have _identical_ proofs, considered each of them as properties of an abstract monoid-action-as-fold, we don't have the same abstraction/similarity between `*-distribʳ-+` and `^-distribʳ-*`?

I think this is because `_*_` is defined recursive on its _first_ argument, whereas `_^_` is recursive on its _second_ argument.

PR #2836, opened 2025-10-07
Factor antisymmetry proof for sublist

andreasabel:
Commits:

- Simplify Sublist/Heterogeneous/antisym: collapse two cases:
   Removes an unnecessary case split.
  
- Simplify Sublist/Heterogeneous/antisym: factor out contradiction:
   Pulls the duplicated proof into a lemma.
  
These can be squashed, but I worked in two steps since I am unsure about the style policy.

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
@@ -400,18 +400,19 @@ module Antisymmetry
 
   open ℕ.≤-Reasoning
 
+  private
+    antisym-lemma : ∀ xs ys₁ y → Sublist R xs ys₁ → ¬ Sublist S (y ∷ ys₁) xs
+    antisym-lemma xs ys₁ y rs ss = ℕ.<-irrefl ≡.refl (begin
+      length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
+      length xs        ≤⟨ length-mono-≤ rs ⟩
+      length ys₁       ∎)
+
   antisym : Antisym (Sublist R) (Sublist S) (Pointwise E)
   -- impossible cases
   antisym (_∷ʳ_ {xs} {ys₁} y rs) ss =
-    contradiction (begin
-    length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
-    length xs        ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
+    case (antisym-lemma xs ys₁ y rs ss) of λ()
   antisym (_∷_ {x} {xs} {y} {ys₁} r rs)  (_∷ʳ_ {ys₂} {zs} z ss) =
-    contradiction (begin
-    length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
-    length xs        ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
+    case (antisym-lemma xs ys₁ y rs ss) of λ()
   -- diagonal cases
   antisym []        []        = []
   antisym (r ∷ rs)  (s ∷ ss)  = rs⇒e r s ∷ antisym rs ss
--- a/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
@@ -401,18 +401,17 @@ module Antisymmetry
   open ℕ.≤-Reasoning
 
   private
-    antisym-lemma : ∀ xs ys₁ y → Sublist R xs ys₁ → ¬ Sublist S (y ∷ ys₁) xs
-    antisym-lemma xs ys₁ y rs ss = ℕ.<-irrefl ≡.refl (begin
-      length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
-      length xs        ≤⟨ length-mono-≤ rs ⟩
-      length ys₁       ∎)
+    antisym-lemma : Sublist R xs ys → ¬ Sublist S (y ∷ ys) xs
+    antisym-lemma {xs} {ys} {y} rs ss = ℕ.<-irrefl ≡.refl $ begin
+      length (y ∷ ys) ≤⟨ length-mono-≤ ss ⟩
+      length xs       ≤⟨ length-mono-≤ rs ⟩
+      length ys       ∎
 
   antisym : Antisym (Sublist R) (Sublist S) (Pointwise E)
   -- impossible cases
-  antisym (_∷ʳ_ {xs} {ys₁} y rs) ss =
-    contradiction′ (antisym-lemma xs ys₁ y rs) ss
+  antisym (_ ∷ʳ rs) ss = contradiction′ (antisym-lemma rs) ss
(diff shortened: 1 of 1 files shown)


src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -400,26 +400,22 @@ module Antisymmetry
 
   open ℕ.≤-Reasoning
 
+  private
+    antisym-lemma : ∀ xs ys₁ y → Sublist R xs ys₁ → ¬ Sublist S (y ∷ ys₁) xs
+    antisym-lemma xs ys₁ y rs ss = ℕ.<-irrefl ≡.refl (begin
+      length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
+      length xs        ≤⟨ length-mono-≤ rs ⟩
+      length ys₁       ∎)
```
andreasabel: Conservatively, I made this lemma private.
There could potentially be some use of this lemma in other contexts, but then it would need a proper name.

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -400,26 +400,22 @@ module Antisymmetry
...
+    case (antisym-lemma xs ys₁ y rs ss) of λ()
   antisym (_∷_ {x} {xs} {y} {ys₁} r rs)  (_∷ʳ_ {ys₂} {zs} z ss) =
-    contradiction (begin
-    length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
-    length xs        ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
+    case (antisym-lemma xs ys₁ y rs ss) of λ()
+  -- diagonal cases
+  antisym []        []        = []
+  antisym (r ∷ rs)  (s ∷ ss)  = rs⇒e r s ∷ antisym rs ss
```
andreasabel, line 417: Cases are reordered so as to avoid an early split on `ss`.

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -400,26 +400,22 @@ module Antisymmetry
...
-    length zs        ≤⟨ ℕ.n≤1+n (length zs) ⟩
-    length (z ∷ zs)  ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
-  antisym (_∷ʳ_ {xs} {ys₁} y rs) (_∷_ {y} {ys₂} {z} {zs} s ss)  =
-    contradiction (begin
-    length (z ∷ zs) ≤⟨ length-mono-≤ rs ⟩
-    length ys₁      ≤⟨ length-mono-≤ ss ⟩
-    length zs       ∎) $ ℕ.<-irrefl ≡.refl
+  antisym (_∷ʳ_ {xs} {ys₁} y rs) ss =
+    case (antisym-lemma xs ys₁ y rs ss) of λ()
```
MatthewDaggitt: We usually prefer to use `contradiction` in cases like this to make the proof strategy more explicit

MatthewDaggitt (changes requested):
Apart from that looks good to me!

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
andreasabel: So, `contradiction' (antisym-lemma xs ys₁ y rs) ss` ?

MatthewDaggitt: Yup

andreasabel: Did that.

MatthewDaggitt (approved):

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
jamesmckinna: Regardless of possible reuse (could be treated downstream?), suggest to use the `variable`s already declared, and `_$_` to reduce the parenthesis cognitive load:
```suggestion
    antisym-lemma : Sublist R xs ys → ¬ Sublist S (y ∷ ys) xs
    antisym-lemma {xs} {ys} {y} rs ss = ℕ.<-irrefl ≡.refl $ begin
      length (y ∷ ys) ≤⟨ length-mono-≤ ss ⟩
      length xs       ≤⟨ length-mono-≤ rs ⟩
      length ys       ∎
```

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -400,26 +400,22 @@ module Antisymmetry
...
-    length zs        ≤⟨ ℕ.n≤1+n (length zs) ⟩
-    length (z ∷ zs)  ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
-  antisym (_∷ʳ_ {xs} {ys₁} y rs) (_∷_ {y} {ys₂} {z} {zs} s ss)  =
-    contradiction (begin
-    length (z ∷ zs) ≤⟨ length-mono-≤ rs ⟩
-    length ys₁      ≤⟨ length-mono-≤ ss ⟩
-    length zs       ∎) $ ℕ.<-irrefl ≡.refl
+  antisym (_∷ʳ_ {xs} {ys₁} y rs) ss =
+    contradiction′ (antisym-lemma xs ys₁ y rs) ss
```
jamesmckinna: If you take the above suggestion, then this simplifies to
```suggestion
  antisym (_ ∷ʳ rs) ss = contradiction′ (antisym-lemma rs) ss
```

src/Data/List/Relation/Binary/Sublist/Heterogeneous/Properties.agda
```diff
@@ -400,26 +400,22 @@ module Antisymmetry
...
   antisym (_∷_ {x} {xs} {y} {ys₁} r rs)  (_∷ʳ_ {ys₂} {zs} z ss) =
-    contradiction (begin
-    length (y ∷ ys₁) ≤⟨ length-mono-≤ ss ⟩
-    length xs        ≤⟨ length-mono-≤ rs ⟩
-    length ys₁       ∎) $ ℕ.<-irrefl ≡.refl
+    contradiction′ (antisym-lemma xs ys₁ y rs) ss
```
jamesmckinna: And, similarly
```suggestion
    contradiction′ (antisym-lemma rs) ss
```

jamesmckinna (approved):
Nitpicks (mostly cosmetic), but using implicits/variables makes proofs more robust (I think!?)

jamesmckinna:
@andreasabel I would merge with my suggestions, unless you feel strongly that they make things worse?

Separately, you allude to (the possibility of) a default squash/merge policy for `stdlib`, and my `git` fu is weak in this regard: AFAIU, we squash commits while merging, but I'm conscious also that this is not the case for the main Agda codebase. 

I'd be grateful for pointers to how to understand the trade offs between these approaches, and what conclusions, if any, we might draw for stdlib maintenance/development?

andreasabel:
I committed your suggestions, thanks!

Merging by rebase would preserve the original commits.  We do this in the Agda codebase if they are all meaningful and pass the build (and maybe the tests) individually.  This can have the advantage of smaller commits too look at when bisecting which commit introduced a bug.  
However, for the standard library there is probably no gain from small commits, so squashing is fine as default policy.

PR #2837, opened 2025-10-09
[ deprecate ] names for deMorgan lemmas in `Data.Fin.Properties`

jamesmckinna:
This is a coda/precursor to #2801 which does the deprecation of names outlined there.


jamesmckinna:
As with #2838 the stakes here are small enough that I'm going to merge now.

PR #2838, opened 2025-10-11
[ deprecate ] names for deMorgan lemmas in `Relation.Nullary.Negation`

jamesmckinna:
See #2831 and #2832 ...
... suggest the additional (implicit) lemmas from the latter PR be merged into this one?


jamesmckinna:
There's little at stake here, so merging now...

PR #2841, opened 2025-10-23
Fix universe level typo in `Function.Construct.Constant`

jacquescomeaux:
Fix typo preventing `Function.Construct.Constant.function` from being as general as possible

### DIFF
--- a/src/Function/Construct/Constant.agda
+++ b/src/Function/Construct/Constant.agda
@@ -53,7 +53,7 @@ module _
 ------------------------------------------------------------------------
 -- Setoid bundles
 
-module _ (S : Setoid a ℓ₂) (T : Setoid b ℓ₂) where
+module _ (S : Setoid a ℓ₁) (T : Setoid b ℓ₂) where
 
   open Setoid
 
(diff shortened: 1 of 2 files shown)


src/Function/Construct/Constant.agda
```diff
@@ -53,7 +53,7 @@ module _
 ------------------------------------------------------------------------
 -- Setoid bundles
 
-module _ (S : Setoid a ℓ₂) (T : Setoid b ℓ₂) where
+module _ (S : Setoid a ℓ₁) (T : Setoid b ℓ₂) where
```
jamesmckinna, line 56: I'm happy with this as well. 

The operation is recommended as the replacement for the deprecated `Function.Equality.const`, which *did* have a more `Level`-polymorphic type, so I think that this one should as well. 

That said, there do not seem to be any other uses of either operation in `stdlib`, so it's a bit hard to tell what the client consequences of this might be, but... absent telepathy, not much we can do about that!

jamesmckinna (approved):

jamesmckinna:
Thanks @jacquescomeaux for your first PR!

PR #811, opened 2019-06-06
n-ary decidable equality transformer

gallais:
I got tired of writing `uncurry (cong₂ _∷_)` and invoking `_×-dec_` in #799
so here we go.

Issue #803, opened 2019-06-04
'Gradual' decidable equality combinators


gallais:
When you define decidable equality for e.g.

```agda
data Rose (A : Set) : Set where
  node : List (Rose A) → Rose A
```

you can't reuse `Data.List.Properties`' `≡-dec` because it makes the
termination checker scream. However you would not have to redo all the
work if only we provided

```agda
∷-dec : Dec (x ≡ y) → Dec (xs ≡ ys) → Dec (x ∷ xs ≡ y ∷ ys)
∷-dec x≟y xs≟ys = map′ (uncurry (cong₂ _∷_)) ∷-injective (x≟y ×-dec xs≟ys)
```

Indeed mutually defining

```agda
_≟_  : Decidable {A = Rose A} _≡_
_≟s_ : Decidable {A = List (Rose A)} _≡_
```

then becomes easy:

```agda
node xs ≟ node ys = map′ (cong node) node-injective (xs ≟s ys)

[]       ≟s []       = yes refl
(x ∷ xs) ≟s (y ∷ ys) = ∷-dec (x ≟ y) (xs ≟s ys)

[] ≟s (x ∷ ys) = no (λ ())
(x ∷ xs) ≟s [] = no (λ ())
```

I have used this trick twice in #799 to simplify decidable equality proofs
involving `Arg` and `Abs` and there are probably numerous other places where
we can define these combinators in the stdlib.

Edit: [self-contained gist](https://gist.github.com/gallais/22fa12efb9ef42835054d13eaec911c8)

jamesmckinna:
Towards a general solution to `deriving (Eq)`-style automated support?
(I'm told that this is a frequently-solved problem among commercial Agda developers)

JacquesCarette:
'commercial Agda developers' ???

jamesmckinna:
Was this issue effectively put to bed by #811 ? If so, we should close, but perhaps document the pattern in the library design/style guide documentation?

JacquesCarette:
Agreed. Can we even test its use internally, or would that introduce dependencies that are a bit over the top?

jamesmckinna:
Stub content added to `doc/style-guide.md` in #2270, so closing this now.

PR #799, opened 2019-06-03
[ fix #698 ] Safe reflection module

gallais:
Now that Agda internalises all the properties we need, we should be able to
finally turn the `--safe` option on in the `Reflection` module.

PR #2843, opened 2025-10-26
[ fix ] `Relation.Nullary.Decidable.Core` names for combinators

### LINKED ISSUE #2842: Use consistent names for `Decidable` combinators
https://github.com/agda/agda-stdlib/blob/2ce5eb3310d87f3fb757e680603c2c83f456df75/src/Relation/Nullary/Decidable/Core.agda#L91-L122 defines
```agda
T? : ∀ x → Dec (T x)
¬? : Dec A → Dec (¬ A)
⊤-dec : Dec {a} ⊤
_×-dec_ : Dec A → Dec B → Dec (A × B)
⊥-dec : Dec {a} ⊥
_⊎-dec_ : Dec A → Dec B → Dec (A ⊎ B)
_→-dec_ : Dec A → Dec B → Dec (A → B)
```
when, presumably (?), we should be consistent and instead write
```agda
-- Interaction with negation, sum, product etc.

infixr 1 _⊎?_
infixr 2 _×?_ _→?_

T? : ∀ x → Dec (T x)
T? x = x because T-reflects x

¬? : Dec A → Dec (¬ A)
does  (¬? a?) = not (does a?)
proof (¬? a?) = ¬-reflects (proof a?)

⊤? : Dec {a} ⊤
does  ⊤? = true
proof ⊤? = ⊤-reflects

_×?_ : Dec A → Dec B → Dec (A × B)
does  (a? ×? b?) = does a? ∧ does b?
proof (a? ×? b?) = proof a? ×-reflects proof b?

⊥? : Dec {a} ⊥
does  ⊥?  = false
proof ⊥?  = ⊥-reflects

_⊎?_ : Dec A → Dec B → Dec (A ⊎ B)
does  (a? ⊎? b?) = does a? ∨ does b?
proof (a? ⊎? b?) = proof a? ⊎-reflects proof b?

_→?_ : Dec A → Dec B → Dec (A → B)
does  (a? →? b?) = not (does a?) ∨ does b?
proof (a? →? b?) = proof a? →-reflects proof b?
```
and deprecate the old names?

jamesmckinna:
Fixes #2842 

Issues:
* lots of knock-on viscosity
* lots of other occurrences of `-dec` as a suffix which could/should be changed to `?`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,6 +146,17 @@ Additions to existing modules
   ≟-≡          : (eq : i ≡ j) → (i ≟ j) ≡ yes eq
   ≟-≡-refl     : (i : Fin n) → (i ≟ i) ≡ yes refl
   ≟-≢          : (i≢j : i ≢ j) → (i ≟ j) ≡ no i≢j
+  inject-<     : inject j < i
+
+  record Least⟨_⟩ (P : Pred (Fin n) p) : Set p where
+    constructor least
+    field
+      witness : Fin n
+      example : P witness
+      minimal : ∀ {j} → .(j < witness) → ¬ P j
+
+  search-least⟨_⟩  : Decidable P → Π[ ∁ P ] ⊎ Least⟨ P ⟩
+  search-least⟨¬_⟩ : Decidable P → Π[ P ] ⊎ Least⟨ ∁ P ⟩
   ```
 
 * In `Data.Nat.ListAction.Properties`
--- a/src/Data/Fin/Properties.agda
+++ b/src/Data/Fin/Properties.agda
@@ -48,13 +48,13 @@ open import Relation.Binary.PropositionalEquality.Properties as ≡
 open import Relation.Binary.PropositionalEquality as ≡
   using (≡-≟-identity; ≢-≟-identity)
 open import Relation.Nullary.Decidable as Dec
-  using (Dec; _because_; yes; no; _×?_; _⊎?_; map′)
+  using (Dec; _because_; yes; no; _×?_; _⊎?_; map′; decidable-stable)
 open import Relation.Nullary.Negation.Core
   using (¬_; contradiction; contradiction′)
 open import Relation.Nullary.Reflects using (invert)
 open import Relation.Unary as U
-  using (U; Pred; Decidable; _⊆_; Satisfiable; Universal)
-open import Relation.Unary.Properties using (U?)
+  using (U; Pred; Decidable; _⊆_; ∁; Satisfiable; Universal)
+open import Relation.Unary.Properties using (U?; ∁?)
 
 private
   variable
@@ -470,6 +470,10 @@ i<1+i = ℕ.n<1+n ∘ toℕ
 toℕ-inject : ∀ {i : Fin n} (j : Fin′ i) → toℕ (inject j) ≡ toℕ j
 toℕ-inject {i = suc i} zero    = refl
 toℕ-inject {i = suc i} (suc j) = cong suc (toℕ-inject j)
+
+inject-< : ∀ {i : Fin n} (j : Fin′ i) → inject j < i
+inject-< {i = suc i} zero    = z<s
+inject-< {i = suc i} (suc j) = s<s (inject-< j)
 
 ------------------------------------------------------------------------
 -- inject₁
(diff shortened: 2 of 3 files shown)


src/Data/Product/Relation/Binary/Pointwise/NonDependent.agda
```diff
@@ -88,7 +88,7 @@ Pointwise R S (a , c) (b , d) = (R a b) × (S c d)
 ... | inj₂ y₁∼x₁ | inj₁ x₂∼y₂ = inj₁ (sym₁ y₁∼x₁ , x₂∼y₂)
 
 ×-decidable : Decidable R → Decidable S → Decidable (Pointwise R S)
```
gallais, line 90: Should this be called `pointwise?`?

src/Data/Product/Relation/Binary/Lex/Strict.agda
```diff
@@ -111,9 +111,9 @@ module _ {_≈₁_ : Rel A ℓ₁} {_<₁_ : Rel A ℓ₂} {_<₂_ : Rel B ℓ
                 Decidable _<ₗₑₓ_
   ×-decidable dec-≈₁ dec-<₁ dec-≤₂ x y =
```
gallais, line 112: Should this be called `<ₗₑₓ?`?

src/Data/Vec/Properties.agda
```diff
@@ -76,7 +76,7 @@ toList-injective m≡n (x ∷ xs) (y ∷ ys) xs=ys =
 ≡-dec : DecidableEquality A → DecidableEquality (Vec A n)
```
gallais, line 76: Should this be called `≡?`?

src/Relation/Binary/Construct/Intersection.agda
```diff
@@ -87,7 +87,7 @@ module _ (≈ : Rel A ℓ₁) (L : Rel A ℓ₂) (R : Rel A ℓ₃) where
 module _ {L : REL A B ℓ₁} {R : REL A B ℓ₂} where
 
   decidable : Decidable L → Decidable R → Decidable (L ∩ R)
```
gallais, line 89: Should this be called `_∩?_`?

gallais (commented):
Thank you for that. Wondering whether there are a few more opportunities for consistent naming.

jamesmckinna:
> Thank you for that. Wondering whether there are a few more opportunities for consistent naming.

yes, hence the OP reference to "lots of other occurrences of `-dec`"... but I thought it better to handle those downstream.

I already encountered the need for some careful qualified importing in order to avoid name clashes, and I didn't want to get into deeper rabbit holes by widening the impact of this PR. 

~~But happy to do so if reviewers agree, although~~ Suggest we move such things downstream, as I think/worry it might open *many* more cans of worms? The ones you identify arise as direct consequences of this PR, but I fear the general pattern (and its violation) may be even more widespread...

src/Data/Product/Relation/Binary/Pointwise/NonDependent.agda
jamesmckinna: Arguably! But see ~~above~~ below...

src/Data/Product/Relation/Binary/Lex/Strict.agda
jamesmckinna: Ditto.

src/Data/Vec/Properties.agda
jamesmckinna: Ditto. (this one seems especially egregious, given that conventionally, we make the different choice of `_≟_` instead... one reason I wanted to leave this alone, until we could agree a/the 'blessed'/'right' name!) #2845

src/Relation/Binary/Construct/Intersection.agda
jamesmckinna: Indeed.

JacquesCarette:
Wow, I was surprised that you even attempted this as v2.4.

src/Data/Product/Relation/Binary/Lex/Strict.agda
JacquesCarette: I also notice that the type is `Decidable _<₂_` but the argument called `dec-≤₂` which is rather confusing!

jamesmckinna:
> While I agree that there are quite a few more things that ought to be changed, I think that it is fine to do this kind of work incrementally. So what @gallais points out could form the basis of 2+ future PRs. Similarly with the wart that I point out.

See also:
* the `∷-dec` combinator, and friends, introduced in #799 / #803 / #811 
* ... others? (we should make a list on #2842 else as separate issues)

UPDATED this PR closes #2842 so lifted out the additional comments/examples to #2846. Merging now.

jamesmckinna:
Ugh. Forgot to to fix the alignment in `CHANGELOG`.

MatthewDaggitt:
That's okay, there's always a tidying up pass before release.

src/Data/Vec/Properties.agda
jamesmckinna: See #2952 , follow ups to that PR.

src/Data/Product/Relation/Binary/Lex/Strict.agda
jamesmckinna: See comments on #2846 :
* I think it's worth fixing the assumptions name
* but not the combinator name?

src/Relation/Binary/Construct/Intersection.agda
jamesmckinna: This one I *do* think worth doing!?

src/Data/Product/Relation/Binary/Lex/Strict.agda
JacquesCarette: fixing the assumption name is non-breaking, so that's a free one to do.

PR #2844, opened 2025-10-26
Added README for tests directory

MatthewDaggitt:
As requested by the JOSS review.


jamesmckinna (approved):
Looks good to me.

I note, for future reference, that we haven't updated/factored out, the RTS flags for either of `cabal` or `agda`, but I assume that the current (old!) values still work OK, both remotely on GitHub, and locally (I tend to run with a higher `-M` and `-H` value, to reflect the 16GB RAM in my laptop...)

JacquesCarette:
Note that the reviewers are [having a hard time with this](https://github.com/openjournals/joss-reviews/issues/9241#issuecomment-3453058570).

JacquesCarette (approved):

JacquesCarette:
I'm guessing the issue might be that the reviewer did `make testsuite` and did not pass the other arguments?

Issue #2848, opened 2025-10-28
v2.3 webpage is not quite v2.3


shhyou:
## Expected

Version 2.3 library in the webpage: http://agda.github.io/agda-stdlib/v2.3/README.html

## Actual

It appears to be some post-v2.3 version:

(Link to GitHub: https://github.com/agda/agda-stdlib/blob/gh-pages/v2.3/README.html)

```
------------------------------------------------------------------------
-- The Agda standard library, version 2.4-dev
```

Perhaps some version between 2.3 and HEAD is accidentally deployed.

jamesmckinna:
Hmmm... the version of `doc/README.Agda` on [the v2.3 release branch](https://github.com/agda/agda-stdlib/blob/v2.3-release/doc/README.agda) seems to correct, but the generated pages seem to be out of sync. 

This may need further investigation! 

But a useful prompt ahead of the 'upcoming' v2.4 release in any case...

shhyou:
Given https://github.com/agda/agda-stdlib/commit/6b8bbb7bd10d12f241037e696eb55eeefa177a67 and https://github.com/agda/agda-stdlib/commit/f58379c834c82f243dadc6e9f7c713502801a71f, I suspect the per-version HTML pages are deployed manually.

shhyou:
Here:
https://github.com/agda/agda-stdlib/blob/dd64a3a1b5d73153db61cb68936652c9d6a4dc9f/doc/release-guide.txt#L63-L75

shhyou:
Aha, figured how to send PR to a non-primary branch. And it turns out that I almost forgot to switch to the `v2.3-release` tag before running `agda -i. -idoc -isrc --html index.agda` from above.

JacquesCarette:
One 'approve' from me on that PR. Nicely spotted!

PR #2849, opened 2025-10-29
[ release ] Re-generate doc of version 2.3

shhyou:
The current v2.3 doc is ahead of the v2.3 release, causing it to be slightly out-of-sync, e.g., `Data.Fin.Properties` shows extra, newer definitions and `README` refers to 2.4-dev.

This commit re-generates the doc of agda-stdlib 2.3. Closes #2848.

PR #2850, opened 2025-10-29
Add two steps to Release Guide

shhyou:
* Make sure to create a (final) release branch in addition to creating tags.
* Before generating the docs, make sure the directory contains the release branch (related to #2848).

### DIFF
--- a/doc/release-guide.txt
+++ b/doc/release-guide.txt
@@ -47,6 +47,11 @@ procedure should be followed:
 
     git push --follow-tags
 
+* Create a branch for the release (since tags can't be branches)
+
+    git switch -c v${VERSION}-release
+    git push origin v${VERSION}-release
+
 * Make a new release on Github at https://github.com/agda/agda-stdlib/releases
 
 * Submit a pull request to update the version of standard library on Homebrew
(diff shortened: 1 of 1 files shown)


doc/release-guide.txt
```diff
@@ -47,6 +47,11 @@ procedure should be followed:
 
     git push --follow-tags
 
+* Create a branch for the release (since tags can't be branches)
+
+    git switch -c v${VERSION}-release
+    git push origin v${VERSION}-release
```
jamesmckinna, line 53: Is this still (canonically) called `origin`, or should it be abstracted out as (the value of) a shell variable?

(I think in my setup, it's called something else, but that may be mere idiosyncrasy...)

jamesmckinna:
Thanks for posting a fix to #2848 on #2849 and for the defensive fix here. 
I think as perhaps the source of the original bug, I'd better stand back from reviewing this!

doc/release-guide.txt
shhyou: Yes, by default `origin` refers to where the local repository is cloned from. If one directly clones `https://github.com/agda/agda-stdlib.git` then `origin` is `agda/agda-stdlib`. And yes, when working with forks, if one originally cloned the fork then it is possible to add `agda/agda-stdlib` as a remote other than `origin` (like `upstream`).

Given that the earlier steps in Release Guide do not specify the remote, I hope `origin` would be an okay choice.

PR #2765, opened 2025-07-12
Ring reasoning

JacquesCarette:
Reasoning combinators for monoid, renamed and adapted for Ring. Builds on #2692 .

PR #2851, opened 2025-10-29
Algebra property reexports

Taneb:
* Note that a `RingWithoutOne` is a `SemiringWithoutOne`
* Introduce new module `Algebra.Properties.SemiringWithoutOne`
* Re-export many algebraic properties (appropriately renamed) into the property files for larger algebraic structures, including all the new semigroup properties for semirings, both for addition and multiplication

### DIFF
--- a/src/Algebra/Properties/Ring.agda
+++ b/src/Algebra/Properties/Ring.agda
@@ -14,8 +14,26 @@ open Ring ring
 open import Relation.Binary.Reasoning.Setoid setoid
 
 ------------------------------------------------------------------------
--- Export properties of rings without a 1#.
+-- Export properties of semirings and rings without a 1#.
 
+open import Algebra.Properties.Semiring semiring public
+  using
+    ( 1#-unique
+    ; 1#-comm
+
+    ; *-elimʳ
+    ; *-elimˡ
+    ; *-introʳ
+    ; *-introˡ
+    ; *-introᶜ
+
+    ; *-cancelʳ
+    ; *-cancelˡ
+    ; *-insertˡ
+    ; *-insertʳ
+    ; *-cancelᶜ
+    ; *-insertᶜ
+    )
 open import Algebra.Properties.RingWithoutOne ringWithoutOne public
 
 ------------------------------------------------------------------------
--- a/src/Algebra/Properties/Semiring.agda
+++ b/src/Algebra/Properties/Semiring.agda
@@ -13,13 +13,24 @@ module Algebra.Properties.Semiring
   {c ℓ} (semiring : Semiring c ℓ) where
 
 open Semiring semiring
-import Algebra.Consequences.Setoid setoid as Consequences
-open import Relation.Binary.Reasoning.Setoid setoid
 
-------------------------------------------------------------------------
--- Re-export binomial expansion
+open import Algebra.Properties.SemiringWithoutOne semiringWithoutOne public
+open import Algebra.Properties.Monoid *-monoid public
+  using ()
+  renaming
+    ( ε-unique to 1#-unique
+    ; ε-comm to 1#-comm
 
-binomial-expansion : ∀ w x y z →
-                     (w + x) * (y + z) ≈ w * y + w * z + x * y + x * z
-binomial-expansion = Consequences.binomial-expansion +-cong +-assoc distrib
+    ; elimʳ to *-elimʳ
+    ; elimˡ to *-elimˡ
+    ; introʳ to *-introʳ
+    ; introˡ to *-introˡ
+    ; introᶜ to *-introᶜ
 
+    ; cancelʳ to *-cancelʳ
+    ; cancelˡ to *-cancelˡ
+    ; insertˡ to *-insertˡ
+    ; insertʳ to *-insertʳ
+    ; cancelᶜ to *-cancelᶜ
+    ; insertᶜ to *-insertᶜ
+    )
--- a/src/Algebra/Properties/SemiringWithoutOne.agda
+++ b/src/Algebra/Properties/SemiringWithoutOne.agda
@@ -1,0 +1,128 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Some basic properties of Semirings without a multiplicative identity
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (SemiringWithoutOne)
+
+module Algebra.Properties.SemiringWithoutOne {c ℓ} (semiring : SemiringWithoutOne c ℓ) where
+
+open SemiringWithoutOne semiring
+
+import Algebra.Consequences.Setoid setoid as Consequences
+open import Algebra.Properties.CommutativeMonoid (+-commutativeMonoid) public
+  using ()
+  renaming
+    ( x∙yz≈xy∙z to x+[y+z]≈[x+y]+z
+    ; alternativeˡ to +-alternativeˡ
+    ; alternativeʳ to +-alternativeʳ
+    ; alternative to +-alternative
+    ; flexible to +-flexible
+
+    ; uv≈w⇒xu∙v≈xw to u+v≈w⇒[x+u]+v≈x+w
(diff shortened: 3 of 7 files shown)


Taneb:
Ah - this is failing, in part because I assumed that what I was calling `+-cancelʳ` (specialization of `Properties.Semigroup.cancelʳ` to addition) matches the preexisting `+-cancelʳ` (addition is right cancellative). There needs to be some rethinking of names here.

jamesmckinna:
> Ah - this is failing, in part because I assumed that what I was calling `+-cancelʳ` (specialization of `Properties.Semigroup.cancelʳ` to addition) matches the preexisting `+-cancelʳ` (addition is right cancellative). There needs to be some rethinking of names here.

~~See also #2654 ... I think there this is (yet another) instance of systematic clash of different authors' mental models of what the qualifiers denote... ?~~

Er, no. It's because we accepted the `Algebra.Properties.Monoid.cancel*` in #2692 without thinking through the consequences wrt `Cancellative` operations. Sigh.

src/Algebra/Properties/SemiringWithoutOne.agda
```diff
@@ -0,0 +1,128 @@
...
+    ; ε-unique to 0#-unique
+    
+    ; elimʳ to +-elimʳ
+    ; elimˡ to +-elimˡ
+    ; introʳ to +-introʳ
+    ; introˡ to +-introˡ
+    ; introᶜ to +-introᶜ
+
+    ; cancelʳ to +-cancelʳ
+    ; cancelˡ to +-cancelˡ
```
jamesmckinna, line 88: Suggest, in view of the clash between naming of `cancel` properties, that you perhaps make these
```suggestion
    ; cancelʳ to cancelʳ-+
    ; cancelˡ to cancelˡ-+
```
ahead of trying to figure out a solution to his mess. 

Shoutout to @JacquesCarette for input on this (the names come from `agda-categories` via Jacques and his students, I think).

src/Algebra/Properties/SemiringWithoutOne.agda
```diff
@@ -0,0 +1,128 @@
...
+    ; insertˡ to +-insertˡ
+    ; insertʳ to +-insertʳ
+    ; cancelᶜ to +-cancelᶜ
```
jamesmckinna, line 91:
```suggestion
    ; cancelᶜ to cancelᶜ-+
```
for consistency!?

src/Algebra/Properties/Semiring.agda
```diff
@@ -13,13 +13,24 @@ module Algebra.Properties.Semiring
...
-                     (w + x) * (y + z) ≈ w * y + w * z + x * y + x * z
-binomial-expansion = Consequences.binomial-expansion +-cong +-assoc distrib
+    ; elimʳ to *-elimʳ
+    ; elimˡ to *-elimˡ
+    ; introʳ to *-introʳ
+    ; introˡ to *-introˡ
+    ; introᶜ to *-introᶜ
 
+    ; cancelʳ to *-cancelʳ
+    ; cancelˡ to *-cancelˡ
```
jamesmckinna, line 31: See below for the additive versions.

src/Algebra/Properties/Semiring.agda
```diff
@@ -13,13 +13,24 @@ module Algebra.Properties.Semiring
...
+    ; insertˡ to *-insertˡ
+    ; insertʳ to *-insertʳ
+    ; cancelᶜ to *-cancelᶜ
```
jamesmckinna, line 34: Ditto

src/Algebra/Properties/Ring.agda
```diff
@@ -14,8 +14,26 @@ open Ring ring
...
+    ; 1#-comm
+
+    ; *-elimʳ
+    ; *-elimˡ
+    ; *-introʳ
+    ; *-introˡ
+    ; *-introᶜ
+
+    ; *-cancelʳ
+    ; *-cancelˡ
```
jamesmckinna, line 31: move to a `renaming` block?

src/Algebra/Properties/Ring.agda
```diff
@@ -14,8 +14,26 @@ open Ring ring
...
+    ; *-insertˡ
+    ; *-insertʳ
+    ; *-cancelᶜ
```
jamesmckinna, line 34: ditto.?

jamesmckinna (changes requested):
I've made (frankly unsatisfactory) suggestions as to how to rename things, otherwise I'd simply approve. This all looks like useful refactoring/filling in the hierarchy, but we do need to do something about these `cancel` lemmas...

jamesmckinna:
Actually, I think it  *might* be easier to simply *not* export the 'offending' names, on the basis that 
* this should guarantee that testing succeeds (?)
* that this doesn't commit to 'nonsense' names ahead of any refactoring/renaming in `Algebra.Properties.Monoid`

Thoughts?

JacquesCarette (commented):
In theory, this is good. In practice, clearly there are all sorts of problems. Which require non-trivial thinking, which I can't do right now. But will come back to this when I can.

jamesmckinna:
> In theory, this is good. In practice, clearly there are all sorts of problems. Which require non-trivial thinking, which I can't do right now. But will come back to this when I can.

Hope you manage to find space/time/brainspace to do this... you have far-reaching insights into how to improve `Algebra.*` general, and `stdlib` needs them! #2765 etc.

My current thinking about the `cancel*` properties is that they be renamed to `inverse⇒cancel*` (I know you don't like long names, but such 'implicational' forms do, at least, reflect the logical structure), and they certainly could be, locally here, in order to not obstruct progress? See #2885 for specimen usage.

src/Algebra/Properties/Ring.agda
jamesmckinna: See the strategy employed in #2885 for dealing with this.

JacquesCarette:
I'm not too adverse to `inverse⇒cancel*`. It has a degree of "self documentation" that I like, despite its verbosity.

jamesmckinna:
@Taneb do you want to try to refactor this (via `renaming`) for v2.4, or mark it as blocking on a PR to rename the `Algebra.Properties.Monoid.cancel*` lemmas?

jamesmckinna:
@Taneb @JacquesCarette what are the interactions between this PR and #2765 ?
What chance of clearing these from the v2.4 docket sooner rather than later?

One reason to do so is that if we (as already agreed) plan `breaking` changes for v3.0, we could consider any suitable refactoring/renaming there?

JacquesCarette:
I'm going to try to refocus some of my coding urges back to stdlib. So there's a hope!

jamesmckinna:
I think it would be good to try to achieve consensus on names, so that this can go through without too much more pain.

MatthewDaggitt (approved):

jamesmckinna:
@JacquesCarette @Taneb what's our path forward on this and #2765 ?

jamesmckinna:
I fixed the whitespace (at least) in order to force re-checking, as the old failing logs were no longer accessible.

And now... a failure arising from `Algebra.Properties.BooleanRing`? @Taneb Suggest rebasing and proceeding until a 'real' error occurs? This PR seems in a horribly inconsistent state wrt current `master`...?

JacquesCarette:
Hopefully we can make progress on this via asynchronous discussion on here. I do have some brainspace for this. But I will need some help to re-figure out what the issues were!

jamesmckinna:
> Hopefully we can make progress on this via asynchronous discussion on here. I do have some brainspace for this. But I will need some help to re-figure out what the issues were!

@JacquesCarette My recall is that:
* `Algebra.Properties.Monoid` added suboptimal names in #2692 
* which then clash with names introduced here
* ... PR stalls... problematic names don't get fixed... opinions differ...
* PR bit rots

The issue: we need to find new/better names for the `cancel` properties introduced in #2692 . I've previously made suggestions, either as shims to permit progress on this PR, or more permanently as back-patches for `Algebra.Properties.Monoid`. But without some consensus, progress is blocked!

Separately, some remedial work (eg rebasing) is also necessary to get past the bitrot aspects...

Issue #2472, opened 2024-09-03
Find all single letter publicly-exported modules and make sure they are private


MatthewDaggitt:
Currently we have many single letter modules that are unintentionally being exported publicly by other modules. In v3.0 we should do a sweep and make them all private (or rename them to something more meaningful if we really do want to export them!).

MatthewDaggitt:
See https://github.com/agda/agda-stdlib/pull/2407#discussion_r1735449110

jamesmckinna:
Ha!/Oops! :man_facepalming: 

The final merged version of #2407 in fact *does* make `R` a `private` module definition now... so we have just merged a `breaking` change... sorry about that! ...
... but hopefully a harmless one (given that the `Algebra.Solver.*` machinery appears to have been superseded by `Tactic.*Solver`, but without being deprecated... so perhaps there are clients uses out there, hopefully they don't explicitly refer to `R`!?)

If anyone reports `breaking` behaviour, I'll post a follow-up PR, but perhaps better to let sleeping dogs lie? Let me know!

MatthewDaggitt:
> A quick pass via grep suggests that the Solver export might have been the only such instance? Additional eyes welcome on this...

That is a little surprising. I definitely remember having seen the pattern used in `Data.List.Relation` hierarchy, but given that this is a  v3.0 change, I'm not going to spend too long digging now. We should just bear it in mind when writing new code for the moment.

jamesmckinna:
No need to dig now, but we could make a task list here?

jamesmckinna:
Haven't found any on my travels recently, but we *should* have a note in `style-guide` about this!

JacquesCarette:
```
rg -e "[^a-zA-Z]module [a-zA-Z] "
```
where 'rg' is 'ripgrep' does find quite a lot of stuff. I'm quite sure a number of those would be unintentionally part of that module's interface. There will also be a lot of false positives in there.

PR #2852, opened 2025-10-29
Add subgroups and submodules

Taneb:
Pulled out of work on #2729 that can be merged straight away hopefully

### DIFF
--- a/src/Algebra/Construct/Sub/Group.agda
+++ b/src/Algebra/Construct/Sub/Group.agda
@@ -10,13 +10,13 @@ open import Algebra.Bundles using (Group; RawGroup)
 
 module Algebra.Construct.Sub.Group {c ℓ} (G : Group c ℓ) where
 
-private
-  module G = Group G
-
 open import Algebra.Structures using (IsGroup)
 open import Algebra.Morphism.Structures using (IsGroupMonomorphism)
 import Algebra.Morphism.GroupMonomorphism as GroupMonomorphism
 open import Level using (suc; _⊔_)
+
+private
+  module G = Group G
 
 record Subgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
   field
--- a/src/Algebra/Module/Construct/Sub/Bimodule.agda
+++ b/src/Algebra/Module/Construct/Sub/Bimodule.agda
@@ -9,7 +9,14 @@
 open import Algebra.Bundles using (Ring)
 open import Algebra.Module.Bundles using (Bimodule; RawBimodule)
 
-module Algebra.Module.Construct.Sub.Bimodule {cr ℓr cs ℓs cm ℓm} {R : Ring cr ℓr} {S : Ring cs ℓs}  (M : Bimodule R S cm ℓm) where
+module Algebra.Module.Construct.Sub.Bimodule
+  {cr ℓr cs ℓs cm ℓm} {R : Ring cr ℓr} {S : Ring cs ℓs}  (M : Bimodule R S cm ℓm)
+  where
+
+open import Algebra.Module.Structures using (IsBimodule)
+open import Algebra.Module.Morphism.Structures using (IsBimoduleMonomorphism)
+import Algebra.Module.Morphism.BimoduleMonomorphism as BimoduleMonomorphism
+open import Level using (suc; _⊔_)
 
 private
   module R = Ring R
@@ -17,10 +24,6 @@ private
   module M = Bimodule M
 
 open import Algebra.Construct.Sub.Group M.+ᴹ-group
-open import Algebra.Module.Structures using (IsBimodule)
-open import Algebra.Module.Morphism.Structures using (IsBimoduleMonomorphism)
-import Algebra.Module.Morphism.BimoduleMonomorphism as BimoduleMonomorphism
-open import Level using (suc; _⊔_)
 
 record Subbimodule cm′ ℓm′ : Set (cr ⊔ cs ⊔ cm ⊔ ℓm ⊔ suc (cm′ ⊔ ℓm′)) where
   field


src/Algebra/Construct/Sub/Group.agda
```diff
@@ -0,0 +1,40 @@
...
+  module G = Group G
+
+open import Algebra.Structures using (IsGroup)
+open import Algebra.Morphism.Structures using (IsGroupMonomorphism)
+import Algebra.Morphism.GroupMonomorphism as GroupMonomorphism
+open import Level using (suc; _⊔_)
+
+record Subgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    Sub : RawGroup c′ ℓ′
```
JacquesCarette: I think the name `Algebra.Construct.Sub.*` is fine, even good. I think the name `Sub` as a field name is not.

`group` ? Because it is a group, it has an independent existing as a group, and acquires a new one when inserted into this record.

src/Algebra/Module/Construct/Sub/Module.agda
```diff
@@ -0,0 +1,50 @@
...
+
+open import Algebra.Construct.Sub.Group M.+ᴹ-group
+open import Algebra.Module.Structures using (IsModule)
+open import Algebra.Module.Morphism.Structures using (IsModuleMonomorphism)
+import Algebra.Module.Morphism.ModuleMonomorphism as ModuleMonomorphism
+open import Level using (suc; _⊔_)
+
+record Submodule cm′ ℓm′ : Set (c ⊔ cm ⊔ ℓm ⊔ suc (cm′ ⊔ ℓm′)) where
+  field
+    Sub : RawModule R.Carrier cm′ ℓm′
```
JacquesCarette: in a similar vein, I would want `module`, but that can't work. `moduleOn`, since it is parametrized?

src/Algebra/Construct/Sub/Group.agda
Taneb: I'm willing to change the name but I don't like the name `group` - there's already another group lurking in the background and I'm worried they'll be confused

JacquesCarette: Fair. I'll try to come up with other suggestions. (I did not propose `rawgroup` because I dislike that quite a bit myself.)

jamesmckinna:
I agree with @JacquesCarette that `Sub` is a bad name, but earlier I came up short trying think of a better one. But let's try harder!

It occurs to me that, in one sense, the name *cannot* matter, as we could insist that it only ever be accessed as the underlying (computed) `rawGroup` manifest field of the `group` manifest field of `Subgroup`, but that may be too much cognitive gymnastics for most readers/users...

... I've argued, persuasively or not, for 'canonicalising' the name `\iota` for the (name of the) monomorphism defining the subobject, so it would be good to have a canonical name for its domain (and not to be fussy about standardising apart each such name according to what algebraic signature/structure it is intended to support; that's handled by the manifest field names)... so how about:
* `domain` (I'm largely indifferent to capitalisation, but I can see `Domain` as perhaps preferable)
* `subobject` (a marginal improvement on `Sub`? ditto. re capitalisation)
* ... `I` (this might cause trouble!? but hints we should take care when referring to it... and we definitely need capitalisation here!)
* `Iota` (or even `\Iota`!?)

I realise this is a vexed topic, and I want to stress the last two are not intended vexatiously (nor facetiously); we need a name for a thing which shouldn't really ever be referred to directly outwith the `record` declaration. Choosing a 'jarring' field name might be a way to achieve that?

Avoiding prior convention, Freyd/Sčedrov in their Categories/Allegories book, used prefix and suffix 'box' on (names of arrows) to denote 'domain' and 'codomain' (with moreover the advantage of being minimum-ink), so `\box\iota` could work, I suppose. 

Early morning thoughts of a chronic insomniac!

Taneb:
`domain` was something I was thinking of suggesting.

Another suggestion: `H` (for group) or `N` (for module), i.e. the second dummy variable name for the structure

jamesmckinna:
> `domain` was something I was thinking of suggesting.

OK, good to maybe achieve some convergence on this?

> Another suggestion: `H` (for group) or `N` (for module), i.e. the second dummy variable name for the structure

yes, that would be very much in the spirit of old-school typographical convention in mathematics, 

(sidebar: did we actually get this from mathematicians, or from the extraordinarily gifted typesetters/printers at, particularly, Oxford University Press, Springer, and Van Nostrand Reinhold, who left such a mark on *published* Anglophone/German mathematics in the pre-digital, pre- frantic M&A consolidation years ie up to about 1980?)

but as @JacquesCarette often points out, we need to move past such orthodoxies, and if we're going to keep the same name for the monomorphism, then we need to keep the same name for its domain? (eyes on a generative solution downstream, cf. #2287 etc.)

src/Algebra/Module/Construct/Sub/Module.agda
```diff
@@ -0,0 +1,50 @@
...
+  ⟨module⟩ = record { isModule = isModule }
+
+  open Module ⟨module⟩ public hiding (isModule)
+
+  subgroup : Subgroup cm′ ℓm′
+  subgroup = record
+    { Sub = Sub.+ᴹ-rawGroup
+    ; ι = ι
+    ; ι-monomorphism = ι.+ᴹ-isGroupMonomorphism
+    }
```
jamesmckinna: In the spirit of not explicitly naming things which don't need naming... surely
```suggestion
    { 
    ; ι-monomorphism = ι.+ᴹ-isGroupMonomorphism
    }
```
suffices here?

src/Algebra/Construct/Sub/Group.agda
Taneb: Looking what I actually wrote, I use `group` for the embellishment of `Sub` with laws

Taneb:
I've changed the name from `Sub` to `domain`, but the name of the (private) _module_ `Sub` to `H`/`N`

jamesmckinna:
This may be the sweet spot... the internal module names are `private`, so the one-character name doesn't violate #2472 ... while the potentially-exportable  name `domain` is consistently repeatable and sufficiently distinct ... looks good to me!

src/Algebra/Module/Construct/Sub/Module.agda
```diff
@@ -0,0 +1,46 @@
...
+--
+-- Definition of submodules
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (CommutativeRing)
+open import Algebra.Module.Bundles using (Module; RawModule)
+
+module Algebra.Module.Construct.Sub.Module {c ℓ cm ℓm} {R : CommutativeRing c ℓ} (M : Module R cm ℓm) where
```
Taneb: I've realised that later on in my development of modular arithmetic, I don't want sub _modules_ per se, but sub R-R-_bimodules_ (I don't want to assume that `R` is commutative). Or possibly sub left (eq. right) modules.

jamesmckinna: Hmmm... but that could be added downstream, or else (easily?) as part of this one?

jamesmckinna:
What's the reason for *not* including the versions also for `Module`? Eventually we'll want all of these things?

Taneb:
> What's the reason for _not_ including the versions also for `Module`? Eventually we'll want all of these things?

The principle of "only adding only what I need immediately". I worry that if this becomes the "add every possible sub-whatever" PR it'll take about a month to get done

jamesmckinna:
> > What's the reason for _not_ including the versions also for `Module`? Eventually we'll want all of these things?
> 
> The principle of "only adding only what I need immediately". I worry that if this becomes the "add every possible sub-whatever" PR it'll take about a month to get done

Absolutely fair enough. I was only surpised that given you did have working code for the other case, it might have been easier to leave it in. Definitely not asking for the full tapestry of subXYZmodules!

jamesmckinna:
Happy to merge now, but I'd be happier still if the module imports/private blocks were tidied up in the spirit of the style-guide... but I think we can nitpick all of that downstream in round (n+1) of library refactoring...

Taneb:
I'd rather fix the style points before this gets merged. I should be able to do that today

src/Algebra/Construct/Sub/Group.agda
```diff
@@ -0,0 +1,40 @@
...
+
+module Algebra.Construct.Sub.Group {c ℓ} (G : Group c ℓ) where
+
+private
+  module G = Group G
+
+open import Algebra.Structures using (IsGroup)
+open import Algebra.Morphism.Structures using (IsGroupMonomorphism)
+import Algebra.Morphism.GroupMonomorphism as GroupMonomorphism
+open import Level using (suc; _⊔_)
```
jamesmckinna:
```suggestion
open import Algebra.Structures using (IsGroup)
open import Algebra.Morphism.Structures using (IsGroupMonomorphism)
import Algebra.Morphism.GroupMonomorphism as GroupMonomorphism
open import Level using (suc; _⊔_)

private
  module G = Group G
```

src/Algebra/Module/Construct/Sub/Bimodule.agda
```diff
@@ -0,0 +1,47 @@
...
+--
+-- Definition of submodules
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Algebra.Bundles using (Ring)
+open import Algebra.Module.Bundles using (Bimodule; RawBimodule)
+
+module Algebra.Module.Construct.Sub.Bimodule {cr ℓr cs ℓs cm ℓm} {R : Ring cr ℓr} {S : Ring cs ℓs}  (M : Bimodule R S cm ℓm) where
```
jamesmckinna:
```suggestion
module Algebra.Module.Construct.Sub.Bimodule
  {cr ℓr cs ℓs cm ℓm} {R : Ring cr ℓr} {S : Ring cs ℓs}  (M : Bimodule R S cm ℓm)
  where
```

src/Algebra/Module/Construct/Sub/Bimodule.agda
```diff
@@ -0,0 +1,47 @@
...
+private
+  module R = Ring R
+  module S = Ring S
+  module M = Bimodule M
+
+open import Algebra.Construct.Sub.Group M.+ᴹ-group
+open import Algebra.Module.Structures using (IsBimodule)
+open import Algebra.Module.Morphism.Structures using (IsBimoduleMonomorphism)
+import Algebra.Module.Morphism.BimoduleMonomorphism as BimoduleMonomorphism
+open import Level using (suc; _⊔_)
```
jamesmckinna:
```suggestion
open import Algebra.Module.Structures using (IsBimodule)
open import Algebra.Module.Morphism.Structures using (IsBimoduleMonomorphism)
import Algebra.Module.Morphism.BimoduleMonomorphism as BimoduleMonomorphism
open import Level using (suc; _⊔_)

private
  module R = Ring R
  module S = Ring S
  module M = Bimodule M

open import Algebra.Construct.Sub.Group M.+ᴹ-group
```

src/Algebra/Module/Construct/Sub/Bimodule.agda
```diff
@@ -0,0 +1,47 @@
...
+  private
+    module N = RawBimodule domain
+
+  field
+    ι : N.Carrierᴹ → M.Carrierᴹ
+    ι-monomorphism : IsBimoduleMonomorphism domain M.rawBimodule ι
+
+  module ι = IsBimoduleMonomorphism ι-monomorphism
+
+  isBimodule : IsBimodule R S N._≈ᴹ_ N._+ᴹ_ N.0ᴹ N.-ᴹ_ N._*ₗ_ N._*ᵣ_
```
jamesmckinna, line 41: Comment (no action required):
if I ever get my way with #2252 then this would simplify to
```agda
  isBimodule : IsBimodule R S domain
```
which... I would see as a win ;-)

src/Algebra/Module/Construct/Sub/Bimodule.agda
Taneb: Styleguide says if you're splitting the module declaration, `where` should be on its own line

jamesmckinna: touch'e!

jamesmckinna:
I (still; further) don't want to get in the way of this PR, but I think that given that the `Algebra.Module.*` definitions all assume *commutativity* of the underlying additive structure on which the `*Ring` multiplication variously acts, that this could usefully be refactored once #2854 gets merged so that:
* `Subgroup`s of `AbelianGroup`s are automatically `Normal` (moreover directly, not going via `abelian⇒subgroup-normal`)
* hence, `Sub*module`s (of `Ring`-based `*module`s; perhaps there's a story to tell about `Semi` such things, with an underlying `CommutativeMonoid`?) are automatically `Normal` 
* and, `Quotient`s of `AbelianGroup`s need only take a `Subgroup` as second argument to the construction
* hence `Ideal`s are automatically `Normal`, and thus give rise to suitable quotients, without having to relitigate the abelian quotient structure
* ... etc.

Accordingly, it's tempting to suggest adding `Algebra.Construct.Sub.Group`, `Algebra.NormalSubgroup`, and `Algebra.Construct.Sub.AbelianGroup` first, and only then add `Subbimodule` and `Algebra.Ideal`... but as I say, this could be a downstream refactoring instead... cf. https://github.com/agda/agda-stdlib/pull/2854#discussion_r2489296090

JacquesCarette (approved):
still looking good to me.

jamesmckinna:
Still looking good, and it would be good to merge, but what's happening with the cosmetic fix-ups @Taneb ?

jamesmckinna:
Looks good to me!

Issue #2847, opened 2025-10-28
Simple Bool-contradiction lemma?


andreasabel:
Would there be a place for this simple lemma in the standard library?
```agda
Bool-contradiction : ∀ {ℓ} {A : Set ℓ} {b : Bool} → b ≡ true → b ≡ false → A
Bool-contradiction refl ()
```
I found it convenient when refuting impossible cases concerning functions into `Bool`.

jamesmckinna:
In the interests of non-duplication/avoiding redundancy, you could instead write
```agda
open import Data.Bool.Properties using (not-¬)

Bool-contradiction : ∀ {ℓ} {A : Set ℓ} {b : Bool} → b ≡ true → b ≡ false → A
Bool-contradiction ≡tt ≡ff with () ← not-¬ ≡tt ≡ff
```
(or inline `with () ← not-¬ ≡tt ≡ff` where appropriate in your code?)

But not sure whether that's *really* the answer to your question?

andreasabel:
My question is whether it is within the policy to add a new lemma
```agda
Bool-contradiction : ∀ {ℓ} {A : Set ℓ} {b : Bool} → b ≡ true → b ≡ false → A
```
to `Data.Bool.Properties`.
Of course, this lemma can be proven in many ways, and everything than can be proven with the lemma can be proven without it in some other way, but it strikes me as a natural lemma to look for when working with boolean functions.

E.g. here is a proof that the unit type is not isomorphic to the booleans (not using the Iso definition from the standard library):
```agda
record _≅_ (A B : Set) : Set where
  constructor iso
  field
    raw : A ⇔ B
  open _⇔_ raw public
  field
    from∘to : ∀{a} → from (to a) ≡ a
    to∘from : ∀{b} → to (from b) ≡ b

¬Unit≅Bool : ¬(⊤ ≅ Bool)
¬Unit≅Bool (iso _ _ _ to∘from) =
  Bool-contradiction (to∘from {b = true}) (to∘from {b = false})
```
Here, we basically want to combine the two equations `f record{} ≡ true` and `f record{} ≡ false`.

Here is another example proving that minimization is deterministic:
```agda
  Mu : (f : ℕ → Bool) → ℕ → Set
  Mu f zero    = f zero ≡ true
  Mu f (suc n) = f zero ≡ false × Mu (f ∘ suc) n

  Mu-functional :  ∀ n m → Mu f n → Mu f m → n ≡ m
[...]

PR #2853, opened 2025-10-30
[ add ] `Data.Bool.contradiction` fixing #2847

jamesmckinna:
NB.
* use of `Whatever` as conventional wrt *ex falso*-style typing in `Data.Empty` etc.
* deliberate (but questionable?) use of name `contradiction`, avoiding the OP's redundant `Bool-` prefix (potentially already invoked in any usage context via qualified `import Data.Bool as Bool`, if needed, to avoid name clash with `Relation.Nullary.Negation.Core.contradiction`)

Not really to my taste, but offered as a challenge to my own stubbornness on such issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,11 @@ Additions to existing modules
   ```
   NB. the latter is based on `IsCommutativeRing`, with the former on `IsSemiring`.
 
+* In `Data.Bool`:
+  ```agda
+  contradiction : b ≡ true → b ≡ false → Whatever
+  ```
+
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
I would have bet a large amount of money that this name would cause a lot of clashes with the original `contradiction`, but I clearly would have lost it all...

JacquesCarette (approved):

jamesmckinna:
@andreasabel 's latest comment on #2847 nudges me towards abandoning this.
As a first (more hesitant) step, I'll just move it to DRAFT... and remove it form the v2.4 milestone.

Issue #1966, opened 2023-05-10
Kernels and images in `Algebra`


jamesmckinna:
Discussing @Taneb 's [most recent comments on PR #1767 ](https://github.com/agda/agda-stdlib/pull/1767#issuecomment-1542438586), it is clear that we are (also!) missing some fairly crucial definitions from the `Algebra` library/hierarchy: ie given a `XHomomorphism h`, define
* *kernel*, ie via the property ` InKernel = λ x → h x ≈ 0` (suitably interpreted)
* *image*, ie via the property ` InImage = λ y → ∃ x → h x ≈ y`

together with their evident closure properties as substructures, cf. issue #1899 . 

Note also that while we have definitions of structures (but not yet bundles! #1960) defining `IsXMonomorphism` and `IsXIsomorphism` based on `Function.Definitions.Injective` and `Function.Definitions.Core2.Surjective`, these might better (best?) be rephrased/proved equivalent to, properties involving the kernel and image, eg. 
* `IsXMonomorphism h` iff `∀ {x} → InKernel h x → x ≈ 0`, etc.

jamesmckinna:
30 of what, exactly? And which 12 were implemented?

jamesmckinna:
Regarding whether this is a "dangerous rabbit hole", I've already argued (#1899 ) that we *at least* need these two concepts, and hence the characterisation of mono/surjective homomorphisms by the corresponding (short) exact sequences. 

What is unsatisfying about the properties in #1767 is that they are not phrased in such terms, but without their being already present in the library, it's perhaps easy to see how such mathematically less 'efficient' formulations can come about.

JacquesCarette:
30 constructions from the universal algebra literature (viewed very broadly). A quick overview on slide 21 of [Yasmine Sharoda's PhD defense slides](https://github.com/ysharoda/PhD-Thesis/blob/master/defence/slides.pdf).  Some of them are detailed in Chapter 3 of [her thesis](https://github.com/ysharoda/PhD-Thesis); chapter 9 lists the ones that are implemented. (I may have had the numbers a tiny bit off, memory is funny that way).

jamesmckinna:
Apologies for not taking the trouble to go back to Yasmine's thesis for chapter-and verse... 
[...]

PR #2854, opened 2025-10-31
Normal subgroups and quotient groups

Taneb:
Builds off #2852, continuing towards #2729 in bitesize chunks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,9 +79,13 @@ Deprecated names
 New modules
 -----------
 
+* `Algebra.Cosntruct.Quotient.Group` for the definition of quotient groups.
+
 * `Algebra.Construct.Sub.Group` for the definition of subgroups.
 
 * `Algebra.Module.Construct.Sub.Bimodule` for the definition of subbimodules.
+
+* `Algebra.NormalSubgroup` for the definition of normal subgroups.
 
 * `Algebra.Properties.BooleanRing`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@ Deprecated names
 New modules
 -----------
 
-* `Algebra.Cosntruct.Quotient.Group` for the definition of quotient groups.
+* `Algebra.Construct.Quotient.Group` for the definition of quotient groups.
 
 * `Algebra.Construct.Sub.Group` for the definition of subgroups.
 
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -18,7 +18,7 @@ open import Algebra.Morphism.Structures using (IsGroupHomomorphism)
 open import Algebra.Properties.Monoid monoid
 open import Algebra.Properties.Group G using (⁻¹-anti-homo-∙)
 open import Algebra.Structures using (IsGroup)
-open import Data.Product.Base using (_,_; proj₁; proj₂)
+open import Data.Product.Base using (_,_)
 open import Function.Definitions using (Surjective)
 open import Level using (_⊔_)
 open import Relation.Binary.Core using (_⇒_)
@@ -28,7 +28,7 @@ open import Relation.Binary.Reasoning.Setoid setoid
 
 private
   module N = NormalSubgroup N
-open NormalSubgroup N using (ι; module ι; normal)
+open NormalSubgroup N using (ι; module ι; conjugate; normal)
 
 infix 0 _by_
 
@@ -66,19 +66,19 @@ open AlgDefs _≋_
 ≋-∙-cong : Congruent₂ _∙_
 ≋-∙-cong {x} {y} {u} {v} (g by ιg∙x≈y) (h by ιh∙u≈v) = g N.∙ h′ by begin
   ι (g N.∙ h′) ∙ (x ∙ u) ≈⟨ ∙-congʳ (ι.∙-homo g h′) ⟩
-  (ι g ∙ ι h′) ∙ (x ∙ u) ≈⟨ uv≈wx⇒yu∙vz≈yw∙xz (normal h x .proj₂) (ι g) u ⟩
+  (ι g ∙ ι h′) ∙ (x ∙ u) ≈⟨ uv≈wx⇒yu∙vz≈yw∙xz (normal h x) (ι g) u ⟩
   (ι g ∙ x) ∙ (ι h ∙ u)  ≈⟨ ∙-cong ιg∙x≈y ιh∙u≈v ⟩
   y ∙ v                  ∎
-  where h′ = normal h x .proj₁
+  where h′ = conjugate h x
 
 ≋-⁻¹-cong : Congruent₁ _⁻¹
 ≋-⁻¹-cong {x} {y} (g by ιg∙x≈y) = h by begin
-  ι h ∙ x ⁻¹        ≈⟨ normal (g N.⁻¹) (x ⁻¹) .proj₂ ⟩
+  ι h ∙ x ⁻¹        ≈⟨ normal (g N.⁻¹) (x ⁻¹) ⟩
   x ⁻¹ ∙ ι (g N.⁻¹) ≈⟨ ∙-congˡ (ι.⁻¹-homo g) ⟩
   x ⁻¹ ∙ ι g ⁻¹     ≈⟨ ⁻¹-anti-homo-∙ (ι g) x ⟨
   (ι g ∙ x) ⁻¹      ≈⟨ ⁻¹-cong ιg∙x≈y ⟩
   y ⁻¹              ∎
-  where h = normal (g N.⁻¹) (x ⁻¹) .proj₁
+  where h = conjugate (g N.⁻¹) (x ⁻¹)
 
 quotientIsGroup : IsGroup _≋_ _∙_ ε _⁻¹
 quotientIsGroup = record
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -19,6 +19,7 @@ open import Algebra.Properties.Monoid monoid
 open import Algebra.Properties.Group G using (⁻¹-anti-homo-∙)
 open import Algebra.Structures using (IsGroup)
 open import Data.Product.Base using (_,_)
+open import Function.Base using (_∘_)
 open import Function.Definitions using (Surjective)
 open import Level using (_⊔_)
 open import Relation.Binary.Core using (_⇒_)
@@ -90,15 +91,9 @@ quotientIsGroup = record
         }
       ; assoc = λ x y z → ≈⇒≋ (assoc x y z)
       }
-    ; identity = record
-      { fst = λ x → ≈⇒≋ (identityˡ x)
-      ; snd = λ x → ≈⇒≋ (identityʳ x)
-      }
+    ; identity = ≈⇒≋ ∘ identityˡ , ≈⇒≋ ∘ identityʳ
     }
-  ; inverse = record
-    { fst = λ x → ≈⇒≋ (inverseˡ x)
-    ; snd = λ x → ≈⇒≋ (inverseʳ x)
-    }
+  ; inverse = ≈⇒≋ ∘ inverseˡ , ≈⇒≋ ∘ inverseʳ
   ; ⁻¹-cong = ≋-⁻¹-cong
   }
 
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -66,15 +66,14 @@ open AlgDefs _≋_
 ≋-∙-cong : Congruent₂ _∙_
 ≋-∙-cong {x} {y} {u} {v} (g by ιg∙x≈y) (h by ιh∙u≈v) = g N.∙ h′ by begin
   ι (g N.∙ h′) ∙ (x ∙ u) ≈⟨ ∙-congʳ (ι.∙-homo g h′) ⟩
-  (ι g ∙ ι h′) ∙ (x ∙ u) ≈⟨ uv∙wx≈u[vw∙x] (ι g) (ι h′) x u ⟩
-  ι g ∙ ((ι h′ ∙ x) ∙ u) ≈⟨ uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx (normal h x .proj₂) u) (ι g) ⟨
+  (ι g ∙ ι h′) ∙ (x ∙ u) ≈⟨ uv≈wx⇒yu∙vz≈yw∙xz (normal h x .proj₂) (ι g) u ⟩
   (ι g ∙ x) ∙ (ι h ∙ u)  ≈⟨ ∙-cong ιg∙x≈y ιh∙u≈v ⟩
   y ∙ v                  ∎
   where h′ = normal h x .proj₁
 
 ≋-⁻¹-cong : Congruent₁ _⁻¹
 ≋-⁻¹-cong {x} {y} (g by ιg∙x≈y) = h by begin
-  ι h ∙ x ⁻¹        ≈⟨ normal (g N.⁻¹) (x ⁻¹) .proj₂ ⟨
+  ι h ∙ x ⁻¹        ≈⟨ normal (g N.⁻¹) (x ⁻¹) .proj₂ ⟩
   x ⁻¹ ∙ ι (g N.⁻¹) ≈⟨ ∙-congˡ (ι.⁻¹-homo g) ⟩
   x ⁻¹ ∙ ι g ⁻¹     ≈⟨ ⁻¹-anti-homo-∙ (ι g) x ⟨
   (ι g ∙ x) ⁻¹      ≈⟨ ⁻¹-cong ιg∙x≈y ⟩
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -105,6 +105,9 @@ quotientIsGroup = record
 quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
 quotientGroup = record { isGroup = quotientIsGroup }
 
+_/_ : Group c (c ⊔ ℓ ⊔ c′)
+_/_ = quotientGroup
+
 project : Group.Carrier G → Group.Carrier quotientGroup
 project x = x -- because we do all the work in the relation
 
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -35,8 +35,11 @@ infix 0 _by_
 data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
   _by_ : ∀ g → ι g ∙ x ≈ y → x ≋ y
 
+≈⇒≋ : _≈_ ⇒ _≋_
+≈⇒≋ {x} {y} x≈y = N.ε by trans (∙-cong ι.ε-homo x≈y) (identityˡ y)
+
 ≋-refl : Reflexive _≋_
-≋-refl {x} = N.ε by trans (∙-congʳ ι.ε-homo) (identityˡ x)
+≋-refl = ≈⇒≋ refl
 
 ≋-sym : Symmetric _≋_
 ≋-sym {x} {y} (g by ιg∙x≈y) = g N.⁻¹ by begin
@@ -57,9 +60,6 @@ data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
   ; sym = ≋-sym
   ; trans = ≋-trans
   }
-
-≈⇒≋ : _≈_ ⇒ _≋_
-≈⇒≋ {x} {y} x≈y = N.ε by trans (∙-cong ι.ε-homo x≈y) (identityˡ y)
 
 open AlgDefs _≋_
 
--- a/src/Algebra/Construct/Quotient/Group.agda
+++ b/src/Algebra/Construct/Quotient/Group.agda
@@ -1,0 +1,127 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Quotient groups
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group)
+open import Algebra.NormalSubgroup using (NormalSubgroup)
+
+module Algebra.Construct.Quotient.Group  {c ℓ} (G : Group c ℓ) {c′ ℓ′} (N : NormalSubgroup G c′ ℓ′) where
+
+open Group G
+
+import Algebra.Definitions as AlgDefs
+open import Algebra.Morphism.Structures using (IsGroupHomomorphism)
+open import Algebra.Properties.Monoid monoid
+open import Algebra.Properties.Group G using (⁻¹-anti-homo-∙)
+open import Algebra.Structures using (IsGroup)
+open import Data.Product.Base using (_,_; proj₁; proj₂)
+open import Function.Definitions using (Surjective)
+open import Level using (_⊔_)
+open import Relation.Binary.Core using (_⇒_)
+open import Relation.Binary.Definitions using (Reflexive; Symmetric; Transitive)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+private
+  module N = NormalSubgroup N
+open NormalSubgroup N using (ι; module ι; normal)
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → ι g ∙ x ≈ y → x ≋ y
+
+≋-refl : Reflexive _≋_
+≋-refl {x} = N.ε by trans (∙-congʳ ι.ε-homo) (identityˡ x)
+
+≋-sym : Symmetric _≋_
+≋-sym {x} {y} (g by ιg∙x≈y) = g N.⁻¹ by begin
+  ι (g N.⁻¹) ∙ y      ≈⟨ ∙-cong (ι.⁻¹-homo g) (sym ιg∙x≈y) ⟩
+  ι g ⁻¹ ∙ (ι g ∙ x)  ≈⟨ cancelˡ (inverseˡ (ι g)) x ⟩
+  x                   ∎
+
+≋-trans : Transitive _≋_
+≋-trans {x} {y} {z} (g by ιg∙x≈y) (h by ιh∙y≈z) = h N.∙ g by begin
+  ι (h N.∙ g) ∙ x ≈⟨ ∙-congʳ (ι.∙-homo h g) ⟩
+  (ι h ∙ ι g) ∙ x ≈⟨ uv≈w⇒xu∙v≈xw ιg∙x≈y (ι h) ⟩
+  ι h ∙ y         ≈⟨ ιh∙y≈z ⟩
+  z               ∎
+
+≋-isEquivalence : IsEquivalence _≋_
+≋-isEquivalence = record
+  { refl = ≋-refl
+  ; sym = ≋-sym
+  ; trans = ≋-trans
+  }
+
+≈⇒≋ : _≈_ ⇒ _≋_
+≈⇒≋ {x} {y} x≈y = N.ε by trans (∙-cong ι.ε-homo x≈y) (identityˡ y)
+
+open AlgDefs _≋_
+
+≋-∙-cong : Congruent₂ _∙_
+≋-∙-cong {x} {y} {u} {v} (g by ιg∙x≈y) (h by ιh∙u≈v) = g N.∙ h′ by begin
+  ι (g N.∙ h′) ∙ (x ∙ u) ≈⟨ ∙-congʳ (ι.∙-homo g h′) ⟩
+  (ι g ∙ ι h′) ∙ (x ∙ u) ≈⟨ uv∙wx≈u[vw∙x] (ι g) (ι h′) x u ⟩
+  ι g ∙ ((ι h′ ∙ x) ∙ u) ≈⟨ uv≈w⇒xu∙v≈xw (uv≈w⇒u∙vx≈wx (normal h x .proj₂) u) (ι g) ⟨
+  (ι g ∙ x) ∙ (ι h ∙ u)  ≈⟨ ∙-cong ιg∙x≈y ιh∙u≈v ⟩
+  y ∙ v                  ∎
+  where h′ = normal h x .proj₁
+
+≋-⁻¹-cong : Congruent₁ _⁻¹
+≋-⁻¹-cong {x} {y} (g by ιg∙x≈y) = h by begin
+  ι h ∙ x ⁻¹        ≈⟨ normal (g N.⁻¹) (x ⁻¹) .proj₂ ⟨
+  x ⁻¹ ∙ ι (g N.⁻¹) ≈⟨ ∙-congˡ (ι.⁻¹-homo g) ⟩
+  x ⁻¹ ∙ ι g ⁻¹     ≈⟨ ⁻¹-anti-homo-∙ (ι g) x ⟨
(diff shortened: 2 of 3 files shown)


src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,35 @@
...
+open import Algebra.Construct.Sub.Group G using (Subgroup)
+open import Data.Product.Base using (∃-syntax; _,_)
+open import Level using (suc; _⊔_)
+
+private
+  module G = Group G
+
+-- every element of the subgroup commutes in G
+Normal : ∀ {c′ ℓ′} → Subgroup c′ ℓ′ → Set (c ⊔ ℓ ⊔ c′)
+Normal subgroup = ∀ n g → ∃[ n′ ] ι n′ G.∙ g G.≈ g G.∙ ι n
```
jamesmckinna: On the `Structures`/`Bundles` convention, should we call this `IsNormal`?

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,35 @@
...
+  where open Subgroup subgroup
+
+record NormalSubgroup c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
```
jamesmckinna, line 27: And hence this becomes (simply?) `Normal`, as the `Bundled` version of `IsNormal`?

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,35 @@
...
+  field
+    subgroup : Subgroup c′ ℓ′
+    normal : Normal subgroup
```
jamesmckinna:
```suggestion
    isNormal : IsNormal subgroup
```
?

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,35 @@
...
+
+  open Subgroup subgroup public
+
+abelian⇒subgroup-normal : ∀ {c′ ℓ′} → Commutative G._≈_ G._∙_ → (subgroup : Subgroup c′ ℓ′) → Normal subgroup
```
jamesmckinna: Again
```suggestion
abelian⇒subgroup-normal : ∀ {c′ ℓ′} → Commutative G._≈_ G._∙_ → (subgroup : Subgroup c′ ℓ′) → IsNormal subgroup
```

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,126 @@
...
+    { fst = λ x → ≈⇒≋ (inverseˡ x)
+    ; snd = λ x → ≈⇒≋ (inverseʳ x)
+    }
+  ; ⁻¹-cong = ≋-⁻¹-cong
+  }
+
+quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
+quotientGroup = record { isGroup = quotientIsGroup }
+
+project : Group.Carrier G → Group.Carrier quotientGroup
```
jamesmckinna: My personal preference is for `π` (for projection), echoing `ι` (for inclusion), but I won't fight for it...
... maybe I should?

jamesmckinna (approved):
This all looks great. I've made suggestions, but nothing is a deal-breaker.

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,126 @@
...
+open import Relation.Binary.Reasoning.Setoid setoid
+
+private
+  module N = NormalSubgroup N
+open NormalSubgroup N using (ι; module ι; normal)
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → ι g ∙ x ≈ y → x ≋ y
```
jamesmckinna, line 37: Are you sticking with this version, or the symmetric one which multiplies on each side of the equation?
```suggestion
data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
  by: ∀ g h → ι g ∙ x ≈  ι h ∙ y → x ≋ y
```

plus some smart constructors/pattern synonyms to achieve the various properties?

src/Algebra/NormalSubgroup.agda
jamesmckinna: How does flipping the old definition improve matters?
Curious...

Taneb: If you look at the diff of the commit that flipped this (https://github.com/agda/agda-stdlib/pull/2854/commits/e3baf591bc94deaf20b6c359fbd95639f3a31708#diff-4b2ec40733e56b6ad2e835149fee88672cc51d5ec131be90521ea0098f4813cc) in the quotient groups module, previously every time I was using `normal` I had to `sym` it, this way I can use it directly

src/Algebra/Construct/Quotient/Group.agda
Taneb: I want to experiment with the symmetric version more, but:
* it means we need to carry around more data, operationally (minor issue)
* I think it works out pretty close to being equivalent to this construction on the Grothendieck group for a cancellative monoid, and if that's the case I'd rather have that in two steps explicitly

src/Algebra/Construct/Quotient/Group.agda
Taneb: I've got no strong feeling here

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,35 @@
...
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group)
+
+module Algebra.NormalSubgroup {c ℓ} (G : Group c ℓ)  where
+
+open import Algebra.Definitions
+open import Algebra.Construct.Sub.Group G using (Subgroup)
+open import Data.Product.Base using (∃-syntax; _,_)
```
jamesmckinna: is this the best way to represent `Normal`, or would having a bespoke `record` definition be better for eg. type inference?

As well as eg. widening such a record with additional manifest fields (not sure what's needed here, but... extensibility considerations in general make the use of Sigma-types an anti-pattern for me)

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,126 @@
...
+    }
+  ; inverse = record
+    { fst = λ x → ≈⇒≋ (inverseˡ x)
+    ; snd = λ x → ≈⇒≋ (inverseʳ x)
+    }
+  ; ⁻¹-cong = ≋-⁻¹-cong
+  }
+
+quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
+quotientGroup = record { isGroup = quotientIsGroup }
```
jamesmckinna, line 101: cf. #2855 I'd like an additional public export here
```suggestion
quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
quotientGroup = record { isGroup = quotientIsGroup }

module _/_ = Group quotientGroup
```

so that the quotient *construction* may always be referred to via `G / N` in any client context-of-use

src/Algebra/NormalSubgroup.agda
JacquesCarette: My instincts also go in that direction, i.e. that `Normal` should be a record.

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,41 @@
...
+open import Level using (suc; _⊔_)
+
+private
+  module G = Group G
+
+-- every element of the subgroup commutes in G
+record IsNormal {c′ ℓ′} (subgroup : Subgroup c′ ℓ′) : Set (c ⊔ ℓ ⊔ c′) where
+  open Subgroup subgroup
+  field
+    normal : ∀ n g → ∃[ n′ ] ι n′ G.∙ g G.≈ g G.∙ ι n
```
Taneb: Now that this is a record, should this field be split into two, one which gives the value, and one which gives the proof?

src/Algebra/NormalSubgroup.agda
Taneb: I'd rather that this remained `NormalSubgroup`, because we're using `Normal` as an adjective, not a noun (unlike the records in `Algebra.Bundles`, which are all nouns)

src/Algebra/Construct/Quotient/Group.agda
Taneb: I don't think this should be a _module_ here. It would be awkward to use as such, you'd end up with things like `G /.inverseˡ N` which doesn't read clearly. I will add this as a _value_ however.

src/Algebra/NormalSubgroup.agda
jamesmckinna: Absolutely!
The use of Sigma/exists fossilises the two fields as fst and snd, or whatever they've (locally) been renamed to, but a record permits both more informative names, and the possibility to extend with additional structure if necessary/desirable. 

Separately, the use of bespoke records may help elsewhere with type inference...

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+open import Function.Definitions using (Surjective)
+open import Level using (_⊔_)
+open import Relation.Binary.Core using (_⇒_)
+open import Relation.Binary.Definitions using (Reflexive; Symmetric; Transitive)
+open import Relation.Binary.Structures using (IsEquivalence)
+open import Relation.Binary.Reasoning.Setoid setoid
+
+private
+  module N = NormalSubgroup N
+open NormalSubgroup N using (ι; module ι; conjugate; normal)
```
jamesmckinna, line 32: Can simplify this to:
```suggestion
private
  open module N = NormalSubgroup N using (ι; module ι; conjugate; normal)
```

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,37 @@
...
+  field
+    subgroup : Subgroup c′ ℓ′
+    isNormal : IsNormal subgroup
+
+  open Subgroup subgroup public
+  open IsNormal isNormal public
+
+abelian⇒subgroup-normal : ∀ {c′ ℓ′} → Commutative G._≈_ G._∙_ → (subgroup : Subgroup c′ ℓ′) → IsNormal subgroup
+abelian⇒subgroup-normal ∙-comm subgroup = record { normal = λ n g → ∙-comm (ι n) g }
+  where open Subgroup subgroup
```
jamesmckinna, line 37: I'd maybe be tempted to lift this out to `Algebra.Construct.Quotient.AbelianGroup` to bundle up the quotient-out-of-any-subgroup construction, and simplify the downstream imports in #2855 ? See #2859 .

CHANGELOG.md
```diff
@@ -79,10 +79,14 @@ Deprecated names
 New modules
 -----------
 
+* `Algebra.Cosntruct.Quotient.Group` for the definition of quotient groups.
```
jamesmckinna: Typo
```suggestion
* `Algebra.Construct.Quotient.Group` for the definition of quotient groups.
```

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+quotientGroup = record { isGroup = quotientIsGroup }
+
+_/_ : Group c (c ⊔ ℓ ⊔ c′)
+_/_ = quotientGroup
+
+project : Group.Carrier G → Group.Carrier quotientGroup
+project x = x -- because we do all the work in the relation
+
+project-isHomomorphism : IsGroupHomomorphism rawGroup (Group.rawGroup quotientGroup) project
+project-isHomomorphism = record
```
jamesmckinna: Naming?
```suggestion
project-isGroupHomomorphism : IsGroupHomomorphism rawGroup (Group.rawGroup quotientGroup) project
project-isGroupHomomorphism = record
```

Twofold reasons:
* Subsequently, for `Ring R / Ideal I`, you also define `project-isHomomorphism : IsRingHomomorphism`, and maybe that's enough (let the `X` path context of the algebra determine what kind of `IsXHomomorphism` is defined
* BUT, for `Ring`, what you actually need is to separate out the `isMonoidHomomorphism` field, so why not also `isGroupHomomorphism`?

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+  { isMonoidHomomorphism = record
+    { isMagmaHomomorphism = record
+      { isRelHomomorphism = record
+        { cong = ≈⇒≋
+        }
+      ; homo = λ _ _ → ≋-refl
+      }
+    ; ε-homo = ≋-refl
+    }
```
jamesmckinna, line 119: This should be lifted out as a separate (prior) definition
```agda
project-isMagmaHomomorphism : IsMagmaHomomorphism rawMagma (Group.rawMagma quotientGroup) project
project-isMagmaHomomorphism = record
  { isRelHomomorphism = record
    { cong = ≈⇒≋
    }
  ; homo = λ _ _ → ≋-refl
  }

project-isMonoidHomomorphism : IsMonoidHomomorphism rawMonoid (Group.rawMonoid quotientGroup) project
project-isMonoidHomomorphism = record
  { isMagmaHomomorphism = project-isMagmaHomomorphism
  ; ε-homo = ≋-refl
  }

project-isGroupHomomorphism : IsGroupHomomorphism rawGroup (Group.rawGroup quotientGroup) project
project-isGroupHomomorphism = record
  { isMonoidHomomorphism = project-isMonoidHomomorphism
  ; ⁻¹-homo = λ _ → ≋-refl
  }
```

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,37 @@
...
+-- Definition of normal subgroups
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group)
+
+module Algebra.NormalSubgroup {c ℓ} (G : Group c ℓ)  where
+
+open import Algebra.Definitions
```
jamesmckinna: Unused!

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+≋-isEquivalence : IsEquivalence _≋_
+≋-isEquivalence = record
+  { refl = ≋-refl
+  ; sym = ≋-sym
+  ; trans = ≋-trans
+  }
+
+≈⇒≋ : _≈_ ⇒ _≋_
+≈⇒≋ {x} {y} x≈y = N.ε by trans (∙-cong ι.ε-homo x≈y) (identityˡ y)
+
```
jamesmckinna:
```suggestion
```
In view of the above refactoring of reflexivity...

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+  module N = NormalSubgroup N
+open NormalSubgroup N using (ι; module ι; conjugate; normal)
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → ι g ∙ x ≈ y → x ≋ y
+
+≋-refl : Reflexive _≋_
+≋-refl {x} = N.ε by trans (∙-congʳ ι.ε-homo) (identityˡ x)
```
jamesmckinna:
```suggestion
≈⇒≋ : _≈_ ⇒ _≋_
≈⇒≋ {x} {y} x≈y = N.ε by trans (∙-cong ι.ε-homo x≈y) (identityˡ y)

≋-refl : Reflexive _≋_
≋-refl = ≈⇒≋ refl
```
Unsurprisingly, these proofs are definitionally equal to the previous versions, but put `≈⇒≋` ('abstract' reflexivity...) higher up the food chain.

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+
+module Algebra.Construct.Quotient.Group  {c ℓ} (G : Group c ℓ) {c′ ℓ′} (N : NormalSubgroup G c′ ℓ′) where
+
+open Group G
+
+import Algebra.Definitions as AlgDefs
+open import Algebra.Morphism.Structures using (IsGroupHomomorphism)
+open import Algebra.Properties.Monoid monoid
+open import Algebra.Properties.Group G using (⁻¹-anti-homo-∙)
+open import Algebra.Structures using (IsGroup)
```
jamesmckinna, line 20: Do we need this import? Cf. #2391 and see below.

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,129 @@
...
+    }
+  ; inverse = record
+    { fst = λ x → ≈⇒≋ (inverseˡ x)
+    ; snd = λ x → ≈⇒≋ (inverseʳ x)
+    }
+  ; ⁻¹-cong = ≋-⁻¹-cong
+  }
+
+quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
+quotientGroup = record { isGroup = quotientIsGroup }
```
jamesmckinna, line 101: And now inline the definition above?
```suggestion
quotientGroup = record
  { isGroup = ...
  }
```

jamesmckinna:
In lieu of any further review comments, please see #2859 (purely!) as a comparison point, and feel free to grab anything from there which you might find useful... it just seemed the easiest way to encapsulate all the thoughts I had had about your groups-rings-modules development

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: On reflection, my response to @JacquesCarette 's request for comments on #2859 suggest that I perhaps row back on this suggestion/request, in favour of suitable `open` declarations to bring these three things into scope... cf. #2391

src/Algebra/NormalSubgroup.agda
Taneb: No it's not, that's where `Commutative` is exported from. I'll add an explicit import

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: Then let me push for it!

Taneb: Done!

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,124 @@
...
+open import Relation.Binary.Reasoning.Setoid setoid
+
+private
+  module N = NormalSubgroup N
+open NormalSubgroup N using (ι; module ι; conjugate; normal)
+
+infix 0 _by_
+
+data _≋_ (x y : Carrier) : Set (c ⊔ ℓ ⊔ c′) where
+  _by_ : ∀ g → ι g ∙ x ≈ y → x ≋ y
```
Taneb, line 37: If this is instead `_by_ : ∀ g → x ∙ ι g ≈ y → x ≋ y` some things line up nicer when we get to integers (in particular it matches `Data.Integer.DivMod.a≡a%n+[a/n]*n`)

jamesmckinna: Well, so be it, but be careful of having a tail wag the dog. It wouldn't be the first time that existing `stdlib` modules expose argument order/definitional inconsistency when exposed to new additions/refactoriings (the saga of `Algebra.Properties.*.Divisibility` and `Data.Nat.DivMod.*` being a case in point)...

Taneb: The alternative here would be putting more lemmas in `Data.Integer.DivMod`, which I want to do anyway

jamesmckinna:
FTR, if you don't feel like augmenting this with the `Sub.AbelianGroup` additions (plus their consequences for `NormalSubgroup`s and `Quotient.AbelianGroup` in #2859 , I'll be sure to post a follow-up PR ;-)

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: Well, after my recent flurry of activity around `Central`/`Centre.*` #2872 #2873 #2885 I've been turning over-and-over this question of what is the mathematically/formalistically 'best' account of the equivalence relation on the `Carrier` such that `id` is the function underlying the quotient map, *and* such that the identification of the `Kernel` of the quotient homomorphism with the `Normal` subgroup which gives rise to it. 

Envisaging `Kernel`s #2871 #1966  being designed/defined on top of the recent addition `Relation.Binary.Morphism.Construct.On`  in #2872 , the conclusion I come to is that the symmetric form of the `by` constructor would be more streamlined...

... and avoid having to commit you to one or other 'handed' choice...

* the carrier of (the kernel of a hom `h` onto) the quotient becomes triples `(m,x,n)` st `h(m)x = xh(n)`,
* this is obviously `Normal`... the conjugate swaps `m` and `n` (up to `h`...)

conversely, given a `Normal` subgroup, we obtain an element of the kernel as `(m, x, conjugate(m))`...

etc. details to work out, but I think overall the symmetric choice for the `by` constructor would work more smoothly.

src/Algebra/NormalSubgroup.agda
```diff
@@ -0,0 +1,37 @@
...
+
+private
+  module G = Group G
+
+-- every element of the subgroup commutes in G
+record IsNormal {c′ ℓ′} (subgroup : Subgroup c′ ℓ′) : Set (c ⊔ ℓ ⊔ c′) where
+  open Subgroup subgroup
+  field
+    conjugate : ∀ n g → Carrier
+    normal : ∀ n g → ι (conjugate n g) G.∙ g G.≈ g G.∙ ι n
```
jamesmckinna, line 25: As this PR matures, I can't help thinking that while this definition is perfectly fine *for `Group`s*, nevertheless it still embodies a left/right handedness which we should try as far as possible to resist committing to: calling it `IsNormal` for the sake of argument, and rename its fields to:
```agda
record IsNormalʳ {c′ ℓ′} (subgroup : Subgroup c′ ℓ′) : Set (c ⊔ ℓ ⊔ c′) where
  open Subgroup subgroup using (ι)
  field
    _ι∙ʳ_ : ∀ g n → typeOf n

    normalʳ : ∀ n g → ι (g ι∙ʳ n) G.∙ g G.≈ g G.∙ ι n
```
(there's some notation design to consider here! #2894 ).

By symmetry we should also consider 
```agda
    _ι∙ˡ_ : ∀ n g → typeOf n
    normalˡ : ∀ n g → g G.∙ ι (n ι∙ˡ g) G.≈ ι n G.∙ g
```
and, again *for `Group`s*, there is a relationship between `g ι∙ʳ n` and `n ι∙ˡ g` (roughly: 'conjugate of the inverse is the inverse of the conjugate'), but in general, we might expect no such relationship?

I'd first thought that 'conjugate of the inverse is the inverse of the conjugate' should be reified as a manifest field (and still could: to simplify certain arguments, if nothing else), but the more I thought about it, the more I thought that 'left-normal' and 'right-normal' are free-standing concepts... ?

Cf. left- and right-actions #2350 (normal subgroups, defined by monomorphism, induce left- and right- actions on their ambient group... with certain properties...)

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: For comparison:
```agda
-- Quotient groups

module Quotient {c′ ℓ′} (N : NormalSubgroup c′ ℓ′) where

  private

    open G
      using (_≈_; _∙_; ε; _⁻¹)

    open module N = NormalSubgroup N
      using (ι; module ι; conjugate; normal)

-- Definitions

-- symmetric form

  data _≋_ x y : Set (c ⊔ ℓ ⊔ c′) where
     ≋-intro : ∀ g h → ι g ∙ x ≈ y ∙ ι h → x ≋ y

-- left-biased form

  data _≋ˡ_ x y : Set (c ⊔ ℓ ⊔ c′) where
     ≋ˡ-intro : ∀ g → ι g ∙ x ≈ y → x ≋ˡ y

-- right-biased form

  data _≋ʳ_ x y : Set (c ⊔ ℓ ⊔ c′) where
     ≋ʳ-intro : ∀ h → x ≈ y ∙ ι h → x ≋ʳ y

-- smart constructors

  ≋ˡ⇒≋ : _≋ˡ_ ⇒ _≋_
  ≋ˡ⇒≋ (≋ˡ-intro g gx≈y) = ≋-intro g N.ε (G.trans gx≈y (G.sym (elimʳ ι.ε-homo _)))

  ≋ʳ⇒≋ : _≋ʳ_ ⇒ _≋_
  ≋ʳ⇒≋ (≋ʳ-intro h x≈yh) = ≋-intro N.ε h (G.trans (elimˡ ι.ε-homo _) x≈yh)

  ≋⇒≋ˡ : _≋_ ⇒ _≋ˡ_
  ≋⇒≋ˡ {x} {y} (≋-intro g h gx≈yh) = ≋ˡ-intro ((h′ N.⁻¹) N.∙ g) $ begin
    ι (h′ N.⁻¹ N.∙ g) ∙ x   ≈⟨ G.∙-congʳ (ι.∙-homo (h′ N.⁻¹) g) ⟩
    (ι (h′ N.⁻¹) ∙ ι g) ∙ x ≈⟨ G.assoc _ _ _ ⟩
    ι (h′ N.⁻¹) ∙ (ι g ∙ x) ≈⟨ G.∙-congˡ gx≈yh ⟩
    ι (h′ N.⁻¹) ∙ (y ∙ ι h) ≈⟨ G.∙-cong (G.sym (ι.⁻¹-homo h′)) (normal h y) ⟨
    (ι h′) ⁻¹ ∙ (ι h′ ∙ y)  ≈⟨ cancelˡ (G.inverseˡ _) y ⟩
    y ∎
    where h′ = conjugate y h

  ≋⇒≋ʳ : _≋_ ⇒ _≋ʳ_
  ≋⇒≋ʳ {x} {y} (≋-intro g h gx≈yh) = ≋ʳ-intro (h N.∙ (g′ N.⁻¹)) $ begin
    x                           ≈⟨ x≈z//y _ _ _ xg′≈gx ⟩
    (ι g ∙ x) ∙ ι g′ ⁻¹         ≈⟨ G.∙-congʳ gx≈yh ⟩
    ((y ∙ (ι h)) ∙ (ι g′) ⁻¹)   ≈⟨ G.∙-congˡ (ι.⁻¹-homo g′) ⟨
    (y ∙ (ι h)) ∙ ι (g′ N.⁻¹)   ≈⟨ G.assoc _ _ _ ⟩
    (y ∙ ((ι h) ∙ ι (g′ N.⁻¹))) ≈⟨ G.∙-congˡ (ι.∙-homo h (g′ N.⁻¹)) ⟨
    y ∙ ι (h N.∙ g′ N.⁻¹)       ∎
    where
    g′ = conjugate (x ⁻¹) g
    xg′≈gx : x ∙ ι g′ ≈ ι g ∙ x
    xg′≈gx = ∙-cancelʳ (x ⁻¹) _ _ $ begin
      (x ∙ ι g′) ∙ x ⁻¹ ≈⟨ G.assoc _ _ _ ⟩
      x ∙ (ι g′ ∙ x ⁻¹) ≈⟨ G.∙-congˡ (normal g (x ⁻¹)) ⟩
      x ∙ (x ⁻¹ ∙ ι g)  ≈⟨ cancelˡ (G.inverseʳ _) _ ⟩
      ι g               ≈⟨ cancelʳ (G.inverseʳ _) _ ⟨
      (ι g ∙ x) ∙ x ⁻¹  ∎

-- Properties

  ≈⇒≋ : _≈_ ⇒ _≋_
  ≈⇒≋ {x = x} {y = y} x≈y = ≋-intro N.ε N.ε $ begin
    ι N.ε ∙ x ≈⟨ G.∙-cong ι.ε-homo x≈y ⟩
    ε ∙ y     ≈⟨ ε-central y ⟩
    y ∙ ε     ≈⟨ G.∙-congˡ ι.ε-homo ⟨
    y ∙ ι N.ε ∎

  refl : Reflexive _≋_
  refl = ≈⇒≋ G.refl

  sym : Symmetric _≋_
  sym {x} {y} (≋-intro g h gx≈yh) = ≋-intro (g N.⁻¹) (h N.⁻¹) $
    ∙-cancelˡ (ι g) _ _ $ ∙-cancelʳ (ι h) _ _ $ begin
    ι g ∙ (ι (g N.⁻¹) ∙ y) ∙ ι h
      ≈⟨ G.∙-congʳ (cancelˡ (G.trans (G.∙-congˡ (ι.⁻¹-homo g)) (G.inverseʳ _)) y) ⟩
    y ∙ ι h
      ≈⟨ gx≈yh ⟨
    ι g ∙ x
      ≈⟨ uv≈w⇒xu∙v≈xw (cancelʳ (G.trans (G.∙-congʳ (ι.⁻¹-homo h)) (G.inverseˡ _)) x) _ ⟨
    ι g ∙ (x ∙ ι (h N.⁻¹)) ∙ ι h ∎

  trans : Transitive _≋_
  trans {x} {y} {z} (≋-intro g h gx≈yh) (≋-intro m n my≈zn) =
    ≋-intro (m N.∙ g) (n N.∙ h) $ begin
      ι (m N.∙ g) ∙ x ≈⟨ uv≈w⇒u∙vx≈wx (G.sym (ι.∙-homo m g)) x ⟨
      ι m ∙ (ι g ∙ x) ≈⟨ G.∙-congˡ gx≈yh ⟩
      ι m ∙ (y ∙ ι h) ≈⟨ uv≈w⇒u∙vx≈wx my≈zn (ι h) ⟩
      (z ∙ ι n) ∙ ι h ≈⟨ uv≈w⇒xu∙v≈xw (G.sym (ι.∙-homo n h)) z ⟩
      z ∙ ι (n N.∙ h) ∎

  ⁻¹-cong : Congruent₁ _≋_ _⁻¹
  ⁻¹-cong {x} {y} (≋-intro g h gx≈yh) =
    ≋-intro h g $ ∙-cancelˡ y _ _ $ ∙-cancelʳ x _ _ $ begin
    (y ∙ (ι h ∙ x ⁻¹)) ∙ x ≈⟨ uv≈w⇒xu∙v≈xw (cancelʳ (G.inverseˡ x) (ι h)) y ⟩ 
    y ∙ ι h                ≈⟨ gx≈yh ⟨
    ι g ∙ x                ≈⟨ G.∙-congʳ (cancelˡ (G.inverseʳ y) (ι g)) ⟨
    (y ∙ (y ⁻¹ ∙ ι g)) ∙ x ∎

  ∙-cong : Congruent₂ _≋_ _∙_
  ∙-cong {x} {y} {u} {v} x≋y u≋v
    with ≋ˡ-intro g gx≈y ← ≋⇒≋ˡ x≋y | ≋ʳ-intro h u≈vh ← ≋⇒≋ʳ u≋v
    = ≋-intro g h $ begin
      ι g ∙ (x ∙ u) ≈⟨ uv≈w⇒u∙vx≈wx gx≈y u ⟩
      y ∙ u         ≈⟨ uv≈w⇒xu∙v≈xw (G.sym u≈vh) y ⟨
      (y ∙ v) ∙ ι h ∎
```
...
and each of these proofs I think ca be further simplified with a bit more thought/refactoring... ;-)

jamesmckinna:
@Taneb do you want to try to finish this (and related PRs) for v2.4? It's currently unbadged as to milestone...

src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,124 @@
...
+  x ⁻¹ ∙ ι g ⁻¹     ≈⟨ ⁻¹-anti-homo-∙ (ι g) x ⟨
+  (ι g ∙ x) ⁻¹      ≈⟨ ⁻¹-cong ιg∙x≈y ⟩
+  y ⁻¹              ∎
+  where h = conjugate (g N.⁻¹) (x ⁻¹)
+
+quotientIsGroup : IsGroup _≋_ _∙_ ε _⁻¹
+quotientIsGroup = record
+  { isMonoid = record
+    { isSemigroup = record
+      { isMagma = record
```
gallais, line 88: Lots of nested records :s

JacquesCarette: What the 1lab is doing is to make the 'official' version of stuff nested, but then provide a helper "flat" record for users to fill in (and a smart constructor to shove things over). The helper also omits fields when they are derivable.

Taneb: I'm happy to unnest these once their contents is determined

Issue #2804, opened 2025-08-04
[ design ] Re-exports from `Algebra.Properties.X`


jamesmckinna:
Reading over #2695 and its antecedent PRs #2688 and #2692 , I have the feeling that we still haven't quite got the API right for `Properties` modules:
* it's not obvious that we do, but I think we should, reexport the axioms from the underlying `isX : IsX ...` substructure, otherwise we are left in the (slightly?) uneasy position of having to make multiple re-imports to get the 'widest possible' API (so perhaps the design question is more: "how wide should the API be?")
* we certainly don't consistently re-export `Algebra.Properties.Y y` for each sub-bundle `y :Y` which is re-exported publicly by bundle `X`, but again, I think that we should (which, if nothing else, would encourage us to improve consistency in naming such properties)

`Group` is the first really intricate version of the phenomenon, where some recent retrofitting #2251 manages to re-export the underlying `Loop` and `Quasigroup` properties, but we still don't re-export the `Monoid` (and hence `Semigroup`) properties for the underlying `monoid` object.

We should improve our documentation of what we (think that we should) do, and try to ensure that we do, indeed, do it!

MatthewDaggitt:
I agree with both of those suggestions. The only question is whether this is a breaking change because if people have opened both `IsX` and `Properties.X` will they get clashes if `Properties.X` also exports `IsX` fields? I suspect we probably will...

jamesmckinna:
Happy to badge this as v3.0, not least because that would give us longer to discuss/reflect. 

I think my instincts in this context are to make the API as wide as possible cf. #2762 but perhaps there are possibilities to offer users 'narrow' and 'wide' versions (for `Structures`), and to refactor `Properties` in similar fashion? 

But yes, I think you're right that trying to amalgamate `Properties` 'in one place' will end up being `breaking`...

JacquesCarette:
I am definitely a big fan of offering both narrow and wide versions of APIs.

Library developers are going to want to be as precise as possible, with as few dependencies as possible. Users who are merely looking to "get things done" generally want a wide version. It's the difference between "the theory (presentation) of *a* group" (something rather small) and "Group Theory" (a book's worth of stuff).

jamesmckinna:
[...]

PR #2855, opened 2025-10-31
Add ideals and quotient rings

Taneb:
Add definition of ideals and quotient rings. Based on #2854. Taken from #2729

### DIFF
--- a/src/Algebra/Construct/Quotient/Ring.agda
+++ b/src/Algebra/Construct/Quotient/Ring.agda
@@ -11,6 +11,11 @@ open import Algebra.Ideal using (Ideal)
 
 module Algebra.Construct.Quotient.Ring {c ℓ} (R : Ring c ℓ) {c′ ℓ′} (I : Ideal R c′ ℓ′) where
 
+open import Algebra.Morphism.Structures using (IsRingHomomorphism)
+open import Algebra.Properties.Ring R
+open import Algebra.Structures
+open import Level
+
 open Ring R
 private module I = Ideal I
 open I using (ι; normalSubgroup)
@@ -18,14 +23,8 @@ open I using (ι; normalSubgroup)
 open import Algebra.Construct.Quotient.Group +-group normalSubgroup public
   using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; π; π-surjective)
   renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
-
 open import Algebra.Definitions _≋_
-open import Algebra.Morphism.Structures using (IsRingHomomorphism)
 open import Algebra.Properties.Semiring semiring
-open import Algebra.Properties.Ring R
-open import Algebra.Structures
-open import Function.Definitions using (Surjective)
-open import Level
 open import Relation.Binary.Reasoning.Setoid setoid
 
 ≋-*-cong : Congruent₂ _*_


src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,31 @@
...
+record Ideal c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    -- a (two sided) ideal is exactly a subbimodule of R considered as a bimodule over itself
+    subbimodule : Subbimodule (TU.bimodule {R = R}) c′ ℓ′
+
+  open Subbimodule subbimodule public
+
+  normalSubgroup : NormalSubgroup c′ ℓ′
+  normalSubgroup = record
+    { subgroup = subgroup
```
jamesmckinna: Is this first field inferrable by Agda from the second one?
if so, is it better, or worse, to avoid mentioning it explicitly?

Apparently, it's not... oh well.

src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,31 @@
...
+open Ring R using (+-group; +-comm)
+
+open import Algebra.Module.Construct.Sub.Bimodule using (Subbimodule)
+import Algebra.Module.Construct.TensorUnit as TU
+open import Algebra.NormalSubgroup (+-group)
+open import Level
+
+record Ideal c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    -- a (two sided) ideal is exactly a subbimodule of R considered as a bimodule over itself
```
jamesmckinna: I'd put the comment *above* the `record` declaration, and even embellish it to:
```suggestion
------------------------------------------------------------------------
-- Definition
-- a (two-sided) ideal is exactly a sub-bimodule of R considered as a bimodule over itself

record Ideal c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
  field
```

src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,31 @@
...
+    subbimodule : Subbimodule (TU.bimodule {R = R}) c′ ℓ′
```
jamesmckinna: I personally find the parametrisation in `TensorUnit` to make this kind of thing more ugly to read than is wholly necessary (I'd perhaps downstream even reconsider some of the implicit vs. explicit quantification there), and, because `Subbimodule` already takes an `{R : Ring _ _}` parameter, I'd be tempted to write instead:
```suggestion
    subbimodule : Subbimodule {R = R} (TU.bimodule) c′ ℓ′
```
which would then permit the earlier import of `TU` to actually be handled in place as
```agda
open import Algebra.Module.Construct.TensorUnit using (bimodule)
```

I'd even be tempted to go further: conventionally, the canonical name for the tensor unit (wrt whatever tensor product structure) is `\bI`, so I'd even make this a `renaming` import

```agda
open import Algebra.Module.Construct.TensorUnit renaming (bimodule to 𝕀)
```

but this is being hyper-nitpicky! (maybe!?)

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+
+module Algebra.Construct.Quotient.Ring {c ℓ} (R : Ring c ℓ) {c′ ℓ′} (I : Ideal R c′ ℓ′) where
+
+open Ring R
+private module I = Ideal I
+open I using (ι; normalSubgroup)
+
+open import Algebra.Construct.Quotient.Group +-group normalSubgroup public
+  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; project; project-surjective)
+  renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
```
jamesmckinna, line 25: Given that any `Algebra.Construct.Quotient` module is going to take two explicit parameters
* the thing-to-be-quotiented
* the thing-to-be-quotiented-by

I'd like it if each such module declared for public export a binary infix 
```agda
module _/_ = ...
```
with definiens the bundle being defined, so that one could instead here write
```suggestion
open import Algebra.Construct.Quotient.Group using (module _/_)
private
  module R = Ring R
  module I = Ideal I
  module R/I = R.+-group / I.normalSubgroup

open R
open I using (ι)

open R/I public
  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; project; project-surjective)
  renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
```

etc. 
(obviously some further nitpicking need to localise what else of `R` gets used, so maybe the full `open R` is not actually called for?)

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+
+open import Algebra.Definitions _≋_
+open import Algebra.Morphism.Structures using (IsRingHomomorphism)
+open import Algebra.Properties.Semiring semiring
+open import Algebra.Properties.Ring R
```
jamesmckinna: A pity that we don't (yet) have it so that the second presupposes the first cf. #2804 ?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+open import Algebra.Structures
+open import Function.Definitions using (Surjective)
+open import Level
+open import Relation.Binary.Reasoning.Setoid setoid
```
jamesmckinna, line 28: Ditto. should this be part of the API exported by `Algebra.Properties.X`? Do we ever invoke algebraic properties *not* in a context where we have access to equational reasoning wrt that algebra?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+open import Algebra.Construct.Quotient.Group +-group normalSubgroup public
+  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; project; project-surjective)
+  renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
+
+open import Algebra.Definitions _≋_
+open import Algebra.Morphism.Structures using (IsRingHomomorphism)
+open import Algebra.Properties.Semiring semiring
+open import Algebra.Properties.Ring R
+open import Algebra.Structures
+open import Function.Definitions using (Surjective)
```
jamesmckinna: Unused, AFAICT?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+open Ring R
+private module I = Ideal I
+open I using (ι; normalSubgroup)
+
+open import Algebra.Construct.Quotient.Group +-group normalSubgroup public
+  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; project; project-surjective)
+  renaming (≋-∙-cong to ≋-+-cong; ≋-⁻¹-cong to ≋‿-‿cong)
+
+open import Algebra.Definitions _≋_
+open import Algebra.Morphism.Structures using (IsRingHomomorphism)
```
jamesmckinna: Should be bumped up above the `private` block?

src/Algebra/Construct/Quotient/Ring.agda
JacquesCarette: I support that idea.

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+open import Algebra.Properties.Semiring semiring
+open import Algebra.Properties.Ring R
+open import Algebra.Structures
+open import Function.Definitions using (Surjective)
+open import Level
+open import Relation.Binary.Reasoning.Setoid setoid
+
+≋-*-cong : Congruent₂ _*_
```
JacquesCarette, line 30: I would be tempted to use a local module (I forget the syntax for doing just inside the proof part) where `I` is opened so that all these qualifiers can go away. It's an idiom used a lot in 1lab that I need to integrate into my own Agda.

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+
+quotientIsRing : IsRing _≋_ _+_ _*_ (-_) 0# 1#
+quotientIsRing = record
+  { +-isAbelianGroup = record
+    { isGroup = quotientIsGroup
+    ; comm = λ x y → ≈⇒≋ (+-comm x y)
+    }
+  ; *-cong = ≋-*-cong
+  ; *-assoc = λ x y z → ≈⇒≋ (*-assoc x y z)
+  ; *-identity = record
```
JacquesCarette, line 58: as this is 'just' a pair, use `_,_` instead of a record? And maybe also eta contract?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+    { fst = λ x → ≈⇒≋ (*-identityˡ x)
+    ; snd = λ x → ≈⇒≋ (*-identityʳ x)
+    }
+  ; distrib = record
```
JacquesCarette, line 62: `_,`_ would work here too (but not eta)

src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,30 @@
...
+
+  open Subbimodule subbimodule public
+
+  normalSubgroup : NormalSubgroup c′ ℓ′
+  normalSubgroup = record
+    { isNormal = abelian⇒subgroup-normal +-comm subgroup
```
jamesmckinna, line 32: 🎉

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Ring; RawRing)
+open import Algebra.Ideal using (Ideal)
+
+module Algebra.Construct.Quotient.Ring {c ℓ} (R : Ring c ℓ) {c′ ℓ′} (I : Ideal R c′ ℓ′) where
+
+open Ring R
+private module I = Ideal I
+open I using (ι; normalSubgroup)
```
jamesmckinna, line 21: You can, somewhat counterintuitively, simplify this to
```suggestion
private open module I = Ideal I using (ι; normalSubgroup)
```

src/Algebra/Construct/Quotient/Ring.agda
jamesmckinna: However, much as I would like such an extension, Agda seems to complain about trying to use an *infix* `module` name when instantiated... but maybe that simply means it needs an additional fixity declaration...?
cf. https://github.com/agda/agda/issues/8191

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+    { fst = λ x y z → ≈⇒≋ (distribˡ x y z)
+    ; snd = λ x y z → ≈⇒≋ (distribʳ x y z)
+    }
+  }
+
+quotientRing : Ring c (c ⊔ ℓ ⊔ c′)
+quotientRing = record { isRing = quotientIsRing }
+
+project-isHomomorphism : IsRingHomomorphism rawRing quotientRawRing project
+project-isHomomorphism = record
```
jamesmckinna:
```suggestion
project-isRingHomomorphism : IsRingHomomorphism rawRing quotientRawRing project
project-isRingHomomorphism = record
```
As with #2854 maybe have the algebra name `X` in the name, as well as the type?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+open import Algebra.Ideal using (Ideal)
+
+module Algebra.Construct.Quotient.Ring {c ℓ} (R : Ring c ℓ) {c′ ℓ′} (I : Ideal R c′ ℓ′) where
+
+open Ring R
+private module I = Ideal I
+open I using (ι; normalSubgroup)
+
+open import Algebra.Construct.Quotient.Group +-group normalSubgroup public
+  using (_≋_; _by_; ≋-refl; ≋-sym; ≋-trans; ≋-isEquivalence; ≈⇒≋; quotientIsGroup; quotientGroup; project; project-surjective)
```
jamesmckinna: How much of this list actually needs to be exported?
Yes, you use `≋-refl` in particular, as well as `≈⇒≋` but the rest of the generic `≋-isEquivalence` structure is not used, and could be recreated if needed from `quotientGroup`? (variations on the theme of #2391 )

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+quotientRawRing : RawRing c (c ⊔ ℓ ⊔ c′)
+quotientRawRing = record
+  { Carrier = Carrier
+  ; _≈_ = _≋_
+  ; _+_ = _+_
+  ; _*_ = _*_
+  ; -_ = -_
+  ; 0# = 0#
+  ; 1# = 1#
+  }
```
jamesmckinna, line 48:
```suggestion
  { RawRing R.rawRing hiding (_≈_)
  ; _≈_ = _≋_
  }
```
avoids redundant DRY ...?

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+
+quotientIsRing : IsRing _≋_ _+_ _*_ (-_) 0# 1#
+quotientIsRing = record
+  { +-isAbelianGroup = record
+    { isGroup = quotientIsGroup
+    ; comm = λ x y → ≈⇒≋ (+-comm x y)
+    }
```
jamesmckinna, line 55: Don't need to create, nor import, `quotientIsGroup` here; can be recreated on the fly:
```suggestion
    { isGroup = isGroup
    ; comm = λ x y → ≈⇒≋ (+-comm x y)
    } where open Group quotientGroup using (isGroup)
```

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,90 @@
...
+    ; snd = λ x → ≈⇒≋ (*-identityʳ x)
+    }
+  ; distrib = record
+    { fst = λ x y z → ≈⇒≋ (distribˡ x y z)
+    ; snd = λ x y z → ≈⇒≋ (distribʳ x y z)
+    }
+  }
+
+quotientRing : Ring c (c ⊔ ℓ ⊔ c′)
+quotientRing = record { isRing = quotientIsRing }
```
jamesmckinna, line 69: Similarly, can we inline the definition of `quotientIsRing` here?

src/Algebra/Construct/Quotient/Ring.agda
jamesmckinna: So... no, for the time being!

jamesmckinna:
In lieu of any further review comments, please see #2859 (purely!) as a comparison point, and feel free to grab anything from there which you might find useful... it just seemed the easiest way to encapsulate all the thoughts I had had about your groups-rings-modules development

src/Algebra/Construct/Quotient/Ring.agda
Taneb: I personally find lambdas on the left hand side of an operator (such as `_,_`) _really_ ugly, so I use the record syntax to avoid it where I can...

src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,33 @@
...
+
+open import Algebra.Module.Construct.Sub.Bimodule using (Subbimodule)
+import Algebra.Module.Construct.TensorUnit as TU
+open import Algebra.NormalSubgroup (+-group)
+open import Level
+
+------------------------------------------------------------------------
+-- Definition
+
+-- a (two sided) ideal is exactly a subbimodule of R considered as a bimodule over itself
```
jamesmckinna, line 23: I agree with @JacquesCarette on #2859 
```suggestion
-- a (two sided) ideal is exactly a sub-bimodule of R considered as a bimodule over itself
```

src/Algebra/Ideal.agda
```diff
@@ -0,0 +1,33 @@
...
+record Ideal c′ ℓ′ : Set (c ⊔ ℓ ⊔ suc (c′ ⊔ ℓ′)) where
+  field
+    subbimodule : Subbimodule {R = R} TU.bimodule c′ ℓ′
+
+  open Subbimodule subbimodule public
+
+  normalSubgroup : NormalSubgroup c′ ℓ′
+  normalSubgroup = record
```
jamesmckinna, line 31: I'd move this to `Subbimodule`, via `Sub.AbelianGroup`, as in #2859

src/Algebra/Construct/Quotient/Ring.agda
JacquesCarette: I can get on board with that. There are a variety of generalized composition combinators in `Function.Base` that might let you not use a lambda, if you feel so inclined.

src/Algebra/Construct/Quotient/Ring.agda
Taneb: I do actually want most of these to be reexported! They might not be used in this module but I deliberately make them part of this module's interface

jamesmckinna: See also the discussion https://github.com/agda/agda-stdlib/pull/2859#discussion_r2507742656

Making these names public pollutes the namespace, when they can be obtained, with the 'usual' names, by suitable `open` directives on the constructed quotient group. My personal view (with qualified names to avoid ambiguity/clash if necessary) is that the extra names are a DRY failure.

Issue #2858, opened 2025-11-04
[ design ] Should `Reasoning.Binary.Setoid setoid` be part of the public API exported by `Algebra.Properties.X`? 


jamesmckinna:
... should this be part of the API exported by `Algebra.Properties.X`? Do we ever invoke algebraic properties *not* in a context where we have access to equational reasoning wrt that algebra?

_Originally posted by @jamesmckinna in https://github.com/agda/agda-stdlib/pull/2855#discussion_r2483685387_
            
If we do this, and with #2804  in mind, we could do this right at the bottom with `Algebra.Properties.Magma`, and then re-export it right up through the hierarchy...

JacquesCarette:
Interesting question. When importing `Algebra.Properties.X`, what are you putting in the scope? Having both tools for reasoning (over an `X`) as well as a bunch of properties about `X` does make sense. And `Algebra.Properties.X` are already "fat" modules, so this further broadening of the API size does not bother me so much. [Although I'll probably want to re-compose some `Algebra.Properties.X` from smaller pieces in the future, once findability is under control.]

jamesmckinna:
NB. I'm not clear whether or not adding the `Reasoning` infrastructure would in fact be a `breaking` change?

JacquesCarette:
I'd say that normally it would not be, but if doing so causes a lot of ambiguous names downstream (because now things are imported twice), then it would be.

MatthewDaggitt:
If you exported it unopened then I don't think you'd get much savings in terms of code. If you exported it opened, I suspect that as @JacquesCarette says we would end up with a very large number of ambiguous names. Particularly, if you wanted to both open the algebraic structure of a concrete datatype (e.g. naturals) and then also wanted to perform non-standard reasoning over the structure (e.g. partial order reasoning).

jamesmckinna:
Thanks @MatthewDaggitt I had been slowly coming to the same conclusion myself (esp. wrt the setting where you have multiple reasoning environments coexisting... such as in an algebra and in some quotient of it at the same time), so happy to close this now.
Thanks to you both for spending time thinking about it though!

PR #2859, opened 2025-11-05
[ draft ] Alternative design for sub- and quotient- `(Abelian)Group`s

jamesmckinna:
Alternative to #2852 and #2854 , based on @Taneb's original design, but:
* `Normal` is now *below* `Algebra.Construct.Sub.Group`
* `Algebra.Construct.Sub.AbelianGroup`s added, and are automatically `Normal`

Hence:
* downstream, `Algebra.Module.Construct.Sub.Bimodule`s are automatically suitable as `Ideal`s
* ... *below* `Algebra.Construct.Sub.Ring`
* simplifying `Algebra.Construct.Quotient.Ring`

The high-level differences between the two approaches mostly consist in:
* this one is more aggressively optimised than @Taneb 's original in the spirit of #2391 , in terms of making large single definitions of nested `record`s, and `open`ing them to expose substructure
* in avoiding, wherever possible, redundant imports/type signatures solely for the sake of making explicit the types of 'implementation details', so that the resulting (public) API for the various constructions is (or at least aims to be) as abstract as possible from a client perspective
* exact location of concepts within the `Algebra` hierarchy, and their mutual dependency (aiming to flatten/linearise the latter, and making the former follow a more conventional inheritance based directory/sub-directory tree structure
* making the notion of 'sub-Abelian group' explicit, and using that to streamline the development of sub- and quotient- structure of richer algebras (such as `Ring`) which are based on an underlying (additive) Abelian group structure, which slightly alters the conceptual, as well as concrete dependency graph compared to Taneb's #2852 #2854 and #2855

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@ Deprecated names
 New modules
 -----------
 
-* `Algebra.Construct.Quotient.{Abelian}Group` for the definition of quotient (Abelian) groups.
+* `Algebra.Construct.Quotient.{{Abelian}Group|Ring}` for the definition of quotient (Abelian) groups and rings.
 
 * `Algebra.Construct.Sub.{Abelian}Group` for the definition of sub-(Abelian)groups.
 


src/Algebra/Construct/Quotient/Group.agda
```diff
@@ -0,0 +1,126 @@
...
+  y ⁻¹              ∎
+  where h = conjugate (g N.⁻¹) (x ⁻¹)
+
+quotientGroup : Group c (c ⊔ ℓ ⊔ c′)
+quotientGroup = record
+  { isGroup = record
+    { isMonoid = record
+      { isSemigroup = record
+        { isMagma = record
+          { isEquivalence = record
```
JacquesCarette, line 81: I know this is just an experiment, but I'd certainly wish that this `IsEquivalence` record was pulled out.

src/Algebra/Construct/Sub/Group/Normal.agda
```diff
@@ -0,0 +1,32 @@
...
+open import Level using (suc; _⊔_)
+
+private
+  module G = Group G
+
+-- every element of the subgroup commutes in G
+record IsNormal {c′ ℓ′} (subgroup : Subgroup c′ ℓ′) : Set (c ⊔ ℓ ⊔ c′) where
+  open Subgroup subgroup using (ι)
+  field
+    conjugate : ∀ n g → _
```
JacquesCarette: That `_` is perverse. Sure, it can be reconstructed -- so what? Record declarations serve as documentation too!

src/Algebra/Construct/Sub/Ring/Ideal.agda
```diff
@@ -0,0 +1,26 @@
...
+
+module Algebra.Construct.Sub.Ring.Ideal {c ℓ} (R : Ring c ℓ) where
+
+open import Algebra.Module.Construct.Sub.Bimodule using (Subbimodule)
+open import Algebra.Module.Construct.TensorUnit using (bimodule)
+open import Level
+
+------------------------------------------------------------------------
+-- Definition
+-- a (two sided) ideal is exactly a subbimodule of R considered as a bimodule over itself
```
JacquesCarette: minor: sub-bimodule please. I find 'subbimodule' very hard to parse (as a human) properly.

src/Algebra/Construct/Sub/AbelianGroup.agda
```diff
@@ -0,0 +1,48 @@
...
+    using (ι; ι-monomorphism)
+
+  abelianGroup : AbelianGroup c′ ℓ′
+  abelianGroup = record
+    { isAbelianGroup = isAbelianGroup ι-monomorphism G.isAbelianGroup }
+
+  open AbelianGroup abelianGroup public
+
+  isNormal : IsNormal subgroup
+  isNormal = record { normal = λ n → G.comm (ι n) }
```
JacquesCarette, line 58: eta contract?

JacquesCarette:
I'm sure this alternative design makes a lot of sense to you @jamesmckinna and to @Taneb , but since I don't have the original in mind, your description relative to it does not really inform me in any substantial way of the pros and cons of each.

i.e. could the PR description be made less relative to knowing the original, but rather have more information about both designs in the description please?

src/Algebra/Construct/Quotient/Group.agda
jamesmckinna: One thing that impressed me about @Taneb 's code was the extent to which, in various places, it manages to produces huge nested `record` blocks which nevertheless 'do the right thing'.

I've flip-flopped over the years between this style, and a much more incremental, define every sub`-record` at top-level, put the pieces together step-by-step style, which the library has to date favoured.

One reason (perhaps two or more!) I am now turning towards the 'one big record' style arises from #2391 : 
* a single large object, suitably re-opened as `public`, (re-)introduces all the 'correct' substructure, moreover with the 'rectified'/'official' names, rather than trying to invent proxy names for the top-level substructure in the other style
* when any client imports the new constructions, they aren't faced with a (DRY) choice between cherry-picking the ad-hoc names (which may be the 'correct' ones), or facing a clash/choicepoint between such names and those names introduced by `open` of substructures

In this instance (as perhaps elsewhere...) the definition of distinct `isEquivalence : IsEquivalence ...` fields at top-level is starting to emerge as an anti-pattern for me: once an `Algebra.Structures.IsX` object is in scope, or has been defined, then `isEquivalence` is available as a canonical projection from that, and indeed that should be the preferred mode-of-access (via `open` if necessary) for a substructure which exists solely
* to export canonical names `refl`, `sym`, `trans` for the properties, and their derived forms
* to construct `Setoid`s from which `Relation.Binary.Reasoning.Setoid` syntax may then be brought into scope
* ...

... much of which functionality/behaviour can be achieved at a higher-level of the nesting hierarchy by suitable re-organisation of the `Algebra.Properties.X` hierarchy #2804 #2858 etc.

src/Algebra/Construct/Sub/Group/Normal.agda
jamesmckinna: Thanks for "perverse"... ;-)

I surely agree that 
* `record` declarations should be self-documenting 
but, here, I think that
* an-line comment would here do as much work in explaining why a name isn't necessary...
* ... rather than, as now, having to introduce a private manifest field defining a `module` (or else expose a new name for a subfield of the `subgroup` object), solely for the sake of projecting out the name of its `Carrier`.

So... at rock-bottom, I chose this design as a optimisation for space (as well as that of *my* cognitive load, but it seems YMMV) and like all such, doing this prematurely might be harmful.

But as with the comment above about `isEquivalence`, I am starting to find that when we create/`open` substructures of this sort, it is usually in a context in which the 'missing'/'hidden' name of the type is already in scope *with another pre-existing name* which is more salient in that deployment context.

So the `private` module name, and or that of its sub-fields, is being created *solely* for the purpose of documentation.

Accordingly, in this setting, I made a design choice to avoid committing to creating such an 'ephemeral' name. But an inline comment should, I think, solve this problem?

I'd still be OK if you disagreed with my above thought process (which should, in any case, be documented towards any eventual codification of a `library-design` philosophy ;-)), and I guess part of the point of offering an 'alternative' design for these fundamantal structures was to create an opportunity for debate around such questions.

src/Algebra/Construct/Sub/Ring/Ideal.agda
jamesmckinna: Inherited from the original, but I whole-heartedly agree!

src/Algebra/Construct/Sub/AbelianGroup.agda
jamesmckinna: Yes! In that, it is itself an eta-contraction of the original
```agda
isNormal = record { normal = λ n g → G.comm (ι n) g }
```
Is it harmful to have done this?

Or are you asking whether the definition could be contracted further? Perhaps, but only via one of those odd `Function.Base` composition operators which compose only in one argument position of an `n+2`-ary function, which I find obscure rather than illuminate (my point-free fu is not always as strong as it might be).

jamesmckinna:
I think that the best answer to the above comment might be to give a more detailed commentary in the source? (See ongoing/recent commits)

I think that this is probably more portable/surveyable than a complex interweaving of the GitHub comment trail? This still seems to be a major bugbear in using git to debate alternatives, and to document discussion around such things, as we have observed in the past, elsewhere in this project.

src/Algebra/Construct/Sub/Group.agda
```diff
@@ -0,0 +1,101 @@
...
+-- the typechecker by offering instead a placeholder, in exactly the same way
+-- that `domain` exists as a named field solely to express the dependency
+-- in the type of `ι-monomorphism`. In that sense, it is an ephemeral form
+-- derived from `domain`, which moreover will *never* play a role in client
+-- uses of this module (eg exemplarily in `Algebra.Construct.Sub.Group.Normal),
+-- except as a typing constraint on any application of `ι`, a constraint
+-- which is already inherited (by definitional equality!) from the type of
+-- the `⟦_⟧` parameter to `IsXHomomorphism`...
+
+    ι              : _ → G.Carrier -- where _ = RawGroup.Carrier domain
```
jamesmckinna, line 67: A last thought on this (or two, depending on how you count; maybe it's 3/2 of a thought): either of the following would also typecheck:
```suggestion
    ι              : _ → _ 
```
or
```suggestion
    ι              : _ -- where _ = RawGroup.Carrier domain → G.Carrier
```

src/Algebra/Construct/Quotient/Group.agda
JacquesCarette: Defining every sub-record independently also gets out of hand. I've settled on an approach more 'in the middle' where some particularly re-usable records are given top-level names.

The big record always exists, so users can choose that one if that's what they want, or the big one and `open` it.

I guess I was thinking of the "developer API" here rather than the "user API". So I think that's where our difference of opinion comes from, we had different modes of use in mind.

This is where natural language *and* Agda are insufficiently precise.

src/Algebra/Construct/Sub/Group/Normal.agda
JacquesCarette: I think I would be sufficiently placated with an inline comment.

src/Algebra/Construct/Sub/AbelianGroup.agda
JacquesCarette: Oh, I had not noticed that this had been contracted once already! I was thinking that this would be a straightforward composition. Hmm, now I wonder if the eta expanded version isn't easier to understand!

src/Algebra/Construct/Sub/Group/Normal.agda
```diff
@@ -0,0 +1,97 @@
...
+-- to a level one *below* that of `Algebra.Construct.Sub.Group`,
+-- for the 'obvious' inheritance hierarchy 'reason' that a
+-- `NormalSubgroup` is a specialisation of the notion of `Subgroup`.
+
+-- `Normal` as a root name avoids recapitulation of the parent name
+-- `Subgroup`, but also permits/does not frustrate development of a
+-- companion `Construct` sub-hierarchy (intersections etc.) below
+-- `Algebra/Construct/Sub/Group/Normal` as a sub-directory.
+
+module Algebra.Construct.Sub.Group.Normal {c ℓ} (G : Group c ℓ)  where
```
JacquesCarette, line 24: But why is this under `Construct`? This defines what being a `Normal` `Subgroup` is -- it does not construct anything!

`Algebra.Gropu.Sub.Normal` ?

src/Algebra/Construct/Sub/Ring/Ideal.agda
```diff
@@ -0,0 +1,40 @@
...
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Ring)
+
+-- As with `Algebra.Construct.Sub.Group.Normal`, this module
+-- relocates Taneb's original `Algebra.Ideal`
+-- to a level one *below* that of `Algebra.Construct.Sub.Ring`,
+-- notwithstanding that we have yet to define `Sub.XRing`s...
+
+module Algebra.Construct.Sub.Ring.Ideal {c ℓ} (R : Ring c ℓ) where
```
JacquesCarette, line 19: Same comment about this file not containing a construction but rather a definition. And it isn't `Sub` either: an `Ideal` is defined as a property of a `Ring` R. This happens to use a `Sub-bimodule` as part of its definition, sure, but it is not a `Subring`. So I think both `Construct` and `Sub` don't belong in the name.

src/Algebra/Construct/Sub/AbelianGroup.agda
```diff
@@ -0,0 +1,78 @@
...
+-- and avoids the lemma `Algebra.NormalSubgroup.abelian⇒subgroup-normal`
+-- introduced in PR https://github.com/agda/agda-stdlib/pull/2854
+-- in favour of the direct definition of the field `normal` below.
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (AbelianGroup)
+
+module Algebra.Construct.Sub.AbelianGroup {c ℓ} (G : AbelianGroup c ℓ) where
```
JacquesCarette, line 17: This is better named: it does construct an `AbelianGroup` out of a (normal) subgroup. This seems like it is in the right place.

src/Algebra/Construct/Sub/Group.agda
```diff
@@ -0,0 +1,101 @@
...
+--
+-- Based on Nathan van Doorn (Taneb)'s original PR
+-- https://github.com/agda/agda-stdlib/pull/2852
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Group; RawGroup)
+
+module Algebra.Construct.Sub.Group {c ℓ} (G : Group c ℓ) where
```
JacquesCarette, line 14: Again: this does not `Construct` anything, it gives a definition. So `Sub` fits, `Construct` does not.

JacquesCarette (changes requested):
I do really like the extensive documentation.

I think some modules are in the 'wrong' place (see detailed comments).

jamesmckinna:
@JacquesCarette regarding the `Algebra.Construct.Sub` 'mis-identification', sadly this is also present in #2852 (now merged), so we should do some retrofitting if you want to distinguish the
* hierarchy of `SubX` *definitions* (`Magma`, `Monoid`, `Group`, `Ring`...)
* hierarchy of `SubX` *constructions* (`CommutativeMagma`, `CommutativeMonoid`, `AbelianGroup`, `CommutativeRing`...)

I guess I took at face value that we 'construct' a subgroup by supplying a monomorphism... but I see your point. It's (potentially) a lot more complicated to do so, so we perhaps need to discuss further with @Taneb on a fresh issue/PR?

src/Algebra/Construct/Sub/Group/Normal.agda
jamesmckinna: I ended up just being explicit, and saving the indirection... ;-)

src/Algebra/Construct/Sub/AbelianGroup.agda
jamesmckinna: Could be: but work on #2863 suggests that this, plus commutativity in the `Centre` constructions, might better be plumbed in by other means. Further experimentation required!

src/Algebra/Construct/Sub/Ring/Ideal.agda
jamesmckinna: Yeah, I wondered about this. 
Do we regard `NormalSubgroup`, `Ideal` etc. as 'special' kinds of `Sub`object? 
Or is the specialness that hey give rise to `Quotient`s?
I'd previously suggested to @Taneb a `Quotientable` hierarchy, and then `Quotient.X` takes an `X` and a `Quotientable.X` as parameters... but got pushback on that idea.

#2854 and #2855 put these things at the top of `Algebra`, and that doesn't seem right to me either.

src/Algebra/Construct/Sub/AbelianGroup.agda
jamesmckinna: Yes (but still: sigh).

src/Algebra/Construct/Sub/Group.agda
jamesmckinna: Inherited from https://github.com/agda/agda-stdlib/pull/2852 so we've got some retrofitting to do...

src/Algebra/Construct/Sub/Group/Normal.agda
jamesmckinna: Well, it's under `Sub.Group`... so if that moves, so too will/should this.

JacquesCarette:
Re: Algebra.Construct.Sub 'mis-identification', and #2852 -- those are not yet released, so there is still time!

And yes, I hope that @Taneb will chyme in soon. I think it is worthwhile to use different terminology to indicate "definition" and "construction". I was happy to see that in Voevodsky's later work, he was quite careful in labelling things, including making this difference.

src/Algebra/Construct/Sub/Ring/Ideal.agda
JacquesCarette: Being `Normal` is a property (of a `Subgroup`). An `Ideal` is different because it involves structure. I think many (including Gonthier) have established that the classical mathematician's view of `Sub` is too restricted, particularly when it comes to formalization in type theory. [A subX should really be an X in its own right, with a monomorphism into the thing that it is a `Sub` of.]

jamesmckinna: But the various `Algebra.Morphism.XMonomorphism` modules indeed show how to construct precisely the relevant '`X` in its own right' out of the monomorphism witnessing being a 'sub`X`'...?

JacquesCarette: Not saying they don't, saying we should take that seriously throughout.

PR #2863, opened 2025-11-06
[ add ] `Centre` of a `Group` etc.

jamesmckinna:
Modulo spelling... ;-)

Currently: blocked on #2854 / #2859 
Need to rebase against #2854 ?

TODO: add/refactor to use
- [x] `Semigroup`
- [x] `Monoid`
- [ ] `Ring`
- [ ] ...
- [x] `Magma` seems not to make sense, because no `assoc`, but definition of `Central`/`Center` make sense without?

UPDATED: refactor to make use of #2872 #2873

### DIFF
--- a/src/Algebra/Construct/Centre/Magma.agda
+++ b/src/Algebra/Construct/Centre/Magma.agda
@@ -45,7 +45,7 @@ _≈_ = G._≈_ on ι
 ι-isRelHomomorphism : IsRelHomomorphism _≈_ G._≈_ ι
 ι-isRelHomomorphism = record { cong = id }
 
-ι-isRelMonomorphism : IsRelMonomorphism _≈_ G._≈_ ι
+ι-isRelMonomorphism : IsRelMonomorphism _≈_ _ _
 ι-isRelMonomorphism = record
   { isHomomorphism = ι-isRelHomomorphism
   ; injective = id
--- a/src/Algebra/Construct/Centre/Monoid.agda
+++ b/src/Algebra/Construct/Centre/Monoid.agda
@@ -50,12 +50,6 @@ domain = record { RawMagma Z.domain ; ε = ε }
 monoid : Monoid _ _
 monoid = Submonoid.monoid record { ι-monomorphism = ι-isMonoidMonomorphism }
 
-
-------------------------------------------------------------------------
--- Public exports
-
-module ι = IsMonoidMonomorphism ι-isMonoidMonomorphism
-
 commutativeMonoid : CommutativeMonoid _ _
 commutativeMonoid = record
   { isCommutativeMonoid = record
@@ -63,6 +57,12 @@ commutativeMonoid = record
     ; comm = ∙-comm
     }
   } where open Monoid monoid using (isMonoid)
+
+
+------------------------------------------------------------------------
+-- Public exports
+
+module ι = IsMonoidMonomorphism ι-isMonoidMonomorphism
 
 Z[_] = commutativeMonoid
 
--- a/src/Algebra/Construct/Centre/Semigroup.agda
+++ b/src/Algebra/Construct/Centre/Semigroup.agda
@@ -46,7 +46,7 @@ central (g ∙ h) = λ k → begin
 domain : RawMagma _ _
 domain = record { _≈_ = Z._≈_; _∙_ = _∙_ }
 
-ι-isMagmaMonomorphism : IsMagmaMonomorphism domain G.rawMagma ι
+ι-isMagmaMonomorphism : IsMagmaMonomorphism domain _ _
 ι-isMagmaMonomorphism = record
   { isMagmaHomomorphism = record
     { isRelHomomorphism = Z.ι.isHomomorphism
@@ -61,12 +61,6 @@ submagma = record { ι-monomorphism = ι-isMagmaMonomorphism }
 semigroup : Semigroup _ _
 semigroup = Subsemigroup.semigroup submagma
 
-
-------------------------------------------------------------------------
--- Public exports
-
-module ι = IsMagmaMonomorphism ι-isMagmaMonomorphism
-
 commutativeSemigroup : CommutativeSemigroup _ _
 commutativeSemigroup = record
   { isCommutativeSemigroup = record
@@ -74,5 +68,11 @@ commutativeSemigroup = record
     ; comm = ∙-comm
     }
   } where open Semigroup semigroup using (isSemigroup)
+
+
+------------------------------------------------------------------------
+-- Public exports
+
+module ι = IsMagmaMonomorphism ι-isMagmaMonomorphism
 
 Z[_] = commutativeSemigroup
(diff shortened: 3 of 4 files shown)


JacquesCarette:
As a Canadian whose idea of spelling is an ungodly mixture of American and English, I will refrain on opining between `Centre` and `Center`. (But it is *colour*, for sure.)

src/Algebra/Construct/Centre/Magma.agda
```diff
@@ -0,0 +1,44 @@
...
+--
+-- Yet a Magma *is* sufficient to define such a type, and thus can be
+-- inherited up through the whole `Algebra.Construct.Centre.X` hierarchy.
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles using (Magma)
+
+module Algebra.Construct.Centre.Magma {c ℓ} (G : Magma c ℓ) where
```
JacquesCarette, line 18: semi-facetiously: `Magma` or `RawSemigroup` ?

But actually: one has the data of the equivalence being a congruence, the other does not. So if you're going to go 'minimal', maybe `RawMagma` rather than `Magma`?

src/Algebra/Construct/Centre/Monoid.agda
```diff
@@ -0,0 +1,68 @@
...
+--
+-- Definition of the centre of a Group
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+open import Algebra.Bundles
+  using (CommutativeMonoid; Monoid; RawMonoid; RawMagma)
+
+module Algebra.Construct.Centre.Monoid {c ℓ} (G : Monoid c ℓ) where
```
JacquesCarette, line 12: Comments say `Group`, code says `Monoid` ?

src/Algebra/Construct/Centre/Semigroup.agda
```diff
@@ -0,0 +1,74 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Definition of the centre of a Group
```
JacquesCarette: `Semigroup`

src/Algebra/Construct/Quotient/AbelianGroup.agda
```diff
@@ -0,0 +1,35 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Quotient Abelian groups
```
JacquesCarette: Quotient *by a subgroup*.

src/Algebra/Construct/Quotient/Ring.agda
```diff
@@ -0,0 +1,76 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Quotient rings
```
JacquesCarette: *by an ideal*

jamesmckinna:
> I find the code duplication between the definitions of `Centre` appalling. It may well be Agda's fault.

(not sure I quite understand?)

Actually, I was trying to avoid code duplication, by exporting `Center` etc. publicly through the hierarchy. Is there an alternative approach? When I started with `Group`, everything was in one place, so the 'inheritance' hierarchy wasn't clear. What I've tried to do here is
* identify the 'local' increment at each stage (I could even do this in a `private` block, to avoid export...?)
* identify the global increment for export at each stage wrt the `*-IsHomomorphism` properties and bundles

I welcome suggestions for a better design! (Maybe my modelling skills need sharpening for the `Algebra` hierarchy?)

The thing which makes me sad is re-duplication of the `injective` proofs... but that could be fixed with a suitably shadow hierarchy of smart constructors for `*-IsMonomorphism`, ... which we haven't developed yet. Maybe we should (sigh...).

UPDATED: I now think I agree, and it seems to be express itself in my having to re-export not only `Center` etc., but also the `∙-comm` proof as well, when I'd rather be able to inherit the `commutativeY` sub-bundle straightaway when defining the `commutativeX` bundles, but that may be because the `SubX` hierarchy isn't quite right yet, either... Advice welcome!

src/Algebra/Construct/Centre/Magma.agda
jamesmckinna: Well, OK... I dithered *a lot* about whether to even have an entry at this level of the hierarchy... but it seemed the best place to 'hide' the definition of `Center`, so that it is never subsequently explicit referred to except via its `ι; central; ι-isXMonomorphism` interface.

But in the absence of associativity, it is ridiculous at some level to have this. Above `Magma` (ie `Semigroup` and beyond), we do need `assoc`, so it doesn't really make sense to avoid also `cong` etc.? But maybe we should consider 'central' notions only wrt the underlying `Raw` structure throughout? is that what you're suggesting? 

As for `RawSemigroup`, well we don't have it, because that just 'is' (would be) `RawMagma` (as in the `IsXHomomorphism` hierarchy, which only uses `IsRawX` structures to drive the definitions, and rightly so).

But, yes: `draft`, so come back at me down the road!

src/Algebra/Construct/Centre/Monoid.agda
jamesmckinna: Lazy cut-n-pastafication ;-)

src/Algebra/Construct/Centre/Semigroup.agda
jamesmckinna: Ditto.

src/Algebra/Construct/Quotient/AbelianGroup.agda
jamesmckinna: Yes! But at the moment, this is based on my #2859 so the architecture is slightly different to the now-merged #2852 and I'll wait for #2855 before refactoring/adding quotients for Abelian groups ... and while I'm happy enough with #2852 I think that my approach in #2859 is a preferable path for subgroups-of-Abelian-groups-are-Normal to that of #2855 ...

... so maybe what I should do is lift out `Centre`s to a stand-alone PR (but then that blocks on *some* version of `NormalSubgroup`... but this is still a draft, so...

src/Algebra/Construct/Quotient/Ring.agda
jamesmckinna: Ditto. These comments really belong on whatever version of `Quotient`s we end with, but sure.

jamesmckinna:
> As a Canadian whose idea of spelling is an ungodly mixture of American and English, I will refrain on opining between `Centre` and `Center`. (But it is _colour_, for sure.)

Busted!

As a UK-English user, I've embodied by cultural preference in the *folder*/module-name hierarchy, while in deference to US-English for mathematical concepts, I have the formal object `Center`... which I then try to avoid referring to in terms of the public exported API ;-)

jamesmckinna:
Looping back to @JacquesCarette 's
> I find the code duplication between the definitions of `Centre` appalling. It may well be Agda's fault.

There may be a way out of this via #2876 and keeping all the definitions inside one (nested) module, so the that 'nested/iterated' re-export is less painful.

I think that means that I should return to this later, once some of the other infrastructural PRs have landed.

JacquesCarette:
In a sense, the largest amount of duplication is in the setup code. So many lines of identical prelude for eventually saying so little. That "little" does need to be said!

src/Algebra/Construct/Centre/Magma.agda
JacquesCarette: This isn't the best place to have this discussed (it's actually quite non-trivial but made apparent in this code): where do definitions belong? Should they be hooked up at the lowest level of *necessary vocabulary* (i.e. `RawMagma`) or at the level where there is enough structure around so that the definition is somehow 'useful' (i.e. `Semigroup` in this case).

PR #2866, opened 2025-11-08
[ add ] properties of `Relation.Unary` adjoints

jamesmckinna:
Follows on from #2840 . 

Tweaks some of the text introduced there.

### DIFF
--- a/src/Relation/Unary.agda
+++ b/src/Relation/Unary.agda
@@ -261,21 +261,26 @@ P ⊥ Q = P ∩ Q ⊆ ∅
 _⊥′_ : Pred A ℓ₁ → Pred A ℓ₂ → Set _
 P ⊥′ Q = P ∩ Q ⊆′ ∅
 
--- Update/functorial change-of-base
-
--- The notation captures the Martin-Löf tradition of only mentioning
--- updates to the ambient context when describing a context-indexed
--- family P e.g. (_, σ) ⊢ Tm τ is
+-- Update/preimage/inverse image/functorial change-of-base
+--
+-- The notation, which elsewhere might be rendered
+-- * `f⁻¹ P`, for preimage/inverse image
+-- * `f* P`, for change-of-base/pullback along `f`
+-- captures the Martin-Löf tradition of only mentioning updates to
+-- the ambient context when describing a context-indexed family P:
+-- e.g. (_, σ) ⊢ Tm τ is
 -- "a term of type τ in the ambient context extended with a fresh σ".
 
 _⊢_ : (A → B) → Pred B ℓ → Pred A ℓ
 f ⊢ P = λ x → P (f x)
 
 -- Change-of-base has left- and right- adjoints (Lawvere).
+--
 -- We borrow the 'diamond'/'box' notation from modal logic, cf.
 -- `Relation.Unary.Closure.Base`, rather than Lawvere's ∃f, ∀f.
--- In some settings, the left adjoint is called 'image' or 'pushforward',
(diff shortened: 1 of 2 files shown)


src/Function/Related/TypeIsomorphisms.agda
```diff
@@ -360,6 +361,6 @@ Related-cong {A = A} {B = B} {C = C} {D = D} A≈B C≈D = mk⇔
 -- restriction that the quantified variable is equal to the given one
 
 ∃-≡ : ∀ (P : A → Set b) {x} → P x ↔ (∃[ y ] y ≡ x × P y)
-∃-≡ P {x} = mk↔ₛ′ (λ Px → x , refl , Px) (λ where (_ , refl , Py) → Py)
+∃-≡ P {x} = mk↔ₛ′ (⟨ id ⟩⊢⁻ id) (⟨ id ⟩⊢⁺ id)
```
JacquesCarette, line 364: Nice. I wish the proofs could get a similar treatment!

src/Relation/Unary.agda
```diff
@@ -261,24 +261,36 @@ P ⊥ Q = P ∩ Q ⊆ ∅
 _⊥′_ : Pred A ℓ₁ → Pred A ℓ₂ → Set _
 P ⊥′ Q = P ∩ Q ⊆′ ∅
 
--- Update.
+-- Update/preimage/inverse image/functorial change-of-base
```
JacquesCarette, line 264: Nice!

src/Function/Related/TypeIsomorphisms.agda
jamesmckinna: Well, ok... but downstream?

I think when I last looked at this, the proofs for the left adjoint behave nicely, while the corresponding ones for the right need (lots of) extensionality, to the point of being unilluminating/incomprehensible.

Also, they should be generalised to
'property of `f` implies related property of adjoint diagram'
(iso; injective; surjective... adjoint), but I got a headache working out the details...

The special case `f = id` is almost too special to be worth singling out, but... baby steps. It's one reason I went back on my original thought to rewrite the lemma *statement* to use the adjoint notation... it felt like overkill for too-small stakes.

jamesmckinna: Oh, and the use of `id` to proxy for its definitionally-equal `⊆-refl` (Yoneda lemma!) is a bit cheeky, but arguably it looks worse by using the (conceptually) type-correct version, cf. the functoriality proofs, which I wrote Yoneda-style, rather than flattening out the compositions to re duplicate the pattern-matching versions which precede them...

JacquesCarette: Absolutely downstream - that's why you got an 'Approve' rather than a 'Request changes'!

The full Yoneda-style probably belongs in `agda-categories` rather than `agda-stdlib`?

jamesmckinna: Re: Yoneda. Maybe? In this instance, I think it could be handled `Setoid`-theoretically without too much pain, but perhaps you're right. Alternatively, I could try leaning back in to: how about merging more of `agda-categories` into `stdlib`... ;-)

JacquesCarette: > how about merging more of agda-categories into stdlib

still on my radar, but this term's been crazy.

jamesmckinna:
This is ready to go, so if someone is willing either to offer an additional review, or else support merging as-is, please do so!

UPDATED: merging now in the absence of further comment either way.

PR #2868, opened 2025-11-11
Share the proofs of ×-comm and ⊎-comm of type isomorphisms and add a pairing operation

shhyou:
The functions `Function.Related.TypeIsomorphism.⊎-comm` and `Data.Sum.Properties.swap-↔` are the same, thus this PR points one to the other. `Function.Related.TypeIsomorphism.∃∃↔∃∃` is similar as well.

To make things parallel, this PR keeps only the name of `Function.Related.TypeIsomorphism.×-comm` and moves its implementation to `Data.Product.Properties`.

Finally, this PR adds `_,′-↔_ : A ↔ C → B ↔ D → (A × B) ↔ (C × D)` that pairs two type isomorphisms.

(The type of `_,′-↔_` can't be expressed as `_×_ Preserves₂ _↔_ ⟶ _↔_ ⟶ _↔_` because that requires `A` and `C` to have identical `Level`s and similarly `B` and `D` would need to have the same `Level`.)

### DIFF
--- a/src/Function/Related/TypeIsomorphisms.agda
+++ b/src/Function/Related/TypeIsomorphisms.agda
@@ -26,6 +26,7 @@ open import Data.Product.Base as Product
 open import Data.Product.Function.NonDependent.Propositional
 import Data.Product.Properties as Product
 open import Data.Sum.Base as Sum
+  using (_⊎_; inj₁; inj₂; [_,_]; [_,_]′)
 import Data.Sum.Properties as Sum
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):
I was afraid (from the description) that this might make a mess of the dependency graph. But it seems not - great.

So this looks like a nice simplification.

shhyou:
@MatthewDaggitt Is this good to go?

Taneb (approved):

shhyou:
Thanks!

PR #2872, opened 2025-11-12
[ add ] `Relation.Binary.Morphism.Construct.On`

### LINKED ISSUE #2871: [ add ] `Relation.Binary.Morphism.Construct.On`
Given a `Setoid`/`IsEquivalence` bundle/structure on type `B`, and `h : A → B`, to construct
* `_≈_ : Rel A _`,
  by `= B._≈_ on h`
* `isRelHomomorphism : IsRelHomomorphism _≈_ B._≈_ h`,
  by `isRelHomomorphism = record { cong = id }`
* `isRelMonomorphism : IsRelMonomorphism _≈_ B._≈_ h`, 
  by `isRelMonomorphism = record { isRelHomomorphism = isRelHomomorphism; injective = id }`

And... that's just about it!

UPDATED: After discussion (incl. concerning what module path/name this should live under) on the associated PR #2872 , I realise that this is the germ (but only as far as constructing the mono) of the `Kernel` of `h` construction (and the Isomorphism Theorems attaching to that), so a sensible followup would be to develop `Algebra.Morphism.Construct.Kernel` (again: path/name?) on a sound footing, with this buried somewhere at the bottom of such a development... cf. #1899 etc.

jamesmckinna:
Fixes #2871 .

### DIFF
--- a/src/Relation/Binary/Morphism/Construct/On.agda
+++ b/src/Relation/Binary/Morphism/Construct/On.agda
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Construct Morphisms from a relation and a function via `_on_`
+-- Construct monomorphism from a relation and a function via `_on_`
 ------------------------------------------------------------------------
 
 {-# OPTIONS --cubical-compatible --safe #-}
@@ -30,6 +30,3 @@ isRelMonomorphism = record
   { isHomomorphism = isRelHomomorphism
   ; injective = id
   }
-
-module ι = IsRelMonomorphism isRelMonomorphism
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,8 +106,8 @@ New modules
   ```
 
 * `Relation.Binary.Morphism.Construct.On`: given a relation `_∼_` on `B`,
-  and a function `f : A → B`, lift to various `IsRelHomomorphism`s between
-  `_∼_ on f` and `_∼_`.
+  and a function `f : A → B`, construct the canonical `IsRelMonomorphism`
+  between `_∼_ on f` and `_∼_`, witnessed by `f` itself.
 
 Additions to existing modules
 -----------------------------


src/Relation/Binary/Morphism/Construct/On.agda
```diff
@@ -0,0 +1,35 @@
...
+-- The Agda standard library
+--
+-- Construct IsRelHomomorphisms from a relation and a function
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Relation.Binary.Core using (Rel)
+
+module Relation.Binary.Morphism.Construct.On
```
JacquesCarette, line 11: I was expecting a construction of something called `On`. Instead, what is really built is a `IsRelHomomorphism`. Or maybe it is the relation `_≈_` that is being 'built'? (The code and the comment and the name don't quite agree with each other.)

JacquesCarette (changes requested):

src/Relation/Binary/Morphism/Construct/On.agda
jamesmckinna: Hmmm... this won't have been the first time where I seem to have misunderstood how `Construct`s work, or what to call them: here
* I picked the `On` name, because the salient underlying relation is `_∼_ on f`
* ... which in turn is treated systematically in `Relation.Binary.Construct.On` in terms of constructions of properties, structures, and bundles
* ... and here the 'construction' is of the `Morphism`s between `_∼_ on f` and `_∼_`

I've tweaked the comment text, but suggestions welcome as to how to reconcile any remaining incoherence between this and the code and name which you may identify...

jamesmckinna: Perhaps this should be added as additional (`Morphism.`)`Structure` in `Relation.Binary.Construct.On` instead, then?

src/Relation/Binary/Morphism/Construct/On.agda
```diff
@@ -0,0 +1,35 @@
...
+isRelHomomorphism : IsRelHomomorphism _≈_ _∼_ f
+isRelHomomorphism = record { cong = id }
+
+isRelMonomorphism : IsRelMonomorphism _≈_ _∼_ f
+isRelMonomorphism = record
+  { isHomomorphism = isRelHomomorphism
+  ; injective = id
+  }
+
+module ι = IsRelMonomorphism isRelMonomorphism
```
MatthewDaggitt: This looks a little non-standard. Do we create such a named module for any other structures elsewhere in the library?

src/Relation/Binary/Morphism/Construct/On.agda
MatthewDaggitt: I agree with @jamesmckinna that `On` is the consistent name. I also think that it should go here rather than under `Relation.Binary.Construct.On`

src/Relation/Binary/Morphism/Construct/On.agda
jamesmckinna: No, I think you're right, this doesn't belong here (it was factored out as part of a larger design for #2863 , where I was trying to re-export only this module, in some forlorn DRY hope that from it all the other pieces could be reconstructed... will delete!)

jamesmckinna:
@JacquesCarette hopefully we can merge now, if you're happy with the latest round of tweaks.

As with my updated comment on #2871 I could imagine this getting refactored as part of a development of `Kernel`s... but downstream!

PR #2873, opened 2025-11-12
[ add ] `Algebra.Definitions.Central` plus `Consequences` for `Identity` and `Zero`

jamesmckinna:
Lifted out of #2863 for low-effort review and merging?

### DIFF
--- a/src/Algebra/Definitions.agda
+++ b/src/Algebra/Definitions.agda
@@ -50,7 +50,8 @@ Associative _∙_ = ∀ x y z → ((x ∙ y) ∙ z) ≈ (x ∙ (y ∙ z))
 Commutative : Op₂ A → Set _
 Commutative _∙_ = ∀ x y → (x ∙ y) ≈ (y ∙ x)
 
--- An element is called `Central` for a binary operation if it commutes with all other elements.
+-- An element is called `Central` for a binary operation
+-- if it commutes with all other elements.
 Central : Op₂ A → A → Set _
 Central _∙_ x = ∀ y → (x ∙ y) ≈ (y ∙ x)
 


CHANGELOG.md
```diff
@@ -118,12 +118,21 @@ Additions to existing modules
   ```agda
   binomial-expansion : Associative _∙_ → _◦_ DistributesOver _∙_ →
     ∀ w x y z → ((w ∙ x) ◦ (y ∙ z)) ≡ ((((w ◦ y) ∙ (w ◦ z)) ∙ (x ◦ y)) ∙ (x ◦ z))
+  identity⇒central   : Identity e _∙_ → Central _∙_ e
```
JacquesCarette, line 121: why does this appear twice in the `CHANGELOG`?

src/Algebra/Definitions.agda
```diff
@@ -50,6 +50,9 @@ Associative _∙_ = ∀ x y z → ((x ∙ y) ∙ z) ≈ (x ∙ (y ∙ z))
 Commutative : Op₂ A → Set _
 Commutative _∙_ = ∀ x y → (x ∙ y) ≈ (y ∙ x)
 
+Central : Op₂ A → A → Set _
```
JacquesCarette, line 55: Maybe a good opportunity to leave a comment to help readers?
```suggestion
-- An element is called `Central` for a binary operation if it commutes with all other elements.
Central : Op₂ A → A → Set _
```

CHANGELOG.md
jamesmckinna: Perhaps this is a misunderstanding on my part, but the separation of `Algebra.Consequences` into `Base`, `Setoid`, and `Propositional`, means that if we add a new lemma to (the API offered by) `Setoid`, then we *also* add a new lemma to the `propositional` API, so each one gets a `CHANGELOG` entry?

src/Algebra/Definitions.agda
jamesmckinna: Nice text, and nice to offer more than bare definitions to the reader. Happy to add!

CHANGELOG.md
jamesmckinna: Tangentially, hence #2874 ... which I stumbled over while writing this PR.

jamesmckinna:
Hmmm... possible (downstream?) refactoring opportunity for `Algebra.Properties,Monoid` to replace
```agda
ε-comm : ∀ a → a ∙ ε ≈ ε ∙ a
ε-comm a = trans (identityʳ a) (sym (identityˡ a))
```
with 
```agda
ε-central : Central _∙_ ε
ε-central = Consequences.identity⇒central identity
```
instead?

CHANGELOG.md
JacquesCarette: Ah, I see the reason for this duplication. Now makes sense.

PR #2884, opened 2025-11-19
[ fix ] the test runner for v2.0-joss-submission

gallais:
Unfortunately we never worried about pinning down the versioning story
and agda makes that quite difficult.

I am hand-writing cabal files for these. We should then port these changes
over to the main branch, with updated bounds to be able to use more modern
versions of GHC & accompanying packages.

* [X] Get the test runner to build & run
         This now works when calling `AGDA_EXEC=agda-2.6.4 GHC_EXEC=ghc-9.2.8 make testsuite`
* [x] Fix all the tests (probably need the same GHC-pinning & test-specific `cabal` file rigmarole)
* [x] Document the process
* [x] Use this new approach for the CI

### DIFF
--- a/tests/_config/config.sh
+++ b/tests/_config/config.sh
@@ -7,7 +7,7 @@
 #
 # Usage: . ../../config.sh
 
-set -eux
+set -eu
 
 # Ugh, paths are relative to the script sourcing this file!
 . ../../_config/version-numbers.sh


MatthewDaggitt:
Wow this looks hairy. Thanks so much @gallais! Let me know when we can merge this in...

gallais:
I'll polish it this afternoon (specifically the part documenting what needs to
be changed after releases / changes in the dependencies) and then will mark
it ready for review.

gallais:
CI work is hell

Unfortunately did not have time to finish it today. I'm busy tomorrow & then
will be traveling so this may not come to a conclusion before early / mid December.

gallais:
Nevermind, it looks like continuing to hack while watching Limmy's show did deliver! 🥳

jamesmckinna:
Heroic refactoring!
Hopefully re-exportable going forward?
And... who knew that a 3pp. placeholder documentation effort would end up consuming *so* many resources!? 
(:sadface:, :facepalm: etc.)

gallais:
> Hopefully re-exportable going forward?

Yes, I tried to design it so that it could be easily upgraded & extended to the new tests
once cherry-picked into `master`.

> consuming so many resources

I think it was a good thunk-forcing function for stuff we should have done a while ago
(in this instance actually figuring out how to reliably build these examples with external
dependencies).

jamesmckinna:
Re: resources. I'd been thinking of the human ones, much more than the machine ones, but yes... ;-)

MatthewDaggitt:
Okay, have just downloaded and run this and it works for me! Merging in...

Thanks again @gallais!

PR #2883, opened 2025-11-19
[ rename/deprecate ] Rename `Algebra.Properties.Monoid.ε-comm`

jamesmckinna:
Fixes #2882 .

UPDATED: Moved issues to the OP #2882 and leaving that open after merging this...

Issue #2695, opened 2025-04-11
Implementation of consequences and lemmas for Groups


jmougeot:
As I have already started with Semigroups in PR #2688 and Monoids in PR #2692, the next step seems to be Groups.

I’m considering whether we should define specific left and right inverse properties within the Monoid structure and then generalize these in the Group structure.

jamesmckinna:
Consider looking first at `Quasigroup` (without identity) and then `Loop` (with) although I'm not entirely sure how much more is provable for each of those than is already covered under `Properties`...

... beyond the general question of considering `Solver`-like normal forms within each structure, relative to the associated `Free` instance of the structure, some of which depend on decidability and/or commutativity of the underlying `Carrier` to be effective : (I'm hazy on how well this all works out in practice)
* `List.NonEmpty` for `Semigroup` / non-empty lists-with-natural number multiplicity (run-length encoding, in the decidable case)
* `List` for `Monoid` / run-length encoded lists
* `Sign`ed lists, for `Group`/ integer-run-length-encoding (free Z-module on the generators...)
* ...

Others with greater expertise might care to weigh in here...

jamesmckinna:
@Taneb 's #2729 could indeed benefit from this, not least because `Algebra.Properties.Group` currently does *not* include `Algebra.Properties.{Semigroup|Monoid}` for the relevant sub-bundles...
... NB. possible clash/cognitive of names `cancelʳ` (from `Monoid`) and `∙-cancelʳ` (from `Group`) etc.?

JacquesCarette:
This is both a good direction to take, and feels like it is going to get derailed right away. Basically because Agda's features with respect to concrete Universal Algebra constructions are sub-optimal. You run into namespace management issues for theory presentations quite quickly down this path.

PR #2885, opened 2025-11-19
[ add ] `Algebra.Construct.Centre.X` of an algebra `X`, following #2863

jamesmckinna:
This PR is a clean do-over of #2863 , avoiding any dependency/coupling on ~~either #2872 or~~ #2854 .

Issues:
* keeps the 'iterated' design from #2863 ie. the centre of a `Group` is defined in terms of the centre of its underlying `Monoid`, iterated down as far as a bare ~~`Center`~~ `Centre` type based on the (components of) an underlying `RawMagma`.
* doesn't explicitly plumb in the construction of `Sub.X`  (eg as in #2859 ) until the discussion on that PR and #2854 stabilises with an agreed hierarchy for such sub-algebras.
* can (easily) be refactored when and if #2872 and/or #2883 get merged, etc. UPDATED: DONE
* there's probably a lot more that can be done to simplify/refactor `Algebra.Construct.Centre.Ring` (and develop the infrastructure for the 'missing' intermediate algebras...)

NB. the hierarchy of `IsXMonomorphism` structures unfortunately (;-)/!?) goes against the grain of how these things are all built up successively here, so downstream we might think about refactoring that hierarchy, or else introducing more 'smart' constructors for such things.


src/Algebra/Construct/Centre/Group.agda
```diff
@@ -0,0 +1,85 @@
...
+  { isAbelianGroup = record
+    { isGroup = isGroup isGroupMonomorphism G.isGroup
+    ; comm = ∙-comm
+    }
+  }
+
+Z[_] = abelianGroup
+
+{-
+  normalSubgroup : NormalSubgroup _ _
```
JacquesCarette: delete

src/Algebra/Construct/Centre/Group.agda
```diff
@@ -0,0 +1,85 @@
...
+open import Algebra.Construct.Centre.Monoid G.monoid as Z public
+  using (Center; ι; ∙-comm)
+
+-- Now, can define a commutative sub-Group
+
+_⁻¹ : Op₁ Center
+g ⁻¹ = record
+  { ι = ι g G.⁻¹
+  ; central = λ k → ∙-cancelʳ (ι g) _ _ $ begin
+      (ι g G.⁻¹ G.∙ k) G.∙ (ι g) ≈⟨ G.assoc _ _ _ ⟩
```
JacquesCarette: these assoc/cong, cong/assoc lines are exactly what the push/pull combinators from `agda-categories` were all about. They help take this proof from 7 lines down to 4. This comment is more of an educational comment than a "please change this".

JacquesCarette (changes requested):
Only change needed: delete commented out code.

src/Algebra/Construct/Centre/Group.agda
jamesmckinna: Of course! I hadn't previously appreciated 'the point' of these combinators (other than in the abstract), and hadn't studied their use in anger, cf. my interpolation of `Quasigroup`/`Loop` properties into those for `Group`. Sheer intellectual blinkers on my part in writing the 7 lines 'from first principles' ... so do please feel free to offer the 4 line proof as a GitHub `suggestion`!

Elsewhere, I have echoing on my mind an observation made somewhere else in GitHub (how to find such things!?) that the `Group` case can be further streamlined by observing something like (is this true? scratches head for a moment): any `MonoidHomomorphism` between (the underlying `Monoid`) `Group`s must, in fact, be a `GroupHomomorphism`, also by a 'cancellability' argument...?

src/Algebra/Construct/Centre/Group.agda
jamesmckinna: yes, I'd left these in a stubs in case #2854 landed first!

src/Algebra/Construct/Centre/Group.agda
jamesmckinna: Ah... absent the `Algebra.Properties.Group` refactoring #2695 along the lines already introduced by you and your students for `Semigroup` and `Monoid`... maybe they better should go there, and we come back and refactor later?

jamesmckinna: As for 'educational': thanks for the nudge! Latest commit simplifies...

jamesmckinna: But in the spirit of the above comment (lifted out as issue #2889 now), and knowing that `ι` defines a monoid *mono*morphism, can we further simplify the construction of `Centre.Group`? Or is that the 'real' semantic content of the appeal to `GroupMonomorphism.isGroup` as the 'smart constructor' I'm looking for?

JacquesCarette (approved):

jamesmckinna:
This is ready to go, so if someone is willing either to offer an additional review, or else support merging as-is, please do so!

JacquesCarette:
I guess we need to explicitly ping people like @Taneb and @MatthewDaggitt since I've already done my part!

src/Algebra/Construct/Centre/Semigroup.agda
```diff
@@ -0,0 +1,75 @@
...
+isMagmaMonomorphism = record
+  { isMagmaHomomorphism = isMagmaHomomorphism
+  ; injective = id
+  }
+
+-- And hence a CommutativeSemigroup
+
+commutativeSemigroup : CommutativeSemigroup _ _
+commutativeSemigroup = record
+  { isCommutativeSemigroup = record
```
gallais, line 69: `isCommutativeSemigroup` should be a separate toplevel definition

src/Algebra/Construct/Centre/Group.agda
```diff
@@ -0,0 +1,74 @@
...
+
+isGroupMonomorphism : IsGroupMonomorphism domain G.rawGroup ι
+isGroupMonomorphism = record
+  { isGroupHomomorphism = isGroupHomomorphism
+  ; injective = id
+  }
+
+abelianGroup : AbelianGroup _ _
+abelianGroup = record
+  { isAbelianGroup = record
```
gallais, line 72: Same here, we should have a separate `isAbelianGroup` definition.

src/Algebra/Construct/Centre/Monoid.agda
```diff
@@ -0,0 +1,67 @@
...
+isMonoidMonomorphism = record
+  { isMonoidHomomorphism = isMonoidHomomorphism
+  ; injective = id
+  }
+
+-- And hence a CommutativeMonoid
+
+commutativeMonoid : CommutativeMonoid _ _
+commutativeMonoid = record
+  { isCommutativeMonoid = record
```
gallais, line 63: To be lifted to the toplevel.

src/Algebra/Construct/Centre/Ring.agda
```diff
@@ -0,0 +1,107 @@
...
+  ; _+_ = _+_
+  ; _*_ = _*_
+  ; -_ = -_
+  ; 0# = 0#
+  ; 1# = 1#
+  } where open RawMonoid Z.domain renaming (ε to 1#; _∙_ to _*_)
+
+isRingHomomorphism : IsRingHomomorphism domain R.rawRing ι
+isRingHomomorphism = record
+  { isSemiringHomomorphism = record
```
gallais: Same with all of these nested structures.

gallais (changes requested):

src/Algebra/Construct/Centre/Center.agda
```diff
@@ -0,0 +1,39 @@
...
+open import Algebra.Definitions _∼_ using (Central)
+open import Function.Base using (id; _on_)
+open import Level using (_⊔_)
+import Relation.Binary.Morphism.Construct.On as On
+
+
+------------------------------------------------------------------------
+-- Definitions
+
+record Center : Set (c ⊔ ℓ) where
```
Taneb: We generally use UK spelling... so `Centre`?

Taneb: The module description says "centre", and the module name uses _both_ spellings. We should definitely fall on one side or the other

src/Algebra/Construct/Centre/Semigroup.agda
jamesmckinna: OK, I guess.
I think I had been thinking that any client module of `Centre.X` would (should?) `open X Z[...]` to expose its fields (cf. #2255 ), so that I didnt feel the need to do so explicitly here, but, fine, happy to do so...

... at the cost of quite a bit more involved `import` preamble...

... which really doesn't seem as though it adds anything but noise?

src/Algebra/Construct/Centre/Center.agda
jamesmckinna: Yes, I was conflicted over this, so hedged my bets by using British (preferred) for module names, but US for the underlying `Carrier` type, in case someone wanted to push the needle towards that spelling... ;-). Will fix.

src/Algebra/Construct/Centre/Ring.agda
jamesmckinna: Sigh... but yes, OK.

Oh... but no. It turns out that `Algebra.Structures.IsRing`/`Algebra.Bundles.Ring` still don't quite export enough substructures/bundles (in this instance `RawSemiring` among other things), and I'd rather deal with that downstream.

Plus: This PR doesn't consider `Centre` constructions for all the intermediate algebraic structure between `Group` and `Ring`... so that should all be handled... separately!?

jamesmckinna:
So hopefully I've now dealt with the feedback above, EXCEPT:
* trying to unpack `isRingHomomorphism` exposes that the exports of `Ring` fail to capture *all* the necessary nested records
* that I'm *not* convinced that writing out all the sub-records makes sense (discuss?)
* BUT that it might also be useful to explicitly export a bundled `XMonomorphism` from `Z[ X ]` to `X` (but we've never quite resolved to date where the unique morphisms arising from `Algebra.Construct.Y` should live?)

I suggest it's better to tackle these downstream!

UPDATED: I've now been persuaded to take a more maximalist approach to exports ... not clear to me, until these constructions are exercised in anger, that we see exactly what is and is not necessary here.

src/Algebra/Construct/Centre/Semigroup.agda
jamesmckinna: UPDATED: actually I think I should have had the courage of my convictions. I'll reimplement with `open ... public`. cf. #2391

JacquesCarette (approved):
I'm still good on this. Only need @gallais to re-review.

CHANGELOG.md
```diff
@@ -181,6 +181,11 @@ New modules
   Data.List.NonEmpty.Membership.Setoid
   ```
 
+* Added tactic ring solvers for rational numbers (issue #1879):
```
MatthewDaggitt, line 184: Quoting is weird.

src/Algebra/Construct/Centre/Semigroup.agda
```diff
@@ -0,0 +1,81 @@
...
+    { isSemigroup = isSemigroup isMagmaMonomorphism X.isSemigroup
+    ; comm = ∙-comm
+    }
+  }
+
+-- Public export of the sub-X-structures/bundles
+
+open CommutativeSemigroup commutativeSemigroup public
```
MatthewDaggitt, line 77: Agda dev meeting, we think we should have an explicit `using` list for this and similar statements.

MatthewDaggitt (changes requested):

CHANGELOG.md
jamesmckinna: Not sure what this comment means, but maybe could/should be fixed on #2974 ?
Not actually part of this PR, IIRC?

src/Algebra/Construct/Centre/Semigroup.agda
jamesmckinna: Can @gallais and @Taneb check that I now have the correct exports listed throughout?

Taneb: @jamesmckinna what's your criterion for what to export?

jamesmckinna: @Taneb great question!

So I *think* the gist of the discussion today was to the effect that
* unrestricted `open ... public` would bring *too much* into scope
* the use of `open` was (intended) to follow #2391 in *not* creating `record`s solely for the sake of the eventual constructions, but instead to obtain them by public re-export

So here: I'm trying only to re-export those things which previously @gallais wanted me to define explicitly...
... but the example of `Ring`, where I seem to re-export a bit more than is strictly necessary by that criterion, but I myself wasn't completely sure what the right answer *should* be!?

CHANGELOG.md
jamesmckinna: This seems to be an artefact of the merge commit on March 17, so not 'properly' part of this PR...

src/Algebra/Construct/Centre/Monoid.agda
```diff
@@ -68,6 +69,7 @@ commutativeMonoid = record
 -- Public export of the sub-X-structures/bundles
 
 open CommutativeMonoid commutativeMonoid public
+  using (isCommutativeMonoid; isMonoid)
```
gallais: Should you also have (commutative?) (magma and semigroup)?

jamesmckinna: So... I certainly *could*...

I've been see-sawing over this: 
* your original requested change would have been the absolutely minimal set of exported/exportable definitions
* my pushback took the absolutely maximal approach
* I've tried to reset to your original request, as in commit 8f08ff1ec623bd66b6da25240c25268c50d51a45
* but this new suggestion (and my own for `Ring`) suggests some intermediate middle way...
* ... because the proposed additions aren't *direct* sub-bundles (such things don't exist as inherited fields) of `CommutativeMonoid`, but do need to be *created* as derived/manifest fields

That said, I'm not against the suggestion, just unclear where to stop?

And *that* said, we could always expand the `using` list in subsequent PRs downstream? And at no great additional cost... so I'd tend towards minimality now?

jamesmckinna: Oh, and then `Centre.Semigroup` would also need to re-export `magma` and `commutativeMagma` etc.

jamesmckinna: Sticking with the 'minimal' version for now.

gallais: The argument (IIUC) was that having a single definition would not make
the other substructures more difficult to access but the minimal approach
of not re-exporting (at least) the sub bundles means we have a strictly less
ergonomic module now and we're back to having to manually extract the
relevant structure from one that's too big a structure.

Compare `Data.Nat.Properties` re-exporting various bundles:
https://agda.github.io/agda-stdlib/v2.3/Data.Nat.Properties.html#24740

vs. this PR only exporting `AbelianGroup` (and opening it only to re-export
some `isX` sub-properties).

jamesmckinna: Hmmm, I think we have very different perspectives/understanding of 'the argument'.
I agree that something like `Data.Nat.Properties` builds a lot of stuff, moreover incrementally.

But these `Construct`s are really designed to do *just one thing* (well, two, depending on how you count):
* construct a unique-up-to `CommutativeX` out of given `X`, and uniformly name it `Z[ X ]`
* together with the unique-up-to `XMonomorphism` which exhibits `Z[ X ]` as a `SubX` of the given algebra `X`.

*Everything else* can be derived by client modules from these two bits of construction. 

It feels unnatural (to me) to want to have these modules be responsible for such exports in addition to the above? Maybe the right thing to do is to remove the `public` reexports as well!?

gallais: > Everything else can be derived by client modules from these two bits of construction.

I know, and this is what the stdlib modules used to do too.

And then at every use site you needed to figure out the right incantation to
manually extract the relevant substructure because the module you're invoking
takes a monoid, not an Abelian group.

As a beginner, you'd often end up with an ungodly chain of projection (because
that's what `C-c C-a` finds and you typically don't know that you can `open` a
module and obtain a lot of things not explicitly listed in the `record` definition
but nonetheless re-exported by another record deep in the hierarchy).

jamesmckinna: Hmmm, that helps me grasp your earlier appeal to UX/ergonomics. 

I think that what I *don't* want to have to do at this stage, is try to show that the underlying `CommutativeMagma` of the centre `Z[ S ]` for a `Semigroup` `S` is in fact (definitionally equal to, or only iso?) `Z[ M ]` for the underlying `Magma` of `S`... even though I think that such things could/should be provable. (Are they, question?)

So I'm reluctant to export such sub-bundles, as they muddy the water as to which `Centre` construction is being invoked where, but maybe you're right, we'll (eventually) want all of this. I think I want to hide behind a @JacquesCarette -like "this is enough for now, we can add more later (don't let making this perfect be the enemy of it already being good enough)"?

jamesmckinna: I've done this (export sub-bundles as well as sub-structures) as far as `Group`... but the export list for `Ring` was too much for me! UPDATED but I plucked up the courage, and now they're all there.

CHANGELOG.md
jamesmckinna: Now (hopefully!?) fixed on #2974 .

MatthewDaggitt: You have one set of triple quotes at the bottom and then individual single quotes on each line. I think we should remove the individual single quotes, and add another matching set of triple quotes at the top.

MatthewDaggitt (approved):

CHANGELOG.md
jamesmckinna: See above: fixed on #2974

jamesmckinna: ... and as it turns out, already fixed on `master`... !?

jamesmckinna:
@gallais back to you now...

src/Algebra/Construct/Centre/Monoid.agda
JacquesCarette: I certainly would say something like that ;)

But also: I remain of the opinion that we need minimal records for some purposes, and maximally rich versions of those same records for other purposes. Certainly not a v2.4 issue though.

gallais (approved):

PR #2891, opened 2025-11-23
Attempt at the INT construction

### LINKED ISSUE #2890: [ add ] `INT` view of `Data.Integer.Base.ℤ`
See discussion on #2879 .

Taneb:
Closes #2890 (eventually)

The proofs, while purely algebraic, are often very tedious, especially `*-cong`.

I've got basic arithmetic, ordering with addition, and divmod working!

### DIFF
--- a/src/Data/Integer/IntConstruction/DivMod.agda
+++ b/src/Data/Integer/IntConstruction/DivMod.agda
@@ -1,0 +1,108 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Division with remainder for integers constructed as a pair of naturals
+------------------------------------------------------------------------
+
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Data.Integer.IntConstruction.DivMod where
+
+open import Data.Fin.Base as Fin using (Fin)
+import Data.Fin.Properties as Fin
+open import Data.Integer.IntConstruction
+open import Data.Integer.IntConstruction.Properties
+open import Data.Nat.Base as ℕ using (ℕ)
+import Data.Nat.Properties as ℕ
+import Data.Nat.DivMod as ℕ
+open import Function.Base
+open import Relation.Binary
+open import Relation.Binary.PropositionalEquality
+open import Relation.Nullary.Decidable
+
+-- should the divisor also be an integer?
+record DivMod (dividend : ℤ) (divisor : ℕ) : Set where
+  field
+    quotient : ℤ
+    remainder : ℕ
+    remainder<divisor : remainder ℕ.< divisor
+    property : dividend ≃ ⁺ remainder + quotient * ⁺ divisor
+
+divMod : ∀ i n .{{_ : ℕ.NonZero n}} → DivMod i n
+divMod (a ⊖ b) n = record
+  { quotient = quotient
+  ; remainder<divisor = remainder<n
+  ; property = property
+  }
+  where
+    open ≡-Reasoning
+
+  -- either the actual quotient or one above it
+    quotient′ : ℤ
+    quotient′ = (a ℕ./ n) ⊖ (b ℕ./ n)
+
+    remainder′ : ℤ
+    remainder′ = (a ℕ.% n) ⊖ (b ℕ.% n)
+
+    property′ : a ⊖ b ≃ remainder′ + quotient′ * ⁺ n
+    property′ = mk≃ $ cong₂ ℕ._+_ left right
+      where
+        left : a ≡ a ℕ.% n ℕ.+ (a ℕ./ n ℕ.* n ℕ.+ b ℕ./ n ℕ.* 0)
+        left = begin
+          a                                             ≡⟨ ℕ.m≡m%n+[m/n]*n a n ⟩
+          a ℕ.% n ℕ.+ a ℕ./ n ℕ.* n                     ≡⟨ cong (a ℕ.% n ℕ.+_) (ℕ.+-identityʳ (a ℕ./ n ℕ.* n)) ⟨
+          a ℕ.% n ℕ.+ (a ℕ./ n ℕ.* n ℕ.+ 0)             ≡⟨ cong (λ z → a ℕ.% n ℕ.+ (a ℕ./ n ℕ.* n ℕ.+ z)) (ℕ.*-zeroʳ (b ℕ./ n)) ⟨
+          a ℕ.% n ℕ.+ (a ℕ./ n ℕ.* n ℕ.+ b ℕ./ n ℕ.* 0) ∎
+
+        right : b ℕ.% n ℕ.+ (a ℕ./ n ℕ.* 0 ℕ.+ b ℕ./ n ℕ.* n) ≡ b
+        right = begin
+          b ℕ.% n ℕ.+ (a ℕ./ n ℕ.* 0 ℕ.+ b ℕ./ n ℕ.* n) ≡⟨ cong (λ z → b ℕ.% n ℕ.+ (z ℕ.+ b ℕ./ n ℕ.* n)) (ℕ.*-zeroʳ (a ℕ./ n)) ⟩
+          b ℕ.% n ℕ.+ (0 ℕ.+ b ℕ./ n ℕ.* n)             ≡⟨ cong (b ℕ.% n ℕ.+_) (ℕ.+-identityˡ (b ℕ./ n ℕ.* n)) ⟩
+          b ℕ.% n ℕ.+ b ℕ./ n ℕ.* n                     ≡⟨ ℕ.m≡m%n+[m/n]*n b n ⟨
+          b                                             ∎
+
+    quotient : ℤ
+    quotient with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes _ = quotient′
+    ... | no _ = pred quotient′
+
+    remainder : ℕ
+    remainder with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes _ = a ℕ.% n ℕ.∸ b ℕ.% n
+    ... | no _ = n ℕ.∸ (b ℕ.% n ℕ.∸ a ℕ.% n)
+
+    remainder<n : remainder ℕ.< n
+    remainder<n with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes _ = ℕ.≤-<-trans (ℕ.m∸n≤m (a ℕ.% n) (b ℕ.% n)) (ℕ.m%n<n a n)
+    ... | no b%n≰a%n = ℕ.∸-monoʳ-< {o = 0} (ℕ.m<n⇒0<n∸m (ℕ.≰⇒> b%n≰a%n)) (ℕ.≤-trans (ℕ.m∸n≤m (b ℕ.% n) (a ℕ.% n)) (ℕ.<⇒≤ (ℕ.m%n<n b n)))
+
+    property : a ⊖ b ≃ ⁺ remainder + quotient * ⁺ n
+    property with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes p = ≃-trans property′ (+-cong rem-prop ≃-refl)
+      where
+        rem-prop : remainder′ ≃ ⁺ (a ℕ.% n ℕ.∸ b ℕ.% n)
+        rem-prop = mk≃ $ begin
+          a ℕ.% n ℕ.+ 0                   ≡⟨ ℕ.+-identityʳ (a ℕ.% n) ⟩
+          a ℕ.% n                         ≡⟨ ℕ.m∸n+n≡m p ⟨
+          a ℕ.% n ℕ.∸ b ℕ.% n ℕ.+ b ℕ.% n ∎
+    ... | no b%n≰a%n = ≃-trans property′ $ mk≃ $ begin
+      (w ℕ.+ x) ℕ.+ (y ℕ.+ (n ℕ.+ z))                 ≡⟨ cong ((w ℕ.+ x) ℕ.+_) (ℕ.+-assoc y n z) ⟨
+      (w ℕ.+ x) ℕ.+ ((y ℕ.+ n) ℕ.+ z)                 ≡⟨ cong (λ k → (w ℕ.+ x) ℕ.+ (k ℕ.+ z)) (ℕ.+-comm y n) ⟩
+      (w ℕ.+ x) ℕ.+ ((n ℕ.+ y) ℕ.+ z)                 ≡⟨ cong ((w ℕ.+ x) ℕ.+_) (ℕ.+-assoc n y z) ⟩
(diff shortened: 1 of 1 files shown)


src/Data/Integer/IntConstruction.agda
```diff
@@ -0,0 +1,77 @@
...
+  field
+    minuend : ℕ
+    subtrahend : ℕ
+
+infix 4 _≃_ _≤_ _≥_ _<_ _>_
+
+record _≃_ (i j : ℤ) : Set where
+  constructor mk≃
+  field
+    eq : ℤ.minuend i ℕ.+ ℤ.subtrahend j ≡ ℤ.minuend j ℕ.+ ℤ.subtrahend i
```
jamesmckinna, line 29: First observation from `Properties`: use of `sym` to flip rewrites in the RHS of such equations... so suggest instead
```suggestion
    eq : ℤ.minuend i ℕ.+ ℤ.subtrahend j ≡ ℤ.subtrahend i ℕ.+ ℤ.minuend j
```
?

src/Data/Integer/IntConstruction.agda
```diff
@@ -0,0 +1,77 @@
...
+
+pred : ℤ → ℤ
+pred (a ⊖ b) = a ⊖ ℕ.suc b
+
+0ℤ : ℤ
+0ℤ = 0 ⊖ 0
+
+1ℤ : ℤ
+1ℤ = 1 ⊖ 0
+
```
jamesmckinna, line 54: Also:
```suggestion
1ℤ : ℤ
1ℤ = 1 ⊖ 0
-1ℤ : ℤ
-1ℤ = 0 ⊖ 1
```

src/Data/Integer/IntConstruction.agda
```diff
@@ -0,0 +1,77 @@
...
+-_ : ℤ → ℤ
+- (a ⊖ b) = b ⊖ a
+
+infix 8 ⁻_ ⁺_
+
+⁺_ : ℕ → ℤ
+⁺ n = n ⊖ 0
+
+⁻_ : ℕ → ℤ
+⁻ n = 0 ⊖ n
```
jamesmckinna, line 73: The `⁻` symbol is in particular quite hard to discern as distinct form `-`. Would it be better to use `Sign` here (and reimplement the corresponding `signAbs`/`SignAbs` plumbing)?

src/Data/Integer/IntConstruction/DivMod.agda
```diff
@@ -0,0 +1,108 @@
...
+    property : a ⊖ b ≃ ⁺ remainder + quotient * ⁺ n
+    property with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes p = ≃-trans property′ (+-cong rem-prop ≃-refl)
+      where
+        rem-prop : remainder′ ≃ ⁺ (a ℕ.% n ℕ.∸ b ℕ.% n)
+        rem-prop = mk≃ $ begin
+          a ℕ.% n ℕ.+ 0                   ≡⟨ ℕ.+-identityʳ (a ℕ.% n) ⟩
+          a ℕ.% n                         ≡⟨ ℕ.m∸n+n≡m p ⟨
+          a ℕ.% n ℕ.∸ b ℕ.% n ℕ.+ b ℕ.% n ∎
+    ... | no b%n≰a%n = ≃-trans property′ $ mk≃ $ begin
```
jamesmckinna, line 88: Oh. Tempting to use a `(Semi)Ring` solver here?

src/Data/Integer/IntConstruction/IntProperties.agda
```diff
@@ -0,0 +1,72 @@
...
+
+open import Data.Integer.Base
+open import Data.Integer.IntConstruction as INT using (_≃_; mk≃)
+open import Data.Integer.IntConstruction.Tmp
+open import Data.Integer.Properties
+import Data.Nat.Base as ℕ
+open import Data.Product.Base
+open import Function.Base
+open import Relation.Binary.PropositionalEquality
+
```
jamesmckinna, line 19: is it easier to derive each of these properties from the 'adjoint equivalence'/'graph relation' forms:
```agda
fromINT i ≡ j → i ≃ toINT j
i ≃ toINT j → fromINT i ≡ j
```
(plus, perhaps, some additional lemmas to 'explain' these two graph relations)?
(I think even the `cong`s follow from this, but I'd need to check that...)

src/Data/Integer/IntConstruction.agda
Taneb: I mostly got annoyed at Agda not being able to tell if a section `(x +_)` was `_+_ x` or `x (+_)`, so I wanted different symbols for them... the superscript-minus symbol is used in APL's number syntax for a negative, and that's what inspired the use of these here. I'm not attached to them, though.

src/Data/Integer/IntConstruction/DivMod.agda
Taneb: Do those work well when I'm using monus? Not that I ever really can manage to get them to work well for me...

src/Data/Integer/IntConstruction.agda
Taneb: The reason I did it this way was so that reflexivity is exactly `refl`. But you're right, that doesn't seem to be worth the tradeoff

src/Data/Integer/IntConstruction.agda
Taneb: Easy enough, but does it pass the Fairbairn threshold? `-1ℤ = - 1ℤ`, after all

src/Data/Integer/IntConstruction/IntProperties.agda
Taneb: The second one is easy enough, it's hidden in my `surjective` proof. I'll have a go at the first one later.

As weak evidence that the `cong`s _don't_ follow from this: they're separate fields in the `IsInverse` structure in `Function.Structures`

src/Data/Integer/IntConstruction/DivMod.agda
```diff
@@ -0,0 +1,108 @@
...
+    remainder with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes _ = a ℕ.% n ℕ.∸ b ℕ.% n
+    ... | no _ = n ℕ.∸ (b ℕ.% n ℕ.∸ a ℕ.% n)
+
+    remainder<n : remainder ℕ.< n
+    remainder<n with b ℕ.% n ℕ.≤? a ℕ.% n
+    ... | yes _ = ℕ.≤-<-trans (ℕ.m∸n≤m (a ℕ.% n) (b ℕ.% n)) (ℕ.m%n<n a n)
+    ... | no b%n≰a%n = ℕ.∸-monoʳ-< {o = 0} (ℕ.m<n⇒0<n∸m (ℕ.≰⇒> b%n≰a%n)) (ℕ.≤-trans (ℕ.m∸n≤m (b ℕ.% n) (a ℕ.% n)) (ℕ.<⇒≤ (ℕ.m%n<n b n)))
+
+    property : a ⊖ b ≃ ⁺ remainder + quotient * ⁺ n
```
JacquesCarette, line 79: I wonder if we could do a giant "control inversion": the computation of `remainder` and `quotient` and this here `property` all do a case analysis *on the same thing*. Maybe we could case once and then return all 3 pieces of information?

src/Data/Integer/IntConstruction/DivMod.agda
```diff
@@ -0,0 +1,108 @@
...
+    property : dividend ≃ ⁺ remainder + quotient * ⁺ divisor
+
+divMod : ∀ i n .{{_ : ℕ.NonZero n}} → DivMod i n
+divMod (a ⊖ b) n = record
+  { quotient = quotient
+  ; remainder<divisor = remainder<n
+  ; property = property
+  }
+  where
+    open ≡-Reasoning
```
JacquesCarette, line 38: Given that all of the reasoning below is done in a ℕ context and the `+`, `*`, etc of ℤ are never used, might it make sense to qualify ℤ and not ℕ instead? That would definitely shorten a lot of things.

src/Data/Integer/IntConstruction/DivMod.agda
```diff
@@ -0,0 +1,108 @@
...
+      ((n ℕ.+ (v ℕ.∸ (v ℕ.∸ w))) ℕ.+ x) ℕ.+ (y ℕ.+ z) ≡⟨ cong (λ k → (k ℕ.+ x) ℕ.+ (y ℕ.+ z)) (ℕ.+-∸-assoc n (ℕ.m∸n≤m v w)) ⟨
+      (((n ℕ.+ v) ℕ.∸ (v ℕ.∸ w)) ℕ.+ x) ℕ.+ (y ℕ.+ z) ≡⟨ cong (λ k → (k ℕ.+ x) ℕ.+ (y ℕ.+ z)) (ℕ.+-∸-comm v (ℕ.≤-trans (ℕ.m∸n≤m v w) (ℕ.<⇒≤ (ℕ.m%n<n b n)))) ⟩
+      (((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ v) ℕ.+ x) ℕ.+ (y ℕ.+ z) ≡⟨ cong (ℕ._+ (y ℕ.+ z)) (ℕ.+-assoc (n ℕ.∸ (v ℕ.∸ w)) v x) ⟩
+      ((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ (v ℕ.+ x)) ℕ.+ (y ℕ.+ z) ≡⟨ cong (λ k → ((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ k) ℕ.+ (y ℕ.+ z)) (ℕ.+-comm v x) ⟩
+      ((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ (x ℕ.+ v)) ℕ.+ (y ℕ.+ z) ≡⟨ cong (ℕ._+ (y ℕ.+ z)) (ℕ.+-assoc (n ℕ.∸ (v ℕ.∸ w)) x v) ⟨
+      (((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ x) ℕ.+ v) ℕ.+ (y ℕ.+ z) ≡⟨ ℕ.+-assoc (n ℕ.∸ (v ℕ.∸ w) ℕ.+ x) v (y ℕ.+ z) ⟩
+      ((n ℕ.∸ (v ℕ.∸ w)) ℕ.+ x) ℕ.+ (v ℕ.+ (y ℕ.+ z)) ∎
+      where
+        w = a ℕ.% n
+        x = a ℕ./ n ℕ.* n ℕ.+ b ℕ./ n ℕ.* 0
```
JacquesCarette, line 105: can't the above proof be greatly simplified by first getting rid of those `* 0` in `x` and `y`, and then reconstructing them later?

Also, isn't it the case that z + v = b ?

src/Data/Integer/IntConstruction/IntProperties.agda
```diff
@@ -0,0 +1,72 @@
...
+fromINT-cong : ∀ {i j} → i ≃ j → fromINT i ≡ fromINT j
+fromINT-cong {a INT.⊖ b} {c INT.⊖ d} (mk≃ a+d≡c+b) = begin
+  a ⊖ b                       ≡⟨ m-n≡m⊖n a b ⟨
+  + a - + b                   ≡⟨ cong (_- + b) (+-identityʳ (+ a)) ⟨
+  (+ a + 0ℤ) - + b            ≡⟨ cong (λ z → (+ a + z) - + b) (+-inverseʳ (+ d)) ⟨
```
JacquesCarette, line 24: identity followed by inverse are exactly the (opposite direction) of the cancel lemmas.

src/Data/Integer/IntConstruction/IntProperties.agda
```diff
@@ -0,0 +1,72 @@
...
+  (+ a + (+ d - + d)) - + b   ≡⟨ cong (_- + b) (+-assoc (+ a) (+ d) (- + d)) ⟨
+  (+ (a ℕ.+ d) - + d) - + b   ≡⟨ cong (λ z → (+ z - + d) - + b) a+d≡c+b ⟩
+  (+ (c ℕ.+ b) - + d) - + b   ≡⟨ cong (_- + b) (+-assoc (+ c) (+ b) (- + d)) ⟩
```
JacquesCarette, line 27: aren't these 2 lines some kind of middle4 ?

src/Data/Integer/IntConstruction/IntProperties.agda
```diff
@@ -0,0 +1,72 @@
...
+  (+ c + (+ b - + d)) - + b   ≡⟨ cong (λ z → (+ c + z) - + b) (+-comm (+ b) (- + d)) ⟩
+  (+ c + (- + d + + b)) - + b ≡⟨ cong (_- + b) (+-assoc (+ c) (- + d) (+ b)) ⟨
+  ((+ c - + d) + + b) - + b   ≡⟨ +-assoc (+ c - + d) (+ b) (- + b) ⟩
+  (+ c - + d) + (+ b - + b)   ≡⟨ cong₂ _+_ (m-n≡m⊖n c d) (+-inverseʳ (+ b)) ⟩
```
JacquesCarette, line 31: cancel!

src/Data/Integer/IntConstruction.agda
```diff
@@ -0,0 +1,77 @@
...
+_>_ : Rel ℤ _
+_>_ = flip _<_
+
+suc : ℤ → ℤ
+suc (a ⊖ b) = ℕ.suc a ⊖ b
+
+pred : ℤ → ℤ
+pred (a ⊖ b) = a ⊖ ℕ.suc b
+
+0ℤ : ℤ
```
JacquesCarette, line 49: I wonder if it would be useful to also have non-normalized 0 and 1, i.e. `0ℤn {n} = n ⊖ n` with the n in the name being a super/sub-script? I think you are prematurely normalizing in places.

src/Data/Nat/Properties.agda
```diff
@@ -1834,6 +1834,10 @@ m≤n⇒∣m-n∣≡n∸m {n = _}     (s≤s m≤n) = m≤n⇒∣m-n∣≡n∸m
 ∣m+n-m+o∣≡∣n-o∣ zero    n o = refl
 ∣m+n-m+o∣≡∣n-o∣ (suc m) n o = ∣m+n-m+o∣≡∣n-o∣ m n o
 
+∣m+o-n+o∣≡∣m-n∣ : ∀ m n o → ∣ m + o - n + o ∣ ≡ ∣ m - n ∣
```
JacquesCarette, line 1837: maybe add parens in the lhs to make it easier to parse for humans?

JacquesCarette (changes requested):
Some are requests for changes, some are suggestions.

I have quite a few more things I'd like simplified, but I feel that maybe I should do those myself as they require some level of experimenting.

jamesmckinna:
The library always turns up surprising things: have you looked at `Relation.Binary.HeterogeneousEquality.Quotients.Examples`!?

Taneb:
Wow, I don't think I'd ever even noticed that module before!

Issue #2892, opened 2025-11-24
[ refactor/design ] left- and right- actions of normal subgroups on their ambient group


jamesmckinna:
As #2854 matures, I can't help thinking that while this definition is perfectly fine *for `Group`s*, nevertheless it still embodies a left/right handedness which we should try as far as possible to resist committing to: calling it `IsNormalʳ` for the sake of argument, and rename its fields to:
```agda
record IsNormalʳ {c′ ℓ′} (subgroup : Subgroup c′ ℓ′) : Set (c ⊔ ℓ ⊔ c′) where
  open Subgroup subgroup using (ι)
  field
    _ι∙ʳ_ : ∀ g n → typeOf n

    normalʳ : ∀ n g → ι (g ι∙ʳ n) G.∙ g G.≈ g G.∙ ι n
```
(there's some notation design to consider here!).

By symmetry we should also consider 
```agda
    _ι∙ˡ_ : ∀ n g → typeOf n
    normalˡ : ∀ n g → g G.∙ ι (n ι∙ˡ g) G.≈ ι n G.∙ g
```
and, again *for `Group`s*, there is a relationship between `g ι∙ʳ n` and `n ι∙ˡ g` (roughly: 'conjugate of the inverse is the inverse of the conjugate'), but in general, we might expect no such relationship?

I'd first thought that 'conjugate of the inverse is the inverse of the conjugate' should be reified as a manifest field (and still could: to simplify certain arguments, if nothing else), but the more I thought about it, the more I thought that 'left-normal' and 'right-normal' are free-standing concepts... ?

Cf. left- and right-actions #2350 (normal subgroups, defined by monomorphism, induce left- and right- actions on their ambient group... with certain properties...)

NB. For some reason, this hasn't been quoted as a `reference in new issue`, but/and I can develop/refine this further, so consider this issue currently as `DRAFT`... ;-)

JacquesCarette:
Why does `_ι∙ʳ_` take a `g` are all, since it is ignored?

jamesmckinna:
> Why does `_ι∙ʳ_` take a `g` at all, since it is ignored?

[...]

PR #2894, opened 2025-11-24
[ add ] left- and right- actions induced by a `Magma` homomorphism

jamesmckinna:
A simple piece of 'infrastructure' for `Algebra`.

NB. used extensively (implicitly) in #2854 etc. cf. #2892 and #2350 .

Issues:
* need to sort out the re-exporting strategy, plus suitable renaming (NOW: DONE?)

### DIFF
--- a/src/Algebra/Morphism/Structures.agda
+++ b/src/Algebra/Morphism/Structures.agda
@@ -267,13 +267,20 @@ module NearSemiringMorphisms (R₁ : RawNearSemiring a ℓ₁) (R₂ : RawNearSe
       *-homo : Homomorphic₂ ⟦_⟧ _*₁_ _*₂_
 
     open +.IsMonoidHomomorphism +-isMonoidHomomorphism public
-      renaming (homo to +-homo; ε-homo to 0#-homo; isMagmaHomomorphism to +-isMagmaHomomorphism; ⟦_⟧∙_ to ⟦_⟧+_; _∙⟦_⟧ to _+⟦_⟧)
+      renaming (homo to +-homo; ε-homo to 0#-homo
+               ; isMagmaHomomorphism to +-isMagmaHomomorphism
+               ; ⟦_⟧∙_ to ⟦_⟧+_; _∙⟦_⟧ to _+⟦_⟧
+               )
 
     *-isMagmaHomomorphism : *.IsMagmaHomomorphism ⟦_⟧
     *-isMagmaHomomorphism = record
       { isRelHomomorphism = isRelHomomorphism
       ; homo = *-homo
       }
+
+    open *.IsMagmaHomomorphism *-isMagmaHomomorphism public
+      using ()
+      renaming (⟦_⟧∙_ to ⟦_⟧*_; _∙⟦_⟧ to _*⟦_⟧)
 
   record IsNearSemiringMonomorphism (⟦_⟧ : A → B) : Set (a ⊔ ℓ₁ ⊔ ℓ₂) where
     field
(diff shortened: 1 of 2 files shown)


Taneb:
I think this is a good addition, but if these are getting reexported by the various variations of ring morphisms, they should be renamed `⟦_⟧+_` and `_+⟦_⟧` (and we should also have `⟦_⟧*_` and `_*⟦_⟧`, and probably something for module and lattice homomorphisms)

jamesmckinna:
> I think this is a good addition, but if these are getting reexported by the various variations of ring morphisms, they should be renamed `⟦_⟧+_` and `_+⟦_⟧` (and we should also have `⟦_⟧*_` and `_*⟦_⟧`, and probably something for module and lattice homomorphisms)

Latest commit adds some of these export renamings. More to follow later, or downstream... most of the multiplicative substructure isn't opened in `Structures`, and I'm not sure what is gained by trying to chase down client `open`ings... but agree that `Lattice` and `Module` might/should benefit from these things as well. 

UPDATED: added renamed exports for `Lattice` and `Module` morphisms; the multiplicative structure of ring-like morphisms doesn't actually seem ever to be opened in `stdlib`, so no renaming export changes needed, it would seem?

Taneb:
Even if they're not currently opened, it might be worth opening them specifcally to export `⟦_⟧*_` and `_*⟦_⟧`?

jamesmckinna:
> Even if they're not currently opened, it might be worth opening them specifcally to export `⟦_⟧*_` and `_*⟦_⟧`?

Good suggestion! Incoming commits...

src/Algebra/Morphism/Structures.agda
```diff
@@ -89,6 +89,12 @@ module MagmaMorphisms (M₁ : RawMagma a ℓ₁) (M₂ : RawMagma b ℓ₂) wher
     open IsRelHomomorphism isRelHomomorphism public
       renaming (cong to ⟦⟧-cong)
 
+    ⟦_⟧∙_ : A → B → B
```
JacquesCarette, line 92: Fairbairn threshold? Do we really gain anything by adding these synonyms? The code already says we lose a bit (via the need to add `hiding`).

Personally I think that `IsMagmaHomomorphism` should be more minimalistic, and we should also have accompanying "extra kit" modules that we can use when needed. I've partially refactored `agda-categories` in that direction. And, in that case, seriously shrank the size of `.agdai` files and load times. I don't expect such an effect for `Magma`-related stuff, but this more "kitchen-sink" design approach is exactly what made working with Monoidal Categories so heavy.

jamesmckinna:
Thanks @JacquesCarette for yet further ammunition in the perennial 'wide vs. narrow' API debate... but echoing comments of yours elsewhere, I don't regard the inefficiencies in Agda's allocation/usage of records in terms of memory (and hence time) as an *intrinsic* quality of any design for `stdlib`, rather that I guess I'm more willing (certainly than you) to trade off DRY (and 'standardised' vocabulary to go along with it) against YAGNI (or if you are going to need it, install an extender/connector module to afford such functionality). Of course, at least if understand your nudge correctly, you envisage these things coexisting, albeit at the cost of (proliferation and import) of such extender modules...

I (perhaps) agree that `stdlib` somehow should aim to accommodate both ends of this spectrum, but as the `Algebra.*` hierarchy develops, it does seem clear that a number of infrastructural definitional extensions of the the 'basic' signature(s) seem to emerge as pervasive once we move to developing (short) exact sequences and their associated sub-/quotient structures cf. the PRs identified in the preamble. 

Against that, I do (perhaps) accept the appeal to Fairbairn, although again, once we get to action and wreath products, I'm not so sure... these 'crossed'/'semidirect' reindexed actions seem to arise everywhere? And the programme of constructive explicitation of mathematics which our proof assistants instrument seems concomitantly to expose the need to identify the infrastructure otherwise implicit in many branches of axiomatic algebra? (Maybe not; it's a discussion point, just as this PR was in part a prompt to such discussion...)

jamesmckinna:
After sleeping on this, I'll convert to `DRAFT` in the interim, against refactoring along the lines suggested by @JacquesCarette ...

JacquesCarette:
I'm a big believer in Kolmogorov complexity (and its simpler cousin, Minimum Description Length). Practically in code bases this means that if an idiom occurs in practice quite a number of times, especially if the idiom has real semantic content over its lower level direct implementation, then it makes sense to introduce a name for that idiom.

So my instinct then is to wait until enough downstream code would be improved by this addition, rather than to 'predict' now that it will be useful later.

PR #2895, opened 2025-11-25
[ new ] unix install script for the standard library

gallais:
* [x] Start by checking the required unix utils are installed
* [x] Run through https://www.shellcheck.net/
* [x] Add documentation (including in release checklist)
* [x] Add `curl`-based command for users to run

For testing purposes, using the latest version of the script on my branch,
if you're a user of a unix-style operating system, you should be able to just
copy/paste the following in a terminal and hit return!

```
sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/gallais/agda-stdlib/refs/heads/install-script/stdlib-install.sh)"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ On Unix-style systems, it should be as simple as running the following command i
 a shell:
 
 ```shell
-sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh))"
+sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh)"
 ```
 
 ### Manual installation
--- a/doc/installation-guide.md
+++ b/doc/installation-guide.md
@@ -13,7 +13,7 @@ a version of the standard library compatible with your installed
 version of Agda.
 
 ```shell
-sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh))"
+sh -c "$(curl --proto '=https' --tlsv1.2 -s https://raw.githubusercontent.com/agda/agda-stdlib/refs/heads/master/stdlib-install.sh)"
 ```
 
 ## Manual installation


stdlib-install.sh
```diff
@@ -0,0 +1,141 @@
...
+# Adding the standard library to the list of installed and default libraries
+STDLIB_PATH="$AGDA_DIR/agda-stdlib-$STDLIB_VERSION/standard-library.agda-lib"
+AGDA_LIBRARIES_FILE="libraries-$AGDA_VERSION"
+
+if ! grep -Eq "$STDLIB_PATH" "$AGDA_LIBRARIES_FILE"; then
+  echo "$STDLIB_PATH" >> "$AGDA_LIBRARIES_FILE"
+fi
+
+if ! grep -Eq "^standard-library$" "defaults-$AGDA_VERSION"; then
+  echo "standard-library" >> "defaults-$AGDA_VERSION"
```
gallais: Should we specify the version number?

stdlib-install.sh
```diff
@@ -0,0 +1,141 @@
...
+if [ -z ${AGDA_EXEC-} ]; then
+    read -p "What's the name of your Agda executable (default: agda)? " AGDA_EXEC
+    if [ -z "$AGDA_EXEC" ]; then
+        AGDA_EXEC=agda
+    fi
+fi
+
+# Double check that the command exists
+if ! [ -x "$(command -v $AGDA_EXEC)" ]; then
+  throwError "'$AGDA_EXEC' is not a valid executable"
```
gallais, line 64: Should we check if e.g. `$HOME/.cabal/bin/$AGDA_EXEC` exists and suggest that the user should be adding `$HOME/.cabal/bin/` to their `PATH`?

MatthewDaggitt:
In theory this looks great.

How cross platform is this and what do we need to do to support a new release? It looks like we just update the lookup table?

gallais:
It won't work on windows but I'm keeping it as simple as possible in the hope it'd work on
any unix style system. So far we only need simple things like `/bin/sh`, `grep`, `echo`, `true`,
`read`.

And, indeed, the only update would be to the lookup table.

stdlib-install.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# Setting up the Agda install directory
+mkdir -p "$AGDA_DIR"
+cd "$AGDA_DIR"
+mkdir -p logs
+
+# Downloading and extracting the standard library
+STDLIB_TARBALL_NAME="/tmp/agda-stdlib-$STDLIB_VERSION.tar.gz"
+STDLIB_TARBALL_URL="https://github.com/agda/agda-stdlib/archive/$STDLIB_TAG.tar.gz"
+wget -O "$STDLIB_TARBALL_NAME" "$STDLIB_TARBALL_URL" -o logs/wget
```
gallais: Catch errors and report them?

stdlib-install.sh
Taneb: You can get `cabal`'s default install path from `cabal` with `cabal path --installdir`

stdlib-install.sh
gallais: I have decided against it.

gallais:
I am quite happy with the current script.

I have included a command you can copy/paste in a terminal to test it.

stdlib-install.sh
JacquesCarette: Yes... but let's not let perfection stop us from having very useful functionality now.

gallais: It turned out to be easy enough to do.

gallais:
Hmm. I think `defaults-2.9.0` is not actually supported by Agda. Meaning the only solution
is to shove `standard-library` in the global `defaults` even though it may not be installed for
all the compiler's versions. :(

I'll open a PR to Agda to add support for `defaults-X.Y.Z` but that will only work for future
versions. :(

gallais:
The semester is about to start, it'd be convenient to have the installer merged in :)

jamesmckinna (approved):
@gallais I've only eyeballed the `stdlib-install.sh` script, and perhaps am not the best person to comment, but I accept the appeal to Getting Stuff Done now... and your students will quickly find out if the script is broken, so merging now.

PR #2904, opened 2025-12-22
Adding extrema to a non-strict order respects equality

Taneb:
These match corresponding lemmata for strict orders, and I based the implementation on those.

### DIFF
--- a/src/Relation/Binary/Construct/Add/Supremum/NonStrict.agda
+++ b/src/Relation/Binary/Construct/Add/Supremum/NonStrict.agda
@@ -15,14 +15,16 @@ module Relation.Binary.Construct.Add.Supremum.NonStrict
   {a ℓ} {A : Set a} (_≤_ : Rel A ℓ) where
 
 open import Level using (_⊔_)
+open import Data.Product.Base as Product using (_,_)
 open import Data.Sum.Base as Sum
 open import Relation.Binary.Structures
   using (IsPreorder; IsPartialOrder; IsDecPartialOrder; IsTotalOrder; IsDecTotalOrder)
 open import Relation.Binary.Definitions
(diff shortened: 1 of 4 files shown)


src/Relation/Binary/Construct/Add/Supremum/NonStrict.agda
```diff
@@ -76,6 +78,15 @@ data _≤⁺_ : Rel (A ⁺) (a ⊔ ℓ) where
 ≤⁺-antisym-≡ antisym (_ ≤⊤⁺) (_ ≤⊤⁺) = refl
 ≤⁺-antisym-≡ antisym [ p ] [ q ]     = cong [_] (antisym p q)
 
+≤⁺-respˡ-≡ : _≤⁺_ Respectsˡ _≡_
+≤⁺-respˡ-≡ = subst (_≤⁺ _)
```
JacquesCarette, line 82: Is that the 'only' proof? `subst` is such a pain to deal with downstream, I think it's a good idea to avoid it whenever possible.

JacquesCarette (changes requested):
It all looks good, except for the use of `subst`.

src/Relation/Binary/Construct/Add/Supremum/NonStrict.agda
Taneb: I think this is the only implementation, more or less. It's also what we're already doing for strict orders - see, for example, https://agda.github.io/agda-stdlib/master/Relation.Binary.Construct.Add.Supremum.Strict.html#3718

JacquesCarette: The argument "the other implementation is bad too" carries no weight with me. "I've thought about it and don't see any other way" does.

JacquesCarette (approved):

MatthewDaggitt (approved):

PR #2914, opened 2026-01-23
[ add ] `Data.List.Fresh.Membership.DecSetoid`

jkopanski:
I needed to inline `Decidable` because otherwise Agda wasn't able to figure out `r`.

### DIFF
--- a/src/Data/List/Fresh/Membership/DecSetoid.agda
+++ b/src/Data/List/Fresh/Membership/DecSetoid.agda
@@ -1,0 +1,40 @@
+------------------------------------------------------------------------
+-- The Agda standard library
+--
+-- Decidable setoid membership over fresh lists
+------------------------------------------------------------------------
+
+{-# OPTIONS --cubical-compatible --safe #-}
+
+open import Relation.Binary.Bundles using (DecSetoid)
+
+module Data.List.Fresh.Membership.DecSetoid {a ℓ} (DS : DecSetoid a ℓ) where
+
+open import Level using (Level)
+open import Data.List.Fresh.Relation.Unary.Any using (any?)
+import Data.List.Fresh.Membership.Setoid as MembershipSetoid
+open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Definitions using (Decidable)
+open import Relation.Nullary.Decidable using (¬?)
+
+open DecSetoid DS using (setoid; _≟_) renaming (Carrier to A)
+
+private
+  variable
+    r : Level
+    R : Rel A r
+
+
+------------------------------------------------------------------------
+-- Re-export contents of Setoid membership
+open MembershipSetoid setoid public
+
+------------------------------------------------------------------------
+-- Other operations
+infix 4 _∈?_ _∉?_
+
+_∈?_ : Decidable (_∈_ {R = R})
+x ∈? xs = any? (x ≟_) xs
+
+_∉?_ : Decidable (_∉_ {R = R})
+x ∉? xs = ¬? (x ∈? xs)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,8 @@ New modules
 
 * `Algebra.Properties.Semiring`.
 
+* `Data.List.Fresh.Membership.DecSetoid`.
+
 * `Data.List.Relation.Binary.Permutation.Algorithmic{.Properties}` for the Choudhury and Fiore definition of permutation, and its equivalence with `Declarative` below.
 
 * `Data.List.Relation.Binary.Permutation.Declarative{.Properties}` for the least congruence on `List` making `_++_` commutative, and its equivalence with the `Setoid` definition.


src/Data/List/Fresh/Membership/DecSetoid.agda
```diff
@@ -0,0 +1,39 @@
...
+
+module Data.List.Fresh.Membership.DecSetoid {a ℓ} (DS : DecSetoid a ℓ) where
+
+open import Level using (Level)
+open import Data.List.Fresh using (List#)
+open import Data.List.Fresh.Relation.Unary.Any using (any?)
+open import Relation.Binary.Core using (Rel)
+open import Relation.Nullary.Decidable using (¬?; Dec)
+
+open DecSetoid DS renaming (Carrier to A)
```
jamesmckinna: This will bring `setoid` into scope, so L24 below can be simplified accordingly?
And if you wanted to sharpen the scope of imported declarations, you could go further and write
```suggestion
open DecSetoid DS using (setoid; _≟_) renaming (Carrier to A)
```

src/Data/List/Fresh/Membership/DecSetoid.agda
```diff
@@ -0,0 +1,39 @@
...
+
+------------------------------------------------------------------------
+-- Re-export contents of propositional membership
```
jamesmckinna:
```suggestion
-- Re-export contents of Setoid membership
```

We usually restrict the designation 'propositional' to refer to the canonical `Setoid` structure induced by `PropositionalEquality`.

src/Data/List/Fresh/Membership/DecSetoid.agda
```diff
@@ -0,0 +1,39 @@
...
+
+open import Data.List.Fresh.Membership.Setoid (DecSetoid.setoid DS) public
```
jamesmckinna: As above
```suggestion
open import Data.List.Fresh.Membership.Setoid setoid public
```

src/Data/List/Fresh/Membership/DecSetoid.agda
```diff
@@ -0,0 +1,39 @@
...
+-- Other operations
+
+infix 4 _∈?_ _∉?_
+
+private
+  variable
+    r : Level
+
+_∈?_ : {R : Rel A r} → (x : A) → (xs : List# A R) → Dec (x ∈ xs)
+x ∈? xs = any? (x ≟_) xs
```
jamesmckinna: The explicit expansion to use `Dec` might indeed be annoying. I'm not sure it would necessarily be an improvement (readability?), though, to `open import Relation.Binary.Definitions using (Decidable)`, and then have
```agda
 private
  variable
    r : Level
    R : Rel A r


-- Re-export contents of propositional membership

open import Data.List.Fresh.Membership.Setoid setoid public

-- Other operations

infix 4 _∈?_ _∉?_

_∈?_ : Decidable (_∈_ {R = R})
x ∈? xs = any? (x ≟_) xs
```
instead?
```

File: src/Data/List/Fresh/Membership/DecSetoid.agda [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+
+_∉?_ : {R : Rel A r} → (x : A) → (xs : List# A R) → Dec (x ∉ xs)
```
jamesmckinna: Ditto.
```agda
_∉?_ : Decidable (_∉_ {R = R})
```

src/Data/List/Fresh/Membership/DecSetoid.agda
```diff
@@ -0,0 +1,39 @@
...
+open import Data.List.Fresh.Membership.Setoid (DecSetoid.setoid DS) public
+
+------------------------------------------------------------------------
+-- Other operations
+
+infix 4 _∈?_ _∉?_
+
+private
+  variable
+    r : Level
```
jamesmckinna: I think the `style-guide` recommends that such `private` blocks, where possible, be moved to a position below the main `import` block, and above any subsequent definitions.

More debatable (as in the suggestions above) is whether it's legitimate to also make a global declaration of `R : Rel A r` in this block as well.

src/Data/List/Fresh/Membership/DecSetoid.agda
jkopanski: > The explicit expansion to use Dec might indeed be annoying

It's not that annoying imho. In the opening message I was just trying to address question that I thought might pop up.

jamesmckinna: But thanks for encouraging the question!

It may be that *I* am irritated by having to supply so much quantifier prefix/inlined terms in order to get past the typechecker, so I thought it worth exploring the alternatives, notwithstanding @JacquesCarette 's reservations concerning #697 ...

See #2916 for such an exploration, following my suggestions above.

jamesmckinna: And thanks for taking my suggestions... I think it does look cleaner now. I think we can merge this now, but suggest waiting till Monday in case any of the other maintainers have strong views on the matter!

jkopanski: > It may be that I am irritated by having to supply so much quantifier prefix/unlined fern in order to get past the typechecker, so I thought it worth exploring the alternative

Sure, although personally I have a knee-jerk reaction whenever I have to apply implicit to operator thinking that I must have done something wrong.

jamesmckinna: See also https://github.com/agda/agda/issues/8067

PR #701, opened 2019-04-14
[ refactor ] Experimenting with variables in `Data.List`

MatthewDaggitt:
An experiment with the new `variable` key word informed by the discussion in #697. This was enormously satisfying to do :smile: Comments welcome.

PR #2916, opened 2026-01-23
[ refactor ] use `variable`s more systematically in `Data.List.Fresh{.*}`

jamesmckinna:
@jkopanski 's recent #2914 nudged me to think about refactoring the associated modules
* to be more systematic about using `variable`s (UPDATED: it seems the only reasonable way to do this for the whole library is ... little-by-little, at least wrt only 'small' groups of cognate modules?)
* ~~to extend the range of permitted such declarations to also range over `Rel` and `Pred` arguments... (elsewhere discussed as a `style-guide`/`library-design` issue #697 #701 etc.)~~

~~DRAFT, for discussion at this stage.~~ Now open for review (including a self-review... ;-))

UPDATED: `CHANGELOG` now records the new notation `x #[ R ] xs`

### DIFF
--- a/src/Data/List/Fresh/Membership/Setoid/Properties.agda
+++ b/src/Data/List/Fresh/Membership/Setoid/Properties.agda
@@ -83,7 +83,7 @@ module _ (R⇒≉ : ∀[ R ⇒ _≉_ ]) (≉⇒R : ∀[ _≉_ ⇒ R ]) where
   fresh-remove {xs = cons x xs pr} (here x≈y)   = fresh-respectsˡ resp (sym x≈y) pr
     where
     resp : R Respectsˡ _≈_
-    resp x≈y Rxz = ≉⇒R (R⇒≉ Rxz ∘′ trans x≈y) 
+    resp x≈y Rxz = ≉⇒R (R⇒≉ Rxz ∘′ trans x≈y)
   fresh-remove {xs = cons x xs pr} (there x∈xs) =
     ≉⇒R (distinct x∈xs (fresh⇒∉ R⇒≉ pr)) , fresh-remove x∈xs
 


jamesmckinna:
### @MatthewDaggitt , you wrote
> I like these changes! The only comment I have is that according to the style guide, `P` should not be a variable as it is not involved in the type of the data definition.
> 
> EDIT: Ah I see you already commented above. So `R` is covered under the [style guide](https://github.com/agda/agda-stdlib/blob/master/doc/style-guide.md#variables) already. I think the current style-guide rule is a good balance, and I don't see how one could extend it in a principled manner?

Thanks! 

UPDATED: TL;DR: you were right, and I was wrong. The additional commits for the `Relation.Unary` and `Properties` sub-modules don't, AFAICT, violate the existing guidelines, so there's no need to entertain any expansion of them.

(IGNORE FROM HERE) I (largely) agree, but in my more 'maximalist' style, I am reaching the point where I would like the `style-guide` to row back on the two `Example`s under `Types` in the style-guide for `variable` declarations, as follows:
* as you point out, here `R` is OK, because it is part of the type definition, where `P` is not;
* BUT, I think we should (at least) entertain the possibility that whenever a type arises as a *parameter* to a definition, or as a *definiendum* itself, then *so too do all the dependent types which may derive from it*.

That is to say, in this example, once we have `A` in scope as an allowable range of implicit quantification, all of `Pred A _`, `Rel A _` etc. come with it, too... *but also* (but this may indeed be going too far), also `Pred (Pred A _) _` etc. (So there is a first-order/higher-order distinction here which may simply reflect my own instincts/feelings about cognitive overload... as well as bitter experience with unsolved metas and the limitations of Agda's lack of higher-order unification ;-)).

Now, this would be a significant widening of the style-guide recommendations, for sure, so how to keep things under reasonable control so that `variable` blocks don't proliferate to ridiculous lengths/logical complexity? Two possible 'sensible' (!?) restrictions:
* stay first order, so `Pred A _` is OK, similarly `Pred (List A) _` under `Data.List.*` etc. would be OK, but not `Pred (Pred A _) _` etc..
* allow `Pred` etc. `variable`s in `Properties` modules, but not in `Base`? (except where already covered as with `R` in `List A R`)

The alternative (as we do now, but sometimes it gets *very* creaky, at least, that was how I felt doing this piece of work) is to be (more) disciplined about using
```agda
module _ {P : Pred A _} where
```
scoping mechanisms, but the creakiness comes, as with this PR when we pass to `Properties`, I might claim, when you end up with *every* significant block of functionality requiring such a prefix... at which point, a global `variable` becomes the 'obvious' choice to make, both for implementation (and the prenex scope discipline which derives from that) *and* for documentation... "we're going to be talking about predicates and relations in general about the types and functions defined in `Base`, so it will be convenient to standardise identifiers/lexemes for such things as follows:..."

I'm sure there are good reasons to reject the above ideas, but... testing the envelope!

UPDATED: I think the above can (mostly) stand, but I did at least re-read my own code contributed here, and agree with you. I was being an idiot! `P` is only really used for `filter`/`takeWhile`/`dropWhile`, so I should take my medicine.

jamesmckinna:
I think in the interests of review overhead, I won't contribute the corresponding refactoring of `Properties` to this PR, but it's an obvious next step, unless the sensible thing is to do so now (I'd have to do those changes, of course!). 
UPDATED: `Properties` hardly took any effort to add, so worth it, I think. And potentially the rest to follow for both the `Relation` and `Membership` subhierarchies...
UPDATED: now done, with some interesting minor gripes wrt the actual intended generalisation!

src/Data/List/Fresh.agda
```diff
@@ -24,7 +24,7 @@ open import Data.Nat.Base using (ℕ; zero; suc)
 open import Function.Base using (_∘′_; flip; id; _on_)
 open import Relation.Nullary using (does)
 open import Relation.Unary as U using (Pred)
```
jamesmckinna: Given that `U` is only used *once* qualified, for `U.Decidable`, I would rather:
* retain the qualified import, but as `Unary`
* use `Decidable` in-place, unqualified

```suggestion
open import Relation.Unary as Unary using (Pred; Decidable)
```

src/Data/List/Fresh.agda
```diff
@@ -24,7 +24,7 @@ open import Data.Nat.Base using (ℕ; zero; suc)
...
-open import Relation.Binary.Core using (Rel)
+open import Relation.Binary.Core using (Rel; REL)
 import Relation.Binary.Definitions as B using (Reflexive)
```
jamesmckinna: Similarly, here:
```suggestion
open import Relation.Binary.Definitions as Binary using (Reflexive)
```

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
+pattern [_] x = x ∷# []
 
-fromMaybe : {R : Rel A r} → Maybe A → List# A R
+fromMaybe : Maybe A → List# A R
 fromMaybe nothing  = []
-fromMaybe (just a) = [ a ]
+fromMaybe (just x) = [ x ]
 
-module _ {R : Rel A r} (R-refl : B.Reflexive R) where
+module _ (R-refl : B.Reflexive {A = A} R) where
```
jamesmckinna: Thus:
```suggestion
module _ (R-refl : Reflexive {A = A} R) where
```
or even
```suggestion
module _ (refl : Reflexive {A = A} R) where
```
because there's no clash with `Relation.Binary.PropostionalEquality.refl`?
```

File: src/Data/List/Fresh.agda [hunk 4]
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
 
-  replicate zero    a = []
-  replicate (suc n) a = cons a (replicate n a) (replicate-# n a)
+  replicate zero    x = []
+  replicate (suc n) x = cons x (replicate n x) (replicate-# n x)
 
-  replicate-# zero    a = _
-  replicate-# (suc n) a = R-refl , replicate-# n a
+  replicate-# zero    x = _
+  replicate-# (suc n) x = R-refl , replicate-# n x
```
jamesmckinna: and
```suggestion
  replicate-# (suc n) x = refl , replicate-# n x
```

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
-drop : {R : Rel A r} → ℕ → List# A R → List# A R
-drop zero    as        = as
+drop : ℕ → List# A R → List# A R
+drop zero    xs        = xs
 drop (suc n) []        = []
-drop (suc n) (a ∷# as) = drop n as
+drop (suc n) (x ∷# xs) = drop n xs
 
-module _ {P : Pred A p} (P? : U.Decidable P) where
+module _ {P : Pred A p} (P? : U.Decidable {A = A} P) where
```
jamesmckinna: Ditto. regarding the qualified name, but this is where Matthew's comment bites: (and arguably, this version *is* more readable!)
```suggestion
module _ {P : Pred A p} (P? : Decidable P) where
```

src/Data/List/Fresh.agda
jamesmckinna: I've now done us throughout the rest of the PR, so I'll just commit this change here too.

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
 
 -- Convenient notation for freshness making A and R implicit parameters
 infix 5 _#_
-_#_ : {R : Rel A r} (a : A) (as : List# A R) → Set r
+_#_ : REL A (List# A R) _
 _#_ = fresh _ _
```
jamesmckinna: The whole intention behind this PR is to better support generalisation over `R` when using `List#`, and specifically the ability to generalise over arguments `xs : List# A R` correctly.

Unfortunately, this sometime means that *some* statements involving `x # xs` fail to generalise correctly wrt the implied/implicit argument `R` relative to other generalisation sites in the type. My current solution is to nudge the type checker by sprinkling a few `{R = R}` implicit bindings where needed, but this seems slightly less satisfactory, when the real culprit seems to be the *relation* `_#_`

I see two possible solutions:
* delegate to using the *explicit* `fresh` operation where needed for disambiguation; but this changes the 'visible'/'legible' API when stating lemmas
* introduce a small additional piece of notation, eg `x #[ R ] xs` where `R` needs to be made visible (and for some reason I couldn't seem to do this as `syntax`...?)

Thoughts? Eg. as
```agda
-- Convenient notation for freshness making A (and R) implicit
infix 5 _#[_]_ _#_

_#[_]_ : A → (R : Rel A r) → Pred (List# A R) _
x #[ R ] xs = fresh _ R x xs

_#_ : REL A (List# A R) _
x # xs = x #[ _ ] xs
```

src/Data/List/Fresh.agda
jamesmckinna: Done!

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
 head = Maybe.map proj₁ ∘′ uncons
 
-tail : {R : Rel A r} → List# A R → Maybe (List# A R)
+tail : List# A R → Maybe (List# A R)
 tail = Maybe.map proj₂ ∘′ uncons
 
-take   : {R : Rel A r} → ℕ → List# A R → List# A R
-take-# : {R : Rel A r} → ∀ n a (as : List# A R) → a # as → a # take n as
+take   : ℕ → List# A R → List# A R
+take-# : ∀ n y (xs : List# A R) → y # xs → y # take n xs
```
jamesmckinna: Eg if we were to introduce `_#[_]_` then this could typecheck as:
```suggestion
take-# : ∀ n y xs → y #[ R ] xs → y # take n xs
```

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
-drop (suc n) (a ∷# as) = drop n as
+drop (suc n) (x ∷# xs) = drop n xs
 
-module _ {P : Pred A p} (P? : U.Decidable P) where
+module _ {P : Pred A p} (P? : Decidable P) where
 
-  takeWhile   : {R : Rel A r} → List# A R → List# A R
-  takeWhile-# : ∀ {R : Rel A r} a (as : List# A R) → a # as → a # takeWhile as
+  takeWhile   : List# A R → List# A R
+  takeWhile-# : ∀ y (xs : List# A R) → y # xs → y # takeWhile xs
```
jamesmckinna: Similarly
```suggestion
  takeWhile-# : ∀ y xs → y #[ R ] xs → y # takeWhile xs
```

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
-  dropWhile : {R : Rel A r} → List# A R → List# A R
+  dropWhile : List# A R → List# A R
   dropWhile []            = []
-  dropWhile aas@(a ∷# as)  = if does (P? a) then dropWhile as else aas
+  dropWhile xxs@(x ∷# xs)  = if does (P? x) then dropWhile xs else xxs
 
-  filter   : {R : Rel A r} → List# A R → List# A R
-  filter-# : ∀ {R : Rel A r} a (as : List# A R) → a # as → a # filter as
+  filter   : List# A R → List# A R
+  filter-# : ∀ y (xs : List# A R) → y # xs → y # filter xs
```
jamesmckinna:
```suggestion
  filter-# : ∀ y xs → y #[ R ] xs → y # filter xs
```
etc.

src/Data/List/Fresh.agda
```diff
@@ -64,145 +68,145 @@ module _ {a} (A : Set a) (R : Rel A r) where
...
+  ... | true  = p , filter-# y xs ps
+  ... | false = filter-# y xs ps
 
 ------------------------------------------------------------------------
 -- Relationship to List and AllPairs
 
-toList : {R : Rel A r} → List# A R → ∃ (AllPairs R)
-toAll  : ∀ {R : Rel A r} {a} as → fresh A R a as → All (R a) (proj₁ (toList as))
+toList : List# A R → ∃ (AllPairs R)
+toAll  : (xs : List# A R) → x # xs → All (R x) (proj₁ (toList xs))
```
jamesmckinna:
```suggestion
toAll  : ∀ xs → x #[ R ] xs → All (R x) (proj₁ (toList xs))
```

src/Data/List/Fresh.agda
```diff
@@ -23,32 +23,36 @@ open import Data.Maybe.Base as Maybe using (Maybe; just; nothing)
...
+
 
 ------------------------------------------------------------------------
 -- Basic type
 
 -- If we pick an R such that (R a b) means that a is different from b
 -- then we have a list of distinct values.
 
-module _ {a} (A : Set a) (R : Rel A r) where
+module _ (A : Set a) (R : Rel A r) where
```
jamesmckinna, line 48: I keep scratching my head as to whether `A` could be bound implicitly here, or indeed *should*.
That would obviously be a `breaking` change to the API, so out-of-scope for this PR, but still...

src/Data/List/Fresh/Relation/Unary/Any.agda
```diff
@@ -9,28 +9,35 @@
...
+
 
 module _ {A : Set a} {R : Rel A r} (P : Pred A p) where
 
   data Any : List# A R → Set (p ⊔ a ⊔ r) where
     here  : ∀ {x xs pr} → P x → Any (cons x xs pr)
     there : ∀ {x xs pr} → Any xs → Any (cons x xs pr)
 
-module _ {R : Rel A r} {P : Pred A p} {x} {xs : List# A R} {pr} where
+module _ {pr : fresh A R x xs} where
```
jamesmckinna: This is the one glaring wart in the refactoring, but I couldn't find a clean way out of the bottle, other than to introduce the abbreviation `pr : x #[ R ] xs`. Suggestions welcome!

UPDATED: this seems to be a problem with the relative order in which `P` gets generalised... But simple-minded permutation of the `varaiable` block doesn't seem to fix it. Oh well.

src/Data/List/Fresh/Membership/Setoid.agda
```diff
@@ -10,22 +10,26 @@ open import Relation.Binary.Bundles using (Setoid)
...
+
+infix 4 _∈_ _∉_
 
-_∈_ : {R : Rel A r} → A → List# A R → Set _
+_∈_ : REL A (List# A R) _
 x ∈ xs = Any (x ≈_) xs
 
-_∉_ : {R : Rel A r} → A → List# A R → Set _
+_∉_ : REL A (List# A R) _
 x ∉ xs = ¬ (x ∈ xs)
```
gallais, line 35: Unrelated, but I wonder whether we should prioritise a more constructive approach here
and have `All (x ≉_) xs` instead. Probably part of a wider 3.0 discussion on constructive
negation à la https://bentnib.org/data-types-negation-msfp.html

src/Data/List/Fresh/Relation/Unary/Any/Properties.agda
```diff
@@ -30,37 +30,40 @@ private
...
+Any⇒NonEmpty {xs = cons x xs pr} p  = cons x xs pr
 
 ------------------------------------------------------------------------
 -- Correspondence between Any and All
 
-module _ {R : Rel A r} {P : Pred A p} {Q : Pred A q} (P⇒¬Q : ∀[ P ⇒ ∁ Q ]) where
+module _ (P⇒¬Q : ∀[ P ⇒ ∁ Q ]) where
 
-  Any⇒¬All : {xs : List# A R} → Any P xs → ¬ (All Q xs)
+  Any⇒¬All : Any P xs → ¬ (All Q xs)
```
gallais: You could go all in with

```suggestion
  Any⇒¬All : ∀[ Any P ⇒ ∁ (All Q) ]
```

src/Data/List/Fresh/Relation/Unary/Any.agda
```diff
@@ -50,32 +57,28 @@ module _ {R : Rel A r} {P : Pred A p} {x} {xs : List# A R} {pr} where
...
   ⊎⇔Any = mk⇔ fromSum toSum
 
-module _ {R : Rel A r} {P : Pred A p} {Q : Pred A q} where
-
-  map : {xs : List# A R} → ∀[ P ⇒ Q ] → Any P xs → Any Q xs
-  map p⇒q (here p)  = here (p⇒q p)
-  map p⇒q (there p) = there (map p⇒q p)
-
-module _ {R : Rel A r} {P : Pred A p} where
+map : ∀[ P ⇒ Q ] → Any P xs → Any Q xs
```
gallais: I guess it's breaking due to the position of the implicit argument but

```suggestion
map : ∀[ P ⇒ Q ] → ∀[ Any P ⇒ Any Q ]
```

is more elegant I think.

src/Data/List/Fresh/Relation/Unary/Any.agda
```diff
@@ -50,32 +57,28 @@ module _ {R : Rel A r} {P : Pred A p} {x} {xs : List# A R} {pr} where
...
+map p⇒q (here p)  = here (p⇒q p)
+map p⇒q (there p) = there (map p⇒q p)
 
-  witness : {xs : List# A R} → Any P xs → ∃ P
-  witness (here p)   = -, p
-  witness (there ps) = witness ps
+witness : Any P xs → ∃ P
```
gallais: Isn't this typically using `Satisfiable` as the return type?

src/Data/List/Fresh/Relation/Unary/Any.agda
```diff
@@ -50,32 +57,28 @@ module _ {R : Rel A r} {P : Pred A p} {x} {xs : List# A R} {pr} where
...
 
 infixl 4 _─_
 _─_ = remove
 
-module _ {R : Rel A r} {P : Pred A p} (P? : Decidable P) where
+module _ (P? : Decidable P) where
 
-  any? : (xs : List# A R) → Dec (Any P xs)
-  any? []        = no (λ ())
+  any? : ∀ xs → Dec (Any {R = R} P xs)
```
gallais:
```suggestion
  any? : Decidable (Any {R = R} P)
```

src/Data/List/Fresh/Relation/Unary/Any.agda
MatthewDaggitt: I know I say this a lot but, in my opinion, this use of concise notation for everyday operations such as quantification and functions deeply hurts the readability of the code base for non-advanced users...

src/Data/List/Fresh/Membership/Setoid.agda
jamesmckinna: Nice!

Unfortunately, I suspect that 'conventional' usage would favour the status quo, but we should/could instead show equivalence with the 'positive' version `All (x ≉_) xs` under `Properties`?

UPDATED: now added.

src/Data/List/Fresh/Relation/Unary/Any.agda
jamesmckinna: Yes, at least following #2954 etc. Good catch! And it shouldn't be called `witness` in that case, either!

src/Data/List/Fresh/Relation/Unary/Any.agda
jamesmckinna: Well @MatthewDaggitt , for better or worse, it already uses `∀[ P ⇒ Q ]` in the assumptions, on `master`.

To be frank, though, I don't know why we use the above form in preference to writing `P ⊆ Q` so that we could have, instead
```suggestion
map : P ⊆ Q → Any P ⊆ Any Q
```
modulo @gallais 's comment about the `breaking` change to the quantification order...

... because that makes clearer that this is the functorial action of `Any` wrt the `_⊆_` categorical structure...

src/Data/List/Fresh/Relation/Unary/Any.agda
jamesmckinna: Nice!

src/Data/List/Fresh/Relation/Unary/Any/Properties.agda
jamesmckinna: Better, I think:
```suggestion
  Any⇒¬All : Any P  ⊆ ∁ (All Q)
```

jamesmckinna:
Any thoughts @gallais @MatthewDaggitt concerning my suggested introduction of `x #[ R ] xs` as additional notation to disambiguate `x # xs` when `R` isn't inferrable?

UPDATED: by inserting appropriate `{R = R}` bindings, it seems the notation can be avoided altogether, but I've added it nonetheless, and done some experiments using it (sparingly)

src/Data/List/Fresh.agda
jamesmckinna: RESOLVED: it seems that inserting `{R = R}` bindings solves this problem in the cases of interest.

Nevertheless, most recent commits introduce the notation and experiment with it.

src/Data/List/Fresh/Relation/Unary/Any.agda
gallais: I like that suggestion

gallais:
I do like that `x #[ R ] xs` syntax

jamesmckinna:
Regarding the discussions above about the type of `map`, I think that on balance it *is* worth making the `breaking` change to shift the implicit quantifier from prenex to the subset inclusion on `Pred`... but I guess that can wait to a followup under v3.0?

It's worth noting that at least as far as `stdlib`/`make test` is concerned, the change isn't observable... and indeed, such a change would bring `Data.List.Fresh` into line with `Data.List` wrt `All` and `Any`... sigh.

Similarly, we should followup this PR with one which is more systematic wrt use of `variable`s in `Data.List.*` and `Data.Vec.*`?

src/Data/List/Fresh/Relation/Unary/Any.agda
JacquesCarette: As do I!

JacquesCarette:
I also like the `x #[ R ] xs` syntax.

Issue #2918, opened 2026-01-23
Typo in ``Rational.Properties.nonPos*nonPos⇒nonPos``


mechvel:
The name has to be `nonPos*nonPos⇒nonNeg`

jamesmckinna:
Well spotted! Thanks very much.

PR #2920, opened 2026-01-24
[ fix ] issue #2918

jamesmckinna:


jamesmckinna:
> Do we need to add a deprecation warning?

Well, it's now listed under `Bug fixes` in the `CHANGELOG`, and my understanding @MatthewDaggitt was that such things, while technically `breaking`, should be considered... otherwise... ie if it were to break code in any client of the v2.4 release, then I guess they should look at the `CHANGELOG`, but yes, I suppose belt-and-braces we could add a deprecation warning. 

Will do so, and can always revert later (else simply add it to the `Non-backwards-compatible` list for v3.0 ;-)).

Issue #2917, opened 2026-01-23
[ deprecate ] one of `Function.Core` *or* `Algebra.Core`


jamesmckinna:
It's not obvious to me that we ever make use of this module. 

Moreover, its functionality is largely duplicated by `Algebra.Core`, while `Morphism` as a definition seems deprecated/unused everywhere else, except as a `public` re-export in `Relation.Binary.Morphism.Definitions` cf. #2875 / #2922 UPDATED: ah... but see #1206 / #1213

The alternative would be to deprecate (?), or else simply leave as a stub, `Algebra.Core`, as:
```agda
module Algebra.Core where

------------------------------------------------------------------------
-- Unary and binary operations

open Function.Core public
  using ()
  renaming (Fun₁ to Op₁; Fun₂ to Op₂)
```

MatthewDaggitt:
We have found `Algebra.Core` and `Relation.Binary.Core` very useful in breaking dependency cycles. I therefore added `Function.Core` under the assumption that it would be similarly useful. If we want to remove it, I guess we should at least try to understand why the former two are useful but the latter is not?

jamesmckinna:
So my understanding (thanks for the context/history update) would be: we only *really* have `Function.Core` for `Morphism`, as requested (for pedagogical purposes) by @pnlph . So in that sense, it remain 'useful' (whereas `Fun₁` and `Fun₂` seem... moot? They're *really* never used anywhere, except for their redundant recapitulation in `Algebra.Core`).

I can believe that, at least in principle, having a `Z.Core` module at the root of hierarchy is a convenient placeholder for *extremely* common definitions which would otherwise create dependency cycles/DRY violations. And there very well may be more to come in the future...

But the DRY violation in `Algebra.Core` (`Op₁` and `Op₂` in place of/duplicating `Fun₁` and `Fun₂`) is perhaps the one stand-out (and why I modified the title of the issue, as my knowledge of/instinct towards the issue shifted), and I guess is also (a)historical, but now, *could* be refactored away, as indicated, if we opt to retain `Function.Core`.

Arguably, this is pure bikeshedding, but downstream/eventually/maybe never, the current setup seems a wart which we could 'tidy up'.

But happy to file this under `hypothetical-rewrite`... it surely can't be a deal-breaker ;-)

JacquesCarette:
[...]

Issue #2875, opened 2025-11-12
[ fix ] parametrisation of `Relation.Binary.Morphism.Definitions` and `Algebra.Morphism.Definitions`


jamesmckinna:
In the first case:
* the top-level module parameters bind `A : Set a` and `B : Set b` *explicitly*
* and then defines `Homomorphic₂` taking explicit arguments `(_∼₁_ : Rel A ℓ₁)` and `(_∼₂_ : Rel B ℓ₂)`
  
It seems 'obvious' that `A` and `B` could be given *implicitly* instead (but: `breaking`!), moreover that `Homomorphic₂` could be given by delegation to ~~`f Preserves _∼₁_ ⟶ _∼₂_` from `Relation.Binary.Core`~~ UPDATED `Function.Definitions.Congruent`.

Indeed, that the whole ~~module-level parametrisation could be rejigged using `variable`s to streamline things more smoothly?~~ module could be left as a stub, purely for compatibility cf. #1206 / #1213 . UPDATED: see #2922 

In the second case:
* the top-level module parameters bind `A : Set a` and `B : Set b` *explicitly*, together with `(_∼₂_ : Rel B ℓ₂)`
* and then defines `Homomorphic₂` taking explicit argument `(f : A → B)` (and the binary operations)

It seems 'obvious' here that `B` at least could be given *implicitly* instead (but: `breaking`!).

Are these 'bugs'/design decisions worth fixing?

Consequences: 
* harmless within `stdlib` in the deployment contexts for `Algebra.Morphism.Definitions` (but: knock-ons need fixing!)
* possibly more radical for `Relation.Binary.Morphism.Definitions` but could be an improvement (avoid supplying 'bogus' explicit arguments `A`, `B` anywhere on qualified instantiated import...?)

JacquesCarette:
I would be curious as to the scale of the knock-ons. For both. These are certainly excellent questions to ask. I can't help but think that, in practice, this won't be a 100% win -- but that's just a gut feel! (Trivial relations from which you can't figure out the actual carrier come to mind as times you might need to be specific. But it sure might be a 99% win.)

jamesmckinna:
So the curiosity seems to be:
* don't bother 'fixing' the `Algebra.Morphism.Definitions` (hardly worth the effort)?
* whereas `Relation.Binary.Morphism.Definitions` can be eliminated (not-quite) entirely! #2922

jamesmckinna:
[...]

Issue #2318, opened 2024-03-13
Can we unify Relation.Binary.Morphism.Definitions and Function.Definitions?


Taneb:
`Relation.Binary.Morphism.Definitions` has as module parameters the domain and codomain of the functions over which the definitions are made, whereas `Function.Definitions` has them in a `variable` block. The former module only contains `Homomorphic₂`, which is exactly the same as the latter's `Congruent`, and a re-export of `Morphism` from `Function.Core`, which I think is otherwise unused in the library (and is just a synonym for the function arrow).

Can we deprecate `Relation.Morphism.Definitions` in favour of `Function.Definitions`?

jamesmckinna:
"Can"? or "Should"?

It seems as though (modulo dependencies/dependency cycles? not sure about this...) we can.

Not obvious to me that we should... partly because of the ergonomics of the whole `Algebra` hierarchy... that eventually rests on `Relation.Binary` (and `IsRelHomomorphism` specifically), and swapping out a `Relation.Binary` component in favour of a `Function` one, even if semantically equivalent/justified, seems like a rupture (however small) in the design hierarchy? Maybe this is a nonsense argument, but the "what are all the things I need to import to get task X done?" question (allied with the lack of suitable search tools) means that I like not having to look 'outside the box' for functionality if I can avoid it... but this is off the top of my head, without a serious analysis of the situation.

MatthewDaggitt:
I agree with @jamesmckinna. We absolutely can, but I don't think we should because it would be violating the standard design of hierarchies in the library. Obviously if we could do it for every hierarchy in the library that would be a different question....

jamesmckinna:
... And the above remarks apply also to `Algebra.Consequences`, where it would (perhaps?) be better ergonomically not to have to reimport `Function.Definitions` instantiated twice with `_≈_`...

jamesmckinna:
See also #2581 for yet another variation on this theme...

JacquesCarette:
About your two different `Foo` modules above: my impression was that, at the usage site, there are the same. So if they are not, that should warrant an explanation in the official documentation.

jamesmckinna:
I'll refine the example to explain how I think they *could* be different at usage sites... later

jamesmckinna:
[...]

PR #2922, opened 2026-01-24
[ refactor ] make `Relation.Binary.Morphism.Definitions` obsolete

jamesmckinna:
This tackles the first part of #2875 , retaining the module solely for compatibility per #1206 / #1213 as requested by @pnlph , so no actual `deprecation` at this stage.

Knock-on consequences include:
* hopefully simplifying the dependency graph wrt imports of `Function.Definitions` vs. alternatives
* replacing the definition `Homomorphic₂` throughout by `Function.Definitions.Congruent` (but the name is still exported as `public` by the obsolete stub; and hence also exported by `Relation.Binary.Morphism`, but this could be removed/deprecated?)
* etc.

Apologies to @Taneb for having changed my mind regarding your original #2318 but this PR should now close that issue as 'fixed'!

TL;DR: the 'cause' of the problem is yet again [ DRY ] with some basic concepts duplicated across the `Function`/`Relation`/`Algebra` hierarchies as various specialisations (or not even that) of 'common' (very) general abstractions (eg 'monotonicity'/'congruence'/'respects') which different authors have, at different times, identified as `Core` to one or other hierarchy... cf. #2917


src/Relation/Binary/Morphism/Bundles.agda
```diff
@@ -9,17 +9,18 @@
 module Relation.Binary.Morphism.Bundles where
 
 open import Level using (Level; _⊔_)
-open import Relation.Binary.Core using (_Preserves_⟶_)
 open import Relation.Binary.Bundles using (Setoid; Preorder; Poset)
+open import Relation.Binary.Consequences using (mono⇒cong)
+open import Relation.Binary.Definitions using (Monotonic₁)
```
jamesmckinna, line 14: NB, this too could be replaced by
```suggestion
open import Function.Definitions using (Congruent)
```
But at the cost of straying outside the `Relation.*` hierarchy...

src/Relation/Binary/Morphism/Structures.agda
```diff
@@ -1,35 +1,34 @@
...
-open import Relation.Binary.Core using (Rel)
-
 module Relation.Binary.Morphism.Structures
   {a b} {A : Set a} {B : Set b}
   where
 
 open import Data.Product.Base using (_,_)
-open import Function.Definitions using (Injective; Surjective; Bijective)
+open import Function.Definitions
+  using (Congruent; Injective; Surjective; Bijective)
```
jamesmckinna, line 15: Alternatively, 
```suggestion
  using (Injective; Surjective; Bijective)
  renaming (Congruent to Homomorphic₂)
```

src/Relation/Binary/Morphism/Bundles.agda
JacquesCarette: Since morphisms tend to be functions, using `Function.Definitions` here does not seem out of place.

JacquesCarette (commented):
Putting as comment, since I don't want to block merge on just this.

src/Relation/Binary/Morphism/Bundles.agda
jamesmckinna: Although here, as the definition is being used in a context where the monotonicity of the function is being appealed to, I think is makes slightly more sense (cognitive load? 'documentation'?) to prefer `Monotonic₁` in this instance?

JacquesCarette: I agree that the context means monotonicity makes more sense to emphasize. So I guess you didn't really mean your own suggestion... but good discussion!

jamesmckinna: Once I'm on a DRY kick, it can easily get out-of-hand wanting *really* to have 'only one thing, in one place', whereas *some* contextual embedding is I think important. So yes, arguing-out-loud with myself may seem a bit ridiculous, so I appreciate a second look from outside!

src/Relation/Binary/Morphism/Structures.agda
jamesmckinna: This turns on whether we any consider the *name* `Homomorphic₂` as somehow more 'informative' than `Congruent`, but I suspect we no longer do so?

JacquesCarette: That's a really complex question. `Congruent` is very generic, and somehow not as well-known in mathematics (because it is pervasively assumed to be "just true"). So I'm not quite sure which name is more informative. Plus `Homomorphic` is somehow more fitting in the context of morphisms - but I could be convinced that that's not true.

jamesmckinna: The thing which clinched it for me is that `Algebra.Morphism.Definitions` defines `Homomorphic₀`, `Homomorphic₁`, and `Homomorphic₂` based on the *arity* of the operation wrt which the source and target 'equality' relations witness *being a homomorphism*, whereas `Relation.Binary.Morphism.Structures` uses subscripted `Homomorphic₂` for two *relations*... and that's a cognitive dissonance I can no longer tolerate!

JacquesCarette: Yes, I agree that that cognitive dissonance is a serious problem. That should be bumped to a higher level: we should be more coherent as to what we indicate in subscripts, and using it for both arity of the operation and arity of the current function seems like too much of an overload.

src/Relation/Binary/Morphism/Definitions.agda
```diff
@@ -28,5 +20,6 @@ open import Function.Core public
 ------------------------------------------------------------------------
 -- Basic definitions
 
-Homomorphic₂ : Rel A ℓ₁ → Rel B ℓ₂ → (A → B) → Set _
-Homomorphic₂ _∼₁_ _∼₂_ ⟦_⟧ = ∀ {x y} → x ∼₁ y → ⟦ x ⟧ ∼₂ ⟦ y ⟧
+open import Function.Definitions public
+  using ()
+  renaming (Congruent to Homomorphic₂)
```
MatthewDaggitt, line 25: If this is a stub module, should we be deprecating `Homomorphic₂`? Otherwise, where is it going to live?

src/Algebra/Morphism/Bundles.agda
```diff
@@ -16,8 +16,6 @@ open import Algebra.Bundles.Raw using
   ; RawRingWithoutOne; RawRing)
 open import Algebra.Morphism.Structures
 open import Level using (Level; suc; _⊔_)
---open import Relation.Binary.Morphism using (IsRelHomomorphism)
---open import Relation.Binary.Morphism.Bundles using (SetoidHomomorphism)
```
MatthewDaggitt, line 20: Delete?

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ Minor improvements
   an alias, and should be backwards compatible, but does improve the behaviour
   of the termination checker for some `Vector`-defined operations.
 
+* `Relation.Binary.Morphism.Definitions` is no longer imported by any module,
+  but retained as a stub for compatibility with external users.
```
MatthewDaggitt: Should it be deprecated then?

src/Relation/Binary/Morphism/Definitions.agda
jamesmckinna: My only reason for *not* deprecating, is that there seemed to be a request from @pnlph a number of years ago, to retain this name for client compatibility (and hence the `public` re-export from this module, where it originated). If there is no longer any such demand, then I'd be more radical wrt the surgery being carried out here, and mark the whole module as `DEPRECATED`...

src/Algebra/Morphism/Bundles.agda
jamesmckinna: Blast, I thought that I had caught these two.

jamesmckinna: Wait, won't this PR do the deletion when merged?

jamesmckinna: And yes, they are commented out (never mind unused) on `master`, so *should* be deleted!?

src/Relation/Binary/Morphism/Structures.agda
jamesmckinna: The conclusion I draw is that we stick with `Congruent`, and avoid renaming here.
That only leaves `Homomorphic₂` exported from `Relation.Binary.Morphism.Definitions`, but we can remove/deprecate that, too, if there is no longer demand for it.

CHANGELOG.md
jamesmckinna: Well, see my comment on the module. I'd be happy to, if we no longer have external clients requesting this.

But now I think about it, let's just deprecate anyway?

src/Relation/Binary/Morphism/Structures.agda
jamesmckinna: See now #2959 to lift this issue out

CHANGELOG.md
jamesmckinna: I tried a middle course, deprecating the two names, but not the whole module, but now, against expectations, a warning is raised whenever `Function.Definitions.Congruent` is used... ???

UPDATED: seemingly fixed now, but let's see what `make test` does remotely...

jamesmckinna:
Is this an Agda bug now? Or simply that I don't understand the semantics of `WARNING_ON_USE`?

src/Relation/Binary/Morphism/Definitions.agda
jamesmckinna: UPDATED: I've deprecated the lemmas, but not the module, and we'll do that in v3.0

JacquesCarette (approved):

Taneb:
See previous discussion: https://github.com/agda/agda-stdlib/issues/2318

MatthewDaggitt (changes requested):
Developer meeting: we decided not to deprecate `Homomorphism₂` and `Morphism` but instead just redefine them as `Congruent` etc. Still reduces duplication greatly.

src/Relation/Binary/Morphism/Definitions.agda
jamesmckinna: So I think we agreed to *NOT* deprecate now.

jamesmckinna:
Apologies to @Taneb for having ignored my own arguments on your earlier #2318 !

jamesmckinna:
@MatthewDaggitt over to you to re-review before merging.

src/Relation/Binary/Morphism/Definitions.agda
```diff
@@ -1,24 +1,17 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Basic definitions for morphisms between algebraic structures
+-- Issue #2318/#2875:
+-- this is now a stub module, retained solely for compatibility
```
MatthewDaggitt: I think we should undo this change. We don't have any concept of a stub module anywhere else.

src/Data/List/Relation/Unary/Unique/Setoid/Properties.agda
```diff
@@ -47,12 +47,12 @@ module _ (S : Setoid a ℓ₁) (R : Setoid b ℓ₂) where
   open Setoid S renaming (_≈_ to _≈₁_)
   open Setoid R renaming (_≈_ to _≈₂_)
 
-  map⁺ : ∀ {f} → (∀ {x y} → f x ≈₂ f y → x ≈₁ y) →
-         ∀ {xs} → Unique S xs → Unique R (map f xs)
+  map⁺ : ∀ {f} → Injective _≈₁_ _≈₂_ f →
+         Unique S ⊆ Unique R ∘ map f
```
MatthewDaggitt: I like the `Injective` change as it reduces duplication, but I don't think we should be bundling changes to from point to point free notation under the hood in this PR? I'm not sure I'd agree this change is an improvement...

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@ Minor improvements
   an alias, and should be backwards compatible, but does improve the behaviour
   of the termination checker for some `Vector`-defined operations.
 
+* `Relation.Binary.Morphism.Definitions` is no longer imported by any module,
+  but retained as a compatibility stub for external users, and for uniformity
+  with the design of existing hierarchies.
```
MatthewDaggitt: Remove?

src/Algebra/Morphism/Bundles.agda
MatthewDaggitt: Yes, they should be deleted if they are commented out on master.

MatthewDaggitt (changes requested):

src/Data/List/Relation/Unary/Unique/Setoid/Properties.agda
jamesmckinna: Yup. Done.

src/Relation/Binary/Morphism/Definitions.agda
jamesmckinna: OK, fine. Done.

CHANGELOG.md
jamesmckinna: Fine. It's in the spirit of all the other reversions you've suggested. Done

jamesmckinna:
Thanks @MatthewDaggitt ! PR should now follow your suggestions.

MatthewDaggitt (approved):

JacquesCarette (approved):

Issue #2626, opened 2025-02-27
Redefine `Algebra.Consequences.Setoid.wlog`


jamesmckinna:
The trouble with [the definition](https://github.com/agda/agda-stdlib/blob/2bce915db17fc284c7ed7a13d91cffe7a058a1e1/src/Algebra/Consequences/Setoid.agda#L311-L325):
```agda
------------------------------------------------------------------------
-- Without Loss of Generality

module _ {p} {P : Pred A p}
         (≈-subst : Substitutive _≈_ p)
         (∙-comm : Commutative _∙_)
         where

  subst∧comm⇒sym : Symmetric (λ a b → P (a ∙ b))
  subst∧comm⇒sym = ≈-subst P (∙-comm _ _)

  wlog : ∀ {r} {_R_ : Rel _ r} → Total _R_ →
         (∀ a b → a R b → P (a ∙ b)) →
         ∀ a b → P (a ∙ b)
  wlog r-total = BinaryConsequences.wlog r-total subst∧comm⇒sym
```
 is twofold:
* the assumption `≈-subst : Substitutive _≈_ p` is needlessly restrictive, when `≈-resp : P Respects _≈_` would suffice
* moreover, as even pointed out in `Relation.Binary.Definitions`, a relation `_≈_` only ever satisfies `Substitutive _≈_ p` when it *is* `PropositionalEquality` (or a derivative of it), so the `Setoid` version of `wlog` can only ever be deployed in the `Propositional` case!

Arguably this is `bug`, so should be fixed as non-backwards-compatible change in v2.3 ;-) but v3.0 is probably more reasonable, or at least, less unreasonable...

Accordingly, propose instead:
```agda
------------------------------------------------------------------------
-- Without Loss of Generality

module _ {p} {P : Pred A p}
         (≈-resp : P Respects _≈_)
         (∙-comm : Commutative _∙_)
         where

  resp∧comm⇒sym : Symmetric (λ a b → P (a ∙ b))
  resp∧comm⇒sym = ≈-resp (∙-comm _ _)

  wlog : ∀ {r} {_R_ : Rel _ r} → Total _R_ →
         (∀ a b → a R b → P (a ∙ b)) →
         ∀ a b → P (a ∙ b)
  wlog r-total = BinaryConsequences.wlog r-total resp∧comm⇒sym
```
with marginal knock-on consequences for the `Propositional` case (`resp P` instead of `subst`)

gallais:
[...]

PR #2924, opened 2026-01-26
[ add ] Nat lemmas with `_∸_`, `_⊔_` and `_⊓_`

jkopanski:
Some stuff that I've needed. I have no idea if those belong here in stdlib.  I don't know if I've put those in the right place, or named them properly.  Hence the draft and lack of changlog entry.

### DIFF
--- a/src/Data/Nat/Properties.agda
+++ b/src/Data/Nat/Properties.agda
@@ -1640,6 +1640,15 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
   suc ((m + n) ∸ o) ≡⟨ cong suc (+-∸-assoc m o≤n) ⟩
   suc (m + (n ∸ o)) ∎
 
+m∸n+o≡m∸[n∸o] : ∀ {m n o} → .(n ≤ m) → .(o ≤ n) → (m ∸ n) + o ≡ m ∸ (n ∸ o)
+m∸n+o≡m∸[n∸o] {m}     {zero}  {zero}  _     _ = +-identityʳ m
+m∸n+o≡m∸[n∸o] {suc m} {suc n} {zero}  _     _ = +-identityʳ (m ∸ n)
+m∸n+o≡m∸[n∸o] {suc m} {suc n} {suc o} sn≤sm so≤sn = begin-equality
+  suc m ∸ suc n + suc o ≡⟨ +-suc (m ∸ n) o ⟩
+  suc (m ∸ n + o)       ≡⟨ cong suc (m∸n+o≡m∸[n∸o] (s≤s⁻¹ sn≤sm) (s≤s⁻¹ so≤sn)) ⟩
+  suc (m ∸ (n ∸ o))     ≡⟨ ∸-suc (≤-trans (m∸n≤m n o) (s≤s⁻¹ sn≤sm)) ⟨
+  suc m ∸ (n ∸ o)       ∎
+
 m≤n+o⇒m∸n≤o : ∀ m n {o} → m ≤ n + o → m ∸ n ≤ o
 m≤n+o⇒m∸n≤o      m  zero    le = le
 m≤n+o⇒m∸n≤o zero    (suc n)  _ = z≤n
@@ -1728,10 +1737,29 @@ even≢odd (suc m) (suc n) eq = even≢odd m n (suc-injective (begin-equality
 ------------------------------------------------------------------------
 -- Properties of _∸_ and _⊓_ and _⊔_
 
+m∸n≤m⊔n : ∀ m n → m ∸ n ≤ m ⊔ n
+m∸n≤m⊔n m n = ≤-trans (m∸n≤m m n) (m≤m⊔n m n)
+
 m⊓n+n∸m≡n : ∀ m n → (m ⊓ n) + (n ∸ m) ≡ n
 m⊓n+n∸m≡n zero    n       = refl
 m⊓n+n∸m≡n (suc m) zero    = refl
 m⊓n+n∸m≡n (suc m) (suc n) = cong suc $ m⊓n+n∸m≡n m n
+
+m⊔n∸[m∸n]≡n : ∀ m n → m ⊔ n ∸ (m ∸ n) ≡ n
+m⊔n∸[m∸n]≡n zero    n       = cong (n ∸_) (0∸n≡0 n)
+m⊔n∸[m∸n]≡n (suc m) zero    = n∸n≡0 m
+m⊔n∸[m∸n]≡n (suc m) (suc n) = begin-equality
+  suc  (m ⊔ n) ∸ (m ∸ n)  ≡⟨ ∸-suc (m∸n≤m⊔n m n) ⟩
+  suc ((m ⊔ n) ∸ (m ∸ n)) ≡⟨ cong suc (m⊔n∸[m∸n]≡n m n) ⟩
+  suc n                   ∎
+
+m⊔n≡m∸n+n : ∀ m n → m ⊔ n ≡ m ∸ n + n
+m⊔n≡m∸n+n zero    n       = sym (cong (_+ n) (0∸n≡0 n))
+m⊔n≡m∸n+n (suc m) zero    = sym (cong suc (+-identityʳ m))
+m⊔n≡m∸n+n (suc m) (suc n) = begin-equality
+  suc (m ⊔ n)     ≡⟨ cong suc (m⊔n≡m∸n+n m n) ⟩
+  suc (m ∸ n + n) ≡⟨ +-suc (m ∸ n) n ⟨
+  m ∸ n + suc n   ∎
 
 [m∸n]⊓[n∸m]≡0 : ∀ m n → (m ∸ n) ⊓ (n ∸ m) ≡ 0
 [m∸n]⊓[n∸m]≡0 zero zero       = refl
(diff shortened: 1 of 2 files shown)


src/Data/Nat/Properties.agda
```diff
@@ -1722,11 +1730,24 @@ even≢odd (suc m) (suc n) eq = even≢odd m n (suc-injective (begin-equality
 ------------------------------------------------------------------------
 -- Properties of _∸_ and _⊓_ and _⊔_
 
+m∸n≤m⊔n : ∀ m n → m ∸ n ≤ m ⊔ n
```
Taneb, line 1740: This property doesn't feel natural to me, and the definition is short enough I don't really think it needs a name

jkopanski: It looks kind of similar to the [m⊔n≤m+n](https://agda.github.io/agda-stdlib/v2.3/Data.Nat.Properties.html#41624) that is already here.

jamesmckinna:
Monus `m∸n` is characterised #1949 by a pair of adjoint properties wrt the ordering `_≤_`, so it seems as though these properties also ought to be obtainable from those and the corresponding characterisations of `_⊓_` and `_⊔_` as glb (product) and lub (copoduct) in that category/preorder.

src/Data/Nat/Properties.agda
```diff
@@ -1844,6 +1872,17 @@ m∸n≤∣m-n∣ m n with ≤-total m n
 ∣m-n∣≤m⊔n (suc m) zero    = ≤-refl
 ∣m-n∣≤m⊔n (suc m) (suc n) = m≤n⇒m≤1+n (∣m-n∣≤m⊔n m n)
 
+∣m-n∣≡m⊔n∸m⊓n : ∀ m n → ∣ m - n ∣ ≡ m ⊔ n ∸ m ⊓ n
```
jamesmckinna, line 1881: The structure of this proof recalls that of `∣m-n∣≡[m∸n]∨[n∸m]` (L1863). 
Is there a refactoring which allows you to prove each result from some other, more general lemma?

Separately, you might like to consider refactoring *this* proof to *not* use `with`, but (more) simply via `[ branch1 , branch2 ]′ $ ≤-total m n` for suitable subproofs `branch1`, `branch2`, should you ever need to reason about this operation. cf. '`with` (sometimes) considered harmful' #2123

jkopanski: > The structure of this proof recalls that of `∣m-n∣≡[m∸n]∨[n∸m]` (L1863).
> Is there a refactoring which allows you to prove each result from some other, more general lemma?

Is this actual question or more like a _teacher_ question where you know there is and you probing me to look for it? I'm asking as I can't really find it.
I could rewrite it where instead of `≤-total` I would match on the result of `∣m-n∣≡m⊔n∸m⊓n` but that would only handle lhs, for rhs I would still need to recover the `m≤n` and `n≤m` from the result using `∣m-n∣≡m∸n⇒n≤m` and `∣m-n∣≡n∸m⇒m≤n`*

*) seems this one is missing, let me add it here

jamesmckinna: @jkopanski Thanks!

While it possible that I do indeed ask questions in a 'teacher-like' style (as friends and colleagues often point out to me... you can't easily take the teacher out of me ;-)), here I must say that my question was one of curiosity!

I think that there is a general principle for `stdlib` that we try not to 'over-duplicate' or generate (too much) additional redundancy between lemmas. Here I could see that there is a 'similar' result, so I was hoping there might be a way to exploit the similarity. But if not, then not!

And if there is, then we can always refactor downstream... ;-)

UPDATED: happy to leave this for later, but perhaps this could be refactored using `Relation.Binary.Consequences.wlog`...

src/Data/Nat/Properties.agda
JacquesCarette: I'm okay - lets one do proofs at a higher level with fewer steps.

JacquesCarette (approved):
I'm fine with adding these. We can have later passes of simplification of proofs.

jamesmckinna:
@jkopanski Shall we take this out of draft and try to get it in shape to merge? Or do you still want time to work on it further?

src/Data/Nat/Properties.agda
Taneb: Can you provide an example of an "interesting" proof where this lemma is a useful step?

jkopanski: I don't know if that counts as interesting, but the only place where I used it is down below in the `m⊔n∸[m∸n]≡n`.

Just to be clear I'm fine with removing that. In this PR I've just included stuff that I've tried to lookup in the stdlib first and written down once I couldn't find it.

jkopanski:
> @jkopanski Shall we take this out of draft and try to get it in shape to merge? Or do you still want time to work on it further?

yes, I would like to get it going. I've rebased it on recent changes and added changlog entry.

There's the case of `m∸n≤m⊔n` to be decided.

I've only now noticed your update regarding `wlog`. I don't think edits trigger notification here. Anyway that is something that I've never used so it may take me some time to figure out, if we were to push on it in this pr.

jamesmckinna:
> > @jkopanski Shall we take this out of draft and try to get it in shape to merge? Or do you still want time to work on it further?
> 
> ...
> I've only now noticed your update regarding `wlog`. I don't think edits trigger notification here. Anyway that is something that I've never used so it may take me some time to figure out, if we were to push on it in this pr.

No, as @JacquesCarette observes, we can handle that as a downstream refactoring. And `wlog` itself could do with some fresh thought... since #523 #2577 #2626

src/Data/Nat/Properties.agda
```diff
@@ -1640,6 +1640,15 @@ m≤n⇒n∸m≤n (s≤s m≤n) = m≤n⇒m≤1+n (m≤n⇒n∸m≤n m≤n)
...
   suc (m + (n ∸ o)) ∎
 
+m∸n+o≡m∸[n∸o] : ∀ {m n o} → n ≤ m → o ≤ n → (m ∸ n) + o ≡ m ∸ (n ∸ o)
+m∸n+o≡m∸[n∸o] {m}     {zero}  {zero}  z≤n       _         = +-identityʳ m
+m∸n+o≡m∸[n∸o] {suc m} {suc n} {zero}  (s≤s n≤m) z≤n       = +-identityʳ (m ∸ n)
+m∸n+o≡m∸[n∸o] {suc m} {suc n} {suc o} (s≤s n≤m) (s≤s o≤n) = begin-equality
+  suc m ∸ suc n + suc o ≡⟨ +-suc (m ∸ n) o ⟩
+  suc (m ∸ n + o)       ≡⟨ cong suc (m∸n+o≡m∸[n∸o] n≤m o≤n) ⟩
+  suc (m ∸ (n ∸ o))     ≡⟨ ∸-suc (≤-trans (m∸n≤m n o) n≤m) ⟨
+  suc m ∸ (n ∸ o)       ∎
```
jamesmckinna: So... noting that it took me a long time to internalise this idiom, `stdlib` design tends to favour:
* making at least `m` and `n` explicit (because not directly inferrable from `_∸ _`), as in the lemma immediately following this one, but probably also `o` as well, for the same reason
* making arguments of (proof-irrelevant) propositions `n ≤ m` and `o ≤ n` *definitionally* irrelevant where possible
So, here:

```suggestion
m∸n+o≡m∸[n∸o] : ∀ m n o → n ≤ m → .(o ≤ n) → (m ∸ n) + o ≡ m ∸ (n ∸ o)
m∸n+o≡m∸[n∸o] m        zero   zero    _     _     = +-identityʳ m
m∸n+o≡m∸[n∸o] (suc m) (suc n) zero    sn≤sm _     = +-identityʳ (m ∸ n)
m∸n+o≡m∸[n∸o] (suc m) (suc n) (suc o) sn≤sm so≤sn = begin-equality
  suc m ∸ suc n + suc o ≡⟨ +-suc (m ∸ n) o ⟩
  suc (m ∸ n + o)       ≡⟨ cong suc (m∸n+o≡m∸[n∸o] m n o (s≤s⁻¹ sn≤sm) (s≤s⁻¹ so≤sn)) ⟩
  suc (m ∸ (n ∸ o))     ≡⟨ ∸-suc (≤-trans (m∸n≤m n o) (s≤s⁻¹ sn≤sm)) ⟨
  suc m ∸ (n ∸ o)       ∎
```

And in fact, `n ≤ m` can also be made irrelevant, after refactoring lemma `∸-suc`...

jkopanski: > making at least m and n explicit (because not directly inferrable from _∸ _)

Oh I thought the `m n o` were inferred from the inequalities passed.

> as in the lemma immediately following this one

indeed that seems to be the case in the `m≤n+o⇒m∸n≤o`, but the lemmas directly above: `+-∸-assoc`, `∸-+-assoc` and `+-∸-comm` seems to follow the convention I mentioned: variables that are used in the inequalities are implicit, explicit ones are only the ones that aren't used in any arguments to the lemma.

Does making arguments proof irrelevant changes anything with that regard?

jamesmckinna: D'oh, yes, you're quite correct, each of `m`, `n`, and `o`, is inferrable from the hypotheses. Apologies!

jamesmckinna: But... see also #2790 : if you have an irrelevant (relation) argument, and the values on which it depends are implicit, sometimes Agda can't quite figure them out during inference when *used*, rather than *defined*. Puzzling...

jamesmckinna:
@MatthewDaggitt are you happy that your review requests have been answered satisfactorily?
if so, I think we could move to merging this in?

src/Data/Nat/Properties.agda
jamesmckinna: Having pondered this for a while, I'm leaning towards @Taneb 's position more than @JacquesCarette 's.

src/Data/Nat/Properties.agda
jamesmckinna: Something to think about for another day: to try to characterise what's going on here in terms of `∣ m - n ∣` as a symmetric difference operator, and any possible candidate `IsBoolean(Semi)Ring` structure on Nat...?

src/Data/Nat/Properties.agda
jkopanski: > And in fact, `n ≤ m? can also be made irrelevant, after refactoring lemma `∸-suc`

I've noticed you've merged the #2939. So I've updated this pr with irrelevant arguments

src/Data/Nat/Properties.agda
jamesmckinna: Namely, ... TL;DR: `Fairbairn threshhold`, I think. It's used only once.

src/Data/Nat/Properties.agda
jamesmckinna: Thanks! Sorry not to have updated the discussion here to mention that, but Sunday/yesterday saw a flurry of merge activity which I was having trouble staying on top of.

PR #2925, opened 2026-01-28
[ add ] inequalities for `2 ^ log₂ n`

t-wissmann:
This adds the inequalities `2 ^ ⌊log₂ n ⌋ ≤ n` for non-zero `n` and `n ≤ 2 ^ ⌈log₂ n ⌉` for all natural numbers `n`. Similar to the existing properties of the logarithm, the main proofs that keep track of the accumulators are added to `Data.Nat.Logarithm.Core`. Both use a lemma that `2 * (suc n) ≡ 2 + (n + n)` which is added to `Data.Nat.Logarithm.Core` as well.

The respective simplified versions on `⌊log₂_⌋` and `⌈log₂_⌉` are added directly to `Data.Nat.Logarithm`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,7 +238,7 @@ Additions to existing modules
 
 * In `Data.Nat.Logarithm`
   ```agda
-  2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
+  2^⌊log₂n⌋≤n : ∀ n .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
   n≤2^⌈log₂n⌉ : ∀ n → n ≤ 2 ^ ⌈log₂ n ⌉
   ```
 


src/Data/Nat/Logarithm/Core.agda
```diff
@@ -81,6 +81,22 @@ open import Relation.Binary.PropositionalEquality.Properties
   suc n                                 ∎
   where open ≡-Reasoning
 
+2*suc-n≡2+n+n : ∀ n → 2 * (ℕ.suc n) ≡ 2 + (n + n)
```
gallais: The convention is to write `2*suc[n]≡2+n+n` for this type of lemma.
It should probably be moved to `Data.Nat.Properties`.

t-wissmann: Thanks! I've renamed the lemma to `2*suc[n]≡2+n+n` and have moved it to `Data.Nat.Properties`. Is the location in there OK? I put it right under `*-suc`

src/Data/Nat/Logarithm/Core.agda
```diff
@@ -81,6 +81,19 @@ open import Relation.Binary.PropositionalEquality.Properties
   suc n                                 ∎
   where open ≡-Reasoning
 
+2^⌊log2n⌋≤n : ∀ n → (acc : Acc _<_ (1 + n)) → 2 ^ (⌊log2⌋ (1 + n) acc) ≤ 1 + n
+2^⌊log2n⌋≤n ℕ.zero _ = s≤s z≤n
+2^⌊log2n⌋≤n (ℕ.suc n) (acc rs) =
+  begin
```
jamesmckinna: This is very nice, but I think that it would be 'better' (for a suitable definition of 'better' ;-) certainly 'more in line with the `std-lib` library style'...) to use the `NonZero` predicate as an (irrelevant) `instance` as part of the type, rather than relying on the knowledge that `suc n`/`1 + n` is automatically `NonZero`...

Accordingly:

```suggestion
2^⌊log2n⌋≤n : ∀ n .{{_ : NonZero n}} → (acc : Acc _<_ n) → 2 ^ (⌊log2⌋ n acc) ≤ n
2^⌊log2n⌋≤n (ℕ.suc n) (acc rs) =
  begin
```

I haven't checked that the rest of the proof goes through unchanged, but... it shouldn't require much more work to fix things?

src/Data/Nat/Logarithm/Core.agda
```diff
@@ -124,3 +137,17 @@ open import Relation.Binary.PropositionalEquality.Properties
   1 + ⌈log2⌉ (2 ^ n) (<-wellFounded _)  ≡⟨ cong suc (⌈log2⌉2^n≡n n) ⟩
   suc n                                 ∎
   where open ≡-Reasoning
+
+n≤2^⌈log2n⌉ : ∀ n → (acc : Acc _<_ n) → n ≤ 2 ^ (⌈log2⌉ n acc)
+n≤2^⌈log2n⌉ ℕ.zero _ = z≤n
+n≤2^⌈log2n⌉ (ℕ.suc ℕ.zero) _ = s≤s z≤n
+n≤2^⌈log2n⌉ (ℕ.suc (ℕ.suc n)) (acc rs) =
```
jamesmckinna: We do have the `2+_` pattern also now?

src/Data/Nat/Logarithm/Core.agda
```diff
@@ -124,3 +137,17 @@ open import Relation.Binary.PropositionalEquality.Properties
...
+  begin
+  2 + n                                ≡⟨ cong (2 +_) (⌊n/2⌋+⌈n/2⌉≡n _) ⟨
+  2 + (⌊ n /2⌋ + ⌈ n /2⌉)              ≤⟨ +-monoʳ-≤ 2 (+-monoˡ-≤ _ (⌊n/2⌋≤⌈n/2⌉ n)) ⟩
+  2 + (⌈ n /2⌉ + ⌈ n /2⌉)              ≡⟨ 2*suc[n]≡2+n+n _ ⟨
+  2 * (1 + ⌊ 1 + n /2⌋)                ≤⟨ *-monoʳ-≤ 2 (n≤2^⌈log2n⌉ (1 + ⌊ 1 + n /2⌋) _) ⟩
+  2 * 2 ^ (⌈log2⌉ (1 + ⌊ 1 + n /2⌋) _) ≡⟨⟩
```
jamesmckinna: Given that this is in `Core` (ie notionally not anything that will be exported to 'client-facing' code) you could, again 'better', replace all the `1 +_` occurrences with `ℕ.suc`?

Are there particular reasons for the choices you make?

src/Data/Nat/Logarithm.agda
```diff
@@ -41,6 +41,9 @@ open import Relation.Binary.PropositionalEquality.Core using (_≡_)
 ⌊log₂[2^n]⌋≡n : ∀ n → ⌊log₂ (2 ^ n) ⌋ ≡ n
 ⌊log₂[2^n]⌋≡n n = ⌊log2⌋2^n≡n n
 
+2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
+2^⌊log₂n⌋≤n (ℕ.suc n) = 2^⌊log2n⌋≤n n (<-wellFounded (ℕ.suc n))
```
jamesmckinna: If you implement the above changes, then this could/should now simply be:

```suggestion
2^⌊log₂n⌋≤n n = 2^⌊log2n⌋≤n n (<-wellFounded n)
```

src/Data/Nat/Properties.agda
```diff
@@ -774,6 +774,9 @@ m+n≮m m n = subst (_≮ m) (+-comm n m) (m+n≮n n m)
   suc (m + (n + m * n)) ≡⟨⟩
   suc m + suc m * n     ∎
 
+2*suc[n]≡2+n+n : ∀ n → 2 * (ℕ.suc n) ≡ 2 + (n + n)
+2*suc[n]≡2+n+n n = trans (cong ((1 + n) +_) (+-identityʳ _)) (+-suc (1 + n) n)
```
jamesmckinna: Two comments:
* `1 + n` could/should be `suc n`, as above?
* it's clearly 'only' a helper lemma, so its proof doesn't need to be *readable*, but consider whether a proof with equational reaosning combinators would be easier to *understand*...?

src/Data/Nat/Logarithm/Core.agda
t-wissmann: My proof needs the case
```
2^⌊log2n⌋≤n ℕ.zero _ = s≤s z≤n
```
at some point because the proof is by induction on `n` and I can't image how the induction should work without the base case. I tried it but I don't see how to update the proof such that it still compiles for your suggested type
```
2^⌊log2n⌋≤n : ∀ n .{{_ : NonZero n}} → (acc : Acc _<_ n) → 2 ^ (⌊log2⌋ n acc) ≤ n
```

src/Data/Nat/Logarithm/Core.agda
t-wissmann: The rest of the file uses `(suc (suc n))` instead of `2+_`. But I have replaced `ℕ.suc` with `suc` such that it fits better with the existing patterns in the file.

(If you want me to switch to `2+`, I would need to also update the `using`-list in the import accordingly.)

src/Data/Nat/Logarithm/Core.agda
t-wissmann: I just tried to find the notation that is most easy to read. But sure, I can also replace `1 +` with `suc` (I've just pushed the respective commit).

src/Data/Nat/Properties.agda
t-wissmann: Thanks! Here, `suc` is really more readable than `1 +_`. Also, I've upgraded it to equational reasoning :-)

Shall I also extract the first step as a separate lemma?
```
2*n≡n+n : ∀ n → 2 * n ≡ n + n
```
I couldn't find `2*n≡n+n` in the stdlib but would have needed it a couple of times already.

src/Data/Nat/Logarithm/Core.agda
jamesmckinna: No need!

src/Data/Nat/Properties.agda
jamesmckinna: Well certainly that seems to be a more reusable/generic/useful property?

jamesmckinna:
I notice as well that your changes are on your `master` branch, rather than first cloning a fork of `agda/master` and then opening a new branch for the PR (as in the `HACKING` guide...). 

This is (more or less) fine for a first PR, but generally not recommended as a future-proof strategy!

src/Data/Nat/Logarithm/Core.agda
jamesmckinna: Yes, you're quite right. See below!

t-wissmann:
Thanks for the pointer to `HACKING`! I will choose a different name next time.

src/Data/Nat/Properties.agda
t-wissmann: Does this mean that I shall add it? (Regardless of the decision on that, I still use `2*suc[n]≡2+n+n` in both log-inequalities)

jamesmckinna (changes requested):
Nearly there!

src/Data/Nat/Logarithm/Core.agda
```diff
@@ -15,7 +15,7 @@ open import Data.Nat.Properties
 open import Data.Nat.Induction using (<-wellFounded)
 open import Induction.WellFounded using (Acc; acc)
 open import Relation.Binary.PropositionalEquality.Core
-  using (_≡_; refl; cong; sym)
+  using (_≡_; refl; cong; sym; trans)
```
jamesmckinna: Is this needed now?

t-wissmann: Right, it's not needed :-) . I've just removed it.

src/Data/Nat/Logarithm/Core.agda
jamesmckinna:
```agda
2^⌊log2n⌋≤n : ∀ n .{{_ : NonZero n}} → (acc : Acc _<_ n) → 2 ^ (⌊log2⌋ n acc) ≤ n
2^⌊log2n⌋≤n (suc zero)     _       = ≤-refl
2^⌊log2n⌋≤n (suc (suc n)) (acc rs) = begin
  2 ^ (⌊log2⌋ (suc (suc n)) (acc rs)) ≡⟨⟩
  2 * 2 ^ (⌊log2⌋ (suc ⌊ n /2⌋) _)    ≤⟨ *-monoʳ-≤ 2 (2^⌊log2n⌋≤n _ _) ⟩
  2 * (suc ⌊ n /2⌋)                   ≡⟨ 2*suc[n]≡2+n+n _ ⟩
  2 + (⌊ n /2⌋ + ⌊ n /2⌋)             ≤⟨ +-monoʳ-≤ (2 + ⌊ n /2⌋) (⌊n/2⌋≤⌈n/2⌉ _)  ⟩
  2 + (⌊ n /2⌋ + ⌈ n /2⌉)             ≡⟨ cong (2 +_) (⌊n/2⌋+⌈n/2⌉≡n _) ⟩
  2 + n
  ∎
  where open ≤-Reasoning
```

NB. `style-guide` mandates that `begin` should be on the *same* line as the LHS of the definition.

CHANGELOG.md
```diff
@@ -236,6 +236,18 @@ Additions to existing modules
...
 
+* In `Data.Nat.Logarithm`
+  ```agda
+  2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
+  n≤2^⌈log₂n⌉ : ∀ n → n ≤ 2 ^ ⌈log₂ n ⌉
+  ```
+
+* In `Data.Nat.Logarithm.Core`
+  ```agda
+  2^⌊log2n⌋≤n : ∀ n → (acc : Acc _<_ (1 + n)) → 2 ^ (⌊log2⌋ (1 + n) acc) ≤ 1 + n
```
jamesmckinna: Again...
```suggestion
  2^⌊log2n⌋≤n : ∀ n .{{_ : NonZero n}} → (acc : Acc _<_ n) → 2 ^ (⌊log2⌋ n acc) ≤ n
```

jamesmckinna (changes requested):
I fixed your proof.
Oh, for some reason, the suggestion didn't get recorded?
```agda
2^⌊log2n⌋≤n : ∀ n .{{_ : NonZero n}} → (acc : Acc _<_ n) → 2 ^ (⌊log2⌋ n acc) ≤ n
2^⌊log2n⌋≤n (suc zero)     _       = ≤-refl
2^⌊log2n⌋≤n (suc (suc n)) (acc rs) = begin
  2 ^ (⌊log2⌋ (suc (suc n)) (acc rs)) ≡⟨⟩
  2 * 2 ^ (⌊log2⌋ (suc ⌊ n /2⌋) _)    ≤⟨ *-monoʳ-≤ 2 (2^⌊log2n⌋≤n _ _) ⟩
  2 * (suc ⌊ n /2⌋)                   ≡⟨ {!2*suc[n]≡2+n+n _!} ⟩
  2 + (⌊ n /2⌋ + ⌊ n /2⌋)             ≤⟨ +-monoʳ-≤ (2 + ⌊ n /2⌋) (⌊n/2⌋≤⌈n/2⌉ _)  ⟩
  2 + (⌊ n /2⌋ + ⌈ n /2⌉)             ≡⟨ cong (2 +_) (⌊n/2⌋+⌈n/2⌉≡n _) ⟩
  2 + n
  ∎
  where open ≤-Reasoning
```

NB `style-guide` mandates that `begin` be on the *same* line as the LHS of the definition...

src/Data/Nat/Properties.agda
jamesmckinna: Hmmm. It seems as though only one of then should be added to the `Data.Nat.Properties` API, but then get used/specialised as needed? There's a frequently-occurring issue of the `Fairbairn threshold` which you might google...

src/Data/Nat/Logarithm/Core.agda
t-wissmann: Thanks! I've updated it (also in the Changelog). However, I find the definition in `Logarithm.agda` then more difficult to understand, because it is not visible that the property of `n` being `NonZero` is passed to the call:
```
2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
2^⌊log₂n⌋≤n n = 2^⌊log2n⌋≤n n (<-wellFounded n)
```
But it's fine for me.  :-)

src/Data/Nat/Properties.agda
t-wissmann: Either is fine for me. If I remove `2*suc[n]≡2+n+n`, then each of the equational chains of `2^⌊log2n⌋≤n` and `n≤2^⌈log2n⌉` get longer by one step.

src/Data/Nat/Logarithm/Core.agda
jamesmckinna: Ah!

There are two ways to think about this problem (there may be others... I offer only two here!?), of which the first is what I take to be your 'difficulty' above:
* `instance` arguments are mysterious, because we neither see them explicitly bound on the LHS, nor see them explicitly passed as arguments on the RHS;
* `instance` arguments are fantastically useful, especially as here, when used as *irrelevant* arguments, because we neither see them explicitly bound on the LHS, nor see them explicitly passed as arguments on the RHS.

Spelt out, the definition above amounts to (and I believe this is what the type checker is more or less doing):
```agda
2^⌊log₂n⌋≤n n {{nz}} = 2^⌊log2n⌋≤n n (<-wellFounded n) {{something}}
  where
  something = <a proof of NonZero n>
```
in such a way that:
* `something` can be *uniquely* computed from the 'visible' arguments, and any other declared `instance`s, in scope (this is the requirement for instance inference to succeed)
* argument `nz` cannot actually be used in a *relevant* way, but that's OK, because here, the function `2^⌊log2n⌋≤n` also expects its instance argument to be used in an irrelevant way
* so... taking `something = nz` seems the only available choice, and moreover it satisfies the conditions above.

Ie.
```agda
2^⌊log₂n⌋≤n n {{nz}} = 2^⌊log2n⌋≤n n (<-wellFounded n) {{nz}}
```
or even, to emphasise that we are *not allowed* even to consider using `nz` in a relevant way
```agda
2^⌊log₂n⌋≤n n {{_}} = 2^⌊log2n⌋≤n n (<-wellFounded n) {{_}}
```
where the use of `_` on the LHS means "I don't care", while on the RHS means "you figure it out", in the dialogue game between user ("I") and typechecker ("you").

Once we arrive at the version with underscores, well, then they really are just noise, so...

CHANGELOG.md
```diff
@@ -236,6 +236,18 @@ Additions to existing modules
   map : P ⊆ Q → All P xs → All Q xs
   ```
 
+* In `Data.Nat.Logarithm`
+  ```agda
+  2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
```
jamesmckinna: No need to separate the arguments with an additional arrow
```suggestion
  2^⌊log₂n⌋≤n : ∀ n .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
```

src/Data/Nat/Logarithm.agda
```diff
@@ -41,6 +41,9 @@ open import Relation.Binary.PropositionalEquality.Core using (_≡_)
 ⌊log₂[2^n]⌋≡n : ∀ n → ⌊log₂ (2 ^ n) ⌋ ≡ n
 ⌊log₂[2^n]⌋≡n n = ⌊log2⌋2^n≡n n
 
+2^⌊log₂n⌋≤n : ∀ n → .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
```
jamesmckinna: As with the `CHANGELOG`
```suggestion
2^⌊log₂n⌋≤n : ∀ n .{{ _ : NonZero n }} → 2 ^ ⌊log₂ n ⌋ ≤ n
```
The additional arrow is noise. If you want leave it in, then the parser would also happily accept

```suggestion
2^⌊log₂n⌋≤n : ∀ n → .{{NonZero n}} → 2 ^ ⌊log₂ n ⌋ ≤ n
```
which is differently less 'noisy'...

CHANGELOG.md
```diff
@@ -248,6 +260,7 @@ Additions to existing modules
   ≟-≢   : (m≢n : m ≢ n) → (m ≟ n) ≡ no m≢n
   ∸-suc : m ≤ n → suc n ∸ m ≡ suc (n ∸ m)
   ^-distribʳ-* : ∀ m n o → (n * o) ^ m ≡ n ^ m * o ^ m
+  2*suc[n]≡2+n+n : ∀ n → 2 * (suc n) ≡ 2 + (n + n)
```
jamesmckinna, line 263: Happy to stick with this formulation!

src/Data/Nat/Properties.agda
jamesmckinna: Happy with things as they are.

t-wissmann:
Thank you for accepting! I also commited your suggestions regarding the removal of unnecessary →.
I'm not so familiar with the instances arguments, so I really appreciate your explanations on that! :-)

Issue #2905, opened 2026-01-04
Inline `Data.Vec.Functional.map` for Better Termination Checking


david-davies:
I ran across an issue with termination checking a recursive function on a data-type using functional vectors, and have simplified to this mwe:
```agda
data Stuff (A : Set a) : Set a where
  stuff : (i : ℕ) → Vector (Stuff A) i → Stuff A

mapStuff : (A → B) → Stuff A → Stuff B
mapStuff f (stuff i xs) = stuff i (Vector.map (mapStuff f) xs)
```
The termination checker gets upset with the call to `mapStuff f`.
This simplifies to `λ x → mapStuff f (xs x)`, which means the recursive call is smaller, so ideally it should pass the termination check.
Forcing this simplification by adding an inline pragma (without any change to the definition) resolves this:
```agda
map2 : (A → B) → ∀ {n} → Vector A n → Vector B n
map2 f xs = f ∘ xs
{-# INLINE map2 #-}

map2Stuff : (A → B) → Foo A → Foo B
map2Stuff f (stuff i xs) = stuff i (map2 (map2Stuff f) xs)
```
This does rely on `_∘_` being inlined also.

JacquesCarette:
I'm seriously torn about whether to do this. This is obviously a hack to get around a current limitation of Agda, which increases technical debt in `stdlib`. It also makes goals 'less semantic', which makes writing proofs that are not fragile harder.

I'm not actually saying "don't do this". I think I'm saying we should leave very clear markers in the source code when we choose to do this, so that they are easier to revisit when Agda improves. This particularly one is certainly quite tempting, as I've gotten bit by it too.

david-davies:
Yeah, it definitely feels like a work-around rather than a solution. I don't have strong feelings on this, I just thought it best to have it written down somewhere should others run into the same issue.

jamesmckinna:
Thanks @david-davies for the provocation to think... hopefully hard enough to arrive at a sensible resolution of this (kind of) issue.

[...]

PR #2927, opened 2026-01-30
[ fix ] issue #2905

jamesmckinna:
Only outstanding questions in my mind:
* (possibly PR-creep for this one) should we also inline the other aliases for `Function.Base` operations, eg `_⊛_` in `Data.Vec.Functional`?
* (possible downstream PR) should we, as policy, allow/mandate 'hereditary inlining' for such definitions in general? (eg. the `INLINE` specialised operations in `Tactic.RingSolver.Reasoning` could arguably be applied to their definitions in `Algebra.Structures`, etc. and hence make the former obsolete?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,10 +32,10 @@ Minor improvements
   properties (and their proofs). In particular, `truncate-irrelevant` is now
   deprecated, because definitionally trivial.
 
-* The function `Data.Vec.Functional.map` have marked with the `INLINE` pragma.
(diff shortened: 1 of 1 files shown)


jamesmckinna:
Thanks @JacquesCarette : any thoughts on subsequent inlinings in this module, or elsewhere?

JacquesCarette:
I think we should be sparing with our inlinings, and do them when someone actively notices that it is a hindrance.

PR #2928, opened 2026-02-04
[ deprecate ] `README.Inspect`

jamesmckinna:
See recent discussion on #1580 . No `CHANGELOG`.


JacquesCarette (approved):

doc/README/Inspect.agda
```diff
@@ -1,14 +1,17 @@
 ------------------------------------------------------------------------
 -- The Agda standard library
 --
--- Explaining how to use the inspect idiom and elaborating on the way
--- it is implemented in the standard library.
+-- This module is DEPRECATED.
```
MatthewDaggitt, line 4: Could we add a pointer towards the Agda documentation on the `with` idiom instead?

MatthewDaggitt (changes requested):

doc/README/Inspect.agda
jamesmckinna: Yes. I'd been going to add that, as previously, to the deprecation section of `Relation.Binary.PropositionalEquality.Core`, but happy to do so here too.

But NB. the upstream change will only arise in v2.9.0, but AFAIK, there aren't readthedocs pages for that release... yet... so there's a more annoying synchronisation problem, by contrast with #2931 / #2932

UPDATED: latest commit makes a forward-in-time pointer to the v2.9.0, assuming we sort out a sensible v2.4 release schedule for which this makes sense...

jamesmckinna: Going back over #1930, I realised that I'd failed to push the claimed `README.WithIn` to explain the new syntax, on the model of the previous `README.Inspect`.

One thing to do might be to reinstate it as part of this PR, but I'd need to find it, and work out if it still makes sense against the new semantics!

Let me know whether you think this worthwhile before I ~~push~~ look for, never mind push, anything else!

JacquesCarette:
@MatthewDaggitt I believe that @JacquesCarette has indeed done what you asked - your re-review is what is holding this up from getting merged.

jamesmckinna:
@JacquesCarette did you mean to shoutout to *me* @jamesmckinna ?
In any case, this is a weird one, because if they *haven't* released v2.9.0, then this is technically blocking on their upstream release?

It seems that `stable` (rather than `latest`) is one way to achieve a version-neutral link to readthedocs? Will update accordingly.

jamesmckinna:
Actually, I've gone back on myself: I think that until v2.9.0 is released (and with it, the crucial PR https://github.com/agda/agda/pull/8318 which finally laid `inspect` to rest), we *shouldn't* merge this PR, or its companion #2934 .

Suggest removing this from the milestone?

jamesmckinna:
Ugh, I'm making a mess of trying to bring this up-to-date. Converting to `DRAFT`.

Closing now in favour of a single PR bringing this and #2934  under one roof (after screwing up an attempt to bring each up-to-date... :-()

JacquesCarette:
Oops, yes, wrong tag.

PR #2932, opened 2026-02-05
[ fix ] make `Data.Nat.Primality.prime⇒nonZero` more strict in its explicit. argument

### LINKED ISSUE #2931: Don't be so eager to _, especially in the presence of instances
See agda/agda#8367 for the details. That points to the [exact code to change](https://github.com/agda/agda-stdlib/blob/18055fbe7985f661a596c46d5e9f2f4b3a16e9bf/src/Data/Nat/Primality.agda#L214).

jamesmckinna:
Fixes #2931 . No `CHANGELOG` unless anyone thinks we need one... the [upstream issue](https://github.com/agda/agda/pull/8367) is the cause.

### DIFF
--- a/src/Data/Nat/Primality.agda
+++ b/src/Data/Nat/Primality.agda
@@ -211,7 +211,7 @@ prime[2] : Prime 2
 prime[2] = prime 2-rough
 
 prime⇒nonZero : Prime p → NonZero p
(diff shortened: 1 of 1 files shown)


JacquesCarette (approved):

JacquesCarette:
This is a straight bug-from-the-future bug fix, so I don't think it needs a CHANGELOG either.

Issue #2909, opened 2026-01-17
Suggestion with `Preserves` for two predicates


mechvel:
lib-2.3  has `Preserves` for the pair of a binary relation and a map.             
But such is also needed for one or two predicates.                                        
For example, I define                                                           
```
Preserves11 :     ∀ {β p q}{B : Set β} → (A → B) → Pred A p → Pred B q → Set _                  
Preserves11 f P Q =  ∀ {x} → P x → Q (f x)                                      
                                                                                
_Preserves_ :  ∀ {p} → (A → A) → Pred A p → Set (α ⊔ p)                         
f Preserves P =  ∀ {x} → P x → P (f x)                                          
```                                                                             
                                                                                
For example,  ``Preserves11 Integer.+_ EvenNatural EvenInteger``                 
                                                                                
means (if `{n : ℕ}` is `even-inℕ`, then ``(+ n) : ℤ``  is `even-inℤ`).                    
                                                                                
And this cannot be named `Respects*` or `Congruent*` due to the library style.                                 
May Standard library need this?

jamesmckinna:
I could imagine this being a useful addition to `Relation.Unary`, so if you raise a PR, we would look at it.

That said, please also look at the existing definitions #2840 and properties #2866 of unary adjoints... where your `Preserves11` property is/would be expressible in terms of existing `Relation.Unary` constructions as `P ⊆ f ⊢ Q`, so I'd be just as likely to say that the library doesn't need (Fairbairn threshold) to add this as a named definition.

Similarly, one could argue that a *more* general definition would be better, namely in terms of `Relation.Binary._⟶_Respects_`, as `Preserves11 f P Q = P ⟶ Q Respects λ x y → y ≡ f x`, using the *graph* binary relation associated with `f`, which at least generalises to the case of non-`Propositional` equality `λ x y → y ≈ f x`. 

I definitely agree that the library is 'biased' towards definitions and properties of `Binary` relations more than `Unary` predicates, so an initiative to refactor both to make them more uniform/comparable might well yield an improved overall design...

mechvel:
On 2026-01-22 13:22, jamesmckinna wrote:
> jamesmckinna left a comment (agda/agda-stdlib#2909) [1]
> 
> I could imagine this being a useful addition to Relation.Unary, so if
> you raise a PR, we would look at it.
> [..]

So far, I am going to use Preserves11 of my own.
But if you find that this my suggestion is reasonable and thus make the 
[...]

PR #2934, opened 2026-02-06
[ deprecate ] `Relation/Binary.PropositionalEquality.inspect` (#1580 / #1630 / #1930 redux)

### LINKED ISSUE #1580: Deprecate `inspect` function in favour of new `with ... in` construct

jamesmckinna:
Fixes #1580 see the discussion there and on the previous attempts #1630 #1930. 
`CHANGELOG` based on text from #1930

NB 
* This does not touch the `Relation/Binary.HeterogeneousEquality.*` modules, which should be tackled in a separate PR, once we can agree how best to deal with them.
* As with #2928 (on which this PR is `blocking` ;-)) forward pointers to the `stable` version of the documentation ~~of the Agda v2.9.0 release.~~

UPDATED: added `CHANGELOG`, but cancelled most recent `make test` to save cycles.


jamesmckinna:
I further experimented with the following 'alternative' (well: hardly!)  design, which is less dramatic, but does achieve (by the back door) deprecation of the bugbear `[_]` constructor:
```agda
module Graph {A : Set a} {B : A → Set b} (f : (x : A) → B x) (x : A) where

  record View (y : B x) : Set (a ⊔ b) where
    field fx≡y : f x ≡ y

  view : View (f x)
  view .View.fx≡y = refl

  pattern [_] eq = record { fx≡y = eq } -- can be omitted?

-- Version 2.4

module _ {A : Set a} {B : A → Set b} where

  Reveal_·_is_ : (f : (x : A) → B x) (x : A) (y : B x) → Set (a ⊔ b)
  Reveal f · x is y = Graph.View f x y

  inspect : (f : (x : A) → B x) (x : A) → Reveal f · x is f x
  inspect = Graph.view
  {-# WARNING_ON_USAGE inspect
  "Warning: inspect was deprecated in v2.4.
  Please use Graph.view instead. "
  #-}
  {-# WARNING_ON_USAGE Reveal_·_is_
  "Warning: Reveal_·_is_ was deprecated in v2.4.
  Please use Graph.View instead. "
  #-}
```

That way we still retain a *useful* notion (the graph relation associated with a function, cf. my comment on #2909; this is Conor's `Im f` family from their later papers about DTFP/views/Epigram, in disguise), while omitting
* the problematic constructor (well, not at present with the `pattern` synonym?)
* the mumbo-jumbo associated with the names `Reveal` and `inspect`, which I have always considered part of a different, cognitive/discoverability, problem.

Perhaps the thing to do is to lift this out to a separate module, or else in `Relation.Binary.Definitions`?

JacquesCarette (approved):

JacquesCarette:
I rather like the `Graph` relation. Yes, I think that should be available in `stdlib` somewhere (without the `[_]`).

jamesmckinna:
> I rather like the `Graph` relation. Yes, I think that should be available in `stdlib` somewhere (without the `[_]`).

Thanks for the endorsement. I realised that to try and work out a 'proper' design wrt arbitrary (indexed!) relations, and naming, was a bit beyond me (I have yet another wicked chest/throat infection atm after what seems like 7 continuous weeks of freezing rain here), so I'll make some rearrangements to the above code, and then leave it be... but I'd be happy either way if the most recent commit b23264ea0af8aac9ba574f300e478421493f1575 were reverted.

I think it's time to stop picking at this particular wart!

jamesmckinna:
As with #2928 if we aim to release v2.4 after v2.9.0, I'll try to get this ready for the v2.4 milestone... but for now, removing it.

Closing now in favour of a single PR bringing this and #2928 under one roof (after screwing up an attempt to bring each up-to-date... :-()

